PR #3277, opened 2025-05-13
fix: decimal byte parts scalar at incorrectly casted when it should transmute

joseph-isaacs:


robert3005 (approved):
Hmm, I thought that's how the cast would work

PR #3282, opened 2025-05-13
Remove optional metadata since empty vec is zero cost

gatesn:
And rename "decode" to clearer "build"

### DIFF
--- a/vortex-layout/src/layouts/dict/reader.rs
+++ b/vortex-layout/src/layouts/dict/reader.rs
@@ -34,7 +34,11 @@ impl DictReader {
             vortex_panic!("Mismatched layout ID")
         }
         let metadata = ProstMetadata::<DictLayoutMetadata>::deserialize(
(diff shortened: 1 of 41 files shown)


vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -34,7 +34,11 @@ impl DictReader {
             vortex_panic!("Mismatched layout ID")
         }
         let metadata = ProstMetadata::<DictLayoutMetadata>::deserialize(
-            layout.metadata().as_ref().map(|b| b.as_ref()),
+            layout
+                .metadata()
+                .as_ref()
+                .map(|b| b.as_ref())
+                .unwrap_or(&[]),
```
robert3005: maybe explicit error here or flip the flow?

gatesn: Flatbuffer serializes empty metadata as optional empty I believe.

robert3005 (approved):

PR #3283, opened 2025-05-13
feat: is constant with cost enum

joseph-isaacs:
Can more precisely specify how much work to carry out when running is_constant?

Should we move the constant compute function to an array trait method?


encodings/fastlanes/src/bitpacking/compute/is_constant.rs
```diff
@@ -16,8 +16,11 @@ impl IsConstantKernel for BitPackedVTable {
     fn is_constant(
         &self,
         array: &BitPackedArray,
-        _opts: &IsConstantOpts,
+        opts: &IsConstantOpts,
     ) -> VortexResult<Option<bool>> {
+        if opts.is_constant() {
```
gatesn: Hmmmm, I changed my mind. I think we should call this "Trivial", it's way too confusing otherwise!

vortex-array/src/arrays/bool/compute/is_constant.rs
```diff
@@ -5,7 +5,11 @@ use crate::compute::{IsConstantKernel, IsConstantKernelAdapter, IsConstantOpts};
 use crate::register_kernel;
 
 impl IsConstantKernel for BoolVTable {
-    fn is_constant(&self, array: &BoolArray, _opts: &IsConstantOpts) -> VortexResult<Option<bool>> {
+    fn is_constant(&self, array: &BoolArray, opts: &IsConstantOpts) -> VortexResult<Option<bool>> {
+        if opts.is_constant() {
```
gatesn: Could actually allow len <= 64 to be constant time, since we're just comparing the boolean buffer as a single u64 to the zero value.

encodings/fastlanes/src/bitpacking/compute/is_constant.rs
joseph-isaacs: How about `constant_time`?
I think trivial is an ambiguous and confusing.

vortex-array/src/arrays/bool/compute/is_constant.rs
joseph-isaacs: Sure

encodings/fastlanes/src/bitpacking/compute/is_constant.rs
gatesn: Yeah that's an improvement. I would still argue that "constant_<anything>" is more confusing that not using the word "constant". Immediate? Negligible?

vortex-array/src/arrays/varbin/compute/is_constant.rs
```diff
@@ -9,8 +9,11 @@ impl IsConstantKernel for VarBinVTable {
     fn is_constant(
         &self,
         array: &VarBinArray,
-        _opts: &IsConstantOpts,
+        opts: &IsConstantOpts,
     ) -> VortexResult<Option<bool>> {
+        if opts.is_constant() {
```
gatesn: I wonder if we should instead be writing something like `if opts.cost <= Cost::Constant`? Will we ever add more? Feel free to say no

vortex-array/src/compute/is_constant.rs
```diff
@@ -35,14 +35,26 @@ pub fn is_constant(array: &dyn Array) -> VortexResult<Option<bool>> {
...
             inputs: &[array.into()],
             options,
         })?
         .unwrap_scalar()?
         .as_bool()
-        .value())
+        .value();
+
+    // TODO(joe): add is_constant for ListArray
+    if options.cost == Cost::Canonicalize && !array.is::<ListVTable>() {
```
gatesn: Move this logic inside the `impl ComputeFn for Constant`, in general we shouldn't assume callers will go through this function. They could go via FFI for example.

vortex-array/src/arrays/varbin/compute/is_constant.rs
gatesn: Actually, this would solve the `is_constant` naming problem too

joseph-isaacs:
I like `Negligible`

joseph-isaacs:
`opts.is_cost_negligible()`

a10y:
Perhaps we should also add an option to give your compute functions a tip for all the hard work they do

```
pub enum TipPercent {
  Fifteen,
  TwentyFive,
  ThirtyFive,
  Custom(NonZeroUsize),
}
```

a10y:
I guess I'm not sure why we need an enum with 3 options. I sort of just group the use-cases into two buckets in my head

1. I can do some optimization if I know something is constant, so check the stat/encoding in constant time
2. I need to know if something is constant for correctness reasons, so I trigger calculating the stat

Why expand this?

joseph-isaacs:
1. I want the quickest is_constant method (this is currently missing).
2. I am happy to have an unpredictable runtime but its not required I get back a value (for scanning)
3. I want an answer (for writing)

We are currently missing (1). There are multiple places where we want this

vortex-array/src/compute/is_constant.rs
```diff
@@ -231,16 +243,32 @@ impl<'a> TryFrom<&InvocationArgs<'a>> for IsConstantArgs<'a> {
...
 }
 
+/// When calling `is_constant` the children are all checked for constantness.
+/// This enum decide at each precision/cost level the constant check should run as.
+/// The cost increase as we move down the list.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Cost {
+    /// Only apply constant time computation to estimate constantness.
+    Negligible,
+    /// Allow the encoding to do a linear amount of work to decide is constant.
```
gatesn: I think the crucial thing to document here is that it can short-circuit before doing a linear amount of work. Whereas canonicalizing is >O(N)

joseph-isaacs: updated the doc str

vortex-array/src/array/mod.rs
```diff
@@ -363,7 +360,7 @@ impl<V: VTable> Array for ArrayAdapter<V> {
         // computing derived stats and merging them in.
         // TODO(ngates): skip the is_constant check here, it can force an expensive compute.
         // TODO(ngates): provide a means to slice an array _without_ propagating stats.
-        let derived_stats = (self.encoding_id() != ConstantEncoding.id()).then(|| {
+        let derived_stats = self.0.is_constant_opts(Cost::Negligible).then(|| {
```
gatesn: I think you've inverted this condition?

joseph-isaacs:
```suggestion
        let derived_stats = (!self.0.is_constant_opts(Cost::Negligible)).then(|| {
```

joseph-isaacs: yep, fixed

gatesn (approved):
Follow up to move is_constant into ArrayOperationsVTable

PR #3295, opened 2025-05-14
break: minimum Python 3.11

gatesn:


robert3005 (commented):
Are we sure the build is actually compatible with 3.11+ or just 3.11? I'm not certain PyO3 uses minimal abi by default

gatesn:
Yes abi3 is the stable API, with minimum of 3.11.

Without abi3 you target a specific Python minor version

robert3005 (approved):

PR #3213, opened 2025-05-05
wip: GeoVortex

a10y:
This branch serves as an implementation sketch for a new series of extension types around geospatial vector data. This is largely an exercise to help flush out what's lacking in the currently extension type infra in a more "real" way than our temporal types. Still very raw

* Provides Vortex encodings that canonicalize into [GeoArrow](https://geoarrow.org/format.html) and ability to zero-copy load geoarrow data into Vortex
* Add pluggability of Arrow <> Vortex type conversion via a similar mechanism to the KernelRef traits
  * This also necessitates that we convert between DType and a Field rather than a DataType, as Field carries the Arrow extension type metadata

Another question this begs: should stats be pluggable in the same way layouts are? One very simple way we can do much better than GeoParquet, which only prunes at the whole-file level, is by implementing a spatial bloom filter (with [S2 cell IDs](http://s2geometry.io/devguide/s2cell_hierarchy.html) instead of hash functions). Or maybe you don't need a new Stat for this and it's just a Layout?

PR #3304, opened 2025-05-14
feat: extension pluggability

a10y:
Extracting this out of #3213.

This adds a new `ArrowTypeConversion` trait, which provides a way to support externally pluggable extension types for Vortex. Downstream crates can register these implementations using the `DType::register_extension_type` associated function. I originally went with `inventory` crate but there are a whole slew of issues with using it, that I tried to document in [this repo](https://github.com/a10y/testinv/blob/main/plugin_use/src/main.rs).

ArrowTypeConversion provides optional conversion in both directions between Vortex extension types and Arrow types. It also fixes up a few code paths to ensure that any struct conversions will attach the Arrow extension type/metadata

### DIFF
--- a/vortex-array/src/arrow/convert.rs
+++ b/vortex-array/src/arrow/convert.rs
@@ -234,18 +234,14 @@ impl FromArrowArray<&ArrowStructArray> for ArrayRef {
         let mut field_arrays = Vec::with_capacity(value.fields().len());
 
         for (field, column) in value.fields().iter().zip(value.columns()) {
-            let array = match field.extension_type_name() {
-                Some(_) => {
-                    match arrow_field_to_dtype(field).vortex_expect("arrow_field_to_dtype") {
-                        Some(DType::Extension(ext_dtype)) => {
-                            let storage = Self::from_arrow(column.clone(), field.is_nullable());
-                            ExtensionArray::new(ext_dtype, storage).into_array()
-                        }
-                        _ => Self::from_arrow(column.clone(), field.is_nullable()),
-                    }
+            let mut array = Self::from_arrow(column.clone(), field.is_nullable());
+            if field.extension_type_name().is_some() {
+                if let Some(DType::Extension(ext_dtype)) =
+                    arrow_field_to_dtype(field).vortex_expect("arrow_field_to_dtype")
+                {
+                    array = ExtensionArray::new(ext_dtype, array).into_array()
                 }
-                None => Self::from_arrow(column.clone(), field.is_nullable()),
-            };
+            }
             field_names.push(field.name().to_string().into());
             field_arrays.push(array);
         }
--- a/vortex-array/src/arrow/convert.rs
+++ b/vortex-array/src/arrow/convert.rs
@@ -166,8 +166,8 @@ where
         }
         DataType::Time32(time_unit) => TemporalArray::new_time(arr, time_unit.into()).into(),
         DataType::Time64(time_unit) => TemporalArray::new_time(arr, time_unit.into()).into(),
-        DataType::Date32 => TemporalArray::new_date(arr, TimeUnit::D).into(),
-        DataType::Date64 => TemporalArray::new_date(arr, TimeUnit::Ms).into(),
+        DataType::Date32 => TemporalArray::new_date(arr, TimeUnit::Day).into(),
+        DataType::Date64 => TemporalArray::new_date(arr, TimeUnit::Milli).into(),
         DataType::Duration(_) => unimplemented!(),
         DataType::Interval(_) => unimplemented!(),
         _ => vortex_panic!("Invalid temporal type: {}", T::DATA_TYPE),
--- a/vortex-array/src/arrow/convert.rs
+++ b/vortex-array/src/arrow/convert.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use arrow_array::array::{
     Array as ArrowArray, ArrayRef as ArrowArrayRef, ArrowPrimitiveType,
     BooleanArray as ArrowBooleanArray, GenericByteArray, NullArray as ArrowNullArray,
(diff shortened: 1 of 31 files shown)


vortex-array/src/arrow/convert.rs
```diff
@@ -227,14 +230,29 @@ impl FromArrowArray<&ArrowBooleanArray> for ArrayRef {
...
+                    match arrow_field_to_dtype(field).vortex_expect("arrow_field_to_dtype") {
+                        Some(DType::Extension(ext_dtype)) => {
+                            let storage = Self::from_arrow(column.clone(), field.is_nullable());
+                            ExtensionArray::new(ext_dtype, storage).into_array()
+                        }
+                        _ => Self::from_arrow(column.clone(), field.is_nullable()),
+                    }
+                }
+                None => Self::from_arrow(column.clone(), field.is_nullable()),
+            };
```
danking:
```suggestion
            let mut array = Self::from_arrow(column.clone(), field.is_nullable())
            if field.extension_type_name().is_some() {
                if let Some(DType:Extension(ext_dtype) =             arrow_field_to_dtype(field).vortex_expect("arrow_field_to_dtype") {
                    array = ExtensionArray::new(ext_dtype, array).into_array()
                }
            }
```

danking: Code golfing as procrastination from design docs 🤦

gatesn (changes requested):
Gonna intercept this until I can go through it if that's ok! Feel free to dismiss if it's blocking.

vortex-error/src/lib.rs
```diff
@@ -418,6 +418,21 @@ macro_rules! vortex_bail {
     };
 }
 
+/// Convenience macro that bails if an assertion fails.
+#[macro_export]
+macro_rules! vortex_assert {
```
gatesn, line 423: I'm not a huge fan of this... because assert! macros typically panic, but this one returns a result?

vortex-array/src/arrow/convert.rs
```diff
@@ -227,14 +230,25 @@ impl FromArrowArray<&ArrowBooleanArray> for ArrayRef {
...
+            field_arrays.push(array);
+        }
+
         StructArray::try_new(
-            value.column_names().iter().map(|s| (*s).into()).collect(),
-            value
-                .columns()
-                .iter()
-                .zip(value.fields())
-                .map(|(c, field)| Self::from_arrow(c.clone(), field.is_nullable()))
```
gatesn, line 236: Seems like we should just change `Self::from_arrow(ArrowArray, is_nullable: bool)` to `Self::from_arrow(ArrowArray, Field)` ?

vortex-dtype/src/arrow.rs
```diff
@@ -131,35 +260,66 @@ impl FromArrowType<(&DataType, Nullability)> for DType {
...
+                "failed to resolve ArrowTypeConversion for unrecognized extension type \"{}\"",
+                ext_type
+            );
+        }
+
         Self::from_arrow((field.data_type(), field.is_nullable().into()))
     }
 }
 
+/// Registry of Arrow extension type conversions.
```
gatesn, line 282: This was my question around whether extension dtypes should have vtables. This registry works, but it's very specific to arrow conversion. Vs saying you can register an extension dtype, and then possibly even define things like ordering semantics etc. 🤷 Maybe that's being too generic too early.

gatesn: The other thing is we just made `to_arrow` a generic compute function, so there's no special handling of it whatsoever. Seems a shame to not try and plumb this conversion through the same logic, but I really don't know how or whether it's worth it / the correct thing to do.

vortex-dtype/src/arrow.rs
```diff
@@ -189,34 +349,121 @@ impl DType {
...
+        }
+    }
+
+    Ok(None)
+}
+
+/// Convert an Arrow [`Field`] to a `DType` using a [registered extension][ArrowTypeConversion].
+///
+/// If no suitable conversion has been registered, `Ok(None)` is returned.
+pub fn arrow_field_to_dtype(field: &Field) -> VortexResult<Option<DType>> {
```
gatesn, line 419: This is specifically for arrow _extension fields_ right?

vortex-dtype/src/arrow.rs
```diff
@@ -189,34 +349,121 @@ impl DType {
...
+
+/// Conversions between Arrow [`Field`] type and Vortex logical `DType`.
+///
+/// This crate provides infra for registering and discovering extension types.
+/// Once you implement this trait, you need to register it using [`DType::register_extension_type`]
+/// to have it get discovered.
+///
+/// See also: [`DType::register_extension_type`]
+pub trait ArrowTypeConversion: 'static + Send + Sync + Debug {
+    /// Convert the given Arrow [`Field`] to a Vortex [`DType`].
```
gatesn, line 437: Can you document what the `None` result means?

vortex-dtype/src/arrow.rs
```diff
@@ -189,34 +349,121 @@ impl DType {
...
+    /// extension type, to preserve round-trip between the two.
+    #[allow(clippy::disallowed_types)]
+    fn to_arrow(&self, _dtype: &ExtDType) -> VortexResult<Option<Field>> {
+        Ok(None)
+    }
+}
+
+/// Conversion token
+#[derive(Debug)]
+pub struct ArrowTypeConversionRef(pub ArcRef<dyn ArrowTypeConversion>);
```
gatesn, line 455: We tend to just type alias `XXRef` things. Then you get all the other traits impl'd by ArcRef.

vortex-error/src/lib.rs
a10y: yea i'm happy to change the name i just find this much more readable than the long-form if pattern

vortex-dtype/src/arrow.rs
a10y: I patterned this off of the XYZKernelRef structs which all do this, unless something has changed in the last 48 hours

a10y:
@gatesn curious for your thoughts as i'm hitting some friction points in this refactor.

I have the ExtensionType trait defined which is the in-memory deserialized/owned version of an ext type. It's not a dyn ptr, it's just a trait. The idea is that you can downcast from an ExtDType (which remains the in-memory serialized form of an extension type) into a T: ExtensionType (the in-memory deserialized form of metadata).

image


Temporal types keep the same binary metadata serde they have right now (time unit + timezone, if it's a Timestamp).

I am modeling all arrow extension types has a single implementation since they share a common serde (to/from JSON map).

One thing I'm not sure of is where the logic for converting an extensiontype to an Arrow field should go? I can think of two places

1. A new Kernel-like type that is dynamically registerable, similar to how this PR was originally structured with the `ArrowConversionRef`
2. As a method on the `ExtensionType` itself, e.g. `ExtensionType::to_field()` from the screenshot above

I'm kinda concerned about just heaping all extensibility into the trait methods though.

gatesn:
1) I think we may want to move towards having an in-memory ExtType dyn-trait, along with an ExtTypeEncoding, and ExtType vtable. Same setup as array, layout, and soon expressions.
2) I don't think we need to convert dtypes to arrow. The ToArrow compute function should be sufficient?

a10y:
I think you still need to support conversion to an Arrow schema outside of `ToArrow`. e.g. for Datafusion?

gatesn:
I think in those limited (and lossy!) cases, we should fall back to `to_arrow(&Canonincal::empty(dtype))?.data_type()`

PR #3324, opened 2025-05-16
feat: add is_null expression

delta003:


vortex-expr/src/is_null.rs
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    fn children(&self) -> Vec<&ExprRef> {
+        vec![&self.child]
+    }
+
+    fn replacing_children(self: Arc<Self>, children: Vec<ExprRef>) -> ExprRef {
+        assert_eq!(children.len(), 0);
+        self
```
robert3005: I think you want to replace the single child you have here?

delta003: good catch!

vortex-expr/src/is_null.rs
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    fn children(&self) -> Vec<&ExprRef> {
+        vec![&self.child]
+    }
+
+    fn replacing_children(self: Arc<Self>, children: Vec<ExprRef>) -> ExprRef {
+        assert_eq!(children.len(), 1);
+        Self::new_expr(children[0].clone())
```
robert3005: I would `.pop()` here and then expect the option you get since you have an assert before. You could also remove the assert and just do `Self::new_expr(children.pop().vortex_expect(..))` which should be the same

robert3005 (approved):

PR #3325, opened 2025-05-16
chore: Enforce usage of parking_lot

AdamGS:
Mostly because of better ergonomics, but also has better performance in some cases and some useful additional features.


lwwmanning:
should maybe put the parking lot Mutex/RwLock in `aliases`? the same way we do with hashbrown Hashmap.

AdamGS:
I thought about that but the API is slightly different, the hashmap thing became a general thing in Rust because the std itself exposes `hashbrown` which is generic over the hasher

gatesn:
Also aliases is more for re-exporting things on our own API, e.g. the paste macro is used from within the exported vtable macro.

gatesn (approved):
Curious if we see any performance change

robert3005:
Generally on linux parking_lot is only better on uncontended locks. For MacOS the difference is a lot more prononuced since there's no good allocation free locking apis in the system

PR #3273, opened 2025-05-12
VTable Trait

gatesn:
Arrays to encodings are like instances to classes in OOP. This PR fixes a couple of things to make this more true:

* Encodings can now hold heap-allocated data, in the same way that arrays can. This makes it possible to e.g. tie a Python class object to an encoding by holding a reference to the Python object.
* Both encoding and array (class and instance) functions now appear on the _same_ vtable trait, with encoding and array functions being passed their respective heap-allocated data as the first argument. This trait is implemented for a "<Encoding>VTable" struct, and so is hidden from users of the "<Encoding>Encoding" and "<Encoding>Array" structs.
* Because the vtable trait is now both independent _and_ non-object-safe, we can auto-implement lots of things (see e.g.  ValidityVTableForValidityChild which automatically delegates all validity calls to a child array).
* ArrayImpl and its ugly underscore-prefixed functions are gone because behaviours are not implemented directly on the array structs (which previously exposed this behaviour as semi-public API). We therefore retain the ability to perform pre- and post-condition checks when we `impl Array for ArrayAdapter<VTable>`.
* `dyn Array` and `dyn Encoding` are now sealed traits, meaning we control that all implementations must go via a vtable trait and therefore we can enforce the pre- and post-conditions, vs previously where someone could work around them. worked 

Possible downsides:
* The array structs, e.g. PrimitiveArray, don't themselves `impl Array`, therefore cannot be passed as `&primitive_array` into a function that accepts a `&dyn Array`. They do however both `Deref` and `AsRef<dyn Array>`, so they can be passed using `primitive_array.as_ref()` and all `Array` trait functions can be called directly.
* Downcasting / interpretting arrays can only be done via the VTable structs, e.g. `array.is::<PrimitiveVTable>` or `array.is_opt::<ConstantVTable>`. This is because the actual type that impls the array trait is `ArrayAdapter<VTable>`, and we cannot describe this type without the `VTable` generic parameter.

Alternative... we go all the way and just hold a "real" vtable. The problem is we'd just end up introducing a trait similar to this one to automate the downcasting of the array/encoding parameters.
```rust
struct VTable {
  len: fn(&dyn Array) -> usize,
  serde: Optional<SerdeVTable>,
}
```

PR #3331, opened 2025-05-18
feat: Layout VTable

gatesn:
Similar refactor as #3273 but for layouts.

Again, this introduces the idea that a `LayoutEncoding` can be independently heap allocated, and is instantiated as a `Layout`. Slight difference to arrays is that the `Layout` tree is lazy in terms of parsing children, so child accessors are fallible.


vortex-layout/src/vtable.rs
```diff
@@ -1,60 +1,125 @@
...
+use crate::segments::{SegmentId, SegmentSource};
+use crate::{
+    IntoLayout, Layout, LayoutChildType, LayoutEncoding, LayoutEncodingRef, LayoutId,
+    LayoutReaderRef, LayoutRef,
+};
 
-/// A reference to a layout VTable, either static or arc'd.
-pub type LayoutVTableRef = ArcRef<dyn LayoutVTable>;
+pub trait VTable: 'static + Sized + Send + Sync + Debug {
+    type Layout: 'static + Send + Sync + Clone + Deref<Target = dyn Layout> + IntoLayout;
```
robert3005: If we want to have Debug for Layouts we should add Debug here so it correctly recurses?

vortex-layout/src/lib.rs
```diff
@@ -20,6 +27,8 @@ mod writer;
 
 pub type LayoutId = ArcRef<str>;
 
+// FIXME(ngates): remove these
```
robert3005: we shouldn't merge the fixme

vortex-layout/src/layouts/dict/mod.rs
```diff
@@ -1,43 +1,139 @@
...
     ) -> VortexResult<()> {
-        layout
-            .child(1, layout.dtype().clone(), "codes")?
-            .register_splits(field_mask, row_offset, splits)
+        layout.codes.register_splits(field_mask, row_offset, splits)
+    }
+
+    fn new_reader(
+        layout: &Self::Layout,
+        name: &Arc<str>,
```
robert3005, line 82: How about we make this ArcRef<str>? Though that's likely minor. Could also create a `const UNNAMED: Arc<str> = Arc::from("");` and clone that instead of creating new ones

gatesn: We tend not to expose ArcRef on the public API. We could alias it to LayoutName I suppose

robert3005 (approved):

PR #3336, opened 2025-05-19
fix: duckdb tls on linux machine as default

joseph-isaacs:
Use the makefile to ensure `-ftls-model=global-dynamic`

### DIFF
--- a/duckdb-vortex/Makefile
+++ b/duckdb-vortex/Makefile
@@ -20,7 +20,7 @@ export VCPKG_TOOLCHAIN_PATH := ${PROJ_DIR}vcpkg/scripts/buildsystems/vcpkg.cmake
 export BUILD_MAIN_DUCKDB_LIBRARY=0
 export DISABLE_BUILTIN_EXTENSIONS=1
 
-# This is not needed on macOS, we don't see a tls there.
+# This is not needed on macOS, we don't see a tls error on load there.
 ifeq ($(shell uname), Linux)
     export CFLAGS=-ftls-model=global-dynamic
 endif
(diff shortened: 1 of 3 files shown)


duckdb-vortex/Makefile
```diff
@@ -20,4 +20,9 @@ export VCPKG_TOOLCHAIN_PATH := ${PROJ_DIR}vcpkg/scripts/buildsystems/vcpkg.cmake
 export BUILD_MAIN_DUCKDB_LIBRARY=0
 export DISABLE_BUILTIN_EXTENSIONS=1
 
+# TODO(joe): fix me on windows
```
0ax1: Given that we don't support Windows anywhere else right now, I'd say we don't need this comment.

Can we add a comment though we don't need to set this flag for macOS? As in, we know that there's sufficient TLS storage when using the standard TLS mode.

joseph-isaacs:
```suggestion
# This is not needed on macOS, we don't see a tls there.
```

duckdb-vortex/Makefile
```diff
@@ -20,4 +20,9 @@ export VCPKG_TOOLCHAIN_PATH := ${PROJ_DIR}vcpkg/scripts/buildsystems/vcpkg.cmake
 export BUILD_MAIN_DUCKDB_LIBRARY=0
 export DISABLE_BUILTIN_EXTENSIONS=1
 
+# This is not needed on macOS, we don't see a tls there.
```
joseph-isaacs:
```suggestion
# This is not needed on macOS, we don't see a tls error on load there.
```

0ax1:
```suggestion
# This is not needed on macOS, we don't see a thread-local storage issue there.
```

0ax1 (approved):

PR #3339, opened 2025-05-19
fix: compression benchmark

joseph-isaacs:
Build the duckdb extension explicitly

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -107,6 +107,15 @@ jobs:
       - name: Install gzip
         run: sudo apt-get update && sudo apt-get install -y gzip ninja-build
 
+      - name: DuckDB extension build
+        env:
+          # Build DuckDB and the Vortex extension with `-march=native`.
+          # The `NATIVE_ARCH` environment variable is picked up by `duckdb/Makefile`.
+          NATIVE_ARCH: 1
+          DUCKDB_PLATFORM: linux_amd64
+        run: GEN=ninja make release
+        working-directory: ${{ github.workspace }}/duckdb-vortex
+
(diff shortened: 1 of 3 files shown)


bench-vortex/src/datasets/tpch_l_comment.rs
```diff
@@ -17,8 +17,6 @@ impl Dataset for TPCHLCommentChunked {
     }
 
     async fn to_vortex_array(&self) -> ArrayRef {
```
joseph-isaacs, line 19: I removed this since we don't often use this locally and its more confusing to have this implicit build

.github/workflows/bench.yml
```diff
@@ -107,6 +107,15 @@ jobs:
       - name: Install gzip
         run: sudo apt-get update && sudo apt-get install -y gzip ninja-build
 
+      - name: DuckDB extension build
```
gatesn, line 110: Do you know if this does, or can, use SCCACHE?

joseph-isaacs: maybe its mostly the internal rust stuff that is slow

0ax1: @gatesn Note that we are caching the compiled binary as of today:
```
      - name: Cache duckdb compiled binary
        uses: runs-on/cache@v4
        id: cache-duckdb-binary
        with:
          fail-on-cache-miss: true
          key: "${{ runner.os }}-duckdb-linux_amd64-${{ env.DUCKDB_VERSION }}"
          path: ${{ github.workspace }}/duckdb/build/release/duckdb
```

0ax1 (approved):

PR #3343, opened 2025-05-19
Cache FSST compressor for running compare

gatesn:
Alternative would be to change fsst crate to hold `Cow` of symbols internally and be cheap to construct without validation?

### DIFF
--- a/encodings/fsst/src/array.rs
+++ b/encodings/fsst/src/array.rs
@@ -1,4 +1,7 @@
-use fsst::{Decompressor, Symbol};
+use std::fmt::{Debug, Formatter};
+use std::sync::{Arc, LazyLock};
+
+use fsst::{Compressor, Decompressor, Symbol};
 use vortex_array::arrays::VarBinArray;
 use vortex_array::stats::{ArrayStats, StatsSetRef};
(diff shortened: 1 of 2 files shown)


encodings/fsst/src/array.rs
```diff
@@ -42,6 +45,21 @@ pub struct FSSTArray {
     /// Lengths of the original values before compression, can be compressed.
     uncompressed_lengths: ArrayRef,
     stats_set: ArrayStats,
+
+    /// Memoized compressor used for push-down of compute by compressing the RHS.
+    compressor: Arc<LazyLock<Compressor, Box<dyn Fn() -> Compressor + Send>>>,
```
a10y, line 50: interesting, why does it need the `Send` hint? Compressor is a struct, shouldn't it be able to infer that it's send?

gatesn: Not sure, it complains that the future isn't send, therefore the arc isn't sync

a10y (approved):
In the limit I'm curious how we evict stuff like this from memory, but the symbol table is of fixed and pretty small size so this seems fine

PR #3345, opened 2025-05-19
Faster list filter

gatesn:


a10y (approved):

a10y:
So this is less about Arrow having a bad kernel and more so just being able to pushdown the elements filter right?

PR #3348, opened 2025-05-19
Reduce list contains with binary search

gatesn:


vortex-array/src/compute/list.rs
```diff
@@ -133,20 +139,37 @@ fn list_contains_null(list_array: &ListArray) -> VortexResult<ArrayRef> {
...
-        .collect();
+    // TODO(ngates): what threshold is reasonable here?
+    let num_matches = matches.count_set_bits();
+    if num_matches < 20 || (num_matches as f64 / matches.len() as f64) < 0.025 {
+        let mut result = MutableBuffer::from_len_zeroed(ends.len().div_ceil(8));
+        for idx in matches.set_indices() {
+            let end = ends
+                .search_sorted(
+                    &T::from_usize(idx).vortex_expect("invalid index"),
+                    SearchSortedSide::Left,
```
gatesn, line 155: I think this should be Right...?

stuhood:
Thanks again for looking at this!

FWIW: I have an (unstable) sketch of this that looks like:
```rust
    let mut mask = BooleanBufferBuilder::new(ends.len() - 1);

    // Sparse array: binary search in offsets and set individual bits.
    let mut set_bits = matches.set_indices();
    let mut ends_offset = 0;
    mask.resize(ends.len() - 1);
    while let Some(bit) = set_bits.next() {
        if bit < ends[ends_offset].as_() {
            // This bit is in a list that we have already visited.
            continue;
        }

        // This bit is in a list that we haven't visited yet.
        let idx = match (&ends[ends_offset..]).binary_search_by_key(&bit, |t| t.as_()) {
            Ok(idx) => idx,
            Err(idx) => idx - 1,
        };
        mask.set_bit(ends_offset + idx, true);
        ends_offset += idx + 1;
    }

    BoolArray::new(mask.into(), validity).into_array()
```

... because it shrank the domain of the binary search on each iteration.

PR #3352, opened 2025-05-20
Return null-d array in `between`

blaginin:
Now on dev: 

```rust

    let a = vortex::arrays::PrimitiveArray::from_iter([0, 0, 2, 0, 2]);
    let b = vortex::arrays::PrimitiveArray::from_iter([0, 0, 2, 0, 2]);
    let c = ConstantArray::new(Scalar::null(DType::Primitive(PType::I32, Nullability::Nullable)), 5);

    let opts = BetweenOptions {
        lower_strict: StrictComparison::NonStrict,
        upper_strict: StrictComparison::NonStrict,
    };

    let t = vortex::compute::between(
        &b.to_array(),
        &a.to_array(),
        &c.to_array(),
        &opts
    ).unwrap().to_bool().unwrap();
```
```
thread 'main' panicked at src/main.rs:44:7:
called `Result::unwrap()` on an `Err` value: Internal error: compute function between returned a result of length 0 but expected 5
```

### DIFF
--- a/vortex-array/src/compute/between.rs
+++ b/vortex-array/src/compute/between.rs
@@ -312,7 +312,7 @@ mod tests {
         .to_bool()
         .unwrap();
 
-        let indices = to_int_indices(matches);
+        let indices = to_int_indices(matches).unwrap();
         assert_eq!(indices, expected);
     }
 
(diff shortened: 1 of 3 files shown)


joseph-isaacs (commented):
Looks good once we can get CI to pass 🔥

vortex-array/src/compute/between.rs
joseph-isaacs, line 1: Some tests with nullable lower and upper would be nice, but not a blocker

joseph-isaacs (approved):

joseph-isaacs:
Thanks for contributing!

PR #3354, opened 2025-05-20
fix: only assign thread local file idx on pinning

0ax1:


duckdb-vortex/src/vortex_scan.cpp
```diff
@@ -273,6 +272,7 @@ static void CreateScanPartitions(ClientContext &context, const BindData &bind, S
 		};
 
 		if (PinFileToThread(global_state)) {
```
joseph-isaacs: You can pull out this check (before the loop) to avoid reevaluating.

Also assign the value once out there?

0ax1: Well we did have the out if once. But sure we can add an additional one.

0ax1: done

duckdb-vortex/src/vortex_scan.cpp
```diff
@@ -263,7 +263,10 @@ static void CreateScanPartitions(ClientContext &context, const BindData &bind, S
 
 	const auto partition_count = std::max(static_cast<uint64_t>(1), row_count / partition_size);
 	global_state.partitons_total += partition_count;
-	local_state.thread_local_file_idx = file_idx;
+
+	if (PinFileToThread(global_state)) {
```
joseph-isaacs: Why do you reeval this in each iteration below?

0ax1: Extracted this into a var if thats what you meant.

joseph-isaacs (approved):
LG

PR #3356, opened 2025-05-20
Clean up stats propagation for slicing

gatesn:
Reduces the amount we copy some stats (by removing into_iter that forces a full stats copy), and also makes duckdb databases idempotent.


vortex-array/src/array/mod.rs
```diff
@@ -352,41 +381,21 @@ impl<V: VTable> Array for ArrayAdapter<V> {
...
-            }
-            if is_strict_sorted == Some(Precision::Exact(true)) {
-                stats.set(Stat::IsStrictSorted, Precision::exact(true));
+        // Propagate some stats from the original array to the sliced array.
+        let mut stats = StatsSet::default();
+        self.statistics().with_iter(|iter| {
+            for (stat, value) in iter {
+                match stat {
+                    Stat::IsSorted | Stat::IsStrictSorted | Stat::IsConstant => {
+                        stats.set(*stat, value.clone());
```
lwwmanning: this is incorrect! it could have been non-constant before but is constant now. previous version had careful handling of this

lwwmanning (dismissed):
see comment

vortex-array/src/array/mod.rs
joseph-isaacs: +1

gatesn: I think we found a reasonable compromise

vortex-array/src/array/mod.rs
```diff
@@ -397,19 +393,33 @@ impl<V: VTable> Array for ArrayAdapter<V> {
...
-            let mut stats = sliced.statistics().to_owned();
-            stats.combine_sets(&derived_stats, self.dtype())?;
-            for (stat, val) in stats.into_iter() {
-                sliced.statistics().set(stat, val)
-            }
+        // Propagate some stats from the original array to the sliced array.
+        if !sliced.is::<ConstantVTable>() {
+            self.statistics().with_iter(|iter| {
+                sliced.statistics().inherit(iter.filter(|stat| {
+                    match stat {
```
lwwmanning, line 408: nit: probably simpler to write `matches!(stat, Stat::IsConstant | Stat::IsSorted | Stat::IsStrictSorted) && value.as_bool().vortex_expect("must be a bool").unwrap_or_default()`

vortex-array/src/arrays/constant/mod.rs
```diff
@@ -58,11 +58,10 @@ impl ConstantArray {
         S: Into<Scalar>,
     {
         let scalar = scalar.into();
-        let stats = StatsSet::constant(scalar.clone(), len);
```
robert3005, line 64: I think we need the compute functions otherwise constant arrays have no stats

gatesn: As in, we need to implement the compute functions?

robert3005: Yes, implement the compute functions

gatesn: Done. Most of the compute functions already had special case for `is::<ConstantVTable>`

robert3005 (dismissed):

vortex-array/src/array/mod.rs
```diff
@@ -397,19 +393,27 @@ impl<V: VTable> Array for ArrayAdapter<V> {
...
-            for (stat, val) in stats.into_iter() {
-                sliced.statistics().set(stat, val)
-            }
+        // Propagate some stats from the original array to the sliced array.
+        if !sliced.is::<ConstantVTable>() {
+            self.statistics().with_iter(|iter| {
+                sliced.statistics().inherit(iter.filter(|stat| {
+                    matches!(stat, (
+                               Stat::IsConstant | Stat::IsSorted | Stat::IsStrictSorted,
+                               Precision::Exact(value),
```
lwwmanning: this is still potentially incorrect?

robert3005: This matches previous code in behaviour

vortex-tui/src/browse/ui/layouts.rs
```diff
@@ -162,14 +162,17 @@ fn render_array(app: &AppState, area: Rect, buf: &mut Buffer, is_stats_table: bo
...
-        let rows = array.statistics().into_iter().map(|(stat, value)| {
-            let value = value.into_scalar(
-                stat.dtype(array.dtype())
-                    .vortex_expect("stat invalid for dtype"),
-            );
-            let stat = Cell::from(Text::from(format!("{stat}")));
-            let value = Cell::from(Text::from(format!("{value}")));
-            Row::new(vec![stat, value])
+        let rows = array.statistics().with_iter(|iter| {
+            iter.map(|(stat, value)| {
```
lwwmanning, line 166: I don't understand why you changed to
```
...with_iter(|iter| {
            iter.map(...)
}
```

gatesn:
I have no recollection of this PR 😊

vortex-tui/src/browse/ui/layouts.rs
robert3005: This is a borrowed iterator so it only lives for the duration of the closure

robert3005: Previously into iter made a copy

vortex-array/src/array/mod.rs
```diff
@@ -388,19 +385,28 @@ impl<V: VTable> Array for ArrayAdapter<V> {
...
+                           ) if value
+                            .as_bool()
+                            .vortex_expect("must be a bool")
+                            .unwrap_or_default())
```
a10y: is there a less obnoxiously clever way to write this lol

vortex-array/src/stats/stats_set.rs
```diff
@@ -47,38 +47,6 @@ impl StatsSet {
         StatsSet::new_unchecked(vec![(Stat::NullCount, Precision::exact(0))])
     }
 
-    pub fn constant(scalar: Scalar, length: usize) -> Self {
```
a10y, line 50: was this unused?

vortex-array/src/arrays/constant/compute/sum.rs
```diff
@@ -13,21 +13,28 @@ use crate::stats::Stat;
...
         let scalar = array.scalar();
+        let sum_value = match scalar.dtype() {
+            DType::Bool(_) => ScalarValue::from(match scalar.as_bool().value() {
+                None | Some(false) => 0u64,
+                Some(true) => array.len() as u64,
+            }),
+            DType::Primitive(ptype, _) => {
+                match_each_native_ptype!(
+                    ptype,
+                    integral: |T| { sum_integral::<u64>(scalar.as_primitive(), array.len())?.into() },
```
a10y: is this right? what about for sum(i32) with negative values?

a10y (changes requested):
If I'm reading this right, there might be a bug in Sum impl for constantarray with negative values. Otherwise mostly stylistic requests/q's

vortex-array/src/stats/stats_set.rs
robert3005: This was only ever used in ConstantArray::new which no longer by default populates stats and instead relies on compute functions

vortex-array/src/arrays/constant/compute/sum.rs
robert3005: Oh well... this is what I lost in the rebase

vortex-array/src/array/mod.rs
robert3005: I tried

vortex-array/src/compute/sum.rs
```diff
@@ -129,77 +129,8 @@ pub fn sum_impl(
     }
 
     // If the array is constant, we can compute the sum directly.
-    if let Some(mut constant) = array.as_constant() {
```
robert3005, line 132: I realize we had constant array sum kernel and we implemented it here as well

vortex-array/src/compute/sum.rs
```diff
@@ -129,77 +129,8 @@ pub fn sum_impl(
     }
 
     // If the array is constant, we can compute the sum directly.
```
robert3005: fix comment

vortex-array/src/compute/sum.rs
robert3005: I am happy to move this out of this pr but I found this while chasing down constant array sum behaviour

vortex-array/src/compute/sum.rs
```diff
@@ -128,78 +128,9 @@ pub fn sum_impl(
         };
     }
 
-    // If the array is constant, we can compute the sum directly.
-    if let Some(mut constant) = array.as_constant() {
-        if constant.is_null() {
```
robert3005, line 133: @gatesn any reason you went with sum of all null being 0? I wonder if we need an option to be sql compliant

a10y:
@claude is this true

a10y (approved):

PR #3358, opened 2025-05-20
feat[ffi]: add a session used for caching

joseph-isaacs:
For now this is a footer cache but we can cache more if we want too

### DIFF
--- a/duckdb-vortex/src/include/vortex_session.hpp
+++ b/duckdb-vortex/src/include/vortex_session.hpp
@@ -3,6 +3,8 @@
 #include "vortex.hpp"
 #include "duckdb/storage/object_cache.hpp"
 
+namespace vortex {
+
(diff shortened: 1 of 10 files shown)


gatesn (commented):
Is this because you don't want to cache these in C++?

joseph-isaacs:
What would you cache in C++ open files or expose a footer?

gatesn (approved):

duckdb-vortex/src/include/vortex_session.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "vortex.hpp"
+#include "duckdb/storage/object_cache.hpp"
+
+class VortexSession : public duckdb::ObjectCacheEntry {
```
0ax1, line 8: `namespace vortex`

PR #3359, opened 2025-05-20
feat: VTables for extension types

a10y:


vortex-dtype/src/extension/vtable/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+use crate::{ExtID, ExtMetadata};
+
+/// Convert a typed value into a type-erased handle to an extension type.
+pub trait IntoExtensionTypeRef {
+    /// Consume the receiver and return a new cloneable handle to an extension type.
+    fn into_extension_type_ref(self) -> ExtensionTypeRef;
+}
+
+/// VTable for extension types.
+pub trait ExtensionVTable: Send + Sync + Debug + 'static {
```
a10y, line 21: i really don't like these. something feels off, there has to be a less complex way to do all of this

gatesn: I think this is how it would look without them:

```
pub trait ExtTypeEncoding { 
  fn decode(&self, metadata: &[u8]) -> Result<Arc<dyn ExtType>>;
}

pub trait ExtType {
  fn encode(&self) -> ExtTypeData; // Or whatever functions we have. 
}
```

The problems we wanted to work around are:
* We want to make some assertions about the implementation (post-conditions). e.g. that the result of the decode function is actually the type we expected. This is optional I suppose, but is nice for our sanity and debugging.
* We often want to perform pre-conditions and other shared logic. e.g. we do bounds checking in the array slice function, or we use an associated metadata type to perform deserialization into a concrete type.
* We can auto-implement portions of the vtable, e.g. see how we do things like `ValidityFromValidityChild`.
* We can the implementation trait API separate from the exposed API. This kind of covers the points above too, but in practice, you can thing of this as having java's `protected` keyword. The vtable functions are protected, and the Array trait is the public interface that calls into them. The separation of the public API trait from the adapter trait is purely to allow us to benefit from associated types and implement shared logic in a monomorphised way.

In this case, we don't have an associated metadata type, and we don't (yet?) have functions that require post-conditions. So I agree it looks a tad superfluous...

PR #3360, opened 2025-05-20
duckdb: drain local queues to enable work-stealing

0ax1:


0ax1:
Closed, as we will push the scan into Rust.

PR #3362, opened 2025-05-21
Idempotent DuckDB bench

gatesn:


bench-vortex/src/bin/clickbench.rs
```diff
@@ -89,15 +88,13 @@ struct DataFusionCtx {
...
-    tmp_dir: TempDir,
 }
 
 impl DuckDBCtx {
     pub fn duckdb_file(&self, format: Format) -> PathBuf {
-        self.tmp_dir
-            .path()
-            .to_path_buf()
-            .join(format!("hits-{format}.db"))
+        let dir = Path::new("bench-vortex/data/duckdb");
```
0ax1, line 95: @gatesn  Shouldn't the `dir` path be relative to the Git root, not relative to the current working dir? For local runs, we can't assume they're the same.

gatesn: Yep, laziness on my part :(

PR #3363, opened 2025-05-21
File pruning when set

joseph-isaacs:
Add whole file pruning to the vortex-file/ffi


vortex-layout/src/scan/mod.rs
```diff
@@ -161,6 +174,26 @@ impl<A: 'static + Send> ScanBuilder<A> {
...
+        if let Some(((file_stats, filter), struct_dtype)) = self
+            .file_stats
+            .zip(self.filter.as_ref())
+            .zip(layout_reader.dtype().as_struct())
+        {
+            if let Some(predicate) = PruningPredicate::try_new(filter) {
+                if let Some(struct_row) = extract_struct_row(&predicate, &file_stats, struct_dtype)?
+                {
+                    if let Some(result) = predicate
+                        .evaluate(&struct_row)?
```
gatesn: Can you fix the evaluate function to take a trait? Maybe it looks like:

```
StatsSource {
 fn fetch(&self, field_path: &FieldPath, stat: &Stat) -> Option<ArrayRef>;
}
```

That way you don't need to construct the full stats array if we only end up using one field.

vortex-layout/src/scan/mod.rs
```diff
@@ -48,6 +53,8 @@ pub struct ScanBuilder<A> {
     /// The executor used to spawn each split task.
     executor: Option<Arc<dyn TaskExecutor>>,
     metrics: VortexMetrics,
+    /// Should we try to prune the file (using stats) on open.
+    file_stats: Option<Arc<[StatsSet]>>,
```
gatesn, line 53: I think you could actually create a `FileStatsLayoutReader` and provide a `impl PruningEvaluator`. Then inside `VortexFile::layout_reader()` you can wrap up the root layout with the `FileStatsLayoutReader`, that way the scan never even knows.

joseph-isaacs: We agreed this does extra work, so going to add a method on the vortex file itself

vortex-layout/src/scan/mod.rs
joseph-isaacs: We already do this, would you prefer this method to live in a impl/trait?

vortex-ffi/src/file.rs
```diff
@@ -243,19 +237,32 @@ pub unsafe extern "C-unwind" fn vx_file_dtype(file: *const vx_file_reader) -> *m
...
-        let layout_reader = unsafe { layout_reader.as_ref().vortex_expect("null layout reader") };
-        let mut scan_builder = ScanBuilder::new(layout_reader.inner.clone());
+        let file_reader = unsafe { file_reader.as_ref().vortex_expect("null file reader") };
 
         let scan_options = unsafe { opts.as_ref() }.map_or_else(
             || Ok(ScanOptions::default()),
             |options| options.process_scan_options(),
         )?;
 
+        if let Some(expr) = &scan_options.filter_expr {
```
0ax1, line 252:
```
       if let Some(expr) = &scan_options.filter_expr {
            if file_reader.inner.can_prune(expr)? {
                let dtype = file_reader.inner.dtype().clone();
                let empty_iter = ArrayIteratorAdapter::new(dtype, std::iter::empty());

                return Ok(Box::into_raw(Box::new(vx_array_iterator {
                    inner: Some(Box::new(empty_iter)),
                })));
            }
        };
```

?

duckdb-vortex/src/vortex_scan.cpp
```diff
@@ -292,8 +290,7 @@ static void CreateScanPartitions(ClientContext &context, const BindData &bind, S
 	D_ASSERT(global_state.files_partitioned <= global_state.expanded_files.size());
 }
 
-static unique_ptr<ArrayIterator> OpenArrayIter(ScanGlobalState &global_state,
-                                               std::shared_ptr<LayoutReader> &layout_reader,
+static unique_ptr<ArrayIterator> OpenArrayIter(ScanGlobalState &global_state, shared_ptr<FileReader> &layout_reader,
```
0ax1: rename to `file_reader`?

duckdb-vortex/src/vortex_scan.cpp
```diff
@@ -380,18 +377,24 @@ static void VortexScanFunction(ClientContext &context, TableFunctionInput &data,
...
-				    OpenFileAndVerify(FileSystem::GetFileSystem(context), *bind_data.session, file_name, bind_data);
-				global_state.layout_readers[file_idx] = LayoutReader::CreateFromFile(vortex_file.get());
-				CreateScanPartitions(context, bind_data, global_state, local_state, file_idx, vortex_file);
+				if (file_idx == 0) {
+					global_state.file_readers[0] = bind_data.initial_file;
+				} else {
+					auto file_name = global_state.expanded_files[file_idx];
+					auto vortex_file =
+					    OpenFileAndVerify(FileSystem::GetFileSystem(context), *bind_data.session, file_name, bind_data);
+					global_state.file_readers[file_idx] = std::move(vortex_file);
```
0ax1: This is correct but in general I think we should be *very* careful with using `move` unless there's a clear perf upside.

vortex-ffi/src/file.rs
```diff
@@ -274,7 +281,9 @@ pub unsafe extern "C-unwind" fn vx_layout_reader_scan(
             scan_builder = scan_builder.with_split_by(split_by_value);
         }
 
-        vx_array_iterator(scan_builder.into_array_iter()?)
+        Ok(Box::into_raw(Box::new(vx_array_iterator {
+            inner: Some(Box::new(scan_builder.into_array_iter()?) as _),
```
0ax1:
```suggestion
            inner: Some(Box::new(scan_builder.into_array_iter()?)),
```

vortex-file/src/pruning.rs
```diff
@@ -0,0 +1,54 @@
...
+            return Ok(None);
+        };
+
+        let field_idx = struct_dtype.find(field)?;
+        let field_dtype = struct_dtype.field_by_index(field_idx)?;
+        let Some(cols) = stats_set[field_idx]
+            .iter()
+            .filter(|(stat, _)| stats.contains(stat))
+            .map(|(stat, value)| {
+                if let Precision::Exact(value) = value {
```
0ax1: This gets hard to read due to the formatting. Could consider extracting the closure, and move it outside of the loop.

vortex-ffi/src/file.rs
joseph-isaacs: This is needed, I believe

0ax1: Works on my machine tm.

vortex-file/src/pruning.rs
joseph-isaacs: I have refined

0ax1 (approved):

PR #3364, opened 2025-05-21
layout writer + strategy refactor for parallel writes

onursatici:


vortex-file/src/segments/ordered.rs
```diff
@@ -0,0 +1,115 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::task::Waker;
+
+use vortex_array::aliases::hash_map::HashMap;
+use vortex_buffer::ByteBuffer;
+use vortex_error::{VortexExpect, VortexResult, vortex_bail, vortex_err};
+use vortex_layout::segments::SegmentId;
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Hash, Debug)]
+pub(super) struct Section(Vec<usize>);
```
gatesn: DOCS

Cargo.toml
```diff
@@ -84,6 +85,7 @@ datafusion-execution = { version = "47" }
 datafusion-expr = { version = "47" }
 datafusion-physical-expr = { version = "47" }
 datafusion-physical-plan = { version = "47" }
+educe = { version = "0.6.0" }
```
gatesn: I'd rather we stay explicit tbh

encodings/dict/src/builders/mod.rs
```diff
@@ -22,7 +22,7 @@ pub const UNCONSTRAINED: DictConstraints = DictConstraints {
     max_len: usize::MAX,
 };
 
-pub trait DictEncoder: Send {
+pub trait DictEncoder: Send + Sync {
```
gatesn: Seems weird to require `Sync` here, why should all dictionary encoders be thread-safe?

vortex-file/src/segments/writer.rs
```diff
@@ -1,35 +1,75 @@
...
 
 use crate::footer::SegmentSpec;
 
-/// A segment writer that holds buffers in memory until they are flushed by a writer.
-#[derive(Default)]
-pub(crate) struct BufferedSegmentWriter {
-    /// A Vec byte buffers for segments
-    segments_buffers: Vec<Vec<ByteBuffer>>,
-    next_id: SegmentId,
+pub struct SerialSegmentWriter {
```
gatesn, line 13: Docs plz

vortex-file/src/writer.rs
```diff
@@ -1,44 +1,60 @@
+#![allow(dead_code)]
```
gatesn: Remove

vortex-file/src/writer.rs
```diff
@@ -1,44 +1,60 @@
...
         }
     }
 }
 
 impl VortexWriteOptions {
     /// Replace the default layout strategy with the provided one.
-    pub fn with_strategy<S: LayoutStrategy>(mut self, strategy: S) -> Self {
-        self.strategy = Box::new(strategy);
+    pub fn with_strategy<S: LayoutStrategy>(self, _strategy: S) -> Self {
+        todo!("make strategy configurable again");
```
gatesn: Bring this back

vortex-file/src/writer.rs
```diff
@@ -59,60 +75,57 @@ impl VortexWriteOptions {
 
 impl VortexWriteOptions {
     /// Perform an async write of the provided stream of `Array`.
-    pub async fn write<W: VortexWrite, S: ArrayStream + Unpin>(
+    pub async fn write<W: VortexWrite, S: ArrayStream + Unpin + Send + 'static>(
```
gatesn, line 68: Seems a shame that we now need a sendable array stream, I guess it makes sense for the tokio use-case.

vortex-layout/src/strategy.rs
```diff
@@ -3,29 +3,27 @@
...
-            Ok(FlatLayoutWriter::new(ctx.clone(), dtype.clone(), Default::default()).boxed())
-        }
-    }
+pub trait LayoutStrategy: 'static + Send + Sync {
+    fn write_stream(
+        &self,
+        ctx: &ArrayContext,
+        dtype: &DType,
+        segment_writer: Arc<dyn SegmentWriter>,
+        stream: SequentialArrayStream,
```
gatesn: I thought we didn't need a SequentialArrayStream if we supported "split_off" on the segment writer?

gatesn: If we really really need it, ideally it would also have a `dtype()` function on it, same as ArrayStream. Then remove the dtype parameter from this function

vortex-layout/src/segments/sink.rs
```diff
@@ -11,5 +12,5 @@ pub trait SegmentWriter {
     //  if we know we're going to read an entire FlatLayout together, then we should probably
     //  serialize it into a single segment that is 512 byte aligned? Or else, we should guarantee
     //  to align the the first segment to 512, and then assume that coalescing captures the rest.
-    fn put(&mut self, buffer: &[ByteBuffer]) -> SegmentId;
+    fn put(&self, segment_id: SegmentId, buffer: Vec<ByteBuffer>) -> VortexResult<()>;
```
gatesn: Why does this take a segment_id now?

vortex-layout/src/layouts/chunked/writer.rs
```diff
@@ -28,87 +28,43 @@ impl Default for ChunkedLayoutStrategy {
...
-        if children.len() == 1 {
-            return Ok(children.pop().vortex_expect("child layout"));
-        }
+    fn write_stream(
+        &self,
+        ctx: &ArrayContext,
+        dtype: &DType,
+        segment_writer: Arc<dyn SegmentWriter>,
+        mut stream: SequentialArrayStream,
+    ) -> Pin<Box<dyn LayoutWriter>> {
```
gatesn: See `SendableArrayStream`, I think you should create type alias for `SendableLayoutWriter`

vortex-layout/src/segments/sink.rs
onursatici: so it can still be not async and not block. Current implementation enforces `put` has been called with sequential segment ids but we could have one that buffers until complete segmentId runs are collected and flush those, or an implementation that doesn't need flushing and writes these segments to a KV store

encodings/dict/src/builders/mod.rs
onursatici: nice catch this is an artifact from my previous design, reverting

vortex-layout/src/strategy.rs
onursatici: yea we either need each chunk in every stream to be associated with a segment writer, or a sequence id. We need this to be able to globally compare each chunk with all others.
And if we are going to split the segment writer down to every single chunk, then it becomes a bit weird that those are still called segment writers, because they can just write one chunk really when they reach a leaf writer. So I did extract the ordering stuff from the segment writer and pass around chunks associated with sequence ids

onursatici:
Follow ups:
- move dict writer above coalescing, I want to see the effect of this on benchmarks because it will compress a larger initial chunk to see if dict encoding the rest is viable. Trade off here is better but slower compression.
- stats layout to precompute stats in parallel. Similar to above, I want to see the isolated effect on this
- dict layout to have pipeline parallelism with its children. This is a clear perf improvement but I still have some work to make it work on single threaded compression, the way I did this relies on spawned futures to drive themselves.

vortex-layout/src/strategy.rs
```diff
@@ -3,29 +3,98 @@
...
 use vortex_error::VortexResult;
 
-use crate::layouts::flat::writer::FlatLayoutWriter;
-use crate::layouts::struct_::writer::StructLayoutWriter;
-use crate::writer::{LayoutWriter, LayoutWriterExt};
+use crate::SendableLayoutWriter;
+use crate::segments::SequenceWriter;
+use crate::sequence::SequenceId;
+
+pub trait SequentialStream: Stream<Item = VortexResult<(SequenceId, ArrayRef)>> {
```
gatesn, line 19: I don't know where you make use of a SequentialStream that _isn't_ Send, therefore could just make the typeddef = Pin<Box<...>>.

vortex-layout/src/writer.rs
```diff
@@ -1,60 +1,17 @@
...
-        self.flush(segment_writer)?;
-        self.finish(segment_writer)
-    }
-}
+// Allow async blocks to impl LayoutWriter, this would change if more methods
+// are added to LayoutWriter.
+impl<F: Future<Output = VortexResult<LayoutRef>> + ?Sized + Send> LayoutWriter for F {}
 
-impl<L: LayoutWriter + ?Sized> LayoutWriterExt for L {}
+pub type SendableLayoutWriter = Pin<Box<dyn LayoutWriter + Send>>;
```
gatesn, line 17: Same here, can this just be `pub type LayoutWriter = Pin<Box<dyn Future....>>`?

vortex-file/src/segments/writer.rs
gatesn: This is still intended for public use right? Therefore should have docstrings.

vortex-array/src/context.rs
```diff
@@ -50,7 +50,7 @@ impl ArrayRegistry {
 #[derive(Debug, Clone)]
 pub struct VTableContext<T>(Arc<RwLock<Vec<T>>>);
 
-impl<T: Clone + Eq> VTableContext<T> {
+impl<T: Clone + Eq + Display> VTableContext<T> {
```
robert3005, line 53: I think this is a leftover? Not sure if anything here is using display

vortex-file/src/segments/writer.rs
robert3005: We need docs here

vortex-file/src/writer.rs
```diff
@@ -59,64 +65,47 @@ impl VortexWriteOptions {
...
-            self.strategy.new_writer(&ctx, stream.dtype())?,
-            stream.dtype(),
+        let dtype = stream.dtype().clone();
+        let (segment_writer, flusher) = SerialSegmentWriter::create();
+        let sequence_writer = SequenceWriter::new(Box::new(segment_writer));
+
+        let stream = stream.filter(|item| match item {
+            Ok(chunk) => future::ready(!chunk.is_empty()),
+            Err(_) => future::ready(true),
+        });
```
robert3005, line 83: use `try_filter` here

vortex-file/src/writer.rs
```diff
@@ -59,64 +65,47 @@ impl VortexWriteOptions {
...
-        segment_writer
-            .flush_async(&mut write, &mut segment_specs)
-            .await?;
-
-        // Finish the layouts and flush the finishing messages into the file
-        let layout = layout_writer.finish(&mut segment_writer)?;
-        segment_writer
-            .flush_async(&mut write, &mut segment_specs)
-            .await?;
+        let io_fut = async { flusher.flush(write).await };
```
robert3005, line 98: no need for the async block if you're just awaiting

vortex-layout/src/scan/executor.rs
```diff
@@ -66,3 +66,14 @@ impl TaskExecutor for tokio::runtime::Handle {
             .boxed()
     }
 }
+
+pub struct LocalExecutor {}
```
robert3005, line 70:
```suggestion
pub struct LocalExecutor;
```

vortex-file/src/writer.rs
```diff
@@ -27,7 +33,7 @@ pub struct VortexWriteOptions {
 impl Default for VortexWriteOptions {
     fn default() -> Self {
         Self {
-            strategy: Box::new(VortexLayoutStrategy),
+            strategy: VortexLayoutStrategy::with_executor(Arc::new(LocalExecutor {})),
```
robert3005, line 36:
```suggestion
            strategy: VortexLayoutStrategy::with_executor(Arc::new(LocalExecutor)),
```

vortex-file/src/strategy.rs
```diff
@@ -211,95 +153,61 @@ struct BufferedStrategy {
...
-    fn new_writer(&self, ctx: &ArrayContext, dtype: &DType) -> VortexResult<Box<dyn LayoutWriter>> {
-        let child = self.child.new_writer(ctx, dtype)?;
-        Ok(BufferedWriter {
-            chunks: Default::default(),
-            nbytes: 0,
-            buffer_size: self.buffer_size,
-            child,
-        }
-        .boxed())
+impl BufferedStrategy {
```
robert3005, line 156: if we are here already let's move this to a separate file

vortex-layout/src/layouts/file_stats.rs
```diff
@@ -1,32 +1,47 @@
...
 use vortex_error::{VortexExpect, VortexResult};
 
 use crate::layouts::zoned::zone_map::StatsAccumulator;
-use crate::segments::SegmentWriter;
-use crate::{LayoutRef, LayoutWriter};
+use crate::sequence::SequenceId;
+use crate::{SendableSequentialStream, SequentialStreamAdapter, SequentialStreamExt};
 
-/// A layout writer that computes aggregate statistics for all fields.
+pub fn accumulate_stats(
```
robert3005, line 16: Is this function still used?

vortex-layout/src/segments/sink.rs
```diff
@@ -11,5 +20,55 @@ pub trait SegmentWriter {
...
+    pub fn new_sequential(&self, stream: SendableArrayStream) -> SendableSequentialStream {
+        let sequence_pointer = self.tail();
+        SequentialStreamAdapter::new(
+            stream.dtype().clone(),
+            stream.scan(sequence_pointer, |pointer, item| match item {
+                Ok(chunk) => future::ready(Some(Ok((pointer.advance(), chunk)))),
+                Err(e) => future::ready(Some(Err(e))),
+            }),
+        )
+        .sendable()
```
robert3005, line 83:
```suggestion
        let mut sequence_pointer = self.tail();
        SequentialStreamAdapter::new(
            stream.dtype().clone(),
            stream.map_ok(move |chunk| (sequence_pointer.advance(), chunk)),
        ).sendable()
```

vortex-layout/src/layouts/struct_/writer.rs
```diff
@@ -1,197 +1,272 @@
...
+    T: Unpin,
+{
+    type Item = VortexResult<T>;
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut guard = self.state.lock();
+        if let Some(item) = guard.buffers[self.index].pop_front() {
+            return Poll::Ready(Some(item));
+        }
+
+        // support non fused streams
```
robert3005, line 152: this comment is a bit confusing. You support non fused stream and this transpose acts as a fuse?

vortex-layout/src/layouts/struct_/writer.rs
```diff
@@ -1,197 +1,272 @@
...
+        stream: SendableSequentialStream,
+    ) -> SendableLayoutWriter {
+        let dtype = stream.dtype().clone();
+        let Some(struct_dtype) = stream.dtype().as_struct().cloned() else {
+            // nothing we can do if dtype is not struct
+            return self.child.write_stream(ctx, sequence_writer, stream);
+        };
+        if HashSet::from_iter(struct_dtype.names().iter()).len() != struct_dtype.names().len() {
+            return Box::pin(async {
+                Err(vortex_err!("StructLayout must have unique field names"))
```
robert3005, line 47:
```suggestion
                vortex_bail!("StructLayout must have unique field names")
```

vortex-layout/src/layouts/zoned/writer.rs
```diff
@@ -32,110 +39,91 @@ impl Default for ZonedLayoutOptions {
...
-        self.data_writer.push_chunk(segment_writer, chunk)
-    }
+impl LayoutStrategy for ZonedStrategy {
+    fn write_stream(
+        &self,
+        ctx: &ArrayContext,
+        sequence_writer: SequenceWriter,
+        stream: SendableSequentialStream,
+    ) -> SendableLayoutWriter {
+        let present_stats: Arc<[Stat]> = self.options.stats.iter().sorted().copied().collect();
```
robert3005, line 69: Something has gone terribly wrong with this line. We don't need it, the replace line 72 with `&self.options.stats`

vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -31,98 +43,380 @@ impl Default for DictLayoutOptions {
...
-struct DelegatingDictLayoutWriter {
-    ctx: ArrayContext,
-    strategy: DictStrategy,
-    dtype: DType,
-    writer: Option<Box<dyn LayoutWriter>>,
+fn dict_encode_stream(
+    input: SendableSequentialStream,
+    constraints: DictConstraints,
+) -> DictionaryStream {
+    Box::pin(try_stream! {
```
robert3005, line 159: This macro is funky, took me a second to understand what yield meant, not sure if there's a simpler way of defining it though

gatesn (approved):

vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -31,98 +43,380 @@ impl Default for DictLayoutOptions {
...
+impl Drop for DictEncodedRunStream {
+    fn drop(&mut self) {
+        if let Some(tx) = self.input_tx.take() {
+            let _ = tx.send(self.input.take());
         }
     }
 }
 
+type Sequenced = Pin<Box<dyn Stream<Item = VortexResult<(SequenceId, ArrayRef)>> + Send>>;
+async fn call_for_first_item<T>(
```
robert3005, line 385: I think there's a nicer way of doing this but ownership is quite annoying

vortex-layout/src/layouts/repartition.rs
```diff
@@ -37,124 +26,149 @@ pub struct RepartitionWriterOptions {
...
+                        let output_chunks = chunks.collect_exact_blocks()?;
+                        assert!(!output_chunks.is_empty());
+                        let chunked_array = to_canonical_chunked(output_chunks, &dtype_clone)?;
+                        if !chunked_array.is_empty() {
+                            yield (sequence_pointer.advance(), chunked_array)
+                        }
+                    }
+                }
+                if canonical_stream.as_mut().peek().await.is_none() {
+                    let to_flush = to_canonical_chunked(chunks.data.drain(..).collect(), &dtype_clone)?;
```
robert3005, line 85: I think you can do the check before canonicalizing

vortex-layout/src/layouts/dict/writer/mod.rs
robert3005: I think there's a way to use peek, however, all it does is mostly hide this logic behind a library

robert3005 (commented):
Left couple mostly style comments about streams

vortex-layout/src/layouts/file_stats.rs
onursatici: yeap here: https://github.com/vortex-data/vortex/blob/03260ee44389c39395bea2835c7ed4e40ab6c018/vortex-file/src/writer.rs#L88

vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -31,98 +43,380 @@ impl Default for DictLayoutOptions {
...
-                    .boxed()
-                };
-                writer.push_chunk(segment_writer, chunk)?;
-                self.writer = Some(writer);
-                Ok(())
+    ) -> Vec<VortexResult<DictionaryChunk>> {
+        match self.try_encode(labeler, chunk) {
+            Ok(chunks) => chunks,
+            Err(e) => vec![Err(e)],
+        }
```
robert3005, line 209: this is just `unwrap_or_else`

vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -31,98 +43,380 @@ impl Default for DictLayoutOptions {
...
+        };
+        let (input_tx, input_rx) = oneshot::channel();
+        self.input = Some(input_rx);
+
+        let (values_tx, values_rx) = oneshot::channel();
+        let values_future = async {
+            match values_rx.await {
+                Ok(values) => values,
+                Err(_) => Err(vortex_err!("sender dropped")),
+            }
```
robert3005, line 310: use `unwrap_or_else`

vortex-layout/src/layouts/zoned/writer.rs
onursatici: ah good catch, yea this is a leftover from an experiment I did with precomputing stats in parallel but failed to cleanup

vortex-layout/src/layouts/dict/writer/mod.rs
onursatici: yea I would like to get rid of it also because it effectively disables rustfmt on the block it contains

vortex-layout/src/layouts/dict/writer/mod.rs
onursatici: yea indeed peekable wasn't quite ergonomic here but I didn't try hard enough

robert3005: I tried using peek, it's not any better. At least this is explicit

PR #3367, opened 2025-05-21
fix: Decimal ToArrow conversion

a10y:
Fixup several lingering Decimal issues:

- DecimalBuilder<T> -> DecimalBuilder. This allows us to create a DecimalBuilder and append any `NativeDecimalType` value into it, with the value being properly cast to the decimal type
- jni: fix `JNIArray.exportToArrow` to support Decimal128/256
- error message improvements


joseph-isaacs (approved):

vortex-scalar/src/decimal.rs
```diff
@@ -207,7 +207,7 @@ macro_rules! decimal_scalar_unpack {
                 Ok(match value.value {
                     None => None,
                     Some(DecimalValue::$arm(v)) => Some(v),
-                    _ => vortex_bail!("Cannot extract decimal as "),
+                    v => vortex_bail!("Cannot extract decimal {:?} as {}", v, stringify!($ty)),
```
robert3005, line 225:
```suggestion
                    v => vortex_bail!("Cannot extract decimal {v:?} as {}", stringify!($ty)),
```
same for below

robert3005: Can you try putting `stringify!($ty)` inline as well?

robert3005 (approved):

vortex-array/src/builders/mod.rs
```diff
@@ -208,15 +208,17 @@ pub trait ArrayBuilderExt: ArrayBuilder {
...
+                let self_dtype = self.dtype().clone();
                 macro_rules! append_decimal {
                     ($self:expr, $ty:ty, $value:expr) => {{
                         $self
                             .as_any_mut()
                             .downcast_mut::<DecimalBuilder<$ty>>()
                             .ok_or_else(|| {
                                 ::vortex_error::vortex_err!(
-                                    "Cannot append decimal scalar of type {} to builder of another type",
+                                    "Cannot append decimal scalar of type {} to builder of type {}",
```
robert3005: Inline self_dtype here

a10y: I'm going to remove it, I'm using this branch to iterate on the Azure Iceberg thing. It turns out there are quite a few broken things with Decimals :P

a10y: Converted the branch back to draft to clarify the intention, once I finish fixing the latent issues will promote to normal PR again

a10y:
This PR has morphed into a general fix-up of Decimal array handling as part of a push to get working round-trip of TPC-DS data from Parquet -> Arrow -> Vortex.

Some FLUPs to accomplish

- [x] Make ToArrowCanonical avoid re-casting the decimal values type, instead should canonicalize directly into an i128/i256 builder



----

Capturing some Slack commentary

image

image

image

encodings/decimal-byte-parts/src/decimal_byte_parts/mod.rs
```diff
@@ -82,27 +80,6 @@ impl DecimalBytePartsArray {
...
-            vortex_bail!(
-                "cannot represent a decimal {decimal_dtype} as primitive parts {:?}, decimal bit width {}, primitive bit width {}",
-                iter::once(&msp)
-                    .chain(&lower_parts)
-                    .map(|a| a.dtype())
-                    .collect_vec(),
-                decimal_dtype.required_bit_width(),
-                primitive_bit_width
-            )
-        }
```
a10y, line 104: This causes errors when you're just holding values of a narrowed integer type

vortex-array/src/arrow/compute/to_arrow/canonical.rs
```diff
@@ -398,4 +426,54 @@ mod tests {
...
+            .as_any()
+            .downcast_ref::<Decimal128Array>()
+            .unwrap();
+        assert_eq!(arrow_decimal.value(0), 10);
+        assert_eq!(arrow_decimal.value(1), 11);
+        assert_eq!(arrow_decimal.value(2), 12);
+    }
+
+    #[rstest]
+    #[case(0i8)]
```
a10y, line 467: rstest doesn't let you do the divan-style type parameter spec, so best we can do is supply a dummy value that propagates the T

vortex-array/src/arrow/convert.rs
```diff
@@ -98,24 +98,24 @@ impl_from_arrow_primitive!(Float32Type);
...
-    fn from_arrow(array: &ArrowPrimitiveArray<Decimal256Type>, _nullable: bool) -> Self {
+    fn from_arrow(array: &ArrowPrimitiveArray<Decimal256Type>, nullable: bool) -> Self {
         let decimal_type = DecimalDType::new(array.precision(), array.scale());
         let buffer = Buffer::from_arrow_scalar_buffer(array.values().clone());
         // SAFETY: Our i256 implementation has the same bit-pattern representation of the
         //  arrow_buffer::i256 type. It is safe to treat values held inside the buffer as values
         //  of either type.
         let buffer =
             unsafe { std::mem::transmute::<Buffer<arrow_buffer::i256>, Buffer<i256>>(buffer) };
-        let validity = nulls(array.nulls(), false);
```
a10y, line 118: this was a bug

vortex-array/src/arrays/decimal/mod.rs
```diff
@@ -44,8 +44,8 @@ impl VTable for DecimalVTable {
 #[derive(Clone, Debug)]
 pub struct DecimalEncoding;
 
-/// Maps a decimal precision into the small type that can represent it.
-pub fn precision_to_storage_size(decimal_dtype: &DecimalDType) -> DecimalValueType {
+/// Maps a decimal precision into the smallest type that can represent it.
+pub fn smallest_storage_type(decimal_dtype: &DecimalDType) -> DecimalValueType {
```
gatesn, line 48: Any reason this isn't a function on DecimalDType?

vortex-array/src/arrow/compute/to_arrow/canonical.rs
```diff
@@ -48,6 +48,22 @@ impl Kernel for ToArrowCanonical {
...
             .unwrap_or_else(|| array.dtype().to_arrow_dtype())?;
 
+        if matches!(arrow_type, DataType::Decimal128(..)) {
+            let arrow_array = to_arrow_decimal128(array.to_decimal()?)?;
+            return Ok(Some(
+                ArrowArray::new(arrow_array, array.dtype().nullability())
+                    .into_array()
+                    .into(),
+            ));
+        } else if matches!(arrow_type, DataType::Decimal256(..)) {
```
gatesn: nit: prefer not to have `else if` when the previous branch returns

vortex-array/src/arrow/compute/to_arrow/canonical.rs
```diff
@@ -48,6 +48,22 @@ impl Kernel for ToArrowCanonical {
             .map(Ok)
             .unwrap_or_else(|| array.dtype().to_arrow_dtype())?;
 
+        if matches!(arrow_type, DataType::Decimal128(..)) {
+            let arrow_array = to_arrow_decimal128(array.to_decimal()?)?;
```
gatesn: I don't understand why this is a special case? You're still calling `array.to_decimal()` which effectively canonicalizes no?

vortex-array/src/arrow/compute/to_arrow/canonical.rs
```diff
@@ -206,10 +225,15 @@ fn to_arrow_decimal256(array: DecimalArray) -> VortexResult<ArrowArrayRef> {
...
-        DecimalValueType::I32 => array.buffer::<i8>().into_iter().map(|x| x.as_()).collect(),
-        DecimalValueType::I64 => array.buffer::<i8>().into_iter().map(|x| x.as_()).collect(),
-        DecimalValueType::I128 => array.buffer::<i8>().into_iter().map(|x| x.as_()).collect(),
+        DecimalValueType::I16 => array.buffer::<i16>().into_iter().map(|x| x.as_()).collect(),
+        DecimalValueType::I32 => array.buffer::<i32>().into_iter().map(|x| x.as_()).collect(),
+        DecimalValueType::I64 => array.buffer::<i64>().into_iter().map(|x| x.as_()).collect(),
+        DecimalValueType::I128 => array
+            .buffer::<i128>()
+            .into_iter()
+            // SAFETY: Vortex i256 is just a newtype wrapper around Arrow i256
```
gatesn: I thought you implemented `From` trait?

vortex-array/src/builders/decimal.rs
```diff
@@ -149,8 +319,16 @@ impl<T: NativeDecimalType> ArrayBuilder for DecimalBuilder<T> {
             );
         }
 
-        self.values
-            .extend_from_slice(array.buffer::<T>().as_slice());
+        match array.values_type() {
+            DecimalValueType::I8 => self.values.extend_from_buffer(array.buffer::<i8>()),
```
gatesn: I think there's a match_each macro for this?

vortex-array/src/builders/decimal.rs
```diff
@@ -1,37 +1,196 @@
...
 use crate::builders::ArrayBuilder;
 use crate::builders::lazy_validity_builder::LazyNullBufferBuilder;
 use crate::validity::Validity;
 use crate::{Array, ArrayRef, IntoArray, ToCanonical};
 
-pub struct DecimalBuilder<T> {
-    values: BufferMut<T>,
+#[derive(Debug)]
+enum InnerDecimalBuilder {
+    I8(BufferMut<i8>),
```
gatesn, line 20: Is it worse to just hold ByteBufferMut and a DecimalValueType, switching each time you need to downcast the ByteBuffer?

vortex-scalar/src/bigint.rs
```diff
@@ -151,20 +159,58 @@ impl CheckedSub for i256 {
...
+    ($typ:ty) => {
+        impl AsPrimitive<$crate::i256> for $typ {
+            fn as_(self) -> $crate::i256 {
+                $crate::i256::from_i128(self.as_())
+            }
+        }
+
+        impl AsPrimitive<$typ> for $crate::i256 {
+            fn as_(self) -> $typ {
+                self.maybe_i128()
```
gatesn: Seems weird to call maybe_i128 here?

vortex-scalar/src/bigint.rs
```diff
@@ -151,20 +159,58 @@ impl CheckedSub for i256 {
...
+        self
+    }
+}
+
+// All signed types convert up to i256
+macro_rules! impl_as_i256 {
+    ($typ:ty) => {
+        impl AsPrimitive<$crate::i256> for $typ {
+            fn as_(self) -> $crate::i256 {
+                $crate::i256::from_i128(self.as_())
```
gatesn: Why convert via i128?

vortex-array/src/builders/decimal.rs
```diff
@@ -1,37 +1,196 @@
...
+
+pub struct DecimalBuilder {
+    values: InnerDecimalBuilder,
     nulls: LazyNullBufferBuilder,
     dtype: DType,
 }
 
 const DEFAULT_BUILDER_CAPACITY: usize = 1024;
 
-impl<T: NativeDecimalType> DecimalBuilder<T> {
```
gatesn, line 23: I think I'm missing the reasoning behind removing this generic?

gatesn (changes requested):
I think I'm missing the reasoning behind removing the generics from the builder? Other stuff all looks good though

encodings/decimal-byte-parts/src/decimal_byte_parts/mod.rs
joseph-isaacs: What validation will you add instead?

vortex-array/src/arrays/decimal/mod.rs
joseph-isaacs: It's an implementation detail

a10y:
Removing the type parameter was mostly about consistency, since our decimal type doesn't specify an exact PType for the buffer holding the values.

vortex-scalar/src/bigint.rs
a10y: All of the signed types upcast to i128 without loss of precision, and building an i256 from i128 is very cheap

gatesn:
But we still have to pick _one_ decimal type for the builder? And for exporting to Arrow, we could even create an i128 builder and then use append_to_builder if we wanted to avoid the copy in the middle.

Is the problem that the downcasting becomes hard? If so, we could always have `DecimalBuilder(TypedDecimalBuilder<D>)` such that you can downcast in two steps.

vortex-array/src/arrays/decimal/mod.rs
a10y: Yea, the way our decimal types work now the PType is a concern mostly for how a canonical DecimalArray holds the values rather than anything that's bound by the DecimalDType

a10y:
I think we can add the type parameter back. The important thing is that we preserve the ability to extend from a slice of some other T: NativeDecimalType.

I think we also should be using checked casting rather than AsPrimitive. This might actually be easier with the other i256 type from your PR...

gatesn:
Agree with all that, and yeah the other PR has an i256 that implements NumCast, so we get back an Option<T> to perform checked casts.

Do you need the other PR to make that change easier? I already do that "append anything to anything" thing here: https://github.com/vortex-data/vortex/pull/3373/files#diff-e3c7c38c80720dc1fd02f75e9f7d273b03ddb1fb0d7658b7e3f83b7f57c71893R25-R66

You can merge that PR into this one if it's easier.

a10y:
Following up from our slack convo, decided not to go with i256 due to questions around byte representation stability and FFI.

I've added new `ToPrimitive` and `BigCast` traits into our vortex-scalar here which gives us the ability to do checked casts between i256 and all other primitive types. I've also implemented

Also I ended up keeping the untyped `DecimalBuilder` but I did clean it up a bit with the help of a macro so it's less visually cluttersome

encodings/decimal-byte-parts/src/decimal_byte_parts/mod.rs
a10y: What sort of validation do you reckon we need here? This felt superfluous to me. Only thing I can think of is checking the opposite: none of the values exceeds the number of digits that can be represented with current precision. That's a full element-wise scan though, not sure if we want to check that

vortex-array/src/arrow/compute/to_arrow/canonical.rs
a10y: Yea this is wrong actually, I need to make this canonicalize into an i128 builder directly

vortex-array/src/builders/decimal.rs
```diff
@@ -1,61 +1,186 @@
...
+    };
+}
+impl_from_buffer!(i8, I8);
+impl_from_buffer!(i16, I16);
+impl_from_buffer!(i32, I32);
+impl_from_buffer!(i64, I64);
+impl_from_buffer!(i128, I128);
+impl_from_buffer!(i256, I256);
+
+macro_rules! delegate_fn {
```
a10y, line 50: i think this might be a better way to write a lot of our `match_each_XYZ` functions, because it's hygenic, which means that you get IDE support _inside_ of the closure body!

vortex-array/src/arrow/compute/to_arrow/canonical.rs
gatesn: Not sure it's a huge difference honestly. We can work on it if it does end up being important.

gatesn:
Hmm, I still think the typed builder would be better. The internal enum adds another level of dispatch to most of the builder functions.

a10y:
The issue I couldn't find an easy way around was how to implement `append_scalar` from ArrayBuilderExt generically, b/c you don't have enough information written down anywhere to infer the right `T` for `.downcast_mut::<DecimalBuilder<T>>`

FWIW I did at one point measure the difference between Box<dyn T> and enum dispatch overhead and the dynamic dispatch overhead is strictly worse, by a factor of at least 2-3x.

gatesn:
That's what I was saying about Dow casting first to DecimalBuilder, then you have a values type to downcast further to a TypedDecimalBuilder.

a10y:
Could you maybe share more what you're thinking? Is the idea being to get some long-lived handle to the typed thing to amortize the downcasting overhead? AFAICT we already pretty much always use it behind a Box<dyn ArrayBuilder> anyway so there is a cost for every method call. ChunkedArray canonicalize amortizes this cost a bit b/c it extends_from_array

gatesn:
Yah, played with it, it's just as annoying to hold a typed thing internally.

gatesn (approved):

vortex-array/src/builders/decimal.rs
a10y: c.f. #3401

PR #3369, opened 2025-05-21
clean: annotate rust c abi types with `#[allow(non_camel_case_types)]`

0ax1:


0ax1:
Unclear why ~~clippy~~ [rustc didn't catch this](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html?search=#non-camel-case-types), but I get a warning for those type names in context of Zed.

vortex-ffi/src/file.rs
```diff
@@ -35,6 +35,7 @@ pub struct vx_file_reader {
 
 /// Options supplied for opening a file.
 #[repr(C)]
```
joseph-isaacs, line 37: I bet this implies non camel case

0ax1: Let me double check whether this isn't an editor issue..

gatesn:
Can we figure out why clippy didn't catch this? Otherwise it'll keep getting out of sync

0ax1:
> Can we figure out why clippy didn't catch this? Otherwise it'll keep getting out of sync

Will have a look. But also realized that this is a rustc lint, not clippy: https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html?search=#non-camel-case-types

vortex-ffi/src/file.rs
0ax1: Oddly, it is mandatory on:
```
#[repr(C)]
#[allow(non_camel_case_types)]
pub enum vx_log_level {
    LOG_LEVEL_OFF = 0,
    LOG_LEVEL_ERROR = 1,
    LOG_LEVEL_WARN = 2,
    LOG_LEVEL_INFO = 3,
    LOG_LEVEL_DEBUG = 4,
    LOG_LEVEL_TRACE = 5,
}
```

0ax1: Which is though bc of the enum values not the type.

0ax1:
I assume Zed used the wrong rust-analyzer version, defaulting to the system wide installation. Turning that option off presumably leads to Zed using the rust-analyzer version matching what we define in the rust-toolchain.toml. The warnings don't show up in that case any more.

https://zed.dev/docs/languages/rust#binary

PR #3377, opened 2025-05-23
Rename pyvortex to vortex-python

gatesn:
This has been annoying me for a while...

The PyPi package is now called vortex-data, and this is largely a regular Rust crate. So let's stick with the same naming convention as the others (e.g. vortex-jni)


robert3005 (approved):
sure, have a rename. I always found pyvortex cuter but we can be consistent

PR #3381, opened 2025-05-27
Find latest benchmark result

gatesn:
Compare bench results to nearest common ancestor of the PR commit and develop

### DIFF
--- a/scripts/compare-benchmark-jsons.py
+++ b/scripts/compare-benchmark-jsons.py
@@ -5,45 +5,101 @@
 #   "tabulate",
 # ]
 # ///
-import sys
+import argparse
+import subprocess
 
 import pandas as pd
 
-base = pd.read_json(sys.argv[1], lines=True)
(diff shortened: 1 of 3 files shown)


scripts/compare-benchmark-jsons.py
```diff
@@ -5,45 +5,109 @@
...
+    )
+
+
+def main(args):
+    base = pd.read_json(args.base, lines=True)
+    pr = pd.read_json(args.results, lines=True)
+
+    pr_commit_id = set(pr["commit_id"].unique())
+    assert len(pr_commit_id) == 1, "PR results must have exactly one commit ID."
+    pr_commit_id = next(iter(pr_commit_id))
```
AdamGS, line 69: this looks weird but maybe I'm just not of a pythonista

gatesn: I just moved whatever was there before.. it's not crazy weird though

AdamGS (approved):

PR #3385, opened 2025-05-27
feat: add a sequence vector `a[i] = b + m * i`

joseph-isaacs:


robert3005:
How about we call the directory just `sequence`?

encodings/sequence/src/array.rs
```diff
@@ -0,0 +1,262 @@
...
+use vortex_mask::Mask;
+use vortex_scalar::{Scalar, ScalarValue};
+
+vtable!(Sequence);
+
+#[derive(Clone, Debug)]
+/// An array representing the equation `A[i] = base + i * multiplier`.
+pub struct SequenceArray {
+    base: ScalarValue,
+    multiplier: ScalarValue,
```
robert3005: You could put PValue here (slightly smaller than ScalarValue) since this array can only represent primitive types

encodings/sequence/src/array.rs
```diff
@@ -0,0 +1,262 @@
...
+        length: usize,
+    ) -> Self {
+        let dtype = DType::Primitive(ptype, NonNullable);
+        Self {
+            base,
+            multiplier,
+            dtype,
+            length,
+            // TODO(joe): add stats, on construct or on use?
+            stats_set: Default::default(),
```
robert3005, line 72: The general change @gatesn have made is that stats are always handled on access instead of frontloading. Which makes sense if the array is going to be shortlived (most are so far)

encodings/sequence/src/array.rs
```diff
@@ -0,0 +1,262 @@
...
+        array.stats_set.to_ref(array.as_ref())
+    }
+}
+
+impl CanonicalVTable<SequenceVTable> for SequenceVTable {
+    fn canonicalize(array: &SequenceArray) -> VortexResult<Canonical> {
+        let prim = match_each_native_ptype!(array.ptype(), |$P| {
+            let base = <$P>::try_from(&array.base)?;
+            let multi = <$P>::try_from(&array.multiplier)?;
+            PrimitiveArray::from_iter((0..array.len()).map(|i| base + <$P>::from_usize(i).vortex_expect("") * multi))
```
robert3005:
```suggestion
            PrimitiveArray::from_iter((0..array.len()).map(|i| base + <$P>::from_usize(i).vortex_expect("must fit") * multi))
```

vortex-dtype/src/ptype.rs
```diff
@@ -56,6 +56,7 @@ pub trait NativePType:
     + RefUnwindSafe
     + Num
     + NumCast
+    + Bounded
```
robert3005: where are we using this?

robert3005 (changes requested):
Looks good though I always like renames to be put in separate prs. Just need to change the base and multiplier type

encodings/sequence/src/array.rs
```diff
@@ -0,0 +1,292 @@
...
+        base: PValue,
+        multiplier: PValue,
+        ptype: PType,
+        length: usize,
+    ) -> VortexResult<PValue> {
+        match_each_integer_ptype!(ptype, |$P| {
+            let len_t = <$P>::from_usize(length-1)
+                .ok_or_else(|| vortex_err!("cannot convert length {} into {}", length, ptype))?;
+
+            let base = <$P>::try_from(base)?;
```
robert3005: there's a slightly nicer syntax you can use now with `let base = base.as_primitive::<$P>()?;`, same for multiplier. Here and on line 114-115 and 168-169

vortex-dtype/src/ptype.rs
joseph-isaacs: gone

robert3005 (approved):

robert3005:
I think you have to rebase due to changes in match_each macros

encodings/sequence/src/array.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        Ok(Canonical::Primitive(prim))
+    }
+}
+
+impl OperationsVTable<SequenceVTable> for SequenceVTable {
+    fn slice(array: &SequenceArray, start: usize, stop: usize) -> VortexResult<ArrayRef> {
+        Ok(SequenceArray::unchecked_new(
+            array.index_value(start)?,
+            array.multiplier.clone(),
```
robert3005: clippy yells at you here that the clone is redundant since PValue is Copy

PR #3386, opened 2025-05-27
duckdb: bump to `v1.3.0`

0ax1:


duckdb-vortex/src/vortex_scan.cpp
```diff
@@ -480,7 +483,11 @@ void RegisterScanFunction(DatabaseInstance &instance) {
 		auto global_state = make_uniq<ScanGlobalState>();
 
 		// TODO(joe): do this expansion gradually in the scan to avoid a slower start.
-		global_state->expanded_files = bind.file_list->GetAllFiles();
+		auto file_infos = bind.file_list->GetAllFiles();
+		global_state->expanded_files.reserve(file_infos.size());
+		for (const auto &file_info : file_infos) {
+			global_state->expanded_files.push_back(file_info.path);
+		}
```
joseph-isaacs, line 496: For me understanding, why do this?

.github/workflows/bench-pr.yml
```diff
@@ -1,7 +1,7 @@
 name: PR Benchmarks
 
 env:
-  DUCKDB_VERSION: v1.2.2
+  DUCKDB_VERSION: v1.3.0
```
joseph-isaacs, line 4: there are 3

joseph-isaacs:
LG, but I think you might have missed a few 1.2.2 strs

duckdb-vortex/src/vortex_scan.cpp
0ax1: We can't simply assign anymore as we need to access `.path`. Alternatively, we change the vector type for `expanded_files`.

.github/workflows/bench-pr.yml
0ax1: good catch!

joseph-isaacs (approved):

PR #3389, opened 2025-05-27
Use i256 crate

gatesn:


a10y (changes requested):
let's please not do this until after #3367

gatesn:
This crate isn't guaranteed to be byte-stable, and also fails to cast i256->i256... soooooo... not for us thank you very much

Issue #3382, opened 2025-05-27
Multi-file Layout Reader


gatesn:
It would be nice if we can construct a `LayoutReader` that encapsulates reading multiple files (similar to multi-file datasets in Arrow, etc.).

This may require some changes to the Layout / LayoutReader APIs, particularly around not know exact row counts up-front. We don't want to block on opening say ~100 files before we can finish planning and return the first row. This also allows us to internalize more of the concurrency inside a single scan.

Changes:

- [x] Remove `Deref<Target=dyn Layout>` from `LayoutReader` trait.
- [x] Return approximate row count from `LayoutReader` instead of exact row count.
- [ ] Do not eagerly construct splits for scanning. Need an alternate that allows for sensible streaming?
- [ ] Some way for multi-file layout reader to read multiple files in parallel, allowing each file to have affinity to a worker thread.

gatesn:
On reflection, I think we may instead want a `MultiReaderScan`.

PR #3400, opened 2025-05-28
Layout reader updates

gatesn:
See #3382


robert3005:
Which part is this? Remove the `Deref`?

gatesn:
This removes Deref and makes row_count approximate

robert3005 (approved):

PR #3401, opened 2025-05-28
chore: better match_each_XYZ macros

a10y:
Now with 100% more IDE support


robert3005 (approved):

a10y:
One of the nice things that falls out of this is that clippy will run on all of the stuff inside the macro calls

PR #3373, opened 2025-05-22
DuckDB Exporter

gatesn:
Instead of slicing arrays and converting them, we track the row offset and export from a single ArrayIterator

Also fixes #3405

PR #3405, opened 2025-05-29
fix[duckdb]: scan return correct number of rows

joseph-isaacs:


0ax1 (approved):

gatesn:
Can we punt this? I think I also fixed it in #3373

PR #3409, opened 2025-05-29
[experiment] Using Cuckoo runtime

AdamGS:


AdamGS:
the first batch of results here isn't valuable because I left a bunch of debug prints, there's another run starting now.

AdamGS:
Right now it seems like its significantly slower in many queries, I might spend some more time profiling the results here and see if there's any obvious improvements that I can find.

vortex-ffi/src/file.rs
```diff
@@ -307,7 +307,7 @@ pub unsafe extern "C-unwind" fn vx_file_reader_scan(
         // This means any thread that calls into the iterator will use the same runtime to make
         // progress on _all_ the spawned tasks (across all scans!)
         let scan_builder = scan_builder
-            .with_concurrency(100)
+            .with_concurrency(10000)
```
a10y, line 310: science has gone too far this time

PR #3410, opened 2025-05-29
Adds `list_contains` expression

delta003:
And fix bugs in compute::list_contains


vortex-array/src/compute/list.rs
```diff
@@ -77,8 +77,12 @@ pub fn list_contains(array: &dyn Array, value: Scalar) -> VortexResult<ArrayRef>
     // Fast path: no elements match.
     if let Some(pred) = matches.as_constant() {
         if matches!(pred.as_bool().value(), None | Some(false)) {
-            // TODO(aduffy): how do we handle null?
```
delta003, line 80: @a10y did you have some concerns here with this approach?

a10y: No i think you're right, this looks like it was a bug.

a10y: Actually, I'm not sure this quite fixes it? The constant array is still a constantarray of `false?` but it doesn't carry any validity buffer. Do we need to build a sparsearray with null fill value?

delta003: Haha that's another bug. I was fixing type, didn't realize this handles None. How about a constant null array?

delta003: Just to validate my understanding: None here means that matches is constant array of nulls, which must mean that the given scalar (rhs) is null. I'm going to through in this function when scalar is null. There is another function that should be used.

delta003: Or all elements of the list are null, but for some reason that doesn't hit this branch. I think my brain is too tired for this now, I'll pick it up next week.

gatesn (dismissed):
Hmmm, `is_in` != `list_contains`

`list_contains([[0, 1], [1, 2]], 2) -> [false, true]`
`is_in([0, 0, 4, 6], [[0, 1], [2, 3], [4, 5], [6, 7]]) -> [true, false, true, true]`

Although same as list_contains, we would typically only see a ConstantArray on the RHS of an is_in expression.

See https://duckdb.org/docs/stable/sql/expressions/in.html

(This is just a naming concern)

robert3005:
Can you elaborate how the is_in case is evaluated? I struggle to figure out how you arrived at the expected answer, the dimensions seem wrong? And yes is_in is ‘forall items: list_contains(<list>, item)’

vortex-array/src/compute/list.rs
delta003: Fixed this, I've added some more tests which uncovered more bugs, so this resulted in a bugger change.

delta003:
Changed to `list_contains`, but found more bugs in compute implementation so fixed those in the same PR

a10y (approved):
LGTM!

a10y:
Merged to resolve conflict with Cast expr

PR #3412, opened 2025-05-29
fix: Decimal append_scalar_value

a10y:
append_scalar_value should not match on the smallest_storage_type, but rather just unwrap the DecimalScalar enum and handle it correctly.

### DIFF
--- a/vortex-scalar/src/decimal.rs
+++ b/vortex-scalar/src/decimal.rs
@@ -316,17 +316,34 @@ decimal_scalar_pack!(i256, i256, I256);
 
 #[macro_export]
 macro_rules! match_each_decimal_value {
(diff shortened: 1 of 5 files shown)


robert3005 (approved):

vortex-scalar/src/decimal.rs
```diff
@@ -316,17 +316,34 @@ decimal_scalar_pack!(i256, i256, I256);
 
 #[macro_export]
 macro_rules! match_each_decimal_value {
-    ($self:expr, | $_:tt $value:ident | $($body:tt)*) => ({
-        macro_rules! __with__ {( $_ $value:ident ) => ( $($body)* )}
+    ($self:expr, | $value:ident | $body:block) => {{
```
a10y, line 319: straggler from #3401

PR #3413, opened 2025-05-29
fix: DecimalValue comparison should be ptype-blind

a10y:
Fixes issue found in Fuzzer run: https://github.com/vortex-data/vortex/actions/runs/15313610267/job/43083230856#step:7:762

DecimalValue should compare the same way that PValue does, by upcasting to the largest signed type that can hold any values and doing comparison in the upcasted space.

### DIFF
--- a/vortex-scalar/src/decimal.rs
+++ b/vortex-scalar/src/decimal.rs
@@ -3,10 +3,77 @@ use std::fmt;
 use std::fmt::{Display, Formatter};
 
 use vortex_dtype::{DType, DecimalDType, Nullability};
-use vortex_error::{VortexError, VortexResult, vortex_bail};
+use vortex_error::{VortexError, VortexExpect, VortexResult, vortex_bail};
 
 use crate::scalar_value::InnerScalarValue;
-use crate::{BigCast, Scalar, ScalarValue, i256};
+use crate::{BigCast, Scalar, ScalarValue, ToPrimitive, i256};
+
+#[macro_export]
+macro_rules! match_each_decimal_value {
+    ($self:expr, | $value:ident | $body:block) => {{
+        match $self {
+            DecimalValue::I8(v) => {
+                let $value = v;
+                $body
+            }
+            DecimalValue::I16(v) => {
+                let $value = v;
+                $body
+            }
+            DecimalValue::I32(v) => {
+                let $value = v;
+                $body
+            }
+            DecimalValue::I64(v) => {
+                let $value = v;
+                $body
+            }
+            DecimalValue::I128(v) => {
+                let $value = v;
+                $body
+            }
+            DecimalValue::I256(v) => {
+                let $value = v;
+                $body
+            }
+        }
+    }};
+}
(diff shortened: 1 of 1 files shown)


a10y:
Clean fuzzer run after this change
image

vortex-scalar/src/decimal.rs
```diff
@@ -3,10 +3,77 @@ use std::fmt;
 use std::fmt::{Display, Formatter};
 
 use vortex_dtype::{DType, DecimalDType, Nullability};
-use vortex_error::{VortexError, VortexResult, vortex_bail};
+use vortex_error::{VortexError, VortexExpect, VortexResult, vortex_bail};
 
 use crate::scalar_value::InnerScalarValue;
-use crate::{BigCast, Scalar, ScalarValue, i256};
+use crate::{BigCast, Scalar, ScalarValue, ToPrimitive, i256};
```
a10y, line 10: oh this looks like cruft, let me rebase

a10y: oh nvm, i moved this to top of the file so i could use it in the impls

vortex-scalar/src/decimal.rs
```diff
@@ -31,6 +98,42 @@ pub enum DecimalValue {
...
+        self_upcast == other_upcast
+    }
+}
+
+impl Eq for DecimalValue {}
+
+impl PartialOrd for DecimalValue {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        let self_upcast = match_each_decimal_value!(self, |v| {
+            v.to_i256()
```
a10y, line 126: every signed type must losslessly upcast to i256

robert3005 (approved):

vortex-scalar/src/decimal.rs
```diff
@@ -314,69 +417,25 @@ decimal_scalar_pack!(u64, i128, I128);
...
-            }
-            DecimalValueType::I256 => {
-                type $enc = i256;
-                $body
-            }
-            ty => unreachable!("unknown decimal value type {:?}", ty),
-        }
-    }};
+    use crate::{DecimalValue, i256};
+
```
a10y, line 440: i tried to be somewhat creative with these

robert3005: don't worry, the fuzzer will find the rest of the examples

vortex-scalar/src/decimal.rs
```diff
@@ -418,7 +430,10 @@ decimal_scalar_pack!(i128, i128, I128);
 decimal_scalar_pack!(i256, i256, I256);
 
 #[cfg(test)]
+#[allow(clippy::disallowed_types)]
```
robert3005, line 433: just use the aliases, they're only a reexport

a10y: they're only exported from the vortex-array crate though, which depends on vortex-scalar, so we can't import here

PR #3414, opened 2025-05-29
bench global queue

0ax1:
bench run only


0ax1:
ddb vortex clickbench benefits from this with speedups up to 40% for: q36-q42

Closing the pr, as we move the scan logic to Rust.

PR #3422, opened 2025-05-30
feat & fix: Scalar::zero & from_arrow handles nullable struct fields

danking:
Consider this nullable struct type with a non-nullable field.

```
{
    non_nullable_inner=i32
}?
```

In Arrow, a nullable field *does not mean* the corresponding array has no nulls. Instead, it means that the array has nulls at only those positions that some containing struct array has a null.

For example, this is a valid arrow array of that type:

```
StructArray
-- validity:
[
  null,
]
[
-- child 0: "non_nullable_inner" (Int32)
PrimitiveArray<Int32>
[
  null,
]
]
```

In this PR, we use `fill_null`, with the newly implemented `Scalar::zero`, to convert a nullable Arrow field array into a valid non-nullable Vortex array. If a function like `fill_null_arbitrary` existed, some arrays could simply drop their validity mask without modifying their data.

### DIFF
--- a/vortex-array/src/arrow/convert.rs
+++ b/vortex-array/src/arrow/convert.rs
@@ -19,13 +19,14 @@ use vortex_buffer::{Alignment, Buffer, ByteBuffer};
 use vortex_dtype::datetime::TimeUnit;
 use vortex_dtype::{DType, DecimalDType, NativePType, PType};
 use vortex_error::{VortexExpect as _, vortex_panic};
-use vortex_scalar::i256;
+use vortex_scalar::{Scalar, i256};
 
 use crate::arrays::{
     BoolArray, DecimalArray, ListArray, NullArray, PrimitiveArray, StructArray, TemporalArray,
     VarBinArray, VarBinViewArray,
 };
 use crate::arrow::FromArrowArray;
+use crate::compute::fill_null;
 use crate::validity::Validity;
 use crate::{ArrayRef, IntoArray};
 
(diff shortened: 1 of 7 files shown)


vortex-array/src/arrow/convert.rs
```diff
@@ -226,17 +227,31 @@ impl FromArrowArray<&ArrowBooleanArray> for ArrayRef {
...
+                .map(|s| (*s).into())
+                .collect(),
+            struct_array
                 .columns()
                 .iter()
-                .zip(value.fields())
-                .map(|(c, field)| Self::from_arrow(c.as_ref(), field.is_nullable()))
+                .zip(struct_array.fields())
+                .map(|(c, field)| {
+                    let vtx = Self::from_arrow(c.as_ref(), true);
```
gatesn, line 242: I'm not sure I quite understand what you're doing here. 

I think our definition of struct validity should be the same as Arrow's:
https://arrow.apache.org/docs/format/Columnar.html#struct-validity

In which case, I think the code from before is correct? I might misunderstand though, happy to jump on a call.

vortex-scalar/src/scalar_value/mod.rs
```diff
@@ -258,6 +269,35 @@ impl InnerScalarValue {
...
+    /// Produces a "zero" value of the given type.
+    ///
+    /// Even recursive values (i.e. structs) transitively will not contain any nulls.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if given the nullable data type.
+    pub(crate) fn zero(dtype: &DType) -> Self {
+        match &dtype {
+            DType::Null => vortex_panic!("the null data type has no 'zero' value"),
```
gatesn: Surely this is just Self::Null?

gatesn (changes requested):

vortex-array/src/arrow/convert.rs
a10y: > This is illustrated in the example above, one of the child arrays has a valid entry 'alice' for the null struct but it is “hidden” by the struct array’s validity bitmap. However, when treated independently, corresponding entries of the children array will be non-null.

This seems pretty conclusive yea

danking: Develop panics on the test I added. In particular, it tries to construct a non-nullable PrimitiveArray from an Arrow Int32Array containing a null [1].

How can we preserve the Arrow struct array type from the PR description when we create a Vortex struct array? In Arrow, the non-nullability of a field is stored in the `Field` of the struct array. In Vortex, AFAIK, the nullability of a field array must match the nullability of the field's type. I thought a nullable PrimitiveArray which contains nulls is not of type `DType::Primitive(..., NonNullable)`. Is that wrong?

---

[1] [This assertion on line 284 of arrow/convert.rs](https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/arrow/convert.rs#L284) fails when we try to construct the validity of the field's primitive array [on line 81 of arrow/convert.rs](https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/arrow/convert.rs#L81).

gatesn: I can't quite tell if you agree with me or not? 

If our semantics are the same as arrow, then it's fair to say that we can just recursively convert the arrow arrays zero-copy (i.e. not running fill_null) and the null bitmaps will all be correct in the same places they were in the arrow array.

vortex-array/src/arrow/convert.rs
```diff
@@ -370,3 +385,35 @@ impl FromArrowArray<&dyn ArrowArray> for ArrayRef {
...
+
+    use crate::ArrayRef;
+    use crate::arrow::FromArrowArray as _;
+
+    #[test]
+    pub fn nullable_may_contain_non_nullable() {
+        let fields = Fields::from(vec![Field::new(
+            "non_nullable_inner",
+            DataType::Int32,
+            false,
```
gatesn, line 405: I don't see why you're allowed to mark a field as non-nullable when it contains nulls?

vortex-scalar/src/scalar_value/mod.rs
danking: Done.

PR #3423, opened 2025-05-30
fix: Handle comparing VarBinArray to other non VarBin Utf8/BinaryDType arrays

robert3005:


gatesn:
Is there a test you can craft that hits this? So we don't regress _again_

robert3005:
Added a test that would catch some of it. The difficulty writing real one is that we don't have a 3rd encoding outside of VarBinView and Constant that can be used to encode Utf8/Binary in the core crate. If this does come again we can add a test encoding here specifically to emulate dict/fsst or pr arrow upstream

onursatici (approved):

vortex-array/src/arrays/varbin/compute/compare.rs
```diff
@@ -146,4 +147,25 @@ mod test {
...
+        );
+        let result = compare(array.as_ref(), vbv.as_ref(), Operator::Eq)
+            .unwrap()
+            .to_bool()
+            .unwrap();
+
+        assert_eq!(
+            result.boolean_buffer(),
+            &BooleanBuffer::from_iter([false, true, true])
+        );
```
a10y, line 169: Is our `compare` well-defined over NULLs?

robert3005: the assertion is a bit lazy. The result has a validity that would null out this field.

joseph-isaacs: The middle value here should be undefined

robert3005: Here you have more assertions for the tests https://github.com/vortex-data/vortex/pull/3441

PR #3424, opened 2025-05-30
fix: Fuzzer generates valid DecimalArrays

robert3005:


vortex-scalar/src/bigint/mod.rs
```diff
@@ -157,12 +159,60 @@ impl CheckedAdd for i256 {
     }
 }
 
+impl WrappingAdd for i256 {
```
robert3005, line 162: all of this is necessary to implement `Int` from `Arbitrary` crate

vortex-scalar/src/bigint/mod.rs
```diff
@@ -15,6 +16,7 @@ use num_traits::{CheckedAdd, CheckedSub, ConstZero, One, Zero};
 #[derive(Debug, Copy, Clone, Default, Eq, PartialEq, Hash, PartialOrd, Ord)]
 pub struct i256(arrow_buffer::i256);
 
+#[allow(clippy::same_name_method)]
```
robert3005, line 19: This is only the case if you compile with `arbitrary` feature which is only enabled in the fuzzer

a10y (approved):

vortex-array/src/arrays/arbitrary.rs
```diff
@@ -69,13 +70,17 @@ fn random_array(u: &mut Unstructured, dtype: &DType, len: Option<usize>) -> Resu
                     PType::F64 => random_primitive::<f64>(u, *n, chunk_len),
                 },
                 DType::Decimal(decimal, n) => {
-                    // TODO(aduffy): also do i256.
-                    let chunk: Vec<i128> = arbitrary_vec_of_len(u, chunk_len)?;
```
a10y, line 73: 🤦

PR #3425, opened 2025-05-30
Vortex FFI Updates

gatesn:
Some updates to the FFI library + docs. More to come.


vortex-ffi/src/lib.rs
```diff
@@ -41,3 +46,134 @@ pub(crate) unsafe fn to_string_vec(ptr: *const *const c_char, len: c_int) -> Vec
...
+            }
+
+            #[doc = r" Clone a borrowed [`" $ffi_ident "`], returning an owned [`" $ffi_ident "`].\n\n"]
+            #[doc = r" Must be released with [`" $ffi_ident "_free`]."]
+            #[unsafe(no_mangle)]
+            pub unsafe extern "C-unwind" fn [<$ffi_ident _clone>](ptr: *const $ffi_ident) -> *const $ffi_ident {
+                if ptr.is_null() {
+                    vortex::error::vortex_panic!("null pointer");
+                }
+                $ffi_ident::new($ffi_ident::into_arc(ptr.cast_mut()).clone())
```
a10y: won't this drop the wrapping Box? should this instead be `Arc::clone($ffi_ident::as_ref(ptr))` ?

vortex-ffi/src/lib.rs
```diff
@@ -41,3 +46,134 @@ pub(crate) unsafe fn to_string_vec(ptr: *const *const c_char, len: c_int) -> Vec
...
+
+                /// Extract a borrowed reference from a const pointer.
+                pub(crate) fn as_ref<'a>(ptr: *const $ffi_ident) -> &'a std::sync::Arc<$T> {
+                    use vortex::error::VortexExpect;
+                    &unsafe { ptr.as_ref() }
+                        .vortex_expect("null pointer")
+                        .0
+                }
+
+                /// Extract an owned reference from a mutable pointer.
```
a10y: comment is outdated i think

vortex-ffi/src/lib.rs
```diff
@@ -41,3 +46,134 @@ pub(crate) unsafe fn to_string_vec(ptr: *const *const c_char, len: c_int) -> Vec
...
+
+            #[doc = r" Clone a borrowed [`" $ffi_ident "`], returning an owned [`" $ffi_ident "`].\n\n"]
+            #[doc = r" Must be released with [`" $ffi_ident "_free`]."]
+            #[unsafe(no_mangle)]
+            pub unsafe extern "C-unwind" fn [<$ffi_ident _clone>](ptr: *const $ffi_ident) -> *mut $ffi_ident {
+                if ptr.is_null() {
+                    vortex::error::vortex_panic!("null pointer");
+                }
+                unsafe { std::sync::Arc::increment_strong_count(ptr) };
+                ptr.cast_mut()
```
a10y: I don't think this is right? if you increment the strong count of the Arc, that doesn't mean you can just treat it as a unique pointer now

vortex-ffi/src/lib.rs
gatesn: Yes it will, but we don't have another choice in the arc_dyn case. If we just incref the internal arc, then two people are going to try and de-allocate the surrounding box.

In the arc_wrapper case, yes we can just incref: https://github.com/vortex-data/vortex/pull/3425/files/0355896381e0264eb379843e3bf5b0665a86d2e0#diff-5965f190b698e12b6c74e058b0e756b46f95b8adb8c80f6b6c0f91131b4f090eR165

vortex-ffi/src/lib.rs
gatesn: Yeah, this should return `*const ffi_ident`, but it's correct that it's now "owned"

vortex-ffi/src/lib.rs
a10y: It should just make a new box?

```
vx_array *array1; // array1 under the hood is a Box<Arc<dyn Array>>
vx_array* array2 = vx_array_clone(array1);
// now, array1's Box was dropped, which means attempts to access it will be use-after-free
```

Am I reading this wrong?

gatesn: Nope, you're completely right.

Man cast_mut is dangerous. But also, very glad this is centralized now. Won't merge until I add some tests for all these.

duckdb-vortex/src/include/vortex_error.hpp
```diff
@@ -8,19 +8,20 @@
...
+		auto msg = std::string(vx_string_ptr(msg_str), vx_string_len(msg_str));
 		vx_error_free(error);
 		throw duckdb::InvalidInputException(msg);
 	}
 }
 
 template <typename Func>
 auto Try(Func func) {
-	vx_error *error = nullptr;
+	const vx_error *error = nullptr;
```
0ax1: As we assign the error as an output parameter, this shouldn't be const, right?

0ax1: Or its at least a little misleading.

vortex-ffi/src/error.rs
```diff
@@ -1,18 +1,19 @@
...
 /// The error structure populated by fallible Vortex C functions.
+// NOTE(ngates): our errors are passed back out as opaque structs, so while we currently alias
+// `vx_string`, we could change this to a different type in the future without breaking the API.
 #[allow(non_camel_case_types)]
-pub struct vx_error {
-    code: c_int,
-    message: CString,
-}
+#[allow(dead_code)]
+pub struct vx_error(vx_string);
```
gatesn: This should just be a box_wrapper!(VortexError)

duckdb-vortex/src/include/vortex_error.hpp
gatesn: I made vx_error a boxed object, although in theory it's correct to be const since we provide no way to mutate an error after construction.

0ax1 (approved):
_Very nice! ✨_

PR #3426, opened 2025-05-30
fix: Teach from_arrow to handle non-nullable fields with nulls

danking:
Arrow permits non-nullable struct array fields to contain nulls if the containing struct array "masks" this null. This is explicitly checked in
[StructArray::try_new](https://github.com/apache/arrow-rs/blob/55.1.0/arrow-array/src/array/struct_array.rs#L169-L180). Lists appear to allow the same situation.

As written, Vortex panics if a null appears in any non-nullable array, even if the array is a field of a struct array. This PR fixes that. It, lamentably, introduces `ArrowNullability` an enum representing these three states: nullable, not nullable, and "not nullable, but a null might appear because this is a field array". I *do not* check that the containing struct or list masks this null because Arrow checked that on creation.

See also: https://spiraldb.slack.com/archives/C07BV3GKAJ2/p1748562330640079

### DIFF
--- a/bench-vortex/src/tpch/mod.rs
+++ b/bench-vortex/src/tpch/mod.rs
@@ -11,7 +11,7 @@ use itertools::Itertools;
 use object_store::ObjectStore;
 use url::Url;
 use vortex::arrays::ChunkedArray;
-use vortex::arrow::FromArrowArray;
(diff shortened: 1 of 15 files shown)


gatesn:
Fixed by #3427

PR #3430, opened 2025-06-01
Fix labels API endpoint

gatesn:
Of course you must read labels from issues not pulls endpoint


a10y:
I've always found it weird that GH puts issues and PRs in the same ID space. Probably some weird historical architecture decision there related to their unsharded MySQL or something

PR #3436, opened 2025-06-02
apply dict layout after coalescing, not before

onursatici:


gatesn:
Heh, curious this has noticably worse performance!

joseph-isaacs:
I think its defiantly work looking at why this is?

onursatici:
yea I will look at this, after all the changes we did I don't think this reorder will alter perf much, I agree this is something we want

onursatici:
regression is not as bad as before, but q25 still seems worse with this, as well as CmsProvider in compression

a10y:
The PR doesn't have any commentary on why, so just ti clarify: Is the goal here to set some upper bound on what the dict expands into when we canonicalize it?

robert3005:
Sorry, I'm rebasing it with hopes that we fixed up the reason things have gotten slower. It's weird that you would create a dict array and then coalesce the input instead of other way around, i.e. you should be able to first get enuogh values to fill a segment and only then do a dictionary encoding. But maybe the partitioning logic is stilly and we get too few values before dict encoding such that we end up with too small chunks since we overestimate their size and they compress very well.

robert3005:
ok, the only thing here is to investigate why CMSProvider regresses which I think will tell us something about our partitioning logic. Otherwise looks roughly flat or slighly better

robert3005:
Let's close it and revisit in the future, we have some writer changes planned that will make this easier to analyse

PR #3438, opened 2025-06-02
Have vortex-datafusion depend only on vortex crate

gatesn:
Also removes the in-memory DataFusion execution plans.


vortex/Cargo.toml
```diff
@@ -20,9 +20,10 @@ workspace = true
 vortex-alp = { workspace = true }
 vortex-array = { workspace = true }
 vortex-btrblocks = { workspace = true }
-vortex-buffer = { workspace = true }
+vortex-buffer = { workspace = true, features = ["memmap2"] }
```
AdamGS: this is weird?

vortex-datafusion/src/lib.rs
```diff
@@ -3,20 +3,13 @@
 #![allow(clippy::cast_possible_truncation)]
```
robert3005: delete this while you're here, there's no truncation

vortex-datafusion/src/persistent/format.rs
```diff
@@ -451,7 +430,7 @@ mod tests {
     async fn fail_table_config() {
         let dir = TempDir::new().unwrap();
 
-        let factory = VortexFormatFactory::default_config();
+        let factory: VortexFormatFactory = Default::default();
```
robert3005, line 434: We should have some guidance on `let var = <T>::default` vs `let var: T = Default::default()`. Nothing to change here, just noticed this when I was reviewing this line

vortex-datafusion/src/persistent/sink.rs
```diff
@@ -137,7 +144,7 @@ mod tests {
     async fn test_insert_into() {
         let dir = TempDir::new().unwrap();
 
-        let factory = VortexFormatFactory::default_config();
+        let factory = VortexFormatFactory::default();
```
robert3005, line 138: then this is inconsistent with the earlier examples

vortex/Cargo.toml
gatesn: Why? "Vortex" the library supports memory mapped files. The vortex-buffer crate has optional support for it

AdamGS: this makes it impossible to opt out of that feature

robert3005: I think what Adam is suggesting is that this should be forwarded to the feature of this crate?

gatesn: 🤷 doesn't seem like a huge feature. Depends if we want the vortex crate to be batteries included or not

robert3005 (approved):

vortex/Cargo.toml
AdamGS: I'm not sure I understand what "batteries included" means in this context, but I do think that if that's a core design principle of Vortex, it's worth expanding upon somewhere in the repo.

a10y: Agree with Adam, at the very least, it should probably be forwarded to a feature of this crate and enabled by default, but can be disabled with `default-features = false` ?

PR #3439, opened 2025-06-02
feat[vortex-expr]: let expression

joseph-isaacs:


vortex-expr/src/like.rs
```diff
@@ -107,9 +107,9 @@ impl VortexExpr for Like {
         self
     }
 
-    fn unchecked_evaluate(&self, batch: &dyn Array) -> VortexResult<ArrayRef> {
-        let child = self.child().evaluate(batch)?;
-        let pattern = self.pattern().evaluate(&child)?;
+    fn unchecked_evaluate(&self, ctx: &EvaluationContext) -> VortexResult<ArrayRef> {
+        let child = self.child().unchecked_evaluate(ctx)?;
+        let pattern = self.pattern().unchecked_evaluate(ctx)?;
```
joseph-isaacs: child was a bug

vortex-expr/src/var.rs
```diff
@@ -0,0 +1,141 @@
...
+            Ok(Kind::Var(kind::Var {
+                var: self.var.to_string(),
+            }))
+        }
+    }
+}
+
+impl Display for Var {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "[{}]", self.var)
```
gatesn: I tihnk this should be `"${var}"` where the root scope is the empty string

vortex-expr/src/transform/partition.rs
```diff
@@ -445,4 +450,6 @@ mod tests {
             )
         )
     }
+
+    // #[te// }
```
gatesn: ?

vortex-expr/src/lib.rs
```diff
@@ -73,40 +77,43 @@ pub trait ExprSerializable {
...
 /// Represents logical operation on [`ArrayRef`]s
 pub trait VortexExpr: Debug + Send + Sync + DynEq + DynHash + Display + ExprSerializable {
     /// Convert expression reference to reference of [`Any`] type
     fn as_any(&self) -> &dyn Any;
 
     /// Compute result of expression on given batch producing a new batch
     ///
-    fn evaluate(&self, batch: &dyn Array) -> VortexResult<ArrayRef> {
-        let result = self.unchecked_evaluate(batch)?;
+    fn evaluate(&self, ctx: &EvaluationContext) -> VortexResult<ArrayRef> {
```
gatesn: `EvaluationContext` -> `Scope`
`DTypeEvaluationContext` -> `ScopeDType`

vortex-expr/src/lib.rs
```diff
@@ -73,40 +77,43 @@ pub trait ExprSerializable {
...
             "Expression {} returned dtype {} but declared return_dtype of {}",
             self,
             result.dtype(),
-            self.return_dtype(batch.dtype())?,
+            self.return_dtype(&ctx.into())?,
         );
         Ok(result)
     }
 
+    fn evaluate_array(&self, array: &dyn Array) -> VortexResult<ArrayRef> {
```
gatesn: Remove this, make a `Scope::new(ArrayRef)`, which is concise for the common case.
Also `ScopeDType::new(DType)`

gatesn: Rename `ident()` -> `root()`

vortex-expr/src/context.rs
```diff
@@ -0,0 +1,148 @@
...
+use vortex_dtype::DType;
+use vortex_error::{VortexError, VortexExpect, VortexResult, vortex_err};
+
+use crate::IDENTITY_IDENTIFIER;
+
+pub type Identifier = Arc<str>;
+pub type ExprScope<T> = HashMap<Identifier, T>;
+
+#[derive(Clone, Default)]
+pub struct ValuesScope(ExprScope<ArrayRef>);
```
gatesn: I wouldn't bother wrapping / exposing these types, keep them internal to Context (Scope)

vortex-expr/src/context.rs
```diff
@@ -0,0 +1,148 @@
...
+#[derive(Clone, Default)]
+pub struct ValuesDTypeScope(ExprScope<DType>);
+#[allow(dead_code)]
+#[derive(Clone, Default)]
+pub struct VarsScope(ExprScope<Arc<dyn Any>>);
+
+#[derive(Clone, Default)]
+pub struct EvaluationContext {
+    array_len: usize,
```
gatesn: Probably hold an explicit `root: ArrayRef` just for easy access / no hash map when we don't have any vars

vortex-expr/src/context.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    pub fn values(&self, id: &Identifier) -> VortexResult<&ArrayRef> {
+        self.values
+            .0
+            .get(id)
+            .ok_or_else(|| vortex_err!("cannot find {} in values scope", id))
+    }
+
+    #[allow(clippy::len_without_is_empty)]
```
gatesn: Just add is_empty

vortex-expr/src/context.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+impl DTypeEvaluationContext {
+    pub fn new_identity(dtype: DType) -> Self {
+        Self {
+            types: ValuesDTypeScope(HashMap::from([(IDENTITY_IDENTIFIER.into(), dtype)])),
+        }
+    }
+
+    pub fn type_(&self, id: &Identifier) -> VortexResult<&DType> {
```
gatesn: dtype

vortex-expr/src/var.rs
robert3005: When we next bump rust this will be a builtin lint. I manually fixed it one off to simplify future migration

gatesn (approved):

PR #3444, opened 2025-06-03
fix: ddb build

0ax1:
Follow up to https://github.com/vortex-data/vortex/pull/3425. Fixes clippy issues, and wiring in the build to the workspace.


vortex-duckdb-ext/build.rs
joseph-isaacs, line 1: can you use a cbindgen.toml file?

joseph-isaacs (commented):
cbindgen.toml and we are good to go.

joseph-isaacs (approved):

PR #3445, opened 2025-06-03
fix: struct array must skip its validity child

danking:


delta003 (approved):

robert3005:
This was a silly bug, we never exercise serde for struct since we have a layout for it

gatesn:
We do, but only in the stats table. So makes sense the bugs kept showing up there

PR #3448, opened 2025-06-04
list_contains to use any semantics

joseph-isaacs:


vortex-array/src/arrays/bool/test.rs
```diff
@@ -0,0 +1,16 @@
+use itertools::Itertools;
+use vortex_error::VortexResult;
+
+use crate::arrays::BoolArray;
+
+impl BoolArray {
+    pub fn opt_iter(&self) -> VortexResult<impl IntoIterator<Item = Option<bool>>> {
```
gatesn, line 7: Return impl Iterator instead, you're collecting to vec right now

vortex-expr/src/list_contains.rs
```diff
@@ -92,11 +87,14 @@ impl VortexExpr for ListContains {
     }
 
     fn unchecked_evaluate(&self, scope: &Scope) -> VortexResult<ArrayRef> {
-        compute_list_contains(self.list.evaluate(scope)?.as_ref(), self.value.clone())
+        let Some(scalar) = self.value.evaluate(scope)?.as_constant() else {
+            vortex_bail!("not implemented list contains of a value array")
```
delta003:
```suggestion
            todo!("not implemented list contains of a value array")
```

vortex-expr/src/list_contains.rs
```diff
@@ -6,32 +6,31 @@ use std::sync::Arc;
...
     }
 
-    pub fn value(&self) -> &Scalar {
+    pub fn value(&self) -> &ExprRef {
         &self.value
     }
 }
 
-pub fn list_contains(list: ExprRef, value: impl Into<Scalar>) -> ExprRef {
+pub fn list_contains(list: ExprRef, value: impl Into<ExprRef>) -> ExprRef {
```
delta003:
```suggestion
pub fn list_contains(list: ExprRef, value: ExprRef) -> ExprRef {
```

delta003: This is more consistent with other expressions

vortex-array/src/compute/list.rs
```diff
@@ -332,8 +311,8 @@ mod tests {
     // Case 2: list(utf8?) with NULL search scalar
```
gatesn: Comment is stale now?

vortex-array/src/compute/list.rs
```diff
@@ -6,26 +6,28 @@ use num_traits::AsPrimitive;
...
 use crate::{Array, ArrayRef, IntoArray, ToCanonical};
 
-/// Compute a `Bool`-typed array the same length as `array` where elements are `true` if the list
-/// item contains the `value`, or `false` otherwise.
+/// Compute a `Bool`-typed array the same length as `array` where elements is `true` if the list
+/// item contains the `value`, `false` otherwise.
 ///
-/// If the ListArray is nullable, then the result will contain nulls matching the null mask
-/// of the original array.
+/// If the ListArray or Element is nullable, then the result will contain nulls matching the union
```
delta003: What's an Element here? Can you rewrite this so it's more clear

vortex-array/src/compute/list.rs
```diff
@@ -52,6 +54,15 @@ pub fn list_contains(array: &dyn Array, value: Scalar) -> VortexResult<ArrayRef>
     if &**elem_dtype != value.dtype() {
```
gatesn, line 53: Might be worth promoting this to an actual compute function? That way it would have return type assertions

vortex-array/src/compute/list.rs
delta003: This is now a wrong description

vortex-array/src/compute/list.rs
```diff
@@ -357,14 +336,20 @@ mod tests {
...
-        bool_array(vec![false, true, true], None)
+        bool_array(vec![false, true, true], Some(vec![false, false, false]))
     )]
     // Case 7: list(utf8?) with empty + NULL elements and search scalar
     #[case(
         null_strings(vec![vec![], vec![None, None], vec![None, None, None]]),
         Some("a"),
         bool_array(vec![false, false, false], None)
     )]
+    // Case 8: list(utf8?) with null and some match
```
delta003:
```suggestion
    // Case 8: list(utf8?) with NULL and some match
```

delta003: for now reason other than all other case descriptions use NULL

vortex-array/src/arrays/bool/test.rs
joseph-isaacs: Does this matter is the test_harness?

vortex-array/src/compute/list.rs
```diff
@@ -6,26 +6,28 @@ use num_traits::AsPrimitive;
...
 /// ## Null scalar handling
 ///
-/// When the search scalar is `NULL`, then the resulting array will be a `BoolArray` containing
-/// `true` if the list contains any nulls, and `false` if the list does not contain any nulls,
-/// or `NULL` for null lists.
+/// When the value or list is `NULL` the result is `NULL` otherwise is will be a non-nullable value.
+///
+/// ## Format semantics
+///
+/// list_contains(list, elem)
```
a10y, line 26: could we wrap in markdown so this renders nicely for the docs page?

vortex-array/src/compute/list.rs
joseph-isaacs: removed

vortex-array/src/compute/list.rs
joseph-isaacs: Shall we leave this as a todo

vortex-array/src/compute/list.rs
```diff
@@ -329,11 +307,10 @@ mod tests {
         Some("a"),
         bool_array(vec![false, true, true], None)
     )]
-    // Case 2: list(utf8?) with NULL search scalar
```
delta003, line 332: Haha well this doesn't really solve it, like how does this inconsistency not bother you! Update a comment instead of deleting it please 🙏

vortex-array/src/compute/list.rs
```diff
@@ -357,14 +334,19 @@ mod tests {
...
-        bool_array(vec![false, true, true], None)
+        bool_array(vec![false, true, true], Some(vec![false, false, false]))
     )]
     // Case 7: list(utf8?) with empty + NULL elements and search scalar
     #[case(
         null_strings(vec![vec![], vec![None, None], vec![None, None, None]]),
         Some("a"),
         bool_array(vec![false, false, false], None)
     )]
+    #[case(
```
delta003: Missing Cast 8: what's the intention of this test case, it should be clear that this isn't already tested above

delta003 (commented):
I'm more okay with deleting all test descriptions rather than the current state

vortex-array/src/compute/list.rs
joseph-isaacs: Ill add them back :)

delta003 (approved):

PR #3449, opened 2025-06-04
feat[vortex-layout]: add row_id layout reader to support row_id expr queries

joseph-isaacs:
Add the following well defined Scope value 

```
dtype("$vx.row_id") = {file_index: u64, file_row_number: u64}
```

This can be used in the scan, and file pruning


vortex-expr/src/lib.rs
```diff
@@ -192,6 +202,25 @@ impl Eq for dyn VortexExpr {}
 
 dyn_hash::hash_trait_object!(VortexExpr);
 
+/// An expression wrapper that performs pointer equality.
+/// NOTE(ngates): we should consider if this shoud live in vortex-expr crate?
```
onursatici: we can remove this comment now

vortex-expr/src/lib.rs
```diff
@@ -118,20 +119,29 @@ pub trait VortexExpr: Debug + Send + Sync + DynEq + DynHash + Display + ExprSeri
 
 pub trait VortexExprExt {
     /// Accumulate all field references from this expression and its children in a set
-    fn references(&self) -> HashSet<FieldName>;
+    fn root_references(&self) -> HashSet<FieldName>;
```
onursatici: why this rename? I thought `references` was collecting all referenced fields in the expressions, from the name `root_references` I think it would only collect the references in the direct children of root

vortex-expr/src/transform/immediate_access.rs
```diff
@@ -20,81 +20,36 @@ pub fn immediate_scope_accesses<'a>(
     expr: &'a ExprRef,
     scope_dtype: &'a StructFields,
 ) -> VortexResult<FieldAccesses<'a>> {
-    ImmediateScopeAccessesAnalysis::<'a>::analyze(expr, scope_dtype)
+    AccessesAnalysis::analyze(expr, move |node| {
+        assert!(
+            !node.as_any().is::<Select>(),
+            "cannot analyse select, simply the expression"
```
onursatici: nit: I know this is a copy, but `simply -> simplify` while we are here

vortex-expr/src/transform/var_partition.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    pub(crate) fn field_idx_name(field: &FieldName, idx: usize) -> FieldName {
+        let mut hasher = SPLITTER_RANDOM_STATE.build_hasher();
+        field.hash(&mut hasher);
+        idx.hash(&mut hasher);
+        hasher.finish().to_string().into()
+    }
+
+    fn split_all(expr: &ExprRef) -> VortexResult<VarPartitionedExpr> {
+        Self::split(expr, |f| f.clone())
```
onursatici: nit: `f -> identifier`

vortex-layout/src/layouts/row_id.rs
```diff
@@ -0,0 +1,366 @@
...
+            Some(if matches!(dtype, DType::Bool(_)) {
+                // If the partition evaluates to a boolean, we can evaluate it as a mask which
+                // can often be more efficient since nulls are turned into `false` early on,
+                // and layouts can perform predicate pruning / indexing.
+                FieldEval::Mask(self.child.filter_evaluation(row_range, expr)?)
+            } else {
+                // Otherwise, we evaluate the projection as an array, and combine the results
+                // at the end.
+                FieldEval::Array(self.child.projection_evaluation(row_range, expr)?)
+            })
```
onursatici, line 209: A follow up but having this both on struct reader and here I wonder if this warrants to be more generic

vortex-layout/src/layouts/row_id.rs
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    fn filter_evaluation(
+        &self,
+        row_range: &Range<u64>,
+        expr: &ExprRef,
+    ) -> VortexResult<Box<dyn MaskEvaluation>> {
+        let arr_len = row_range.clone().count();
+        let partitioned = self.partition_expr(expr);
+        let Some(row_id) = partitioned.find_partition(&ROW_ID) else {
```
onursatici, line 191: If the array already contains a `row_id` column I suspect we will override it with the sequence array in our evaluations?

vortex-expr/src/lib.rs
joseph-isaacs: This function is very error prone, I think we should remove it. It should return FieldPaths.

vortex-layout/src/layouts/row_id.rs
joseph-isaacs: This splits on the variable domain, not the struct field path domain, so there cannot be a clash

vortex-layout/src/layouts/row_id.rs
joseph-isaacs: I think that this should disappear, I have a PR https://github.com/vortex-data/vortex/pull/3487

.github/workflows/sql-benchmarks.yml
```diff
@@ -96,13 +96,36 @@ jobs:
...
+          CXX: clang++
+          GEN: ninja
+          NATIVE_ARCH: 1
+          DUCKDB_PLATFORM: linux_amd64
+          LTO: thin
+        run: |
+          git clone https://github.com/duckdb/duckdb
+          cd duckdb
+          git checkout "$DUCKDB_VERSION"
+          make release
```
onursatici: leftover debugging diff?

duckdb-vortex/src/vortex_expr.cpp
```diff
@@ -246,44 +249,59 @@ scalar::Scalar *into_vortex_scalar(Arena &arena, const Value &value, bool nullab
...
 	column->set_id(GET_ITEM_ID);
 	auto kind = column->mutable_kind();
 	auto get_item = kind->mutable_get_item();
 	get_item->mutable_path()->assign(s);
 
 	auto id = column->add_children();
-	id->mutable_kind()->mutable_var()->set_var("");
 	id->set_id(VAR_ID);
+	if (s == "file_row_number" || s == "file_index") {
+		id->mutable_kind()->mutable_var()->set_var(VX_ROW_ID_COL_ID);
```
onursatici, line 301: pardon my duckdb knowledge but can row_number and file_index be both  set in the same expression? If so maybe we should use different variable names for these two

vortex-expr/src/lib.rs
```diff
@@ -123,20 +124,30 @@ pub trait VortexExpr:
...
         self.accept(&mut collector).vortex_unwrap();
         collector.into_fields()
     }
 
+    fn vars(&self) -> HashSet<Identifier> {
+        let mut collector = VarsCollector::new();
+        // The collector is infallible, so we can unwrap the result
+        self.accept(&mut collector).vortex_unwrap();
+        collector.into_vars()
+        // println!("vars{:?}", va);
```
onursatici:
```suggestion
```

vortex-layout/src/layouts/row_id.rs
```diff
@@ -0,0 +1,454 @@
...
+            return self.child.projection_evaluation(row_range, expr);
+        };
+        let rest = partitioned.find_partition(&Identifier::Identity);
+
+        let row_id_scope = Scope::empty(arr_len).with_array_pair(self.row_id_scope(row_range));
+
+        let res = row_id.evaluate(&row_id_scope)?;
```
onursatici: should we do this on the returned `RowIdArrayEvaluation`?

vortex-layout/src/layouts/row_id.rs
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    pub fn row_id_scope_dtype() -> (Identifier, DType) {
+        (
+            ROW_ID.clone(),
+            DType::Struct(
+                Arc::new(StructFields::from_iter([
+                    (
+                        FieldName::from("file_row_number"),
+                        DType::Primitive(PType::I64, Nullability::NonNullable),
```
onursatici:
```suggestion
                        PType::I64.into(),
```

duckdb-vortex/src/vortex_expr.cpp
joseph-isaacs: The `VX_ROW_ID_COL_ID` its a struct array with two cols [file_row_number, file_index]

onursatici (approved):

duckdb-vortex/src/vortex_expr.cpp
gatesn: Why does this know about files? I would have thought any multi-file reader would just provide the correct row offset

joseph-isaacs: This was added to pave the way for the multi-file reader, i'm happy to revisit this if desired

gatesn: Yeah I think we need a global row offset, otherwise you'll struggle to return a row idx to DuckDB, and row idx expressions would make less sense

PR #3451, opened 2025-06-04
chore: remove unused dependencies

0ax1:
Removes unused deps based on https://github.com/bnjbvr/cargo-machete & https://github.com/Boshen/cargo-shear.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -225,11 +225,6 @@ features = ["wasm_js"]
 package = "getrandom"
 version = "0.3.1"
 
-[workspace.dependencies.getrandom_v02]
-features = ["js"]
-package = "getrandom"
-version = "0.2.15"
-
 [workspace.lints.rust]
 macro_use_extern_crate = "deny"
 redundant_lifetimes = "deny"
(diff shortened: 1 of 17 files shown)


vortex-file/src/writer.rs
```diff
@@ -22,7 +22,7 @@ use crate::{EOF_SIZE, MAGIC_BYTES, MAX_FOOTER_SIZE, VERSION, VortexLayoutStrateg
 /// Configure a new writer, which can eventually be used to write an [`ArrayStream`] into a sink that implements [`VortexWrite`].
 ///
 /// By default, the [`LayoutStrategy`] will be the [`VortexLayoutStrategy`], which includes re-chunking and will also
-/// uncompress all data back to its canonical form before compressing it using the [`BtrBlocksCompressor`](vortex_btrblocks::BtrBlocksCompressor).
+/// uncompress all data back to its canonical form before compressing it using the `vortex_btrblocks::BtrBlocksCompressor`.
```
0ax1, line 25: We only had `vortex_btrblocks` as a dep in `vortex_file` because of this comment.

Cargo.toml
```diff
@@ -225,11 +225,6 @@ features = ["wasm_js"]
 package = "getrandom"
 version = "0.3.1"
 
-[workspace.dependencies.getrandom_v02]
```
0ax1, line 228: Unified the workspace to only use `v03`.

joseph-isaacs (approved):
LG

Cargo.toml
robert3005: Glad that it worked. v02 and v03 existed only while some dependencies were still using the old one

PR #3452, opened 2025-06-04
chore: lint for #3451 changes

danking:


danking:
Vortex is too complex for this.

PR #3455, opened 2025-06-04
chore: layout cleanups

a10y:
Mostly drive-bys for readability

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,10 +67,8 @@ arrow-ord = "55"
 arrow-schema = "55"
 arrow-select = "55"
 arrow-string = "55"
-async-once-cell = "0.5.4"
 async-stream = "0.3.6"
 async-trait = "0.1.88"
-backtrace = "0.3.74"
 bindgen = "0.71.1"
 bit-vec = "0.8.0"
 bitvec = "1.0.1"
@@ -86,10 +84,6 @@ crossterm = "0.28"
 dashmap = "6.1.0"
 datafusion = { version = "47", default-features = false }
 datafusion-common = { version = "47" }
-datafusion-datasource = { version = "47" }
-datafusion-execution = { version = "47" }
-datafusion-expr = { version = "47" }
-datafusion-physical-expr = { version = "47" }
 datafusion-physical-plan = { version = "47" }
 divan = { package = "codspeed-divan-compat", version = "2.8.0" }
 duckdb = { path = "duckdb-vortex/duckdb-rs/crates/duckdb", features = [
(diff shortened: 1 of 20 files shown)


Cargo.toml
```diff
@@ -67,10 +67,8 @@ arrow-ord = "55"
 arrow-schema = "55"
 arrow-select = "55"
 arrow-string = "55"
-async-once-cell = "0.5.4"
```
a10y, line 70: RustRover tells me all of these are unused

vortex-layout/src/layout.rs
```diff
@@ -49,9 +49,9 @@ pub trait Layout: 'static + Send + Sync + Debug + private::Sealed {
 
     fn new_reader(
         &self,
-        name: &Arc<str>,
-        segment_source: &Arc<dyn SegmentSource>,
-        ctx: &ArrayContext,
```
a10y, line 54: The first thing that every single Layout impl does is clone all of these, but idk

vortex-layout/src/layouts/zoned/reader.rs
```diff
@@ -136,18 +139,16 @@ impl ZonedReader {
             .clone()
     }
 
-    /// Return the zone range covered by a row range.
-    pub(crate) fn zone_range(&self, row_range: &Range<u64>) -> Range<usize> {
-        let zone_start = usize::try_from(row_range.start / self.layout.zone_len as u64)
-            .vortex_expect("Invalid zone start");
-        let zone_end = usize::try_from(row_range.end.div_ceil(self.layout.zone_len as u64))
-            .vortex_expect("Invalid zone end");
```
a10y, line 144: we now defer doing the u64 -> usize truncation until we're actually building the Mask, which I think is slightly more correct but unlikely to come up in practice. Reduces a lot of usize::try_from at least 🤷

vortex-layout/src/layouts/zoned/reader.rs
```diff
@@ -232,20 +236,20 @@ impl LayoutReader for ZonedReader {
     }
 }
 
-struct StatsPruningEvaluation {
+struct ZoneMapPruningEvaluation {
```
a10y, line 239: This naming seems more consistent with the other layouts

vortex-layout/src/writer.rs
```diff
@@ -6,9 +6,7 @@ use crate::LayoutRef;
 
 /// A future created by a strategy to yield a layout. It is its own
 /// trait to be potentially extended with new methods.
-// [layout writer]
```
gatesn, line 9: Was this not used in the documentation?

gatesn (approved):

vortex-layout/src/writer.rs
a10y: I'm not sure, haven't seen that syntax before

a10y: oh i see it now 
image


I can revert this part

gatesn: I wonder if we can get Sphinx to fail if that include block doesn't match anything

a10y: The Python docs CI check failed I believe for this reason

PR #3456, opened 2025-06-04
Zstd encoding

mwlon:
optionally supports finer-grained frames that share a dictionary for faster single-row or slice decompression


mwlon:
@gatesn Do I need to do something to get github actions to run consistently?

robert3005:
We have a bit of a restrictive setup that for new contributors requires every run to be approved

mwlon:
I'm not sure what this wasm issue is. Something about wasm+zstd

robert3005:
I think the issue with wasm is this https://stackoverflow.com/questions/54082459/fatal-error-bits-libc-header-start-h-no-such-file-or-directory-while-compili. We compile for wasm which is 32 bits but we don't have the necessary libc. However, I'm not sure that standard zstd works on wasm so might be we need to omit it and come up with a workaround. Quick internet search suggests it should be possible to make it work but that's not a priority right now.

mwlon:
+1 to omitting zstd on wasm for a workaround. Maybe we make zstd a default feature that we omit during wasm builds?

gatesn:
There's a couple of crates already that are omitted from the wasm build, e.g. take a look at line ~180 in the ci.yml where we have exclusions for the WASM build.

mwlon:
Trying to fix the DCO check (how the heck am I supposed to write these messages in every time?), it appears I messed up the branch state

mwlon:
Looks like `-s` is supposed to do this, so I guess I'll make a bash alias for `git commit -a -s -m`

mwlon:
Ok, branch is sane again

mwlon:
@gatesn maybe this builds now? Will someone be assigned to review?

mwlon:
I realize I forgot to collect the valid elements and compress only those. Will fix later

robert3005:
@mwlon I have pushed a commit that should fix the wasm integration test. We not only check that we can build the wasm target but also that we can take it as a dependency that's runtime specific. The second part needed dependency adjustment.

mwlon:
Ok, now it only compresses values instead of all possibly null elements.

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    pub fn from_primitive(
+        parray: &PrimitiveArray,
+        level: i32,
+        rows_per_frame: usize,
+    ) -> VortexResult<Self> {
+        let dtype = parray.dtype().clone();
+        let byte_width = parray.ptype().byte_width();
+        let validity = parray.validity().clone();
```
robert3005: Generally when we want to perform some compute operation on validity we extract the `Mask` via `validity_mask()`, i.e. you get the canonical validity while this is only used to hold the value.

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+        let rows_per_frame = if rows_per_frame > 0 {
+            rows_per_frame
+        } else {
+            n_rows
+        };
+        let frame_row_indices = (0..n_rows).step_by(rows_per_frame).collect::<Vec<_>>();
+        let n_frames = frame_row_indices.len();
+
+        // We compress only the valid elements.
+        let values = parray.collect_values()?;
```
robert3005: This is implemented via `filter(parray, parray.validity_mask()?)?.to_primitive()?`. There's slight potential improvement by stripping validity from parray but not sure it matters overall

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+        let values = values.byte_buffer();
+        let value_bytes = values.inner();
+
+        // Would-be sample sizes if we end up applying zstd dictionary
+        let sample_sizes: Vec<usize> = value_indices
+            .windows(2)
+            .map(|pair| (pair[1] - pair[0]) * byte_width)
+            .filter(|&size| size > 0)
+            .collect();
+        debug_assert!(sample_sizes.iter().sum::<usize>() == value_bytes.len());
```
robert3005:
```suggestion
        debug_assert_eq!(sample_sizes.iter().sum::<usize>(), value_bytes.len());
```

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+        };
+        let frame_row_indices = (0..n_rows).step_by(rows_per_frame).collect::<Vec<_>>();
+        let n_frames = frame_row_indices.len();
+
+        // We compress only the valid elements.
+        let values = parray.collect_values()?;
+        let mut value_indices = validity.value_indices_for_rows(&frame_row_indices)?;
+        value_indices.push(values.len()); // for convenience
+        let values = values.byte_buffer();
+        let value_bytes = values.inner();
```
robert3005, line 143: There's really no reason to use the inner of buffer unless you need some specific `Bytes` api access, i.e. just use `values` directly

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+            let buf_stop = frame_start_row + usize::try_from(frame_meta.n_rows)?;
+            if frame_start_row < self.slice_start {
+                frame_idx_lb = i;
+                row_offset = self.slice_start - frame_start_row
+            }
+            if frame_start_row < self.slice_stop {
+                frame_idx_ub = i + 1
+            }
+            frame_start_row = buf_stop;
+        }
```
robert3005, line 229: This could be a binary search? Though I would start with a shortcircuit of `slice_n_rows == self.len()`

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+            .try_into()?;
+
+        let mut decompressor = if let Some(dictionary) = &self.dictionary {
+            zstd::bulk::Decompressor::with_dictionary(dictionary)
+        } else {
+            zstd::bulk::Decompressor::new()
+        }?;
+
+        // we could make this empty initialized for better performance
+        let mut frames_values_bytes = vec![0; total_uncompressed_size];
```
robert3005: We have our own buffer to be able to easily do those unitiliazed writes. 
```suggestion
        let mut frames_values_bytes = ByteBufferMut::with_capacity_aligned(
            total_uncompressed_size,
            Alignment::new(byte_width),
        );
```
then before decompress_to_buffer we need
```
unsafe { frames_values_bytes.set_len(stop_byte) };
```
and once again at the end of the loop
```
unsafe { frames_values_bytes.set_len(total_uncompressed_size) };
```

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+        let frames_validity = self
+            .validity
+            .slice(self.slice_start - row_offset, self.slice_stop)?;
+        let frames_values_start_stop =
+            frames_validity.value_indices_for_rows(&[row_offset, row_offset + slice_n_rows])?;
+        let slice_value_bytes = frames_values_bytes
+            [frames_values_start_stop[0] * byte_width..frames_values_start_stop[1] * byte_width]
+            .to_vec();
+        let slice_values_buffer =
+            ByteBuffer::from(slice_value_bytes).aligned(Alignment::new(byte_width));
```
robert3005: We can then simplify these two into 
```suggestion
        let slice_values_buffer = frames_values_bytes.freeze().slice(
            frames_values_start_stop[0] * byte_width..frames_values_start_stop[1] * byte_width,
        );
```

encodings/zstd/src/array.rs
```diff
@@ -0,0 +1,321 @@
...
+            slice_n_rows,
+        )?;
+
+        Ok(primitive.into_array())
+    }
+
+    fn _slice(&self, start: usize, stop: usize) -> VortexResult<ZstdArray> {
+        if self.slice_start + cmp::max(start, stop) > self.slice_stop {
+            vortex_bail!("Cannot slice beyond end of ZstdArray")
+        }
```
robert3005: top level slice does this validation already

vortex-array/src/arrays/primitive/mod.rs
```diff
@@ -115,6 +115,41 @@ impl PrimitiveArray {
...
+                    if bool_buffer.value(i) {
+                        bytes.extend_from_slice(
+                            &valid_elems_buffer[valid_i * byte_width..(valid_i + 1) * byte_width],
+                        );
+                        valid_i += 1;
+                    } else {
+                        bytes.extend_from_slice(&vec![0; byte_width]);
+                    }
+                }
+                ByteBuffer::from(bytes)
```
robert3005: I think this is simpler expressed if we prefill the invalid values and then only copy the valid value, however, the optimal way is dependent on valid values density. I would suggest something like
```
let bool_array = is_valid.to_canonical()?.into_bool()?;
let bool_buffer = bool_array.boolean_buffer();
let mut bytes =
    ByteBufferMut::zeroed_aligned(n_rows * byte_width, Alignment::new(byte_width));
for (i, valid_i) in bool_buffer.set_indices().enumerate() {
    bytes[valid_i * byte_width..(valid_i + 1) * byte_width]
        .copy_from_slice(&valid_elems_buffer[i * byte_width..(i + 1) * byte_width])
}
bytes.freeze()
```
for now

vortex-array/src/arrays/primitive/mod.rs
```diff
@@ -115,6 +115,41 @@ impl PrimitiveArray {
...
+    pub fn from_values_byte_buffer(
+        valid_elems_buffer: ByteBuffer,
+        ptype: PType,
+        validity: Validity,
+        n_rows: usize,
+    ) -> VortexResult<Self> {
+        let byte_width = ptype.byte_width();
+        let buffer = match &validity {
+            Validity::AllValid | Validity::NonNullable => valid_elems_buffer,
+            Validity::AllInvalid => ByteBuffer::from(vec![0; n_rows * byte_width]),
```
robert3005:
```suggestion
            Validity::AllInvalid => ByteBuffer::zeroed_aligned(n_rows * byte_width, Alignment::new(byte_width)),
```

vortex-array/src/arrays/primitive/mod.rs
```diff
@@ -115,6 +115,41 @@ impl PrimitiveArray {
...
+                        bytes.extend_from_slice(&vec![0; byte_width]);
+                    }
+                }
+                ByteBuffer::from(bytes)
+            }
+        };
+
+        match_each_native_ptype!(ptype, |T| {
+            Ok(Self::new::<T>(Buffer::from_byte_buffer(buffer), validity))
+        })
```
robert3005: Since we align the buffers in the previous step we can make this jusst
```
Ok(Self::from_byte_buffer(buffer, ptype, validity))
```

vortex-array/src/arrays/primitive/mod.rs
```diff
@@ -229,6 +264,44 @@ impl PrimitiveArray {
         Ok(PrimitiveArray::new(buffer.freeze(), validity.clone()))
     }
 
+    /// Filter down to only valid elements, returning a new non-nullable array.
+    ///
+    /// The resulting array's count of valid elements will be the same, but its
+    /// count of rows may be less.
+    pub fn collect_values(&self) -> VortexResult<PrimitiveArray> {
```
robert3005: As commented before this is a filter function, if we use filter this can be removed

vortex-array/src/arrays/primitive/mod.rs
robert3005: Part of me is wondering if this is simpler to understand if we do match_each_ptype to avoid size calculations

vortex-array/src/validity.rs
```diff
@@ -376,6 +376,37 @@ impl Validity {
...
+                for &next_row_idx in row_indices {
+                    while row_idx < next_row_idx {
+                        row_idx += 1;
+                        value_idx += bool_iter
+                            .next()
+                            .ok_or_else(|| vortex_err!("Row indices exceed array length"))?
+                            as usize;
+                    }
+                    value_indices.push(value_idx);
+                }
```
robert3005:
```suggestion
                let mut bool_iter = bool_array.boolean_buffer().set_indices().peekable();
                let mut value_indices = Vec::with_capacity(row_indices.len());
                let mut value_idx = 0;
                for &next_row_idx in row_indices {
                    if let Some(&next_set_idx) = bool_iter.peek() {
                        if next_set_idx < next_row_idx {
                            value_idx += bool_iter
                                .by_ref()
                                .peeking_take_while(|&set_idx| set_idx < next_row_idx)
                                .count();
                        }
                    }
                    value_indices.push(value_idx);
                }
```
I think something like this would work better where we only iterate set indices. You very rarely want to iterate every bit.

robert3005 (changes requested):
I left comments where we can use some features of our library to simplify the code or make it faster. Mostly around using the features of our Buffer and BooleanBuffer for faster iteration

encodings/zstd/src/array.rs
robert3005: It would then make sense to have `value_indices_for_rows` be a method on a Mask and not validity

vortex-array/src/validity.rs
```diff
@@ -376,6 +376,37 @@ impl Validity {
     pub fn is_array(&self) -> bool {
         matches!(self, Validity::Array(_))
     }
+
+    /// Given monotonically increasing `row_indices` in [0, n_rows], returns the
+    /// count of valid elements up to each row_index.
+    ///
+    /// This is O(n_rows).
+    pub fn value_indices_for_rows(&self, row_indices: &[usize]) -> VortexResult<Vec<usize>> {
```
robert3005: maybe `offsets_value_counts` is a better name? It feels like the name has two level of references that makes this hard to name and then rename `row_indices` to `row_offsets`?

encodings/zstd/src/array.rs
mwlon: The binary search wouldn't buy us anything, since we need to do a linear cumulative sum to begin with. The number of buffers should be pretty small in comparison to the data, so this part should be cheap.

vortex-array/src/validity.rs
mwlon: I actually think the original will be more performant:
1. set_indices is linearly iterating over every bit to begin with
2. There are a lot fewer Options and complex iterator types in the original. Some of these will probably get zero-costed away, but I'd guess not all.

encodings/zstd/src/array.rs
robert3005: Right, I was mostly trying to confirm my understanding!

vortex-array/src/validity.rs
mwlon: Renamed to `valid_counts_for_indices`, given how `indices` refers to rows in this file. How's that?

vortex-array/src/arrays/primitive/mod.rs
mwlon: :+1:

vortex-array/src/arrays/primitive/mod.rs
mwlon: :+1:

vortex-array/src/arrays/primitive/mod.rs
mwlon: Did your suggestion above, but couldn't mutably dereference the buffer so I used a Vec<u8>.

There's would be some performance gain from using compile-time type sizes allowing shifts instead of multiplications in the type-specialized version, but I think it would be fairly minor and it's better to avoid unnecessary binary bloat in this case. My opinion is weak though.

encodings/zstd/src/array.rs
mwlon: :+1:

encodings/zstd/src/array.rs
mwlon: :+1:

encodings/zstd/src/array.rs
mwlon: :+1:

vortex-array/src/arrays/primitive/mod.rs
robert3005: There's two types `ByteBufferMut` is the type that implements `DerefMut` that you then freeze to get `ByteBuffer`

robert3005 (dismissed):

vortex-array/src/arrays/primitive/mod.rs
mwlon: :+1:

encodings/zstd/src/array.rs
mwlon: :+1:

encodings/zstd/src/array.rs
mwlon: :+1:

vortex-array/src/arrays/primitive/mod.rs
mwlon: Oh, I overlooked the `Mut`. Fixing this now

encodings/zstd/src/array.rs
mwlon: :+1:

encodings/zstd/src/array.rs
mwlon: :+1:

mwlon:
I think all the comments are addressed now

robert3005 (approved):

robert3005:
I resolved the one conflict that crept in in the meantime. Will merge as soon as CI passes

PR #3458, opened 2025-06-05
bump rust version v1.89.0-nightly

joseph-isaacs:


AdamGS:
Is there a specific reason to bump it? Every time we do move our msrv it makes it harder for people to use the Rust bindings

joseph-isaacs:
There is something strange about that specific rust version, its picked up differently on macOS 2025-02-21 and CI (linux) 2025-02-24, this leads to clippy differences

0ax1:
> Is there a specific reason to bump it? Every time we do move our msrv it makes it harder for people to use the Rust bindings

Given that we use a toolchain file, when is the actually an issue? This can only be a problem in case the bumped version is not supported on a user's system, right?

0ax1 (approved):
lgtm when clippy is happy

joseph-isaacs:
https://github.com/rust-lang/rust/issues/141492

a10y:
I tend to agree with Adam about bumping MSRV

PR #3461, opened 2025-06-05
feat: list_contains constant list, array value

joseph-isaacs:
Support `list_contains({[1,2,3]}, {1, 2, 4}) = {t,t,f}`

This is pretty slow, but that can be refined


vortex-array/src/compute/list.rs
```diff
@@ -31,41 +31,85 @@ use crate::{Array, ArrayRef, IntoArray, ToCanonical};
...
+        let res = fill_null(
+            &compare(
+                ConstantArray::new(element, len).as_ref(),
+                values,
+                Operator::Eq,
+            )?,
+            &false_scalar,
+        )?;
+        if let Some(acc) = result {
+            result = Some(boolean(&acc, &res, BooleanOperator::Or)?)
```
a10y: there is an `or` top-level function for this as well

vortex-array/src/compute/list.rs
```diff
@@ -31,41 +31,85 @@ use crate::{Array, ArrayRef, IntoArray, ToCanonical};
...
         )
         .to_array());
     }
 
+    let nullability = array.dtype().nullability() | value.dtype().nullability();
+
+    if let Some(value_scalar) = value.as_constant() {
+        list_contains_scalar(array, &value_scalar, nullability)
+    } else if let Some(list_scalar) = array.as_constant() {
+        list_scalar_contains_array(&list_scalar.as_list(), value, nullability)
```
a10y: so this is just an optimization for when the lhs happens to be constant? is that common?

vortex-array/src/compute/list.rs
```diff
@@ -31,41 +31,85 @@ use crate::{Array, ArrayRef, IntoArray, ToCanonical};
...
+    } else {
+        todo!("unsupported list contains with list and element as arrays")
+    }
+}
+
+fn list_scalar_contains_array(
```
a10y: the name confuses what this actually does. how about `constant_contains` and a short comment indicating that this is the codepath for when the LHS target array is constantarray

vortex-array/src/compute/list.rs
joseph-isaacs: Yep, this will be used for `row_id in [....]`

vortex-array/src/compute/list.rs
```diff
@@ -68,13 +68,14 @@ pub fn list_contains(array: &dyn Array, value: &dyn Array) -> VortexResult<Array
...
-        list_scalar_contains_array(&list_scalar.as_list(), value, nullability)
+        list_scalar_contains_constant_array(&list_scalar.as_list(), value, nullability)
     } else {
         todo!("unsupported list contains with list and element as arrays")
     }
 }
 
-fn list_scalar_contains_array(
+// Then there is a constant list scalar (haystack) being compared to an array of needles.
+fn list_scalar_contains_constant_array(
```
a10y: to me this still reads like "check if the list scalar contains a constanarray as a subarray" or something, but i'm not gonna block on this

a10y (approved):

vortex-array/src/compute/list.rs
robert3005: Wait, this is not what it means?

joseph-isaacs: happy to rename to anythingg

PR #3465, opened 2025-06-05
chore: split pruning.rs into four files.

danking:
1. field_or_identity.rs: just cut-and-paste.
2. relation.rs: cut-and-paste but then I added `map(&self) -> &HashMap...`
3. pruning_predicate.rs: I fixed some nits but also made this somewhat larger change:
```diff
@@ -95,26 +32,15 @@ impl PruningPredicate {
         let (expr, required_stats) = convert_to_pruning_expression(original_expr);
         if let Some(lexp) = expr.as_any().downcast_ref::<Literal>() {
             // Is the expression constant false, i.e. prune nothing
-            if lexp
-                .value()
-                .as_bool_opt()
-                .and_then(|b| b.value())
-                .map(|b| !b)
-                .unwrap_or(false)
-            {
-                None
-            } else {
-                Some(Self {
-                    expr,
-                    required_stats,
-                })
+            if lexp.value().as_bool_opt().and_then(|b| b.value()) == Some(false) {
+                return None;
             }
-        } else {
-            Some(Self {
-                expr,
-                required_stats,
-            })
         }
+
+        Some(Self {
+            expr,
+            required_stats,
+        })
     }
```
4. pruning_predicate_rewriter.rs: cut-and-paste but I lifted a type definition to the top: `type PruningPredicateStats = (ExprRef, Relation<FieldOrIdentity, Stat>);`


vortex-expr/src/pruning/relation.rs
```diff
@@ -0,0 +1,66 @@
+use std::fmt::Display;
+use std::hash::Hash;
+
+use itertools::Itertools as _;
+use vortex_array::aliases::hash_map::HashMap;
+use vortex_array::aliases::hash_set::HashSet;
+
+#[derive(Debug, Clone)]
+pub struct Relation<K, V> {
```
a10y: Relation can be pub(crate) or pub(super) or something not fully pub

vortex-expr/src/pruning/relation.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            self.map.iter().format_with(",", |(k, v), fmt| {
+                fmt(&format_args!("{k}: {{{}}}", v.iter().format(",")))
+            })
+        )
+    }
+}
```
a10y: I don't think we need this Display impl, if we kill the PruningPredicate display impl

vortex-expr/src/pruning/relation.rs
```diff
@@ -0,0 +1,66 @@
...
+
+impl<K: Hash + Eq, V: Hash + Eq> Relation<K, V> {
+    pub fn new() -> Self {
+        Relation {
+            map: HashMap::new(),
+        }
+    }
+
+    pub fn union(mut iter: impl Iterator<Item = Relation<K, V>>) -> Relation<K, V> {
```
a10y: this method is not used anywhere and can be deleted

vortex-expr/src/pruning/pruning_predicate.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl Display for PruningPredicate {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "PruningPredicate({}, {{{}}})",
+            self.expr, self.required_stats
+        )
+    }
+}
```
a10y: I think we should probably kill this, seems like it's only used for tests and logging, in which case no reason we can't use the Debug impl

vortex-expr/src/pruning/pruning_predicate.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub fn required_stats(&self) -> &HashMap<FieldOrIdentity, HashSet<Stat>> {
+        self.required_stats.map()
+    }
+
+    /// Evaluate this predicate against a per-chunk statistics table.
+    ///
+    /// Returns Ok(None) if any of the required statistics are not present in metadata.
+    /// If it returns Ok(Some(array)), the array is a boolean array with the same length as the
+    /// metadata, and a true value means the chunk _can_ be pruned.
+    pub fn evaluate(&self, metadata: &dyn Array) -> VortexResult<Option<ArrayRef>> {
```
a10y, line 47: this is not relevant for this PR, but i'm wondering if the control flow is inverted.

I kind of thing that a thing that is prunable should receive a PruningPredicate, and then do something with it, rather than forcing the PruningPredicate to receive this type-erased thing that we pretty-promise looks like a stats table defined in the other crate

a10y (commented):
I was on my own branch messing around with this module, just porting some comments over based on things I found

vortex-expr/src/pruning/relation.rs
danking: Done.

vortex-expr/src/pruning/relation.rs
danking: Done.

vortex-expr/src/pruning/relation.rs
danking: Done.

vortex-expr/src/pruning/pruning_predicate.rs
danking: Done.

vortex-expr/src/pruning/pruning_predicate.rs
danking: Agreed.

vortex-duckdb-ext/src/cpp.rs
```diff
@@ -333,6 +331,19 @@ pub type __darwin_pthread_once_t = _opaque_pthread_once_t;
 pub type __darwin_pthread_rwlock_t = _opaque_pthread_rwlock_t;
 pub type __darwin_pthread_rwlockattr_t = _opaque_pthread_rwlockattr_t;
 pub type __darwin_pthread_t = *mut _opaque_pthread_t;
+pub type u_int8_t = ::std::os::raw::c_uchar;
```
a10y: should we revert these?

a10y (approved):
not sure about those vortex-duckdb-ext changes but the rest is gtg

PR #3466, opened 2025-06-05
Minor touchups to vortex-expr

AdamGS:
There's a lot more work to do there, but I think that's a start.


danking (approved):

danking:
Changing expr adjacent things so I didn't want to hit any conflicts! Sorry if I jumped the gun Joe.

PR #3467, opened 2025-06-05
chore: check-in vortex-duckdb-ext generated code

danking:


a10y (approved):
image

vortex-duckdb-ext/src/cpp.rs
```diff
@@ -333,6 +331,19 @@ pub type __darwin_pthread_once_t = _opaque_pthread_once_t;
 pub type __darwin_pthread_rwlock_t = _opaque_pthread_rwlock_t;
 pub type __darwin_pthread_rwlockattr_t = _opaque_pthread_rwlockattr_t;
 pub type __darwin_pthread_t = *mut _opaque_pthread_t;
+pub type u_int8_t = ::std::os::raw::c_uchar;
```
0ax1, line 334: Not sure why this is happening, but bindgen seems to be flip-flopping between these 2 versions.

danking: 🤦

PR #3469, opened 2025-06-06
fix: ArrowUtils conversion from Arrow decimal -> Spark decimal

a10y:


a10y:
Tested this via Iceberg and it fixes the issue.

robert3005 (approved):

PR #3471, opened 2025-06-06
feat[vortex-expr]: add an expr analysis trait and use it to impl pruning

joseph-isaacs:
Cleaned up pruning
Supports nested struct

todo:
Update the FieldPath to contain an identifier and can go from (identifier, fieldpath) -> fieldpath


vortex-expr/src/binary.rs
```diff
@@ -82,6 +83,66 @@ pub(crate) mod proto {
...
+            Operator::Eq => {
+                let min_lhs = self.lhs.min(catalog);
+                let max_lhs = self.lhs.max(catalog);
+
+                let min_rhs = self.rhs.min(catalog);
+                let max_rhs = self.rhs.max(catalog);
+
+                let left = min_lhs.zip_with(max_rhs, gt);
+                let right = min_rhs.zip_with(max_lhs, gt);
+                left.into_iter().chain(right).reduce(or)
```
danking, line 97: Hmm. Perhaps I'm missing something subtle but I find the zip_withs and chains and reduces harder to read than ~~MONADIC DO NOTATION~~ question marks:

```suggestion
                let min_lhs = self.lhs.min(catalog)?;
                let max_lhs = self.lhs.max(catalog)?;

                let min_rhs = self.rhs.min(catalog)?;
                let max_rhs = self.rhs.max(catalog)?;

                or(gt(min_lhs, max_rhs), gt(min_rhs, max_lhs))
```

vortex-expr/src/binary.rs
```diff
@@ -82,6 +83,66 @@ pub(crate) mod proto {
     }
 }
 
+impl AnalysisExpr for BinaryExpr {
+    fn prune_expr(&self, catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```
danking: I really dislike this name. I know it's what we were calling it before, but I feel like it doesn't convey well what is happening.

I like `falsification` or `falsifiable_expr` (Edara & Pasumansky "Big Metadata" call these "falsifiable expressions").

vortex-expr/src/binary.rs
```diff
@@ -82,6 +83,66 @@ pub(crate) mod proto {
...
+            }),
+            Operator::Lt | Operator::Lte => self.lhs.min(catalog).and_then(|min| {
+                self.rhs.max(catalog).map(|max| {
+                    let inverse = self
+                        .operator
+                        .inverse()
+                        .vortex_expect("inverse of le & lte defined");
+                    BinaryExpr::new_expr(min, inverse, max)
+                })
+            }),
```
danking:
```suggestion
            Operator::Lt  => gte(self.lhs.min(catalog)?, self.rhs.max(catalog)?),
            Operator::Lte => gt(self.lhs.min(catalog)?, self.rhs.max(catalog)?),
```

I know this is mostly a copy from the old code but the noise of the inverse and vortex_expect hurts legibility, imo pretty badly.

vortex-expr/src/binary.rs
```diff
@@ -82,6 +83,66 @@ pub(crate) mod proto {
...
+            // We can short circuit pruning expr for and
+            Operator::And => self
+                .lhs
+                .prune_expr(catalog)
+                .into_iter()
+                .chain(self.rhs.prune_expr(catalog))
+                .reduce(or),
```
danking, line 118: To my point about naming, I think this reads a lot more clearly with a name like `falsification`:
```
or(
  self.lhs.falsification(catalog),
  self.rhs.falsification(catalog),
)
```
Clearly if you can falsify either the lhs or the rhs, you falsify the and.

vortex-expr/src/pruning/pruning_predicate.rs
```diff
@@ -17,14 +21,33 @@ pub struct PruningPredicate {
...
     pub fn try_new(original_expr: &ExprRef) -> Option<Self> {
-        let (expr, required_stats) = convert_to_pruning_expression(original_expr);
+        let (expr, required_stats) = pruning_expr(original_expr)?;
+
         if let Some(lexp) = expr.as_any().downcast_ref::<Literal>() {
             // Is the expression constant false, i.e. prune nothing
             if lexp.value().as_bool_opt().and_then(|b| b.value()) == Some(false) {
                 return None;
             }
         }
```
danking: I think we should delete this block. Constant false was used to represent what you now do with `None`.

If we are worried about trivial pruning expressions in general, I think we should first `simplify` and then check for literal false.

vortex-expr/src/pruning/pruning_predicate.rs
```diff
@@ -67,3 +90,279 @@ impl PruningPredicate {
...
+
+        // Expected: a_max <= 10 OR a_min >= 50
+        assert_eq!(
+            &predicate,
+            &or(
+                lt_eq(get_item_scope(FieldName::from("a_max")), lit(10)),
+                gt_eq(get_item_scope(FieldName::from("a_min")), lit(50))
+            ),
+        );
+    }
```
danking, line 359:
```suggestion
    }

    fn test_gt_eq_with_booleans() {
        // Consider this unusual, but valid (in Arrow, BooleanArray implements ArrayOrd), filter expression:
        //
        // x > (y > z)
        //
        // The x column is a Boolean-valued column. The y and z columns are numeric. True > False.
        // Suppose we had a Vortex zone whose min/max statistics for each column were:
        //
        // x: [True, True]
        // y: [1, 2]
        // z: [0, 2]
        //
        // The pruning predicate will convert the aforementioned expression into:
        //
        // x_max <= (y_min > z_min)
        //
        // If we evaluate that pruning expression on our zone we get:
        //
        // x_max <= (y_min > z_min)
        // x_max <= (1     > 0    )
        // x_max <= True
        // True <= True
        // True
        //
        // If a pruning predicate evaluates to true then, as stated in PruningPredicate::evaluate:
        //
        // > a true value means the chunk can be pruned.
        //
        // But, the following record lies within the above intervals and *passes* the filter expression! We
        // cannot prune this zone because we need this record!
        //
        // {x: True, y: 1, z: 2}
        //
        // x > (y > z)
        // True > (1 > 2)
        // True > False
        // True
        let expr = gt_eq(col("x"), gt(col("y"), col("z")));
        assert!(PruningPredicate::try_new(&expr).is_none());
        // TODO(DK): a sufficiently complex pruner would produce: `x_max <= (y_max > z_min)`
    }
```

vortex-expr/src/pruning/mod.rs
```diff
@@ -1,295 +1,39 @@
...
+        None
     }
+}
 
-    #[test]
-    fn unprojectable_expr() {
-        let or_expr = not(lt(get_item_scope("a"), get_item_scope("b")));
-        assert!(PruningPredicate::try_new(&or_expr).is_none());
+pub trait AnalysisExpr {
+    fn prune_expr(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```
danking: I think these methods need doc strings because all this stuff is pretty theoretically complex! This is what I used in my PR:

```suggestion
    /// An expression over zone-statistics which is true if-and-only-if `expr` is false for all
    /// records in the zone.
    ///
    /// # Examples
    ///
    /// - An expression over one variable: `x > 0` is false for all records in a zone if the maximum
    ///   value of the column `x` in that zone is less than or equal to zero: `max(x) <= 0`.
    /// - An expression over two variables: `x > y` becomes `max(x) <= min(y)`.
    /// - A conjunctive expression: `x > y AND z < x` becomes `max(x) <= min(y) OR min(z) >= max(x).
    ///
    /// Some expressions, in theory, have falsifications but this function does not support them
    /// such as `x < (y < z)` or `x LIKE "needle%"`.
    fn prune_expr(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```

vortex-expr/src/pruning/mod.rs
```diff
@@ -1,295 +1,39 @@
...
-        let expected = HashMap::from([(
-            FieldOrIdentity::from("a"),
-            HashSet::from([Stat::Min, Stat::Max]),
-        )]);
-
-        assert_eq!(
-            PruningPredicate::try_new(&expr).unwrap().required_stats(),
-            &expected
-        );
+    fn max(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```
danking: I think max is the best name we've got, but I think we should clarify in the doc string that this is merely an upper bound. It's neither the least upper bound nor necessarily a value in the column (e.g. truncated string max).

```suggestion
    /// If an expression is returned, its value is an upper bound on the value of `expr`.
    ///
    /// We may return `None` for values which have no upper bound or values for which knowing the
    /// upper bound is difficult.
    fn max(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```

vortex-expr/src/pruning/mod.rs
```diff
@@ -1,295 +1,39 @@
...
-        let expected = HashMap::from([(
-            FieldOrIdentity::from("a"),
-            HashSet::from([Stat::Min, Stat::Max]),
-        )]);
-
-        assert_eq!(
-            PruningPredicate::try_new(&expr).unwrap().required_stats(),
-            &expected
-        );
+    fn min(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```
danking: This is what I wrote for min 🤷 

```suggestion
    /// If an expression is returned, its value is a lower bound on the value of `expr`.
    ///
    /// We may return `None` for values which have no lower bound or values for which knowing the
    /// lower bound is difficult.
    fn min(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```

danking (changes requested):
This looks great. Much much better than what we had. I am mostly picking stylistic nits that I think make the code a bit more obviously correct to the reader. I also think this needs careful doc strings because it's such subtle stuff! I copied over suggestions from my PR but happy to go with something of your own creation too!

vortex-expr/src/pruning/mod.rs
```diff
@@ -1,295 +1,39 @@
...
+    fn stats_ref(&mut self, _id: &Identifier, _field: &FieldPath, _stat: Stat) -> Option<ExprRef> {
+        None
     }
+}
 
-    #[test]
-    fn unprojectable_expr() {
-        let or_expr = not(lt(get_item_scope("a"), get_item_scope("b")));
-        assert!(PruningPredicate::try_new(&or_expr).is_none());
+pub trait AnalysisExpr {
```
danking: On the subject of names, should this be `FalsifiableExpr`? or something to that effect?

vortex-expr/src/binary.rs
joseph-isaacs: The zip and reduce will find more possible exprs, (in theory), another valid pruning expr is `gt(min_lhs, max_rhs)` if the rhs cannot be created

vortex-expr/src/binary.rs
joseph-isaacs: `falsification` is vague, it not clear what that means either, but I am not against it

vortex-expr/src/pruning/mod.rs
joseph-isaacs: I want this to work for thing other than pruning (like I would even add simplify here).

joseph-isaacs: I am not sure if we would want to split that into more traits but that add the complexity of adding a new expr

joseph-isaacs: Maybe I should move this trait into the root crate?

vortex-expr/src/binary.rs
danking: Hmm. I want a name that conveys that we are (fuzzily) negating this expression. Given an expression `e`, we give you expression `e2` such that `e2 implies (not e)`. ChatGPT suggests:
- exclusion expression
- sufficient condition for negation
- sufficient for not
- refutation

I'm not sure I like any of those better than `falsification`.

Edara and Pasmunasky describe it like this in section 5.2:

> Our approach is based on the notion of a falsifiable expression which is a boolean expression derived from the query filters. It satisfies the following property: _For a given block, if the falsifiable expression evaluates to true, that block does not need to be scanned by the original query._

vortex-expr/src/binary.rs
danking: Ah right. Grr. Man that's pretty subtle. OK.

vortex-expr/src/binary.rs
```diff
@@ -82,6 +83,66 @@ pub(crate) mod proto {
...
+            Operator::NotEq => {
+                let min_lhs = self.lhs.min(catalog);
+                let max_lhs = self.lhs.max(catalog);
+
+                let min_rhs = self.rhs.min(catalog);
+                let max_rhs = self.rhs.max(catalog);
+
+                let left = min_lhs.zip_with(max_rhs, eq);
+                let right = max_lhs.zip_with(min_rhs, eq);
+                left.into_iter().chain(right).reduce(and)
```
danking: Now that I see what the reduce is doing, I think this is wrong. Just because the right-hand-side doesn't have a min or max doesn't mean that we can use `left` alone. Specifically, `min(lhs) == max(rhs)`, alone, is not sufficient to conclude that `lhs != rhs`. For example: $lhs \in [1, 2], rhs \in [0, 1]$

danking: I think, except for the `or` cases, we should use question mark notation both for the correct semantics (we need both sides of the and) and for legibility.

danking (changes requested):

vortex-expr/src/binary.rs
joseph-isaacs: maybe `stat_falsification`? I would be happy with `falsification` but it might clash with other expr analysis that do falsification not using stats

vortex-expr/src/binary.rs
joseph-isaacs: If we have try2, it would be nicer

vortex-expr/src/pruning/pruning_predicate.rs
```diff
@@ -67,3 +90,279 @@ impl PruningPredicate {
...
+        expr = get_item(name, expr);
+        self.usage
+            .insert((id.clone(), field.clone(), stat), expr.clone());
+        Some(expr)
+    }
+}
+
+#[allow(clippy::type_complexity)]
+// TODO: remove (Id, FieldPath) when updating FieldPath
+pub fn pruning_expr(expr: &ExprRef) -> Option<(ExprRef, Relation<(Identifier, FieldPath), Stat>)> {
```
a10y, line 112: dear god this return type

vortex-expr/src/binary.rs
joseph-isaacs: correct, this is wrongg

vortex-expr/src/pruning/pruning_predicate.rs
joseph-isaacs: its gettingg better

joseph-isaacs: I have a follow up that create an access path = (Identifier, FieldPath)

danking (approved):
v. nice.

PR #3474, opened 2025-06-06
Use foldhash for counting instances

AdamGS:
`Itertools::counts()` uses the default std hasher which we know is worse, this just inlines it and uses foldhash from hashbrown.

### DIFF
--- a/encodings/alp/src/alp_rd/mod.rs
+++ b/encodings/alp/src/alp_rd/mod.rs
@@ -346,11 +346,12 @@ fn build_left_parts_dictionary<T: ALPRDFloat>(
     );
 
     // Count the number of occurrences of each left bit pattern
-    let counts = samples
+    let mut counts = HashMap::new();
+    samples
         .iter()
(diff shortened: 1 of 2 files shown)


encodings/alp/src/alp_rd/mod.rs
```diff
@@ -346,11 +346,12 @@ fn build_left_parts_dictionary<T: ALPRDFloat>(
     );
 
     // Count the number of occurrences of each left bit pattern
-    let counts = samples
+    let mut counts = HashMap::new();
```
a10y, line 349: nice. did you find this in profiling or just noticed it? should we add it to our clippy as a disallowed-methods https://doc.rust-lang.org/clippy/lint_configuration.html#disallowed-methods ?

AdamGS: I'll add it, got it from profiling a benchmark of some synthetic data

a10y (approved):

PR #3476, opened 2025-06-06
Add an `AccessPath` (Identity + FieldPath)

joseph-isaacs:


vortex-expr/src/lib.rs
```diff
@@ -156,6 +156,45 @@ impl VortexExprExt for ExprRef {
     }
 }
 
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct AccessPath {
```
danking, line 159: I'm glad you found this name. I've had a heckin' time thinking of a good name for this back when I added FIeldOrIdentity.

vortex-expr/src/pruning/pruning_predicate.rs
```diff
@@ -59,80 +40,93 @@ impl PruningPredicate {
...
-            .flat_map(|(key, value)| value.iter().map(|stat| key.stat_field_name_string(*stat)))
-            .collect::<HashSet<_>>();
+            .flat_map(|(path, stats)| stats.iter().map(|s| (path.clone(), s.name().into())))
+            .collect::<HashSet<(AccessPath, FieldName)>>();
+
         let missing_stats = required_stats.difference(&known_stats).collect::<Vec<_>>();
 
         if !missing_stats.is_empty() {
             return Ok(None);
         }
```
danking, line 72: I guess this is a nice check but it feels like we don't need any of this any more. The code should be written to take whatever stats we have and pass it to the pruner. Can be follow up ofc.

vortex-expr/src/pruning/pruning_predicate.rs
```diff
@@ -59,80 +40,93 @@ impl PruningPredicate {
...
+        let name = access_path_field_name(access_path, stat);
         expr = get_item(name, expr);
-        self.usage
-            .insert((id.clone(), field.clone(), stat), expr.clone());
+        self.usage.insert((access_path.clone(), stat), expr.clone());
         Some(expr)
     }
 }
 
+pub fn access_path_field_name(access_path: &AccessPath, stat: Stat) -> FieldName {
```
danking: This is like the access_path_stat_field_name, right?

danking (approved):
LGTM, a couple nits, take or leave them.

vortex-expr/src/pruning/pruning_predicate.rs
joseph-isaacs: That is a good idea, Ill add a TODO.

PR #3477, opened 2025-06-06
feat: change values and vars to take &str

danking:
It is annoying to have to construct an Arc just to get the root identifier.


AdamGS:
Can you wait just a second with this? I have something that might make it a bit nicer 🤞

joseph-isaacs:
I was going to change the Identifier to be static arced

danking:
Sure. @AdamGS Can you fix ScopeDType as well? I just hit that one too.

PR #3478, opened 2025-06-06
feat: Make `Identifier` an enum

AdamGS:
Using an empty string as a sentinel value for the "identity" identifier seems bug-prune, and the API isn't very ergonomic.

The naming in this PR is a bit atrocious, but I'm more than open to suggestions.

Also added some comments that made sense to me but might not be 100% correct or clear.

### DIFF
--- a/vortex-expr/src/scope.rs
+++ b/vortex-expr/src/scope.rs
@@ -6,18 +6,49 @@ use vortex_array::{Array, ArrayRef};
 use vortex_dtype::DType;
 use vortex_error::{VortexResult, vortex_err};
 
-use crate::IDENTITY_IDENTIFIER;
-
-pub type Identifier = Arc<str>;
 type ExprScope<T> = HashMap<Identifier, T>;
 
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub enum Identifier {
+    Identity,
+    Other(Arc<str>),
+}
+
+impl From<String> for Identifier {
+    fn from(value: String) -> Self {
+        Self::Other(value.into())
+    }
+}
+
+impl From<&str> for Identifier {
+    fn from(value: &str) -> Self {
+        Self::Other(value.into())
+    }
+}
+
+impl Identifier {
+    pub fn is_identity(&self) -> bool {
+        matches!(self, Self::Identity)
+    }
+}
+
+impl std::fmt::Display for Identifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Identifier::Identity => write!(f, "$"),
+            Identifier::Other(v) => write!(f, "{}", v),
+        }
+    }
+}
+
+/// The evaluation scope for an expression, all variables are evaluated relatively to the data it contains.
+/// It allows for expressions to access fields and previously defined data by [`Identifier`].
 #[derive(Clone, Default)]
 pub struct Scope {
     array_len: usize,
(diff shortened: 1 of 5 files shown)


vortex-proto/proto/expr.proto
```diff
@@ -51,7 +51,18 @@ message Kind {
...
   message Var {
-    string var = 1;
+    Identifier ident = 1;
+  }
+
+  message Identifier {
+    oneof kind {
+      Identity identity = 1;
+      string other = 2;
+    }
```
joseph-isaacs: this is not good

AdamGS: elaborate?

vortex-expr/src/scope.rs
```diff
@@ -6,18 +6,49 @@ use vortex_array::{Array, ArrayRef};
...
 
-use crate::IDENTITY_IDENTIFIER;
-
-pub type Identifier = Arc<str>;
 type ExprScope<T> = HashMap<Identifier, T>;
 
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub enum Identifier {
+    Identity,
+    Other(Arc<str>),
```
danking, line 15: There's also `IdentityOrField` in the pruning system which is maybe isomorphic to this. Should this just be a FieldPath? @joseph-isaacs are you working on something like that?

vortex-expr/src/scope.rs
```diff
@@ -6,18 +6,49 @@ use vortex_array::{Array, ArrayRef};
...
+}
+
+impl From<String> for Identifier {
+    fn from(value: String) -> Self {
+        Self::Other(value.into())
```
joseph-isaacs: this might clash with ""

joseph-isaacs (changes requested):
We this not safe as is

vortex-expr/src/scope.rs
joseph-isaacs: https://github.com/vortex-data/vortex/pull/3476

a10y (commented):
I agree this is better than using empty string

vortex-expr/src/scope.rs
AdamGS: I'll change that into `TryFrom` or `FromStr` and not allow the empty string

vortex-proto/proto/expr.proto
joseph-isaacs: It just seems strange to expose this outside of vortex internals

AdamGS: Empty strings are problematic with protobuf because its a default value and some bugs will just make it default to that. IDK if there's a better way to express that in proto but within my somewhat limited experience with it it seems reasonable.

vortex-expr/src/scope.rs
```diff
@@ -1,23 +1,55 @@
...
+impl Identifier {
+    pub fn is_identity(&self) -> bool {
+        matches!(self, Self::Identity)
+    }
+}
+
+impl std::fmt::Display for Identifier {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Identifier::Identity => write!(f, "$"),
```
AdamGS: This makes some tests fail because it expects it to display an empty string, I guess keeping that behavior is acceptable because we don't allow empty string identifiers.

joseph-isaacs (approved):
NIce thanks

PR #3479, opened 2025-06-06
nit: AnalysisExpr docs are not quite true

danking:
It is not and if-and-only-if, the implication only goes one way.


joseph-isaacs (approved):

AdamGS:
@danking I can fix your label but you have to sign for your commits (👀?)

PR #3480, opened 2025-06-06
Use set instead of map to count distinct values in float compressor

AdamGS:
Just seems like the values aren't actually used anywhere and its extra work that popped out while profiling some benchmark


a10y:
So we actually might want to keep it as HashMap, the counts are used to determine if there's a dominant value to then use SparseScheme. I think right now we don't try and do Sparse directly on floats, we assume that it needs to go through Int first, but maybe we should consider coming back around and just implementing Sparse encoding for floats

a10y:
Meh, we can change it to HashSet for now and change it back later if we come around to implementing Sparse, it's a small change anyway

a10y (approved):

PR #3482, opened 2025-06-07
chore: Fix comments that could be rustdoc

AdamGS:
I was hoping to find a regex that can find those reliably, but there are too many false positives. For the record the last ripgrep command I used was;
```bash
rg -U '^\t*// [A-Za-z].*\n\t*(fn|pub|struct|trait|enum|type)' -g '*.rs'
```


gatesn:
Honestly, it may also be worth a pass with an LLM to add comments and enable deny(missing_docs)

AdamGS:
On what scope? Seems like vortex-expr breaks like every day and a lot of other things are in flux

robert3005 (approved):

Issue #2254, opened 2025-02-06
Epic: Apache DataFusion Integration


AdamGS:
## Vortex Issues
- [x] https://github.com/vortex-data/vortex/issues/6269
- [x] #1507 
- [x] Output ordering support (reverting #2429) 
- [ ] Respecting more global configurations (or explicitly error if they are used), and making more vortex-specific things configurable.
    - [x] `FileSinkConfig::table_partition_cols` and `FileSinkConfig::keep_partition_by_columns` (Not sure if that's still relevant)
- [x] Once released in DataFusion, support decimal32/64 conversion to Arrow. https://github.com/vortex-data/vortex/issues/4668
- [ ] Support Dynamic filter expressions https://github.com/vortex-data/vortex/issues/4034
- [ ] Metadata table function - https://github.com/vortex-data/vortex/issues/5144

## DataFusion issues we filed
- https://github.com/apache/datafusion/issues/15809
- https://github.com/apache/datafusion/issues/13720 (I think `FileMetadataCache` might've solved this issue, need to make sure)
- https://github.com/apache/datafusion/issues/13325 (Potentially solved)
- https://github.com/apache/datafusion/issues/20078
- https://github.com/apache/datafusion/pull/20111

## DataFusion issues we're interested in
- https://github.com/apache/datafusion/issues/14993
- https://github.com/apache/datafusion/issues/20135

## Related
- [x] If there's willingness, support Vortex in datafusion-dft to have an easy demo https://github.com/datafusion-contrib/datafusion-dft/issues/336

## Older solved issues
- [x] Migrating from `DataSink` to `FileSink` once the next major release (45) goes out.
- [x] `FileScanConfig::limit`
- [x] `FileScanConfig::table_partition_cols`

haohuaijin:
this https://github.com/vortex-data/vortex/discussions/5481 maybe relate

hengfeiyang:
Any plan to separate datafusion feature to a separate repo? because any datafusion change need to waiting a new version from vortex. that is a long way.

robert3005:
I think the ideal long term solution is that vortex-DataFusion lives in DataFusion but right now apis aren’t stable enough for that to be possible. We have started the work on coming up with a abstraction that would allow us to avoid exposing whole api surface to DataFusion users

hengfeiyang:
Yeah, i also think so that will be great. Thank you @robert3005 👍

PR #3483, opened 2025-06-07
feat: Initial limit pushdown in the scan

### LINKED ISSUE #2593: Limit push-down for scanning
Most "table provider" interfaces, including DataFusion, Arrow Dataset, Polars, etc all support pushing down limits.

AdamGS:
Figured that `SELECT * FROM tbl LIMIT 10` is a pretty common use case for data exploration and is probably worth including. I think this closes #2593 and is also part of #2254.

Supporting limit with filters with our heavy use of `async move` seems much harder because the limit has to be shared in some way. Doing that for the stream (like when `into_stream`) is trivial with a `StreamExt::scan` but for the current return type of `ScanBuilder::build` that will require shared state and I'm not even sure that would be correct given the ordering isn't guaranteed.

### DIFF
--- a/vortex-datafusion/src/lib.rs
+++ b/vortex-datafusion/src/lib.rs
@@ -1,4 +1,4 @@
-//! Connectors to enable DataFusion to read Vortex data.
+//! Connectors to enable [DataFusion](https://docs.rs/datafusion/latest/datafusion/) to read [`Vortex`](https://docs.rs/crate/vortex/latest) data.
 #![deny(missing_docs)]
 use std::fmt::Debug;
 
@@ -8,7 +8,9 @@ use datafusion::logical_expr::{Expr, Operator};
 use vortex::stats::Precision;
 
 mod convert;
-pub mod persistent;
+mod persistent;
+
+pub use persistent::*;
 
 const SUPPORTED_BINARY_OPS: &[Operator] = &[
     Operator::Eq,
--- a/vortex-datafusion/src/persistent/mod.rs
+++ b/vortex-datafusion/src/persistent/mod.rs
@@ -35,6 +35,7 @@ mod tests {
         ListingOptions, ListingTable, ListingTableConfig, ListingTableUrl,
     };
     use datafusion::prelude::SessionContext;
+    use rstest::rstest;
     use tempfile::tempdir;
     use tokio::fs::OpenOptions;
     use vortex::IntoArray;
@@ -46,8 +47,11 @@ mod tests {
 
     use crate::persistent::VortexFormat;
 
+    #[rstest]
+    #[case(Some(1))]
+    #[case(None)]
     #[tokio::test]
-    async fn query_file() -> anyhow::Result<()> {
+    async fn query_file(#[case] limit: Option<usize>) -> anyhow::Result<()> {
         let temp_dir = tempdir()?;
         let strings = ChunkedArray::from_iter([
             VarBinArray::from(vec!["ab", "foo", "bar", "baz"]).into_array(),
@@ -99,8 +103,17 @@ mod tests {
         let listing_table = Arc::new(ListingTable::try_new(config)?);
 
         ctx.register_table("vortex_tbl", listing_table as _)?;
-        let row_count = ctx.table("vortex_tbl").await?.count().await?;
-        assert_eq!(row_count, 8);
+        let total_row_count = ctx.table("vortex_tbl").await?.count().await?;
+        assert_eq!(total_row_count, 8);
+
+        let row_count = ctx
+            .table("vortex_tbl")
+            .await?
+            .limit(0, limit)?
+            .count()
+            .await?;
+
+        assert_eq!(row_count, limit.unwrap_or(total_row_count));
 
         Ok(())
     }
--- a/vortex-layout/src/scan/mod.rs
+++ b/vortex-layout/src/scan/mod.rs
@@ -159,6 +159,12 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
         if self.filter.is_some() && self.limit.is_some() {
             vortex_bail!("Vortex doesn't support scans with both a filter and a limit")
         }
+
+        // The ultimate short circuit
+        if self.limit.is_some_and(|l| l == 0) {
+            return Ok(Vec::new());
+        }
+
         // Spin up the root layout reader, and wrap it in a FilterLayoutReader to perform
         // conjunction splitting if a filter is provided.
         let layout_reader = if self.filter.is_some() {
--- a/vortex-layout/src/scan/mod.rs
+++ b/vortex-layout/src/scan/mod.rs
@@ -186,7 +186,7 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
         // Create a task that executes the full scan pipeline for each split.
         let split_tasks = splits
             .into_iter()
-            .map(|split_range| {
+            .filter_map(|split_range| {
                 let ctx = Arc::new(TaskContext {
                     row_range: self.row_range.clone(),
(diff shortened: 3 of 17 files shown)


vortex-datafusion/src/lib.rs
```diff
@@ -10,6 +10,8 @@ use vortex::stats::Precision;
 mod convert;
 pub mod persistent;
 
+pub use persistent::*;
```
AdamGS, line 13: Now that's the only module in the repo, figured I'll re export everything there.

vortex-mask/src/lib.rs
```diff
@@ -509,6 +509,38 @@ impl Mask {
             _ => None,
         }
     }
+
+    /// Limit the mast to the first `limit` true values
```
gatesn: Typo: mask

vortex-mask/src/lib.rs
```diff
@@ -509,6 +509,38 @@ impl Mask {
...
+                Self::from_iter([Self::new_true(limit), Self::new_false(len - limit)])
+            }
+            Mask::AllFalse(_) => self,
+            Mask::Values(mask_values) => {
+                let existing_buffer = mask_values.boolean_buffer();
+                let mut limit = limit;
+                let mut new_buffer_builder = BooleanBufferBuilder::new(mask_values.len());
+                new_buffer_builder.append_n(existing_buffer.len(), false);
+
+                for index in existing_buffer.set_indices() {
```
gatesn: .iter().take(n)?

vortex-layout/src/scan/mod.rs
```diff
@@ -258,6 +266,11 @@ impl<A: 'static + Send> ScanBuilder<A> {
                     if mask.all_false() {
                         Ok(None)
                     } else {
+                        // If there's a limit, we only apply it at the end
+                        let mut mask = mask;
+                        if let Some(limit) = self.limit {
```
gatesn: This applies the same limit to each row split?

gatesn (commented):
Feels like you're missing the bit where you can avoid constructing scan tasks entirely when there is a limit and no filter expression? 

This avoids traversing the layout tree and also prefetching unnecessary segments.

vortex-layout/src/scan/mod.rs
AdamGS: that's true 🤦

gatesn: If there isn't a filter expression, then the masks should be limited at the very beginning of the scan builder (this would automatically prune splits that appeared after the limit)

If there _is_ a filter expression, then this is a reasonable approach! Apply the limit to each split's initial row mask independently. This would create an upper bound. But you also then need to add a StreamExt::scan to the end of the array stream to limit the result.

AdamGS:
I was trying to figure that part out and it wasn't immediately obvious to me

vortex-layout/src/scan/mod.rs
AdamGS: I was really hoping to avoid `StreamExt::scan` 😂

AdamGS: I need to think about it some more, maybe tomorrow/monday when I'm sharper

AdamGS: ~I have something but that I'm not very happy with but I think is correct? I'll push it so its visible but I feel like its the sort of thing that is just going to be obvious to me on Monday morning.~ Its not even correct, realized that the limit was just being copied around in a bunch of places which is obviously wrong

vortex-datafusion/src/persistent/mod.rs
```diff
@@ -99,8 +103,17 @@ mod tests {
...
         ctx.register_table("vortex_tbl", listing_table as _)?;
-        let row_count = ctx.table("vortex_tbl").await?.count().await?;
-        assert_eq!(row_count, 8);
+        let total_row_count = ctx.table("vortex_tbl").await?.count().await?;
+        assert_eq!(total_row_count, 8);
+
+        let row_count = ctx
+            .table("vortex_tbl")
+            .await?
+            .limit(0, limit)?
```
onursatici, line 112: I know this PR is not finished, but it is surprising to me that it passes this test given the limit is not used in scan builder at all

AdamGS: Seems like `DataSourceExec` which is the physical plan node for `DataSource`'s takes care of it on its own, even if the underlying thing doesn't respect the config.

AdamGS: I guess that does mean that there's a minimal version of this PR which is just allowing that in datafusion, but I still want to figure it out in the scan because its useful for many other cases AND will probably make the datafusion integration better

AdamGS: I happens [here](https://github.com/apache/datafusion/blob/54d85fc27796771533ce946c07a692cc0a6bb581/datafusion/datasource/src/file_stream.rs#L228), the config is pushed both to the underlying source and to the stream that drives it.

vortex-layout/src/scan/tasks.rs
```diff
@@ -55,7 +56,18 @@ pub(super) fn split_exec<A: 'static + Send + Sync>(
...
         // No filter == immediate task
-        None => ok(row_mask).boxed(),
+        None => {
+            let row_mask = if let Some(limit) = limit.filter(|l| **l > 0_usize) {
+                let true_count = row_mask.true_count();
+                let row_mask = row_mask.limit(*limit);
+                *limit -= usize::min(*limit, true_count);
+                row_mask
+            } else {
+                row_mask
```
onursatici: Shouldn't we return all false here if limit == 0, but return row_mask if limit is None. Feels like we need one more branch

vortex-layout/src/scan/tasks.rs
```diff
@@ -55,7 +56,18 @@ pub(super) fn split_exec<A: 'static + Send + Sync>(
 
     let filter = match ctx.filter.as_ref() {
         // No filter == immediate task
-        None => ok(row_mask).boxed(),
+        None => {
```
a10y, line 60: should we be doing an early return here or something? like if the remaining limit is 0, we shouldn't build the projection_eval (which spawns some prefetching of the row range I think?)

vortex-layout/src/scan/tasks.rs
```diff
@@ -29,6 +29,7 @@ pub type TaskFuture<A> = BoxFuture<'static, VortexResult<A>>;
 pub(super) fn split_exec<A: 'static + Send + Sync>(
     ctx: Arc<TaskContext<A>>,
     split: Range<u64>,
+    limit: Option<&mut usize>,
```
robert3005, line 33: there's something weird here - you will still do a bunch of work even if you have `Some(0)` as limit

AdamGS:
Pushed a change that deals with the a limit of 0,  but by pre-filtering tasks in the top loop in `ScanBuilder::build` and at the `split_exec` level when we build the futures.

vortex-mask/src/lib.rs
```diff
@@ -509,6 +509,36 @@ impl Mask {
...
+            Mask::AllFalse(_) => self,
+            Mask::Values(ref mask_values) => {
+                if limit >= mask_values.true_count() {
+                    return self;
+                }
+
+                let existing_buffer = mask_values.boolean_buffer();
+
+                let mut new_buffer_builder = BooleanBufferBuilder::new(mask_values.len());
+                new_buffer_builder.append_n(existing_buffer.len(), false);
```
robert3005: nit: use mask_values.len() here, it's confusing that existing buffer is used here to mean the same thing as mask_values line above

robert3005 (approved):

Issue #3463, opened 2025-06-05
Python modules are private, docs aren't tested


mwlon:
I don't see a way to write a vortex file from Python right now? [The docs](https://github.com/vortex-data/vortex/blob/develop/docs/guides/python-integrations.md) say

```python
>>> vx.io.write(arr, 'example.vortex')
```

But the io module is private. Is there a way to write a vortex file from python today, and can you add `--doctest-glob=**/*python*.md` or something to your CI pytests?

I'm using `vortex-data==0.33.2`.

robert3005:
I think you're hitting https://github.com/vortex-data/vortex/issues/2668#issuecomment-2718346462. You will need to change the way things are imported

gatesn:
I wonder if actually we should make sure this works as modules, rather than packages.

Folks will draw a similarity to Parquet/Arrow and therefore expect the imports to work the same way.

mwlon:
Ah `from vortex import io as vio` works; I hadn't thought to try that. I'm in favor of having these as modules, but it would also suffice to fix the docs.

Issue #2668, opened 2025-03-12
bug: AttributeError: module 'vortex' has no attribute 'io'


shenlei149:
The version of python package vertex_array is 0.25.2.

repro steps:
1. pip3 install vortex-array
2. python code: `vortex.io`

error message:
```
>>> vortex.io
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'vortex' has no attribute 'io'
```

gatesn:
Apologies, I didn't realise folks were making use of the Python bindings! I will bring back this behaviour.

Note also there have been other breaking changes to the Python API, mostly moving some pieces around.

gatesn:
To be a little more helpful, `vortex.io` is now a Python package, and therefore you need to import it as `from vortex import io` or `import vortex.io; vortex.io.blah(...)`

`import vortex; vortex.io` only works when `io` is an attribute of the `vortex` module

shenlei149:
> To be a little more helpful, `vortex.io` is now a Python package, and therefore you need to import it as `from vortex import io` or `import vortex.io; vortex.io.blah(...)`
> 
> `import vortex; vortex.io` only works when `io` is an attribute of the `vortex` module

Thanks~

gatesn:
I actually want to fix this so that `import vortex; vortex.io` works. I think I prefer it vs the current Python package model where you must separately import the io package!

blaginin:
since @gatesn mentioned he wanted to fix, assigning to him

but @claude, maybe you want to give it a go?

PR #3485, opened 2025-06-08
[experimental] Playing around with python bindings

AdamGS:
Saw #3463 and #2668 which made me want to understand better what we're doing in python and maybe get a nicer import experience with less code

Current status:
1. Completely autogenerated stubs (with [pybind11-stubgen](https://github.com/sizmailov/pybind11-stubgen) (aside from formatting them with ruff)
2. It generates packages, so examples like `vortex.io.write` do work. On the other hand - I'm not sure that's better/desired, I don't have a very good sense as to what's the nicest API here and what's intuitive to other python users.
3. All tests pass
4. Slightly less mypy errors compared to develop
5. I tried using `pyo3-stub-gen` but ran into some linking issues that I couldn't solve (still have it in some local branch if anyone wants to take a look)


AdamGS:
IDK if this is any better, I'm closing it because I don't feel like there's a real need for this large of a change that doesn't actually make that much of a difference in usability or experience.

Issue #3442, opened 2025-06-03
bug[vortex-layout]: cannot map nullable bool array to mask in `filter_evaluation`


joseph-isaacs:
The following expression is an example: `is_null((($.a = 0i32) and ($.b = 0i32)))`

joseph-isaacs:
https://github.com/vortex-data/vortex/pull/3443

joseph-isaacs:
@onursatici  and I came up with what we think is the solution


LayoutReader changes the filter_evaluation to 

```rust
fn filter_evaluation(
  &self, 
  row_range: &Range<u64>,
  expr: &ExprRef,
  null_semantics: {NullToFalse | PreserveNulls
)
```

mask evalution invoke changes to
```rust
MaskEvaluation::invoke(self, mask) -> { Mask | MaskWithValidity }
```

PR #3487, opened 2025-06-09
fix[vortex-layout]: remove unsound layout filter expr

joseph-isaacs:
It is unsound cast `null -> false` in  `filter_evaluation` logic, this is the only reason to have a filter and a projection, so use the projection to eval boolean exprs. This requires moving bool shortcuts into the projection function

see https://github.com/vortex-data/vortex/pull/3443


joseph-isaacs:
Clickbench q2 is the only regression, 2x Ill take a look

joseph-isaacs:
closed in favour of: https://github.com/vortex-data/vortex/issues/3442#issuecomment-2962825878

PR #3488, opened 2025-06-09
dict layout writer to parallelise encoding and compression

onursatici:


vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -100,36 +108,62 @@ impl LayoutStrategy for DictStrategy {
...
+                        )
+                        .await?;
+                    children.push(DictLayout::new(values_layout, codes_layout).into_layout());
+                }
+                Ok(children)
+            };
+
+            // join dict encoding task
+            let (mut children, _) = try_join!(child_layouts_fut, encode_handle)?;
+
```
onursatici, line 166: might be worth hiding whitespace while reviewing this bit

robert3005 (approved):

PR #3489, opened 2025-06-09
chore: cleanup scan task execution to be clearer

a10y:
This offers a clearer straight-line implementation of each async scan task. I personally find this much more readable and traceable/editable.

### DIFF
--- a/vortex-expr/src/transform/immediate_access.rs
+++ b/vortex-expr/src/transform/immediate_access.rs
@@ -65,7 +65,7 @@ impl<'a> NodeVisitor<'a> for ImmediateScopeAccessesAnalysis<'a> {
     fn visit_down(&mut self, node: &'a Self::NodeTy) -> VortexResult<TraversalOrder> {
         assert!(
             !node.as_any().is::<Select>(),
-            "cannot analyse select, simply the expression"
+            "cannot analyze select, simplify the expression"
         );
         if let Some(get_item) = node.as_any().downcast_ref::<GetItem>() {
             if is_root(get_item.child()) {
--- a/vortex-layout/src/scan/mod.rs
+++ b/vortex-layout/src/scan/mod.rs
@@ -190,7 +190,7 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
                     selection: self.selection.clone(),
                     filter: self.filter.clone(),
                     reader: layout_reader.clone(),
-                    projection: self.projection.clone(),
+                    projection: projection.clone(),
                     mapper: self.map_fn.clone(),
                     task_executor: None,
                 });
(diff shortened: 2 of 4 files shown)


vortex-expr/src/transform/immediate_access.rs
```diff
@@ -65,7 +65,7 @@ impl<'a> NodeVisitor<'a> for ImmediateScopeAccessesAnalysis<'a> {
     fn visit_down(&mut self, node: &'a Self::NodeTy) -> VortexResult<TraversalOrder> {
         assert!(
             !node.as_any().is::<Select>(),
-            "cannot analyse select, simply the expression"
+            "cannot analyze select, simplify the expression"
```
a10y, line 68: 🇺🇸 🇺🇸 🇺🇸

vortex-layout/src/scan/mod.rs
```diff
@@ -177,98 +181,25 @@ impl<A: 'static + Send> ScanBuilder<A> {
...
-                    .boxed();
-                    Ok::<_, VortexError>((row_range, mask_fut))
-                })
-                .try_collect()?;
-
-            // Map the row masks through the filter evaluation
-            row_masks
-                .into_iter()
-                .map(|(row_range, mask_fut)| {
-                    let eval = layout_reader.filter_evaluation(&row_range, filter)?;
```
a10y, line 233: one thing that is lost by this PR is that the evaluators before were all built eagerly. I'm not sure how important that is.

gatesn: It was kind of the whole point, to register the segments from each evaluator for prefetching

a10y: Where is the prefetching happening?

a10y: Ok, so the prefetching is very deep into the segmentsource and dependent on the order that you call the layout_reader. I wonder if there's a way to hoist control up a level to make it a bit more obvious what's going on...

a10y:
I _think_ this is at parity? The Q6 for S3 is 1.36x but everything else is the same, so not sure if that's a flake or not?

a10y:
Ok some of Clickbench like Q41 is 1.4x as well. It's hard to tell if those are real regressions or just noise. I propose we merge this and if they cause regressions on develop run we just revert?

AdamGS (approved):
I think that some performance pain is fine, we'll figure out the issues (if its not a flake) and clearer code will make it easier to improve over time.

robert3005:
This seems to be a regression on couple of clickbench queries. I think we changed the order of registering scans in a subtle way

joseph-isaacs:
If this being looked at, I think we really want the previous behaviour back

robert3005:
@a10y is looking at it, you're correct we want the performance back

PR #3490, opened 2025-06-09
feat: add vortex-common a create for shared defs

joseph-isaacs:


lwwmanning:
this might be better named `vortex-utils` or `vortex-common`? if I saw a `vortex-core` crate, I'd imagine it to look a lot like current `vortex-array` (albeit a bit pruned)

joseph-isaacs:
`vortex-common` was used for commwasn, I guess we can reuse (https://crates.io/crates/vortex-common)?

utils is a banned keyword, but I personally prefer it

joseph-isaacs:
updated

lwwmanning (approved):

PR #2273, opened 2025-02-07
Improve expression performance

gatesn:
* Only simplify once in the scanner
* Manually implement return_dtype instead of expensive Canonical::Empty (FLUP: we should add `return_nullability()` expr too)
* Remove `Field::Index` weirdness.
* More consistent APIs for Struct{Scalar,Array,DType}

PR #3491, opened 2025-06-09
feat: FieldPath::resolve, ::exists, and AccesPath::new

danking:
For Spiral, mainly.


vortex-dtype/src/field.rs
```diff
@@ -110,6 +126,62 @@ impl FieldPath {
...
+    /// let resolved = path.resolve(dtype).unwrap();
+    /// assert_eq!(resolved, DType::Primitive(PType::U32, NonNullable));
+    /// ```
+    pub fn resolve(&self, mut dtype: DType) -> VortexResult<DType> {
+        for field in &self.0 {
+            dtype = match (dtype, field) {
+                (DType::Struct(fields, _), Field::Name(name)) => fields.field(name)?,
+                (DType::List(element_dtype, _), Field::ElementType) => DType::clone(&element_dtype),
+                (other, f) => {
+                    vortex_bail!("FieldPath: invalid index {:?} for DType {:?}", f, other)
```
joseph-isaacs, line 172: I think we want an option here, or maybe both. Since vortex_bail, create a backtrace

vortex-dtype/src/field.rs
```diff
@@ -59,7 +63,19 @@ impl Field {
...
+/// A sequence of field selectors representing a path through zero or more layers of `DType`.
+///
+/// # Examples
+///
+/// The empty path references the root:
+///
+/// ```
+/// use vortex_dtype::*;
+///
+/// let i32 = DType::Primitive(PType::I32, Nullability::NonNullable);
```
a10y: I didn't realize this was even valid syntax. Can we call it i32_col or something. Idk why but it's jarring to me :P

vortex-dtype/src/field.rs
```diff
@@ -110,6 +126,62 @@ impl FieldPath {
...
+    ///         ),
+    ///     )])),
+    ///     NonNullable,
+    /// );
+    ///
+    /// let path = FieldPath::from(vec![Field::from("a"), Field::ElementType, Field::from("b")]);
+    /// let resolved = path.resolve(dtype).unwrap();
+    /// assert_eq!(resolved, DType::Primitive(PType::U32, NonNullable));
+    /// ```
+    pub fn resolve(&self, mut dtype: DType) -> VortexResult<DType> {
```
a10y, line 166: This feels flipped. A DType should be able to resolve a FieldPath to get back a DType imo?

vortex-dtype/src/field.rs
joseph-isaacs: Or Maybe this is a case for a Result<>, and then a vortex result on exists

vortex-dtype/src/field.rs
```diff
@@ -110,6 +126,62 @@ impl FieldPath {
...
+                    vortex_bail!("FieldPath: invalid index {:?} for DType {:?}", f, other)
+                }
+            }
+        }
+
+        Ok(dtype)
+    }
+
+    /// Does the field referenced by the field path exist in the given dtype?
+    pub fn exists(&self, dtype: DType) -> bool {
```
a10y, line 181: Again, this feels inverted to me

a10y (commented):
Some weakly held opinions

vortex-dtype/src/field.rs
```diff
@@ -110,6 +126,62 @@ impl FieldPath {
...
+    ///
+    /// Note that a nullable DType may contain a non-nullable DType. This function returns the
+    /// literal nullability of the child.
+    ///
+    /// # Examples
+    ///
+    /// Extract the type of the "b" field from `struct{a: list(struct{b: u32})?}`:
+    ///
+    /// ```
+    /// use std::sync::Arc;
```
a10y, line 140: Bless up for doc tests 🙏

vortex-dtype/src/field.rs
a10y: Like this is like saying Path::exists(FileSystem) to me

vortex-dtype/src/field.rs
danking: I also think this is best as an `Option<DType>` valued function but I think `StructFields::field` should likewise be `Option<DType>`

vortex-dtype/src/field.rs
danking: @joseph-isaacs thoughts on `DType::lookup(&FieldPath)` instead?

vortex-dtype/src/field.rs
danking: Done.

vortex-dtype/src/field.rs
joseph-isaacs: I think of a field path as a regex like thing analogous to str matching, so what ever that implies

vortex-dtype/src/field.rs
danking: Hmm. We seem to have explicitly moved away from Option-returning methods in the name of performance: https://github.com/vortex-data/vortex/pull/2273

vortex-dtype/src/field.rs
joseph-isaacs: I'd can see both ways, I don't really  have a strong preference

vortex-dtype/src/field.rs
danking: OK, my conclusion: this PR is currently aligned with how StructFields works. That might be wrong but we should change them all together in a separate PR.

vortex-dtype/src/field.rs
danking: I think I lean towards Andrew's conception, so I went with his approach.

joseph-isaacs: https://docs.rs/regex/latest/regex/ I think we should follow this

joseph-isaacs: so keep as is?

joseph-isaacs (dismissed):
can we keep them on the matcher?

a10y (approved):

joseph-isaacs:
happy for you to overrule and dismiss just like the consistency with regex libs

danking:
Reverted to Joe's preference.

PR #3498, opened 2025-06-10
chore[vortex-layout]: use scope dtype in layouts (not dtype)

joseph-isaacs:
This is a bit anyoing that you keep having to unwrap the dtype().

Maybe another function that panics if there is no value for the ident. I don't think we want to assume there is always a Identity identifier

### DIFF
--- a/vortex-file/src/footer/mod.rs
+++ b/vortex-file/src/footer/mod.rs
@@ -22,7 +22,8 @@ pub use segment::*;
 use vortex_array::stats::StatsSet;
 use vortex_array::{ArrayContext, ArrayRegistry};
 use vortex_dtype::DType;
-use vortex_error::{VortexResult, vortex_bail, vortex_err};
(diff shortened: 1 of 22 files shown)


vortex-file/src/footer/mod.rs
```diff
@@ -114,7 +115,10 @@ impl Footer {
 
     /// Returns the [`DType`] of the file.
     pub fn dtype(&self) -> &DType {
-        self.root_layout.dtype()
+        self.root_layout
+            .scope_dtype()
+            .dtype(&Identifier::Identity)
```
onursatici, line 120: maybe this should be a `root_unchecked` method on the scoped dtype?

joseph-isaacs:
https://github.com/vortex-data/vortex/pull/3512

gatesn:
Why don't we want to assume there is always an identity? We always did before, and it's trivial to create an empty struct.

It simplifies all this downstream stuff

PR #3500, opened 2025-06-10
stats calculation parallelism in layout writer

onursatici:


vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -37,7 +38,7 @@ impl Default for DictLayoutOptions {
                 max_bytes: 1024 * 1024,
                 max_len: u16::MAX as usize,
             },
-            encoded_buffer_size: 512,
+            encoded_buffer_size: 8,
```
onursatici, line 41: unrelated but felt like 512 chunks per column is too much to potentially keep in memory

onursatici: I guess any number > 1 would preserve perf, the consuming end of this channel shouldn't be bursty so we shouldn't need that much of a buffer

vortex-layout/src/layouts/zoned/reader.rs
```diff
@@ -325,6 +326,7 @@ mod test {
                 block_size: 3,
                 ..Default::default()
             },
+            Arc::new(LocalExecutor {}),
```
robert3005:
```suggestion
            Arc::new(LocalExecutor),
```

robert3005 (approved):

PR #3501, opened 2025-06-10
feat[vortex-expr]: remove pruning expr in favor of `checked_pruning_expr`.

joseph-isaacs:
This allows the creation of pruning expression with only some stats available, this also verifies that the stats exist before the predicate creation

### DIFF
--- a/vortex-dtype/src/field.rs
+++ b/vortex-dtype/src/field.rs
@@ -10,6 +10,7 @@ use std::sync::Arc;
 
 use itertools::Itertools;
 use vortex_error::{VortexResult, vortex_bail};
+use vortex_utils::aliases::hash_set::HashSet;
 
 use crate::DType;
 
@@ -205,6 +206,28 @@ impl From<Vec<Field>> for FieldPath {
 impl Display for FieldPath {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         Display::fmt(&self.0.iter().format("."), f)
+    }
+}
+
+#[derive(Default, Clone, Debug)]
+/// Contains a set of field paths, and can answer efficient field path contains queries.
+pub struct FieldPathSet {
+    /// While this is currently a set wrapper it can be replaced with a trie.
+    // TODO(joe): this can be replaced with a `FieldPath` trie
+    set: HashSet<FieldPath>,
+}
+
+impl FieldPathSet {
+    /// Checks if a set contains a field path
+    pub fn contains(&self, path: &FieldPath) -> bool {
+        self.set.contains(path)
+    }
+}
+
+impl FromIterator<FieldPath> for FieldPathSet {
+    fn from_iter<T: IntoIterator<Item = FieldPath>>(iter: T) -> Self {
+        let set = HashSet::from_iter(iter);
+        Self { set }
     }
 }
 
--- a/vortex-expr/src/pruning/pruning_expr.rs
+++ b/vortex-expr/src/pruning/pruning_expr.rs
@@ -2,97 +2,55 @@ use std::iter;
 
 use itertools::Itertools;
 use vortex_array::stats::Stat;
-use vortex_array::{Array, ArrayRef};
 use vortex_dtype::{Field, FieldName, FieldPath};
-use vortex_error::{VortexExpect, VortexResult};
 use vortex_utils::aliases::hash_map::HashMap;
-use vortex_utils::aliases::hash_set::HashSet;
 
 use super::relation::Relation;
-use crate::{AccessPath, ExprRef, Scope, StatsCatalog, get_item, var};
+use crate::{AccessPath, ExprRef, ScopeFieldPathSet, StatsCatalog, get_item, var};
 
-#[derive(Debug, Clone)]
-pub struct PruningPredicate {
-    expr: ExprRef,
-    required_stats: Relation<AccessPath, Stat>,
-}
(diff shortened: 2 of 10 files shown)


vortex-dtype/src/field.rs
```diff
@@ -208,6 +209,28 @@ impl Display for FieldPath {
     }
 }
 
+#[derive(Default, Clone, Debug)]
+/// Contains a set of field paths, and can answer efficient field path contains queries.
+pub struct FieldPathSet {
+    /// While this is currently a set wrapper it can be replaced with a trie.
+    // TODO(joe): this can be replaced with a `FieldPath` trie
```
danking, line 216: tries! 🥳

vortex-expr/src/pruning/pruning_expr.rs
```diff
@@ -120,13 +79,37 @@ pub fn pruning_expr(expr: &ExprRef) -> Option<(ExprRef, Relation<AccessPath, Sta
     Some((expr, relation))
 }
 
+// Build a pruning expr mask an existing bundle of stats
```
danking: why is this a normal comment and not a doc comment?

vortex-expr/src/pruning/pruning_expr.rs
```diff
@@ -120,13 +79,37 @@ pub fn pruning_expr(expr: &ExprRef) -> Option<(ExprRef, Relation<AccessPath, Sta
...
+    scope_field_paths: &ScopeFieldPathSet,
+) -> Option<(ExprRef, RequiredStats)> {
+    let mut catalog = ScopeStatsCatalog {
+        any_catalog: Default::default(),
+        scope_field_paths,
+    };
+
+    let expr = expr.stat_falsification(&mut catalog)?;
+
+    // TODO: filter by used exprs
```
danking: Why is this TODO? Is that not what lines 98-101 are doing?

vortex-expr/src/scope.rs
```diff
@@ -27,6 +27,35 @@ impl FromStr for Identifier {
...
+    fn from(value: &str) -> Self {
+        if value.is_empty() {
+            Identifier::Identity
+        } else {
+            Identifier::Other(Arc::from(value))
+        }
+    }
+}
+
+impl From<&Arc<str>> for Identifier {
```
danking, line 49: I'm not really a Rustacean yet but it seems a bit odd to me to define these two `From` implementations on references to values when they both require ownership (which they do via cloning).

vortex-file/src/pruning.rs
```diff
@@ -2,54 +2,57 @@ use std::sync::Arc;
...
+use vortex_error::{VortexResult, vortex_bail};
+use vortex_expr::pruning::field_path_stat_field_name;
 use vortex_scalar::Scalar;
+use vortex_utils::aliases::hash_map::HashMap;
+use vortex_utils::aliases::hash_set::HashSet;
 
-pub fn extract_relevant_stat_as_struct_row(
-    predicate: &PruningPredicate,
-    stats_set: &Arc<[StatsSet]>,
+pub fn extract_relevant_file_stat_as_struct_row(
```
danking:
```suggestion
pub fn extract_relevant_file_stats_as_struct_row(
```

Right? It extracts a variable number of stats.

vortex-file/src/pruning.rs
```diff
@@ -2,54 +2,57 @@ use std::sync::Arc;
...
+            };
+            if stat == &Stat::Max || stat == &Stat::Min {
+                columns.push((
+                    field_path_stat_field_name(field_path, *stat),
+                    ConstantArray::new(Scalar::new(field_dtype.clone(), stat_value.clone()), 1)
+                        .to_array(),
+                ))
+            } else {
+                todo!("unsupported file prune stat {stat}")
+            }
```
danking, line 52: Man yeah this is such an unfortunate thing to have to write. It'd be nice if the file stats were just a length one array of stats.

danking (changes requested):
Just a few nits; looks good to me!

vortex-expr/src/pruning/pruning_expr.rs
joseph-isaacs: Nope, there is no filtering by expr being in the falsification expr just yet

vortex-file/src/pruning.rs
joseph-isaacs: one row, but many stats, ye

vortex-file/src/pruning.rs
joseph-isaacs: yep, but I think this place is where to do that

vortex-expr/src/scope.rs
joseph-isaacs: hmm, you would prefer From<Arc<str>> (and clone) on the outside

joseph-isaacs: I can see the point of this, but I guess should be either remove this from the code base or embrace clone on Arc is cheep

danking: If this is the style of Vortex let’s stick with it. I think my argument for caller clones is that when you write new code you see the clones and reckon with their effect.

danking (approved):

vortex-expr/src/scope.rs
robert3005: This impl is never used in the codebase

PR #3502, opened 2025-06-10
duckdb: extract schema in bind function

0ax1:


vortex-duckdb-ext/src/convert/dtype.rs
```diff
@@ -1,12 +1,302 @@
...
+                let child_names: Vec<CString> = struct_type
+                    .names()
+                    .iter()
+                    .map(|field_name| {
+                        CString::new(field_name.as_ref())
+                            .map_err(|e| vortex_err!("Invalid field name '{field_name}': {e}"))
+                    })
+                    .collect::<Result<_, _>>()?;
+
+                return LogicalType::struct_type(child_types, child_names);
```
joseph-isaacs, line 77: cann this take an iterator to avoid a double collect?

0ax1: ~~Hmm, I tried that and concluded that it is more complex / harder to read.~~

Ah you didn't mean `zip` here but passing the iterator into `LogicalType::struct_type`.

vortex-duckdb-ext/src/convert/dtype.rs
```diff
@@ -1,12 +1,302 @@
...
+use vortex::error::{VortexError, vortex_err};
+
+use crate::cpp::{self, duckdb_logical_type};
 use crate::duckdb::LogicalType;
 
+impl LogicalType {
+    fn struct_type(
+        child_types: Vec<LogicalType>,
+        child_names: Vec<CString>,
+    ) -> Result<LogicalType, VortexError> {
```
joseph-isaacs:
```suggestion
    ) -> VortexResult<LogicalType> {
```

joseph-isaacs (approved):
some small changes and we are good to go

vortex-duckdb-ext/src/convert/dtype.rs
0ax1: Alternative would be passing iterators would look like
```
            DType::Struct(struct_type, _) => {
                return struct_type
                    .fields()
                    .map(|field_dtype| LogicalType::try_from(&field_dtype))
                    .process_results(|child_types_iter| {
                        struct_type
                            .names()
                            .iter()
                            .map(|field_name| {
                                CString::new(field_name.as_ref()).map_err(|e| {
                                    vortex_err!("Invalid field name '{field_name}': {e}")
                                })
                            })
                            .process_results(|child_names_iter| {
                                LogicalType::struct_type(child_types_iter, child_names_iter)
                            })?
                    })?;
            } // TODO: Add support for Decimal, Extension, List
```
Issue being that the `Result` needs to be unpacked. I think the added complexity isn't justified here.

But I changed the struct construction to
```
impl LogicalType {
    fn struct_type<T, N>(child_types: T, child_names: N) -> VortexResult<LogicalType>
    where
        T: IntoIterator<Item = LogicalType>,
        N: IntoIterator<Item = CString>,
```

PR #3504, opened 2025-06-10
fix: Arrow timestamps are in an _unknown_ timezone

danking:
The current logic assumes a timezone-naive vortex.timestamp has the UTC timzeone. However, Arrow [specifically says it is in an unknown timezone](https://github.com/apache/arrow/blob/main/format/Schema.fbs#L304-L322):

> Timestamps with an unset / empty timezone
> -----------------------------------------
>
> If a Timestamp column has no timezone value, its epoch is
> 1970-01-01 00:00:00 (January 1st 1970, midnight) in an *unknown* timezone.
>
> Therefore, timestamp values without a timezone cannot be meaningfully
> interpreted as physical points in time, but only as calendar / clock
> indications ("wall clock time") in an unspecified timezone.
>
>   For example, the timestamp value 0 with an empty timezone string
>   corresponds to "January 1st 1970, 00h00" in an unknown timezone: there
>   is not enough information to interpret it as a well-defined physical
>   point in time.
>
> One consequence is that timestamp values without a timezone cannot
> be reliably compared or ordered, since they may have different points of
> reference.  In particular, it is *not* possible to interpret an unset
> or empty timezone as the same as "UTC".

This bug most prominently appears in the Vortex cli browser which shows the incorrect timezone for the min/max statistics of a timestamp column:

```
╭───────────────────────────────────────────────────────────────────
│ File Layout │ Segments
│╭Layout Info───────────────────────────────────────────────────────
││Kind: vortex.flat
││Row Count: 1
││Schema: {max=ext(vortex.timestamp, i64, ExtMetadata([1, 0, 0]))?,
││Children: 0
││Segment data size: 1.32 kB
││FlatBuffer Size: 1.32 kB
││
││
│╰──────────────────────────────────────────────────────────────────
│╭Array Info────────────────────────────────────────────────────────
││chunk                         max                           max_is
││0                             2025-06-11T02:24:37.054366Z   false
││
││
││
│╰──────────────────────────────────────────────────────────────────
╰───────────────────────────────────────────────────────────────────
```

After this change, the year, month, day, hour, etc. are unchanged, but the timezone is correctly reported as unknown.

```
╭───────────────────────────────────────────────────────────────────
│ File Layout │ Segments
│╭Layout Info───────────────────────────────────────────────────────
││Kind: vortex.flat
││Row Count: 1
││Schema: {max=ext(vortex.timestamp, i64, ExtMetadata([1, 0, 0]))?,
││Children: 0
││Segment data size: 1.32 kB
││FlatBuffer Size: 1.32 kB
││
││
│╰──────────────────────────────────────────────────────────────────
│╭Array Info────────────────────────────────────────────────────────
││chunk                         max                          max_is
││0                             2025-06-11T02:24:37.054366   false
```


vortex-dtype/src/datetime/temporal.rs
```diff
@@ -39,8 +39,8 @@ pub enum TemporalJiff {
     Time(Time),
     /// A date value.
     Date(Date),
-    /// A timestamp value.
-    Timestamp(Timestamp),
+    /// A zone-naive timestamp value.
+    Unzoned(DateTime),
     /// A zoned timestamp value.
     Zoned(Zoned),
```
danking, line 45: I think Jiff's docs are a bit confusing, but I have convinced myself this is correct. Zoned is an instant in time _with some timezone_. Timestamp is an instant in time _in the UTC timezone_. DateTime is a timezone-naive instant in time.

[From the overview](https://docs.rs/jiff/latest/jiff/#overview):
>The primary type in this crate is [Zoned](https://docs.rs/jiff/latest/jiff/struct.Zoned.html). A Zoned value is a datetime that corresponds to a precise instant in time in a particular geographic region. Users of this crate may find it helpful to think of a Zoned as a triple of the following components:
> 
> - A [Timestamp](https://docs.rs/jiff/latest/jiff/struct.Timestamp.html) is a 96-bit integer of nanoseconds since the [Unix epoch](https://en.wikipedia.org/wiki/Unix_time). A timestamp is a precise instant in time.
> - A [civil::DateTime](https://docs.rs/jiff/latest/jiff/civil/struct.DateTime.html) is an inexact calendar date and clock time. The terms “civil”, “local”, “plain” and “naive” are all used in various places to describe this same concept.
> 
> [...]

a10y (approved):

vortex-dtype/src/datetime/temporal.rs
robert3005: Should really rename DateTime to LocalDateTime

BurntSushi: Jiff author here.

I suspect the thing that has made this extremely confusing is the difference in terminology used by Arrow versus most other datetime libraries. A "timestamp" or a "Unix timestamp" is _always_ a physical instant in time. That's what a `jiff::Timestamp` is. It's not 100% precisely correct to say that a `jiff::Timestamp` is an instant in time in the UTC time zone, although that is a short-hand that is unlikely to get you into much trouble. A `jiff::Timestamp` is just an instant in time in the the Unix time scale (not UTC time scale). It's somewhat of a category error to say that it is "in" a specific time zone. This is why when you print a `jiff::Timestamp`, it uses the `Z` offset indicator, which means "the offset from UTC is not known."

For `civil::DateTime`, it is _not_ a "timezone-naive instant in time." It _can't_ be an instant in time on its own. It can only _become_ an instant in time when you pair it with a time zone. For example, `2025-06-11T09:17:00` is a civil datetime. It is _common_ but _not universal_ for it to be a local datetime. While `2025-06-11T09:17:00` is my current local time, it may not be _your_ local time. (Hence the confusion around the "local" terminology and why Jiff doesn't use it.)

I comment here only to try and help improve understanding.

If there's anything I can add to the docs that would have helped you more efficiently, I'd be happy to hear that. :-)

> Should really rename DateTime to LocalDateTime

See https://github.com/BurntSushi/jiff/issues/12 for more on that.

PR #3508, opened 2025-06-10
docs: more docs around AnalysisExpr and boolean operators

danking:


robert3005 (approved):

vortex-array/src/compute/boolean.rs
```diff
@@ -181,11 +181,66 @@ impl ComputeFnVTable for Boolean {
...
+    /// | ----- | ----- | ------ |
+    /// | true  | true  | true   |
+    /// | true  | false | false  |
+    /// | true  | null  | null   |
+    /// | false | true  | false  |
+    /// | false | false | false  |
+    /// | false | null  | null   |
+    /// | null  | true  | null   |
+    /// | null  | false | null   |
+    /// | null  | null  | null   |
```
joseph-isaacs: Can we embed the actual table from the wikipdia


| A ∧ B |     | **B** |   |   |
|:-----:|:---:|:-----:|:-:|:-:|
|       |     | **F** | **U** | **T** |
| **A** | **F** | F | F | F |
|       | **U** | F | U | U |
|       | **T** | F | U | T |

joseph-isaacs: I used an llm to go from html to md

vortex-expr/src/analysis.rs
```diff
@@ -35,15 +35,20 @@ pub trait AnalysisExpr {
         None
     }
 
-    /// If an expression is returned, its value is an upper bound on the value of `expr`.
+    /// An expression for the upper non-null bound of this expression, if available.
```
joseph-isaacs, line 38: Why do we want this? returning null here implies no knowledge, it seems good that `max(lit(Scalar::null)) -> None` not `lit(Scalar::null)`.

I think this will be hard for people to impl correctly

vortex-expr/src/analysis.rs
```diff
@@ -35,15 +35,20 @@ pub trait AnalysisExpr {
...
     ///
-    /// We may return `None` for values which have no upper bound or values for which knowing the
-    /// upper bound is difficult.
+    /// This function returns None if there is no upper bound or it is difficult to compute.
```
joseph-isaacs, line 40: Sorry nitpick, but I think the point is *unknown* (due to any reason, including not implemented)
```suggestion
    /// This function returns None if there the upper bound is unknown (non existent or hard to compute).
```

vortex-expr/src/analysis.rs
```diff
@@ -35,15 +35,20 @@ pub trait AnalysisExpr {
...
     fn max(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
         None
     }
-    /// If an expression is returned, its value is an upper bound on the value of `expr`.
-    /// see `AnalysisExpr::max`
+
+    /// An expression for the lower non-null bound of this expression, if available.
+    ///
+    /// See [AnalysisExpr::max] for important details.
     fn min(&self, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef> {
```
joseph-isaacs, line 52: What do you think about making this fn `fn stat_of(&self, stat: Stat, _catalog: &mut dyn StatsCatalog) -> Option<ExprRef>` (once fn for all stats)

joseph-isaacs (changes requested):
Please do make the unknown change

vortex-array/src/compute/boolean.rs
joseph-isaacs: | A ∨ B |     | **B** |   |   |
|:-----:|:---:|:-----:|:-:|:-:|
|       |     | **F** | **U** | **T** |
| **A** | **F** | F | U | T |
|       | **U** | U | U | T |
|       | **T** | T | T | T |

joseph-isaacs: | **A** | **¬A** |
|:-----:|:------:|
| F     | T      |
| U     | U      |
| T     | F      |

joseph-isaacs: The grid much easier to read than column for me

vortex-expr/src/analysis.rs
danking: Sure, I suppose undefined (max of an empty list) is just another kind of unknown.

vortex-array/src/compute/boolean.rs
danking: Agreed. All four converted.

vortex-expr/src/analysis.rs
danking: In develop there are three possible results of this method:
1. `None`
2. `Some(e)` where e evaluates to a non-null value.
3. `Some(e)` where e evaluates to a null value.

I think we very clearly defined the meaning of the first two (1: the max is unknown, 2: the max is this known value). The doc string does not clarify what the third case means. I am trying to clarify that in this docstring. I am not trying to forbid returning `None` when, at expression-creation-time, we know the max is unknown.

The specific use-case is that in Spiral, we have an array of file statistics. For some files, we know the maximum. For some files, we _do not_ know the maximum. In the latter case, the `max` column for that row is `null`.

vortex-expr/src/analysis.rs
danking: Seems fine too, though I think best to do as a separate PR.

vortex-expr/src/analysis.rs
joseph-isaacs: I see I think that when I read non-null I read non `None`. Maybe 
```suggestion
    /// An expression for the upper non-null scalar bound of this expression, if available.
```

vortex-expr/src/analysis.rs
joseph-isaacs: Agreed, but since you where defining semantics I thought it was a good idea to chat about it

joseph-isaacs (approved):
thanks

PR #3509, opened 2025-06-11
fix: deserialize ident proto expressions as the new Var("")

a10y:
I had considered removing Ident from the proto and changing all of the Spark/Iceberg protobuf serde, but I'm honestly not convinced we aren't going to keep changing/remove Var/Let so I'd prefer to just deserialize Ident proto


a10y:
Concretely the problem I'm trying to fix is this error

image

robert3005 (approved):

PR #3512, opened 2025-06-11
feat[vortex-layout]: add a virtual layout reader node not persisted

joseph-isaacs:
Layouts seem to be mapped directly to seconds on a file, this PR distinguishes between (physical) layout and (new) virtual layouts, these can be used to include extra reading logic or virtual columns

### DIFF
--- a/vortex-layout/src/layouts/filter.rs
+++ b/vortex-layout/src/layouts/filter.rs
@@ -19,7 +19,7 @@ use crate::{ArrayEvaluation, MaskEvaluation, PruningEvaluation};
 /// The selectivity histogram quantile to use for reordering conjuncts. Where 0 == no rows match.
 const DEFAULT_SELECTIVITY_QUANTILE: f64 = 0.1;
 
-/// A [`LayoutReader`] that splits boolean expressions into individual conjunctions, tracks
+/// A [`VirtualLayoutReader`] that splits boolean expressions into individual conjunctions, tracks
 /// statistics about selectivity, and uses this information to reorder the evaluation of the
 /// conjunctions in an attempt to minimize the work done.
 ///
--- a/vortex-layout/src/layouts/filter.rs
+++ b/vortex-layout/src/layouts/filter.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeSet;
 use std::iter;
 use std::ops::{BitAnd, Range};
 use std::sync::Arc;
@@ -9,14 +8,13 @@ use dashmap::DashMap;
 use itertools::Itertools;
 use parking_lot::RwLock;
 use sketches_ddsketch::DDSketch;
-use vortex_array::stats::Precision;
-use vortex_dtype::{DType, FieldMask};
 use vortex_error::{VortexExpect, VortexResult, vortex_err, vortex_panic};
-use vortex_expr::ExprRef;
 use vortex_expr::forms::cnf::cnf;
+use vortex_expr::{ExprRef, ScopeDType};
 use vortex_mask::Mask;
 
-use crate::{ArrayEvaluation, LayoutReader, LayoutReaderRef, MaskEvaluation, PruningEvaluation};
+use crate::virtual_reader::{VirtualLayoutReader, VirtualLayoutReaderRef};
+use crate::{ArrayEvaluation, MaskEvaluation, PruningEvaluation};
 
 /// The selectivity histogram quantile to use for reordering conjuncts. Where 0 == no rows match.
 const DEFAULT_SELECTIVITY_QUANTILE: f64 = 0.1;
@@ -28,12 +26,12 @@ const DEFAULT_SELECTIVITY_QUANTILE: f64 = 0.1;
 /// This reader does not have a corresponding layout in the file, as it merely implements
 /// expression rewrite logic at read-time.
 pub struct FilterLayoutReader {
-    child: LayoutReaderRef,
+    child: VirtualLayoutReaderRef,
     cache: DashMap<ExprRef, Arc<FilterExpr>>,
 }
 
 impl FilterLayoutReader {
-    pub fn new(child: LayoutReaderRef) -> Self {
+    pub fn new(child: VirtualLayoutReaderRef) -> Self {
         Self {
             child,
             cache: Default::default(),
@@ -41,27 +39,13 @@ impl FilterLayoutReader {
     }
 }
 
-impl LayoutReader for FilterLayoutReader {
+impl VirtualLayoutReader for FilterLayoutReader {
     fn name(&self) -> &Arc<str> {
         self.child.name()
     }
 
-    fn dtype(&self) -> &DType {
-        self.child.dtype()
-    }
-
-    fn row_count(&self) -> Precision<u64> {
-        self.child.row_count()
-    }
-
-    fn register_splits(
-        &self,
-        field_mask: &[FieldMask],
-        row_offset: u64,
-        splits: &mut BTreeSet<u64>,
-    ) -> VortexResult<()> {
-        // Pass-through the splits to the child layout reader.
-        self.child.register_splits(field_mask, row_offset, splits)
+    fn scope_dtype(&self) -> &ScopeDType {
+        self.child.scope_dtype()
     }
 
     fn pruning_evaluation(
--- a/vortex-layout/src/lib.rs
+++ b/vortex-layout/src/lib.rs
@@ -22,5 +22,6 @@ pub mod scan;
 pub mod segments;
 pub mod sequence;
 mod strategy;
+mod virtual_reader;
 pub mod vtable;
 mod writer;
--- a/vortex-layout/src/scan/mod.rs
+++ b/vortex-layout/src/scan/mod.rs
@@ -26,6 +26,7 @@ use vortex_metrics::VortexMetrics;
 use crate::LayoutReader;
 use crate::layouts::filter::FilterLayoutReader;
 use crate::scan::tasks::{TaskContext, split_exec};
+use crate::virtual_reader::{PhysicalToVirtualLayoutAdaptor, VirtualLayoutReaderRef};
 
 mod executor;
 pub mod row_mask;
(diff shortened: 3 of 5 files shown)


gatesn:
Can you explain the difference?

I agree current named are bad, but Layout describes a node serialised in the file, and LayoutReader represents an instantiated node in the scan plan.

Is LayoutReader not sufficient to model such "virtual" things? We already do for filtering for example.

joseph-isaacs:
The big change is that virtual layout readers operate on virtual and physical columns (nodes not directly persisted to disk) (`ScopeDType`s) whereas layout readers operate on only layouts on disk (`DType`s). While we can change layout readers to operate on `ScopeDType` is error prone, since physical layouts only really have a DType and then other elements of the scope might be discarded if they are embedded deep in the layout hierarchy.

I also feels like current layout/layoutreaders are too loosely defined I think this change makes it more clear that they are (They operate on a persisted hierarchy of lazy arrays)

https://github.com/vortex-data/vortex/pull/3498 Is where I adapted layouts to use ScopeDType (and therefore operator on possible nested virtual columns)

gatesn:
I don't think this is moving in the right direction.. 

These concepts should be _more_ unified, not less.

Am I right in thinking this all stems from the way we implemented Let/Var?

In what way is it error prone for everyone to operate on ScopeDType?

joseph-isaacs:
Can you suggest a different direction, that can support virtual columns. 

There are parts to this suggestion.

1. Layout/LayoutReader are a little ill defined
2. LayoutReaders have `row_count` and `register_splits` which are not really part of a reader, just a physical file reader
3. `LayoutReader::dtype` needs to return a ScopeDType (if layouts support virtual columns [not in the file]).

Regarding the scope I guess on further reflection is not error prone, but to get the dtype of a file (with virtual columns) you will need to walk the whole layout tree for child dtypes since at any point in the tree a virtual column might be added altering the layout tree.

a10y:
Yea I think I'm also curious why this requires a new concept. In the case where I want to scan the file with schema `{a:i32, b:i32}` with row_idx, that just seems like a new LayoutReader that wraps an existing LayoutReader. I'm not sure why we need to distinguish.

a10y:
> to get the dtype of a file (with virtual columns) you will need to walk the whole layout tree for child dtypes since at any point in the tree a virtual column might be added altering the layout tree.

It feels like it's the other way around: based on the virtual columns you want to scan, you construct the LayoutReader tree that will read that for you. But you should be able to look at an expression upfront (including one which brings in virtual columns) and be able to determine the output DType surely?

joseph-isaacs:
What does this LayoutReader (with a virtual row_id) return as its input DType? Used in `expr.return_type(...)`

Either:
 1. DType -- I think this should be true for all persisted arrays.
 2. ScopeDType -- This is required if Layout/LayoutReaders can contain nested virtual columns
 3. We move virtual column into the DType and nested it in a struct. Then who is responsible for lift and unlifting?

gatesn:
I think I've realised a better name for `LayoutReader` that might better indicate it's role. Or at least.. how I see it's role...

Let's call it a `ScanNode`, similar in functionality to a DataFusion `ExecNode`. It's a node of a physical scan plan.

The tree of layouts in a file are loaded into a tree of scan nodes. We can then wrap up that scan node with our `FilterNode` (currently FilterLayoutReader), or other nodes that have evaluation logic but are not necessarily serialized into the file.

@joseph-isaacs had a good PR up at one point that removed `filter_evaluation` from the LayoutReader trait. I like this because it leaves the LayoutReader (ScanNode) with just two modes of execution: an approximate pruning_evaluation, and an exact projection_evaluation. Or `prune` and `evaluate` if we want to be more succinct.

I think in this model, it maybe makes it clearer why `VirtualLayoutReader` feels a little out of place. Because all scan nodes are "virtual", it's just that some happen to align to nodes of the layout tree.

Looking then at what the API of a ScanNode should be, and it doesn't make sense to me why a ScanNode should be _returning_ a `ScopeDType`. The `ScanNode::dtype()` function should return the dtype of the array that is returned from `ScanNode::evaluate`.

> What does this LayoutReader (with a virtual row_id) return as its input DType?

A LayoutReader / ScanNode does not return it's _input_ dtype, only it's output dtype.

The virtual columns you're referring to I think should look something like this (excuse the psuedo-syntax):

```rust
struct RowIdxScanNode {
  child: Arc<dyn ScanNode>,
}

impl ScanNode for RowIdxScanNode {
  fn evaluate(&self, row_range: Range<u64>, expr: Expr) -> Evaluation {
    // Create a sequence array representing the current row idx range.
    let row_range_array = SequenceArray::new(row_range);
    // Wrap the expression with a "let" to introduce the row_idx array into the expression scope.
    let wrapped_expr = LetExpr::new(expr, HashMap::from({"__row_idx": row_range_array}));
    // Pass the wrapped expression to our child to evaluate. Now any child expressions have access to the row_idx var.
    self.child.evaluate(row_range, wrapped_expr)
  }
}
```

joseph-isaacs:
would `ScanNode` have a `fn return_dtype(expr: Expr)` method as well?

joseph-isaacs:
Also how would you implement `FilterLayoutReader` match on DType in eval, if it is bool split conjuncts and call children with  each conjunct or move this into the outer scan?

gatesn:
> Also how would you implement FilterLayoutReader match on DType in eval, if it is bool split conjuncts and call children with each conjunct or move this into the outer scan?

Interestingly, the same optimisations we use to evaluate a filter can also be used to optimise a projection! I think we'd actually already handle this because the CNF thing fails when the expression cannot be split.

onursatici:
Currently `LayoutReader::dtype` is used as the input dtype:
https://github.com/vortex-data/vortex/blob/develop/vortex-layout/src/scan/mod.rs#L150-L154

I think we should have two different tree's on the scan, one of `ScanNode`'s that we use for expression evaluation, also to get the return_dtype per expression ( so scan nodes can decide to add virtual columns via scoped dtypes if they want), the other tree being layouts that the scan uses to register splits, and get the underlying dtype from the root layout.

onursatici:
If scan's knew about the root layout we could remove the `dtype` method from the layout readers, `register_splits` could live under layouts until we switch to a stream based scan

gatesn:
A Scan should not know about a Layout. I think we need to fix the crate organization / dependencies to make this obvious.

A layout is purely some information serialised into a file that is used to construct a LayoutReader/ScanNode. 

If you want a funky way of thinking about it, a Layout is like a serialised physical plan containing only array literals (inside each flat layout). So why should a Scan require both a serialised plan (Layout), and an in-memory plan (ScanNode)?

The problem I think we have, is that this analogy isn't quite correctly implemented. I had confused it a bit in the past (e.g. making LayoutReader deref to Layout). So we should work towards fixing this to be the case.

The big difference right now is that a LayoutReader is actually more like a ScanNodeFactory. You pass in an Expr to evaluate_projection and get back an evaluation. 

That's why the DType function on a LayoutReader is weird. What you're actually asking for is `fn dtype(&self, &Expr) -> DType`. 

---

Maybe ScanNode isn't quite right either then? As I say, the current LayoutReader is more like a ScanNodeFactory.

A couple of ideas come to mind:
- Perhaps name it something like EvaluationFactory... a bit gross?
- Bind the expression to the layout reader earlier, e.g. `VortexFile::new_layout_reader(&self, &Expr) -> LayoutReader`, then have something like `ScanBuilder::new().with_projection(reader).with_filter(reader)`. Also a bit weird?

onursatici:
Yea binding the expression earlier makes sense to me, then we can have the dtype method on the reader return the output dtype. ScanBuilder already takes the reader today in its construction so it is not that weird. We can rename that to ScanPlan that exposes the projection and filter from its given expression

 @joseph-isaacs lets discuss today about the api / naming and see if we can solve both concerns

joseph-isaacs:
I think we should go with this: 

```rust
ScanBuilder::new().with_projection(reader).with_filter(reader)
```

Having a single expr per `LayoutReader/ScanNode` seems like a nice idea, it will get rid of the expr caches all the way down?

onursatici:
Maybe to clarify the point on layout readers needing to know the scope dtype required for their children we can think about how struct layout needs to split the expressions per column.

If we have an expression like `(a == 3) or (b == row_id)`, and have a layout reader graph of `struct -> row_id -> flat`, struct reader while constructing the filter evaluation will try to split this expression. It needs to partition this by column references, and assign a dtype per partition. To get the dtype for the `b == row_id` split it would call `return_dtype` with the only dtype it knows of, which is the struct dtype from the layout. This would fail though as the row_id scope is absent from this dtype. 

To fix this it needs to be aware of all potential scopes any child under it would need. So we either make the expression bound to the layout readers so they all know their return_dtype and we can bubble that up, or we leave the current layout readers as expr eval factories but then they need a way for their parents to know what potential virtual scopes they need

gatesn:
The splitting gets really complex with let/var expressions I think.

In your example @onursatici, we would never have struct -> row-id -> flat, we would always wrap it as row-id -> struct -> flat.

That means the let(row-id) node is already in the expression passed to the struct layout reader.

---

@joseph-isaacs on the single-expr readers, yes it's nice, but I wonder if there's value in this "factory" level of abstraction (basically, maybe we want both!). Without it, it's quite hard to make sense of how filter columns/segments can be reused in the projection. It assumes the I/O and SegmentSource layers are aware enough to provide reuse. Whereas the current layout reader provides a SharedArrayFuture for explicitly reusing the arrays that appear in both the filter and project.

I think the easiest path forwards is to keep layout reader as a factory (I.e. each function takes an expression), then experiment with changing this later. Lumping it together with the row-id stuff will be very complicated because it has implications with re-use and prefetching that you won't want to be distracted with.

onursatici:
Yeap the way done in https://github.com/vortex-data/vortex/pull/3449 works by putting row_id on top. It works but the scan and the can_prune functions need to be aware of the row_id scope, which is a bit fragile and leaky but not too bad. I think we will merge that as is and do this potential refactor on our next virtual column reader

gatesn:
I don't understand why they need to be aware of the scope? The scope gets created by the let expression.

All layout readers assume an empty scope. We shouldn't even pass a scope into a layout reader. Flat layout creates its own scope

onursatici:
By scope I meant the scoped dtype, not the virtual array data. 

in https://github.com/vortex-data/vortex/pull/3449, in the scan for example, to get the dtype of the returned array stream, we cannot simply call `projection.return_dtype(layout_reader.dtype())` anymore. The input dtype that we have to pass to the projection has to include all possible virtual column dtypes, and there is currently no way to get that from the layout_reader. So we are doing:

```rust
         self.projection.return_dtype(
             &ScopeDType::new(self.layout_reader.dtype().clone())
                 .with_dtype_element(RowIdLayoutReader::row_id_scope_dtype()),
         )
     }
```
This works but it means the `row_id_scope_dtype` can't be contained only in the row id layout reader, we use it now in the scan, and also in duckdb file pruning for stats.

gatesn:
No it doesn't, the virtual columns are self-contained within the expression.

**Only** the let expression should mutate the scope. No one else, no layout readers.

The row idx layout reader should wrap up the expression with a let expression and pass it down

onursatici:
I agree with you that it is how it should be, and I don't have a strong opinion on this, but I don't think that it is quite possible to express this with the current layout reader api.

I think I have trouble understanding how these two constraints can co-exist
- only the let expression can mutate the scope, and the let expression scope is created in the row idx layout reader
- layout readers do not expose information about scopes (no scopes, no scoped dtypes)

With these two how would you get the result dtype from a scan? if the projection refers to row_id, then we need to know the dtype of it somewhere to return the right dtype, there are two ways I can think of:
- layout readers expose scoped_dtypes they need, which violates constraint 2
- scan calls return_dtype of the projection expression with the union of all virtual column dtypes in all layout readers it uses, currently only row_id. This violates constraint 1

PR #3513, opened 2025-06-11
duckdb: basic vortex scan

0ax1:


vortex-duckdb-ext/cpp/table_function.cpp
```diff
@@ -33,6 +33,11 @@ struct CTableGlobalData final : GlobalTableFunctionState {
 	}
 
 	optional_ptr<vortex::CData> ffi_data;
+
+	idx_t MaxThreads() const override {
+		return 1;
```
0ax1, line 38: For the current development, keep the thread count at 1.

gatesn: Just shove this on the table function vtable and make it a const on the TableFunction trait.

May want to make it non-const later, but for now that's still trivial to plumb through

vortex-duckdb-ext/src/scan.rs
```diff
@@ -74,43 +81,63 @@ impl TableFunction for VortexTableFunction {
...
             _column_names: column_names,
             _column_types: column_types,
         })
     }
 
-    fn function(
-        _bind_data: &Self::BindData,
-        _local_init: &mut Self::InitLocalData,
-        global_init: &mut Self::InitGlobalData,
+    fn scan(
```
joseph-isaacs, line 90: maybe in the next PR, call this `next_chunk`? Or something, scan is not really correct here

joseph-isaacs (approved):

vortex-duckdb-ext/src/scan.rs
0ax1: I'll meditate on that 🙂

vortex-duckdb-ext/cpp/table_function.cpp
0ax1: `MaxThreads` needs to be defined in terms of overriding
```
struct GlobalTableFunctionState {
public:
	// value returned from MaxThreads when as many threads as possible should be used
	constexpr static const int64_t MAX_THREADS = 999999999;

public:
	DUCKDB_API virtual ~GlobalTableFunctionState();

	virtual idx_t MaxThreads() const {
		return 1;
	}
```
in the subclass. So `MaxThreads` is called on the `GlobalTableFunctionState` instance we pass to 
```
extern "C" duckdb_state duckdb_vx_tfunc_register(duckdb_connection ffi_conn,
                                                 const duckdb_vx_tfunc_vtab_t *vtab) {
	if (!ffi_conn || !vtab) {
		return DuckDBError;
	}

	auto conn = reinterpret_cast<Connection *>(ffi_conn);
	auto tf = new TableFunction(vtab->name, {}, c_function, c_bind, c_init_global, c_init_local);
```

constness is given by the superclass and means that fields of the instance are not modified when this function is called.

I'd need to check but I assume `MaxThreads` is called once by DuckDB during query planning and is not dynamically adjusted. That said the constness of the member function doesn't prevent us from changing the thread count between scans.

My suggestion here would be pass another parameter when constructing `CTableGlobalData`, as the table function vtable is not accessible in a `GlobalTableFunctionState` subclass.

0ax1: @gatesn I'll merge this is in as is for now, and gonna bring this into shape based on further discussion in a follow up.

PR #2909, opened 2025-04-04
Scan row range

gatesn:
Useful for partitioning Vortex files.

PR #3514, opened 2025-06-11
fix: row_range vs row_indices pushdown in JNI

a10y:
Sometime around #2909, there must've been some semantic merge conflict or other careless change on my part.

Iceberg performs row-splitting for Vortex files and so it passes `rowIndices` in the form of `new long[2]{start, end}` for task creation.

However at some point we started passing this as the `rowIndices` and we removed `rowRange` as an option from the JNI. This adds it back, so that I can fix it in Iceberg.

cc @ashvina


a10y:
We should add better tests for Java so this doesn't break again because it was very annoying to track down

robert3005 (approved):

PR #3515, opened 2025-06-11
feat: scope dynamic symbol resolution to `vortex-python`

0ax1:
This PR scopes dynamic symbol resolution to `vortex-python`. When unresolved symbols lead to crashes at runtime, no context is provided why or where the code crashed, with the only error shown being: `SIGSEGV (11). Invalid memory access`.


robert3005 (approved):

0ax1:
Thanks @robert3005 ❤️

PR #3518, opened 2025-06-12
feat[vortex-dtype]: vx struct and list dtype ctors

joseph-isaacs:
I think this reduces the complexity of defining deeply nested structs


danking (approved):
Definitely better

PR #3521, opened 2025-06-12
fix: Expose Decimal Literal over Java

a10y:
We need to support pushing down expressions that reference Decimal literals for filters. Once available this will be exposed in https://github.com/spiraldb/iceberg/blob/main/vortex/src/main/java/org/apache/iceberg/vortex/ConvertFilterToVortex.java

### DIFF
--- a/java/vortex-jni/src/main/java/dev/vortex/api/expressions/proto/EndianUtils.java
+++ b/java/vortex-jni/src/main/java/dev/vortex/api/expressions/proto/EndianUtils.java
@@ -34,12 +34,22 @@ final class EndianUtils {
 
         // Determine target size (1, 2, 4, 8, 16, or 32 bytes)
         int targetSize;
-        if (bigEndianBytes.length <= 1) targetSize = 1;
-        else if (bigEndianBytes.length <= 2) targetSize = 2;
-        else if (bigEndianBytes.length <= 4) targetSize = 4;
-        else if (bigEndianBytes.length <= 8) targetSize = 8;
-        else if (bigEndianBytes.length <= 16) targetSize = 16;
-        else targetSize = 32;
+        if (bigEndianBytes.length <= 1) {
+            targetSize = 1;
+        } else if (bigEndianBytes.length <= 2) {
+            targetSize = 2;
+        } else if (bigEndianBytes.length <= 4) {
+            targetSize = 4;
+        } else if (bigEndianBytes.length <= 8) {
+            targetSize = 8;
+        } else if (bigEndianBytes.length <= 16) {
+            targetSize = 16;
+        } else if (bigEndianBytes.length <= 32) {
+            targetSize = 32;
+        } else {
+            throw new IllegalArgumentException(
+                    "BigDecimal with " + bigEndianBytes.length + " bytes overflows maximum Vortex decimal size");
+        }
 
         byte[] result = new byte[targetSize];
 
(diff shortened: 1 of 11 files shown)


vortex-proto/proto/scalar.proto
```diff
@@ -21,6 +21,7 @@ message ScalarValue {
     uint64 uint64_value = 4;
     float f32_value = 5;
     double f64_value = 6;
+    bytes decimal_value = 10;
```
robert3005: Why is this not bytes type?

a10y: That's the current behavior, but if you look at the TryFrom impl for pb::ScalarValue we lose the type information and always cast it back into InnerScalarValue::Bytes, and then you get a Scalar(dtype = decimal, inner = Bytes) which is a broken state

robert3005: why is that a broken state? That's the idea of scalarvalues is that they need to be interpretable to a scalar dtype and binary clearly can be, you can get `Scalar(dtype=float16, inner=u16)` which is handled just fine

robert3005: That is to say the type information of scalarvalue shouldn't be relevant

a10y: Ah I see, this is handled in `Decimal::as_scalar`

vortex-layout/src/segments/events.rs
```diff
@@ -151,7 +151,7 @@ impl SegmentSource for EventsSegmentSource {
 }
 
 /// A pending segment returned by the [`SegmentSource`].
-pub struct PendingSegment {
```
a10y, line 154: drive by, doesn't need to be pub

vortex-proto/proto/scalar.proto
a10y: Reverted the proto changes

java/vortex-jni/src/main/java/dev/vortex/api/expressions/proto/LiteralToScalar.java
```diff
@@ -174,6 +175,15 @@ public ScalarProtos.Scalar visitFloat64(Double literal) {
         }
     }
 
+    @Override
+    public ScalarProtos.Scalar visitDecimal(BigDecimal decimal, int precision, int scale) {
```
a10y, line 179: we need to pass precision/scale from the DType b/c they precision is not preserved through BigDecimal, and because `decimal` might be null for a null Decimal scalar

vortex-array/src/canonical.rs
```diff
@@ -10,16 +10,41 @@ use crate::arrays::{
 use crate::builders::builder_with_capacity;
```
a10y: wait these shouldn't be here

robert3005 (approved):
Had to remeber that java is bigendian

java/vortex-jni/src/main/java/dev/vortex/api/expressions/proto/EndianUtils.java
```diff
@@ -0,0 +1,60 @@
...
+        byte[] bigEndianBytes = unscaled.toByteArray();
+
+        // Determine target size (1, 2, 4, 8, 16, or 32 bytes)
+        int targetSize;
+        if (bigEndianBytes.length <= 1) targetSize = 1;
+        else if (bigEndianBytes.length <= 2) targetSize = 2;
+        else if (bigEndianBytes.length <= 4) targetSize = 4;
+        else if (bigEndianBytes.length <= 8) targetSize = 8;
+        else if (bigEndianBytes.length <= 16) targetSize = 16;
+        else targetSize = 32;
```
lwwmanning: is it possible to be bigger than this...? should we be checking that `bigEndianBytes.length <= 32`?

a10y: good point

PR #3524, opened 2025-06-12
feat: do not swallow errors from streams being written

danking:


vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -419,9 +421,21 @@ async fn call_for_first_item<T>(
...
 ) -> (BoxStream<'static, SequencedChunk>, VortexResult<T>) {
-    let Some(Ok((sequence_id, first_chunk))) = stream.next().await else {
+    let Some(result) = stream.next().await else {
         return (stream.boxed(), Err(vortex_err!("empty stream")));
     };
+    let (sequence_id, first_chunk) = match result {
+        Ok((sequence_id, first_chunk)) => (sequence_id, first_chunk),
+        Err(err) => {
+            return (
+                stream.boxed(),
```
onursatici: do we want to return `once(async {Err(err)})` here?

onursatici (approved):

PR #3525, opened 2025-06-12
Pco encoding

mwlon:


mwlon:
@gatesn @robert3005 I think this is ready for review

encodings/pco/src/array.rs
```diff
@@ -0,0 +1,357 @@
...
+        )?;
+        Ok(primitive.into_array())
+    }
+
+    #[allow(clippy::unwrap_in_result, clippy::unwrap_used)]
+    fn decompress_values_typed<T: Number>(&self) -> VortexResult<ByteBuffer> {
+        // To start, we figure out which chunks and pages we need to decompress, and with
+        // what value offset into the first such page.
+        let slice_n_rows = self.slice_stop - self.slice_start;
+        let mask = self.validity.to_mask(self.n_rows)?;
```
robert3005: there's `self.validity_mask()?`

encodings/pco/src/array.rs
```diff
@@ -0,0 +1,357 @@
...
+                    let chunk_end = cmp::min(n_values, chunk_start + values_per_chunk);
+                    let values = values.buffer::<T>();
+                    let chunk = &values.as_slice()[chunk_start..chunk_end];
+                    fc
+                        .chunk_compressor(chunk, &chunk_config)
+                        .map_err(vortex_err_from_pco)?
+                }
+            );
+
+            let mut chunk_meta_buffer = Vec::with_capacity(cc.chunk_meta_size_hint());
```
robert3005, line 189: I wanted to suggest replacing this with `ByteBufferMut` but it doesn't implement `Write` https://github.com/vortex-data/vortex/issues/3533

encodings/pco/src/test.rs
```diff
@@ -0,0 +1,105 @@
...
+    // check full decompression works
+    let decompressed = compressed.decompress().unwrap().to_primitive().unwrap();
+    assert_eq!(decompressed.as_slice::<i32>(), &data);
+
+    // check slicing works
+    let slice = compressed.slice(100, 105).unwrap();
+    for i in 0_i32..5 {
+        assert_nth_scalar!(slice, i as usize, 100 + i);
+    }
+    match slice.to_canonical() {
```
robert3005: there's shortcuts for these, i.e. `slice.to_primitive()`

robert3005 (approved):
Only a couple of nits

encodings/pco/src/array.rs
mwlon: So this doesn't work because I actually need the unsliced validity here in order to line up values with their indices in the primitive array. I'll change the naming to make that clearer.

But also this highlighted that the validity vtable here (and in zstd) won't work, since that expects the sliced validity. I'll see if there's a workaround there.

vortex/src/session.rs
```diff
@@ -45,6 +46,7 @@ impl Default for VortexSession {
             EncodingRef::new_ref(SequenceEncoding.as_ref()),
             EncodingRef::new_ref(SparseEncoding.as_ref()),
             EncodingRef::new_ref(ZigZagEncoding.as_ref()),
+            EncodingRef::new_ref(PcoEncoding.as_ref()),
```
lwwmanning: nit: the list is currently alphabetized, could you put PCoEncoding just before RunEndEncoding?

lwwmanning: relatedly @robert3005 @a10y do we want to feature flag PCoEncoding for now, or just ship straight to default?

mwlon: :+1:

encodings/pco/src/test.rs
mwlon: :+1:

vortex/src/session.rs
robert3005: I don’t think we need a compile time flag. Until we wire it into compressor adding an encoding doesn’t do anything to existing users.

encodings/pco/src/array.rs
mwlon: I'll add a new kind of validity helper for this

mwlon: done

vortex/src/session.rs
gatesn: Yes, we just need to be careful when we formalize "editions" to define the set of included encodings.

mwlon:
I think all comments are addressed now

PR #3528, opened 2025-06-13
feat[vortex-array]: add list contains compute fn

joseph-isaacs:
Make `list_contains` a compute fn, which specialises on the elements being search, not the list array.


encodings/sequence/Cargo.toml
```diff
@@ -27,6 +27,7 @@ vortex-scalar = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["full"] }
+vortex-array = { workspace = true, features = ["test-harness"] }
```
robert3005: this has to be a path dependency to not break publish

encodings/sequence/src/compute/compare.rs
```diff
@@ -65,6 +56,44 @@ impl CompareKernel for SequenceVTable {
...
+            .vortex_expect("base pvalue matching already validated");
+        let multiplier = multiplier
+            .as_primitive::<P>()
+            .vortex_expect("multiplier pvalue matching already validated");
+
+        find_intersection(base, multiplier, len, intercept)
+    })
+}
+
+fn find_intersection<P: NativePType + PartialOrd + Eq>(
```
robert3005: Use methods from NativePType, is_eq, is_ge, is_le. It is the same definition as the builin traits but avoids people following the same patter in other places

encodings/sequence/src/compute/compare.rs
```diff
@@ -26,23 +25,15 @@ impl CompareKernel for SequenceVTable {
         };
 
         // Check if there exists an integer solution to const = base + (0..len) * multiplier.
-        let set_idx: Option<usize> = match_each_integer_ptype!(lhs.ptype(), |P| {
-            let c = constant
+        let set_idx: Option<usize> = find_intersection_scalar(
```
robert3005: you don't need the annotaion here anymore since this is an explicit function call

robert3005 (approved):
just small nits

encodings/sequence/src/compute/compare.rs
joseph-isaacs: I wish clippy could help here

PR #3532, opened 2025-06-13
duckdb: file list via glob

0ax1:
As of now, this only handles local disk globs.


duckdb-vortex/src/vortex_scan.cpp
```diff
@@ -545,7 +545,6 @@ void RegisterScanFunction(DatabaseInstance &instance) {
 
 	vortex_scan.init_global = [](ClientContext &context,
 	                             TableFunctionInitInput &input) -> unique_ptr<GlobalTableFunctionState> {
-		auto &nonbind = input.bind_data->Cast<ScanBindData>();
```
0ax1, line 548: this var was unused

vortex-duckdb-ext/src/lib.rs
```diff
@@ -55,110 +55,3 @@ pub extern "C" fn vortex_extension_version() -> *const c_char {
     }
     .as_ptr()
 }
-
```
0ax1, line 58: these tests were moved to scan.rs, no changes

joseph-isaacs (approved):

PR #3535, opened 2025-06-13
fix: teach dict layout how to evaluate expressions with non-trivial nullability

danking:
The dict layout currently evaluates expressions on its values and then constructs a dict array of the result. This is semantically correct for non-nullable arrays. For nullable arrays, this forces the null semantics to be the trivial "poisoning" semantics: any expression whose leaves evaluate to null must also evaluate to null. This is not true for at least two expressions in our expression language:
- `is_null(...)`
- `pack{k1: v1, ...}`.

This PR introduces a test case for each expression and makes the most conservative change, simply evaluating the expression on the constructed dictionary array. If individual expressions can evaluate more cleverly, they should!

### DIFF
--- a/vortex-layout/src/layouts/dict/reader.rs
+++ b/vortex-layout/src/layouts/dict/reader.rs
@@ -159,11 +159,12 @@ impl LayoutReader for DictReader {
         row_range: &Range<u64>,
         expr: &ExprRef,
     ) -> VortexResult<Box<dyn ArrayEvaluation>> {
-        let values_eval = self.values_eval(expr.clone());
+        let values_eval = self.values_eval(root());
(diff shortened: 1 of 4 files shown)


vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -239,14 +241,111 @@ impl ArrayEvaluation for DictArrayEvaluation {
                 values_result
```
joseph-isaacs: Not part of this PR, but we need to do this push down into all containers all the way down.

onursatici (approved):

robert3005 (approved):

PR #3540, opened 2025-06-16
feat: add `StructArray::remove_column`

blaginin:
`remove_column` in arrow is so neat, can we have it too? `StructArray` isn't fully cheap to clone so made edit in-place


vortex-array/src/arrays/struct_/mod.rs
```diff
@@ -211,6 +211,30 @@ impl StructArray {
             self.validity().clone(),
         )
     }
+
+    /// Removes and returns a column from the struct array by name.
+    /// If the column does not exist, returns `None`.
+    pub fn remove_column(&mut self, name: impl AsRef<str>) -> Option<ArrayRef> {
```
joseph-isaacs: How do you expect to you this, usually StructArray are wrapped in an Arc

vortex-dtype/src/struct_.rs
```diff
@@ -268,6 +268,34 @@ impl StructFields {
...
     }
+
+    /// Returns a new [`StructFields`] without the field at the given index.
+    /// 
+    /// ## Panics
+    /// Panics if the index is out of bounds for the struct fields.
+    pub fn without_field(&self, index: usize) -> Self {
+        if index >= self.nfields() {
+            vortex_panic!("index out of bounds for struct fields");
+        }
```
joseph-isaacs, line 279: We seem to both panic and return option/result here, why did you choose to panic?

vortex-array/src/arrays/struct_/mod.rs
blaginin: `ArrayRef::to_struct` returns `StructArray` directly -- and that seems like the main way to get a struct?

joseph-isaacs: This https://github.com/vortex-data/vortex/blob/10a80eaba27d256bcbc8ef55819070ed358359c1/vortex-array/src/canonical.rs#L229 is a very expensive operation, that partially uncompress the array

vortex-dtype/src/struct_.rs
blaginin: no strong opinion here. the idea was that `without_field` is semantically equivalent to [remove](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove) which panics. also, user knows the element exists given that they're passing the index

vortex-array/src/arrays/struct_/mod.rs
blaginin: fair! I guess if I _know_ my array is a struct I should use `.as_::<StructVTable>`? if true feels like a good thing to add to the docs - happy to do that

but on your original point, if I have reference to `StructArray` this method would still implicitly clone all the fields. struct clone isn't that expensive but I don't want to do that implicitly, and also the method signature would be much uglier `(Option<ArrayRef>, ArrayRef)` so I think as is is better - but feel free to disagree 🙂

joseph-isaacs: It depends what you want to do, what do you want to do with the method?

robert3005: into_struct would not decompress the array, given we don't have other struct encodings this is fine for now. The change here would be to have a struct like view over other arrays.

robert3005: FWIW the project function (which is the general version of this function) has `pub fn project(&self, projection: &[FieldName]) -> VortexResult<Self>` signature

joseph-isaacs: It the array is a `ChunkedArray` there will be a partial decompression

joseph-isaacs (approved):

PR #3541, opened 2025-06-16
duckdb: impl clone for bind data

0ax1:


vortex-duckdb-ext/src/duckdb/table_function/bind.rs
```diff
@@ -19,6 +19,29 @@ pub(super) unsafe extern "C" fn bind_callback<T: TableFunction>(
     })
 }
 
+/// The native copy callback for bind data.
+pub(super) unsafe extern "C" fn clone_bind_data_callback<T: TableFunction>(
+    bind_data: *const std::ffi::c_void,
+    error_out: *mut cpp::duckdb_vx_error,
+) -> cpp::duckdb_vx_data {
+    try_or_null(error_out, || {
+        let bind_data = unsafe { &*(bind_data as *const T::BindData) };
```
joseph-isaacs: can we use .as_ref()?

vortex-duckdb-ext/src/scan.rs
```diff
@@ -9,14 +9,33 @@ use crate::duckdb::{
...
-    _column_names: Vec<String>,
+    first_file: VortexFile,
+    file_paths: Vec<PathBuf>,
+    column_names: Vec<String>,
     _column_types: Vec<LogicalType>,
 }
 
-pub struct VortexGlobalData {}
+impl std::fmt::Debug for VortexBindData {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
joseph-isaacs, line 38: why do you impl this for only a few fields?

vortex-duckdb-ext/src/scan.rs
```diff
@@ -140,8 +173,19 @@ impl TableFunction for VortexTableFunction {
...
 
-    fn init_global(_init: &TableInitInput<Self>) -> VortexResult<Self::GlobalState> {
-        Ok(VortexGlobalData {})
+    fn init_global(init: &TableInitInput<Self>) -> VortexResult<Self::GlobalState> {
+        let bind_data = init.bind_data();
+        let file_paths = SegQueue::new();
+
+        // Skip the first file path, as the file is opened during bind.
+        for path in bind_data.file_paths.iter().skip(1) {
+            file_paths.push(path.clone());
```
joseph-isaacs, line 187: can you not Extend?

joseph-isaacs (commented):
a few comment. will finish the review tomz. 

bit confused about why you clone and eq on only a few fields in global?

vortex-duckdb-ext/src/scan.rs
0ax1: `extend` doesn't exist on the lockfree `SegQueue`.

vortex-duckdb-ext/src/duckdb/table_function/bind.rs
0ax1: Hmm maybe. Though `as_ref` returns an `Option` and we're in an FFI function.

We could change all instances to do sth. along the lines of:

```
        let bind_data = unsafe {
            (bind_data as *const T::BindData)
                .as_ref()
                .ok_or(vortex_err!("bind_data is nullptr"))?
        };
```

vortex-duckdb-ext/src/scan.rs
0ax1: pr is still in draft, but i wanted to get the unit tests working which require the asserted types to impl Debug.

0ax1: done for all relevant fields by now

vortex-duckdb-ext/src/duckdb/table_function/bind.rs
0ax1: changed that for `bind_data_clone_callback`

vortex-duckdb-ext/cpp/table_function.cpp
```diff
@@ -24,6 +24,25 @@ struct CTableBindData final : TableFunctionData {
 	    : info(std::move(info_p)), ffi_data(ffi_data_p) {
 	}
 
+	bool Equals(const FunctionData &other_p) const override {
+		assert(info->vtab.bind_data_eq != nullptr);
+
+		auto &other = (CTableBindData &)other_p;
+		return info->vtab.bind_data_eq(this, &other);
+	}
```
joseph-isaacs: where is this used by duckdb?

0ax1: parquet doesn't use this => dropped.

vortex-duckdb-ext/src/scan.rs
```diff
@@ -14,13 +14,39 @@ use crate::exporter::ArrayIteratorExporter;
...
+    column_names: Vec<String>,
+    column_types: Vec<LogicalType>,
+}
+
+impl Clone for VortexBindData {
+    /// `VortexBindData` is cloned in case of multiple scan nodes.
+    fn clone(&self) -> Self {
+        Self {
+            _first_file: self._first_file.clone(),
+            filter_exprs: vec![],
```
joseph-isaacs, line 29: maybe explain why we clear this

joseph-isaacs (approved):

vortex-duckdb-ext/src/scan.rs
0ax1: sure

vortex-duckdb-ext/src/duckdb/table_function/mod.rs
```diff
@@ -21,7 +21,7 @@ use crate::{cpp, duckdb_try};
 ///
 /// This trait does not yet cover the full C++ API, see table_function.hpp.
 pub trait TableFunction: Sized + Debug {
-    type BindData: Send;
+    type BindData: Send + Clone;
```
gatesn, line 24: Why is this bound required? This isn't required by DuckDB, so it must mean our TableFunction implementation requires it?

gatesn (commented):
I don't see why bind data needs to be cloneable

vortex-duckdb-ext/cpp/table_function.cpp
```diff
@@ -24,6 +24,18 @@ struct CTableBindData final : TableFunctionData {
 	    : info(std::move(info_p)), ffi_data(ffi_data_p) {
 	}
 
+	unique_ptr<FunctionData> Copy() const override {
```
gatesn, line 27: I guess I don't understand why this needs a copy constructor?

0ax1: `BindData` gets cloned in case multiple scan nodes are used.

vortex-duckdb-ext/src/duckdb/table_function/mod.rs
0ax1: In order to work with multiple scan nodes, extensions need to override
```
unique_ptr<FunctionData> Copy() const override {
```
which in Rust land translates to `impl Clone`.

vortex-duckdb-ext/cpp/table_function.cpp
0ax1: See the Parquet ext for ref: https://github.com/duckdb/duckdb/blob/23f4de5996157fe435b6160e6bb88b5821b1fcab/extension/parquet/parquet_multi_file_info.cpp#L25

0ax1: It's not a copy ctor by the way. But part of the interface of `TableFunctionData` in form of a virtual func subclasses can override.

gatesn: Oh that's interesting!

I thought we wanted the scan concurrency to be handled in Rust though, not by DuckDB?

0ax1: Sorry by scan node I meant DuckDB scan node here. This is effectively just about making the bind data available across threads, and beyond that doesn't affect how we orchestrate the scan in Rust.

0ax1: So we're just hooking into the mechanism DuckDB provides for extensions to share bind data.

PR #3544, opened 2025-06-16
feat: migrate some vortex-expr transforms to use AccessPath

danking:
I have not audited this module, but I suspect there are more bugs lurking here. I will submit fixes as I encounter them in Spiral.

### DIFF
--- a/vortex-expr/src/transform/partition.rs
+++ b/vortex-expr/src/transform/partition.rs
@@ -3,14 +3,14 @@ use std::hash::{BuildHasher, Hash, Hasher};
 use std::sync::LazyLock;
 
 use itertools::Itertools;
-use vortex_dtype::{DType, FieldName, FieldNames, Nullability, StructFields};
+use vortex_dtype::{DType, Field, FieldName, FieldNames, Nullability, StructFields};
 use vortex_error::{VortexExpect, VortexResult, vortex_bail, vortex_err};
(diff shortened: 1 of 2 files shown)


vortex-expr/src/transform/partition.rs
```diff
@@ -253,14 +253,25 @@ impl MutNodeVisitor for ScopeStepIntoFieldExpr {
...
+                    [] => {
+                        return Ok(TransformResult::yes(pack(
+                            [(self.0.clone(), root())],
+                            Nullability::NonNullable,
+                        )));
+                    }
+                    [Field::Name(first_field_access), ..] => {
+                        assert_eq!(first_field_access, &self.0);
+                        return Ok(TransformResult::yes(root()));
+                    }
```
robert3005, line 270: Oh, ok you have to take the child that has `self.0` name. However, not every expr has names

vortex-expr/src/transform/partition.rs
```diff
@@ -281,6 +292,24 @@ impl MutNodeVisitor for ReplaceAccessesWithChild {
                 return Ok(TransformResult::yes(item.child().clone()));
             }
         }
```
robert3005: I guess you can remove this and replace the single element case

vortex-expr/src/transform/partition.rs
robert3005: but also this has to continue transforming

danking:
I'll retry this in a slightly different form.

PR #3548, opened 2025-06-17
perf[vortex-duckdb-ext]: add column pruning/projection

joseph-isaacs:


vortex-duckdb-ext/src/scan.rs
```diff
@@ -262,46 +310,70 @@ mod tests {
...
         assert_eq!(value, 42);
     }
 
     #[tokio::test]
     async fn test_vortex_scan_booleans() {
         let flags = vec![true, false, true, true, false];
         let flags_array = BoolArray::new(flags.into(), Validity::NonNullable);
-        let file = write_vortex_file("flag", flags_array).await;
-        let true_count: i64 = scan_vortex_file(
+        let file = write_single_column_vortex_file("flag", flags_array).await;
```
0ax1, line 398: Preference for only having one func for write and scan. The func names don't make this shorter either `write_single_column_vortex_file`.

0ax1: Could consider only using `scan_vortex_file` and `write_vortex_file` everywhere.

vortex-duckdb-ext/src/scan.rs
```diff
@@ -246,6 +281,19 @@ mod tests {
...
     }
 
+    fn scan_vortex_file<T>(tmp_file: NamedTempFile, query: &str) -> Result<Vec<T>, String>
+    where
+        T: duckdb::types::FromSql,
+    {
+        let conn = database_connection();
+        conn.prepare(query)
+            .unwrap()
+            .query_and_then([tmp_file.path().to_string_lossy()], |row| row.get(0))
```
0ax1: I do wonder whether we should pass in the column index to retrieve. Rather implicit using the constant here `row.get(0)`. wdyt?

0ax1 (approved):

Issue #3546, opened 2025-06-16
Support building on Stable Rust


robert3005:
Currently we use a number of nightly features. We should evaluate whether any of them is strictly necessary and whether there's a simpler alternative. More importantly removing one nightly feature doesn't offer an improvement unless we can get rid of all of them. Hence, if anyone wants to look into this issue they should start with `generic_const_exprs`

- vortex-buffer
- [ ] #![feature(min_specialization)]
- [ ] #![feature(trusted_len)]
- encodings/fastlanes
- [ ] #![feature(generic_const_exprs)]

gatesn:
A workable alternative here is that we actually provide a Rust API over Vortex FFI.

For many out-of-the-box use-cases (I.e. not doing any custom trait impls) this would be sufficiently powerful.

a10y:
Hmm, so an FFI-portable API over something like [abi_stable](https://docs.rs/abi_stable/latest/abi_stable/) could be interesting. But I feel like even the simpler case of using Vortex inside another codebase without forcing the entire project to nightly toolchain is also very useful. DataFusion and LanceDB both build on stable, we're probably putting ourselves at an adoption disadvantage by forcing users to nightly.

robert3005:
I think if we dynamically load encodings then we can have the core library be built on stable and then all the unstable use cases can be distributed as shared library if consumers don’t want use nightly

gatesn:
The big ones imo are trusted_len and min_specialization.

These are used in the stdlib to make vectors fast. We use the almost identically to make buffers fast. Removing them means our buffers aren't fast 🤷‍♂️

robert3005:
For that we can do the thing that arrow does and add some unsafe functions since trustedlen is a marker interface. However, users would need to know to call it. Right now just extend has the correct and optimal behaviour without any new apis

a10y:
At the end of last week I started working on another part of this, which I think is slightly controversial so I want to get people on board before I spend more time on it: I think we should move our `portable_simd` code over to instead use architecture-specific kernels, which get selected based on runtime feature detection (something that is already a problem which I made a ticket for in #3580).

Currently there's only one function we need to port, which is the TakeFn for PrimitiveArray. This mostly affects the performance of dict-encoded primitive arrays and is tracked via the dict_encode microbenchmark.

[...]

PR #3549, opened 2025-06-17
chore: remove most nightly features usage

a10y:
Part of addressing #3546 

Takes us from 18 nightly flags -> 5. Some were just totally unused, others were very easy to port to equivalent stable code.

This still leaves const generic exprs (for fastlanes), portable_simd for one function, and trustedlen specialization for vortex-buffer. I will deal with those in a follow up.

### DIFF
--- a/vortex-error/src/lib.rs
+++ b/vortex-error/src/lib.rs
@@ -95,15 +95,6 @@ pub enum VortexError {
     /// A wrapper for errors from the FlatBuffers library.
     #[cfg(feature = "flatbuffers")]
     FlatBuffersError(flatbuffers::InvalidFlatbuffer, Backtrace),
-    /// A wrapper for reader errors from the FlexBuffers library.
-    #[cfg(feature = "flexbuffers")]
-    FlexBuffersReaderError(flexbuffers::ReaderError, Backtrace),
-    /// A wrapper for deserialization errors from the FlexBuffers library.
-    #[cfg(feature = "flexbuffers")]
-    FlexBuffersDeError(flexbuffers::DeserializationError, Backtrace),
-    /// A wrapper for serialization errors from the FlexBuffers library.
-    #[cfg(feature = "flexbuffers")]
-    FlexBuffersSerError(flexbuffers::SerializationError, Backtrace),
     /// A wrapper for formatting errors.
     FmtError(fmt::Error, Backtrace),
     /// A wrapper for IO errors.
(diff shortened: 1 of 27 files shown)


vortex-layout/src/reader.rs
```diff
@@ -131,9 +131,27 @@ impl LazyReaderChildren {
         if idx >= self.cache.len() {
             vortex_bail!("Child index out of bounds: {} of {}", idx, self.cache.len());
         }
-        self.cache[idx].get_or_try_init(|| {
```
gatesn, line 134: You could implement get_or_try_init in an extension crate in vortex-common?

a10y:
Gets us down to just these:

image

vortex-expr/src/binary.rs
```diff
@@ -92,8 +92,8 @@ impl AnalysisExpr for BinaryExpr {
                 let min_rhs = self.rhs.min(catalog);
                 let max_rhs = self.rhs.max(catalog);
 
-                let left = min_lhs.zip_with(max_rhs, gt);
-                let right = min_rhs.zip_with(max_lhs, gt);
+                let left = min_lhs.zip(max_rhs).map(|(a, b)| gt(a, b));
+                let right = min_rhs.zip(max_lhs).map(|(a, b)| gt(a, b));
```
joseph-isaacs, line 96: Does 
```suggestion
                let left = min_lhs.zip(max_rhs).map(gt);
                let right = min_rhs.zip(max_lhs).map(gt);
```
work?

a10y: it doesn't, b/c the input to the closure is a tuple and the input to `gt` is two separate args

a10y:
I think we need to change how we capture backtraces and get off of `thiserror` since it forces you to nightly for backtrace

vortex-layout/src/reader.rs
a10y: I spent about five min poking and it is annoying to implement the logic of get_or_try_init identically without private field access

robert3005: The alternative is to use the once_cell crate https://docs.rs/once_cell/1.21.3/once_cell/sync/struct.OnceCell.html#method.get_or_try_init

a10y: I like that idea

a10y:
Alright, now these are all we have left: 
image

AdamGS:
the only idea I have for errors is to try `snafu`, I'll give it a spin tonight and see how it works and what the experience is like.
If it helps narrow down where errors occurred enough, I think giving up on backtraces until that interface is stable is OK - its not the best experience but its basically as-expected in stable rust

robert3005:
The whole backtraces in stable rust situation is very sad. Errors team doesn't want to make further changes but library team doesn't want to stabilise existing api because they think its slow. This has been the case for last 2 years. In the meantime the error has been moved out of std to core and whole reason the provide api exists is to workaround the fact that error is core and backtrace is std. ehhh... the greptime solution with manually capturing the locations instead of using backtrace could be the answer

a10y:
@AdamGS the errors were already redone, just needed to impl the things thiserror auto-derived without impling the nightly stuff

vortex-error/src/lib.rs
```diff
@@ -67,190 +66,82 @@ impl From<Infallible> for VortexError {
...
+    FlexBuffersReaderError(flexbuffers::ReaderError, Backtrace),
     /// A wrapper for deserialization errors from the FlexBuffers library.
     #[cfg(feature = "flexbuffers")]
-    #[error(transparent)]
-    FlexBuffersDeError(
-        #[from]
-        #[backtrace]
-        flexbuffers::DeserializationError,
-    ),
+    FlexBuffersDeError(flexbuffers::DeserializationError, Backtrace),
```
robert3005: if you're here remove the flexbuffers feature, it's no longer used

robert3005 (approved):

PR #3551, opened 2025-06-17
fix: add default fields when `canonicalize`-ing nullable structs

blaginin:
See https://github.com/vortex-data/vortex/issues/3553 for context


vortex-array/src/arrays/constant/canonical.rs
```diff
@@ -91,14 +91,24 @@ impl CanonicalVTable<ConstantVTable> for ConstantVTable {
...
-                });
+                        .collect::<Vec<_>>(),
+                    None => {
+                        debug_assert!(scalar.is_null());
+                        struct_dtype
+                            .fields()
+                            .map(|dt| {
+                                ConstantArray::new(Scalar::null(dt), array.len()).into_array()
+                            })
+                            .collect::<Vec<_>>()
```
blaginin: @joseph-isaacs I feel we can add nullable columns - `StructArray` assumes fields to never be empty:

https://github.com/vortex-data/vortex/blob/99ff1d1a3a7ec64c338d5c930c803289bc768b66/vortex-array/src/arrays/struct_/mod.rs#L101

vortex-scalar/src/lib.rs
```diff
@@ -178,6 +178,35 @@ impl Scalar {
...
+            }
+            DType::Utf8(nullability) => Self::utf8("", nullability),
+            DType::Binary(nullability) => Self::binary(Buffer::empty(), nullability),
+            DType::Struct(sf, nullability) => {
+                let fields: Vec<_> = sf.fields().map(Scalar::default_value).collect();
+                Self::struct_(DType::Struct(sf, nullability), fields)
+            }
+            DType::List(dt, nullability) => Self::list(dt, vec![], nullability),
+            DType::Extension(dt) => {
+                let scalar = Self::default_value(dt.storage_dtype().clone());
```
joseph-isaacs, line 206: Not sure this is a good default? I don't know a better one however

blaginin: I believe this copies the logic from `append_zeros`: 
https://github.com/vortex-data/vortex/blob/2e816f09a7265350ef7dadc70badcf266cefb7d1/vortex-array/src/builders/extension.rs#L71

vortex-array/src/arrays/constant/canonical.rs
```diff
@@ -91,14 +91,28 @@ impl CanonicalVTable<ConstantVTable> for ConstantVTable {
...
-                    fields
+                let fields: Vec<_> = match value.fields() {
+                    Some(fields) => fields
                         .into_iter()
                         .map(|s| ConstantArray::new(s, array.len()).into_array())
-                        .collect::<Vec<_>>()
-                });
+                        .collect(),
+                    None => {
+                        debug_assert!(validity.all_invalid()?);
```
joseph-isaacs: shall we make this an assert?

vortex-array/src/arrays/constant/canonical.rs
```diff
@@ -91,14 +91,28 @@ impl CanonicalVTable<ConstantVTable> for ConstantVTable {
...
+                        struct_dtype
+                            .fields()
+                            .map(|dt| {
+                                let scalar = match dt.nullability() {
+                                    Nullability::NonNullable => Scalar::default_value(dt),
```
joseph-isaacs, line 106: can you add a todo to remove this split once we decide on the default_value of nullable?

vortex-scalar/src/lib.rs
```diff
@@ -567,4 +597,32 @@ mod test {
             "{result:?}"
         );
     }
+
+    #[test]
+    fn zero_for_complex_dtype() {
+        let struct_dtype = DType::Struct(
+            Arc::new(vortex_dtype::StructFields::new(
```
joseph-isaacs: you can clean this up using `DType::struct_`

vortex-scalar/src/lib.rs
```diff
@@ -567,4 +597,32 @@ mod test {
...
+            Nullability::NonNullable,
+        );
+
+        let scalar = Scalar::default_value(struct_dtype.clone());
+        assert_eq!(scalar.dtype(), &struct_dtype);
+
+        let scalar = scalar.as_struct();
+
+        let a_field = scalar.field("a").unwrap();
+        assert_eq!(a_field.as_primitive().pvalue().unwrap(), PValue::I32(0));
```
joseph-isaacs, line 623: This value is really undefined and shouldn't relied, we could change this to return `ConstantArray(1)` tomorrow and it would still be correct, since there is a validity that is unapplied.

I think instead of want to assert that that the values of this field with a struct validity applied is all_invalid and also that the dtype of the fields are what you expect

joseph-isaacs (commented):
Just a few changes please

joseph-isaacs:
Nearly ready!

vortex-array/src/arrays/constant/canonical.rs
blaginin: sure, updated :)

vortex-array/src/arrays/constant/canonical.rs
blaginin: done!

vortex-scalar/src/lib.rs
blaginin: I don't think we have `DType::struct_` currently (we have `Scalar::struct_` but I need to create dtype here)

vortex-scalar/src/lib.rs
blaginin: I do agree that we shouldn't rely on the primitive value but instead check validity and `ptype` - for that I've added `test_canonicalize_nullable_struct` and `write_nullable_nested_struct`. 

However, this test doesn't test `canonicalize` but `default_value` directly - and here I think we should actually check the returned default value

joseph-isaacs: Oops, my bad.

vortex-scalar/src/lib.rs
joseph-isaacs: Try rebasing

blaginin: so much better!! 🙏

joseph-isaacs (approved):

PR #3555, opened 2025-06-17
feat: optionally include row index during scans

danking:


robert3005 (approved):
We have to figure out how to avoid adding all the boolean flags

danking:
Joe says I need to fix the dtype too

robert3005:
RowIdLayoutReader should return a correct dtype?

vortex-layout/src/scan/mod.rs
```diff
@@ -57,6 +57,8 @@ pub struct ScanBuilder<A> {
     file_stats: Option<Arc<[StatsSet]>>,
     /// Maximal number of rows to read (after filtering)
     limit: Option<usize>,
+    /// Include the within-file row index as a non-nullable u64 column with this name.
+    include_row_index_in_scope: bool,
```
a10y, line 61: with_row_index

vortex-layout/src/scan/mod.rs
```diff
@@ -90,6 +92,11 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
         self
     }
 
+    pub fn with_row_index_in_scope(mut self) -> Self {
```
a10y, line 95: with_row_index

PR #3559, opened 2025-06-18
fix: taiki-e/install-action nextest install

0ax1:
`taiki-e/install-action@v2` fails to install the `latest` nextest binary (`v0.9.99`). As an intermediate fix, we pin nextest to `v0.9.98`.


AdamGS (approved):

0ax1:
issue: https://github.com/taiki-e/install-action/issues/1005

PR #3560, opened 2025-06-18
datafusion 48

onursatici:


vortex-datafusion/src/lib.rs
```diff
@@ -47,26 +49,24 @@ fn supported_data_types(dt: DataType) -> bool {
...
-        },
-        Expr::Like(like) => {
-            can_be_pushed_down(&like.expr, schema) && can_be_pushed_down(&like.pattern, schema)
-        }
-        Expr::Literal(lit) => supported_data_types(lit.data_type()),
-        _ => {
-            log::debug!("DataFusion expression can't be pushed down: {expr:?}");
-            false
-        }
+fn can_be_pushed_down(expr: &PhysicalExprRef, schema: &Schema) -> bool {
```
AdamGS, line 52: this is a nice

AdamGS, line 52: was just thinking last night due we support new things here? We have `In` now too right? (@joseph-isaacs?)

AdamGS (approved):

AdamGS:
I think the tests fails because datafusion used to default to `Utf8` and now it defaults to `Utf8View`, and that was implicit somewhere in our conversion logic

AdamGS:
IDK what happened to query 6, but we're 20x worse and parquet is 25x worse. There were a bunch of changes to statistics and repartitioning recently and I wonder if that's the issue there, the query is:
```sql
SELECT MIN("EventDate"), MAX("EventDate") FROM hits;
```

vortex-datafusion/src/lib.rs
joseph-isaacs: But its a pretty slow rn, what type would it run over?

AdamGS: whatever is available I guess? if its not a perf win right now we can add that later

robert3005:
Wonder if they no longer correctly prune on file level statistics when they added partition group statistics

AdamGS:
@robert3005 thats my suspicion, parts of that were already in v47 but v48 has at least one more big change around that

onursatici:
I am also curious, will have a look, query 0 is also 10x @ parquet and 3x @ vortex

robert3005:
I filed an issue upstream https://github.com/apache/datafusion/issues/16444

AdamGS:
I think this fine to merge, we can re-enable the datafusion config (+ #3569) and once apache/datafusion#16447 is released it'll be the default behavior again.

PR #3563, opened 2025-06-18
chore: Cleanup some effectively unused deps

AdamGS:
Inspired by a conversation with @joseph-isaacs.

The next candidate for some dependency cleanup might be `jiff`, which is now tied with `zstd-sys` to be the biggest non arrow or vortex dependency of the core crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,7 +119,7 @@ memmap2 = "0.9.5"
 mimalloc = "0.1.42"
 moka = { version = "0.12.10", default-features = false }
 num-traits = "0.2.19"
-num_enum = {version = "0.7.3", default-features = false}
+num_enum = { version = "0.7.3", default-features = false }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -119,7 +119,7 @@ memmap2 = "0.9.5"
 mimalloc = "0.1.42"
 moka = { version = "0.12.10", default-features = false }
 num-traits = "0.2.19"
-num_enum = "0.7.2"
+num_enum = {version = "0.7.3", default-features = false}
```
a10y: suspect taplo will want you to reformat this

a10y (approved):

Cargo.toml
joseph-isaacs: `a10y -- cargo taplo`

robert3005 (approved):

PR #3571, opened 2025-06-19
chore[fuzz]: added decimal array ops

joseph-isaacs:
Would be nice to be able to be generic over a `NativePType` and a `NativeDecimalType`

### DIFF
--- a/vortex-array/src/arrays/struct_/mod.rs
+++ b/vortex-array/src/arrays/struct_/mod.rs
@@ -104,8 +104,8 @@ impl StructArray {
         for field in fields.iter() {
             if field.len() != length {
                 vortex_bail!(
-                    "Expected all struct fields to have length {length}, found {:?}",
-                    fields.iter().map(|f| f.len()).collect_vec(),
+                    "Expected all struct fields to have length {length}, found {}",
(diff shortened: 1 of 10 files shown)


AdamGS:
semver wise this isn't a feature - it doesn't really affect the public API

robert3005:
While you're here can you eliminate the fallthrough cases and enumerate them so when adding a new dtype you will have to at least acknowledge you're not implementing them. I completely missed this when reviewing decimaldtype prs

vortex-array/src/arrays/struct_/mod.rs
```diff
@@ -104,8 +104,8 @@ impl StructArray {
         for field in fields.iter() {
             if field.len() != length {
                 vortex_bail!(
-                    "Expected all struct fields to have length {length}, found {}",
-                    field.len()
+                    "Expected all struct fields to have length {length}, found {:?}",
+                    fields.iter().map(|f| f.len()).collect_vec(),
```
robert3005:
```suggestion
                    "Expected all struct fields to have length {length}, found {}",
                    fields.iter().map(|f| f.len()).format(","),
```
You will also need to import itertools::Itertools

robert3005 (approved):

PR #3576, opened 2025-06-19
chore[fuzzer]: add u256 to implemented `Int` for i256 correctly

joseph-isaacs:


robert3005 (approved):
And I thought I could just fake it with i256

joseph-isaacs:
I think it can lead to a negative range len which is added to a start producing an empty range

PR #3578, opened 2025-06-19
bug[decimal-byte-parts]: unconvertible scalar in compare

joseph-isaacs:


robert3005:
You really want to write a unit test

encodings/decimal-byte-parts/Cargo.toml
```diff
@@ -26,3 +26,6 @@ vortex-dtype = { workspace = true }
 vortex-error = { workspace = true }
 vortex-mask = { workspace = true }
 vortex-scalar = { workspace = true }
+
+[dev-dependencies]
+vortex-array = { workspace = true, features = ["test-harness"] }
```
robert3005: this has to be a path dependency

robert3005: because of silly publish reasons

joseph-isaacs: Why is the rule, here any dev dep or if there is a cycle?

joseph-isaacs: Would be nice to have a linter

robert3005: It creates a publish ordering cycle. Cargo is not clever about dev-dependencies.

robert3005: Or rather it doesn't properly register publish dependencies

robert3005 (approved):

PR #3579, opened 2025-06-19
feat: migrate take_primitive_simd to stable AVX2 kernel

a10y:
An implementation of TakeKernel for PrimitiveArray that uses AVX2 explicit instructions, falling back to Scalar. For non-x86_64 platforms the `portable_simd` implementation is still loaded if nightly compiler is being used.

This is part of the #3546 series of PRs.

Additionally: fixed a soundness issue with lack of bounds checking of indices for portable_simd impl. This requires us to do a full scan of the indices upfront before running the kernel to avoid out of bounds memory access.

## Implementation

The biggest source of added complexity in this PR is the new `avx2` module which implements a take kernel for primitive indices/values that uses the AVX2 GATHER operation.

Intel ISA provides 4 different gather intrinsics:

- `_mm256_i32gather_epi32` -> gathering 8x 32bit values with 32bit indices
- `_mm_i32gather_epi64` -> gathering 4x 64bit values with 32bit indices
- `_mm_i64gather_epi32` -> gathering 8x 32bit values with 32bit indices
- `_mm256_i32gather_epi32` -> gathering 8x 32bit values with 32bit indices

We implement a generic inner loop with a trait parameter `GatherFn<I, V>`, and allow specialization to insert the proper loop logic for each valid index/value type combination.


robert3005:
I wouldn't merge this and instead leave simd impl behind a feature flag. Portable simd probably gives us more than just avx2 impl?

vortex-array/src/arrays/primitive/compute/take/x86.rs
```diff
@@ -0,0 +1,363 @@
...
+
+/// AVX2-optimized take operation dispatch.
+///
+/// This returns None if the AVX2 feature is not detected at runtime, signalling to the caller
+/// that it should fall back to the scalar implementation.
+///
+/// If AVX2 is available, this returns a PrimitiveArray containing the result of the take operation
+/// accelerated using AVX2 instructions.
+#[cfg(target_arch = "x86_64")]
+pub fn take_primitive_avx2<I, V>(
```
a10y: this is still unsafe, b/c it doesn't prevent out of bounds memory access of indices into values.

we should probably do a straight-line check to see that all of the indices are  < values.len(). If indices are nullable, we also need to handle null positions which could be anything

vortex-array/src/arrays/primitive/compute/take/x86.rs
```diff
@@ -0,0 +1,363 @@
...
+
+/// AVX2 implementation for i32 values with u32 indices
+///
+/// # Safety:
+///
+/// Caller must ensure that all of the indices point to valid elements in the values array.
+/// Failure to do so will result in potentially accessing out of bounds memory.
+#[cfg(target_arch = "x86_64")]
+#[target_feature(enable = "avx2")]
+unsafe fn take_u32_i32_avx2(indices: &[u32], values: &[i32]) -> Buffer<i32> {
```
a10y: all of these are similar enough that we should probably macro them (or otherwise generify)

vortex-array/src/arrays/primitive/compute/take/x86.rs
```diff
@@ -0,0 +1,363 @@
...
+    let output_ptr: *mut i32 = buffer.spare_capacity_mut().as_mut_ptr().cast();
+    let values_ptr = values.as_ptr();
+
+    // Process chunks of 8 elements
+    let chunks = indices_len / SIMD_WIDTH;
+    for chunk_idx in 0..chunks {
+        let offset = chunk_idx * SIMD_WIDTH;
+
+        // Load the next 8 indices into a vector
+        let indices_vec = unsafe { _mm256_loadu_si256(indices.as_ptr().add(offset).cast()) };
```
a10y: interestingly, [tantivy uses the `_mm256_lddqu_si256`](https://github.com/quickwit-oss/tantivy/blob/436ec6caea6094ad71b1084fc9715b08705c6579/bitpacker/src/filter_vec/avx2.rs#L5) intrinsic instead, which according to the [Intel docs](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/mm256-lddqu-si256.html) seems to indicate that it's very similar to [mm256_loadu_si256](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/mm256-loadu-si256.html), and indeed [this StackOverflow answer](https://stackoverflow.com/a/47426790) also backs that up:

> There's no reason to ever use _mm256_lddqu_si256, consider it a synonym for _mm256_loadu_si256. lddqu only exists for historical reasons as x86 evolved towards having better unaligned vector load support, and CPUs that support the AVX version run them identically. There's no AVX512 version.

a10y:
@robert3005 I probably don't have as good intuition about this as you or Alex, but AFAICT, the most beneficial part of the existing portable_simd implementation is the [gather](https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/arrays/primitive/compute/take.rs#L104) operation, which exists on AVX2 but has no equivalent on NEON.

Similarly, from other things I've read, AVX512 generally executes a 512-bit load as two instructions instead of one, so the speedup is not really 2x it's something short of that. I think that might be why things like Tantivy which implement direct SIMD support only have avx2 kernels and don't bother with avx512.

a10y:
I'm also open to just shoving existing impl behind some nightly-only feature flag, if that's possible. I was just hoping that if this were valuable we can preserve it for everyone, since it is doable.

robert3005:
Ok, avx2 gather is probably the only widespread use simd implementation of this function so might be worth having a stable version

joseph-isaacs:
Are there any benchmarks for this?

0ax1:
> Are there any benchmarks for this?

This is covered by https://github.com/vortex-data/vortex/blob/develop/encodings/dict/benches/dict_compress.rs

0ax1:
> @robert3005 I probably don't have as good intuition about this as you or Alex, but AFAICT, the most beneficial part of the existing portable_simd implementation is the [gather](https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/arrays/primitive/compute/take.rs#L104) operation, which exists on AVX2 but has no equivalent on NEON.
> 
> Similarly, from other things I've read, AVX512 generally executes a 512-bit load as two instructions instead of one, so the speedup is not really 2x it's something short of that. I think that might be why things like Tantivy which implement direct SIMD support only have avx2 kernels and don't bother with avx512.

Yep, this is only about gather. Would be interesting though to also compare perf on macOS between portable simd and non-SIMD to double check that moving away from portable SIMD doesn't introduce a regression there. So the assumption is there shouldn't be, as there's no gather equivalent for NEON.

robert3005:
Interesting that i64 improved

a10y:
That's b/c the latest commit I hand-rolled a kernel for u8 indices and i64 values, now the AVX pathway is actually being hit

image

vortex-array/src/arrays/primitive/compute/take/x86.rs
```diff
@@ -100,6 +120,100 @@ where
...
+/// AVX2 implementation for i64 values with u8 indices
+///
+/// # Safety:
+///
+/// Caller must ensure that all of the indices point to valid elements in the values array.
+/// Failure to do so will result in potentially accessing out of bounds memory.
+#[cfg(target_arch = "x86_64")]
+#[target_feature(enable = "avx2")]
+unsafe fn take_u8_i64_avx2(indices: &[u8], values: &[i64]) -> Buffer<i64> {
+    const SIMD_WIDTH: usize = 4; // 256 bits / 32 bits per element
```
a10y: should say 64 bits per element

a10y:
Alright, I think I've convinced myself that this PR adds value.

One important thing to note: the old `portable_simd` implementation was unsound, b/c it failed to bounds-check the indices before issuing the gather instructions, meaning that you could very easily trigger unsafe memory access. So no matter what, bounds checking needs to be added and that will undoubtedly hurt performance a bit from the baseline numbers.

I did two CodSpeed runs:

1. **Run 1**: portable_simd Kernel only with bounds check added: https://codspeed.io/vortex-data/vortex/runs/compare/685c2abfbd0c18c2abcc32be..685c56398a5bab413a016f8d
2. **Run 2**: Using AVX2 with bounds check added: https://codspeed.io/vortex-data/vortex/runs/compare/685c2abfbd0c18c2abcc32be..685c4aa98a5bab413a016f07

Note that in Run1, all of the decode_primitives benchmarks regress considerably. In Run 2, a few regress (less than in Run1) but the majority speed up by up to 20% _despite_ the added bounds checking.

vortex-array/src/arrays/primitive/compute/take/avx2.rs
```diff
@@ -0,0 +1,350 @@
...
+
+                    // Write the vec out to dst.
+                    unsafe { $store(dst.cast(), values_vec) };
+                }
+            }
+        )*
+    };
+}
+
+impl_gather!(u8,
```
a10y, line 122: we should implement f32 and f64 as value types since there are relevant gather instructions for them. This PR is probably enough to review so I'd prefer to do in follow up

vortex-array/src/arrays/primitive/compute/take/avx2.rs
```diff
@@ -0,0 +1,350 @@
...
+    ///
+    /// This function can read up to `STRIDE` elements through `indices`, and read/write up to
+    /// `WIDTH` elements through `src` and `dst` respectively.
+    unsafe fn gather(indices: *const Idx, src: *const Values, dst: *mut Values);
+}
+
+/// AVX2 version of GatherFn defined for 32- and 64-bit value types.
+enum AVX2Gather {}
+
+unsafe fn identity<T>(input: T) -> T {
```
a10y: some of the impls don't need an `extend` operation (when the indices and values are the same size) so we use this and it should get optimized away

a10y:
Hmm, I'm not seeing the nice improvements anymore. I'd like to merge https://github.com/vortex-data/vortex/pull/3653 first so that way we have an apples-to-apples comparison

a10y:
Some nice speedups for several of these

image

Some slight regression for f32 (~8%). Those can be addressed pretty easily though.

vortex-array/src/arrays/primitive/compute/take/avx2.rs
```diff
@@ -0,0 +1,545 @@
...
+    fn take(
+        &self,
+        values: &PrimitiveArray,
+        indices: &PrimitiveArray,
+        validity: Validity,
+    ) -> VortexResult<ArrayRef> {
+        if values.ptype() != PType::F16
+            && indices.dtype().is_unsigned_int()
+            && indices.all_valid()?
+            && values.all_valid()?
```
robert3005: You don't need these, f16 can be reinterpreted casted to u16 and back. We can adapt the logic from the portable_simd kernel that I made

vortex-array/src/arrays/primitive/compute/take/avx2.rs
robert3005: use std::convert::identity instead?

robert3005 (changes requested):
I think apart from the mismatch between avx2 and portable simd on when we can use simd everything else looks good

vortex-array/src/arrays/primitive/compute/take/avx2.rs
a10y: Only 32/64 bit values are eligible for the kernel for now. There is a way to extend it to types narrower than dword but it's complex.

I have updated the kernel to add impls for f32/f64 though and updated the test macro to generate test cases for them

robert3005: Ah I missed that part. Portable simd makes this look very easy

robert3005 (approved):

PR #3583, opened 2025-06-20
chore[fuzzer]: add array casting op

joseph-isaacs:


fuzz/src/lib.rs
```diff
@@ -186,7 +190,21 @@ impl<'a> Arbitrary<'a> for FuzzArrayAction {
                         ExpectedValue::Array(current_array.to_array()),
                     )
                 }
-                _ => unreachable!(),
+                6 => {
```
robert3005, line 196: You need to adjust `actions_for_dtype` and `ALL_ACTIONS`

robert3005 (approved):

PR #3584, opened 2025-06-20
chore: add `StructFields::merge`

blaginin:
Forth time I had to write this logic 😀 Maybe we can have it here?

### DIFF
--- a/vortex-dtype/src/struct_.rs
+++ b/vortex-dtype/src/struct_.rs
@@ -1,7 +1,10 @@
 use std::hash::Hash;
 use std::sync::Arc;
 
-use vortex_error::{VortexExpect, VortexResult, VortexUnwrap, vortex_err, vortex_panic};
+use itertools::Itertools;
+use vortex_error::{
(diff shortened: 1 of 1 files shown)


vortex-dtype/src/struct_.rs
```diff
@@ -283,6 +286,33 @@ impl StructFields {
 
         StructFields::from_fields(names, dtypes)
     }
+
+    /// Merge two [`StructFields`] instances into a new one.
+    /// Order of fields in arguments is preserved
+    ///
+    /// # Errors
+    /// Returns an error if the merged struct would have duplicate field names.
+    pub fn merge(&self, other: &Self) -> VortexResult<Self> {
```
joseph-isaacs: shall we call this `disjoint_merge`?

joseph-isaacs (commented):
Later/now it might be nice to have a version of this that handles duplicates in a different way

joseph-isaacs (approved):

PR #3586, opened 2025-06-20
chore[fuzzer]: add projection and filter expressions to the fuzzer

joseph-isaacs:
I will refactor the fuzz into two [file and array] in a future PR


vortex-array/src/arrays/arbitrary.rs
```diff
@@ -38,7 +38,7 @@ fn split_number_into_parts(n: usize, parts: usize) -> Vec<usize> {
         .collect()
 }
 
-fn random_array(u: &mut Unstructured, dtype: &DType, len: Option<usize>) -> Result<ArrayRef> {
+pub fn random_array(u: &mut Unstructured, dtype: &DType, len: Option<usize>) -> Result<ArrayRef> {
```
robert3005: no need to be pub?

fuzz/src/lib.rs
```diff
@@ -224,3 +227,78 @@ fn actions_for_dtype(dtype: &DType) -> HashSet<usize> {
...
+    Ok(match u.int_in_range(0..=5)? {
+        0 => vortex_expr::Operator::Eq,
+        1 => vortex_expr::Operator::NotEq,
+        2 => vortex_expr::Operator::Gt,
+        3 => vortex_expr::Operator::Gte,
+        4 => vortex_expr::Operator::Lt,
+        5 => vortex_expr::Operator::Lte,
+        _ => unreachable!("range 0..=5"),
+    })
+}
```
robert3005: I would move this to the expr crate

fuzz/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(error_generic_member_access)]
+extern crate core;
```
robert3005: this doesn't need to be here

robert3005 (approved):

PR #3590, opened 2025-06-21
fix: make dylib rpath available for all targets

0ax1:
The dynamic library search path is defined relative to the executable location. Executables either reside under `target/<build_config>` or `target/<build_config>/deps`.

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -129,6 +129,9 @@ jobs:
 
       - name: Run ${{ matrix.benchmark.name }} benchmark
         shell: bash
+        env:
+          RUST_BACKTRACE: full
+          LD_LIBRARY_PATH: target/duckdb-v1.3.0
         run: |
           target/release_debug/${{ matrix.benchmark.id }} -d gh-json | tee ${{ matrix.benchmark.id }}.json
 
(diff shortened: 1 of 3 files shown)


.github/workflows/bench-pr.yml
```diff
@@ -129,6 +129,9 @@ jobs:
 
       - name: Run ${{ matrix.benchmark.name }} benchmark
         shell: bash
+        env:
+          RUST_BACKTRACE: full
+          LD_LIBRARY_PATH: target/duckdb-v1.3.0
```
0ax1, line 134: Setting this env var is not required locally. It's not clear to me yet why this is necessay on CI. The current setup otherwise works locally on macOS and Linux (Docker container).

vortex-duckdb-ext/build.rs
```diff
@@ -109,6 +109,7 @@ fn main() {
     // Link against DuckDB dylib.
     println!("cargo:rustc-link-search=native={}", lib_path.display());
     println!("cargo:rustc-link-lib=dylib=duckdb");
+    println!("cargo:rustc-link-arg=-Wl,-rpath,{}", lib_path.display());
```
0ax1, line 112: `vortex-duckdb-ext` also sets the rpath value in order to be self-contained in terms of linker flags. Ideally, this rpath value would propagate to other crates linking against `vortex-duckdb-ext` which is not the case though.

.github/workflows/bench-pr.yml
0ax1: I did a CI run, running `ls` as part of a build task to check whether the current working dir and duckdb dylib location is correct, which was the case.

joseph-isaacs (approved):
woo

PR #3592, opened 2025-06-22
chore: simplify rpath handling

0ax1:


bench-vortex/build.rs
```diff
@@ -0,0 +1,13 @@
+#![allow(clippy::unwrap_used)]
+#![allow(clippy::expect_used)]
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    const DUCKDB_VERSION: &str = "v1.3.0";
+
+    let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
```
gatesn, line 23: Would it be easier to statically link in vortex-duckdb-ext and avoid consumers having to configure anything?

0ax1: No, bc the duckdb static lib isn't self contained.

0ax1: As in, it pulls in symbols that are not defined in the static lib itself.

0ax1: Plus linking large static libs at compile time is slow.

joseph-isaacs (approved):
Does this mean we need to add this build rs each time we want to use duckdb-rs+vx

0ax1:
> Does this mean we need to add this build rs each time we want to use duckdb-rs+vx

Yes, but I also don't see other consumers besides the benchmarks linking against vortex-duckdb-ext coming up. Note that the actual extension build we will ship to ddb community is different.

PR #3593, opened 2025-06-22
duckdb: marshal query results via C API

0ax1:
Replaces duckdb-rs query result handling with custom marshaling via DuckDB C API. This change removes the dependency on duckdb-rs and provides more fine-grained control over querying and result processing.

Changes:
- Add `QueryResult` wrapper for C API marshaling
- Replace Connection::execute*() methods with query() returning `QueryResult`
- Implement result extraction via `TryFrom`
- Remove duckdb-rs fork dependency
- Update tests to use new interface


joseph-isaacs:
Mind adding an description explaining the motivation of this change.

joseph-isaacs:
I wonder if this has much overhead compared to the arrow exporter, maybe marshalling can be done after timing the run?

0ax1:
> I wonder if this has much overhead compared to the arrow exporter, maybe marshalling can be done after timing the run?

Marshelling is opt-in where the extend of the marshelling is given by how `QueryResult` is consumed.

E.g. in clickbench ddb2 we skip the marshelling:
```
    pub fn execute_query(&self, query: &str) -> Result<Duration> {
        // TODO: handle multiple queries
        trace!("execute duckdb query: {}", query);
        let time_instant = Instant::now();
        self.connection.query(query)?;
        let query_time = time_instant.elapsed();
        trace!("query completed in {:.3}s", query_time.as_secs_f64());

        Ok(query_time)
    }
```

The result pointer of a query always gets assigned in `cpp::duckdb_query(self.as_ptr(), query_cstr.as_ptr(), &mut result)`.

---

If you look at `develop`, e.g.
```
#[test]
fn test_scan_function_registration() {
    let conn = database_connection();
    let result: String = conn
        .prepare("SELECT function_name FROM duckdb_functions() WHERE function_name = 'vortex_scan'")
        .unwrap()
        .query_row([], |row| row.get(0))
        .unwrap();
    assert_eq!(&result, "vortex_scan");
}
```
`query_row` seems to have much more overhead due to the arrow layer, and under the hood eventually calls into:
```
    /// NOTE: if execute failed, we shouldn't call any other methods which depends on result
    pub fn execute(&mut self) -> Result<usize> {
        self.reset_result();
        unsafe {
            let mut out: ffi::duckdb_arrow = ptr::null_mut();
            let rc = ffi::duckdb_execute_prepared_arrow(self.ptr, &mut out);
            result_from_duckdb_arrow(rc, out)?;

            let rows_changed = ffi::duckdb_arrow_rows_changed(out);
            let mut c_schema = Rc::into_raw(Rc::new(FFI_ArrowSchema::empty()));
            let rc = ffi::duckdb_query_arrow_schema(out, &mut c_schema as *mut _ as *mut ffi::duckdb_arrow_schema);
            if rc != ffi::DuckDBSuccess {
                Rc::from_raw(c_schema);
                result_from_duckdb_arrow(rc, out)?;
            }
            self.schema = Some(Arc::new(Schema::try_from(&*c_schema).unwrap()));
            Rc::from_raw(c_schema);

            self.result = Some(out);
            Ok(rows_changed as usize)
        }
    }

```

vortex-duckdb-ext/src/duckdb/query_result.rs
joseph-isaacs, line 1: Do we have any tests for this, or should you say e2e is enough?

vortex-duckdb-ext/src/vortex_scan_tests.rs
joseph-isaacs, line 1: Did this change at all?

a file rename makes this harder to look at

vortex-duckdb-ext/src/duckdb/query_result.rs
joseph-isaacs, line 1: Is this a complete set of types? Does this require a lot more marshalling to get the rest?

joseph-isaacs:
This is a good find! 

It does however rely on another marshalling layer from duckdb to rust primitive think this can be unified with the vx conversion, maybe using vortex scalar or vx-scalar using this marshalling.

vortex-duckdb-ext/src/vortex_scan_tests.rs
0ax1: No the file didn't change, except for that I scoped it to
```
+#[cfg(test)]
+mod tests {
```

vortex-duckdb-ext/src/duckdb/query_result.rs
0ax1: This is not complete, but we can expand on this on demand, as this is only  used in unit tests or when we want to verify the results for tpch or clickbench.

0ax1: I think the tests in https://github.com/vortex-data/vortex/pull/3593/files#diff-26b97f72d3d96a34817dd141b73fcc801b8a45b104bc4d3f6a7798b19d4bbd47R150 are a good start. But yeah, maybe we want to expand on this over time to get full coverage, and less e2e style.

0ax1: Or as an intermediate step for marshelling to Vortex as you mentioned.

joseph-isaacs (approved):

PR #3594, opened 2025-06-22
bug[vortex-array]: take nullability chunked array not respects validity

joseph-isaacs:
Chunked array current ignores indices validity

### DIFF
--- a/vortex-array/src/arrays/chunked/compute/take.rs
+++ b/vortex-array/src/arrays/chunked/compute/take.rs
@@ -1,7 +1,6 @@
 use vortex_buffer::BufferMut;
 use vortex_dtype::{DType, Nullability, PType};
 use vortex_error::VortexResult;
-use vortex_mask::Mask;
 
 use crate::arrays::chunked::ChunkedArray;
 use crate::arrays::{ChunkedVTable, PrimitiveArray};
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/chunked/compute/take.rs
```diff
@@ -1,44 +1,111 @@
...
+            indices,
+            &DType::Primitive(PType::U64, indices.dtype().nullability()),
+        )?
+        .to_primitive()?;
+
+        if indices.dtype().is_nullable() {
+            take_nullable(array, indices.as_slice::<u64>(), indices.validity_mask()?)
+        } else {
+            take_non_nullable(array, indices.as_slice::<u64>())
+        }
```
robert3005: I'm curious if there's a significant slowdown to adding a mask slice if the mask is alltrue.

vortex-array/src/arrays/chunked/compute/take.rs
```diff
@@ -1,44 +1,111 @@
...
+    let mut prev_chunk_idx = array.find_chunk_idx(indices[0].try_into()?).0;
+    for idx in indices {
+        let idx = usize::try_from(*idx)?;
+        let (chunk_idx, idx_in_chunk) = array.find_chunk_idx(idx);
+
+        if chunk_idx != prev_chunk_idx {
+            // Start a new chunk
+            let indices_in_chunk_array = PrimitiveArray::new(
+                indices_in_chunk.clone().freeze(),
+                Validity::Array(indices_validity.slice(start, stop - start).into_array()),
```
robert3005: This should be `Validity::from(...)` instead of Array

vortex-array/src/arrays/chunked/compute/take.rs
joseph-isaacs: Hmm. Yep. I'll include and rename the non null arm

robert3005: as in - do we need to distinguish nullable and non nullable branches at all? In non nullable case the Mask is going to be AllTrue and we just pay a little bit of extra overhead of calling slice to resize it

joseph-isaacs: There is definitely some overhead, hard to know how much, since the fn is pretty small shall we keep it?

robert3005: I would keep it simple for now and refactor if necessary in the future

joseph-isaacs: Sure if you prefer

robert3005 (approved):

PR #3596, opened 2025-06-22
chore[vortex-expr]: move core expressions into their own module

joseph-isaacs:
do we want to call these core, or something else?


robert3005:
I don't get what `core` here means. You want to sperate expressions from the utilities for transforming them?

joseph-isaacs:
I want to put all the (core) expression definitions into their own crate. Not sure what a good name is. The current setup is a little confusing as to what is what. I want to make a distinction between expr trait & definitions and specific expression instances.

robert3005:
call it `exprs` and we can get it merged

robert3005 (approved):

PR #3597, opened 2025-06-23
fix[fuzzer/vortex-array]: use optional indicies in fuzzer `take`/`Decimal::take`

joseph-isaacs:


vortex-array/src/arrays/primitive/compute/take.rs
```diff
@@ -42,6 +41,9 @@ impl TakeKernel for PrimitiveVTable {
             });
         }
 
+        // TODO(joe): if the take indicies validity is tc is lower,
+        // we should merge the validity and value lookup
```
robert3005: What does this comment mean?

joseph-isaacs: rewritten it

robert3005 (approved):

PR #3598, opened 2025-06-23
perf[vortex-array]: chunked take all-valid special case

joseph-isaacs:


vortex-array/src/arrays/chunked/compute/take.rs
```diff
@@ -16,58 +17,102 @@ impl TakeKernel for ChunkedVTable {
...
-                    indices_in_chunk_array.as_ref(),
-                )?);
-                indices_in_chunk.clear();
-                start = stop;
-            }
-
-            indices_in_chunk.push(idx_in_chunk as u64);
-            stop += 1;
-            prev_chunk_idx = chunk_idx;
+        if indices.dtype().is_nullable() {
```
robert3005: Nullability is the logical concept. We should switch here whether `indices.all_valid()` and then propagate nullability

robert3005:
I am running some profiles to figure out if there's a different fix, i.e. is there some overhead we don't know about

joseph-isaacs:
lets see what we can see and go from there

robert3005:
Seems that it was a one off slowdown correlated with the previous commit merging. There's no sign of this logic in profiles

PR #3599, opened 2025-06-23
feat: pack & merge display their nullability

danking:
Also, expose the nullability of a pack.

### DIFF
--- a/vortex-expr/src/merge.rs
+++ b/vortex-expr/src/merge.rs
@@ -52,7 +52,11 @@ impl Display for Merge {
             .iter()
             .format_with(", ", |expr, f| f(expr))
             .fmt(f)?;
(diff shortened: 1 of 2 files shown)


joseph-isaacs:
Merge seems to drop the validity mask of each struct being merged. Do we want this? This seems either like a bug or at least undocumented.

robert3005 (approved):

danking:
I agree merge's nullability seems wrong but insofar as it has it I'd like to see that fact 😉

joseph-isaacs:
does empty pack look like this {}?

joseph-isaacs:
this is the same as the dtype {}?, maybe we should disambiguate?

joseph-isaacs:
can follow up

PR #3601, opened 2025-06-23
chore: More idiomatic display impls for expressions

robert3005:


robert3005:
@danking small followup I noticed after you made your pr

vortex-expr/src/merge.rs
```diff
@@ -47,16 +47,12 @@ pub fn merge(
...
-            .format_with(", ", |expr, f| f(expr))
-            .fmt(f)?;
-        f.write_str("}")?;
-        if Nullability::Nullable == self.nullability {
-            f.write_str("?")?;
-        }
-        Ok(())
+        write!(
+            f,
+            "{{{}}}{}",
```
joseph-isaacs: I think we should use `merge({...})` or something similar

joseph-isaacs (commented):
{{{}}} is crazy

vortex-expr/src/merge.rs
robert3005: sure, I guess same for pack

vortex-expr/src/merge.rs
```diff
@@ -47,16 +47,12 @@ pub fn merge(
...
-            .format_with(", ", |expr, f| f(expr))
-            .fmt(f)?;
-        f.write_str("}")?;
-        if Nullability::Nullable == self.nullability {
-            f.write_str("?")?;
-        }
-        Ok(())
+        write!(
+            f,
+            "merge({}){}",
```
robert3005, line 52: I removed `{...}` from formatting since there's no names here

joseph-isaacs (approved):

PR #3602, opened 2025-06-23
chore: kill duckdb ci build tasks

0ax1:


robert3005 (approved):

.github/workflows/sql-benchmarks.yml
```diff
@@ -76,8 +76,8 @@ jobs:
         run: |
           cargo build --bin ${{ matrix.binary_name }} --package bench-vortex --profile release_debug
 
-      - name: Install ninja & gzip
```
robert3005, line 127: do we still need gzip?

robert3005: we do

0ax1: kept that

bench-vortex/src/tpch/duckdb.rs
```diff
@@ -94,15 +94,10 @@ pub fn generate_tpc(opts: DuckdbTpcOptions) -> Result<PathBuf> {
...
-    let is_local_duckdb = opts.duckdb_path.is_some();
-
     let mut command = Command::new(opts.duckdb_path.unwrap_or_else(|| PathBuf::from("duckdb")));
-
-    let vortex_path = vortex_duckdb_extension_path();
     command
-        .arg("-unsigned")
         .arg("-c")
-        .arg(format!("load \"{}\";", vortex_path.to_string_lossy()));
+        .arg("install vortex from community; load vortex;");
```
joseph-isaacs, line 100: I am not sure we want to do this this will regress all our benchmarks

joseph-isaacs (dismissed):
I noticed that 

```rust
            if !is_local_duckdb {
                command
                    .arg("-c")
                    .arg("install vortex from community; load vortex;");
            }
```

Means that we will in fact use the up to date impl of vortex-duckdb

bench-vortex/src/tpch/duckdb.rs
joseph-isaacs: You can see this in clickbench

0ax1: Interesting this is only about writing the files though.

0ax1: Not about scanning.

0ax1: I agree that this is an issue. This would be go away though once we implement the write path in v2.

PR #3603, opened 2025-06-23
feat[vortex-expr]: CheckedAdd operator

danking:


vortex-expr/src/binary.rs
```diff
@@ -381,6 +390,33 @@ where
     iter.reduce(and)
 }
 
+/// Create a new `BinaryExpr` using the `CheckedAnd` operator.
```
robert3005: typo

danking: done.

vortex-expr/src/exprs/operators.rs
```diff
@@ -16,6 +16,9 @@ pub enum Operator {
     // boolean algebra
     And,
     Or,
+    // arithmetic
+    /// The sum of the arguments, erring if the sum would overflow or underflow.
+    CheckedAdd,
```
joseph-isaacs, line 25: What are the null semantics?

danking: I documented the behavior. It inherits the behavior from the compute function which inherit from arrow which is: if the index is null in either argument, it is null in the result.

robert3005 (approved):
naming is hard. Would like there to be regular Add but at this level this is fine

gatesn:
What if this was just "add" with unchecked as part of the options?

robert3005:
Is it wrapping or saturating?

robert3005:
I think we can call the default version just add. Which would follow more rust naming

gatesn:
Yeah, maybe better to be explicit. In which case this should still be called "Add" since checked is the default?

robert3005:
Rust is funky but we should call this Add. Rust is Checked in debug mode and Wrapping in release mode

PR #3605, opened 2025-06-23
feat: teach ScanBuilder with_row_index

danking:


vortex-layout/src/scan/mod.rs
```diff
@@ -146,15 +155,23 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
...
+
+    /// Returns the output [`ScopeDType`] of the scan.
+    pub fn scope_dtype(&self) -> ScopeDType {
+        let scope_dtype = ScopeDType::new(self.layout_reader.dtype().clone());
+        if self.row_index {
+            scope_dtype.with_dtype_element(RowIdLayoutReader::row_id_scope_dtype())
+        } else {
+            scope_dtype
+        }
     }
```
robert3005: Can this logic live in `RowIdLayoutReader`?

danking: The row_id_scope_dtype is already there. The only way I could use a hypothetical `RowIdLayoutReader::scope_dtype` is if I construct an actual RowIdLayoutReader that wraps `self.layout_reader`. I'll push that change but it seems a bit silly to create the layout reader just to get its scope dtype.

I considered mutating the layout_reader in `ScanBuilder::with_row_index`, but I think that means it's possible for the filter expression to not have row index in its scope. I think I need the RowIdLayoutReader to be the outer-most layout reader. cc: @joseph-isaacs for thoughts?

danking: Okay pushed the change. For posterity, this was what we had before. I'm not convinced the current version is much better
```rust
        let scope_dtype = ScopeDType::new(self.layout_reader.dtype().clone());
        if self.row_index {
            scope_dtype.with_dtype_element(RowIdLayoutReader::row_id_scope_dtype())
        } else {
            scope_dtype
        }
```

vortex-layout/src/scan/mod.rs
```diff
@@ -146,15 +155,22 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
...
         }
     }
 
     /// Returns the output [`DType`] of the scan.
     pub fn dtype(&self) -> VortexResult<DType> {
-        self.projection.return_dtype(
-            &ScopeDType::new(self.layout_reader.dtype().clone())
-                .with_dtype_element(RowIdLayoutReader::row_id_scope_dtype()),
-        )
+        self.projection.return_dtype(&self.scope_dtype())
```
robert3005: I thought this could be `self.projection.return_dtype(&ScopeDType::new(self.layout_reader.dtype().clone()))`

robert3005: and then rest of this logic moves to `RowIdLayoutReader::dtype`

robert3005 (commented):
I think something got lost in translation. Why do we need extra methods here instead of fixing `LayoutReader::dtype` for `RowIdLayoutReader`

gatesn:
This has become pretty weird imo.

We used to have let/var expression. This is a super general purpose way of introducing extra fields into the scope. The let expression would mutate the scope before invoke its child expression. (In other words, the caller doesn't mutate the scope! In fact, mutation should be pub(crate))

I think we decided to remove let because it's a bit too powerful for what we need, and makes transformations much more difficult to implement.

But in that case, we should also remove var. Otherwise var just becomes a way of referencing things we hope to exist in the scope at invocation time. Let/Var is self-contained. The expression is only valid if the corresponding let exists for each var. Without let, the expression is under-defined and the caller, at all call sites, must know which items to introduce into the scope.

If we don't like the complexity of let, then another option is to hard-code row_idx, since it's a pretty common and core functionality. The Scope then holds a row_offset and the caller is only responsible for configuring the offset. The row_idx expression then returns a Sequence array based on the scope's row offset.

The final option, is to have a RowIdx expression, essentially just a placeholder logical expression. Prior to execution, it must be turned into a physical expression, e.g. replaced with an ArrayLiteral, or even internally mutated (Ergh). This is permitted because Vortex expressions are a mutant combination of logical and physical expressions. But still, given a "physical" expression, a caller doesn't need any knowledge to build a scope and invoke the expression.

Side note: row_idx should not know anything about file_idx. Files aren't a thing in the context. The type should be u64.

danking:
cc: @robert3005. I think Nick's comment essentially captures the full scope (heh) of the problems.

What I ultimately need is two things:
1. A way to ask ScanBuilder to include a row index.
2. A way to know the scope (which means: the input array dtype and the dtypes of any other bound names, e.g. `$vx.row_id`).

I need the latter because I, Spiral, want to compute the return type of my projection expression before I evaluate the scan.

I cannot use only `LayoutReader::dtype` because that only tells me about the input array. To make `LayoutReader::dtype` work, RowIdLayoutReader must return something like `{"row_id": u64, "data": child.dtype()}`. That has the annoying side-effect that the VortexExpr must change depending on whether you are or are not using row ids.

I can unblock myself in Spiral by inlining some code over there. I'm reticent to block Spiral on resolving this PR because it seems to me consensus will take some time.

gatesn:
LayoutReader::dtype should disappear.

It's a hangover from when a LayoutReader: Deref<Target=Layout>.

What you actually want is LayoutReader::dtype(&Expr)

That should give the return_dtype of the expression as evaluated by the layout reader.

danking:
`LayoutReader::dtype(&Expr)` is tolerable as long as either there is a `ScanBuilder::dtype(&Expr)` (or it is my job to give ScanBuilder a RowIdLayoutReader, which is fine).

joseph-isaacs:
Shall we continue the larger discussion in one place, maybe here: https://spiraldb.slack.com/archives/C07BV3GKAJ2/p1749223688531319

gatesn:
ScanBuilder::dtype returns the dtype using the projection expression.

danking:
@joseph-isaacs @robert3005 @gatesn. I added `scope_dtype` to `LayoutReader`. Does this look better?

`LayoutReader::dtype(...)` doesn't feel right because we'd have to pass the layout reader to, for example, `simplify_merge` and `simplify_select`.

gatesn:
@joseph-isaacs we should discuss here in the open rather than Slack

@danking not really... what does scope_dtype mean? A layout reader doesn't have a scope?

danking:
@gatesn  I don't have opinions on whether they should or should not, but LayoutReaders indeed have a scope. In particular RowIdLayoutReader evaluates its (projection, filter, pruning) expressions in a scope that includes a row index.

gatesn:
I guess that's my objection to this PR then. 

A LayoutReader should not mutate a scope. In fact, it shouldn't even have a scope? It just passes expressions (possibly rewriting them) to child layout readers.

A RowIdxLayoutReader could rewrite the expression and replace any RowIdx expressions (currently Var("row_id")) with an ArrayLiteral representing the row indices.

Or it could wrap the expression tree in a LetExpression, which itself would mutate the scope.

But I don't see a world in which a RowIdxLayoutReader would even be evaluating an expression.

Happy to jump on a call tomorrow to discuss if that's an easier medium.

Edit: I see that we partition the expression into a RowIdx part and a non-RowIdx part. This may/may not be the best approach (non-trivial to benchmark, and it works for now, so let's run with it). But I don't see therefore why anyone needs to know about the scope outside of this Evaluation? The LayoutReader::dtype(Expr) function would do the same partitioning, and construct a ScopeDType when invoking.

onursatici:
relevant prior discussion: https://github.com/vortex-data/vortex/pull/3512

danking:
> The LayoutReader::dtype(Expr) function would do the same partitioning, and construct a ScopeDType when invoking.

Consider lines 179-182 of `scan/mod.rs`: 
```
        let ctx = ScopeDType::new(layout_reader.dtype().clone());

        // Normalize and simplify the expressions.
        let projection = simplify_typed(self.projection.clone(), &ctx)?;
```
To simplify an expression I need a ScopeDType. A DType is insufficient. Spiral has similar needs. In `develop`, the type of a Vortex expression depends on the data types of all the free variables: both Identity and the named ones (like `$vx.row_id`).

We could pass a LayoutReader instead of ScopeDType, but ScopeDType is the parsimonious representation of the necessary information.

Maybe your point is that RowIdLayoutReader needs to be rewritten in a different way? That's fine, but in the meantime I would like to make the smallest/easiest change to Vortex that allows Spiral to use the row index machinery. Is there some other small change that's acceptable?

Another alternative is for Spiral to just skip the ScanBuilder machinery entirely which is how [vortex-ffi](https://github.com/vortex-data/vortex/blob/263f28a63245e307479552d521af908f62ed7397/vortex-ffi/src/file.rs#L254) avoids these issues.

gatesn:
I'm saying there shouldn't ever be free variables. That's what the LetExpr was for, to ensure within an expression the variable is at least bound to a type. Removing LetExpr created the problem of free variables.

There are two other ways to remove the row_idx free variable:
1. Hard-code row-idx as a concept in Vortex expressions. There is a RowIdx expression, and it returns a SequenceArray based on Scope::row_offset().
2. Have RowIdxLayoutReader rewrite the expression tree _replacing_ the free variable with an ArrayLiteral containing a SequenceArray.

danking:
Should pass DCO now

robert3005 (approved):
We know this is not the implementation we want but in order to unblock functionality dependent on this feature we can merge this and work on the replacement which is #3671

PR #3606, opened 2025-06-23
bug[vortex-duckdb-ext]: fix the scan termination condition

joseph-isaacs:
A whole export can be empty so need to keep looking for another chunk before returning

We cannot currently tests this, however I manually verified that a few queries where correct

### DIFF
--- a/vortex-duckdb-ext/src/duckdb/query_result.rs
+++ b/vortex-duckdb-ext/src/duckdb/query_result.rs
@@ -1,5 +1,6 @@
 use std::ffi::CStr;
 
+use bitvec::macros::internal::funty::Fundamental;
 use vortex::error::{VortexResult, vortex_err};
 
 use crate::{cpp, wrapper};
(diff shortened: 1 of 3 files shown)


vortex-duckdb-ext/src/duckdb/query_result.rs
```diff
@@ -94,6 +95,12 @@ impl QueryResult {
         T::try_from(value)
     }
 
+    pub fn cell_as_str(&self, col_idx: usize, row_idx: usize) -> VortexResult<String> {
+        self.check_bounds(col_idx, row_idx)?;
+
+        unsafe { QueryResultCell::new(self.as_ptr(), col_idx, row_idx) }.to_string()
```
0ax1, line 101: you can just call
```
result.get::<String>(1, 0)
```
no need for an extra func

joseph-isaacs: That will not cast (e.g. i8 to string)

joseph-isaacs: `duckdb_value_string` does but the try impl uses `duckdb_value_varchar` is that a bug?

0ax1 (approved):

PR #3612, opened 2025-06-23
chore: Rename CheckedAdd operator to Add

robert3005:
Based on conversation in https://github.com/vortex-data/vortex/pull/3603


lwwmanning (approved):

danking:
The new docs still clarify that it's checked, so maybe that's enough. I found it annoying to hunt down all the concrete implementations of `NumericOperator::Add` to convince myself it really was a checked operator. IMO, APIs meant for engineers (rather than scientists / data-scientists / machine learners) should favor clarity before concision, so I would have gone with CheckedAdd, SaturatingAdd, and WrappingAdd.

robert3005:
It's an annoying balance to strike where you want the short name but also make it descriptive enough. I think it would be more obvious when we add other kinds of addition, i.e. saturating and wrapping

PR #3613, opened 2025-06-23
chore[vortex-layout]: remove file index from `RowIdLayoutReader`

joseph-isaacs:


vortex-layout/src/layouts/row_id.rs
```diff
@@ -75,18 +69,12 @@ impl RowIdLayoutReader {
...
-                    "file_index",
-                    ConstantArray::new(self.file_index, arr_len).to_array(),
-                ),
-            ])
+            StructArray::from_fields(&[(
+                "file_row_number",
+                SequenceArray::typed_new(row_range.start, 1, arr_len)
+                    .vortex_expect("cannot be out of bounds")
+                    .to_array(),
+            )])
```
danking, line 77: IMO, the struct is less obvious than having the row index evaluate directly to the number.

vortex-ffi/src/file.rs
```diff
@@ -35,6 +36,23 @@ arc_wrapper!(
...
+            }
+            if let Some(var) = get_item.as_any().downcast_ref::<Var>() {
+                if var.var() == "$vx.row_id" {
+                    return Ok(TransformResult::yes(lit(file_idx)));
+                }
+            }
+        }
+        return Ok(TransformResult::no(e));
+    }))
+    .map(|r| r.into_inner())
```
danking, line 53: I feel like this should just be a new variable in the scope, right?

danking (changes requested):

vortex-ffi/src/file.rs
joseph-isaacs: yes, in the future, however we don't have the APIs for that now, right?

vortex-layout/src/layouts/row_id.rs
joseph-isaacs: ye, I was lazy here. Ill change this once we have the expr reduction logic in place

joseph-isaacs: need to map `2 == 2 and e => true`

PR #3616, opened 2025-06-23
feat: teach ScopeDType to Display

danking:
The debug output is way too verbose to be useful.


vortex-expr/src/scope.rs
```diff
@@ -1,7 +1,9 @@
 use std::any::Any;
+use std::fmt::Display;
 use std::str::FromStr;
 use std::sync::Arc;
 
+use itertools::Itertools as _;
```
robert3005:
```suggestion
use itertools::Itertools;
```

robert3005 (approved):

PR #3623, opened 2025-06-24
chore: Enable runs-on s3 magic cache

robert3005:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,6 @@ jobs:
         with:
           submodules: "recursive"
       - uses: ./.github/actions/cleanup
-      - uses: rui314/setup-mold@v1
       - uses: ./.github/actions/setup-rust
       - name: Install uv
         uses: spiraldb/actions/.github/actions/setup-uv@0.11.0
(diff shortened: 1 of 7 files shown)


AdamGS (approved):

robert3005:
might be sscache is not happy with this cache :/

robert3005:
looks like it's a flake though, will retrigger

.github/workflows/ci.yml
```diff
@@ -130,8 +130,10 @@ jobs:
       - family=m7i+m7i-flex+m7a
       - cpu=8
       - image=ubuntu24-full-x64
+      - extras=s3-cache
       - tag=rust-docs
     steps:
+      - uses: runs-on/action@v2
```
robert3005, line 135: we were missing the action setup

robert3005:
Turns out sccache requires special setting https://github.com/runs-on/action?tab=readme-ov-file#sccache

robert3005:
Ok, I made us use setup-rust steup everywhere since it's annoying to manage all the one off job configs where we have rust everywhere but not sccache

PR #3628, opened 2025-06-24
bug[vortex-array]: take compare nullable bugs

joseph-isaacs:


vortex-array/src/compute/mod.rs
```diff
@@ -113,10 +113,15 @@ impl ComputeFn {
...
-                "Internal error: compute function {} returned a result of type {} but expected {}",
+                "Internal error: compute function {} returned a result of type {} but expected {}\n{}",
                 self.id,
                 output.dtype(),
                 &expected_dtype,
+                args.inputs
+                    .iter()
+                    .filter_map(|input| input.array())
+                    .map(|array| array.tree_display())
+                    .join(",")
```
robert3005: I think you want format_with so we don’t eagerly construct this string

joseph-isaacs: do you mean this?

robert3005 (approved):

PR #3629, opened 2025-06-24
fix: don't leak selection vector

0ax1:
`duckdb_vx_vector_slice_to_dictionary` does not take ownership.


0ax1:
Double checked that the leak goes away with: `leaks -atExit  -fullContent -- ./clickbench -i 1 --targets duckdb:vortex`.

joseph-isaacs (approved):
I think we might be able to catch this next time with a *mut vs (*const)?

PR #3630, opened 2025-06-24
fix: `duckdb_logical_type` leak

0ax1:
`duckdb_create_vector` does not take ownership of the heap allocated memory but copies the logical type by value.


0ax1:
Verified via: `leaks -atExit  -fullContent -- ./clickbench -i 1 --targets duckdb:vortex`

lwwmanning (approved):

PR #3631, opened 2025-06-24
fix: duckdb `varchar` leak

0ax1:
The caller is responsible for free-ing the memory with `duckdb_free`.

### DIFF
--- a/vortex-duckdb/src/duckdb/value.rs
+++ b/vortex-duckdb/src/duckdb/value.rs
@@ -71,8 +71,14 @@ impl Value {
         unsafe { Self::own(cpp::duckdb_create_date(cpp::duckdb_date { days })) }
     }
 
-    pub fn as_string(&self) -> &CStr {
-        unsafe { CStr::from_ptr(cpp::duckdb_get_varchar(self.as_ptr())) }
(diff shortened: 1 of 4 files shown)


0ax1:
Verified by running: `leaks -atExit  -fullContent -- ./clickbench -i 1 --targets duckdb:vortex`.

vortex-duckdb/src/duckdb/value.rs
```diff
@@ -71,8 +71,14 @@ impl Value {
         unsafe { Self::own(cpp::duckdb_create_date(cpp::duckdb_date { days })) }
     }
 
-    pub fn as_string(&self) -> &CStr {
-        unsafe { CStr::from_ptr(cpp::duckdb_get_varchar(self.as_ptr())) }
+    pub fn as_string(&self) -> String {
```
gatesn, line 74: IMO you should return a CString and allow the caller to decide how to handle utf-8 validation errors.

gatesn: Eh, nah, this is more practical!

gatesn (approved):

PR #3636, opened 2025-06-24
chore: move ci to duckdb v1.3.1

0ax1:
Corresponding Vortex extension build: https://github.com/duckdb/community-extensions/commit/b654106bd842cf6ebaed1bb8802a4d293381959a


robert3005 (approved):

robert3005:
I need to get rid of that gzip install. We can replace it with regular tar commands

PR #3637, opened 2025-06-24
list is_constant

### LINKED ISSUE #3622: bug: is constant in array vortex.struct({value=list(i64)?}?, len=8192) canonicalize returned None
when I convert a parquet file to vortex file, an exception is thrown when executing `vortex.io.write`
```
thread '<unnamed>' panicked at vortex-array/src/compute/is_constant.rs:78:13:
is constant in array vortex.struct({value=list(i64)?}?, len=8192) canonicalize returned None
```

code
```rust
let value = is_constant_impl(array, options, kernels)?;

// TODO(joe): add is_constant for ListArray
if options.cost == Cost::Canonicalize && !array.is::<ListVTable>() {
    // When we run linear canonicalize, there we must always return an exact answer.
    assert!(
        value.is_some(),
        "is constant in array {array} canonicalize returned None"
    );
}
```

so we should implement `is_constant` interface for `ListArray`？

onursatici:
fixes https://github.com/vortex-data/vortex/issues/3622

### DIFF
--- a/vortex-array/src/arrays/list/compute/is_constant.rs
+++ b/vortex-array/src/arrays/list/compute/is_constant.rs
@@ -5,22 +5,39 @@ use crate::arrays::{ListArray, ListVTable};
 use crate::compute::{IsConstantKernel, IsConstantKernelAdapter, IsConstantOpts, numeric};
 use crate::register_kernel;
 
+const SMALL_ARRAY_THRESHOLD: usize = 64;
+
 impl IsConstantKernel for ListVTable {
     fn is_constant(&self, array: &ListArray, opts: &IsConstantOpts) -> VortexResult<Option<bool>> {
         // At this point, we're guaranteed:
         // - Array has at least 2 elements
         // - All elements are valid (no nulls)
 
+        let manual_check_until = std::cmp::min(SMALL_ARRAY_THRESHOLD, array.len());
+
+        let first_list_len = array.offset_at(1) - array.offset_at(0);
+        for i in 1..manual_check_until {
+            let current_list_len = array.offset_at(i + 1) - array.offset_at(i);
+            if current_list_len != first_list_len {
+                return Ok(Some(false));
+            }
+        }
+
         if opts.is_negligible_cost() {
             return Ok(None);
         }
 
-        let start_offsets = array.offsets.slice(0, array.len())?;
-        let end_offsets = array.offsets.slice(1, array.len() + 1)?;
-        let list_lengths = numeric(&end_offsets, &start_offsets, NumericOperator::Sub)?;
+        if array.len() > SMALL_ARRAY_THRESHOLD {
+            // check the rest of the element lengths
+            let start_offsets = array.offsets.slice(SMALL_ARRAY_THRESHOLD, array.len())?;
+            let end_offsets = array
+                .offsets
+                .slice(SMALL_ARRAY_THRESHOLD + 1, array.len() + 1)?;
+            let list_lengths = numeric(&end_offsets, &start_offsets, NumericOperator::Sub)?;
 
-        if !list_lengths.is_constant() {
-            return Ok(Some(false));
+            if !list_lengths.is_constant() {
+                return Ok(Some(false));
+            }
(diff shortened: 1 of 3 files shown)


vortex-array/src/arrays/list/compute/is_constant.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn is_constant(&self, array: &ListArray, opts: &IsConstantOpts) -> VortexResult<Option<bool>> {
+        // At this point, we're guaranteed:
+        // - Array has at least 2 elements
+        // - All elements are valid (no nulls)
+
+        if opts.is_negligible_cost() {
+            return Ok(None);
+        }
+
+        let first_list_len = array.offset_at(1) - array.offset_at(0);
```
gatesn, line 18: Can you do something like `array.offsets().is_constant()`?

vortex-array/src/arrays/list/compute/is_constant.rs
```diff
@@ -0,0 +1,134 @@
...
+        for i in 1..array.len() {
+            let current_list_len = array.offset_at(i + 1) - array.offset_at(i);
+            if current_list_len != first_list_len {
+                return Ok(Some(false));
+            }
+        }
+
+        // If all lists have the same length, compare the actual list contents
+        let first_scalar = array.scalar_at(0)?;
```
gatesn, line 44: Comparing list scalars is quite expensive... but this is where we would want some sort of `reduce_by(ALL_EQ)` compute function / operator...

vortex-array/src/arrays/list/compute/is_constant.rs
robert3005: annoyingly offsets will not be constant for a constant list because of leading 0. You can slice off the first element

vortex-array/src/arrays/list/compute/is_constant.rs
onursatici: I guess once we have reduce_by it can be the fallback impl for is_constant

vortex-array/src/arrays/list/compute/is_constant.rs
onursatici: we need `offsets.diff().is_constant()` here right? We need some sort of `scan` reduction compute function to do this I think

onursatici: actually I guess I can do something like
```
offsets_end = offsets.slice(1..)
binary_numeric(offset_end, offset, SUB).is_constant()
```
maybe that is better?

robert3005: right, this will be incrementing by constant amount, so the delta has to be constant. we really want list view

onursatici: I changed it to use compute functions now. We have lost the early exit but still might be better for some encodings

vortex-array/src/arrays/list/compute/is_constant.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn is_constant(&self, array: &ListArray, opts: &IsConstantOpts) -> VortexResult<Option<bool>> {
+        // At this point, we're guaranteed:
+        // - Array has at least 2 elements
+        // - All elements are valid (no nulls)
+
+        if opts.is_negligible_cost() {
+            return Ok(None);
+        }
+
+        let start_offsets = array.offsets.slice(0, array.len())?;
```
joseph-isaacs: I wonder if it is worth checking the first few list lens or values of index zero as a fast exit.

vortex-array/src/arrays/list/compute/is_constant.rs
onursatici: now we fail fast for the first 64 lists, and check the rest with compute functions

robert3005 (approved):
We should start working on reduction expressions, also filed https://github.com/vortex-data/vortex/issues/3641 for better scalar_at in this case

PR #3638, opened 2025-06-24
Fix: delay getting benchmarks s3 token until after binary has been built

robert3005:


0ax1 (approved):
good to go if ci agrees on the bench build tasks

robert3005:
I have some ideas on what to do if this doesn't work but ideally I don't have to resort to that

lwwmanning (approved):

Issue #3633, opened 2025-06-24
Add DType structs + trait impls


gatesn:
We currently don't have a trait representing a DType. This makes it hard to write generic code over Vortex DTypes, instead forcing us to always use combinations of num-traits or native types. I'm hoping this makes it easier to write code against Rust-based extension types too, e.g. temporal types.

See Arrow's data type structs for an example: https://docs.rs/arrow/latest/arrow/datatypes/index.html#structs

connortsui20:
Is this something we still want? I think it is actually more ideal that we keep the different types as separate as possible

robert3005:
I think this is outdated given everything is lazy now, this was a useful idea if things are eager to so you can generify your expressions but with lazy expressions this is all resolved at runtime anyway

PR #3639, opened 2025-06-24
experiment: Encode dtype structure into primitive arrays

AdamGS:
Been thinking about it today after spending some time with arrow and seeing that a bunch of its timestamp-related types are actually primitive arrays which I thought was nice + thinking about type nullability and how often we seem to mess it up in one way or another and then #3633 gave me a push

### DIFF
--- a/vortex-array/src/arrays/primitive2/mod.rs
+++ b/vortex-array/src/arrays/primitive2/mod.rs
@@ -1,0 +1,99 @@
+use arrow_buffer::BooleanBufferBuilder;
+use vortex_buffer::{Buffer, BufferMut};
+use vortex_dtype::{DType, Nullability, PType};
+
+use crate::stats::ArrayStats;
+use crate::validity::Validity;
+
+pub trait PrimitiveType<const NULLABLE: bool>: Sized {
+    const DTYPE: DType;
+
+    type Storage;
+
+    fn zero() -> Self::Storage;
+
+    fn width() -> usize {
+        size_of::<Self::Storage>()
+    }
+
+    fn nullability() -> Nullability {
+        if NULLABLE {
+            Nullability::Nullable
+        } else {
+            Nullability::NonNullable
+        }
+    }
+}
+
+// Just a very easy type to use here, I expect some complications down the road but
+// the built-in integers should be able to work with a macro.
+struct U8DType<const NULLABLE: bool>;
+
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/primitive2/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+#![allow(unused_variables)]
+
+use arrow_buffer::BooleanBufferBuilder;
+use vortex_buffer::{Buffer, BufferMut};
+use vortex_dtype::{DType, Nullability, PType};
+
+use crate::stats::ArrayStats;
+use crate::validity::Validity;
+
+pub trait PrimitiveType<const NULLABLE: bool>: Sized {
```
robert3005, line 11: Could this be the Nullability enum? the bool reads very weird, should be functionaly the same

AdamGS: cool idea, I think it should work

AdamGS: trying it now

AdamGS: turns out - https://github.com/rust-lang/rust/issues/95174

AdamGS: If we change it to a pseudo enum with associated consts it can work. It doesn't have to replace the enum I guess? It can be an additional type that can be converted into `Nullability`

robert3005: This is such a classic rust outcome… I sort of expected it when I asked

robert3005: Oh this has been open only for 3 years

gatesn:
I do like the idea of this, but a couple of points:
* In many places, we have PrimitiveArray where we know it holds e.g. "indices". This must be an unsigned integer type. How can I hold onto a `PrimitiveArray<IndexType>` without propagating generics all the way up (e.g. I don't think we want `DictArray<Index, Value>` do we? Maybe we do?
* I probably wouldn't bother encoding nullability into the generics

AdamGS:
I need to wrap up some stuff before I come back to this.
I think dicts and fastlanes (or maybe ALP) are the next targets to see how this thing looks. I'm pretty confident that byte/views are easy if those are solved.

PR #3642, opened 2025-06-25
chore: Do we need gzip install?

robert3005:


robert3005:
I checked the AMI we use and it has gzip already

0ax1 (approved):

PR #3645, opened 2025-06-25
fix[vortex-array]: struct validity + runend validity

joseph-isaacs:


vortex-array/src/arrays/struct_/compute/mod.rs
```diff
@@ -3,24 +3,47 @@ mod filter;
...
                 .iter()
-                .map(|field| take(field, indices))
+                .map(|field| {
+                    take(
+                        field,
+                        // The validity is applied to the struct validity,
+                        &fill_null(
+                            indices,
+                            &Scalar::default_value(indices.dtype().with_nullability(NonNullable)),
+                        )?,
```
robert3005: why do you need to fill null here?

joseph-isaacs: otherwise you will apply nullability to the fields, changing the dtype .

joseph-isaacs: pull out the compute fn to do once

robert3005: but take can change nullability?

joseph-isaacs: There is something interesting about top level nullable struct dtypes (`{a: I32 - nonnullable} nullable`) where the inner field non-nullable, but in fact the fields are undefined by the struct validity

robert3005: it's an interesting point whether take on struct should change nullability of the fields...

joseph-isaacs: I think take/filter should only affect the top level nullability

robert3005: I guess they should not? I could be convinced either way

joseph-isaacs: It would impl that it would cascade all the way down all containers, 

{x: {a: I32 NonNullable} NonNullable} -> {x: {a: I32 Nullable} Nullable}

robert3005 (approved):

PR #3646, opened 2025-06-25
fix: bound column name ref leak

0ax1:
`duckdb_vx_expr_get_bound_column_ref_get_name` returns a heap allocated string that needs to freed with `duckdb_free`.


0ax1:
verified via: ` leaks -atExit -- ./clickbench -i 1 --targets duckdb:vortex`

vortex-duckdb/src/duckdb/expr.rs
```diff
@@ -30,9 +30,13 @@ impl Expression {
...
                         unsafe { cpp::duckdb_vx_expr_get_bound_column_ref_get_name(self.as_ptr()) };
-                    ExpressionClass::BoundColumnRef(BoundColumnRef {
-                        name: unsafe { CStr::from_ptr(ptr) },
-                    })
+
+                    let name = unsafe { CStr::from_ptr(ptr) }
+                        .to_string_lossy()
+                        .into_owned();
+
+                    unsafe { cpp::duckdb_free(ptr.cast_mut().cast()) };
```
0ax1: I'll add a Rust wrapper around &CStr such that we don't need to "clone".

joseph-isaacs: not sure it matters so much

joseph-isaacs (approved):

PR #3647, opened 2025-06-25
chore: Always use primitive array simd take

robert3005:
Fixed the cases where we determined we can't use a simd take. Also noticed that
our simd take was unchecked which is not correct as we didn't validate the
indices

### DIFF
--- a/vortex-array/src/arrays/primitive/compute/take.rs
+++ b/vortex-array/src/arrays/primitive/compute/take.rs
@@ -15,6 +15,10 @@ use crate::compute::{TakeKernel, TakeKernelAdapter, cast};
 use crate::vtable::ValidityHelper;
 use crate::{Array, ArrayRef, IntoArray, ToCanonical, register_kernel};
 
+// SIMD types larger than the SIMD register size are beneficial for
+// performance as this leads to better instruction level parallelism.
+const SIMD_WIDTH: usize = 64;
+
 impl TakeKernel for PrimitiveVTable {
     #[allow(clippy::cognitive_complexity)]
     fn take(&self, array: &PrimitiveArray, indices: &dyn Array) -> VortexResult<ArrayRef> {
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/primitive/compute/take.rs
```diff
@@ -1,67 +1,66 @@
...
-        {
-            // TODO(alex): handle nullable codes & values
-            match_each_unsigned_integer_ptype!(indices.ptype(), |C| {
+        let unsigned_indices = match indices.dtype() {
+            DType::Primitive(p, n) => {
+                if p.is_unsigned_int() {
+                    indices.to_primitive()?
+                } else {
+                    // This will fail if all values cannot be converted to unsigned
+                    cast(indices, &DType::Primitive(p.to_unsigned(), *n))?.to_primitive()?
```
joseph-isaacs, line 31: why do we all signed integers in take?

vortex-array/src/arrays/primitive/compute/take.rs
```diff
@@ -1,67 +1,66 @@
...
+
+        let validity = array.validity().take(unsigned_indices.as_ref())?;
+        if array.ptype() != PType::F16 {
+            match_each_unsigned_integer_ptype!(unsigned_indices.ptype(), |C| {
                 match_each_native_simd_ptype!(array.ptype(), |V| {
                     // SIMD types larger than the SIMD register size are beneficial for
                     // performance as this leads to better instruction level parallelism.
                     let decoded = take_primitive_simd::<C, V, 64>(
-                        indices.as_slice(),
+                        unsigned_indices.as_slice(),
```
joseph-isaacs: `unsigned_indices` might have undefined values (due to validity) that are out of bounds of the values (array)

robert3005: You have this problem already in the existing code. We assume all over that invalid values in indices are still valid take indices, however, I changed the simd implementation to do the checked gather so it will mask out out of bounds values

vortex-array/src/arrays/primitive/compute/take.rs
robert3005: the original reason is that a lot of systems coming from c/c++ use signed integers for performance reasons. We could outlaw it for which I had a pr at some point. However, the more I think about it the more I think we should add an Indices helper like Mask for take uses and have that do signed to unsigned.

a10y (approved):

vortex-array/src/arrays/primitive/compute/take.rs
joseph-isaacs: agreed, just making sure its still sound

a10y:
CodSpeed reported some slowdowns with this, which are probably unavoidable

image

robert3005:
Yup, this is the cost of checking indices

vortex-array/src/arrays/primitive/compute/take.rs
```diff
@@ -164,4 +163,13 @@ mod test {
         // the third index is null
         assert_eq!(actual.scalar_at(2).unwrap(), Scalar::null_typed::<i32>());
     }
+
+    #[test]
+    fn test_take_out_of_bounds() {
+        let indices = vec![2_000_000u32; 64];
+        let values = vec![1i32];
+
+        let result = take_primitive_simd::<u32, i32, 64>(&indices, &values);
```
a10y, line 172: on reflection, i think this is the wrong behavior. or at least, it's inconsistent with other take impls which panic/err on out of bounds index access

PR #3652, opened 2025-06-25
dict layout writer to allow empty input stream

### LINKED ISSUE #3600: DictLayout doesn't handle empty input streams
I think this happens if you try to write an all null array (courtesy of the fuzzer). We fail in https://github.com/vortex-data/vortex/blob/8c1700afa25dd23daaae8a695576b15d02d3f782/vortex-layout/src/layouts/dict/writer/mod.rs#L422-L424 with the stream being empty

onursatici:
Fix https://github.com/vortex-data/vortex/issues/3600

### DIFF
--- a/vortex-layout/src/layouts/dict/writer/mod.rs
+++ b/vortex-layout/src/layouts/dict/writer/mod.rs
@@ -100,7 +100,7 @@ impl LayoutStrategy for DictStrategy {
                 None => true, // empty stream
                 Some(chunk) => {
                     let compressed = BtrBlocksCompressor.compress(&chunk)?;
-                    compressed.is_encoding(DictEncoding.id())
+                    !compressed.is_encoding(DictEncoding.id())
                 }
             };
             if should_fallback {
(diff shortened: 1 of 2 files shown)


vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -99,8 +99,8 @@ impl LayoutStrategy for DictStrategy {
             })
             .await?;
             let stream = SequentialStreamAdapter::new(dtype.clone(), stream).sendable();
-            if !is_dict_encoding? {
-                // first chunk did not compress to dict, skip dict layout
+            if !is_dict_encoding.unwrap_or(false) {
```
robert3005: I think we are swallowing errors here

onursatici: yea you are right

onursatici: looks better now I think, should have done it like this the first time. Only if Stream::peekable was easier to use here

robert3005 (approved):

robert3005:
I think you have some tests failures because layout trees changed

PR #3653, opened 2025-06-25
fix: add bounds checking in take_primitive_simd

a10y:
Fix soundness issue with `take_primitive_simd`.

We have bounds-checking implicitly for the old scalar implementation since it goes through `std::slice` access.

Before, if we had invalid take indices we'd do invalid memory accesses 😱


a10y:
The regressions are unfortunately necessary

a10y:
Looks like on a cold run the stats check is worse than iterating the indices directly, hard to tell how often the stats will be cached

robert3005:
You can simd validate the indices which is what my pr does

0ax1:
In C++ voice: _bounds checking?_

a10y:
Closing in favor of #3647

PR #3654, opened 2025-06-26
bug[encodings/sparse]: sparse array doesn't respect indicies validity on fill vector

joseph-isaacs:
Indices validity only applied to patches, this requires an expensive intersection of patches-indices with validity set bits


robert3005:
fixed in #3677

PR #3657, opened 2025-06-26
fix[vortex-layout]: correctly write struct (w/o field) length to layout

joseph-isaacs:


vortex-layout/src/layouts/struct_/writer.rs
```diff
@@ -270,4 +283,43 @@ mod tests {
...
+                        {
+                            Ok((
+                                SequenceId::root().downgrade(),
+                                StructArray::try_new([].into(), vec![], 5, Validity::NonNullable)
+                                    .unwrap()
+                                    .into_array(),
+                            ))
+                        },
+                    ]),
+                )
```
onursatici: here each element of the stream should have an increasing `SequenceId`
I would do something like:
```rust
let mut pointer = SequenceId::root();
...
  stream::iter([
    Ok((pointer.advance(), .. )),
    Ok((pointer.downgrade(), ..)),
  ])
```

onursatici: In this test it doesn't matter, but if both elements in the stream has the same sequence_id there is no ordering guarantee for processing them in the writers

joseph-isaacs: why that order?

joseph-isaacs: or are you saying, make sure there is a strict order, would 

```
    Ok((pointer.downgrade(), .. )),
    Ok((pointer.advance(), ..)),
```

work?

onursatici (approved):

vortex-layout/src/layouts/struct_/writer.rs
```diff
@@ -57,6 +57,19 @@ impl LayoutStrategy for StructStrategy {
             Ok((sequence_id, chunk))
         });
 
+        // There are now fields so this is the layout leaf
```
robert3005, line 60: I”no” instead of “now”?

PR #3661, opened 2025-06-26
Run clickbench file download on a local rayon pool

robert3005:


robert3005:
@0ax1 this should remove usage of rayon global thread pool and leaking of it

a10y (approved):

PR #3662, opened 2025-06-27
 feat: teach SparseArray to canonicalize Null & Decimal

danking:
I wasn't sure what to do when the patch values and the fill value differ in type. I decided to cast
the fill value to the patch value type since there's only one fill value.

@a10y I'll keep working through the other arrays too.

### DIFF
--- a/vortex-array/src/arrays/decimal/mod.rs
+++ b/vortex-array/src/arrays/decimal/mod.rs
@@ -9,7 +9,7 @@ mod serde;
 use arrow_buffer::BooleanBufferBuilder;
 use vortex_buffer::{Buffer, BufferMut, ByteBuffer};
 use vortex_dtype::{DType, DecimalDType};
-use vortex_error::{VortexResult, vortex_bail, vortex_panic};
+use vortex_error::{VortexExpect as _, VortexResult, vortex_bail, vortex_panic};
 use vortex_scalar::{DecimalValueType, NativeDecimalType};
 
 use crate::builders::ArrayBuilder;
@@ -63,14 +63,8 @@ pub fn smallest_storage_type(decimal_dtype: &DecimalDType) -> DecimalValueType {
 }
 
 /// Errors unless the value type can represent every value of the given dtype.
-pub fn compatible_storage_type(
-    value_type: DecimalValueType,
-    dtype: &DecimalDType,
-) -> VortexResult<()> {
-    if value_type < smallest_storage_type(dtype) {
-        vortex_bail!("{:?} cannot represent every value in {}", value_type, dtype)
-    }
-    Ok(())
+pub fn compatible_storage_type(value_type: DecimalValueType, dtype: DecimalDType) -> bool {
+    value_type >= smallest_storage_type(&dtype)
 }
 
 /// Array for decimal-typed real numbers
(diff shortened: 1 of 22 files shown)


robert3005:
The dtypes should be the same for fill and values

robert3005:
I guess you mean storage types

robert3005:
I would cast to the one that’s more numerous

danking:
Given the decimal dtype, isn't there a unique smallest type which fits all values of that decimal dtype? Could I just pick that one and cast things from unnecessarily larger types into that smaller one?

robert3005:
There is one, there should be a function to get it. I only wonder if its an unnecessary copy if you already have values of certain width

danking:
Okay now:
1. Compute smallest storage type for the given DecimalDType.
2. Create a buffer of that type.
3. Fill the buffer with the fill value (casted to that type).
4. Patch the buffer with the patch values (casted to that type, panic'ing if the value is too large to be represented).

I also added a check to DecimalArray::new that the type of the buffer is big enough to represent every value in the DecimalDType.

vortex-array/src/arrays/decimal/mod.rs
```diff
@@ -72,30 +78,47 @@ pub struct DecimalArray {
...
+
+    /// Creates a new [`DecimalArray`] from a [`Buffer`] and [`Validity`], without checking
+    /// any invariants.
+    pub fn new<T: NativeDecimalType>(
+        buffer: Buffer<T>,
+        decimal_dtype: DecimalDType,
+        validity: Validity,
+    ) -> Self {
+        Self::try_new(buffer, decimal_dtype, validity).vortex_expect("DecimalArray::new")
     }
```
robert3005: I would remove this and if we want unchecked version call it `new_unchecked`

vortex-array/src/arrays/decimal/patch.rs
```diff
@@ -0,0 +1,71 @@
...
+    patch_indices_offset: usize,
+    patch_values: Buffer<PatchDVT>,
+    patched_validity: Validity,
+) -> VortexResult<DecimalArray>
+where
+    I: NativePType + ArrowNativeType,
+    PatchDVT: NativeDecimalType,
+    ValuesDVT: NativeDecimalType,
+{
+    for (idx, value) in itertools::zip_eq(patch_indices, patch_values) {
```
robert3005: personal bias for 
```suggestion
    for (idx, value) in patch_indices.iter().zip_eq(patch_values) {
```

vortex-array/src/arrays/decimal/mod.rs
```diff
@@ -72,30 +78,47 @@ pub struct DecimalArray {
...
         }
 
-        Self {
+        if !compatible_storage_type(T::VALUES_TYPE, decimal_dtype) {
+            vortex_bail!(
+                "{:?} cannot represent every value in {}",
+                T::VALUES_TYPE,
+                decimal_dtype
+            )
+        }
```
robert3005: I guess we can assert that in the patch. T could be smaller than the smallests possible width for dtype

vortex-array/src/arrays/decimal/patch.rs
danking: Done.

vortex-array/src/arrays/decimal/mod.rs
danking: Done.

vortex-array/src/arrays/decimal/mod.rs
danking: I've moved it into patch.

vortex-array/src/arrays/decimal/mod.rs
```diff
@@ -72,30 +78,43 @@ pub struct DecimalArray {
...
 impl DecimalArray {
-    /// Creates a new [`DecimalArray`] from a [`Buffer`] and [`Validity`], without checking
-    /// any invariants.
-    pub fn new<T: NativeDecimalType>(
+    /// Creates a new [`DecimalArray`] from a [`Buffer`] and [`Validity`].
+    ///
+    /// # Errors
+    ///
+    /// Errors if the validity length is not compatible with the buffer length.
+    pub fn try_new<T: NativeDecimalType>(
```
robert3005: If we are not adding the dtype width check then I would keep it as is. The error here is not useful imho and clearly a bug for a user

danking: done.

robert3005 (approved):

PR #3663, opened 2025-06-27
fix[vortex-array]: check NaN for inferring is constant with min/max

### LINKED ISSUE #3660: The internal mutability of stats set leak between array/is very confusing
Consider 

```
    let array_data = ChunkedArray::from_iter([
        PrimitiveArray::from_iter([f16::ONE, f16::NAN]).to_array(),
        PrimitiveArray::from_iter([f16::ZERO]).to_array(),
    ])
    .to_array();
    let expected_array = array_data.clone();

    println!(
        "before write {:?}, children {}",
        expected_array.statistics().get(IsConstant),
        expected_array
            .as_::<ChunkedVTable>()
            .chunks()
            .iter()
            .map(|c| format!("{:?}", c.statistics().get(IsConstant)))
            .join(",")
    );

    let _full_buff = VortexWriteOptions::default()
        .write(ByteBufferMut::empty(), array_data.to_array_stream())
        .await
        .vortex_unwrap();

    println!(
        "after write {:?}, children {}",
        expected_array.statistics().get(IsConstant),
        expected_array
            .as_::<ChunkedVTable>()
            .chunks()
            .iter()
            .map(|c| format!("{:?}", c.statistics().get(IsConstant)))
            .join(",")
    );
```

The output is 

```
before write None, children None,None
after write None, children Some(Exact(ScalarValue(Bool(true)))),Some(Exact(ScalarValue(Bool(true))))
```

joseph-isaacs:
I think that in the future min == max doesn't imply is constant for all types e.g. truncated str. I think we should reconsider this.

Closes https://github.com/vortex-data/vortex/issues/3660

### DIFF
--- a/vortex-array/src/compute/is_constant.rs
+++ b/vortex-array/src/compute/is_constant.rs
@@ -139,17 +139,21 @@ fn is_constant_impl(
     let min = array.statistics().get_scalar(Stat::Min, array.dtype());
     let max = array.statistics().get_scalar(Stat::Max, array.dtype());
 
+    println!(
+        "min {:?}, max {:?}",
+        array.statistics().get_as::<f32>(Stat::Min),
+        array.statistics().get_as::<f32>(Stat::Max),
+    );
+
     if let Some((min, max)) = min.zip(max) {
         // min/max are equal and exact and there are no NaNs
         if min.is_exact()
             && min == max
-            && array
+            && if array
                 .dtype()
-                .is_float()
-                .then(|| {
+                .is_float() { {
                     array.statistics().get_as::<u64>(Stat::NaNCount) == Some(Precision::exact(0u64))
-                })
-                .unwrap_or(true)
+                } } else { true }
         {
             return Ok(Some(true));
         }
@@ -291,13 +295,20 @@ impl IsConstantOpts {
 #[cfg(test)]
 mod tests {
     use crate::arrays::PrimitiveArray;
+    use crate::stats::Stat;
 
     #[test]
     fn is_constant_min_max_no_nan() {
         let arr = PrimitiveArray::from_iter([0, 1]);
+        arr.statistics()
+            .compute_all(&[Stat::Min, Stat::Max])
+            .unwrap();
         assert!(!arr.is_constant());
 
         let arr = PrimitiveArray::from_iter([0, 0]);
+        arr.statistics()
+            .compute_all(&[Stat::Min, Stat::Max])
+            .unwrap();
         assert!(arr.is_constant());
 
         let arr = PrimitiveArray::from_option_iter([Some(0), Some(0)]);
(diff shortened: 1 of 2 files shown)


vortex-array/src/compute/is_constant.rs
```diff
@@ -136,17 +136,21 @@ fn is_constant_impl(
...
-        if min == max {
+        // min/max are equal and exact and there are no NaNs
+        if min.is_exact()
+            && min == max
+            && array
+                .dtype()
+                .is_float()
+                .then(|| {
+                    array.statistics().get_as::<u64>(Stat::NaNCount) == Some(Precision::exact(0u64))
+                })
```
AdamGS: we already know the nan-count is 0 here

joseph-isaacs: Well this fixes a bug

joseph-isaacs: Where is that check?

AdamGS: lines 126-136

joseph-isaacs: Nans are valid

AdamGS: oh those are float-nans, got it

vortex-array/src/compute/is_constant.rs
```diff
@@ -136,17 +136,19 @@ fn is_constant_impl(
...
+    let min = array.statistics().get_scalar(Stat::Min, array.dtype());
+    let max = array.statistics().get_scalar(Stat::Max, array.dtype());
 
     if let Some((min, max)) = min.zip(max) {
-        if min == max {
+        // min/max are equal and exact and there are no NaNs
+        if min.is_exact()
+            && min == max
+            && if array.dtype().is_float() {
+                array.statistics().get_as::<u64>(Stat::NaNCount) == Some(Precision::exact(0u64))
```
robert3005: nan_count(array)? == 0? You still have to compute the nan_count otherwise you might get a false positive still?

joseph-isaacs: in what case if there is no stat this equality will not hold, so is constant will not be set

gatesn: I think the question is: if you're this deep into the short-circuit call (min max are present, and equal), then you may as well perform actual nan_count compute instead of also requiring that stat to be pre-computed.

robert3005: I think the overall thing is that it's faster to fallback to is_constant than doing nan_count here. Is there a nicer way of expressing this than an if withing an if condition?

joseph-isaacs: Is there something you want to change here? 

By all means push a small change it you want so we can merge

robert3005 (approved):

PR #3664, opened 2025-06-27
fix[compressor]: Enable float run-end compressor scheme

AdamGS:
Seems like an omission? Couldn't find any reference on slack to it


robert3005:
I think our float runend doesn't follow the paper. We should also try to compress the values. @a10y was this an accidental omission?

a10y:
It was probably a mistake. It's possible that the original intention was to force the floats through an int transformation first, but as long as the compression time is still fast I think it's ok

a10y (approved):

AdamGS:
This seems worse for most benchmarks

PR #3665, opened 2025-06-27
feat: multiversion for primitive take

a10y:
Doesn't affect the CodSpeed runs since we already run our benchmarks with `target-feature=avx2` but probably helps library clients out

### DIFF
--- a/vortex-array/src/arrays/primitive/compute/take.rs
+++ b/vortex-array/src/arrays/primitive/compute/take.rs
@@ -77,7 +77,7 @@ register_kernel!(TakeKernelAdapter(PrimitiveVTable).lift());
 /// # Returns
 /// A `PrimitiveArray` containing the gathered values where each index has been replaced with
 /// the corresponding value from the source array.
-#[multiversion(targets("x86_64+avx+avx2"))]
+#[multiversion(targets("x86_64+avx2", "x86_64+avx"))]
 fn take_primitive_simd<I, V, const LANE_COUNT: usize>(indices: &[I], values: &[V]) -> Buffer<V>
 where
     I: simd::SimdElement + AsPrimitive<usize>,
--- a/vortex-array/src/arrays/primitive/compute/take.rs
+++ b/vortex-array/src/arrays/primitive/compute/take.rs
@@ -1,6 +1,7 @@
 use std::mem::{MaybeUninit, transmute};
 use std::simd;
 
+use multiversion::multiversion;
 use num_traits::AsPrimitive;
 use simd::num::SimdUint;
 use vortex_buffer::{Alignment, Buffer, BufferMut};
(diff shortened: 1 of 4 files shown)


vortex-array/src/arrays/primitive/compute/take.rs
```diff
@@ -72,11 +73,11 @@ register_kernel!(TakeKernelAdapter(PrimitiveVTable).lift());
 /// # Parameters
 /// * `indices` - Indices to gather values from
 /// * `values` - Source values to index
-/// * `nullability` - Nullability of the resulting array
 ///
 /// # Returns
 /// A `PrimitiveArray` containing the gathered values where each index has been replaced with
 /// the corresponding value from the source array.
+#[multiversion(targets("x86_64+avx+avx2"))]
```
0ax1: This is one version besides the fallback right? Would it make sense to add
`x86_64+avx2` and `x86_64+avx` as 2 targets? avx2 implies avx I think.

a10y: Yea this is one version. I wasn't sure how explicit I needed to be, e.g. Daft https://github.com/Eventual-Inc/Daft/blob/5886fd734e56932abc3634ec1222f97d0f328338/src/arrow2/src/compute/hash.rs#L24

I can split them into two separate targets, one avx and one avx2

0ax1 (approved):
Curious whether there's any perf diff for macOS when compiling without arch native and adding `aarch64+neon` to the target list.

lgtm

a10y:
Yea that's probably also a good idea, I did a small test and it looks like for Neon even without the gather instruction being available, just doing vectorized loading seems to make a noticeable improvement. https://github.com/a10y/neontest

Let me update to add that branch and then will merge

vortex-array/src/arrays/primitive/compute/take.rs
```diff
@@ -72,11 +73,11 @@ register_kernel!(TakeKernelAdapter(PrimitiveVTable).lift());
 /// # Parameters
 /// * `indices` - Indices to gather values from
 /// * `values` - Source values to index
-/// * `nullability` - Nullability of the resulting array
 ///
 /// # Returns
 /// A `PrimitiveArray` containing the gathered values where each index has been replaced with
 /// the corresponding value from the source array.
+#[multiversion(targets("x86_64+avx2", "x86_64+avx", "aarch64+neon"))]
```
robert3005, line 80: you could kill the `x86_64+avx` variant. These are cpus produced before 2013? All the modern cpus that don't support avx2 also don't support avx

0ax1: Yeah, I guess that's worth considering. It's also bit annoying that we can't add avx512 rn, as this crashes codspeed unless there's a runtime flag with which we can force multiversion to pick avx2 even if 512 is available.

PR #3669, opened 2025-06-27
chore: Only pull shallow copy of submodule

AdamGS:


AdamGS:
doesn't seem to work :/

gatesn:
I think you might actually only need the header files anyway?

They might be distributed as a zip somewhere

0ax1:
> I think you might actually only need the header files anyway?

We need the clone for our DuckDB C++ includes we use in `vortex-duckdb/cpp/*`. 
E.g. https://github.com/vortex-data/vortex/blob/develop/vortex-duckdb/cpp/table_filter.cpp

0ax1:
~~In theory, we could get away with only vendoring the header files in some way but that would also imply that we lose editor LSP support for the headers as the compilation database is based on cpp files.~~

Obs the DuckDB source files don't show up in the compilation db anymore as we use the precompiled lib.

gatesn:
So this largely works, the only problem is the hpp file in the dylib distribution doesn't include all the bound expressions. Weirdly it _does_ include ColumnRefBoundExpression, which makes me think they might accept a PR to expose the others too?

https://github.com/vortex-data/vortex/compare/ngates/duckdb-header-only?expand=1

0ax1:
> So this largely works, the only problem is the hpp file in the dylib distribution doesn't include all the bound expressions. Weirdly it _does_ include ColumnRefBoundExpression, which makes me think they might accept a PR to expose the others too?
> 
> https://github.com/vortex-data/vortex/compare/ngates/duckdb-header-only?expand=1

Given that `duckdb.hpp` is 50k lines long there's a good chance it'll tank compile times, in particular if included in multiple cpp, even if it mostly contains declarations. That said I'm in general of course all in favor of getting rid of submodules.

AdamGS:
what about pulling it in the relevant `build.rs`?

0ax1:
yeah we can think about some build time vendoring here

Issue #3671, opened 2025-06-27
Remove VarExpr in favor of ScopeVars


gatesn:
Our foray into Let/Var became quite complex. I think there's a simpler solution that provides the functionality we need without any of the complexity of dealing with expression rewriting. 

To recap the requirements:
* We want a way for users to reference RowIdx in the filter or projection expressions of a scan.
* We eventually want a way to implement a zipped scan. That is, to provied an input ArrayStream to the scan with cardinality equal to the Vortex file such that the filter or projection expressions can reference data from the input stream.

In both cases, we can note that the value of these "variables" is a constant literal by the time the evaluation split of the scan is created. And further that the dtype of the variables are known when the scan is constructed.

I'll ignore the zipped scan from now, since it actually reduces down to the RowIdx use-case except we assign arbitrary arrays instead of a SequenceArray representing the current row offsets.

The proposal simplifies the expression evaluation Scope:
```rust
struct Scope {
  root: ArrayRef,
  vars: ScopeVars,
}
```

Where `ScopeVars` is a `TypeMap` (a HashMap keyed by type ID, commonly used to attach context in middleware-style frameworks, e.g. actix).

We then setup our RowIdx expression to look like this:
```rust
struct RowIdxExpr {}

impl Expr for RowIdxExpr {
  fn dtype(&self, scope: &DType) -> DType {
    DType::Primitive(PType::U64, NonNullable)
  }

  fn evaluate(&self, scope: &Scope) -> VortexResult<ArrayRef> {
    let offset = scope.get_var::<RowIdxVar>();
    Ok(SequenceArray::new(offset, scope.len()).into_array())
  }
}

struct RowIdxVar {
  row_offset: u64
}

impl ScopeVar for RowIdxVar {
  // Some as_any functions
}
```

In our RowIdxLayoutReader, we wrap the passed expression in a `RowIdxBindExpr` to set the variable. This allows us to pass in data at runtime without reconstructing the full expression tree each time.

```rust
struct RowIdxBindExpr {
  child: ExprRef
  row_offset: u64,
}

impl Expr for RowIdxBindExpr {
  fn dtype(...) -> { self.child.dtype(...) }

  fn evaluate(&self, scope: &Scope) -> VortexResult<ArrayRef> {
    scope.set_var(RowIdxVar { row_offset: self.row_offset });
    self.child.evaluate(scope);
    scope.remove_var::<RowIdxVar>();
  }
}
```

danking:
[...]

PR #3672, opened 2025-06-28
Add scope variables to expression scope

gatesn:
Part of #3671 

Adds a new way of defining scope variables (before we deprecate the old method) whereby variables are keyed by type. This is typical for e.g. attaching context to an HTTP request using "middleware", which is essentially what we're doing here during expression evaluation.

### DIFF
--- a/vortex-expr/src/scope_vars.rs
+++ b/vortex-expr/src/scope_vars.rs
@@ -1,0 +1,56 @@
+use std::any::{Any, TypeId};
+use std::hash::{BuildHasherDefault, Hasher};
+
+use vortex_utils::aliases::hash_map::HashMap;
+
+/// A TypeMap based on `https://docs.rs/http/1.2.0/src/http/extensions.rs.html#41-266`.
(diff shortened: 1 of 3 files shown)


robert3005 (approved):
I'm always confused by this pattern because it makes it hard to apply the same thing twice but I guess you never want to do that.

vortex-expr/src/scope_vars.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn write(&mut self, _: &[u8]) {
+        unreachable!("TypeId calls write_u64");
+    }
+
+    #[inline]
+    fn write_u64(&mut self, id: u64) {
+        self.0 = id;
+    }
+}
+
```
a10y, line 30: for future reference, there is a popular crate for this already: https://crates.io/crates/nohash-hasher

this is what rerun uses for their typemap https://github.com/emilk/egui/blob/main/crates/egui/src/util/id_type_map.rs

PR #3673, opened 2025-06-29
chore: Remove Patches#into_parts

robert3005:


vortex-array/src/patches.rs
```diff
@@ -124,11 +124,6 @@ impl Patches {
         }
     }
 
-    // TODO(ngates): remove this...
-    pub fn into_parts(self) -> (usize, usize, ArrayRef, ArrayRef) {
```
robert3005, line 128: there's on place where we needed it where we had owned patches so this should go

a10y (commented):
What's the benefit? Just the fact it was only used once?

a10y (approved):

robert3005:
It’s a bad api imho. We never could leverage the owned part of it

a10y:
I think into_parts and to_parts are very common rust idioms, but in this case maybe weren't that helpful

Issue #2692, opened 2025-03-13
Consider ValidityArray


robert3005:
ValidityArray could wrap an existing array by overlaying validity buffer. You could think of it as a special case of SparseArray where only the validity is sparse

joseph-isaacs:
I think this is worth trying out.

Would we remove validity for other arrays such as Primitive?

It might have a large change on how to implement compute functions

robert3005:
It's there as a MaskedArray

PR #3677, opened 2025-06-30
fix: SparseArray handles take with null indices if the fill value is not null

robert3005:

### DIFF
--- a/vortex-array/src/patches.rs
+++ b/vortex-array/src/patches.rs
@@ -7,8 +7,12 @@ use num_traits::{NumCast, ToPrimitive};
 use serde::{Deserialize, Serialize};
 use vortex_buffer::BufferMut;
 use vortex_dtype::Nullability::NonNullable;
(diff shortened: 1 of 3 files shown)


vortex-array/src/patches.rs
```diff
@@ -309,30 +313,54 @@ impl Patches {
             < Self::PREFER_MAP_WHEN_PATCHES_OVER_INDICES_LESS_THAN
     }
 
+    /// Take the indicies from the patches
+    ///
+    /// Any nulls in take_indices are added to the resulting patches.
+    pub fn take_with_nulls(&self, take_indices: &dyn Array) -> VortexResult<Option<Self>> {
```
robert3005, line 319: I'm not sure if we should add a parameter to the top level function or have two functions like here. Most of the time you want regular take and this is used so far only in sparse array

a10y (approved):

robert3005:
There's an optimisation here where if take_indices is very sparse you would want to use #2692

PR #3678, opened 2025-06-30
feat: teach StructFields empty()

danking:
Pretty minor convenience, but convenient nonetheless.


robert3005 (approved):

robert3005:
I think you might want to create a Default impl as well?

AdamGS:
might as well derive `Default`? (in addition to `empty`)

a10y (approved):

PR #3679, opened 2025-06-30
feat: teach StructFields to Default to empty

danking:
I dunno what's best practice. Should (a) empty call default, (b) default call empty, or (c) should there be no empty? I kinda feel like `empty` is the semantically meaningful name and defaulting to `Self::empty()` seems semantically right? Curious for your thoughts.


AdamGS (approved):

robert3005:
I think you want the default to call your function which is the thing you've done

PR #3683, opened 2025-06-30
feat: simplify again after removing select & merge

danking:
Spiral (after some forthcoming changes) will `merge` two fairly large structs. After `remove_merge`, these become _enormous_ packs. These huge packs are almost always the result of `get_item("field", merge(col("data"), col("aux")))`. This relatively small and readable expression explodes into a huge expression even though it really should simplify to:

    get_item("field", col("data"))

Running one more simplify returns us to a concise expression.


danking:
This is mostly useful because that expression might get sent into the bowels of Vortex machinery and then cause an issue there. It's much easier to figure out what's going on with `$.data.field` rather than `{apple: $.data.apple, ... 200 more fields ...}.field` especially when the field names are long random identifiers.

robert3005 (approved):
All those transformations

PR #3684, opened 2025-06-30
fix: FFI example fixed up, hooked into CI

a10y:
Looks like the FFI test binary got out of date, this brings it back into form.

It also hooks into the cargo examples run setup and I added a CI check to make sure it doesn't break again.

### DIFF
--- a/vortex-ffi/examples/hello-vortex.c
+++ b/vortex-ffi/examples/hello-vortex.c
@@ -3,40 +3,65 @@
 #include "vortex.h"
 
 int main(int argc, char *argv[]) {
-    // enable logging
-    vortex_init_logging(LOG_LEVEL_INFO);
-
     if (argc < 2) {
         printf("Usage: %s <VORTEX_FILE_URI>\n", argv[0]);
         return 1;
     }
 
+    vx_error *error = NULL;
+    vx_session *session = vx_session_new();
+    if (session == NULL) {
+        fprintf(stderr, "vx_session_new return NULL\n");
+        return -1;
+    }
+
     // Open the file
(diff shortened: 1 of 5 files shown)


vortex-ffi/examples/hello-vortex.c
```diff
@@ -3,40 +3,65 @@
...
-
     if (argc < 2) {
         printf("Usage: %s <VORTEX_FILE_URI>\n", argv[0]);
         return 1;
     }
 
+    vx_error *error = NULL;
+    vx_session *session = vx_session_new();
+    if (session == NULL) {
+	    fprintf(stderr, "vx_session_new return NULL\n");
```
lwwmanning: nit: too much whitespace?

lwwmanning (approved):

vortex-ffi/examples/hello-vortex.c
a10y: b/c of the tab?

vortex-ffi/examples/hello_vortex.rs
```diff
@@ -0,0 +1,108 @@
...
+
+const BIN_NAME: &str = "hello_vortex";
+
+pub fn main() -> VortexResult<()> {
+    let bin_path = PathBuf::new()
+        .join(env!("CARGO_MANIFEST_DIR"))
+        .join("examples")
+        .join(BIN_NAME);
+
+    let exit = Command::new("cc")
```
a10y, line 34: i originally tried using the `cc` crate but that doesn't let you build static executables, only .so's

PR #3690, opened 2025-07-01
fix: duckdb scalar date and time conversion

0ax1:
All queries pass the  tpch verification at this point, except for q6.


joseph-isaacs (approved):
Nice

PR #3693, opened 2025-07-01
feat[vortex-duckdb]: add copy from function bindings and vortex impl

joseph-isaacs:


vortex-duckdb/Cargo.toml
```diff
@@ -21,14 +21,16 @@ crate-type = ["staticlib", "cdylib", "rlib"]
...
+arrow = { workspace = true }
 bitvec = { workspace = true }
 crossbeam-queue = { workspace = true }
 glob = { workspace = true }
 itertools = { workspace = true }
 log = { workspace = true }
 num-traits = { workspace = true }
 tempfile = { workspace = true }
-tokio = { workspace = true, features = ["macros", "rt"] }
+tokio = { workspace = true, features = ["macros", "rt", "rt-multi-thread"] }
```
0ax1, line 32: double check whether we need `rt-multi-thread`.

vortex-duckdb/cpp/copy_function.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "duckdb/main/client_context.hpp"
+#include "duckdb/main/connection.hpp"
+#include "duckdb/parser/parsed_data/create_copy_function_info.hpp"
+#include "duckdb_vx/error.hpp"
+
+using namespace duckdb;
+
+namespace vortex {
+
+struct CCopyFunctionInfo {
```
0ax1: What's the value of wrapping `duckdb_vx_copy_func_vtab_t` into `CCopyFunctionInfo`? Can we use `duckdb_vx_copy_func_vtab_t` directly?

vortex-duckdb/cpp/copy_function.cpp
```diff
@@ -0,0 +1,163 @@
...
+    duckdb_vx_error error_out = nullptr;
+    bind.info->vtab.copy_to_finalize(bind.ffi_data->DataPtr(), global.ffi_data->DataPtr(), &error_out);
+    if (error_out) {
+        throw ExecutorException(IntoErrString(error_out));
+    }
+}
+
+extern "C" duckdb_vx_copy_func_vtab_t *get_vtab_one() {
+    if (copy_vtab_one == nullptr) {
+        copy_vtab_one = new duckdb_vx_copy_func_vtab_t;
```
0ax1: Strictly speaking, this will show up as a leak.

0ax1: We could have `copy_vtab_one` as a value type and atomically init once with:

```
 static std::once_flag flag;
 std::call_once(flag, initializer());
```

vortex-duckdb/src/convert/vector.rs
```diff
@@ -0,0 +1,317 @@
...
+    let len = chunk.len();
+
+    let columns = (0..chunk.column_count())
+        .zip(field_names.iter())
+        .map(|(i, name)| {
+            let mut vector = chunk.get_vector(i);
+            flat_vector_to_arrow_array(&mut vector, len.as_usize())
+                .map(|array_data| {
+                    assert_eq!(array_data.len(), chunk.len().as_usize());
+                    // let array: Arc<dyn Array> = Arc::new(array_data);
```
0ax1: drop dead code

vortex-duckdb/src/convert/vector.rs
```diff
@@ -0,0 +1,317 @@
...
+                ))),
+            );
+
+            Ok(Arc::new(structs))
+        }
+        _ => todo!(),
+    }
+}
+
+/// converts a `DataChunk` to arrow `RecordBatch`
```
0ax1: Consider full sentences for doc strings. Lower case first char `.` at the end.

vortex-duckdb/src/convert/vector.rs
```diff
@@ -0,0 +1,317 @@
+/// Copied of duckdb-rs (https://github.com/duckdb/duckdb-rs/blob/main/crates/duckdb/src/vtab/arrow.rs)
```
0ax1, line 1: Any unit tests we can copy too? :)

vortex-duckdb/src/copy.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    fn copy_to_finalize(
+        _bind_data: &Self::BindData,
+        init_global: &mut Self::GlobalState,
+    ) -> VortexResult<()> {
+        RUNTIME.block_on(async {
+            if let Some(sink) = init_global.sink.take() {
+                drop(sink)
+            }
```
0ax1, line 84: Can we add some docs for the future reader here how the streaming copy roughly works. Without context, one wonders why tokio is looped in here.

vortex-duckdb/src/copy.rs
```diff
@@ -0,0 +1,134 @@
...
+            Ok(())
+        })
+    }
+
+    fn init_global(
+        bind_data: &Self::BindData,
+        file_path: String,
+    ) -> VortexResult<Self::GlobalState> {
+        // The channel size 32 was chosen arbitrarily.
+        let (sink, rx) = mpsc::channel(32);
```
0ax1, line 100: I assume this is unbounded though, and pre-allocates 32 channels?

vortex-duckdb/src/copy.rs
```diff
@@ -0,0 +1,134 @@
...
+    use crate::{Connection, Database};
+
+    fn database_connection() -> Connection {
+        let db = Database::open_in_memory().unwrap();
+        let connection = db.connect().unwrap();
+        crate::register_table_functions(&connection).unwrap();
+        connection
+    }
+
+    #[test]
```
0ax1: Can we do a test where we read back a written file and check whether the results are correct?

vortex-duckdb/src/duckdb/copy_function/callback.rs
```diff
@@ -0,0 +1,109 @@
...
+use itertools::Itertools;
+use vortex::error::VortexExpect;
+
+use crate::cpp;
+use crate::cpp::{
+    duckdb_data_chunk, duckdb_logical_type, duckdb_vx_copy_func_bind_input, duckdb_vx_error,
+};
+use crate::duckdb::{CopyFunction, Data, DataChunk, LogicalType, try_or, try_or_null};
+
+pub(super) unsafe extern "C" fn bind_callback<T: CopyFunction>(
```
0ax1: I guess this is equivalent to `pub(crate)`? I find it's a bit easier to reason about absolute visibility def rather than relative. wdyt?

vortex-duckdb/src/duckdb/vector.rs
```diff
@@ -51,6 +51,28 @@ impl Vector {
         unsafe { std::slice::from_raw_parts_mut(ptr.cast::<T>(), length) }
     }
 
+    pub fn as_slice_with_len<T>(&self, length: usize) -> &[T] {
+        let ptr = unsafe { cpp::duckdb_vector_get_data(self.as_ptr()) };
+        unsafe { std::slice::from_raw_parts_mut(ptr.cast::<T>(), length) }
+    }
+
+    pub fn row_is_null(&self, row: u64) -> bool {
+        // use idx_t entry_idx = row_idx / 64; idx_t idx_in_entry = row_idx % 64; bool is_valid = validity_mask[entry_idx] & (1 « idx_in_entry);
```
0ax1, line 71: >   // use idx_t entry_idx = row_idx / 64; idx_t idx_in_entry = row_idx % 64; bool is_valid = validity_mask[entry_idx] & (1 « idx_in_entry);

what is this about?

vortex-duckdb/src/convert/vector.rs
joseph-isaacs: I think we should just move over to duckdb -> vortex directly so I would way there is less value in porting tests over

vortex-duckdb/Cargo.toml
joseph-isaacs: can we keep this for now

vortex-duckdb/cpp/copy_function.cpp
```diff
@@ -0,0 +1,163 @@
...
+    unique_ptr<CData> ffi_data;
+};
+
+static duckdb_vx_copy_func_vtab_t *copy_vtab_one;
+
+unique_ptr<FunctionData> c_bind_one(ClientContext &context, CopyFunctionBindInput &info,
+                                    const vector<string> &column_names,
+                                    const vector<LogicalType> &column_types) {
+
+    auto c_column_names = vector<char *>();
```
0ax1, line 44: Drop the const cast in favor of:
```
    auto c_column_names = vector<const char *>();
    for (const auto &col_id : column_names) {
        c_column_names.push_back(col_id.c_str());
    }
```

vortex-duckdb/src/lib.rs
```diff
@@ -1,12 +1,16 @@
 #![allow(clippy::missing_safety_doc)]
 use std::ffi::{CStr, c_char};
+use std::sync::LazyLock;
 
 // **WARNING begin this includes duckdb-rs, which is required to link in the symbol from libduckdb-sys.
 #[allow(unused_imports)]
```
0ax1: I guess this is a leftover from previous PR: `#[allow(unused_imports)]`

vortex-duckdb/cpp/copy_function.cpp
```diff
@@ -0,0 +1,151 @@
...
+    auto &bind = bind_data.Cast<CCopyBindData>();
+    auto &global = gstate.Cast<CCopyGlobalData>();
+    duckdb_vx_error error_out = nullptr;
+    bind.vtab.copy_to_finalize(bind.ffi_data->DataPtr(), global.ffi_data->DataPtr(), &error_out);
+    if (error_out) {
+        throw ExecutorException(IntoErrString(error_out));
+    }
+}
+
+extern "C" duckdb_vx_copy_func_vtab_t *get_vtab_one() {
```
0ax1, line 113: I guess strictly speaking we don't need a separate function, we can declare the global static var in the header and drop the get_vtab_one function and just use the variable.

joseph-isaacs: I prefer this fn

0ax1 (approved):
👌

PR #3701, opened 2025-07-01
chore: polish benchmark row count verification

0ax1:
In particular, this commit adds row count verification for the clickbench benchmark. Further, each row count gets verified on the fly rather than collected and verified at the end.


bench-vortex/src/bin/tpch.rs
```diff
@@ -306,7 +246,13 @@ async fn bench_main(
                         })
                         .await;
 
-                    row_counts.push((query_idx, format, row_count));
+                    // Query 15 is flaky for DataFusion: https://github.com/vortex-data/vortex/issues/2395.
+                    if query_idx != 15 {
```
robert3005: we fixed this

0ax1: removed the condition

robert3005 (approved):

PR #3704, opened 2025-07-01
chore: auto-gen c header via bindgen for vortex-ffi

0ax1:


0ax1:
I'll make the header "transient" and add it to `.gitignore`.

a10y (approved):

a10y:
@0ax1 can we leave it checked in actually? I think it's better for consumers if they can just vendor the header file into their own repo

robert3005 (approved):

robert3005:
I think the only option you have here is to exclude vortex-ffi from miri...

robert3005:
If we are going to check it in we need to add a check that it's up to date

0ax1:
> If we are going to check it in we need to add a check that it's up to date

I think I can simply write the gen header conditionally.

PR #3492, opened 2025-06-09
chore: make session the One True Entrypoint

a10y:
Signed-off-by: Andrew Duffy <andrew@a10y.dev>

Issue #3453, opened 2025-06-04
Expressions becomes enum, ComputeFn becomes ScalarFn


gatesn:
This issue encapsulates a few changes, but I believe the end state makes sense!

* We should redefine `ComputeFn` to mean `ScalarFn`. This let's us make some nice assumptions and keep the trait tightly defined.
* Add an `AggFn` to cover our current aggregation functions, e.g. sum, count, min, max.
* Make a Vortex Expression and enum with a `ScalarFn` variant.
* Any expression that _can_ be implemented as a ScalarFn should be. Note that functions that require short-circuiting, e.g. AND, OR, should have their own entry in the enum, rather than delegate to the scalar function. 
* While `ScalarFn::Options` doesn't need to be serializable (we should support invoking compute functions with opaque runtime options data), a ScalarFn wrapped in an Expression _does_ need to be serializable. This is because...
* ...Expression is serializable as a well-defined protobuf union. This is one way we can create expressions over the C API.
* FilterFn and TakeFn should be added to the ArrayOperations trait (returns `Result<Option<Array>>` since they're optionally implemented) since they're not scalar functions.
* Vortex Session holds a registry of compute functions that are used when deserializing the protobuf to load the actual implementation.

joseph-isaacs:
I think the scalar compute fn changes will reduce the amount of code we have to write. 

I have a few qs/comments. 

What is the relation between agg and scalar exprs, can one contain the other?

Do you think window functions/exprs will follow?

Also there will likely be an overhead to this implementation technique we cannot monomophorize specific scalar function impl since it goes via dynamic dispatch. Do you think we would keep common compute fns (e.g. compare) as is?

joseph-isaacs:
I think the scalar compute fn changes will reduce the amount of code we have to write. 

I have a few qs/comments. 

What is the relation between agg and scalar exprs, can one contain the other?

Do you think window functions/exprs will follow?

Also there will likely be an overhead to this implementation technique we cannot monomophorize specific scalar function impl since it goes via dynamic dispatch. Do you think we would keep common compute fns (e.g. compare) as is?

gatesn:
> What is the relation between agg and scalar exprs, can one contain the other?

There's no AggExpr for now, just an AggFn. So its input is an actual Array. 

In the future, we may have one. In which case, AggExpr could contain Expr (implied ScalarExpr): `SUM(a + 1)`

[...]

PR #3713, opened 2025-07-01
VortexExpr VTables

gatesn:
Adds the same vtable machinery as arrays and layouts already use. It uses the "Encoding" naming scheme from arrays and layouts. I don't particularly like it, but it's consistent. Open to renames later.

Further, adds an expression registry to the Vortex session that will be used for deserialization.

Expressions only decide their "options" serialization. So in theory, can support many container formats, not just proto, provided each expression can deserialize their own options format.


a10y:
If we're doing this, we should probably revive #3492 as well and just push more of the public API to go through the session object, so the session owns all of the vtables and everything that needs to do dynamic dispatch borrows from it

0ax1:
image

"Guys, @gatesn is back"

gatesn:
Or we go in the direction of making ComputeFns extensible, and not expressions... https://github.com/vortex-data/vortex/issues/3453

vortex-expr/src/exprs/binary.rs
```diff
@@ -1,87 +1,145 @@
...
+use crate::{
+    AnalysisExpr, ExprEncodingRef, ExprId, ExprRef, IntoExpr, Operator, Scope, ScopeDType,
+    StatsCatalog, VTable, vtable,
+};
+
+vtable!(Binary);
 
-#[derive(Debug, Clone, Eq, Hash)]
 #[allow(clippy::derived_hash_with_manual_eq)]
+#[derive(Debug, Clone, Hash)]
```
robert3005, line 21: Why no Eq?

vortex-expr/src/encoding.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::fmt::{Debug, Display, Formatter};
+
+use arcref::ArcRef;
+use vortex_array::DeserializeMetadata;
+use vortex_error::{VortexExpect, VortexResult};
+
+use crate::{ExprRef, IntoExpr, VTable};
+
+pub type ExprId = ArcRef<str>;
+pub type ExprEncodingRef = ArcRef<dyn ExprEncoding>;
```
robert3005, line 14: The encoding for an array made a logical sense in my mind, encoding for expression is very forced analogy. We can rename it later

robert3005 (approved):

PR #3721, opened 2025-07-02
chore: remove unused `worker` dep from repo

a10y:
Noticed recent renovate PR to bump `worker`, but we don't have any use for it.


robert3005 (approved):

AdamGS:
yet another problem with thiserror - too easy to create these specialized error types that make it harder to realize you don't use some dependency

PR #3730, opened 2025-07-02
chore[benchmark]: enable duckdb:vortex for tpch (local) and clickbench (local)

joseph-isaacs:


bench-vortex/src/bin/tpch.rs
```diff
@@ -392,18 +392,7 @@ fn verify_duckdb_tpch_results(
...
-        .filter(|entry| {
-            queries
-                .as_ref()
-                .is_none_or(|queries| queries.contains(&(entry.0 + 1)))
-                // Skip query 6 until fixed.
-                && entry.0 != 5
-        })
-        .map(|query_file| query_file.1)
-    {
+    for query_file in query_files.iter() {
```
0ax1: We probably still want the query filter such that we only verify queries that were run as part of the benchmark.

Consider only dropping:
```
     // Skip query 6 until fixed.
                && entry.0 != 5
```

0ax1 (approved):
📈

PR #3735, opened 2025-07-02
feat: public constants for row id fields

danking:
I want to use these in Spiral rather than magic strings.

cc: @blaginin


robert3005:
We want to remove file_index https://github.com/vortex-data/vortex/pull/3613 but we can expose it for now

robert3005 (approved):

PR #3736, opened 2025-07-02
feat[encodings/dict]: allow different nullability of codes and values in dict

joseph-isaacs:


encodings/dict/src/array.rs
```diff
@@ -46,27 +46,11 @@ pub struct DictArray {
 pub struct DictEncoding;
 
 impl DictArray {
-    pub fn try_new(mut codes: ArrayRef, values: ArrayRef) -> VortexResult<Self> {
+    pub fn try_new(codes: ArrayRef, values: ArrayRef) -> VortexResult<Self> {
```
0ax1, line 86: Can we add a unit test checking for diff nullability of codes and values?

robert3005 (changes requested):
There's a small optimisation possible at the reader level but at this level it's just a matter of casts. I think this would make more sense if we had dict array as a selection vector.

robert3005:
If we are going to do this then don't you also want the nulls to live in values by default?

joseph-isaacs:
By default, do you mean the in writer?

robert3005:
I mean the dictbuilder in btrblocks compressor and the dictwriter

robert3005 (approved):
Let's do it, this was the original dict implementation before we simplified it. Before we see any benefit from it we have to make sure to change the writer and compressor

PR #3739, opened 2025-07-02
chore[benchmark]: add -o option to specify where the write results

joseph-isaacs:
This change allow CI to print more diagnostics and still output a result file


a10y (approved):
This seems fine, the other way I could imagine doing this is to instead send other diagnostics to stderr instead

PR #3742, opened 2025-07-02
chore: update toolchain to nightly-2025-06-26

danking:


vortex-btrblocks/src/float.rs
```diff
@@ -74,9 +69,6 @@ struct ALPRDScheme;
 #[derive(Debug, Copy, Clone)]
 struct DictScheme;
 
-#[derive(Debug, Copy, Clone)]
-struct RunEndScheme;
```
danking, line 78: Apparently this is completely unused?

vortex-btrblocks/src/float/stats.rs
```diff
@@ -48,6 +48,7 @@ pub struct FloatStats {
     pub(super) null_count: u32,
     // cache for validity.true_count()
     pub(super) value_count: u32,
+    #[allow(dead_code)]
     pub(super) average_run_length: u32,
```
danking, line 52: This seems to only be used in tests, so clippy is upset.

vortex-flatbuffers/src/lib.rs
```diff
@@ -55,6 +57,7 @@ pub mod dtype;
 #[allow(clippy::unwrap_used)]
 #[allow(clippy::borrow_as_ptr)]
 #[allow(dead_code)]
+#[allow(mismatched_lifetime_syntaxes)]
```
danking, line 66: Even after updating to latest flatc, it still generated bad code, so I just ignore it.

a10y (approved):
Idk why they added this lint, feels unnecessary to add back all the '_ but 🤷

a10y (approved):
Idk why they added this lint, feels unnecessary to add back all the '_ but 🤷

vortex-btrblocks/src/float/stats.rs
robert3005: @AdamGS tried doing this before. The RunEndScheme for floats is unused and if you enable it you degrade performance. I would just delete it and file an issue

robert3005:
oh wait I forgot we are stuck on the GCE docs issue

robert3005:
@a10y there's a fascinating discussion that there are cases where the lifetime ellided syntax is unsound, however, we never have that case :/

PR #3745, opened 2025-07-03
chore: make flavor part of clickbench dataset

0ax1:


bench-vortex/src/bin/clickbench.rs
```diff
@@ -165,6 +166,7 @@ fn main() -> anyhow::Result<()> {
             format,
             &base_url,
             args.single_file,
+            args.flavor,
```
joseph-isaacs: can you create the enum in this fn instead of passing args?

0ax1: done

joseph-isaacs (approved):

PR #3747, opened 2025-07-03
feat: Hide as many of FieldNames' Arcs as possible

AdamGS:
The main purpose of this PR is to hide the implementation details of `FieldNames` from the public API, and pushing all the required `from/into` calls internally.

The change in API can be seen in all the test changes, which I think are nicer.

### DIFF
--- a/vortex-dtype/src/dtype.rs
+++ b/vortex-dtype/src/dtype.rs
@@ -1,5 +1,6 @@
 use std::fmt::{Debug, Display, Formatter};
 use std::hash::Hash;
+use std::ops::Index;
 use std::sync::Arc;
 
 use DType::*;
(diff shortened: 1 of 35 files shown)


a10y (approved):

vortex-dtype/src/dtype.rs
```diff
@@ -13,8 +13,151 @@ use crate::{ExtDType, FieldDType, PType, StructFields};
...
+    }
+}
+
+impl AsRef<[FieldName]> for FieldNames {
+    fn as_ref(&self) -> &[FieldName] {
+        &self.0
+    }
+}
+
+impl std::ops::Index<usize> for FieldNames {
```
robert3005: import this

robert3005 (approved):

PR #3758, opened 2025-07-03
chore[ci]: java ci is slow

joseph-isaacs:


.github/workflows/ci.yml
```diff
@@ -342,10 +342,8 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 120
     steps:
-      - name: Checkout sources
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
         with:
-          fetch-depth: 0
```
a10y, line 348: this just avoids fetching duckdb?

joseph-isaacs: is seems to make the fetch much slower (maybe due to caching)

joseph-isaacs: checkout goes from 45s to 6s

a10y: I think we need this otherwise the version tags aren't pulled

0ax1: There seems to be an extra option for tags: https://github.com/actions/checkout
```
   # Whether to fetch tags, even if fetch-depth > 0.
    # Default: false
    fetch-tags: ''
```

a10y:
I tried to see if this works locally. Here's what I want to see

```
# from my normal vortex checkout
$ ./gradlew printVersion
> Task :printVersion
0.41.2-1-g9321799
```

Now I make a new checkout and try there

```
$ git clone --depth=1 https://github.com/vortex-data/vortex.git vortex2
$ cd vortex2/java
$ ./gradlew printVersion
> Task :printVersion
9321799

# let's try fetch_tags
$ git fetch --tags
$ ./gradlew printVersion
> Task :printVersion
9321799
```

[This SO](https://stackoverflow.com/questions/35691922/how-can-i-use-git-describe-match-sensibly-in-combination-with-a-shallow-clon) seems to indicate that git needs the commit history at least until the most recent tag, which it varies how many commits we have in between tags. For that reason I think we need to keep the fetch settings and look for other ways to speed up java CI

.github/workflows/ci.yml
0ax1: `fetch-tags: true` might do the trick

robert3005:
I was confused. You do need the history for git version to work but do we need the version only for publish and can live with wrong version for tests?

a10y:
Hmm, we probably can actually. I'm trying to remember if I did this for a reason but maybe it's fine if we remove

a10y (approved):

PR #3764, opened 2025-07-04
Compact Compression

### LINKED ISSUE #3611: Compact/heavyweight compression strategy
The current BtrBlocks compression strategy is pretty quick and can reduce memory footprint, but I think it is important to have a CompressionStrategy for storage, where compression ratio is more important. Generally we should use Zstd for string-like types and Pco for numerical types, but I could use some help sorting out the best way to set this up in Vortex:

* Currently the one compression strategy is just implicitly the default, and a lot of APIs don't accept a compression strategy (e.g. the Python `vortex.io.write`). Should we make these APIs accept a compression strategy argument?
* Dictionary+Pco would typically hurt compression ratio. I'm not certain and haven't measured whether Dictionary+Zstd offers any benefits over just Zstd in terms of compressed size, but my guess is that it also won't help, since Zstd identifies these patterns easily. Are we happy omitting Dictionary here? There may be a possibility that in the future Vortex caches Zstd-decompressed frames, which could require much more memory usage than Dictionary-encoded data. That said, my leaning is that these are minor risks, so keeping it simple and avoiding Dictionary may be the best option for now.
* How finely should we paginate the data by default? I.e. if a user wants to read one row, how many unique values should they need to decompress? This a trades off vs (de)compression speed and compressed size. I'd start with a few k values (maybe 10-100us decompression time for numerical data, possibly longer for long strings) if that sounds reasonable.

mwlon:
Added a CompactCompression strategy that generally uses Zstd for binary/string data and Pco for numeric data types.

Size comparison, using the NYC Taxi dataset and running e.g. `cargo run --release -p vortex-tui convert fhvhv_tripdata_2023-04.parquet --strategy compact` to produce files:
```
433M fhvhv_tripdata_2023-04_btrblocks.vortex
334M fhvhv_tripdata_2023-04_compact_8192.vortex
321M fhvhv_tripdata_2023-04_compact_inf.vortex
469M fhvhv_tripdata_2023-04.parquet
```

Here the two compact strategies are using up to 8192 values per page versus "inf", as many as possible. Using 8192 (the default I put in the code) slightly increases size but allows for faster access into slices and lines up with statistics nicely for potential pushdown filters.

NOT HAPPENING IN THIS PR: compression for bool or decimal arrays; or dict encoding for variable-length types. These can be improvements for later work.

Fixes #3611 .

Other changes this incurred:

* Added --strategy arg to convert command (and simplified flags stuff a bit)
* Added VarBinView support for Zstd encoding
* Added unit tests for all the new functionality


mwlon:
Closing this because fixing signoff is impossible

robert3005:
FWIW the fix is that you must rewrite every commit message with a rebaes which is a bit annoying

mwlon:
@robert3005 I'm not sure how to do that after a merge. It seems it requires a rebase, and the rebase command the bot recommends doesn't work (signs off other people's commits and creates a huge commit difference).

gatesn:
Honestly I find the DCO check incredibly annoying and quite a blunt instrument for achieving what it's trying to achieve (i.e. imlpicitly agreeing to licensing terms). I'll look into some alternatives...

PR #3765, opened 2025-07-04
Compact Compression

### LINKED ISSUE #3611: Compact/heavyweight compression strategy

mwlon:
Added a CompactCompression strategy that generally uses Zstd for binary/string data and Pco for numeric data types.

Size comparison, using the NYC Taxi dataset and running e.g. `cargo run --release -p vortex-tui convert fhvhv_tripdata_2023-04.parquet --strategy compact` to produce files:

```
433M fhvhv_tripdata_2023-04_btrblocks.vortex
334M fhvhv_tripdata_2023-04_compact_8192.vortex
321M fhvhv_tripdata_2023-04_compact_inf.vortex
469M fhvhv_tripdata_2023-04.parquet (zstd compressed)
```

Here the two compact strategies are using up to 8192 values per page versus "inf", as many as possible. Using 8192 (the default I put in the code) slightly increases size but allows for faster access into slices and can (in the non-null, non-list case) line up with statistics nicely for potential pushdown filters.

NOT HAPPENING IN THIS PR (leaving this to future work):

* compression for bool arrays
* compression for decimal array 
* dict encoding for variable-length types

Other changes this incurred:

* Added --strategy arg to convert command (and simplified flags stuff a bit)
* Added VarBinView support for Zstd encoding
* Added unit tests for all the new functionality

Fixes https://github.com/vortex-data/vortex/issues/3611 .


encodings/zstd/src/array.rs
```diff
@@ -88,11 +94,46 @@ fn choose_max_dict_size(uncompressed_size: usize) -> usize {
...
+    while offset < buffer.len() {
+        let str_len = u32::from_le_bytes(
+            buffer
+                .get(offset..offset + 4)
+                .ok_or_else(|| vortex_err!("Zstd buffer for VarBinView was corrupt"))?
+                .try_into()?,
+        ) as usize;
+        offset += 4;
+        let value = &buffer[offset..offset + str_len];
+        // slightly surprising that binary views only support u32 offsets?
```
gatesn: I agree... I think it means that binary views _cannot_ support values larger than 4GB?

encodings/zstd/src/serde.rs
```diff
@@ -24,6 +26,16 @@ pub struct ZstdMetadata {
     pub dictionary_size: u32,
     #[prost(message, repeated, tag = "2")]
     pub frames: Vec<ZstdFrameMetadata>,
+    // Tracks the canonical array type stored in the compressed data
+    #[prost(enumeration = "CanonicalArrayType", tag = "3")]
+    pub canonical_array_type: i32,
```
gatesn: Is this not implied by the array's DType?

encodings/zstd/src/array.rs
mwlon: Yeah I think this means depending on the implementation, the largest possible vbv might be 8GB (its final string starts from a 4GB offset and is itself 4GB long).

gatesn: Well you also get u32 different buffer IDs, so I think _total_ VBV size is u64

mwlon: Ah true

encodings/zstd/src/serde.rs
mwlon: Ah indeed it is; simplified it now.

mwlon:
@lwwmanning @gatesn any next thoughts on this? I probably should have made this 2 PRs to be more digestible

encodings/zstd/src/serde.rs
```diff
@@ -15,6 +15,8 @@ use crate::{ZstdArray, ZstdEncoding, ZstdVTable};
 pub struct ZstdFrameMetadata {
```
gatesn, line 15: I wonder how big this metadata is expected to be?

I missed this before, but seems like ZstdMetadata scales with num frames. I wonder if it's worth storing this in a `Buffer<FrameMetadata>` rather than in the protobuf?

Relatedly, is it worth even storing nvalues for primitive typed frames? We could have `uncompressed_sizes: Buffer<u64>; frame_lens: Buffer<u32>` as two buffer children of this array?

encodings/zstd/src/array.rs
```diff
@@ -80,6 +80,12 @@ pub struct ZstdArray {
     slice_stop: usize,
 }
 
+struct Frames {
+    dictionary: Option<Buffer<u8>>,
```
gatesn: Nit: we have a ByteBuffer alias

encodings/zstd/src/array.rs
```diff
@@ -88,11 +94,46 @@ fn choose_max_dict_size(uncompressed_size: usize) -> usize {
...
 
-fn collect_valid(parray: &PrimitiveArray) -> VortexResult<PrimitiveArray> {
+fn collect_valid_primitive(parray: &PrimitiveArray) -> VortexResult<PrimitiveArray> {
     let mask = parray.validity_mask()?;
     filter(&parray.to_array(), &mask)?.to_primitive()
 }
 
+fn collect_valid_vbv(vbv: &VarBinViewArray) -> VortexResult<(Buffer<u8>, Vec<usize>)> {
+    let mask = vbv.validity_mask()?;
+    let mut buffer = BufferMut::empty();
```
gatesn: We know the capacity for this up-front.

encodings/zstd/src/array.rs
```diff
@@ -88,11 +94,46 @@ fn choose_max_dict_size(uncompressed_size: usize) -> usize {
...
+    let mut value_byte_indices = Vec::new();
+    for i in 0..vbv.len() {
+        if mask.value(i) {
```
gatesn: I think you might want `match mask.boolean_buffer() { AllTrue ... AllFalse .. BooleanBuffer }` to handle fast paths. BooleanBuffer can then use an iterator to avoid a function call for each value.

encodings/zstd/src/array.rs
```diff
@@ -88,11 +94,46 @@ fn choose_max_dict_size(uncompressed_size: usize) -> usize {
...
+            value_byte_indices.push(buffer.len());
+            // here's where we write the string lengths
+            let value = vbv.bytes_at(i);
```
gatesn: Similarly, there's an array accessor trait to iterate bytes from the varbinview

encodings/zstd/src/array.rs
```diff
@@ -115,32 +156,20 @@ impl ZstdArray {
...
-        };
-
-        let mut frame_value_starts = (0..n_values).step_by(values_per_frame).collect::<Vec<_>>();
-        let n_frames = frame_value_starts.len();
-        frame_value_starts.push(values.len()); // for convenience, include the stop of the last frame
-        let value_bytes = values.byte_buffer();
+        n_values: usize,
+    ) -> VortexResult<Frames> {
+        let n_frames = frame_byte_starts.len();
+        frame_byte_starts.push(value_bytes.len()); // for convenience, include the stop of the last frame
```
gatesn: Seems odd that the caller would have allocated the vector, for you to push one more element into it. Possibly triggering a re-allocation?

encodings/zstd/src/array.rs
```diff
@@ -160,17 +189,57 @@ impl ZstdArray {
...
         for i in 0..n_frames {
-            let uncompressed = &value_bytes
-                .slice(frame_value_starts[i] * byte_width..frame_value_starts[i + 1] * byte_width);
+            let uncompressed = &value_bytes.slice(frame_byte_starts[i]..frame_byte_starts[i + 1]);
             let compressed = compressor
                 .compress(uncompressed)
                 .map_err(|err| VortexError::from(err).with_context("while compressing"))?;
             frame_metas.push(ZstdFrameMetadata {
                 uncompressed_size: uncompressed.len() as u64,
+                n_values: values_per_frame.min(n_values - i * values_per_frame) as u64,
```
gatesn, line 215: This suggests you might actually only need to store `frame_len` and `last_frame_len`? Or are you allowing for a world with variable frame lengths?

encodings/zstd/src/array.rs
```diff
@@ -249,34 +404,61 @@ impl ZstdArray {
...
         // Last, we slice the exact values requested out of the decompressed data.
         let slice_validity = self
             .unsliced_validity
             .slice(self.slice_start, self.slice_stop)?;
-        let slice_values_buffer = decompressed.freeze().slice(
-            slice_uncompressed_start - skipped_uncompressed
-                ..slice_uncompressed_stop - skipped_uncompressed,
-        );
-
-        let primitive = PrimitiveArray::from_values_byte_buffer(
```
gatesn, line 272: I must have missed this before. By only compressing "valid" values, you always perform a copy for decompression, and then a second copy to splat the non-null values.

Is it not worth just inserting the zero length strings / zero primitive values into the zstd stream and allowing it to deal with them? I would have thought it could handle it quite well.

gatesn (commented):
Sorry, some of my questions relate to the previous zstd PR! But otherwise this looks good

encodings/zstd/src/array.rs
mwlon: Filtering them out helps compression noticeably, even if they're constant. But these null values could be anything, even uniformly distributed bytes, so this especially helps the worst case.

encodings/zstd/src/array.rs
mwlon: I am allowing for future flexibility in choosing the frame boundaries. It could be helpful to couple frame boundaries and stats in the future.

encodings/zstd/src/serde.rs
mwlon: Are there any operations that would actually benefit from moving this to a buffer? This should be much smaller than the compressed frames, something like 10 bytes per frame meta vs 16kB+ (8k*2B) per compressed frame for e.g. uniform u16s.

I think simplicity is valuable here, and it's nice to ensure the uncompressed sizes and frame lens line up by making them part of the same struct.

encodings/zstd/src/array.rs
mwlon: set capacity now

encodings/zstd/src/array.rs
mwlon: fixed

encodings/zstd/src/array.rs
mwlon: What trait are you referring to? I don't see it

mwlon: Ah never mind, found it.

encodings/zstd/src/array.rs
mwlon: Yeah... got rid of this but had to use loops instead of iterators

encodings/zstd/src/array.rs
mwlon: Done, but there's no way to make a fast path for `All`.

encodings/zstd/src/array.rs
a10y: Technically––TECHNICALLY––the spec says that it's an i32, so really take all of those numbers and cut them in half. But non-Java implementations play a bit [fast and loose](https://github.com/apache/arrow-rs/issues/6172) with the rules here...

a10y: This is maybe a bigger concern for us now that we 

a) care about working with large strings, and
b) have a Java API which uses Arrow FFI to exchange data

encodings/zstd/src/serde.rs
gatesn: How big are frames roughly? Just wondering if there's any value in these being recursively compressed. Maybe overkill though

gatesn (approved):

mwlon:
> How big are frames roughly? Just wondering if there's any value in these being recursively compressed. Maybe overkill though

It's currently configured by `values_per_frame` or `values_per_page`, which defaults to 8192.

PR #3779, opened 2025-07-05
Remove filter_evaluation in favor of projection_evaluation

gatesn:
It's functionally almost identical, if there's no performance regression, then it's nicer not to have it.

### DIFF
--- a/vortex-layout/src/layouts/filter.rs
+++ b/vortex-layout/src/layouts/filter.rs
@@ -13,6 +13,7 @@ use itertools::Itertools;
 use parking_lot::RwLock;
 use sketches_ddsketch::DDSketch;
 use vortex_array::stats::Precision;
+use vortex_array::{ArrayRef, IntoArray};
 use vortex_dtype::{DType, FieldMask};
 use vortex_error::{VortexExpect, VortexResult, vortex_err, vortex_panic};
 use vortex_expr::forms::cnf::cnf;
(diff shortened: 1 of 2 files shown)


vortex-layout/src/layouts/filter.rs
```diff
@@ -106,25 +107,44 @@ impl LayoutReader for FilterLayoutReader {
...
     }
 
     fn projection_evaluation(
         &self,
         row_range: &Range<u64>,
         expr: &ExprRef,
     ) -> VortexResult<Box<dyn ArrayEvaluation>> {
-        // Pass-through all projection expressions to the child layout reader.
-        self.child.projection_evaluation(row_range, expr)
+        // If the expression resolves to a boolean, we split it into conjuncts for evaluation.
```
gatesn, line 124: This isn't quite right, but it's enough to test for a performance regression.

We might actually need to maintain a boolean array instead of a mask, and then check for true_count to bail early. i.e. "not false and not null", in order to maintain the correct null semantics of the expression.

robert3005: fwiw this is the error
```
FAILED test/test_dataset.py::test_to_table - pyo3_runtime.PanicException: Masks must have the same length
Backtrace:
   0: vortex_mask::bitops::<impl core::ops::bit::BitAnd for &vortex_mask::Mask>::bitand
   1: <vortex_layout::layouts::filter::FilterEvaluation as vortex_layout::reader::ArrayEvaluation>::invoke::{{closure}}
   2: vortex_layout::scan::tasks::split_exec::{{closure}}
   3: vortex_layout::scan::tasks::split_exec::{{closure}}
   4: <vortex_array::stream::adapter::ArrayStreamAdapter<S> as futures_core::stream::Stream>::poll_next
```

PR #3784, opened 2025-07-07
chore[bench]: store all run time measurements

joseph-isaacs:


0ax1 (approved):
👍

bench-vortex/src/measurements.rs
```diff
@@ -206,7 +213,8 @@ pub struct QueryMeasurementJson {
     pub storage: String,
     pub dataset: BenchmarkDataset,
     pub unit: String,
-    pub value: MeasurementValue,
+    pub value: u128,
```
0ax1, line 216: Guess the field names could be more descriptive for `value` and `values`.

joseph-isaacs: I cannot change value, maybe values can be all_runs?

0ax1: That'd be a nice change.

joseph-isaacs: `all_runtimes`

PR #3789, opened 2025-07-07
fix: DecimalByteParts compare with overflow and nullable elements

a10y:
Fuzzer is [currently failing](https://github.com/vortex-data/vortex/actions/runs/16105049077/job/45439258523) on an optimization pathway in DecimalBytePartsArray, where the constant cannot be represented in the narrowed PType of DBP's narrowed PType.

So for example if you have a nullable DecimalByteParts with values `None, Some(1i32), Some(2i32)` and you do the comparison `<= ConstantArray(Some(999999999999999i128))`, we'd first try and cast `999999999999999i128` to an `i32`, and when that fails we'd assume that it is larger than all values in the original array, so immediately return `ConstantArray(true)`.

However, that disregards the validity information. I've updated the branch so that the optimized pathway is only taken if the array and the constant are non-null

Separately I removed the references to lower_parts, it was a lot of cruft and idk when we're going to implement it so just put a note that we're receiving a field for it in the struct and metadata.

### DIFF
--- a/encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
+++ b/encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
@@ -20,10 +20,6 @@ impl CompareKernel for DecimalBytePartsVTable {
         rhs: &dyn Array,
         operator: Operator,
     ) -> VortexResult<Option<ArrayRef>> {
-        // TODO(joe): support compare with lower parts
-        if !lhs.lower_parts.is_empty() {
-            return Ok(None);
-        }
         let Some(rhs_const) = rhs.as_constant() else {
             return Ok(None);
         };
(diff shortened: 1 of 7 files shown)


encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
```diff
@@ -43,12 +39,25 @@ impl CompareKernel for DecimalBytePartsVTable {
...
-                ConstantArray::new(unconvertible_value(sign, operator, nullability), lhs.len())
-                    .to_array(),
-            )),
+
+            Err(sign) => {
+                // If the MSP and the constant are non-null, we know that failing to coerce the
+                // constant into the MSP bit-width means that it is larger/smaller
+                // (depending on the `sign`) than all values in MSP.
+                // If the LHS or the RHS contain nulls, then we must fallback to the canonicalized
+                // implementation which does null-checking instead.
```
robert3005, line 48: I don't understand the must here. Sure the current implementation is not correct but nulls have well defined ordering. This is one of the cases where yu want #2692

a10y: I don't think NULL ordering is a helpful property in this case, but validity array would be helpful, or if we just had the ability to do a ConstantArray with a validity that would also help

robert3005: Indeed the compare result would be null. We could make it sparse but I think it really depends on density. We can revisit in the future

robert3005 (approved):

PR #3791, opened 2025-07-08
fix: make Java test gradle self-contained

### LINKED ISSUE #3775: Gradle tasks do not setup tests themselves
Out of the box running `./gradlew test` fails, which also means running tests from IntelliJ fails.

CI suggests there are a few manual steps required to setup the test environment: https://github.com/vortex-data/vortex/blob/cc0bdbc3a3868dbefd2a5b8c9061055ea10a637e/.github/workflows/ci.yml#L350-L354

Ideally these would be implemented as gradle tasks.

a10y:
Fixes #3775 

The gradle tasks call into the `xtask` command, directly from the `test` task.

### DIFF
--- a/java/vortex-jni/build.gradle.kts
+++ b/java/vortex-jni/build.gradle.kts
@@ -2,7 +2,7 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
-import com.vanniktech.maven.publish.SonatypeHost
+import org.gradle.kotlin.dsl.support.serviceOf
 
 plugins {
     `java-library`
@@ -41,7 +41,7 @@ testing {
 
 mavenPublishing {
     coordinates(groupId = "dev.vortex", artifactId = "vortex-jni", version = "${rootProject.version}")
-    publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL)
+    publishToMavenCentral()
 
     signAllPublications()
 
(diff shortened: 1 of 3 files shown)


java/vortex-jni/build.gradle.kts
```diff
@@ -41,7 +41,7 @@ testing {
 
 mavenPublishing {
     coordinates(groupId = "dev.vortex", artifactId = "vortex-jni", version = "${rootProject.version}")
-    publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL)
```
a10y, line 44: Apparently OSSRH is dead and this is the default now, according to a bunch of deprecation warnings I was getting

java/vortex-jni/build.gradle.kts
```diff
@@ -109,6 +109,45 @@ tasks.build {
...
+            from("${rootProject.projectDir.absoluteFile.parentFile}/target/debug/libvortex_jni.dylib")
+            into("$projectDir/src/main/resources/native/darwin-aarch64")
+        }
+
+        execOps.exec {
+            workingDir = rootProject.projectDir.absoluteFile.parentFile
+            executable = "cargo"
+            args("xtask", "java-test-files")
+        }
+    }
```
a10y, line 144: I was a bit surprised since gradle is usually so declarative, but apparently the ordering here is preserved

robert3005: this is how a plugin body actually looks like

robert3005 (approved):

java/vortex-jni/build.gradle.kts
```diff
@@ -109,6 +109,49 @@ tasks.build {
...
+}
+
+tasks.named("processResources").all {
```
robert3005: this should be configureEach?

java/vortex-jni/build.gradle.kts
a10y: sure, i've only ever written plugins in the Java DSL

Issue #3720, opened 2025-07-02
C++ API


gatesn:
Vortex currently has bindings for Rust, with limited support for Python, Java, and C.

The C++ could go one of two ways:
* Use something like [cxx](https://cxx.rs/) to auto-generate a large C++ API wrapping the Rust types.
* Hand-roll an API that wraps the Vortex C API.

I'm leaning towards the second option, as it gives us more control to expose a canonical C++ API, vs whatever code happens to be generated by cxx.

Assuming we wrap the C API, we should setup a convention for having C++ objects that wrap pointers, and are aware of whether they're owned/borrowed in order to correctly call destructors.

Let's keep this list up to date, but here is my rough understanding of the priority of features:
1. Expose a Vortex file object, and the ability to turn it into an Arrow ArrayStream.
1. Add support to configure the scan with proto serialized filter + projection expressions.
1. Add ability to write Vortex files using an ArrowArrayStream

---

Toolchain for the C++ API:

- Build system: CMake is pretty much the default, so going with CMake makes it easy for other projects to integrate.
- Unit tests: Either [Catch](https://github.com/catchorg/Catch2) or [Google Test](https://github.com/google/googletest). The former is a bit more lightweight and can be integrated as a single header library.
- Linting: [Clang-Tidy](https://clang.llvm.org/extra/clang-tidy/)
- Formatting: [Clang-Format](https://clang.llvm.org/docs/ClangFormat.html)

XinyuZeng:
I am not sure whether async feature is needed for C++ API but just want to add it here as a candidate. There is a [cxx_async](https://github.com/pcwalton/cxx-async) and an example in [opendal](https://github.com/apache/opendal/blob/main/bindings/cpp/tests/async_test.cpp). It requires C++20 though.

XinyuZeng:
After trying out both options, I feel the first option is safer and more ergonomic, e.g,. you can just write safe Rust code and you don't need to write some C++ macros to handroll the RAII logic etc from the C API. However, it may indeed have overlapping code with vortex-ffi. I have a working draft at https://github.com/XinyuZeng/vortex/tree/cxx/vortex-cxx.

XinyuZeng:
In `vortex-ffi`, I was also trying to directly expose ArrowArrayStream by:

```rust
/// Convert a Vortex array iterator into an Arrow array stream.
/// The ownership of the iterator is transferred to the stream.
#[unsafe(no_mangle)]
[...]

PR #3792, opened 2025-07-08
Vortex - C++ API

XinyuZeng:
This PR uses the cxx crate to bridge Rust and C++. The trade-offs between cxx and directly wrapping the vortex-ffi C API is roughly the follows:

Pros of cxx:
1. Safety is guaranteed by cxx, no need to directly use the unsafe C API (not all cases though). In a sense that the Rust-side code is easier to write and maintain.
2. Better ergonomic by providing Rust data structure [bindings](https://cxx.rs/bindings.html) with RAII support in C++ like `rust::Box` and `rust::Str` .

Cons:
1. Less flexibility because we cannot control what FFI is generated, and we may lack some type bindings support across the FFI.
2. May eventually have duplicate code as `vortex-ffi`

I also tried to find best practices in other repos and seems [opendal](https://github.com/apache/opendal/tree/main/bindings) uses cxx for C++ and cbindgen for C with separate code paths, too. I think we can follow the same approach at least for now for a limited C++ API (e.g., to/from Arrow). We can refactor if we need direct-access to C FFI in the future. 

*The first PR intends to setup a C++ directory structure and convention instead of fixing to use cxx.*

cc @royguo


vortex-cxx/include/vortex.hpp
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  explicit VortexFile(rust::Box<ffi::VortexFile> impl)
+      : impl_(std::move(impl)) {}
+
+  uint64_t row_count() const { return ffi::file_row_count(*impl_); }
+
+  std::pair<ArrowArray, ArrowSchema> scan_to_arrow() const;
+
+  arrow::Result<std::shared_ptr<arrow::RecordBatchReader>> scan_to_stream() const;
```
XinyuZeng: Returning native Arrow C++ data structure requires adding Arrow as a dependency, perhaps we can just return the Arrow C Data Interface

0ax1:
@XinyuZeng Thanks for the draft! I'll get around having a closer look tomorrow.

vortex-cxx/.clang-format
```diff
@@ -0,0 +1,37 @@
...
+SpacesInConditionalStatement: false
+AllowShortLambdasOnASingleLine: Inline
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakTemplateDeclarations: Yes
+NamespaceIndentation: Inner
+IncludeBlocks: Regroup
+Language: Cpp
+AccessModifierOffset: -4
+---
+Language: Java
```
0ax1: Let's drop the Java section here.

vortex-cxx/.clang-tidy
```diff
@@ -0,0 +1,53 @@
...
+
+Checks:          '-*,clang-diagnostic-*,bugprone-*,performance-*,google-explicit-constructor,google-build-using-namespace,google-runtime-int,misc-definitions-in-headers,modernize-use-nullptr,modernize-use-override,-bugprone-macro-parentheses,readability-braces-around-statements,-bugprone-branch-clone,readability-identifier-naming,hicpp-exception-baseclass,misc-throw-by-value-catch-by-reference,-bugprone-signed-char-misuse,-bugprone-misplaced-widening-cast,-bugprone-sizeof-expression,-bugprone-easily-swappable-parameters,google-global-names-in-headers,llvm-header-guard,misc-definitions-in-headers,modernize-use-emplace,modernize-use-bool-literals,-performance-inefficient-string-concatenation,-performance-no-int-to-ptr,readability-container-size-empty,cppcoreguidelines-pro-type-cstyle-cast,-llvm-header-guard,-performance-enum-size,cppcoreguidelines-pro-type-const-cast,cppcoreguidelines-interfaces-global-init,cppcoreguidelines-slicing,cppcoreguidelines-rvalue-reference-param-not-moved,cppcoreguidelines-virtual-class-destructor,-readability-identifier-naming,-bugprone-exception-escape,-bugprone-unused-local-non-trivial-variable,-bugprone-empty-catch'
```
0ax1: Reformat this such that we have one check per line:
```
Checks: >
  -*,
  clang-diagnostic-*,
  bugprone-*,
  performance-*,
  google-explicit-constructor,
  google-build-using-namespace,
  google-runtime-int,
  google-global-names-in-headers,
  misc-definitions-in-headers,
...
  ```

File: vortex-cxx/.clang-tidy [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+WarningsAsErrors: '*'
+HeaderFilterRegex: 'include/.*'
+FormatStyle:     none
+CheckOptions:
```
0ax1, line 48: This is a nice addition. What clang-tidy config are these based on?

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
+# SPDX-License-Identifier: Apache-2.0
```
0ax1, line 1: We need to add a cxx-build job here to CI: https://github.com/vortex-data/vortex/blob/develop/.github/workflows/ci.yml#L23. Note that the CI machines have a CXX toolchain with Clang preinstalled. As a start, we should build though CMake in debug and release, and run the unit tests.

0ax1: This can be along the lines of how we build: 
```
  build-rust:
    name: "Rust build (${{matrix.config.name}})"
    timeout-minutes: 120
    runs-on:
      - runs-on=${{ github.run_id }}
      - family=m7i+m7i-flex+m7a
      - cpu=8
      - image=ubuntu24-full-x64
      - extras=s3-cache
```

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+cmake_minimum_required(VERSION 3.16)
+
+if(DEFINED ENV{VCPKG_TOOLCHAIN_PATH})
```
0ax1: I'd like the CXX build to be more self-contained without relying on additional tools. We should try using `FetchContent_Declare` & `FetchContent_MakeAvailable` instead of relying on vcpkg.

I picture this something like: 
```
set(ARROW_BUILD_WARNINGS OFF)
set(ARROW_BUILD_STATIC ON)
set(ARROW_SIMD_LEVEL NONE)

FetchContent_Declare(Arrow
  GIT_REPOSITORY https://github.com/apache/arrow.git
  GIT_TAG apache-arrow-20.0.0
  GIT_SHALLOW TRUE SOURCE_SUBDIR cpp
)

FetchContent_Declare(GoogleTest
  GIT_REPOSITORY https://github.com/google/googletest.git
  GIT_TAG        v1.17.0
  GIT_SHALLOW    TRUE
)

set(CMAKE_CXX_FLAGS_BACKUP ${CMAKE_CXX_FLAGS})
set(CMAKE_CXX_FLAGS "-w")
FetchContent_MakeAvailable(GoogleTest)
FetchContent_MakeAvailable(Arrow)
set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS_BACKUP})
```

Either way we need to explicitly pin the library versions we link against.

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+cmake_minimum_required(VERSION 3.16)
```
0ax1: Let's align this with what we use in vortex-duckdb => `3.22`.

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+
+if(DEFINED ENV{VCPKG_TOOLCHAIN_PATH})
+    set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_TOOLCHAIN_PATH}")
+else()
```
0ax1: In general, the project structure should roughly align with vortex-duckdb for consistency. So a `cpp` dir with source files with all includes in `include`. The cxx tests should also move to the `cpp` dir.

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+    message(WARNING "VCPKG_TOOLCHAIN_PATH environment variable is not set. Dependencies may not be found correctly. See README.md for vcpkg setup instructions.")
+endif()
+
+project(vortex-cxx LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+if (NOT CMAKE_BUILD_TYPE)
```
0ax1, line 11: This is nice! (And wild that CMake does not default to `Debug`)

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+                    ${Arrow_INCLUDE_DIRS})
+
+file(GLOB CPP_SOURCE_FILE "src/*.cpp")
+file(GLOB CPP_HEADER_FILE "include/*.hpp" "src/*.hpp")
+
+if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+    list(APPEND CARGO_BUILD_FLAGS "--release")
+endif()
+
+add_custom_target(cargo_build
```
0ax1: I'd suggest offloading the CMake to Cargo logic with https://github.com/corrosion-rs/corrosion, like we do in vortex-duckdb.

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+target_include_directories(vortex_cxx PUBLIC ${CPP_INCLUDE_DIR})
+target_link_libraries(vortex_cxx PUBLIC ${RUST_LIB})
+target_link_libraries(vortex_cxx PRIVATE ${CMAKE_DL_LIBS} Arrow::arrow_static)
+set_target_properties(vortex_cxx
+    PROPERTIES ADDITIONAL_CLEAN_FILES ${CARGO_TARGET_DIR}
+)
+add_dependencies(vortex_cxx cargo_build)
+
+# Platform-specific configuration
+if(WIN32)
```
0ax1: We neither build or test against Windows as of now => Drop

vortex-cxx/tests/basic_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+        ASSERT_EQ(int32_array_b->Value(0), 10);
+        ASSERT_EQ(int32_array_b->Value(1), 20);
+        ASSERT_EQ(int32_array_b->Value(2), 30);
+        ASSERT_EQ(int32_array_b->Value(3), 40);
+        ASSERT_EQ(int32_array_b->Value(4), 50);
+    }
+};
+
+TEST_F(VortexTest, ScanToArray) {
+    auto file = vortex::VortexFile::Open("../target/debug/build/test_data.vortex");
```
0ax1: It'd be nice to make less assumptions about where the test file lives. 

How about we move `generate_test_vortex_file();` to `lib.rs`, drop it from `build.rs`, call `generate_test_vortex_file` via FFI in a unit test fixture and put the test file in the same dir as the test binary?

0ax1 (commented):
Thanks again for taking a stab at this!

I had an initial pass on the code having a look at the build system setup which I think is probably the most tricky part to get clean. Once we cleared that part we can move on to other aspects.

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+    USES_TERMINAL
+    COMMENT "Running cargo build for vortex-cxx..."
+)
+
+add_library(vortex_cxx STATIC ${CPP_SOURCE_FILE} ${RUST_BRIDGE_CPP})
+target_sources(vortex_cxx PUBLIC ${CPP_HEADER_FILE})
+target_sources(vortex_cxx PRIVATE ${RUST_HEADER_FILE})
+target_include_directories(vortex_cxx PUBLIC ${CPP_INCLUDE_DIR})
+target_link_libraries(vortex_cxx PUBLIC ${RUST_LIB})
+target_link_libraries(vortex_cxx PRIVATE ${CMAKE_DL_LIBS} Arrow::arrow_static)
```
0ax1: What is `CMAKE_DL_LIBS` used for? Do we currently use or expect to use `dlopen` & `dlclose`?

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+project(vortex-cxx LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+if (NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Debug)
+endif()
+
+option(VORTEX_ENABLE_TESTING "Enable building test binary for vortex-cxx" OFF)
```
0ax1, line 17: We should enable `-Wall -Wextra -Wpedantic` for the CXX build.

vortex-cxx/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+
+# Get cargo target dir
```
0ax1: Set `DNDEBUG` for release builds: `set(CMAKE_CXX_FLAGS_RELEASE "-DNDEBUG")`.

vortex-cxx/.clang-tidy
XinyuZeng: Both the clang-tidy and clang-format are directly borrowed from DuckDB. I have a section in README to explain this.

XinyuZeng:
Thanks for the review!

> Once we cleared that part we can move on to other aspects.

Yep I would also like to first setup the build system, framework, and convention etc. before moving on other parts. Instead of a giant PR I'd like to break it into smaller pieces. I will address the comments tomorrow in my time.

The initial setup mimics the ones in [opendal](https://github.com/apache/opendal/blob/main/bindings/cpp/CMakeLists.txt) and [cxx-cmake-example](https://github.com/XiangpengHao/cxx-cmake-example), just to ramp up. I will checkout vortex-duckdb to make them consistent.

vortex-cxx/CMakeLists.txt
0ax1: Ignore, turns out CMake sets this by default for release.

vortex-cxx/CMakeLists.txt
0ax1: Alternatively we can also instruct vcpkg from CMake. I think the goal here is that everything should only require a single CMake call without manual upfront setup.

vortex-cxx/CMakeLists.txt
XinyuZeng: removed

vortex-cxx/CMakeLists.txt
XinyuZeng: Added with an extra `-Wno-dollar-in-identifier-extension` to prevent warnings from cxx generated code.

vortex-cxx/CMakeLists.txt
XinyuZeng: I now build the Arrow C++ from scratch. Will your CI cache the build output? It turns out the C++ CI becomes the slowest one now, especially building both the debug and release.

vortex-cxx/CMakeLists.txt
XinyuZeng: The reviews about build setup are all addressed

vortex-cxx/CMakeLists.txt
gatesn: Can we not take a dependency on Arrow at all? As in, we can just export to a C ArrowArrayStream so the consumers of the library can do with it as they please

XinyuZeng: Yes the api only exposes Arrow C Data now. Arrow C++ is used for unit tests only. Let me see if we can get rid of that too.

gatesn: If the implementation is largely in rust, then I might just ignore the testing on C++ side and assume arrow-rs export is correct

XinyuZeng: I removed arrow-cpp and only use nanoarrow which is more lightweight.

vortex-cxx/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+    input_stream: *mut u8,
+    path: &str,
+) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let stream_reader =
+        ArrowArrayStreamReader::from_raw(input_stream as *mut FFI_ArrowArrayStream)?;
+
+    let rt = get_runtime();
+    let path = path.to_string();
+
+    // TODO: this actually materializes all the data into memory, should expose async write API to C++, see vortex-ffi/src/sink.rs
```
XinyuZeng: The problem here is that if the input_stream comes from the C++ API,  then the `write_array_stream` and `scan_builder_into_stream` share the same runtime, which leads to nested `block_on` calls and panics. There are two options here: 1. do the similar thing as `vortex-ffi/src/sink.rs` by exposing an api to push one array at a time to sink instead of passing an arraystream 2. we can have another `static RUNTIME` for writing. Do you have any suggestions here? @0ax1

vortex-cxx/cpp/file.cpp
```diff
@@ -0,0 +1,113 @@
...
+ScanBuilder &ScanBuilder::operator=(ScanBuilder &&other) noexcept {
+    if (this != &other) {
+        impl_ = std::move(other.impl_);
+    }
+    return *this;
+}
+
+ScanBuilder::~ScanBuilder() = default;
+
+ScanBuilder &ScanBuilder::SetFilter(std::string_view filter) {
```
0ax1: I guess we're going for something along the lines of a builder pattern here as we return `ScanBuilder &`. How about renaming this to `WithFilter` and `WithLimit`?

vortex-cxx/CMakeLists.txt
0ax1: Great, thanks for iterating on this!

vortex-cxx/src/lib.rs
XinyuZeng: <details>

<summary>Actually this happens the same in vortex-python:</summary>

```python
In [1]: import vortex as vx

In [2]: a = vx.array([
   ...: 
   ...:     {'name': 'Joseph', 'age': 25},
   ...: 
   ...:     {'name': None, 'age': 31},
   ...: 
   ...:     {'name': 'Angela', 'age': None},
   ...: 
   ...:     {'name': 'Mikhail', 'age': 57},
   ...: 
   ...:     {'name': None, 'age': None},
   ...: 
   ...: ])

In [4]: import vortex.io as vortex_io

In [5]: vx.io.write(a, "a.vortex")

In [6]: vxf = vx.open("a.vortex")

In [7]: stream=vxf.scan()

In [8]: vx.io.write(stream, "b.vortex")

thread '<unnamed>' panicked at vortex-python/src/iter/stream.rs:17:14:
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
---------------------------------------------------------------------------
PanicException                            Traceback (most recent call last)
Cell In[8], line 1
----> 1 vx.io.write(stream, "b.vortex")

PanicException: Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```
</details>

gatesn: Could you file an issue for this? Need to figure out the right model here. We probably need sync APIs that "do the right thing"

vortex-cxx/.clang-format
joseph-isaacs, line 1: Shall we merge this clang format with `vortex/vortex-duckdb/.clang-format` putting them in the top level?

vortex-cxx/cpp/file.cpp
joseph-isaacs: Can you either remove the `WithFilter` or add a test (likely using proto generation)

vortex-cxx/cpp/file.cpp
```diff
@@ -0,0 +1,113 @@
...
+        ffi::scan_builder_set_filter(
+            *impl_->rust_impl,
+            rust::Slice<const uint8_t>(reinterpret_cast<const uint8_t *>(filter.data()), filter.size()));
+    } catch (const rust::cxxbridge1::Error &e) {
+        throw VortexException(e.what());
+    }
+    return *this;
+}
+
+ScanBuilder &ScanBuilder::SetLimit(uint64_t limit) {
```
joseph-isaacs: `WithLimit` please

vortex-cxx/cpp/include/vortex/file.hpp
```diff
@@ -0,0 +1,73 @@
...
+    std::pair<ArrowArray, ArrowSchema> IntoArray();
+
+private:
+    friend class VortexFile;
+
+    struct Impl;
+    explicit ScanBuilder(std::unique_ptr<Impl> impl);
+
+    std::unique_ptr<Impl> impl_;
+};
```
joseph-isaacs: can this live in a different file `scan`?

vortex-cxx/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+    let mut arrays: Vec<ArrayRef> = Vec::new();
+    let schema = stream_reader.schema();
+    let dtype = DType::from_arrow(schema);
+
+    // Convert Arrow RecordBatches to Vortex Arrays synchronously
+    for batch_result in stream_reader {
+        let batch = batch_result?;
+        let vortex_array = ArrayRef::from_arrow(batch, false);
+        arrays.push(vortex_array);
+    }
```
joseph-isaacs: This will buffer a lot of data in memory, can convert the stream  into `SendableArrayStream = Pin<Box<dyn ArrayStream + Send>>` and write that to the file.

vortex-cxx/src/lib.rs
```diff
@@ -0,0 +1,339 @@
...
+/// https://github.com/dora-rs/dora/blob/42775e3612b34d3998b1f7feb7e5df7ec3f8a7bd/apis/c%2B%2B/node/src/lib.rs#L211-L212
+///
+/// # Safety
+///
+/// out_array should be properly aligned according to C ABI and valid for write.
+/// out_schema should be properly aligned according to C ABI and valid for write.
+unsafe fn scan_builder_into_arrow(
+    builder: Box<VortexScanBuilder>,
+    out_array: *mut u8,
+    out_schema: *mut u8,
```
joseph-isaacs, line 161: are these really `*mut u8` or `*mut c_void`?

joseph-isaacs (commented):
Looking very close, just a few more things

vortex-cxx/cpp/file.cpp
XinyuZeng: I removed `WithFilter` and add `WithRowRange`. I'd like to have the filter and projection tests with proto generation to be a seprate PR to not make this one too big.

vortex-cxx/.clang-format
XinyuZeng: I used symbolic link now. Feeling weird to have .clang-format at top level for a rust project.

vortex-cxx/src/lib.rs
XinyuZeng: Unfortunately `c_void` is not an allowed type for cxx. Do they make a difference in terms of performance and safety?

vortex-cxx/src/lib.rs
XinyuZeng: I tried converting the stream to SendableArrayStream before https://github.com/vortex-data/vortex/pull/3792/commits/8e3965f7ab957be60a7bd73f501177781b6eab21#diff-7185d60946e047abb9c984a9f8984aa4cb37983ac3d5d2528ef03909c3dfba49R158-R169

But it has the same problem as https://github.com/vortex-data/vortex/issues/3877

vortex-cxx/cpp/include/vortex/file.hpp
XinyuZeng: Done. I also reorg the directory so that `include/` is public headers and `src/` is for private headers and source files.

vortex-cxx/src/lib.rs
joseph-isaacs: Can we not use `out_array: *mut FFI_ArrowArray` or due cxx not understand that?

vortex-cxx/src/lib.rs
joseph-isaacs: We can at least use a model (writing via a channel) similar to hear: https://github.com/vortex-data/vortex/blob/6271c6e54371c650183a9d4f61f34a000651a863/vortex-ffi/src/sink.rs#L71 which means that the whole array is not buffered in memory

vortex-cxx/cpp/file.cpp
joseph-isaacs: Did you push these changes?

vortex-cxx/cpp/include/vortex/file.hpp
joseph-isaacs: Great, but not sure its in this PR

vortex-cxx/cpp/file.cpp
XinyuZeng: Yes. I pushed the changes yesterday. Maybe the page is not refreshed?

vortex-cxx/cpp/include/vortex/file.hpp
XinyuZeng: It is in. Commit here: https://github.com/vortex-data/vortex/pull/3792/commits/9482b526f27f2190cbdeb64861a6fbc8c0f614b7

vortex-cxx/.clang-format
joseph-isaacs: We can move it there at a later date

vortex-cxx/src/lib.rs
XinyuZeng: @gatesn shall we wait for the threading model mentioned in https://github.com/vortex-data/vortex/issues/3890 to be cleaned up before merging this PR?

vortex-cxx/src/lib.rs
XinyuZeng: cxx does not understand it. You have to define another struct with the same layout as `FFI_ArrowArray`, and then it can be used as function parameters.
``` rust
#[cxx::bridge(namespace = "vortex::ffi")]
mod ffi {
    // C-compatible structs for Arrow C ABI
    struct CArrowArray {
        length: i64,
        null_count: i64,
        offset: i64,
        n_buffers: i64,
        n_children: i64,
        buffers: usize,
        children: usize,
        dictionary: usize,
        release: usize,
        private_data: usize,
    }
//...
    extern "Rust" {
    fn file_scan_to_array(file: &VortexFile) -> Result<CArrowArray>;
    // ...
    }
}
```
Then in the C++ side we have to reinterpret_cast the output to Arrow's C ArrowArray.  I've done that before but switched to the current approach because it has fewer conversions across different naming of the same ArrowArray structure.

vortex-cxx/src/lib.rs
joseph-isaacs: We are currently working on this, so might be best to wait

gatesn: You (possibly?) want to use a thread pool, or else just a single threaded iterator, e.g. Python now does this: https://github.com/vortex-data/vortex/pull/3899/files#diff-6d8c26c26b4bfda9d821b89b05fe83dd1168c47fe5b0757c51853594ce1445f1R34

I still need to follow up to fix the writer APIs, but that shouldn't block you I don't think...

XinyuZeng: I see. Thanks! Will check it out.

XinyuZeng: Ive fixed it. Would you mind take a look again?

joseph-isaacs:
Mind merging in main and we can run full CI and then merge

joseph-isaacs (approved):

joseph-isaacs:
@XinyuZeng Thanks for very for the contribution 🚀

joseph-isaacs:
Please could you write somewhere what the follow ups from the PR are?

XinyuZeng:
> Please could you write somewhere what the follow ups from the PR are?

I added follow-ups here https://github.com/vortex-data/vortex/issues/3720#issuecomment-3089550044

0ax1:
@XinyuZeng Great work!!

PR #3795, opened 2025-07-08
Add expression transformation to replace all occurrences of a node

gatesn:


vortex-expr/src/exprs/pack.rs
```diff
@@ -199,7 +199,7 @@ impl Display for PackExpr {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "pack({{{}}}){}",
```
robert3005, line 202: not a fan of curly braces? :P

robert3005 (approved):

vortex-expr/src/transform/replace.rs
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_error::{VortexExpect, VortexResult};
+
+use crate::traversal::{MutNodeVisitor, Node, TransformResult, TraversalOrder};
+use crate::{ExprRef, VortexExpr};
+
+/// Replaces all occurrences of `needle` in the expression `expr` with `replacement`.
```
joseph-isaacs, line 9: This is not technically true, it will only remove the least deep node

gatesn: But no child of the needle expression can also be equal to the needle

PR #3797, opened 2025-07-08
fix: Remove redundant copy from delta compression

robert3005:

### DIFF
--- a/encodings/fastlanes/src/delta/compress.rs
+++ b/encodings/fastlanes/src/delta/compress.rs
@@ -141,7 +141,7 @@ where
             // Initialize the base vector for this chunk
             Delta::undelta(
                 chunk,
-                unsafe { &*(transposed[0..T::LANES].as_ptr() as *const [_; T::LANES]) },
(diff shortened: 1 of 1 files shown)


encodings/fastlanes/src/delta/compress.rs
```diff
@@ -57,25 +53,22 @@ where
...
+            bases.extend(&transposed[0..T::LANES]);
 
             deltas.reserve(1024);
             let delta_len = deltas.len();
             unsafe {
                 deltas.set_len(delta_len + 1024);
                 Delta::delta(
                     &transposed,
-                    &base,
+                    &*(transposed[0..T::LANES].as_ptr() as *const [_; T::LANES]),
```
robert3005, line 71: this is what `array_ref` macro does but because it generates functions the `T::LANES` const is not accessible

joseph-isaacs: can this be done with a transmute?

robert3005: You cannot since the compiler things the types are of different widths

joseph-isaacs (approved):

PR #3801, opened 2025-07-08
ci: run Rust tests with ASAN

a10y:
Not sure if this works. It fails locally but might be a mac problem.

Nice for catching things like the bug introduced with AVX2 change last week (and Miri can't execute SIMD kernels so this is our only way to catch potentially safety issues)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,7 +319,7 @@ jobs:
       - name: Rust Tests
         # vortex-duckdb-ext currently fails linking for cargo test targets.
         run: |
-          cargo nextest run --locked --workspace --all-features --no-fail-fast
+          cargo nextest run --locked --workspace --all-features --no-fail-fast --target x86_64-unknown-linux-gnu
 
       - name: Generate coverage report
         if: github.ref == 'refs/heads/develop'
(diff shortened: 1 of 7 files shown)


robert3005:
This is a mac bug, the linker doesn't seem to like asan but no clue what's the fix

a10y:
Nice, so without slowing the job down, when I artificially reintroduce the AVX2 bug it is caught by the ASAN run, along with some other DuckDB test cleanup issues that should be easy enough to fix.

vortex-duckdb/src/duckdb/logical_type.rs
```diff
@@ -42,6 +42,30 @@ impl LogicalType {
             )
         }
     }
+
+    pub fn array_child_type(&self) -> Self {
```
0ax1, line 46: ❤️

.github/workflows/ci.yml
```diff
@@ -294,7 +294,7 @@ jobs:
         shell: bash
         if: github.ref == 'refs/heads/develop'
         run: |
-          echo "RUSTFLAGS=-Cinstrument-coverage -A warnings">> $GITHUB_ENV
```
0ax1, line 297: We should `apt-get install llvm` to resolve addresses to source locations. libfuzzer relies on https://llvm.org/docs/CommandGuide/llvm-symbolizer.html to do that.

robert3005 (approved):

PR #3802, opened 2025-07-09
chore[bench]: add tpchgen-rs benchmark generation

joseph-isaacs:
hopefully this will reduce memory usage


bench-vortex/src/tpch/tpchgen.rs
```diff
@@ -0,0 +1,356 @@
...
+use vortex::stream::ArrayStreamAdapter;
+
+use crate::utils::file_utils::idempotent_async;
+use crate::{Format, IdempotentPath};
+
+/// Configuration for TPC-H data generation
+#[derive(Debug, Clone)]
+pub struct TpchGenOptions {
+    /// Scale factor (1, 10, 100, 1000)
+    pub scale_factor: f64,
```
0ax1, line 37: Are decimal scale factors allowed or should this be i64?

bench-vortex/src/tpch/tpchgen.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Output directory
+    pub output_dir: PathBuf,
+    /// Output format
+    pub format: Format,
+    /// Batch size for streaming
+    pub batch_size: usize,
+    /// Number of partitions for parallel generation
+    pub num_partitions: usize,
```
0ax1: Given that we always cast `num_partitions` and `partition` to i32 should this be i32 in the first place?

bench-vortex/src/tpch/tpchgen.rs
```diff
@@ -0,0 +1,356 @@
...
+            let output_dir = options.output_dir.join(Format::OnDiskVortex.to_string());
+            fs::create_dir_all(&output_dir)?;
+            (output_dir.join(format!("{}.vortex", table_name)), "Vortex")
+        }
+        f @ Format::Csv => {
+            anyhow::bail!("{f} format is not supported by tpchgen");
+        }
+    };
+
+    let format = options.format;
```
0ax1: Voting for just matching on `options.format` like above, rather than introducing a new variable.

0ax1 (commented):
Couple of thoughts inline. Looks good.

bench-vortex/src/tpch/tpchgen.rs
joseph-isaacs: yes

0ax1 (approved):
Happy if CI is happy.

joseph-isaacs:
running generation and we will see

vortex-array/src/arrow/convert.rs
```diff
@@ -434,6 +435,18 @@ impl FromArrowArray<&dyn ArrowArray> for ArrayRef {
...
+    fn from_arrow(array: RecordBatch, nullable: bool) -> Self {
+        ArrayRef::from_arrow(&arrow_array::StructArray::from(array), nullable)
+    }
+}
+
+impl FromArrowArray<&RecordBatch> for ArrayRef {
+    fn from_arrow(array: &RecordBatch, nullable: bool) -> Self {
+        ArrayRef::from_arrow(&arrow_array::StructArray::from(array.clone()), nullable)
+    }
+}
```
robert3005: We also have a `TryIntoArray` trait which whole purpose is to implement this. Can you delete & replace usages while you're doing this?

robert3005: I have done this separately in #3804

PR #3803, opened 2025-07-09
Generalize expression partition to annotationed expressions

gatesn:
Previously, expression partitioning only worked for fields accessed on the root scope. This change allows an expression to be partitioned by an arbitrary set of annotations.

This means custom expressions can be grouped into a partition and pulled out separately from the rest.


vortex-expr/src/transform/annotations.rs
```diff
@@ -0,0 +1,86 @@
...
+        let annotations = (self.annotate)(node);
+        if annotations.is_empty() {
+            // If the annotate fn returns empty, we do not annotate this node.
+            Ok(TraversalOrder::Continue)
+        } else {
+            self.annotations
+                .entry(node)
+                .or_default()
+                .extend(annotations);
+            Ok(TraversalOrder::Skip)
```
joseph-isaacs, line 77: I find this unexpected from the name of the visitor

joseph-isaacs: I would expect annotate to either continue down or return a  TraversalOrder here

robert3005 (dismissed):

vortex-expr/src/transform/annotations.rs
robert3005: This is refactoring of the previous logic where we would always continue on empty annotations and skip immediately when we had one present

joseph-isaacs: That is not true, this has been generalised

gatesn: Yeah it's perhaps a little odd to skip at the first sign of an annotation..

robert3005 (approved):

PR #3804, opened 2025-07-09
chore: Remove TryIntoArray trait that was only used to convert RecordBatch into Vortex Array

robert3005:


AdamGS (approved):
🙏

PR #3805, opened 2025-07-09
ci: update target dir for coverage runs

a10y:
The `-C instrument-coverage` flag doesn't play well with ASAN, because both instruments require hooking into shutdown so one ends up overriding the other.


robert3005 (approved):
I assume this fixes CI

a10y:
Yea I'm just verifying now locally that this actually is the fix

PR #3810, opened 2025-07-09
WIP feat: add ViewLayout for variable length views

a10y:
- [x] VTable
- [x] Reader skeleton
- [x] Pruning implementations
- [x] FilterEval
- [x] ProjectionEval
- [x] Writer / Strategy
- [ ] Potentially replacing DictLayout?
- [ ] Benchmarking

### DIFF
--- a/vortex-array/src/arrays/varbinview/mod.rs
+++ b/vortex-array/src/arrays/varbinview/mod.rs
@@ -52,14 +52,14 @@ impl Inlined {
 
 #[derive(Clone, Copy, Debug)]
 #[repr(C, align(8))]
-pub struct Ref {
+pub struct Outlined {
     size: u32,
     prefix: [u8; 4],
     buffer_index: u32,
     offset: u32,
 }
 
-impl Ref {
+impl Outlined {
     pub fn new(size: u32, prefix: [u8; 4], buffer_index: u32, offset: u32) -> Self {
         Self {
             size,
@@ -101,12 +101,12 @@ pub union BinaryView {
     inlined: Inlined,
 
     // Reference type: strings > 12 bytes.
-    _ref: Ref,
+    outlined: Outlined,
 }
 
 assert_eq_size!(BinaryView, [u8; 16]);
 assert_eq_size!(Inlined, [u8; 16]);
-assert_eq_size!(Ref, [u8; 16]);
+assert_eq_size!(Outlined, [u8; 16]);
 assert_eq_align!(BinaryView, u128);
 
 impl BinaryView {
@@ -120,7 +120,7 @@ impl BinaryView {
     /// Adapted from arrow-rs <https://github.com/apache/arrow-rs/blob/f4fde769ab6e1a9b75f890b7f8b47bc22800830b/arrow-array/src/builder/generic_bytes_view_builder.rs#L524>
     /// Explicitly enumerating inlined view produces code that avoids calling generic `ptr::copy_non_interleave` that's slower than explicit stores
     #[inline(never)]
-    pub fn make_view(value: &[u8], block: u32, offset: u32) -> Self {
+    pub fn new_view(value: &[u8], block: u32, offset: u32) -> Self {
         match value.len() {
             0 => Self {
                 inlined: Inlined::new::<0>(value),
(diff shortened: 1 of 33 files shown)


vortex-array/src/arrays/varbinview/mod.rs
```diff
@@ -101,12 +101,12 @@ pub union BinaryView {
     inlined: Inlined,
 
     // Reference type: strings > 12 bytes.
-    _ref: Ref,
+    outlined: Outlined,
```
a10y, line 104: Thought this was a better name as an analogue to `Inlined`

vortex-layout/src/layouts/view/reader/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+type BinaryViewFuture = Shared<BoxFuture<'static, SharedVortexResult<Buffer<BinaryView>>>>;
+type BufferFuture = Shared<BoxFuture<'static, SharedVortexResult<ByteBuffer>>>;
+
+/// Handle for fetching a buffer by index.
+#[derive(Clone)]
+pub(crate) struct FetchBuffers {
+    name: Arc<str>,
+    segment_source: Arc<dyn SegmentSource>,
+    buffer_ids: Arc<[SegmentId]>,
+    buffers: Vec<OnceLock<BufferFuture>>,
```
gatesn: I _think_ (given how horrible the implicit prefetching is...) you might want to just create all these futures up-front.

Creation causes registration for prefetching.

Only when you first _poll_ them does the segment get marked as required.

By deferring here with a segment source, you're precluding these buffers from ever being prefetched.

gatesn (commented):
Small note, we will probably have a ListViewLayout at some point.

Don't know if that changes your mind on the name!

vortex-layout/src/layouts/view/reader/mod.rs
a10y: Yea I was thinking about that, but I thought part of the goal here is that if we can prune using the views some of the buffers may not need to be fetched _at all_. Maybe it's not worth it to lose the prefetching though

robert3005: I think what Nick means you can register but never poll them and that will give you a slightly better behaviour instead of registering them later

gatesn: These are the kind of behaviour bugs I really hate... 

Luckily, Rust lets us solve them at compile time!

We should pass in &dyn SegmentSource instead of Arc<dyn SegmentSource>, because the resulting evaluation needs to have a static lifetime, it would be impossible for someone to delay getting a segment future.

a10y:
ListView seems like a less useful thing to break off into a Layout, because there's no pruning opportunity?

robert3005:
For listview you have the same pruning opportunity as for string view. You have a mask from somewhere else and are able to fetch only child buffers that are referenced by the mask

gatesn:
Yeah, you're never probably filtering on a ListView (unless you're also doing a full scan? E.g. vectors as fixed length lists)

But if you had an existing filter mask, the projection could benefit a lot for large lists.

a10y:
ListView doesn't have multiple buffers https://arrow.apache.org/docs/format/Columnar.html#listview-layout

robert3005:
You’re correct, I completely missed that. We will need to come up with a layout that supports lazy loading of list values.

gatesn:
Yah sorry, I had in my head for lists we'd need our own canonical "ListView" format since Arrow's isn't great..

vortex-layout/src/layouts/view/reader/mod.rs
gatesn: Here's an example of what I mean. Not sure if it's the right thing to do though... https://github.com/vortex-data/vortex/pull/3858

vortex-layout/src/layouts/view/reader/mod.rs
```diff
@@ -0,0 +1,524 @@
...
+
+        let unsupported4 = LikeExpr::new_expr(col("name"), lit("%"), false, false); // only %
+        let result = super::ViewReader::analyze_for_view_pruning(&unsupported4);
+        assert!(result.is_none());
+
+        let unsupported5 = LikeExpr::new_expr(col("name"), lit("ABC%"), false, true); // case-insensitive
+        let result = super::ViewReader::analyze_for_view_pruning(&unsupported5);
+        assert!(result.is_none());
+
+        let unsupported6 = LikeExpr::new_expr(col("name"), lit("ABC%"), true, false); // NOT LIKE
```
a10y, line 468: should implement pruning support for NOT LIKE in a flup

PR #3814, opened 2025-07-10
fix[bench]: update action and website

joseph-isaacs:


.github/workflows/sql-benchmarks.yml
```diff
@@ -43,10 +43,6 @@ on:
 jobs:
   bench:
     timeout-minutes: 120
-    # S3 is shared state here, and we want to make sure only one of each job runs at a time
```
0ax1, line 46: Why did this get dropped?

.github/workflows/nightly-bench.yml
```diff
@@ -44,6 +44,8 @@ jobs:
           }
         ]
     strategy:
+      # A single run should kill the others
```
0ax1: A single failed task?

0ax1 (approved):
One question inline but feel free to land.

.github/workflows/sql-benchmarks.yml
joseph-isaacs: Its not true any more

Issue #3609, opened 2025-06-23
Fix regression caused by scan execution logic refactor in #3489


robert3005:
There's a small regression in #3489 for highly selective scans

AdamGS:
Seems like https://github.com/vortex-data/vortex/pull/3815 fixed it, or at least got a significant part of the perf back

PR #3815, opened 2025-07-10
fix: Propagate executor to scan tasks

AdamGS:


vortex-layout/src/scan/mod.rs
```diff
@@ -287,7 +287,8 @@ impl ScanBuilder<ArrayRef> {
     ///
     /// All work will be performed on the current thread, with tasks interleaved per the
     /// configured concurrency. Any configured executor will be ignored.
-    pub fn into_array_iter(self) -> VortexResult<impl ArrayIterator + 'static> {
+    pub fn into_array_iter(mut self) -> VortexResult<impl ArrayIterator + 'static> {
+        self.executor = None;
```
gatesn, line 293: Maybe a comment to explain why?

AdamGS: added something

gatesn (approved):

robert3005:
I guess this is the fix for https://github.com/vortex-data/vortex/issues/3609?

vortex-layout/src/scan/mod.rs
```diff
@@ -217,7 +217,7 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
                     reader: layout_reader.clone(),
                     projection: projection.clone(),
                     mapper: self.map_fn.clone(),
-                    task_executor: None,
+                    task_executor: self.executor.clone(),
```
a10y, line 220: 👍

PR #3819, opened 2025-07-10
Row idx layout reader

gatesn:
Fixes #3613 

* Implements a RowIdx layout reader without using scope vars (in preparation of removing them).
* Works the same way, by partitioning the expression into the row index and non-row index parts.
* Abstracts out partitioned evaluations (shared with struct layouts that partition by field)
* Introduces a RowIdx expression ensuring identity doesn't resolve to include this meta column.


vortex-layout/src/scan/mod.rs
```diff
@@ -183,12 +182,14 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
         // conjunction splitting if a filter is provided.
         let mut layout_reader = self.layout_reader;
 
+        // Enrich the layout reader to support RowIdx expressions.
+        // Note that this is applied below the filter layout reader since it can perform
```
a10y, line 186: This is also for correctness right? Doing RowIdx above the filter node would assign the wrong row indices

vortex-layout/src/scan/mod.rs
```diff
@@ -60,10 +60,9 @@ pub struct ScanBuilder<A> {
     file_stats: Option<Arc<[StatsSet]>>,
     /// Maximal number of rows to read (after filtering)
     limit: Option<usize>,
-    /// Include the row and file index in the scope of the scan.
-    ///
-    /// See also [crate::layouts::row_id].
-    row_index: bool,
+    /// The row-offset assigned to the first row of the file. Used by the `row_idx` expression,
+    /// but not by the scan [`Selection`] which remains relative.
+    row_offset: u64,
```
a10y, line 65: why is this here? some sort of global row offset for multi-file scans?

vortex-layout/src/scan/mod.rs
gatesn: The filter node doesn't itself do any filtering. It just reorders conjunctions by selectivity

vortex-layout/src/scan/mod.rs
gatesn: Yup, similar to Polars API for returning a row index column.

robert3005 (approved):
PartitionedMaskEvaluation is a bit funky with two evaluation functions but in current flow there's no easy way around it (discussed offline).

PR #3823, opened 2025-07-10
Fix empty take for `ChunkedArray`

blaginin:
Got a panic when running take 

image


vortex-array/src/arrays/chunked/compute/take.rs
```diff
@@ -13,14 +13,20 @@ use crate::{Array, ArrayRef, IntoArray, ToCanonical, register_kernel};
...
+        let return_dtype = array.dtype().clone().union_nullability(nullability);
+
         let indices = cast(
             indices,
             &DType::Primitive(PType::U64, indices.dtype().nullability()),
         )?
         .to_primitive()?;
 
+        if indices.is_empty() {
+            return Ok(ChunkedArray::try_new(vec![], return_dtype)?.to_array());
```
robert3005: `Canonical::empty(return_dtype)`?

robert3005: Also do this before the cast

robert3005: Tbh, do this in the top level take function so it's handled once and for all

vortex-array/src/compute/take.rs
```diff
@@ -12,9 +12,13 @@ use crate::arrays::ConstantArray;
 use crate::compute::{ComputeFn, ComputeFnVTable, InvocationArgs, Kernel, Output};
 use crate::stats::{Precision, Stat, StatsProviderExt, StatsSet};
 use crate::vtable::VTable;
-use crate::{Array, ArrayRef, IntoArray};
+use crate::{Array, ArrayRef, Canonical, IntoArray};
 
 pub fn take(array: &dyn Array, indices: &dyn Array) -> VortexResult<ArrayRef> {
+    if indices.is_empty() {
+        return Ok(Canonical::empty(array.dtype()).into_array());
```
robert3005: You still need the nullability union here, if the original array is non nullable but take is nullable it should end up with a nullable array

blaginin: good that we have a test that checks it!

robert3005 (approved):

PR #3825, opened 2025-07-10
feat: Add optimize ArrayOp with VBView implementation

a10y:
Adds a new OperationVTable method `optimize` which will return a new array with unused data garbage-collected. This is similar to the arrow-rs `StringViewArray::gc` method. See doc comments in the PR for more info.

This implements a simple optimize operation that will fully rebuild a VarBinViewArray if < 50% of the buffer data is still being referenced.

----

## Original PR text

Initial Claude Code prompt:

```
> i'd like you to work on a new ComputeFn called `Optimize`. This is a compute function that can be
  applied to an array to garbage collect/reduce its size. The default implementation for most types
  will be effectively a no-op, e.g. there's really no way to take a PrimitiveArray and optimize it.

  It's REALLY useful though for variable-length arrays or arrays with variable buffer counts. For
  example, VarBinViewArray can be filtered, and the filter implemented in
  @vortex-array/src/arrays/varbinview/compute/filter.rs will just filter the views (16-byte prefix +
   len + metadata for strings, or fully inlined short strings). After aggressive filtering you can
  end up with a situation where the VarBinViewArray holds a small number of views, and many buffers
  that are unreferenced, or a large buffer being kept alive by just a few string values.

  I want you to implement a new compute function, for reference on how to do this use the TakeKernel
   as an example, see @vortex-array/src/compute/take.rs for the Take compute function and the
  corresponding TakeKernel trait, and an example implementation for particular array encoding in
  @vortex-array/src/arrays/varbinview/compute/take.rs for VarBinViewArray.

  Implement a new Optimize compute function, that also defines an OptimizeKernel trait. To start
  with, implement the OptimizeKernel for VarBinViewArray to build a new compact VarBinViewArray that
   uses only a single buffer.

  Give me a plan for how you'll approach this so i can make sure you know what you're doing, then
  get going.
```

It did great at the ComputeFn boilerplate, I had to fix up a number of things and tweak (e.g. it was failing tests b/c it didn't know how to unwrap scalars, had an overcomplicated initial implementation of VBView compaction). Leaving in draft until I've had a chance to review every line.


vortex-array/src/arrays/varbinview/compute/optimize.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::builders::{ArrayBuilder, VarBinViewBuilder};
+use crate::compute::{OptimizeKernel, OptimizeKernelAdapter};
+use crate::{ArrayRef, register_kernel};
+
+impl OptimizeKernel for VarBinViewVTable {
+    fn optimize(&self, array: &VarBinViewArray) -> VortexResult<ArrayRef> {
+        // If there are no buffers or only one buffer, no optimization needed
+        if array.buffers().len() <= 1 {
+            return Ok(array.to_array());
+        }
```
a10y: Should probably also calculate total # of bytes referenced by all the views VS total number of bytes held in buffers, have some threshold e.g. 50% to kick in

vortex-array/src/compute/optimize.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    // Otherwise, canonicalize and try again.
+    if !array.is_canonical() {
+        log::debug!(
+            "No optimize implementation found for {}",
+            array.encoding_id()
+        );
+        let canonical = array.to_canonical()?;
+        return optimize(canonical.as_ref());
+    }
```
a10y: it should just do nothing if there is no kernel defined

vortex-array/src/arrays/varbinview/compute/optimize.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        // Iterate through the array and append each element
+        for i in 0..array.len() {
+            if !array.is_valid(i)? {
+                builder.append_null();
+            } else {
+                let bytes = array.bytes_at(i);
+                builder.append_value(bytes.as_slice());
+            }
+        }
```
a10y: there should be an optimized pathway for AllValid or NonNullable arrays to avoid the branch on every value

vortex-array/src/arrays/varbinview/compute/optimize.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        // Verify the data is still correct
+        assert_eq!(optimized_array.len(), 2);
+        assert_eq!(
+            &*optimized_array
+                .scalar_at(0)
+                .unwrap()
+                .as_utf8()
+                .value()
+                .unwrap(),
```
a10y: i don't like how annoying it is to unwrap a scalar to a rust value. maybe we need some better accessors on Scalar

robert3005:
FWIW arrow has gc for StringView arrays so the most trivial thing you could do is to_arrow -> gc -> to_vortex

robert3005:
since we also have a builder this isn't less code overall, ignore me

vortex-array/src/arrays/varbinview/compute/optimize.rs
robert3005: We have a lot of TryFrom implementations, you could `BufferString::try_from(optimized_array.scalar_at(0))`

robert3005: or `optimized_array.scalar_at(0).try_into::<BufferString>().unwrap()`

gatesn:
I think this ought to be an ArrayOperation, not a compute function. I want to move in the direction where ComputeFn == ScalarFn.

So e.g. FilterFn + TakeFn also move to ArrayOperation (although probably as their own VTable thingys)

a10y:
I see, so concretely that means just moving this as another method on OperationsVTable and only impl'ing it for VarbinViewArray?

a10y:
Think I did the ArrayOp thing right

vortex-array/src/arrays/varbinview/ops.rs
```diff
@@ -24,4 +26,243 @@ impl OperationsVTable<VarBinViewVTable> for VarBinViewVTable {
...
+                // SAFETY: in this branch the view is not inlined.
+                unsafe { view._ref }.size as u64
+            }
+        })
+        .sum();
+
+    let buffer_total_size: u64 = array.buffers.iter().map(|buf| buf.len() as u64).sum();
+
+    // If the majority of buffer space is unused, attempt to repack
+    bytes_referenced < buffer_total_size / 2
```
a10y: is this too conservative? is it too aggressive?

vortex-array/src/arrays/varbinview/ops.rs
```diff
@@ -24,4 +26,211 @@ impl OperationsVTable<VarBinViewVTable> for VarBinViewVTable {
...
+}
+
+fn should_compact(array: &VarBinViewArray) -> bool {
+    // If the array is entirely inlined strings, do not attempt to compact.
+    if array.nbuffers() == 0 {
+        return false;
+    }
+
+    // Scan the views, calculating the total buffer size that is referenced.
+    let bytes_referenced: u64 = array
```
a10y: i need to skip null views when i do this

vortex-array/src/arrays/varbinview/ops.rs
gatesn: I kinda think we should always repack if there's any extra bytes whatsoever?  For now, assume / document that this is aggressive.

robert3005: If you write to parquet you would only write present values

a10y: I made it more aggressive

vortex-array/src/array/mod.rs
```diff
@@ -62,6 +62,8 @@ pub trait Array: 'static + private::Sealed + Send + Sync + Debug + ArrayVisitor
     /// Fetch the scalar at the given index.
     fn scalar_at(&self, index: usize) -> VortexResult<Scalar>;
 
+    fn optimize(&self) -> VortexResult<ArrayRef>;
```
gatesn, line 68: docs docs docs

vortex-array/src/array/mod.rs
```diff
@@ -430,6 +436,10 @@ impl<V: VTable> Array for ArrayAdapter<V> {
         Ok(scalar)
     }
 
+    fn optimize(&self) -> VortexResult<ArrayRef> {
+        <V::OperationsVTable as OperationsVTable<V>>::optimize(&self.0)
```
gatesn: Is it worth asserting that dtype and len remain unchanged? Possibly even that nbytes is <= prev_nbytes?

vortex-array/src/vtable/operations.rs
```diff
@@ -28,4 +28,25 @@ pub trait OperationsVTable<V: VTable> {
...
+    ///
+    /// This operation can be called to return a new copy of an array with the same encoding,
+    /// but with all unreferenced data unlinked.
+    ///
+    /// ## Default behavior
+    ///
+    /// For most arrays that do not contain variable buffer counts, such as the canonical
+    /// arrays, the default implementation will not attempt to perform compaction and instead
+    /// return the original array.
+    fn optimize(array: &V::Array) -> VortexResult<ArrayRef> {
```
gatesn: This should return `VortexResult<V::Array>` to ensure people don't be sneaky and change the encoding

gatesn (changes requested):

gatesn:
We probably also want this for DictArray with unused values in the codes, but that can be another PR.

vortex-array/src/array/mod.rs
a10y: there are more docs in the OperationsVTable, I can add a short sentence here

vortex-array/src/array/mod.rs
a10y: the nbytes might be a step too far? maybe as a debug assertion. i can assert the dtype/len though

vortex-array/src/array/mod.rs
gatesn: Yeah, these are docs for the caller. 

The vtable docs are for the implementer

gatesn (approved):

PR #3827, opened 2025-07-10
feat: display

### LINKED ISSUE #3821: Array formatting
Right now, the default `{:?}` implementation is hard to read:

`PrimitiveArray { dtype: Primitive(U64, Nullable), buffer: Buffer<u8> { length: 0, alignment: Alignment(8), as_slice: [] }, validity: AllValid, stats_set: ArrayStats { inner: RwLock { data: StatsSet { values: [] } } } }, self.key_space_idxs: [ChunkedArray { dtype: Primitive(U64, Nullable), len: 8192, chunk_offsets: Buffer<u64> { length: 9, alignment: Alignment(8), as_slice: [0, 1024, 2048, 3072, 4096, 5120, 6144, 7168, 8192] }, chunks: [PrimitiveArray { dtype: Primitive(U64, Nullable), buffer: Buffer<u8> { length: 8192, alignment: Alignment(8), as_slice: [0, 32, 0, 0, 0, 0, 0, 0, 1, 32, 0, 0, 0, 0...`

We want to allow users to choose the verbosity level.

The modes we could support:
- Human: unwraps all internal layers and prints the final logical values (e.g. just the final decoded array, even if it's `ChunkArray(DictArray(PrimitiveArray...`
- Verbose: preserves structure but formats each buffer more clearly (e.g. multi-line, decoded values, truncated if needed)
- Debug: same as current behavior (full structure, raw buffers)

Implementation wise, we can add `Array::display_as` and a new formatting VTable?

danking:
Closes https://github.com/vortex-data/vortex/issues/3821

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -542,6 +542,7 @@ dependencies = [
  "opentelemetry",
  "opentelemetry-otlp",
  "opentelemetry_sdk",
+ "parking_lot",
  "parquet",
  "paste",
  "rand 0.9.1",
@@ -551,6 +552,7 @@ dependencies = [
  "serde",
  "serde_json",
  "similar",
+ "sysinfo",
  "tabled",
  "tar",
  "target-lexicon",
@@ -3665,6 +3667,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ntapi"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e8a3895c6391c39d7fe7ebc444a87eb2991b2a0bc718fdabd071eec617fc68e4"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "nu-ansi-term"
 version = "0.46.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3782,6 +3793,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5c7398b9c8b70908f6371f47ed36737907c87c52af34c268fed0bf0ceb92ead9"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "objc2-core-foundation"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1c10c2894a6fed806ade6027bcd50662746363a9589d3ec9d9bef30a4e4bc166"
+dependencies = [
+ "bitflags",
+]
+
+[[package]]
+name = "objc2-io-kit"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "71c1c64d6120e51cd86033f67176b1cb66780c2efe34dec55176f77befd93c0a"
+dependencies = [
+ "libc",
+ "objc2-core-foundation",
 ]
 
 [[package]]
@@ -5413,6 +5443,20 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.104",
+]
+
+[[package]]
+name = "sysinfo"
+version = "0.36.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aab138f5c1bb35231de19049060a87977ad23e04f2303e953bc5c2947ac7dec4"
+dependencies = [
+ "libc",
+ "memchr",
+ "ntapi",
+ "objc2-core-foundation",
+ "objc2-io-kit",
+ "windows",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -159,6 +159,7 @@ simplelog = "0.12"
 sketches-ddsketch = "0.3.0"
 static_assertions = "1.1"
 strum = "0.27"
+sysinfo = "0.36.0"
 tabled = { version = "0.20.0", default-features = false }
 taffy = "0.8.0"
 tar = "0.4"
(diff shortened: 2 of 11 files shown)


vortex-array/src/vtable/display.rs
```diff
@@ -0,0 +1,113 @@
...
+        assert_eq!(s.display().to_string(), "[{},null,{}]");
+
+        let x = buffer![1, 2, 3, 4].into_array();
+        let y = buffer![-1, -2, -3, -4].into_array();
+        let s = StructArray::from_fields(&[("x", x), ("y", y)])
+            .unwrap()
+            .into_array();
+        assert_eq!(
+            s.display().to_string(),
+            "[{x: 1i32, y: -1i32},{x: 2i32, y: -2i32},{x: 3i32, y: -3i32},{x: 4i32, y: -4i32}]"
```
blaginin: so much better!

vortex-array/src/vtable/display.rs
```diff
@@ -0,0 +1,113 @@
...
+            "{}",
+            (0..array.len())
+                .map(|i| array.scalar_at(i).vortex_expect("index is in bounds"))
+                .format(",")
+        )?;
+        write!(f, "]")
+    }
+}
+
+#[cfg(test)]
```
blaginin: can we also test wrapped types here? like dict of primitive

vortex-array/src/vtable/display.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use itertools::Itertools as _;
+use vortex_error::VortexExpect as _;
+
+use crate::Array;
+use crate::vtable::VTable;
+
+pub struct DisplayArray<'a>(pub &'a dyn Array);
+
+impl Display for DisplayArray<'_> {
```
blaginin: should just array itself implement display?

vortex-array/src/vtable/display.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        assert!(matches!(options, DisplayOptions::CommaSeparatedScalars));
+
+        write!(f, "[")?;
+        write!(
+            f,
+            "{}",
+            (0..array.len())
+                .map(|i| array.scalar_at(i).vortex_expect("index is in bounds"))
```
blaginin: you probs want to write each scalar independently so `{:#?}` will work?

vortex-array/src/vtable/display.rs
blaginin: also, should we do spaces after commas?

danking: personally I like super compact but I'll add an option and default to a space

vortex-array/src/vtable/display.rs
danking: `dyn Array` does implement Display but its the MetadataOnly option. I could change that default to the comma separated values one?

vortex-array/src/vtable/display.rs
danking: Dict has to go in the dict crate but, sure, added!

vortex-array/src/vtable/display.rs
danking: I don't think I follow; can you give me a more complete diff?

gatesn:
If you're just looking for debug printing the "logical values", don't we only need to implement this for Canonical arrays?

As in, each encoding doesn't need to figure out its own logical display.

Concretely, I wonder if we should put this in `impl dyn Array + _`, match over some display enum, and only call into the vtable for display modes that require encoding-specific information.

I would also move TreeDisplay into this enum while we're at it.

blaginin:
> If you're just looking for debug printing the "logical values", don't we only need to implement this for Canonical arrays?

i think in the future it could actually be useful to add a mode which shows nested arrays yet not as verbose as debug?

robert3005:
Sometimes it's useful to go through the whole vtable dispatch, however, these are cases where the array is somehow buggy which is not the good thing to optimise for. Maybe something like #3633 would help here?

danking:
@robert3005  What cost does the vtable introduce? I don't think we're optimizing for the case that, say, dictionary array wants to render itself as codes and values, but we're making it possible at all.

gatesn:
I don't thin that arrays (beyond canonical) should even have the ability to decide how they're rendered for the large part. Tree display takes common components of arrays and lays them out. This "data" display should take the logical data and lay it out.

It's hard to say whether `array.display(...)` should e.g. show codes/values for a dictionary. Or show the underlying logical data. Tree display does the former, this should do the latter imo.

vortex-array/src/vtable/display.rs
blaginin: https://doc.rust-lang.org/std/fmt/index.html - there's a `#` sign to do pretty-printing. But i believe for it to work you need to pass each element individually and not as a one big string

robert3005:
@gatesn the problem is that tree_display is not quite that. So we now know that we want a logical display, tree display with data and tree display with metadata only (current tree display).

gatesn:
We might be speaking past one another and saying the same thing.

All I'm saying is that we shouldn't add anything to the array vtable in order to implement this.

vortex-array/src/vtable/display.rs
danking: Ah, yes, pretty-printing would be best but I don't want the scalars printed in debug mode because I don't want to see, for example, a buffer string printed as an array of u8.

danking:
Continued here: https://github.com/vortex-data/vortex/pull/3866

PR #3843, opened 2025-07-11
chore: better naming for SendableLayoutFuture

a10y:
It's not really a writer, if anything the LayoutStrategy is the LayoutWriter, the thing it returns should be a `SendableLayoutFuture`

### DIFF
--- a/vortex-layout/src/writer.rs
+++ b/vortex-layout/src/writer.rs
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use std::pin::Pin;
-
+use futures::future::BoxFuture;
 use vortex_error::VortexResult;
 
 use crate::LayoutRef;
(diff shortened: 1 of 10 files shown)


vortex-layout/src/writer.rs
```diff
@@ -11,11 +10,5 @@ use crate::LayoutRef;
 /// trait to be potentially extended with new methods.
 // Tag for Python docs:
 // [layout writer]
-pub trait LayoutWriter: Future<Output = VortexResult<LayoutRef>> {}
+pub type SendableLayoutFuture = BoxFuture<'static, VortexResult<LayoutRef>>;
```
a10y, line 13: Every layout is currently a Box::pin, and that doesn't seem likely to change soon

robert3005 (approved):

gatesn:
But isn't it the LayoutWriter that actually awaits on the SegmentWriter?

I guess I saw it as a convenience that it _also_ returns the Layout definition.

PR #3844, opened 2025-07-11
feat[compressor]: wire up sequence array into the compressor

joseph-isaacs:
Add sequence array into the compressor.

We might in the future want to add patches or a validity array, but these are omitted for now, and can be added without a break

### DIFF
--- a/encodings/sequence/src/compress.rs
+++ b/encodings/sequence/src/compress.rs
@@ -3,46 +3,53 @@
 
 use vortex_array::ArrayRef;
 use vortex_array::arrays::PrimitiveArray;
-use vortex_dtype::{NativePType, match_each_integer_ptype};
+use vortex_dtype::{NativePType, Nullability, match_each_integer_ptype};
 use vortex_error::VortexResult;
 use vortex_scalar::PValue;
 
 use crate::SequenceArray;
 
 pub fn sequence_encode(primitive_array: &PrimitiveArray) -> VortexResult<Option<ArrayRef>> {
-    if primitive_array.len() == 0 {
+    if primitive_array.is_empty() {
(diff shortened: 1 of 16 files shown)


encodings/sequence/src/compress.rs
```diff
@@ -0,0 +1,81 @@
...
+use vortex_dtype::{NativePType, match_each_integer_ptype};
+use vortex_error::VortexResult;
+use vortex_scalar::PValue;
+
+use crate::SequenceArray;
+
+pub fn sequence_encode(primitive_array: &PrimitiveArray) -> VortexResult<Option<ArrayRef>> {
+    if primitive_array.len() == 0 {
+        // we cannot encode an empty array
+        return Ok(None);
```
a10y, line 22: can we lump in AllValid at least?

joseph-isaacs: I thought about that, but the dtype would change

joseph-isaacs: In fact maybe not

a10y: Why would the dtype change?

a10y: Oh, there's no validity array. but we store DType anyway for the SequenceArray, I think it can still accept a nullable primitive as long as there are no actual nulls to store?

a10y (approved):

encodings/sequence/src/compress.rs
joseph-isaacs: Exactly

robert3005:
I think you want to check that your sequence array is valid somewhere higher up since right now the construction fails instead.

PR #3846, opened 2025-07-11
fix: teach SparseArray to account for null values when the fill is null

danking:


danking:
@joseph-isaacs now with more zoomies.

robert3005 (approved):

PR #3847, opened 2025-07-11
chore: Add `new_expr` to all built-in expressions in the name of consistency

AdamGS:


vortex-dtype/src/field.rs
```diff
@@ -177,7 +177,7 @@ impl FieldPath {
     }
 
     /// Does the field referenced by the field path exist in the given dtype?
-    pub fn exists(&self, dtype: DType) -> bool {
+    pub fn exists_in(&self, dtype: DType) -> bool {
```
AdamGS, line 180: I've found these too to be somewhat confusing, so also suggesting these small naming changes

robert3005 (approved):

PR #3848, opened 2025-07-11
feat: teach utf8-type SparseArray to canonicalize

danking:
Waiting on https://github.com/vortex-data/vortex/pull/3846

### DIFF
--- a/encodings/sparse/src/canonical.rs
+++ b/encodings/sparse/src/canonical.rs
@@ -2,22 +2,25 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use itertools::Itertools;
+use num_traits::NumCast;
 use vortex_array::arrays::{
-    BoolArray, BooleanBuffer, ConstantArray, NullArray, PrimitiveArray, StructArray,
-    smallest_storage_type,
+    BinaryView, BoolArray, BooleanBuffer, ConstantArray, NullArray, PrimitiveArray, StructArray,
+    VarBinViewArray, smallest_storage_type,
 };
 use vortex_array::builders::{ArrayBuilder as _, DecimalBuilder};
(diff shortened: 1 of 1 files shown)


encodings/sparse/src/canonical.rs
```diff
@@ -199,12 +216,59 @@ fn canonicalize_sparse_decimal<D: NativeDecimalType>(
...
+    let fill = if let Some(buffer) = &fill_value.value() {
+        buffers.push(buffer.inner().clone());
+        BinaryView::make_view(
+            buffer.as_ref(),
+            u32::try_from(n_patch_buffers).vortex_expect("too many buffers"),
+            0,
+        )
+    } else {
+        // any <=12 character value will do
+        BinaryView::make_view("FILL_VALUE".as_ref(), 0, 0)
```
a10y: you can just pass &[], it will make a new Inlined view that is zeroed

encodings/sparse/src/canonical.rs
```diff
@@ -199,12 +216,59 @@ fn canonicalize_sparse_decimal<D: NativeDecimalType>(
...
+    };
+
+    let mut view_builder = BufferMut::<BinaryView>::with_capacity(len);
```
a10y: can you instead just initialize the buffer to the fill like

```
buffer_mut![fill; len]
```

and then iterate the patch indices to insert the patches over it? that seems cleaner

encodings/sparse/src/canonical.rs
danking: Done.

encodings/sparse/src/canonical.rs
danking: oh yeah that's much better. I was still thinking of this as an array builder.

robert3005 (approved):

PR #3851, opened 2025-07-11
feat: Minimal support for extension array casting

AdamGS:
I think this is the most minimal example of extension array casting that's easy and safe to execute? Basically just handles casting within the same extension dtype, changing nullability when appropriate. 

Actually casting the underlying storage array seems dangerous given that we don't know how its actually used.

### DIFF
--- a/encodings/datetime-parts/src/compute/cast.rs
+++ b/encodings/datetime-parts/src/compute/cast.rs
@@ -4,26 +4,28 @@
 use vortex_array::compute::{CastKernel, CastKernelAdapter, cast};
 use vortex_array::{Array, ArrayRef, IntoArray, register_kernel};
 use vortex_dtype::DType;
-use vortex_error::{VortexResult, vortex_bail};
+use vortex_error::VortexResult;
 
 use crate::{DateTimePartsArray, DateTimePartsVTable};
 
 impl CastKernel for DateTimePartsVTable {
-    fn cast(&self, array: &DateTimePartsArray, dtype: &DType) -> VortexResult<ArrayRef> {
+    fn cast(&self, array: &DateTimePartsArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         if !array.dtype().eq_ignore_nullability(dtype) {
-            vortex_bail!("cannot cast from {} to {}", array.dtype(), dtype);
+            return Ok(None);
         };
 
-        Ok(DateTimePartsArray::try_new(
-            dtype.clone(),
-            cast(
-                array.days().as_ref(),
-                &array.days().dtype().with_nullability(dtype.nullability()),
-            )?,
-            array.seconds().clone(),
-            array.subseconds().clone(),
-        )?
-        .into_array())
+        Ok(Some(
+            DateTimePartsArray::try_new(
+                dtype.clone(),
+                cast(
+                    array.days().as_ref(),
+                    &array.days().dtype().with_nullability(dtype.nullability()),
+                )?,
+                array.seconds().clone(),
+                array.subseconds().clone(),
+            )?
+            .into_array(),
+        ))
     }
 }
 
@@ -86,10 +88,10 @@ mod tests {
         let array = date_time_array(validity);
         let result = cast(&array, &DType::Bool(Nullability::NonNullable));
         assert!(
-            result
-                .as_ref()
-                .is_err_and(|err| err.to_string().contains("cannot cast from")),
-            "{result:?}"
+            result.as_ref().is_err_and(|err| err.to_string().contains(
+                "No compute kernel to cast array vortex.ext with dtype ext(vortex.timestamp, i64, ExtMetadata([2, 3, 0, 85, 84, 67]))? to bool"
+            )),
+            "Got error: {result:?}"
         );
 
         let result = cast(
@@ -100,7 +102,7 @@ mod tests {
             result.as_ref().is_err_and(|err| err
                 .to_string()
                 .contains("invalid cast from nullable to non-nullable")),
-            "{result:?}"
+            "Got error: {result:?}"
         );
     }
 }
--- a/vortex-array/src/arrays/constant/compute/cast.rs
+++ b/vortex-array/src/arrays/constant/compute/cast.rs
@@ -9,8 +9,11 @@ use crate::compute::{CastKernel, CastKernelAdapter};
 use crate::{ArrayRef, IntoArray, register_kernel};
 
 impl CastKernel for ConstantVTable {
-    fn cast(&self, array: &ConstantArray, dtype: &DType) -> VortexResult<ArrayRef> {
-        Ok(ConstantArray::new(array.scalar().cast(dtype)?, array.len()).into_array())
+    fn cast(&self, array: &ConstantArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        match array.scalar().cast(dtype) {
+            Ok(scalar) => Ok(Some(ConstantArray::new(scalar, array.len()).into_array())),
+            Err(_e) => Ok(None),
+        }
     }
 }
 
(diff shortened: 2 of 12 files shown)


AdamGS:
This obviously doesn't allow other extensions to register their own casting logic, but I think it can serve as the entry point/default behavior.

vortex-array/src/arrays/extension/compute/cast.rs
```diff
@@ -0,0 +1,124 @@
...
+use crate::compute::{CastKernel, CastKernelAdapter, cast};
+use crate::{ArrayRef, IntoArray, register_kernel};
+
+impl CastKernel for ExtensionVTable {
+    fn cast(&self, array: &ExtensionArray, dtype: &DType) -> vortex_error::VortexResult<ArrayRef> {
+        if !array.dtype().eq_ignore_nullability(dtype) {
+            vortex_bail!("Cannot cast {} to {}", array.dtype(), dtype);
+        }
+
+        let DType::Extension(ext_dtype) = dtype else {
```
gatesn, line 20: I think you can panic here (as_ext?), since you checked above it's equal ignoring nullability.

gatesn: In fact, I think the rest of this function can disappear probably? 

You're basically implementing a kernel that provides nullability casting (only) for extension dtypes by delegating to casting of the storage array.

This is nice! Provides generic behaviour that all extension types should exact. Plugins containing custom extension types can just register their own complex kernels if they wish (e.g. geospatial -> string or vice versa)

AdamGS: not sure I follow which part you think can be dropped

encodings/datetime-parts/src/compute/cast.rs
```diff
@@ -9,21 +9,23 @@ use vortex_error::{VortexResult, vortex_bail};
 use crate::{DateTimePartsArray, DateTimePartsVTable};
 
 impl CastKernel for DateTimePartsVTable {
-    fn cast(&self, array: &DateTimePartsArray, dtype: &DType) -> VortexResult<ArrayRef> {
+    fn cast(&self, array: &DateTimePartsArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         if !array.dtype().eq_ignore_nullability(dtype) {
             vortex_bail!("cannot cast from {} to {}", array.dtype(), dtype);
```
gatesn:
```suggestion
            return Ok(None);
```

vortex-array/src/arrays/constant/compute/cast.rs
```diff
@@ -9,8 +9,10 @@ use crate::compute::{CastKernel, CastKernelAdapter};
 use crate::{ArrayRef, IntoArray, register_kernel};
 
 impl CastKernel for ConstantVTable {
-    fn cast(&self, array: &ConstantArray, dtype: &DType) -> VortexResult<ArrayRef> {
-        Ok(ConstantArray::new(array.scalar().cast(dtype)?, array.len()).into_array())
+    fn cast(&self, array: &ConstantArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        Ok(Some(
+            ConstantArray::new(array.scalar().cast(dtype)?, array.len()).into_array(),
```
gatesn: If the cast fails, I think you should return `Ok(None)`

vortex-array/src/arrays/extension/compute/cast.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::{ArrayRef, IntoArray, register_kernel};
+
+impl CastKernel for ExtensionVTable {
+    fn cast(
+        &self,
+        array: &ExtensionArray,
+        dtype: &DType,
+    ) -> vortex_error::VortexResult<Option<ArrayRef>> {
+        if !array.dtype().eq_ignore_nullability(dtype) {
+            vortex_bail!("Cannot cast {} to {}", array.dtype(), dtype);
```
gatesn:
```suggestion
            return Ok(None);
```

vortex-array/src/arrays/extension/compute/cast.rs
```diff
@@ -0,0 +1,134 @@
...
+            unreachable!("Already verified we have an extension dtype");
+        };
+
+        let new_nullability = dtype.nullability();
+
+        let storage_array = array.storage().clone();
+        let storage_dtype = ext_dtype.storage_dtype();
+        let storage_nullability = ext_dtype.storage_dtype().nullability();
+
+        let new_storage_dtype = if storage_nullability == new_nullability {
```
gatesn: `let new_storage_dtype = dtype.as_ext().storage()` ?

vortex-array/src/arrays/extension/compute/cast.rs
```diff
@@ -0,0 +1,134 @@
...
+        };
+        let new_storage = match cast(storage_array.as_ref(), &new_storage_dtype) {
+            Ok(arr) => arr,
+            Err(e) => {
+                log::warn!("Failed to cast storage array: {e}");
+                return Ok(None);
+            }
+        };
+
+        let new_ext_dtype = ExtDType::new(
```
gatesn: This is just `dtype`

vortex-array/src/arrays/list/compute/cast.rs
```diff
@@ -10,7 +10,7 @@ use crate::vtable::ValidityHelper;
 use crate::{ArrayRef, register_kernel};
 
 impl CastKernel for ListVTable {
-    fn cast(&self, array: &Self::Array, dtype: &DType) -> VortexResult<ArrayRef> {
+    fn cast(&self, array: &Self::Array, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         let Some(target_element_type) = dtype.as_list_element() else {
             vortex_bail!("cannot cast {} to {}", array.dtype(), dtype);
```
gatesn: Ok(None)

vortex-array/src/arrays/primitive/compute/cast.rs
```diff
@@ -13,7 +13,7 @@ use crate::vtable::ValidityHelper;
 use crate::{ArrayRef, IntoArray, register_kernel};
 
 impl CastKernel for PrimitiveVTable {
-    fn cast(&self, array: &PrimitiveArray, dtype: &DType) -> VortexResult<ArrayRef> {
+    fn cast(&self, array: &PrimitiveArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         let DType::Primitive(new_ptype, new_nullability) = dtype else {
             vortex_bail!(MismatchedTypes: "primitive type", dtype);
```
gatesn: Ok(None)

vortex-array/src/arrays/varbin/compute/cast.rs
```diff
@@ -10,21 +10,23 @@ use crate::vtable::ValidityHelper;
 use crate::{ArrayRef, IntoArray, register_kernel};
 
 impl CastKernel for VarBinVTable {
-    fn cast(&self, array: &VarBinArray, dtype: &DType) -> VortexResult<ArrayRef> {
+    fn cast(&self, array: &VarBinArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         if !array.dtype().eq_ignore_nullability(dtype) {
             vortex_bail!("Cannot cast {} to {}", array.dtype(), dtype);
```
gatesn: Ok(None)

vortex-array/src/arrays/varbinview/compute/cast.rs
```diff
@@ -10,21 +10,23 @@ use crate::vtable::ValidityHelper;
 use crate::{ArrayRef, IntoArray, register_kernel};
 
 impl CastKernel for VarBinViewVTable {
-    fn cast(&self, array: &VarBinViewArray, dtype: &DType) -> VortexResult<ArrayRef> {
+    fn cast(&self, array: &VarBinViewArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         if !array.dtype().eq_ignore_nullability(dtype) {
             vortex_bail!("Cannot cast {} to {}", array.dtype(), dtype);
```
gatesn: Ok(None)

gatesn (approved):

PR #3852, opened 2025-07-11
feat: call optimize in compressor

a10y:
BtrBlocksCompressor calls optimize before sending array to compression.

### DIFF
--- a/vortex-btrblocks/src/lib.rs
+++ b/vortex-btrblocks/src/lib.rs
@@ -276,7 +276,8 @@ pub struct BtrBlocksCompressor;
 
 impl BtrBlocksCompressor {
     pub fn compress(&self, array: &dyn Array) -> VortexResult<ArrayRef> {
-        self.compress_canonical(array.to_canonical()?)
+        // Optimize the array before compressing it.
+        self.compress_canonical(array.optimize()?.to_canonical()?)
     }
 
     pub fn compress_canonical(&self, array: Canonical) -> VortexResult<ArrayRef> {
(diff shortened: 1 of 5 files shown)


robert3005:
I think you miss the call to optimise for list array or binary arrays?

gatesn:
I thought the problem in the issue was caused by this flow (perhaps this PR wasn't intended to specifically fix it? Since it's a good idea regardless)

```
array = VarBinView::new(massive string array)
array = array.slice(0, 1)

fsst_array = compress(array)
```

So the FSST array correctly has length 1. But it naively slurped up all the data buffers and compressed them.

I wonder if the place to call optimize is actually here: https://github.com/vortex-data/vortex/blob/416b15ae7aa4c4b70af2ba3dd42fc085f5c3651c/vortex-btrblocks/src/lib.rs#L279

robert3005:
The original issue mentions just struct and list arrays. I think all of our compressors only compress present data. The naive way in which we compress lists right now grabs all the buffers though

vortex-btrblocks/src/lib.rs
```diff
@@ -276,7 +276,8 @@ pub struct BtrBlocksCompressor;
 
 impl BtrBlocksCompressor {
     pub fn compress(&self, array: &dyn Array) -> VortexResult<ArrayRef> {
-        self.compress_canonical(array.to_canonical()?)
+        // Optimize the array before compressing it.
+        self.compress_canonical(array.optimize()?.to_canonical()?)
```
gatesn: Think you want to canonicalize first

a10y: Ah yea, that's right

robert3005 (approved):

PR #3856, opened 2025-07-13
fix: Drain prefetch buffer

a10y:
As I was reviewing the prefetch logic, I noticed that there were no codepaths that would set `is_prefetched` to true, I think there is a bug where the prefetch buffer would get filled and then never drained.

### DIFF
--- a/vortex-file/src/driver.rs
+++ b/vortex-file/src/driver.rs
@@ -20,6 +20,14 @@ use vortex_utils::aliases::hash_map::HashMap;
 
 use crate::SegmentSpec;
 
+macro_rules! trace_log {
+    ($($tts:tt)*) => {
+        if ::log::log_enabled!(::log::Level::Trace) {
+            ::log::trace!($($tts)*);
+        }
+    };
+}
+
 /// An I/O driver that assembles coalesced requests based on a performance hint and a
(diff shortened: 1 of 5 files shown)


vortex-file/src/driver.rs
```diff
@@ -102,13 +98,13 @@ impl CoalescedDriver {
...
-
-        // If the segment has been pre-fetched, we can remove its bytes from the buffer.
+        let state = self.segment_state_mut(id);
         assert!(!state.is_prefetched, "segment already prefetched");
+
+        // Account for these new bytes in the prefetch buffer
         if !state.polled {
+            state.is_prefetched = true;
             self.prefetched_bytes += self.segment_spec(id).length as i64;
-            self.segment_state_mut(id).is_prefetched = false;
```
a10y, line 111: I think b/c we never were setting this true, the branch in unmark_as_prefetched below was probably never being taken

gatesn:
This doesn't seem to have much impact at all... 

Might be worth trying to get some visibility into this event stream / state machine.

a10y:
Yea it doesn't seem to have any significant impact. Still a bug so should probably be fixed?

robert3005 (approved):

PR #3858, opened 2025-07-14
LayoutReader to take &dyn SegmentSource

gatesn:
I'm not sure about this change. On the one hand, it makes sure layout readers only live while we are constructing tasks for the scan, since both VortexFile and Evaluation tasks have static lifetimes, this all works. But a layout reader cannot be held onto and reused later - this ensures correct use of the pre-fetching API by ensuring evaluation tasks eagerly construct their static SegmentFutures from the non-static &dyn SegmentSource.

The downsides are that this forces all scan tasks to be constructed up-front (which they current are anyway).

Btw, a LayoutReader is still a terrible name. It's more like an EvaluationFactory...


a10y:
I think I prefer the existing way over this, personally

PR #3861, opened 2025-07-14
feat[bench]: inter-query memory tracking 

joseph-isaacs:
Add peak memory usage (per query)
Screenshot 2025-07-14 at 12 04 13


bench-vortex/src/memory.rs
```diff
@@ -0,0 +1,246 @@
...
+            assert!(stats.virtual_memory > 0);
+        }
+    }
+
+    #[test]
+    fn test_memory_measurement() {
+        let measurement = MemoryMeasurement::start();
+
+        // Allocate some memory
+        let _data: Vec<u8> = vec![0; 1024 * 1024]; // 1MB
```
0ax1, line 236: Can we assert that it is at least 1MB?

bench-vortex/src/memory.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub physical_memory_delta: i64,
+    /// Change in virtual memory usage in bytes (can be negative)
+    pub virtual_memory_delta: i64,
+}
+
+/// Thread-safe memory tracker using sysinfo
+pub struct MemoryTracker {
+    system: Mutex<System>,
+    pid: u32,
+    peak_physical_memory: Mutex<u64>,
```
0ax1, line 49: Could consider AtomicU64 here.

bench-vortex/src/memory.rs
```diff
@@ -0,0 +1,246 @@
...
+    peak_virtual_memory: Mutex<u64>,
+}
+
+impl Default for MemoryTracker {
```
0ax1: I think we can proc macro derive default.

bench-vortex/src/memory.rs
```diff
@@ -0,0 +1,246 @@
...
+        system.refresh_processes(
+            ProcessesToUpdate::Some(&[sysinfo::Pid::from_u32(self.pid)]),
+            true,
+        );
+
+        if let Some(process) = system.process(sysinfo::Pid::from_u32(self.pid)) {
+            let current_stats = MemoryStats::new(process.memory(), process.virtual_memory());
+
+            // Update peak memory if current usage is higher
+            {
```
0ax1, line 94: Yeah I think it'd be nice going with atomics, then we can get rid of the extra scopes.

0ax1 (commented):
Couple of nits / remarks inline. Looks good!

bench-vortex/src/memory.rs
joseph-isaacs: can we omit this

0ax1 (approved):

PR #3863, opened 2025-07-14
feature: multi-file-scan thread-safe iterator

0ax1:
This PR implements a multi-scan operator which can be driven from multiple threads through `MultiScanIterator`s. Work-stealing will be implemented in a follow up. Further, the implementation will to a global conversion cache shared across all threads.


vortex-duckdb/src/scan.rs
```diff
@@ -54,17 +60,17 @@ impl std::fmt::Debug for VortexBindData {
...
 pub struct VortexGlobalData {
-    file_paths: SegQueue<PathBuf>,
-    is_first_file_processed: AtomicBool,
     /// We currently use a conversion cache to cache converted arrays, this id is used to
     /// ensure that each cache created has a unique id used to name those arrays
     conversion_cache_id: AtomicU64,
-    filter_expr: Option<ExprRef>,
-    projection_expr: ExprRef,
+    multi_file_iter: Arc<MultiFileIterator>,
+    thread_id: AtomicUsize,
```
joseph-isaacs:
```suggestion
    thread_count: AtomicUsize,
```

vortex-duckdb/src/scan.rs
```diff
@@ -54,17 +60,17 @@ impl std::fmt::Debug for VortexBindData {
...
 }
 
 pub struct VortexLocalData {
-    exporter: Option<ArrayIteratorExporter>,
+    exporter: Option<ArrayExporter>,
+    cache: Option<ConversionCache>,
```
joseph-isaacs: I think you want a cache per file

vortex-duckdb/src/scan.rs
```diff
@@ -210,82 +216,79 @@ impl TableFunction for VortexTableFunction {
...
-                        Some(exporter_for_file(&bind_data.first_file, cache_id)?);
-                }
-                // Retrieve a file path from the shared lock-free queue.
-                else if let Some(file_path) = global_state.file_paths.pop() {
-                    let file = VortexOpenOptions::file()
-                        .open_blocking(&file_path)
-                        .map_err(|e| vortex_err!("Failed to open Vortex file: {}", e))?;
+                let Some(array_result) = global_state.multi_file_iter.next(local_state.thread_id)
+                else {
+                    return Ok(());
```
joseph-isaacs, line 210: can you bail out of the loop, instead of returning, there is post loop conditions we should apply `assert!(!chunk.is_empty());`

vortex-duckdb/src/scan.rs
```diff
@@ -210,82 +216,79 @@ impl TableFunction for VortexTableFunction {
...
+                };
 
+                if local_state.cache.is_none() {
                     let cache_id = global_state.conversion_cache_id.fetch_add(1, SeqCst);
```
joseph-isaacs: the cache id should be the file_idx

vortex-duckdb/src/scan.rs
```diff
@@ -210,82 +216,79 @@ impl TableFunction for VortexTableFunction {
...
         let filter_expr = extract_table_filter_expr(init_input, init_input.column_ids())?;
 
+        let num_threads = thread::available_parallelism()
+            .map(|p| p.get())
+            .expect("Failed to get number of available threads");
+
+        let multi_file_iter = MultiFileIterator::new(num_threads)
+            .with_file_paths(&bind_data.file_paths[1..])
+            .with_vortex_files([bind_data.first_file.clone()])
+            .with_projection_expr(Some(projection_expr))
```
joseph-isaacs: can this be a stream of VortexFiles?

joseph-isaacs (approved):
Here are some comments, let see the perf before we think about merging

vortex-scan/src/iterator.rs
```diff
@@ -0,0 +1,184 @@
...
+use vortex_array::ArrayRef;
+use vortex_dtype::DType;
+use vortex_error::VortexResult;
+use vortex_expr::ExprRef;
+use vortex_file::{VortexFile, VortexOpenOptions};
+
+type ArrayFuture = BoxFuture<'static, VortexResult<Option<ArrayRef>>>;
+
+pub struct MultiFileIterator {
+    tokio_runtime: tokio::runtime::Runtime,
```
joseph-isaacs: Can we try a `LocalPool` per thread

vortex-duckdb/src/scan.rs
0ax1: done

vortex-duckdb/src/scan.rs
0ax1: can iterate on this in a follow up.

vortex-scan/src/iterator.rs
0ax1: done

vortex-duckdb/src/scan.rs
0ax1: done

vortex-scan/src/iterator.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub fn with_vortex_files<I>(self, vortex_files: I) -> Self
+    where
+        I: IntoIterator<Item = VortexFile>,
+    {
+        for vortex_file in vortex_files.into_iter() {
+            self.files.push(vortex_file);
+        }
+
+        self
+    }
```
joseph-isaacs: can we replace this is a stream of `VortexReadAt`?

0ax1: We could but it seems like that this would make the implementation non-trivially more complex. Prototyped a solution here: https://github.com/vortex-data/vortex/commit/d9f874c4c08e24cd83c090f3abf9df6e1fa4df98. 

In partulcar, driving the stream from diff threasds would require something other than `LocalPool`.

0ax1: I think in context of this PR it's ok to go with the simpler open a file per thread model. But in general I agree, that it'd be nice to drive opening multiple files at the same time whenever the async runtime is making progress.

vortex-duckdb/src/scan.rs
0ax1: This part I'll add tomorrow before landing the PR.

.github/workflows/ci.yml
```diff
@@ -181,7 +181,7 @@ jobs:
             target: wasm32-unknown-unknown
             env:
               rustflags: 'RUSTFLAGS="-A warnings --cfg getrandom_backend=\"wasm_js\""'
-            args: "--target wasm32-unknown-unknown --exclude vortex --exclude vortex-datafusion --exclude vortex-duckdb --exclude vortex-tui --exclude vortex-zstd"
+            args: "--target wasm32-unknown-unknown --exclude vortex --exclude vortex-datafusion --exclude vortex-duckdb --exclude vortex-scan --exclude vortex-tui --exclude vortex-zstd"
```
gatesn: Why doesn't this work in WASM?

vortex-duckdb/Cargo.toml
```diff
@@ -33,6 +33,7 @@ tokio = { workspace = true, features = ["macros", "rt", "rt-multi-thread"] }
 tokio-stream = { workspace = true }
 vortex = { workspace = true, features = ["files"] }
 vortex-file = { workspace = true, features = ["tokio"] }
+vortex-scan = { workspace = true }
```
gatesn: Should get this via vortex crate

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -12,7 +12,7 @@ mod sequence;
 mod temporal;
 mod varbinview;
 
-use cache::*;
+pub use cache::ConversionCache;
```
gatesn, line 15: ?

vortex-duckdb/src/scan.rs
```diff
@@ -54,17 +55,17 @@ impl std::fmt::Debug for VortexBindData {
...
 
 pub struct VortexGlobalData {
-    file_paths: SegQueue<PathBuf>,
-    is_first_file_processed: AtomicBool,
     /// We currently use a conversion cache to cache converted arrays, this id is used to
     /// ensure that each cache created has a unique id used to name those arrays
     conversion_cache_id: AtomicU64,
-    filter_expr: Option<ExprRef>,
-    projection_expr: ExprRef,
+    multi_file_iter: Arc<MultiFileIterator>,
```
gatesn: Can this just be Box<dyn ArrayIterator>?

vortex-duckdb/src/scan.rs
gatesn: You get a cache-per-file already because there's one inside the ArrayExporter. I don't think you need another one outside the exporter?

vortex-duckdb/src/scan.rs
```diff
@@ -210,82 +201,81 @@ impl TableFunction for VortexTableFunction {
...
+            .with_filter_expr(filter_expr);
+
         Ok(VortexGlobalData {
-            file_paths,
-            is_first_file_processed: AtomicBool::new(false),
             conversion_cache_id: AtomicU64::new(0),
-            filter_expr,
-            projection_expr,
+            multi_file_iter: Arc::new(multi_file_iter),
+            thread_count: AtomicUsize::new(0),
```
gatesn: Can you call this `next_thread_id`

vortex-scan/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
+name = "vortex-scan"
+authors.workspace = true
+description = "Multi-file scanning operations for Vortex"
```
gatesn: I think this should be all scanning operations for Vortex. Either keep multi-file in vortex-layout for now, or move everything over (perhaps in a PR beforehand?)

vortex-scan/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+rust-version = { workspace = true }
+version = { workspace = true }
+
+[dependencies]
+vortex-array = { workspace = true }
+vortex-buffer = { workspace = true }
+vortex-dtype = { workspace = true }
+vortex-error = { workspace = true }
+vortex-expr = { workspace = true }
+vortex-file = { workspace = true, features = ["tokio"] }
```
gatesn: Scan cannot depend on vortex-file. We need to design this so the dependency is inverted

vortex-scan/src/iterator.rs
```diff
@@ -0,0 +1,192 @@
...
+use futures::future::BoxFuture;
+use futures::stream::{FuturesUnordered, StreamExt};
+use vortex_array::ArrayRef;
+use vortex_error::VortexResult;
+use vortex_expr::ExprRef;
+use vortex_file::{VortexFile, VortexOpenOptions};
+
+type ArrayFuture = BoxFuture<'static, VortexResult<Option<ArrayRef>>>;
+
+pub struct MultiFileIterator {
```
gatesn: Docs docs docs

vortex-scan/src/iterator.rs
```diff
@@ -0,0 +1,192 @@
...
+    files: SegQueue<VortexFile>,
+    filter_expr: Option<ExprRef>,
+    local_pools: DashMap<usize, LocalPool>,
+    paths: SegQueue<PathBuf>,
```
gatesn: I think this needs to be more like `Box<dyn FnOnce() -> ScanBuilder<ArrayRef>>`, or perhaps even `-> Arc<dyn LayoutReader>`, that way we remove the dependency on vortex-file

vortex-scan/src/iterator.rs
```diff
@@ -0,0 +1,192 @@
...
+        self.projection_expr = projection_expr;
+        self
+    }
+
+    pub fn with_filter_expr(mut self, filter_expr: Option<ExprRef>) -> Self {
+        self.filter_expr = filter_expr;
+        self
+    }
+
+    pub fn push_scan_task<I, F>(&self, thread_id: usize, items: I)
```
gatesn: Don't think this should be pub

vortex-scan/src/iterator.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl MultiFileIterator {
+    /// `next` is not implemented in terms of `impl Iterator` as `self`
+    /// needs to be immutable in order to be shared across threads.
+    pub fn next(&self, thread_id: usize) -> Option<VortexResult<ArrayRef>> {
```
gatesn: Hmmm, this is a bit non-canonical.

How about something like

```rust
struct MultiScan {
  ...
}

// Configure the scan
impl MultiScan {
  pub fn with_source(self, src: Box<dyn FnOnce() -> VortexResult<Arc<dyn LayoutReader>>>) -> Self { .. }

  pub fn with_projection(self, expr: Expr) -> Self { .. }

  pub fn into_iterator(self) -> MultiScanIterator;
}

struct MultiScanIterator {
  state: Arc<MultiScanState>,
  instance_id: AtomicUsize,
}

// Create a new iterator to drive the scan
impl Clone for MultiScanIterator {
  fn clone(&self) -> Self {
    let instance_id = self.state.new_driver();
    Self {
      instance_id,
      state: self.state.clone()
    }
  }
}

impl Iterator for MultiScanIterator { .. }

impl ArrayIterator for MultiScanIterator { .. }
```

gatesn (changes requested):

.github/workflows/ci.yml
robert3005: I tried to figure this out yesterday, there's something wrong with the dependencies declared in vortex-file that make taking dependency on it from another crate enable tokio features that are not supposed by wasm. However, after spending 30 mintues poking around build graph I couldn't figure out where was it going wrong.

vortex-duckdb/src/scan.rs
0ax1: This is about the conversion cache we pass to `ArrayExporter` within scan.

```
 local_state.exporter = Some(ArrayExporter::try_new(
                    &array_result?.to_struct()?,
                    local_state
                        .cache
                        .as_mut()
                        .vortex_expect("cache should exist"),
                )?);
```

So the `ArrayExporter` has its own cache but it's passed when calling `try_new`.

0ax1: It's the `ArrayIteratorExporter` constructing its own cache.

vortex-duckdb/src/exporter/mod.rs
0ax1: https://github.com/vortex-data/vortex/pull/3863#discussion_r2207074488

vortex-duckdb/src/scan.rs
0ax1: Strictly speaking I guess this could even be just `MultiFileIterator`.

vortex-scan/Cargo.toml
0ax1: Moved all scan code into vortex-scan.

vortex-scan/src/iterator.rs
0ax1: Agreed, I'll give this a cleanup pass.

vortex-duckdb/Cargo.toml
0ax1: done

vortex-datafusion/Cargo.toml
```diff
@@ -27,6 +27,7 @@ tokio = { workspace = true, features = ["rt-multi-thread", "fs"] }
 tokio-stream = { workspace = true }
 tracing = { workspace = true, optional = true }
 vortex = { workspace = true, features = ["object_store"] }
+vortex-scan = { workspace = true, features = ["tokio"] }
```
gatesn: Can access this via vortex crate, may need to enable tokio feature on that

vortex-scan/src/iterator.rs
```diff
@@ -0,0 +1,138 @@
...
+            match result {
+                Some(Ok(array)) => return Some(Ok(array)),
+                Some(Err(e)) => return Some(Err(e)),
+                None => continue, // Try next batch of futures
+            }
+        }
+    }
+}
+
+unsafe impl Send for MultiFileIterator {}
```
gatesn: Remove these, I think they should be derived automatically

vortex-duckdb/.gitignore
```diff
@@ -5,3 +5,4 @@ src/cpp.rs
 
 # Symlinked DuckDB source is ignored. It is downloaded, extracted and symlinked as part of build.rs
 duckdb
+!src/duckdb
```
gatesn, line 8: Is it better the make the path above more explicit than try and negate this one?

vortex-duckdb/src/scan.rs
```diff
@@ -210,82 +199,91 @@ impl TableFunction for VortexTableFunction {
...
 
     fn init_global(init_input: &TableInitInput<Self>) -> VortexResult<Self::GlobalState> {
         let bind_data = init_input.bind_data();
-        let file_paths = create_file_paths_queue(bind_data);
         let projection_expr = extract_projection_expr(init_input);
         let filter_expr = extract_table_filter_expr(init_input, init_input.column_ids())?;
+        // let num_threads = thread::available_parallelism().map(|p| p.get())?;
+        let is_first_file_queued = Arc::new(std::sync::atomic::AtomicBool::new(false));
+
+        let closures = bind_data.file_paths.clone().into_iter().map(move |path| {
```
gatesn, line 249: Why is this cloned?

vortex-duckdb/src/scan.rs
```diff
@@ -210,82 +199,91 @@ impl TableFunction for VortexTableFunction {
...
+            let filter_expr = filter_expr.clone();
+            let projection_expr = projection_expr.clone();
+            let is_first_file_queued = is_first_file_queued.clone();
+
+            move || {
+                let file = if !is_first_file_queued.swap(true, SeqCst) {
+                    first_file
+                } else {
+                    VortexOpenOptions::file()
+                        .open_blocking(&path)
```
gatesn, line 262: Hmmm, maybe it should return a `Future<LayoutReader>`

vortex-duckdb/src/scan.rs
```diff
@@ -210,82 +199,91 @@ impl TableFunction for VortexTableFunction {
...
+        let is_first_file_queued = Arc::new(std::sync::atomic::AtomicBool::new(false));
+
+        let closures = bind_data.file_paths.clone().into_iter().map(move |path| {
+            let first_file = bind_data.first_file.clone();
+            let filter_expr = filter_expr.clone();
+            let projection_expr = projection_expr.clone();
+            let is_first_file_queued = is_first_file_queued.clone();
+
+            move || {
+                let file = if !is_first_file_queued.swap(true, SeqCst) {
```
gatesn, line 256: This is a bit odd.. Since we _must_ have at least one, maybe have MultiScan take the first one in `new`, and the later ones in `with_extra_scans`

vortex-python/Cargo.toml
```diff
@@ -38,3 +38,4 @@ pyo3-log = { workspace = true }
 tokio = { workspace = true, features = ["fs", "rt-multi-thread"] }
 url = { workspace = true }
 vortex = { workspace = true, features = ["object_store", "python"] }
+vortex-scan = { workspace = true, features = ["tokio"] }
```
gatesn: Remove this, access via vortex dependency with tokio feature

vortex-scan/src/multi_scan.rs
```diff
@@ -0,0 +1,126 @@
...
+use crate::ScanBuilder;
+
+type ArrayFuture = BoxFuture<'static, VortexResult<Option<ArrayRef>>>;
+
+/// Coordinator to orchestrate multiple scan operations.
+///
+/// `MultiScan` allows to queue multiple scan operations in order to execute
+/// them in parallel. In particular, this enables scanning multiple files.
+pub struct MultiScan {
+    scan_builder_fns: Arc<SegQueue<Box<dyn FnOnce() -> ScanBuilder<ArrayRef> + Send + Sync>>>,
```
gatesn: Extract a type alias for the ScanBuilderFactory.

vortex-scan/src/multi_scan.rs
```diff
@@ -0,0 +1,126 @@
...
+                    let split_tasks = scan_builder_fn().build().ok()?.1;
+                    for task in split_tasks {
+                        self.task_queue.push(Box::pin(task));
+                    }
+                }
+                // TODO(Alex): worksteal tasks from other threads
+            }
+
+            // Poll one future at a time. Polling multiple futures at
+            // the same time leads to contention within a layout reader.
```
gatesn: Why would it?

vortex-duckdb/src/scan.rs
```diff
@@ -54,17 +54,16 @@ impl std::fmt::Debug for VortexBindData {
...
+    cache_id: AtomicU64,
 }
 
 pub struct VortexLocalData {
-    exporter: Option<ArrayIteratorExporter>,
+    multi_scan_iterator: MultiScanIterator,
+    exporter: Option<ArrayExporter>,
+
+    // TODO(Alex): replace with global conversion cache
+    conversion_cache: ConversionCache,
```
gatesn, line 66: Just remove this. We know we need to plumb it through MultiScan as data attached to an individual scan task.

vortex-duckdb/src/scan.rs
0ax1: The closure wants to own the `PathBuf`.

vortex-duckdb/.gitignore
0ax1: Pretty sure I tried `./duckdb` for the symlink which didn't work.

vortex-duckdb/src/scan.rs
0ax1: Do we really though? That seems like an artificial constraint. What if a user calls into vortex-duckdb-scan with e.g. a wrong glob that doesn't match any files. Shouldn't that still work but return no results?

vortex-duckdb/src/scan.rs
0ax1: I need some cache to live somewhere for the time being as this is passed to 
```
                local_state.exporter = Some(ArrayExporter::try_new(
                    &array_result?.to_struct()?,
                    &mut local_state.conversion_cache,
                )?);
```
Just passing the newly created cache directly to `&mut local_state.conversion_cache` crashes.

vortex-duckdb/src/scan.rs
0ax1: What's the benefit? I kinda conceptually like passing a *Scan*Builder to the scan module/crate. 🤷

vortex-scan/src/multi_scan.rs
0ax1: That's the only way how I could explain a slowdown for some queries when increasing the future count to be polled concurrently.

vortex-duckdb/src/scan.rs
0ax1: I think the really implicit thing here is that the first `path` is skipped, as we already opened the file for it in bind.

vortex-scan/src/multi_scan.rs
0ax1: Happy to drop the comment.

vortex-duckdb/src/scan.rs
gatesn: Not really, because we _need_ to know the DType up front. So if your glob matches no files, then you cannot even create a MultiScan, you would have to just error or return empty vector of no fields.

vortex-tui/Cargo.toml
```diff
@@ -27,6 +27,7 @@ taffy = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread"] }
 vortex = { workspace = true, features = ["tokio", "parquet"] }
 vortex-layout = { workspace = true }
+vortex-scan = { workspace = true, features = ["tokio"] }
```
gatesn: Remove this, and also vortex-layout

gatesn (approved):
It's a reasonable place to start. I think once we add work-stealing it may be the case that we actually unify Scan and MultiScan, it may also be the case that we want better support for async. e.g. we have a blocking call to open files at the moment.

PR #3866, opened 2025-07-14
feat: display

### LINKED ISSUE #3821: Array formatting

danking:
Closes https://github.com/vortex-data/vortex/issues/3821

### DIFF
--- a/vortex-array/src/array/display/mod.rs
+++ b/vortex-array/src/array/display/mod.rs
@@ -1,9 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+mod tree;
+
 use std::fmt::Display;
 
 use itertools::Itertools as _;
+use tree::TreeDisplayWrapper;
 use vortex_error::VortexExpect as _;
 
 use crate::Array;
@@ -17,10 +20,50 @@ impl Display for DisplayArray<'_> {
 }
 
 pub enum DisplayOptions {
-    /// `EncodingId(dtype, len=123)`
+    /// ```
+    /// use vortex_array::display::DisplayOptions;
+    /// use vortex_array::IntoArray;
+    /// use vortex_buffer::buffer;
+    ///
+    /// let array = buffer![0_i16, 1, 2, 3, 4].into_array();
+    /// let opts = DisplayOptions::MetadataOnly;
+    /// assert_eq!(
+    ///     format!("{}", array.display_as(&opts)),
+    ///     "vortex.primitive(i16, len=5)",
+    /// );
+    /// ```
     MetadataOnly,
-    /// `[1i32, 2i32, 3i32]`
+    /// ```
+    /// use vortex_array::display::DisplayOptions;
+    /// use vortex_array::IntoArray;
+    /// use vortex_buffer::buffer;
+    ///
+    /// let array = buffer![0_i16, 1, 2, 3, 4].into_array();
+    /// let opts = DisplayOptions::default();
+    /// assert_eq!(
+    ///     format!("{}", array.display_as(&opts)),
+    ///     "[0i16, 1i16, 2i16, 3i16, 4i16]",
+    /// );
+    /// assert_eq!(
+    ///     format!("{}", array.display_as(&opts)),
+    ///     format!("{}", array.display()),
+    /// );
+    /// ```
     CommaSeparatedScalars { space_after_comma: bool },
+    /// ```
+    /// use vortex_array::display::DisplayOptions;
+    /// use vortex_array::IntoArray;
+    /// use vortex_buffer::buffer;
+    ///
+    /// let array = buffer![0_i16, 1, 2, 3, 4].into_array();
+    /// let opts = DisplayOptions::TreeDisplay;
+    /// let expected = "root: vortex.primitive(i16, len=5) nbytes=10 B (100.00%)
+    ///   metadata: EmptyMetadata
+    ///   buffer (align=2): 10 B (100.00%)
+    /// ";
+    /// assert_eq!(format!("{}", array.display_as(&opts)), expected);
+    /// ```
+    TreeDisplay,
 }
 
 impl Default for DisplayOptions {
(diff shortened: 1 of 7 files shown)


vortex-array/src/array/display/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+
+impl Display for DisplayArrayAs<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt_as(f, self.1)
+    }
+}
+
+impl Display for dyn Array + '_ {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.fmt_as(f, &DisplayOptions::MetadataOnly)
```
blaginin, line 136: i think normally, when you print hashmap / list / etc you get the data? Should it really be metadata here?

danking: @gatesn thoughts? This was how `develop` currently displays an array.

danking:
OK, I'm done fiddling with doc strings.

vortex-array/src/array/display/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+            }
+            DisplayOptions::CommaSeparatedScalars { space_after_comma } => {
+                write!(f, "[")?;
+                let sep = if *space_after_comma { ", " } else { "," };
+                write!(
+                    f,
+                    "{}",
+                    (0..self.len())
+                        .map(|i| self.scalar_at(i).vortex_expect("index is in bounds"))
+                        .format(sep)
```
robert3005, line 216: Can this do some minimum truncation so we don't accidentally overload the output?

robert3005: Arrow has https://github.com/apache/arrow-rs/blob/52fd59c9f05ad8c8dbb62a4ef82a03daf453f1b6/arrow-array/src/array/mod.rs#L897 but that's not public

robert3005 (approved):

vortex-array/src/array/display/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+/// # use vortex_buffer::buffer;
+/// let array = buffer![0_i16, 1, 2, 3, 4].into_array();
+/// assert_eq!(
+///     format!("{}", array.display()),
+///     "[0i16, 1i16, 2i16, 3i16, 4i16]",
+/// )
+/// ```
+///
+/// See also:
+/// [Array::display](http://localhost:8000/doc/vortex_array/trait.Array.html#method.display),
```
gatesn, line 33: We'll need to figure out the relative URL here

danking:
YOLO

vortex-array/src/array/display/mod.rs
gatesn: Yeah, I think compact is better by default

vortex-array/src/array/display/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// # use vortex_array::IntoArray;
+    /// # use vortex_buffer::buffer;
+    /// let array = buffer![0_i16, 1, 2, 3, 4].into_array();
+    /// assert_eq!(
+    ///     format!("{}", array.display()),
+    ///     "[0i16, 1i16, 2i16, 3i16, 4i16]",
+    /// )
+    /// ```
+    pub fn display(&self) -> DisplayArray {
+        DisplayArray(self)
```
gatesn, line 155: Why is this called `display`? When it's different from `impl Display`? It's quite confusing. I would just use display_as always

vortex-array/src/array/display/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// assert_eq!(format!("{}", array.display_as(&opts)), expected);
+    /// ```
+    pub fn tree_display(&self) -> impl Display {
+        TreeDisplayWrapper(self.to_array())
+    }
+
+    /// Format an array as specified by the options.
+    ///
+    /// See [DisplayOptions] for examples.
+    pub fn fmt_as(
```
gatesn, line 193: Why have `display_as` _and_ `fmt_as`?

vortex-array/src/array/display/mod.rs
gatesn: Yup, the options should have a default max number of values

gatesn: Also implementing this via scalar_at is kinda wild... we should warn how insanely inefficient it is. Maybe even `#cfg(debug)`

vortex-array/src/array/display/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// # use vortex_buffer::buffer;
+    /// let array = buffer![0_i16, 1, 2, 3, 4].into_array();
+    /// let opts = DisplayOptions::TreeDisplay;
+    /// let expected = "root: vortex.primitive(i16, len=5) nbytes=10 B (100.00%)
+    ///   metadata: EmptyMetadata
+    ///   buffer (align=2): 10 B (100.00%)
+    /// ";
+    /// assert_eq!(format!("{}", array.display_as(&opts)), expected);
+    /// ```
+    pub fn tree_display(&self) -> impl Display {
```
gatesn, line 186: Similarly we should remove this in favor of `display_as`

vortex-array/src/array/display/mod.rs
robert3005: In the end if you want the array you likely want the whole array. Sure you can make it idiot proof. Instead of adding a feature flag we can make it faster if its necessary

PR #3867, opened 2025-07-14
fix[layout]: row id layout reader cannot stop if inner layout registering its splits

joseph-isaacs:


vortex-layout/src/layouts/row_idx/mod.rs
```diff
@@ -131,12 +131,7 @@ impl LayoutReader for RowIdxLayoutReader {
         row_offset: u64,
         splits: &mut BTreeSet<u64>,
     ) -> VortexResult<()> {
-        // Since RowIdx isn't a field, we only need to register splits for the child layout
-        // if there are any fields in the mask at all.
-        if !field_mask.is_empty() {
-            self.child.register_splits(field_mask, row_offset, splits)?;
-        }
-        Ok(())
```
a10y, line 139: can you say more why this is wrong?

robert3005 (approved):

robert3005:
But probably need to have a test for empty projection and no filter with row id

vortex-layout/src/layouts/row_idx/mod.rs
joseph-isaacs: Cause we never ask the child layout for its splits

robert3005: Empty mask doesn't mean we reading no rows, just that the rows have 0 fields/cells

PR #3869, opened 2025-07-14
Fix intra-doc links

gatesn:
This isn't yet supported by Cargo doc, see https://github.com/rust-lang/rust/issues/74563#issuecomment-702499503


danking (approved):
🙈

PR #3870, opened 2025-07-14
chore[bench]: move tpcds into new benchmark format

joseph-isaacs:


bench-vortex/src/bin/query_bench.rs
```diff
@@ -123,6 +127,26 @@ struct TpcHArgs {
     scale_factor: String,
 }
 
+#[derive(Parser, Debug)]
+struct TpcDSArgs {
+    #[command(flatten)]
+    common: CommonArgs,
+
+    #[arg(long, value_delimiter = ',', value_parser = value_parser!(Target),
+        default_values = vec![
```
0ax1, line 136: I guess we could at some point consider a global const array of targets shared across the diff bench types.

0ax1 (approved):

PR #3871, opened 2025-07-14
chore: Consitent naming for display functions

robert3005:


encodings/dict/src/display.rs
```diff
@@ -20,7 +21,10 @@ mod test {
         .into_array();
 
         assert_eq!(
-            x.display().to_string(),
+            x.display_as(DisplayOptions::CommaSeparatedScalars {
+                space_after_comma: true
```
gatesn: I'm afraid Dan is in the minority here. This should be the default

fuzz/fuzz_targets/array_ops.rs
```diff
@@ -82,7 +83,7 @@ fuzz_target!(|fuzz_action: FuzzArrayAction| -> Corpus {
                 if let Err(e) = assert_array_eq(&expected.array(), &cast_result, i) {
                     vortex_panic!(
                         "Failed to cast {} to dtype {to}\nError: {e}",
-                        current_array.tree_display()
+                        current_array.display_as(DisplayOptions::TreeDisplay)
```
gatesn: I guess I would allow display_tree() just so it's adjacent in the auto-complete

vortex-array/src/array/display/mod.rs
```diff
@@ -12,40 +12,10 @@ use vortex_error::VortexExpect as _;
...
-        self.0.fmt_as(f, &DisplayOptions::default())
-    }
-}
-
 /// Describe how to convert an array to a string.
 ///
 /// See also:
-/// [Array::display](http://localhost:8000/doc/vortex_array/trait.Array.html#method.display),
-/// [Array::display_as](http://localhost:8000/doc/vortex_array/trait.Array.html#method.display_as),
+/// [Array::display_as](./trait.Array.html#method.display_as),
```
gatesn: Have you tested? This is in the display module, would have thought this was ../

fuzz/fuzz_targets/array_ops.rs
robert3005: you're killing me

vortex-array/src/array/display/mod.rs
robert3005: I'm guessing tbh

encodings/dict/src/display.rs
robert3005: What do you mean? That the value display should be the default display?

vortex-array/src/array/display/mod.rs
robert3005: Yes, this is correct

gatesn (approved):

PR #3873, opened 2025-07-14
fix: Fuzzer does not generate filter on struct

a10y:
Fix the fuzzer to not try and generate compare ops for struct fields.


joseph-isaacs:
Before the fuzzer found a bug when using any possible field name. 

I think we shouldn't change it to this.

robert3005:
I think we shouldn’t change this. It is a good check whether we do an incorrect assumption about field names

a10y:
Alright, I can revert the name change part

robert3005 (approved):
Yay, I was a bit too lazy to do this

PR #3879, opened 2025-07-15
[Experimental] Use error-sets crate

AdamGS:
Lack of backtraces is probably a blocker, but there's an [issue](https://github.com/mcmah309/error_set/issues/12) talking/describing it, so hopefully it'll be available soon?

There's a second step here of narrowing the error types in places where returning the top-level one doesn't make sense, but I think I'm done for today.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2243,6 +2243,28 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


robert3005:
I would revisit this when there's backtrace support. Right now I like my backtraces

PR #3881, opened 2025-07-16
feat: build with stable rust

### LINKED ISSUE #3546: Support building on Stable Rust

a10y:
Fixes #3546 

MSRV is 1.87.0.

Stable is far ahead of the February 2025 nightly we've been pegged to, so on top of the regular changes there were a ton of new lints to go whack-a-mole.

I left rustfmt with its unstable options, which means that you'll need to continue to run rustfmt on the nightly channel but the project will be on the stable channel. You can run formatting from CLI like this

```
cargo +nightly fmt
```

For RustRover you can configure rustfmt to execute with nightly in settings:

image

For VSCode and others you'll need to figure that one out.


.github/actions/setup-rust/action.yml
```diff
@@ -24,7 +24,7 @@ runs:
       with:
         toolchain: "${{ steps.rust-version.outputs.version }}"
         targets: "${{inputs.targets || ''}}"
-        components: clippy, rustfmt, miri, llvm-tools-preview
+        components: clippy, rustfmt
```
a10y: gotta flag these on when we're using nightly

encodings/fastlanes/src/lib.rs
```diff
@@ -3,7 +3,6 @@
 
 #![allow(incomplete_features)]
```
robert3005: should delete incomplete features

vortex-buffer/src/trusted_len.rs
```diff
@@ -0,0 +1,61 @@
+//  SPDX-License-Identifier: Apache-2.0
+//  SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+/// Trait for all types which have a known upper-bound.
+///
+/// Functions that receive a `TrustedLen` iterator can assume that it's `size_hint` is exact,
```
a10y, line 6: Technically this is a stricter bound than the std TrustedLen, which just guarantees that the upper bound is either `Some(exact)` or `None` if unknown. This is what we want for our `extend_trusted` though.

The nice part about defining our version of this trait is that we can leave `extend_trusted` as a safe method on BufferMut, which should never fail at runtime. If we ever wanna add a new case we just add a new impl here. But I don't expect many more if any.

vortex-buffer/src/trusted_len.rs
```diff
@@ -0,0 +1,61 @@
...
+/// # Safety
+///
+/// The type which implements this trait must provide an exact `Some` upper-bound for its
+/// `size_hint` method. Failure to do so can trigger undefined behavior in users of the trait.
+pub unsafe trait TrustedLen: Iterator {}
+
+macro_rules! impl_for_range {
+    ($($typ:ty),*) => {
+        $(
+            unsafe impl TrustedLen for std::ops::Range<$typ> {}
```
a10y, line 19: I can't just do `impl<T> for Range<T>` because they introduce a type-bound on _another_ unstable `TrustedStep` trait...which is really just implemented for the integer types anyway

vortex-dtype/src/struct_.rs
```diff
@@ -304,7 +304,7 @@ impl StructFields {
             .names
             .iter()
             .enumerate()
-            .filter(|&(i, _)| (i != index))
+            .filter(|&(i, _)| i != index)
```
a10y, line 307: lints caught when i switched compilers, dunno why nightly clippy wouldn't be a superset of the stable clippy errors 🤷

vortex-error/src/lib.rs
```diff
@@ -68,71 +68,74 @@ impl From<Infallible> for VortexError {
     }
 }
 
+const _: () = {
+    assert!(size_of::<VortexError>() < 128);
+};
```
a10y, line 73: One lint that popped off when I switched to stable was that apparently with all features activated, VortexError was a whopping 140 bytes, which was tripping the large_err check.

I boxed the Backtraces to keep the size down, now it's around ~88B. Still large but better.

vortex-array/build.rs
```diff
@@ -1,10 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: Copyright the Vortex contributors
-
-#[rustversion::nightly]
```
a10y, line 4: This file is now superfluous, instead I opted to do what Tokio does when you wanna unlock unstable nightly-only things and just add a new cfg `vortex_nightly` that gates all of the portable_simd stuff.

To run it, you can use a nigthly compiler and pass the flags, e.g.

```
RUSTFLAGS="--cfg vortex_nightly" cargo +nightly build -p vortex-array --all-features
```

vortex-buffer/src/buffer.rs
```diff
@@ -177,8 +177,10 @@ impl<T> Buffer<T> {
     }
 
     /// Returns an iterator over the buffer of elements of type T.
-    pub fn iter(&self) -> impl Iterator<Item = &T> + '_ {
-        self.as_slice().iter()
+    pub fn iter(&self) -> Iter<'_, T> {
```
a10y, line 180: I think this technically isn't a breaking change actually, b/c we're going from unnamed -> named type.

But the idea is that we have to impl `TrustedLen` for this, so you can BufferMut::extend_trusted from another Buffer.

So I just defined a concrete type that wraps std::slice::Iter and delegates the important methods to it.

vortex-buffer/src/buffer_mut.rs
```diff
@@ -418,10 +418,66 @@ impl<T> AsMut<[T]> for BufferMut<T> {
     }
 }
 
+impl<T> BufferMut<T> {
```
a10y, line 421: I didn't change any logic here, this is just moving methods around

vortex-error/src/lib.rs
robert3005: we should aggressively trim it down

a10y: Yes but this is not new, the lint is just new.

a10y: Filed #3892 as a follow up

vortex-dtype/src/struct_.rs
a10y: i forgot that our "nightly" version is pinned to 6 months ago, lol

a10y:
Clickbench/TPC-H seem relatively unchanged, some things got up to 20% faster but I think that could just be noise. One of the compression benchmarks got 18% slower, I retriggered to see if that's real or not

vortex-array/src/builders/varbinview.rs
```diff
@@ -107,14 +107,14 @@ impl VarBinViewBuilder {
...
-        views: impl IntoIterator<Item = BinaryView>,
+        buffer: &[ByteBuffer],
+        views: &Buffer<BinaryView>,
         validity_mask: Mask,
     ) {
         self.flush_in_progress();
 
-        self.completed.extend(buffer);
-        self.views_builder.extend(views);
+        self.completed.extend(buffer.iter().cloned());
```
a10y, line 116: this should be an extend_trusted

a10y:
It looks like that one slowdown persisted, I can't replicate it though (at least not on M1)

develop:

image


PR:

image

vortex-array/src/builders/varbinview.rs
a10y: ah nvm, this is a Vec not a BufferMut

robert3005:
Is this one of the cases where we call extend instead of extendtrusted?

robert3005:
I found that the extend changes didn’t manifest on macOS but would on Linux

a10y:
I just ran the experiment again on a c5.12xlarge and the results were similar to mac, if anything slightly faster on the PR branch

image

robert3005:
Ok, then let’s assume this a flake and move on

.github/workflows/ci.yml
```diff
@@ -364,14 +364,17 @@ jobs:
...
+          components: "rust-src, rustfmt, clippy, llvm-tools-preview"
       - name: Install nextest
         uses: taiki-e/install-action@v2
         with:
           tool: nextest@0.9.98
       - name: Rust Tests
         # vortex-duckdb-ext currently fails linking for cargo test targets.
         run: |
-          cargo nextest run --locked --workspace --all-features --no-fail-fast --target x86_64-unknown-linux-gnu
+          cargo +nightly nextest run --locked --workspace --all-features --no-fail-fast --target x86_64-unknown-linux-gnu
```
robert3005, line 377: interesting that this uses nightly, I thought nextest would be fine on stable

a10y: This is only for the Rust testing with ASAN job. That requires unstable `-Z sanitizer` flag

robert3005 (approved):
part of me thinks we shouldn't raise MSRV but part of me thinks maybe we can?

a10y:
I think we can probably bump it whenever we want/whenever we need new features

PR #3891, opened 2025-07-16
fix: Include size of self in nbytes

### LINKED ISSUE #3885: Rename `nbytes` and improve docs
Right now, when calculating `nbytes`, unlike in arrow, we only sum the buffers but not the type size. As a result, buffer less types now have zero size:

```rust
let array =
    vortex::encodings::sequence::SequenceArray::typed_new(2i8, 3, Nullability::NonNullable, 4)
        .unwrap();

assert_eq!(array.nbytes(), 0);
```

See the discussion below: right now that's the desired behavior, but the method name / docs / tests should reflect that.

robert3005:
This is definitely an overestimation but should never be an underestimation. fix #3885 


robert3005:
Ok we shouldn't do this, the caller can do it themselves

PR #3894, opened 2025-07-16
fix: `DictMaskEvaluation` to respect validity

blaginin:
Fixes https://github.com/vortex-data/vortex/actions/runs/16310529975/job/46065222914


robert3005 (approved):
Oh this seems annoying. Thanks

PR #3899, opened 2025-07-17
Remove async API for scanning

gatesn:
We only know of DataFusion that intermingles I/O and CPU work on the same Tokio runtime, so we still expose the ability to do that but the implementation has moved to vortex-datafusion.

Some things to note:
* I/O is still all performed on an I/O dispatcher, which by default runs all I/O on a process-wide tokio runtime.
* VortexWriteOptions has a bad API now that also intermingles CPU and I/O work in the same async function. So we need to fix this up. i.e. it should be a blocking call when the underlying I/O is blocking.

All other implementations should use:
* `into_array_iter()` - blocking, single threaded
* `into_array_thread_pool_iter(ThreadPool)` - blocking, multi-threaded

--

*NOTE* - there is a slight performance regression in this PR that will be addressed by work-stealing iterators


vortex-datafusion/src/persistent/opener.rs
```diff
@@ -150,7 +161,7 @@ impl FileOpener for VortexFileOpener {
                             .map(Ok),
                     )
                 })
-                .map_err(|e: VortexError| ArrowError::ExternalError(Box::new(e)))
+                //.map_err(|e: VortexError| ArrowError::ExternalError(Box::new(e)))
```
AdamGS: dead code

vortex-file/src/tests.rs
```diff
@@ -290,25 +280,23 @@ async fn unequal_batches() {
...
             .unwrap()
-            .into_array_stream()
+            .into_array_iter()
             .unwrap()
     );
 
     let mut item_count = 0;
 
-    while let Some(array) = stream.next().await {
+    for array in stream.by_ref() {
```
AdamGS: rename to `iter` or something?

AdamGS: there are a bunch of those

vortex-scan/src/lib.rs
```diff
@@ -227,16 +199,54 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
...
-        Ok((dtype, split_tasks))
+        Ok(split_tasks)
+    }
+
+    /// Returns a blocking iterator over the scan.
+    ///
+    /// All work will be performed on the current thread, with tasks interleaved per the
+    /// configured concurrency. Any configured executor will be ignored.
+    #[allow(clippy::should_implement_trait)]
+    pub fn into_iter(self) -> VortexResult<impl Iterator<Item = VortexResult<A>> + 'static> {
```
AdamGS, line 213: this is so ugly 😢

vortex-scan/src/lib.rs
```diff
@@ -227,16 +199,54 @@ impl<A: 'static + Send + Sync> ScanBuilder<A> {
...
+
+        let split_tasks = self.build()?;
+        let mut stream = stream::iter(split_tasks)
+            .map(move |task| {
+                spawner
+                    .spawn_local_with_handle(task)
+                    .map_err(|e| vortex_err!("Failed to spawn task: {e}"))
+                    .vortex_expect("Failed to spawn task")
+            })
+            .buffered(2)
```
AdamGS, line 228: should respect some config from `self`?

vortex-scan/src/lib.rs
robert3005: But the clippy warning is real - why not fix the clippy warning and implement IntoIter trait?

vortex-scan/src/multi_scan.rs
```diff
@@ -83,7 +83,9 @@ impl Iterator for MultiScanIterator {
             // Queue up tasks if the thread local queue is almost empty.
             if self.task_queue.len() <= 4 {
                 if let Some(scan_builder_fn) = self.scan_builder_factory.pop() {
-                    let split_tasks = scan_builder_fn().build().ok()?.1;
+                    let scan_builder = scan_builder_fn();
+                    // FIXME(ngates): this should not call ok()
+                    let split_tasks = scan_builder.build().ok()?;
                     for task in split_tasks {
                         self.task_queue.push(Box::pin(task));
                     }
```
robert3005:
```suggestion
                    let split_tasks = scan_builder_fn().build();
                    match split_tasks {
                        Ok(tasks) => {
                            for task in tasks {
                                self.task_queue.push(Box::pin(task));
                            }
                        }
                        Err(e) => return Some(Err(e)),
                    }
```

vortex-scan/src/lib.rs
gatesn: Can't return result from that trait

PR #3908, opened 2025-07-17
chore: exclude arbitrary from coverage

lwwmanning:


AdamGS (approved):
There are a bunch of other arbitrary modules IIRC

PR #3909, opened 2025-07-17
feat: impl optimize for more arrays

a10y:
Implements `optimize` for RunEnd, Patches and SparseArray by shifting the offset out of the indices. For ListArray and StructArray, it recursively optimizes the children.

This augments #3852 

Signed-off-by: Andrew Duffy <andrew@a10y.dev>Signed-off-by: Andrew Duffy <andrew@a10y.dev>Signed-off-by: Andrew Duffy <andrew@a10y.dev>

### DIFF
--- a/vortex-array/src/arrays/chunked/ops.rs
+++ b/vortex-array/src/arrays/chunked/ops.rs
@@ -53,7 +53,10 @@ impl OperationsVTable<ChunkedVTable> for ChunkedVTable {
             .map(|chunk| chunk.optimize())
             .collect::<VortexResult<Vec<_>>>()?;
 
-        ChunkedArray::try_new(optimized_chunks, array.dtype().clone())
+        Ok(ChunkedArray::new_unchecked(
+            optimized_chunks,
+            array.dtype().clone(),
+        ))
     }
 }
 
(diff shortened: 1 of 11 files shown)


vortex-expr/src/forms/nnf.rs
```diff
@@ -136,9 +136,8 @@ impl FolderMut for NNFVisitor {
 
 #[cfg(test)]
 mod tests {
-    use crate::{and, lit, or};
-
     use super::*;
+    use crate::{and, lit, or};
```
robert3005, line 140: hm, this is annoying that we have to change formatting after not using nightly toolchain

robert3005 (approved):
We really should figure out how to take owned/mutable arrays so we don't have to copy every time

vortex-array/src/arrays/chunked/ops.rs
```diff
@@ -45,6 +45,16 @@ impl OperationsVTable<ChunkedVTable> for ChunkedVTable {
...
     }
+
+    fn optimize(array: &ChunkedArray) -> VortexResult<ChunkedArray> {
+        let optimized_chunks = array
+            .chunks()
+            .iter()
+            .map(|chunk| chunk.optimize())
+            .collect::<VortexResult<Vec<_>>>()?;
+
+        ChunkedArray::try_new(optimized_chunks, array.dtype().clone())
```
robert3005: new_unchecked?

a10y: didn't know that was a thing now

robert3005: we have it in some places for cases where we know the operations must preserve the properties of the array

PR #3918, opened 2025-07-18
fix: mitigate duckdb threading issue

0ax1:
Set to 1 until we figure out the DuckDB threading issue which can lead to incorrect results.


vortex-duckdb/cpp/table_function.cpp
```diff
@@ -50,7 +50,7 @@ struct CTableGlobalData final : GlobalTableFunctionState {
     unique_ptr<vortex::CData> ffi_data;
 
     idx_t MaxThreads() const override {
-        return GlobalTableFunctionState::MAX_THREADS;
+        return 1;
```
joseph-isaacs, line 53: shall we hook this into the rust ext?

0ax1: probably better to fix forward if we can than landing this

PR #3926, opened 2025-07-18
feat: Refactor traversal logic

AdamGS:
The main rationale for this change is that I would like to use this logic elsewhere, but it wasn't touched in a while and needed some polishing.

I've moved most of the code that deals with actually applying operations in a certain order into the default implementation of `Node::accept`, `Node::rewrite` and the `transform_` functions.
A future user of `Node` will only need to fill in the structure of their tree, getting the fiddly bits OOTB.

This is not a fully-fledged solution for any possible usecase, but its an attempt at tightening the existing abstractions and leaving something that could be extended to new or more advanced patterns.


delta003:
Is an abstract tree traversal module something we want Vortex to expose? Feels a bit random.

vortex-expr/src/traversal/mod.rs
```diff
@@ -1,292 +1,361 @@
...
-/// There are control traversal controls `TraversalOrder`:
-/// - `Skip`: Skip visiting the children of the current node.
-/// - `Stop`: Stop visiting any more nodes in the traversal.
-/// - `Continue`: Continue with the traversal as expected.
+/// Signal to control a traversal's flow
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum TraversalOrder {
     /// In a top-down traversal, skip visiting the children of the current node.
-    /// In the bottom-up phase of the traversal this does nothing (for now).
+    /// In the bottom-up phase of the traversal, skip the next
```
joseph-isaacs: the next what?

vortex-expr/src/traversal/mod.rs
```diff
@@ -1,292 +1,361 @@
...
     Skip,
-
     /// Stop visiting any more nodes in the traversal.
     Stop,
-
     /// Continue with the traversal as expected.
```
joseph-isaacs, line 29: We should expand this

vortex-expr/src/traversal/mod.rs
```diff
@@ -1,292 +1,361 @@
...
+                        t.changed |= transformed.changed;
+                        t
+                    }),
+                TraversalOrder::Skip => {
+                    transformed.order = TraversalOrder::Continue;
+                    Ok(transformed)
+                }
+                TraversalOrder::Stop => Ok(transformed),
+            }
+        }
```
joseph-isaacs: we can define this in term of rewrite

joseph-isaacs (approved):

vortex-expr/src/transform/remove_select.rs
```diff
@@ -3,26 +3,21 @@
...
-struct RemoveSelectTransform<'a> {
-    ctx: &'a DType,
-}
-
-impl MutNodeVisitor for RemoveSelectTransform<'_> {
-    type NodeTy = ExprRef;
-
-    fn visit_up(&mut self, node: ExprRef) -> VortexResult<TransformResult<Self::NodeTy>> {
-        if let Some(select) = node.as_opt::<SelectVTable>() {
+fn remote_select_transformer(node: ExprRef, ctx: &DType) -> VortexResult<Transformed<ExprRef>> {
```
onursatici:
```suggestion
fn remove_select_transformer(node: ExprRef, ctx: &DType) -> VortexResult<Transformed<ExprRef>> {
```

AdamGS: good catch

vortex-expr/src/traversal/mod.rs
AdamGS: done

vortex-expr/src/traversal/mod.rs
AdamGS: expanded

PR #3927, opened 2025-07-18
feat: duckdb workstealing

0ax1:


vortex-scan/src/multi_scan.rs
```diff
@@ -20,12 +24,18 @@ type ScanBuilderFactory<T> = Arc<SegQueue<Box<dyn FnOnce() -> ScanBuilder<T> + S
 #[derive(Default)]
 pub struct MultiScan<T> {
     scan_builder_factory: ScanBuilderFactory<T>,
+    stealers: Arc<DashMap<usize, Stealer<ArrayFuture<T>>>>,
```
gatesn: Can this be `Arc<RwLock<Vec<Stealer<...>>>>`, then we can remove the next_construction_id

vortex-scan/src/multi_scan.rs
```diff
@@ -46,44 +56,65 @@ impl<T> MultiScan<T> {
...
+                    }
+                }
             }
-            // TODO(Alex): worksteal tasks from other threads
         }
 
-        let task = self.task_queue.pop()?;
+        let task = self.worker.pop()?;
 
         self.local_pool.run_until(task).transpose()
```
gatesn, line 115: We should figure out next how to run ~2 tasks on the pool, rather than just one. That way we can do CPU while waiting for I/O

vortex-scan/src/multi_scan.rs
```diff
@@ -46,44 +56,65 @@ impl<T> MultiScan<T> {
...
                         for task in tasks {
-                            self.task_queue.push(Box::pin(task));
+                            self.worker.push(Box::pin(task));
                         }
                     }
                     Err(err) => return Some(Err(err)),
                 }
+            } else {
+                for _ in 0..self.stealers.len() {
+                    // Round robin to ensure work is not always stolen from the same worker.
```
gatesn, line 101: Would it be easier to just have every worker steal from their next neighbour?

vortex-scan/src/multi_scan.rs
0ax1: Totally, we can bring back FuturesUnordered and see how that affects perf.

vortex-scan/src/multi_scan.rs
0ax1: I think it'd be good trying to steal globally. A worker terminates if it has no more work. Trickling work down from a far away neighbour would take a couple of scan calls which would not be invoked as the worker already "finished". E.g. only one worker getting a larger file would lead to worse perf in that scenario.

vortex-scan/src/multi_scan.rs
0ax1: I quite like going with DashMap here. I think this boils to down to taste. DashMap maybe has slightly less contention I think. I'm kinda wired to atomics first before mutexes as they're often easier to get wrong in my experience.

vortex-scan/src/multi_scan.rs
0ax1: Saying only checking the next neighbour won't lead to global workstealing.

vortex-scan/src/multi_scan.rs
0ax1: We can just push n futures into `FuturesUnordered` and poll on that.

vortex-scan/src/multi_scan.rs
0ax1: Let's go with `Arc<RwLock<Vec<Stealer<...>>>>`. Feel free to push this to the branch.

0ax1: done

vortex-scan/src/multi_scan.rs
```diff
@@ -46,44 +54,63 @@ impl<T> MultiScan<T> {
...
-                            self.task_queue.push(Box::pin(task));
+                            self.worker.push(Box::pin(task));
                         }
                     }
                     Err(err) => return Some(Err(err)),
                 }
+            } else {
+                let stealer_count = self.stealers.read().len();
+
+                for _ in 0..stealer_count {
```
gatesn, line 100: Something still feels super racy about this.

Multiple threads can increment the next_stealer_id, and each thread loops over only the number of stealers they witnessed prior to the loop. So it's definitely not guaranteed that each worker will try every stealer.

But will all stealers be tried by at least one worker?

And what if one worker pops the next scan builder, but hasn't pushed the tasks yet. So it has nothing to steal. The other workers see no more scans, and no tasks to steal, so they break out of the loop and terminate. Now we don't get work stealing? 

I'm kind of convinced this will always terminate... but not totally convinced it behaves the way we want.

Any ideas on making it more obviously correct? Otherwise we could try and fuzz it maybe.

gatesn (approved):
Let's merge it for now, but we should revisit when we pull this out for regular scans too.

vortex-scan/src/multi_scan.rs
0ax1: I was just thinking, maybe should always try to steal from the worker with the most items.

0ax1: I don't really see how correctness ties into this, rather than perf. As long as no work is completely dropped this is correct.

0ax1: > And what if one worker pops the next scan builder, but hasn't pushed the tasks yet. So it has nothing to steal.

Yes in that case they terminate too early. In the C++ version we tracked whether all tasks from all files had been created and pushed. We should do the same here.

0ax1:
@gatesn Oddly, the DashMap solution performed significantly better than the Vector if you look at the benchmark results from yesterday TPC-H compared to today.

gatesn:
That is indeed curious... we definitely have assertions that the results are correct? We weren't terminating early?

Only because... TPCH isn't multi-file (edit: doh, obviously, that's the point!)

0ax1:
> That is indeed curious... we definitely have assertions that the results are correct? We weren't terminating early?
> 
> Only because... TPCH isn't multi-file (edit: doh, obviously, that's the point!)

TPC-H results are fully checked.

0ax1:
> That is indeed curious... we definitely have assertions that the results are correct? We weren't terminating early?
> 
> Only because... TPCH isn't multi-file (edit: doh, obviously, that's the point!)

Turns out we currently only check row counts: https://github.com/vortex-data/vortex/commit/373522a9e59240937d9ce5d48b11ac3c5ed40620#r162432166

PR #3931, opened 2025-07-18
fix: remove optimize Op, replace with compact_buffers function

a10y:
Fix both fuzzers.

The first is failing b/c we're not running it with nightly toolchain.

The second is failing due to some issues with how Optimize operation is implemented for RunEndArray. Currently Optimize for sliced types subtracts out the offset, but that actually forces a decompression, which is triggering our debug assertion that calling `optimize` never makes the array smaller.

After chatting with @gatesn , think for now we should move compacting to just be part of `impl Canonical`, as the main use-case is to call right before compression. Removing it now lets us keep the API surface clean while we stew on the right way to introduce this more widely (if at all).

Tested with the fuzzer artifact that were failing before this change:

image


lwwmanning (approved):

robert3005:
If you were to change runend array optimize to call subtract compute function this would all be fine? Though we might need to reencode after applying the compute function

gatesn:
Subtract may still decompress.

It's more that 1) it's a misnamed function, 2) unless it's canonical, it's never invoked anyway... so why keep so much dead code around

robert3005:
Yes, that’s why we would need to reencode. We can deal with it once we do recompression (or however that is going to be called)

PR #3938, opened 2025-07-19
perf: fine tune workstealing

0ax1:
This change makes sure that workers are not terminated as long as there's more work to steal.

### DIFF
--- a/vortex-duckdb/src/scan.rs
+++ b/vortex-duckdb/src/scan.rs
@@ -269,7 +269,7 @@ impl TableFunction for VortexTableFunction {
                 });
 
         Ok(VortexGlobalData {
-            multi_scan: MultiScan::new().with_scan_builders(closures),
+            multi_scan: MultiScan::new(bind_data.file_paths.len()).with_scan_builders(closures),
         })
     }
 
--- a/vortex-duckdb/src/scan.rs
+++ b/vortex-duckdb/src/scan.rs
@@ -269,7 +269,7 @@ impl TableFunction for VortexTableFunction {
                 });
 
         Ok(VortexGlobalData {
-            multi_scan: MultiScan::new(bind_data.file_paths.len()).with_scan_builders(closures),
+            multi_scan: MultiScan::new(closures),
         })
     }
 
--- a/vortex-duckdb/src/scan.rs
+++ b/vortex-duckdb/src/scan.rs
@@ -269,7 +269,7 @@ impl TableFunction for VortexTableFunction {
                 });
 
         Ok(VortexGlobalData {
-            multi_scan: MultiScan::new(bind_data.file_paths.len()).with_scan_builders(closures),
+            multi_scan: MultiScan::new().with_scan_builders(closures),
         })
     }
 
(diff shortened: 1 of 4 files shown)


vortex-duckdb/src/scan.rs
```diff
@@ -269,7 +269,7 @@ impl TableFunction for VortexTableFunction {
                 });
 
         Ok(VortexGlobalData {
-            multi_scan: MultiScan::new().with_scan_builders(closures),
+            multi_scan: MultiScan::new(bind_data.file_paths.len()).with_scan_builders(closures),
```
gatesn: It's probably best to just take a Vec<closures>

gatesn: And then internally store a VecDeq?

vortex-scan/src/multi_scan.rs
```diff
@@ -91,27 +101,34 @@ impl<T: Send + Sync + 'static> Iterator for MultiScanIterator<T> {
...
                         }
+                        self.scan_builders_constructed.fetch_add(1, Relaxed);
                     }
                     Err(err) => return Some(Err(err)),
                 }
             } else {
-                let stealer_count = self.stealers.read().len();
-
-                for _ in 0..stealer_count {
+                'outer_loop: while self.scan_builders_constructed.load(Relaxed)
```
gatesn: You sit in a really hot loop here, right?

I don't know how you do this without some kind of condition variable / lock.

The easiest abstraction is probably to look into mpmc channels where we can call close after all scans have been constructed. That way, each thread can wait on the channel's condition var for either the next task or the closure.

gatesn (commented):
I can confirm this fixes the work stealing problem I was seeing where workers weren't picking up tasks

vortex-duckdb/src/scan.rs
0ax1: As in vec of scan builders fns? Sure.

vortex-scan/src/multi_scan.rs
0ax1: How about just calling yield if we fail to steal a task?
```
let stealers = self.stealers.read();
let steal_id = self.next_stealer_id.fetch_add(1, SeqCst);
for idx in 0..stealers.len() {
    let idx = (steal_id + idx) % stealers.len();
    let stealer = &stealers[idx];
    if let Steal::Success(_) = stealer.steal_batch(&self.worker) {
        break 'outer_loop;
    }
}
std::thread::yield_now();
````

0ax1: > The easiest abstraction is probably to look into mpmc channels where we can call close after all scans have been constructed. 

This is not enough, we also need to be sure no stealers have items anymore.

vortex-duckdb/src/scan.rs
0ax1: done

vortex-scan/src/multi_scan.rs
0ax1: A hot loop might actually be fine here. The longest duration to spin in a loop is if all other stealers have no more tasks to steal from and the last scan builder isn't constructed yet.

0ax1: Every other case will break out of the loop in the first iteration.

gatesn (approved):

PR #3940, opened 2025-07-20
Update workstealing PR

gatesn:
* Moves shared state behind a single Arc
* Tries all stealers and accounts for retry.
* Yields thread inside the hot loop.

### DIFF
--- a/vortex-scan/src/multi_scan.rs
+++ b/vortex-scan/src/multi_scan.rs
@@ -26,7 +26,6 @@ pub struct MultiScan<T> {
     state: Arc<State<T>>,
 }
 
-#[allow(dead_code)]
 struct State<T> {
     /// A queue of factories that lazily produce [`ScanBuilder`] instances.
     scan_builders: SegQueue<ScanBuilderFactory<T>>,
@@ -37,9 +36,6 @@ struct State<T> {
     /// into a worker queue.
     num_scans_constructed: AtomicUsize,
 
-    /// The next ID to issue for a worker/stealer.
-    /// FIXME(ngates): remove if we use vec and not dashmap.
-    next_id: AtomicUsize,
     /// The vector of stealers, one for each worker.
     stealers: RwLock<Vec<Stealer<ArrayFuture<T>>>>,
 }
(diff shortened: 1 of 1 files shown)


vortex-scan/src/multi_scan.rs
```diff
@@ -56,75 +95,76 @@ impl<T> MultiScan<T> {
...
-                    // Round robin to ensure work is not always stolen from the same worker.
-                    for idx in 0..stealers.len() {
-                        let idx = (steal_id + idx) % stealers.len();
-                        if let Steal::Success(_) = stealers[idx].steal_batch(&self.worker) {
-                            break 'outer_loop;
-                        }
-                    }
+                // Otherwise, if we have constructed all scans, _and_ there are no tasks
+                // left to steal, then we can terminate.
+                if self.state.num_scans_constructed.load(Relaxed) >= self.state.num_scans {
```
0ax1, line 157: @gatesn This is a race. We can fail to steal a task and construct a scan builder on a different thread before checking `self.state.num_scans_constructed.load(Relaxed) >= self.state.num_scans`.

gatesn: Should this block move to line 127?

0ax1: This needs
```
           let did_steal = iter::repeat_with(|| {
                // This collect tries all stealers, exits early on the first successful steal,
                // or else tracks whether any steal requires a retry.
                self.state
                    .stealers
                    .read()
                    .iter()
                    .map(|stealer| stealer.steal_batch(&self.worker))
                    .collect::<Steal<()>>()
            })
            .find(|steal| !steal.is_retry())
            .and_then(|steal| steal.success())
            .is_some();
```
needs to run after 

```
let all_scan_builders_constructed = self.state.num_scans_constructed.load(Relaxed) >= self.state.num_scans;
```

with the termination condition

```
if  all_scan_builders_constructed && did_steal { return None; }
```

0ax1: I don't really like how implicit the logic got here with 2 continues and 2 returns, also depending looping back to the beginning. Hot take that it might be worth contemplating to revert `next` to the previous logic.

PR #3946, opened 2025-07-21
Improve work stealing loop

gatesn:
There was a slight race still where works would exit early. The while loop now ensures we spin until the very end

### DIFF
--- a/vortex-scan/src/multi_scan.rs
+++ b/vortex-scan/src/multi_scan.rs
@@ -6,13 +6,14 @@ use std::sync::Arc;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::{Relaxed, SeqCst};
 
-use crate::ScanBuilder;
 use crossbeam_deque::{Steal, Stealer, Worker};
 use crossbeam_queue::SegQueue;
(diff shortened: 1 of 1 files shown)


vortex-scan/src/multi_scan.rs
```diff
@@ -119,52 +135,24 @@ impl<T: Send + Sync + 'static> Iterator for MultiScanIterator<T> {
...
-                    // tasks are available.
-                    std::thread::yield_now();
+        if self.worker.is_empty() {
+            if !self.state.load_next_scan(&self.worker) {
+                // If there are no more scans to load, then there is at least one worker
+                // constructing a scan and about to push some tasks.
+                // We sit in a loop trying to steal some of those tasks, or else bail out when
+                // all scans have been constructed, and we didn't manage to steal anything. To avoid
+                // spinning too hot, we yield the thread each time we fail to steal work.
+                while self.state.num_scans_constructed.load(Relaxed) < self.state.num_scans {
```
0ax1: This exits too early if we already constructed all scan builders. We won't attempt to steal then.

0ax1 (approved):
Nice! Let's wait though for the benchmarks to agree.

PR #3947, opened 2025-07-21
chore: Propagate traces to task executor and a bunch of other places

AdamGS:


Cargo.toml
```diff
@@ -177,7 +177,6 @@ tpchgen = { git = "https://github.com/clflushopt/tpchgen-rs.git", rev = "d849ff4
 tpchgen-arrow = { git = "https://github.com/clflushopt/tpchgen-rs.git", rev = "d849ff430cd52250f6891ed4d5e3adad77bb2698" }
 tracing = { version = "0.1.41" }
 tracing-chrome = "0.7.2"
-tracing-futures = "0.2.5"
```
AdamGS, line 180: AFAICT we don't use it in anyway that we can't get directly from `tracing`

vortex-datafusion/src/persistent/format.rs
```diff
@@ -171,7 +172,7 @@ impl FileFormat for VortexFormat {
             .map(|o| {
                 let store = store.clone();
                 let cache = self.file_cache.clone();
-                tokio::task::spawn(async move {
+                SpawnedTask::spawn(async move {
```
AdamGS, line 175: Its an internal DF utility that gives us two things:
1. Propagates tracing
2. Gets cancelled when dropped

vortex-layout/Cargo.toml
```diff
@@ -64,8 +61,7 @@ vortex-layout = { path = ".", features = ["tokio"] }
 
 [features]
 roaring = ["dep:roaring"]
-tokio = ["dep:tokio", "vortex-error/tokio"]
-tracing = ["dep:tracing", "dep:tracing-futures"]
```
onursatici, line 68: This feature would still exist with the same name, because tracing is still optional and cargo would create a feature per optional dependency right?

onursatici (approved):

vortex-layout/Cargo.toml
AdamGS: yes, there are a couple of other features that depend on it IMO

PR #3948, opened 2025-07-21
chore[bench]: assert partial taxi take result correctness

joseph-isaacs:


bench-vortex/src/bin/random_access.rs
```diff
@@ -85,7 +115,7 @@ fn random_access(
             &runtime,
             iterations,
             || indices.clone(),
-            |indices| async { take_vortex_tokio(&taxi_vortex, indices).await },
+            |indices| async { take_vortex_tokio(&taxi_vortex, indices, Some(validate)).await },
```
robert3005: you always pass the validate here, worst case you pass identity instead of None?

robert3005 (approved):

PR #3951, opened 2025-07-21
DuckDB Table Filters

gatesn:
Adds accessors for the remaining DuckDB table filters.

An interesting side effect of this is that we actually push-down IN LIST operators, whereas previously DuckDB would provide us with a min/max filter over the values in the list. Curious if this is faster or slower than before...

### DIFF
--- a/vortex-duckdb/src/convert/table_filter.rs
+++ b/vortex-duckdb/src/convert/table_filter.rs
@@ -53,12 +53,11 @@ pub fn try_from_table_filter(
             // Optional expressions are optional not yet supported.
             return try_from_table_filter(&child, col, scope_dtype).or_else(|_err| {
                 // Failed to convert the optional expression, but it's optional, so who cares?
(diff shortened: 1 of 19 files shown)


vortex-duckdb/src/convert/table_filter.rs
```diff
@@ -0,0 +1,74 @@
...
+            assert!(
+                !scalars.is_empty(),
+                "IN filter must have at least one value"
+            );
+            let dtype = scalars[0].dtype().clone();
+            let list_scalar = Scalar::list(Arc::new(dtype), scalars, Nullability::Nullable);
+            list_contains(lit(list_scalar), col.clone())
+        }
+        TableFilterClass::Dynamic(_) => {
+            // Dynamic expressions are optional and not yet supported.
```
joseph-isaacs, line 68: This is not true

joseph-isaacs (approved):

PR #3952, opened 2025-07-21
Add size of self to nbytes

### LINKED ISSUE #3885: Rename `nbytes` and improve docs

robert3005:
Closes https://github.com/vortex-data/vortex/issues/3885#issuecomment-3089080441

### DIFF
--- a/vortex-array/src/array/mod.rs
+++ b/vortex-array/src/array/mod.rs
@@ -272,6 +272,7 @@ impl dyn Array + '_ {
     pub fn nbytes(&self) -> u64 {
         let mut nbytes = 0;
         for array in self.depth_first_traversal() {
+            nbytes += size_of_val(array.as_ref()) as u64;
             for buffer in array.buffers() {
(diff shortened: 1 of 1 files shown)


gatesn (changes requested):
I really don't understand why we would want this

robert3005:
I'm trying to figure out a solution to #3885. This pr wasn't ready for review, I only opened it to get benchmark numbers

robert3005:
If we add this it should be done in a separate function. FWIW the results don't seem to be affected but this changes compression of very small arrays to almost always not be worth it.

PR #3960, opened 2025-07-22
chore: benchmark lowering min target partition size

robert3005:


robert3005:
This setting doesn't seem to matter that much

PR #3965, opened 2025-07-22
Fix DuckDB timestamp values

gatesn:
Adds a more typed way of accessing DuckDB values. We were actually fetching the timestamp data incorrectly, so unclear how DuckDB was able to return correct results unless it also post-filters?

### DIFF
--- a/vortex-duckdb/src/convert/scalar.rs
+++ b/vortex-duckdb/src/convert/scalar.rs
@@ -19,10 +19,8 @@
 //! | `Binary` | `BLOB` |
 //! | `ExtScalar` (temporal) | `DATE`/`TIME`/`TIMESTAMP` |
 
-use std::ffi::CStr;
 use std::sync::Arc;
 
-use vortex::buffer::ByteBuffer;
 use vortex::dtype::Nullability::Nullable;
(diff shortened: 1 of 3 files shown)


vortex-duckdb/src/convert/scalar.rs
```diff
@@ -220,125 +217,79 @@ impl TryFrom<&Value> for Scalar {
...
             )),
-            DUCKDB_TYPE::DUCKDB_TYPE_TIMESTAMP_NS => Ok(Scalar::extension(
+            Val::Timestamp(micros) => Ok(Scalar::extension(
                 Arc::new(ExtDType::new(
                     TIMESTAMP_ID.clone(),
                     Arc::new(DType::Primitive(I64, Nullable)),
-                    Some(TemporalMetadata::Timestamp(TimeUnit::Ns, None).into()),
+                    Some(TemporalMetadata::Timestamp(TimeUnit::Us, None).into()),
                 )),
-                Scalar::new(DType::Primitive(I64, Nullable), value.as_i64().into()),
```
gatesn, line 310: These as_i64 were just returning i64::MIN

0ax1 (approved):

PR #3966, opened 2025-07-22
DuckDB Dynamic Expressions

gatesn:
Adds support for dynamic expressions to DuckDB extension. We may need to rethink this if/when other query engines support dynamic expressions, but it works well enough for now.

Interesting potential follow up:
* If we run ahead in the background and open all files, then we get file stats. 
* IFF we have a dynamic filter, we can assume the bound will tighten during execution. Therefore we should re-order our files/splits to make the most of the dynamic filter, i.e. for ORDER BY event_time DESC, we'd scan files in reverse order in an attempt to populate the highest event_time first.


vortex-duckdb/cpp/include/duckdb_vx/table_filter.h
```diff
@@ -57,7 +57,7 @@ void duckdb_vx_table_filter_get_conjunction_and(duckdb_vx_table_filter ffi_filte
                                                 duckdb_vx_table_filter_conjunction *out);
 
 typedef struct duckdb_vx_dynamic_filter_data_ *duckdb_vx_dynamic_filter_data;
-void duckdb_vx_dynamic_filter_data_drop(duckdb_vx_dynamic_filter_data ffi_data);
+void duckdb_vx_dynamic_filter_data_drop(duckdb_vx_dynamic_filter_data *ffi_data);
```
joseph-isaacs: free?
```suggestion
void duckdb_vx_dynamic_filter_data_free(duckdb_vx_dynamic_filter_data *ffi_data);
```

vortex-duckdb/src/duckdb/table_filter.rs
```diff
@@ -264,9 +264,12 @@ wrapper!(
     /// A handle to mutable dynamic filter data.
     DynamicFilterData,
     cpp::duckdb_vx_dynamic_filter_data,
-    |ptr: &mut cpp::duckdb_vx_dynamic_filter_data| unsafe { cpp::duckdb_vx_dynamic_filter_data_drop(*ptr) }
+    cpp::duckdb_vx_dynamic_filter_data_drop
 );
 
+unsafe impl Send for DynamicFilterData {}
+unsafe impl Sync for DynamicFilterData {}
```
joseph-isaacs, line 272:
```suggestion
// Since `duckdb::DynamicFilterData` is Send + Sync 
unsafe impl Send for DynamicFilterData {}
unsafe impl Sync for DynamicFilterData {}
```

vortex-expr/src/exprs/dynamic.rs
```diff
@@ -0,0 +1,293 @@
...
+            && self.operator == other.operator
+            && self.rhs == other.rhs
+            && self.default == other.default
+    }
+}
+impl Eq for DynamicComparisonExpr {}
+
+/// Hash and PartialEq are implemented based on the ptr of the value function, such that the
+/// internal value doesn't impact the hash of an expression tree.
+struct Rhs {
```
joseph-isaacs, line 59: Can we name this better?

vortex-expr/src/exprs/dynamic.rs
```diff
@@ -0,0 +1,293 @@
...
+use vortex_scalar::{Scalar, ScalarValue};
+
+use crate::traversal::{Node, NodeVisitor, TraversalOrder};
+use crate::{
+    AnalysisExpr, ExprEncodingRef, ExprId, ExprRef, IntoExpr, Scope, StatsCatalog, VTable, vtable,
+};
+
+vtable!(DynamicComparison);
+
+/// A dynamic comparison expression can be used to capture a comparison to a value that can change
```
joseph-isaacs, line 24: Make clear that its always refining a range

vortex-expr/src/exprs/dynamic.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub fn version(&self) -> u64 {
+        let mut guard = self.prev_versions.lock();
+
+        let mut updated = false;
+        for (i, expr) in self.exprs.iter().enumerate() {
+            let current = (expr.rhs.value)().map(|v| Scalar::new(expr.rhs.dtype.clone(), v));
+            if current != guard.1[i] {
+                // At least one expression has been updated.
+                // We don't bail out early in order to avoid false positives for future calls
+                // to `is_updated`.
```
joseph-isaacs, line 278: Maybe debug_assert the value is >= last value

gatesn: There's sort of nothing that assumes that right now? But also... I don't think we implement compare for all types of scalars.

vortex-expr/src/exprs/dynamic.rs
gatesn: 🤷

joseph-isaacs (approved):

PR #3973, opened 2025-07-22
fix: Pruning expressions check NanCount where appropriate

### LINKED ISSUE #3958: Pushdown compare with NaN broken
We implement our `compare` binary comparison operation using the arrow_ord kernels. This uses the f32/f64 `total_cmp`, which includes NaN as >= all positive normal numbers.

This causes inconsistent behavior when pushing down a comparison operation against NaN to a file column that contains NaN.

Directly evaluating the following comparison results in `[false, false, true, true]`:

```
PrimitiveArray([1.0, 2.0, NaN, NaN]) >= f32::NAN
```

However, this is not the case when the filter is pushed down into a scan, because our min/max stats do not consider NaN. The pruning expression that gets pushed down is

```
$.max < f32::NAN --> 2.0 < f32::NAN = true
```

which causes the whole thing to get pruned and the result becomes `[false, false, false, false]`

https://github.com/vortex-data/vortex/blob/bba0e63e3e9bb271816edde5ed3fcc0e1924db5b/vortex-array/src/arrays/primitive/compute/min_max.rs#L43-L48

We need to either

1. Make the stats contain NaN, and preserve the `total_cmp` ordering in pushdown
2. Have a fallback `compare` kernel for float arrays with NaNCount > 0 that masks out the NaNs to false

a10y:
Fixes #3958 

NanCount stat is checked in the pruning evaluation if it is present (i.e. for float columns and literals)

- [x] Verify fixes fuzz failure
- [x] Add unit tests

### DIFF
--- a/vortex-array/src/arrays/struct_/mod.rs
+++ b/vortex-array/src/arrays/struct_/mod.rs
@@ -241,7 +241,7 @@ impl StructArray {
 
     /// Removes and returns a column from the struct array by name.
     /// If the column does not exist, returns `None`.
-    pub fn drop(&mut self, name: impl Into<FieldName>) -> Option<ArrayRef> {
+    pub fn remove_column(&mut self, name: impl Into<FieldName>) -> Option<ArrayRef> {
         let name = name.into();
 
         let Some(struct_dtype) = self.dtype.as_struct() else {
@@ -410,7 +410,7 @@ mod test {
         )
         .unwrap();
 
-        let removed = struct_a.drop("xs").unwrap();
+        let removed = struct_a.remove_column("xs").unwrap();
         assert_eq!(
             removed.dtype(),
             &DType::Primitive(PType::I64, Nullability::NonNullable)
@@ -434,7 +434,7 @@ mod test {
             [4u64, 5, 6, 7, 8]
         );
 
-        let empty = struct_a.drop("non_existent");
+        let empty = struct_a.remove_column("non_existent");
         assert!(
             empty.is_none(),
             "Expected None when removing non-existent column"
--- a/vortex-array/src/arrays/struct_/mod.rs
+++ b/vortex-array/src/arrays/struct_/mod.rs
@@ -244,11 +244,7 @@ impl StructArray {
     pub fn remove_column(&mut self, name: impl Into<FieldName>) -> Option<ArrayRef> {
         let name = name.into();
 
-        let Some(struct_dtype) = self.dtype.as_struct() else {
-            unreachable!(
-                "struct arrays must have be a DType::Struct, this is likely an internal bug."
-            )
-        };
+        let struct_dtype = self.struct_fields().clone();
 
         let position = struct_dtype
             .names()
@@ -264,17 +260,9 @@ impl StructArray {
     }
 
     /// Create a new StructArray by appending a new column onto the existing array.
-    pub fn with_column(
-        &mut self,
-        name: impl Into<FieldName>,
-        array: ArrayRef,
-    ) -> VortexResult<Self> {
+    pub fn with_column(&self, name: impl Into<FieldName>, array: ArrayRef) -> VortexResult<Self> {
         let name = name.into();
-        let Some(struct_dtype) = self.dtype.as_struct() else {
-            unreachable!(
-                "struct arrays must have be a DType::Struct, this is likely an internal bug."
-            )
-        };
+        let struct_dtype = self.struct_fields().clone();
 
         let names = struct_dtype.names().iter().cloned().chain(once(name));
         let types = struct_dtype.fields().chain(once(array.dtype().clone()));
--- a/vortex-array/src/arrays/struct_/mod.rs
+++ b/vortex-array/src/arrays/struct_/mod.rs
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use std::fmt::Debug;
+use std::iter::once;
 
 use itertools::Itertools;
 use vortex_dtype::{DType, FieldName, FieldNames, StructFields};
@@ -240,7 +241,7 @@ impl StructArray {
 
     /// Removes and returns a column from the struct array by name.
     /// If the column does not exist, returns `None`.
-    pub fn remove_column(&mut self, name: impl Into<FieldName>) -> Option<ArrayRef> {
+    pub fn drop(&mut self, name: impl Into<FieldName>) -> Option<ArrayRef> {
         let name = name.into();
 
         let Some(struct_dtype) = self.dtype.as_struct() else {
@@ -260,6 +261,29 @@ impl StructArray {
         self.dtype = DType::Struct(new_dtype, self.dtype.nullability());
 
         Some(field)
+    }
+
+    /// Create a new StructArray by appending a new column onto the existing array.
+    pub fn with_column(
+        &mut self,
+        name: impl Into<FieldName>,
+        array: ArrayRef,
+    ) -> VortexResult<Self> {
+        let name = name.into();
+        let Some(struct_dtype) = self.dtype.as_struct() else {
+            unreachable!(
+                "struct arrays must have be a DType::Struct, this is likely an internal bug."
+            )
+        };
+
+        let names = struct_dtype.names().iter().cloned().chain(once(name));
+        let types = struct_dtype.fields().chain(once(array.dtype().clone()));
+        let new_fields = StructFields::new(names.collect(), types.collect());
+
+        let mut children = self.fields.clone();
+        children.push(array);
+
+        Self::try_new_with_dtype(children, new_fields, self.len, self.validity.clone())
     }
 }
 
@@ -386,7 +410,7 @@ mod test {
         )
         .unwrap();
 
-        let removed = struct_a.remove_column("xs").unwrap();
+        let removed = struct_a.drop("xs").unwrap();
         assert_eq!(
             removed.dtype(),
             &DType::Primitive(PType::I64, Nullability::NonNullable)
@@ -410,7 +434,7 @@ mod test {
             [4u64, 5, 6, 7, 8]
         );
 
-        let empty = struct_a.remove_column("non_existent");
+        let empty = struct_a.drop("non_existent");
         assert!(
             empty.is_none(),
             "Expected None when removing non-existent column"
--- a/vortex-datafusion/src/persistent/source.rs
+++ b/vortex-datafusion/src/persistent/source.rs
@@ -170,10 +170,20 @@ impl FileSource for VortexSource {
         VORTEX_FILE_EXTENSION
     }
 
-    fn fmt_extra(&self, _t: DisplayFormatType, f: &mut Formatter) -> std::fmt::Result {
-        if let Some(ref predicate) = self.predicate {
-            write!(f, ", predicate={predicate}")?;
-        };
+    fn fmt_extra(&self, t: DisplayFormatType, f: &mut Formatter) -> std::fmt::Result {
+        match t {
+            DisplayFormatType::Default | DisplayFormatType::Verbose => {
+                if let Some(ref predicate) = self.predicate {
+                    write!(f, ", predicate: {predicate}")?;
+                }
+            }
+            // Use TreeRender style key=value formatting to display the predicate
+            DisplayFormatType::TreeRender => {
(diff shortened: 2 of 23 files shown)


vortex-layout/src/layouts/zoned/reader.rs
```diff
@@ -105,12 +113,35 @@ impl ZonedReader {
...
+                                Scalar::primitive(0u64, Nullability::NonNullable),
+                                nzones,
+                            )
+                            .into_array(),
+                        )?;
+                    }
+
                     // SAFETY: This is only fine to call because we perform validation above
-                    Ok(ZoneMap::unchecked_new(zones_array, present_stats))
+                    Ok(ZoneMap::new_unchecked(zones_array, present_stats))
```
a10y, line 129: make the naming more idiomatic

vortex-array/src/arrays/struct_/mod.rs
```diff
@@ -240,7 +241,7 @@ impl StructArray {
 
     /// Removes and returns a column from the struct array by name.
     /// If the column does not exist, returns `None`.
-    pub fn remove_column(&mut self, name: impl Into<FieldName>) -> Option<ArrayRef> {
+    pub fn drop(&mut self, name: impl Into<FieldName>) -> Option<ArrayRef> {
```
a10y: again, more idiomatic naming

a10y: Actually, this probably conflicts with the Drop trait impl...maybe `drop_column` ?

vortex-expr/src/pruning/pruning_expr.rs
```diff
@@ -60,41 +70,29 @@ pub fn field_path_stat_field_name(field_path: &FieldPath, stat: Stat) -> FieldNa
...
-    let expr = expr.stat_falsification(&mut catalog)?;
-
-    // TODO(joe): filter access by used exprs
-    let mut relation: Relation<FieldPath, Stat> = Relation::new();
-    for ((field_path, stat), _) in catalog.usage.into_iter() {
-        relation.insert(field_path, stat)
-    }
-
-    Some((expr, relation))
-}
```
a10y, line 78: this was only being used in the tests below, I don't think there's a use for a version of this function that doesn't check for stat availability so I just removed this

vortex-datafusion/src/persistent/source.rs
```diff
@@ -176,9 +175,6 @@ impl FileSource for VortexSource {
             DisplayFormatType::Default | DisplayFormatType::Verbose => {
                 if let Some(predicate) = &self.predicate {
                     write!(f, ", predicate={predicate}")?;
-                    if let Some((pruning_predicate, _)) = pruning_expr(predicate) {
```
robert3005, line 179: brutal... if you get rid of this branch then all the cases are the same

vortex-array/src/arrays/struct_/mod.rs
```diff
@@ -261,6 +262,29 @@ impl StructArray {
...
+        name: impl Into<FieldName>,
+        array: ArrayRef,
+    ) -> VortexResult<Self> {
+        let name = name.into();
+        let Some(struct_dtype) = self.dtype.as_struct() else {
+            unreachable!(
+                "struct arrays must have be a DType::Struct, this is likely an internal bug."
+            )
+        };
+
```
robert3005: `self.struct_fields()`

vortex-array/src/arrays/struct_/mod.rs
```diff
@@ -261,6 +262,29 @@ impl StructArray {
 
         Some(field)
     }
+
+    /// Create a new StructArray by appending a new column onto the existing array.
+    pub fn with_column(
+        &mut self,
```
robert3005: why is this mut if you make a copy?

vortex-array/src/arrays/struct_/mod.rs
a10y: nice, I just copied this from `remove_column` a few lines up, I'll switch them both over

vortex-array/src/arrays/struct_/mod.rs
a10y: holdover, fixed now

vortex-datafusion/src/persistent/source.rs
a10y: merged them

robert3005: I guess for pretty printing cases you want to not print `, ` tree case

a10y: ah yea, it looks like the TreeRender formatting is slightly different, needs `=` separated key-values pairs. I'll fix this

vortex-expr/src/exprs/binary.rs
```diff
@@ -138,6 +138,20 @@ impl Display for BinaryExpr {
 
 impl AnalysisExpr for BinaryExpr {
```
joseph-isaacs, line 139: We can add an example transformed predicate?

a10y: There isn't really any documentation in this file, I've updated the pruning_expr unit tests. I agree could be some more documentation here

vortex-expr/src/exprs/binary.rs
```diff
@@ -138,6 +138,20 @@ impl Display for BinaryExpr {
...
+            catalog: &mut dyn StatsCatalog,
+        ) -> Option<ExprRef> {
+            lhs.nan_count(catalog)
+                .into_iter()
+                .chain(rhs.nan_count(catalog))
+                .map(|nans| eq(nans, lit(0u64)))
+                .reduce(and)
+        }
+
         match self.operator {
```
joseph-isaacs, line 175: can we apply the non nan check outside the match?

vortex-expr/src/exprs/get_item.rs
joseph-isaacs, line 1: we need to add nan_count to `extract_relevant_file_stats_as_struct_row`

vortex-expr/src/exprs/literal.rs
```diff
@@ -112,6 +112,28 @@ impl AnalysisExpr for LiteralExpr {
...
+        match_each_float_ptype!(value.ptype(), |T| {
+            match value.typed_value::<T>() {
+                None => Some(lit(0u64)),
+                Some(value) => {
+                    if value.is_nan() {
+                        Some(lit(1u64))
+                    } else {
+                        Some(lit(0u64))
+                    }
+                }
```
joseph-isaacs: Some(value) 
```suggestion
                Some(value) if value.is_nan() => Some(lit(1u64))
                 _ => Some(lit(0u64))
                }
```

vortex-expr/src/exprs/binary.rs
joseph-isaacs: Even this example would really help
`a > 2.0 ==> (a < 2.0) and a_nan_count == 0`

vortex-expr/src/exprs/binary.rs
a10y: Not easily no, b/c the AND/OR/numeric operators don't apply and they're all grouped together in the enum

vortex-expr/src/exprs/get_item.rs
a10y: done (i think)

vortex-expr/src/exprs/binary.rs
a10y: I added a comment in the function body, let me know if you think we need more

vortex-expr/src/exprs/binary.rs
```diff
@@ -157,12 +190,55 @@ impl AnalysisExpr for BinaryExpr {
...
+
+                if let Some(nan_check) = make_nan_predicate(self.lhs(), self.rhs(), catalog) {
+                    Some(and(nan_check, min_max_check))
+                } else {
+                    Some(min_max_check)
+                }
+            }
+            Operator::Gte => {
+                // NaN is not captured by the min/max stat, so we must check NaNCount before pruning
+                let min_max_check = lt(self.lhs.max(catalog)?, self.rhs.min(catalog)?);
```
a10y, line 214: This PR also cheekily fixes broken pruning predicate logic, this was `lt_eq` before 🫢

robert3005 (approved):

vortex-expr/src/exprs/binary.rs
joseph-isaacs: I dont see how?

joseph-isaacs: its always the same chain and reduce and

vortex-expr/src/exprs/literal.rs
joseph-isaacs: is this is no?

joseph-isaacs (changes requested):
Can we try and simplify the logic here

vortex-expr/src/exprs/literal.rs
a10y: no i incorporated this change, I just did it by hand to get formatting right

vortex-expr/src/exprs/binary.rs
a10y: I don't know how to do this without accessing the catalog spuriously. And the catalog is stateful, so I only want to access it in branches where the NanCount stat is actually needed.

joseph-isaacs: I see, maybe make a lambda what make and expr and applies the and (nan) only for these.

On the other handle the stats catalogue was designed to allow for checking if the expr in actually in the final prune expr

vortex-expr/src/exprs/binary.rs
```diff
@@ -138,6 +138,20 @@ impl Display for BinaryExpr {
...
+            lhs: &ExprRef,
+            rhs: &ExprRef,
+            catalog: &mut dyn StatsCatalog,
+        ) -> Option<ExprRef> {
+            lhs.nan_count(catalog)
+                .into_iter()
+                .chain(rhs.nan_count(catalog))
+                .map(|nans| eq(nans, lit(0u64)))
+                .reduce(and)
+        }
```
joseph-isaacs: Why is this different to the other nan checks

a10y: i think you're looking at an old revision, can you refresh? these line numbers are different now

vortex-expr/src/exprs/binary.rs
```diff
@@ -148,7 +162,14 @@ impl AnalysisExpr for BinaryExpr {
...
                 let right = min_rhs.zip(max_lhs).map(|(a, b)| gt(a, b));
-                left.into_iter().chain(right).reduce(or)
+
+                let min_max_check = left.into_iter().chain(right).reduce(or);
+
+                // NaN is not captured by the min/max stat, so we must check NaNCount before pruning
+                make_nan_predicate(self.lhs(), self.rhs(), catalog)
+                    .into_iter()
+                    .chain(min_max_check)
+                    .reduce(and)
```
joseph-isaacs: Here

vortex-expr/src/exprs/binary.rs
joseph-isaacs: I meant the one below (169)

vortex-expr/src/exprs/binary.rs
a10y: that's a good idea let me try that

joseph-isaacs: https://github.com/vortex-data/vortex/blob/30fd768f7d60a9125a2bf122f0f0630d55126f48/vortex-expr/src/pruning/pruning_expr.rs#L71 here we can check if the expr in in the final pruning expr and drop other stats loads

vortex-expr/src/pruning/pruning_expr.rs
```diff
@@ -284,14 +297,14 @@ mod tests {
         assert_eq!(&converted, &(expected_expr));
     }
 
-    #[test]
-    pub fn pruning_lt_column() {
+    #[rstest::rstest]
```
robert3005:
```suggestion
    #[rstest]
```

vortex-expr/src/exprs/binary.rs
a10y: i just centralized the wrapping logic in one place, it was easier

joseph-isaacs (approved):
thanks

PR #3979, opened 2025-07-23
fix: Prevent compressor from choosing SequenceArray for dictionary codes

### LINKED ISSUE #3976: Don't compress array with dictionary encoding if codes are a sequence
We are just shifting plain values into code lookups for no added benefit. This is likely caused by the fact that SequenceArray has nbytes of 0

robert3005:
fix #3976

### DIFF
--- a/vortex-btrblocks/src/integer.rs
+++ b/vortex-btrblocks/src/integer.rs
@@ -589,7 +589,7 @@ impl Scheme for DictScheme {
         let dict = dictionary_encode(stats)?;
 
         // Cascade the codes child
-        let mut new_excludes = vec![DICT_SCHEME];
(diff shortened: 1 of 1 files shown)


vortex-btrblocks/src/integer.rs
```diff
@@ -589,7 +589,7 @@ impl Scheme for DictScheme {
         let dict = dictionary_encode(stats)?;
 
         // Cascade the codes child
-        let mut new_excludes = vec![DICT_SCHEME];
+        let mut new_excludes = vec![DICT_SCHEME, SEQUENCE_SCHEME];
```
AdamGS, line 593: maybe add comment here documenting why? feels like its very likely that in a few months we'll randomly revisit it

AdamGS (approved):

PR #3981, opened 2025-07-23
fix: Defer creation of ZonedMap cached results until they're necessary

robert3005:


robert3005:
This seems unnecessary

PR #3984, opened 2025-07-23
prevent double stealing

onursatici:


vortex-scan/src/work_queue.rs
```diff
@@ -134,12 +134,10 @@ impl<T> Iterator for WorkQueueIterator<T> {
                 // all scans have been constructed, and we didn't manage to steal anything. To avoid
                 // spinning too hot, we yield the thread each time we fail to steal work.
                 while self.state.num_factories_constructed.load(Relaxed) < self.state.num_factories
```
0ax1, line 140: We can't do that as factories might already have been created but there are still tasks to steal. We would then not hit the body of the while loop.

onursatici: ah yea you are right

0ax1: "no more tasks to steal" needs to be part of the while condition.

0ax1: I think the proper solution is writing an idempotent `is_empty` function.

onursatici: yea there doesn't seem to be a clean way to get if there are tasks to steal without attempting to steal

0ax1 (approved):
Looks good! But lets add a comment why anything not being `is_retry` is fine to break out of the loop.

PR #3991, opened 2025-07-23
fix: list array builder respects sliced list arrays

danking:
We previously assumed the first offset was always zero.

### DIFF
--- a/vortex-array/src/builders/list.rs
+++ b/vortex-array/src/builders/list.rs
@@ -163,26 +163,17 @@ impl<O: OffsetPType> ArrayBuilder for ListBuilder<O> {
 
         let index_dtype = self.index_builder.dtype();
 
-        let n_leading_junk_values = offsets.scalar_at(0)?.cast(index_dtype)?;
+        let n_leading_junk_values_scalar = offsets.scalar_at(0)?.cast(index_dtype)?;
+        let n_leading_junk_values = usize::try_from(&n_leading_junk_values_scalar)?;
 
         let casted_offsets = cast(&offsets.slice(1, offsets.len())?, index_dtype)?;
         let offsets_without_leading_junk =
-            sub_scalar(&casted_offsets, n_leading_junk_values.clone())?;
+            sub_scalar(&casted_offsets, n_leading_junk_values_scalar)?;
         let offsets_into_builder =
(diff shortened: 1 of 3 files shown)


vortex-array/src/builders/list.rs
```diff
@@ -119,37 +144,52 @@ impl<O: OffsetPType> ArrayBuilder for ListBuilder<O> {
...
+        let casted_offsets = cast(&offsets.slice(1, offsets.len())?, index_dtype)?;
+        let offsets_without_leading_junk =
+            sub_scalar(&casted_offsets, n_leading_junk_values.clone())?;
+        let offsets_into_builder =
+            add_scalar(&offsets_without_leading_junk, n_already_added_values.into())?;
+
+        let n_leading_junk_values = n_leading_junk_values
+            .as_primitive()
+            .pvalue()
+            .vortex_expect("list offsets must be non-null");
```
robert3005: you don't need need this, usize::try_form(<scalar>) is a thing

vortex-array/src/builders/list.rs
```diff
@@ -119,37 +144,52 @@ impl<O: OffsetPType> ArrayBuilder for ListBuilder<O> {
...
+        let n_leading_junk_values =
+            usize::try_from(n_leading_junk_values).vortex_expect("list offsets fit in usize");
```
robert3005: move this up, do `usize::try_from(&n_leading_junk_values)` and then remove clone on 170

vortex-array/src/builders/list.rs
```diff
@@ -119,37 +144,52 @@ impl<O: OffsetPType> ArrayBuilder for ListBuilder<O> {
...
+        let last_offset = offsets
+            .scalar_at(offsets.len() - 1)?
+            .as_primitive()
+            .pvalue()
+            .vortex_expect("list offsets must be non-null");
```
robert3005: unnecessary

robert3005 (commented):
Some simplifications. I'm impressed we have enough compute function to do this generically

vortex-array/src/builders/list.rs
danking: Ah, nice.

robert3005 (approved):

PR #3992, opened 2025-07-23
feat: teach SparseArray to canonicalize lists

danking:
Kept me busy while I was aloft the Atlantic.

Stacked on https://github.com/vortex-data/vortex/pull/3991

### DIFF
--- a/encodings/sparse/src/canonical.rs
+++ b/encodings/sparse/src/canonical.rs
@@ -15,9 +15,7 @@ use vortex_array::builders::{
 use vortex_array::patches::Patches;
 use vortex_array::validity::Validity;
 use vortex_array::vtable::CanonicalVTable;
-use vortex_array::{
-    Array, ArrayRef, Canonical, IntoArray as _, ToCanonical as _, match_smallest_offset_type,
-};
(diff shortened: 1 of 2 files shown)


robert3005 (changes requested):
I would unwrap this a level instead of going through scalars. If the fill value is null then you take the values of the patches and intersperse offsets with nulls. If the fill value is not null then I think you still will be better off constructing the values and offsets manually.

danking:
You mean instead of going through the ListBuilder, right? Like, use an index builder and values builder directly?

robert3005:
Yes, build the components directly and assemble them into the listarray. In the common case of null fill value it should be a lot simpler and roughly the same for the non null fill

danking:
@robert3005 OK. I added `canonicalize_sparse_lists_inner_with_null_fill_value` which handles the null fill value case. In that case, we avoid copying the sparse values' elements. I didn't change the non-null fill value case because it seems to me there is no benefit over the ListBuilder. Do you agree?

danking:
@robert3005 bump

a10y (approved):

robert3005 (approved):
There's something deeply unsatisfying about list builders

PR #3651, opened 2025-06-25
feat: teach SparseArray to canoncialize structs

danking:

PR #3993, opened 2025-07-23
feat: teach SparseArray to canonicalize Binary arrays

danking:
More flight work.


danking:
@a10y I think this leaves only sparse arrays of Extension DType as broken. I'm not entirely sure I understand extension dtypes. Can I just canonicalize them as if they are their representation? 

https://github.com/vortex-data/vortex/pull/3651#discussion_r2167581783

robert3005 (approved):

PR #3997, opened 2025-07-23
chore[docs]: deny missing_docs on vortex-array

lwwmanning:


lwwmanning:
going to close this since @gatesn has a large inbound refactor, will give it another go after that

PR #4002, opened 2025-07-24
fix: Implement compact for ListArray

### LINKED ISSUE #3640: issue: vortex file is too large than parquet
After I convert parquet file to vortex file, I find that vortex file is too large than I expected.
```
2019809413   seq.parquet
9463239652   seq.vortex
```

I extract one column which type is struct<struct<list<int64>>>, and this issue can be reproduced, but the compressed vortex array size is **157182932**, a little bit bigger than parquet.
```
150732338  single_column_multigroups.parquet
628437016  single_column_multigroups.vortex
```

If I slice first 1000 line, the file size is very close.
```
6315041  single_column_1000.parquet
6577728  single_column_1000.vortex
```

I just follow [quick start](https://docs.vortex.dev/quickstart/python) to do these things.

I have uploaded test file to google drive - https://drive.google.com/file/d/1Fuk7KUO9PHJzbsmOsGJw2UpzztHeCDU4/view?usp=sharing , and its md5 is e8adb6e279c6a3cfbd2845c5370da869.

robert3005:
fix #3640

### DIFF
--- a/vortex-array/src/arrays/list/mod.rs
+++ b/vortex-array/src/arrays/list/mod.rs
@@ -106,8 +106,8 @@ impl ListArray {
     /// Panics if the index is out of bounds.
     pub fn offset_at(&self, index: usize) -> usize {
         assert!(
-            index < self.len() + 1,
-            "Index {index} out of bounds 0..{}",
+            index <= self.len(),
+            "Index {index} out of bounds 0..={}",
             self.len()
         );
 
--- a/vortex-array/src/arrays/list/mod.rs
+++ b/vortex-array/src/arrays/list/mod.rs
@@ -106,7 +106,7 @@ impl ListArray {
     /// Panics if the index is out of bounds.
     pub fn offset_at(&self, index: usize) -> usize {
         assert!(
-            index < self.len(),
+            index < self.len() + 1,
             "Index {index} out of bounds 0..{}",
             self.len()
         );
--- a/vortex-array/src/arrays/list/mod.rs
+++ b/vortex-array/src/arrays/list/mod.rs
@@ -9,7 +9,7 @@ use std::sync::Arc;
 use itertools::Itertools;
 use num_traits::{AsPrimitive, PrimInt};
 use vortex_dtype::{DType, NativePType, match_each_native_ptype};
-use vortex_error::{VortexResult, vortex_bail};
+use vortex_error::{VortexResult, VortexUnwrap, vortex_bail};
 use vortex_scalar::Scalar;
 
 use crate::arrays::PrimitiveVTable;
@@ -101,10 +101,15 @@ impl ListArray {
         })
     }
 
-    pub fn offset_at(&self, index: usize) -> VortexResult<usize> {
-        if index > self.len() + 1 {
-            vortex_bail!(OutOfBounds: index, 0, self.len() + 1)
-        }
+    /// Returns the offset at the given index from the list array.
+    ///
+    /// Panics if the index is out of bounds.
+    pub fn offset_at(&self, index: usize) -> usize {
+        assert!(
+            index < self.len(),
+            "Index {index} out of bounds 0..{}",
+            self.len()
+        );
 
         self.offsets()
             .as_opt::<PrimitiveVTable>()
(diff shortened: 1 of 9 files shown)


vortex-array/src/arrays/varbinview/compact.rs
```diff
@@ -11,68 +11,67 @@ use crate::builders::{ArrayBuilder, VarBinViewBuilder};
...
-    }
+impl VarBinViewArray {
+    /// Returns a compacted copy of the input array, where all wasted space has been cleaned up. This
+    /// operation can be very expensive, in the worst cast copying all existing string data into
+    /// a new allocation.
+    ///
+    /// After slicing/taking operations `VarBinViewArray`s can continue to hold references to buffers
+    /// that are no longer visible. We detect when there is wasted space in any of the buffers, and if
+    /// so, will aggressively compact all visile outlined string data into a single new buffer.
+    pub fn compact_buffers(&self) -> VortexResult<VarBinViewArray> {
```
robert3005, line 22: I have arbitrarily made some functions here memeber functions of the array

vortex-array/src/arrays/list/mod.rs
```diff
@@ -100,40 +101,58 @@ impl ListArray {
...
     }
 
-    // TODO: fetches the elements of the array ignoring validity
+    /// Returns the elements array.
     pub fn elements(&self) -> &ArrayRef {
         &self.elements
     }
+
+    /// Create a copy of this array by adjusting offsets to start at 0 and removing elements not referenced by the offsets
+    pub fn offset_to_0(&self) -> VortexResult<Self> {
```
robert3005: not convinced by the name. In the end though its changing starting offset to 0 is why you need to prune elements

vortex-array/src/arrays/list/mod.rs
```diff
@@ -100,40 +101,58 @@ impl ListArray {
         })
     }
 
-    // TODO: merge logic with varbin
-    // TODO(ngates): should return a result if it requires canonicalizing offsets
-    pub fn offset_at(&self, index: usize) -> usize {
```
a10y: this feels like a weird use of result

vortex-array/src/arrays/list/mod.rs
a10y: I'd call this `reset_offsets()`

a10y:
Does rezeroing the offsets actually do anything to reduce the space usage?

a10y:
Ah yea, in the case of list it does, elements is left intact and offsets are sliced

vortex-array/src/arrays/list/mod.rs
```diff
@@ -346,4 +367,66 @@ mod test {
...
+                    Arc::new(I32.into()),
+                    vec![4.into(), 5.into(), 6.into()],
+                    Nullability::NonNullable,
+                )
+                .as_list(),
+            )
+            .vortex_unwrap();
+        builder
+            .append_value(
+                Scalar::list(
```
a10y, line 403: We could probably change Scalar::list to take an `Into<Arc<DType>>` instead of `Arc<DType>`

a10y:
The top-level BtrBlocksCompressor::compress function is only called once per column in the write path, so we might want to recursively compact the List array? Same for StructArray actually for when we write nested structs

robert3005:
In order to prune the elements you need to rewrite offsets

vortex-array/src/arrays/list/mod.rs
robert3005: Indeed, we moved the varbin to return result so I made this match. Can swap both tbh

vortex-array/src/arrays/list/mod.rs
```diff
@@ -100,40 +101,64 @@ impl ListArray {
...
     }
 
-    // TODO: merge logic with varbin
-    // TODO(ngates): should return a result if it requires canonicalizing offsets
+    /// Returns the offset at the given index from the list array.
+    ///
+    /// Panics if the index is out of bounds.
     pub fn offset_at(&self, index: usize) -> usize {
+        assert!(
+            index < self.len() + 1,
```
a10y: Could you make these <= Len? Len+1 looks fishy

a10y (approved):

Issue #3890, opened 2025-07-16
Clean up Vortex layouts and threading model


gatesn:
Despite a previous version of this issue, the plan now is to lean _more_ into Rust's async/await syntax for performing scans. There's a key change however that fixes a few problems we've been having: CPU work must be spawned.

This essentially allows us to answer the question "are you ready to run". Currently, in order to check if a future is ready, or whether it requires I/O, we poll it. If it was indeed ready, then tough luck, it has already started performing its expensive CPU operation.

By spawning CPU tasks, our futures are left essentially as state machines that perform "planning", i.e. they figure out which I/O to wait on, and which CPU can be spawned. Our scan can then be driven from worker threads using a smol executor to perform planning tasks, delegating I/O to our existing background dispatcher threads (we need more of these, see IoSource below), and interleaving work-stealing CPU task execution whenever we have performed enough planning to sufficiently fill our I/O queues.

## Handle

We introduce the idea of a `Handle` which allows our scheduling futures to spawn CPU work to be processed at some later stage.

```rust
trait Handle<'a> {
  fn spawn_task(&self, task: Box<dyn FnOnce(WorkerId) + Send>, hint: Option<WorkerId>) -> BoxFuture<'static, ()>;
}
```

## IoSource

An I/O source is tied to a background worker thread that drives a local runtime to exclusively drive I/O. Currently this is a single thread, but if we find we're reaching capacity, these threads should be able to launch additional workers.

The aim is that we can mark these I/O threads as high priority and hope to be scheduled as hyper-threads alongside workers. If not, in general, these threads aren't expected to be busy 100% of the time and we almost always want to interupt work to process I/O and unblock future planning. This assumption may prove to be false... in which case we can migrate to interleaving I/O scheduling onto worker threads, similar to how this is described in the AnyBlob paper.

I expect a separate I/O source to be created per set of resource constraints. For example, we may want a source for S3, separate from a source for GCS, separate from an EBS source, separate from an NVMe source. Each may be implemented using a different runtime, e.g. async-io, monoio, io_uring, etc.

The API is roughly:

```rust
trait IoSource: Clone {
  fn enqueue(&self, request: IoRequest) -> impl Future<()> // Async enqueue the request, returns pending if the queue is full (as determined by size / bytes relative to current measured throughput).

  fn metrics(&self) -> // Throughput / latency metrics. Useful for printing periodically.
}
```

## Threading Models

We support three main threading models for driving scans:
[...]

PR #4005, opened 2025-07-24
Remove scan async API

gatesn:
Related to https://github.com/vortex-data/vortex/issues/3890


joseph-isaacs (approved):
lg

vortex-scan/src/lib.rs
```diff
@@ -55,14 +49,12 @@ pub struct ScanBuilder<A> {
     /// The selection mask to apply to the selected row range.
     // TODO(joe): replace this is usage of row_id selection, see
     selection: Selection,
-    /// How to split the file for concurrent processing.
+    /// How to split the file f§    or concurrent processing.
```
robert3005, line 52: Something weird happened here

blaginin:
IMO, for prs like this, it would be really good to put some reasoning in the pr body 🙂

updating vortex for me is failing on a bunch of `into_array_stream` deleted and I can't easily pick up the philosophy i need to use with the new version

PR #4012, opened 2025-07-24
chore: validate ListArray::try_new arguments

lwwmanning:
TODO:

- [x] add new_unchecked
- [ ] audit calls to try_new and use new_unchecked where appropriate

### DIFF
--- a/vortex-array/src/arrays/list/mod.rs
+++ b/vortex-array/src/arrays/list/mod.rs
@@ -70,16 +70,13 @@ impl ListArray {
     ///
     /// A ListArray is valid if all of the following are true:
     /// - Offsets are sorted, non-empty, and non-negative.
-    /// - The first offset is less than or equal to the length of the elements array.
-    /// - The final offset points to an element in the elements list, pointing to zero if the elements list is empty.
+    /// - The final offset is less than or equal to the length of the elements array. If the elements array is empty, the final offset must be zero.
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/list/mod.rs
```diff
@@ -64,15 +65,14 @@ pub trait OffsetPType: NativePType + PrimInt + AsPrimitive<usize> + Into<Scalar>
...
-
 impl ListArray {
+    /// Creates a new ListArray from the provided elements, offsets, and validity.
+    ///
+    /// A ListArray is valid if all of the following are true:
+    /// - Offsets are sorted, non-empty, and non-negative.
+    /// - The first offset is less than or equal to the length of the elements array.
+    /// - The final offset points to an element in the elements list, pointing to zero if the elements list is empty.
+    /// - The size of the validity is the size-1 of the offset array.
     pub fn try_new(
```
robert3005, line 75: whenever you are adding validation like this you need to include `new_unchecked` version so cheaply recreate the list for operations that are internal that avoid running whole validation

gatesn: This is already eating into random access performance (in that case SparseArray::try_new).

So I think the decode impls should use unchecked construction, and just panic later?

robert3005:
This was done as part of removing results from arrayops

PR #4017, opened 2025-07-25
perf[bench]: dont delete duckdb.db between db runs

joseph-isaacs:


gatesn (approved):
Can this not use the idempotent wrapper?

PR #4018, opened 2025-07-25
Empty chunks don't affect arrays validity, empty Mask is always both all true and all false

robert3005:

### DIFF
--- a/vortex-array/src/arrays/chunked/compute/is_constant.rs
+++ b/vortex-array/src/arrays/chunked/compute/is_constant.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use vortex_error::{VortexExpect, VortexResult};
+use vortex_error::VortexResult;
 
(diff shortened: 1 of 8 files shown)


vortex-array/src/arrays/chunked/compute/is_constant.rs
```diff
@@ -13,9 +13,12 @@ impl IsConstantKernel for ChunkedVTable {
         array: &ChunkedArray,
         opts: &IsConstantOpts,
     ) -> VortexResult<Option<bool>> {
-        let mut chunks = array.chunks().iter().skip_while(|c| c.is_empty());
+        let mut chunks = array.non_empty_chunks();
 
-        let first_chunk = chunks.next().vortex_expect("Must have at least one value");
+        let Some(first_chunk) = chunks.next() else {
+            // Empty chunked array
+            return Ok(Some(false));
```
lwwmanning: Empty is constant

robert3005: len == 0 => not constant
when we added `as_constant` those were the semantics we agreed on. But really I should restore old code, there's entry point that checks len == 0

joseph-isaacs (approved):

PR #4019, opened 2025-07-25
Refactor read I/O

gatesn:
This makes VortexReadAt -> ReadAt a sealed trait, where the previous implementations use the TokioDispatched implementation.

Doing this lets us guarantee that futures returned from ReadAt are safe to run on any runtime (i.e. we can use executors::block_on to await them).


a10y:
> Doing this lets us guarantee that futures returned from ReadAt are safe to run on any runtime

Is there a reason this is a problem now? I thought this was always a property that the IoDispatcher gave us

gatesn:
> Is there a reason this is a problem now? I thought this was always a property that the IoDispatcher gave us

It does, the idea here is to encapsulate the dispatching inside the ReadAt trait, so anywhere we take `R: ReadAt` on the API, we guarantee it's been dispatched. Right now it's easy to use a TokioFile with a non-Tokio runtime for instance.

gatesn:
Going to abort this PR. There's something weird about having a free-roaming coalescing driver thread that means "dispatcher" is overloaded to mean "spawn this background task", and also "spawn a !Send future that can be runtime-aware".... One for another day.

PR #4022, opened 2025-07-25
chore: optionally delete the duckdb database on engine creation.

danking:
We previously deleted this file (see 45200f1). Reusing the file substantially improved TPC-H S3 SF=1 (but neither clickbench nor other scale factors).

It seems we should at least be able to choose this option.


lwwmanning (approved):

danking:
Made some changes requested by Joe over slack

PR #4031, opened 2025-07-28
zip compute function

onursatici:


vortex-array/src/compute/zip.rs
```diff
@@ -0,0 +1,244 @@
...
+            if_true,
+            if_false,
+        } = ZipArgs::try_from(args)?;
+
+        if mask.all_true() {
+            return Ok(if_true.to_array().into());
+        }
+
+        if mask.all_false() {
+            return Ok(if_false.to_array().into());
```
blaginin: you may potentially need to update nullability here?

blaginin (approved):
should we add a test for that?

vortex-array/src/compute/zip.rs
```diff
@@ -0,0 +1,244 @@
...
+use vortex_error::{VortexError, VortexResult, vortex_bail, vortex_err};
+use vortex_mask::{AllOr, Mask};
+
+use super::{ComputeFnVTable, InvocationArgs, Output, cast};
+use crate::builders::builder_with_capacity;
+use crate::compute::{ComputeFn, Kernel};
+use crate::vtable::VTable;
+use crate::{Array, ArrayRef};
+
+pub fn zip(mask: &Mask, if_true: &dyn Array, if_false: &dyn Array) -> VortexResult<ArrayRef> {
```
a10y, line 21: Possible to get a doc comment?

vortex-array/src/compute/zip.rs
```diff
@@ -0,0 +1,244 @@
...
+
+pub static ZIP_FN: LazyLock<ComputeFn> = LazyLock::new(|| {
+    let compute = ComputeFn::new("zip".into(), ArcRef::new_ref(&Zip));
+    for kernel in inventory::iter::<ZipKernelRef> {
+        compute.register_kernel(kernel.0.clone());
+    }
+    compute
+});
+
+pub struct Zip;
```
a10y: Does this need to be pub?

a10y:
I think Numpy calls this operation "[where](https://numpy.org/doc/stable/reference/generated/numpy.where.html)". I'm curious where zip came from

blaginin:
arrow uses zip

a10y:
Ah nice https://docs.rs/arrow/latest/arrow/compute/kernels/zip/fn.zip.html

👍

robert3005:
For future reader - numpy's where is a generalized zip for N-dimensions. They themseleves say that 1-D version is a zip.

blaginin:
btw, should we maybe add a helper to perform coalesce directly? which would build on top of this function

onursatici:
for coalesce I would hold off for now because I am not sure if the implementation using `zip` would be enough for us, if it turns out to be fine happy to add here

a10y:
@robert3005 FWIW the numpy docs do not say this is equivalent to `zip`, they merely show a way that you can use `zip` as part of implementing this for 1-D:

image

robert3005:
Indeed, I made a mental shortcut

PR #4035, opened 2025-07-28
use arc slice for buffers in varbinview

onursatici:
`varbinview.clone()` should be constant, with the Vec for buffers it was linear with the number buffers before

### DIFF
--- a/vortex-array/src/arrays/varbinview/mod.rs
+++ b/vortex-array/src/arrays/varbinview/mod.rs
@@ -379,14 +379,8 @@ impl VarBinViewArray {
 
     /// Iterate over the underlying raw data buffers, not including the views buffer.
     #[inline]
-    pub fn buffers(&self) -> &[ByteBuffer] {
+    pub fn buffers(&self) -> &Arc<[ByteBuffer]> {
         &self.buffers
-    }
-
-    /// Share ownership of the underlying raw data buffers, not including the views buffer.
-    #[inline]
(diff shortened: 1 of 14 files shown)


robert3005 (approved):

vortex-array/src/arrays/varbinview/mod.rs
```diff
@@ -316,7 +317,7 @@ impl VarBinViewArray {
 
         Ok(Self {
             dtype,
-            buffers,
```
robert3005: should the input buffers also be an arc slice? Right now you have to create copy of the buffers slice in order to create new one

vortex-array/src/arrays/varbinview/mod.rs
```diff
@@ -382,6 +383,12 @@ impl VarBinViewArray {
         &self.buffers
     }
 
+    /// Share ownership of the underlying raw data buffers, not including the views buffer.
+    #[inline]
+    pub(crate) fn buffers_arc(&self) -> Arc<[ByteBuffer]> {
```
robert3005: If you made `buffers(&self)` return `&Arc<[ByteBuffer]>` wouldn't that almost always work as intended? Since it could deref to `&[ByteBuffer]`

onursatici: yea thought about that but found it a bit weird to return a &Arc, but yea that should cover both cases

PR #4036, opened 2025-07-28
varbinview builder buffer deduplication

onursatici:


robert3005:
should this be optional feature of the builder? Not sure you care all the time to have it deduplicated

robert3005 (approved):

robert3005:
I wonder where the slowdown comes from if we kept the default the same?

onursatici:
@robert3005 I believe it was because of the extra branch I added per view, now moved it to once per `extend_from_array` call here: https://github.com/vortex-data/vortex/pull/4036/commits/28b120788fa1a1696c4e19d989694a10ab1c4f27

PR #4038, opened 2025-07-28
chore: Timeout data downloads in benchmarks

AdamGS:
Values are arbitrary and I assume are reasonable, open to suggestions if anyone thinks they should be more or less generous

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -142,7 +142,7 @@ rand_distr = "0.5"
 ratatui = "0.29"
 rayon = "1.10.0"
 regex = "1.11.0"
-reqwest = { version = "0.12.0", features = [
+reqwest = { version = "0.12.4", features = [
     "charset",
     "http2",
     "macos-system-configuration",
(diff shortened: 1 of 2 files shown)


danking (approved):
Sure

bench-vortex/src/clickbench/clickbench_benchmark.rs
```diff
@@ -80,7 +81,9 @@ impl Benchmark for ClickBenchBenchmark {
         match self.data_url.scheme() {
             "file" => {
                 let basepath = clickbench_flavor(self.flavor).to_data_path();
-                let client = reqwest::blocking::Client::default();
+                let client = reqwest::blocking::Client::builder()
+                    .timeout(Duration::from_secs(60 * 15))
```
a10y: can we be more aggressive? i bet we can cap this at 1 minute

AdamGS: yeah apparently for the blocking client its a different behavior and it defaults to 30 seconds, which I'm happy to keep

AdamGS: for the async client it is the total timeout, from initial connect to reading the full body, so I think the 15 minutes is reasonable but probably a bit generous

PR #4042, opened 2025-07-28
chore: add missing `{doctest}`

### LINKED ISSUE #3463: Python modules are private, docs aren't tested

blaginin:
Closes https://github.com/vortex-data/vortex/issues/3463

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,7 +15,7 @@ log = logging.getLogger("vortex.docs.conf")
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = "Vortex"
-copyright = "2025, Linux Foundation"
+copyright = "The Vortex contributors"
 author = "Vortex contributors"
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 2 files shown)


blaginin:
I wanted to force all python blocks to be docs but sphinx is so clumsy.... should we switch to mintlify

docs/conf.py
```diff
@@ -15,8 +15,8 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = "Vortex"
-copyright = "2024, Spiral"
-author = "Spiral"
+copyright = "2025, Linux Foundation"
```
lwwmanning: If this is the Copyright text for the docs, should be the same as source files, where we put "Copyright the Vortex contributors". (Original authors mostly retain copyrights to their contributions.)

The project name "Vortex" itself is "Copyright © Vortex a Series of LF Projects, LLC."

blaginin: Zen Browser 2025-07-28 21 55 44

ok like this?

lwwmanning (approved):

PR #4049, opened 2025-07-29
fix: Correctly calculate the size of FlatLayouts array flatbuffer size

### LINKED ISSUE #3977: FlatBuffer size doesn't seem correct in TUI
From what I can observe on develop it's always the same as the segment size which is not the information we want to display

robert3005:
This was lost in some refactors, we parse the flat layouts segment to pick up
the arrays metadata bytes

fix #3977

### DIFF
--- a/vortex-tui/src/browse/app.rs
+++ b/vortex-tui/src/browse/app.rs
@@ -7,13 +7,15 @@ use std::sync::Arc;
 use ratatui::prelude::Size;
 use ratatui::widgets::ListState;
 use vortex::dtype::DType;
-use vortex::error::{VortexExpect, VortexResult};
+use vortex::error::{VortexExpect, VortexResult, VortexUnwrap};
 use vortex::file::{Footer, SegmentSpec, VortexFile, VortexOpenOptions};
 use vortex::layout::LayoutRef;
 use vortex::layout::layouts::flat::FlatVTable;
 use vortex::layout::layouts::zoned::ZonedVTable;
-use vortex::layout::segments::SegmentId;
+use vortex::layout::segments::{SegmentId, SegmentSource};
+use vortex::serde::ArrayParts;
 
+use crate::TOKIO_RUNTIME;
 use crate::browse::ui::SegmentGridState;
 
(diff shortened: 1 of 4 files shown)


vortex-tui/src/browse/app.rs
```diff
@@ -72,22 +81,28 @@ impl LayoutCursor {
...
     pub fn flatbuffer_size(&self) -> usize {
         let segment_id = self.layout.as_::<FlatVTable>().segment_id();
-        self.segment_spec(segment_id).length as usize
+        let segment = TOKIO_RUNTIME
+            .block_on(async { self.segment_source.request(segment_id, &("".into())).await })
+            .vortex_unwrap();
+        ArrayParts::try_from(segment)
+            .vortex_unwrap()
+            .metadata()
+            .len()
```
onursatici, line 104: isn't this the size of the metadata buffer in the flatbuffer? Do we want to return the number of bytes in the entire flatbuffer or just the number of bytes in the metadata buffer in the root ArrayNode?

vortex-tui/src/browse/app.rs
```diff
@@ -72,22 +81,28 @@ impl LayoutCursor {
...
+        Self::new_with_path(self.footer.clone(), self.segment_source.clone(), path)
     }
 
     /// Get the size of the backing flatbuffer for this layout.
     ///
     /// NOTE: this is only safe to run against a FLAT layout.
     pub fn flatbuffer_size(&self) -> usize {
         let segment_id = self.layout.as_::<FlatVTable>().segment_id();
-        self.segment_spec(segment_id).length as usize
+        let segment = TOKIO_RUNTIME
```
gatesn: Just use futures block_on

vortex-tui/src/browse/app.rs
robert3005: This is only for FlatLayouts where we want to show size of the arrays flatbuffer. I think we might need to rethink what we show but for now this fixes the display value

onursatici (approved):

PR #4050, opened 2025-07-29
chore:  C++ API improvement

XinyuZeng:
1. split the single `lib.rs` to `read.rs` and `write.rs`.
2. Add an `IntoStreamDriver` API to the `ScanBuilder`, which internally is the thread-safe cloneable `WorkStealingArrayIterator`. I am struggled with the naming, calling it `StreamDriver` for now but  happy to change.
3. Add two builder option `WithIncludeByIndex` for row selection and `WithOutputSchema` to pass in the output Arrow schema. How should `WithOutputSchema` interact with projection is left for next PR.
4. Removed the stale `IntoArray` and thread-pool related configs.
5. Moved the testing data generation logic from Rust to C++.

part of #3720


vortex-cxx/CMakeLists.txt
```diff
@@ -24,14 +27,16 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(Corrosion)
 
+set(RUST_SOURCE_FILE lib.rs read.rs write.rs)
+
+# gen_test_data has been moved to C++ side and the following is deprecated.
```
0ax1: Remove the commented out code.

vortex-cxx/CMakeLists.txt
```diff
@@ -87,8 +91,9 @@ if (VORTEX_ENABLE_TESTING)
     FetchContent_MakeAvailable(googletest)
 
     enable_testing()
-    add_executable(vortex_cxx_test cpp/tests/basic_test.cpp)
+    add_executable(vortex_cxx_test cpp/tests/basic_test.cpp cpp/tests/test_data_generator.cpp)
     target_include_directories(vortex_cxx_test PUBLIC ${CPP_INCLUDE_DIRS})
+    target_include_directories(vortex_cxx_test PRIVATE cpp/tests)
     target_link_libraries(vortex_cxx_test PRIVATE gtest_main vortex nanoarrow_static)
     # unit tests link to the bridge for the gen_test_data function only
```
0ax1: This comment is outdated now right?

vortex-cxx/cpp/include/vortex/file.hpp
```diff
@@ -5,11 +5,7 @@
 
 #include <nanoarrow/common/inline_types.h>
 
-#include <cstdint>
 #include <memory>
-#include <string>
-#include <string_view>
```
0ax1, line 11: Appreciate the include cleanup. :)

vortex-cxx/cpp/include/vortex/scan.hpp
```diff
@@ -7,9 +7,40 @@
...
+    ///
+    /// Within each thread, the record batches will be emitted in the original order they are within
+    /// the scan. Between threads, the order is not guaranteed.
+    ///
+    /// Example: If the scan contains batches [b0, b1, b2, b3, b4, b5] and two threads call this
+    /// function respectively to make progress on their own stream, Thread 1 might receive [b0,
+    /// b2, b4] and Thread 2 might receive [b1, b3, b5]. Each thread maintains order within its
+    /// subset, but overall ordering between threads is not guaranteed (e.g., Thread 2 could emit b1
+    /// before Thread 1 emits b0).
+    ArrowArrayStream CreateArrayStream() const;
```
0ax1, line 36: I think it's worth adding that this under the hood hooks into the `WorksStealingIterator` which in this case returns batches via `builder.inner.into_record_batch_reader(schema)?;`.

vortex-cxx/cpp/include/vortex/scan.hpp
```diff
@@ -20,19 +51,24 @@ class ScanBuilder {
...
+    ScanBuilder &WithIncludeByIndex(const uint64_t *indices, std::size_t size);
+
+    /// Set the limit on the number of rows to scan out.
     ScanBuilder &WithLimit(uint64_t limit);
 
-    /// Consume the scan builder to a stream of record batches.
-    /// The scan builder is consumed and cannot be used after this call.
+    /// Set the output schema on the scan builder.
+    /// TODO: should decide to input full schema or schema after adding projection.
+    ScanBuilder &WithOutputSchema(ArrowSchema &output_schema);
```
0ax1: Just a thought: Might be nice to express the factory functions in terms of move.

`ScanBuilder &&WithOutputSchema(ArrowSchema &output_schema) &&;`

vortex-cxx/cpp/tests/basic_test.cpp
```diff
@@ -36,164 +40,298 @@ class VortexTest : public ::testing::Test {
...
+
+        // Basic properties validation
+        ASSERT_EQ(actual_schema->n_children, ref_schema->n_children);
+
+        auto actual_view = CreateArrayView(actual_array, actual_schema);
+        auto ref_view = CreateArrayView(ref_array, ref_schema);
+
+        if (row_indices.empty()) {
+            // Full array comparison
+            ASSERT_EQ(actual_array->length, ref_array->length);
```
0ax1: Let's extract the logic in both cases into 2 new functions.

0ax1: Overall, the logic is a tad too complex for a unit test for my taste.

0ax1 (commented):
Thanks for spending some time on polishing the C++ API! 

Looks good overall from my end. Left a couple of remarks inline.

vortex-cxx/cpp/tests/basic_test.cpp
XinyuZeng: Indeed, I actually feel the unit tests do not need to be very sophisticated because we already test the function on the Rust side.

XinyuZeng: I addressed the reviews. PTAL if you have time

0ax1 (approved):
Thanks for iterating on this! I think we're good to go landing this PR.

A good follow up would be adding one or multiple ASAN options `option(VORTEX_ENABLE_ASAN "Enable AddressSanitizer" OFF)` to run the example and unit tests with on CI.

vortex-cxx/cpp/src/scan.cpp
```diff
@@ -47,15 +67,41 @@ ArrowArrayStream ScanBuilder::IntoStream() {
...
+        return StreamDriver(std::make_unique<StreamDriver::Impl>(std::move(reader)));
     } catch (const rust::cxxbridge1::Error &e) {
         throw VortexException(e.what());
     }
 }
+
+StreamDriver::StreamDriver(std::unique_ptr<Impl> impl) : impl_(std::move(impl)) {
+}
+
+StreamDriver::StreamDriver(StreamDriver &&other) noexcept : impl_(std::move(other.impl_)) {
```
0ax1, line 83: I think the move ctor could be: `StreamDriver::StreamDriver(StreamDriver &&other) noexcept = default;`.

vortex-cxx/src/read.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/// File operations - using blocking operations for simplicity
+/// TODO(xinyu): object store (see vortex-ffi)
+fn open_file(path: &str) -> Result<Box<VortexFile>, Box<dyn std::error::Error + Send + Sync>> {
+    let file = VortexOpenOptions::file().open_blocking(std::path::Path::new(path))?;
+    Ok(Box::new(VortexFile { inner: file }))
+}
+
+fn file_row_count(file: &VortexFile) -> u64 {
```
0ax1, line 64: I guess this should be def as a member func on `VortexFile`.

vortex-cxx/cpp/src/scan.cpp
XinyuZeng: To clean it up in next PR :)

vortex-cxx/src/read.rs
XinyuZeng: Also, gonna clean it up in next PR

vortex-cxx/cpp/include/vortex/scan.hpp
XinyuZeng: I recently encountered a problem. Suppose we want to conditionally pass options to the builder, like a similar pattern in Rust
```cpp
auto builder = file.CreateScanBuilder();
if (limit > -1) {
  builder = std::move(builder).WithLimit(limit);
}
```
This looks ok, but because we return rvalue ref in `WithLimit`, this assignment is actually a **self-move-assignment**`builder = std::move(builder)`, which clears everything inside the builder.

To correctly use this API, the caller should do:
```cpp
auto builder = file.CreateScanBuilder();
if (limit > -1) {
  std::move(builder).WithLimit(limit);
}
```
which, in my opinion, isn't canonical on the C++ convention and confuses the users.

Two solutions to allow you actually have the code `builder = std::move(builder).WithLimit(limit);`, and at the same time can do function chaining like in Rust:
1. use a custom move assignment ctor to prevent self-move-assignment
```cpp
ScanBuilder &operator=(ScanBuilder &&other) noexcept {
  if (this != &other) {
    impl_ = std::move(other.impl_);
  }
  return *this;
} 
```

2. Return object
```cpp
ScanBuilder WithOutputSchema(ArrowSchema &output_schema) &&;
```

I prefer the second one. As the first one introduces a branch. The second one requires an object creation. However, I think the `ScanBuilder` is quite simple (only has one `impl_` member) and the compiler can certainly do RVO to reduce the cost. What are your thoughts? @0ax1

XinyuZeng: I am now leaning towards returning lvalue-ref as we previously did. Returning object still has an 8-byte copy of the inside `rust::Box`. I don't think we need to make the API rusty. It can just be the C++ way.

0ax1: Sorry just saw this. 👀 @XinyuZeng

0ax1: Implemented here: https://github.com/vortex-data/vortex/pull/4306

PR #4051, opened 2025-07-29
fix: benchmark compress hang

0ax1:


vortex-scan/src/multi_thread.rs
```diff
@@ -74,12 +74,20 @@ impl ScanBuilder<ArrayRef> {
             //  head-room to run ahead and figure out the I/O demands of subsequent tasks.
             .buffered(num_workers * concurrency);
 
-        Ok(
-            iter::from_fn(move || block_on(stream.next())).filter_map(|result| {
+        Ok(Box::new(
```
danking: do you need the box?

0ax1: gone

onursatici (commented):
we also use `futures::executor::block_on` on the work stealing array iterator. I don't think now we do use that within a tokio runtime, but should we add the same fix there for the world we start using that for non-duckdb scans?

0ax1:
> we also use `futures::executor::block_on` on the work stealing array iterator. I don't think now we do use that within a tokio runtime, but should we add the same fix there for the world we start using that for non-duckdb scans?

Yeah, I guess that's worth considering. Mind if I split this from this PR though?

robert3005 (approved):

danking:
weird

> tpch_q22/DataFusion:vortex-file-compressed | 12531250 | 1.64994e+07 | 0.759497

danking:
compress appears fixed though still quite slow

danking (approved):

a10y (approved):

0ax1:
fails on: https://github.com/vortex-data/vortex/actions/runs/16599758689/job/46955816468?pr=4051#step:9:93

which will be fixed once https://github.com/duckdb/community-extensions/pull/517 lands

vortex-scan/src/multi_thread.rs
```diff
@@ -74,12 +74,18 @@ impl ScanBuilder<ArrayRef> {
...
-                    .map_err(|e| vortex_err!("Failed to join on a spawned scan task {e}"))
-                    .vortex_expect("Failed to join on a spawned scan task")
-            }),
-        )
+        Ok(iter::from_fn(move || {
+            // Use runtime-aware blocking strategy
+            if tokio::runtime::Handle::try_current().is_ok() {
+                tokio::task::block_in_place(|| CPU_RUNTIME.handle().block_on(stream.next()))
+            } else {
+                block_on(stream.next())
```
gatesn, line 82: Isn't the bug that this should always be CPU_RUNTIME.block_on and not futures::executor::block_on? 

It seems even weird to use whatever the current tokio handle is?

PR #4054, opened 2025-07-29
varbinview zip kernel

onursatici:
Uses the buffer deduplicating builder to construct the zipped array. This guards against the pathological case where we are zipping two varbinview arrays with a mask that has lots of contiguous slices. Each `builder.extend_from_array(input.slice(..))` would duplicate the entire buffers of `input`, and each slice in the mask would add the same buffers to the result array over and over again.

### DIFF
--- a/vortex-array/src/arrays/varbinview/compute/zip.rs
+++ b/vortex-array/src/arrays/varbinview/compute/zip.rs
@@ -3,26 +3,25 @@
 
 use vortex_error::VortexResult;
 
-use crate::arrays::VarBinViewArray;
+use crate::arrays::{VarBinViewArray, VarBinViewVTable};
 use crate::builders::VarBinViewBuilder;
-use crate::compute::{ZipKernelAdapter, zip_impl_with_builder, zip_return_dtype};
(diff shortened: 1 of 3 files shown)


blaginin (approved):
should we add tests 🥺

vortex-array/src/arrays/varbinview/compute/zip.rs
```diff
@@ -32,3 +31,45 @@ impl ZipKernel for VarBinViewVTable {
...
+            }
+            builder.finish()
+        };
+
+        // [1,2,4,5,7,8,..]
+        let mask = Mask::from_indices(100, (0..100).filter(|i| i % 3 != 0).collect());
+
+        let zipped = zip(&if_true, &if_false, &mask).unwrap();
+        let zipped = zipped.as_opt::<VarBinViewVTable>().unwrap();
+        assert_eq!(zipped.nbuffers(), 2);
```
blaginin, line 75: i wonder if we should do snapshot testing for cases like this like in spiral 🤔 so it'll be easier to assert the actual values

blaginin (approved):
nice!

PR #4057, opened 2025-07-29
feat: `ArrayEquals` kernel

### LINKED ISSUE #1424: Implement array_equals compute fn and PartialEq for ArrayData

blaginin:
Closes https://github.com/vortex-data/vortex/issues/1424


vortex-scalar/src/scalar_value/mod.rs
```diff
@@ -43,6 +43,23 @@ pub(crate) enum InnerScalarValue {
     List(Arc<[ScalarValue]>),
 }
 
+impl PartialEq for ScalarValue {
```
robert3005: We generally held that comparing scalarvalues is meaningless and you should compare them via Scalar wrapper. You can see PartialEq for Scalar. The problem you run into is that while two scalar values might be trivially different they should be considered the same given a dtype

blaginin: fair! but then is there any reasons for `StatsProvider::get` to return `ScalarValue` and not `Scalar`?

robert3005: It's meant to be a optimisation to make stats smaller. We don't have to store the dtype and you can always grab a dtype from the array.

robert3005:
I always found this weird to be a kernel. I think once we land #4188 this will be easier to implement in a reasonable way. Otherwise I would define this as PartialEq for Canonical  since you really shouldn't have to dispatch this to the encoding

blaginin:
Agreed! Joe and Nick mentioned the kernel trait may change quite a bit, so holding this off until it's merged

joseph-isaacs:
I would say define partial eq for canonical. I think this will be useful

PR #4058, opened 2025-07-29
Use correct block-on in multithread iter

gatesn:
@0ax1 I believe we should just always use the CPU runtime since that's how we spawned the tasks in the first place?

### DIFF
--- a/vortex-scan/src/multi_thread.rs
+++ b/vortex-scan/src/multi_thread.rs
@@ -4,7 +4,6 @@
 use std::iter;
 use std::sync::{Arc, LazyLock};
 
-use futures::executor::block_on;
 use futures::{StreamExt, stream};
 use tokio::runtime::{Builder, Runtime};
 use vortex_array::ArrayRef;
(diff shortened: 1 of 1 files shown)


0ax1:
Sounds reasonable. Gonna give it a ✔️ after the benchmarks pass.

gatesn:
Ooooo the failure is interesting. Seems like we cannot launch the CPU_RUNTIME if we're already inside a runtime. Which doesn't really make sense to me unless Tokio is full of global state...

I guess we need to find a different runtime/way to run our multi-threaded scans...

0ax1:
> Ooooo the failure is interesting. Seems like we cannot launch the CPU_RUNTIME if we're already inside a runtime. Which doesn't really make sense to me unless Tokio is full of global state...
> 
> I guess we need to find a different runtime/way to run our multi-threaded scans...

This should work if we keep the outer: `tokio::task::block_in_place`. It is possible to re-enter an async context from within `block_in_place`.

gatesn:
Isn't it to do with who first accesses the lazy lock runtime? If that's already within a tokio runtime, the it fails.

0ax1:
> Isn't it to do with who first accesses the lazy lock runtime? If that's already within a tokio runtime, the it fails.

Double checking here: https://github.com/vortex-data/vortex/actions/runs/16608222655/job/46985371641?pr=4059

gatesn:
I mean, it seems like it fixes it 🤷 doesn't make sense to me based on the docs for block_in_place... I think we should still follow up with a thread-pool based version of this. But it can be later.

PR #4068, opened 2025-07-30
Clean up decompress benchmarks

gatesn:


bench-vortex/src/compress/vortex.rs
```diff
@@ -23,11 +21,14 @@ pub async fn vortex_compress_write(array: &dyn Array, buf: &mut Vec<u8>) -> anyh
...
 #[inline(never)]
-pub async fn vortex_decompress_read(buf: Bytes) -> anyhow::Result<Vec<RecordBatch>> {
+pub async fn vortex_decompress_read(buf: Bytes) -> anyhow::Result<usize> {
     let scan = VortexOpenOptions::in_memory().open(buf)?.scan()?;
     let schema = Arc::new(scan.dtype()?.to_arrow_schema()?);
 
-    Ok(scan
-        .into_record_batch_reader_multithread(schema)?
-        .try_collect()?)
+    let iter = scan.into_record_batch_reader_multithread(schema)?;
```
gatesn, line 28: This is now the same code as Parquet, otherwise we end up benchmarking how fast we can write back to main memory.

danking (approved):

PR #4071, opened 2025-07-30
fix: compression benchmark short argument must not conflict

danking:
Rather than error, clap appears to give the short option to the first argument starting with that letter rather.


danking:
Benchmarks currently use `-o` to specify the output file. `--ops` has a default argument which is both compression and decompression, as intended by the benchmarks. Clap thinks we are trying to set the ops to "compress.json".

lwwmanning:
duplicate of #4070

Issue #3539, opened 2025-06-16
Dict array different nullability code and values


joseph-isaacs:
We should allow dict array have to have different nullability in the codes and values, where the dict itself, is the union of the children nullability.

This will allow more flexibility to the dict array, and allow optimisations if the codes or values a NonNullable.

We could push down compute of the values of a dict if the codes are non-nullable

joseph-isaacs:
Maybe we should limit nullability to just one child array?

onursatici:
before https://github.com/vortex-data/vortex/pull/2341 It was like that I believe

gatesn:
We need to support nullability in both if we want compatibility with Arrow, it also tends to be more powerful in general.

I'm not convinced either by optimisation opportunities, since there's no runtime difference in NonNull vs AllValid.

joseph-isaacs:
In a dict layout reader if the codes are non-nullable we can push down an expr over the values without needing to fetch the codes and examine the validity. 

We can only do this for nullable dicts if the codes can be non nullable

joseph-isaacs:
Why did you make the change @onursatici linked?

robert3005:
I don't think this leads to any meaningful simpler interface, all it does is make it so things that are nullable but allvalid do not get optimised. So imho we need a way to have nullability casts be supported. For sake of pushdown all we need to do is `&& NullCount(x) == 0` otherwise you will still skip AllValid cases

gatesn:
We should revive this.

* I think we should support nulls in _either_ values or codes. We will need to store extra metadata to signify this.
* During compression / layout, we should prefer to put nulls in the values, rather than codes. This allows expressions to be fully evaluated over just the values, with codes being applied back on top.

robert3005:
We have made those changes

PR #4074, opened 2025-07-31
Remove FilterEvaluation

gatesn:
* Use same projection logic to evaluate filter expressions.
* Some other speeds ups e.g. to Dict::fill_null
* Starts to fix up dict layout reader to push-down over values (required Dict(Struct)::canonicalize), but blocked on https://github.com/vortex-data/vortex/issues/3539


vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -93,7 +93,14 @@ impl DictReader {
             .entry(expr.clone())
             .or_insert_with(|| {
                 self.values_array()
-                    .map(move |array| expr.evaluate(&Scope::new(array?)).map_err(Arc::new))
+                    .map(move |array| {
+                        // TODO(ngates): recursive canonicalize?
```
robert3005: why the canonicalize? Because we do the same thing twice?

PR #4075, opened 2025-07-31
feat: Enable protobuf names to enable working with google.protobuf.Any

AdamGS:
This change is required to enable serializing arbitrary messages through the well-known [Any](https://docs.rs/prost-types/latest/prost_types/struct.Any.html#method.from_msg) type (which I'm pretty sure will be useful for something i'm working on).


AdamGS:
this is actually a binary break! so closing the PR

PR #4077, opened 2025-07-31
feat: Add serialization support for Select expression

AdamGS:


vortex-expr/src/exprs/select.rs
```diff
@@ -171,10 +186,9 @@ impl SelectField {
     }
 
     pub fn fields(&self) -> &FieldNames {
-        match self {
-            SelectField::Include(fields) => fields,
-            SelectField::Exclude(fields) => fields,
-        }
+        let (SelectField::Include(fields) | SelectField::Exclude(fields)) = self;
```
a10y, line 188: I don't necessarily like this better, but it is cute

a10y (approved):

vortex-expr/src/exprs/select.rs
AdamGS: honestly, I was just wondering if it works and thought it was cool

PR #4080, opened 2025-07-31
chore: No longer need to mount debugfs and tracefs

robert3005:


robert3005:
Since https://github.com/runs-on/runner-images-for-aws/releases/tag/linux-20250731123551 got released we no longer need to do the manual thing

AdamGS (approved):

PR #4081, opened 2025-07-31
chore: Make vortex pretty

AdamGS:
Adds a top-level feature to enable pretty-printing features in the underlying crates.


blaginin (approved):

blaginin:
pretty is a strong word 😀 you sure it's shouldn't be like "tabular"?

AdamGS:
I don't shame anythings for how they look

PR #4082, opened 2025-07-31
fix: Add select-expr deserialization

AdamGS:
Apparantly I got too excited by fancy pattern matching in #4077 to remember I also need to deserialize expressions


a10y (approved):
oh yea

AdamGS:
I'm working on a followup PR that will have tests to check for that

PR #4086, opened 2025-07-31
feat: implement Cast Kernel everywhere

lwwmanning:
⏺ Summary of Changes

  The changes implement comprehensive cast conformance testing across the Vortex array system:

  Core Changes:
  - Added a new cast conformance test framework in vortex-array/src/compute/conformance/cast.rs (541 lines)
  - Implemented CastKernel for 17 different array encodings across the codebase

  Encodings with Cast Support Added:
  - Compressed formats: ALP, BitPacking, FOR, FSST, RunEnd, Sparse, ZigZag
  - Core types: Bool, Chunked, Constant, Decimal, Extension, List, Null, Primitive, Struct, VarBin, VarBinView
  - Dictionary encoding: Dict
  - Temporal: DateTime-Parts

  Key Features:
  - Each encoding implements optimized casting logic specific to its format
  - Tests cover numeric type conversions (widening/narrowing), nullability changes, string conversions, and edge cases
  - Many encodings can perform casts without full decompression when possible (e.g., ZigZag preserves encoding for nullability-only changes)

  The implementation ensures type safety and performance across Vortex's compressed array formats while maintaining data integrity during type conversions.

### DIFF
--- a/encodings/dict/src/compute/cast.rs
+++ b/encodings/dict/src/compute/cast.rs
@@ -25,10 +25,10 @@ register_kernel!(CastKernelAdapter(DictVTable).lift());
 #[cfg(test)]
 mod tests {
     use rstest::rstest;
+    use vortex_array::IntoArray;
     use vortex_array::arrays::PrimitiveArray;
     use vortex_array::compute::cast;
     use vortex_array::compute::conformance::cast::test_cast_conformance;
-    use vortex_array::IntoArray;
     use vortex_buffer::buffer;
     use vortex_dtype::{DType, Nullability, PType};
 
--- a/encodings/dict/src/compute/cast.rs
+++ b/encodings/dict/src/compute/cast.rs
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_array::compute::{CastKernel, CastKernelAdapter, cast};
+use vortex_array::{ArrayRef, IntoArray, register_kernel};
+use vortex_dtype::DType;
+use vortex_error::VortexResult;
+
+use crate::{DictArray, DictVTable};
+
+impl CastKernel for DictVTable {
+    fn cast(&self, array: &DictArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        // Dictionary encoding doesn't change the logical type, so we delegate to the decoded array
+        let decoded = array.to_canonical()?.into_array();
+        cast(&decoded, dtype).map(Some)
+    }
+}
+
+register_kernel!(CastKernelAdapter(DictVTable).lift());
+
+#[cfg(test)]
+mod tests {
+    use rstest::rstest;
+    use vortex_array::arrays::PrimitiveArray;
+    use vortex_array::compute::cast;
+    use vortex_array::compute::conformance::cast::test_cast_conformance;
+    use vortex_array::{IntoArray, ToCanonical};
+    use vortex_buffer::buffer;
+    use vortex_dtype::{DType, Nullability, PType};
+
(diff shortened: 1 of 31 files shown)


lwwmanning:
@claude please review

encodings/dict/src/compute/cast.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl CastKernel for DictVTable {
+    fn cast(&self, array: &DictArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        // Cast the dictionary values to the target type
+        let casted_values = cast(array.values(), dtype)?;
+
+        // Create a new dictionary array with the same codes but casted values
+        Ok(Some(
+            DictArray::try_new(array.codes().clone(), casted_values)?.into_array(),
+        ))
```
robert3005: until we support mixed nullability dict children you have to also cast the codes

robert3005: Basically if you write a test that casts AllValid dict to NonNullable or back this will fail

encodings/fastlanes/src/bitpacking/compute/cast.rs
```diff
@@ -0,0 +1,109 @@
...
+                .validity()
+                .clone()
+                .cast_nullability(dtype.nullability())?;
+            return Ok(Some(
+                unsafe {
+                    BitPackedArray::new_unchecked_with_offset(
+                        array.packed().clone(),
+                        dtype.as_ptype(),
+                        new_validity,
+                        match array.patches() {
```
robert3005: this is just a map

vortex-array/src/arrays/chunked/compute/cast.rs
```diff
@@ -65,4 +68,31 @@ mod test {
...
+            PrimitiveArray::from_option_iter([Some(1.5f32), None, Some(2.5)]).into_array(),
+            PrimitiveArray::from_option_iter([Some(3.5f32), Some(4.5)]).into_array()
+        ],
+        DType::Primitive(PType::F32, Nullability::Nullable)
+    ).unwrap().into_array())]
+    #[case(ChunkedArray::try_new(
+        vec![buffer![42u8].into_array()],
+        DType::Primitive(PType::U8, Nullability::NonNullable)
+    ).unwrap().into_array())]
+    fn test_cast_chunked_conformance(#[case] array: crate::ArrayRef) {
```
robert3005, line 95: Meta point: You should be able to express all of these with a common set of arrays that you want to convert to the target encoding and then run the test on and compare to the result of applying it on the original

vortex-array/src/arrays/null/compute/cast.rs
```diff
@@ -0,0 +1,58 @@
...
+use vortex_error::{VortexResult, vortex_bail};
+
+use crate::arrays::{NullArray, NullVTable};
+use crate::compute::{CastKernel, CastKernelAdapter};
+use crate::{ArrayRef, register_kernel};
+
+impl CastKernel for NullVTable {
+    fn cast(&self, array: &NullArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        // Null can only be cast to Null
+        match dtype {
```
robert3005: I think you can support arbitrary cast here? You just create AllInvalid array of a given dtype matching the length? I wonder what arrow does for these. Maybe we should remove this array and say null arrays don't roundtrip

robert3005 (changes requested):
I think you have one bug. It would be nice to make sure every array is using the sample input arrays

robert3005 (approved):

joseph-isaacs:
Did we not run the benchmarks?

robert3005:
We should note require benchmarks to run but clearly looks like we are missing some coverage

PR #4089, opened 2025-08-01
Inline the logic of the filter layout reader

gatesn:
Follow up work is to store expressions as `Vec<Vec<Expr>>` (including the projection expression as the last elemet). Then if an expression ever hits 0% selectivity, we extract it into its own round trip, e.g. `select C where A == 10 and B > 2` would become `[[A == 10], [B > 2, C]]` if `A == 10` proves to be very selective.

This allows us to delay I/O for columns B and C until we have evaluated over A.


vortex-layout/src/layouts/zoned/reader.rs
```diff
@@ -11,23 +11,22 @@ use dashmap::DashMap;
 use futures::future::{BoxFuture, Shared};
```
gatesn: I reverted this back to before the dynamic filter PR. This is handled in the scan/filter now

robert3005 (approved):

gatesn:
Just going to wait for benchmarks to make sure dynamic expressions still work

vortex-scan/src/filter.rs
joseph-isaacs, line 1: Maybe this is out of scope, but maybe we want to also randomly try new orders, to not get stuck in a local maxima

gatesn:
Ok, dynamic filters are fixed

PR #4107, opened 2025-08-04
TransposedStream in struct writer to wake other potentially waiting streams

onursatici:
Previously other `TransposedStream` instances might have stuck hanging because they previously returned `Poll::Pending`

### DIFF
--- a/vortex-tui/src/convert.rs
+++ b/vortex-tui/src/convert.rs
@@ -9,6 +9,7 @@ use futures_util::StreamExt;
 use indicatif::ProgressBar;
 use parquet::arrow::ParquetRecordBatchStreamBuilder;
 use tokio::fs::File;
+use tokio::runtime::Handle;
 use vortex::ArrayRef;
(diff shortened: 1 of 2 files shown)


vortex-tui/src/convert.rs
```diff
@@ -76,12 +76,12 @@ pub async fn exec_convert(flags: Flags) -> VortexResult<()> {
             .boxed();
     }
 
-    let executor = Arc::new(LocalExecutor);
```
onursatici, line 79: This change is separate, but I also included it here because we are already in a tokio runtime at this point, so we can just use the workers there for conversion

robert3005 (approved):

joseph-isaacs:
Can we check that this allow `vx convert` for clickbench hits (pq -> vortex)

onursatici:
tested on a hits.parquet file in clickbench_partitioned and worked as expected

robert3005:
I validated locally that it works for clickbench files

PR #4113, opened 2025-08-05
fix: error-prone f16 Scalar serde

lwwmanning:
⏺ Summary of Changes

  The diff shows comprehensive changes to fix scalar casting bugs and ensure backwards compatibility for serialized scalar values. Here are the
  key changes:

  1. Core Bug Fix: Floating-Point Conversion (vortex-scalar/src/pvalue.rs)

  - Fixed critical bug where u64 to f64 conversion was using bit reinterpretation (f64::from_bits()) instead of numeric conversion
  - Changed to use NumCast::from() for proper numeric conversions

  2. Type Coercion System (vortex-scalar/src/lib.rs)

  - Added coerce_value() method in Scalar::new() to handle backwards compatibility
  - Specifically coerces f16 values stored as u64 (for backwards compatibility with serialized data)
  - Includes recursive coercion for structs and lists containing f16 fields
  - Protected against overflow (only coerces u64 to f16 if value ≤ u16::MAX)

  3. Protobuf Support for f16 (vortex-proto/proto/scalar.proto)

  - Added dedicated f16_value field to protobuf schema (field 10)
  - Updated serialization to use this dedicated field instead of storing f16 as u64

  4. Consistent Scalar Construction

  - Updated all direct Scalar { dtype, value } constructions to use Scalar::new()
  - Ensures coercion logic is always applied
  - Changed in:
    - bool.rs: into_scalar(), bool(), From<bool>
    - primitive.rs: primitive_value(), From<$T> for Scalar
    - struct_.rs: field_by_idx(), cast(), struct_()
    - list.rs: element(), list(), list_empty()

  5. Test Coverage (vortex-scalar/src/lib.rs and proto.rs)

  - Fixed test_canonicalize_scalar_values to create f16 values properly
  - Added comprehensive coercion tests:
    - f16 coercion from u64
    - No coercion for other type combinations
    - Struct field coercion
    - List element coercion
    - Overflow protection
  - Added ScalarValue protobuf roundtrip tests:
    - Direct f16 roundtripping
    - Value preservation for all primitive types
    - Handling of type consolidation (u8/u16/u32 → u64, i8/i16/i32 → i64)

  6. Key Design Decision

  - Only f16 from u64 is coerced (minimal approach for backwards compatibility)
  - Other floating-point types (f32, f64) are NOT coerced
  - This preserves backwards compatibility while fixing the root cause

  The changes ensure that:
  1. Scalar casting now uses numeric conversion instead of bit reinterpretation
  2. Old serialized f16 values (stored as u64) are properly handled
  3. All scalar constructions go through the coercion logic
  4. Comprehensive tests verify the behavior

### DIFF
--- a/vortex-array/src/arrays/constant/canonical.rs
+++ b/vortex-array/src/arrays/constant/canonical.rs
@@ -287,14 +287,14 @@ mod tests {
 
     #[test]
     fn test_canonicalize_scalar_values() {
-        let f16_scalar = Scalar::primitive(f16::from_f32(5.722046e-6), Nullability::NonNullable);
-        let scalar = Scalar::new(
-            DType::Primitive(PType::F16, Nullability::NonNullable),
-            Scalar::primitive(96u8, Nullability::NonNullable).into_value(),
-        );
-        let const_array = ConstantArray::new(scalar.clone(), 1).into_array();
+        let f16_value = f16::from_f32(5.722046e-6);
+        let f16_scalar = Scalar::primitive(f16_value, Nullability::NonNullable);
+
+        // Create a ConstantArray with the f16 scalar
+        let const_array = ConstantArray::new(f16_scalar.clone(), 1).into_array();
         let canonical_const = const_array.to_primitive().unwrap();
-        assert_eq!(canonical_const.scalar_at(0).unwrap(), scalar);
+
+        // Verify the scalar value is preserved through canonicalization
         assert_eq!(canonical_const.scalar_at(0).unwrap(), f16_scalar);
     }
 
--- a/vortex-proto/proto/scalar.proto
+++ b/vortex-proto/proto/scalar.proto
@@ -27,6 +27,7 @@ message ScalarValue {
     string string_value = 7;
     bytes bytes_value = 8;
     ListValue list_value = 9;
+    uint64 f16_value = 10;
   }
 }
 
--- a/vortex-scalar/src/lib.rs
+++ b/vortex-scalar/src/lib.rs
@@ -1008,4 +1008,111 @@ mod tests {
             _ => panic!("Expected U64 value to remain unchanged when too large for F16"),
         }
     }
+
+    #[test]
+    fn test_extension_dtype_coercion() {
+        // Create an extension type with f16 storage
+        let ext_id = ExtID::new("test_f16_ext".into());
+        let storage_dtype = Arc::new(DType::Primitive(PType::F16, Nullability::NonNullable));
+        let ext_dtype = Arc::new(ExtDType::new(ext_id, storage_dtype, None));
+        
+        // Test f16 value stored as u64 gets coerced through extension type
+        let f16_value = f16::from_f32(0.42);
+        let u64_bits = f16_value.to_bits() as u64;
+        
+        let scalar = Scalar::new(
+            DType::Extension(ext_dtype.clone()),
+            ScalarValue(InnerScalarValue::Primitive(PValue::U64(u64_bits))),
+        );
+        
+        // Verify the value was coerced to f16
+        match scalar.value() {
+            ScalarValue(InnerScalarValue::Primitive(PValue::F16(v))) => {
+                assert_eq!(*v, f16_value);
+            }
+            _ => panic!("Expected F16 value after extension type coercion"),
+        }
+    }
+
+    #[test]
+    fn test_extension_dtype_no_coercion() {
+        // Create an extension type with u32 storage
+        let ext_id = ExtID::new("test_u32_ext".into());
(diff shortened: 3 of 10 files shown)


lwwmanning:
@claude could you please review this PR?

robert3005:
Which part is the fix here? Why do u8, u16, i8 and i16 don't have the same kind of bug?

robert3005:
Can we instead of coercing scalars separate the conversion logic and cast logic, i.e. the logic for converting u64 scalar value to f16 scalar is different than casting u64 scalar to f64

vortex-proto/proto/scalar.proto
```diff
@@ -27,6 +27,7 @@ message ScalarValue {
     string string_value = 7;
     bytes bytes_value = 8;
     ListValue list_value = 9;
+    uint64 f16_value = 10;
```
joseph-isaacs, line 30: Why is this uint64 and not bytes or something smaller. I guess its all the same size due to the kind

vortex-scalar/src/lib.rs
```diff
@@ -66,12 +66,70 @@ pub struct Scalar {
...
+            (DType::List(elem_dtype, _), InnerScalarValue::List(elements)) => {
+                let coerced_elements: Result<Vec<ScalarValue>, _> = elements
+                    .iter()
+                    .map(|elem| Self::coerce_value(elem_dtype, elem.clone()))
+                    .collect();
+                Ok(ScalarValue(InnerScalarValue::List(
+                    coerced_elements?.into(),
+                )))
+            }
+            // No coercion needed for other types
```
joseph-isaacs, line 132: Do we need a todo for extension types?

vortex-scalar/src/pvalue.rs
```diff
@@ -309,16 +309,18 @@ impl TryFrom<PValue> for f64 {
...
-        // We serialize f64 as u64, but this can also sometimes be narrowed down to u8 if e.g. == 0
         match value {
-            PValue::U8(u) => Some(Self::from_bits(u as u64)),
-            PValue::U16(u) => Some(Self::from_bits(u as u64)),
-            PValue::U32(u) => Some(Self::from_bits(u as u64)),
-            PValue::U64(u) => Some(Self::from_bits(u)),
+            PValue::U8(u) => <Self as NumCast>::from(u),
+            PValue::U16(u) => <Self as NumCast>::from(u),
+            PValue::U32(u) => <Self as NumCast>::from(u),
+            PValue::U64(u) => <Self as NumCast>::from(u),
```
lwwmanning, line 316: @robert3005 the crux of the problem was the internal inconsistency here and in the functions below, because this makes PValues not round-trip with Rust numeric types... which makes it super easy to accidentally lose type info when manipulating them

vortex-proto/proto/scalar.proto
lwwmanning: ya, it's all the same, and I wanted to minimize the number of different cases (since we currently serialize f16 as u64)

vortex-scalar/src/pvalue.rs
robert3005: yes, this function is incorrect if you want to use it to cast values which when we implemented cast should have been tested. However, this is the function to get an arbitrary scalar value and coerce it to the scalars target dtype. We should come up with a separate function to create a correct pvalue. We cannot keep adding variants to the serialized format for any new case, i.e. if I add TF32 does this need another variant?

vortex-array/src/arrays/constant/canonical.rs
```diff
@@ -287,14 +287,14 @@ mod tests {
 
     #[test]
     fn test_canonicalize_scalar_values() {
-        let f16_scalar = Scalar::primitive(f16::from_f32(5.722046e-6), Nullability::NonNullable);
-        let scalar = Scalar::new(
-            DType::Primitive(PType::F16, Nullability::NonNullable),
-            Scalar::primitive(96u8, Nullability::NonNullable).into_value(),
```
lwwmanning, line 293: this old behavior is so gross when you look at it this way. it would reinterpret cast a 96u8 into f16 on construction (rather than doing numeric casting), if and only if the target type was f16. if it was an integer, f32, or f64, it would do numeric casting.

robert3005:
Scalar should just be a tag for value and dtype, it's imho weird that we are doing work when creating a new one. I thought we can do the work when we are constructing the primitive scalar instead, this way people don't have to pay the cost of backcompat. If we are doing the work in primitive scalar then we don't need additional tag on the value since we know at that point what type the value is meant to be (from the dtype) and we can convert it to the value matching that representation.

danking (approved):

robert3005:
https://github.com/vortex-data/vortex/pull/4135

PR #4132, opened 2025-08-05
WIP fsst+

a10y:
Was glancing through FSST+, the meat is the cleaving algorithm in https://github.com/cwida/fsst_plus/blob/main/src/cleaving/cleaving.h#L75-L159 which dictates the optimal point to split the sorted blocks by maximum prefix.

Far from complete. Also the test case is just for development ease, in actuality it seems that the cleaving happens _after_ FSST encoding, not before. So you split the codes by prefix rather than the source strings. This probably makes sense given the LCP is faster to compute over compressed strings than full strings.

An alternative to this is the very simplistic splitting in FSSTView from LiquidCache: https://github.com/XiangpengHao/liquid-cache/blob/main/dev/design/00-fsst-view.md

### DIFF
--- a/encodings/fsst/src/prefix.rs
+++ b/encodings/fsst/src/prefix.rs
@@ -1,0 +1,171 @@
+//! An implementation of the FSST+ extension, original at https://github.com/cwida/fsst_plus/.
+//!
+//! FSST+ augments FSST with the addition.
+
+#![allow(unused)]
+
+#[derive(Debug)]
+pub struct SimilarityBlock {
+    start_idx: usize,
+    prefix_len: usize,
+}
+
+/// Maximum shared prefix length.
+pub const MAX_PREFIX: usize = 128;
+
+/// Find the longest-common-prefix between adjacent encoded texts.
+pub fn longest_common_prefix(codes: &[&[u8]]) -> Vec<usize> {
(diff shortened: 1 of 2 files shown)


robert3005:
Does fsst guarantee that no code is a prefix of another? I was always unsure about that part and lcp would be hard if it wasn’t the case

robert3005:
Should we maybe put this in the fsst repo?

a10y:
The algorithm isn't really FSST-dependent AFAICT, it treats the input as opaque byte strings. The handling code seems to do this step after FSST-encoding all of the values.

So I actually think this makes more sense in Vortex than in the core fsst crate

encodings/fsst/src/prefix.rs
```diff
@@ -0,0 +1,171 @@
...
+#![allow(unused)]
+
+#[derive(Debug)]
+pub struct SimilarityBlock {
+    start_idx: usize,
+    prefix_len: usize,
+}
+
+/// Maximum shared prefix length.
+pub const MAX_PREFIX: usize = 128;
```
yan-alex, line 14: The maximum prefix size should be 255 here :)
128 is the batch size, i.e. how many rows at a time the chunk_by_similarity() function would process.

But nice start, you guys are fast

PR #4135, opened 2025-08-06
Only coerce scalar values when necessary

robert3005:

### DIFF
--- a/vortex-scalar/src/primitive.rs
+++ b/vortex-scalar/src/primitive.rs
@@ -62,12 +62,15 @@ impl<'a> PrimitiveScalar<'a> {
     /// Returns an error if the data type is not a primitive type or if the value
     /// cannot be converted to the expected primitive type.
     pub fn try_new(dtype: &'a DType, value: &ScalarValue) -> VortexResult<Self> {
-        let ptype = dtype.as_ptype();
+        let ptype = PType::try_from(dtype)?;
 
         // Read the serialized value into the correct PValue.
         // The serialized form may come back over the wire as e.g. any integer type.
         let pvalue = match_each_native_ptype!(ptype, |T| {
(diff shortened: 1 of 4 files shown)


vortex-scalar/src/primitive.rs
```diff
@@ -62,16 +62,12 @@ impl<'a> PrimitiveScalar<'a> {
     /// Returns an error if the data type is not a primitive type or if the value
     /// cannot be converted to the expected primitive type.
     pub fn try_new(dtype: &'a DType, value: &ScalarValue) -> VortexResult<Self> {
-        let ptype = PType::try_from(dtype)?;
+        let ptype = dtype.as_ptype();
```
lwwmanning: this should return error, not panic

lwwmanning (approved):

lwwmanning:
I really want the `impl TryFrom<&pb::Scalar> for Scalar` in proto.rs to coerce as well, which happens currently on develop but not here.

lwwmanning:
I think there are also some dumb subtleties here where two scalars with the same Primitive DType could be unequal (because the value of one hasn't been coerced), even though converting them both to PrimitiveScalar would yield two equal PrimitiveScalars

lwwmanning (changes requested):
see other comments

robert3005:
You don’t have enough information at that point, we coerce where we use the scalar not when we construct them but when we use them

robert3005:
The equal on a scalar uses primitive scalar coercion. You cannot compare two scalar values

vortex-scalar/src/primitive.rs
robert3005: I got a little bit trigger happy with panics here

lwwmanning (approved):
I can live with this, though I don't like it

PR #4137, opened 2025-08-06
VTable metadata is owned

robert3005:


robert3005:
Not sure this is better, it appeals to my less clone aesthetic but unclear if there's any benefit to it

PR #4141, opened 2025-08-06
fix: pr benchmark results

0ax1:
To disambiguate different TPC-H runs, include the scale factor in the dataset name. In addition, we filter duplicate entries which can result from nightly runs if the commit is identical to the base commit of the PR to avoid duplicate result rows.

### DIFF
--- a/bench-vortex/src/datasets/mod.rs
+++ b/bench-vortex/src/datasets/mod.rs
@@ -39,12 +39,12 @@ pub enum BenchmarkDataset {
 }
 
 impl BenchmarkDataset {
-    pub fn name(&self) -> &str {
+    pub fn name(&self) -> String {
         match self {
-            BenchmarkDataset::TpcH { .. } => "tpch",
-            BenchmarkDataset::TpcDS { .. } => "tpcds",
-            BenchmarkDataset::ClickBench { .. } => "clickbench",
-            BenchmarkDataset::PublicBi { .. } => "public-bi",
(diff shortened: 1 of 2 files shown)


robert3005:
This either requires manually fixing the json or merging in the frontend

0ax1:
> This either requires manually fixing the json or merging in the frontend

In my understanding we match on
```
#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
pub enum BenchmarkDataset {
    #[serde(rename = "tpch")]
    TpcH { scale_factor: String },
    #[serde(rename = "tpcds")]
    TpcDS { scale_factor: String },
    #[serde(rename = "clickbench")]
    ClickBench { single_file: bool, flavor: Flavor },
    #[serde(rename = "public-bi")]
    PublicBi { name: String },
}
```
on the bench website, not the name.

robert3005:
So we added the scale factor the benchmark website but since we kept the name and don't use it on the website this is only ever wrong if you trying to do a pr diff because that uses the name

0ax1:
> So we added the scale factor the benchmark website but since we kept the name and don't use it on the website this is only ever wrong if you trying to do a pr diff because that uses the name

Yeah the PR needs a base commit which has that bench name such that it can be compared.

robert3005:
Eh, let's get this merged and then we can redo how we serialize json. There's a lot of random duplicate information here

robert3005 (approved):

PR #4144, opened 2025-08-06
[WIP] Datafusion-related crimes

AdamGS:
this code is mostly horrible right now, but its also pretty close. need to cleanup some of the expression translation which can't really happen the same way at this scope unfortunately, but definitely seems possible.


AdamGS:
note to self - https://github.com/vortex-data/vortex/actions/runs/16785004147

AdamGS:
note to self - problem is that because of splitting/annotations, the inner expr references the original names while the rest of the system expects something like:
```
[vortex-datafusion/src/convert/dynamic.rs:125:9] &children = [
    GetItemExpr {
        field: "EventTime_0",
        child: GetItemExpr {
            field: "EventTime",
            child: RootExpr,
        },
    },
    GetItemExpr {
        field: "SearchPhrase_0",
        child: GetItemExpr {
            field: "SearchPhrase",
            child: RootExpr,
        },
    },
]
```

AdamGS:
This is probably salvageable, but the opaqueness of the underlying datafusion expression makes it hard, and the amount of conflicts is too much to bother.

PR #4149, opened 2025-08-07
refactor the LayoutStrategy

a10y:
Breaking this out as it was some refactoring I was already doing as part of #3810

We are already passing around BoxFuture with our `SendableLayoutFuture`, using the async_trait stuff just makes it more ergonomic.

The other thing I did is that I've updated our existing `LayoutStrategy` impls to be parameterized by the children. This actually is completely fine for us and doesn't harm extensibility. I also find it a bit clearer too, and I think there is also a benefit of fewer unnecessary layers of indirection.

I added more doc comments, and also put `VortexLayoutStrategy` behind a builder, which unifies the compressor configuration.


bench-vortex/src/lib.rs
```diff
@@ -191,13 +189,11 @@ pub enum CompactionStrategy {
...
-            CompactionStrategy::Compact => {
-                let executor = Arc::new(LocalExecutor);
-                let compressor = CompactCompressor::default();
-                let compact_strategy =
-                    VortexLayoutStrategy::compact_with_executor(executor, compressor);
-                options.with_strategy(compact_strategy)
-            }
+            CompactionStrategy::Compact => options.with_strategy(
+                WriteStrategyBuilder::new()
+                    .with_compressor(CompactCompressor::default())
```
a10y, line 194: LocalExecutor is the default if you don't override it

onursatici: there is https://github.com/vortex-data/vortex/pull/4118 but I am thinking of not merging that for now until I better understand the hang we sometimes see in `vx convert`

onursatici (approved):

PR #4152, opened 2025-08-07
ScalarValue operations must go via Scalar

robert3005:
While ScalarValues are nice for storage since the dtype is somewhere
else any computation and transformation has to operate in Scalar space


vortex-array/src/array/mod.rs
```diff
@@ -402,8 +402,9 @@ impl<V: VTable> Array for ArrayAdapter<V> {
                         stat,
                         Stat::IsConstant | Stat::IsSorted | Stat::IsStrictSorted
                     ) && value.as_ref().as_exact().is_some_and(|v| {
-                        v.as_bool()
-                            .vortex_expect("must be a bool")
+                        Scalar::new(DType::Bool(Nullability::NonNullable), v.clone())
```
lwwmanning, line 405: `Scalar::bool`

a10y:
This is a fairly large diff, just briefly what actually changed here? Anything in particular you want looked at?

vortex-array/src/stats/stats_set.rs
```diff
@@ -107,6 +95,44 @@ impl StatsSet {
...
+    /// Get scalar value of a given dtype
+    pub fn get_as<T: for<'a> TryFrom<&'a Scalar, Error = VortexError>>(
+        &self,
+        stat: Stat,
+        dtype: &DType,
+    ) -> Option<Precision<T>> {
+        self.get(stat).map(|v| {
+            v.map(|v| {
+                T::try_from(&Scalar::new(dtype.clone(), v)).unwrap_or_else(|err| {
+                    vortex_panic!(
```
lwwmanning, line 126: if the Scalar is somehow null or empty for the stat (as opposed to missing), we'll panic?

I think this is fine, because we don't allow Null scalars for Min, Max, etc... we just store NullCount instead. But just thinking out loud, should make sure we document this well / have good tests for it. (Separate from this PR)

robert3005:
Should have provided more of a description - I was talking to @lwwmanning how ScalarValues are a footgun. I thought the footguns are well contained and handled. However, letting people convert ScalarValues to primitives is not correct ever. This pr removes any ScalarValue downcasting and replaces logic in stats to always wrap things in a Scalar. The changes in stats are the most interesting ones

lwwmanning (dismissed):
😍

vortex-array/src/array/mod.rs
robert3005: This is one of the bad parts of this pr - the iterator yields you ScalarValues and I removed any way to unwrap a ScalarValue so you have to wrap it in a Scalar to unwrap

vortex-array/src/stats/stats_set.rs
robert3005: If the stat is missing then the top level get will return None and the map doesn't run but we should add some tests to this

lwwmanning (commented):
love this conceptually, but we have some benchmark regressions that we should look at before merging

robert3005:
@lwwmanning I don't think there's a true performance degradation here. Maybe something small but rerun doesn't show anything lasting. The only thing that differs now is we clone some dtypes but nothing shows up in profiles

robert3005:
Rerunning clickbench shows no difference

lwwmanning (approved):
Ship it

PR #4155, opened 2025-08-07
chore[bench-website]: add back tpc-ds to query_bench

joseph-isaacs:
tpcds is omitted from CI since we currently fail on some queries


bench-vortex/src/tpch/duckdb.rs
```diff
@@ -148,3 +149,74 @@ pub fn generate_tpc(opts: DuckdbTpcOptions) -> Result<PathBuf> {
 
     Ok(output_dir)
 }
+
+/// Generate TPC-DS data using a DuckDB connection
+pub fn generate_tpcds_with_connection(
```
0ax1: > with_connection

Strip from the name.

0ax1 (approved):
alright, but fix up the name for `generate_tpcds_with_connection`

PR #4156, opened 2025-08-07
vortex-array: support converting arrow DictionaryArray

asubiotto:
This is a pretty simple change with a simple test attached. The thing I'm not so sure about and requires careful review is that I had to move the majority of `vortex-dict` in order to be able to construct a `DictArray` in `vortex-array` (there was a circular dependency otherwise) while maintaining the public API. I'm happy to do whatever the maintainers recommend here, I just went ahead and moved the code since it's what made most sense to me.

Also happy to add more tests, just let me know where.


robert3005:
I have made a minimal change that doesn't plug into conversion logic in #4157. We don't want to move the dict array to vortex-array. The long term solution here that requires moving quite a few files around is to create separate vortex-arrow package

asubiotto:
Wonderful, thank you! Happy to take a look at that when I have time. What do you envision would/wouldn't be included in `vortex-arrow`?

robert3005:
vortex-arrow should have all the logic of converting Arrow->Vortex and Vortex->Arrow. Right now because of orphan rules this is split all over and has issues like the one you run into here

PR #4157, opened 2025-08-07
feature: Support converting Arrow DictArray into vortex

robert3005:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6501,6 +6501,7 @@ dependencies = [
 name = "vortex-dict"
 version = "0.1.0"
 dependencies = [
+ "arrow-array",
  "arrow-buffer",
  "codspeed-divan-compat",
  "num-traits",
--- a/encodings/dict/Cargo.toml
+++ b/encodings/dict/Cargo.toml
@@ -19,8 +19,8 @@ test-harness = ["rand", "vortex-fsst"]
 arrow = ["dep:arrow-array"]
 
 [dependencies]
-arrow-buffer = { workspace = true }
 arrow-array = { workspace = true, optional = true }
+arrow-buffer = { workspace = true }
 num-traits = { workspace = true }
 prost = { workspace = true }
 # test-harness
--- a/encodings/dict/Cargo.toml
+++ b/encodings/dict/Cargo.toml
@@ -16,9 +16,11 @@ version = { workspace = true }
 
 [features]
 test-harness = ["rand", "vortex-fsst"]
+arrow = ["dep:arrow-array"]
 
 [dependencies]
 arrow-buffer = { workspace = true }
+arrow-array = { workspace = true, optional = true }
 num-traits = { workspace = true }
 prost = { workspace = true }
 # test-harness
--- a/encodings/dict/src/arrow.rs
+++ b/encodings/dict/src/arrow.rs
@@ -1,0 +1,16 @@
+use arrow_array::types::ArrowDictionaryKeyType;
+use arrow_array::{AnyDictionaryArray, DictionaryArray};
+use vortex_array::ArrayRef;
+use vortex_array::arrow::FromArrowArray;
+use vortex_error::VortexUnwrap;
+
+use crate::DictArray;
+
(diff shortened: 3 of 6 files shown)


AdamGS (approved):
🥳

asubiotto:
Tried out this test today which I expected to succeed after this PR (running on ac773cb0376c6ad33fcaa42c1b8515013cb98c20)
```
    fn test_dictionary_array_conversion() {
        let schema = Schema::new(vec![Field::new_dictionary(
            "string_dict",
            DataType::UInt8,
            DataType::Utf8,
            true,
        )]);
        let keys = UInt8Array::from(vec![0u8, 1u8, 0u8]);
        let values = StringArray::from(vec!["a", "b"]);
        let dict_array = DictionaryArray::try_new(keys, Arc::new(values)).unwrap();
        let batch = RecordBatch::try_new(Arc::new(schema), vec![Arc::new(dict_array)]).unwrap();

        ArrayRef::from_arrow(batch, false);
    }
 ```
 This panics with:
 ```
 Array encoding not implemented for Arrow data type Dictionary(UInt8, Utf8)
 ```
 Because the record batch shells out to `ArrayRef::from_arrow` for the inner dictionary array (not `DictArray::from_arrow`). Is there some other way I should be converting arrow record batches?

**Comment by robert3005** [comment]
Until we move the logic around you will have to inline some of the RecordBatch translation logic, i.e. RecordBatch is just
```
impl FromArrowArray<RecordBatch> for ArrayRef {
    fn from_arrow(array: RecordBatch, nullable: bool) -> Self {
        ArrayRef::from_arrow(&arrow_array::StructArray::from(array), nullable)
    }
}
```
And StructArray
```
impl FromArrowArray<&ArrowStructArray> for ArrayRef {
    fn from_arrow(value: &ArrowStructArray, nullable: bool) -> Self {
        StructArray::try_new(
            value.column_names().iter().copied().collect(),
            value
                .columns()
                .iter()
                .zip(value.fields())
                .map(|(c, field)| {
                    // Arrow pushes down nulls, even into non-nullable fields. So we strip them
                    // out here because Vortex is a little more strict.
                    if c.null_count() > 0 && !field.is_nullable() {
                        let stripped = make_array(remove_nulls(c.into_data()));
                       // match on field.data_type() here and call Dictionary conversion logic directly
                       // make sure to call back into your own conversion logic for structs and lists
                        Self::from_arrow(stripped.as_ref(), false)
                    } else {
                       // match on field.data_type() here and call Dictionary conversion logic directly
                       // make sure to call back into your own conversion logic for structs and lists 
                       Self::from_arrow(c.as_ref(), field.is_nullable())
                    }
                })
                .collect(),
            value.len(),
            nulls(value.nulls(), nullable),
        )
        .vortex_expect("Failed to convert Arrow StructArray to Vortex StructArray")
        .into_array()
    }
}
```
you will also need this remove_nulls function
```
fn remove_nulls(data: arrow_data::ArrayData) -> arrow_data::ArrayData {
    if data.null_count() == 0 {
        // No nulls to remove, return the array as is
        return data;
    }

    let children = match data.data_type() {
        DataType::Struct(fields) => Some(
            fields
                .iter()
                .zip(data.child_data().iter())
                .map(|(field, child_data)| {
                    if field.is_nullable() {
                        child_data.clone()
                    } else {
                        remove_nulls(child_data.clone())
                    }
                })
                .collect_vec(),
        ),
        DataType::List(f)
        | DataType::LargeList(f)
        | DataType::ListView(f)
        | DataType::LargeListView(f)
        | DataType::FixedSizeList(f, _)
            if !f.is_nullable() =>
        {
            // All list types only have one child
            assert_eq!(
                data.child_data().len(),
                1,
                "List types should have one child"
            );
            Some(vec![remove_nulls(data.child_data()[0].clone())])
        }
        _ => None,
    };

    let mut builder = data.into_builder().nulls(None);
    if let Some(children) = children {
        builder = builder.child_data(children);
    }
    builder
        .build()
        .vortex_expect("reconstructing array without nulls")
}
```

PR #4159, opened 2025-08-07
chore: Bump msrv to 1.89

AdamGS:
Mostly motivated by wanting to remove the horrible `ResutExt::unnest` I introduced a few months ago, but also let chains are cool.


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.88.0"
+channel = "stable"
```
AdamGS: I think this should always be stable, the `rust-version` field is the thing that actually pins the code and behavior.

bench-vortex/src/engines/df/mod.rs
```diff
@@ -138,6 +138,12 @@ pub async fn execute_query(
     let plan = ctx.sql(query).await?;
     let (state, plan) = plan.into_parts();
     let physical_plan = state.create_physical_plan(&plan).await?;
+    println!(
+        "{}",
+        DisplayableExecutionPlan::with_metrics(physical_plan.as_ref())
+            .set_show_schema(true)
+            .tree_render()
+    );
```
robert3005: Remove this

robert3005 (approved):

rust-toolchain.toml
a10y: This might cause some friction when new lints get introduced

a10y (approved):

rust-toolchain.toml
AdamGS: yeah I reverted it (to 1.89), rob notes that he thinks some caching is dependent on this value, so might as well have something stable.

PR #4161, opened 2025-08-07
feat(python): Write PyArrow types directly to Vortex

### LINKED ISSUE #4160: How do I append to a vortex file from python using io.write?
Just head about vortex today! My goal is to benchmark vortex for 10B+ records as an alternative to parquet and lance for ML applications. Looking to sequentially or concurrently write to a vortex file from an ETL pipeline. Currently seems like I need to have the entire dataset as an vortex array in memory to write to disk?

a10y:
Fixes: #4160

Extend PyVortex write API to accept PyArrow Tables and RecordBatchReaders
directly, enabling streaming writes without loading entire datasets into memory.

Key changes:
- Support PyArrow Table and RecordBatchReader objects via ArrowArrayStreamReader
- Stream Arrow RecordBatches directly to Vortex ArrayIterator


gatesn (approved):

kylebarron:
👋 Just saw this and have a small suggestion: this current implementation imports `pyarrow` explicitly and thus will only work if `pyarrow` itself is installed in the user's environment. But since Arrow is a stable format, you really don't need `pyarrow`; this is why the [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) was created. Any Python object that exposes an `__arrow_c_stream__` method can be called to get a PyCapsule with an Arrow C Stream pointer. 

There are a [bunch of libraries](https://github.com/apache/arrow/issues/39195#issuecomment-2245718008) that already support this, so if you implemented support for the PyCapsule interface then you could take in data from Polars/DuckDB/DataFusion etc even if pyarrow isn't installed in the environment.

You can use [`pyo3-arrow`](https://docs.rs/pyo3-arrow) (disclaimer: my library) which has [some benefits](https://docs.rs/pyo3-arrow/latest/pyo3_arrow/#why-not-use-arrow-rss-python-integration) over the arrow-rs Python integration. Or, I think the arrow-rs integration [should use it automatically](https://github.com/apache/arrow-rs/pull/5070), and you shouldn't need to import pyarrow explicitly.

(I could make an issue to discuss this if you prefer)

PR #4162, opened 2025-08-07
Support exporting Arrow RunArray to Vortex

robert3005:


encodings/runend/src/ops.rs
```diff
@@ -13,15 +13,8 @@ impl OperationsVTable<RunEndVTable> for RunEndVTable {
     fn slice(array: &RunEndArray, start: usize, stop: usize) -> VortexResult<ArrayRef> {
         let new_length = stop - start;
 
-        let (slice_begin, slice_end) = if new_length == 0 {
-            let values_len = array.values().len();
-            (values_len, values_len)
```
robert3005, line 18: I think this fixed a bug at some point but we have better binary search handling now that we no longer need this

lwwmanning (approved):

PR #4168, opened 2025-08-08
fix: duckdb dyn expr race on filter constant

0ax1:


vortex-duckdb/src/convert/table_filter.rs
```diff
@@ -89,16 +89,14 @@ pub fn try_from_table_filter(
                 ),
             };
             let data = dynamic.data;
+            let scalar_value = data
```
gatesn: you cant get the value outside the closure!

0ax1: `Value` isn't `Send`

gatesn: I think we can unsafe impl it for Value? We basically have to check the C++ code and make sure it's true, but I can't imagine a world in which it's not.

The whole point is we get the latest value each time we invoke the closure, you've now made the expression _not_ dynamic 😅

0ax1: _touché_

0ax1: I guess we need to move `dynamic` into the closure.
```
                move || {
                    let data = dynamic.data;
                    let scalar_value = data
                        .latest()
                        .and_then(|v| Scalar::try_from(&v).ok().map(|s| s.into_value()));
                    scalar_value.clone()
                },
```
to keep the memory alive.

gatesn: I think `data` is an owned shared_ptr no?

0ax1: If the assignment would transfer ownership we wouldn't have issues based on the code we currently have on `develop` in my understanding.

0ax1: (That said I also had 3 successful runs returning a heap allocated & owned value from `duckdb_vx_dynamic_filter_data_get_value`. That _should_ not really change here anything though.)

robert3005:
I think in this world we don't get any dynamic expression pushdown since the expression value is just a constant.

robert3005 (changes requested):
Screenshot 2025-08-08 at 23 42 32
Just making sure we don't merge this regression

0ax1:
> Screenshot 2025-08-08 at 23 42 32 Just making sure we don't merge this regression

Yeah we all agree that `latest` should be called in the closure. It's not clear yet what's the best way to do that though: https://github.com/vortex-data/vortex/pull/4168#discussion_r2263407762

0ax1:
Guess we want to own here then:
```
   let ptr = unsafe { cpp::duckdb_vx_dynamic_filter_data_get_value(self.as_ptr()) };
        if ptr.is_null() {
            return None;
        }
        Some(unsafe { Value::borrow(ptr) })
```

vortex-duckdb/cpp/table_filter.cpp
```diff
@@ -107,10 +107,16 @@ extern "C" duckdb_value duckdb_vx_dynamic_filter_data_get_value(duckdb_vx_dynami
         return nullptr;
     }
     auto data_wrapper = reinterpret_cast<DynamicFilterDataWrapper *>(ffi_data);
+
```
gatesn, line 113: Do you also need the lock for line 94?

0ax1: yes

robert3005 (approved):

PR #4169, opened 2025-08-08
feat: teach Vortex how to export list arrays to duckdb

danking:


vortex-duckdb/src/duckdb/vector.rs
```diff
@@ -187,6 +198,26 @@ impl Vector {
...
+
+    pub fn list_vector_get_child(&self) -> Self {
+        unsafe { Self::borrow(cpp::duckdb_list_vector_get_child(self.as_ptr())) }
+    }
+
+    pub fn list_vector_set_size(&self, size: u64) -> VortexResult<()> {
+        let state = unsafe { cpp::duckdb_list_vector_set_size(self.as_ptr(), size) };
+        match state {
+            cpp::duckdb_state::DuckDBSuccess => Ok(()),
+            cpp::duckdb_state::DuckDBError => vortex_bail!("vector was nullptr!"),
```
joseph-isaacs, line 218: is that the only reason?

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -186,27 +184,34 @@ fn new_array_exporter(
...
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `len` is less than or equal to the capacity of this vector.
+    unsafe fn set_validity(&mut self, mask: &Mask, offset: usize, len: usize) -> bool;
 }
 
 impl VectorExt for Vector {
-    fn set_validity(&mut self, mask: &Mask, offset: usize, len: usize) -> bool {
+    unsafe fn set_validity(&mut self, mask: &Mask, offset: usize, len: usize) -> bool {
```
joseph-isaacs, line 195: where does the unsafe stop?

vortex-duckdb/src/exporter/list.rs
```diff
@@ -0,0 +1,183 @@
...
+                .ok_or_else(|| vortex_err!("list offsets must fit in u64"))?;
+            let len = end - start;
+            sum_of_list_lens += len;
+            *destination = cpp::duckdb_list_entry {
+                offset: start - start_offset,
+                length: len,
+            };
+        }
+
+        vector.list_vector_reserve(sum_of_list_lens)?;
```
joseph-isaacs, line 78: This is not a blocker, but the multiple calls to reserve (one for each export) seems unnecessarily you could allocate once a larger vector and export the values once

vortex-duckdb/src/exporter/list.rs
```diff
@@ -0,0 +1,183 @@
...
+            offset_type: PhantomData::<T>,
+        }) as Box<dyn ColumnExporter>
+    });
+    Ok(boxed)
+}
+
+impl<T: NativePType> ColumnExporter for ListExporter<T> {
+    fn export(&self, offset: usize, len: usize, vector: &mut Vector) -> VortexResult<()> {
+        // Set validity if necessary.
+        if unsafe { vector.set_validity(&self.validity, offset, len) } {
```
joseph-isaacs, line 45: This is not you but the name of this functions is confusing vs what is does. It also check if all values are null.

vortex-duckdb/src/exporter/list.rs
joseph-isaacs, line 1: We really want a list view

vortex-duckdb/src/exporter/list.rs
```diff
@@ -0,0 +1,183 @@
...
+            // All values are null, so no point copying the data.
+            return Ok(());
+        }
+
+        let offsets = &self.offsets.as_slice::<T>()[offset..][..(len + 1)];
+        let start_offset = offsets[0]
```
joseph-isaacs, line 52: why is this non empty

joseph-isaacs (approved):
LG

vortex-duckdb/src/duckdb/vector.rs
danking: According to the docs, yes: https://duckdb.org/docs/stable/clients/c/vector.html#duckdb_list_vector_set_size

vortex-duckdb/src/exporter/mod.rs
danking: Yeah I don't know what to do here. It seems like everything about this interface is unsafe because we don't have the length. Maybe we should wrap it up in a thing that has the length and then only extract an actual duckdb vector at the end?

vortex-duckdb/src/exporter/list.rs
danking: Are you suggesting that I export the entire elements child array even though this export call might only be from some subset of the lists? Am I allowed to assume the child vector holding the elements doesn't change across multiple calls to `export`?

vortex-duckdb/src/exporter/list.rs
danking: agreed.

vortex-duckdb/src/exporter/list.rs
joseph-isaacs: I think yes for the second part, 
I am not sure about this first part I think this is good for now. Maybe add a comment about this?

vortex-duckdb/src/exporter/list.rs
joseph-isaacs: Want to rename it?

vortex-duckdb/src/exporter/mod.rs
joseph-isaacs: I don't know either

vortex-duckdb/src/exporter/list.rs
danking: I added a TODO

vortex-duckdb/src/exporter/list.rs
danking: Hmm. I spent some time thinking and I can't think of any names that aren't unwieldy like `set_validity_get_is_all_null`.

PR #4171, opened 2025-08-08
fix(pyvortex): encoding-specific dispatch

a10y:
Before this downcasting was always failing b/c we failed to go throught the ArrayAdapter. Also added some simple unit tests.

**Before PR**

image

**After PR**

image


vortex-python/src/arrays/native.rs
```diff
@@ -223,7 +223,8 @@ impl<V: EncodingSubclass> AsArrayRef<<V::VTable as VTable>::Array> for PyRef<'_,
         self.as_super()
             .inner()
             .as_any()
-            .downcast_ref::<<V::VTable as VTable>::Array>()
+            .downcast_ref::<ArrayAdapter<V::VTable>>()
```
a10y, line 226: this was the fix, the rest is just tests and other cleanup

robert3005 (approved):

vortex-python/src/arrays/builtins/struct_.rs
```diff
@@ -22,4 +23,15 @@ impl PyStructArray {
         let field = self_.as_array_ref().field_by_name(name)?.clone();
         Ok(PyArrayRef::from(field))
     }
+
+    /// Get an ordered list of field names for the struct fields.
```
danking, line 27: I find the adjective ordered more confusing than elucidating here but perhaps that's just me.

danking (approved):

vortex-python/src/arrays/builtins/struct_.rs
a10y: i guess i wanted to clarify that it's a list[] and not a set[], so it's in the schema field order. i'm also happy to change the wording if you have a suggestion.

PR #4173, opened 2025-08-08
build: Use Create cc to link to libc++

ssmccoy:
Rather than manually selecting libstdc++ when the `$TARGET` is "linux", simply set the `cpp(true)` flag when calling `cc::Build` so that it delegates stdlib linking to the compiler which will always yield the correct choice, and also allow for manual override via the `$CXXSTDLIB` environment variable [1].

This fixes the out of the box test suite on Linux

 [1]: https://docs.rs/cc/latest/cc/#c-support

### DIFF
--- a/vortex-duckdb/build.rs
+++ b/vortex-duckdb/build.rs
@@ -216,12 +216,6 @@ fn main() {
         extraced_lib_path.display()
     );
 
-    if env::var("TARGET").unwrap().contains("linux") {
-        println!("cargo:rustc-link-lib=stdc++");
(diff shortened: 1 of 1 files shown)


lwwmanning:
hi @ssmccoy -- thanks for the contribution! you can run `git rebase origin/develop --signoff` and then force push to the branch to make the DCO check pass

0ax1 (approved):

ssmccoy:
It wasn't a chore, I had fun.

PR #4174, opened 2025-08-08
fix[scan]: don't convert filter to `cnf` just extract conjunct list

joseph-isaacs:
Cnf can cause a large blow up in expression size (e.g. `(p₁ AND q₁) OR (p₂ AND q₂) OR (p₃ AND q₃)`) goes to 2^n nodes
This PR replaces cnf extract with `and` unpacking.
We could look later at either trying to cnf transform with a conversion expression size cap?


joseph-isaacs:
Look like a small perf win

danking:
I don’t think any of our benchmarks have particularly large queries that would trigger this kind of behavior.

joseph-isaacs:
tpcds 41 does

danking:
I think we need to add tcpds to sql-benchmarks.yml so we can see the difference.

joseph-isaacs:
it doesn't terminate with q41 after a few mins

AdamGS:
I don't have an opinion either way, but at least locally it only seems to hang on `duckdb:vortex`, datafusion works great with both vortex and datafusion (on q41, it has a bunch of other failures including the bug from Friday).

AdamGS:
seems like Datafusion only tries to push a pretty small filter to Vortex, I'll look into it.

joseph-isaacs:
But that else should we do instead of this?

joseph-isaacs:
We currently cannot handle certain query filters due to this exponential increase in formula size. We are lucky that datafusion doesn't pass this down, however if it did we would have the same problem as duckdb.

We can look at handling the transformation more carefully in the future if we want.

gatesn (approved):
Seems like a reasonable step in the right direction, even if we ultimately want a hybrid approach to rewriting expressions.

PR #4177, opened 2025-08-08
feat: ArrayOperations infallible, eager validation + new_unchecked

a10y:
ArrayOperations currently return VortexResult<>, but they really should just be infallible. A failed array op is generally indicative of programmer or encoding error. There's really nothing interesting we can do to handle an out-of-bounds slice() or scalar_at.

There's a lot that falls out of this, like fixing a bunch of tests, tweaking our scalar value casting to return Option instead of Result, etc.


gatesn:
The only thing to note is where we might perform the validation.

Can someone craft a bad file, send it to a server running vortex, and crash it?

a10y:
I guess this is really two questions

1. How does this change our behavior when reading corrupted files
2. Does this make it easier/harder to write corrupted files

For (1) this will panic instead of returning an error, but most server frameworks already catch_unwind to prevent panics inside of handlers from crashing the whole server, so I'm not too concerned about this. FWIW, Arrow does do this as well.

So then the question is (2): does doing this mean that we're doing less validation than before. I think the answer is no? We currently trust the buffers coming back out of a file, just like our in-memory Array constructors only do constant-time validation of their inputs. So for example, if someone tries to build a VarBinViewArray of type Utf8, we never validate that the bytes actually hold UTF8. This I think is probably bad, and we should probably be eagerly checking preconditions. And then have a set of unchecked constructors to call internally when we're doing operations on trusted data that we know the operation maintains the invariant.

I don't think returning Result is actually doing anything currently to prevent writing corrupted files though, so I see it as orthogonal?

encodings/alp/src/alp/array.rs
```diff
@@ -52,23 +52,19 @@ pub struct ALPEncoding;
 
 impl ALPArray {
     // TODO(ngates): remove try_new and panic on wrong DType?
```
robert3005: remove the todo?

robert3005:
Pushed a commit to fix the last doc failure

a10y:
@gatesn I think you're right and we probably need better validation in our `try_from` constructors that check for invariants in the **values** as well as just in metadata. Then the _only_ thing IMO that should be fallible is the `SerdeVTable::build` call. Then we have a nice property that once we hold an instance of an encoding in-memory, we can assume all of its preconditions have been met. Examples:

- RunEnd validates ends.max() < len
- Bool validates the buffer is exactly sized
- Dict validates codes.max() < values.len()
- Varbin/List validates offsets are monotonic and ends at `len`
- VBView scans views to check all of the buffer/offset ptrs are in-bounds.

Even before this change, there are a ton of places that we check these things lazily and would panic, not return error. I think it'd be worth a modest penalty to check some of these things eagerly. The alternative is that you can potentially read and propagate corrupt data and won't know unless you run a a compute function that would trigger it somewhere. That does bring up the question of how to handle unparseable things/messages that your client doesn't know about but wants to propagate unmodified...

a10y:
@gatesn I'm working over in #4216 to improve validation. I think the ideal guarantee to provide clients is

- Deserializing untrusted bytes on file read should **never** panic, it should always return an error
- When `SerdeVTable::build()` returns an `Ok(V::Array)` back, you can assume that it is well-formed. For example, ListArray is guaranteed to have non-negative offsets that only point to valid elements range


But things get a bit weird when we factor extensions into the mix, because in theory someone can add an encoding that doesn't semantically meet the full contract, but compiles fine and may cause a panic at runtime.

For example with ListArray, I'd like to verify that offsets children are non-negative and monotonic, and don't point past the end of `elements`. Once I validate that the offsets are non-nullable, I want to do something like `offsets.statistics().compute_min().vortex_expect("array has no nulls")`. If the encoding fails to properly implement the min stat, then this panics. On the one hand, this is an encoding programmer error and it's probably correct to panic. On the other hand, it can be a bad UX for clients. I find it hard to design around external encodings since they are so rare at the moment.

gatesn:
If you're loading extensions from native code shared libraries.... I think all bets are off anyway!

The bigger concern I have is I simply don't understand how you can perform validation during deserialization without effectively decompressing everything. How can you demonstrate that offsets are unique and monotonic? Even asking for min(X) defers to stats, which we load from the file and never verify. Asserting monotonicity is essentially iterating over the array.

I think you're holding the validation to be a little too strong. Why does monotonicity matter? We care about it to perform e.g. search_sorted, but if it's not sorted, then the results are garbage. Maybe that's fine? It's not a panic, it's just nonsense.

Basically, what are the terms of this validation contract? Is it just panic prevention?

a10y:
I think the terms of the validation contract I'm proposing are: when I build an XYZArray, either from components or from deserialization, and it succeeds, the thing I am holding is a valid XYZEncoding and satisfies its constraints.

I have two concerns with the status quo

1. I'm concerned that we either have or can very conceivably have memory safety violations in the presence of maliciously crafted files, e.g. assuming a view pointer is valid or assuming list offsets are monotonic
2. It is currently trivially easy to write corrupt data to a file, because we don't check any of the invariants. You can happily write a VarBinArray with some bad offsets to disk and it will accept it, then error when you try to read it. I think ideally the system prevents you from writing corrupted data altogether.

This is hard, because I agree with you, that in the limit you can't really do better than decoding to primitive array, or iterating VBView views buffer. But, and FWIW, this is exactly what our ToArrowCanonical kernel already does. We use the arrow-rs safe constructors, which internally do validation on all offsets buffers, BinaryView pointers, utf-8 codepoints, etc. But we don't do any of this when scanning into Vortex.

I want to run full SQL benchmarks but what I'm actually wondering is if we interleave the IO and CPU operations properly, can we do this without destroying scan performance? It'd impose a cost on time-to-first-batch but if you can amortize it with the IO then maybe it actually doesn't impose a big penalty?

robert3005 (approved):

PR #4180, opened 2025-08-09
feat: Support more datafusion features

AdamGS:
After spending some time today debugging a Datafusion issue with @joseph-isaacs and @robert3005 (ended up not being related to us), I figured I'll use the momentum to add some of the missing features:
1. Support for basic arithmetic expression pushdown (which exists in TPC-DS), including subtraction to `vortex-expr`.
2. Support files with different (but compatible) schemas in the same table. This is currently only supported if a `PhysicalExprAdapterFactory` is provided (either through the `FileScanConfig` or directly).
3. Support hive-style partitioning columns.

Both features aren't fully fleshed out and probably require more tests, I want to figure out if we can use datafusion's own test suite to close that gap, they have pretty cool tooling to test SQL.
One of the changes it required is making the source and opener more datafusion-y, doing most of the conversion work into Vortex in the actual `FileOpenFuture`. It allows us to utilize more of datafusion's tools for rewriting expressions over specific files.


vortex-datafusion/src/convert/exprs.rs
```diff
@@ -1,18 +1,45 @@
...
+];
+
+// If we cannot convert an expr to a vortex expr, we run no filter, since datafusion
+// will rerun the filter expression anyway.
+pub(crate) fn make_vortex_predicate(predicate: &[&Arc<dyn PhysicalExpr>]) -> Option<ExprRef> {
+    // This splits expressions into conjunctions and converts them to vortex expressions.
+    // Any inconvertible expressions are dropped since true /\ a == a.
+    predicate
+        .iter()
+        .filter_map(|e| ExprRef::try_from_df(e.as_ref()).ok())
```
joseph-isaacs: Can datafusion give us exprs we don't support here? Should we error?

vortex-datafusion/src/convert/exprs.rs
```diff
@@ -53,6 +80,8 @@ impl TryFromDataFusion<DFOperator> for Operator {
             DFOperator::GtEq => Ok(Operator::Gte),
             DFOperator::And => Ok(Operator::And),
             DFOperator::Or => Ok(Operator::Or),
+            DFOperator::Plus => Ok(Operator::Add),
+            DFOperator::Minus => Ok(Operator::Sub),
```
joseph-isaacs: do we know if the overflow semantics are the same?

vortex-expr/src/exprs/operators.rs
```diff
@@ -92,7 +100,8 @@ impl Display for Operator {
             Operator::Lte => "<=",
             Operator::And => "and",
             Operator::Or => "or",
-            Operator::Add => "+w",
+            Operator::Add => "+",
+            Operator::Sub => todo!(),
```
joseph-isaacs: do we want this?

vortex-array/src/arrow/record_batch.rs
```diff
@@ -30,7 +33,9 @@ impl StructArray {
 
     pub fn into_record_batch_with_schema(self, schema: &Schema) -> VortexResult<RecordBatch> {
         let data_type = DataType::Struct(schema.fields.clone());
-        let array_ref = to_arrow(self.as_ref(), &data_type)?;
+        let dtype = DType::from_arrow((&data_type, Nullability::NonNullable));
+        let casted = cast(self.as_ref(), &dtype)?;
```
robert3005: this cast seems fishy

vortex-expr/src/exprs/operators.rs
robert3005:
```suggestion
            Operator::Sub => "-",
```

vortex-array/src/arrow/record_batch.rs
AdamGS: I don't mind pulling it out to the stream, but I think it makes sense given datafusion's behavior at that level.
You might have a u32 column, but a file can store a u16 and you have to expand it somewhere.

vortex-datafusion/src/convert/exprs.rs
AdamGS: good point, I assumed they will just error but seems like the default is actually to do a wrapping_add/sub. I'll remove it here and hopefully we'll be able to do it again in a later version.

vortex-array/src/arrow/record_batch.rs
robert3005: While we should support it I don't think this should live at the dafatusion level. We can push the desired schema into the reader and then during decompression we can adjust the target. I understand it's nice to fill out the matrix of features but this one is actually useful and we can provide a good implementation that is not a data copy.

AdamGS: I'm all for doing it well, I'll look into it. I'm in no rush to merge any of this.

robert3005: I think this might be easier to do after #4188 (and follow up) lands

vortex-datafusion/src/convert/exprs.rs
AdamGS: We now error, originally it was fine but the way I understand the current version of predicate pushdown in Datafusion, if we say we support something we must push it down.

vortex-array/src/arrow/record_batch.rs
AdamGS: As we discussed, removed it from the core vortex codebase and delegated to datafusion's `SchemaAdapter`.

vortex-datafusion/src/convert/exprs.rs
```diff
@@ -1,18 +1,44 @@
...
 
+use std::sync::Arc;
+
+use datafusion::arrow::datatypes::{DataType, Schema};
 use datafusion::logical_expr::Operator as DFOperator;
-use datafusion::physical_expr::{PhysicalExpr, expressions};
+use datafusion::physical_expr::{PhysicalExpr, PhysicalExprRef, expressions};
 use vortex::error::{VortexResult, vortex_bail, vortex_err};
-use vortex::expr::{BinaryExpr, ExprRef, IntoExpr, LikeExpr, Operator, get_item, lit, root};
+use vortex::expr::{ExprRef, Operator, and};
```
robert3005: This is partially overlapping with imports on line 40

AdamGS: cleaned up

vortex-datafusion/src/persistent/source.rs
```diff
@@ -115,32 +156,13 @@ impl FileSource for VortexSource {
...
-        };
-
-        let mut source = self.clone();
-        source.projection = Some(projection_expr);
-        source.arrow_schema = Some(arrow_schema);
-        source.projected_statistics = Some(statistics);
-
-        Arc::new(source)
+    fn with_projection(&self, _config: &FileScanConfig) -> Arc<dyn FileSource> {
+        Arc::new(self.clone())
```
robert3005, line 165: Should this do something?

robert3005 (approved):
Left one question that looks like an omission

vortex-datafusion/src/persistent/source.rs
AdamGS: No, we do everything later and hope someone can figure out how to untangle this API upstream.

vortex-datafusion/src/persistent/source.rs
```diff
@@ -54,21 +60,28 @@ impl VortexSource {
...
+            arrow_file_schema: None,
+            schema_adapter_factory: None,
+            expr_adapter_factory: None,
             _unused_df_metrics: Default::default(),
             layout_readers: Arc::new(DashMap::default()),
         }
     }
 
-    /// Sets a [`VortexExpr`] as a predicate
-    pub fn with_predicate(&self, predicate: Arc<dyn VortexExpr>) -> Self {
```
xudong963, line 68: I saw we removed the method in the PR, how to set the `predicate` for `VortexSource`?

AdamGS: I think the `TableProvider` should use `FileSource::trt_pushdown_filters`, which is a relatively new API (from DF 49?)

PR #4182, opened 2025-08-09
fix: Fix some minor typos and remove the "single-file" feature from clickbench

AdamGS:


bench-vortex/src/engines/ddb/mod.rs
```diff
@@ -157,7 +157,7 @@ impl DuckDBCtx {
             }
             BenchmarkDataset::ClickBench { single_file, .. } => {
                 let file_glob = if *single_file {
-                    format!("{base_dir}hits.{extension}")
+                    format!("{base_dir}hits_0.{extension}")
```
robert3005: We should really kill the single_file version

robert3005 (approved):

bench-vortex/src/engines/ddb/mod.rs
AdamGS: I guess that's a good sign as any that no one uses it, I'll add that to this PR

robert3005 (approved):

PR #4191, opened 2025-08-11
struct layout strategy to use channels

onursatici:
Because we now spawn consuming from upstream, and this is the very first strategy in the default vortex strategy, we have to be careful in sizing the channel. If it is too low then the channel won't fit 2Mb of each column, so buffered strategy won't be able to make progress. If it is too high then we increase our memory footprint unnecessarily. 

Previously we were only buffering only the minimum required chunk amount that would unblock the buffered writer, because everything was poll based, now we have to size the channel ourselves

### DIFF
--- a/vortex-file/src/strategy.rs
+++ b/vortex-file/src/strategy.rs
@@ -122,18 +122,26 @@ impl WriteStrategyBuilder {
             executor.clone(),
         );
 
-        // 1. split columns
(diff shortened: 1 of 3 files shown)


onursatici:
I am leaning towards merging https://github.com/vortex-data/vortex/pull/4193 instead, which doesn't require us to get the channel size right

PR #4207, opened 2025-08-12
perf: Use sum compute function

viirya:
This patch finishes two todo items to replace with `SUM` compute function in `Validity`'s `all_valid` and `all_invalid` functions.

### DIFF
--- a/vortex-array/src/validity.rs
+++ b/vortex-array/src/validity.rs
@@ -91,10 +91,13 @@ impl Validity {
         Ok(match self {
             Validity::NonNullable | Validity::AllValid => true,
             Validity::AllInvalid => false,
-            Validity::Array(array) => {
-                // TODO(ngates): replace with SUM compute function
-                array.to_bool()?.boolean_buffer().count_set_bits() == array.len()
-            }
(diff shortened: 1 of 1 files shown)


vortex-array/src/validity.rs
```diff
@@ -103,8 +108,13 @@ impl Validity {
...
             Validity::Array(array) => {
-                // TODO(ngates): replace with SUM compute function
-                array.to_bool()?.boolean_buffer().count_set_bits() == 0
+                sum(array)
+                    .map(|v| {
+                        v.as_primitive()
+                            .typed_value::<u64>()
+                            .map(|count| count as usize == 0)
+                    })?
+                    .ok_or(vortex_err!("Failed to compute sum for validity array"))?
```
robert3005: `vortex_err!` is not free as we will force capture a backtrace. You want to do `ok_or_else` here

viirya: okay

robert3005 (approved):

viirya:
Thank you @robert3005 @joseph-isaacs

PR #4214, opened 2025-08-12
feat: streaming stores for decoding run-end `PrimitiveArray`

0ax1:


vortex-buffer/src/buffer_mut.rs
```diff
@@ -307,6 +312,81 @@ impl<T> BufferMut<T> {
         self.length += n;
     }
 
+    /// docs
```
AdamGS: nice

0ax1:
Ultimately, whether for streaming or non-streaming simd stores, we're memory bound for run-end decoding. The benefit of simd instructions only really kicks in for types that are 1 byte and no larger. Further, the run length needs be large enough to outweigh the extra complexity of the  `avx::` version compared to `scalar::`. `avx::push_n_unchecked` requires more setup.

The tipping point of where we start to see wins is somewhere between run-length 1024 (no perf change) and  run-length 4096 (2x win). At that point, we start seeing the benefit of using less store instructions. For types larger than 1 byte, we're memory bound, such that simd stores don't bring any benefit.

PR #4215, opened 2025-08-12
Ji/vectors

joseph-isaacs:


encodings/fastlanes/src/for/pipeline.rs
```diff
@@ -0,0 +1,352 @@
...
+
+        Some(Arc::new(ScalarCompareOperator::new(
+            self.children()[0].clone(),
+            compare.op,
+            new_ref,
+        )))
+    }
+}
+
+// TODO(ngates): we should try putting the const bit width as a generic here, to avoid
```
gatesn: ?

encodings/fastlanes/src/for/pipeline.rs
```diff
@@ -0,0 +1,352 @@
...
+        Ok(())
+    }
+
+    fn step(
+        &mut self,
+        ctx: &dyn KernelContext,
+        selected: BitView,
+        out: &mut ViewMut,
+    ) -> Poll<VortexResult<()>> {
+        let vec = ctx.vector(self.child);
```
gatesn, line 143: You say you're in-place, but then we take a child? Does the planner ignore in-place for now?

vortex-array/src/array/mod.rs
```diff
@@ -121,6 +122,9 @@ pub trait Array: 'static + private::Sealed + Send + Sync + Debug + ArrayVisitor
     /// The [`DType`] of the builder must match that of the array.
     fn append_to_builder(&self, builder: &mut dyn ArrayBuilder) -> VortexResult<()>;
 
+    /// Returns a pipeline for the array.
+    fn to_pipeline_plan(&self) -> VortexResult<Arc<dyn Operator>>;
```
gatesn: to_operator

vortex-array/src/pipeline/operators/constant.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::pipeline::{Kernel, KernelContext};
+
+#[derive(Debug)]
+pub struct ConstantOperator {
+    pub(crate) scalar: Scalar,
+}
+
+impl ConstantOperator {
+    pub fn new(scalar: Scalar) -> Self {
+        Self { scalar }
```
gatesn: Need to check for null.

vortex-array/src/pipeline/query/buffers.rs
```diff
@@ -0,0 +1,337 @@
...
+
+use vortex_error::{VortexExpect, VortexResult};
+
+use crate::pipeline::query::Pipeline;
+use crate::pipeline::query::dag::DagNode;
+use crate::pipeline::types::VType;
+use crate::pipeline::vector::{Vector, VectorId};
+
+#[derive(Debug)]
+pub(in crate::pipeline) struct VectorAllocationPlan {
```
gatesn: Can these be pub(super)?

vortex-array/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+#![allow(dead_code)]
```
gatesn: Kill it

vortex-array/src/pipeline/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+pub mod canonical;
+pub mod operators;
+pub mod query;
+pub mod selection;
+pub mod types;
+pub mod vector;
+pub mod view;
+
+/// The number of elements in each step of a Vortex evaluation pipeline.
+pub const PIPELINE_STEP_COUNT: usize = 1024;
```
gatesn: Put it back to N! Everyone knows what it means after 3 seconds, and it's just much cleaner everywhere.

vortex-layout/src/layouts/flat/reader.rs
```diff
@@ -168,8 +174,45 @@ impl MaskEvaluation for FlatEvaluation {
         // Now we await the array .
         let mut array = self.array.clone().await?;
 
+        if array.to_pipeline_plan().is_ok() {
```
gatesn: Can this return `Result<Option<Operator>>` to avoid generating the backtrace?

vortex-array/src/pipeline/mod.rs
joseph-isaacs: This needs to be exported out of pipeline, then is actually not nice

joseph-isaacs: `SC`? I think a global `N` will clash

encodings/fastlanes/src/for/pipeline.rs
joseph-isaacs: Look like we never implemented that

PR #4216, opened 2025-08-12
feat: better and more consistent validation in SerdeVTable::build

a10y:
FLUP to #4177

This removes **all** possible sources of panic inside the deserialization pathway, returning errors. This also adds more explicit and complete validation of all encodings within their `try_new` operation.

Currently, we do "best effort" on these validations, including things like

- Checking Validity is the appropriate length for the array
- Checking min/max values on offsets
- UTF-8 validation on VarBinView and validating that all views reference valid ranges of the buffers

A lot of this was new validation, some of it was moving validation from just SerdeVTable into the array type's `try_new` constructor.

Tried to add sufficient doc comments.

### DIFF
--- a/encodings/fsst/src/canonical.rs
+++ b/encodings/fsst/src/canonical.rs
@@ -16,13 +16,16 @@ use crate::{FSSTArray, FSSTVTable};
 impl CanonicalVTable<FSSTVTable> for FSSTVTable {
     fn canonicalize(array: &FSSTArray) -> VortexResult<Canonical> {
         let (buffer, views) = fsst_decode_views(array, 0);
-        VarBinViewArray::try_new(
-            views,
-            Arc::new([buffer]),
-            array.dtype().clone(),
-            array.codes().validity().clone(),
-        )
-        .map(Canonical::VarBinView)
+        // SAFETY: FSST already validates the bytes for binary/UTF-8. We build views directly on
+        //  top of them, so the view pointers will all be valid.
+        unsafe {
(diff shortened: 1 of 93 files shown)


gatesn:
Can we talk through performance before this merges?

encodings/fsst/src/canonical.rs
```diff
@@ -3,47 +3,45 @@
...
     fn canonicalize(array: &FSSTArray) -> VortexResult<Canonical> {
-        fsst_into_varbin_view(array.decompressor(), array, 0).map(Canonical::VarBinView)
+        let (buffer, views) = fsst_decode_views(array, 0);
+        VarBinViewArray::try_new(
+            views,
+            Arc::new([buffer]),
+            array.dtype().clone(),
+            array.codes().validity().clone(),
+        )
+        .map(Canonical::VarBinView)
```
robert3005: There's something missing in this pr - this shouldn't need to validate if you validated the FSST array. In general when we transform from one array to another we know that we have a valid array

a10y: Yep agree. The validation is really mostly needed for initial deserialization, assuming our implementation is correct we shouldn't re-validate here. I added a few unchecked constructors, will add some more

a10y:
Given that this doesn't **seem** to give any performance penalty (presumbly it's because we can do `new_unchecked` in our implementations since all of the inputs are checked once upfront), I'm going to merge this into my other branch and we can continue the conversation over there.

PR #4224, opened 2025-08-13
feat(c++): support passing filter and projection expression in scan

XinyuZeng:
I found that direct binding doesn't seem too burdensome compared to using protobuf serde. Performance-wise, using binding avoids the serde cost but requires putting the Rust data structures on the heap using `Box::new`, not sure whether this will be a problem if we have complex nested expr.  On the other hand, it will make things easier is if vortex-expr can consume substrait.

Scalar, DType, and Expr support are not complete (e.g., decimal/date/time is not included), but you can create DType using `from_arrow` and cast a Scalar from primitive types to another.


vortex-cxx/Cargo.toml
```diff
@@ -21,14 +21,19 @@ workspace = true
...
 cxx = "1.0"
 futures = { workspace = true, features = ["thread-pool"] }
+paste = { workspace = true }
 prost = { workspace = true }
 take_mut = { workspace = true }
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros"] }
 vortex = { workspace = true, features = ["tokio"] }
+vortex-dtype = { workspace = true }
+vortex-expr = { workspace = true }
+vortex-scalar = { workspace = true }
```
joseph-isaacs: We expose dtype expr and scalar from the vortex crate.

vortex-cxx/src/expr.rs
```diff
@@ -0,0 +1,81 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_dtype::FieldName;
+use vortex_expr::ExprRef;
+
+use crate::scalar::Scalar;
+
+pub(crate) struct Expr {
+    pub(crate) inner: ExprRef,
+}
```
joseph-isaacs, line 11: Do you need this or can you just expose ExprRef to cxx (since its just an arc)?

vortex-cxx/src/expr.rs
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_dtype::FieldName;
+use vortex_expr::ExprRef;
```
joseph-isaacs: can you use the vortex crate

vortex-cxx/cpp/include/vortex/expr.hpp
```diff
@@ -0,0 +1,46 @@
...
+    static Expr is_null(Expr expr);
+    static Expr eq(Expr lhs, Expr rhs);
+    static Expr not_eq_(Expr lhs, Expr rhs);
+    static Expr gt(Expr lhs, Expr rhs);
+    static Expr gt_eq(Expr lhs, Expr rhs);
+    static Expr lt(Expr lhs, Expr rhs);
+    static Expr lt_eq(Expr lhs, Expr rhs);
+    static Expr and_(Expr lhs, Expr rhs);
+    static Expr or_(Expr lhs, Expr rhs);
+    static Expr checked_add(Expr lhs, Expr rhs);
```
joseph-isaacs: what are you thoughts on moving these in the namespace `vortex::expr`? 

Having to write `vortex::Expr::eq` or even `Expr::eq` vs `eq` is very noisy

joseph-isaacs (commented):
Thanks for doing this :)

vortex-cxx/src/expr.rs
XinyuZeng: Arc is not supported by cxx: https://github.com/dtolnay/cxx/issues/1451 Everything passed across the ffi [needs to be defined as a struct first](https://cxx.rs/extern-rust.html#opaque-rust-types) and then Boxed/referenced. :( Maybe it's time to switch to vortex-ffi where we have the arc_wrapper

vortex-cxx/cpp/include/vortex/expr.hpp
```diff
@@ -0,0 +1,46 @@
...
+public:
+    Expr() = delete;
+    Expr(Expr &&other) noexcept = default;
+    Expr &operator=(Expr &&other) noexcept = default;
+    ~Expr() = default;
+
+    Expr(const Expr &) = delete;
+    Expr &operator=(const Expr &) = delete;
+
+    static Expr literal(Scalar scalar);
```
0ax1: I think all static functions introduced as part of this PR should become free standing functions, living in a nested a namespace within vortex. 

We can define the nested namespace in one go:
```
namespace vortex::expr {
    //...
}
```

vortex-cxx/cpp/include/vortex/scalar.hpp
```diff
@@ -0,0 +1,46 @@
...
+    Scalar() = delete;
+    Scalar(Scalar &&other) noexcept = default;
+    Scalar &operator=(Scalar &&other) noexcept = default;
+    ~Scalar() = default;
+
+    Scalar(const Scalar &) = delete;
+    Scalar &operator=(const Scalar &) = delete;
+
+    // Factory functions for creating scalar values
+    static Scalar bool_(bool value);
```
0ax1: namespace to `vortex::scalar`

vortex-cxx/cpp/include/vortex/dtype.hpp
```diff
@@ -0,0 +1,66 @@
...
+    DType() = delete;
+    DType(DType &&other) noexcept = default;
+    DType &operator=(DType &&other) = default;
+    ~DType() = default;
+
+    DType(const DType &) = delete;
+    DType &operator=(const DType &) = delete;
+
+    // Factory functions
+    static DType null();
```
0ax1: namespace to `vortex::dtype` and make the functions free-standing.

vortex-cxx/cpp/include/vortex/expr.hpp
XinyuZeng: Thanks, fixed

vortex-cxx/cpp/include/vortex/expr.hpp
XinyuZeng: Thanks, fixed

joseph-isaacs:
Once we remove `vortex_...` deps we are good to go

vortex-cxx/src/expr.rs
XinyuZeng: Done.

XinyuZeng:
> Once we remove `vortex_...` deps we are good to go

It's removed. Would you mind taking a look again?

joseph-isaacs (approved):
LG, thanks for this

PR #4225, opened 2025-08-13
fix: runend decompress benchmarks

0ax1:


robert3005 (approved):
The previous implementation was definitely very wrong, now the length means the compressed length

PR #4232, opened 2025-08-13
Fix: Return currect field names in `SelectExpr::as_include`

AdamGS:


robert3005 (approved):

vortex-expr/src/exprs/select.rs
```diff
@@ -229,7 +244,7 @@ impl SelectField {
             SelectField::Include(fields) => Ok(fields.clone()),
             SelectField::Exclude(exc_fields) => Ok(field_names
                 .iter()
-                .filter(|f| exc_fields.iter().contains(f))
+                .filter(|f| !exc_fields.iter().contains(f))
```
danking, line 247: lol

danking (approved):

PR #4235, opened 2025-08-13
chore: speed up the benchmark page loading

a10y:
At least locally this makes a big difference for me

image


a10y:
I'm seeing 15s -> 2s speedup in load time with this

danking (approved):

PR #4236, opened 2025-08-13
fix[bench-site]: Compression Size group missing charts

lwwmanning:


lwwmanning:
Before: 
Screenshot 2025-08-13 at 15 07 49

After:
Screenshot 2025-08-13 at 15 11 00

danking (approved):

PR #4237, opened 2025-08-14
chore[vortex-array]: facilitate third-party ToArrowKernel implementations

asubiotto:
This commit has two changes that were required to write a third-party ToArrowKernel to convert vortex to dictionary arrays:

1. Re-exporting vortex-array::arrow::array publicly so that an ArrowArray can be constructed.
2. Support converting an arrow DictionaryType to a canonical DType (this is just the underlying values type).

### DIFF
--- a/vortex-array/src/arrow/mod.rs
+++ b/vortex-array/src/arrow/mod.rs
@@ -14,6 +14,7 @@ mod datum;
 mod iter;
 mod record_batch;
 
+pub use array::*;
 pub use datum::*;
 pub use iter::*;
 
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrow/mod.rs
```diff
@@ -7,7 +7,7 @@ use arrow_array::ArrayRef as ArrowArrayRef;
 use arrow_schema::DataType;
 use vortex_error::VortexResult;
 
-mod array;
+pub mod array;
```
joseph-isaacs: Mind reexporting this instead?

`pub use array::*`

asubiotto: Good call, done.

robert3005 (approved):

PR #4244, opened 2025-08-14
fix: Validity export using shifted u64 copy

robert3005:


vortex-duckdb/src/exporter/mod.rs
```diff
@@ -208,17 +210,21 @@ impl VectorExt for Vector {
...
+                    unsafe {
+                        std::slice::from_raw_parts(
+                            byte_slice.as_ptr() as _,
+                            byte_slice.len().div_ceil(8),
+                        )
                     }
-                }
-                null_count == len
+                };
+                let sliced_bits = &arr_bits.view_bits::<Lsb0>()[offset..][0..len];
```
AdamGS: nit, `[offset..][..len]`

lwwmanning (commented):
(Have Claude) write a test

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -208,17 +210,21 @@ impl VectorExt for Vector {
...
+                            byte_slice.len().div_ceil(8),
+                        )
                     }
-                }
-                null_count == len
+                };
+                let sliced_bits = &arr_bits.view_bits::<Lsb0>()[offset..][..len];
+
+                let validity = unsafe { self.ensure_validity_slice(len) };
+                validity[offset..][..len].copy_from_bitslice(sliced_bits);
```
robert3005: This should be `validity[..len]`, we always append from zero of the buffer

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -208,18 +210,178 @@ impl VectorExt for Vector {
...
+                        )
                     }
-                }
-                null_count == len
+                };
+                let sliced_bits = &arr_bits.view_bits::<Lsb0>()[offset..][..len];
+
+                let validity = unsafe { self.ensure_validity_slice(len) };
+                validity[..len].copy_from_bitslice(sliced_bits);
+                sliced_bits.count_ones() == len
```
joseph-isaacs: shall we count ones before the copy and skip the alloc and copy if possible?

robert3005: like this?

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -208,18 +210,186 @@ impl VectorExt for Vector {
...
+                if true_count == len {
+                    if let Some(validity) = unsafe { self.validity_slice_mut(len) } {
+                        validity.fill(true);
                     }
+                } else if true_count == 0 {
+                    unsafe { self.ensure_validity_slice(len) }.fill(false);
+                } else {
+                    // SAFETY: Caller guaranteees this.
+                    let validity = unsafe { self.ensure_validity_slice(len) };
+                    validity[..len].copy_from_bitslice(sliced_bits);
```
joseph-isaacs, line 233: do you need [..len]?

robert3005: Looks like I do not. I was being defensive

robert3005: You do need the len here since values of the mask are sliced `mask[offset..][..len]` and we request len so validity here will be offset bits longer

joseph-isaacs (approved):

vortex-duckdb/src/exporter/mod.rs
robert3005: ok the actual reason you need the len is that validity buffer here will always be a multiple of 64 while len doesn't need to be

PR #4250, opened 2025-08-15
Use perfetto tracing

gatesn:
This backend generates actual Perfetto protobufs, and so we get events showing up on the correct threads!!

See 
Screenshot 2025-08-15 at 10 33 14

Vs the old chrome:
Screenshot 2025-08-15 at 10 33 20

### DIFF
--- a/bench-vortex/src/query_bench.rs
+++ b/bench-vortex/src/query_bench.rs
@@ -7,9 +7,9 @@ use std::fs::File;
 use std::io::{Write, stdout};
 use std::path::PathBuf;
 
+use crate::Target;
 use crate::display::{DisplayFormat, print_measurements_json, render_table};
 use crate::measurements::{MemoryMeasurement, QueryMeasurement};
-use crate::{Target, default_env_filter};
 
 /// Common benchmark configuration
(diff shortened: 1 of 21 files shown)


0ax1 (approved):

bench-vortex/src/query_bench.rs
```diff
@@ -25,17 +25,13 @@ pub struct BenchmarkConfig {
 }
 
 /// Initialize logging/tracing for a benchmark
-pub fn setup_logging_and_tracing(
-    verbose: bool,
-    trace_file: &str,
-) -> anyhow::Result<Option<tracing_chrome::FlushGuard>> {
+pub fn setup_logging_and_tracing(verbose: bool, trace_file: &str) -> anyhow::Result<()> {
```
robert3005: this never errors (it also didn't error before). You could return nothing

gatesn: Gotta create a file

robert3005: Genuinely the version I was looking at didn't create a file but now it does, ignore me

PR #2456, opened 2025-02-21
feat: basic BoolBuffer / BoolBufferMut

a10y:
It's very annoying to do random in-place setting of bits in Arrow BooleanBufferBuilder. It's very append focused. For something like SparseArray in particular, you want to create a BoolArray against uninitialized memory and only set the bits corresponding to positions of non-null values.

PR #4251, opened 2025-08-15
feat: Remove 'parquet' feature and some other dependency bloat

AdamGS:
This is a step in the right direction, removes two of our slowest-to-build dependencies which are not really needed. I just re-implemented the cast logic we care about and parquet was just there for an error variant.

To measure the difference, I ran
```bash
cargo clean; cargo build --release -p vortex --timings --all-features
```

Currently on develop `vortex` has 413 dependencies, with the top 10 in build time being:
Screenshot 2025-08-15 at 10 58 51

With this PR it has 327 dependencies and the top 10 are:
Screenshot 2025-08-15 at 11 03 50

`arrow-ord` and `arrow-select` do some heavier lifting that I think will be harder to replace, but we should also do that at some point.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,7 +121,7 @@ moka = { version = "0.12.10", default-features = false }
 multiversion = "0.8.0"
 num-traits = "0.2.19"
 num_enum = { version = "0.7.3", default-features = false }
-object_store = { version = "0.12.3", features = ["aws"] }
+object_store = { version = "0.12.3", default-features = false }
 once_cell = "1.21"
 opentelemetry = "0.30.0"
 opentelemetry-otlp = "0.30.0"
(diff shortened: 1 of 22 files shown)


Cargo.toml
```diff
@@ -121,7 +121,7 @@ moka = { version = "0.12.10", default-features = false }
 multiversion = "0.8.0"
 num-traits = "0.2.19"
 num_enum = { version = "0.7.3", default-features = false }
-object_store = { version = "0.12.3", features = ["aws"] }
+object_store = { version = "0.12.3", default-features = false }
```
robert3005, line 124: where is this feature enabled now?

AdamGS: I think this was enabled here for duckdb, so moved it there. Benchmarks and `vortex-ffi` already enabled a bunch of cloud variants.

robert3005:
with https://github.com/vortex-data/vortex/pull/2456 and moving compute functions out of core crate we could remove all of arrow from critical path

robert3005 (approved):

PR #4252, opened 2025-08-15
chore[fastlanes]: remove FoR unsigned transmute

joseph-isaacs:


robert3005 (changes requested):
I don't think you can just do this for exisiting files. The dtypes will be wrong?

encodings/fastlanes/src/for/serde.rs
```diff
@@ -41,7 +41,7 @@ impl SerdeVTable<FoRVTable> for FoRVTable {
         }
 
         let ptype = PType::try_from(dtype)?;
-        let encoded_dtype = DType::Primitive(ptype.to_unsigned(), dtype.nullability());
+        let encoded_dtype = DType::Primitive(ptype, dtype.nullability());
```
robert3005: You can remove these completely, encoded value has the same dtype

robert3005 (approved):

PR #4254, opened 2025-08-15
feat[vortex-dict]: implement ToArrowKernel for DictVTable

asubiotto:


asubiotto:
cc @robert3005

vortex-array/src/arrow/compute/to_arrow/dict.rs
```diff
@@ -17,3 +21,40 @@ impl<K: ArrowDictionaryKeyType> FromArrowArray<&DictionaryArray<K>> for DictArra
...
+            ),
+            Some(DataType::Dictionary(codes_type, values_type)) => (
+                array.codes().clone().into_arrow(codes_type)?,
+                array.values().clone().into_arrow(values_type)?,
+            ),
+            _ => {
+                // Unsupported type.
+                return Ok(None);
+            }
+        };
```
robert3005, line 48: I think you do want to have the existing behaviour where the resulting array will be the decompressed array unless you give it a dictionary arrow_type

asubiotto: This fall-back behavior already exists. If no kernel returns a `Some` result, we fall back to canonicalizing and then converting see: https://github.com/vortex-data/vortex/blob/4f6ea563cf3e3132675093f05420e6fbba68f45a/vortex-array/src/arrow/compute/to_arrow/mod.rs#L103-L114

encodings/dict/src/arrow.rs
```diff
@@ -17,3 +21,40 @@ impl<K: ArrowDictionaryKeyType> FromArrowArray<&DictionaryArray<K>> for DictArra
         DictArray::try_new(keys, values).vortex_unwrap()
     }
 }
+
+inventory::submit!(ToArrowKernelAdapter(DictVTable).lift());
```
a10y: We have a wrapper macro `register_kernel!` in the vortex_array crate that I think you'll want instead

https://docs.rs/vortex-array/latest/vortex_array/macro.register_kernel.html

vortex-array/src/arrow/compute/to_arrow/dict.rs
robert3005: yes, so we should return `Ok(None)` when arrow_type is not specified and only return dictionary if arrow type is the dictionary

asubiotto: AFAIU when an `arrow_type` is not specified the contract is that the vortex array is converted to an arrow array based on the encoding's preferred `DataType` (https://github.com/vortex-data/vortex/blob/e5ae65dd2c6b47fe1b16a38cedde29dc7932f95b/vortex-array/src/arrow/compute/to_arrow/mod.rs#L36). In the case of dictionary encoding it seems like if no arrow type is specified, converting to the implicit arrow dictionary type should be preferred rather than falling back to canonicalizing and then converting, since the latter is more expensive.

asubiotto:
Thanks for the reviews. I will address comments and CI on Monday.

encodings/dict/src/arrow.rs
asubiotto: Done.

asubiotto:
I'm not sure how to best fix CI. The core issue is that the consistency tests call out to compute kernels that sometimes use `ArrayRef::from_arrow` (https://github.com/vortex-data/vortex/blob/d578fe02e147ebef2f6e49596a33ca168ec2425c/vortex-array/src/compute/mask.rs#L134) which isn't implemented for dictionary types. We implemented from arrow conversions as `FromArrowArray` on `DictArray` in `vortex-dict` and can't define it on an `ArrayRef` due to the orphan rule. I could possibly fix CI by not implementing to_arrow conversions for preferred types but I would like to have this behavior so that ultimately the datafusion `VortexSource` gives me back dictionary-encoded arrow arrays without requiring two conversions.

robert3005:
In order for us to return dictionaries by default we need to move code around so arrow conversion doesn’t live in the main crate. I will do that today which would let you fix the ci problems

robert3005:
I attempted the move and it's a lot more intertwined/involved than I originally assumed. There's some more work required in order to decouple compute from arrays. In the meantime to get the test passing you would need to default to not returning dict array if the type is not specified.

asubiotto:
Thanks! No problem, this behavior is needed for us so I don't mind waiting to merge this and running off a fork in the meantime.

asubiotto:
@robert3005 with the recent IO changes, this change will soon be the only thing requiring us to use a fork of vortex. Is there anything I can do to help with the move to `vortex-arrow`?

robert3005:
I was halfway through doing this but realised there's a bit of API design change necessary. What we want to do is move compute functions out of the main crate so we don't have to always recompile them and keep them optional. Arrow is just another compute function and that crate cna take dependency on other vortex crates. However, the problem we have is that stats are part of the Array trait and stats are defined using compute functions. Therefore we need to change how stats use compute functions, likely means there's a trait you implement that can delegate to compute functions or not but haven't had time to think through this yet.

gatesn:
Actually the aim was to make vortex-arrow its own crate, at which point ToArrow _shouldn't_ be a compute function, and it can just be a regular Rust trait.

robert3005:
Once you split compute from the main crate you can split arrow from compute

asubiotto:
Making some progress thanks to @onursatici's recent changes 🥳 Seems like there's now a failure on tpch queries probably due to the preferred encoding of string dictionaries resolving to dict<_, utf8view>. I'll take a look when I have a moment:
```
thread 'main' panicked at /home/runner/_work/vortex/vortex/bench-vortex/src/benchmark_driver.rs:182:33:
query: 1 failed with: Arrow error: Compute error: Internal Error: Cannot cast Utf8View to StringArray of expected type
Backtrace:
   0: bench_vortex::benchmark_driver::execute_queries::<bench_vortex::tpch::tpch_benchmark::TpcHBenchmark>::{closure#0}::{closure#0}::{closure#0}::{closure#0}
   ```

asubiotto:
This PR is now ready for a review again. Everything is green apart from the benchmarks. The main changes since the initial PR are:
- Coerce dictionary encoded utf8view values to dict<_, utf8> since dict<_, utf8view> essentially double-encodes dictionaries and isn't well supported by arrow (this was the TPC-H failure, missing arrow cast arm).
- In compare/numeric, I force canonicalization when falling back to arrow to force dictionary expansion. The benchmark regressions are because of this. I don't think the regressions matter that much since we're talking about arrow fallback and the canonicalization was happening anyway for dict-encoded arrays. I might be wrong though. I can add more special-casing to make the benchmarks look good.

cc @onursatici

gatesn:
Am I right in thinking for you use case you will always be passing an Arrow target DataType?

asubiotto:
> Am I right in thinking for you use case you will always be passing an Arrow target DataType?

No. The datafusion `VortexOpener` converts to the preferred arrow target type. We *could* plumb it down, but I also don't think it's right for dictionary encoding on disk to be expanded when converting to arrow when not specifying a target type. i.e. dictionary encoding should "prefer" to maintain zero-copy when decoding to arrow.

gatesn:
I didn't think DataFusion handled dictionary / other encodings that weren't == the declared schema? Or am I out of date here?

asubiotto:
> I didn't think DataFusion handled dictionary / other encodings that weren't == the declared schema? Or am I out of date here?

You're right, but that's taken care of in the schema adapter step with an arrow batch cast to cast the file batches to the table schema if there is a difference. So it's correct but could be more performant. I think there are two questions here:
1) Should we pass in a target arrow type from datafusion to vortex? Yes, probably. This will solve our use-case and seems like generally a thing we should do. I agree here.
2) Should vortex dictionary arrays be able to zero-copy to arrow when no target type is specified? I think they should regardless of datafusion and/or our use-case

gatesn:
Ahhh, gotcha.

1. Yes 100%
2. Yes 100%

asubiotto:
This PR is not needed anymore now that the operator path is on by default.

PR #4261, opened 2025-08-16
chore: Fix all rustdoc warnings and enforce no warning in CI

AdamGS:


vortex-file/src/strategy.rs
```diff
@@ -37,7 +37,7 @@ impl Default for WriteStrategyBuilder {
 
 impl WriteStrategyBuilder {
     /// Create a new empty builder. It can be further configured, and then finally built
-    /// yielding the [`WriteStrategy`].
```
AdamGS, line 40: @a10y there is something going on here with the naming but I'm missing context.

vortex-datafusion/src/persistent/source.rs
```diff
@@ -34,9 +34,9 @@ use super::metrics::PARTITION_LABEL;
 use super::opener::VortexOpener;
 use crate::convert::exprs::can_be_pushed_down;
 
-/// A config for [`VortexFileOpener`]. Used to create [`DataSourceExec`] based physical plans.
+/// Execution plan for reading one or more Parquet files, intended to be consumed by [`DataSourceExec`].
```
a10y: Vortex files?

vortex-file/src/strategy.rs
a10y: You're right this is more consistent. At one point I had changed the name but then I un-changed it and this was just a leftover mistake that somehow didn't trip CI checks

vortex-datafusion/src/persistent/source.rs
AdamGS: probably

robert3005 (approved):

PR #4263, opened 2025-08-16
experiment: turn arbitrary into a cfg instead of a feature

AdamGS:
Inspired by [this](https://www.reddit.com/r/rust/comments/1mrmdft/speed_wins_when_fuzzing_rust_code_with/) reddit discussion, IDK what I think about it but figured I want to see how it looks.

We don't derive Arbitrary anywhere (even though I think we could for something like ptype?), so its probably not a huge difference either way, but it does remove a bunch of dependencies when running stuff like clippy with `all-features`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ version = "0.1.0"
 
 [workspace.dependencies]
 anyhow = "1.0.95"
-arbitrary = "1.3.2"
+arbitrary = "1.4.2"
 arcref = "0.2.0"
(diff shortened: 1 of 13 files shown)


robert3005:
Is this pr ready for review?

AdamGS:
sort of? Its red for some reason but the general concept is clear if you just want to take a look. I'll take a look into what's broken.

AdamGS:
I don't think this will make any difference for us, and there's something here I can't figure out, so closing it.

PR #4266, opened 2025-08-18
Report doc coverage on PRs

AdamGS:


AdamGS:
the action and underlying command have some issues with the workspace, but I think I can get something working around it, closing it for now to put it on the back burner.

PR #4269, opened 2025-08-18
Enable TPC-DS benchmarks for datafusion

### LINKED ISSUE #4268: Enable TPC-DS benchmarks for datafusion (Parquet + Vortex).
We just [merged](https://github.com/vortex-data/vortex/pull/4267) an update to datafusion 49.0.1, which enables running the TPC-DS benchmark with datafusion.

The only unresolved issue there is that query 72 has this piece of SQL:
```sql
AND d3.d_date > d1.d_date + 5
``` 
and datafusion doesn't like it:
```
Error during planning: Cannot coerce arithmetic expression Date32 + Int64 to valid types
```
We can either skip it, but maybe just adding some casting to the query is an acceptable change here.

AdamGS:
Closes #4268

### DIFF
--- a/.github/workflows/sql-benchmarks.yml
+++ b/.github/workflows/sql-benchmarks.yml
@@ -42,7 +42,7 @@ on:
               "id": "tpcds-nvme",
               "subcommand": "tpcds",
               "name": "TPC-DS SF=1 on NVME",
-              "targets": "duckdb:parquet,duckdb:vortex,duckdb:duckdb",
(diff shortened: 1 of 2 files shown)


joseph-isaacs (approved):
lets wait until the benchmarks pass?

AdamGS:
From running it locally, it seems like datafusion is having a hard time with some of these queries but vortex is doing pretty well compared to parquet, will be easier to see trends once we have more runs.

PR #4276, opened 2025-08-18
feat: teach vortex.io.write how to configure zone_size and compressio…

danking:
…n_block_min_bytesize

### DIFF
--- a/vortex-file/src/strategy.rs
+++ b/vortex-file/src/strategy.rs
@@ -27,6 +27,12 @@ const ROW_BLOCK_SIZE: usize = 8192;
 pub struct WriteStrategyBuilder {
     executor: Option<Arc<dyn TaskExecutor>>,
     compressor: Option<Arc<dyn CompressorPlugin>>,
+    /// The minimum size, in bytes, of a block of values which are compressed together.
+    compression_block_min_bytesize: Option<u64>,
+    /// The size of zones, each of which carries statistics such as the minimum value.
+    ///
+    /// Smaller zones allow for finer-grained filter push down at the cost of more metadata.
+    zone_size: Option<usize>,
 }
 
 impl Default for WriteStrategyBuilder {
@@ -42,6 +48,8 @@ impl WriteStrategyBuilder {
         Self {
             executor: None,
             compressor: None,
+            compression_block_min_bytesize: None,
+            zone_size: None,
         }
     }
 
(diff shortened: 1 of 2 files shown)


danking:
Tagging a bunch of folks because I think this is controversial.

vortex-file/src/strategy.rs
```diff
@@ -27,6 +27,12 @@ const ROW_BLOCK_SIZE: usize = 8192;
 pub struct WriteStrategyBuilder {
     executor: Option<Arc<dyn TaskExecutor>>,
     compressor: Option<Arc<dyn CompressorPlugin>>,
+    /// The minimum size, in bytes, of a block of values which are compressed together.
+    compression_block_min_bytesize: Option<u64>,
+    /// The size of zones, each of which carries statistics such as the minimum value.
+    ///
+    /// Smaller zones allow for finer-grained filter push down at the cost of more metadata.
+    zone_size: Option<usize>,
```
a10y: rows? bytes? hectares?

danking: addressed.

vortex-file/src/strategy.rs
```diff
@@ -63,15 +71,30 @@ impl WriteStrategyBuilder {
         self
     }
 
+    /// Override the minimum bytesize of a block for compression.
+    pub fn with_compression_block_min_bytesize(mut self, min_bytesize: Option<u64>) -> Self {
```
a10y, line 75: a bit wordy? maybe something like

```suggestion
    pub fn with_compression_block_size(mut self, block_size: u64) -> Self {
```

Also I think this should take u64 instead of Option<u64>

vortex-file/src/strategy.rs
```diff
@@ -63,15 +71,30 @@ impl WriteStrategyBuilder {
...
+        self.compression_block_min_bytesize = min_bytesize;
+        self
+    }
+
+    /// Override the size, in rows, of zone-map zones.
+    pub fn with_zone_size(mut self, zone_size: Option<usize>) -> Self {
```
a10y, line 81:
```suggestion
    pub fn with_zone_size(mut self, rows: usize) -> Self {
```

vortex-file/src/strategy.rs
```diff
@@ -63,15 +71,30 @@ impl WriteStrategyBuilder {
...
+        let compression_block_min_bytesize = self.compression_block_min_bytesize.unwrap_or(ONE_MEG);
+        let block_size = self.zone_size.unwrap_or(ROW_BLOCK_SIZE);
+
         let executor = self.executor.unwrap_or_else(|| Arc::new(LocalExecutor));
 
         // 7. for each chunk create a flat layout
         let chunked = ChunkedLayoutStrategy::new(FlatLayoutStrategy::default());
         // 6. buffer chunks so they end up with closer segment ids physically
-        let buffered = BufferedStrategy::new(chunked, 2 * ONE_MEG); // 2MB
+        let buffered = BufferedStrategy::new(chunked, 2 * compression_block_min_bytesize); // 2MB
```
a10y, line 97:
```suggestion
        let buffered = BufferedStrategy::new(chunked, compression_block_min_bytesize); // 2MB
```

vortex-file/src/strategy.rs
pavanramkumar: rows would be nice to have. coming from lance, i'm able to specify some high level compression settings like this: https://github.com/slaf-project/slaf/blob/main/slaf/data/converter.py#L1221

danking:
This needs more thought. Maybe we start by exposing compact and standard?

PR #4278, opened 2025-08-18
Chore: Cleanup in `take`

connortsui20:


a10y (approved):

vortex-array/src/compute/take.rs
```diff
@@ -71,21 +71,19 @@ impl ComputeFnVTable for Take {
...
         }
 
+        let taken_array = take_impl(array, indices, kernels)?;
+
         // We know that constant array don't need stats propagation, so we can avoid the overhead of
         // computing derived stats and merging them in.
-        let derived_stats = (!array.is_constant()).then(|| derive_take_stats(array));
-
-        let taken = take_impl(array, indices, kernels)?;
+        if !array.is_constant() {
```
robert3005:
```suggestion
        if !taken_array.is::<ConstantVTable>() {
```

vortex-array/src/compute/take.rs
```diff
@@ -71,21 +71,19 @@ impl ComputeFnVTable for Take {
...
 
+        let taken_array = take_impl(array, indices, kernels)?;
+
         // We know that constant array don't need stats propagation, so we can avoid the overhead of
         // computing derived stats and merging them in.
-        let derived_stats = (!array.is_constant()).then(|| derive_take_stats(array));
-
-        let taken = take_impl(array, indices, kernels)?;
+        if !taken_array.is::<ConstantVTable>() {
+            let derived_stats = derive_take_stats(array);
```
robert3005, line 79: I was looking at this code and what you want to happen here is still run the `combine_sets` or some variant of that function. I think we need to change the methods to not require a StatsSet but instead operate on some borrowed type like StatsSetRef

PR #4283, opened 2025-08-19
chore: monad vs. monoid

connortsui20:
A monad is a monoid in the category of endofunctors.

### DIFF
--- a/vortex-array/src/compute/take.rs
+++ b/vortex-array/src/compute/take.rs
@@ -84,7 +84,7 @@ impl ComputeFnVTable for Take {
             stats.combine_sets(&derived_stats, array.dtype())?;
 
             for (stat, val) in stats {
-                // Alternatively, use a monadic reducer here to set `stat = val`, or if it already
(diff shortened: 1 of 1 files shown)


vortex-array/src/compute/take.rs
```diff
@@ -84,7 +84,7 @@ impl ComputeFnVTable for Take {
             stats.combine_sets(&derived_stats, array.dtype())?;
 
             for (stat, val) in stats {
-                // Alternatively, use a monadic reducer here to set `stat = val`, or if it already
+                // Alternatively, use a monoidal pattern here to set `stat = val`, or if it already
```
danking, line 87: If we're being pedantic here do we even need an identity?

danking (approved):

vortex-array/src/compute/take.rs
connortsui20: Technically the empty map is the identity so nothing gets updated

PR #4285, opened 2025-08-19
feat[vortex-array]: support converting to timestamp arrow arrays with tz

asubiotto:


asubiotto:
cc @robert3005

vortex-array/src/arrow/compute/to_arrow/temporal.rs
```diff
@@ -112,3 +112,24 @@ where
...
     Ok(Arc::new(ArrowPrimitiveArray::<T>::new(values, nulls)))
 }
+
+fn to_arrow_timestamp<T: ArrowTimestampType>(
+    array: &TemporalArray,
+    arrow_tz: &Option<Arc<str>>,
+) -> VortexResult<ArrowArrayRef>
+where
+    T::Native: NativePType,
+{
```
robert3005, line 128: This shouldn't be a completely separate function, change `to_arrow_temporal` to accept timezone instead?

robert3005 (changes requested):
we should remove some code duplication

vortex-array/src/arrow/compute/to_arrow/temporal.rs
asubiotto: Sounds good, will fix tomorrow. I was debating whether it made sense given the other types don't have a notion of a timezone.

asubiotto: Done, although I had to use a shared function. `to_arrow_temporal` can't do anything with a timezone since it is generic over `ArrowPrimitiveType` and `with_timezone` is only available with `ArrowTimestampType`s. I felt like casting to remove code duplication was the wrong tradeoff. The other option would be to remove `to_arrow_temporal` and call `to_arrow_temporal_inner` in all the callsites, but that requires some boilerplate (maybe there's a macro somewhere I can use, can also write). Happy to do whatever you prefer.

robert3005: You don't need any casting since ArrowTimestampType requires ArrowTemporalType. https://github.com/vortex-data/vortex/commit/523329aa8d29bb484a1c45876e97fc5afd13a43e is a slight change to your pr that I think is quite nice

robert3005 (approved):

vortex-array/src/arrow/compute/to_arrow/temporal.rs
asubiotto: I was referring to casting in order to have only one function that did everything (instead of the current three). Your modification sounds good! Cherry picked and pushed.

asubiotto:
Thanks for the review!

PR #4286, opened 2025-08-19
chore: infinite publish timeout

a10y:


robert3005 (approved):

.github/workflows/publish.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       - name: Release
         run: |
-          cargo +nightly publish -Z package-workspace --no-verify --allow-dirty --workspace \
```
lwwmanning, line 29: We don't need package-workspace anymore?

lwwmanning (approved):

.github/workflows/publish.yml
robert3005:
```
warning: flag `-Z package-workspace` has been stabilized in the 1.89 release, and is no longer necessary
  Workspace packaging and publishing (a.k.a. `-Zpackage-workspace`) is now always enabled.
```

Issue #1366, opened 2024-11-19
Create issues from fuzzing failures


AdamGS:
Instead of persisting fuzzing failures into the corpus which makes the fuzzer fail again and again, we should create issues with the reproduction artifact, allowing the fuzzer to keep exploring the space for more bugs. This will also make failures more visible.

robert3005:
This exists, though we need it to properly deduplicate

PR #4287, opened 2025-08-19
chore: slack message on fuzz failure

a10y:
Sometimes we don't notice when the fuzzer has found a bug for awhile. This will post into a new #ext-vortex-fuzz on our internal Slack instance for now

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -121,3 +121,20 @@ jobs:
       - name: Fail job if fuzz run found a bug
         if: steps.fuzz.outcome == 'failure'
         run: exit 1
+  slack_alert:
+    runs-on: ubuntu-latest
+    if: always() && (needs.io_fuzz.result == 'failure' || needs.ops_fuzz.result == 'failure')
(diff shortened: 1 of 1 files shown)


.github/workflows/fuzz.yml
```diff
@@ -121,3 +121,20 @@ jobs:
...
         run: exit 1
+  slack_alert:
+    runs-on: ubuntu-latest
+    if: always() && (needs.io_fuzz.result == 'failure' || needs.ops_fuzz.result == 'failure')
+    needs: [io_fuzz, ops_fuzz]
+    steps:
+      - name: Send slack message
+        uses: slackapi/slack-github-action@v2.1.1
+        with:
+          webhook: ${{ secrets.SLACK_FUZZ_WEBHOOK_URL }}
```
a10y, line 132: I've added this secret in the repo

robert3005:
I am not a fan of using slack notifications for this. How do you know that the issue is a new one vs not? Since fuzzing is deterministic we should use the seeds as a unique identifier and create issues as in #1366

PR #4291, opened 2025-08-19
chore: `DType` documentation

connortsui20:
Also moves the `FieldName` and `FieldNames` to `field_names.rs` (it was confusing going to `dtype.rs` and only seeing `DType` defined halfway down).

I'm not sure if this is too much documentation, but it feels better than having just 2 lines?


robert3005:
Move the FieldNames to a separate file?

robert3005 (approved):

PR #4300, opened 2025-08-20
chore: move `field_names` tests to `field_names.rs`

connortsui20:
Also clears up some variable names in match statements.

Somehow I missed this in #4291...

### DIFF
--- a/vortex-dtype/src/dtype.rs
+++ b/vortex-dtype/src/dtype.rs
@@ -102,18 +102,16 @@ impl DType {
 
     /// Check if the DType is nullable
     pub fn is_nullable(&self) -> bool {
-        use crate::nullability::Nullability::*;
-
         match self {
             Null => true,
             Extension(ext_dtype) => ext_dtype.storage_dtype().is_nullable(),
-            Bool(n)
-            | Primitive(_, n)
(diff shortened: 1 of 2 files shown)


vortex-dtype/src/dtype.rs
```diff
@@ -107,13 +107,13 @@ impl DType {
...
-            | Binary(n)
-            | Struct(_, n)
-            | List(_, n) => matches!(n, Nullable),
+            Bool(nullability)
+            | Primitive(_, nullability)
+            | Decimal(_, nullability)
+            | Utf8(nullability)
+            | Binary(nullability)
+            | Struct(_, nullability)
+            | List(_, nullability) => matches!(nullability, Nullable),
```
robert3005: I'm not convinced the longer name actually helps here

robert3005 (approved):

vortex-dtype/src/dtype.rs
connortsui20: I can change it to something else, but at a quick glance `n` is super confusing

connortsui20: Maybe just `null`?

PR #4301, opened 2025-08-20
fix: downscale FSST offsets

a10y:
Similar to what we do for ListArray and VarBin, the compressor should be downscaling before calling into the IntCompressor.

### DIFF
--- a/vortex-array/src/arrays/varbin/compute/take.rs
+++ b/vortex-array/src/arrays/varbin/compute/take.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::iter::Sum;
+
 use num_traits::PrimInt;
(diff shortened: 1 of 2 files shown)


robert3005 (approved):

danking (approved):

danking:
Hmm. VarBin assumes the post-take offsets fit inside its offset type, but that's not true. Trivially, the array: `["x"]` has offset size 0u8 but a take vector `[0 for _ in range(1 << 16)]` produces an array whose offsets extend into u16. I think maybe you need to compute the sum of the offset of the taken indices first to determine the necessary size?

danking:
But this is also not a this-PR-problem, this is a latent bug.

a10y:
I just switched to u32 to skirt around the issue, they will be downscaled later anyway once they are sent through the compressor

danking:
I created a tracking issue https://github.com/vortex-data/vortex/issues/4302

PR #4303, opened 2025-08-20
chore: Make downscale_integer_array a method on PrimitiveArray

robert3005:


vortex-array/src/arrays/primitive/downscale.rs
```diff
@@ -0,0 +1,82 @@
...
+use vortex_dtype::{DType, PType};
+use vortex_error::VortexResult;
+
+use crate::ToCanonical;
+use crate::arrays::PrimitiveArray;
+use crate::compute::{cast, min_max};
+use crate::validity::Validity;
+
+impl PrimitiveArray {
+    pub fn downscale(&self) -> VortexResult<PrimitiveArray> {
```
danking: If we're rewriting things anyway, maybe this should be called `downcast`? `to_min_bitwidth`?

danking (approved):

vortex-array/src/arrays/primitive/downscale.rs
robert3005: let's go with downcast

PR #4306, opened 2025-08-21
chore: return lvalue ref for cxx's scanbuilder functions

XinyuZeng:
rationales described here: https://github.com/vortex-data/vortex/pull/4050#discussion_r2279001069 https://github.com/vortex-data/vortex/pull/4050#discussion_r2281669025


vortex-cxx/cpp/include/vortex/scan.hpp
```diff
@@ -56,23 +56,23 @@ class ScanBuilder {
     ScanBuilder &operator=(const ScanBuilder &) = delete;
 
     /// Only include rows that match the filter expressions.
-    ScanBuilder &&WithFilter(Expr expr) &&;
+    ScanBuilder &WithFilter(Expr&& expr);
```
0ax1: Let's pass Expr to the scan builder via "perfect forwarding" such that they can be either rvalues or lvalues

```
template<typename ExprType>
ScanBuilder& WithFilter(ExprType&& expr) {
    impl_->with_filter(std::forward<ExprType>(expr).IntoImpl());
    return *this;
}
```
and add function docs that the expected type is `Expr`.

vortex-cxx/cpp/include/vortex/scan.hpp
```diff
@@ -56,23 +56,23 @@ class ScanBuilder {
...
 
     /// Only include columns that match the projection expressions.
-    ScanBuilder &&WithProjection(Expr expr) &&;
+    ScanBuilder &WithProjection(Expr&& expr);
```
0ax1: See: https://github.com/vortex-data/vortex/pull/4306/files#r2290474736

vortex-cxx/cpp/include/vortex/scan.hpp
```diff
@@ -56,23 +56,23 @@ class ScanBuilder {
...
 
     /// Only include rows in the range [row_range_start, row_range_end).
-    ScanBuilder &&WithRowRange(uint64_t row_range_start, uint64_t row_range_end) &&;
+    ScanBuilder &WithRowRange(uint64_t row_range_start, uint64_t row_range_end);
```
0ax1: We should implement both variants for the builder functions.
```
ScanBuilder &&WithRowRange(uint64_t row_range_start, uint64_t row_range_end) &&
ScanBuilder &WithRowRange(uint64_t row_range_start, uint64_t row_range_end) &
```
This allows for conditional settings on the scan builder as well as constructing the scan builder directly in one go. The functions are pretty short, so it's ok to have these in 2 variants.

Same for the `IntoStream` functions.

Turns out the key to overloading is adding the ref qualifier also for 
`ScanBuilder &WithRowRange(uint64_t row_range_start, uint64_t row_range_end) &` at the end.

0ax1: Apparently C++ 23 supports auto-deducing Self.
```
template<typename Self>
auto WithLimit(this Self&& self, uint64_t limit) {
    self.impl_->with_limit(limit);
    return std::forward<Self>(self);  // Perfect forwarding of *this
}
```
C++23 is too recent though to enforce as a requirement for vortex-cxx.

vortex-cxx/examples/hello-vortex.cpp
```diff
@@ -13,8 +13,16 @@ int main(int argc, char *argv[]) {
         return 1;
     }
     std::string vortex_file = argv[1];
-    auto file = vortex::VortexFile::Open(vortex_file);
-    auto stream = file.CreateScanBuilder().WithLimit(100).IntoStream();
+    // 1. bad, dangling lvalue reference!
+    // auto& builder = vortex::VortexFile::Open(vortex_file).CreateScanBuilder().WithLimit(3);
+    // auto stream = std::move(builder).IntoStream();
+    // 2. good, prefer
+    auto builder = vortex::VortexFile::Open(vortex_file).CreateScanBuilder();
```
0ax1: Any combination of this should work without subtle errors.

vortex-cxx/cpp/include/vortex/scan.hpp
```diff
@@ -56,23 +56,23 @@ class ScanBuilder {
     ScanBuilder &operator=(const ScanBuilder &) = delete;
```
0ax1, line 61: I like the idea you mentioned in the other PR guarding against self move assignment.

```
ScanBuilder &operator=(ScanBuilder &&other) noexcept {
  if (this != &other) {
    impl_ = std::move(other.impl_);
  }
  return *this;
} 
```
Let's add this in addition to the other aspects mentioned. Ideally, the API would be impossible to misuse.

vortex-cxx/cpp/include/vortex/scan.hpp
XinyuZeng: Great suggestion. Fixed.

vortex-cxx/cpp/include/vortex/scan.hpp
XinyuZeng: I don't think we should add lvalue variant for `IntoStream` functions, as internally the Rust part can only consume the builder. The ScanBuilder does not implement Clone.

vortex-cxx/examples/hello-vortex.cpp
XinyuZeng: Great. I modified the examples.

vortex-cxx/cpp/include/vortex/scan.hpp
XinyuZeng: fixed

vortex-cxx/cpp/src/scan.cpp
```diff
@@ -4,33 +4,96 @@
...
+template ScanBuilder &ScanBuilder::WithFilter(expr::Expr &&expr) &;
+template ScanBuilder &ScanBuilder::WithFilter(const expr::Expr &expr) &;
+
+template <typename ExprType>
+ScanBuilder &&ScanBuilder::WithFilter(ExprType &&expr) && {
+    if constexpr (std::is_rvalue_reference_v<ExprType &&>) {
+        impl_->with_filter(std::forward<ExprType>(expr).IntoImpl());
+    } else {
+        impl_->with_filter_ref(std::forward<ExprType>(expr).Impl());
+    }
```
XinyuZeng: cxx does not support function overload (although I think technically it can do it), so we have to add separate ffi functions and do static dispatch here.

0ax1: ~~I think the forwarding part is less important than the "reference collapsing" aspect such that the builder function can handle both rvalues and lvalues. Let's drop `std::forward` and a keep a single `with_xyz` not adding a ref version.~~

Just saw that you already added the ref versions. I think that looks good, even if a bit redundant.

vortex-cxx/cpp/include/vortex/scan.hpp
```diff
@@ -45,33 +46,49 @@ class StreamDriver {
...
+        return *this;
+    }
     ~ScanBuilder() = default;
 
     ScanBuilder(const ScanBuilder &) = delete;
     ScanBuilder &operator=(const ScanBuilder &) = delete;
 
     /// Only include rows that match the filter expressions.
-    ScanBuilder &&WithFilter(Expr expr) &&;
+    /// Expect `ExprType` to be `expr::Expr &&` or `const expr::Expr &`
```
0ax1: nit: `const expr::Expr &` doesn't need to be const, right?

vortex-cxx/cpp/src/scan.cpp
XinyuZeng: I think we should drop `std::forward` and simply leave the overload variants for both rvalue and lvalue ref of `Expr`. I am working on that now.

XinyuZeng: If there are only two variants, and we still need specialization for both of them (to call the `with_filter_ref` ffi), then we do not need forward I think

XinyuZeng: Would you mind taking a look at the modifications of my latest commit?  I think it is cleaner

0ax1: > If there are only two variants, and we still need specialization for both of them (to call the with_filter_ref ffi), then we do not need forward I think

Yes, this only works if the function we're forwarding to also supports r and l values. Thanks to C++ that's a silly amount of redudancy but this is fine for now. The most important part is that the API is hard to misuse now. We can iterate on making this more compact down the line.

0ax1: I'll have a second thought here also on whether C++23 might be ok to use so we could express this with reference collapsing and auto-deducing `this` in order to have a single func definition.

0ax1 (approved):
Thanks! @XinyuZeng

vortex-cxx/cpp/src/scan.cpp
XinyuZeng: I think C++ 20 may be ok but C++23 is too aggressive. As I know someone is still using C++17 (e.g., my current employer). Arrow-cpp recently had a [vote](https://lists.apache.org/thread/48zlj0dn2y0f53y2k37qsr90y781wfnj) to switch to 20 but it has not happened yet.

PR #4307, opened 2025-08-21
feat[dtype]: ptype visitor

joseph-isaacs:


gatesn (approved):
There are probably some places we could use this in place of match_each_native_ptype

joseph-isaacs:
Ye, going to come back to this

vortex-dtype/src/ptype.rs
```diff
@@ -104,6 +104,80 @@ pub trait NativePType:
...
+
+    /// Downcast the provided object to a type-specific instance.
+    fn downcast<V: PTypeVisitor + ?Sized>(visitor: &V) -> V::Output<Self>;
+
+    /// Downcast the provided object to a type-specific instance.
+    fn downcast_mut<V: PTypeVisitorMut + ?Sized>(visitor: &mut V) -> V::Output<Self>;
+}
+
+/// A visitor trait for converting a `NativePType` to another parameterized type.
+#[allow(missing_docs)] // Kind of obvious..
```
robert3005, line 116: yet you added the docs? ;) Same below

joseph-isaacs: @gatesn if from a bigger pr

PR #4316, opened 2025-08-21
fix: Spawn datafusion write tasks

### LINKED ISSUE #4315: Datafusion VortexSink hangs when writing > 40_960 entries
While experimenting with `vortex-datafusion`, I observed that any attempt to write more than 40_960 entries (regardless of schema) causes the execution to hang indefinitely.

Minimized example that reproduces the issue:
```rust
use datafusion::arrow::array::Int8Array;
use datafusion::arrow::datatypes::{DataType, Field, Schema};
use datafusion::arrow::record_batch::RecordBatch;
use datafusion::datasource::file_format::format_as_file_type;
use datafusion::logical_expr::LogicalPlanBuilder;
use datafusion::prelude::SessionContext;
use std::default::Default;
use std::sync::Arc;
use tempfile::TempDir;
use vortex_datafusion::VortexFormatFactory;

const ENTRIES: usize = 40_961; // Works as expected for 0 to 40960, hangs for 40961

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let dir = TempDir::new()?;

    let session = SessionContext::new();

    let data = session.read_batch(RecordBatch::try_new(
        Arc::new(Schema::new(vec![Field::new("a", DataType::Int8, false)])),
        vec![Arc::new(Int8Array::from(vec![0i8; ENTRIES]))],
    )?)?;

    let logical_plan = LogicalPlanBuilder::copy_to(
        data.logical_plan().clone(),
        dir.path().to_str().unwrap().to_string(),
        format_as_file_type(Arc::new(VortexFormatFactory::default())),
        Default::default(),
        vec![],
    )?
    .build()?;

    session
        .execute_logical_plan(logical_plan)
        .await?
        .collect()
        .await?;

    Ok(())
}
```

> [!NOTE]
> Changing `ENTRIES` to `40_960` (or lower) allows the process to succeed

AdamGS:
Closes https://github.com/vortex-data/vortex/issues/4315, spawns writer tasks so the upstream data producer doesn't block.

### DIFF
--- a/vortex-datafusion/src/persistent/sink.rs
+++ b/vortex-datafusion/src/persistent/sink.rs
@@ -105,7 +105,8 @@ impl FileSink for VortexSink {
             let object_store = object_store.clone();
             let dtype = DType::from_arrow(self.config.output_schema.clone());
 
-            // We need to spawn work because there's a dependency between
+            // We need to spawn work because there's a dependency between the different files. If one file has too many batches buffered,
+            // the demux task might deadlock itself.
             file_write_tasks.spawn(async move {
                 let stream = ReceiverStream::new(rx).map(move |rb| {
                     row_counter.fetch_add(rb.num_rows() as u64, Ordering::Relaxed);
--- a/vortex-datafusion/src/persistent/sink.rs
+++ b/vortex-datafusion/src/persistent/sink.rs
@@ -8,7 +8,7 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use arrow_schema::SchemaRef;
 use async_trait::async_trait;
 use datafusion_common::{DataFusionError, Result as DFResult};
-use datafusion_common_runtime::SpawnedTask;
+use datafusion_common_runtime::{JoinSet, SpawnedTask};
 use datafusion_datasource::file_sink_config::{FileSink, FileSinkConfig};
 use datafusion_datasource::sink::DataSink;
 use datafusion_datasource::write::demux::DemuxedStreamReceiver;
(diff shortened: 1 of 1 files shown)


vortex-datafusion/src/persistent/sink.rs
```diff
@@ -95,24 +96,49 @@ impl FileSink for VortexSink {
...
+        // 1. We can probably be better at signaling how much memory we're consuming (potentially when reading too), see ParquetSink::spawn_writer_tasks_and_join.
         while let Some((path, rx)) = file_stream_rx.recv().await {
             let row_counter = row_counter.clone();
-            let stream = ReceiverStream::new(rx).map(move |rb| {
-                row_counter.fetch_add(rb.num_rows() as u64, Ordering::Relaxed);
-                VortexResult::Ok(ArrayRef::from_arrow(rb, false))
+            let object_store = object_store.clone();
+            let dtype = DType::from_arrow(self.config.output_schema.clone());
+
+            // We need to spawn work because there's a dependency between
```
robert3005: dependency between...? The suspense is killing me

AdamGS: 😂

AdamGS: expanded a bit, hard to capture it in a reasonably-sized comment, and its also a private implementation details of datafusion that I assume doesn't really have any guarantees.
I'll try and find some time next week to improve of the public docs there (in datafusion) a bit, maybe will help us get some clarity about the behavior here.

robert3005 (approved):

PR #4317, opened 2025-08-21
deny `clippy::wildcard_enum_match_arm`

connortsui20:
This PR turns on the clippy lint `clippy::wildcard_enum_match_arm`, which prevents catch-all arms in `match` statements. You can view the docs [here](https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_enum_match_arm).

If enums ever get updated (for example `DType`), catch-all arms are quite bad because the compiler won't catch if a match statement needs to be updated with the new variant.
You can opt out of this with `#[allow(clippy::wildcard_enum_match_arm)]`, or you can simply use different syntax (`if let` or `let else` or `matches!`) to avoid this.

---

For now, this is just a subset of the crates to make this reviewable.

The protocol I followed was basically:
- If the match statement clearly has only 1 possible "correct" arm, convert into an `if let` statement.
- If it is very obvious that a new variant will not be added that will affect how the existing match arms match, then we can use `#[allow(clippy::wildcard_enum_match_arm)]` with some comments.
  - _An example of this _not_ being the case if for `VarBinView`, where right now this matches against `Utf8` and `Binary`, but it may be possible for us to add `Utf16` of something in the future..._
- If it is a very large hassle to update the match statement, then we can use `#[allow(clippy::wildcard_enum_match_arm)]` with some comments.
- Otherwise, spell out every single enum variant instead of the catch-all arm.

In some cases where this becomes very verbose, I bring the enum variants all into the function scope `use DType::*;`. Usually this is not great, but I try to make sure that this only happens in a function scope where the variants are unambiguous.

### DIFF
--- a/vortex-array/src/arrays/chunked/decode.rs
+++ b/vortex-array/src/arrays/chunked/decode.rs
@@ -21,6 +21,7 @@ impl CanonicalVTable<ChunkedVTable> for ChunkedVTable {
         if array.nchunks() == 1 {
             return array.chunks()[0].to_canonical();
         }
+
         match array.dtype() {
             DType::Struct(struct_dtype, _) => {
                 let struct_array = swizzle_struct_chunks(
@@ -30,12 +31,18 @@ impl CanonicalVTable<ChunkedVTable> for ChunkedVTable {
                 )?;
                 Ok(Canonical::Struct(struct_array))
             }
-            DType::List(elem, _) => Ok(Canonical::List(pack_lists(
+            DType::List(elem_dtype, _) => Ok(Canonical::List(pack_lists(
                 array.chunks(),
                 Validity::copy_from_array(array.as_ref())?,
-                elem,
+                elem_dtype,
             )?)),
-            _ => {
+            DType::Null
+            | DType::Bool(_)
+            | DType::Primitive(..)
+            | DType::Decimal(..)
+            | DType::Utf8(_)
+            | DType::Binary(_)
+            | DType::Extension(_) => {
                 let mut builder = builder_with_capacity(array.dtype(), array.len());
                 array.append_to_builder(builder.as_mut())?;
                 builder.finish().to_canonical()
--- a/vortex-array/src/arrays/primitive/compute/take/mod.rs
+++ b/vortex-array/src/arrays/primitive/compute/take/mod.rs
@@ -73,17 +73,17 @@ impl TakeImpl for TakeKernelScalar {
 
 impl TakeKernel for PrimitiveVTable {
     fn take(&self, array: &PrimitiveArray, indices: &dyn Array) -> VortexResult<ArrayRef> {
-        let unsigned_indices = match indices.dtype() {
-            DType::Primitive(p, n) => {
-                if p.is_unsigned_int() {
-                    indices.to_primitive()?
-                } else {
-                    // This will fail if all values cannot be converted to unsigned
-                    cast(indices, &DType::Primitive(p.to_unsigned(), *n))?.to_primitive()?
-                }
-            }
-            _ => vortex_bail!("Invalid indices dtype: {}", indices.dtype()),
+        let DType::Primitive(ptype, null) = indices.dtype() else {
+            vortex_bail!("Invalid indices dtype: {}", indices.dtype())
         };
+
+        let unsigned_indices = if ptype.is_unsigned_int() {
+            indices.to_primitive()?
+        } else {
+            // This will fail if all values cannot be converted to unsigned
+            cast(indices, &DType::Primitive(ptype.to_unsigned(), *null))?.to_primitive()?
+        };
+
         let validity = array.validity().take(unsigned_indices.as_ref())?;
         // Delegate to the best kernel based on the target CPU
         PRIMITIVE_TAKE_KERNEL.take(array, &unsigned_indices, validity)
--- a/vortex-array/src/arrays/varbin/compute/min_max.rs
+++ b/vortex-array/src/arrays/varbin/compute/min_max.rs
@@ -20,7 +20,7 @@ impl MinMaxKernel for VarBinVTable {
 register_kernel!(MinMaxKernelAdapter(VarBinVTable).lift());
 
 /// Compute the min and max of VarBin like array.
-pub fn compute_min_max<T: ArrayAccessor<[u8]>>(
+pub(crate) fn compute_min_max<T: ArrayAccessor<[u8]>>(
     array: &T,
     dtype: &DType,
 ) -> VortexResult<Option<MinMaxResult>> {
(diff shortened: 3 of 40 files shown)


vortex-array/src/arrow/compute/to_arrow/varbin.rs
```diff
@@ -21,20 +21,26 @@ impl ToArrowKernel for VarBinVTable {
...
+                    // TODO(connor):                              vvv Is this correct?
+                    U8 | U16 | U32 | I8 | I16 | I32 | F16 | F32 | F64 => to_arrow::<i32>(array),
                 },
-                DType::Utf8(_) => match offsets_ptype {
-                    PType::I64 | PType::U64 => to_arrow::<i64>(array),
-                    _ => to_arrow::<i32>(array),
+                Utf8(_) => match offsets_ptype {
+                    I64 | U64 => to_arrow::<i64>(array),
+                    // TODO(connor):                              vvv Is this correct?
+                    U8 | U16 | U32 | I8 | I16 | I32 | F16 | F32 | F64 => to_arrow::<i32>(array),
```
connortsui20: I have no idea if or why this is correct. Could someone let me know and I can write a comment as to why this is the case?

vortex-array/src/arrays/varbin/compute/min_max.rs
```diff
@@ -44,15 +47,21 @@ pub fn compute_min_max<T: ArrayAccessor<[u8]>>(
...
-        DType::Binary(_) => Scalar::new(dtype.clone(), value.into()),
-        DType::Utf8(_) => {
+        Binary(_) => Scalar::new(dtype.clone(), value.into()),
+        Utf8(_) => {
+            // TODO(connor): This is absolutely not safe!!! We cannot trust the input to a public
+            // function (via `compute_min_max`) to be correct.
             // Safety:
             // We trust the array's dtype here
             let value = unsafe { str::from_utf8_unchecked(value) };
             Scalar::new(dtype.clone(), value.into())
```
connortsui20: Also this is extremely concerning

vortex-array/src/arrays/primitive/compute/take/mod.rs
```diff
@@ -73,17 +73,17 @@ impl TakeImpl for TakeKernelScalar {
...
     fn take(&self, array: &PrimitiveArray, indices: &dyn Array) -> VortexResult<ArrayRef> {
-        let unsigned_indices = match indices.dtype() {
-            DType::Primitive(p, n) => {
-                if p.is_unsigned_int() {
-                    indices.to_primitive()?
-                } else {
-                    // This will fail if all values cannot be converted to unsigned
-                    cast(indices, &DType::Primitive(p.to_unsigned(), *n))?.to_primitive()?
-                }
+        let unsigned_indices = if let DType::Primitive(ptype, null) = indices.dtype() {
```
lwwmanning: nit: this can be `let ... else` and then not be nested

vortex-array/src/arrow/compute/to_arrow/varbin.rs
connortsui20: (Turns out this is incorrect)

vortex-array/src/arrays/varbin/compute/min_max.rs
```diff
@@ -44,15 +44,19 @@ pub fn compute_min_max<T: ArrayAccessor<[u8]>>(
...
+        Utf8(_) => {
+            // SAFETY: We only call `compute_min_max` within `varbin/`, in which we always validate
+            // the arrays, and we always pass `array.dtype()` in as the `dtype` argument.
             let value = unsafe { str::from_utf8_unchecked(value) };
             Scalar::new(dtype.clone(), value.into())
         }
-        _ => unreachable!(),
+        Null | Bool(_) | Primitive(..) | Decimal(..) | List(..) | Struct(..) | Extension(_) => {
+            unreachable!()
+        }
```
joseph-isaacs: This appear a lot which is pretty annoying to both read and write

vortex-array/src/arrays/chunked/decode.rs
```diff
@@ -15,27 +15,30 @@ use crate::{Array as _, ArrayRef, Canonical, IntoArray, ToCanonical};
 
 impl CanonicalVTable<ChunkedVTable> for ChunkedVTable {
     fn canonicalize(array: &ChunkedArray) -> VortexResult<Canonical> {
+        use DType::*;
```
onursatici: I think using a wildcard import for enum variants might be dangerous if one of the variants is removed. If that removed variant has no variables, then the arm with that removed variant becomes a variable declaration with the same name, everything compiles and that match arm matches everything from then on. something like:

```rust
enum Enum {
  // A, 
  B, 
}

use Enum::*;
fn test(a: Enum) {
   match a {
    A => {},   // now this becomes the variable A
    B => {},
  }
}
```

onursatici: so I would strongly prefer explicit variant names, or importing all variants explicitly instead of the wildcard

connortsui20: I think this makes sense, but also if this does happen we would get compile errors as well as the same `wildcard_enum_match_arm` lint

vortex-array/src/arrays/varbin/compute/min_max.rs
connortsui20: I do agree, though I think the point of this PR is that because enums like `DType` are unstable, we want to know immediately where things need to change if more types are added. For example, if we added a `DType::Utf16` type (which we shouldn't do, but for the sake of example), then naming all of these variants will tell us that we need to revisit this `make_scalar` function and update it to handle interpreting UTF-16 bytes.

Maybe this is too heavy-handed? Not sure

vortex-array/src/arrays/chunked/decode.rs
connortsui20: Ah sorry, to be more clear if we include `#[deny(unreachable_patterns)]` as well, things will immediately fail in this case.

Playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6ca5ef09cc8c16e9cc652c4147df3860

vortex-array/src/arrays/varbin/compute/min_max.rs
joseph-isaacs: I think so

vortex-array/src/compute/conformance/consistency.rs
```diff
@@ -656,11 +657,13 @@ fn test_comparison_symmetry_consistency(array: &dyn Array) {
...
+    use DType::*;
+
+    // Skip non-comparable types.
     match array.dtype() {
-        DType::Null | DType::Extension(_) => return,
-        DType::Struct(..) | DType::List(..) => return,
-        _ => {}
+        Null | Extension(_) => return,
+        Struct(..) | List(..) => return,
+        Bool(_) | Primitive(..) | Decimal(..) | Utf8(_) | Binary(_) => {}
```
joseph-isaacs: ?

PR #4328, opened 2025-08-22
chore[bench]: have `--skip-generate` option and only create iterators for tpch when needed

joseph-isaacs:


bench-vortex/src/benchmark_driver.rs
```diff
@@ -37,13 +37,16 @@ pub struct DriverConfig {
     pub show_metrics: bool,
     pub hide_progress_bar: bool,
     pub track_memory: bool,
+    pub skip_generate: bool,
 }
 
 /// Run a benchmark using the provided implementation and configuration
 pub fn run_benchmark<B: Benchmark>(benchmark: B, config: DriverConfig) -> Result<()> {
     // Generate data for each target (idempotent)
```
gatesn, line 45: So was it _not_ idempotent before?

joseph-isaacs: it was, the file was not rewritten, but lots of extra work was gone and then not done

joseph-isaacs: this just allow a future escape hatch

gatesn (approved):

PR #4331, opened 2025-08-22
fix: Java builds with GLIBC 2.31+ compatibility

### LINKED ISSUE #3744: Build vortex-jni on CentOS 7 for lower glibc
We currently build on ubuntu24 which brings in glibc 2.38.

That is newer than the glibc version available on most CentOS based systems (including Amazon Linux) so you can end up with situations like this

Image

We should either

- Build statically linked on CI, i.e. x86_64-unknown-linux-musl
- Build on CI with a sufficiently old glibc to get better compatibility (e.g. via docker image)

a10y:
Fixes #3744

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -136,8 +136,6 @@ jobs:
           distribution: "corretto"
           java-version: "17"
       - uses: ./.github/actions/setup-rust
-        with:
-          targets: ${{ matrix.target.target }}
       - run: cargo build --release --package vortex-jni
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -128,17 +128,8 @@ jobs:
 #          verbose: true
 #          skip-existing: true  # Idempotent releases
 
-  prepare-java:
-    runs-on: ${{ matrix.target.runs-on }}
-    container: ${{ matrix.target.container }}
-    timeout-minutes: 120
-    strategy:
-      fail-fast: false
-      matrix:
-        target:
-          - { os: macos, runs-on: "macos-latest", target: aarch64-apple-darwin, container: {} }
-          - { os: ubuntu, runs-on: "ubuntu-24.04-arm", container: { image: "ubuntu:20.04" }, target: aarch64-unknown-linux-gnu }
-          - { os: ubuntu, runs-on: "ubuntu-latest", container: { image: "ubuntu:20.04" }, target: x86_64-unknown-linux-gnu }
+  prepare-java-macos:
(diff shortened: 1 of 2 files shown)


a10y:
Alright, this configures the Java build to execute in an ubuntu 20.04 container. Anything older than this and the CI actions all failed. While 20.04 is technically EOL, it does provide us compatibility with GLIBC back to 2.31

image

FWIW Lance uses Debian 10 in CI which EOL'ed over a year ago.

Current Amazon Linux AMIs launch with 2.34:

image

After this change

* Amazon Linux 2022 and onward is supported
* Ubuntu 20.04 onward will be supported

.github/workflows/publish.yml
```diff
@@ -125,20 +125,47 @@ jobs:
...
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-java@v5
+        with:
+          distribution: "corretto"
+          java-version: "17"
+      - uses: ./.github/actions/setup-rust
+        with:
+          targets: ${{ matrix.target.target }}
```
lwwmanning: Does this work...? We're not in a matrix afaict

a10y: interestingly this didn't fail https://github.com/vortex-data/vortex/actions/runs/17176981022/job/48734213185?pr=4331

But it's definitely cruft, removing

a10y: done

.github/workflows/publish.yml
```diff
@@ -125,20 +125,47 @@ jobs:
...
+        with:
+          name: "libvortex_jni_aarch64-apple-darwin.zip"
+          path: "target/release/libvortex_jni.dylib"
+          retention-days: 1
+          if-no-files-found: error
+
+  prepare-java-linux:
     runs-on: ${{ matrix.target.runs-on }}
+    container:
+      image: "ubuntu:20.04"
```
a10y, line 150: GH discontinued support for ubuntu 20.04 _runners_ a few months ago sadly, but the container version is pretty much seamless drop-in, save for the extra apt-install step below

docs/api/java/index.rst
```diff
@@ -23,6 +23,22 @@ Installation
 
 The Java API can be included in your project using Gradle or Maven. Please refer to the main documentation for detailed installation instructions.
 
+
+Compatibility
+-------------
+
+The Java bindings are supported on the following architectures:
+
+* x86_64 Linux
```
lwwmanning, line 32: Looks like we prepare/build arm Linux as well?

a10y: See line 33

lwwmanning (approved):

PR #4332, opened 2025-08-22
Clean up match statements and misc

connortsui20:
Cleans up some code in the following crates:

- `vortex-array`
- `vortex-dtype`
- `vortex-scalar`
- `vortex-mask`

These are some things I identified that could be cleaned up when trying to add the clippy lint `wildcard_enum_match_arm` lint. That PR is closed but this is a new one.

This is the ghost of #4317

### DIFF
--- a/vortex-array/src/arrays/varbin/compute/min_max.rs
+++ b/vortex-array/src/arrays/varbin/compute/min_max.rs
@@ -3,7 +3,7 @@
 
 use itertools::Itertools;
 use vortex_dtype::DType;
-use vortex_error::VortexResult;
+use vortex_error::{VortexResult, vortex_panic};
 use vortex_scalar::Scalar;
 
 use crate::accessor::ArrayAccessor;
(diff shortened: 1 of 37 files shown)


vortex-array/src/arrays/varbin/compute/min_max.rs
```diff
@@ -47,12 +47,20 @@ fn make_scalar(dtype: &DType, value: &[u8]) -> Scalar {
...
             Scalar::new(dtype.clone(), value.into())
         }
-        _ => unreachable!(),
+        DType::Null
+        | DType::Bool(_)
+        | DType::Primitive(..)
+        | DType::Decimal(..)
+        | DType::List(..)
+        | DType::Struct(..)
+        | DType::Extension(_) => {
```
robert3005: since this is pub(crate) this should be just `_ =>`

vortex-scalar/src/scalar_value.rs
```diff
@@ -64,8 +63,8 @@ pub(crate) enum InnerScalarValue {
     Bool(bool),
     Primitive(PValue),
     Decimal(DecimalValue),
-    Buffer(Arc<ByteBuffer>),
-    BufferString(Arc<BufferString>),
+    Buffer(Arc<vortex_buffer::ByteBuffer>),
+    BufferString(Arc<vortex_buffer::BufferString>),
```
robert3005: import these?

robert3005 (approved):

PR #4333, opened 2025-08-22
feat[ffi]: add non-panicking accessors to ffi

lwwmanning:
## Summary

Transforms the Vortex FFI layer from a panic-prone interface to a robust, safe-by-default API by replacing all panicking accessor functions with error-handling variants.

### 🛡️ **Core Safety Transformation**
- **Replaced panicking accessors**: All `vx_array_get_*` functions now include `vx_error **error_out` parameter
- **Comprehensive error handling**: Three-layer validation (bounds → null → type checking) 
- **Zero segfaults**: Eliminates panic-induced crashes in FFI boundaries
- **Compiler-enforced migration**: Additional error parameter ensures existing code won't compile without updates
- **Consistent pattern**: UTF-8, binary, and slice functions follow same error handling approach

### 📖 **Enhanced Documentation**
- **Comprehensive C header docs** with safety requirements and usage examples
- **Clear migration guidance** showing before/after patterns
- **Detailed f16 handling** explaining IEEE 754 bit representation

### 🧪 **Robust Testing**
- **Complete test coverage**: Success scenarios, error conditions, edge cases
- **Memory safety validation**: Prevents leaks and use-after-free
- **Concurrent access testing**: Thread-safe error handling

### ✅ **Security**
- Eliminates all identified segfault vulnerabilities
- Prevents buffer overflows through bounds checking
- Safe null value and type mismatch handling

### ✅ **Usability** 
- Clear error messages for debugging
- Predictable behavior (returns 0 on error)
- Excellent documentation with practical examples

### ✅ **Performance**
- Minimal overhead from error checking
- Macro-generated consistent implementation
- No ABI bloat from duplicate functions

## Breaking Changes

All primitive accessor functions now require an additional `vx_error **error_out` parameter:

```c
// Before
int32_t value = vx_array_get_i32(array, index);

// After  
vx_error *error = NULL;
int32_t value = vx_array_get_i32(array, index, &error);
if (error != NULL) {
    // Handle error
    vx_error_free(error);
}
```

## Files Changed

- **`vortex-ffi/src/array.rs`**: Core implementation with safe accessors
- **`vortex-ffi/cinclude/vortex.h`**: Auto-generated header with new signatures
- **`vortex-ffi/examples/hello-vortex.c`**: Updated example demonstrating proper usage
- **Tests**: Comprehensive test coverage for all error scenarios

## Result

**Production-ready FFI interface** with comprehensive error handling, memory safety guarantees, and extensive test coverage. The API transformation ensures that unsafe usage patterns are caught at compile-time, preventing runtime crashes.


vortex-ffi/src/array.rs
```diff
@@ -62,17 +62,69 @@ pub unsafe extern "C-unwind" fn vx_array_get_field(
...
+///     // Use slice - contains elements 10-19
+///     vx_array_free(slice);
+/// }
+/// ```
 #[unsafe(no_mangle)]
 pub unsafe extern "C-unwind" fn vx_array_slice(
     array: *const vx_array,
     start: u32,
     stop: u32,
-    // TODO(aduffy): deprecate this from the FFI API.
```
lwwmanning, line 70: lol, I just saw this comment @a10y -- I started this PR because the C example will just segfault on panic, which is not very easy to debug 😬

vortex-ffi/src/lib.rs
```diff
@@ -80,8 +119,38 @@ pub(crate) unsafe fn to_string_vec(ptr: *const *const c_char, len: c_int) -> Vec
...
+/// vx_session* session2 = vx_session_new();  // Reuses same runtime
+///
+/// vx_session_free(session1);
+/// vx_session_free(session2);
+///
+/// // Explicit cleanup:
+/// vx_try_shutdown_runtime();  // Only succeeds if no sessions active
+/// ```
 #[unsafe(no_mangle)]
 pub extern "C" fn vx_try_shutdown_runtime() {
```
gatesn, line 155: I can't imagine we actually want to commit this to as a stable API...?

We should make it clear (in another PR) the state of this API. Honestly, I'd rather just kill it for now and publish language-specific bindings that are allowed to make lock-step breaking changes to their own C APIs.

lwwmanning: oh yeah, this is gross. it's just that not cleaning up the runtime can lead to segfaults on exit when e.g., used with mimalloc. we could also kill this function and have the runtime be cleaned up when the last VortexSession is dropped

robert3005: Not that I disagree but this is how the api would look like in C anyway? You're sahying we shouldn't have C api and only C++?

lwwmanning:
we should just kill FFI, probably

PR #4334, opened 2025-08-22
Add a developer overview of Vortex

connortsui20:
I don't really know what to call this. This is not really a README, but it is also not really a contributors guide.

Regardless, would love to get feedback on this!

### DIFF
--- a/vortex-array/dev-docs.md
+++ b/vortex-array/dev-docs.md
@@ -1,0 +1,69 @@
+# High-level overview of `Array`s in `vortex-array`
+
+## `Array`
+
+An `Array` is a type-erased, in-memory representation of a (possibly compressed) logical array of data with known length. Rust types (usually a normal Rust `struct`) implement `Array` plus several other required traits.
+
+`Array`s are recursively compressed, which means that we can apply compression multiple times to data. This also means that `Array`s can be recursively **de**compressed, which allows for partial decompression to reduce the amount of computation necessary.
+
+All data is stored in either a `Buffer<T>` (which is a wrapper around an aligned buffer that takes up space in `O(n)`) or an associated `Metadata` (which takes up space in `O(1)`).
+
+## `DType`
+
+`DType` is an enumeration of the different **logical** types that can be represented in a Vortex array. This is different from physical types, which we call **encodings**, that represent the actual layout of data (compressed or uncompressed). Each logical type has at least one physical array encoding that lives in the `vortex-array` crate (or more formally, the set of physical types is surjective into the set of logical types).
+
+Note that a `DType` represents the logical type of the elements in the `Array`s, **not** the logical type of the `Array` itself.
+
+## Encodings
+
+We can think of an encoding as a “class” (in the OOP sense). `Array`s are simply instances of an encoding. And similar to a class, each `Array` implementation also comes with a "virtual" function table `VTable`, which we will explain later.
+
+### Canonical Encodings
+
+For every unique `DType`, there exists a **canonical encoding** of that logical type. We can think of these canonical encodings as the "default" physical type / encoding of a given logical type. 
+
+All `Array`s can be decompressed into canonical encodings based on the logical `DType`. However, note that canonical encodings **do not** have to be fully decompressed, where we consider "fully decompressed" to mean that the `Array` is zero-copyable to Apache Arrow.
+
+For example, a validity child `Array` (the null map) does not have to be fully decompressed in a canonical encoding (_note that the Vortex maintainers are considering requiring it fully decompressed_). Similarly, the `struct` canonical encoding does not require its field `Array`s (children) to be fully decompressed (canonicalized). In other words, we only care about canonical access to the **components** of the `Array`.
+
+All of the canonical encodings are located in the `vortex-array/src/arrays` module. There are a few other **non-canonical encodings** that live in there as well for convenience.
+
+### Non-Canonical Encodings
+
+Non-canonical encodings are simply physical data layouts / encodings that are _not_ the default representation of a logical `DType`.
+
+There are a few non-canonical encodings in the `vortex-array/src/arrays` module (notably `constant`, `datetime`, `chunked`). The remaining non-canonical encodings live in the `encodings/` subcrates. These non-canonical encodings represent faster or more specialized encodings / compression strategies like dictionary and run-end encoding, as well as some more state-of-the-art compression schemes like `ALP` for floating-point numbers and `FSST` for strings.
+
+The non-canonical (faster / advanced) encodings in the `encodings` directory will generally form a tree structure of `Array`s (some of which may have only one root node), where the leaf nodes of this tree are canonical encodings. There are a few exceptions to this (`fastlanes` has `bitpacking`).
+
+All encodings of arrays implement `VTable`, and all have several **compute kernels** for specialized computation (which we will also talk about later).
+
+## `VTable`
+
+Every `Array` implementation must implement `VTable`, which is conceptually similar to virtual tables used for dynamic dispatch in Rust or C++.
+
+Implementations of `Array` (via help from a zero-sized marker type generated with macros) implement the `VTable` trait, which is a trait that combines several other traits for organization. Some of the behavior includes canonicalizing the array (`CanonicalVTable`) or serializing/deserializing the array (`SerdeVTable`).
+
+### Decompression
+
+Decompression happens via calls to the `canonicalize` method, which `Array`s implement via the trait `CanonicalVTable`. Recall that `Array` are _recursively compressed_ data: we can think of the `canonicalize` method as decompression of the topmost "compression layer" of an `Array` into a canonical encoding. Most of the canonical encodings have relatively straightforward decompression implementations. Non-canonical encodings utilize specialized compute kernels to speed up decompression (which we talk about in the next section).
+
+In order to fully decompress an `Array`, we use the `Array::append_to_builder` method. It takes a `&mut dyn ArrayBuilder` of the same `DType`, which further allows the concatenation of multiple arrays.
+
+## Compute Kernels
+
+All encodings implement several `Kernel`s (all of which are traits from `vortex_array::compute`) such as `TakeKernel`, `FilterKernel`, or `FillNullKernel` which return new `Array`s, or `IsSortedKernel` and `MinMaxKernel` which return other things. These kernels are registered with the `inventory` crate via `register_kernel!`.
+
+Generally speaking, any operation over an array of data that takes work in `O(n)` could be a candidate compute kernel. The kernels that turn naive `O(n)` operations into `O(1)` are generally the most useful in speeding up operations.
+
+As an example for how these compute kernels can speed up decompression, suppose we have a `ConstantArray` of `4` and we want to add `2` to each value in the array (so that the resultant array is a array of just `6`s). Instead of decompressing the `ConstantArray` into a full array and adding the number to each element (`O(n)`), we can instead just add `2` to the constant `4` and return a `ConstantArray` of `6`.
+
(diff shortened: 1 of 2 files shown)


vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## `Array`
+
+An `Array` is a type-erased, in-memory representation of a (possibly compressed) logical array of data with known length. Rust types (usually a normal `struct`) implement `Array` as well as several other required traits.
+
+`Array`s are recursively compressed, which means that we can apply compression multiple times to data. This also means that `Array`s can be recursively **de**compressed, which allows for partial decompression to reduce the amount of computation necessary.
+
+## `DType`
+
+`DType` is an enumeration of the different **logical** types that can be represented in a Vortex array. This is different from physical types, which represent the actual layout of data (compressed or uncompressed). The set of physical types/formats (or data layout) is surjective into the set of logical types.
```
danking: I think our term of art for "physical types/formats (or data layout)" is "array encoding(s)".

danking: Agreed, it is surjective though I think I find the opposite direction more illustrative:
> Each logical type has at least one encoding in the vortex-array crate.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Note that a `DType` represents the logical type of the elements in the `Array`s, **not** the logical type of the `Array` itself.
+
+## Encodings
+
+We can think of an encoding as a “class” (in the OOP sense). `Array`s are simply instances of an encoding. And similar to a class, each `Array` implementation also comes with a "virtual" function table `VTable`, which we will explain later.
+
+## Canonical Encodings
+
+For every `DType`, there exists a **canonicalized encoding** of that logical type. We can think of these canonicalized encodings as the "default" physical type / data layout of a given logical type. The set of canonicalized encodings is bijective into the set of logical types under `DType`.
```
danking:
```suggestion
For every `DType`, there exists a **canonical encoding** of that logical type. We can think of these canonical encodings as the "default" physical type / data layout of a given logical type. The set of canonical encodings is bijective into the set of logical types under `DType`.
```

danking: It doesn't form a bijection because both the `UTF8` and `Binary` DTypes map to `VarBinView`. There's a surjective function from DType to Canonical though.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+All `Array`s can be decompressed into canonicalized encodings based on the logical `DType`. However, note that canonicalized encodings **do not** have to be fully decompressed, where we consider "fully decompressed" to mean that the `Array` is zero-copyable to Apache Arrow.
```
danking:
```suggestion
All `Array`s can be decompressed into canonical encodings based on the logical `DType`. However, note that canonical encodings **do not** have to be fully decompressed, where we consider "fully decompressed" to mean that the `Array` is zero-copyable to Apache Arrow.
```
I'll stop commenting but prefer "canonical" everywhere to "canonicalized".

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+For example, a validity child `Array` does not have to be fully decompressed in a canonical encoding (note that we are considering making it fully decompressed). Similarly, the `struct` canonical encoding does not require the its child field `Array`s to be fully decompressed. In other words, we only care about canonical access to the **components** of the `Array`.
```
danking: I might merge this with the last paragraph for clarity of reading.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+All of the canonical encodings are located in the `vortex-array/src/arrays` module. There are a few other **non-canonical encodings** that live in there as well for convenience.
+
+## Non-Canonical Encodings
+
+Non-canonical encodings are simply physical data layouts / physical types of `Array`s that are _not_ the default representation of a logical `DType`.
+
+There are a few non-canonical encodings in the `vortex-array/src/arrays` module (notably `constant`, `datetime`, `chunked`). The remaining non-canonical encodings live in the `encodings/` subcrates. These non-canonical encodings represent faster or more specialized encodings / compression strategies like dictionary and run-end encoding, as well as some more state-of-the-art compression schemes like `ALP` for floating-point numbers and `FSST` for strings.
+
+The non-canonical (faster/advanced) encodings in the `encodings` directory will generally form a tree structure of `Array`s (some of which may have only one root node), where the leaf nodes of this tree are canonical encodings. There are a few exceptions to this (`fastlanes` has `bitpacking`).
```
danking: I'm not sure where it belongs but it might help to mention that eventually all data is stored in either a Buffer (O(N) sized data) or a Metadata (O(1) sized data).

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## `VTable`
+
+Every `Array` implementation must implement `VTable`, which is conceptually similar to virtual tables used for dynamic dispatch in Rust or C++.
+
+Implementations of `Array` (via a zero-sized marker type generated with macros) implement the `VTable` trait, which is a trait that combines several other traits for organization. Some of the behavior includes canonicalizing the array (`CanonicalVTable`) or serializing/deserializing the array (`SerdeVTable`).
+
+## Decompression
+
+Decompression happens via the `canonicalize` method, which `Array`s implement via `CanonicalVTable`. Most of the canonical encodings have straight-forward decompression implementations. Non-canonical encodings utilize specialized compute kernels to speed up decompression.
```
danking: Hmm. Canonicalize is not always decompression, it's just canonicalization of one layer. I think it is best not to conflate these two concepts. 

I'd say something like:

> The canonicalize method converts the first-level of the array into one of the Canonical encodings. It might perform O(N) decompression work, for example, bitpacking must expand its buffer into a buffer of full-width integers.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+One way to fully decompress an `Array` is to use the `to_arrow` function in the `vortex-array/src/arrow/compute/to_arrow/mod.rs` module, which will output a (decompressed) Apache Arrow array.
```
danking: I wouldn't include this line. I think you can have a separate section about Arrow, but the relationship between decompression, Arrow arrays, Canonical Vortex arrays, and Vortex arrays, in general, is subtle. Best to keep it simple here.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The "built-in" way to arbitrarily decompress an `Array` is create an `ArrayBuilder` (which pre-allocates a recursively canonical output array), and then call `Array::append_to_builder`".
```
danking: I'd say: "The `append_to_builder` method fully decompresses an array. It accepts a mut ref to an ArrayBuilder of the same dtype which further allows the concatenation of multiple arrays."

danking: In particular, I'm not sure what "arbitrarily" means in this context.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Compute Kernels
+
+All encodings implement several `Kernel`s (all of which are traits from `vortex_array::compute`) such as `TakeKernel`, `FilterKernel`, or `FillNullKernel` which return new `Array`s, or `IsSortedKernel` and `MinMaxKernel` which return other things. These kernels are registered with the `inventory` crate via `register_kernel!`.
```
danking: I might include something like "In principle, any operation over an array which is O(N) is a candidate Kernel." In particular, we don't plan to implement joins, but we're open to more or less any linear compute over an Array, including aggregations.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+As an example for how these compute kernels can speed up decompression, suppose we want to filter mask elements of a `DictArray`. Instead of decompressing the dictionary and values (what we would call "canonicalizing" the data) into a canonicalized array and then masking out the elements we want to drop, we can instead filter just the codes and keep the values. And even better, the codes might be bitpacked, so the bitpacked codes (reliant on the `fastlanes` encoding) will also use a specialized compute kernel for filtering.
```
danking:
```suggestion
As an example for how these compute kernels can speed up decompression, suppose we want to filter mask elements of a `DictArray`. Instead of decompressing the dictionary and values (what we would call "canonicalizing" the data) into a canonical array and then masking out the elements we want to drop, we can instead filter just the codes and keep the values. And even better, the codes might be bitpacked, so the bitpacked codes (reliant on the `fastlanes` encoding) will also use a specialized compute kernel for filtering.
```

danking (approved):
This seems great, I think there's one factual error (the bit about bijection). My other comments are optional changes that I find elucidating.

vortex-array/README.md
```diff
@@ -0,0 +1,67 @@
+# High-level overview of `Array`s in `vortex-array`
```
danking: I think you cannot call this README.md because, in Cargo.toml, we copy the README from the root. This ensures all published packages have the same crates.io summary page.

CONTRIBUTING.md ? DEV_README.md ?

connortsui20: I think that there definitely should be some sort of DESIGN.md or similar, and what I have here (after I make corrections) should be a part of that. However there definitely should be MUCH MORE considering this doesn't even go through crate/module structure nor the purpose of each thing (the why).

And to be honest there are still some things that I still have unanswered but I'm sort of leaving that for me to explore rather than ask outright right now.

vortex-array/README.md
connortsui20: I put it at the top under `Array`

vortex-array/README.md
connortsui20: That makes sense. I think I'll leave out the O(n) thing since some work could be instant or in the future it could take more than linear time

vortex-array/README.md
connortsui20: Would it be accurate to say this?

> Generally speaking, any operation over an array of data that takes work in `O(n)` could be a candidate compute kernel. The kernels that turn naive `O(n)` operations into `O(1)` are generally the most useful in speeding up operations.

vortex-array/README.md
connortsui20: I think it might be best to leave it here to be honest, as it wasn't immediately obvious to me that this was true until talking to Nick. Do you think the information density is too high here?

vortex-array/dev-docs.md
```diff
@@ -0,0 +1,69 @@
+# High-level overview of `Array`s in `vortex-array`
```
robert3005, line 1: Can we also add a link to this document from the repos readme? Otherwise it's not very discoverable

vortex-array/README.md
danking: Sure!

vortex-array/README.md
danking: I prefer concatenation because the first paragraph makes a not entirely obvious claim, “canonical encoding do not have to be fully decompressed”, and the second paragraph is the evidence. That said it’s quite a minor thing and I trust your judgement as the one most recently encountering these ideas!

danking (approved):

a10y:
This is great!

A few other ideas that I've seen for doing these types of architecture documentation:

- Rust Language Book style [mdbook](https://rust-lang.github.io/mdBook/) setup
- Having a book embedded in Rust doc comments, either in the module root or in a special `book` module, and hosted on docs.rs
- Just having an `architecture` folder in the root with a bunch of markdown files

The first is probably a lot of effort, the second is nice because it's easy to maintain and the doctests ensure it's always up to date, but I kind of like the third. It's also nice because it lends itself well to a type of RFC process where people can discuss proposals in PR comments.

connortsui20:
So I'm going to close this and reopen a new PR that introduces an mdbook (somewhere TBD)

PR #4335, opened 2025-08-22
fix(java): Spark writer properly handles string and array data

a10y:
We need to use setSafe() as the initial allocation might not be large enough to hold all of the string data in the InternalRow, and a realloc might be required.

Also adds support for Spark `ArrayType`, and augments the existing roundtrip writer unit test to include an `array<string>`

### DIFF
--- a/java/vortex-spark/src/main/java/dev/vortex/spark/write/VortexDataWriter.java
+++ b/java/vortex-spark/src/main/java/dev/vortex/spark/write/VortexDataWriter.java
@@ -8,6 +8,7 @@ import dev.vortex.relocated.org.apache.arrow.memory.BufferAllocator;
 import dev.vortex.relocated.org.apache.arrow.memory.RootAllocator;
 import dev.vortex.relocated.org.apache.arrow.vector.*;
 import dev.vortex.relocated.org.apache.arrow.vector.VectorSchemaRoot;
+import dev.vortex.relocated.org.apache.arrow.vector.complex.ListVector;
 import dev.vortex.relocated.org.apache.arrow.vector.ipc.ArrowStreamWriter;
 import dev.vortex.spark.SparkTypes;
 import java.io.ByteArrayOutputStream;
@@ -20,6 +21,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.apache.spark.sql.catalyst.InternalRow;
+import org.apache.spark.sql.catalyst.expressions.SpecializedGetters;
+import org.apache.spark.sql.catalyst.util.ArrayData;
 import org.apache.spark.sql.connector.write.DataWriter;
 import org.apache.spark.sql.connector.write.WriterCommitMessage;
 import org.apache.spark.sql.types.*;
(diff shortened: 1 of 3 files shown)


java/vortex-spark/src/main/java/dev/vortex/spark/write/VortexDataWriter.java
```diff
@@ -203,12 +207,12 @@ private void populateVector(FieldVector vector, DataType dataType, InternalRow r
         } else if (dataType instanceof StringType) {
             UTF8String str = row.getUTF8String(fieldIndex);
             if (str != null) {
-                ((VarCharVector) vector).set(rowIndex, str.getBytes());
+                ((VarCharVector) vector).setSafe(rowIndex, str.getBytes());
```
a10y, line 210: setSafe will allocate a larger buffer if we try to insert more strings than the initial allocation (default is 32KiB)

lwwmanning (approved):

java/vortex-spark/src/main/java/dev/vortex/spark/write/VortexDataWriter.java
```diff
@@ -218,9 +222,19 @@ private void populateVector(FieldVector vector, DataType dataType, InternalRow r
...
                 ((DecimalVector) vector).set(rowIndex, decimal);
             }
+        } else if (dataType instanceof ArrayType) {
+            ArrayType arrayType = (ArrayType) dataType;
+            ArrayData data = row.getArray(fieldIndex);
+            ListVector listVector = ((ListVector) vector);
+            int writtenElements = listVector.getElementEndIndex(listVector.getLastSet());
+            listVector.startNewValue(rowIndex);
+            for (int i = 0; i < data.numElements(); i++) {
+                populateVector(listVector.getDataVector(), arrayType.elementType(), data, i, writtenElements + i);
```
a10y, line 232: you need to be a tad cute about how you populate the inner elements vector that the ListVector wraps

robert3005:
There surely is an api in spark that goes from internalrows to arrow arrays

robert3005:
Aaaand you'd be wrong thinking that there's one. All arrow conversions is really only for python execution

robert3005:
wait no, there's a thing called `ArrowWriter` in spark that does what you did here

a10y:
Yea I cribbed that for some of my implementation here. We need to use the shaded classes, otherwise it fails at runtime since Iceberg injects its own Arrow JARs in

robert3005:
:sadpanda: OKAY I hate shadow jars

a10y:
Sorry to be more specific, the order of things is

- `vortex-jni` exports a shaded copy of `org.apache.arrow`, and all methods that work on Arrow return shaded objects
- `vortex-spark` depends on `vortex-jni`, and so it consumes the shaded objects and must work with them
- `iceberg-vortex` will depend on `vortex-jni`. All iceberg first-party sources **must** internally shade Arrow since iceberg injects its own Arrow JARs at runtime

It's extremely annoying

robert3005:
I fail to understand why they don't resolve application deps with spark deps but that's a problem for a different time

PR #4338, opened 2025-08-22
feat: vortex-python/check.sh

danking:


robert3005 (approved):

robert3005:
We probably want an xtask but I can make that conversion

PR #4346, opened 2025-08-23
Re-open DuckDB context between iterations

gatesn:
There was a suspicious lack of pread in the profiles..


robert3005 (approved):

lwwmanning:
This does indeed slow duckdb down considerably. Confirms the duckdb runs were "cheating" before

PR #4347, opened 2025-08-24
performance: faster utf8 validation by using simd instructions

robert3005:

### DIFF
--- a/vortex-buffer/src/string.rs
+++ b/vortex-buffer/src/string.rs
@@ -70,6 +70,7 @@ impl TryFrom<ByteBuffer> for BufferString {
     fn try_from(value: ByteBuffer) -> Result<Self, Self::Error> {
         let _ = simdutf8::basic::from_utf8(value.as_ref()).map_err(|_| {
             #[allow(clippy::unwrap_used)]
+            // run validation using `compat` package to get more detailed error message
             let err = simdutf8::compat::from_utf8(value.as_ref()).unwrap_err();
             vortex_err!("invalid utf-8: {err}")
         })?;
--- a/vortex-buffer/src/string.rs
+++ b/vortex-buffer/src/string.rs
@@ -68,8 +68,11 @@ impl TryFrom<ByteBuffer> for BufferString {
     type Error = VortexError;
 
     fn try_from(value: ByteBuffer) -> Result<Self, Self::Error> {
-        let _ = simdutf8::basic::from_utf8(value.as_ref())
-            .map_err(|_| vortex_err!("not a valid utf-8 string"))?;
+        let _ = simdutf8::basic::from_utf8(value.as_ref()).map_err(|_| {
+            #[allow(clippy::unwrap_used)]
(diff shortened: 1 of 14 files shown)


vortex-error/src/lib.rs
```diff
@@ -101,8 +101,6 @@ pub enum VortexError {
     FmtError(fmt::Error, Box<Backtrace>),
     /// A wrapper for IO errors.
     IOError(io::Error, Box<Backtrace>),
-    /// A wrapper for UTF-8 conversion errors.
-    Utf8Error(std::str::Utf8Error, Box<Backtrace>),
```
robert3005, line 105: while I was at it I removed this from error enum

robert3005:
not sure why it segfaults on linux :/

vortex-duckdb/src/duckdb/scalar_function.rs
```diff
@@ -14,7 +14,8 @@ impl ScalarFunction {
             let name_ptr = cpp::duckdb_vx_sfunc_name(self.as_ptr());
             std::ffi::CStr::from_ptr(name_ptr)
                 .to_str()
-                .vortex_expect("invalid utf-8")
+                .map_err(|_| vortex_err!("invalid utf-8"))
```
lwwmanning: you now swallow the Utf8Error though, which might have given useful info (e.g., which charpoint in a long string)

robert3005: we can forward the error

vortex-buffer/src/string.rs
```diff
@@ -64,10 +65,14 @@ impl From<&str> for BufferString {
...
 
 impl TryFrom<ByteBuffer> for BufferString {
-    type Error = Utf8Error;
+    type Error = VortexError;
 
     fn try_from(value: ByteBuffer) -> Result<Self, Self::Error> {
-        let _ = std::str::from_utf8(value.as_ref())?;
+        let _ = simdutf8::basic::from_utf8(value.as_ref()).map_err(|_| {
+            #[allow(clippy::unwrap_used)]
+            let err = simdutf8::compat::from_utf8(value.as_ref()).unwrap_err();
```
lwwmanning, line 74: wat? doesn't it pass in the err? you're already in map_err from the same call

robert3005: Basic is basic. The error doesn’t have any information to make it a zero sized type

robert3005: The compat variant has same error reporting as the standard library but they need to produce own struct since you can’t create utf8error yourself

lwwmanning: is compat that much slower? seems weird to run validation twice

lwwmanning: eh, looks like it is slower, and it bails on first error. maybe just leave a comment? I didn't even notice the basic -> compat transition on first read

lwwmanning (approved):

PR #4350, opened 2025-08-25
chore: add PyO3 environment variables

connortsui20:
This should make builds a bit more consistent.

I also believe it prevents rebuilding some things. If you change terminals or if something related to the python installation changes, PyO3 will rebuild everything that is dependent on it. Not sure how important that is in practice, but this seems like a good thing to have.


danking (approved):
This seems great! Unfortunate that we now store `3.11` in both `.python-version` and here, though I'm not sure how to fix that without some pre-build step.

PR #4352, opened 2025-08-25
fix: don't recursively dictionary-encode dictionary codes

asubiotto:
I added an optional `exclude_int_dict_encoding` setting to the `BtrBlocksCompressor` to be minimally invasive but I wonder if we should just call `IntCompressor::compress_no_dict` always? I'm not sure it ever makes sense to dictionary encode integers.

### DIFF
--- a/vortex-file/src/tests.rs
+++ b/vortex-file/src/tests.rs
@@ -17,6 +17,7 @@ use vortex_array::stream::ArrayStreamExt;
 use vortex_array::validity::Validity;
 use vortex_array::{Array, ArrayRef, IntoArray, ToCanonical};
 use vortex_buffer::{Buffer, ByteBufferMut, buffer};
+use vortex_dict::{DictEncoding, DictVTable};
 use vortex_dtype::PType::I32;
 use vortex_dtype::{DType, DecimalDType, Nullability, PType, StructFields};
 use vortex_error::VortexResult;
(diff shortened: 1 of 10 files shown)


a10y:
I think there are some cases where dictionary is the best option, mainly when you have large dynamic range but a few unique values.

Take for example a postings list, where a term appears in many `u32` document IDs. The IDs may be randomly dispersed such that FFOR doesn't compress at all, but dict will.

vortex-file/src/tests.rs
```diff
@@ -1183,3 +1184,26 @@ async fn test_into_tokio_array_stream() -> VortexResult<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_array_stream_no_double_dict_encode() -> VortexResult<()> {
```
onursatici, line 1189: I am curious if this was failing before?

onursatici (commented):
Thanks for the contribution! From what I can tell you are disabling dict encoding for all integers here right? I am curious to see a repro of a recursively dict encoded array, as that is definitely a bug because in the DictScheme we do add the dict encoding to excludes before cascading. If we fix that bug happy to make that the default compressor, so we don't need to expose it as an option even

asubiotto:
> From what I can tell you are disabling dict encoding for all integers here right?

AFAICT I'm only disabling it for dictionary codes since the `compressing` strategy is only used by the `coalescing` strategy which in turn is only used for the codes or fallback in the `DictStrategy`. I'm new to this code so I could be wrong. Note that the `compress_then_flat` strategy does **not** disable dict encoding.

Happy to rework this!

vortex-file/src/tests.rs
asubiotto: Yes, it was.

robert3005 (approved):

onursatici:
@asubiotto yeap that makes sense the first level of dict encoding would be coming from the dict layout, so we want no further dict encoding in the compressor that we put as a child of the dict layout writer

PR #4354, opened 2025-08-25
chore: revert to 20.04, enforce glibc version == 2.31

danking:


.github/workflows/publish.yml
```diff
@@ -164,6 +164,12 @@ jobs:
           apt install -y wget curl build-essential
       - name: Print GLIBC version
         run: ldd --version
+      - name: Verify GLIBC version
+        run: |
+          set -eux -o pipefail
+
+          LDD_VERSION=$(ldd --version | head -n 1 | sed -E 's/.*GLIBC ([0-9]+[.][0-9]+).*/\1/')
+          [ $LDD_VERSION == "2.31" ]
```
a10y: can we be even simpler?

```
ldd --version | grep 'GLIBC 2.31'
```

should do the trick

danking: done.

danking: y u no like my 1337 b4sh h4x0r sk1llz

a10y: actually changed my mind, i think we can write this more elegantly as an elisp macro

```
(defun glibc-version-string ()
  "Get the glibc version string by calling ldd --version."
  (condition-case nil
      (with-temp-buffer
        (call-process "ldd" nil t nil "--version")
        (goto-char (point-min))
        (when (re-search-forward "ldd (GNU libc) \\([0-9]+\\.[0-9]+\\)" nil t)
          (match-string 1)))
    (error nil)))

(defun parse-version (version-string)
  "Parse a version string like '2.31' into a list (major minor)."
  (when version-string
    (let ((parts (split-string version-string "\\.")))
      (when (>= (length parts) 2)
        (list (string-to-number (nth 0 parts))
              (string-to-number (nth 1 parts)))))))

(defun version-compare (v1 v2)
  "Compare two version lists. Returns -1 if v1 < v2, 0 if equal, 1 if v1 > v2."
  (let ((major1 (nth 0 v1)) (minor1 (nth 1 v1))
        (major2 (nth 0 v2)) (minor2 (nth 1 v2)))
    (cond
     ((< major1 major2) -1)
     ((> major1 major2) 1)
     ((< minor1 minor2) -1)
     ((> minor1 minor2) 1)
     (t 0))))

(defmacro glibc-version-check (min-version &optional max-version)
  "Check if the current glibc version meets the specified requirements.
MIN-VERSION should be a string like \"2.31\".
MAX-VERSION is optional and should be a string like \"2.35\".
Returns t if version is within range, nil otherwise.

Examples:
  (glibc-version-check \"2.31\")        ; Check if >= 2.31
  (glibc-version-check \"2.31\" \"2.35\") ; Check if >= 2.31 and <= 2.35"
  `(let* ((current-version-str (glibc-version-string))
          (current-version (parse-version current-version-str))
          (min-ver (parse-version ,min-version))
          (max-ver ,(when max-version `(parse-version ,max-version))))
     (when (and current-version min-ver)
       (let ((min-check (>= (version-compare current-version min-ver) 0))
             (max-check (if max-ver
                           (<= (version-compare current-version max-ver) 0)
                         t)))
         (and min-check max-check)))))

;; Convenience macro for error handling
(defmacro require-glibc-version (min-version &optional max-version error-message)
  "Require glibc version to be within specified range, error if not.
ERROR-MESSAGE is optional custom error message."
  `(unless (glibc-version-check ,min-version ,max-version)
     (error ,(or error-message 
                 (if max-version
                     (format "glibc version must be between %s and %s" min-version max-version)
                   (format "glibc version must be at least %s" min-version))))))
```

a10y: @danking plzfix ?

a10y (approved):
i think this can be more succinctly implemented with an elisp macro

PR #4330, opened 2025-08-22
chore [ci]: nightly build

joseph-isaacs:

PR #4358, opened 2025-08-25
chore[ci]: pin nightly toolchain uses

lwwmanning:
Adds a `rust-toolchain-nightly.toml` file for CI to use where we need nightly. Figured it is cleaner & easier to maintain rather than inlining a string everywhere.

The implicit change here is that specifying this file is essentially imposing a semantic of "run the whole job as if this were `rust-toolchain.toml`", and thus overrides the toolchain for the whole job. I think that's actually less confusing than the current setup, but curious for feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,9 +228,9 @@ jobs:
       - uses: actions/checkout@v5
       - uses: ./.github/actions/setup-rust
         with:
-          toolchain: nightly
+          toolchain-file: rust-toolchain-nightly.toml
       - name: Rust Lint - Format
-        run: cargo +nightly fmt --all --check
+        run: cargo fmt --all --check
       - name: Rustc check
         run: cargo check --locked --all-features --all-targets
       - name: Rust Lint - Clippy All Features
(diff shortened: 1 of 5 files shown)


lwwmanning:
Alternative to #4330

joseph-isaacs:
shall we merge this

rust-toolchain-nightly.toml
joseph-isaacs, line 1: can you add a comment here explain that this is only used in CI, but natively by cargo?

.github/workflows/ci.yml
```diff
@@ -227,9 +227,8 @@ jobs:
           sccache: s3
       - uses: actions/checkout@v5
       - uses: ./.github/actions/setup-rust
-        with:
-          toolchain: nightly
       - name: Rust Lint - Format
+        # this is the one place where we always want latest nightly
         run: cargo +nightly fmt --all --check
```
danking, line 232: Why do we want latest nightly rather than a pin? Latest nightly means develop could start failing without a code change, right?

.github/workflows/ci.yml
```diff
@@ -514,13 +513,13 @@ jobs:
       - uses: actions/checkout@v5
       - uses: ./.github/actions/setup-rust
         with:
-          toolchain: nightly
+          toolchain-file: rust-toolchain-nightly.toml
           components: "rust-src, rustfmt, clippy, miri"
```
danking, line 517: Does Miri need rustfmt and clippy?

danking (changes requested):

PR #4368, opened 2025-08-26
fix: cleanup old to_record_batch_reader

a10y:
Confusing and incomplete API, instead we just expose via the `VortexFile::scan` method

### DIFF
--- a/docs/quickstart/python.rst
+++ b/docs/quickstart/python.rst
@@ -28,22 +28,6 @@ Vortex array:
    >>> vtx.nbytes
    141024
 
-Compress
---------
-
-Use :func:`~vortex.compress` to compress the Vortex array and check the relative size:
-
-.. doctest::
-
-   >>> cvtx = vx.compress(vtx)
-   >>> cvtx.nbytes
-   14298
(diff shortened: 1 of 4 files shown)


docs/quickstart/python.rst
```diff
@@ -28,30 +28,14 @@ Vortex array:
    >>> vtx.nbytes
    141024
 
-Compress
```
a10y, line 31: I think having a Compress section followed by Write makes people think they need to call compress before they write. Python users probably shouldn't be calling compress directly, so I think this is better suited for some kind of Python docs rather than the quickstart guide

vortex-python/src/dataset.rs
```diff
@@ -127,26 +127,20 @@ impl PyVortexDataset {
         Ok(PyArrayRef::from(array))
     }
 
-    #[pyo3(signature = (*, columns = None, row_filter = None, indices = None, split_by = None))]
```
a10y, line 130: disagrees with the Python definitions

danking (approved):

PR #4369, opened 2025-08-26
fix[vortex-datafusion]: add filename context on read error

asubiotto:
This has been useful to me to pinpoint the source of the error, allowing me to download and debug further locally.


AdamGS (approved):
Thank you!

PR #4373, opened 2025-08-26
Pretty tree display for Vortex expressions

AdamGS:
My main motivations here are:
1. Display that is easier to understand quickly, especially in deeply nested cases.
2. Easier to diff when using `insta`.

I've made it hidden behind a feature, but not sure that's actually desirable here, it adds one small dependency and it seems unlikely that the code would be misused in some way that will really bomb perf.

### DIFF
--- a/bench-vortex/src/datasets/struct_list_of_ints.rs
+++ b/bench-vortex/src/datasets/struct_list_of_ints.rs
@@ -39,7 +39,7 @@ impl Dataset for StructListOfInts {
 
     async fn to_vortex_array(&self) -> ArrayRef {
         let names: FieldNames = (0..self.num_columns)
-            .map(|col_idx| (col_idx.to_string()))
+            .map(|col_idx| col_idx.to_string())
             .collect();
         let mut rng = rand::rngs::StdRng::seed_from_u64(0);
 
--- a/vortex-expr/src/exprs/literal.rs
+++ b/vortex-expr/src/exprs/literal.rs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-
 use vortex_array::arrays::ConstantArray;
 use vortex_array::{ArrayRef, DeserializeMetadata, IntoArray, ProstMetadata};
 use vortex_dtype::{DType, match_each_float_ptype};
--- a/vortex-expr/src/exprs/literal.rs
+++ b/vortex-expr/src/exprs/literal.rs
@@ -106,7 +106,7 @@ impl DisplayAs for LiteralExpr {
             DisplayFormat::Tree => {
                 write!(
                     f,
-                    "LiteralExpr(value: {}, dtype: {})",
+                    "Literal(value: {}, dtype: {})",
                     self.value,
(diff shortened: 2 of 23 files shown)


joseph-isaacs:
are there any big examples

AdamGS:
Added one [here](https://github.com/vortex-data/vortex/pull/4373/commits/c681d144f94952a076efed3199bc20695f33e076) (including a rustdoc)

joseph-isaacs:
Does this mean we need to impl Debug + Display and DisplayAs for all new extra? can we avoid one of these?

AdamGS:
I think this is the best we can do now? because we wan't `"{}"` to work, but there's no way to provide a blanket implementation for foreign types, best we can do (which we already do) is format `ExprAdatper` and `dyn VortexExpr`

joseph-isaacs:
Maybe we can define a common display and DisplayAs for ann expr without its child and then use display or display_as for the childrenn and combine?

joseph-isaacs:
Sorry to be annoyinng but its already so much to add a new expr

AdamGS:
100% agree it should be as easy as possible to implement a new expr, I'll try and find a way to make it easier here.

robert3005:
Why do we still require display if we have DisplayAs? Simplest option is to define Display for Expr to go through displayas

AdamGS:
ok apparently just implementing `Display` for `dyn VortexExpr` is enough, so it now defaults to the `Dense` display (absolutely open to changing this name).
I've also removed the feature flag - it felt akward and would've made it harder for 3rd-party expressions to have nice tree display.

bench-vortex/src/datasets/struct_list_of_ints.rs
```diff
@@ -39,7 +39,7 @@ impl Dataset for StructListOfInts {
 
     async fn to_vortex_array(&self) -> ArrayRef {
         let names: FieldNames = (0..self.num_columns)
-            .map(|col_idx| (col_idx.to_string()))
+            .map(|col_idx| col_idx.to_string())
```
AdamGS, line 42: I have no idea why clippy suddenly figured out this thing

vortex-expr/src/exprs/literal.rs
```diff
@@ -98,9 +97,21 @@ impl LiteralExpr {
...
+impl DisplayAs for LiteralExpr {
+    fn fmt_as(&self, df: DisplayFormat, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match df {
+            DisplayFormat::Compact => {
+                write!(f, "{}", self.value)
+            }
+            DisplayFormat::Tree => {
+                write!(
+                    f,
+                    "LiteralExpr(value: {}, dtype: {})",
```
joseph-isaacs: lit()?

joseph-isaacs:
Is the plan here to make this use the full expr name? It quite verbose

vortex-expr/src/exprs/literal.rs
AdamGS: I was considering dropping the `Expr` at the end, but in general I'm in favor of fully spelled out names here.

joseph-isaacs (approved):
LGTM

PR #4379, opened 2025-08-26
WIP: FSSTView

a10y:
Goal is to bring the benefits of views (fast random take/filter, better pushdown) with FSST compression.

DuckDB goes even further here with an explicit DictFSST encoding, not sure we need that

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6902,6 +6902,7 @@ dependencies = [
  "vortex-mask",
  "vortex-scalar",
  "vortex-utils",
+ "zstd",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6901,6 +6901,7 @@ dependencies = [
  "vortex-error",
  "vortex-mask",
  "vortex-scalar",
+ "vortex-utils",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6891,6 +6891,7 @@ version = "0.1.0"
 dependencies = [
  "codspeed-divan-compat",
  "fsst-rs",
+ "num-traits",
  "prost 0.14.1",
  "rand 0.9.2",
  "rstest",
--- a/encodings/fsst/Cargo.toml
+++ b/encodings/fsst/Cargo.toml
@@ -27,6 +27,7 @@ vortex-error = { workspace = true }
 vortex-mask = { workspace = true }
 vortex-scalar = { workspace = true }
 vortex-utils = { workspace = true }
+zstd = { workspace = true }
 
 [dev-dependencies]
 divan = { workspace = true }
--- a/encodings/fsst/Cargo.toml
+++ b/encodings/fsst/Cargo.toml
@@ -18,6 +18,7 @@ workspace = true
 
 [dependencies]
 fsst-rs = { workspace = true }
+num-traits = { workspace = true }
 prost = { workspace = true }
 vortex-array = { workspace = true }
 vortex-buffer = { workspace = true }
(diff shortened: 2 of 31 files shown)


a10y:
Promising results for canonicalizing (this is mix of null, non-null, inlined and FSST strings 64k row groups)

```
canonical     fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ fsst       154.8 µs      │ 313.7 µs      │ 160.8 µs      │ 162 µs        │ 100     │ 100
╰─ fsst_view  48.91 µs      │ 133 µs        │ 49.77 µs      │ 51.9 µs       │ 100     │ 100
```

I want to also enable prefix-match pushdown (`x LIKE 'ABC%'`), something we couldn't do with normal FSST but that we can do here. And I think there are some more optimization tricks we can play with compare when the needle string is <= 12 bytes.

Also we can start deduplicating views on the writer

a10y:
I'm struggling a bit to understand why some of the TPC-H S3 queries get markedly slower, despite the microbenchmarks indicating that all things that should matter (compare, canonicalize) are faster than the old FSST.

There's also some weird bug lurking in here that is being tripped in TPC-DS data gen which I'm struggling to replicate locally

robert3005:
Some extra roundtrips maybe?

a10y:
It was a very dumb bug in the slice impl 🤦

a10y:
So FSSTView seems to give a lot of speed up on microbenchmarks, some operations up to 3x faster as documented above.

However, there are a few issues

* Poor compression for datasets with lots of short strings. In PBI, Arade and HashTags get considerably larger

```
## BEFORE

-rw-r--r--@ 1 aduffy  staff   136M Sep  2 10:56 Arade/vortex/Arade_1.vortex
-rw-r--r--@ 1 aduffy  staff   506M Sep  2 10:57 Bimbo/vortex/Bimbo_1.vortex
-rw-r--r--@ 1 aduffy  staff   482M Sep  2 11:00 CMSprovider/vortex/CMSprovider_1.vortex
-rw-r--r--@ 1 aduffy  staff   485M Sep  2 11:00 CMSprovider/vortex/CMSprovider_2.vortex
-rw-r--r--@ 1 aduffy  staff   166M Sep  2 11:01 Euro2016/vortex/Euro2016_1.vortex
-rw-r--r--@ 1 aduffy  staff    43M Sep  2 11:01 Food/vortex/Food_1.vortex
-rw-r--r--@ 1 aduffy  staff   196M Sep  2 11:01 HashTags/vortex/HashTags_1.vortex

## AFTER

-rw-r--r--@ 1 aduffy  staff   438M Sep  2 10:08 Arade/vortex/Arade_1.vortex
-rw-r--r--@ 1 aduffy  staff   506M Sep  2 10:14 Bimbo/vortex/Bimbo_1.vortex
-rw-r--r--@ 1 aduffy  staff   494M Sep  2 10:16 CMSprovider/vortex/CMSprovider_1.vortex
-rw-r--r--@ 1 aduffy  staff   497M Sep  2 10:16 CMSprovider/vortex/CMSprovider_2.vortex
-rw-r--r--@ 1 aduffy  staff   238M Sep  2 10:17 Euro2016/vortex/Euro2016_1.vortex
-rw-r--r--@ 1 aduffy  staff    43M Sep  2 10:17 Food/vortex/Food_1.vortex
-rw-r--r--@ 1 aduffy  staff   295M Sep  2 10:18 HashTags/vortex/HashTags_1.vortex
```

* NVME benchmarks seem very solid, but some of the S3 queries are a lot worse up, up to 50%. The likely culprit here is that the lineitem SF=1 table is almost 2x larger. Examining several chunks, the `views` buffer is up to 2x larger than the `fsst_compressed` data buffers they refer to!

We need to compress the views buffer at rest. Some immediate things to try after chatting with @robert3005 

* ZSTD the whole thing
* ALP-RD style splitting out components into separate bitpacked streams

Issue #4372, opened 2025-08-26
Tracking Issue: Fixed-Size List Arrays


connortsui20:
# Tracking Issue: Fixed-Size List Arrays

This is a tracking issue for adding fixed-size list arrays to Vortex.

The main difference between lists and fixed-size lists is that fixed-size lists do not need to store an `offset`s buffer, since we are able to use the predictable stride pattern to jump to data. This allows us to more efficiently store data with known size (like matrices or tensors) as well as compute more efficiently with SIMD.

### Public API

```rust
pub enum DType {
    /// A logical fixed-size list type.
    ///
    /// This is parameterized by a `DType` that represents the element type of the inner lists, as
    /// well as a `u32` size that determines the fixed length of each `FixedSizeList` scalar.
    FixedSizeList(Arc<DType>, u32, Nullability),
}

#[cfg(not(target_arch = "wasm32"))]
const_assert_eq!(size_of::<DType>(), 16);

#[cfg(target_arch = "wasm32")]
const_assert_eq!(size_of::<DType>(), 12); // Increase size to 12 bytes because of the `u32` field.
```

### Prior Art

- Apache Arrow Fixed-Size List Overview: https://arrow.apache.org/docs/format/Intro.html#fixed-size-list
- Apache Arrow Fixed-Size List Layout: https://arrow.apache.org/docs/format/Columnar.html#fixed-size-list-layout
- Apache Arrow Variable-Size List Layout: https://arrow.apache.org/docs/format/Columnar.html#variable-size-list-layout
- Apache Arrow Fixed-shape Tensor: https://arrow.apache.org/docs/format/CanonicalExtensions.html#fixed-shape-tensor
- Arrow Rust `FixedSizeListArray`: https://docs.rs/arrow/latest/arrow/array/struct.FixedSizeListArray.html
- `pyarrow.FixedSizeListType`: https://arrow.apache.org/docs/python/generated/pyarrow.FixedSizeListType.html
- DuckDB Array Type (the same as fixed-size list): https://duckdb.org/docs/stable/sql/data_types/array.html
- Adding a new variant (`DType::Decimal`) to Vortex: https://github.com/vortex-data/vortex/pull/3058
- Conversion to numpy array potential optimizations: https://github.com/apache/arrow/issues/ 35622 (so we don't ping that issue)

### Steps / History

- [x] https://github.com/vortex-data/vortex/pull/4371
- [x] https://github.com/vortex-data/vortex/pull/4380
[...]

PR #4380, opened 2025-08-26
Feature: Add `FixedSizeList` support in `vortex-scalar`

connortsui20:

Adds `vortex-scalar` support for `FixedSizeList`.

We reuse `ListScalar` to represent **both** `List` and `FixedSizeList` logical types, as there is no effective difference between a single scalar value of a list or fixed-size list.

Also refactors the unit `tests` module to be a bit more clear in what each submodule is actually testing, plus adds several tests for `FixedSizeList`.

Also fixed some `no_coercion` tests that checked the wrong things.

For reviewers: it might be a good idea to review each commit separately


robert3005 (approved):

robert3005:
FWIW it might be beneficial to split refactor logic into separate prs. You should think of atomic unit of work in github as a pr and not commit (though working with commits in ui has gotten slightly better over time)

PR #4381, opened 2025-08-26
fix: When coercing scalars make sure values don't exceed maximum value

robert3005:

### DIFF
--- a/vortex-scalar/src/pvalue.rs
+++ b/vortex-scalar/src/pvalue.rs
@@ -9,7 +9,9 @@ use num_traits::NumCast;
 use paste::paste;
 use vortex_dtype::half::f16;
 use vortex_dtype::{NativePType, PType, ToBytes};
-use vortex_error::{VortexError, VortexExpect, VortexResult, vortex_bail, vortex_err};
+use vortex_error::{
+    VortexError, VortexExpect, VortexResult, vortex_bail, vortex_ensure, vortex_err,
+};
 
 /// A primitive value that can represent any primitive type supported by Vortex.
 ///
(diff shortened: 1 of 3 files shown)


robert3005:
found as a part of #4380 needs more tests

vortex-scalar/src/pvalue.rs
```diff
@@ -444,8 +446,20 @@ impl CoercePValue for f16 {
         match value {
             PValue::U8(u) => Ok(Self::from_bits(u as u16)),
             PValue::U16(u) => Ok(Self::from_bits(u)),
-            PValue::U32(u) => Ok(Self::from_bits(u as u16)),
-            PValue::U64(u) => Ok(Self::from_bits(u as u16)),
+            PValue::U32(u) => {
+                vortex_ensure!(
```
lwwmanning, line 450: not really part of this PR, but should this just be called `vortex_assert!`...?

vortex-scalar/src/pvalue.rs
```diff
@@ -444,8 +446,20 @@ impl CoercePValue for f16 {
...
+                    u & !((1u32 << 16) - 1) == 0,
```
joseph-isaacs: can we do a
```suggestion
                    u >> 16 == 0,
```

joseph-isaacs: or compare to u16 max?

a10y: +1

a10y (commented):
Putting it at top level since for some reason GH isn't letting me comment on lines not included in the diff:

Now that we're reviewing this, I'm not sure NumCast is appropriate for float conversions, shouldn't we be using to_bits and from_bits to round-trip these? If coercion is about bit patterns and not value conversion

robert3005:
I misspoke when I said it's about bitpatterns. It's about things that are representable using different binary representation. We made somewhat arbitrary choice of treating integers as bit representations and floats as values

vortex-scalar/src/pvalue.rs
robert3005: Hm, ensure returns an error, vortex_assert would panic

danking (approved):
LGTM

robert3005:
I will add some docs on scalarvalue treatment, too many times we have issues with them

PR #4383, opened 2025-08-26
Run pruning expressions concurrently

gatesn:
This should also allow us to coalesce the I/O more effectively


joseph-isaacs:
Any thoughts on the lack of runtime change?

PR #4386, opened 2025-08-26
feat: VortexDataset.get_fragments yields one fragment per split

danking:


vortex-file/src/file.rs
```diff
@@ -133,4 +133,19 @@ impl VortexFile {
...
+        self.layout_reader()?
+            .register_splits(&[FieldMask::All], 0, &mut splits)?;
+        let mut prev = 0_u64;
+        Ok(splits
+            .into_iter()
+            .map(|i| {
+                let range = (prev, i);
+                prev = i;
+                range
+            })
```
a10y: i think you can do this with .scan

vortex-python/python/vortex/dataset.py
```diff
@@ -271,6 +282,7 @@ def take(  # pyright: ignore[reportIncompatibleMethodOverride]
         use_threads: bool | None = None,
         cache_metadata: bool = False,
         memory_pool: pyarrow.MemoryPool | None = None,
+        _row_range: tuple[int, int] | None = None,
```
a10y, line 285: why `_row_range`?

a10y, line 285: I wish there were a `RangeLike` typeclass so you could also take `range()`, list[int], etc.

vortex-file/src/file.rs
robert3005: Is there maybe a way to use SplitBy::splits?

vortex-python/python/vortex/dataset.py
danking: It's not part of the Dataset API and I don't want to commit to supporting it if we change our minds about how this works.

vortex-file/src/file.rs
danking: @robert3005 , if this was public I could:
```

impl SplitBy {
    /// Compute the splits for the given layout.
    // TODO(ngates): remove this once layout readers are stream based.
    pub(crate) fn splits(
        &self,
        layout_reader: &dyn LayoutReader,
        field_mask: &[FieldMask],
    ) -> VortexResult<Vec<Range<u64>>> {
```

danking: @a10y done.

vortex-file/src/file.rs
```diff
@@ -133,4 +134,18 @@ impl VortexFile {
...
+        self.layout_reader()?
+            .register_splits(&[FieldMask::All], 0, &mut splits)?;
+        Ok(splits
+            .into_iter()
+            .scan(0_u64, |prev, i| {
+                let range = (*prev, i);
+                *prev = i;
+                Some(range)
+            })
+            .collect())
```
robert3005: I wonder if we can have a way to avoid duplication and this being out of sync with whatever the scanbuilder does. Maybe this is something that Scan can return?

danking: In here, I only have a LayoutReader. I could create a Scan but I feel like Scan is supposed to package up splitting for you.

It feels like the right way to de-dupe this is to make `SplitBy::splits` public, but:
```rust
impl SplitBy {
    /// Compute the splits for the given layout.
    // TODO(ngates): remove this once layout readers are stream based.
    pub(crate) fn splits(
        &self,
        layout_reader: &dyn LayoutReader,
        field_mask: &[FieldMask],
    ) -> VortexResult<Vec<Range<u64>>> {
```

Should we just make it public and then remove it when Nick finishes the great rewrite of destiny?

robert3005: I think this makes sense to be public. The whole splitting logic will have to be rethought

danking: @robert3005 done.

robert3005 (approved):

PR #4388, opened 2025-08-27
feat: expr -> operator conversion and use in layouts

joseph-isaacs:


vortex-expr/src/operator/mod.rs
```diff
@@ -3,4 +3,49 @@
...
 pub use reduce::*;
+use vortex_array::Array;
+use vortex_array::pipeline::Operator;
+use vortex_error::VortexResult;
+
+use crate::traversal::{FoldUp, NodeFolder};
+use crate::{ExprRef, RootVTable};
+
+pub struct ExprOperatorConverter<'a> {
+    root: &'a dyn Array,
```
gatesn: Can this not just already be the Array's operator?

vortex-layout/src/layouts/flat/reader.rs
```diff
@@ -234,6 +247,55 @@ impl ArrayEvaluation for FlatEvaluation {
     }
 }
 
+fn try_to_operator(
```
gatesn: try_evaluate_using_operator?

robert3005 (approved):

Issue #1643, opened 2024-12-11
PyVortex creating array from python list should create canonical array


robert3005:
Right now we delegate to pyarrow which will for strings/binary produces String/Binary arrays. We want those cases to produce StringView/BinaryView arrays

PR #4390, opened 2025-08-27
fix[vortex-array]: propagate preferred type conversion to nested fields

asubiotto:
When arrow_type is None, vortex to arrow conversion should respect conversion to the encoding's preferred arrow type for all nested fields. Previous to this commit, this preference was lost for any fields within structs or lists.

For example, a struct with a utf8 dictionary field would lose the dictionary encoding.

### DIFF
--- a/vortex-layout/src/layouts/dict/reader.rs
+++ b/vortex-layout/src/layouts/dict/reader.rs
@@ -328,7 +328,8 @@ mod tests {
         .unwrap()
         .into_array();
         let actual = actual.into_arrow_preferred().unwrap();
-        let expected = expected.into_arrow_preferred().unwrap();
+        let expected_arrow_dtype = expected.dtype().to_arrow_dtype().unwrap();
+        let expected = expected.into_arrow(&expected_arrow_dtype).unwrap();
         assert_eq!(actual.data_type(), expected.data_type());
         assert_eq!(&actual, &expected);
     }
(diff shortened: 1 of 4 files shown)


vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -328,7 +328,11 @@ mod tests {
         .unwrap()
         .into_array();
         let actual = actual.into_arrow_preferred().unwrap();
-        let expected = expected.into_arrow_preferred().unwrap();
+        // TODO: The expected array has vortex.varbin encoding, which prefers
```
asubiotto: This was surprising to me: https://github.com/vortex-data/vortex/blob/11c9452e486d57e01f0560f9e7eed29dcf78fac3/vortex-array/src/arrow/compute/to_arrow/varbin.rs#L90. Shouldn't varbin arrays be zero-copy converted into Utf8Views?

a10y: You need to do a linear scan to read every string and construct its view into a new views buffer, so it's actually fairly expensive to do this

asubiotto: urgh sorry, I was getting confused with the `VarBinView` array. Will remove the TODO

robert3005:
Seems like the python conversion changed 
```
File "../vortex-python/python/vortex/__init__.py", line ?, in default
Failed example:
    array.to_arrow_table()
Expected:
    pyarrow.Table
    age: int64
    name: string_view
    ----
    age: [[25,31,33,57]]
    name: [["Joseph","Narendra","Angela","Mikhail"]]
Got:
    pyarrow.Table
    age: int64
    name: string
    ----
    age: [[25,31,33,57]]
    name: [["Joseph","Narendra","Angela","Mikhail"]]
```
I think we were relying on the implicit conversion before and might need an explicit argument somewhere

asubiotto:
> Seems like the python conversion changed

Fixed. Just had to update the expected type.

robert3005:
ah you have managed to hit https://github.com/vortex-data/vortex/issues/1643. Thanks for fixing it

robert3005 (approved):

PR #4391, opened 2025-08-27
chore: Unify `slice` methods to use Range arguments

### LINKED ISSUE #4389: Unify `slice`
Should `slice` take `start` and `end` OR `start` and `length`? We have both:

https://github.com/vortex-data/vortex/blob/8ed46d5d1b7972cfac3be0ba0125bd2d8c8679b6/vortex-mask/src/lib.rs#L469

https://github.com/vortex-data/vortex/blob/8ed46d5d1b7972cfac3be0ba0125bd2d8c8679b6/vortex-array/src/arrays/struct_/mod.rs#L445

Arrow has `start` and `length` - should we unify? In fact, I several times had a silent bug when I replaced arrow array with vortex - compilation passed but it was actually wrong, as the signature should be changed from `array.slice(a, b)` to `array.slice(a, a + b)`

blaginin:
Closes #4389: Unifies inconsistent slice method signatures by introducing Range-based APIs.

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/vortex-mask/src/tests.rs
+++ b/vortex-mask/src/tests.rs
@@ -119,7 +119,7 @@ fn test_mask_false_count() {
 fn test_mask_slice() {
     let mask = Mask::from_buffer(BooleanBuffer::from_iter([true, false, true, true, false]));
 
-    let sliced = mask.slice(1..3);
+    let sliced = mask.slice(1..4);
     assert_eq!(sliced.len(), 3);
     assert_eq!(sliced.true_count(), 2);
     assert!(!sliced.value(0)); // false from index 1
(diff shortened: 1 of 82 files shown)


blaginin:
hmm i'd rather have just one (new) method rather than new and old ones..

vortex-mask/src/tests.rs
```diff
@@ -119,20 +119,20 @@ fn test_mask_false_count() {
 fn test_mask_slice() {
     let mask = Mask::from_buffer(BooleanBuffer::from_iter([true, false, true, true, false]));
 
-    let sliced = mask.slice(1, 3);
+    let sliced = mask.slice(1..4);
```
blaginin, line 122: before it was start, len

robert3005 (approved):

PR #4393, opened 2025-08-27
chore: rust depends on rust, java on java, and python on python

danking:
Currently, Python depends only on the prepare-python jobs, Java depends only on the prepare-java jobs, and Rust depends on both prepare-python and prepare-java (there are no prepare-rust jobs).

After this PR, Rust, Python, and Java can all independently publish as long as their publish sequence succeeds.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,7 +11,6 @@ jobs:
   publish-rust:
     runs-on: ubuntu-latest
     timeout-minutes: 120
-    needs: [prepare-python, prepare-java-macos, prepare-java-linux]
     steps:
       - uses: actions/checkout@v5
         with:


lwwmanning (approved):

a10y (changes requested):
can we fix the glibc problem here too

a10y:
I actually think that I prefer the way it works now: we have prepare phases which compile the artifacts for Java/Python. It's like 2PC. I think to the best effort, our releases should all publish in lockstep.

danking:
@a10y Right now, only Rust depends on the other two. It seems to me there are two reasonable answers:
1. (this PR) Each platform is independent.
2. Every platform depends on every `prepare-`. `publish` should only fail due to network issues (and we really ought to aggressively retry those).

a10y:
I think then that all of the `prepare-*` tasks should fan into a single no-op `prepare` job, and then all of the publish-* jobs should depend on that one prepare job

PR #4395, opened 2025-08-27
fix: Java GLIBC check

a10y:
Apparently @danking you tried to be too safe :P https://github.com/vortex-data/vortex/actions/runs/17223295185


danking:
too slow https://github.com/vortex-data/vortex/pull/4392

PR #4396, opened 2025-08-27
chore: remove gradle inferred processors and update gradle jvm links

robert3005:


a10y (approved):
whoops missed those the first time, nice

PR #4397, opened 2025-08-27
chore: prepare all before publish

danking:
I pinky promised that only renamed steps and moved them around so all the publishes are last.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -129,7 +129,7 @@ jobs:
           if-no-files-found: error
 
   prepare-all:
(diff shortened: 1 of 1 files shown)


a10y (approved):

.github/workflows/publish.yml
```diff
@@ -184,8 +128,70 @@ jobs:
           retention-days: 1
           if-no-files-found: error
 
+  prepare-all:
+    needs: [prepare-python, prepare-java-macos, prepare-java-linux, prepare-rust]
```
lwwmanning: where prepare-rust?

lwwmanning (approved):

PR #4398, opened 2025-08-27
Expand expression pushdown to DataFusion

AdamGS:
Started as a demo for @danking to show how to add more pushdown to Datafusion, now it includes some more expression (`IS NULL`, `IS NOT NULL` and `IN LIST`), which only leaves `substr` and dynamic expressions.
It also allows for pushing more binary operators for arithmetic operations.

`vortex-compact` seems really noise, but clickbench still shows nice improvements in a bunch of queries for the btrblocks compression.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6587,6 +6587,7 @@ dependencies = [
  "datafusion-physical-plan",
  "datafusion-pruning",
  "futures",
+ "insta",
  "itertools 0.14.0",
  "log",
  "moka",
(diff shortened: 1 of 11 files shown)


joseph-isaacs:
this is a filter push down right?

AdamGS:
yes, just that, still seems useful to pushdown if we can prune stuff with zonemaps or get some of them to run on compressed arrays.

robert3005 (approved):

danking:
IIRC, pco and zstd are missing almost all compute kernels so they always fully decompress even though that have fine grained pages

PR #3876, opened 2025-07-15
feat[layout]: replace register_splits with a layout splits stream

joseph-isaacs:
Signed-off-by: Nicholas Gates <nick@nickgates.com>

PR #4399, opened 2025-08-27
feat: prepared scan: accelerate successive range reads to a file

danking:


vortex-expr/src/lib.rs
```diff
@@ -147,7 +147,7 @@ impl dyn VortexExpr + '_ {
     /// Compute result of expression on given batch producing a new batch
     pub fn evaluate(&self, scope: &Scope) -> VortexResult<ArrayRef> {
         let result = self.unchecked_evaluate(scope)?;
-        assert_eq!(
+        debug_assert_eq!(
```
danking: dtype construction is fairly slow maybe 3% overall of the data loader example.

vortex-layout/src/layouts/chunked/reader.rs
```diff
@@ -64,7 +67,7 @@ impl ChunkedReader {
         self.lazy_children.get(
             idx,
             self.layout.dtype(),
-            &format!("{}.[{}]", self.name, idx).into(),
+            &self.child_name, // &format!("{}.[{}]", self.name, idx).into(),
```
danking: Maybe a spicy take but this is also like 3%

vortex-layout/src/layouts/struct_/reader.rs
```diff
@@ -97,8 +101,7 @@ impl StructReader {
             .field_by_index(idx)
             .ok_or_else(|| vortex_err!("Missing field {idx}"))?;
         let name = &self.struct_fields().names()[idx];
-        self.lazy_children
-            .get(idx, &field_dtype, &format!("{}.{}", self.name, name).into())
+        self.lazy_children.get(idx, &field_dtype, &self.child_name)
```
danking: again like 3%

danking:
More things to look at:

1. Executing Pack is like 3%. Can we somehow avoid this if it's just a `select(["name"], root())` getting pushed down to the flat layout?
2. SegmentSource::request is like 20% but totally unclear to me why/what.
3. Even after this change, a lot of repeated work happens in the split_exec (e.g. massaging filters).
4. Can we make slicing any faster?

[target_release_vortex_sample 2025-08-27 17.53 profile.json.gz](https://github.com/user-attachments/files/22014306/target_release_vortex_sample.2025-08-27.17.53.profile.json.gz)

AdamGS:
IMO if string building is so expensive here - we should use `Vec<String>` (or `Arc<str>`)

robert3005:
Before we end up going down the route of microoptimisations here can we revive https://github.com/vortex-data/vortex/pull/3876? The part we care about is not even visiting layouts that are not included in the requested read range. Once we have that we can look at all the small changes?

delta003:
I've asked Dan to try to push this as far as possible, so we can validate fast that the Vortex can work as an MDS format, so basically that design is possible. Happy to try another branch if it's going to give better results. We don't need to merge anything (yet).

robert3005:
I looked into this a ~month ago. I will send you the notes I had from that time

robert3005:
also to be clear I think caching layout readers is good. I was mostly addressing the random allocation changes

vortex-layout/src/layouts/chunked/reader.rs
delta003: Let's revert these micro stuff and figure out how do we merge PreparedScan

delta003 (commented):
Proposal: PreparedScan -> RandomAccessScan / TakeScan / IndicesScan or maybe just Take - it's a scan designed for repeated takes, or range reads. The API should accept everything ScanBuilder accepts that is related to indices/ranges.

Once you know that is how the file is going to be read, you can decide to do https://github.com/spiraldb/spiraldb/pull/2559#discussion_r2305376930

robert3005:
I think RepeatedScan or something in that vein works. I was thinking there's a two parts to this, if you know you're going to read most of the file then this is the way to do it. I was thinking that you also want to reduce the overhead of one off random access by not creating all layout readers

vortex-scan/src/lib.rs
```diff
@@ -186,28 +180,28 @@ impl<A: 'static + Send> ScanBuilder<A> {
         let field_mask: Vec<_> = [filter_mask, projection_mask].concat();
         let splits = self.split_by.splits(layout_reader.as_ref(), &field_mask)?;
```
onursatici, line 183: I expect the main improvement is from moving this so it is only done once with the prepared scan

onursatici (commented):
I agree with @delta003 and @robert3005 . Maybe the api would be like you  call `into_array_iter` on the scan builder you can also call `into_row_accessor` or something like that which will give you the equivalent of the prepared scan, and we can  then call `scalar_at`  or `read_row_range` on that, and internally it reuses the splits and the layout reader

vortex-scan/src/lib.rs
danking: The expression partitioning was also somewhat expensive,  constructing layouts over-and-over again was slow but not as slow as splits and expression manipulation.

danking:
I went with repeated scan since, at least as currently written, it isn't specialized to single-row reads.

vortex-scan/src/lib.rs
```diff
@@ -330,5 +313,101 @@ fn to_field_mask(field: FieldName) -> FieldMask {
     FieldMask::Prefix(FieldPath::from(Field::Name(field)))
 }
 
+pub struct RepeatedScan<A: 'static + Send> {
```
delta003, line 323: I use ScanBuilder to set a bunch of parameters, and I call prepare to get RepeatedScan.

1. Prepare methods to take the row range. What happens if I've set indices or row range in ScanBuilder before calling prepare? It seems to me these are ignored. That seems wrong.
2. Prepare methods that take concurrency. What happens to the concurrency set on ScanBuilder?
 
(2) feels like less of an issue; we could just not accept concurrency, and it must be set on the ScanBuilder. But what about (1)? One way would be to fail on "prepare" in ScanBuilder if indices are set. But that just feels like working around a bad API and deferring an issue for runtime. 

Is using a ScanBuilder useful? What if we have RepeatedScanBuilder? Or what we don't ignore any of the args in ScanBuilder? That's tricky cause now you have to translate from index relative to RepeatedScan to index relative to Scan with some indices. Curious if you have other ideas

delta003: Correction, indices are converted to selection to they propagate. This only applies to row_range arg of ScanBuilder. I guess that's more fine to assert in runtime? Acceptable for now? ScanBuilder is very complicated anyway.

delta003: I won't block on this. I'll see if I can think of something nicer, but I'd accept:
1. Removing concurrency
2. Raising an error if row_range is set on the builder when calling prepare.

danking: After an out-of-band chat we decided to:
1. ScanBuilder.concurrency is sent to RepeatedScan.
2. ScanBuilder.row_range is also sent to RepeatedScan and intersected with the row range argument of each method.

vortex-scan/src/lib.rs
```diff
@@ -330,5 +315,135 @@ fn to_field_mask(field: FieldName) -> FieldMask {
...
+    /// The natural splits of the file.
+    splits: Vec<Range<u64>>,
+    /// The number of splits to make progress on concurrently **per-thread**.
+    concurrency: usize,
+    /// Function to apply to each [`ArrayRef`] within the spawned split tasks.
+    map_fn: Arc<dyn Fn(ArrayRef) -> VortexResult<A> + Send + Sync>,
+    /// Maximal number of rows to read (after filtering)
+    limit: Option<usize>,
+    /// The fields we will read.
+    #[allow(dead_code)]
```
robert3005: is this dead?

robert3005 (approved):
Just delete the dead code

vortex-scan/src/lib.rs
```diff
@@ -330,5 +315,135 @@ fn to_field_mask(field: FieldName) -> FieldMask {
...
+    field_mask: Vec<FieldMask>,
+    /// The dtype of the projected arrays.
+    dtype: DType,
+}
+
+pub fn intersect_ranges(
```
robert3005: pub(crate)

vortex-scan/src/lib.rs
```diff
@@ -330,5 +314,131 @@ fn to_field_mask(field: FieldName) -> FieldMask {
...
+    /// Function to apply to each [`ArrayRef`] within the spawned split tasks.
+    map_fn: Arc<dyn Fn(ArrayRef) -> VortexResult<A> + Send + Sync>,
+    /// Maximal number of rows to read (after filtering)
+    limit: Option<usize>,
+    /// The dtype of the projected arrays.
+    dtype: DType,
+}
+
+impl<A: 'static + Send> RepeatedScan<A> {
+    pub fn execute(
```
delta003, line 344: calling this build was weird, since it's not a Builder, so renamed

vortex-scan/src/lib.rs
```diff
@@ -330,5 +314,131 @@ fn to_field_mask(field: FieldName) -> FieldMask {
...
+                    Some(split_exec(ctx.clone(), split_range.clone(), limit.as_mut()))
+                }
+            })
+            .try_collect()?;
+
+        Ok(split_tasks)
+    }
+
+    #[cfg(feature = "tokio")]
+    pub fn execute_tokio_stream(
```
delta003, line 377: calling this into when it takes &self, instead of self is wrong (JetBrains IDEs warn you about it)

vortex-scan/src/lib.rs
delta003: Removed. Can be added when it's used.

vortex-scan/src/lib.rs
delta003: Made this private

delta003 (approved):

Issue #4400, opened 2025-08-28
IO dispatcher pool overhead makes it unusable in production scenarios


asubiotto:
## Problem

I noticed some unusually long times spent in [`read_byte_range`](https://github.com/vortex-data/vortex/blob/f5473e57f79b36c5a50e9e32a8cb6467b3c3ae38/vortex-io/src/object_store.rs#L45) for a workload using datafusion partitioned 32 ways and doing 100% object reads.

<img width="720" height="397" alt="Image" src="https://github.com/user-attachments/assets/bd5d4d72-6df1-48fd-a06d-ae965ee591ac" />

Sometimes, these spans were as long as ~20s with the majority of time attributed to `idle_ns`. After noticing that the thread pool onto which these IO operations are dispatched only had a single thread https://github.com/vortex-data/vortex/blob/f5473e57f79b36c5a50e9e32a8cb6467b3c3ae38/vortex-file/src/generic.rs#L22-L23, I tried bumping the thread pool size to `num_cpus::get()` to no effect.

Based on tokio console, there was indeed a very high scheduling overhead:

<img width="720" height="195" alt="Image" src="https://github.com/user-attachments/assets/a08a66aa-5c72-4115-86c7-02fab87aa87d" />

At this point I tried removing the dispatch pool and running IO tasks on the same runtime, and performance improved dramatically, confirming that there is high scheduling overhead using the separate IO dispatcher pool.

## Profiles

I collected some profiling data while running the separate IO pool, and I think the majority of the overhead is coming from `flume`, which is the crate that provides the channel onto which IO tasks are scheduled and collected by the IO threads.

A CPU profile shows that the majority of CPU time is spent just managing threads waiting on the channel: https://pprof.me/b6613f12145582bc14829a2232574346

This is the code where the flume channel is read from for reference:
https://github.com/vortex-data/vortex/blob/f5473e57f79b36c5a50e9e32a8cb6467b3c3ae38/vortex-io/src/dispatcher/tokio.rs#L51

Additionally, there seems to be a bug in flume which causes a memory leak which probably doesn't help. After running a couple of queries, the system was in a fully idle state but holding on to ~30GiB of memory all coming from a [flume vector](https://github.com/zesterer/flume/blob/02eda72e14d6805beb3b27559b8e066f8464f490/src/async.rs#L410): https://pprof.me/7f60e6b4ad581c22be6dd3669ba70eb9

## Solution

[...]

PR #4401, opened 2025-08-28
Replace Flume with Kanal

AdamGS:
mostly just curious about some of the insights in https://github.com/vortex-data/vortex/issues/4400.

It seems to have some performance benefit, so I figure its worth merging while other work happens to make bigger changes.


AdamGS:
I think the second iteration of the benchmarks (which is just replacing the dispatcher's channel with Kanal) seems like the most improved? Trying to replace the segment event one seems to have made things worse, so I'm reverting that PR.

robert3005 (approved):

PR #4403, opened 2025-08-28
Run SQL benchmarks on metal instances

AdamGS:


danking:
Something weird with datafusion, no? Some of its queries get much slower.

AdamGS:
I think its a contention issue with the small scale factor, at sf=10 results are much better, I'm not sure what happened to sf=100, I'll try and find some time next week to look deeper into the whole thing 
I think I need to spend some time doing a deeper investigation to reach a useful conclusion here.

AdamGS:
The results are really weird in multiple ways, but much of it seems to be from contention somewhere in our duckdb code, and in multiple queries seems to be more related to the underlying engine itself (like tpcds-q21 with SF=1 which is between 5x to 6x slower than our baseline).

PR #4405, opened 2025-08-28
Feature: Add `FixedSizeListArray` and `FixedSizeListVTable`

connortsui20:

Implements `FixedSizeListArray` and `FixedSizeListVTable` and adds `Canonical::FixedSizeList`.

Additonally adds 33 tests for fixed size list, mostly because we allow the possibility for degenerate fixed-size list array where each list scalar has size 0 (and thus no actual data is stored unless the validity is an array)

### Notes

I decided to split out the different `VTable` traits into separate modules under a `vtable` folder. I know that this is different from the rest of the codebase, but I would argue that this makes much more sense than having a subset of these files and add different vtable implementations under different file names (for example, `serde.rs` having both `SerdeVTable` implementation as well as `VisitorVTable`). I would go as far to say that we should just refactor every array module in `vortex-array` to split each trait implementation into a different file.


vortex-array/src/arrays/fixed_size_list/tests/basic.rs
```diff
@@ -0,0 +1,169 @@
...
+        let first = float_fsl.scalar_at(0);
+        assert_eq!(
+            first,
+            Scalar::fixed_size_list(
+                Arc::new(PType::F32.into()),
+                vec![1.1f32.into(), 2.2f32.into()],
+                Nullability::NonNullable,
+            )
+        );
+    }
```
robert3005: make this 2 tests or just use rstest?

vortex-array/src/arrays/fixed_size_list/vtable/serde.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::ProstMetadata;
+use crate::arrays::FixedSizeListEncoding;
+use crate::serde::ArrayChildren;
+use crate::validity::Validity;
+use crate::vtable::SerdeVTable;
+
+#[derive(Clone, prost::Message)]
+pub struct FixedSizeListMetadata {
+    #[prost(uint64, tag = "1")]
+    len: u64,
```
robert3005: Where is this used?

robert3005 (changes requested):
I think there's quite a lot of short files here. Maybe we combine all the vtables into one since they're trivial? For tests that do multiple things I would split them up and optionally use rstest depending on the complexity of the setup, i.e. it's nicer to get independent failures from the runner.

I am not sure you need any array specific metadata since everything is in the dtype?

vortex-btrblocks/src/lib.rs
```diff
@@ -335,6 +335,7 @@ impl BtrBlocksCompressor {
                 )?
                 .into_array())
             }
+            Canonical::FixedSizeList(_) => todo!("TODO(connor)[FixedSizeList]"),
```
a10y: For now just return the original array?

vortex-array/src/arrays/fixed_size_list/vtable/serde.rs
connortsui20: Removed now

connortsui20:
> I think there's quite a lot of short files here. Maybe we combine all the vtables into one since they're trivial?

I think I still prefer breaking things up into separate files, but I am also okay with having them all in 1 file. My biggest issue is doing something in between these 2 things, where we have 2-3 files with 5-6 implementations: that is **very** confusing when trying to read through the code.

vortex-array/src/arrays/fixed_size_list/tests/basic.rs
connortsui20: I just removed it since I basically do the same thing above and in other tests. I actually intended to remove it before but I forgot to.

vortex-array/src/canonical.rs
```diff
@@ -69,11 +69,11 @@ pub enum Canonical {
     Bool(BoolArray),
     Primitive(PrimitiveArray),
     Decimal(DecimalArray),
-    Struct(StructArray),
+    VarBinView(VarBinViewArray),
     // TODO(joe): maybe this should be a ListView, however this will be annoying in spiral
     List(ListArray),
-    // TODO(connor)[FixedSizeList]
-    VarBinView(VarBinViewArray),
+    FixedSizeList(FixedSizeListArray),
```
a10y, line 76: add this to the doc comment section as well, in the bulleted list

vortex-array/src/arrays/fixed_size_list/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl FixedSizeListArray {
+    /// Creates a new `FixedSizeListArray`. This is simply a wrapper around [`try_new()`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the inputs are invalid. See
+    ///
+    /// [`try_new()`]: Self::try_new
+    pub fn new(elements: ArrayRef, list_size: u32, validity: Validity, len: usize) -> Self {
```
a10y, line 67: We should also add a new_unchecked constructor that you use for the implementations of the operations. We maintain the invariants in our implementation and don't need to check them again on construction

vortex-array/src/arrays/fixed_size_list/vtable/operations.rs
```diff
@@ -0,0 +1,47 @@
...
+        debug_assert!(
+            start <= end && end <= array.len(),
+            "slice [{start}..{end}) out of bounds: then len is {}",
+            array.len()
+        );
+
+        let new_len = end - start;
+        let list_size = array.list_size() as usize;
+
+        FixedSizeListArray::new(
```
a10y: here's an example of where we should use `new_unchecked` instead, since we're maintaining the invariant of slicing the elements child ourselves

a10y:
I think you'll want to [add a branch](https://github.com/vortex-data/vortex/blob/8b8c9ee3eef16719b1d91c8d35ce84f8c1451bc8/vortex-array/src/arrow/compute/to_arrow/canonical.rs#L180) in the implementation of ToArrowCanonical as well, but I guess you can save that for a FLUP if you prefer

vortex-btrblocks/src/lib.rs
connortsui20: Actually I guess I have a question about that, if the elements are already compressed then technically we are done right? It's not like we can compress the list size (whereas in list arrays we can further compress the offsets). If that's true then I'll just implement this now in this PR

connortsui20:
```rust
Canonical::FixedSizeList(fsl_array) => {
    let compressed_elems = self.compress(fsl_array.elements())?;

    Ok(FixedSizeListArray::try_new(
        compressed_elems,
        fsl_array.list_size(),
        fsl_array.validity().clone(),
        fsl_array.len(),
    )?
    .into_array())
}
```

vortex-array/src/arrays/fixed_size_list/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+
+        Ok(Self::new_unchecked(elements, list_size, validity, len))
+    }
+
+    /// Creates a new `FixedSizeListArray`, assuming that the caller has validated the inputs.
+    ///
+    /// See [`try_new()`] for more details on what the validity requirements are.
+    ///
+    /// [`try_new()`]: Self::try_new
+    pub fn new_unchecked(
```
a10y: this should be unsafe, and callers should include a `// SAFETY` comment describing how they know the preconditions for FSLArray are validated before calling this

connortsui20: Do we use `unsafe` for things that aren't memory unsafe and are also just logically incorrect?

From my understanding if someone gives incorrect inputs, the only thing that can go wrong is slicing the child array, and it should be on the child array to do bounds checking regardless.

Or is there other `unsafe` hidden here that I don't know about?

a10y: I think people on the team have different opinions about this. In my view, a properly constructed array is a type and it should maintain some contract with other Rust code that uses it. That means that all of its internal pointers are valid, its validity is properly sized based on its expected len, etc.

The thing that I tend to compare this to is something like `str` or `NonZero<T>`. Constructing a non-utf8 `str` or a zero `NonZeroUsize` is not technically memory unsafe (not immediately anyway) but it breaks the type contract. Consumers of those APIs understand that getting one of those requires some validation, and then you can treat them as having those qualities. For NonZero the compiler does some niche optimization, and for str, anyone can assume the thing it points to is valid UTF8.

connortsui20: Okay that makes sense. I will say though that the main reason I thought this was due to looking at the code already in vortex right now, and I think that "unchecked" behavior is not super consistent (I guess most notable is the transmutes associated with assuming the dtype is always correct)

a10y: All of the array types should have an `unsafe new_unchecked(...)` constructor. Do you have an example of the transmutes you're referring to?

connortsui20: No I meant looking at some of the other unsafe code elsewhere in the codebase (I believe there's a macro generated impl that uses unsafe transmute, and also there was something else last week that I improved the safety comment on that sort of just assumed that the  passed in dtype was correct - that was technically a `pub(crate) fn` but imo was still unsafe even though it wasn't marked)

connortsui20: Also I don't think list has a new unchecked constructor, should that be added?

vortex-btrblocks/src/lib.rs
```diff
@@ -335,6 +337,17 @@ impl BtrBlocksCompressor {
                 )?
                 .into_array())
             }
+            Canonical::FixedSizeList(fsl_array) => {
+                let compressed_elems = self.compress(fsl_array.elements())?;
+
+                Ok(FixedSizeListArray::try_new(
```
robert3005, line 343: this should also be new_unchecked

robert3005 (approved):

connortsui20:
CC @robert3005 @a10y 

So I read through the other canonical arrays to check for `new_unchecked` methods, and they are wildly inconsistent:

- `bool`: No `new_unchecked` even though there is a `try_new`
- `chunked`: `pub unsafe fn new_unchecked`
- `datetime`: Panicking `new` (and no `try_new`; maybe this is okay since it is just checking the time unit but this is still inconsistent)
- `decimal`: No `new_unchecked` even though there is a `try_new`
- `list`: No `new_unchecked` even though there is a `try_new`
- `primitive`: No `new_unchecked` NOR a `try_new`, which is terrible because it only has a `new` that panics if the inputs are invalid
- `struct`: `pub(crate) fn new_unchecked` (this isn't `unsafe`! A function being private does not make it safe!)
- `varbin`: No `new_unchecked` even though there is a `try_new`
- `varbinview`: `pub unsafe fn new_unchecked`

There are only a few that "make sense" to me:

- `constant`: No `new_unchecked` (this makes sense since it is infallible and there is no `try_new`)
- `extension`: No `new_unchecked` (this makes sense since it (seems to be) infallible and there is no `try_new`)
- `null`: No `new_unchecked` (this makes sense since it is infallible and there is no `try_new`)

I'm going to address all of these in a followup PR because this is not great.

Edit: Issue #4420

vortex-btrblocks/src/lib.rs
connortsui20: See https://github.com/vortex-data/vortex/pull/4405#issuecomment-3236232355

vortex-array/src/arrays/fixed_size_list/mod.rs
connortsui20: See https://github.com/vortex-data/vortex/pull/4405#issuecomment-3236232355

PR #4407, opened 2025-08-28
feat: Add BitOrd and Concat for `Mask`

blaginin:
Missing there in Spiral, feels like those would be useful for others

### DIFF
--- a/vortex-mask/src/lib.rs
+++ b/vortex-mask/src/lib.rs
@@ -603,6 +603,32 @@ impl Mask {
             }
         }
     }
+
+    /// Concatenate multiple masks together into a single mask.
+    pub fn concat<'a>(masks: impl Iterator<Item = &'a Self>) -> VortexResult<Self> {
+        let masks: Vec<_> = masks.collect();
+        let len = masks.iter().map(|t| t.len()).sum();
+
+        if masks.iter().all(|t| t.all_true()) {
+            return Ok(Mask::AllTrue(len));
+        }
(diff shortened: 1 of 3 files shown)


vortex-mask/src/lib.rs
```diff
@@ -603,6 +603,32 @@ impl Mask {
             }
         }
     }
+
+    /// Concatenate multiple masks together into a single mask.
+    pub fn concat<'a>(masks: impl Iterator<Item = &'a Self>) -> VortexResult<Self> {
```
lwwmanning, line 608: would be good to add test for `concat`

vortex-mask/src/tests.rs
```diff
@@ -670,3 +670,51 @@ fn test_intersection_indices(
         AllOr::None | AllOr::All => panic!("Unexpected result for intersection"),
     }
 }
+
+// Zip operation tests
```
lwwmanning: oh, you did! but they're just called `zip` here, guessing this is a rename you need to finish :)

vortex-mask/src/lib.rs
blaginin: they are in `vortex-mask/src/tests.rs` 🙂

vortex-mask/src/tests.rs
blaginin: yes! thank you

lwwmanning (approved):

PR #4408, opened 2025-08-28
alignment is cheap?

danking:


vortex-io/src/compio.rs
```diff
@@ -54,7 +54,12 @@ impl VortexReadAt for File {
         alignment: Alignment,
     ) -> io::Result<ByteBuffer> {
         let len = usize::try_from(range.end - range.start).vortex_expect("range too big for usize");
-        let buffer = ByteBufferMut::with_capacity_aligned(len, alignment);
+        let better_alignment = if alignment.is_aligned_to(Alignment::new(8)) {
+            Alignment::new(8)
+        } else {
+            alignment
+        };
+        let buffer = ByteBufferMut::with_capacity_aligned(len, better_alignment);
```
a10y: some segments need to be aligned to 16 (e.g. the views buffers in VarbinView) so I'm not sure this is ok to do?

robert3005:
may I introduce you to https://doc.rust-lang.org/std/alloc/fn.alloc.html?

vortex-io/src/compio.rs
danking: I had the direction flipped. I also expanded it so we try to go to 16 when possible.

danking:
A minor part of the profile I saw was spent copying unaligned segments from the initial read. If we're allocating fresh memory anyway, why not allocate it with a large alignment?

danking:
This is fraught, trying something more targeted. https://github.com/vortex-data/vortex/pull/4418

PR #4409, opened 2025-08-28
chore[duckdb]: lifetime to object cache

joseph-isaacs:


vortex-duckdb/src/duckdb/client_context.rs
```diff
@@ -15,13 +15,13 @@ wrapper!(
 
 impl ClientContext {
     /// Get the object cache for this client context.
-    pub fn object_cache(&self) -> ObjectCache {
+    pub fn object_cache(&self) -> ObjectCacheRef<'static> {
```
joseph-isaacs, line 18: We need this until we can better scope table scans and tokio dispatch

robert3005 (approved):

PR #4410, opened 2025-08-28
Chore: Rename `TimeUnit` variants to be more descriptive

connortsui20:
I feel like there's no good reason to use abbreviations in general, as it makes things harder to read, and also for this case there's no places where it makes the code cleaner.


a10y (approved):
Agree abbreviations are unnecessary here, I was meaning to do this but kept forgetting

PR #4412, opened 2025-08-28
feature: ValidityVTable functions are non fallible

robert3005:
Since we made ArrayOperations non fallible these were only ever fallible because canonical is fallible but since we validate all the arguments these shouldn't be fallible anymore


a10y (approved):
nice

PR #4414, opened 2025-08-28
Feature: Add optional git hooks scripts

connortsui20:
Before I wrote this git hook, I think I forgot to ensure REUSE compliance and DCO (though I've now just aliased `git commit` to `git commit --signoff` for DCO).

Regardless, this git hook ensures a few things:

- Checks that every commit we push has a `--signoff` (this does mean that rebasing will fail if the bot has not signed off, I just use `git push --force-with-lease --no-verify` in that case)
- Checks REUSE compliance (just checks if `reuse` passes)
- `cargo +nightly fmt --all` passes
- `cargo clippy --all-targets --all-features` passes (this can take a while, so you can just comment that section out)

To use this hook, you just copy and paste the file into `.git/hooks`

EDIT: I have separated the hook into 2: `pre-commit` and `pre-push`. `pre-commit` ensures REUSE compliance with the headers and formatting. `pre-push` runs DCO compliance checks on all commits and runs clippy.

I also talked with Dima about if we want people to symlink this by default. I was unable to get symlinking to work on my computer for some reason so I'm leaving it out of this PR. 

Commit output:
```
📜 Checking REUSE compliance
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
✅ REUSE compliant - all files have proper license headers

🎨 Checking code formatting
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
✅ Code formatting looks good

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
🎉 All pre-commit checks passed! Proceeding with push...
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

Push output:
```
🚀 Pre-push checks starting...
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
ℹ️  Checking 1 new commit(s)

✍️  Checking commit sign-offs (DCO compliance)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
✅ All 1 commit(s) properly signed off

📝 Running Clippy linter
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
✅ Clippy found no issues

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
🎉 All pre-push checks passed! Proceeding with push...
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

### DIFF
--- a/scripts/pre-push
+++ b/scripts/pre-push
@@ -1,0 +1,284 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+#
+# Git Pre-Push Hook for Vortex
+# Validates commits being pushed to ensure code quality, REUSE compliance, and DCO.
+#
+# If you want to use this hook, simply copy and paste the file into a new file `.git/hooks/pre-push`
+# and everything should work correctly. If you are in the project root, you can use this command:
+#
+# ```sh
+# cp scripts/pre-push .git/hooks/
+# ```
+# 
+# Note that `cargo clippy --all-warnings --all-features` can take a long time if you haven't cached
+# a build recently. You can simply comment out that section (CHECK 4) if you don't want it.
+
+set -Euo pipefail
+
+# Configuration.
+MAIN_BRANCH="develop"  # The main branch name for vortex.
+
+# Color codes for terminal output.
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+BOLD='\033[1m'
+RESET='\033[0m'
+
+# Output helper functions.
+print_error() {
+    echo -e "${RED}$1${RESET}"
+}
+
+print_success() {
+    echo -e "${GREEN}$1${RESET}"
+}
+
+print_warning() {
+    echo -e "${YELLOW}$1${RESET}"
+}
+
+print_info() {
+    echo -e "${BLUE}$1${RESET}"
+}
+
+print_header() {
+    echo -e "\n${BOLD}$1${RESET}"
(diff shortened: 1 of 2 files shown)


scripts/pre-push
```diff
@@ -0,0 +1,280 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+#
+# Git Pre-Push Hook for Vortex
+# Validates commits being pushed to ensure code quality, REUSE compliance, and DCO.
+#
+# If you want to use this hook, simply copy and paste the file into a new file `.git/hooks/pre-push`
```
blaginin, line 8: 1 - should this be ln -s (so we can update the file later not forcing users to update)
2 - can you give a copy paste command so it'll be a bit easier to setup?

danking (approved):
pre-commit / pre-push hooks are great.

In the past, I've used https://pre-commit.com/; what do you think of managing pre-push hooks that way versus hand-writing them?

joseph-isaacs:
this is not on by default is it?
I want to be able to push a commit without waiting for clippy

robert3005:
I think you generally want to push things to ci and not block on things passing locally.

connortsui20:
Yeah so since everyone likely has their own preferences, the default is that this checks as many things as possible and you can easily comment out checks you don't want to run yourself, as you need to copy and paste these files into `.git/hooks/` yourself anyways.

So personally I want to run as many checks as possible before pushing to CI, but for people who do not want to do that, when they copy and paste it over they can comment it out (or just not copy and paste it at all)

robert3005 (approved):

PR #4415, opened 2025-08-28
feat: teach patches a fast path for size one slice

danking:


delta003 (approved):

robert3005:
Why not scalar_at and return a constant array?

danking:
I'm also trying to eliminate Scalars on the fast path.

robert3005:
I think though that constant array would be better than single element sparse array

PR #4416, opened 2025-08-28
feat: teach Patches to fast-path search sorted on primitive array

danking:


danking:
Maybe search sorted should be an ArrayOperation instead?

vortex-array/src/patches.rs
```diff
@@ -224,6 +224,26 @@ impl Patches {
 
     /// Return the insertion point of `index` in the [Self::indices].
     pub fn search_index(&self, index: usize) -> SearchResult {
+        if self.indices.is_canonical() {
+            let primitive = self
+                .indices
+                .to_canonical()
```
gatesn: I think you can go directly inro_primitive?

gatesn (approved):

vortex-array/src/patches.rs
danking: Done.

PR #4417, opened 2025-08-28
feat: teach DashMap to use a faster hasher

danking:


robert3005 (approved):
Thanks! We should check that in cases of integer keys we are using fxhash (while default should be for anything else)

AdamGS (approved):
sure why not

PR #4418, opened 2025-08-28
feat: align the initial read

danking:


vortex-file/src/generic.rs
```diff
@@ -255,7 +255,7 @@ impl VortexOpenOptions<GenericVortexFile> {
         Ok(self
             .options
             .io_dispatcher
-            .dispatch(move || async move { read.read_byte_range(range, Alignment::none()).await })?
+            .dispatch(move || async move { read.read_byte_range(range, Alignment::new(8)).await })?
```
robert3005: why is this 8 and not something like 64 or something else?

danking: We don't have any 64-byte types but 16-byte alignment is reasonable. I ran into some issues at first with 16 but I can try it again.

danking:
No effect on random access :|

danking:
I suspect this is not worth merging in current state.

PR #4419, opened 2025-08-28
feat: no string allocation, no boxes, no futures on the hot read path of Scan

danking:
Two improvements:
1. Do not repeatedly allocate strings for the names of Struct and Chunked children. (this is minor)
2. No boxing and no futures in the read path. One coroutine loads the rows sequentially. This coroutine could, in theory, be executed concurrently with other coroutines reading ahead to future samples.

filter_row_mask is just copied. split_exec_direct is a copy of split_exec with the boxed futures removed. I think I can better abstract this though.

### DIFF
--- a/vortex-scan/src/lib.rs
+++ b/vortex-scan/src/lib.rs
@@ -398,7 +398,7 @@ impl<A: 'static + Send> RepeatedScan<A> {
             }
 
             if let Some(maybe_value) =
-                split_exec(ctx.clone(), range.clone(), limit.as_mut())?.await?
+                split_exec_direct(ctx.clone(), range.clone(), limit.as_mut()).await?
             {
                 vec.push(maybe_value)
             }
--- a/vortex-scan/src/lib.rs
+++ b/vortex-scan/src/lib.rs
@@ -374,7 +374,7 @@ impl<A: 'static + Send> RepeatedScan<A> {
         Ok(split_tasks)
     }
 
-    pub async fn build_single_threaded(
+    pub async fn execute_single_threaded(
         &self,
         row_range: Option<Range<u64>>,
     ) -> VortexResult<Vec<A>> {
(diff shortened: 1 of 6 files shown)


vortex-scan/src/lib.rs
```diff
@@ -330,5 +316,168 @@ fn to_field_mask(field: FieldName) -> FieldMask {
...
+                } else {
+                    Some(split_exec(ctx.clone(), split_range.clone(), limit.as_mut()))
+                }
+            })
+            .try_collect()?;
+
+        Ok(split_tasks)
+    }
+
+    pub async fn build_single_threaded(
```
delta003: This avoids futures completely, right? Maybe call it blocking? But does this actually have an impact comparing to setting concurrency 1?

delta003 (commented):
I can't quite figure out what the core improvement here is. Is it the blocking function and other changes (in tasks.rs) that are to make that function possible? There are no actual logic changes to e.g. filter_row_mask? Mind writing a PR description so I can take a better look?

vortex-scan/src/lib.rs
danking: I originally called it blocking, but it _is_ async, you could run it concurrently with itself.

Yes, it avoids futures entirely. There's just one coroutine that does all the reading rather than multiple coroutines communicating their results through futures.

danking: And, yeah, this is the bulk of the improvement from ~2 microseconds to ~1 microseconds. Unclear to me that 1 microsecond per sample matters (even an SSD's latency is ~10us), so we could dump this PR and look more at other improvements.

robert3005:
I think the bigger change here would be making names a feature flag and not enabling it by default - AFAIK only TUI cares

danking:
There's something here but I think it's not necessary to achieve our stated goals for the Vortex DataLoader.

PR #4425, opened 2025-08-29
chore: `GetItem` to apply parent nullability

### LINKED ISSUE #4424: `get_item` to apply struct validity
This:
```rust
let expr = get_item("a", root());

let array = StructArray::try_new(
    FieldNames::from(["a"]),
    vec![
        PrimitiveArray::from_iter([1]).to_array()
    ],
    1,
    Validity::AllInvalid
)?.to_array();

let res = expr.evaluate(&Scope::new(array))?;
println!("{}", res.display_values());
```

Will print
```
[1i32]
```

It's very easy to forget to apply parent level nullability afterwards - and in fact we don't do that in [`remove_merge`](https://github.com/vortex-data/vortex/blob/c24b371d8704085977958d281f70e9f013ae7410/vortex-expr/src/transform/remove_merge.rs#L15). 

Let's make `get_item` to always apply parent nullability?

blaginin:
Closes https://github.com/vortex-data/vortex/issues/4424

### DIFF
--- a/vortex-expr/src/exprs/get_item.rs
+++ b/vortex-expr/src/exprs/get_item.rs
@@ -7,7 +7,6 @@ use std::ops::Not;
 
 use vortex_array::compute::{cast, mask};
 use vortex_array::stats::Stat;
-use vortex_array::vtable::ValidityHelper;
 use vortex_array::{Array, ArrayRef, DeserializeMetadata, ProstMetadata, ToCanonical};
 use vortex_dtype::{DType, FieldName, FieldPath, Nullability};
 use vortex_error::{VortexResult, vortex_bail, vortex_err};
--- a/vortex-expr/src/exprs/get_item.rs
+++ b/vortex-expr/src/exprs/get_item.rs
@@ -3,10 +3,12 @@
 
 use std::fmt::{Debug, Formatter};
 use std::hash::Hash;
-
+use std::ops::Not;
+use vortex_array::compute::{cast, mask};
 use vortex_array::stats::Stat;
-use vortex_array::{ArrayRef, DeserializeMetadata, ProstMetadata, ToCanonical};
-use vortex_dtype::{DType, FieldName, FieldPath};
+use vortex_array::vtable::ValidityHelper;
+use vortex_array::{Array, ArrayRef, DeserializeMetadata, ProstMetadata, ToCanonical};
+use vortex_dtype::{DType, FieldName, FieldPath, Nullability};
 use vortex_error::{VortexResult, vortex_bail, vortex_err};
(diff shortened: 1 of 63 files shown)


vortex-expr/src/exprs/get_item.rs
```diff
@@ -83,18 +85,26 @@ impl VTable for GetItemVTable {
...
-            .cloned()
+        let input = expr.child.unchecked_evaluate(scope)?.to_struct()?;
+        let field = input.field_by_name(expr.field()).cloned()?;
+
+        match input.dtype().nullability() {
+            Nullability::NonNullable => Ok(field),
+            Nullability::Nullable => {
+                let result = mask(&field, &input.validity_mask().not())?;
+                // if mask is empty, it will return the original array without changing its nullability,
+                // but we already promised a nullable output...
```
danking: I think this is a bug. The ComputeFnVTable is pretty explicit about what should happen when the mask is all false or all true:
```rust
        if matches!(mask, Mask::AllFalse(_)) {
            // Fast-path for empty mask
            return Ok(cast(array, &array.dtype().as_nullable())?.into());
        }

        if matches!(mask, Mask::AllTrue(_)) {
            // Fast-path for full mask.
            return Ok(ConstantArray::new(
                Scalar::null(array.dtype().clone().as_nullable()),
                array.len(),
            )
            .into_array()
            .into());
        }
```

vortex-expr/src/exprs/get_item.rs
```diff
@@ -83,18 +85,26 @@ impl VTable for GetItemVTable {
...
-            .to_struct()?
-            .field_by_name(expr.field())
-            .cloned()
+        let input = expr.child.unchecked_evaluate(scope)?.to_struct()?;
+        let field = input.field_by_name(expr.field()).cloned()?;
+
+        match input.dtype().nullability() {
+            Nullability::NonNullable => Ok(field),
+            Nullability::Nullable => {
+                let result = mask(&field, &input.validity_mask().not())?;
```
danking: ugh this sucks. I wish we had, like, intersect_validity or something.

danking (changes requested):

vortex-expr/src/exprs/get_item.rs
blaginin: ooo thank you!

vortex-expr/src/exprs/get_item.rs
blaginin: agreed, https://github.com/vortex-data/vortex/issues/4438

danking (approved):

PR #4426, opened 2025-08-29
Add RepeatedScan in vortex-data python package

delta003:


vortex-python/src/file.rs
```diff
@@ -38,8 +39,7 @@ pub(crate) fn init(py: Python, parent: &Bound<PyModule>) -> PyResult<()> {
 #[pyfunction]
 pub fn open(path: &str) -> PyResult<PyVortexFile> {
     let vxf = VortexOpenOptions::file()
-        // TODO(ngates): use a globally shared segment cache for all files
-        .with_segment_cache(Arc::new(MokaSegmentCache::new(256 << 20)))
+        .without_segment_cache()
```
onursatici: should we make this configurable, maybe with a default arg?

vortex-python/python/vortex/file.py
```diff
@@ -144,6 +145,29 @@ def scan(
         """
         return self._file.scan(projection, expr=expr, indices=indices, batch_size=batch_size)
 
+    def prepare(
```
onursatici: I know this mirrors the underlying prepare method, but maybe something like `to_repeated_scan` is more suitable for this?

vortex-python/src/file.rs
delta003: I think so. I couldn't figure out the right API... segment_cache_size_bytes: usize defaulting to 256 << 20? So you can set zero to disable

onursatici: yea I think so and maybe an Option and None disables the cache. But if then we want to have a global cache then it will be a bit weird to have it configurable here

delta003: For the similar reasons, I've decided on `without_segment_cache` which stays even in the case of global cache

vortex-python/python/vortex/file.py
delta003: Much better! And consistent with other methods in that class

onursatici (approved):

vortex-python/python/vortex/file.py
```diff
@@ -144,6 +147,29 @@ def scan(
...
+
+        Parameters
+        ----------
+        projection : :class:`vortex.Expr` | list[str] | None
+            The projection expression to read, or else read all columns.
+        expr : :class:`vortex.Expr` | None
+            The predicate used to filter rows. The filter columns do not need to be in the projection.
+        indices : :class:`vortex.Array` | None
+            The indices of the rows to read. Must be sorted and non-null.
+        batch_size : :class:`int` | None
```
danking, line 168: fwiw, the :class:`xxx` syntax is not necessary. Our Sphinx setup somehow automatically parses these parameter lines and converts things like `int` `Expr` and `list[str]` into the right hyper links.

vortex-python/python/vortex/scan.py
```diff
@@ -0,0 +1,99 @@
...
+
+from ._lib import scan as _scan  # pyright: ignore[reportMissingModuleSource]
+from ._lib.iter import ArrayIterator  # pyright: ignore[reportMissingModuleSource]
+from ._lib.scalar import Scalar  # pyright: ignore[reportMissingModuleSource]
+
+
+@final
+class RepeatedScan:
+    """
+    A prepared scan that is optimized fro repeated execution.
```
danking:
```suggestion
    A prepared scan that is optimized for repeated execution.
```

vortex-python/python/vortex/scan.py
```diff
@@ -0,0 +1,99 @@
...
+    def execute(
+        self,
+        *,
+        row_range: tuple[int, int] | None = None,
+    ) -> ArrayIterator:
+        """Execute the scan returning a :class:`vortex.ArrayIterator`.
+
+        Parameters
+        ----------
+        row_range : tuple[int, int] | slice | None
```
danking: This is inconsistent with the types.

vortex-python/python/vortex/file.py
delta003: Copilot writes it cause it's consistent with the rest of the file 🤷

danking (changes requested):

vortex-python/python/vortex/scan.py
delta003: Fixed. Went back and forth on supporting slice only to realize it's just annoying. Maybe in the future if we add __getitem__

danking (approved):

PR #4429, opened 2025-08-29
test[mask] cannot convert nullable bool array to mask

joseph-isaacs:


vortex-array/src/compute/filter.rs
```diff
@@ -194,6 +194,8 @@ impl<V: VTable + FilterKernel> Kernel for FilterKernelAdapter<V> {
 
 impl From<&BoolArray> for Mask {
     fn from(array: &BoolArray) -> Self {
+        // This should in fact be non-nullable.
+        assert!(array.all_valid());
```
blaginin: just fyi, in spiral i rely on this conversion to work even with nullable arrays, because of

`Mask::Values(validity) => validity.boolean_buffer().bitand(array.boolean_buffer())`

joseph-isaacs: why? This is likely incorrect apart from at the top level of a filter. In which case we should use a specific coerce/convert

blaginin: imagine you have some array (or arbitrary type) containing nulls. you apply some filter condition and get a binary mask with nulls. then you'd want to get a mask of values that pass the condition - for this you'll need to cast nullable bool to mask

joseph-isaacs: I understand why, but that is a very specific operation which should not be inside a from method

joseph-isaacs: filters mapping null to false only works at the top of a filter, this operation should be explicit

blaginin (approved):
Not sure why you got a performance decrease - I think the code is the same just moved? Lgtm

PR #4435, opened 2025-08-29
Chore: Validate `Scalar::new` for null and clean up `DecimalScalar`

### LINKED ISSUE #4378: Scalar::new should validate value for dtype, add Scalar::new_unchecked
Currently we perform lazy validation of scalars. While this is useful it's very errorprone, we would benefit from checked variant

connortsui20:
Progress towards #4378 

~~I don't think that it makes a lot of sense to add `new_unchecked` given that the only check we really need is `!dtype.is_nullable()` (which is cheap), and by making the fields private we can guarantee this invariant.~~

Edit: In the future the `Scalar::new` should validate that the dtype matches the value that is passed in.

Also cleans up the `decimal` module in `vortex-scalar`.


danking (approved):
LGTM.

Agreed something is wrong with append_zeros. Agreed that the zero element of list<t> should be `[]`. Send me a separate PR to fix that?

PR #4444, opened 2025-08-30
chore: Layout names are behind a compile flag

robert3005:

### DIFF
--- a/vortex/Cargo.toml
+++ b/vortex/Cargo.toml
@@ -67,6 +67,7 @@ python = ["vortex-error/python"]
 tokio = ["vortex-file/tokio", "vortex-scan/tokio"]
 zstd = ["dep:vortex-zstd", "vortex-file/zstd", "vortex-layout/zstd"]
 pretty = ["vortex-array/table-display"]
+layout_names = ["vortex-file/layout_names", "vortex-layout/layout_names", "vortex-scan/layout_names"]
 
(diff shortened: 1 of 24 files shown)


vortex/Cargo.toml
```diff
@@ -67,6 +67,11 @@ python = ["vortex-error/python"]
 tokio = ["vortex-file/tokio", "vortex-scan/tokio"]
 zstd = ["dep:vortex-zstd", "vortex-file/zstd", "vortex-layout/zstd"]
 pretty = ["vortex-array/table-display"]
+layout_names = [
+    "vortex-file/layout_names",
+    "vortex-layout/layout_names",
+    "vortex-scan/layout_names",
+]
```
AdamGS, line 74: Should it be part of the "pretty" feature? maybe it should be re-named to "verbose" or something similar

robert3005: I am happy to use a different name but this is not related to pretty. It's for debuggability when you want to add debug logging to your scan

gatesn (changes requested):
Is there no way to either make the names fast (i.e. push down FieldPath stack thing?), or else properly figure out improving the scan startup performance to cache this stuff, or else remove them altogether?

robert3005:
I thought this was the simplest thing to do. Happy to drop this.

PR #4445, opened 2025-08-30
Feature: Add `eq_with_nullability_subset ` and `eq_with_nullability_superset ` and validate `extend_from_array`

connortsui20:
This will be helpful in my refactoring of the `builders` module in `vortex-array` (in draft right now: https://github.com/vortex-data/vortex/pull/4440)

Note that `extend_from_array` is called in many non-test places yet is not validated at all. Given that this is a public method, this is a bug waiting to happen...

I have an interesting thought though: should a `DType::Primitive(U64, nullable)` be a superset of `DType::Primitive(U16, non-nullable)`?

In theory, we should be able to extend an array builder of nullable `U64` with an array of non-nullable `U8`.

I'll leave that to a separate PR since this idea might have some other implications.

### DIFF
--- a/vortex-array/src/arrays/arbitrary.rs
+++ b/vortex-array/src/arrays/arbitrary.rs
@@ -10,7 +10,7 @@ use builders::ListBuilder;
 use vortex_buffer::Buffer;
 use vortex_dtype::{DType, NativePType, Nullability, PType};
 use vortex_error::{VortexExpect, VortexUnwrap};
-use vortex_scalar::arbitrary::{random_decimal, random_scalar};
+use vortex_scalar::arbitrary::random_scalar;
 use vortex_scalar::{Scalar, match_each_decimal_value_type};
 
 use super::{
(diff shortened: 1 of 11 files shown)


robert3005 (changes requested):
Instead of introducing type hierarchy there's maybe a simpler thing to do here where you check for nullability. I don't see how a builder can optimize the append any differently than the caller. This is worth revisiting imho when we add a downcast array.

connortsui20:
That makes sense, I'll make an issue for it. However, this PR doesn't actually introduce a type hierarchy and considering we don't actually use type hierarchy anywhere else I feel like this is fine to merge as is. Or do you think the name needs to change?

robert3005:
I would change the name since people could assume that it does more than what it says. Otherwise I think this change is gtg

connortsui20:
Renamed to `eq_with_nullability_subset` and `eq_with_nullability_superset`

connortsui20:
Also I just realized I messed up the branch chain, none of the other PRs I made last night got merged to develop, they got merged into this branch

robert3005 (approved):

PR #4450, opened 2025-08-31
Chore: Make all `ArrayBuilder`s consistent

connortsui20:
This PR refactors array builders for consistent APIs (and mostly consistent implementations).

## Changes

This is really all of the changes from the other 4 PRs I made right before this...

- `append_value` -> type-specific names (`append_bool`, `append_decimal`, `append_list`)
- All builders now validate dtype compatibility in `extend_from_array`
- `is_superset_of`/`is_subset_of` for type checking
- Standardized method signatures across all builders
- Extracted common constants, improved docs
- Added `finish_into_specific` for all of the builders (this opens the door to implementing `ToCanonical` for every `ArrayBuilder`, avoiding an allocation)

Note that I avoided changing `VarBinViewBuilder` since there's a lot of code in there that I don't understand.

Also note that there are a lot of breaking changes, but to be honest, a bunch of `append_value`s that took completely different parameters is too confusing.

### DIFF
--- a/vortex-array/src/builders/decimal.rs
+++ b/vortex-array/src/builders/decimal.rs
@@ -13,7 +13,7 @@ use crate::arrays::DecimalArray;
 use crate::builders::{ArrayBuilder, DEFAULT_BUILDER_CAPACITY, LazyNullBufferBuilder};
 use crate::{Array, ArrayRef, IntoArray, ToCanonical};
 
-/// An [`ArrayBuilder`] for `Decimal` typed arrays.
+/// The builder for building a [`DecimalArray`].
 ///
 /// The output will be a new [`DecimalArray`] holding values of `T`. Any value that is a valid
 /// [decimal type][NativeDecimalType] can be appended to the builder and it will be immediately
--- a/vortex-array/src/builders/decimal.rs
+++ b/vortex-array/src/builders/decimal.rs
@@ -10,8 +10,7 @@ use vortex_mask::Mask;
 use vortex_scalar::{BigCast, NativeDecimalType, i256, match_each_decimal_value_type};
 
 use crate::arrays::DecimalArray;
-use crate::builders::lazy_validity_builder::LazyNullBufferBuilder;
-use crate::builders::{ArrayBuilder, DEFAULT_BUILDER_CAPACITY};
+use crate::builders::{ArrayBuilder, DEFAULT_BUILDER_CAPACITY, LazyNullBufferBuilder};
 use crate::{Array, ArrayRef, IntoArray, ToCanonical};
 
 /// An [`ArrayBuilder`] for `Decimal` typed arrays.
--- a/vortex-array/src/builders/decimal.rs
+++ b/vortex-array/src/builders/decimal.rs
@@ -103,7 +103,7 @@ impl DecimalBuilder {
     }
 
     /// Appends a decimal `value` to the builder.
-    pub fn append_value<V: NativeDecimalType>(&mut self, value: V) {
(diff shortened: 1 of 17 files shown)


robert3005 (changes requested):
I think append value is the consistent thing. Every builder has a same set of methods that take the primitive value type that the array stores. The inconsistency is in the nested which need to take some kind of IntoIterator type. Since those are all concrete types I don't quite see how renaming methods improves things

connortsui20:
To me it doesn't make sense that we have an `append_value` on each builder that takes in a completely different type... I think I would understand `append_value` if we always passed in a `ScalarValue`, but I think that where `append_scalar` is supposed to come in (which I think probably should not have a default implementation and rely on each builder to implement a more efficient version, but that is a separate issue).

robert3005:
to me having `append_primitive` implies that you might have something else like `append_bool` but you have a PrimitiveBuilder and it doesn't quite make sense to `append_bool` on that builder. But you know you have XBuilder which builds `XArray`. Having `append_primitive` is repetitive in my opinion.

vortex-array/src/builders/primitive.rs
```diff
@@ -35,26 +36,33 @@ impl<T: NativePType> PrimitiveBuilder<T> {
...
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the input is `None` and the builder is non-nullable.
+    pub(crate) fn append_option(&mut self, value: Option<T>) {
         match value {
             Some(value) => {
                 self.values.push(value);
-                self.nulls.append(true);
+                self.nulls.append_non_null();
```
robert3005: self.append_value?

vortex-array/src/builders/primitive.rs
```diff
@@ -35,26 +36,33 @@ impl<T: NativePType> PrimitiveBuilder<T> {
...
+        self.nulls.append_non_null();
     }
 
-    pub fn append_option(&mut self, value: Option<T>) {
+    /// Appends an optional primitive (representing a nullable primitive) to the builder.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the input is `None` and the builder is non-nullable.
+    pub(crate) fn append_option(&mut self, value: Option<T>) {
```
robert3005: why pub(crate)?

vortex-array/src/builders/decimal.rs
```diff
@@ -108,18 +107,6 @@ impl DecimalBuilder {
...
-    ///
-    /// # Panics
-    ///
-    /// This method will panic if the input is `None` and the builder is non-nullable.
-    pub fn append_option<V: NativeDecimalType>(&mut self, value: Option<V>) {
-        match value {
-            Some(value) => self.append_value(value),
-            None => self.append_null(),
-        }
-    }
```
robert3005, line 121: why delete this?

robert3005 (commented):
I think we want a append_value_opt?

vortex-array/src/builders/primitive.rs
connortsui20: I'm about to remove this in the next PR since we _only_ need this for the default implementation for `append_scalar` (which will be gone)

vortex-array/src/builders/decimal.rs
connortsui20: See above

vortex-array/src/builders/primitive.rs
connortsui20: I removed the `.append(true)` because when I first read this I though `true` meant we were appending a null but it turns out the parameter is `not_null` which is super confusing

robert3005 (approved):

PR #4452, opened 2025-09-01
Chore: `ArrayBuilder` refactor

connortsui20:
3 changes (in each commit):

- Fix the `append_null` semantics. Before, we allow non-nullable array builders to call `append_null`, and it only fails at the end when we finish the builder. This change makes it such that if a `non-nullable` array calls `append_null`, it panics
- Adds regression tests for the above
- Removes `ArrayBuilderExt` (which only had `append_scalar`) and merge into `ArrayBuilder`. There wasn't really any reason to have `append_scalar` so this just cleans things up.

### DIFF
--- a/vortex-array/src/builders/fixed_size_list.rs
+++ b/vortex-array/src/builders/fixed_size_list.rs
@@ -181,15 +181,14 @@ impl ArrayBuilder for FixedSizeListBuilder {
     /// - If elements are nullable, we use null values
     /// - If elements are non-nullable, we use zero values
     fn append_nulls(&mut self, n: usize) {
+        assert!(
+            self.dtype.is_nullable(),
+            "tried to append {n} nulls to a non-nullable array builder"
+        );
+
         let element_count = n * self.list_size() as usize;
 
-        // TODO(connor): `append_default()`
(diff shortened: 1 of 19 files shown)


vortex-array/src/builders/fixed_size_list.rs
```diff
@@ -181,15 +180,14 @@ impl ArrayBuilder for FixedSizeListBuilder {
     /// - If elements are nullable, we use null values
     /// - If elements are non-nullable, we use zero values
     fn append_nulls(&mut self, n: usize) {
-        let element_count = n * self.list_size() as usize;
+        assert!(
+            self.dtype.is_nullable(),
+            "tried to append {n} nulls to a non-nullable array builder"
+        );
```
joseph-isaacs: Shall we have a public `append_nulls` that call a `_append_nulls()` and do validation in `append_nulls`?

connortsui20: Actually I think if we're going to do that we might as well also do that for `extend_from_array` since they all do the similar validations

robert3005 (approved):

PR #4455, opened 2025-09-01
Feature: Add `append_option` method to every builder

connortsui20:
Also adds some documentation here and there.

### DIFF
--- a/vortex-array/src/builders/primitive.rs
+++ b/vortex-array/src/builders/primitive.rs
@@ -7,7 +7,6 @@ use std::ops::{Deref, DerefMut};
 
 use vortex_buffer::BufferMut;
 use vortex_dtype::{DType, NativePType, Nullability};
-use vortex_error::vortex_panic;
 use vortex_mask::Mask;
 
 use crate::arrays::PrimitiveArray;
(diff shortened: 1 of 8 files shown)


vortex-array/src/builders/primitive.rs
```diff
@@ -47,12 +47,15 @@ impl<T: NativePType> PrimitiveBuilder<T> {
...
+    /// Appends an optional primitive value to the builder.
+    ///
+    /// If the value is `Some`, it appends the primitive value. If the value is `None`, it appends a
+    /// null.
     ///
     /// # Panics
     ///
     /// This method will panic if the input is `None` and the builder is non-nullable.
-    pub fn append_option(&mut self, value: Option<T>) {
+    pub(crate) fn append_option(&mut self, value: Option<T>) {
```
robert3005: this shouldn't be crate?

vortex-scan/src/tests/loom.rs
```diff
@@ -532,6 +532,7 @@ fn test_steal_work_retry_semantics() {
     });
 }
 
+#[ignore] // Shhh...
```
robert3005: I wanted to reduce the runtime but loom really only lets you use env variables

robert3005: file an issue that we should likely delete this or lower its runtime

robert3005 (approved):

vortex-scan/src/tests/loom.rs
connortsui20: https://github.com/vortex-data/vortex/issues/4458

Issue #4220, opened 2025-08-12
UB in several crates that don't run miri


lwwmanning:
Claude & I gave a go at running miri across the whole repo (partitioned across 8 groups of crates) in #4219. Doing so uncovered several cases of undefined behavior in crates that we haven't run miri on for a long time.

### 1. Strict Provenance Violations
- **Affected groups**: 3, 7
- **Cause**: Using integer-to-pointer casts which are not allowed with `-Zmiri-strict-provenance`
- **Common in**: `once_cell` initialization code
- **Solution**: Either:
  - Remove `-Zmiri-strict-provenance` flag from MIRIFLAGS
  - Fix the code to avoid integer-to-pointer casts
  - Use `ptr::from_exposed_addr` instead

### 2. Borrow Stack Violations
- **Affected groups**: 2, 6
- **Cause**: Invalid memory access patterns, attempting to access memory with tags that don't exist in the borrow stack
- **Common in**: Async/futures code, unsafe buffer manipulations
- **Solution**: Fix unsafe code to properly maintain borrow stack invariants

### 1. Fix CI Configuration (Immediate)
```yaml
# In .github/workflows/ci.yml
env:
  # Remove -Zmiri-strict-provenance flag
  MIRIFLAGS: -Zmiri-symbolic-alignment-check -Zmiri-disable-isolation -Zmiri-env-forward=RUST_BACKTRACE
```

#### vortex-fastlanes Borrow Stack Violations
**Location**: `/vortex-array/src/builders/primitive.rs:192`
**Problem**: Invalid retag operations when creating slices from raw pointers

The error occurs in the `UninitRange::deref` implementation:
```rust
// Line 189: Creates a zero-size retag that doesn't establish proper tags
let start = self.builder.values.as_ptr(); 
// Line 192: Tries to create a slice with tags that don't exist
let dst = std::slice::from_raw_parts(start, self.len);
```

**Affected Tests** (all in vortex-fastlanes bitpacking):
- `test_offset_and_short_chunk_and_patches`
- `test_offset_and_short_chunk_with_chunks_between_and_patches`
- `test_simple_patches`
- `test_one_full_chunk_and_one_short_chunk_with_patches`
- `test_three_full_chunks_with_patches`
- `canonicalize_chunked_of_bitpacked`
- `take_nullable_with_nullables`

[...]

PR #4457, opened 2025-09-01
Chore: Clean up `PrimitiveBuilder` API

connortsui20:
Fixes part of #4220 (not too sure about the strict provenance violations yet)

The `Deref` implementation for `UninitRange` always returned slices from the start of the buffer instead of tracking the current position, causing incorrect memory access when multiple ranges were used sequentially. Additionally, `Deref` inconsistently sliced the entire builder's buffer while `DerefMut` only sliced the uninitialized memory up to capacity.

### Changes

- Removed `Deref`/`DerefMut` implementations entirely to prevent confusion
- Moved `append_mask()` from `PrimitiveBuilder` to `UninitRange` where it logically belongs
- Added explicit methods: `len`, `set_value()`, `copy_from_slice()`, `slice_uninit_mut()` to account for the fact that we don't have `Deref` to a slice anymore (which actually returned the wrong length slice)
- Fixed offset tracking to be relative to each range
- Added a bunch of safety comments for unsafe things
- Add regression tests

### DIFF
--- a/encodings/fastlanes/src/bitpacking/unpack_iter.rs
+++ b/encodings/fastlanes/src/bitpacking/unpack_iter.rs
@@ -128,19 +128,23 @@ impl<T: BitPacked> BitUnpackedChunks<T> {
         )
     }
 
-    /// Unpack full chunks into output range and return the last index that was written to
+    /// Unpack full chunks into output range and return the next local index to write to.
     pub fn decode_full_chunks_into(&mut self, output: &mut UninitRange<T>) -> usize {
         let first_chunk_is_sliced = self.first_chunk_is_sliced();
-        // If there's only one chunk and that chunk is sliced it has been handled already by `header` method
+        // If there's only one chunk and that chunk is sliced it has been handled already by
+        // `header` method
         if first_chunk_is_sliced && self.num_chunks == 1 {
-            return self.len;
+            // Return the length since the header already wrote everything.
+            return CHUNK_SIZE - self.offset;
         }
 
         let last_chunk_is_sliced = self.last_chunk_is_sliced();
         let full_chunks_range =
             (first_chunk_is_sliced as usize)..(self.num_chunks - last_chunk_is_sliced as usize);
 
-        let mut out_idx = if first_chunk_is_sliced {
+        // Track position relative to the start of the UninitRange.
+        let mut local_idx = if first_chunk_is_sliced {
+            // The header already wrote from 0 to (CHUNK_SIZE - self.offset).
             CHUNK_SIZE - self.offset
         } else {
             0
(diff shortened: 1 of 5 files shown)


vortex-array/src/builders/bool.rs
connortsui20, line 1: I somehow missed this on the previous PRs

encodings/fastlanes/src/bitpacking/unpack_iter.rs
connortsui20, line 1: Don't know if this makes it clearer or not tbh

encodings/fastlanes/src/bitpacking/unpack_iter.rs
```diff
@@ -152,13 +156,15 @@ impl<T: BitPacked> BitUnpackedChunks<T> {
             let chunk = &packed_slice[i * elems_per_chunk..][..elems_per_chunk];
 
             unsafe {
+                // SAFETY: We're about to initialize CHUNK_SIZE elements at local_idx.
+                let uninit_dst = output.slice_uninit_mut(local_idx, CHUNK_SIZE);
```
connortsui20, line 160: I'm not 100% that this is the correct safety guarantee but I think it's "good enough"

vortex-array/src/builders/primitive.rs
```diff
@@ -184,57 +176,361 @@ impl<T: NativePType> ArrayBuilder for PrimitiveBuilder<T> {
 
 /// A range of uninitialized values in the primitive builder that can be filled.
 pub struct UninitRange<'a, T> {
-    offset: usize,
     len: usize,
     builder: &'a mut PrimitiveBuilder<T>,
 }
 
-impl<T> Deref for UninitRange<'_, T> {
```
a10y, line 192: Yea this is the right move, these were added optimistically but were not useful after all, and indeed were wrong

vortex-array/src/builders/primitive.rs
```diff
@@ -184,57 +176,361 @@ impl<T: NativePType> ArrayBuilder for PrimitiveBuilder<T> {
...
+    /// Set a value at the given index within this range.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is out of bounds.
+    pub fn set_value(&mut self, index: usize, value: T) {
+        assert!(index < self.len, "index out of bounds");
+        let spare = self.builder.values.spare_capacity_mut();
+        spare[index] = MaybeUninit::new(value);
     }
```
a10y: Can we mark these `#[inline]` since they are small and commonly used across crates

a10y (approved):
Better than what we have

PR #4459, opened 2025-09-01
Remove caching the array on flat layout reader, rely on the segment cache instead

onursatici:
With this we don't hold references to arrays in the flat reader, so we can configure the segment cache to have more control on memory held by a layoutref instance. There doesn't seem to be a noticeable regression, other than one query in clickbench datafusion (q23 - 20%).

Another approach could be to have weak references instead.
LayoutRef would still hold cache zone maps and dict values after this change, I did not remove them because they are somewhat bounded

### DIFF
--- a/vortex-layout/src/layouts/flat/reader.rs
+++ b/vortex-layout/src/layouts/flat/reader.rs
@@ -3,7 +3,7 @@
 
 use std::collections::BTreeSet;
 use std::ops::{BitAnd, Range};
-use std::sync::{Arc, OnceLock};
+use std::sync::Arc;
 
 use async_trait::async_trait;
 use futures::FutureExt;
@@ -37,7 +37,6 @@ pub struct FlatReader {
     layout: FlatLayout,
     name: Arc<str>,
     segment_source: Arc<dyn SegmentSource>,
-    array: OnceLock<SharedArrayFuture>,
 }
 
 impl FlatReader {
@@ -50,7 +49,6 @@ impl FlatReader {
             layout,
             name,
             segment_source,
-            array: Default::default(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


robert3005:
It's great that this is somewhat flat, I would adjust the code of the layout reader since we no longer need a sharedfuture iirc and the comments are no longer correct

gatesn:
Mind if we chat through this? Want to double check it fits into 4406.

The other reason this existed originally was to avoid parsing the serialized arrays from each segment over and over. Flatbuffers verification was killing us

vortex-layout/src/layouts/flat/reader.rs
```diff
@@ -69,21 +67,16 @@ impl FlatReader {
...
+        let ctx = self.layout.ctx.clone();
+        let dtype = self.layout.dtype().clone();
+        Ok(async move {
+            let segment = segment_fut.await?;
+            ArrayParts::try_from(segment)?
+                .decode(&ctx, &dtype, row_count)
+                .map_err(Arc::new)
+        }
+        .boxed()
+        .shared())
```
gatesn: Remove shared, then let's do it!

onursatici:
Unfortunately there is no straightforward way to get rid of shared futures here just yet. Because of the `Sync` bound on `ArrayEvaluation`, I can only store `Sync` futures in `FlatEvaluation`. 

We have to have the Sync bound on both `ArrayEvaluation` and `MaskEvaluation` because of a combination of two things:
- both evaluation traits are `async_trait`, and they have the invoke method which takes `&self`. Futures generated by async_trait are `Send`, and because it captures `&self`, there is the `&self: Send` or `self: Sync` bound
- we store other evaluations in evaluation implementations like in the chunked reader where we store `Vec<Arc<dyn Evaluation>>` these trait objects have to be sync for self to be sync.

robert3005 (approved):
One minor thing is that we no longer need the array_future to return VortexResult (that was a case before this pr, just not cleaned up earlier)

Issue #3788, opened 2025-07-07
Add `CastKernel::can_cast`


blaginin:
Follow up on https://github.com/vortex-data/vortex/issues/3681 to discuss if we want to add a method that checks if cast is possible without performing the actual cast

Arrow equivalent: [`arrow_cast::can_cast_types`](https://arrow.apache.org/rust/arrow_cast/cast/fn.can_cast_types.html)

gatesn:
I don't think we want this, because Vortex doesn't define casting semantics itself

blaginin:
Would it be possible to operate on datatypes without the actual arrays then? In spiral we do that e.g. in `unify_data_types`

blaginin:
Discussed in person - we don't want this for now because:
- in many cases, you already have data, so you can cast directly
- when you do need checking instead of conversion, it gets more complex due to extensions and FFI implementation 
- as a workaround, you can cast e.g. a default value to verify the cast works

Closing the issue for now, but will revisit if there's more demand!

blaginin:
> as a workaround, you can cast e.g. a default value to verify the cast works

hm, not sure that solution works because this passes:

```rust

let data_type1 = DType::Primitive(PType::I32, Nullability::Nullable);
let data_type2 = DType::Primitive(PType::I8, Nullability::Nullable);

let mut builder = builder_with_capacity(&data_type1, 1);
builder.append_scalar(&Scalar::default_value(data_type1.clone())).expect("default value can be appended");

let array = builder.finish();
let res = vortex::compute::cast(&array, &data_type2);
assert!(res.is_ok());

```

PR #4462, opened 2025-09-01
chore: allow casting empty all invalid arrays to non-nullable

blaginin:


vortex-array/src/arrays/struct_/compute/cast.rs
```diff
@@ -133,4 +134,28 @@ mod tests {
         )
         .unwrap()
     }
+
+    #[test]
+    fn cast_nullable_all_invalid() {
```
blaginin, line 128: failing on dev with 

```
called `Result::unwrap()` on an `Err` value: Cannot cast array with invalid values to non-nullable type.
```

vortex-array/src/compute/cast.rs
```diff
@@ -23,6 +23,10 @@ static CAST_FN: LazyLock<ComputeFn> = LazyLock::new(|| {
 ///
 /// Some array support the ability to narrow or upcast.
 pub fn cast(array: &dyn Array, dtype: &DType) -> VortexResult<ArrayRef> {
+    if array.is_empty() {
+        return Ok(Canonical::empty(dtype).into_array());
+    }
+
```
blaginin: The only thing I don't like about this is that our cast would succeed in places where it shouldn't (https://github.com/vortex-data/vortex/issues/3788) but with empty array that feels the right thing to do?

lwwmanning: Yeah, I think of arrays as an ordered collection of values guaranteed to all be within a certain domain (determined by the DType). 

Empty array is trivially covered by all possible domains. Likewise, we can cast nullable u64 array to nonnullable u8 array if all the values are contained in the target domain [0, 255]

lwwmanning: Another example: IIRC we can cast a NullArray to any nullable dtype

vortex-array/src/arrays/struct_/compute/cast.rs
```diff
@@ -133,4 +133,28 @@ mod tests {
...
+        let empty_struct = StructArray::try_new(
+            FieldNames::from(["a"]),
+            vec![
+                PrimitiveArray::new::<i32>(buffer![], crate::validity::Validity::AllInvalid)
+                    .to_array(),
+            ],
+            0,
+            crate::validity::Validity::AllInvalid,
```
lwwmanning: nit: can just import Validity

vortex-array/src/arrays/null/compute/cast.rs
```diff
@@ -78,7 +78,6 @@ mod tests {
     #[case(NullArray::new(5))]
     #[case(NullArray::new(1))]
     #[case(NullArray::new(100))]
-    #[case(NullArray::new(0))]
```
lwwmanning: Why do we need to remove this case? Do we need to update the cast conformance test?

vortex-array/src/arrays/struct_/compute/cast.rs
blaginin: Done in https://github.com/vortex-data/vortex/pull/4462/commits/6f899266579765bd61ebd030ccaf6515540ede37

vortex-array/src/arrays/null/compute/cast.rs
blaginin: Done in https://github.com/vortex-data/vortex/pull/4462/commits/69fd783c9ed36693234450a9b69cae05636c80cc

blaginin:
image


assuming that was flaky after all 🙂

robert3005 (approved):

PR #4463, opened 2025-09-01
chore(deps): Update `tracing-subscriber`

blaginin:
Github says we should update

Zen Browser 2025-09-01 20 03 55

Dependabot has actually done it: https://github.com/vortex-data/vortex/pull/4439 - but it updated only lockfile so now we have

https://github.com/vortex-data/vortex/blob/63b60ff8434d91773c25feb25c765d0666db784d/Cargo.lock#L6249-L6250

https://github.com/vortex-data/vortex/blob/63b60ff8434d91773c25feb25c765d0666db784d/Cargo.toml#L194

(how is that even allowed???)


AdamGS:
it's allowed because by default cargo dependencies are https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements, which is why we have a check that verifies that the minimum version still work.

blaginin:
oo thanks! although i thought renovate should bump the version in the toml file as well

robert3005 (approved):
toml bumps should be human driven, this is where you're raising your minimup supported version

PR #4469, opened 2025-09-02
perf[pipeline]: optimize the execution of prim and bool pipeline

joseph-isaacs:
Remove all memory copies in primitive and bool canoncalise pipeline materalise.

We use iterator for emitting mask slices and validity slices. Since the next borrow from the iterator, the trait `Iterator` cannot be used.


vortex-duckdb/src/convert/scalar.rs
```diff
@@ -209,78 +209,89 @@ impl<'a> TryFrom<ValueRef<'a>> for Scalar {
...
-            ExtractedValue::Varchar(s) => Ok(Scalar::utf8(s, Nullable)),
-            ExtractedValue::Blob(b) => Ok(Scalar::binary(b, Nullable)),
+            ExtractedValue::UTinyInt(v) => Ok(Scalar::primitive(v, NonNullable)),
+            ExtractedValue::USmallInt(v) => Ok(Scalar::primitive(v, NonNullable)),
+            ExtractedValue::UInteger(v) => Ok(Scalar::primitive(v, NonNullable)),
+            ExtractedValue::UBigInt(v) => Ok(Scalar::primitive(v, NonNullable)),
+            ExtractedValue::Float(v) => Ok(Scalar::primitive(v, NonNullable)),
+            ExtractedValue::Double(v) => Ok(Scalar::primitive(v, NonNullable)),
+            ExtractedValue::Varchar(s) => Ok(Scalar::utf8(s, NonNullable)),
+            ExtractedValue::Blob(b) => Ok(Scalar::binary(b, NonNullable)),
```
joseph-isaacs, line 230: return non-nullable, since our kernels can apply operations on the same DType with different nullability

vortex-array/src/arrays/primitive/compute/pipeline.rs
```diff
@@ -105,14 +123,79 @@ impl<T: Element + NativePType> Kernel for PrimitiveKernel<T> {
...
+                )
+            };
+
+            validity_slice[..remaining.div_ceil(u32::BITS as usize)].copy_from_slice(usize_slice);
+
+            self.offset += remaining;
+        }
+
+        // TODO(joe): use mask in copy_from_slice, if faster.
+        // out.select_mask::<T>(&_selected);
```
AdamGS, line 183: dead code

PR #4473, opened 2025-09-02
Feature: add `to_canonical` to `ArrayBuilder`

connortsui20:
It seems somewhat wasteful to always finish the builders into type-erased `ArrayRef`, when all of them can directly finish into a canonical array.

This PR adds `to_canonical` to the `ArrayBuilder` method, which saves an allocation in a few cases. This is mostly a "nice-to-have" rather than something that is necessary.

Side note: I think that it might be worth going through our conversion naming since `finish().to_primitive()` and `to_canonical().into_primitive()` sound like they do different things...

### DIFF
--- a/encodings/dict/src/array.rs
+++ b/encodings/dict/src/array.rs
@@ -329,7 +329,7 @@ mod test {
         array.clone().append_to_builder(builder.as_mut());
 
         let into_prim = array.to_primitive();
-        let prim_into = builder.finish().to_primitive();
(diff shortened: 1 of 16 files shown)


robert3005 (commented):
Maybe call it `finish_to_canonical()`? `to_canonical()` is definitely misleading since it's unclear what happens to the builder state

connortsui20:
I made it `finish_into_canonical` to be consistent with the methods already on each builder

robert3005 (approved):

PR #4475, opened 2025-09-02
Refactor ObjectStoreWriter

AdamGS:
1. Support multiple flush calls.
2. Clarify usage of `shutdown`.
3. Use `FuturesUnordered` instead of `FuturesOrdered`.
4. Remove artificial limit on amount of buffered data.
5. Test using both in-memory object store and local disk.

### DIFF
--- a/vortex-io/src/object_store.rs
+++ b/vortex-io/src/object_store.rs
@@ -7,7 +7,7 @@ use std::os::unix::prelude::FileExt;
 use std::sync::Arc;
 
 use bytes::BytesMut;
-use futures::future::try_join_all;
+use futures::stream::FuturesUnordered;
 use futures_util::StreamExt;
 use object_store::path::Path;
 use object_store::{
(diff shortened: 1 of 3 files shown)


vortex-io/src/object_store.rs
```diff
@@ -136,113 +138,115 @@ impl ObjectStoreWriter {
...
 
-                parts.push(part_fut);
-            }
+            parts.push(part_fut);
+        }
 
-            try_join_all(parts).await?;
+        while let Some(p) = parts.next().await {
+            p?;
         }
```
robert3005: why is this not `parts.await?`? I thought that's the whole point of futuresunordered

AdamGS: as far as I can tell, it doesn't implement `Future` directly, I can probably `try_collect`

robert3005: yes, you have to collect

AdamGS: done

robert3005 (approved):

PR #4486, opened 2025-09-02
chore: `#[inline]` fns vortex-dtype

0ax1:
Annotates small functions in vortex-dtype with `#[inline]` that could benefit from cross crate inlining.

### DIFF
--- a/vortex-dtype/src/ptype.rs
+++ b/vortex-dtype/src/ptype.rs
@@ -598,38 +598,45 @@ macro_rules! match_each_native_simd_ptype {
 
 impl PType {
     /// Returns `true` iff this PType is an unsigned integer type
+    #[inline]
     pub const fn is_unsigned_int(self) -> bool {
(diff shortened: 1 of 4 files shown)


vortex-dtype/src/ptype.rs
```diff
@@ -598,38 +598,45 @@ macro_rules! match_each_native_simd_ptype {
...
     }
 
     /// Returns `true` iff this PType is a floating point type
+    #[inline]
     pub const fn is_float(self) -> bool {
         matches!(self, Self::F16 | Self::F32 | Self::F64)
     }
 
     /// Returns the number of bytes in this PType
+    #[inline]
```
joseph-isaacs, line 626: I would do this for match_each macros?

0ax1: Keeping it for now, as we don't see any regressions, and the assumption is that this compiles to a lookup table.

robert3005 (approved):

PR #4489, opened 2025-09-02
feat: when writing, enable finer grained configuration of layout

danking:


danking:
I seem to have broken flat layouts in this change. I'll investigate tomorrow.

vortex-file/src/strategy.rs
```diff
@@ -19,14 +20,61 @@ use vortex_layout::{LayoutStrategy, LocalExecutor, TaskExecutor};
 const ONE_MEG: u64 = 1 << 20;
 const ROW_BLOCK_SIZE: usize = 8192;
 
+enum ZoneMapConfiguration {
```
robert3005, line 23: I would actually stitch this in vortex-python. At rust level you can combine the strategy yourself?

vortex-file/src/strategy.rs
```diff
@@ -54,12 +104,63 @@ impl WriteStrategyBuilder {
...
+
+    /// Disable inter-chunk dictionary compression.
+    pub fn without_dictionary(mut self) -> Self {
+        self.dictionary_compressor = Some(CompressorConfiguration::NoCompressor);
+        self
+    }
+
+    /// Override the size of the zone maps and the compressor for the zone stats.
+    ///
+    /// If not provided, this uses 8Ki zones with a BtrBlocks-style cascading compressor on the stats.
```
a10y, line 149: For clarity, I think we just say 8192 rows instead of 8Ki

vortex-layout/src/layouts/dict/writer/mod.rs
```diff
@@ -54,18 +54,39 @@ impl Default for DictLayoutOptions {
...
     options: DictLayoutOptions,
     executor: Arc<dyn TaskExecutor>,
 }
 
+impl<Codes, Values> DictStrategy<Codes, Values, Values>
+where
+    Codes: LayoutStrategy,
+    Values: LayoutStrategy,
+{
+    pub fn new(
```
a10y, line 67: I'm not sure I like getting rid of the need to explicitly name a fallback

vortex-file/src/strategy.rs
danking: Nick and Onur suggested Rust-side configurability [in Slack](https://spiraldb.slack.com/archives/C07BV3GKAJ2/p1756836540009419?thread_ts=1756835735.945749&cid=C07BV3GKAJ2).

It seems to me somewhat subtle to properly combine these different strategies which justifies something like this in Rust.

robert3005: Hm, I looked at the dependencies of vortex-file and I thought we should minimize them but then I see that we include all the encodings

PR #4494, opened 2025-09-03
Remove evaluations, use MaskFuture

gatesn:
Next step in removing implicit pre-fetching is to move over to plumbing the scan evaluations together ahead of time using MaskFutures.


vortex-scan/src/tasks.rs
```diff
@@ -150,45 +122,41 @@ pub(super) fn split_exec<A: 'static + Send>(
...
-        .projection_evaluation(&row_range, &ctx.projection)?;
+    let projection_future =
+        ctx.reader
+            .projection_evaluation(&row_range, &ctx.projection, filter_mask.clone())?;
+
     let mapper = ctx.mapper.clone();
     let array_fut = async move {
-        let filtered_mask = filter.await?;
-        if filtered_mask.all_false() {
+        let (array, mask) = try_join!(projection_future, filter_mask)?;
```
gatesn: This is the only behavior change, where we fetch projection I/O alongside waiting for the result of the filter.

joseph-isaacs: What result does this have? its a big change

joseph-isaacs (changes requested):
There are huge clickbench regressions

joseph-isaacs (approved):
LG, now

PR #4495, opened 2025-09-03
Feature: Implement Compute Functions for `FixedSizeListArray`

connortsui20:

Implements the same compute functions as `ListArray`, but for `FixedSizeListArray`.

- Take: Select elements by indices with optimized paths for nullable/non-nullable data
- Filter: Apply boolean masks with density-based expansion strategy (not sure it this is a premature optimization)
- Cast
- IsConstant: Check if all lists in the array are identical (this is super expensive and it might be worth revisiting the compute function characteristics itself, cc @gatesn) 
- Mask
- Fuzzing: Extended arbitrary data generation to support `FixedSizeListArray`
- Arrow conversion
- Refactored `ListArray` filter compute function with a similar density check

The two complicated compute functions here are `filter` and `take`, with `take` being significantly more complicated. Those are accompanied by tests since the logic is a lot harder.

Nice-to-have-but-I'm-too-tired:
- Tests for each compute function (missing for cast, is_constant, and mask)
- Refactor the `ListArray` `take` compute function

### DIFF
--- a/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
@@ -9,6 +9,10 @@ use crate::compute::{CastKernel, CastKernelAdapter, cast};
 use crate::vtable::ValidityHelper;
 use crate::{ArrayRef, register_kernel};
 
+/// Cast implementation for [`FixedSizeListArray`].
+///
+/// Recursively casts the inner elements array to the target element type while preserving the list
+/// structure.
 impl CastKernel for FixedSizeListVTable {
     fn cast(&self, array: &FixedSizeListArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         let Some(target_element_type) = dtype.as_list_element_opt() else {
--- a/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
@@ -26,8 +26,19 @@ impl CastKernel for FixedSizeListVTable {
             .clone()
             .cast_nullability(dtype.nullability())?;
 
-        FixedSizeListArray::try_new(elements, array.list_size(), validity, array.len())
-            .map(|a| Some(a.to_array()))
+        Ok(Some(
+            // SAFETY: The only requirements for safety here are related to lengths, and no lengths
+            // have changed here. So as long as the original array is valid, this is also valid.
+            unsafe {
+                FixedSizeListArray::new_unchecked(
+                    elements,
+                    array.list_size(),
+                    validity,
+                    array.len(),
+                )
+            }
+            .to_array(),
+        ))
     }
 }
 
--- a/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
@@ -10,7 +10,7 @@ use crate::vtable::ValidityHelper;
 use crate::{ArrayRef, register_kernel};
 
 impl CastKernel for FixedSizeListVTable {
-    fn cast(&self, array: &Self::Array, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+    fn cast(&self, array: &FixedSizeListArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
         let Some(target_element_type) = dtype.as_list_element_opt() else {
             return Ok(None);
         };
--- a/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_dtype::DType;
+use vortex_error::VortexResult;
+
+use crate::arrays::{FixedSizeListArray, FixedSizeListVTable};
+use crate::compute::{CastKernel, CastKernelAdapter, cast};
+use crate::vtable::ValidityHelper;
+use crate::{ArrayRef, register_kernel};
+
+impl CastKernel for FixedSizeListVTable {
+    fn cast(&self, array: &Self::Array, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        let Some(target_element_type) = dtype.as_list_element_opt() else {
+            return Ok(None);
+        };
+
+        // The list elements could technically be from either `FixedSizeList` or `List`.
+        if dtype.is_list() {
+            return Ok(None);
+        }
+
+        let elements = cast(array.elements(), target_element_type)?;
+        let validity = array
+            .validity()
+            .clone()
+            .cast_nullability(dtype.nullability())?;
+
+        FixedSizeListArray::try_new(elements, array.list_size(), validity, array.len())
+            .map(|a| Some(a.to_array()))
+    }
+}
+
+register_kernel!(CastKernelAdapter(FixedSizeListVTable).lift());
--- a/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/cast.rs
@@ -15,14 +15,9 @@ use crate::{ArrayRef, register_kernel};
 /// structure.
 impl CastKernel for FixedSizeListVTable {
     fn cast(&self, array: &FixedSizeListArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
-        let Some(target_element_type) = dtype.as_list_element_opt() else {
+        let Some(target_element_type) = dtype.as_fixed_size_list_element_opt() else {
             return Ok(None);
         };
-
-        // The list elements could technically be from either `FixedSizeList` or `List`.
-        if dtype.is_list() {
-            return Ok(None);
-        }
 
         let elements = cast(array.elements(), target_element_type)?;
         let validity = array
--- a/vortex-array/src/arrays/fixed_size_list/compute/filter.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/filter.rs
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_error::VortexResult;
+use vortex_mask::Mask;
+
+use crate::arrays::FixedSizeListVTable;
+use crate::compute::{FilterKernel, FilterKernelAdapter};
+use crate::{ArrayRef, register_kernel};
+
+impl FilterKernel for FixedSizeListVTable {
+    fn filter(&self, array: &Self::Array, mask: &Mask) -> VortexResult<ArrayRef> {
+        todo!()
+    }
+}
+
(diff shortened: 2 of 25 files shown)


vortex-array/src/arrays/fixed_size_list/compute/cast.rs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        let elements = cast(array.elements(), target_element_type)?;
+        let validity = array
+            .validity()
+            .clone()
+            .cast_nullability(dtype.nullability())?;
+
+        FixedSizeListArray::try_new(elements, array.list_size(), validity, array.len())
+            .map(|a| Some(a.to_array()))
```
robert3005: I think you want the unsafe version here

vortex-array/src/arrays/fixed_size_list/compute/filter.rs
```diff
@@ -0,0 +1,151 @@
...
+            .iter()
+            .map(|&idx| {
+                let start = idx * list_size;
+                let end = (idx + 1) * list_size;
+                (start, end)
+            })
+            .collect();
+
+        Mask::from_slices(expanded_len, expanded_slices)
+    }
```
robert3005: I feel this is uncessarily optimized, we can come back to further optimisation once we have it wired into the file layout

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -0,0 +1,227 @@
...
+            array.list_size(),
+            Validity::NonNullable,
+            len,
+        )
+    }
+    .into_array()
+}
+
+/// Returns empty lists with computed validity for the nullable case.
+fn take_degenerate_nullable<I: NativePType>(
```
robert3005: this whole thing is just `array.validity().take(indices_array)`

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -0,0 +1,227 @@
...
+        if array.dtype().is_nullable() || indices_array.dtype().is_nullable() {
+            take_nullable_fsl::<I>(array, indices_array)
+        } else {
+            take_non_nullable_fsl::<I>(array, indices_array)
+        }
+    }
+}
+
+/// Returns empty non-nullable lists with the requested length.
+fn take_degenerate_non_nullable(
```
robert3005: and then you don't need this case

robert3005 (changes requested):
I think we can simplify things a little bit

vortex-array/src/arrays/fixed_size_list/compute/cast.rs
```diff
@@ -0,0 +1,45 @@
...
+impl CastKernel for FixedSizeListVTable {
+    fn cast(&self, array: &FixedSizeListArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        let Some(target_element_type) = dtype.as_list_element_opt() else {
+            return Ok(None);
+        };
+
+        // The list elements could technically be from either `FixedSizeList` or `List`.
+        if dtype.is_list() {
+            return Ok(None);
+        }
```
connortsui20: This seems somewhat error-prone, should I create a `as_fixed_size_list_element_opt` that only returns the type if it is `DType::FixedSizeList` and have `as_list_element_opt` return `None` in that case?

vortex-array/src/arrays/fixed_size_list/compute/filter.rs
```diff
@@ -0,0 +1,131 @@
...
+        .iter()
+        .map(|&idx| {
+            let start = idx * list_size;
+            let end = (idx + 1) * list_size;
+            (start, end)
+        })
+        .collect();
+
+    Mask::from_slices(expanded_len, expanded_slices)
+}
```
connortsui20: This can be inlined

vortex-array/src/arrays/fixed_size_list/compute/is_constant.rs
```diff
@@ -0,0 +1,38 @@
...
+        let first_scalar = array.scalar_at(0); // We checked the array length above.
+
+        // TODO(connor): There must be a more efficient way to do this. Each `scalar_at()` call
+        // makes several allocations...
+        for i in 1..array.len() {
+            let current_scalar = array.scalar_at(i);
+            if current_scalar != first_scalar {
+                return Ok(Some(false));
+            }
+        }
```
connortsui20, line 35: Yeah this is pretty bad. I'm also realizing that this will also check the dtypes are equal _every single time_. Maybe a `eq_value` or something is needed? I'll probably leave this for later.

vortex-array/src/arrays/list/compute/cast.rs
```diff
@@ -15,6 +15,11 @@ impl CastKernel for ListVTable {
             return Ok(None);
         };
 
+        // The list elements could technically be from either `FixedSizeList` or `List`.
+        if dtype.is_fixed_size_list() {
+            return Ok(None);
+        }
+
```
connortsui20: Same problem as above, I think this might be unexpected behavior but having a `as_fixed_size_list_element_opt` might be bloat

vortex-array/src/arrays/list/compute/filter.rs
connortsui20, line 1: This is a refactor where I do a similar optimization based on the density of the mask. However, we cannot simply call `Mask::from_slices` since we have no idea how long each list we are taking is. So instead we just build the buffer in both cases. This might also be premature, but I think that _always_ calling `.slices()` and incurring that initialization every time (line 60-64 in the original) seems pretty expensive.

Edit: (this is not outdated)

vortex-array/src/arrays/list/compute/take.rs
connortsui20, line 1: I think these are just wrong to be honest

Edit: (this is not outdated)

vortex-array/src/arrow/compute/to_arrow/canonical.rs
connortsui20, line 1: I don't fully understand this code to be honest, based on https://github.com/vortex-data/vortex/pull/4481

vortex-array/src/arrays/list/compute/filter.rs
```diff
@@ -6,83 +6,136 @@ use std::ops::AddAssign;
...
+        let (new_elements, new_offsets) = match_each_integer_ptype!(offsets.ptype(), |O| {
+            compute_filtered_elements_and_offsets::<O>(
+                elements.as_ref(),
+                offsets.as_slice::<O>(),
+                selection_mask,
+            )?
+        });
+
+        // TODO(connor): Use `new_unchecked` here.
+        Ok(ListArray::try_new(new_elements, new_offsets.into_array(), new_validity)?.into_array())
```
connortsui20: I'm going to address all of these in a followup PR

vortex-array/src/arrow/compute/to_arrow/canonical.rs
```diff
@@ -386,6 +392,46 @@ fn to_arrow_list<O: NativePType + OffsetSizeTrait>(
...
+            element.clone(),
+        )
+    } else {
+        let values = array.elements().clone().into_arrow_preferred()?;
+        let element_field = Arc::new(Field::new_list_field(
+            values.data_type().clone(),
+            array.elements().dtype().is_nullable(),
+        ));
+        (values, element_field)
+    };
```
connortsui20, line 424: I don't fully understand this code to be honest, based on https://github.com/vortex-data/vortex/pull/4481

vortex-array/src/arrays/fixed_size_list/compute/filter.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn filter(&self, array: &FixedSizeListArray, selection_mask: &Mask) -> VortexResult<ArrayRef> {
+        let new_len = selection_mask.true_count();
+        let null_mask = array.validity_mask();
+
+        // If the entire array is null, then we only need to adjust the length of the array.
+        if let Mask::AllFalse(_) = null_mask {
+            return Ok(
+                ConstantArray::new(Scalar::null(array.dtype().clone()), new_len).into_array(),
+            );
+        }
```
robert3005: this should be handled already

vortex-array/src/arrays/list/compute/take.rs
```diff
@@ -109,12 +109,15 @@ fn _take_nullable<I: NativePType, O: OffsetPType + NativePType + PrimInt>(
...
 ) -> VortexResult<ArrayRef> {
+    // TODO(connor): The primitive builders here have the same ptype...
     let mut new_offsets = PrimitiveBuilder::with_capacity(Nullability::NonNullable, indices.len());
     let mut elements_to_take =
         PrimitiveBuilder::with_capacity(Nullability::NonNullable, 2 * indices.len());
 
     let mut current_offset = O::zero();
     new_offsets.append_zero();
+
+    // TODO(connor): Why is this 2 x length???
```
robert3005: we make arbitrary decision here since we don't know the translation from items to elements

vortex-array/src/arrays/list/compute/take.rs
```diff
@@ -139,6 +142,7 @@ fn _take_nullable<I: NativePType, O: OffsetPType + NativePType + PrimInt>(
 
             elements_to_take.ensure_capacity(elements_to_take.len() + additional);
             for i in 0..additional {
+                // TODO(connor): This should probably be `I::from_usize`?
```
robert3005: this is the offsets so it should be O, we don't really need I in all of these

robert3005 (approved):
Only have some nits

vortex-array/src/arrow/compute/to_arrow/canonical.rs
robert3005: The nested arrow conversion is very... annoying

vortex-array/src/arrays/fixed_size_list/compute/cast.rs
robert3005: I don't think s we should maybe just make this work?

robert3005: Having thought about this more I think you want to separate this two out. If FixedSizeList is a dtype then it's elements are distinct from the list elements

connortsui20: I separated them out:

```rust
    /// Get the inner element dtype if `self` is a [`DType::List`], otherwise returns `None`.
    ///
    /// Note that this does _not_ return `Some` if `self` is a [`DType::FixedSizeList`].
    pub fn as_list_element_opt(&self) -> Option<&Arc<DType>> {
        if let List(edt, _) = self {
            Some(edt)
        } else {
            None
        }
    }

    /// Get the inner element dtype if `self` is a [`DType::FixedSizeList`], otherwise returns
    /// `None`.
    ///
    /// Note that this does _not_ return `Some` if `self` is a [`DType::List`].
    pub fn as_fixed_size_list_element_opt(&self) -> Option<&Arc<DType>> {
        if let FixedSizeList(edt, ..) = self {
            Some(edt)
        } else {
            None
        }
    }

    /// Get the inner element dtype if `self` is **either** a [`DType::List`] or a
    /// [`DType::FixedSizeList`], otherwise returns `None`
    pub fn as_any_size_list_element_opt(&self) -> Option<&Arc<DType>> {
        if let FixedSizeList(edt, ..) = self {
            Some(edt)
        } else if let List(edt, ..) = self {
            Some(edt)
        } else {
            None
        }
    }

```

PR #4497, opened 2025-09-03
optional array cache for layout readers

onursatici:
This creates an array cache that follows the segment cache. This would restore the previous behaviour of caching deserialised arrays from segments, but now the size is bounded with the segment cache size.

I am assuming most of the time the arrayref's would have the references to the ByteBuffer we create them from, because their alignment match, so the extra layer of array cache won't use much memory. There are cases where deserialising an array from a byte buffer allocate memory, but that should not be that common.


gatesn (changes requested):
I'm going to push back on this approach. Main concern is that I think this is solving the right problem at the wrong layer. Once the operator/lazy eval work lands, we'll have much better visibility into which arrays are actually being reused across operators vs. read-once. Caching at the layout reader level right now is a guess about access patterns that we'll be able to make informed decisions about in a few weeks.

Let's punt this until operators are in place.

onursatici:
punting this until operators land

PR #4499, opened 2025-09-03
feat: New vx inspect tool

AdamGS:
This is mostly claude generated, still lacks some finishing touches.

Also removed the `segments` subcommand, as its basically the same functionality.


gatesn:
"dump" is maybe a better name?

robert3005:
I think having more commands could definitely be good since you can use them with other command line tools but on the other hand I also like having more comprehensive tui with more features

gatesn:
Yeah, but this is indeed a perfect Claude task. Maybe we get claude to flesh out the TUI to be a bit more complete

AdamGS:
right now I'm just stashing this tool here while I investigate, once that investigation is done I'll invest some time into making it nice to use.

In general I think that this sort of thing can be useful in both TUI and CLI forms, we don't have to pick one over the other.
The only thing is that last time I looked the TUI needed some untangling to make it easier to add more tabs, but its been a while and before @robert3005 added the segment browser.

vortex-tui/src/inspect.rs
joseph-isaacs, line 1: think there is much duplication between here and vortex file?

joseph-isaacs (approved):

vortex-tui/src/inspect.rs
robert3005: Yeah, this is reverse engineered logic based on the file version (right now only one version). There should be methods for parsing segments that we maybe need to make public?

AdamGS: yeah but the functionality in `vortex-file` isn't public, and I'm not sure we want to be?

PR #4501, opened 2025-09-03
Disable prefetching

gatesn:


robert3005:
there's a bit of regression in S3, but can't tell if it's ranom or not, doesn't look random

gatesn:
Yeah I was getting a baseline to see if anything else I do makes a blind bit of difference 😄

PR #4502, opened 2025-09-03
feat: remove ArrayAccessor, use Iterator/IntoIterator

a10y:
ArrayAccessor was garbage for two reasons:

1. Strange API
2. It received a `&mut dyn Iterator` parameter everywhere, forcing dynamic dispatch to visit every element of the array.

ArrayAccessor is replaced with relevant `Iterator` and `IntoIterator` impls where appropriate, allowing us to dispatch statically in things like FSST compression, sorted/constant compute and the like.

One thing I didn't do here that might be nice is to have separate iters for when null checks can be elided, returning `T` instead of `Option<T>`


a10y:
Apparently eliding the validity check is very important after all

robert3005:
You can pass an enum? Or make the iterator by default unsafe and access elements directly and callers can’t opt into checking validity?(this is what arrow does)

a10y:
Passing an enum I think might give similar levels of indirection, though at least you could (in theory) optimize through the branch? Just wasn't clearly better. Might revisit another time

PR #4504, opened 2025-09-03
fix: #4503

### LINKED ISSUE #4503: bug: pco fails to round-trip the size two i32 array: [null, 1].
I expected this to exit successfully; instead, the `decompress()` call fails.
```rust
    let array = PrimitiveArray::from_option_iter([None, Some(1)]);
    let compressed = PcoArray::from_primitive(&array, 3, 0).unwrap();
    assert_eq!(compressed.scalar_at(0), Scalar::null_typed::<i32>());
    assert_eq!(compressed.scalar_at(1), Scalar::from(Some(1)));
    let decompressed = compressed.decompress();
    assert_eq!(decompressed.scalar_at(0), Scalar::null_typed::<i32>());
    assert_eq!(decompressed.scalar_at(1), Scalar::from(Some(1)));
```
```
thread 'test::test_compress_decompress_big' panicked at /Users/danielking/projects/vortex/vortex-buffer/src/buffer.rs:266:13:
range end out of bounds: 2 <= 1
```

danking:
Fixes #4503 

Pco mistakenly used the number of values rather than the number of _valid_ values.

### DIFF
--- a/encodings/pco/src/test.rs
+++ b/encodings/pco/src/test.rs
@@ -44,7 +44,7 @@ fn test_compress_decompress() {
 }
 
 #[test]
-#[should_panic]
+#[should_panic(expected = "range end out of bounds: 2 <= 1")]
 fn test_compress_decompress_big() {
     let array = PrimitiveArray::from_option_iter([None, Some(1)]);
     let compressed = PcoArray::from_primitive(&array, 3, 0).unwrap();
--- a/encodings/pco/src/test.rs
+++ b/encodings/pco/src/test.rs
@@ -44,7 +44,6 @@ fn test_compress_decompress() {
 }
 
 #[test]
-#[should_panic(expected = "range end out of bounds: 2 <= 1")]
 fn test_compress_decompress_big() {
     let array = PrimitiveArray::from_option_iter([None, Some(1)]);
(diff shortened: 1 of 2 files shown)


encodings/pco/src/test.rs
```diff
@@ -42,6 +43,18 @@ fn test_compress_decompress() {
...
+#[test]
+#[should_panic]
+fn test_compress_decompress_big() {
+    let array = PrimitiveArray::from_option_iter([None, Some(1)]);
+    let compressed = PcoArray::from_primitive(&array, 3, 0).unwrap();
+    assert_eq!(compressed.scalar_at(0), Scalar::null_typed::<i32>());
+    assert_eq!(compressed.scalar_at(1), Scalar::from(Some(1)));
+    let decompressed = compressed.decompress();
+    assert_eq!(decompressed.scalar_at(0), Scalar::null_typed::<i32>());
+    assert_eq!(decompressed.scalar_at(1), Scalar::from(Some(1)));
```
a10y, line 54: Which one should panic? there are 4 different asserts here

danking: It's actually the decompression that panics. I'll add a message check

danking: Actually I found the fix.

robert3005 (approved):

PR #4508, opened 2025-09-03
Concurrent filter 2

gatesn:
Based on ngates/concurrent-filter (but I can't run benchmarks if I set the base branch)


gatesn:
This change is only really better than concurrent filters for SF1 from S3, which we sort of expect, since it avoids a round-trip to run filtering and starts fetching projection columns immediately.

However, if we want to do that, we should be able to write a scan implementation that expands the I/O window until we reach the end of the file and then begins to early evaluate projection columns, without _always_ doing it, which can lead to some regressions on TPCDS vs the other PR.

PR #4512, opened 2025-09-03
feat: Constant compression for strings

a10y:
Arade has two all constant columns, we were encoding this with an FSST chunk with constant components, which is pretty gross

image

After this change, constant encoding is properly getting selected:

image

### DIFF
--- a/vortex-btrblocks/src/string.rs
+++ b/vortex-btrblocks/src/string.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

danking (approved):

gatesn:
Does this mean the column has a single chunk? Or do we need to fix the layout strategy to avoid re-partitioning when consecutive arrays are constant? I think we would still split based on fully uncompressed size.

a10y:
Right now the column has 128 chunks. After this change it still has 128 chunks.

Separately, I would like to repartition based on compressed size. You can accumulate FSST buffers in the writer and flush them once they reach a certain size, or when you need to switch out a new symbol table

Issue #4470, opened 2025-09-02
vortex-io windows platform support


Trojanking123:
```
 cargo run
   Compiling vortex-io v0.51.0
error[E0433]: failed to resolve: could not find `unix` in `os`
  --> C:\Users\wuqia\.cargo\registry\src\rsproxy.cn-e3de039b2554c837\vortex-io-0.51.0\src\tokio.rs:7:14
   |
7  | use std::os::unix::fs::FileExt;
   |              ^^^^ could not find `unix` in `os`
   |
note: found an item that was configured out
  --> C:\Users\wuqia\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\std\src\os\mod.rs:36:9
   |
36 | pub mod unix {}
   |         ^^^^
note: the item is gated here
  --> C:\Users\wuqia\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\std\src\os\mod.rs:28:1
   |
28 | / #[cfg(all(
29 | |     doc,
30 | |     any(
31 | |         all(target_arch = "wasm32", not(target_os = "wasi")),
...  |
34 | | ))]
   | |___^
note: found an item that was configured out
  --> C:\Users\wuqia\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\std\src\os\mod.rs:85:9
   |
85 | pub mod unix;
   |         ^^^^
note: the item is gated here
  --> C:\Users\wuqia\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\std\src\os\mod.rs:84:1
   |
84 | #[cfg(all(not(target_os = "hermit"), any(unix, doc)))]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0599]: no method named `read_exact_at` found for struct `TokioFile` in the current scope
    --> C:\Users\wuqia\.cargo\registry\src\rsproxy.cn-e3de039b2554c837\vortex-io-0.51.0\src\tokio.rs:65:18
     |
24   | pub struct TokioFile(Arc<File>);
     | -------------------- method `read_exact_at` not found for this struct
...
65   |             this.read_exact_at(&mut buffer, range.start)?;
     |                  ^^^^^^^^^^^^^
     |
[...]

PR #4513, opened 2025-09-04
add windows platform support for vortex-io

Trojanking123:
When the `tokio` feature is enabled in `vortex-io`, use `#[cfg(windows)]` or `#[cfg(unix)]` to call different APIs.
To fix bug #4470

### DIFF
--- a/vortex-io/src/tokio.rs
+++ b/vortex-io/src/tokio.rs
@@ -81,10 +81,9 @@ impl VortexReadAt for TokioFile {
         let this = self.clone();
 
         spawn_blocking(move || {
-            use std::os::windows::fs::FileExt; // 导入 Windows 特有的 trait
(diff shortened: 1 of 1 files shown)


vortex-io/src/tokio.rs
```diff
@@ -138,8 +159,18 @@ mod tests {
 
         // Create a function to test if we can read from the file
         let can_read = |file: &File| {
+
             let mut buffer = vec![0; 7];
-            file.read_exact_at(&mut buffer, 0).is_ok()
+            #[cfg(unix)]
+            {
+                use std::os::unix::fs::FileExt;
```
gatesn, line 166: Are we better using https://docs.rs/positioned-io/latest/positioned_io/ do you think?

Sorry, not dropping this PR, we're just making lots of changes to I/O at the moment!

PR #4518, opened 2025-09-04
chore[ci]: natural change benchmark github post 

joseph-isaacs:


scripts/compare-benchmark-jsons.py
```diff
@@ -125,13 +125,17 @@ def format_performance(ratio, target_name):
...
+            emoji = "✅"
+        else:
+            emoji = "❌"
         return f"{ratio:.3f}x {emoji}"
 
 
-overall_performance = (
-    "no data" if pd.isna(geo_mean_ratio) else f"{geo_mean_ratio:.3f}x {'✅' if geo_mean_ratio < 1 else '❌'}"
-)
+overall_performance = "no data" if pd.isna(geo_mean_ratio) else format_performance(geo_mean_ratio, "overall")
```
0ax1, line 138: I assume `isna` is for "is nan". Should we also check for inf?

scripts/compare-benchmark-jsons.py
```diff
@@ -125,13 +125,17 @@ def format_performance(ratio, target_name):
     if pd.isna(ratio):
         return f"no {target_name.lower()} data"
     else:
-        emoji = "✅" if ratio < 1 else "❌"
+        # Use neutral emoji if within threshold
+        if improvement_threshold <= ratio <= regression_threshold:
```
0ax1, line 129: Not sure we really need emojis here. But I guess it is 2025 after all. :)

0ax1 (approved):

scripts/compare-benchmark-jsons.py
joseph-isaacs: maybe?

scripts/compare-benchmark-jsons.py
joseph-isaacs: how else do you convey info with colour?

PR #4520, opened 2025-09-04
Add a Vortex runtime abstraction

gatesn:
For now it doesn't include I/O support and isn't actually used anywhere.

See #4406 for examples of how this will ultimately be used.

I imagine this work will temporarily break WASM support. I plan to add this back after the I/O changes and then make sure we ship a WASM application that demonstrates usage.


vortex-io/src/runtime/handle.rs
```diff
@@ -0,0 +1,102 @@
...
+    recv: oneshot::Receiver<T>,
+    abort_handle: Option<Box<dyn AbortHandle<'rt> + 'rt>>,
+}
+
+impl<'rt, T> Task<'rt, T> {
+    /// Detach the task, allowing it to continue running in the background after being dropped.
+    pub fn detach(mut self) {
+        self.abort_handle = None;
+    }
+}
```
AdamGS: Do we want an explicit abort here?

vortex-io/src/runtime/tokio.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn spawn_cpu(&self, cpu: Box<dyn FnOnce() + Send + 'static>) -> AbortHandleRef<'static> {
+        Box::new(self.0.spawn(async move { cpu() }).abort_handle())
+    }
+}
+
+impl AbortHandle<'_> for tokio::task::AbortHandle {
+    fn abort(self: Box<Self>) {
+        tokio::task::AbortHandle::abort(&self)
+    }
+}
```
AdamGS, line 39: A weird thing here is that the task's lifetime is actually longer than the runtime, because the underlying tokio lifetime may stay alive after `TokioRuntime` is dropped.
Maybe the runtime should also have a `JoinSet` and then spawn them using [`JoinSet::spawn_on`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.spawn_on)

gatesn: I think that's fine right? The tokio case is weird anyway because you can do Handle::current() and it's not clear what lifetime that should have...

vortex-io/src/runtime/handle.rs
```diff
@@ -0,0 +1,102 @@
...
+/// A handle to an active Vortex runtime.
+///
+/// Users should obtain a handle from one of the runtime constructors and use it to spawn new
+/// async tasks or CPU-heavy worker.
+#[derive(Clone)]
+pub struct Handle<'rt>(pub(crate) Arc<dyn Runtime<'rt> + 'rt>);
+
+impl<'rt> Handle<'rt> {
+    /// Spawn a new future onto the runtime.
+    pub fn spawn<Fut, R>(&self, f: Fut) -> Task<'rt, R>
```
joseph-isaacs, line 27: is this for IO or just anything that is not CPU heavy, if so what other tasks should be spawned here?

gatesn: It's when you want parallelism (not concurrency) in your "scheduling" tasks. I/O will be handled with a file system abstraction, CPU is explicitly spawned, so it's sort of for everything else? You're expected to be awaiting either I/O or CPU almost all of the time.

But by spawning, it allows multiple threads to drive your futures at the same time. e.g. we'd spawn each split of a scan.

vortex-io/src/runtime/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+///
+/// Note: users interact with the [`Handle`] API rather than the [`Runtime`] trait.
+///
+// TODO(ngates): I/O is coming in a future change.
+pub(crate) trait Runtime<'rt>: Send + Sync {
+    /// Spawns a future to be executed on the runtime.
+    ///
+    /// The future should continue to be polled in the background by the runtime.
+    /// The returned `AbortHandle` may be used to optimistically cancel the future.
+    fn spawn(&self, fut: BoxFuture<'rt, ()>) -> AbortHandleRef<'rt>;
```
joseph-isaacs, line 40: can you add the above response here?

vortex-io/src/runtime/current.rs
```diff
@@ -0,0 +1,121 @@
...
+use smol::{Executor, block_on};
+
+use crate::runtime::Handle;
+
+/// A current thread runtime allows users to explicitly drive Vortex futures from multiple worker
+/// threads that they manage. This is useful in environments where the user already has a thread
+/// pool and wants to integrate Vortex into that pool, for example query engines.
+pub struct CurrentThreadRuntime;
+
+impl CurrentThreadRuntime {
```
joseph-isaacs, line 16: is this used anywhere?

joseph-isaacs (approved):

vortex-io/src/runtime/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+mod handle;
+pub use handle::*;
+#[cfg(not(target_arch = "wasm32"))]
+pub mod current;
+#[cfg(not(target_arch = "wasm32"))]
+pub mod multi;
+#[cfg(not(target_arch = "wasm32"))]
+pub mod single;
+#[cfg(not(target_arch = "wasm32"))]
+mod smol;
```
asubiotto, line 13: Should the tokio and smol-specific things be behind feature flags?

gatesn: Ah yep, I'll do that in the next one

vortex-io/src/runtime/current.rs
```diff
@@ -0,0 +1,121 @@
...
+        // We create an MPMC result channel and spawn a task to drive the stream and send results.
+        // This allows multiple worker threads to drive the executor while all waiting for results
+        // on the channel.
+        let (result_tx, result_rx) = kanal::unbounded();
+        executor
+            .spawn(async move {
+                futures::pin_mut!(stream);
+                while let Some(item) = stream.next().await {
+                    // Ignore send errors, which happen if all receivers are dropped.
+                    let _ = result_tx.send(item);
```
danking, line 51: It would be a bug if we received a SendError::Closed, right? We should only ever see SendError::ReceiveClosed b/c all the receivers are dropped.

https://docs.rs/kanal/latest/kanal/enum.SendError.html#variants

vortex-io/src/runtime/current.rs
```diff
@@ -0,0 +1,121 @@
...
+        S: Stream<Item = R> + Send + Unpin + 'rt,
+        R: Send + 'rt,
+    {
+        let executor = Arc::new(Executor::new());
+        let stream = f(Handle(executor.clone()));
+
+        // We create an MPMC result channel and spawn a task to drive the stream and send results.
+        // This allows multiple worker threads to drive the executor while all waiting for results
+        // on the channel.
+        let (result_tx, result_rx) = kanal::unbounded();
```
danking, line 45: Why do we use the sync version?

vortex-io/src/runtime/single.rs
```diff
@@ -0,0 +1,220 @@
...
+        };
+
+        BlockingStream { executor, stream }
+    }
+}
+
+/// Since the [`Handle`], and therefore runtime implementation needs to be `Send` and `Sync`,
+/// we cannot just `impl Runtime for LocalExecutor`. Instead, we create channels that the handle
+/// can forward its work into, and we drive the resulting tasks on a [`LocalExecutor`] on the
+/// calling thread.
```
danking, line 117: Does "calling thread" mean the thread which calls `spawn` or `spawn_cpu`?

I'm still trying to wrap my head around the model here. I thought the idea was that there's a single thread which will execute all the work. Callers (which might be on some other random thread) send work to that thread. Do I misunderstand that?

vortex-io/src/runtime/multi.rs
```diff
@@ -0,0 +1,123 @@
...
+use futures::Stream;
+use futures_util::StreamExt;
+use futures_util::stream::BoxStream;
+use smol::{Executor, block_on};
+use vortex_error::VortexExpect;
+
+use crate::runtime::Handle;
+
+/// A runtime that drives work in the background on a dedicated thread pool.
+pub struct MultiThreadRuntime<'rt> {
```
danking, line 16: If I have a MultiThreadRuntime, how do I get a reference to Executor so that I can spawn_cpu?

danking:
Oh oops I somehow missed that this was merged heh

PR #4527, opened 2025-09-04
Add more unchecked constructors

### LINKED ISSUE #4420: `vortex-array` constructor inconsistencies
So I read through the other canonical arrays to check for `new_unchecked` methods, and they are wildly inconsistent:

- `bool`: No `new_unchecked` even though there is a `try_new`
- `chunked`: `pub unsafe fn new_unchecked`
- `datetime`: Panicking `new` (and no `try_new`; maybe this is okay since it is just checking the time unit but this is still inconsistent)
- `decimal`: No `new_unchecked` even though there is a `try_new`
- `list`: No `new_unchecked` even though there is a `try_new`
- `primitive`: No `new_unchecked` NOR a `try_new`, which is terrible because it only has a `new` that panics if the inputs are invalid
- `struct`: `pub(crate) fn new_unchecked` (this isn't `unsafe`! A function being private does not make it safe!)
- `varbin`: No `new_unchecked` even though there is a `try_new`
- `varbinview`: `pub unsafe fn new_unchecked`

There are only a few that "make sense" to me:

- `constant`: No `new_unchecked` (this makes sense since it is infallible and there is no `try_new`)
- `extension`: No `new_unchecked` (this makes sense since it (seems to be) infallible and there is no `try_new`)
- `null`: No `new_unchecked` (this makes sense since it is infallible and there is no `try_new`)

I'm going to address all of these in a followup PR because this is not great.

_Originally posted by @connortsui20 in https://github.com/vortex-data/vortex/issues/4405#issuecomment-3236232355_

connortsui20:
Fixes #4420 

TODO description but mostly ready to review

### DIFF
--- a/vortex-array/src/builders/bool.rs
+++ b/vortex-array/src/builders/bool.rs
@@ -67,7 +67,7 @@ impl BoolBuilder {
             "Null count and value count should match when calling BoolBuilder::finish."
         );
 
-        BoolArray::new(
+        BoolArray::from_bool_buffer(
             self.inner.finish(),
             self.nulls.finish_with_nullability(self.dtype.nullability()),
         )
--- a/encodings/bytebool/src/array.rs
+++ b/encodings/bytebool/src/array.rs
@@ -116,7 +116,7 @@ impl CanonicalVTable<ByteBoolVTable> for ByteBoolVTable {
     fn canonicalize(array: &ByteBoolArray) -> Canonical {
         let boolean_buffer = BooleanBuffer::from(array.as_slice());
         let validity = array.validity().clone();
-        Canonical::Bool(BoolArray::new(boolean_buffer, validity))
+        Canonical::Bool(BoolArray::from_bool_buffer(boolean_buffer, validity))
     }
 }
 
--- a/encodings/datetime-parts/src/compute/cast.rs
+++ b/encodings/datetime-parts/src/compute/cast.rs
@@ -70,7 +70,7 @@ mod tests {
     #[case(Validity::NonNullable, Nullability::NonNullable)]
     #[case(Validity::AllValid, Nullability::NonNullable)]
     #[case(Validity::from_iter([true, true, true]), Nullability::Nullable)]
-    fn test_cast_to_compatibile_nullability(
+    fn test_cast_to_compatible_nullability(
         #[case] validity: Validity,
         #[case] cast_to_nullability: Nullability,
     ) {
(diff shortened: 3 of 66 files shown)


vortex-array/src/arrays/bool/serde.rs
```diff
@@ -50,7 +50,13 @@ impl SerdeVTable<BoolVTable> for BoolVTable {
...
         };
 
-        BoolArray::try_new(buffers[0].clone(), metadata.offset as usize, len, validity)
+        // SAFETY: The deserialization process ensures all invariants:
+        // - Buffer comes from serialized data with the correct byte length.
+        // - Offset and len are validated during metadata deserialization.
+        // - Validity is properly constructed from children or dtype.
+        Ok(unsafe {
+            BoolArray::new_unchecked(buffers[0].clone(), metadata.offset as usize, len, validity)
+        })
```
a10y: serde is pretty much the only place we should use `try_new`. We don't actually enforce the Bool invariants in this function body.

Also, we should check that buffers.len() == 1 and vortex_bail otherwise. As written, this will panic if there are no buffers.

a10y:
Also do you have any sense if the CodSpeed regressions are genuine?

connortsui20:
> Also do you have any sense if the CodSpeed regressions are genuine?

I guess they are genuine as it is consistent across runs. I have no idea why it swings that way though because I would have expected primitive construction to be faster...

vortex-array/src/builders/bool.rs
```diff
@@ -67,7 +67,7 @@ impl BoolBuilder {
             "Null count and value count should match when calling BoolBuilder::finish."
         );
 
-        BoolArray::new(
+        BoolArray::new_from_bool_buffer(
```
gatesn: `BoolArray::from_bool_buffer`

connortsui20:
Relevant benchmark: https://github.com/vortex-data/vortex/pull/4477/

connortsui20:
I think that there is something beyond this PR that is causing just that test to regress. If you look at the changes I made, the code is essentially identical in the `PrimitiveArray` code path. I'll try removing the `freeze` and add inline as a last check but other than that I don't really know what else it could be.

Edit: I don't think I trust these benchmarks anymore

robert3005:
#4546 has fixes for the is_sorted functions

robert3005:
I have fixed issorted and rebased this pr to include that fix

vortex-array/src/builders/bool.rs
a10y: this is a bit awkward since there is also a `From<BooleanBuffer>` impl on BoolArray

a10y (approved):
Nice

PR #4528, opened 2025-09-04
fix: use correct & safe bitwise-slicing in duckdb validity export

danking:
I use align_to to find a prefix, suffix, and 8-byte aligned slices. The 8-byte aligned slice uses the fast `copy_from_bitslice`. The other two loop.

### DIFF
--- a/vortex-duckdb/src/exporter/mod.rs
+++ b/vortex-duckdb/src/exporter/mod.rs
@@ -233,46 +233,11 @@ fn copy_from_byte_slice(
     offset: usize,
     len: usize,
 ) -> usize {
-    assert!(target.len() == len, "{} {}", target.len(), len);
-    assert!(source.len() * 8 >= len, "{} {}", source.len(), len);
-
-    let (start, middle, end) = unsafe { source.align_to::<u64>() };
-    let start = start.view_bits::<Lsb0>();
-    let middle = middle.view_bits::<Lsb0>();
-    let end = end.view_bits::<Lsb0>();
-
-    let start_offset = std::cmp::min(offset, start.len());
-    let start_len = std::cmp::min(len, start.len() - start_offset);
-
-    let start_slice = &start[start_offset..][..start_len];
-    for (slice_index, value) in start_slice.iter().enumerate() {
-        target.set(slice_index, *value);
-    }
-
-    let target_cursor = start_slice.len();
-
-    let remaining_offset = offset.saturating_sub(start.len());
-    let remaining_len = len.saturating_sub(start_slice.len());
-
-    let middle_offset = std::cmp::min(remaining_offset, middle.len());
-    let middle_len = std::cmp::min(remaining_len, middle.len() - middle_offset);
-
-    let middle_slice = &middle[middle_offset..][..middle_len];
-    target[target_cursor..][..middle_len].copy_from_bitslice(middle_slice);
-
-    let target_cursor = target_cursor + middle_len;
-
-    let end_offset = remaining_offset.saturating_sub(middle.len());
-    let end_len = remaining_len.saturating_sub(middle_slice.len());
-
-    assert!(end_offset <= end.len(), "{} {}", end_offset, end.len());
-    assert!(end_len <= end.len(), "{} {}", end_len, end.len());
-
-    let end_slice = &end[end_offset..][..end_len];
-    for (slice_index, value) in end_slice.iter().enumerate() {
-        target.set(target_cursor + slice_index, *value);
-    }
-
+    let (start, middle, end) = unsafe { target.align_to_mut::<u8>() };
+    assert!(start.is_empty());
+    assert!(end.is_empty());
+    assert!(middle.len() == len, "{} {}", middle.len(), len);
+    middle.copy_from_bitslice(&source.view_bits()[offset..][..len]);
     target.count_ones()
 }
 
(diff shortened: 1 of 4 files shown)


vortex-duckdb/src/exporter/mod.rs
```diff
@@ -210,42 +211,82 @@ impl VectorExt for Vector {
...
-                    }
-                } else if true_count == 0 {
-                    unsafe { self.ensure_validity_slice(len) }.fill(false);
-                } else {
-                    // SAFETY: Caller guaranteees this.
-                    let validity = unsafe { self.ensure_validity_slice(len) };
-                    validity[..len].copy_from_bitslice(sliced_bits);
-                }
-                true_count == 0
+                // SAFETY: Caller guaranteees this.
```
a10y:
```suggestion
                // SAFETY: Caller guarantees this.
```

robert3005 (commented):
How about you cast down the u64 to u8 and append the u8 bitslice

danking:
@robert3005 I don't see any methods on [BitSlice](https://docs.rs/bitvec/latest/bitvec/slice/struct.BitSlice.html) which chnage the `T` type parameter nor do I see any methods that return a `&[T]`. Without bigger changes to vortex-duckdb, I don't see how we can get a `BitSlice<u8>` from the `BitSlice<u64>` we have.

danking:
Ah, okay, I can use align_to.

robert3005:
I thought you could unsafe cast?

danking:
@robert3005 

I had a real bad time making the Divan benchmarks work. Do you know how to do this correctly? I have to run the benchmarks with `cargo bench --features bench`. If I add the bench feature as a required-features in `[[bench]]` I get no benchmark output. It's really bizarre.

Anyway, my version is indeed faster but probably not enough to matter. I went with your approach.
```
vortex_duckdb                                  fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ exporter                                                  │               │               │               │         │
   ╰─ benchmarks                                             │               │               │               │         │
      ├─ test_bench_copy_from_byte_slice                     │               │               │               │         │
      │  ├─ 64                                 40.59 ns      │ 13.08 µs      │ 41.59 ns      │ 185.4 ns      │ 100     │ 100
      │  ├─ 1024                               227.4 ns      │ 239.1 ns      │ 230 ns        │ 231.6 ns      │ 100     │ 3200
      │  ├─ 32768                              6.457 µs      │ 7.958 µs      │ 6.499 µs      │ 6.544 µs      │ 100     │ 100
      │  ├─ 1048576                            192.9 µs      │ 256.2 µs      │ 200.5 µs      │ 202.3 µs      │ 100     │ 100
      │  ╰─ 33554432                           6.2 ms        │ 7.506 ms      │ 6.329 ms      │ 6.405 ms      │ 100     │ 100
      ╰─ test_bench_copy_from_byte_slice_fast                │               │               │               │         │
         ├─ 64                                 37.99 ns      │ 51.34 ns      │ 38.65 ns      │ 40.79 ns      │ 100     │ 12800
         ├─ 1024                               155.8 ns      │ 180.5 ns      │ 161 ns        │ 161.8 ns      │ 100     │ 3200
         ├─ 32768                              4.207 µs      │ 4.749 µs      │ 4.249 µs      │ 4.286 µs      │ 100     │ 100
         ├─ 1048576                            136.1 µs      │ 173.8 µs      │ 141.9 µs      │ 142.1 µs      │ 100     │ 100
         ╰─ 33554432                           4.397 ms      │ 4.933 ms      │ 4.535 ms      │ 4.564 ms      │ 100     │ 100
```

danking:
@robert3005 OK, so, this passes my tests but it still fails in the statpopgen benchmark. AFAICT `align_to_mut` only works when you're casting to a bigger type. If you try to cast to a smaller type you blow up [here](https://github.com/ferrilab/bitvec/blob/main/src/ptr/span.rs#L490-L493). In particular, `align_to_mut` wants to return `(BitSlice<u64>, BitSlice<u8>, BitSlice<u64>)` where the middle slice is as big as possible. The last BitSlice is always empty but 7 out of 8 times it is not 8-byte aligned. 

What's the unsafe cast I can do here? In particular, the vortex-duckdb stuff never gives me an `&mut [u64]` it gives me a `BitSlice<u64, Lsb0>`. I'm not sure how to make that into a `BitSlice<u8, Lsb0>`.

```
fn copy_from_byte_slice(
    target: &mut BitSlice<u64, Lsb0>,
    source: &[u8],
    offset: usize,
    len: usize,
) -> usize {
    let (start, middle, end) = unsafe { target.align_to_mut::<u8>() };
    assert!(start.is_empty());
    assert!(end.is_empty());
    assert!(middle.len() == len, "{} {}", middle.len(), len);
    middle.copy_from_bitslice(&source.view_bits()[offset..][..len]);
    target.count_ones()
}
```

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -210,42 +211,47 @@ impl VectorExt for Vector {
...
-                } else if true_count == 0 {
-                    unsafe { self.ensure_validity_slice(len) }.fill(false);
-                } else {
-                    // SAFETY: Caller guaranteees this.
-                    let validity = unsafe { self.ensure_validity_slice(len) };
-                    validity[..len].copy_from_bitslice(sliced_bits);
-                }
-                true_count == 0
+                // SAFETY: Caller guarantees this.
+                let target = &mut unsafe { self.ensure_validity_slice(len) }[..len];
```
robert3005, line 215: @danking I thought you can convert this to a &mut [u8] and then convert this to a bitslice and copy into that?

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -210,42 +211,47 @@ impl VectorExt for Vector {
...
     }
 }
 
+/// Copy the sliced bits from source into target. Returns the number of now set bits in target.
+///
+/// Offset and length are a _bit_ offset and a _bit_ length into source.
+///
+/// `target.len()` must equal `len`.
+fn copy_from_byte_slice(
+    target: &mut BitSlice<u64, Lsb0>,
```
robert3005, line 231: you could take here `&mut [u64]` then add an assert and unsafe cast it to u8?

vortex-duckdb/src/exporter/mod.rs
danking: @robert3005 AFAICT, there's no way to convert a `BitSlice<u64>` to a `BitSlice<u8>` but I can change all these methods to return `&mut [u64]` instead.

danking:
Trying a new PR.

PR #4530, opened 2025-09-04
fix: fix deserialization bug in pco

danking:
The buffer visitor pretty clearly visits all the chunk metadata followed by all the page metadata.

I can see a cache efficiency argument for meta0 meta0_page1 meta0_page2 ... meta1 ... but this PR at least allows deserialization of already-written pco arrays.

### DIFF
--- a/encodings/pco/src/serde.rs
+++ b/encodings/pco/src/serde.rs
@@ -61,17 +61,17 @@ impl SerdeVTable<PcoVTable> for PcoVTable {
             vortex_bail!("PcoArray expected 0 or 1 child, got {}", children.len());
         };
 
-        let mut chunk_metas = vec![];
-        let mut pages = vec![];
-        let mut buffer_idx = 0;
-        for chunk_info in &metadata.chunks {
-            chunk_metas.push(buffers[buffer_idx].clone());
-            buffer_idx += 1;
-            for _ in &chunk_info.pages {
-                pages.push(buffers[buffer_idx].clone());
-                buffer_idx += 1;
-            }
-        }
+        let chunk_metas = buffers[..metadata.chunks.len()]
+            .iter()
+            .cloned()
+            .collect::<Vec<_>>();
+        let pages = buffers[metadata.chunks.len()..]
+            .iter()
+            .cloned()
+            .collect::<Vec<_>>();
+
+        let expected_n_pages = metadata.chunks.iter().map(|info| info.pages.len()).sum();
+        assert_eq!(pages.len(), expected_n_pages);
 
         Ok(PcoArray::new(
             chunk_metas,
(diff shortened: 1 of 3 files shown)


encodings/pco/src/serde.rs
```diff
@@ -61,17 +61,15 @@ impl SerdeVTable<PcoVTable> for PcoVTable {
...
-        }
+        let chunk_metas = buffers[..metadata.chunks.len()].to_vec();
+        let pages = buffers[metadata.chunks.len()..].to_vec();
+
+        let expected_n_pages = metadata
+            .chunks
+            .iter()
+            .map(|info| info.pages.len())
+            .sum::<usize>();
+        assert_eq!(pages.len(), expected_n_pages);
```
a10y: use vortex_bail!

SerdeVTable::build should never panic

encodings/pco/src/serde.rs
```diff
@@ -61,17 +61,15 @@ impl SerdeVTable<PcoVTable> for PcoVTable {
...
-        for chunk_info in &metadata.chunks {
-            chunk_metas.push(buffers[buffer_idx].clone());
-            buffer_idx += 1;
-            for _ in &chunk_info.pages {
-                pages.push(buffers[buffer_idx].clone());
-                buffer_idx += 1;
-            }
-        }
+        let chunk_metas = buffers[..metadata.chunks.len()].to_vec();
+        let pages = buffers[metadata.chunks.len()..].to_vec();
```
a10y, line 66: similarly, add a vortex_ensure! to verify that the buffers has the appropriate length, so we don't panic here

encodings/pco/src/serde.rs
```diff
@@ -61,17 +61,15 @@ impl SerdeVTable<PcoVTable> for PcoVTable {
...
-        let mut buffer_idx = 0;
-        for chunk_info in &metadata.chunks {
-            chunk_metas.push(buffers[buffer_idx].clone());
-            buffer_idx += 1;
-            for _ in &chunk_info.pages {
-                pages.push(buffers[buffer_idx].clone());
-                buffer_idx += 1;
-            }
-        }
+        let chunk_metas = buffers[..metadata.chunks.len()].to_vec();
```
a10y, line 65: it's too late to change this now, but this `chunks` Vec should probably have been another u32 Buffer

robert3005: there's really no difference between this being metadata and a buffer

encodings/pco/src/serde.rs
danking: Done.

encodings/pco/src/serde.rs
danking: Done.

danking:
Okay I think this is ready.

a10y (approved):

PR #4534, opened 2025-09-04
fix: teach vortex-duckdb to safely copy validity

danking:
First attempt here: https://github.com/vortex-data/vortex/pull/4528.

I suspect the reason the API was designed to return BitSlice rather than `&mut [u64]` is that there's an ambiguity as to which bit is bit zero.

I doubt this change affects the benchmark results reported in #4528. This implementation is probably ~140% the runtime of the version that uses u64 copies where possible.

### DIFF
--- a/vortex-duckdb/src/convert/vector.rs
+++ b/vortex-duckdb/src/convert/vector.rs
@@ -467,7 +467,7 @@ mod tests {
 
         // Set middle element as null
         // SAFETY: Vector was created with this length.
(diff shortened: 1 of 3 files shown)


danking:
@robert3005 I actually think this is a bad interface. It exposes the raw `[u64]` even though the number of bits in use might not be a multiple of 64.

robert3005 (approved):

PR #4540, opened 2025-09-05
File I/O abstraction for new runtimes

gatesn:
Again, not hooked up to anything.

This allows us to open files against a runtime, avoids an allocation per I/O request (no async_trait), and allows runtimes to opt-in to using a `!Send` implementation


lwwmanning:
@claude review this PR please

onursatici (commented):
It feels a bit weird that io source has both drive_send and drive_local, feels like that is more of a runtime concern? 

Also now drive methods don't spawn by themselves but the future they return have to be polled for them to make progress? If I create a scan in the runtime world would I get back a stream of split tasks, and this io future that I need to poll myself?

gatesn:
> It feels a bit weird that io source has both drive_send and drive_local, feels like that is more of a runtime concern?

It's _sort of_ a runtime concern. The issue is, some runtimes support !Send futures, and some don't. For a multi-thread Tokio runtime, there simply is no way to run !Send futures if we don't own the entry-point of the runtime (by spawning a new thread-pool for example). So this allows I/O implementations to return an optimized (i.e. without synchronization primitives) version of a reader for runtimes that _do_ support it.

The current thread runtime currently uses a Send smol executor, but I want to eventually build a more custom runtime that pins I/O to a thread so we can run !Send futures.

The example here is io_uring being thread-local, so we can avoid a synchronization primitive for spawning I/O tasks in this world _iff_ the user chooses a runtime that supports it.

> Also now drive methods don't spawn by themselves but the future they return have to be polled for them to make progress? If I create a scan in the runtime world would I get back a stream of split tasks, and this io future that I need to poll myself?

The drive functions don't, but `Handle::open` delegates to `Runtime::spawn_io`, which _does_ spawn the task onto the given runtime. So I/O is driven whenever the runtime is driven.

onursatici (approved):

PR #4544, opened 2025-09-05
fix[duckdb]: handle dict with constant value with validity exporter

joseph-isaacs:


vortex-duckdb/src/exporter/dict.rs
```diff
@@ -33,6 +33,23 @@ pub(crate) fn new_exporter(
 ) -> VortexResult<Box<dyn ColumnExporter>> {
     // Grab the cache dictionary values.
     let values = array.values();
+    if let Some(constant) = values.as_opt::<ConstantVTable>() {
+        let constant_codes = ConstantArray::new(constant.scalar().clone(), array.codes().len());
```
lwwmanning: this isn't really `codes` is it? it's the scalar from `values`, which is constant

joseph-isaacs: was a bad name

robert3005 (approved):

PR #4546, opened 2025-09-05
Fix: IsSortedKernel can return unknown, fix IsSorted for DictArray

robert3005:

### DIFF
--- a/encodings/dict/src/compute/is_sorted.rs
+++ b/encodings/dict/src/compute/is_sorted.rs
@@ -9,21 +9,15 @@ use crate::{DictArray, DictVTable};
 
 impl IsSortedKernel for DictVTable {
     fn is_sorted(&self, array: &DictArray) -> VortexResult<Option<bool>> {
-        if let Some((values_sorted, codes_sorted)) =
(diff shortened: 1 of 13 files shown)


encodings/dict/src/compute/is_sorted.rs
```diff
@@ -8,15 +8,26 @@ use vortex_error::VortexResult;
 use crate::{DictArray, DictVTable};
 
 impl IsSortedKernel for DictVTable {
-    fn is_sorted(&self, array: &DictArray) -> VortexResult<bool> {
-        // TODO(ngates): we should change these kernels to return Option<bool> to allow for "unknown"
-        let is_sorted = is_sorted(array.values())? && is_sorted(array.codes())?;
-        Ok(is_sorted)
+    fn is_sorted(&self, array: &DictArray) -> VortexResult<Option<bool>> {
+        if let Some((values_sorted, codes_sorted)) =
```
joseph-isaacs: can you not match true here?

robert3005: I think I can

joseph-isaacs (approved):

PR #4547, opened 2025-09-05
feat: vortex.mds provides support for the Vortex format in MDS

danking:


delta003 (changes requested):
let's discuss this internally

PR #4548, opened 2025-09-05
feature: Bloom Filters

a10y:
This PR (WIP) augments our existing `DictLayout` to write bloom filters that can then be used for pruning.

This is backwards-compatible, the metadata includes an additional optional protobuf field and if that field is present, it will point at a segment ID that contains the bloom filters.

We implement the Split block Bloom Filter from https://arxiv.org/pdf/2101.01719, with xxHasher. This is effectively what Parquet does (and StarRocks, and a few others).

The bloom filter implementation borrows heavily from the parquet-rs one, as well as the pseudocode in the Parquet documentation https://parquet.apache.org/docs/file-format/bloomfilter/

## Settings

Writing of bloom filters are currently disabled by default, but are enabled for SQL benchmarks TPC-H and FineWeb.

## Tokenization and Pruning

We tokenize the dictionary values upon insertion so we can support pruning of partial-string matches.

Now pruning occurs for the following expression types

- `$ = Literal`
- `$ LIKE 'EXACT'`
- `$ LIKE 'prefix%'`
- `$ LIKE '%suffix'`
- `$ LIKE '%substr%'`

Not-equals cannot be pruned using an approximate membership data structure like a bloom filter.

## Extension

Extending to other data types like datetime, ints, decimals, extension types like geocolumns etc.


gatesn:
Any reason the bloom filters aren't just a binary column in the zone map? In theory, we could write the zone map using a StructLayout such that only the required columns are read. 

Although I guess we'd probably want a layout that allows us to group some set of columns together to avoid too many flat layouts.

joseph-isaacs:
Why did you choose a bloom filter instead of an another approximate set membership data-structure (cuckoo or quotient filter)?

robert3005:
FWIW the SOA is either what you did or https://arxiv.org/pdf/2405.10253, there's no comparison of these two in the literature yet

a10y:
> Why did you choose a bloom filter instead of an another approximate set membership data-structure (cuckoo or quotient filter)?

Yea, I mostly just ran with SBBF since there is a lot of prior art behind it, Cuckoo might be better but the infra is the same regardless. I also haven't done anything to explicitly perf test the hasher or the bloom filter itself, but it deserves its own microbenchmarks. StarRocks has some [AVX2 kernels](https://github.com/StarRocks/starrocks/blob/77a4dac33b199576aa70b77a5b9e29db1c1e187a/be/src/exprs/runtime_filter.h#L78-L80) as well that might be useful to look into in the future

> Any reason the bloom filters aren't just a binary column in the zone map? In theory, we could write the zone map using a StructLayout such that only the required columns are read.

That is true, but then it gets a bit weird: how do you implement pushdown? For ZoneMaps we create pruning expressions and then apply them against the stats array like it were any other array. Following the analogy, we'd write nullable bloom filter Binary columns in the zone map, then implement a new `BloomFilterCheck` expression which receives a Binary child and a Utf8 child, and would deserialize the binary child as well.

You also want all of the bloom filters to be packed into a single segment, roughly, since all of it gets touched in pruning. So then you need to make a special layout for it or whatever. All this vs just shoving it explicitly into a segment and wiring it in as another step in the pruning pipeline...not sure really.

Initial results here look like there are some promising speedups to be gained, particularly against S3. Need to figure out why there are some perf degredations, and probably play around with the filter size

robert3005:
Having fewer special things should be the way. Stats are already at the end of the file. If we want to use bloom filters for anything else it shoudnt be the case that nothing works. The special layout we are suggesting is likely generally useful concept that would get used outside of stats

gatesn:
Not sure what you're suggesting @robert3005 ? I think I agree with @a10y that we can treat bloom filters specially? The problem we may have is that LayoutReaders are a tree, so it's not clear if the bloom filter should come before / after pruning or basically where in the tree? And what about if we add other indexing structures? But that feels like it can be solved later.

robert3005:
Im suggesting they’re a sibling of other stats and essentially a extension dtype on binary dtype. But it’s also easy to add new layouts or stats so we can always implement another version of it.

vortex-expr/src/exprs/like.rs
```diff
@@ -149,7 +149,13 @@ impl DisplayAs for LikeExpr {
     fn fmt_as(&self, df: DisplayFormat, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match df {
             DisplayFormat::Compact => {
-                write!(f, "{} LIKE {}", self.child(), self.pattern())
+                write!(
+                    f,
+                    "{} {}LIKE {}",
+                    self.child(),
+                    if self.negated() { "NOT " } else { "" },
```
a10y, line 156: to render `NOT LIKE` properly

a10y:
I'm struggling to reproduce a lot of the slowdowns that I'm seeing in the NVME runs here

image

(top is branch, bottom is develop)

I'm convinced they're real but I need to find a better way to replicate and track them down.

There's probably something suboptimal in the way that we're laying out the segments. Right now, I have the bloom filter segments for all dict chunks be contiguous. We could change it to be contiguous with the dict values segment, but that ends up causing many small reads for each chunk. Something feels a bit off.

vortex-layout/src/layouts/dict/bloom/sbbf/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub fn new(blocks: BufferMut<Block>) -> Self {
+        Self { blocks }
+    }
+
+    /// Serialize the filter to a new buffer.
+    pub fn serialize(&self) -> ByteBuffer {
+        BufferMut::copy_from(&self.blocks)
+            .freeze()
+            .into_byte_buffer()
+    }
```
robert3005: Could this be just `self.clone().freeze().into_byte_buffer()`? Though it's likely the same thing overall

PR #4557, opened 2025-09-06
Refactor VortexWrite to return Stream<Buffer> and Footer

### LINKED ISSUE #4064: Actually be able to pass a Footer when opening a file
I'd like to pass an externally stored footer when opening a file https://github.com/vortex-data/vortex/blob/029af03f001821c0803bd4f6e15d789f38cbc80a/vortex-file/src/open.rs#L35

Although this is possible in the API, in practice, it's not usable. There is no (good) way to obtain a Footer at write time, and even if I work around that (for example, by opening a written file again), I have no way to store the footer.

### LINKED ISSUE #4552: Pass TaskExecutor (or Handle) into LayoutStrategy::write_stream
We currently wrap up the executor inside the strategy, but this doesn't make sense in a world where the caller gets to arbitrarily choose the runtime handle.

There's no point adding runtime lifetimes to all the strategies, we should just pass a handle into write_stream such that strategies still have the chance to spawn work etc.

gatesn:
I attempted this in #4549 , but it was hanging, which has been resolved by switching over to using the new runtimes API for execution. I could have made this change separately, but both this PR and switching to runtimes adds an argument to write_stream function so would have been a conflict nightmare 😢 

---

This revives an old branch I had worked on with @onursatici about changing the Vortex write API to return a Stream and be completely independent from the actual file sink.

This PR exposes a new write_stream function, and we can remove the old write(VortexWrite, ArrayStream) APIs in a future PR. We should decide whether we want to make that change prior to releasing to avoid multiple user-facing breaks to this Rust API.

In order to make this change easier, we now pass an explicit eof-of-file pointer into the writers that they can use to create new streams that occur "just before the end of the file", as is required for zone maps for example.

We also remove the custom transpose_stream implementation that was quite dependent on the polling behaviour of `try_join_all`, which alarmingly changes when the number of futures reaches 40. 

Closes #4118
Closes #4549 
Fixes #4552 
Fixes #4064


gatesn:
Looks like compress time for wide tables needs some work. I may not be spawning something I should be, will check

vortex-layout/src/layouts/zoned/writer.rs
```diff
@@ -115,29 +109,30 @@ where
...
         let block_size = self.options.block_size;
+
+        // We create a new SequencePointer for the stats table so that we can write it just
+        // before the end of the file.
+        let (stats_eof, eof) = eof.split();
+
         let data_layout = self
             .child
-            .write_stream(&ctx, sequence_writer.clone(), stream)
+            .write_stream(ctx, segment_sink, stream, eof, handle.clone())
```
onursatici: This looks a bit funny to me. I reckon this would only work if you don't use the `eof` at all anywhere in child, and its children. Because `stats_eof` is still alive, calling collapse on any sequence id derived from eof would block forever

AdamGS:
Can we return a type that can (potentially) include other stuff? It would be really nice to return rows or bytes written

vortex-layout/src/layouts/zoned/writer.rs
gatesn: With the new SingleThreadRuntime (vs LocalTaskExecutor), it wouldn't block forever. But I think you're right in that it would cause everything to be buffered in memory until stats are written?

I have swapped it around so stats comes after the eof we pass to the child.

gatesn:
Ideally we return the footer right? 

Any ideas on an API that returns both? Maybe this?

```rust
let footer = VortexWriteOptions::new()
  .write(|buffers| ...)
  .await?;
```

vortex-io/src/runtime/blocking.rs
```diff
@@ -0,0 +1,30 @@
...
+use futures::Stream;
+
+use crate::runtime::Handle;
+
+/// A generic API blocking entry points to runtimes.
+pub trait BlockingRuntime {
+    /// Associated type for the blocking iterator returned by `block_on_stream`.
+    type BlockingIterator<'a, R>: Iterator<Item = R> + 'a
+    where
+        R: 'a;
```
robert3005:
```suggestion
    type BlockingIterator<'a, R: 'a>: Iterator<Item = R> + 'a;
```

vortex-io/src/runtime/tokio.rs
```diff
@@ -9,11 +9,12 @@ use tokio::runtime::Handle as TokioHandle;
 use crate::runtime::{AbortHandle, AbortHandleRef, Handle, IoTask, Runtime};
 
 /// A Vortex runtime that drives all work the currently scoped Tokio runtime.
-pub struct TokioRuntime(TokioHandle);
+#[allow(dead_code)]
```
robert3005, line 12: This shouldn't be dead - we are missing feature flags?

vortex-layout/src/layouts/dict/writer.rs
```diff
@@ -139,31 +140,33 @@ impl LayoutStrategy for DictStrategy {
         let dtype_clone = dtype.clone();
         let child_layouts_fut = async move {
             let mut children = Vec::new();
-            let mut runs = DictEncodedRuns::new(Box::pin(encoded_rx));
+            let mut runs = DictEncodedRuns::new(Box::pin(encoded_rx.into_stream()));
```
robert3005: do you need the additional pin here?

vortex-layout/src/layouts/dict/writer.rs
```diff
@@ -120,16 +121,16 @@ impl LayoutStrategy for DictStrategy {
         let mut dict_stream = dict_encode_stream(stream, options.constraints);
 
         // 2.a spawn encoding codes
-        let (mut encoded_tx, encoded_rx) = mpsc::channel(options.encoded_buffer_size);
-        let encode_handle = executor.spawn({
+        let (encoded_tx, encoded_rx) = kanal::bounded_async(1);
```
robert3005: why is this size 1 now?

robert3005 (commented):
Tried to read through all of it. Roughly makes sense, I left some questions

vortex-io/src/runtime/tokio.rs
gatesn: The TokioRuntime struct is a placeholder, unless you have the tokio feature enabled, then we actually use its TokioHandle field 🤷

vortex-ffi/src/sink.rs
```diff
@@ -59,9 +59,9 @@ pub unsafe extern "C-unwind" fn vx_array_sink_open_file(
         let array_stream = ArrayStreamAdapter::new(file_dtype.clone(), ReceiverStream::new(rx));
 
         let writer = get_runtime().spawn(async move {
-            let file = File::create(path).await?;
+            let mut file = File::create(path).await?;
```
onursatici, line 62: we should call sync_all before we drop this, maybe

vortex-duckdb/src/copy.rs
```diff
@@ -117,14 +116,9 @@ impl CopyFunction for VortexCopyFunction {
             ArrayStreamAdapter::new(bind_data.dtype.clone(), ReceiverStream::new(rx));
 
         let writer = COPY_RUNTIME.spawn(async move {
-            let file = File::create(file_path).await?;
+            let mut file = File::create(file_path).await?;
```
onursatici, line 119: potential `sync_all` here as well

vortex-layout/src/layouts/dict/writer.rs
```diff
@@ -3,34 +3,36 @@
...
+use crate::segments::SegmentSinkRef;
+use crate::sequence::{
+    SendableSequentialStream, SequenceId, SequencePointer, SequentialStream,
+    SequentialStreamAdapter, SequentialStreamExt,
 };
+use crate::{IntoLayout, LayoutRef, LayoutStrategy, OwnedLayoutChildren};
 
 #[derive(Clone)]
 pub struct DictLayoutOptions {
     pub constraints: DictConstraints,
```
onursatici, line 35: if this has max_len > u16::MAX now, we might panic in the dict encoder because we hardcode codes dtype to be u16

vortex-layout/src/layouts/zoned/writer.rs
```diff
@@ -71,65 +68,83 @@ impl ZonedStrategy {
...
                         let (sequence_id, chunk) = chunk?;
                         chunk.statistics().compute_all(&stats)?;
                         VortexResult::Ok((sequence_id, chunk))
-                    }
-                    .boxed()
+                    })
                 })
-                .map(move |stats_future| executor.spawn(stats_future))
-                .buffered(self.options.parallelism),
+                .buffered(self.options.concurrency),
```
onursatici, line 98: nit: isn't this still parallelism, we spawn cpu?

gatesn: Not necessarily, depends on the runtime

vortex-layout/src/layouts/dict/writer.rs
gatesn: Correct, will file a FLUP

onursatici (approved):

PR #4560, opened 2025-09-07
Fix: Don't pushdown binary numeric operations into dict arrays

robert3005:
In cases where codes don't cover all values of the dictionary operations on the dictionary can fail while they wouldn't on regular array. We need to perform a rewrite on the pushed operation but we don't have machinery for it


encodings/dict/src/compute/binary_numeric.rs
```diff
@@ -26,6 +26,32 @@ impl NumericKernel for DictVTable {
...
+                > Scalar::primitive_value(
+                    PValue::U8(0),
+                    array.codes().dtype().as_ptype(),
+                    array.codes().dtype().nullability(),
+                )
+                || min_max.max
+                    < Scalar::primitive_value(
+                        PValue::U32(
+                            u32::try_from(array.values().len())
+                                .vortex_expect("dictionary length didn't fit in u32"),
```
joseph-isaacs: what about u64?

encodings/dict/src/compute/binary_numeric.rs
```diff
@@ -26,6 +26,11 @@ impl NumericKernel for DictVTable {
         };
         let rhs_const_array = ConstantArray::new(rhs_scalar, array.values().len()).into_array();
 
+        // TODO(robert): this can't be pushed down due to overflow behaviour
```
joseph-isaacs:
```suggestion
        // TODO(robert): this can't be pushed down due to overflow behaviour. 
        // Values not referenced by the codes could overflow
```

encodings/dict/src/compute/binary_numeric.rs
```diff
@@ -26,6 +26,12 @@ impl NumericKernel for DictVTable {
         };
         let rhs_const_array = ConstantArray::new(rhs_scalar, array.values().len()).into_array();
 
+        // TODO(robert): this can't be pushed down due to overflow behaviour.
+        // Values not referenced by the codes could overflow
```
a10y: I don't understand why this differs from normal overflow behavior?

a10y: Oh i see, because it can trigger an error that otherwise would not happen

a10y: Why does this only apply to unsigned ints? Signed ints can still overflow no?

robert3005: you're right, I focused on the case I have seen but in principle this is never fine to do

a10y (approved):
We can probably do the pushdown when we know from stats that this will not overflow right? But for now I guess we can drop this. Not sure the performance effects

robert3005:
I filed an issue so we don’t forget to make a better solution but otoh I wasn’t sure of a behaviour preserving way to push things down

PR #4561, opened 2025-09-07
VarBinArray#slice doesn't slice bytes

robert3005:

### DIFF
--- a/vortex-array/src/arrays/varbin/mod.rs
+++ b/vortex-array/src/arrays/varbin/mod.rs
@@ -4,12 +4,11 @@
 use std::fmt::Debug;
 
 pub(crate) use compute::compute_min_max;
-use num_traits::PrimInt;
+use itertools::Itertools;
+use num_traits::{AsPrimitive, PrimInt};
 use vortex_buffer::ByteBuffer;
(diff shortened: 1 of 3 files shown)


robert3005:
This got snuck in another big pr #4527

vortex-array/src/arrays/varbin/mod.rs
```diff
@@ -213,30 +199,27 @@ impl VarBinArray {
...
+                    .windows(2)
+                    .map(|o| (o[0].as_(), o[1].as_()))
+                    .enumerate()
+                {
+                    if validity.is_null(i) {
+                        continue;
+                    }
+
+                    let string_bytes = &bytes.as_ref()[start..end];
+                    simdutf8::basic::from_utf8(string_bytes).map_err(|_| {
```
joseph-isaacs, line 215: why do we not validate the whole array vs just part (as before?

Does this make the slice operations non-constant

robert3005: Slice operation shouldn't have to revalidate array, the change here is to remove validation from slice and while I was here I changed the validation to actually validate the whole array, validating part is useless from the users point of view

joseph-isaacs (approved):

PR #4567, opened 2025-09-08
use the footer on count_rows where possible

onursatici:


vortex-python/src/dataset.rs
```diff
@@ -168,6 +168,14 @@ impl PyVortexDataset {
         split_by: Option<usize>,
         row_range: Option<(u64, u64)>,
     ) -> PyResult<usize> {
+        if row_filter.is_none() {
```
robert3005, line 171: since you're already here - can you also change the main logic of this function to avoid exporting to arrow?

robert3005 (approved):

PR #4572, opened 2025-09-08
Remove transmutes from vortex-io/runtime

gatesn:
See https://without.boats/blog/the-scoped-task-trilemma/ for why we can't just have nice things like structured concurrency 😢 

I therefore gave up scoping handles with artificial lifetimes and fall back to just holding a phantom lifetime in an open file handle. It still doesn't stop you passing a static future back out of a block_on call... which would then deadlock if you tried to poll it, but hey ho.


robert3005:
How did you find that post? It's sad that it explains our problem precisely. !Leak in rust would have been cool.

gatesn:
@onursatici sent it to me. There's a few other explanations on the various runtime issues, e.g. [Tokio's ](https://github.com/tokio-rs/tokio/issues/3162)

robert3005 (approved):

PR #4575, opened 2025-09-09
Make `FieldName` an actual type an not just an alias

AdamGS:
Makes `FieldName` its own type (like `FieldNames`), including many `PartialEq` implementations to make it as ergonomic as possible, while keeping other behaviors.

### DIFF
--- a/vortex-array/src/array/display/mod.rs
+++ b/vortex-array/src/array/display/mod.rs
@@ -330,7 +330,7 @@ mod test {
     #[test]
     fn test_empty_struct() {
         let s = StructArray::try_new(
-            FieldNames::from(vec![]),
+            FieldNames::empty(),
             vec![],
             3,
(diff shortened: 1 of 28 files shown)


gatesn (approved):

joseph-isaacs:
Is it worth adding a iter_str() method to filter names that iter().map(|s| s.as_ref()?

robert3005:
@AdamGS and I figured that this would actually be better if FieldName wasn't a type alias but a wrapper around `Arc<str>`

gatesn:
We have a lot of Arc<str> things right? Spiral has a macro to generate wrapper objects with all the conversions / From / other traits that you need. Wonder if we should just publish as a lib / move into vortex-utils.

AdamGS:
Yeah the spiral macro could work here, it might have some differences around serde, but otherwise should work.

Issue #4034, opened 2025-07-28
Support datafusion dynamic filter expressions


AdamGS:
Starting with DataFusion's 49 release, they support [dynamic filter expressions](https://docs.rs/datafusion/latest/datafusion/physical_expr/expressions/struct.DynamicFilterPhysicalExpr.html), which give a significant performance boost in cases like top-k queries.

A lot of performance wins in DataFusion are driven by this change, so we should push it to Vortex competitive.

The latest PR trying to push this is #5205, but work have stalled a bit.

robert3005:
While they do support them internally they never push them down through optimizer so we have to wait for more datafusion features to land before we do anything here. Let's reopen this once there's a release we can test the feature with

robert3005:
I have forgotten that we filter expression, it's there

AdamGS:
Just to expand on the context - I would love to support datafusion dynamic expressions, but they are extremely flexible. They probably don't fit with our current dynamic expression implementation (that was built for DuckDB).
I've made an [attempt](https://github.com/vortex-data/vortex/pull/4144) that I didn't push all the way to the end to add a new expression just to wrap datafusion dynamic expressions, but I'm not sure how to handle cases where the inner expression is actually something we don't know how to translate, because once we say we accept it, I'm not sure upstream physical operators will run it again.

adriangb:
> I'm not sure how to handle cases where the inner expression is actually something we don't know how to translate, because once we say we accept it, I'm not sure upstream physical operators will run it again

I think in general you should push down the expression and do a best-effort to evaluate it at runtime but say `No`/`Unsupported` for dynamic expressions. This is always safe. Most likely it won't be "re evaluated" upstream e.g. HashJoinExec produces a dynamic physical expression but then doesn't actually use it. TopK I think does use it to pre-filter rows before comparing them in the heap, in theory it could make use of pushdown information to skip this but that seems like a complex API to get right, probably not worth the effort.

[...]

PR #4577, opened 2025-09-09
Bump DataFusion to 50 and arrow to 56

AdamGS:
Just made sure there's no breaking changes coming our way. Includes support for the two new Arrow decimal types (32 and 64 bits), but without converting from Vortex dtype into them.

Codspeed improvements are due to a new inline hint on arrow's `BitIterator::next` (found by @0ax1 🥳).

Open issues:
- [ ] https://github.com/apache/datafusion/issues/17489 - mostly affects if/how much support we want to provide to decimals, or whether we need to provide some other solution for DataFusion.
- [ ] Issue with followups - https://github.com/vortex-data/vortex/issues/4668

### DIFF
--- a/bench-vortex/Cargo.toml
+++ b/bench-vortex/Cargo.toml
@@ -29,6 +29,7 @@ datafusion = { workspace = true, features = [
     "parquet",
     "datetime_expressions",
     "nested_expressions",
+    "unicode_expressions",
 ] }
 datafusion-common = { workspace = true }
 datafusion-physical-plan = { workspace = true }
--- a/vortex-array/src/arrow/compute/to_arrow/canonical.rs
+++ b/vortex-array/src/arrow/compute/to_arrow/canonical.rs
@@ -10,6 +10,7 @@ use arrow_array::types::{
 };
 use arrow_array::{
     Array, ArrayRef as ArrowArrayRef, ArrowPrimitiveType, BooleanArray as ArrowBoolArray,
+    Decimal32Array as ArrowDecimal32Array, Decimal64Array as ArrowDecimal64Array,
     Decimal128Array as ArrowDecimal128Array, Decimal256Array as ArrowDecimal256Array,
     FixedSizeListArray as ArrowFixedSizeListArray, GenericByteArray, GenericByteViewArray,
     GenericListArray, NullArray as ArrowNullArray, OffsetSizeTrait,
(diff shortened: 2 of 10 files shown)


bench-vortex/Cargo.toml
```diff
@@ -29,6 +29,7 @@ datafusion = { workspace = true, features = [
     "parquet",
     "datetime_expressions",
     "nested_expressions",
+    "unicode_expressions",
```
AdamGS, line 32: I'm not sure why this feature wasn't needed before, as tpch q22 uses `substring`.

I posted about in the DF release thread so maybe we'll find an explanation.

AdamGS: we got an answer - https://github.com/apache/datafusion/issues/16799#issuecomment-3270855890

vortex-array/src/arrow/compute/to_arrow/canonical.rs
```diff
@@ -116,6 +117,34 @@ impl Kernel for ToArrowCanonical {
             {
                 to_arrow_primitive::<Float64Type>(array)
             }
+            (Canonical::Decimal(array), DataType::Decimal32(precision, scale)) => {
+                if array.decimal_dtype().precision() != *precision
```
0ax1, line 121: Is this check too restrictive? Why not check whether the target type can hold at least the same precision and scale? Maybe I'm misreading something here.

```
array.decimal_dtype().precision() <= *precision
array.decimal_dtype().scale() <= *scale
```

0ax1 (commented):
One question inline.

vortex-array/src/arrow/compute/to_arrow/canonical.rs
0ax1: As in bail if `array.decimal_dtype().precision() > *precision`.

AdamGS:
Some work here depends on the outcome of https://github.com/apache/datafusion/issues/17489, if DataFusion's support for smaller decimals doesn't improve, I'll have to rethink some stuff here.

vortex-array/src/arrow/compute/to_arrow/canonical.rs
AdamGS: I think I agree, I'll think about it a bit more but probably include it

vortex-datafusion/src/convert/exprs.rs
```diff
@@ -122,6 +123,11 @@ impl TryFromDataFusion<DFOperator> for Operator {
 }
 
 pub(crate) fn can_be_pushed_down(expr: &PhysicalExprRef, schema: &Schema) -> bool {
+    // We currently do not support pushdown of dynamic expressions in DF
```
xudong963: I think the dynamic filter will be gradually stable in the next few versions. How about recording an issue for this? There may be potential contributors to participate

AdamGS: Already opened :) https://github.com/vortex-data/vortex/issues/4034

AdamGS: I guess I should link to the issue in the code, I've also expanded there a bit about my current understanding of the issue, which might not be correct but probably worth externalizing.

vortex-array/src/arrow/compute/to_arrow/canonical.rs
AdamGS: After spending some more time with with arrow's and datafusion's decimal-handling code, I think we shouldn't change precision and scale here. Keeping it this way guarantees predictable behavior and zero-copy, unless explicitly cast.

robert3005 (approved):

PR #4578, opened 2025-09-09
performance: Avoid intersecting ranges multiple times in scan

robert3005:

### DIFF
--- a/vortex-scan/src/lib.rs
+++ b/vortex-scan/src/lib.rs
@@ -202,8 +202,7 @@ impl<A: 'static + Send> ScanBuilder<A> {
             return Ok(vec![]);
         }
 
-        let row_range = self.row_range.clone();
-        self.prepare()?.execute(row_range)
+        self.prepare()?.execute(None)
     }
 
     /// Returns a [`Stream`](futures::Stream) with tasks spawned onto the current Tokio runtime.
@@ -217,8 +216,7 @@ impl<A: 'static + Send> ScanBuilder<A> {
     pub fn into_tokio_stream(
         self,
     ) -> VortexResult<impl futures::Stream<Item = VortexResult<A>> + Send + 'static + use<A>> {
-        let row_range = self.row_range.clone();
-        self.prepare()?.execute_tokio_stream(row_range)
+        self.prepare()?.execute_tokio_stream(None)
     }
 }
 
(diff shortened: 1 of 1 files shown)


vortex-scan/src/lib.rs
```diff
@@ -420,8 +414,6 @@ impl RepeatedScan<ArrayRef> {
         &self,
         row_range: Option<Range<u64>>,
     ) -> VortexResult<impl vortex_array::stream::ArrayStream + Send + 'static> {
-        let row_range = intersect_ranges(self.row_range.as_ref(), row_range);
```
danking, line 423: If you remove this line, `self.row_range` is completely ignored.

vortex-scan/src/lib.rs
```diff
@@ -217,8 +216,7 @@ impl<A: 'static + Send> ScanBuilder<A> {
     pub fn into_tokio_stream(
         self,
     ) -> VortexResult<impl futures::Stream<Item = VortexResult<A>> + Send + 'static + use<A>> {
-        let row_range = self.row_range.clone();
-        self.prepare()?.execute_tokio_stream(row_range)
+        self.prepare()?.execute_tokio_stream(None)
```
danking, line 219: Oops yep. This is definitely cruft from when I merged the first PR.

danking (changes requested):

vortex-scan/src/lib.rs
robert3005: ha, I went too far

robert3005: Nope, it's passed to self.execute_tokio_stream and then to self.execute

robert3005:
in particular row ranges are always intersected at https://github.com/vortex-data/vortex/pull/4578/files#diff-a2f1b2589c6399cd95128e7320423c6983fd77c7a1497074ac9a105f7c5f8a71L348

onursatici (approved):
lgtm, all these methods end up calling self.execute and the row_range is intersected there with self.row_range

vortex-scan/src/lib.rs
danking: ah, yes. Hmm. this is more subtle than I would like.

danking (approved):

PR #4579, opened 2025-09-09
Fix: simplify dict array compute functions and report correct dictarray dtype now that codes and values don't have same nullability

robert3005:


java/testfiles/Cargo.lock
```diff
@@ -210,6 +210,49 @@ dependencies = [
  "tokio",
 ]
 
+[[package]]
```
robert3005: we should figure out how to keep this file up to date

encodings/dict/src/ops.rs
```diff
@@ -33,10 +31,14 @@ impl OperationsVTable<DictVTable> for DictVTable {
         let dict_index: usize = array
             .codes()
             .scalar_at(index)
-            .as_ref()
-            .try_into()
-            .vortex_expect("code overflowed usize");
-        array.values().scalar_at(dict_index)
+            .as_primitive()
+            .as_::<usize>()
+            .vortex_expect("Index overflow usize");
```
joseph-isaacs: null here means the code is null

joseph-isaacs (approved):

PR #4580, opened 2025-09-09
Add results to vortex-bench

gatesn:
Largely courtesy of Claude


danking (approved):
lgtm; I have no strong opinions on panics vs result in this case

PR #4582, opened 2025-09-09
feat: teach Footer to (de)serialize itself as FlatBuffer

danking:
I believe this provides SpiralDB with the remaining functionality necessary to persist a written Vortex Footer into metadata at write time, deserialize it, and use it at read time.

DTypes already have a serialized form, so I didn't include it in the Footer. Moreover, I think Spiral can deduce the DType itself without serializing it.

### DIFF
--- a/vortex-file/src/footer/mod.rs
+++ b/vortex-file/src/footer/mod.rs
@@ -18,16 +18,19 @@ use std::sync::Arc;
 
 pub(crate) use file_layout::*;
 pub(crate) use file_statistics::*;
-use flatbuffers::root;
+use flatbuffers::{FlatBufferBuilder, root};
 use itertools::Itertools;
 pub(crate) use postscript::*;
 pub use segment::*;
-use vortex_array::ArrayRegistry;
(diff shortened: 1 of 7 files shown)


danking:
cc: @delta003 

I wonder if we also want a way to like serialize a `Vec<Footer>` into a columnar representation? I dunno. I can imagine there's a lot of duplicate data.

gatesn (commented):
I'm tempted to say that it's not Vortex's job to give you a way to serialize the entire footer into a single byte buffer... wdyt?

Instead, we would give you a better way to serialize each segment of the footer, and then rebuild a footer from parts.

[edit] The problem is they aren't really self-contained. The Footer flatbuffer itself contains dictionary encoding of array encodings, layout encodings, etc. The Layout flatbuffer uses those dictionaries to construct in-memory objects.

What if... we change PostscriptSegment to use a relative offset rather than an absolute offset? That way, a serialized footer could just be the last N bytes of a VortexFile. And the deserialization logic would be the same thing that happens now during the open call?

Issue #1732, opened 2024-12-19
Improve error reporting and stack tracing


AdamGS:
Current errors tend to be very noisy, with long chains of `Arrow Error:`, mix of internally and externally provided sources and messy stack traces. I ran into [this](https://greptime.com/blogs/2024-05-07-error-rust) blogpost by the greptime people, which outlines a pretty elegant solution with very nice outcomes, both performance wise and UX wise.

I think we should consider adopting a similar approach, either by using whatever they end up extracting into a separate crate or building it on our own.

the main goals as I see them are:
1. Keep the general ergonomics of `thiserror` - nice `From` implementations for external types, ability to specify internal Vortex errors.
2. Better and more relevant stack traces - should inform user about concrete problems and issues, without having 50 lines of tokio/datafusion/random functions
3. Keep the error type(s?) useful for Vortex users
4. Obviously good performance/binary size etc.

PR #4585, opened 2025-09-09
RFC: Error Handling

connortsui20:
Relevant Issue: https://github.com/vortex-data/vortex/issues/1732

[Rendered Markdown](https://github.com/vortex-data/vortex/blob/ct/error-rfc/rfcs/rfc-snafu-error-handling.md)

Relevant blog post by GreptimeDB: [blog](https://greptime.com/blogs/2024-05-07-error-rust)

I don't want to be super critical since I've only worked on Vortex for a few weeks now, but having interacted with a solid percentage of the codebase (under 50%, though that interaction was a lot more than surface-level), I strongly feel that the error-handling situation can and should improve. While the current `VortexError` system seems to allow for good usage (strongly typed), the overwhelming majority of the time it is used in the codebase is by the strawman approach of formatted error strings.

I think it is clear that it can be improved. I guess the real question is if this is worth investing time into. I'm interested to hear other people's opinions.

Note that the second half is mostly claude-generated (claude helped me generate a first draft and then I rewrote the whole first half). If the first section passes the vibe check and it is something that we feel is worth (at the very least) exploring, then I'll go through that second section in more detail (which is really just figuring out an implementation/migration plan)

### DIFF
--- a/rfcs/001-snafu-error-handling.md
+++ b/rfcs/001-snafu-error-handling.md
@@ -180,6 +180,10 @@ GreptimeDB successfully implemented a hierarchical error handling system using s
 
 Their approach of creating per-module error types that form a tree structure directly inspired this RFC. They also demonstrated that migration can be incremental—existing error types can be annotated with `#[stack_trace_debug]` to add location tracking without requiring a full rewrite, allowing teams to adopt the pattern gradually. GreptimeDB's experience demonstrates that this pattern scales well for large Rust projects with complex error propagation requirements.
 
+### `iroh`
+
+They wrote a blog about using `snafu` as well: [blog](https://www.iroh.computer/blog/error-handling-in-iroh). They argue that an even more fine-grained approach is necessary for complex code, and instead of using module-level errors they have _function_-level errors.
+
 ### Other Projects Using Snafu
 
 - **[Vector](https://github.com/vectordotdev/vector)**: A high-performance observability data pipeline that uses snafu throughout their codebase to maintain error context across transformations. Their [error handling patterns](https://github.com/vectordotdev/vector/search?q=snafu) show extensive use of contextual errors across dozens of modules.
--- a/rfcs/001-snafu-error-handling.md
+++ b/rfcs/001-snafu-error-handling.md
@@ -1,5 +1,7 @@
 # RFC: Hierarchical Error Handling with Snafu
 
+Status: Proposed
+
 ## Summary
 
 This RFC proposes migrating Vortex's error handling from a centralized `VortexError` enum to a hierarchical error system using the [`snafu`] library. The migration would proceed incrementally, with each crate defining its own error types that convert into the existing `VortexError`, allowing for a gradual transition without breaking changes. Once the migration is done, we can remove the current `VortexError` and replace it with a top-level, user-facing `VortexError`.
@@ -194,17 +196,11 @@ We could keep the centralized approach but enhance `VortexError` with better con
 
 ### anyhow/color-eyre
 
-These libraries provide excellent error context but are better suited for applications than libraries. They use dynamic dispatch which makes it impossible to match on specific error types, a capability we need to preserve.
+These libraries provide excellent error context, but are better suited for applications than libraries. This would really just be replacing the current system with the same issues.
 
-### thiserror Without Hierarchy
+### `thiserror` Without Hierarchy
 
 We could use thiserror to reduce boilerplate while keeping a flat error structure. This would provide some ergonomic improvements but wouldn't address the core architectural issues around error ownership and context.
-
-## Performance Considerations
-
-The location capture in snafu is designed to be very lightweight, recording only file, line, and column information without symbol resolution. The error path is typically cold code, so the performance impact should be minimal for the success path. However, workloads that generate many errors (such as parsing invalid data) could see performance impacts that would need benchmarking.
-
-The virtual stack traces would actually be more efficient than the current backtrace capture, which performs symbol resolution at runtime. The context propagation adds only pointer-sized overhead at each level.
 
 ## Testing Strategy
 
--- a/rfcs/rfc-snafu-error-handling.md
+++ b/rfcs/rfc-snafu-error-handling.md
@@ -1,0 +1,265 @@
+# RFC: Hierarchical Error Handling with Snafu
+
+## Summary
+
+This RFC proposes migrating Vortex's error handling from a centralized `VortexError` enum to a hierarchical error system using the [`snafu`] library. The migration would proceed incrementally, with each crate defining its own error types that convert into the existing `VortexError`, allowing for a gradual transition without breaking changes. Once the migration is done, we can remove the current `VortexError` and replace it with a top-level, user-facing `VortexError`.
+
+[`snafu`]: https://docs.rs/snafu/latest/snafu/
+
+## Motivation
+
+Vortex currently uses a single `VortexError` enum shared across the entire codebase. While this approach is relatively consistent and easy, it has several limitations that has clearly become increasingly problematic.
+
+### Current Pain Points
+
+When debugging failures, we will likely come across generic error variants like `WithContext` that really only provide limited context about what has actually happened. The error types don't encode which operations or modules can produce which errors, making it difficult to handle specific failure cases programmatically. Of course, it is very possible for us to manually inspect the backtrace, but in my opinion this is not ideal, and we can definitely do better.
+
+Some other (small) things: The 128-byte size limit on `VortexError` constrains how much context can be stored, and the manual backtrace capture adds runtime overhead even when not needed (and I would argue that most of the time we don't _really_ need it, see the [#GreptimeDB] section and the [blog post](https://greptime.com/blogs/2024-05-07-error-rust)).
+
+Basically, the current system makes it difficult to understand error propagation paths through the _entire_ codebase, as any function returning `VortexResult` could theoretically produce any variant of `VortexError`.
+
+I found that the the current `VortexError` also causes a lot of confusion on the developer side as to what error variant to return. I didn't even know that there were so many variants until recently, since the entirety of the code that I have interacted with over the past 3 weeks simply uses `vortex_bail` or `vortex_err!`, which returns a `WithContext` variant with an error string. This is the **exact** definition of a strawman approach to error handling, and it completely disregards the benefits of strong type consistency that Rust provides us.
+
+### Benefits of Hierarchical Errors
+
+A hierarchical error system would establish clear ownership boundaries, with each module maintaining its own error types that accurately reflect its failure modes. This approach provides several key advantages, and is basically taken from GreptimeDB and their [blog post](https://greptime.com/blogs/2024-05-07-error-rust).
+
+Error context would be automatically captured at each level of the call stack, creating a "virtual stack trace" that shows the logical flow of operations rather than just the final failure point. Module-specific error types would make it immediately clear which operations can fail in which ways, improving both API documentation and code comprehension. As a small benefit, the workspace dependency tree would be inverted with respect to errors, with the lower-level crates no longer depending on the `vortex-error` crate.
+
+The snafu library provides ergonomic macros for error creation and context addition, reducing boilerplate while maintaining type safety. I would say it is a mix between `thiserror` and `anyhow`, but my personal experience using `snafu` has been quite different than just smashing those two libraries together.
+
+## Detailed Design
+
+### Error Hierarchy Structure
+
+Each crate would define its own error module with types specific to its domain. For example, `vortex-buffer` would define errors related to memory alignment and bounds checking, while `vortex-array` would have errors for type mismatches and compute operations. These error types would form a natural hierarchy following the module structure of each crate. This is the philosophy that `snafu` tries to cater to.
+
+```rust
+// vortex-buffer/src/error.rs
+use snafu::{Snafu, Location, Backtrace};
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub))]
+pub enum BufferError {
+    #[snafu(display("Buffer alignment error: required {required}, found {actual}"))]
+    Alignment {
+        required: usize,
+        actual: usize,
+        location: Location,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Slice range {start}..{end} out of bounds for buffer of length {len}"))]
+    SliceOutOfBounds {
+        start: usize,
+        end: usize,
+        len: usize,
+        location: Location,
+        backtrace: Backtrace,
+    },
+}
+```
+
+### Context Propagation
+
+Snafu's context selectors would replace the current `vortex_err!` and `vortex_bail!` macros. When an error crosses module boundaries, context would be added to track the operation that failed:
+
+```rust
+// vortex-array/src/compute/cast.rs
+use snafu::{ResultExt, Snafu, Location, Backtrace};
+
+#[derive(Debug, Snafu)]
+enum CastError {
+    #[snafu(display("Cannot cast from {from} to {to}"))]
+    IncompatibleTypes {
+        from: DType,
+        to: DType,
+        location: Location,
+        backtrace: Backtrace,
+    },
+
+    #[snafu(display("Buffer operation failed during cast"))]
+    BufferOperation {
+        #[snafu(source)]
+        source: vortex_buffer::BufferError,
+        location: Location,
+        backtrace: Backtrace,
+    },
+}
+
+fn cast_array(array: &Array, target: DType) -> Result<Array, CastError> {
+    let buffer = array.buffer()
+        .context(BufferOperationSnafu)?;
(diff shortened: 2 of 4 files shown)


rfcs/rfc-snafu-error-handling.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Alternatives Considered
+
+### Enhanced Central Error Type
+
+We could keep the centralized approach but enhance `VortexError` with better context tracking. This would avoid the migration complexity but wouldn't address the fundamental coupling issues or provide module-specific error types.
+
+### anyhow/color-eyre
+
+These libraries provide excellent error context but are better suited for applications than libraries. They use dynamic dispatch which makes it impossible to match on specific error types, a capability we need to preserve.
```
gatesn: "makes it impossible to match on specific error types, a capability we need to preserve."

Is this true? I don't think we ever match on error types now, and I would argue we probably shouldn't in almost all cases?

rfcs/rfc-snafu-error-handling.md
```diff
@@ -0,0 +1,265 @@
...
+
+### thiserror Without Hierarchy
+
+We could use thiserror to reduce boilerplate while keeping a flat error structure. This would provide some ergonomic improvements but wouldn't address the core architectural issues around error ownership and context.
+
+## Performance Considerations
+
+The location capture in snafu is designed to be very lightweight, recording only file, line, and column information without symbol resolution. The error path is typically cold code, so the performance impact should be minimal for the success path. However, workloads that generate many errors (such as parsing invalid data) could see performance impacts that would need benchmarking.
+
+The virtual stack traces would actually be more efficient than the current backtrace capture, which performs symbol resolution at runtime. The context propagation adds only pointer-sized overhead at each level.
```
gatesn: The current backtrace capture is indeed slow, but happens only on error, which is rare.

rfcs/rfc-snafu-error-handling.md
gatesn, line 1: Can you rename `001-snafu-error-handling.md` and add `Status: Proposed` below the title

rfcs/rfc-snafu-error-handling.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Summary
+
+This RFC proposes migrating Vortex's error handling from a centralized `VortexError` enum to a hierarchical error system using the [`snafu`] library. The migration would proceed incrementally, with each crate defining its own error types that convert into the existing `VortexError`, allowing for a gradual transition without breaking changes. Once the migration is done, we can remove the current `VortexError` and replace it with a top-level, user-facing `VortexError`.
+
+[`snafu`]: https://docs.rs/snafu/latest/snafu/
+
+## Motivation
+
+Vortex currently uses a single `VortexError` enum shared across the entire codebase. While this approach is relatively consistent and easy, it has several limitations that has clearly become increasingly problematic.
```
gatesn: I would strike " that has clearly become increasingly problematic.", it's a bit too subjective perhaps?

rfcs/rfc-snafu-error-handling.md
```diff
@@ -0,0 +1,265 @@
...
+
+### Current Pain Points
+
+When debugging failures, we will likely come across generic error variants like `WithContext` that really only provide limited context about what has actually happened. The error types don't encode which operations or modules can produce which errors, making it difficult to handle specific failure cases programmatically. Of course, it is very possible for us to manually inspect the backtrace, but in my opinion this is not ideal, and we can definitely do better.
```
gatesn: "difficult to handle specific failure cases programmatically" - isn't using errors for control flow famously a bad thing?

I guess we use VortexError sort of like an "exception". Where we _do_ care about errors, and specifically that a caller should match on them, we should indeed use error enums.

rfcs/rfc-snafu-error-handling.md
```diff
@@ -0,0 +1,265 @@
...
+
+Some other (small) things: The 128-byte size limit on `VortexError` constrains how much context can be stored, and the manual backtrace capture adds runtime overhead even when not needed (and I would argue that most of the time we don't _really_ need it, see the [#GreptimeDB] section and the [blog post](https://greptime.com/blogs/2024-05-07-error-rust)).
+
+Basically, the current system makes it difficult to understand error propagation paths through the _entire_ codebase, as any function returning `VortexResult` could theoretically produce any variant of `VortexError`.
```
gatesn: Suppose we have a compute function, can't that basically return any error that the entire codebase can return? Buffers, arrays, even custom user errors that they wrap up in their expressions?

gatesn (commented):
I'm not against, but I'm also not yet convinced? 

As an intermediate step, I would propose we remove most of the error variants from VortexError that don't just wrap up external errors (which is just for ergonomics), and then settle on using vortex_err / vortex_bail macros. I agree it's ugly, but it's more consistent than our current half-baked set of error variants that never get properly used.

rfcs/rfc-snafu-error-handling.md
connortsui20: I just woke up, so I'll respond to everything else once I'm more awake, but for this I should clarify because I feel strongly about it: right now I think this is a problem - a manageable one right now, but still somewhat of a problem.

As the codebase grows, and as people start to use this / develop on top of it, I strongly believe that this approach will not scale well. One could argue that it already hasn't scaled well due to the huge amount of semi-inconsistent error formatting strings littered throughout the codebase, as well as the difficulty of parsing backtraces when something goes wrong (due to large amounts of indirection for vortex-array).

rfcs/rfc-snafu-error-handling.md
AdamGS: Not sure how things will look in this new world, but in my mind that's an argument about composing the errors in a more scoped way. While crates that deal with serde should only return serde-related errors (easy to imagine an error type that has two variants - one to wrap flatbuffers error and the other for flexible vortex-specific erros), higher level crates will include that error type + specific errors it cares about (compute has many - out of bounds, wrong types)

rfcs/rfc-snafu-error-handling.md
connortsui20: So this section is AI generated, but at the same time I'm still inclined to agree because matching on error types is not something we need to care about on our side of the library boundary, but it is something that users of vortex (as a library) would want to care about.

For example, I would certainly want to match on an error when it could be a difference between trying to write data with the wrong type to a file versus writing past a boundary, versus an internal vortex error (I don't know if this actually applies to Vortex since I haven't looked at the exact API boundary, but hopefully this gets my idea across).

Regardless, I'm removing this since `anyhow` is certainly not going to improve our error handling.

rfcs/rfc-snafu-error-handling.md
connortsui20: I think that actually boils down my argument: We use `VortexError` as an exception, where Rust as a language has specifically designed error handling to **not** look like exceptions.

As for the programmatic handling of failures, I think using errors for control flow in Rust specifically is not necessarily a bad thing since a `Result` is just a normal enum and not a special language feature (like exceptions in other languages).Off the top of my head in the Rust standard library, I can think of `Poison` from the `sync` library (with mutexes and rwlock), I/O errors from all file I/O, message passing returning errors when the other side has dropped, these are all things that are crucial to control flow, and I would argue are idiomatic rather than a bad pattern

rfcs/rfc-snafu-error-handling.md
connortsui20: I think it would be worth doing an experiment where we see what module-specific errors would look like in Vortex.

To continue off of what @AdamGS said, even in a compute function, we are easily able to construct a tree of hierarchical errors that could be wrapped together in a single `ComputeFnError` (name to be bikeshed). For example, we could have an error type that essentially forwards encoding errors, while adding context based on which part of the compute function failed (wrong inputs, wrong output, an actual compute error, infinite loop error, etc).

Or put another way, hierarchical errors are **strictly more expressive** than what we have right now. It's not like we would lose functionality that returning format-string errors has.

connortsui20:
I had claude help me create a demo file of what errors might look like. It should help visualize what the error hierarchy could look like. If you ctrl F all of the `source: ` fields you'll see what we mean by composing errors in a scoped manner, and why it creates a "tree" of errors.

rfcs/001-snafu-error-handling.md
```diff
@@ -0,0 +1,265 @@
+# RFC: Hierarchical Error Handling with Snafu
+
+Status: Proposed
+
+## Summary
+
+This RFC proposes migrating Vortex's error handling from a centralized `VortexError` enum to a hierarchical error system using the [`snafu`] library. The migration would proceed incrementally, with each crate defining its own error types that convert into the existing `VortexError`, allowing for a gradual transition without breaking changes. Once the migration is done, we can remove the current `VortexError` and replace it with a top-level, user-facing `VortexError`.
```
alamb, line 7: my personal experience with Snafu (we used it in influxdb_iox) is that having module specific error types adds a lot of overhead for very little practical value

rfcs/001-snafu-error-handling.md
```diff
@@ -0,0 +1,265 @@
...
+
+[`snafu`]: https://docs.rs/snafu/latest/snafu/
+
+## Motivation
+
+Vortex currently uses a single `VortexError` enum shared across the entire codebase. While this approach is relatively consistent and easy, it has several limitations that has clearly become increasingly problematic.
+
+### Current Pain Points
+
+When debugging failures, we will likely come across generic error variants like `WithContext` that really only provide limited context about what has actually happened. The error types don't encode which operations or modules can produce which errors, making it difficult to handle specific failure cases programmatically. Of course, it is very possible for us to manually inspect the backtrace, but in my opinion this is not ideal, and we can definitely do better.
```
alamb, line 17: > The error types don't encode which operations or modules can produce which errors, making it difficult to handle specific failure cases programmatically.

While this is true in theory, it is worth considering in what cases you would actually handle the error differently.

Let's take the follow enum from this PR as example. What are the cases where you would programmatically handle the variants differently? For example, do you imagine retying on some of the variants and not others? I suspect most code would probably return a error and a string to the user for all four cases (not react differently to them programatically)

```rust
[derive(Debug, Snafu)]
#[snafu(visibility(pub))]
pub enum DTypeError {
    #[snafu(display("Cannot cast from {from} to {to}: incompatible types"))]
    IncompatibleCast {
        from: String, // Would be DType in real code
        to: String,
        location: Location,
        backtrace: Backtrace,
    },

    #[snafu(display("Failed to parse type from flatbuffer"))]
    FlatbufferParse {
        #[snafu(source)]
        source: Box<dyn std::error::Error + Send + Sync>,
        location: Location,
        backtrace: Backtrace,
    },

    #[snafu(display("Struct field '{field}' not found"))]
    FieldNotFound {
        field: String,
        location: Location,
        backtrace: Backtrace,
    },

    #[snafu(display("Type mismatch: expected {expected}, found {actual}"))]
    TypeMismatch {
        expected: String,
        actual: String,
        location: Location,
        backtrace: Backtrace,
    },
}
```

connortsui20: I agree with you that there isn't too much benefit here if we only treat these errors as things to be matched on. However, one of the issues I see right now is that these variants (namely things like incompatible casts or type mismatch or invalid type for operation) happen _so often_ and yet are always reported _completely differently_ with a different format string.

Being able to handle errors programmatically is not the reason to move away from format string errors. This is about code reuse, making code more explicit in what we are trying to represent (my personal opinion is that returning a typed `IncompatibleCast ` error is far better than having to manually write `vortex_bail!("We tried to cast something of type {:?} to type {:?}")` hundreds of times (and always slightly differently because there's no way to share this logic unless you wrap it in a function, and at that point you might as well just use a new error type).

rfcs/001-snafu-error-handling.md
connortsui20: When you say overhead, do you mean overhead in code bloat / readability? Or do you mean performance or binary size? My personal experience has been that explicit types make things _easier_ to work with across components.

Regardless... I' interested to hear how the alternative has worked for you. Does iox also treat all of the errors like exceptions? From my understanding, this works great for applications but not super well for libraries.

rfcs/001-snafu-error-handling.md
alamb: > (and always slightly differently because there's no way to share this logic unless you wrap it in a function, and at that point you might as well just use a new error type).

I personally prefer the function because it puts less cognitive load on users / future readers

Or if `IncompatibleCast` will be very common, perhaps making `Vortex::IncompatibleCast` is reasonable

I found the snafu based errors in influxdb_iox very frustrating to work on as it required defining a new *type* just to return an error message. So it basically added a tax on all new code for the theoretical benefit of more precise error tracking, though in reality we almost always simply want the message to send back to the user

rfcs/rfc-snafu-error-handling.md
alamb: > Or put another way, hierarchical errors are strictly more expressive than what we have right now. It's not like we would lose functionality that returning format-string errors has.

That is correct -- the downside is that you make adding new code significantly harder (in my opinion) as you need to define new types for all error messages.

rfcs/001-snafu-error-handling.md
connortsui20: That makes sense. Do you think then that it would be better to just use `anyhow` instead? Because that is essentially what we are doing. We could have a `VortexError` with a general `anyhow::Error` variant, and then in the odd cases where we need to do control flow on errors we could add those specifically as variants.

That would certainly be a better system than what we have now, and it also aligns with what you said. However, it also feels incredibly strange to me since the conventional wisdom is "don't use `anyhow` for libraries," as you lose a lot of control over your errors. So hopefully that explains why I feel like this is an issue: as a library we are actively choosing to go in the direction of type erasing all of our errors, which doesn't pass the smell test for me.

connortsui20: The tradeoff here seems to be cognitive load versus explicitness over what can go wrong in a function. I guess I am arguing that the upfront cognitive load is warranted.

My understanding is that influxdb_iox has been around for a while now: what has the experience of _maintaining_ it been like? I can understand that there is an upfront cost to creating all of these error types just to return a single error message (which personally I actually like since it makes it exceedingly obvious what can go wrong to someone reading the code, but I can understand that many people don't), but I would be surprised if that did not give you any benefits later on. Has it really just been painful throughout the entire process?

alamb: > The tradeoff here seems to be cognitive load versus explicitness over what can go wrong in a function. I guess I am arguing that the upfront cognitive load is warranted.

In my opinion the tradeoff is cognitive load (and extra work for new code) vs explicitness

I agree it is an opinion and that there are pros and cons. 

I have now been maintaining Arrow / Parquet / Datafusion for 5 years and the error types there work well in my experience. Here are some examples

* https://docs.rs/arrow/latest/arrow/error/enum.ArrowError.html
* https://docs.rs/datafusion/latest/datafusion/common/enum.DataFusionError.html

There is some structure that allows responding to different types of errors, but mostly they just contain strings

alamb: For DataFusion, a good example of an error type that actually needs its own enum is `ResourcesExhausted` -- as it is very common for the application to try to do something different on that (like retry the query, or run on a different machine, or run later, etc)

alamb: But note it doesn't try to carefully enumerate all possible ways the resources can be exhausted (e.g. in sorts, or hash join, etc)

connortsui20: After thinking about this a bit more, I think you are right in the idealized version: DataFusion seems to have put thought into what exactly should and shouldn't be represented in it's error type (https://github.com/apache/datafusion/issues/7360) and so it can afford to represent most of its errors as format strings.

Though arguably for us, there is somewhat of a lack of discipline in the codebase where our error messages aren't super consistent and also arguably not as detailed as they can be.

The benefit that GreptimeDB and iroh have shown is that they get explicit stack traces that make it much easier to hone in on issues. Vortex has quite a lot of indirection, which makes identifying the issue harder (not impossible, but I'm saying it could be better). Whether or not that is worth the engineering cost is up for debate.

Whatever is the best way, we definitely need to clean up error handling as it is now and refactor some variants (like what @gatesn mentioned). And maybe this problem goes away if we just take some time to think about the broad categories of errors we have (I like the clear distinctions that `DataFusionError` has made)...

Thanks for taking the time to discuss this with me! I'll admit that I'm not fully convinced that this is the best, but I'm also convinced that we have many more pressing problems so I will be leaving this for later.

alamb: I am just trying to save you some heartache based on my experience -- I don't think your eventual decision will affect me greatly either way

connortsui20:
Closing for now in favor of https://github.com/vortex-data/vortex/issues/4611, though once that is done and we've worked with better error handling for a bit, I might revisit this.

PR #4588, opened 2025-09-10
feat: fastlanes rle

0ax1:
Corresponding fastlanes repo commit implementing the kernel: https://github.com/spiraldb/fastlanes/pull/70

Note that I didn't include the `compute` mod as part of this PR to limit the scope.

Comparing the benchmark results between runend and fastlanes rle, we should probably consider to only provide a fused rle + delta kernel, given that fastlanes in the current shape is significantly slower than vortex runend.


0ax1:
@gatesn I'll check whether the asm is equivalent for the rust kernel comparing against the c++ version.

0ax1:
Note that I didn't include the `compute` mod as part of this PR to limit the scope.

encodings/fastlanes/src/rle/serde.rs
```diff
@@ -0,0 +1,183 @@
...
+            2,
+            &DType::Primitive(vortex_dtype::PType::U16, Nullability::NonNullable),
+            children.get(2, &DType::Null, 0)?.len(),
+        )?;
+
+        assert_eq!(indices.dtype().nullability(), Nullability::NonNullable);
+        assert_eq!(
+            unique_counts.dtype().nullability(),
+            Nullability::NonNullable
+        );
```
robert3005: this shouldn't panic

encodings/fastlanes/src/rle/serde.rs
```diff
@@ -0,0 +1,183 @@
...
+                children.get(1, &DType::Null, 0)?.dtype().nullability(),
+            ),
+            len,
+        )?;
+
+        let unique_counts = children.get(
+            2,
+            &DType::Primitive(vortex_dtype::PType::U16, Nullability::NonNullable),
+            children.get(2, &DType::Null, 0)?.len(),
+        )?;
```
robert3005: what is happening here? This looks very... wrong, like you inlined a variable and also dtypes are wrong

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+    ) -> VortexResult<Self> {
+        let dtype = DType::Primitive(values.dtype().as_ptype(), validity.nullability());
+
+        let indices = match indices.dtype() {
+            DType::Primitive(PType::U16, _) => indices,
+            _ => cast(
+                &indices,
+                &DType::Primitive(PType::U16, indices.dtype().nullability()),
+            )?,
+        };
```
robert3005: this shouldn't cast, the caller should do this

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+        Ok(())
+    }
+
+    /// Create a new chunk-based RLE array from its components.
+    ///
+    /// # Arguments
+    ///
+    /// * `values` - Flattened unique values from all chunks
+    /// * `indices` - Flattened chunk-local indices from all chunks
+    /// * `unique_counts` - Number of unique values per chunk
```
robert3005: These clearly aren't flattened if you take an array here. Why would they be flattened?

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    for chunk_idx in 0..array.num_chunks() {
+        let unique_count = unique_counts[chunk_idx] as usize;
+        let chunk_len = array.chunk_length(chunk_idx) as usize;
+
+        let chunk_values = &values[values_offset..values_offset + unique_count];
+        let chunk_indices = &indices[indices_offset..indices_offset + chunk_len];
+        let chunk_validity = validity_mask.slice(validity_offset..validity_offset + chunk_len);
+
+        if chunk_len == 1024 {
```
robert3005: when is this not true? The compression should chunk input into 1024 elements so you'd have at most 2 non 1024 element long chunks? Can we handle this if outside the loop?

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,282 @@
...
+            #[allow(clippy::expect_used)]
+            T::decode(
+                chunk_values,
+                chunk_indices
+                    .try_into()
+                    .expect("Indices must be convertible to &[u16; 1024]"),
+                builder_values
+                    .try_into()
+                    .expect("Builder values must be convertible to &[T; 1024]"),
+            );
```
robert3005: you are making a copy here, array_ref is the solution, checkout delta encoding.

robert3005 (changes requested):
There's some things around serde that need changing, validation seems lacking and is doing things I wouldn't expect it to do. Overall though I think I would need a walkthrough since I am struggling to understand what's happening in compression/decompression compared to the paper

encodings/fastlanes/src/rle/serde.rs
0ax1: ~~Not sure I follow. There's nothing panicking here.~~ Ignore I thought you were looking at the tests. => changed to `vortex_ensure`.

encodings/fastlanes/src/rle/serde.rs
0ax1: What do you mean by inlined a variable? And good catch I initially set Null as placeholders.

encodings/fastlanes/src/rle/mod.rs
0ax1: What I meant was: Fastlanes processes in 1024 chunks. Values and indices are local to the chunk but end up in the same array. As in, the output is not an array of arrays. If you look at the decoding, you can have a look how we step through the chunks.

encodings/fastlanes/src/rle/compress.rs
0ax1: The last chunk can be smaller than 1024.

0ax1: The input can obviously have any length which might not be len % 1024 == 0.

encodings/fastlanes/src/rle/mod.rs
0ax1: removed

encodings/fastlanes/src/rle/compress.rs
0ax1: Are you sure? 

This calls into 
```
    fn try_from(slice: &'a [T]) -> Result<&'a [T; N], TryFromSliceError> {
        slice.as_array().ok_or(TryFromSliceError(()))
    }
```

which in turn calls into 

```
    pub const fn as_array<const N: usize>(&self) -> Option<&[T; N]> {
        if self.len() == N {
            let ptr = self.as_ptr() as *const [T; N];

            // SAFETY: The underlying array of a slice can be reinterpreted as an actual array `[T; N]` if `N` is not greater than the slice's length.
            let me = unsafe { &*ptr };
            Some(me)
        } else {
            None
        }
    }
```

which casts and not copies.

encodings/fastlanes/src/rle/mod.rs
0ax1: Lemme know what you think would help here to clarify the docs.

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,268 @@
...
+
+impl RLEArray {
+    /// Encode a primitive array of unsigned integers using FastLanes RLE.
+    ///
+    /// In case a signed integer array is provided, `encode` attempts to convert
+    /// to the corresponding unsigned type.
+    pub fn encode(array: &dyn Array) -> VortexResult<Self> {
+        let parray = array
+            .as_opt::<PrimitiveVTable>()
+            .ok_or_else(|| vortex_error::vortex_err!("RLE can only encode primitive arrays"))?;
```
robert3005: this should just take PrimitiveArray

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,268 @@
...
+///
+/// Returns an `RLEArray` that contains:
+///    - `values`: Flattened values from all chunks
+///    - `indices`: Flattened chunk-local indices
+///    - `unique_counts`: Number of unique values per chunk (for reconstruction)
+fn rle_encode_typed<T>(array: &PrimitiveArray) -> VortexResult<RLEArray>
+where
+    T: NativePType + RLE + Clone + Hash + Eq,
+{
+    let values = array.as_slice::<T>();
```
robert3005, line 37: You should chunk this into 1024 elements and interate them

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,268 @@
...
+    let mut indices_offset = 0;
+    let mut validity_offset = 0;
+
+    for chunk_idx in 0..array.num_chunks() {
+        let unique_count = unique_counts[chunk_idx] as usize;
+        let chunk_len = array.chunk_length(chunk_idx) as usize;
+
+        let chunk_values = &values[values_offset..values_offset + unique_count];
+        let chunk_indices = &indices[indices_offset..indices_offset + chunk_len];
+        let chunk_validity = validity_mask.slice(validity_offset..validity_offset + chunk_len);
```
robert3005: there's `set_validity` on a builder

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,268 @@
...
+                array_mut_ref![rle_vals, 0, 1024],
+                array_mut_ref![rle_idxs, 0, 1024],
+            )
+        };
+
+        // Reflect the updated length for values and indices.
+        unsafe {
+            values_buf.set_len(values_buf.len() + unique_count);
+            indices_buf.set_len(indices_buf.len() + chunk_len);
+        }
```
robert3005: You pad indices to 1024 multiple always and then on decode you can truncate at the very end

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub fn try_new(
+        values: ArrayRef,
+        indices: ArrayRef,
+        unique_counts: ArrayRef,
+        validity: Validity,
+    ) -> VortexResult<Self> {
```
robert3005: you have to validate that the validity matches the length of the array

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+        let dtype = DType::Primitive(values.dtype().as_ptype(), validity.nullability());
```
robert3005, line 129: `as_ptype()` will panic if this is not a primitive so you should validate first

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+impl RLEArray {
+    fn validate(
+        values: &dyn Array,
+        indices: &dyn Array,
+        unique_counts: &dyn Array,
+    ) -> VortexResult<()> {
+        use vortex_dtype::Nullability::NonNullable;
+
+        vortex_ensure!(
+            matches!(values.dtype(), DType::Primitive(_, _)),
```
robert3005: `is_primitive()`

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+            "RLE values must be a primitive type, got {}",
+            values.dtype()
+        );
+
+        vortex_ensure!(
+            matches!(indices.dtype(), DType::Primitive(PType::U16, NonNullable)),
```
robert3005: you can replace matches with `==`

encodings/fastlanes/src/rle/mod.rs
robert3005: just skip `Flattened`

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,268 @@
...
+fn rle_decode_typed<T>(array: &RLEArray) -> PrimitiveArray
+where
+    T: NativePType + RLE + Clone + Copy,
+{
+    let unique_counts = array
+        .unique_counts()
+        .as_::<PrimitiveVTable>()
+        .as_slice::<u16>();
+    let values = array.values().as_::<PrimitiveVTable>().as_slice::<T>();
+    let indices = array.indices().as_::<PrimitiveVTable>().as_slice::<u16>();
```
robert3005: `.as_::<PrimitiveVTable>()` -> `to_primitive()`

encodings/fastlanes/src/rle/compress.rs
0ax1: Changed to array ref.

encodings/fastlanes/src/rle/mod.rs
0ax1: Added a check to compare the validity length against the array length.

encodings/fastlanes/src/rle/mod.rs
0ax1: swapped the order

encodings/fastlanes/src/rle/mod.rs
0ax1: done

encodings/fastlanes/src/rle/mod.rs
0ax1: done

encodings/fastlanes/src/rle/compress.rs
0ax1: done

encodings/fastlanes/src/rle/compress.rs
0ax1: done

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,269 @@
...
+use vortex_buffer::BufferMut;
+use vortex_dtype::{NativePType, match_each_unsigned_integer_ptype};
+use vortex_error::VortexResult;
+
+use crate::{FL_CHUNK_SIZE, RLEArray};
+
+impl RLEArray {
+    /// Encodes a primitive array of unsigned integers using FastLanes RLE.
+    pub fn encode(array: &PrimitiveArray) -> VortexResult<Self> {
+        vortex_dtype::match_each_unsigned_integer_ptype!(array.ptype(), |T| {
```
robert3005: import

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,269 @@
...
+    for chunk_start_idx in (0..len).step_by(FL_CHUNK_SIZE) {
+        let chunk_end = std::cmp::min(chunk_start_idx + FL_CHUNK_SIZE, len);
+        let chunk_len = chunk_end - chunk_start_idx;
+        let chunk_slice = &values[chunk_start_idx..chunk_end];
+
+        // SAFETY:
+        // `MaybeUninit<T>` and `T` have the same layout.
+        let rle_vals: &mut [T] = unsafe {
+            std::mem::transmute(
+                &mut values_uninit[value_count_acc..value_count_acc + FL_CHUNK_SIZE],
```
robert3005: I think these are nicer expressed as 
```suggestion
                &mut values_uninit[value_count_acc..][..FL_CHUNK_SIZE],
```

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+pub struct RLEEncoding;
+
+impl RLEArray {
+    fn validate(
+        values: &dyn Array,
+        indices: &dyn Array,
+        value_chunks_offsets: &dyn Array,
+        validity: Validity,
+    ) -> VortexResult<()> {
+        use vortex_dtype::Nullability::NonNullable;
```
robert3005: move this up

encodings/fastlanes/src/rle/serde.rs
```diff
@@ -0,0 +1,212 @@
...
+use crate::{RLEArray, RLEVTable};
+
+#[derive(Clone, prost::Message)]
+pub struct RLEMetadata {
+    #[prost(uint64, tag = "1")]
+    pub values_len: u64,
+    #[prost(uint64, tag = "2")]
+    pub indices_len: u64,
+    #[prost(uint64, tag = "3")]
+    pub value_chunk_offsets_len: u64,
```
robert3005, line 24: aren't these two the same?

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    // Pre-allocate for one offset per chunk.
+    let mut value_chunk_offsets = BufferMut::<u64>::with_capacity(len.div_ceil(FL_CHUNK_SIZE));
+
+    let values_uninit = values_buf.spare_capacity_mut();
+    let indices_uninit = indices_buf.spare_capacity_mut();
+    let mut value_count_acc = 0; // Chunk value count prefix sum.
+
+    // Note that the loop iterates the last chunk even if smaller than 1024.
+    for chunk_start_idx in (0..len).step_by(FL_CHUNK_SIZE) {
```
robert3005: Style note: Would this be nicer using `chunks_exact`? Then you avoid the if conditions and just have 1024 chunks and remainder

robert3005 (commented):
Logic looks good, I have some style comments

encodings/fastlanes/src/rle/serde.rs
0ax1: nope

encodings/fastlanes/src/rle/compress.rs
0ax1: done

encodings/fastlanes/src/rle/mod.rs
0ax1: done

encodings/fastlanes/src/rle/compress.rs
0ax1: Yeah, that's better. Changed.

encodings/fastlanes/src/rle/compress.rs
0ax1: done

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -58,11 +56,8 @@ where
 
         // SAFETY:
         // `MaybeUninit<T>` and `T` have the same layout.
-        let rle_vals: &mut [T] = unsafe {
-            std::mem::transmute(
-                &mut values_uninit[value_count_acc..value_count_acc + FL_CHUNK_SIZE],
-            )
-        };
+        let rle_vals: &mut [T] =
```
robert3005, line 55: You have the same thing below

0ax1: done

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -0,0 +1,259 @@
...
+    unsafe {
+        range.finish();
+    }
+
+    let offset_within_chunk = array.offset_in_chunk(array.offset);
+
+    builder.set_validity(array.validity_mask());
+    builder
+        .finish_into_primitive()
+        .slice(offset_within_chunk..(offset_within_chunk + array.len()))
```
robert3005, line 152: this should be 
```suggestion
        .slice(offset_within_chunk..array.len())
```

robert3005: More importantly this seems like a bug? I am surprised this hasn't been hit

robert3005: No, I'm wrong, offset elements are not valid elements

encodings/fastlanes/src/rle/ops.rs
```diff
@@ -0,0 +1,257 @@
...
+        assert_eq!(array.scalar_at(5), 30u32.into());
+        assert_eq!(array.scalar_at(6), 10u32.into());
+
+        assert!(array.scalar_at(1).is_null());
+        assert!(array.scalar_at(4).is_null());
+    }
+
+    #[test]
+    fn test_scalar_at_multiple_chunks() {
+        use vortex_array::{IntoArray, ToCanonical};
```
robert3005: Can we consolidate all the use statements to the top of the module instead of functions? No IDE adds it like this, I'm confused why I am seeing this so much

encodings/fastlanes/src/rle/ops.rs
```diff
@@ -0,0 +1,257 @@
...
+#[cfg(test)]
+mod tests {
+    use vortex_array::arrays::PrimitiveArray;
+    use vortex_array::validity::Validity;
+    use vortex_array::{Array, IntoArray};
+    use vortex_buffer::Buffer;
+
+    use super::*;
+
+    mod fixture {
```
robert3005, line 70: why is this a module?

encodings/fastlanes/src/rle/ops.rs
```diff
@@ -0,0 +1,257 @@
...
+        let array = fixture::rle_array();
+        let sliced = array.slice(1..4);
+
+        assert_eq!(array.dtype(), sliced.dtype());
+    }
+
+    #[test]
+    fn test_slice_across_chunk_boundaries() {
+        use vortex_array::{IntoArray, ToCanonical};
+        use vortex_buffer::Buffer;
```
robert3005: this as well

encodings/fastlanes/src/rle/serde.rs
```diff
@@ -0,0 +1,212 @@
...
+            ProstMetadata(RLEMetadata {
+                values_len: u64::MAX,
+                indices_len: u64::MAX,
+                value_chunk_offsets_len: u64::MAX,
+            }),
+        );
+    }
+
+    #[test]
+    fn test_serde_build() {
```
robert3005: not sure how useful this test is, since you ignore the length and dtype passed into the get method so you're not actually testing your logic but instead test your mocks

encodings/fastlanes/src/rle/ops.rs
0ax1: In general, I prefer to scope imports as narrow as possible which probably comes from my habit of doing the same in C++. I think scoping imports makes it easy to see where the imports are actually used. Plus for the context of the imported symbol the module path is visible as the import is close by. Has nothing to do with Zed. But happy to go here with whatever we do in Vortex.

encodings/fastlanes/src/rle/ops.rs
0ax1: It makes clear that all functions in this module are fixtures. `fixture::rle_array` is nice to read I think. Rather than `rle_array_fixture` or just `rle_array`.

encodings/fastlanes/src/rle/ops.rs
0ax1: moved

encodings/fastlanes/src/rle/ops.rs
0ax1: done

encodings/fastlanes/src/rle/serde.rs
0ax1: dropped

encodings/fastlanes/src/rle/ops.rs
robert3005: Generally my take is that we should keep imports at module level unless there’s conflicts or compile flags that we need to abide by

robert3005 (approved):

lwwmanning:
this is great! as a follow-up, would be good to use the consistency test harness, which covers a bunch of the common fuzzer failures with unit tests

e.g., https://github.com/vortex-data/vortex/blob/develop/encodings/fastlanes/src/bitpacking/compute/mod.rs#L62-L95

PR #4590, opened 2025-09-10
Feature: Add DuckDB conversion for fixed-size lists / ARRAY type

connortsui20:
Tracking issue: https://github.com/vortex-data/vortex/issues/4372

Relevant DuckDB docs for their ARRAY type: https://duckdb.org/docs/stable/sql/data_types/array.html

Adds conversion methods and an exporter for fixed-size list type to DuckDB.

Also adds `vortex-ffi` bindings for fixed-size lists.

### DIFF
--- a/vortex-duckdb/src/exporter/fixed_size_list.rs
+++ b/vortex-duckdb/src/exporter/fixed_size_list.rs
@@ -1,0 +1,323 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Export support for Vortex FixedSizeListArray to DuckDB ARRAY type.
+//!
+//! DuckDB distinguishes between LIST (variable-size) and ARRAY (fixed-size) types.
+//! The ARRAY type in DuckDB corresponds to Vortex's [`DType::FixedSizeList`], where all
+//! lists have the same number of elements.
+//!
+//! [`DType::FixedSizeList`]: vortex_dtype::DType::FixedSizeList
+
+use vortex::arrays::FixedSizeListArray;
+use vortex::error::VortexResult;
+use vortex::mask::Mask;
+
+use super::{ConversionCache, new_array_exporter};
+use crate::duckdb::Vector;
+use crate::exporter::{ColumnExporter, VectorExt};
+
+/// Exporter for converting Vortex [`FixedSizeListArray`] to DuckDB ARRAY vectors.
+struct FixedSizeListExporter {
+    /// Validity mask indicating which lists/arrays are null.
+    validity: Mask,
+    /// Exporter for the underlying elements array.
+    elements_exporter: Box<dyn ColumnExporter>,
+    /// The fixed number of elements in each list.
+    list_size: u32,
+}
+
+/// Creates a new exporter for converting a [`FixedSizeListArray`] to DuckDB ARRAY format.
+pub(crate) fn new_exporter(
+    array: &FixedSizeListArray,
+    cache: &ConversionCache,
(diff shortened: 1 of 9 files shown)


vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -0,0 +1,458 @@
...
+            buffer![1i32, 2, 3, 4, 5, 6].into_array(),
+            2, // list_size
+            Validity::AllValid,
+            3, // len (3 lists)
+        );
+
+        let array_type = LogicalType::new_array(cpp::DUCKDB_TYPE::DUCKDB_TYPE_INTEGER, 2);
+        let mut chunk = DataChunk::new([array_type]);
+
+        new_exporter(&fsl, &ConversionCache::new(0))
```
0ax1: Not introduced as part of this PR but the pattern of passing a mutable vector and then setting separately seems brittle. Maybe there should be something like `export_to_chunk` which bundles this.

vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -0,0 +1,458 @@
...
+        let outer_fsl = FixedSizeListArray::new(
+            inner_fsl.into_array(),
+            3, // outer list_size (3 inner lists per outer list)
+            Validity::AllValid,
+            2, // 2 outer lists
+        );
+
+        // Create the nested array type for DuckDB.
+        // First create the inner array type, then use it for the outer.
+        let inner_array_type = LogicalType::new_array(cpp::DUCKDB_TYPE::DUCKDB_TYPE_INTEGER, 2);
```
0ax1: I think these unit tests are pushing it in terms of complexity. They are valuable though.

vortex-ffi/src/dtype.rs
```diff
@@ -198,6 +215,30 @@ pub unsafe extern "C-unwind" fn vx_dtype_list_element(dtype: *const vx_dtype) ->
     vx_dtype::new_ref(element_dtype)
 }
 
+/// Return the `element` type of a fixed-size list data type.
```
0ax1: How about going with `/// Return the element type of a fixed-size list.`?

vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -0,0 +1,458 @@
...
+        // Verify that offset + len doesn't exceed the validity mask length.
+        assert!(
+            offset + len <= self.validity.len(),
+            "Export range [{}, {}) exceeds validity mask length {}",
+            offset,
+            offset + len,
+            self.validity.len()
+        );
+
+        // TODO(connor): Should `export` be `unsafe` instead? We have no way to verify this without
```
0ax1: I agree that there's an argument that be made for marking it as unsafe, as it crashes / is UB in case the user passes in invalid args which are then forwarded to other unsafe functions called within `set_validity`.

Can reconsider this API for all exporters. (cc: @gatesn)

0ax1 (commented):
Couple of thoughts inline. Looks good!

0ax1:
@claude review

0ax1 (approved):

vortex-ffi/src/dtype.rs
connortsui20: I was mostly following everything around this, but that's better so I'll fix the others as well.

vortex-duckdb/src/exporter/fixed_size_list.rs
connortsui20: I've left a TODO in here and I'll look into this later*

PR #4591, opened 2025-09-10
Chore: Specialize `append_scalar` and remove `append_option`

### LINKED ISSUE #4487: Implement `append_scalar` for each array builder
Somewhat relevant: https://spiraldb.slack.com/archives/C07BV3GKAJ2/p1756822454415319

Right now, `append_scalar` has a default implementation that basically casts up and down to append the correct thing. Each array builder should probably implement this on their own instead of relying on the default implementation.

There is also a question of what values each builder should be allowed to accept. Right now we have this:

```rust
// In bool.rs

    /// Appends a boolean value to the builder.
    pub fn append_value(&mut self, value: bool) {}

    /// Appends an optional boolean value to the builder.
    pub fn append_option(&mut self, value: Option<bool>) {}


// In list.rs

    /// Appends a list `value` to the builder.
    pub fn append_value(&mut self, value: ListScalar) -> VortexResult<()> {}

    pub fn append_option(&mut self, value: Option<ListScalar>) -> VortexResult<()> {}
```

Should we make the `BoolBuilder` accept a `BoolScalar`? The issue here is that `Option<ListScalar>` is redundant because we store the `null` within the `ListScalar` via optional elements.

connortsui20:
Closes https://github.com/vortex-data/vortex/issues/4487

Specializes the `append_scalar` method for all builders.

Additionally, I've removed the `append_option` methods that I added before. Unless we pull nullability outside of scalars, I don't really see a good reason to have `append_option`.

### DIFF
--- a/vortex-array/src/builders/primitive.rs
+++ b/vortex-array/src/builders/primitive.rs
@@ -560,4 +560,47 @@ mod tests {
         assert_eq!(array.len(), 3);
         assert_eq!(array.as_slice::<i32>(), &[10, 20, 30]);
     }
+
+    #[test]
+    fn test_append_scalar() {
+        use vortex_dtype::DType;
+        use vortex_scalar::Scalar;
+
+        let mut builder = PrimitiveBuilder::<i32>::with_capacity(Nullability::Nullable, 10);
+
+        // Test appending a valid primitive value.
+        let scalar1 = Scalar::primitive(42i32, Nullability::Nullable);
+        builder.append_scalar(&scalar1).unwrap();
+
+        // Test appending another value.
+        let scalar2 = Scalar::primitive(84i32, Nullability::Nullable);
+        builder.append_scalar(&scalar2).unwrap();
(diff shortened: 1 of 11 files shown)


vortex-array/src/builders/primitive.rs
```diff
@@ -150,6 +137,24 @@ impl<T: NativePType> ArrayBuilder for PrimitiveBuilder<T> {
...
     }
 
+    fn append_scalar(&mut self, scalar: &Scalar) -> VortexResult<()> {
+        if scalar.dtype() != self.dtype() {
+            vortex_bail!(
+                "PrimitiveBuilder expected scalar with dtype {:?}, got {:?}",
+                self.dtype(),
+                scalar.dtype()
+            );
+        }
```
a10y: style nit: can use the `vortex_ensure!` macro to compact these checks a bit

a10y (approved):

PR #4592, opened 2025-09-10
Chore: Add missing `FixedSizeListEncoding`

connortsui20:

Nearly missed these...

Also I'm not entirely sure that `Action::SearchSorted` is supported for `FixedSizeList` in the fuzzer, but I also don't see if `List` is supported either so it's probably fine?

### DIFF
--- a/fuzz/fuzz_targets/array_ops.rs
+++ b/fuzz/fuzz_targets/array_ops.rs
@@ -7,10 +7,7 @@
 use std::backtrace::Backtrace;
 
 use libfuzzer_sys::{Corpus, fuzz_target};
-use vortex_array::arrays::{
-    BoolEncoding, ConstantArray, FixedSizeListEncoding, ListEncoding, PrimitiveEncoding,
-    StructEncoding, VarBinEncoding, VarBinViewEncoding,
-};
+use vortex_array::arrays::ConstantArray;
 use vortex_array::compute::{cast, compare, filter, take};
 use vortex_array::search_sorted::{SearchResult, SearchSorted, SearchSortedSide};
 use vortex_array::{Array, ArrayRef, IntoArray};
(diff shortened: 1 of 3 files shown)


fuzz/fuzz_targets/array_ops.rs
```diff
@@ -46,14 +45,15 @@ fuzz_target!(|fuzz_action: FuzzArrayAction| -> Corpus {
...
-                if !HashSet::from([
+                if [
                     PrimitiveEncoding.id(),
                     VarBinEncoding.id(),
                     VarBinViewEncoding.id(),
                     BoolEncoding.id(),
                     StructEncoding.id(),
                     ListEncoding.id(),
-                ])
+                    FixedSizeListEncoding.id(),
```
connortsui20: Is this correct? I don't see a "search sorted" compute function for `ListEncoding` but this seems to have been fine?

robert3005: you inverted condition but this really should just be if `!current_array.is_canonical()`. This is mostly to recover the original encoding after sorting

connortsui20: oh whoops

connortsui20: Could you see if the comment I added (in the new commit) makes sense?

robert3005: yes this is correct but the proper fix here is to change the condition here to what I mentioned before so we don't have to fix it again

connortsui20: 🤦 I misread your comment

connortsui20:
I will fix this tomorrow morning when I get back to my laptop (or someone else can push the quick fix)

robert3005 (approved):

connortsui20:
Thanks @robert3005!

PR #4596, opened 2025-09-10
DictionaryEncoder to store nulls in values

gatesn:
Two reasons:
1. We can do much better push-down if our codes are non-null (all scalar operations)
2. We can know up-front the DType of the codes array coming out of DictEncoder. This is important for DictLayout where we currently get lucky that all chunks of a dict column have the same nullability :(

### DIFF
--- a/encodings/dict/src/compute/compare.rs
+++ b/encodings/dict/src/compute/compare.rs
@@ -2,8 +2,8 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use vortex_array::arrays::ConstantArray;
-use vortex_array::compute::{compare, CompareKernel, CompareKernelAdapter, Operator};
-use vortex_array::{register_kernel, Array, ArrayRef, IntoArray};
+use vortex_array::compute::{CompareKernel, CompareKernelAdapter, Operator, compare};
+use vortex_array::{Array, ArrayRef, IntoArray, register_kernel};
 use vortex_error::VortexResult;
 
 use crate::{DictArray, DictVTable};
@@ -48,7 +48,7 @@ register_kernel!(CompareKernelAdapter(DictVTable).lift());
 #[cfg(test)]
 mod tests {
     use vortex_array::arrays::{ConstantArray, PrimitiveArray};
-    use vortex_array::compute::{compare, Operator};
+    use vortex_array::compute::{Operator, compare};
     use vortex_array::validity::Validity;
     use vortex_array::{IntoArray, ToCanonical};
     use vortex_buffer::buffer;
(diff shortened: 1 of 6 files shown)


gatesn:
This is another reason to move to the operator world. These benchmark regressions are meaningless because we don't canonicalize. The actual TPCH / Clickbench queries are largely unchanged suggesting we just moved the logic inside the comparison function.

encodings/dict/src/compute/compare.rs
```diff
@@ -24,25 +19,22 @@ impl CompareKernel for DictVTable {
...
         }
+
         // If the RHS is constant, then we just need to compare against our encoded values.
         if let Some(rhs) = rhs.as_constant() {
             let compare_result = compare(
                 lhs.values(),
                 ConstantArray::new(rhs, lhs.values().len()).as_ref(),
                 operator,
             )?;
-            return if operator == Operator::Eq {
```
robert3005, line 34: I don't think you can delete this, it has nothing to do with nullability but instead with whether we enforce that dictionaries are unique (which we don't)

encodings/dict/src/compute/mod.rs
```diff
@@ -70,10 +70,10 @@ mod test {
 
         let expected: Vec<i32> = (0..65)
             .map(|i| match i % 3 {
-                // Compressor puts 0 as a code for invalid values which we end up using in take
-                // thus invalid values on decompression turn into whatever is at 0th position in dictionary
-                0 | 2 => 42,
+                // Compressor puts 0 as a code for invalid values
```
robert3005, line 73: the invalid value is going to be the position of first null in the array

encodings/dict/src/compute/compare.rs
gatesn: This branch should only ever be an optimization, no? If the RHS is constant, then we can compute against values and slap original codes on top.

If this is for correctness... then we're probably missing some tests and Operator::Neq should probably also get the same treatment.

encodings/dict/src/compute/mod.rs
gatesn: I'm saying that where a value is null, the decompressed array (ignoring the null buffer) will have zero in its place

encodings/dict/src/compute/compare.rs
robert3005: This is an optimisation so I was surprised this was touched at all, the logic should stay the same since it handled nullability in both codes and value. It’s the same logic that would have to be implemented in the operator world

gatesn: Turns out it did it badly... So figured I'd just try without, since the Dict::canonicalize is pretty well optimized anyway.

gatesn: In the operator world it's much easier to benchmark because everyone outputs canonical data. So no cheating.

robert3005 (approved):
I filed an issue for the compare change. The only thing I was curious about was whether the null should be in hashtable or not and whether this matters for encoding perf

PR #4598, opened 2025-09-11
Support independent footer serde

### LINKED ISSUE #4064: Actually be able to pass a Footer when opening a file

gatesn:
This PR provides public APIs for serializing and deserializing a Vortex file footer. This can be used to store footers out-of-band to avoid file I/O when opening a Vortex file.

Fixes #4064


vortex-file/src/footer/deserializer.rs
```diff
@@ -0,0 +1,274 @@
...
+    // The DType, if provided externally.
+    dtype: Option<DType>,
+
+    // Internal state that we accumulate
+
+    // The file size, possibly provided externally.
+    file_size: Option<u64>,
+    // The postscript, once we've parsed it.
+    postscript: Option<Postscript>,
+}
```
AdamGS, line 38: Should be rustdoc

AdamGS (approved):
LGTM, I guess we don't really have a tests but I can do as a followup

PR #4600, opened 2025-09-11
Add back segment paths to the inspect command

gatesn:


vortex-tui/src/inspect.rs
```diff
@@ -197,45 +205,11 @@ impl VortexInspector {
...
-    fn read_footer_segments(
-        &mut self,
-        postscript: &PostscriptInfo,
-    ) -> VortexResult<FooterSegments> {
-        // Read footer segment
-
-        let mut footer_bytes = vec![0u8; postscript.footer.length as usize];
-        self.file
-            .seek(SeekFrom::Start(postscript.footer.offset))
-            .map_err(|e| vortex_err!("Failed to seek to footer: {}", e))?;
```
AdamGS, line 209: The original intent was to be able to parse individual parts, so we could peek into partially corrupted files

AdamGS (approved):

PR #4601, opened 2025-09-11
Chore: Add no-op `IsSorted` and `MinMax` compute kernels for `FixedSizeListVTable`

connortsui20:

This follows the list implementation that is also a no-op (is no-op the right word here?)


robert3005:
I don't think you need these, this is a canonical encoding so there's nothing coming from these returning nothing

robert3005:
They should also not exist for List (or actually be implemented)

connortsui20:
We need this for the genomics benchmark (it will panic without it)

robert3005 (approved):

PR #4602, opened 2025-09-11
row range scans to prune unneeded splits

onursatici:


vortex-scan/src/split_by.rs
```diff
@@ -31,32 +30,21 @@ impl SplitBy {
...
-                    .into_iter()
-                    .tuple_windows()
-                    .map(|(start, end)| start..end)
-                    .collect()
             }
             SplitBy::RowCount(n) => {
                 let row_count = *layout_reader.row_count().to_exact().ok_or_else(|| {
                     vortex_err!("Cannot split layout by row count, row count is not exact")
                 })?;
-                let mut splits =
```
gatesn, line 53: <3 I always found this a weird location for this logic

gatesn (approved):
Nice

PR #4605, opened 2025-09-11
Feature: Migrate genomics benchmark to `FixedSizeList`

connortsui20:

Also fixes a subtle bug in the duckdb exporter, where duckdb `ARRAY` is quite different from their `LIST` type.

This was mostly a proof of concept to see if the `FixedSizeList` holds up against some real data. It found a few things (#4592, #4601), but nothing else significant.

@danking and I observed that for scale factor 1, where most of the data is null, this actually performs worse than the variable-size list arrays. Part of this may be due to the compressor being better able to compress list sequences and "two-level" numeric sequences in nested types.

Note however that fixed-size lists vs Parquet are a lot better (but we think that is because Parquet is just much worse with fixed-size lists because it has to store a bunch of null values before it's black box compression).

I would expect the fixed-size lists to be better than variable size lists on much more dense data.

### DIFF
--- a/bench-vortex/src/statpopgen/builder.rs
+++ b/bench-vortex/src/statpopgen/builder.rs
@@ -4,8 +4,8 @@
 use std::sync::Arc;
 
 use arrow_array::builder::{
-    ArrayBuilder, BooleanBuilder, Float32Builder, Int32Builder, ListBuilder, StringBuilder,
-    UInt64Builder,
+    ArrayBuilder, BooleanBuilder, FixedSizeListBuilder, Float32Builder, Int32Builder, ListBuilder,
+    StringBuilder, UInt64Builder,
 };
 use arrow_array::{ArrayRef, RecordBatch};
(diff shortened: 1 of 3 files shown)


danking:
The other list-y science-y benchmark to consider is https://github.com/iris-hep/adl-benchmarks-index . I think there's fewer missing data but I'm not sure duckdb has all the primitives necessary for those queries.

connortsui20:
@claude review

connortsui20:
I'm thinking we probably don't want to merge this change? Or at least not until we figure out a better way to compress sparse fixed-size list data (which might come up a lot more in scientific computing and mathematics, though maybe not so much in AI/ML workloads).

connortsui20:
Closing since I don't think this really gives us much of a benefit (unless we are strictly talking about performance versus Parquet fixed-size lists).

CC @robert3005

robert3005:
I was hoping it would be faster but I guess we can't efficiently compress these yet

PR #4606, opened 2025-09-12
Fix: restore and correct dict compare where only single value matches the search

### LINKED ISSUE #4599: Restore dict array compare equals optimisation
In #4596 we changed how we encode dictionaries by default and as a result uncovered a bug in compare equals result. We should figure out what the bug was and move that logic to canonicalise call for Boolean dict arrays

robert3005:
This optimisation was generally applicable to canonicalizing DictArray
with bool values. I have moved the logic there and fixed the issue with
not respecting validity of the values array

fix #4599 

### DIFF
--- a/encodings/dict/src/array.rs
+++ b/encodings/dict/src/array.rs
@@ -4,12 +4,9 @@
 use std::fmt::Debug;
 
 use arrow_buffer::BooleanBuffer;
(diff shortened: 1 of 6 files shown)


AdamGS (approved):
Insane perf improvements 👀

PR #4610, opened 2025-09-12
Fix: `FixedSizeList` DuckDB exporter bug

connortsui20:

Changes checked out from: https://github.com/vortex-data/vortex/pull/4605

Fixes a bug in the DuckDB exporter where if we do not flatten / materialize the inner child of the fixed-size list before exporting the vector to DuckDB, DuckDB will segfault after running a function like `scan`.

Also fixes a bug where I didn't correctly append garbage values when the validity is all null (I took that from the list implementation, which doesn't need to append garbage values since null can be an empty list there). I don't actually know if this is an issue to be honest, but it seems wrong?

I was not able to figure out how to write regression tests for both of these. The early return might not actually be an issue, but the flattening at the end definitely is. However, I was not able to create a minimally reproducible test for that, and it only fails in query 7 and 8 in `statpopgen.sql`.

```sql
-- 7. Read just one variant (this is the sixth one).
SELECT *
  FROM statpopgen
 WHERE "CHROM" == 'chr21'
   AND "POS" == 5030278;
-- 8. Read a 700 base-pair window of variants.
SELECT *
  FROM statpopgen
 WHERE "CHROM" == 'chr21'
   AND "POS" >= 5030300
   AND "POS" <= 5031000;
```

If anyone has an idea of how to recreate this easily that would be appreciated!

Finally, I simplified the tests I wrote from before since a lot were redundant.

### DIFF
--- a/vortex-duckdb/src/exporter/fixed_size_list.rs
+++ b/vortex-duckdb/src/exporter/fixed_size_list.rs
@@ -56,12 +56,13 @@ impl ColumnExporter for FixedSizeListExporter {
 
         let list_size = self.list_size as usize;
 
+        // If all values are null, then we don't need to worry about exporting values (similar to
+        // the primitive exporter).
         // SAFETY: We've asserted that offset + len <= self.validity.len(), which ensures we won't
         // read past the validity mask bounds.
-        unsafe { vector.set_validity(&self.validity, offset, len) };
-
-        // TODO(connor): It may be possible to early return here if the null map is completely null,
-        // similar to how `List` does it.
+        if unsafe { vector.set_validity(&self.validity, offset, len) } {
+            return Ok(());
+        }
 
         // Get the child vector for array elements and export the elements directly.
         let mut elements_vector = vector.array_vector_get_child();
--- a/vortex-duckdb/src/exporter/fixed_size_list.rs
+++ b/vortex-duckdb/src/exporter/fixed_size_list.rs
@@ -54,25 +54,27 @@ impl ColumnExporter for FixedSizeListExporter {
             self.validity.len()
         );
 
-        // Set validity if necessary.
-        // SAFETY: We've asserted that offset + len <= self.validity.len(), which ensures
-        // we won't read past the validity mask bounds.
-        if unsafe { vector.set_validity(&self.validity, offset, len) } {
-            // All values are null, so no point copying the data.
-            return Ok(());
-        }
+        let list_size = self.list_size as usize;
 
-        // Get the child vector for array elements.
+        // SAFETY: We've asserted that offset + len <= self.validity.len(), which ensures we won't
+        // read past the validity mask bounds.
+        unsafe { vector.set_validity(&self.validity, offset, len) };
+
+        // TODO(connor): It may be possible to early return here if the null map is completely null,
+        // similar to how `List` does it.
+
+        // Get the child vector for array elements and export the elements directly.
         let mut elements_vector = vector.array_vector_get_child();
-
-        // Export elements directly.
-        // For fixed-size lists: elements start at offset * list_size
-        // and we export len * list_size elements.
-        let element_offset = offset * self.list_size as usize;
-        let element_count = len * self.list_size as usize;
-
         self.elements_exporter
-            .export(element_offset, element_count, &mut elements_vector)
+            .export(offset * list_size, len * list_size, &mut elements_vector)?;
+
+        // CRITICAL: We must flatten the child vector to ensure any child views (namely UTF-8 string
+        // views) are materialized.
+        // See https://github.com/vortex-data/vortex/pull/4610#issuecomment-3286676825 for a
+        // detailed explanation on why we need this.
+        elements_vector.flatten((len * list_size) as u64);
+
+        Ok(())
     }
 }
 
--- a/vortex-duckdb/src/exporter/fixed_size_list.rs
+++ b/vortex-duckdb/src/exporter/fixed_size_list.rs
@@ -69,10 +69,10 @@ impl ColumnExporter for FixedSizeListExporter {
         self.elements_exporter
             .export(offset * list_size, len * list_size, &mut elements_vector)?;
 
-        // CRITICAL: We must flatten the child vector to ensure any child views (namely UTF-8 string
-        // views) are materialized.
+        // TODO(connor): We must flatten the child vector to ensure any child dictionary views
+        // (namely UTF-8 string views in dictionaries) are materialized.
         // See https://github.com/vortex-data/vortex/pull/4610#issuecomment-3286676825 for a
-        // detailed explanation on why we need this.
+        // detailed explanation on why we need this for now.
         elements_vector.flatten((len * list_size) as u64);
 
         Ok(())
--- a/bench-vortex/statpopgen.sql
+++ b/bench-vortex/statpopgen.sql
@@ -18,7 +18,7 @@ SELECT "CHROM", "POS", "REF", "ALT", CAST(LIST_SUM(GT) AS DOUBLE) / (2 * LIST_SU
 SELECT "CHROM", "POS", "REF", "ALT", LIST_SUM(LIST_TRANSFORM(GT, lambda GT: GT IS NOT NULL)) AS n_called FROM statpopgen;
 -- 6.
 --
--- Collect the necessary statistics for a Hardy-Weinberg Equalibrium test. The
+-- Collect the necessary statistics for a Hardy-Weinberg Equilibrium test. The
 -- actual test involves the Levene-Haldane distribution which is somewhat subtle
 -- to implement in SQL.
 SELECT "CHROM", "POS", "REF", "ALT",
--- a/vortex-duckdb/src/e2e_test/vortex_scan_test.rs
+++ b/vortex-duckdb/src/e2e_test/vortex_scan_test.rs
@@ -14,7 +14,10 @@ use jiff::{Span, Timestamp, Zoned, tz};
 use num_traits::AsPrimitive;
 use tempfile::NamedTempFile;
 use vortex::IntoArray;
-use vortex::arrays::{BoolArray, ConstantArray, PrimitiveArray, StructArray, VarBinArray};
+use vortex::arrays::{
+    BoolArray, ConstantArray, FixedSizeListArray, ListArray, PrimitiveArray, StructArray,
+    VarBinArray, VarBinViewArray,
+};
 use vortex::buffer::buffer;
 use vortex::file::VortexWriteOptions;
 use vortex::scalar::Scalar;
@@ -385,4 +388,162 @@ fn test_write_timestamps() {
             .to_zoned(TimeZone::fixed(tz::offset(-7))),
         Zoned::from_str("2025-05-03 16:19:14.338895-07[-07]").unwrap()
     );
+}
+
+#[test]
+fn test_vortex_scan_nested_fixed_size_list_utf8() {
+    // Regression test for a segfault that occurs inside query 7 and 8 of the `statpopgen` benchmark
+    // when running with `FixedSizeList` instead of `List`.
+
+    // Test FixedSizeList of Utf8 to ensure proper materialization.
+
+    let runtime = tokio::runtime::Runtime::new().unwrap();
+    let file = runtime.block_on(async {
+        // Create a large number of strings to stress test.
+        let strings: Vec<&str> = (0..24)
+            .map(|i| match i % 6 {
+                0 => "first",
+                1 => "second",
+                2 => "third",
+                3 => "fourth",
+                4 => "fifth",
+                _ => "sixth",
+            })
+            .collect();
+
+        let strings_array = VarBinViewArray::from_iter_str(strings);
+
+        // Create inner fixed-size lists.
+        let inner_fsl = FixedSizeListArray::new(
+            strings_array.into_array(),
+            4, // 4 strings per inner list
+            Validity::AllValid,
+            6, // 6 inner lists
+        );
+
+        // Create outer fixed-size list of lists.
+        let outer_fsl = FixedSizeListArray::new(
+            inner_fsl.into_array(),
+            3, // 3 inner lists per outer list
+            Validity::AllValid,
+            2, // 2 outer lists
+        );
+
+        write_single_column_vortex_file("nested_string_lists", outer_fsl).await
+    });
+
+    let conn = database_connection();
+    let file_path = file.path().to_string_lossy();
+
+    // Query the nested structure.
+    let result = conn
+        .query(&format!(
+            "SELECT nested_string_lists FROM vortex_scan('{file_path}')"
+        ))
+        .unwrap();
+
+    let mut row_count = 0;
+    for chunk in result {
+        row_count += chunk.len();
+        // Accessing the nested structure should not cause a segfault.
+        let _vec = chunk.get_vector(0);
+    }
+    assert_eq!(row_count, 2, "Should have retrieved 2 outer lists");
+}
+
+#[test]
+fn test_vortex_scan_ultra_deep_nesting() {
+    // Test ultra-deep nesting: Multiple levels of FSL and List combinations with UTF8.
+    // FSL[List[FSL[List[FSL[UTF8]]]]]
+
+    let runtime = tokio::runtime::Runtime::new().unwrap();
+    let file = runtime.block_on(async {
+        // Level 1: Create base UTF8 strings - need a lot for deep nesting.
+        let strings = VarBinViewArray::from_iter_str(
+            (0..360)
+                .map(|i| match i % 10 {
+                    0 => "zero",
+                    1 => "one",
+                    2 => "two",
+                    3 => "three",
+                    4 => "four",
+                    5 => "five",
+                    6 => "six",
+                    7 => "seven",
+                    8 => "eight",
+                    _ => "nine",
+                })
+                .collect::<Vec<_>>(),
+        );
+
+        // Level 2: Inner-most FixedSizeList of strings.
+        let level2_fsl = FixedSizeListArray::new(
+            strings.into_array(),
+            5, // 5 strings per list
+            Validity::AllValid,
+            72, // 72 lists at this level
+        );
+
+        // Level 3: Variable-length lists of level 2 FSLs.
+        let level3_offsets = buffer![0i32, 3, 6, 8, 12, 15, 18, 20, 24, 27, 30, 32, 36];
+        let level3_list = ListArray::try_new(
+            level2_fsl.into_array(),
+            level3_offsets.into_array(),
+            Validity::AllValid,
+        )
+        .unwrap();
+
+        // Level 4: FixedSizeList of level 3 lists.
+        let level4_fsl = FixedSizeListArray::new(
+            level3_list.into_array(),
+            3, // 3 variable lists per FSL
+            Validity::AllValid,
+            4, // 4 FSLs at this level
+        );
+
+        // Level 5: Variable-length lists of level 4 FSLs.
+        let level5_offsets = buffer![0i32, 2, 4];
+        let level5_list = ListArray::try_new(
+            level4_fsl.into_array(),
+            level5_offsets.into_array(),
+            Validity::AllValid,
+        )
+        .unwrap();
+
+        // Level 6: Outermost FixedSizeList.
+        let outermost_fsl = FixedSizeListArray::new(
+            level5_list.into_array(),
+            2, // 2 lists per outermost FSL
+            Validity::AllValid,
+            1, // 1 outermost FSL
+        );
+
+        write_single_column_vortex_file("ultra_deep", outermost_fsl).await
+    });
+
+    let conn = database_connection();
+    let file_path = file.path().to_string_lossy();
+
+    // Query the ultra-deep nested structure.
+    let result = conn
+        .query(&format!("SELECT COUNT(*) FROM vortex_scan('{file_path}')"))
+        .unwrap();
+    let chunk = result.into_iter().next().unwrap();
+    let vec = chunk.get_vector(0);
+    let count = vec.as_slice_with_len::<i64>(chunk.len().as_())[0];
+    assert_eq!(count, 1, "Should have 1 outermost list");
+
+    // Try to access the data - this is the critical test for segfaults.
+    let result = conn
+        .query(&format!(
+            "SELECT ultra_deep FROM vortex_scan('{file_path}')"
+        ))
+        .unwrap();
+
+    let mut row_count = 0;
+    for chunk in result {
+        row_count += chunk.len();
+        let _vec = chunk.get_vector(0);
+    }
+    assert_eq!(row_count, 1, "Should have retrieved 1 row");
 }


vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -54,28 +54,34 @@ impl ColumnExporter for FixedSizeListExporter {
...
         // we won't read past the validity mask bounds.
-        if unsafe { vector.set_validity(&self.validity, offset, len) } {
-            // All values are null, so no point copying the data.
-            return Ok(());
-        }
+        unsafe { vector.set_validity(&self.validity, offset, len) };
+
+        // Note: Unlike variable-size lists, for fixed-size lists (ARRAY type) we must always export
+        // the child data even if all values are null, because the ARRAY type has a fixed size that
+        // must be respected. The child vector needs to have the correct number of elements.
```
connortsui20: Is this actually true? I am unable to write a test that would trigger this incorrect(?) behavior

vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -54,28 +54,34 @@ impl ColumnExporter for FixedSizeListExporter {
...
-
         self.elements_exporter
-            .export(element_offset, element_count, &mut elements_vector)
+            .export(element_offset, element_count, &mut elements_vector)?;
+
+        // CRITICAL: We must flatten the child vector to ensure the data is materialized.
+        // The child vector returned by `array_vector_get_child()` is borrowed and only valid while
+        // the parent is valid. After a `scan` returns, DuckDB accesses this data and will segfault
+        // if it's not materialized.
+        elements_vector.flatten(element_count as u64);
```
danking: AFAIK, we should not need to decompress the data.

I do think you need to call [`duckdb_list_vector_reserve`](https://duckdb.org/docs/stable/clients/c/vector.html#duckdb_list_vector_reserve). I think you also need to call [`duckdb_list_vector_set_size`](https://duckdb.org/docs/stable/clients/c/vector.html#duckdb_list_vector_set_size) on the list vector (not the elements vector) to specify the number of lists.

danking (changes requested):
I think to cause a panic you would need to actually run a duckdb query. Given the failing queries in the benchmark, I suspect you need to do something that filters or takes elements of the table.

vortex-duckdb/src/exporter/fixed_size_list.rs
danking: As discussed offline, this is incorrect.

connortsui20:
So I asked Claude to research into why I was observing what I did, and I had it go look through the duckdb source code as well. Here is the results of that research. TLDR; I think we definitely need the `flatten` as there is no equivalent `reserve` for `ARRAY` type in DuckDB.

cc @danking 

# Understanding the `flatten()` Requirement for FixedSizeList with UTF8 in DuckDB

## Executive Summary

The `flatten()` call on line 77 in `vortex-duckdb/src/exporter/fixed_size_list.rs` is **absolutely critical** for preventing segmentation faults when exporting FixedSizeList arrays containing UTF8 strings to DuckDB. Without it, DuckDB will crash when querying the data.

This requirement is due to fundamental architectural differences between how DuckDB's LIST (variable-size) and ARRAY (fixed-size) types manage their child vectors and string buffers.

## The Problem

When we export a `FixedSizeList<UTF8>` to DuckDB without calling `flatten()`:

- The export appears to succeed
- But when DuckDB queries the data, it **segfaults**
- This ONLY happens with UTF8/string data, not with primitive types like `i32`
- Regular `List<UTF8>` works fine without `flatten()`

## Root Cause: LIST vs ARRAY Architecture in DuckDB

### Key Insight

DuckDB's LIST and ARRAY types have fundamentally different memory management models for their child vectors:

- **LIST**: Dynamically allocates and owns its child vector
- **ARRAY**: Returns a view/reference to an embedded structure

This difference is critical when dealing with string data that requires auxiliary buffers.

## Memory Layout Comparison

### LIST Memory Structure

```
┌──────────────────────────────────────────────────────────────┐
│                         LIST Vector                          │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ Header:                                                  │ │
│ │  - Type: LIST                                            │ │
│ │  - Count: 3                                              │ │
│ └──────────────────────────────────────────────────────────┘ │
│                                                              │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ Validity Mask: [1, 0, 1]  (bit array)                    │ │
│ └──────────────────────────────────────────────────────────┘ │
│                                                              │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ List Entries Array:                                      │ │
│ │  ┌─────────────┬─────────────┐                           │ │
│ │  │ offset: 0   │ length: 3   │ <- List 0                 │ │
│ │  ├─────────────┼─────────────┤                           │ │
│ │  │ offset: 3   │ length: 0   │ <- List 1 (NULL)          │ │
│ │  ├─────────────┼─────────────┤                           │ │
│ │  │ offset: 3   │ length: 2   │ <- List 2                 │ │
│ │  └─────────────┴─────────────┘                           │ │
│ └──────────────────────────────────────────────────────────┘ │
│                                                              │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ Child Vector Pointer: 0x7fff8c024100 ──────────────────┐ │ │
│ └──────────────────────────────────────────────────────────┘ │
└──────────────────────────────────────────────────────────────┘
                                                           |
                                                           v
        ┌─────────────────────────────────────────────────────────────────┐
        │                  Child Vector (HEAP ALLOCATED)                  │
        │ ┌─────────────────────────────────────────────────────────────┐ │
        │ │ Header:                                                     │ │
        │ │  - Type: VARCHAR                                            │ │
        │ │  - Size: 5 (total elements across all lists)                │ │
        │ │  - Capacity: 100 (allocated via list_vector_reserve)        │ │
        │ └─────────────────────────────────────────────────────────────┘ │
        │                                                                 │
        │ ┌─────────────────────────────────────────────────────────────┐ │
        │ │ String Views Array:                                         │ │
        │ │  [0]: {len: 15, ptr: 0x8a001000}  <- "hello world..."       │ │
        │ │  [1]: {len: 20, ptr: 0x8a001100}  <- "another long string"  │ │
        │ │  [2]: {len: 8,  inline: "short"}   <- Inlined (<= 12 bytes) │ │
        │ │  [3]: {len: 25, ptr: 0x8a001200}  <- "more string data..."  │ │
        │ │  [4]: {len: 10, inline: "test data"} <- Inlined             │ │
        │ └─────────────────────────────────────────────────────────────┘ │
        │                                                                 │
        │ ┌─────────────────────────────────────────────────────────────┐ │
        │ │ String Buffer List (OWNED):                                 │ │
        │ │  ┌──────────────────────────────────────────┐               │ │
        │ │  │ Buffer 0: 0x8a001000                     │               │ │
        │ │  │  "hello world..."                        │               │ │
        │ │  ├──────────────────────────────────────────┤               │ │
        │ │  │ Buffer 1: 0x8a001100                     │               │ │
        │ │  │  "another long string"                   │               │ │
        │ │  ├──────────────────────────────────────────┤               │ │
        │ │  │ Buffer 2: 0x8a001200                     │               │ │
        │ │  │  "more string data..."                   │               │ │
        │ │  └──────────────────────────────────────────┘               │ │
        │ └─────────────────────────────────────────────────────────────┘ │
        └─────────────────────────────────────────────────────────────────┘
```

### ARRAY Memory Structure (Without flatten - BROKEN)

```
┌──────────────────────────────────────────────────────────────┐
│                         ARRAY Vector                         │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ Header:                                                  │ │
│ │  - Type: ARRAY                                           │ │
│ │  - Array Size: 3 (fixed elements per array)              │ │
│ │  - Count: 2 (number of arrays)                           │ │
│ └──────────────────────────────────────────────────────────┘ │
│                                                              │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ Validity Mask: [1, 1]  (both arrays valid)               │ │
│ └──────────────────────────────────────────────────────────┘ │
│                                                              │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ NO Offset/Length Data!                                   │ │
│ │ (Array uses fixed positioning: index * array_size)       │ │
│ └──────────────────────────────────────────────────────────┘ │
│                                                              │
│ ┌──────────────────────────────────────────────────────────┐ │
│ │ Child Vector (EMBEDDED IN PARENT STRUCT):                │ │
│ │ ┌────────────────────────────────────────────────────┐   │ │
│ │ │ This is NOT a pointer! It's part of the struct:    │   │ │
│ │ │                                                    │   │ │
│ │ │ struct ArrayVector {                               │   │ │
│ │ │     ValidityMask validity;                         │   │ │
│ │ │     Vector child;  // <- EMBEDDED, not allocated   │   │ │
│ │ │     ...                                            │   │ │
│ │ │ }                                                  │   │ │
│ │ └────────────────────────────────────────────────────┘   │ │
│ │                                                          │ │
│ │ ┌────────────────────────────────────────────────────┐   │ │
│ │ │ Child Vector Data (pre-allocated):                 │   │ │
│ │ │  - Type: VARCHAR                                   │   │ │
│ │ │  - Fixed Size: STANDARD_VECTOR_SIZE * array_size   │   │ │
│ │ │  - This is a VIEW, not owned memory!               │   │ │
│ │ └────────────────────────────────────────────────────┘   │ │
│ │                                                          │ │
│ │ ┌────────────────────────────────────────────────────┐   │ │
│ │ │ String Views (in embedded buffer):                 │   │ │
│ │ │  [0]: {len: 15, ptr: 0x9b002000} <- Points to buff │   │ │
│ │ │  [1]: {len: 20, ptr: 0x9b002100} <- Points to buff │   │ │
│ │ │  [2]: {len: 8,  inline: "short"}                   │   │ │
│ │ │  [3]: {len: 25, ptr: 0x9b002200} <- Points to buff │   │ │
│ │ │  [4]: {len: 10, inline: "test"}                    │   │ │
│ │ │  [5]: {len: 18, ptr: 0x9b002300} <- Points to buff │   │ │
│ │ └────────────────────────────────────────────────────┘   │ │
│ │                                                          │ │
│ │ ┌────────────────────────────────────────────────────┐   │ │
│ │ │ String Buffers (PROBLEM - attached to VIEW!):      │   │ │
│ │ │  These are added via add_string_buffer() but:      │   │ │
│ │ │  - Attached to a view, not owned vector            │   │ │
│ │ │  - Unclear lifetime/ownership                      │   │ │
│ │ │  - May be freed when export function returns!      │   │ │
│ │ │                                                    │   │ │
│ │ │  Buffer 0: 0x9b002000 <- INVALID after return!     │   │ │
│ │ │  Buffer 1: 0x9b002100 <- INVALID after return!     │   │ │
│ │ │  Buffer 2: 0x9b002200 <- INVALID after return!     │   │ │
│ │ │  Buffer 3: 0x9b002300 <- INVALID after return!     │   │ │
│ │ └────────────────────────────────────────────────────┘   │ │
│ └──────────────────────────────────────────────────────────┘ │
└──────────────────────────────────────────────────────────────┘

[WARNING] DANGER: When export function returns, string buffer pointers become invalid!
         DuckDB will SEGFAULT when trying to access these strings!
```

### ARRAY Memory Structure (With flatten - FIXED)

```

┌───────────────────────────────────────────────────────────────┐
│                         ARRAY Vector                          │
│ ┌──────────────────────────────────────────────────────────┐  │
│ │ Header:                                                  │  │
│ │  - Type: ARRAY                                           │  │
│ │  - Array Size: 3                                         │  │
│ │  - Count: 2                                              │  │
│ └──────────────────────────────────────────────────────────┘  │
│                                                               │
│ ┌──────────────────────────────────────────────────────────┐  │
│ │ Child Vector (AFTER flatten()):                          │  │
│ │ ┌────────────────────────────────────────────────────┐   │  │
│ │ │ flatten() creates a NEW FLAT VECTOR:               │   │  │
│ │ │  1. Allocates new buffer                           │   │  │
│ │ │  2. Copies all data                                │   │  │
│ │ │  3. Properly attaches string buffers               │   │  │
│ │ │  4. Ensures ownership and persistence              │   │  │
│ │ └────────────────────────────────────────────────────┘   │  │
│ │                                                          │  │
│ │ ┌────────────────────────────────────────────────────┐   │  │
│ │ │ Flattened Vector Data (PROPERLY OWNED):            │   │  │
│ │ │  - Type: VARCHAR                                   │   │  │
│ │ │  - Vector Type: FLAT_VECTOR                        │   │  │
│ │ │  - All data materialized and owned                 │   │  │
│ │ └────────────────────────────────────────────────────┘   │  │
│ │                                                          │  │
│ │ ┌────────────────────────────────────────────────────┐   │  │
│ │ │ String Buffers (PROPERLY OWNED & PERSISTED):       │   │  │
│ │ │  ┌──────────────────────────────────────────┐      │   │  │
│ │ │  │ Buffer 0: "hello world..."               │      │   │  │
│ │ │  │  - Properly attached to flattened vector │      │   │  │
│ │ │  │  - Survives after export returns         │      │   │  │
│ │ │  ├──────────────────────────────────────────┤      │   │  │
│ │ │  │ Buffer 1: "another long string"          │      │   │  │
│ │ │  │  - Reference counted                     │      │   │  │
│ │ │  │  - Owned by the vector                   │      │   │  │
│ │ │  ├──────────────────────────────────────────┤      │   │  │
│ │ │  │ Buffer 2: "more string data..."          │      │   │  │
│ │ │  │  - Valid for entire query lifetime       │      │   │  │
│ │ │  └──────────────────────────────────────────┘      │   │  │
│ │ └────────────────────────────────────────────────────┘   │  │
│ └──────────────────────────────────────────────────────────┘  │
└────────────────────────────────────────────────────────────────┘

[OK] SUCCESS: String buffers properly materialized and will persist!
```

## Code Comparison

### How LIST Manages Child Vectors

```rust
// list.rs - Working code that doesn't need flatten()
vector.list_vector_reserve(sum_of_list_lens)?;  // <- ALLOCATES a new buffer
let mut elements_vector = vector.list_vector_get_child();  // <- Returns OWNED vector
vector.list_vector_set_size(sum_of_list_lens)?;
self.elements_exporter.export(..., &mut elements_vector)?;
// No flatten() needed - child vector is properly owned!
```

### How ARRAY Manages Child Vectors

```rust
// fixed_size_list.rs - Requires flatten() for strings
let mut elements_vector = vector.array_vector_get_child();  // <- Returns a VIEW
self.elements_exporter.export(..., &mut elements_vector)?;
elements_vector.flatten(element_count as u64);  // <- CRITICAL for strings!
```

## Detailed Export Flow Diagrams

### LIST with UTF8 Export Flow (Success)

```

┌─────────────────────────────────────────────────────────────────────┐
│ Step 1: list_vector_reserve(100)                                    │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  LIST Vector                   Heap Memory                          │
│  ┌──────────┐                 ┌─────────────────────┐               │
│  │ child: ──┼────────────────>│ NEW Vector          │               │
│  └──────────┘                 │ capacity: 100       │               │
│                               │ buffer: allocated   │               │
│                               └─────────────────────┘               │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 2: VarBinView exports strings                                  │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  Child Vector                  String Buffers                       │
│  ┌──────────────┐             ┌─────────────────┐                   │
│  │ views: [...] │             │ Buffer 1        │                   │
│  │ buffers: ────┼────────────>│ "hello world"   │                   │
│  └──────────────┘             ├─────────────────┤                   │
│                               │ Buffer 2        │                   │
│                               │ "test string"   │                   │
│                               └─────────────────┘                   │
│                               (OWNED by child vector)               │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 3: Export function returns                                     │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  [OK] Child vector still exists (owned by LIST)                     │
│  [OK] String buffers still attached and valid                       │
│  [OK] All references remain valid                                   │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 4: DuckDB queries the data                                     │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  DuckDB: "Give me string at position 0"                             │
│  Vector: views[0] -> Buffer 1 -> "hello world" [OK]                 │
│                                                                     │
│  SUCCESS: All data accessible!                                      │
└─────────────────────────────────────────────────────────────────────┘
```

### ARRAY with UTF8 Export Flow WITHOUT flatten (Failure)

```
┌─────────────────────────────────────────────────────────────────────┐
│ Step 1: array_vector_get_child()                                    │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  ARRAY Vector                                                       │
│  ┌────────────────────────┐                                         │
│  │ struct ArrayVector {   │                                         │
│  │   validity: [...]      │                                         │
│  │   child: Vector {      │ <─ Returns pointer to this              │
│  │     data: [...]        │   (embedded struct member)              │
│  │   }                    │                                         │
│  │ }                      │                                         │
│  └────────────────────────┘                                         │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 2: VarBinView exports strings to child VIEW                    │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  Child View                    Temporary Buffers?                   │
│  ┌──────────────┐             ┌─────────────────┐                   │
│  │ views: [...] │             │ Buffer 1        │                   │
│  │ buffers: ────┼────???────> │ "hello world"   │                   │
│  └──────────────┘             └─────────────────┘                   │
│         ^                                                           │
│         │                     [WARNING] Attached to a VIEW!         │
│    Just a view!               Ownership unclear!                    │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 3: Export function returns                                     │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  [X] Child "vector" was just a view - not independently owned       │
│  [X] String buffer references may be freed                          │
│  [X] Pointers now point to invalid memory!                          │
│                                                                     │
│  Stack frame destroyed:                                             │
│  ┌──────────────┐             ┌─────────────────┐                   │
│  │ views: [...] │             │ FREED MEMORY    │                   │
│  │ buffers: ────┼────────────>│ ????????????    │                   │
│  └──────────────┘             └─────────────────┘                   │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 4: DuckDB queries the data                                     │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  DuckDB: "Give me string at position 0"                             │
│  Vector: views[0] -> 0x9b002000 -> ??????                           │
│                         ^                                           │
│                         │                                           │
│                    INVALID POINTER!                                 │
│                                                                     │
│  [CRASH] SEGMENTATION FAULT!                                        │
└─────────────────────────────────────────────────────────────────────┘
```

### ARRAY with UTF8 Export Flow WITH flatten (Success)

```
┌─────────────────────────────────────────────────────────────────────┐
│ Step 1 & 2: Get child and export (same as above)                    │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 3: flatten(element_count)                                      │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  flatten() operation:                                               │
│                                                                     │
│  1. Create new FLAT_VECTOR:                                         │
│     ┌────────────────────┐                                          │
│     │ type: FLAT_VECTOR  │                                          │
│     │ buffer: NEW ALLOC  │                                          │
│     └────────────────────┘                                          │
│                                                                     │
│  2. Copy all data from view -> new buffer                           │
│     [view data] ──────copy──────> [new buffer]                      │
│                                                                     │
│  3. Re-attach string buffers to new vector:                         │
│     ┌────────────────────┐      ┌─────────────────┐                 │
│     │ new vector         │      │ String Buffer 1 │                 │
│     │ buffers: ──────────┼────> │ "hello world"   │                 │
│     └────────────────────┘      └─────────────────┘                 │
│                                  (Now properly owned!)              │
│                                                                     │
│  4. Update parent's child to point to flattened vector              │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 4: Export function returns                                     │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  [OK] Flattened data persists in parent's structure                 │
│  [OK] String buffers properly attached and owned                    │
│  [OK] All references remain valid                                   │
└─────────────────────────────────────────────────────────────────────┘
                                        │
                                        v
┌─────────────────────────────────────────────────────────────────────┐
│ Step 5: DuckDB queries the data                                     │
├─────────────────────────────────────────────────────────────────────┤
│                                                                     │
│  DuckDB: "Give me string at position 0"                             │
│  Vector: views[0] -> Buffer 1 -> "hello world" [OK]                 │
│                                                                     │
│  SUCCESS: All data accessible!                                      │
└─────────────────────────────────────────────────────────────────────┘
```

## Why Only UTF8 Strings Are Affected

### Primitive Types (Work Fine)

```

ARRAY Vector with i32 elements:
┌──────────────────────────────┐
│ Child (embedded):            │
│  ┌────────────────────────┐  │
│  │ i32 data: [1,2,3,4,5]  │  │ <- Data stored INLINE
│  └────────────────────────┘  │   No external buffers needed!
└──────────────────────────────┘   View model works fine
```

### UTF8/String Types (Require flatten())

```
ARRAY Vector with UTF8 elements:
┌──────────────────────────────┐
│ Child (embedded):            │
│  ┌────────────────────────┐  │
│  │ string_t views:        │  │
│  │  [0]: ptr -> ???       │  │ <- Needs external buffer!
│  │  [1]: ptr -> ???       │  │ <- Buffer ownership unclear
│  └────────────────────────┘  │
└──────────────────────────────┘
         [WARNING] Without flatten(), buffer pointers invalid!
```

## What `flatten()` Does

The `flatten()` operation performs these critical steps:

1. **Checks vector type**: Determines if flattening is needed
2. **Allocates new buffer**: Creates properly-sized flat vector
3. **Copies data**: Transfers all data to new buffer
4. **Handles string buffers**:
   - For VARCHAR types, ensures all string buffers are attached
   - For FSST compressed strings, decompresses them
   - Maintains reference counting for buffer ownership
5. **Updates vector type**: Marks as FLAT_VECTOR
6. **Ensures persistence**: Data now owned and will survive

## Implementation Difference (Conceptual C++)

### LIST's Child Vector Management

```cpp
// Simplified conceptual implementation
class ListVector {
    Vector* child = nullptr;  // Heap-allocated child

    void Reserve(size_t size) {
        // ALLOCATES a new child vector
        this->child = new Vector(VARCHAR, size);
        this->child->buffer = allocate_buffer(size * sizeof(string_t));
    }

    Vector* GetChild() {
        return this->child;  // Returns OWNED pointer
    }

    ~ListVector() {
        delete this->child;  // Properly cleans up
    }
};
```

### ARRAY's Child Vector Management

```cpp
// Simplified conceptual implementation
class ArrayVector {
    Vector child;  // EMBEDDED member, not a pointer!
    size_t array_size;

    ArrayVector(size_t array_size, size_t count) {
        this->array_size = array_size;
        // Child is pre-initialized as part of struct
        this->child.data = this->internal_buffer;
        this->child.size = STANDARD_VECTOR_SIZE * array_size;
    }

    Vector* GetChild() {
        return &this->child;  // Returns ADDRESS of member
    }
    // No explicit cleanup needed - child is part of struct
};
```

## Evidence from Testing

The failing tests demonstrate this issue clearly:

### Tests That FAIL Without flatten()

1. **test_vortex_scan_nested_fixed_size_list_utf8**:

   ```
   Structure: FSL[FSL[UTF8]]
   Result: SEGFAULT without flatten()
   Reason: Nested string buffers not properly materialized
   ```

2. **test_vortex_scan_ultra_deep_nesting**:
   ```
   Structure: FSL[List[FSL[List[FSL[UTF8]]]]]
   Result: SEGFAULT without flatten()
   Reason: Deep nesting with strings at leaf level
   ```

### Tests That WORK Without flatten()

3. **Nested integers**:

   ```
   Structure: FSL[FSL[i32]]
   Result: Works fine
   Reason: No external buffers needed for i32
   ```

4. **Variable lists with strings**:

   ```
   Structure: List[UTF8]
   Result: Works fine
   Reason: LIST allocates owned child vector
   ```

5. **Any depth with primitives**:
   ```
   Structure: FSL[List[FSL[List[FSL[i64]]]]]
   Result: Works fine
   Reason: Primitive types stored inline
   ```

## Why DuckDB Made This Design Choice

DuckDB's design reflects different optimization priorities:

### LIST (Variable-Size) Design Goals

- **Flexibility**: Different numbers of elements per row
- **Dynamic allocation**: Allocates exactly what's needed
- **General purpose**: Works for all scenarios
- **Trade-off**: Slightly more overhead for management

### ARRAY (Fixed-Size) Design Goals

- **Performance**: No offset/length tracking needed
- **Cache efficiency**: Predictable memory access patterns
- **SIMD optimization**: Fixed strides enable vectorization
- **Direct indexing**: `element = base + (index * array_size)`
- **Trade-off**: Requires explicit materialization for complex types

## Performance Characteristics

### LIST Access Pattern

```
To access element at list[i][j]:
1. Read list_entry[i] -> {offset: 10, length: 5}
2. Calculate: child_index = offset + j = 10 + j
3. Access: child_vector[child_index]
(Requires offset lookup + addition)
```

### ARRAY Access Pattern

```
To access element at array[i][j]:
1. Calculate: child_index = i * array_size + j
2. Access: child_vector[child_index]
(Direct calculation - no lookup needed!)
```

### Why It's Required

- ARRAY's child vector is an embedded view, not an owned allocation
- String data requires external buffers that must be properly owned
- Without `flatten()`, string buffer references become invalid after export
- DuckDB crashes with segfault when accessing invalid string pointers

### Why It's Not Needed for LIST

- LIST allocates and owns its child vector via `list_vector_reserve()`
- String buffers attached to owned vectors persist properly
- The ownership chain is clear and maintained

### Why It's Not Needed for Primitives

- Primitive types store data inline in the vector's buffer
- No external buffers or pointers involved
- The view/embedded model works fine for inline data

vortex-duckdb/src/exporter/fixed_size_list.rs
connortsui20: See https://github.com/vortex-data/vortex/pull/4610#issuecomment-3286676825

danking:
I don't think I believe the LLM. The docs seem pretty clear that the child vector lives as long as the parent vector.

> [duckdb_array_vector_get_child](https://duckdb.org/docs/stable/clients/c/api#duckdb_array_vector_get_child)
> 
> Retrieves the child vector of an array vector.
> 
> The resulting vector is valid as long as the parent vector is valid. The resulting vector has the size of the parent vector multiplied by the array size.

Moreover, when we export the elements (the strings), the varbinview exporter explicitly increments assigns shared ownership for those buffers to the (child) vector [here](https://github.com/vortex-data/vortex/blob/develop/vortex-duckdb/src/exporter/varbinview.rs#L46). 

I don't know what "VIEW" means in this context, I don't see that terminology in the duckdb docs at all.

connortsui20:
> The docs seem pretty clear that the child vector lives as long as the parent vector.

To be clear, the issue that the LLM came up with was that the child vector does not own the string buffers (what I was referring to as the view) when they get added. Note that it _is_ correct when it says `duckdb_array_vector_get_child` does not allocate (see the duckdb code [here](https://github.com/duckdb/duckdb/blob/f8761864efe92a157a0248bea3fb7446de13489a/src/main/capi/data_chunk-c.cpp#L194)).

So yeah I think you are right, the issue comes from something related to the Utf8 "views" (or ownership/borrowing of the string buffers). Just so I can copy and paste this later, the `add_string_buffer` function call goes down to [`StringVector::AddBuffer`](https://github.com/duckdb/duckdb/blob/f8761864efe92a157a0248bea3fb7446de13489a/src/common/types/vector.cpp#L2178).

I think on slack you said there might be an issue with the dictionaries? Maybe something here is unexpected for us?

```cpp
/* In duckdb-1.3.2/src/include/duckdb/common/types.hpp */

template <class T, typename... ARGS>
buffer_ptr<T> make_buffer(ARGS &&...args) { // NOLINT: mimic std casing
    // I'm assuming this is the same as the constructor for `shared_ptr`.
	return make_shared_ptr<T>(std::forward<ARGS>(args)...);
}

<-- snip -->

/* In duckdb-1.3.2/src/common/types/vector.cpp */

// Returns a shared pointer to a string buffer (I think).
VectorStringBuffer &StringVector::GetStringBuffer(Vector &vector) {
	if (vector.GetType().InternalType() != PhysicalType::VARCHAR) {
		throw InternalException("StringVector::GetStringBuffer - vector is not of internal type VARCHAR but of type %s",
		                        vector.GetType());
	}
	if (!vector.auxiliary) {
		vector.auxiliary = make_buffer<VectorStringBuffer>();
	}
	D_ASSERT(vector.auxiliary->GetBufferType() == VectorBufferType::STRING_BUFFER);
	return vector.auxiliary.get()->Cast<VectorStringBuffer>();
}

<-- snip -->

void StringVector::AddBuffer(Vector &vector, buffer_ptr<VectorBuffer> buffer) {
	D_ASSERT(buffer.get() != vector.auxiliary.get());
	auto &string_buffer = GetStringBuffer(vector);
	string_buffer.AddHeapReference(std::move(buffer));
}

void StringVector::AddHeapReference(Vector &vector, Vector &other) {
	D_ASSERT(vector.GetType().InternalType() == PhysicalType::VARCHAR);
	D_ASSERT(other.GetType().InternalType() == PhysicalType::VARCHAR);

	if (other.GetVectorType() == VectorType::DICTIONARY_VECTOR) {
        // For dictionaries, we add a heap reference to the dictionary values.
        // Don't really know where the codes are added (no call to `DictionaryVector::SelVector`)...
		StringVector::AddHeapReference(vector, DictionaryVector::Child(other));
		return;
	}
	if (!other.auxiliary) {
		return;
	}
	StringVector::AddBuffer(vector, other.auxiliary);
}
```

At the very least this is probably enough to figure out what might be wrong? Though I think I don't know enough about the duckdb part of our codebase to solve this.

cc @danking and @0ax1 since I saw you looked at ownership here a few months back

0ax1:
```
// Returns a shared pointer to a string buffer (I think).
VectorStringBuffer &StringVector::GetStringBuffer(Vector &vector) {
```
This does not copy the shared pointer, but hands out a ref to the shared pointer. The retain count of the shared pointer is therefore not +1 after calling this function

vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -54,25 +54,27 @@ impl ColumnExporter for FixedSizeListExporter {
...
-            // All values are null, so no point copying the data.
-            return Ok(());
-        }
+        let list_size = self.list_size as usize;
 
-        // Get the child vector for array elements.
-        let mut elements_vector = vector.array_vector_get_child();
+        // SAFETY: We've asserted that offset + len <= self.validity.len(), which ensures we won't
+        // read past the validity mask bounds.
+        unsafe { vector.set_validity(&self.validity, offset, len) };
```
robert3005: you could return early here like you used to?

connortsui20: Is that correct? It isn't obvious to me that we do not need to call `export` on the child array. Though I will note that I wasn't able to trigger an error when I had the early return there before (but I also wasn't testing e2e queries).

robert3005: This is the case of the array being all null so we should be fine - we do it everywhere else

danking (approved):

0ax1:
We gotta follow up here on `test_vortex_scan_ultra_deep_nesting` which fires an assert in DuckDB when built in debug: https://github.com/duckdb/duckdb/blob/464ae747bcf33971679ae80bcce481adf34902c3/src/common/types/vector.cpp#L1802

I'll land the PR to opt into DuckDB debug builds before. https://github.com/vortex-data/vortex/compare/ad/duckdb-debug

joseph-isaacs:
@connortsui20 Why didn't you canoncalize the elements vector once in the exporter ctor?

connortsui20:
@joseph-isaacs Shouldn't the `new_array_exporter` function take care of that?

PR #4614, opened 2025-09-12
chore: expose DeserializeStep

danking:
I'm not sure how else to interact with this enum. It is returned by `footer.into_deserializer().deserialize()`.

### DIFF
--- a/vortex-file/src/lib.rs
+++ b/vortex-file/src/lib.rs
@@ -109,7 +109,7 @@ mod writer;
 use std::sync::{Arc, LazyLock};
 
 pub use file::*;
-pub use footer::{DeserializeStep, Footer, SegmentSpec};
+pub use footer::*;
 pub use forever_constant::*;
 pub use generic::*;
 pub use memory::*;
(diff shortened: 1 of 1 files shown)


AdamGS (approved):

AdamGS:
sure, but isn't it exposed through another way? thought clippy would complain about something like this

vortex-file/src/lib.rs
```diff
@@ -109,7 +109,7 @@ mod writer;
 use std::sync::{Arc, LazyLock};
 
 pub use file::*;
-pub use footer::{Footer, SegmentSpec};
+pub use footer::{DeserializeStep, Footer, SegmentSpec};
```
gatesn: Can we just use `footer::*`, our convention tends to be anything public should be fully public.

danking: Done.

danking:
> sure, but isn't it exposed through another way? thought clippy would complain about something like this

I was likewise pretty confused that there was no way to import it. I would've thought this was a compile error, but I guess it *is* marked public, so you can receive that type and, say, pass it somewhere else, you just can't name it. 🤷

Issue #4615, opened 2025-09-12
Python check script does not work without some packages


connortsui20:
I had to install `basedpyright` to get the first steps of the `./vortex-python/check.sh` script to work, but now I am running into these errors and warnings (on `develop`, on my machine):

```
/Users/connor/spiral/vortex-data/vortex-develop/vortex-python/python/vortex/arrays.py
  /Users/connor/spiral/vortex-data/vortex-develop/vortex-python/python/vortex/arrays.py:47:12 - warning: Type of "from_arrays" is partially unknown
    Type of "from_arrays" is "(arrays: Collection[Array[Any] | ChunkedArray[Any]], names: list[str] | None = None, schema: Schema | None = None, metadata: Mapping[Unknown, Unknown] | None = None) -> Table" (reportUnknownMemberType)
/Users/connor/spiral/vortex-data/vortex-develop/vortex-python/python/vortex/dataset.py
  /Users/connor/spiral/vortex-data/vortex-develop/vortex-python/python/vortex/dataset.py:49:9 - error: Method "count_rows" overrides class "Dataset" in an incompatible manner
    Parameter 2 name mismatch: base parameter is named "columns", override parameter is named "filter"
    Parameter 3 name mismatch: base parameter is named "filter", override parameter is named "batch_size"
    Parameter 4 name mismatch: base parameter is named "batch_size", override parameter is named "batch_readahead"
    Parameter 5 name mismatch: base parameter is named "batch_readahead", override parameter is named "fragment_readahead"
    Parameter 6 name mismatch: base parameter is named "fragment_readahead", override parameter is named "fragment_scan_options"
    Parameter 7 name mismatch: base parameter is named "fragment_scan_options", override parameter is named "use_threads"
    Parameter 8 name mismatch: base parameter is named "use_threads", override parameter is named "cache_metadata"
    Parameter 9 name mismatch: base parameter is named "cache_metadata", override parameter is named "memory_pool"
    ... (reportIncompatibleMethodOverride)
  /Users/connor/spiral/vortex-data/vortex-develop/vortex-python/python/vortex/dataset.py:572:9 - error: Method "to_batches" overrides class "Fragment" in an incompatible manner
    Parameter 2 name mismatch: base parameter is named "columns", override parameter is named "schema"
    Parameter 3 name mismatch: base parameter is named "filter", override parameter is named "columns"
    Parameter 4 name mismatch: base parameter is named "batch_size", override parameter is named "filter"
[...]

PR #4617, opened 2025-09-12
Chore: add `uv sync --all-packages` in `check.sh`

connortsui20:
Relevant: https://github.com/vortex-data/vortex/issues/4615

### DIFF
--- a/vortex-python/check.sh
+++ b/vortex-python/check.sh
@@ -4,6 +4,9 @@
 
 set -ex -o pipefail
 
+# Ensure all packages are synced.
+uv sync --all-packages
+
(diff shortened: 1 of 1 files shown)


AdamGS:
you don't need to explicitly sync, you can just add --all-packages to the first command and uv will take care of the rest

connortsui20:
> you don't need to explicitly sync, you can just add --all-packages to the first command and uv will take care of the rest

Do you mean `uv --all-packages`? That doesn't work on my machine, though I'm assuming you mean something else.

AdamGS:
I assumed we used `uv` elsewhere in this script, seeing as we don't, this seems like the right solution

robert3005 (approved):

PR #4618, opened 2025-09-12
Feature: Implement `FixedSizeList` in `PyVortex` bindings

connortsui20:

Implements `FixedSizeList` into the Python bindings.

# Notes

- We expose that we do not make a distinction between `List` scalars and `FixedSizeList` scalars here. This is notably different from how `pyarrow` does it, which has both a [`ListScalar`](https://arrow.apache.org/docs/python/generated/pyarrow.ListScalar.html#pyarrow.ListScalar) and a [`FixedSizeListScalar`](https://arrow.apache.org/docs/python/generated/pyarrow.FixedSizeListScalar.html) which is derived from `ListScalar`. Is this what we want?
- If this _is_ what we want, then the way users create a `FixedSizeList` is by specifying the `DType`. If they pass us a python list of python lists (without a `DType`), then that should be turned into a `ListArray`. 
- There is an opportunity here to allow users to pass us a python list of python tuples, where we simply check that all of the tuples have the same size and give them back a `FixedSizeListArray`. Not sure if this is the best idea to be honest.
- This PR does not implement casting/conversion functionality between `FixedSizeList` and `List`. More specifically, we should be able to cast(?) a `FixedSizeList` to a `List`, but not vice versa. But what does "cast" even mean though? Something that is cheap? If we "cast" a `FixedSizeList` array to a `List` array, then we have the build an offsets buffer (and if it is a `ListView`, a sizes buffer as well). And then in the other direction, if all of the list scalars of a `ListArray` have the same size, then it is even cheaper to just discard the offsets buffer of the `ListArray` to convert it to a `FixedSizeListArray`.
- I would appreciate some scrutiny on the tests at the bottom of this PR. It is mostly generated by Claude, and to me (who has not interacted much with the Python side) it seems mostly right, but I could be very wrong.
- It seems like we are missing some `Decimal` APIs for Python?


vortex-python/python/vortex/__init__.py
```diff
@@ -10,11 +10,13 @@
     ChunkedArray,
     ConstantArray,
     DateTimePartsArray,
+    # DecimalArray # TODO(connor): Is this missing a `DecimalArray`?
```
robert3005, line 13: this is likely the case, it should be here

robert3005:
I don't understand how you can avoid the problem of converting nested builtin types if you were to have different scalar types. List of list is still some arbitrary decision based on what is the default conversion

vortex-python/python/vortex/__init__.py
danking, line 13: Yes, I think it was never added.

vortex-python/test/test_fixed_size_list.py
```diff
@@ -0,0 +1,124 @@
...
+    # Verify it's a ListScalar
+    assert isinstance(scalar, vortex.ListScalar)
+
+    # Access elements
+    assert scalar.element(0) == 1
+    assert scalar.element(1) == 2
+    assert scalar.element(2) == 3
+
+
+def test_round_trip_preservation():
```
danking: this isn't any more a round trip test than the other ones here, I think we should use the same naming scheme as above
```suggestion
def test_fixed_size_list_with_f64_elements():
```

danking (approved):
just a naming nit

robert3005 (approved):

PR #4619, opened 2025-09-12
Buffer more data in object store writer

AdamGS:
Mostly realized that this makes more sense:
1. Buffer is bigger than the chunk size, so we don't submit a part immediately when we pass the chunk size, which should give us some more concurrency.
2. We upload in buffer sizes that are aligned to 16MB, which the AWS perf guide seems to indicate is a good idea when reading it back.
3. Also fixes two typos.

### DIFF
--- a/vortex-io/src/object_store.rs
+++ b/vortex-io/src/object_store.rs
@@ -116,23 +116,24 @@ impl VortexReadAt for ObjectStoreReadAt {
     }
 }
 
-/// Adapter type to write data through a [`ObjectStore`] instace.
+/// Adapter type to write data through a [`ObjectStore`] instance.
 ///
-/// After writing, the caller must make sure to call `shutdonw`, in order to ensure the data is actually persisted.
+/// After writing, the caller must make sure to call `shutdown`, in order to ensure the data is actually persisted.
 pub struct ObjectStoreWriter {
     upload: Box<dyn MultipartUpload>,
     buffer: BytesMut,
     put_result: Option<PutResult>,
 }
 
-const CHUNKS_SIZE: usize = 25 * 1024 * 1024;
+const CHUNK_SIZE: usize = 16 * 1024 * 1024;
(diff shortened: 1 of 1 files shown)


vortex-io/src/object_store.rs
```diff
@@ -147,12 +148,15 @@ impl VortexWrite for ObjectStoreWriter {
         self.buffer.extend_from_slice(buffer.as_slice());
         let parts = FuturesUnordered::new();
 
-        // Split off chunks while buffer is larger than CHUNKS_SIZE
-        while self.buffer.len() > CHUNKS_SIZE {
-            let payload = self.buffer.split_to(CHUNKS_SIZE).freeze();
-            let part_fut = self.upload.put_part(PutPayload::from_bytes(payload));
+        // If the buffer is full
+        if self.buffer.len() > BUFFER_SIZE {
+            // Split off chunks while buffer is larger than CHUNKS_SIZE
```
gatesn, line 153: I don't think we should ever split our buffers. While I agree we should aim for 16MB, for multi-part upload it's nice to keep our existing chunk boundaries, because those are the boundaries we will ultimately read from.

AdamGS: you have to, chunks must all be of equal size and bigger than 5mb (other than the last one).

gatesn: For future readers, looks like that may just be an R2 restriction

robert3005 (approved):

vortex-io/src/object_store.rs
AdamGS: I'm sure S3 (or GCP?) used to have the same limitation, but I can't find it now, not sure if I imagined it or it just changed.

robert3005: I’m convinced that it was the case 5 years ago but not anymore

AdamGS:
I think this is a good change for now. Long term there are different things we can do here, but this improves the state of Vortex and DF writes now.

PR #4624, opened 2025-09-13
Convert handles to hold weak reference to their executor

gatesn:
This fixes a problem where we can spawn a task that holds onto handle (since it has a static lifetime, ergh) can keep a runtime alive even after the caller drops it. For futures that use channels, e.g. the FileRead I/O, this results in read futures that hang forever when the runtime is dropped (because the executor holds the pending recv end of the channel, and the executor is kept alive by the send end of the channel...)

We also impl BlockingRuntime for CurrentThreadRuntime as this will likely become the default runtime for language bindings to use.

### DIFF
--- a/vortex-io/src/runtime/blocking.rs
+++ b/vortex-io/src/runtime/blocking.rs
@@ -17,13 +17,15 @@ pub trait BlockingRuntime {
     ///
     /// The future is provided a [`Handle`] to the runtime so that it may spawn additional tasks
     /// to be executed concurrently.
-    fn block_on<Fut, R>(&self, f: Fut) -> R
+    fn block_on<F, Fut, R>(&self, f: F) -> R
     where
+        F: FnOnce(Handle) -> Fut,
         Fut: Future<Output = R>;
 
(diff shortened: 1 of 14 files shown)


vortex-io/src/runtime/blocking.rs
```diff
@@ -17,13 +17,15 @@ pub trait BlockingRuntime {
     ///
     /// The future is provided a [`Handle`] to the runtime so that it may spawn additional tasks
     /// to be executed concurrently.
-    fn block_on<Fut, R>(&self, f: Fut) -> R
+    fn block_on<F, Fut, R>(&self, f: F) -> R
```
gatesn, line 20: We change the API to take a handle so it's harder to write code that drops the runtime too early.

It would be impossible to write code that drops the runtime too early if we could embed a lifetime in the handle. But alas. My week long dive into Rust revealed this is simply not possible :'(

robert3005 (approved):
the statics inside functions are a bit weird style but they work

PR #4626, opened 2025-09-13
Move VortexRead over to new File I/O

gatesn:
This PR moves our I/O layer over from the old pre-fetching segment source to the new coalescing File I/O implementation on the Vortex runtime.

Note: this PR currently includes #4624 

## Breaking Changes

* `VortexOpenOptions` is no longer generic.
  * Fix: find/replace `VortexOpenOptions::file` -> `VortexOpenOptions::new`
* `VortexOpenOptions::in_memory().open(...)` is replaced by `VortexOpenOptions::new().open_buffer(...)`

## Performance

Generally this PR is about even, or slightly improved. There is some variance on some of the S3 benchmarks.

The big regressions are in DuckDB on S3, where our read path has a custom work-stealing queue that doesn't buffer enough of the scan to perform effective pre-fetching. DataFusion doesn't have this problem, so we see much less of an impact.

I want to follow up to this PR to switch over DuckDB to avoid this getting large, and accept the short-term regression.


vortex-file/src/segments/cache.rs
```diff
@@ -69,14 +70,14 @@ impl SegmentCache for MokaSegmentCache {
 
 /// Segment cache containing the initial read segments.
 pub(crate) struct InitialReadSegmentCache {
-    pub(crate) initial: DashMap<SegmentId, ByteBuffer>,
+    pub(crate) initial: RwLock<HashMap<SegmentId, ByteBuffer>>,
```
robert3005: why this change?

gatesn: There's no concurrent editing, we just have to have a shared handle that we first write to, then forever more we only read. So this has less overhead

vortex-io/src/tokio.rs
```diff
@@ -101,30 +98,29 @@ mod tests {
...
-            .await
-            .unwrap();
-
-        assert_eq!(first_half.as_ref(), "01234".as_bytes());
-        assert_eq!(second_half.as_ref(), "56789".as_bytes());
-    }
+    use crate::TokioFile;
+    //
+    // #[tokio::test]
+    // async fn test_shared_file() {
```
robert3005: delete this?

vortex-io/src/tokio.rs
```diff
@@ -48,35 +44,36 @@ impl Deref for TokioFile {
...
-    }
-
-    #[tracing::instrument(skip_all)]
-    async fn size(&self) -> io::Result<u64> {
-        self.metadata().map(|metadata| metadata.len())
-    }
-}
+//
+// #[async_trait]
+// impl VortexReadAt for TokioFile {
```
robert3005: delete this?

vortex-tui/src/browse/ui/layouts.rs
```diff
@@ -89,16 +90,17 @@ fn render_layout_header(cursor: &LayoutCursor, area: Rect, buf: &mut Buffer) {
...
     let row_count = app.cursor.layout().row_count();
     let reader = app
         .cursor
         .layout()
         .new_reader("".into(), app.vxf.segment_source())
         .vortex_expect("Failed to create reader");
 
-    let array = TOKIO_RUNTIME
-        .block_on(
+    // FIXME(ngates): our TUI app should never perform I/O in the render loop...
```
robert3005, line 101: we probably shouldn't merge a fixme?

vortex-file/src/open.rs
```diff
@@ -3,54 +3,111 @@
...
-    pub(crate) layout_registry: Arc<LayoutRegistry>,
+    layout_registry: Arc<LayoutRegistry>,
     /// An optional, externally provided, file size.
-    pub(crate) file_size: Option<u64>,
+    file_size: Option<u64>,
     /// An optional, externally provided, DType.
-    pub(crate) dtype: Option<DType>,
+    dtype: Option<DType>,
     /// An optional, externally provided, file layout.
     // TODO(ngates): add an optional DType so we only read the layout segment.
```
robert3005: I think this todo is stale

vortex-io/src/file/object_store.rs
```diff
@@ -22,7 +23,7 @@ const COALESCING_WINDOW: CoalesceWindow = CoalesceWindow {
 const CONCURRENCY: usize = 192; // Number of concurrent requests to allow.
 
 #[cfg(feature = "object_store")]
```
robert3005: You probably want to put this on the whole file. I don't believe this compiles without the feature enabled

robert3005 (commented):
Have some small comments

vortex-tui/src/browse/ui/layouts.rs
gatesn: I can file an issue, but it's more serious than a TODO and not something that I've changed the behavior of in this PR: https://github.com/vortex-data/vortex/issues/4641

robert3005: Thanks for filing an issue!

robert3005 (approved):

PR #4635, opened 2025-09-15
Remove pending segments in favor of weak futures

gatesn:
We made a change to remove shared futures from the layout readers, and this had little to no impact on performance @onursatici - the reason was our segment source was performing internal de-duplicating of segment requests.

It's nicer I think for the dedupe logic to happen in the scan / layout reader explicitly, rather than relying on different segment sources to implement this logic.

Prep work for https://github.com/vortex-data/vortex/pull/4626


onursatici (commented):
I think the segment cache should be responsible of deduplicating in flight requests, so we support all future segment requests, not only the ones from the flat layout

PR #4646, opened 2025-09-15
Move scan to use handles

gatesn:
Note, there are some regressions in DuckDB NVMe queries that use dynamic filtering. I would like to fix these in a FLUP since they may impact all scans (incl DataFusion).

FLUPs:
* Fix dynamic filtering in DuckDB scans
* Move each language API over to a CurrentThreadRuntime with a configurable number of background worker threads.


robert3005:
Just so I understand, the slowdown in duckdb comes from missing work stealing that we have to bring back in the different form under new runtime?

vortex-duckdb/src/lib.rs
```diff
@@ -3,11 +3,7 @@
 
 #![allow(clippy::missing_safety_doc)]
 use std::ffi::{CStr, c_char};
-use std::sync::LazyLock;
 
-// **WARNING begin this includes duckdb-rs, which is required to link in the symbol from libduckdb-sys.
-use tokio::runtime;
-use tokio::runtime::Runtime;
 // **WARNING end
```
onursatici, line 7:
```suggestion
```

onursatici (approved):

onursatici:
> Just so I understand, the slowdown in duckdb comes from missing work stealing that we have to bring back in the different form under new runtime?

Because the same iterator is shared among all worker threads that use the same current thread runtime, there is now one queue for tasks that all threads poll from

robert3005:
Looks like recent commits made the slowdown go away and we have some nice improvements as well!

PR #4648, opened 2025-09-15
Inline some dtype code

AdamGS:
Just gradually making more scalar/dtype code to inline.

Looking at the top-changed benchmarks and some of those that are just under the threshold, it seems like it helps bigger functions to get inlined, which I assume improves perf overall, even if sometimes in somewhat unexpected ways.


AdamGS:
Seems like registering the kernels shows up in some runs but not in others, and it accounts for significant latency.

robert3005 (approved):

AdamGS:
After warming up the compute kernels, this seems to make no difference.

PR #4650, opened 2025-09-15
Explicitly register wall kernels in compute vtables in benchmarks

AdamGS:
Noticed that some benchmarks had significant swings just because the ComputeFn lock would appear in some runs compared to others. This PR introduces a function which should force loading all existing kernels for the default functions.

The example I have right now is [this](https://codspeed.io/vortex-data/vortex/branches/adamg%2Finline-some-dtype-code?uri=encodings%2Ffastlanes%2Fbenches%2Fcompute_between.rs%3A%3Abitpack%3A%3Anew_bp_prim_test_between%255Bi16%252C%25202048%255D).

### DIFF
--- a/vortex-array/src/compute/mod.rs
+++ b/vortex-array/src/compute/mod.rs
@@ -72,6 +72,32 @@ pub struct ComputeFn {
     kernels: RwLock<Vec<ArcRef<dyn Kernel>>>,
 }
 
+/// Force all the default [`ComputeFn`] vtables to register all available compute kernels.
+///
+/// Mostly useful for small benchmarks where the overhead might cause noise depending on the order of benchmarks.
+pub fn warm_up_vtables() {
+    crate::arrow::warm_up_vtable();
+    #[allow(unused_qualifications)]
+    between::warm_up_vtable();
+    boolean::warm_up_vtable();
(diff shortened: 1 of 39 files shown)


vortex-array/src/compute/mod.rs
```diff
@@ -72,6 +72,32 @@ pub struct ComputeFn {
     kernels: RwLock<Vec<ArcRef<dyn Kernel>>>,
 }
 
+/// Force all the default [`ComputeFn`] vtables to register all available compute kernels.
+///
+/// Mostly useful for small benchmarks where the overhead might cause noise depending on the order of benchmarks.
+pub fn warm_up_vtables() {
+    crate::arrow::compute::warm_up_vtable();
```
robert3005: maybe reexport this function so the name here is nicer?

AdamGS:
The swing in benchmarks here is wild

vortex-array/src/compute/mod.rs
AdamGS: made it slightly nicer  - `crate::arrow::warm_up_vtable`, its a different module so it seems like a reasonable thing?

robert3005 (approved):

PR #4651, opened 2025-09-15
chore: tune datafusion in benchmarks

lwwmanning:


lwwmanning:
h/t @adriangb

adriangb (approved):
fwiw ✅

gatesn:
Are we worried that these seem to largely regress the Parquet performance..?

adriangb:
Yeah I think some investigation is needed

PR #4652, opened 2025-09-15
feat: FoRArray is sorted if its compressed form is sorted

danking:
I think this is true right? We previously had `[a, b, c, d, ...]` and we now have `[a - min, b - min, c - min, d - min, ...]`. If we know `a - min >= 0` (which is true by definition of min), then sortedness should be preserved because `a - k > b - k` implies `a > b`.

### DIFF
--- a/encodings/fastlanes/src/for/compute/is_sorted.rs
+++ b/encodings/fastlanes/src/for/compute/is_sorted.rs
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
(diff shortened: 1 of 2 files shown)


robert3005 (changes requested):
I think because of wrap around math this is not true i.e., i8::MAX - i8::MIN doesn't fit into i8 and gets put at the end

gatesn:
Wrap around continues to be a bad idea IMO. We should just check bounds

danking:
@robert3005 I think this is correct now. See comment for explanation.

robert3005 (approved):
I wonder why I didn’t convince myself that unsigned types were sorted before but you’re right. Previously we always casted the child to unsigned type so this would have been fine to have

PR #4654, opened 2025-09-15
feat: take for sequence array

danking:


encodings/sequence/src/compute/take.rs
```diff
@@ -0,0 +1,56 @@
...
+use vortex_array::{Array, ArrayRef, IntoArray, ToCanonical, register_kernel};
+use vortex_dtype::{NativePType, match_each_native_ptype};
+use vortex_error::VortexExpect;
+use vortex_error::VortexResult;
+use vortex_mask::Mask;
+
+use crate::{SequenceArray, SequenceVTable};
+
+impl TakeKernel for SequenceVTable {
+    #[allow(clippy::cast_possible_truncation)]
```
robert3005: I don't think you have a trunctaion here?

encodings/sequence/src/compute/take.rs
```diff
@@ -0,0 +1,56 @@
...
+
+fn take<T: NativePType, S: NativePType>(
+    mul: S,
+    base: S,
+    indices: &[T],
+    is_valid: Option<Mask>,
+) -> PrimitiveArray {
+    match is_valid {
+        Some(mask) => {
+            PrimitiveArray::from_option_iter(indices.iter().enumerate().map(|(mask_index, i)| {
```
robert3005: Would it be faster for this to just slam the validity on top separately?

encodings/sequence/src/compute/take.rs
danking: Removed. Bad copy paste.

encodings/sequence/src/compute/take.rs
danking: "all valid indices fit" is not true of invalid indices. I guess I could do:
```
PrimitiveArray::new(Buffer::from_iter(indices...).freeze(), validity)
```
and that might be faster. I'll switch to that.

danking: Done.

robert3005: hmm ok these are the take indices...

encodings/sequence/src/compute/take.rs
```diff
@@ -0,0 +1,60 @@
...
+
+fn take<T: NativePType, S: NativePType>(
+    mul: S,
+    base: S,
+    indices: &[T],
+    is_valid: Option<Mask>,
+) -> PrimitiveArray {
+    match is_valid {
+        Some(mask) => {
+            let buffer = Buffer::from_iter(indices.iter().enumerate().map(|(mask_index, i)| {
```
robert3005: since we can't use unstable apis, can you create the buffer first and then use extend_trusted since this iterator has a known size?

robert3005: ah no, just use `Buffer::from_trusted_len_iter` here

encodings/sequence/src/compute/take.rs
```diff
@@ -0,0 +1,60 @@
...
+                if mask.value(mask_index) {
+                    let i = <S as NumCast>::from::<T>(*i).vortex_expect("all valid indices fit");
+                    base + i * mul
+                } else {
+                    S::zero()
+                }
+            }));
+            PrimitiveArray::new(buffer, Validity::from_mask(mask, Nullability::Nullable))
+        }
+        None => PrimitiveArray::from_iter(indices.iter().map(|i| {
```
robert3005: same as above `Buffer::from_trusted_len_iter` should be preferred

encodings/sequence/src/compute/take.rs
danking: Sure! I had to impl TrustedLen for `Enumerate<I>` but I think that's safe.

robert3005 (approved):

PR #4656, opened 2025-09-15
buffered bytes counter in layout strategy

### LINKED ISSUE #4636: Expose internal buffer size through new push-based write API
With #4608, `bytes_written` was added to the push based writer so that callers could track the in-progress file size:
https://github.com/vortex-data/vortex/blob/4bce0a398088d16ab00fffb48ad70ea69a71681e/vortex-file/src/writer.rs#L312-L316

For some cases, e.g. when writing files with many columns (buffer sizes are defined per column), internal buffer sizes can be significant and are not reflected in `bytes_written`, these internal buffers are only flushed on `finish` and can sometimes be >100% of the last `bytes_written` call.

For example (writing file with 84 columns):
```
        let mut buf = Vec::with_capacity(80_000_000);
        let mut writer =
            VortexWriteOptions::default().writer(&mut buf, chunked_array.dtype().clone());
        for vortex_array in chunked_array.to_array_iterator() {
            writer.push(vortex_array?).await?;
            let bytes_written = writer.bytes_written();
            println!("bytes_written total: {} bytes", bytes_written);
        }
        println!("calling finish");
        let write_summary = writer.finish().await?;
        println!("Finished writing file of {} bytes", write_summary.size());
```

The last couple print lines of output are:
```
bytes_written total: 14405048 bytes
bytes_written total: 14405048 bytes
bytes_written total: 14620736 bytes
bytes_written total: 14841950 bytes
bytes_written total: 19811429 bytes
bytes_written total: 30070188 bytes
bytes_written total: 30070188 bytes
calling finish
Finished writing file of 76050884 bytes
```

cc @gatesn

onursatici:
Fixes https://github.com/vortex-data/vortex/issues/4636

LayoutStrategy to return the total buffered bytes as accurate as it can, but it will still undershoot to some extent, because there is no good way to get the actual size of all buffered data.

Currently a new LayoutStrategy instance is created per `VortexOpenOptions`, so returning the total buffered size across all streams being processed by a LayoutStrategy instance should be fine.

### DIFF
--- a/vortex-file/src/writer.rs
+++ b/vortex-file/src/writer.rs
@@ -222,12 +222,14 @@ impl VortexWriteOptions {
 
         let write = CountingVortexWrite::new(write);
         let bytes_written = write.counter();
+        let strategy = self.strategy.clone();
         let future = self.write(write, arrays).boxed_local().fuse();
 
         Writer {
             arrays: Some(arrays_send),
             future,
             bytes_written,
+            strategy,
         }
     }
 }
@@ -240,6 +242,8 @@ pub struct Writer<'w> {
     future: Fuse<LocalBoxFuture<'w, VortexResult<WriteSummary>>>,
     // The bytes written so far.
     bytes_written: Arc<AtomicU64>,
+    // The layout strategy that is being used for the write.
+    strategy: Arc<dyn LayoutStrategy>,
 }
 
 impl Writer<'_> {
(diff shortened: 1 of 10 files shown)


vortex-file/src/writer.rs
```diff
@@ -317,6 +321,17 @@ impl Writer<'_> {
...
     }
 
+    /// Returns the number of bytes currently buffered by the layout writers.
+    pub fn buffered_bytes(&self) -> u64 {
+        self.strategy.buffered_bytes()
+    }
+
+    /// Returns the total number of bytes (written + buffered).
+    /// This provides a more accurate estimate of the final file size.
+    pub fn total_bytes(&self) -> u64 {
```
gatesn: Remove this fn, I don't know that it reports useful _enough_ information that the caller shouldn't be forced to think about themselves.

gatesn (approved):

PR #4657, opened 2025-09-15
dict layout options to have u16 max size

onursatici:
After https://github.com/vortex-data/vortex/pull/4557 we can only have u16 codes on the dict writer, enforce this on the constraints so we don't panic


robert3005:
Wait, where in #4557 is that limit?

onursatici:
It is currently hardcoded here: https://github.com/vortex-data/vortex/blob/develop/vortex-layout/src/layouts/dict/writer.rs#L381-L384

gatesn (approved):

PR #4662, opened 2025-09-16
experiment: Pass FieldDType around

AdamGS:


vortex-dtype/src/struct_.rs
```diff
@@ -312,8 +894,8 @@ impl StructFields {
     }
 
     /// Returns an ordered iterator over the fields.
-    pub fn fields(&self) -> impl ExactSizeIterator<Item = DType> + '_ {
-        self.0.dtypes.iter().map(|dt| dt.value().vortex_unwrap())
+    pub fn fields(&self) -> impl ExactSizeIterator<Item = &FieldDType> + '_ {
+        self.0.dtypes.iter()
```
AdamGS, line 898: This is the big change here, instead of hiding this unwrap, its now everywhere.

robert3005: If any dtype could be a vieweddtype then I think you get this for free

AdamGS: you're saying - we can just merge `DType` and `FieldDType`?

robert3005: yes BUT now I realize I don't quite know how to make the api nice, since it would no longer be an enum :/

AdamGS: macros

AdamGS:
I think the real takeway here is to do something closer to #4678, so closing this.

PR #4666, opened 2025-09-16
coalescing metrics

onursatici:


vortex-io/src/runtime/handle.rs
```diff
@@ -136,13 +137,22 @@ impl Handle {
 
     /// Open a file for I/O on this runtime.
     pub fn open_read<S: IntoReadSource>(&self, source: S) -> VortexResult<FileRead> {
+        self.open_read_with_metrics(source, Default::default())
+    }
+
+    pub fn open_read_with_metrics<S: IntoReadSource>(
+        &self,
+        source: S,
+        metrics: VortexMetrics,
```
gatesn, line 142: Make this mandatory so it's harder to forget to pass in

gatesn (approved):

PR #4670, opened 2025-09-16
chore: provide shim from VortexReadAt to ReadSource

danking:


gatesn:
I think if you swap VortexReadAt for AsyncRead, then this makes a lot of sense

PR #4673, opened 2025-09-17
Remove local path optimization for object store

### LINKED ISSUE #4672: Make hardcoded local path prefix `/` configurable
The `VortexOpenOptions::open_object_store` method contains a hardcoded local path prefix `/` that should be made configurable to improve flexibility.

```rust
let local_path = Path::new("/").join(path);
if local_path.exists() {
    // Local disk is too fast to justify prefetching.
    self.open(local_path).await
} else {
    self.open_read_at(ObjectStoreReadAt::new(
        object_store.clone(),
        path.into(),
        None,
    ))
    .await
}
```

Proposed Solution: add a `local_path_prefix` field to `VortexFileCache` and thread it through to `VortexOpenOptions`

gatesn:
Closes #4672

This optimization is a bit weird. If the user explicitly asks for object store, and has configured it with telemetry etc, then it's odd for us to then ignore that request and use our own I/O layer.

Also assuming an absolute path from root was too naive.

Users are encouraged to either `open_object_store`, or `open(&Path)`. If we had our own registry of object stores, we could also support `impl IntoReadSource for Url`

### DIFF
--- a/vortex-file/src/open.rs
+++ b/vortex-file/src/open.rs
@@ -295,24 +295,12 @@ impl VortexOpenOptions {
         object_store: &Arc<dyn object_store::ObjectStore>,
         path: &str,
     ) -> VortexResult<VortexFile> {
-        use std::path::Path;
-
         use vortex_io::file::object_store::ObjectStoreReadSource;
 
-        // If the file is local, we much prefer to use TokioFile since object store re-opens the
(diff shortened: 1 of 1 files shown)


AdamGS (approved):
🥳🥳

xudong963:
Thank you @gatesn

PR #4678, opened 2025-09-17
experiment: Try a different FB compiler, now with both owned and borrowed variants

AdamGS:
stashing it here for a bit.

This PR is trying to use the [planus](https://crates.io/crates/planus) flatbuffer compiler, which generates both owned and ref variants of types.

~There are some incompatibilities in the FB files themselves between it and flatc, I'm not sure why, but for now I just got it to the point it actually generates code.~ Seems like this a known issue which hopefully will be fixed soon - https://github.com/planus-org/planus/pull/314


robert3005:
They validate fields on demand instead of upfront which would be very useful though a small gain for now. @gatesn

vortex-flatbuffers/flatbuffers/vortex-dtype/dtype.fbs
```diff
@@ -64,16 +64,16 @@ table Extension {
...
-    Bool = 2,
-    Primitive = 3,
-    Decimal = 4,
-    Utf8 = 5,
-    Binary = 6,
-    Struct_ = 7,
-    List = 8,
-    Extension = 9,
-    FixedSizeList = 10, // This is after `Extension` for backwards compatibility.
+    Null,
```
AdamGS, line 67: We'll probably have to upstream support for this

PR #4679, opened 2025-09-17
experiment: another scalars thing

AdamGS:
I might've been too clever for our own good by trying to get scalars to bee smaller by adding more memory indirection, I think this might be better.


AdamGS:
seems to be really worse in a bunch of cases

PR #4681, opened 2025-09-17
fix: duckdb dict ownership in `export`

0ax1:
The issue with `slice_to_dictionary` is that it neither properly initializes the dict nor holds a strong reference to it.

### DIFF
--- a/vortex-duckdb/cpp/include/duckdb_vx/vector.h
+++ b/vortex-duckdb/cpp/include/duckdb_vx/vector.h
@@ -18,6 +18,11 @@ void duckdb_vx_set_dictionary_vector_id(duckdb_vector dict, const char *id, unsi
 
 void duckdb_vx_set_dictionary_vector_length(duckdb_vector dict, unsigned int len);
 
+/// Turn vector into a dictionary vector. In contrast to `duckdb_vx_vector_slice_to_dictionary` this
+/// call creates a dictionary that holds a strong reference to its children.
+void duckdb_vx_vector_dictionary(duckdb_vector ffi_vector, duckdb_vector ffi_dict, idx_t dictionary_size,
+                                 duckdb_selection_vector ffi_sel_vec, idx_t count);
+
 // Add the buffer to the string vector (basically, keep it alive as long as the vector).
 void duckdb_vx_string_vector_add_buffer(duckdb_vector ffi_vector, duckdb_vx_data buffer);
 
--- a/vortex-duckdb/cpp/include/duckdb_vx/vector.h
+++ b/vortex-duckdb/cpp/include/duckdb_vx/vector.h
@@ -14,14 +14,14 @@ extern "C" {
 void duckdb_vx_vector_slice_to_dictionary(duckdb_vector ffi_vector, duckdb_selection_vector selection_vector,
                                           idx_t selection_vector_length);
 
-void duckdb_vx_set_dictionary_vector_id(duckdb_vector dict, const char *id, unsigned int id_len);
-
-void duckdb_vx_set_dictionary_vector_length(duckdb_vector dict, unsigned int len);
-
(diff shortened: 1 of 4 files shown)


vortex-duckdb/src/duckdb/vector.rs
```diff
@@ -127,6 +127,24 @@ impl Vector {
         unsafe { cpp::duckdb_vx_string_vector_add_buffer(self.as_ptr(), data.into_ptr()) }
     }
 
+    pub fn dictionary(
```
connortsui20: Nit: Could you move this up to where the other dictionary functions are? It just makes it slightly easier to read.

On a slightly different note (not part of this PR), what exactly do these functions do? Are these sort of casts of the given vector to the specified type? In that case I don't really understand what `slice_to_dictionary` does because it takes a `&mut self` (the comment "Slice the vector to a new dictionary vector" is also quite confusing to me, I don't really know what this means).

vortex-duckdb/cpp/include/duckdb_vx/vector.h
```diff
@@ -18,6 +18,11 @@ void duckdb_vx_set_dictionary_vector_id(duckdb_vector dict, const char *id, unsi
 
 void duckdb_vx_set_dictionary_vector_length(duckdb_vector dict, unsigned int len);
 
+/// Turn vector into a dictionary vector. In contrast to `duckdb_vx_vector_slice_to_dictionary` this
+/// call creates a dictionary that holds a strong reference to its children.
+void duckdb_vx_vector_dictionary(duckdb_vector ffi_vector, duckdb_vector ffi_dict, idx_t dictionary_size,
+                                 duckdb_selection_vector ffi_sel_vec, idx_t count);
+
```
connortsui20, line 21: Nit: move this up to under `duckdb_vx_vector_slice_to_dictionary` (to mimic the `vector.cpp` file?)

vortex-duckdb/src/exporter/dict.rs
```diff
@@ -80,10 +80,7 @@ pub(crate) fn new_exporter(
 
 impl<I: NativePType + AsPrimitive<u32>> ColumnExporter for DictExporter<I> {
     fn export(&self, offset: usize, len: usize, vector: &mut Vector) -> VortexResult<()> {
-        // Copy across the dictionary values.
-        vector.reference(&self.values_vector.lock());
```
connortsui20, line 84: This is a somewhat interesting choice of a word... I would not have expected this to copy anything more than a pointer (but maybe this is because I'm Rust-brained?). The point here is that reading this code does not make it obvious that there was no strong count increment and so I understand why this bug was here.

Basically I wonder if it might be good to rename the method `reference`. Is that against our philosophy for how we interop with C++?

connortsui20 (changes requested):
I'm also going to rebase my PR on this and see if the other bug is fixed

connortsui20 (approved):

vortex-duckdb/src/exporter/dict.rs
0ax1: That comment was wrong. This leads to `vector` holding a strong ref to `values_vector`.

vortex-duckdb/src/duckdb/vector.rs
0ax1: > Are these sort of casts of the given vector to the specified type? 

Yes, all vector types are based on the same C++ type and then interpreted in different ways.

see: 
- https://github.com/duckdb/duckdb/blob/main/src/common/types/vector.cpp#L289
- https://github.com/duckdb/duckdb/blob/main/src/common/types/vector.cpp#L311C4-L311C47

0ax1:
@connortsui20 Thanks for having a peek! Bundled the dict funcs by now.

PR #4690, opened 2025-09-18
Rename functions in SelectExpr to make API clearer

AdamGS:
Unfortunately a minor breaking change to the API, but the amount of `fields()` functions we have is just too much.

### DIFF
--- a/vortex-expr/src/exprs/select.rs
+++ b/vortex-expr/src/exprs/select.rs
@@ -206,11 +206,11 @@ impl SelectExpr {
     /// For example:
     /// ```rust
     /// # use vortex_expr::root;
-    /// # use vortex_expr::{SelectExpr, SelectField};
+    /// # use vortex_expr::{FieldSelection, SelectExpr};
     /// # use vortex_dtype::FieldNames;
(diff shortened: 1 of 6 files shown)


vortex-expr/src/exprs/select.rs
```diff
@@ -52,13 +52,13 @@ impl VTable for SelectVTable {
 
     fn metadata(expr: &Self::Expr) -> Option<Self::Metadata> {
         let names = expr
-            .fields()
-            .fields()
```
AdamGS, line 56: Mainly to prevent this, I think this is much clearer now.

vortex-scalar/src/struct_.rs
```diff
@@ -164,7 +164,7 @@ impl<'a> StructScalar<'a> {
     }
 
     /// Returns the fields of the struct scalar, or None if the scalar is null.
-    pub fn fields(&self) -> Option<impl Iterator<Item = Scalar>> {
+    pub fn fields(&self) -> Option<impl ExactSizeIterator<Item = Scalar>> {
```
AdamGS, line 167: just had this and figured its not worth its own PR

robert3005 (approved):

PR #4691, opened 2025-09-18
docs: clarify duckdb dictionary and dictionary slice

0ax1:
I read a bit more in the DuckDB source code concluding that both hold a strong ref to all memory that they use.

### DIFF
--- a/vortex-duckdb/src/duckdb/vector.rs
+++ b/vortex-duckdb/src/duckdb/vector.rs
@@ -53,6 +53,8 @@ impl Vector {
 
     /// Slice the vector to a new dictionary vector, using the current vector's values and
     /// the provided selection vector.
+    ///
+    /// A dictionary slice holds a strong reference to all memory it uses.
     pub fn slice_to_dictionary(&mut self, sel_vec: SelectionVector, sel_vec_length: usize) {
         unsafe {
             cpp::duckdb_vx_vector_slice_to_dictionary(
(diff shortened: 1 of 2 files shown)


vortex-duckdb/src/duckdb/vector.rs
```diff
@@ -562,4 +568,48 @@ mod tests {
...
+        let slice = unsafe { vector.as_slice_mut::<i32>(len) };
+        slice[0] = 10;
+        slice[1] = 20;
+
+        let mut sel_vec = SelectionVector::with_capacity(2);
+        let sel_slice = unsafe { sel_vec.as_slice_mut(2) };
+        sel_slice[0] = 1;
+        sel_slice[1] = 0;
+
+        vector.slice_to_dictionary(sel_vec, 2);
```
connortsui20, line 587: What would happen if the selection vec was also 010 like underneath? Still trying to wrap my head around the difference between `dictionary` and `slice_to_dictionary`...

connortsui20 (approved):
LGTM!

vortex-duckdb/src/duckdb/vector.rs
0ax1: They're the same except for the initialization `Dictionary` does. Just stepped through the code in the debugger to double check.

```
void Vector::Dictionary(Vector &dict, idx_t dictionary_size, const SelectionVector &sel, idx_t count) {
	if (DictionaryVector::CanCacheHashes(dict.GetType()) && !dict.cached_hashes) {
		// Create an empty hash vector for this dictionary, potentially to be used for caching hashes later
		// This needs to happen here, as we need to add "cached_hashes" to the original input Vector "dict"
		dict.cached_hashes = make_buffer<VectorChildBuffer>(Vector(LogicalType::HASH, false, false, 0));
	}
	Reference(dict);
	Dictionary(dictionary_size, sel, count);
}
```

calls into 
```
void Vector::Dictionary(idx_t dictionary_size, const SelectionVector &sel, idx_t count) {
	Slice(sel, count);
	if (GetVectorType() == VectorType::DICTIONARY_VECTOR) {
		buffer->Cast<DictionaryBuffer>().SetDictionarySize(dictionary_size);
	}
}
```

which in turn calls
```
void Vector::Slice(const SelectionVector &sel, idx_t count) {
    //...
```

PR #4695, opened 2025-09-18
canonical builders to deduplicate varbinview buffers

onursatici:


vortex-array/src/builders/mod.rs
```diff
@@ -239,8 +239,11 @@ pub fn builder_with_capacity(dtype: &DType, capacity: usize) -> Box<dyn ArrayBui
                 ))
             })
         }
-        DType::Utf8(n) => Box::new(VarBinViewBuilder::with_capacity(DType::Utf8(*n), capacity)),
-        DType::Binary(n) => Box::new(VarBinViewBuilder::with_capacity(
+        DType::Utf8(n) => Box::new(VarBinViewBuilder::with_buffer_deduplication(
```
joseph-isaacs, line 242: Mind adding a doc comment to the ctor

gatesn: I think this is used in too many internal code paths right now to be the default?

onursatici:
zip_impl default is now deduplicating by default. I think we might want something like https://github.com/vortex-data/vortex/pull/4827 instead of making deduplication by default for now

PR #4696, opened 2025-09-18
Update pyproject.toml to use dependency-groups

AdamGS:
Its been stable for almost a year, and new `uv` versions complain about it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,12 +13,10 @@ build-backend = "hatchling.build"
 [tool.hatch.build.targets.wheel]
 packages = ["dummy"] # Required for workspace project
 
-[tool.uv]
-managed = true
-required-version = ">=0.8.0"
+[dependency-groups]
 # Currently, all dev dependencies live in the root since uv doesn't have transitive dev dependencies.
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -13,12 +13,10 @@ build-backend = "hatchling.build"
 [tool.hatch.build.targets.wheel]
 packages = ["dummy"] # Required for workspace project
 
-[tool.uv]
-managed = true
-required-version = ">=0.8.0"
+[dependency-groups]
 # Currently, all dev dependencies live in the root since uv doesn't have transitive dev dependencies.
 #  See: https://github.com/astral-sh/uv/issues/7541
```
danking, line 18: Relatedly, it sounds like we can push dev dependencies into their appropriate child projects and just use `uv sync --all-packages`.

https://github.com/astral-sh/uv/issues/7541#issuecomment-2487051795

danking (approved):

pyproject.toml
AdamGS: the state of python packaging continues to confuse me

PR #4704, opened 2025-09-19
repartition strategy to divide large chunks

onursatici:
This is not ideal because splitting is based on an heuristic, and we do compact the arrays twice now, first to get a better nbytes estimation and second before compression. 

Actual solution to this would to have a canonical array builder with max size constraints, so we can canonicalise an array with fixed sized blocks. We can then use this to split the array into already compacted and fixed size chunks and compress them


robert3005:
Any reason not to implement the full thing with size limited builders?

onursatici:
Yea I thought this might be good enough and less intrusive, but I am doing the limited builders now

PR #4705, opened 2025-09-19
feat: allow for linking DuckDB debug build

0ax1:


connortsui20 (approved):
It looks fine to me, but I think it would be best if someone more familiar with DuckDB builds looks at this? Also don't really have time to test this locally myself

connortsui20:
oh it merged haha

robert3005:
I think we want to run tests with debug build so we get all the assertions

0ax1:
> I think we want to run tests with debug build so we get all the assertions

Yes.

Issue #4699, opened 2025-09-18
Tracking Issue: Canonicalize `ListView` over `List`


connortsui20:
# Tracking Issue: Canonicalize `ListView` over `List`

This is a tracking issue for 1) adding a `ListView` encoding to Vortex and 2) making the new `ListView` encoding the canonical encoding for the `DType::List` logical type.

One of the limitations / tradeoffs that the current `List` encoding has is that it does not store a `sizes` buffer alongside the `offsets` buffer, where the size of each list scalar is implicitly calculated via the equation `offsets[i + 1] - offsets[i]`.

The issue with this is that we cannot effectively vectorize execution over the current `List` encoding because of the inherent sequential dependencies of the list offsets (next value depends on the previous value). This makes the `List` encoding both less random-access/slice friendly as well as less SIMD friendly, on top of independent data generally providing better branch/loop prediction.

### Public API

```rust
struct ListViewArray { ... }

struct ListViewBuilder<OffsetType, SizeType> 
where
    OffsetType: OffsetPType,
    SizeType: OffsetPType,
{ ... }

#[derive(Debug, Clone)]
pub enum Canonical {
<-- snip -->
    // List(ListArray),
    List(ListViewArray),
<-- snip -->
}
```

### Child Arrays

There are two possible ways to store the `offsets` and `sizes` data. The first way is to store 2 separate child arrays, and the second is to store a single array where each "scalar" is the pair of the offset and the size for a list scalar next to each other.

We have chosen the first layout. The main reason is that it allows better compressibility. It also enables the next section.

### Generic `offsets` and `sizes`

There is technically no reason that the `offsets` array and the `sizes` array cannot be different integer types. As long as the `sizes` integer type is smaller than the `offsets` type (since a size bigger than the largest offset would cause overflow), we can have any combination of `offsets` and `sizes` child arrays.

This is notably different from the Rust Arrow implementation, in which the `GenericListViewBuilder` is only generic over a single `OffsetSize` parameter.

### Binary Compatability

`ListViewArray` will be compatible with Arrow's `ListViewArray` when both the `offsets` and `sizes` are `i32`, and compatible with `LargeListViewArray` when both child arrays are `i64`.

### Prior Art

- Arrow `ListView` overview: https://arrow.apache.org/docs/format/Intro.html#list-view
- Arrow `ListView` layout: https://arrow.apache.org/docs/format/Columnar.html#listview-layout
[...]

PR #4708, opened 2025-09-19
Feature: `ListViewArray`

connortsui20:

Adds a new `ListViewArray` encoding.

The compute functions and the `ListViewBuilder` will come in a future PR.


vortex-array/src/arrays/listview/compact.rs
```diff
@@ -0,0 +1,142 @@
...
+            // even though offsets are sorted. This is because a list with a smaller offset could
+            // extend further than a list with a larger offset if it has a larger size.
+            // Example: offset[0]=0, size[0]=10 (ends at 10) vs offset[1]=5, size[1]=2 (ends at 7).
+            let mut max_end = 0usize;
+            for i in 0..self.len() {
+                let offset = self.offset_at(i);
+                let size = self.size_at(i);
+                let end = offset + size;
+                max_end = max_end.max(end);
+            }
```
connortsui20: We could also do `max(add(offsets, sizes))`. However, I don't actually know if that is more or less expensive in the general case...

vortex-array/src/arrays/listview/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+        if list.is_empty() {
+            let empty_offsets = Canonical::empty(list.offsets().dtype()).into_array();
+            let empty_sizes = Canonical::empty(list.offsets().dtype()).into_array();
+            return Self::try_new(
+                list.elements().clone(),
+                empty_offsets,
+                empty_sizes,
+                list.validity().clone(),
+            );
+        }
```
connortsui20: I might be able to remove this once I migrate `Canonical::List` to use `ListViewArray`?

vortex-array/src/arrays/listview/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+            size_ptype,
+            offset_ptype
+        );
+
+        // Validate all offsets + sizes are within bounds.
+        let elements_len = elements.len() as u64;
+
+        // Check each offset[i] + size[i] <= elements.len().
+        // Use scalar_at to handle both primitive and compressed arrays.
+        match_each_integer_ptype!(offset_ptype, |O| {
```
robert3005, line 224: I would convert both offsets and lengths `to_primitive()` here. This is a trivial validation and scalar_at is likely adding A LOT of overhead (a lot more than to_primitive())

vortex-array/src/arrays/listview/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    /// Returns the elements array.
+    pub fn elements(&self) -> &ArrayRef {
+        &self.elements
+    }
+
+    /// Create a `ListViewArray` from a [`ListArray`](crate::arrays::ListArray) by computing sizes
+    /// from offsets.
+    pub fn from_list_array(list: ListArray) -> VortexResult<Self> {
```
robert3005: This should be the cast compute function

vortex-array/src/arrays/listview/reset.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Create a new version of this array by adjusting `offsets` to start at 0 and removing any
+    /// `elements` not referenced by the `offsets` + `sizes`.
+    ///
+    /// Unlike [`ListArray`], the offsets of a `ListViewArray` are allowed to be out-of-order. This
+    /// means that we cannot simply subtract the first offset from the rest of the offsets (because
+    /// we don't know where the used elements actually are, and we may reorder some scalars).
+    ///
+    /// [`ListArray`]: crate::arrays::ListArray
+    #[allow(clippy::cognitive_complexity)]
+    pub fn reset_offsets(&self) -> VortexResult<Self> {
```
robert3005: I think `reset_offsets` made sense for ListArray since there the change is to have offsets start at 0. In this case you are removing values that are not reference by any offset+size pair. I would call this `compact_values`

vortex-array/src/arrays/listview/compact.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    let mut current_offset = 0usize;
+
+    for i in 0..listview.len() {
+        let offset = listview.offset_at(i);
+        let size = listview.size_at(i);
+
+        // Append the elements for this list to the builder.
+        for j in 0..size {
+            elements_builder.append_scalar(&listview.elements.scalar_at(offset + j))?;
```
robert3005: I need to add api to remove usage of scalar at here

vortex-array/src/arrays/listview/vtable/operations.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::arrays::{ListViewArray, ListViewVTable};
+use crate::vtable::OperationsVTable;
+use crate::{ArrayRef, IntoArray};
+
+impl OperationsVTable<ListViewVTable> for ListViewVTable {
+    fn slice(array: &ListViewArray, range: Range<usize>) -> ArrayRef {
+        let start = range.start;
+        let end = range.end;
+
+        debug_assert!(
```
robert3005: slice already asserts bounds, no need to do it again

robert3005 (commented):
Haven't looked at the tests yet, left some minor style comments

connortsui20:
I just realized my optimization is wrong. Even if the offsets are sorted, it could be "sparse". So I think I'm just going to refactor that function entirely. In fact, I'm probably going to remove it from this PR and do a followup PR

gatesn:
Just confirming this is binary compatible with Arrow's ListView

vortex-array/src/arrays/listview/mod.rs
connortsui20: Do you mean that this should be a special case in the cast compute fn?

vortex-array/src/arrays/listview/compact.rs
connortsui20: Do you mean appending directly to the underlying memory here? I'm actually not too sure how else you would do this when everything is type-erased

vortex-array/src/arrays/listview/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+            elements,
+            offsets,
+            sizes,
+            validity,
+            stats_set: Default::default(),
+        }
+    }
+
+    /// Validates the components that would be used to create a [`ListViewArray`].
+    #[allow(clippy::cognitive_complexity)]
```
joseph-isaacs: Can we have offset or size be a `u64` would that work is usize == 32 bits?

connortsui20: Right now I just check that the `byte_width` of the size of the `sizes` array is less than the `offsets`, so I don't think that should matter?

connortsui20:
> Just confirming this is binary compatible with Arrow's ListView

I'm pretty sure it is? At least when everything is in canonical form

vortex-array/src/arrays/listview/rebuild.rs
connortsui20, line 1: I will implement this in a followup PR. Note that I'm already thinking about several permutations of strategies (optimistic trim, deduplication, running an expensive shuffle/compaction, fixing overlaps, ensuring nulls don't take up any space, etc). Even if I don't implement all of these I still think it is worthwhile having control over this.

vortex-array/src/arrays/listview/mod.rs
robert3005: I think this being an instance function is weird, if you want I would make it a standalone top level function. If this is a ListArray then I don't think this should error? Then we can call this from the cast compute

vortex-array/src/arrays/listview/compact.rs
robert3005: Builders aren't type erased though. I have an idea that I need to prototype. I think in practice you always know the concrete type when you use it

robert3005:
Yes, this is compatible with Arrow ListViewArray when offsets and lengths are u32/i32 or u64/i64

vortex-array/src/arrays/listview/rebuild.rs
connortsui20: Discussed on slack: instead of having this logic here, it should be in the `ListViewBuilder`, and we will require that any time you want to restructure the array you have to pass it through the builder.

vortex-array/src/arrays/listview/mod.rs
robert3005: technically yes but in practice the values still have to be constrained to u32. In practice this means that any in memory array cannot have more than usize elements. If you have an array with u64 offsets or lengths then the io reader has to slice them to usize chunks

connortsui20: Doesn't `ListArray` have the same problem then? Actual I'm not super sure what the issue is here

robert3005: There's no issue, at least nothing you can do at this point

vortex-array/src/arrays/listview/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// Returns the elements array.
+    pub fn elements(&self) -> &ArrayRef {
+        &self.elements
+    }
+}
+
+/// Helper function to validate `offsets` and `sizes` with specific types.
+fn validate_offsets_and_sizes<O, S>(
+    offsets: &dyn Array,
+    sizes: &dyn Array,
```
robert3005: I think this works out nicer if you take `&[O]` and `&[S]` here and do the conversion outside this function

robert3005 (approved):
Only a style comment

PR #4711, opened 2025-09-19
chore: run vortex-duckdb tests in ci

0ax1:


0ax1:
I'll check in a follow up whether I can get the tests to run with ASAN.

.github/workflows/ci.yml
```diff
@@ -378,6 +378,31 @@ jobs:
             --target x86_64-unknown-linux-gnu \
             --verbose
 
+  vortex-duckdb-test:
+    name: "Rust tests (vortex-duckdb)"
+    timeout-minutes: 120
+    runs-on:
+      - runs-on=${{ github.run_id }}
+      - family=c6id.8xlarge
```
0ax1: Picked a beefy machine such that the build task matches the duration of others, not introducing any additional CI wait time.

.github/workflows/ci.yml
```diff
@@ -378,6 +378,31 @@ jobs:
             --target x86_64-unknown-linux-gnu \
             --verbose
 
+  vortex-duckdb-test:
```
robert3005, line 381: Can we run this as part of regular tests?

0ax1: The main test suite runs with ASAN right? Then no. I did bang my head against that ASAN wall for an hour today before without success.

0ax1: We can surely make this work on CI, but it's a lot less trivial to set up than I would have thought. Maybe we can do this in a follow up.

0ax1: Feel free to push on this PR though if you have an idea how to make this work.

.github/workflows/ci.yml
robert3005: I would stick to the regular machines, unless the perf difference is really big? I am always suspicious of this one of setup for ci

.github/workflows/ci.yml
```diff
@@ -378,6 +378,31 @@ jobs:
...
+          sccache: s3
+      - uses: actions/checkout@v5
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ninja-build cmake
+      - uses: ./.github/actions/setup-rust
+      - name: Vortex DuckDB tests
+        run: |
+          cargo test -p vortex-duckdb -- --skip test_vortex_scan_ultra_deep_nesting
```
robert3005: you will get a lot of perf improvement if you use nextest instead of regular test runner

0ax1: yah, that's a good point

0ax1: changed

.github/workflows/ci.yml
0ax1: We have to build DuckDB, so this build task would take significantly longer than others on `family=m7i+m7i-flex+m7a`.

0ax1: Giving it a try just now.

0ax1: Seems totally fine on m7. Maybe it was the nextest change making it faster. => changed

robert3005 (approved):

PR #4712, opened 2025-09-19
Feature: `ListView` Compute Functions

connortsui20:

Adds compute functions to the `ListView` encoding.

Additionally refactors the the `list` `filter` implementation to be a bit more succinct.

Also adds tests, and refactors some of the fixed-size list tests to use more conformance tests.

Unresolved Questions:

~~How much do we care about carrying around extra data? Both `filter` and `cast` can be implemented by pushing down the operation directly to the `offsets` and `sizes` child arrays. This can leave quite a lot of data in the `elements` array.~~

~~There are many things we can do here. In the second commit I trim the ends of the `elements` array using the min and max indexes into the `elements` (derived from computing `min/max` of `offsets` and `offsets + sizes`). This could be more expensive than we want.~~

~~What we could do instead is simply filter/take `offsets`/`sizes` and then allow the caller to compact / "clean up" by passing the final array back through a builder (which can have different strategies to rebuild the array). I think I am in favor of this (though I'm leaving the code in the commit history for now so people can see it).~~


vortex-array/src/arrays/listview/tests/cast.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn to_listview(&self) -> ListViewArray;
+}
+
+impl ToListView for ArrayRef {
+    fn to_listview(&self) -> ListViewArray {
+        self.as_opt::<ListViewVTable>()
+            .unwrap_or_else(|| vortex_panic!("Expected ListViewArray"))
+            .clone()
+    }
+}
```
connortsui20: Maybe can replace this once the canonical migration is done?

connortsui20:
@robert3005 So I made the decision to essentially do no rebuilding of the underlying `elements` array since we are going to have all of the rebuilding logic in the builders regardless. You can look at commit 69dd60b35411cdb5171b92c6608ecd208a1861f6 if you want to see the implementation where I just slice the ends, imo this is kind of pointless and might add a lot of overhead.

robert3005:
That's correct, the idea is that we perform the minimum amount of work required and if the user wants to prune invalid values we give them a function to do so.

vortex-array/src/arrays/fixed_size_list/compute/filter.rs
```diff
@@ -14,7 +14,7 @@ use crate::{ArrayRef, IntoArray, register_kernel};
 /// When the mask density is below this threshold, we use indices. Otherwise, we use slices.
 ///
 /// Note that this is somewhat arbitrarily chosen...
-const FSL_MASK_EXPANSION_DENSITY_THRESHOLD: f64 = 0.1;
+const MASK_EXPANSION_DENSITY_THRESHOLD: f64 = 0.05;
```
joseph-isaacs, line 17: OOI how did you work this out?

vortex-array/src/arrays/listview/compute/cast.rs
```diff
@@ -0,0 +1,41 @@
...
+        let Some(target_element_type) = dtype.as_list_element_opt() else {
+            return Ok(None);
+        };
+
+        // Cast the elements to the target element type.
+        let new_elements = compute::cast(array.elements(), target_element_type)?;
+        let validity = array
+            .validity()
+            .clone()
+            .cast_nullability(dtype.nullability(), array.len())?;
```
joseph-isaacs, line 24: you only need to do this if the nullability changes?

vortex-array/src/arrays/listview/compute/filter.rs
```diff
@@ -0,0 +1,52 @@
...
+/// This implementation is deliberately simple and read-optimized. We just filter the `offsets` and
+/// `sizes` arrays and reuse the original `elements` array. This works because `ListView` (unlike
+/// `List`) allows non-contiguous and out-of-order lists.
+///
+/// We don't slice the `elements` array because it would require computing min/max offsets and
+/// adjusting all offsets accordingly, which is not really worth the small potential memory we would
+/// be able to get back.
+///
+/// The trade-off is that we may keep unreferenced elements in memory, but this is acceptable since
+/// we're optimizing for read performance and the data isn't being copied.
```
joseph-isaacs, line 23: Do we have an `compact` function for this, which would compact the elements array?

vortex-array/src/arrays/listview/compute/is_constant.rs
```diff
@@ -0,0 +1,48 @@
...
+        opts: &IsConstantOpts,
+    ) -> VortexResult<Option<bool>> {
+        // At this point, we're guaranteed:
+        // - Array has at least 2 elements
+        // - All elements are valid (no nulls)
+
+        // First check if all list sizes are constant.
+        if !array.sizes.is_constant() {
+            return Ok(Some(false));
+        }
```
joseph-isaacs: this is not constant time is it? Otherwise move below next check

joseph-isaacs: I also wonder if you want to try to compare the first two elements before checking all sizes?

vortex-array/src/arrays/fixed_size_list/compute/filter.rs
connortsui20: I didn't...

Though I will note that if we have a mask with 10% density, the slices might still be large so I arbitrarily chose 5% instead

vortex-array/src/arrays/listview/compute/filter.rs
connortsui20: That will hopefully come in the next PR that implements builders. The goal would be that we can pass the array back through a builder (or some other structure) that can reorganize the array however we want. For example, we might not want any overlaps, or we might want in-order offsets, or we might want the most compact representation possible, etc.

vortex-array/src/arrays/listview/compute/cast.rs
connortsui20: To be honest I just copied and pasted that from the `list` implementation.

Given that the clone and the cast is cheap in the case where the nullability is the same, I think it should be fine?

vortex-array/src/arrays/listview/compute/is_constant.rs
connortsui20: Are you saying the algorithm should be:

- Get the first two list scalars (`scalar_at`)
- Check the first element of each of the scalars to see if they are the same
- Then check if all the sizes are constant
- Check the remaining scalars

I'm not sure this is faster... Since `sizes` is always going to have a primitive type I think the `is_constant` check might even be faster than decompressing the scalars of the first two lists.

vortex-array/src/arrays/listview/compute/cast.rs
joseph-isaacs: You are correct, I thought this was validity_mask which is slow to compte

vortex-array/src/arrays/listview/compute/is_constant.rs
joseph-isaacs: hmm... maybe. You definitely need to use `is_constant_opts`, not is_constant, but this might be a problem in many places

connortsui20: ~~I guess I might as well go and fix all of those now~~

Actually nevermind this is only relevant for nested types

robert3005 (commented):
Most compute functions have conformance tests. Some of the test logic could be replaced with applying that conformance test logic. They're in vortex_array::compute::conformance::...

connortsui20:
I just saw that there is a `list_contains` compute kernel that for some reason is not under `list` and is only under `compute`. That seems sufficiently complicated that I want to put that in a different PR

robert3005 (approved):

PR #4722, opened 2025-09-22
chore: drop invalid sql test as it triggers exception breakpoints

0ax1:
This lead to poor DX when running `vortex-duckdb` unit tests, as `lldb` stops multiple times on exception breakpoints. It is expected though and correct behavior that DuckDB throws an exception for invalid SQL queries.


0ax1:
same as: https://github.com/vortex-data/vortex/commit/54388b2259b3444cc639a05d7e8820fd1e059107

robert3005 (approved):

PR #4725, opened 2025-09-22
Operator Updates

gatesn:
This PR changes the API a little for operators as well as creates an executor that performs CSE and pipeline conversion.

It breaks many of the old operator tests and benchmarks, many of which are commented out. But to avoid long-lived feature branch @joseph-isaacs and I will continue to merge onto develop.

Execution via operators is gated on a runtime environment variable for local development until we get this prototype fleshed out and performant!


encodings/fastlanes/benches/pipeline_bitpacking.rs
```diff
@@ -68,32 +67,32 @@ pub fn decompress_bitpacking_late_filter<T: NativePType>(bencher: Bencher, fract
...
+//         .bench_local_values(|mask| {
+//             export_canonical_pipeline_expr(
+//                 array.dtype(),
+//                 array.len(),
+//                 array.to_operator().unwrap().unwrap().as_ref(),
+//                 &mask,
+//             )
+//             .unwrap()
+//         });
+// }
```
joseph-isaacs, line 99: comment or remove

encodings/fastlanes/benches/pipeline_bitpacking_compare_scalar.rs
```diff
@@ -74,58 +72,58 @@ pub fn eval<T: NativePType + Into<Scalar>>(bencher: Bencher, fraction_kept: f64)
...
-                &mask,
-            )
-            .unwrap()
-        });
-}
+// #[divan::bench(types = [u8, u16, u32, u64], args = TRUE_COUNT)]
+// pub fn pipeline<T: Element + NativePType + Into<Scalar>>(bencher: Bencher, fraction_kept: f64) {
+//     let mut rng = StdRng::seed_from_u64(0);
+//     let values = (0..100_000)
+//         .map(|_| T::from(rng.random_range(10..100)).unwrap())
```
joseph-isaacs, line 80: same

encodings/fastlanes/benches/pipeline_bitpacking_kernel.rs
```diff
@@ -28,102 +14,102 @@ pub fn main() {
...
-            kernel.step(&ctx, bit_view, &mut output).unwrap();
-
-            assert_eq!(
-                output.as_slice::<T>()[..mask.true_count()],
-                *expect.as_slice::<T>()
-            );
-        });
-}
+//
+// #[divan::bench(types = [i8, i16, i32, i64], args = [0.01, 0.05, 0.1, 0.3, 0.5, 0.7, 0.9, 1.0])]
```
joseph-isaacs, line 19: here

encodings/fastlanes/benches/pipeline_for.rs
```diff
@@ -68,31 +67,31 @@ pub fn decompress_for_early_filter<T: NativePType>(bencher: Bencher, fraction_ke
...
-            export_canonical_pipeline_expr(
-                array.dtype(),
-                array.len(),
-                array.to_operator().unwrap().unwrap().as_ref(),
-                &mask,
-            )
-            .unwrap()
-        });
-}
+// #[divan::bench(types = [i8, i16, i32, i64], args = TRUE_COUNT)]
```
joseph-isaacs, line 71: +

encodings/fastlanes/src/bitpacking/pipeline/mod.rs
```diff
@@ -29,28 +26,67 @@ impl PipelineVTable<BitPackedVTable> for BitPackedVTable {
...
+        if array.offset != 0 {
+            log::trace!("BitPackedVTable does not support non-zero offsets");
+            return Ok(None);
+        }
 
         Ok(Some(Arc::new(array.clone())))
     }
 }
 
+impl Hash for BitPackedArray {
```
joseph-isaacs: do we actually want this? maybe we want an OperatorHash trait?

vortex-expr/src/exprs/binary.rs
```diff
@@ -111,14 +111,15 @@ impl VTable for BinaryVTable {
...
-
-        Some(Arc::new(CompareOperator::new(lhs, rhs, op)) as OperatorRef)
+    fn operator(expr: &BinaryExpr, scope: &OperatorRef) -> VortexResult<Option<OperatorRef>> {
+        let Some(lhs) = expr.lhs.operator(scope)? else {
+            return Ok(None);
+        };
+        let Some(rhs) = expr.rhs.operator(scope)? else {
+            return Ok(None);
+        };
+        let op: compute::Operator = expr.operator.try_into()?;
```
joseph-isaacs: this is an opt None not an Err

encodings/fastlanes/src/bitpacking/pipeline/mod.rs
```diff
@@ -59,200 +95,237 @@ impl Operator for BitPackedArray {
...
-
-        for i in 0..mask.true_count() {
-            assert_eq!(
-                result.scalar_at(i),
-                expect.scalar_at(i),
-                "mismatch at index {}",
-                i,
-            );
-        }
+            assert_eq!(self.offset, 0, "Offset handling not implemented yet");
```
joseph-isaacs: is it that hard to use the unaligned kernel?

vortex-array/src/operator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        _parent: OperatorRef,
+        _child_idx: usize,
+    ) -> VortexResult<Option<OperatorRef>> {
+        Ok(None)
+    }
+
+    // Which fields this operator accesses (for projection pushdown)
+    // fn required_columns(&self) -> Option<Vec<FieldPath>> {
+    //     None
+    // }
```
joseph-isaacs: remove or impl?

vortex-array/src/operator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    /// Whether this operator is aligned 1:1 with its child.
+    ///
+    /// Returns `None` if unknown.
+    fn is_position_preserving(&self, _child_idx: usize) -> Option<bool> {
+        None
+    }
```
joseph-isaacs: I think we might end up with lots of these shall now put them in another place or add todo?

vortex-array/src/operator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl Eq for dyn Operator {}
+
+/// The default execution mode for an operator is batch mode.
+pub trait BatchOperator: Operator {
+    fn bind(&self, ctx: &mut dyn BatchBindCtx) -> VortexResult<BatchExecutionRef>;
+}
+
+pub trait BatchBindCtx {
+    fn take_child(&mut self, idx: usize) -> VortexResult<BatchExecutionRef>;
```
joseph-isaacs: can you add a doc str, also do you expect many impls of this?

vortex-array/src/pipeline/operator/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+                        len: self.len(),
+                        batch_inputs,
+                        vectors,
+                        pipeline,
+                        phantom_data: PhantomData::<T>,
+                    }))
+                })
+            }
+            _ => vortex_bail!(
+                "PipelineOperator currently only supports primitive output types {}",
```
joseph-isaacs: does this support nullable pipelines?

vortex-array/src/pipeline/operator/mod.rs
```diff
@@ -0,0 +1,389 @@
...
+        )?))
+    }
+}
+
+struct PrimitivePipelineExecution<T> {
+    len: usize,
+    batch_inputs: Vec<BatchExecutionRef>,
+    vectors: Vec<RefCell<Vector>>,
+    pipeline: Pipeline,
+    phantom_data: PhantomData<T>,
```
joseph-isaacs, line 357: output ptype?

vortex-array/src/arrays/varbin/operator.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl PartialEq for VarBinArray {
+    fn eq(&self, other: &Self) -> bool {
+        self.dtype == other.dtype
+            && OperatorHash(self.bytes()) == OperatorHash(other.bytes())
+            && OperatorHash(self.offsets()) == OperatorHash(other.offsets())
+            && OperatorHash(self.validity()) == OperatorHash(other.validity())
+    }
+}
+impl Eq for VarBinArray {}
```
joseph-isaacs: I do think we need a separate trait, feels wrong the impl all these methods on array

vortex-array/src/executor.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Attempts to reduce the given operator by applying this optimization rule.
+    fn optimize(&self, operator: OperatorRef) -> VortexResult<Option<OperatorRef>>;
+
+    /// Returns the order in which this rule should be applied.
+    ///
+    /// If `None`, the rule should handle recursion itself.
+    fn apply_order(&self) -> Option<ApplyOrder> {
+        None
+    }
+}
```
joseph-isaacs: when do we use this vs reduce parent/child, maybe a comment on that?

vortex-array/src/operator/compare.rs
```diff
@@ -0,0 +1,446 @@
...
+
+impl<T: Element + NativePType, Op: CompareOp<T> + Send> Kernel
+    for ScalarComparePrimitiveKernel<T, Op>
+{
+    fn step(&mut self, ctx: &KernelContext, out: &mut ViewMut) -> VortexResult<()> {
+        let lhs_vec = ctx.vector(self.lhs);
+        let lhs = lhs_vec.as_slice::<T>();
+        let bools = out.as_slice_mut::<bool>();
+
+        // TODO(ngates): use lhs.len()?
```
joseph-isaacs: can you?

vortex-array/src/operator/compare.rs
```diff
@@ -0,0 +1,446 @@
...
+/// Less than or equal comparison operation.
+pub struct Lte;
+impl<T: PartialOrd> CompareOp<T> for Lte {
+    #[inline(always)]
+    fn compare(lhs: &T, rhs: &T) -> bool {
+        lhs <= rhs
+    }
+}
+
+// #[cfg(test)]
```
joseph-isaacs, line 332: fixup

vortex-array/src/operator/filter.rs
```diff
@@ -0,0 +1,155 @@
...
+        std::mem::discriminant(&self.mask).hash(state);
+        match &self.mask {
+            Mask::AllTrue(len) => len.hash(state),
+            Mask::AllFalse(len) => len.hash(state),
+            Mask::Values(values) => {
+                Arc::as_ptr(values).hash(state);
+            }
+        }
+    }
+}
```
joseph-isaacs: this is a hash that is valid as long as the Arc is alive

vortex-array/src/operator/filter.rs
```diff
@@ -0,0 +1,155 @@
...
+                if child.is_position_preserving(i).unwrap_or_default() {
+                    // Push-down the filter to this child.
+                    Arc::new(FilterOperator::new(child, self.mask.clone()))
+                } else {
+                    child
+                }
+            })
+            .collect();
+
+        Ok(Some(self.child.clone().with_children(children)?))
```
joseph-isaacs: this is not true `is_position_preserving == None` drops the filter on that child even if its unknown if its `is_position_preserving` need to bail out if this is undefined.

vortex-array/src/operator/filter.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn as_batch(&self) -> Option<&dyn BatchOperator> {
+        Some(self)
+    }
+
+    // fn as_pipelined(&self) -> Option<&dyn PipelinedOperator> {
+    //     TODO(ngates): do we decide if we're super sparse that we should be batch executed?
+    //      Seems like a weird place to make that decision though... Although we do have all the
+    //      information here. Pushdown has already happened, so we know exactly which is faster.
+    // Some(self)
+    // }
```
joseph-isaacs: ?

vortex-array/src/operator/metrics.rs
```diff
@@ -0,0 +1,149 @@
...
+use crate::Canonical;
+use crate::operator::{
+    BatchBindCtx, BatchExecution, BatchExecutionRef, BatchOperator, Operator, OperatorId,
+    OperatorRef,
+};
+use crate::pipeline::view::ViewMut;
+use crate::pipeline::{BindContext, Kernel, KernelContext, PipelinedOperator};
+
+#[derive(Debug)]
+pub struct MetricsOperator {
```
joseph-isaacs, line 24: doc str?

vortex-array/src/operator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    // Which fields this operator accesses (for projection pushdown)
+    // fn required_columns(&self) -> Option<Vec<FieldPath>> {
+    //     None
+    // }
+
+    /// Whether this operator is aligned 1:1 with its child.
+    ///
+    /// Returns `None` if unknown.
+    fn is_position_preserving(&self, _child_idx: usize) -> Option<bool> {
```
joseph-isaacs: I am not sure this is the property you want if more like the ith output is directly dependent on the ith input not about alignment, maybe its the same thing, just need to clarification

vortex-array/src/operator/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        None
+    }
+
+    /// Whether this operator preserves the nulls of the given position-preserving child.
+    ///
+    /// Returns `None` if unknown.
+    fn is_null_preserving(&self, _child_idx: usize) -> Option<bool> {
```
joseph-isaacs: so `is_position_preserving(i) == false` implies is_null_preserving(i) == None`?

vortex-array/src/pipeline/operator/buffers.rs
```diff
@@ -0,0 +1,102 @@
...
+struct VectorAllocation {
+    /// Type of elements in this vector
+    element_type: VType,
+}
+
+// ============================================================================
+// Improved Pipeline with vector allocation
+// ============================================================================
+
+/// Allocate vectors with lifetime analysis and zero-copy optimization
```
joseph-isaacs, line 54: I think this might be me but its incorrect now (no zero-copy)

joseph-isaacs (commented):
This is a lot.

I think the main concepts make sense, Hash/Eq needs some work.

vortex-array/src/operator/filter.rs
gatesn: Yeah good point... I guess that means we shouldn't provide a default implementation? Otherwise it's too easy to forget to declare it and then we will hardly ever push-down filters

vortex-array/src/operator/mod.rs
gatesn: No I don't like all these separate traits unless the function impl is expected to be large. These should just all be kept alongside the definition of the operator.

vortex-array/src/pipeline/operator/buffers.rs
gatesn: We still do zero-copy to the output buffer?

vortex-array/src/pipeline/operator/mod.rs
gatesn: == T

vortex-array/src/executor.rs
gatesn: This is never used, I was experimenting with the idea that we could implement rules that aren't at all tied to a specific operator

vortex-array/src/operator/hash.rs
```diff
@@ -1,36 +1,115 @@
...
 use crate::validity::Validity;
 
-/// A wrapper type to implement [`Hash`] using the semantics defined by
-/// [`crate::operator::Operator`].
-pub struct OperatorHash<T>(pub T);
+/// A hash trait for operators that loosens the semantics to permit pointer-based hashing for
+/// data objects such as buffers.
+pub trait OperatorHash {
+    fn operator_hash<H: Hasher>(&self, state: &mut H);
+}
```
joseph-isaacs, line 22: We need to expand on this since ptr addresses are reuse this implies the hash is only valid if you have a strong ref to the OperatorHash.

vortex-array/src/pipeline/operator/buffers.rs
joseph-isaacs: I thought this was about the reuse of input to output with the same type

vortex-array/src/operator/hash.rs
gatesn: I can add to the docs, but it's kind of implied

vortex-array/src/pipeline/operator/mod.rs
joseph-isaacs: was just suggesting we name the element that? Also you want an underscore?

connortsui20:
@claude review

joseph-isaacs (approved):
🤞

connortsui20:
I'm finding it a bit hard to grok the API change (Edit: after reading a bit more I think I understand a bit better, but I'm going to write this anyways. Edit Edit: And since this is merged this is more of a discussion question for later). I may have missed this when reading over everything, but what I think I expected to see at a high level was:

- An operator tree/DAG structure (either built in to the `Array` layout or formed when converting everything to `Operator`)
- Manipulation / optimization of the graph structure. Things like adding a filter at the top of the tree and then performing filter pushdown, deciding which compute functions to use on top of each operator (are compute functions operators? Or is this supposed to be a different concept?), reordering of children based on simple stats (is that in the scope of this PR?). I saw the `optimize.rs` file but it seems to just do pushdowns and rely on trait implementations to figure that out.
- Translation of the graph structure into a physical execution plan, based on the execution mode we want, that can be executed directly (ideally with as little dispatch as possible?)

So I think I can see all of these things here scattered across the codebase, but I'm wondering if it is possible to structure the code in such a way that the 3 things above are essentially in 1 place. Yesterday I said that it makes sense to allow the plan to be "baked in" to our `Array`s, but I'm not so sure anymore.

For example, I'm thinking about if I wanted to link up Vortex to some future query optimizer. I would want to be able to easily translate a logical decompression "plan" from Vortex into whatever IR that optimizer takes as input (could be substrait, for all we know). I think it would be ideal if there is a clear and obvious way to pass these operator trees: is that the case here? (Just to be clear I don't fully understand this change so maybe that is the case already...)

And then on the other side, is it possible for that optimizer to then spit out a physical plan where we take everything in that plan that is below a join and a shuffle (pipelines) and translate that into our own execution plan. Is there a clear and obvious way we would do that?

TLDR; would it be beneficial to have a clear IR for both the operator tree and the execution plan that would be distinct from each other? My understanding (which could be very wrong) is that this changes somewhat conflates the two (because both concepts seem to be methods dispatched via trait implementations), and I'm not sure if that is the right abstraction.

Issue #4492, opened 2025-09-03
`vortex-operator` roadmap


joseph-isaacs:
The operator epic is looking to unify array encodings and compute into a single operator tree that can later be turned into a canonical form. See #4726 .

We are currently working on:

- [x] Implement vortex-vector to provide canonical interchange format between arrays.
- [ ] Figure out how to handle async expressions
- [ ] Implement batch execution kernels for arrays.
- [ ] Support common sub-expression elimination when binding array kernels.
- [ ] Support aggressive CSE when mask is a subset of a previous mask (e.g. for filter -> project).
- [ ] Define optimization functions on the OperatorVTable for the new functionality of `reduce_parent`, `reduce_child`.

PR #4726, opened 2025-09-22
RFC: Operators

gatesn:
See also https://github.com/vortex-data/vortex/issues/4492


rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
+# Array Operators
+
+This RFC proposes to turn Vortex arrays into logical query plans to support multiple modes of execution. Compute
+functions become arrays, and expressions disappear in favor of serialized arrays with placeholder values.
```
lwwmanning: "Expressions disappear in favor of serialized arrays with placeholder values" sounds weird. I feel like what you want to say is what we currently call Expressions are really just "unbound Arrays"? or like, Arrays with unbound children?

Or perhaps this all comes back to "Arrays and Expressions are all just nodes in a Logical Plan"?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+
+## Motivations
+
+1. **Predictable Compute Cost**
+
+Compute functions in Vortex currently take and return ArrayRefs, meaning they have full decision making control (and
+responsibility) over how much or little compute to actually perform.
+
+For example, `filter(DictArray, Mask)` could fully evaluate the filter, or it could push down the filter over the
+dictionary codes with an unknown cost.
```
lwwmanning: Worth noting that the problem here is not just unpredictability... looking at this as a Logical Plan, it tends to make  optimization opportunities local and intermingled with the actual execution, since every layer is somewhat opaque.

This RFC lets us have a separate planning/optimization step

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+- Batch Compute - operates over a full array, returning a canonical array.``
+- Pipelined Compute - operates over vectors of 1024 elements at a time, improving CPU cache locality getting much closer
+  to the performance of fused decompression kernels.
+- GPU Compute - self explanatory
+- JIT Compute - take the logic from pipelined compute and jit-compile.
+
+## Arrays as a Logical Plan
+
+We have long described the scope of Vortex as performing “linear compute”. In other words, any compute that doesn’t
+require shuffling data. In many ways this RFC builds on the idea that really Vortex is fundamentally a subset of a query
```
lwwmanning: I'd frame it as "Vortex implements a subset of functionality usually performed by the query engine. Specifically, we implement a highly optimized physical Scan operation, that supports pushdown of filters, projections, and scalar expressions."

It's worth being very precise here with the relational algebra / common query engine terminology

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+engine and leans into this with somewhat familiar terminology.
+
+Vortex arrays will become nodes in a logical plan where the execution of such a plan returns a canonical array. Recall
```
lwwmanning: Must the execution return a canonical array? How does this work for e.g., duckdb export, where we want an ALP or Dictionary vector?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+### Exporting Compressed Arrays
+
+Many query engines support partially compressed data (e.g. DuckDB, Velox) rather than fully decompressed Arrow (e.g.
+DataFusion, Polars).
+
+For these engines, we propose that the export logic inspects the root node of the array for compatible compression
+codecs and deconstructs the array as required.
+
+For example, DuckDB supports dictionary-encoded vectors. So we would try to downcast the root node as a DictArray, and
+if successful, canonicalize the codes and values children separately.
```
lwwmanning: This example is helpful, but it feels like we should carry it further. Since on its face, the prior section seems slightly in conflict... I think you're just missing a sentence to the effect of "This effectively splits the original execution plan into multiple execution plans, to meet the prior invariant that each execution returns a canonical array"

But then it raises the question... if they're now disjoint, how does that square with Common Subtree Elimination below?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+
+array.reduce_children() -> DictArray:
+    codes: IntArray
+    values: ArithmeticArray:
+        op: subtract
+        lhs: IntArray
+        rhs: Constant{value = 1}
+```
+
+**Example 3: Reduce Parent - Pull Down**
```
lwwmanning, line 232: these examples are great

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+```rust
+trait Array {
+    fn len(&self) -> usize;
+    fn dtype(&self) -> &DType;
+
+    ...
+
+    // Inspection APIs for push-down / pull-up optimization
+    fn children(&self) -> &[ArrayRef];
+    fn with_children(self: Arc<dyn Self>) -> VortexResult<ArrayRef>;
```
lwwmanning: what does with_children do...?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+optimizations within a batch evaluation.
+
+### Pipelined Execution
+
+For deeply nested array trees there can be significant benefit to passing a small chunk of data through the entire
+pipeline of array operators at once, keeping much of the data in the L1 CPU cache.
+
+Early experiments have shown up to 4x improvements in performance.
+
+Pipelines operator over vectors of 1024 elements at a time, until the final vector which may be shorter. They are not
```
lwwmanning: typo: s/operator/operate/

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+```
+
+### GPU Execution
+
+Similar to pipelined execution, the executor will find subgraphs of the array tree that are eligible for execution on
+the GPU (ideally, we reach full compatibility and the entire graph can be moved).
+
+During the bind phase, arrays buffers are moved over from CPU to GPU memory. Note that these buffers contain fully
+compressed data since they live in the leaves of the compressed array tree. In the future, we should be able to hint to
+the vortex-scan logic that buffers should be loaded directly from files/network into the GPU, avoiding the compressed
```
lwwmanning, line 403: 🔥

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+use the operator tree for evaluation inside Vortex scan.
+
+Once we decide we have sufficient coverage, we would port the implementation of into_canonical to use the same executor.
+
+Finally, we would need to big-bang the migration of the array trait and compute function entry points such that
+`take(array) -> TakeArray{ child: Array }` instead of performing any eager compute.
+
+## Open Questions
+
+- Is the canonical return value to performance-restrictive
```
lwwmanning: typo: s/to/too/

rfcs/0001-operators.md
```diff
@@ -0,0 +1,369 @@
...
+
+Note: we are yet to decide whether Vortex canonical representations will diverge from Arrow representations, but for now
+they are all identical to some Arrow form giving us zero-copy export.
+
+### Exporting Compressed Arrays
+
+Many query engines support partially compressed data (e.g. DuckDB, Velox) rather than fully decompressed Arrow (e.g.
+DataFusion, Polars).
+
+For these engines, we propose that the export logic inspects the root node of the array for compatible compression
```
blaginin, line 128: Can you please elaborate on this a bit more? 

In spiral, if there are several fragment files, we would always combine them using vortex `zip` function. Say, it would become an operator. Then we'll always need to decompress data on (almost) every single read in vortex. Then we won't be able to use, for example, FSST in spiral at all?

rfcs/0001-operators.md
gatesn: The next motivation point no?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+
+## Expressions as an Abstract Plan
+
+As part of this change, the `vortex-expr` crate can be significantly simplified.
+
+* Each expression becomes an abstract node in a tree containing an `id`, `metadata`, and `children`. The leaves of
+  this tree are represented by the placeholder expression called `scope` (often denoted as `$`).
+* Expressions can be converted to arrays by first converting their children to arrays, then calling the `build`
+  function of the array encoding indicated by the `id` of the expression, along with the serialized `metadata`. This
+  process results in a bound array now with known `len` and `dtype`.
```
connortsui20, line 142: By "bound" we mean that the expression now has data bound to it, right? I think that might be worth spelling out

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+* All evaluation logic (for executing the expression), and analysis logic (for producing pruning predicates) will be
+  moved to the `Array` trait.
+
+Expressions simply provide a way of describing an unbound query over a root scope array and make up the user-facing
```
connortsui20, line 146: I am not sure I understand what a "scope" is super well, above it says that scopes are at the leaves of the expression tree, but this says it is at the root?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+
+## Layouts and Scanning
+
+Vortex scans currently operate by converting a `Layout` into a `LayoutReader`. In practice, this is largely equivalent
+to a current Vortex `Array`, except the execution function of the `LayoutReader` is async. In the new operator model,
+arrays _are_ async, and so we can remove `LayoutReader` entirely in favor of regular Vortex arrays.
+
+This change further allows us to simplify scanning logic and avoid the current expression partitioning. Instead of
+splitting expressions and pushing the partitions down over the fields of a struct layout, we simply apply the expression
+to a row-range of a layout to produce an array: `Layout::apply(Expr, RowRange) -> ArrayRef`. A `FlatLayout` would be
```
connortsui20, line 157: This probably stems from my lack of understanding about how expressions and layouts currently work, but I thought that in this new model we get the expression **from** our layouts, not apply a layout to the expression. Basically, where is the expression coming from if not from the layout?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+modelled as a node in the array tree allowing us to perform partial optimizations over the tree, such as pushing
+`get_item` expressions through a `StructLayout`.
```
connortsui20, line 159: Not sure I understand why `FlatLayout` being a node in the array tree is the specific reason that allows us to push `get_item` through a `StructLayout`, isn't `FlatLayout` supposed to be the leaf nodes before we are able to "resolve" it to something more specific?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+
+The scan would then find all `FlatLayout` nodes and resolve their segment IDs to load an array for each segment. These
+nodes will be swapped out for the loaded arrays, the optimizer re-run to push compute into the flat layouts, and then
+the root array can be evaluated.
+
+## Optimizations
+
+### Common Subtree Elimination
```
connortsui20, line 167: I think this means that we do not have to cache _any_ computed data / never do duplicate work, is that true or are there exceptions?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+struct ChildInfo {
+    // N.B. These properties will evolve as we attempt to implement optimization rules.
+    aligned: bool, // Whether the child elements align 1:1 with the parent elements.
+    propagates_null: bool, // Whether the child nulls align with the parent nulls.
+}
+```
+
+## Array API
+
+This change essentially collapses the API for an array down to just an async “into_vector” function. Almost everything
```
connortsui20, line 295: Should `into_vector` be added to `Array` above?

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+can be expressed via this function, and it means we can strip down the surface area required for array implementations.
+
+For example, functions relating to validity can be removed and implemented using an IsNull operator pushed down over the
+array tree and evaluated into a boolean vector. Even scalar_at could be implemented with a singleton selection array,
+evaluated to a vector, and then converted to a scalar.
```
connortsui20, line 300: I think that it is additionally worth considering refactoring our scalars to be a lot closer to vectors, rather than what we currently have which is opaque views into (physical) type-erased `Array`s. If we say that scalars become the elements of our vectors (in which we can additionally have references into our vectors), I bet we would have _massive_ performance gains from the reduction in (imo) wasteful allocations we do to keep everything opaque.

rfcs/0001-operators.md
```diff
@@ -0,0 +1,495 @@
...
+Pipelines operate over vectors of 1024 elements at a time, until the final vector which may be shorter. They are not
+allowed to perform any I/O or async operations.
+
+The execution engine looks at the tree of arrays and finds subgraphs where all nodes support pipelined execution. These
+subgraphs are collapsed into a single PipelineBatchExecution node that executes the subgraph step by step.
+
+During construction, the pipeline goes through common sub-tree elimination, followed by a vector allocation stage. This
+stage computes the minimal number of vectors required for execution based on the topological sort of the pipeline DAG (
+essentially the same problem as register allocation in a compiler). By minimizing the number of intermediate vectors, we
+can maximize our use of CPU caches. Where possible, pipeline operators will write directly into the output vector.
```
connortsui20, line 360: Are we planning on implementing the reuse of memory (after we use a vector we can reuse its memory for a future vector, this is what `collect` does and is probably more complicated to implement?).

connortsui20 (changes requested):
I think it is worth considering refactoring `vortex-scalar` as well to truly support a "canonicalized vectors are our IR" world, but I'm not sure if it fits into this timeline?

lwwmanning (approved):

PR #4730, opened 2025-09-23
Minimal docs for vortex-btrblocks

AdamGS:


joseph-isaacs (approved):

joseph-isaacs:
Can we now require doc strs on this crate?

gatesn (commented):
Wanna go the extra mile and add deny_missing_docs?

AdamGS:
added a bunch of docs, and reduced the public API a bit. I think we might want to make it even smaller and/or make the compressor more configurable/modular.

PR #4734, opened 2025-09-23
chore: bump duckdb to 1.4.0

0ax1:
try the 1.3.2 matching vortex commit, don't merge


0ax1:
takeaway: tpch broke exclusively due to changes in duckdb since version 1.3.2 not vortex

robert3005:
But are we doing something wrong with the duckdb integration?

PR #4737, opened 2025-09-23
Add benchmarks for StructArray scalar_at and take

AdamGS:
Some of these code paths are known to be slow, and this will make it easier to benchmark and improve directly without having to pull other repos or create one-off demos.


AdamGS:
I guess I need to do something to get this to run on codspeed? I'll take a look tomorrow.

danking:
@AdamGS looks good to me but it'd be nice to see the codspeed first. Do I need to trigger that?

robert3005:
it's there https://github.com/vortex-data/vortex/pull/4737/checks?check_run_id=51065228379 no clue why there's no comment

AdamGS:
I've made the benchmarks bigger by an order of magnitude just to make differences less likely to be noise.
@robert3005  - I think we codspeed now comments only when there's a difference above some threshold.

robert3005 (approved):

robert3005:
Maybe add width to the name of the benchmark?

PR #4740, opened 2025-09-23
Pipeline row selection

gatesn:
This PR changes the API for a pipeline to:
* Have all nodes of a pipeline reference the same selection mask for a single step.
* Make kernels stateless, passing the step idx
* Break pipelines wherever the row selection domain changes


encodings/fastlanes/src/bitpacking/operator/aligned_kernel.rs
```diff
@@ -41,27 +38,26 @@ where
...
 {
-    fn step(&mut self, _ctx: &KernelContext, out: &mut ViewMut) -> VortexResult<()> {
+    fn step(
+        &self,
+        _ctx: &KernelContext,
+        chunk_idx: usize,
+        _selection: &BitView,
+        out: &mut ViewMut,
+    ) -> VortexResult<()> {
+        assert_eq!(N, 1024, "BitPackedKernel assumes N=1024");
```
joseph-isaacs: remove?

encodings/fastlanes/src/for/pipeline.rs
```diff
@@ -77,7 +80,7 @@ impl Operator for FoROperator {
         &self.dtype
     }
 
-    fn len(&self) -> usize {
+    fn bounds(&self) -> LengthBounds {
         todo!()
     }
```
joseph-isaacs, line 85: Does this make sense? Is this the child's bounds?

encodings/fastlanes/src/for/pipeline.rs
```diff
@@ -178,11 +188,17 @@ where
...
     E: NativePType + Element,
 {
-    fn step(&mut self, ctx: &KernelContext, out: &mut ViewMut) -> VortexResult<()> {
+    fn step(
+        &self,
+        ctx: &KernelContext,
+        _chunk_idx: usize,
+        _selection: &BitView,
+        out: &mut ViewMut,
+    ) -> VortexResult<()> {
```
joseph-isaacs, line 197: there is something strange about only leave nodes caring about chunk_idx?

encodings/fastlanes/src/for/pipeline.rs
```diff
@@ -178,11 +188,17 @@ where
     T: NativePType + Element + WrappingAdd,
     E: NativePType + Element,
 {
-    fn step(&mut self, ctx: &KernelContext, out: &mut ViewMut) -> VortexResult<()> {
+    fn step(
+        &self,
+        ctx: &KernelContext,
+        _chunk_idx: usize,
+        _selection: &BitView,
```
joseph-isaacs, line 195: what about selection do we just not use it yet

encodings/fastlanes/src/for/pipeline.rs
```diff
@@ -178,11 +188,17 @@ where
     T: NativePType + Element + WrappingAdd,
     E: NativePType + Element,
 {
-    fn step(&mut self, ctx: &KernelContext, out: &mut ViewMut) -> VortexResult<()> {
+    fn step(
```
joseph-isaacs, line 191: shall we replace this with a arithmetic kernel?

encodings/sequence/src/operator.rs
```diff
@@ -123,21 +131,41 @@ impl PipelinedOperator for SequenceArray {
...
+                values[i] = self.base
+                    + T::from_usize(offset + i)
+                        .ok_or_else(|| vortex_err!("Overflow converting usize to ptype"))?;
+                Ok(())
+            })?;
+        } else {
+            for i in 0..N {
+                values[i] = self.base
+                    + T::from_usize(offset + i)
+                        .ok_or_else(|| vortex_err!("Overflow converting usize to ptype"))?;
```
joseph-isaacs, line 160: we could check if there is overflow and then do with without an error being thrown in the loop?

encodings/fastlanes/src/for/pipeline.rs
gatesn: Yah, need to choose the threshold. Added a comment

encodings/fastlanes/src/for/pipeline.rs
gatesn: Perhaps, or perhaps not 🤷

encodings/fastlanes/src/for/pipeline.rs
gatesn: Eventually yes

encodings/sequence/src/operator.rs
gatesn: Unclear if that's faster? Worth a benchmark later

vortex-array/src/pipeline/operator/mod.rs
```diff
@@ -299,95 +368,224 @@ impl BatchOperator for PipelineOperator {
...
+                            "Mismatched input lengths for pipeline domain inputs: {:?}",
+                            lens
+                        );
+                    }
+                }
+            }
+            LenProvider::Fixed(len) => *len,
+            LenProvider::Mask => {
+                let selection_mask = batch_inputs
+                    .last()
```
joseph-isaacs: this needs documenting

vortex-array/src/pipeline/operator/mod.rs
```diff
@@ -299,95 +368,224 @@ impl BatchOperator for PipelineOperator {
...
+                    let mut elements_view = output.view_mut(position);
+                    self.pipeline.step(
+                        &ctx,
+                        chunk_idx,
+                        &selection.as_view(),
+                        &mut elements_view,
+                    )?;
+
+                    // Flatten the elements view such that the selected elements are at the front.
+                    elements_view.flatten::<O::Element>(&selection.as_view());
```
joseph-isaacs, line 574: why do we need this, are they not already at the front?

gatesn: Not necessarily right? Depends what the root node of the pipeline writes into it's output vector?

joseph-isaacs (approved):

PR #4742, opened 2025-09-24
Zero copy from any Vec into a Buffer

AdamGS:
I just want to have nice things


AdamGS:
ok this annoys miri a bit, need to try and fix it

AdamGS:
ok this works now

robert3005 (approved):

PR #4746, opened 2025-09-24
fix[array]: nested Operator format fix

joseph-isaacs:


vortex-array/src/operator/mod.rs
```diff
@@ -82,10 +84,22 @@ pub trait Operator: 'static + Send + Sync + Debug + DynOperatorHash + DynOperato
     }
 
     /// Override the default formatting of this operator.
-    fn fmt_as(&self, _df: DisplayFormat, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+    fn fmt_as(&self, _df: DisplayFormat, f: &mut Formatter) -> fmt::Result {
         write!(f, "{}", type_name::<Self>())
     }
 
+    fn fmt_all(&self) -> Tree<String> {
```
gatesn: Seems weird that our operator API returns a treenode::Tree object?

gatesn (commented):
Is there a way to do this without exposing the treenode API?

vortex-array/src/operator/mod.rs
joseph-isaacs: Done, but it create lots of string, which we can opt later.

gatesn (approved):

PR #4747, opened 2025-09-24
Zip Fragmentation on `take`

blaginin:


vortex-array/src/compute/zip.rs
```diff
@@ -260,4 +263,134 @@ mod tests {
...
+          buffer (align=1): 28 B (0.63%)
+          buffer (align=1): 29 B (0.65%)
+          buffer (align=1): 28 B (0.63%)
+          buffer (align=1): 29 B (0.65%)
+          buffer (align=1): 28 B (0.63%)
+          buffer (align=1): 29 B (0.65%)
+          buffer (align=1): 28 B (0.63%)
+          buffer (align=1): 29 B (0.65%)
+          buffer (align=1): 28 B (0.63%)
+          buffer (align=1): 29 B (0.65%)
```
blaginin: @onursatici this shouldn't be the case, right?

blaginin: I remember you worked on something similar in https://github.com/vortex-data/vortex/pull/4054

vortex-array/src/compute/zip.rs
```diff
@@ -75,6 +75,15 @@ impl ComputeFnVTable for Zip {
         // TODO(os): add invert_mask opt and check if if_false has a kernel like:
         //           kernel.invoke(Args(if_false, if_true, mask, invert_mask = true))
 
+        if !if_true.is_canonical() || !if_false.is_canonical() {
```
robert3005, line 78: what is the thing that is actually changing here? Doesn't this indicate that our default impl is bad?

blaginin: It makes us use `impl ZipKernel for VarBinViewVTable`. Default impl is indeed doesn't handle VarBinViews properly, that's why Onur did https://github.com/vortex-data/vortex/pull/4054

onursatici (commented):
I think this makes sense, but we should just make zip_impl, the default implementation when no kernels are found, to be the same as the varbinview kernel. We can then remove the varbinview kernel as well, because it will be the default for two already canonicalised arrays anyway.

we might need to branch in creating the builder to create the deduplicating builder for varbinviews, and the regular canonical builders for others

vortex-array/src/compute/zip.rs
```diff
@@ -260,4 +276,87 @@ mod tests {
...
+        let if_true = {
+            let mut builder = VarBinViewBuilder::new(
+                DType::Utf8(Nullability::NonNullable),
+                10,
+                Default::default(),
+                BufferGrowthStrategy::fixed(64 * 1024),
+            );
+            for _ in 0..100 {
+                builder.append_value("Hello");
+                builder.append_value("Hello this is a long string that won't be inlined.");
```
blaginin: just moved this test

vortex-array/src/builders/varbinview.rs
```diff
@@ -284,13 +284,13 @@ impl ArrayBuilder for VarBinViewBuilder {
...
 
 pub enum CompletedBuffers {
-    Default(Vec<ByteBuffer>),
+    Plain(Vec<ByteBuffer>),
     Deduplicated(DeduplicatedBuffers),
 }
 
 impl Default for CompletedBuffers {
     fn default() -> Self {
-        Self::Default(Vec::new())
```
blaginin, line 293: @onursatici I didn't change `builder_with_capacity` because IMO it would be non-intuitive if it returned something rather than the default builder. Changed it here instead - feel like we can [afford](https://github.com/vortex-data/vortex/pull/4747#issuecomment-3333653486) `Deduplicated` to be the default right?

vortex-array/src/builders/varbinview.rs
```diff
@@ -284,13 +284,13 @@ impl ArrayBuilder for VarBinViewBuilder {
 }
 
 pub enum CompletedBuffers {
-    Default(Vec<ByteBuffer>),
+    Plain(Vec<ByteBuffer>),
```
blaginin: renamed otherwise it's strange that it is not chosen by default

vortex-array/src/compute/zip.rs
blaginin: I made the default behavior different, but i think needs to still be here, because it follows the pattern from other functions ([1](https://github.com/vortex-data/vortex/blob/d2bb6e243173c78b57703a9c2bf6d6744235ef0f/vortex-array/src/compute/nan_count.rs#L142), [2](https://github.com/vortex-data/vortex/blob/d2bb6e243173c78b57703a9c2bf6d6744235ef0f/vortex-array/src/compute/filter.rs#L123), [3](https://github.com/vortex-data/vortex/blob/d2bb6e243173c78b57703a9c2bf6d6744235ef0f/vortex-array/src/compute/take.rs#L175))

onursatici (commented):
I want to turn on deduplication by default but we use it in some many places to make the switch easy: https://github.com/vortex-data/vortex/pull/4695
I think the default should remain the old default for now because of that

blaginin:
> I think the default should remain the old default for now because of that

Sure, in that case, let's handle this case specifically for `VarBinViewVTable` (like we do with other compute functions)? Later, once https://github.com/vortex-data/vortex/pull/4695 is merge, we'll handle it properly for all arrays

AdamGS (approved):
This LGTM, IDK if anyone else wants to weigh in.

blaginin:
Will discuss with @onursatici once he is back and do changes of top as needed :)

PR #4753, opened 2025-09-24
Bug: DuckDB List Validation assertion failure

connortsui20:
On develop, if we run all 3 of these tests with `VX_DUCKDB_DEBUG=1`, we get an assertion failure when DuckDB validates the list view entries. So for now, we should ignore them.

For context, I have been migrating `List` to use `ListViewArray` instead of `ListArray`, and I reimplemented `export` under `vortex-duckdb/src/exporter/list.rs` to export directly from a `ListView`.

It seems like we might be interpreting memory incorrectly? Here is the relevant code (on develop):

###### vortex-duckdb/src/exporter/list.rs
```rust
pub struct duckdb_list_entry {
    pub offset: u64,
    pub length: u64,
}

<-- snip -->

let offsets_slice: &mut [cpp::duckdb_list_entry] =
    unsafe { vector.as_slice_mut::<cpp::duckdb_list_entry>(len) };

<-- snip -->

// Ordering doesn't really seem to matter here but maybe it does?
vector.list_vector_reserve(sum_of_list_lens)?;
let mut elements_vector = vector.list_vector_get_child();
vector.list_vector_set_size(sum_of_list_lens)?;

self.elements_exporter.export(
    usize::try_from(start_offset)?,
    usize::try_from(sum_of_list_lens)?,
    &mut elements_vector,
)
```

Here is an example assertion failure:

```
---- e2e_test::vortex_scan_test::test_vortex_scan_list_of_ints stdout ----

thread 'e2e_test::vortex_scan_test::test_vortex_scan_list_of_ints' panicked at vortex-duckdb/src/e2e_test/vortex_scan_test.rs:488:10:
called `Result::unwrap()` on an `Err` value: Failed to execute query:
>   INTERNAL Error: Assertion triggered in file "/Users/connor/spiral/vortex-data/vortex-develop/target/duckdb-source-v1.4.0/duckdb-1.4.0/src/common/types/vector.cpp"
>   on line 1802: le.offset + le.length <= child_size

<-- snip -->
4        duckdb::InternalException::InternalException<char const*, int, char const*>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, char const*, int, char const*) + 80
5        duckdb::DuckDBAssertInternal(bool, char const*, char const*, int) + 124
6        duckdb::Vector::Verify(duckdb::Vector&, duckdb::SelectionVector const&, unsigned long long) + 4572
7        duckdb::Vector::Verify(unsigned long long) + 56
8        duckdb::DataChunk::Verify() + 152
9        duckdb::PipelineExecutor::EndOperator(duckdb::PhysicalOperator&, duckdb::optional_ptr<duckdb::DataChunk, true>) + 84
<-- snip -->
```

I double checked that the values that we write into that vector are correct (both on develop and on my branch where the list view `offsets` and `sizes` are definitely correct). So it is likely the case that we are writing correct values, but writing it in the wrong place (or at least duckdb is expecting it somewhere else).

I'd like to go into lldb and figure out what exactly `le.offset` and `le.length` are, but I wasn't able to figure out how to do that easily since a new process is spawned.

CC @0ax1 @danking 

---

Also adds a test that is slightly simpler than `test_vortex_scan_nested_fixed_size_list_utf8`, simply removes the outer fixed-size list layer (and the other assertion error `child.GetVectorType() == VectorType::FLAT_VECTOR` still triggers)

---

Edit: Somewhat resolved in comments below?

Something of note that is VERY weird that happened to me more than once (but not every time):

- If I do a clean build it fails immediately. 
- If I make a trivial change anywhere (add `dbg!("");` somewhere) and run with only `VX_DUCKDB_DEBUG`, **it sometimes stops failing, pretty consistently**
- If I add `VX_DUCKDB_ASAN` and compile tests again, it starts to fail again

There seems to be something strange going on with the builds as well because in certain cases the tests will start passing again seemingly randomly, and I cannot replicate it consistently.

### DIFF
--- a/vortex-duckdb/src/e2e_test/vortex_scan_test.rs
+++ b/vortex-duckdb/src/e2e_test/vortex_scan_test.rs
@@ -391,11 +391,63 @@ fn test_write_timestamps() {
 }
 
 #[test]
+#[ignore = "Assertion fail in DuckDB (`child.GetVectorType() == VectorType::FLAT_VECTOR`) with debug"]
+fn test_vortex_scan_fixed_size_list_utf8() {
+    // Test a simple FixedSizeList of Utf8 strings to ensure proper materialization.
+
+    let runtime = tokio::runtime::Runtime::new().unwrap();
+    let file = runtime.block_on(async {
+        // Create a large number of strings to stress test.
+        let strings: Vec<&str> = (0..24)
+            .map(|i| match i % 6 {
+                0 => "first",
+                1 => "second",
+                2 => "third",
+                3 => "fourth",
+                4 => "fifth",
+                _ => "sixth",
+            })
+            .collect();
+
+        let strings_array = VarBinViewArray::from_iter_str(strings);
+
+        // Create fixed-size lists of strings.
+        let fsl = FixedSizeListArray::new(
+            strings_array.into_array(),
+            4, // 4 strings per list
+            Validity::AllValid,
+            6, // 6 lists total
+        );
+
+        write_single_column_vortex_file("string_lists", fsl).await
+    });
+
+    let conn = database_connection();
+    let file_path = file.path().to_string_lossy();
+
+    // Query the structure.
+    let result = conn
+        .query(&format!(
+            "SELECT string_lists FROM vortex_scan('{file_path}')"
+        ))
+        .unwrap();
+
+    let mut row_count = 0;
+    for chunk in result {
+        row_count += chunk.len();
+        // Accessing the structure should not cause a segfault.
+        let _vec = chunk.get_vector(0);
+    }
+    assert_eq!(row_count, 6, "Should have retrieved 6 lists");
+}
+
+#[test]
+#[ignore = "Assertion fail in DuckDB (`child.GetVectorType() == VectorType::FLAT_VECTOR`) with debug"]
 fn test_vortex_scan_nested_fixed_size_list_utf8() {
     // Regression test for a segfault that occurs inside query 7 and 8 of the `statpopgen` benchmark
     // when running with `FixedSizeList` instead of `List`.
 
-    // Test FixedSizeList of Utf8 to ensure proper materialization.
+    // Test FixedSizeList of FixedSizeList of Utf8 to ensure proper materialization.
 
     let runtime = tokio::runtime::Runtime::new().unwrap();
     let file = runtime.block_on(async {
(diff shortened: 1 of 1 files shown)


0ax1:
> Something of note that is VERY weird that happened to me more than once (but not every time):

Only explanation I can come up with is that your editor races on the build without having set `VX_DUCKDB_DEBUG`. 
After the editor does a non-debug build, you trigger the next test run, which does not trigger a DuckDB rebuild.

In general, changing the DuckDB env vars is picked up by our `vortex-duckdb/build.rs` https://github.com/vortex-data/vortex/blob/develop/vortex-duckdb/build.rs#L281. I think setting the asan env var makes the build aware to rebuild. Probably the same would happen if you set `VX_DUCKDB_DEBUG=0` and back to `1` again. So my hunch is that your editor somehow manages to slip through a non-debug build. I don't think the issue is in DuckDB's C++ build system. 

I'll have a look and see if I can repro this.

--- 

Edit: We could repro this on Joe's machine. Oddly, it was linking against the debug lib though. When you run into this issue again, can you check for the output of `otool -L target/duckdb-lib/libduckdb.dylib`?

For debug it'll give you:

```
otool -L target/duckdb-lib/libduckdb.dylib
target/duckdb-lib/libduckdb.dylib:
        @rpath/libduckdb.0.0.dylib (compatibility version 0.0.0, current version 0.0.1)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 2000.63.0)
        @rpath/libclang_rt.tsan_osx_dynamic.dylib (compatibility version 0.0.0, current version 0.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1356.0.0)
```

where `.0.0` is the confirmation that we're linking against the debug build.

connortsui20:
You're probably right with it being something to do with the editor, I tried making trivial changes with vim and no LSP and wasn't able to reproduce (tests still failed), and then when I opened zed and allowed `rust-analyzer` to finish building, the tests passed!

I will note though that I have my `rust-analyzer` set up to use a separate build directory in `./target/rust-analyzer/` so that my editor doesn't take the lock on the actual build directory when I want to do other stuff. Seems like `duckdb-lib` is just under `target/`, not under `target/duckdb-lib`, so my `rust-analyzer` is probably rebuilding DuckDB incorrectly?

danking (approved):

PR #4754, opened 2025-09-24
Feature: Canonicalize `ListView`

connortsui20:

This PR migrates the `Canonical::List` encoding from `ListArray` to `ListViewArray`.

There were a lot of small things that had to be updated, as well as tests that assumed `ListArray` was the default encoding.

# Unresolved

- [x] ~~Bug in TPC-DS (`Failed to downcast`) due to `list_contains` kernel not expecting a `ListViewArray`~~
- [X] I am not entirely sure that my refactoring of `encodings/sparse/src/canonical.rs` and `vortex-array/src/arrays/chunked/decode.rs` is correct, so I would appreciate scrutiny on that. (I cleaned it up but also filed https://github.com/vortex-data/vortex/issues/4777)
- [x] Should the `take` compute function on `ListArray` return a `ListViewArray`? We might also want to defer this question to the operator changes later. (yes)
- [X] The preferred arrow encoding for lists is still the normal `GenericListArray`, not their `GenericListViewArray`. There are some tests that rely on this, not sure if should just remove parts of those tests. (`ToArrow` is now implemented for `ListArray`
- [ ] The statpopgen benchmark slows down by 1.6-2.1x. That should be investigated, but is performance in the scope of this PR? There are a lot of compute functions that `ListView` needs to operate well (to make use of the fact that it is a view).


connortsui20:
A few bugs still remaining but those shouldnt impact the entire PR

vortex-array/src/arrays/chunked/decode.rs
```diff
@@ -83,56 +83,64 @@ fn swizzle_struct_chunks(
...
+        // Track the current position in our combined elements array.
+        let current_elements_offset = elements_builder.len() as u64;
+
+        // Append all elements from this chunk.
+        elements_builder.extend_from_array(chunk.elements());
+
+        // Append offsets and sizes, adjusting offsets to point into the combined array.
+        for i in 0..chunk.len() {
+            offsets.push(current_elements_offset + offsets_slice[i]);
+            sizes.push(sizes_slice[i]);
```
robert3005: this doesn't have to happen in the loop, do it once for all sizes

vortex-array/src/arrays/list/compute/take.rs
```diff
@@ -14,6 +13,8 @@ use crate::validity::Validity;
 use crate::vtable::ValidityHelper;
 use crate::{Array, ArrayRef, OffsetPType, ToCanonical, register_kernel};
 
+// TODO(connor)[ListView]: Should `take` on a `ListArray` return a `ListViewArray` instead?
```
robert3005: I think so - you can return whatever is more convenient, there's no requirement that the function returns the same encoding

vortex-array/src/canonical.rs
```diff
@@ -500,6 +500,7 @@ mod test {
     }
 
     #[test]
+    #[ignore = "TODO(connor)[ListView]: Arrow prefers `List` over `ListView`"]
```
robert3005: all you need here is an `ToArrow` kernel for ListArray so it can support roundtrip

vortex-duckdb/src/exporter/list.rs
```diff
@@ -3,86 +3,99 @@
...
+
+            duckdb_list_views[i] = cpp::duckdb_list_entry { offset, length };
         }
 
-        // TODO(DK): This calls `list_vector_reserve` once for each call to `export`. Moreover, we
-        // copy slices of the elements once for each call to `export`. Would `export` be faster if
-        // we copied all the elements on the first call to `export` so that subsequent calls need
-        // only copy the correct slice of offsets?
-        vector.list_vector_reserve(sum_of_list_lens)?;
+        // TODO(connor): This is quite bad: we have to export the entire elements array every single
```
robert3005: @joseph-isaacs might have a solution here

robert3005 (commented):
Answered some ofthe inline comments. I think we should keep ToArrow kernel for list array and then we are good to go

vortex-array/src/canonical.rs
connortsui20: Actually where do I add that?

robert3005: there's already `impl ToArrowKernel for VarBinVTable` which you can use as a template

vortex-duckdb/src/exporter/list.rs
joseph-isaacs: You try what we do in the dict array, where we create a single elements vector in `new_exporter` and then `reference` that as the element child. 

I wonder if

```
let mut child = vector.list_vector_get_child()
child.reference(self.duckdb_elements)
```

Will just work, you might want to include `vector.list_vector_set_size(self.num_elements as u64)?` after

vortex-duckdb/src/exporter/list.rs
```diff
@@ -3,86 +3,99 @@
...
         let mut elements_vector = vector.list_vector_get_child();
-        vector.list_vector_set_size(sum_of_list_lens)?;
-        self.elements_exporter.export(
-            usize::try_from(start_offset)?,
-            usize::try_from(sum_of_list_lens)?,
-            &mut elements_vector,
-        )
+
+        self.elements_exporter
+            .export(0, self.num_elements, &mut elements_vector)
```
joseph-isaacs: Do this after writing the elements to the child?
```suggestion
let mut elements_vector = vector.list_vector_get_child();

        self.elements_exporter
            .export(0, self.num_elements, &mut elements_vector)
            
        vector.list_vector_set_size(self.num_elements as u64)?;
```

vortex-duckdb/src/exporter/list.rs
connortsui20: ~~I think the set_size call might be unnecessary?~~ Edit: it was definitely necessary

vortex-duckdb/src/exporter/dict.rs
```diff
@@ -52,6 +52,7 @@ pub(crate) fn new_exporter(
     let values_vector = match cached_vector {
         Some(vector) => vector,
         None => {
+            // TODO(connor)[ListView]: Do we not need to reserve this?
```
joseph-isaacs: ? reserve what?

connortsui20:
@claude review

vortex-array/src/compute/list_contains.rs
```diff
@@ -268,19 +266,44 @@ fn list_contains_scalar(
...
+                })
+                .collect();
+
+            Ok(BoolArray::from_bool_buffer(
+                mask,
+                list_array.validity().clone().union_nullability(nullability),
+            )
+            .into_array())
+        })
     })
```
joseph-isaacs: can we move this into its own function. In not a great idea to have large macro bodies

vortex-duckdb/src/exporter/list.rs
```diff
@@ -3,86 +3,105 @@
...
-            offset_type: PhantomData::<T>,
-        }) as Box<dyn ColumnExporter>
+    let sizes = array.sizes().to_primitive();
+
+    // Create a duckdb elements vector up front so that future exports can reference it.
+    let elements = array.elements();
+
+    let mut duckdb_elements = Vector::with_capacity(elements.dtype().try_into()?, elements.len());
+    let elements_exporter = new_array_exporter(elements, cache)?;
+    elements_exporter.export(0, elements.len(), &mut duckdb_elements)?;
```
joseph-isaacs: I know I suggested this, but it worth at least nothing the trade off here, this will export all elements, even those that are unreachable. Another approach can be seen here (https://github.com/vortex-data/vortex/issues/4767), this is the possible regression from this approach (https://github.com/vortex-data/vortex/pull/4759)

joseph-isaacs:
This list view exporter is really slow, do you think we should think about adding a bound value to list elements. This would record the min and max element used by the offsets in the list view. This we can be refined by slice and filter

connortsui20:
Since the conversation is getting long and I need to rebase a bunch of stuff I'm closing this and will reopen a fresh PR soon.

PR #4759, opened 2025-09-25
perf[duckdb]: fsl canonicalize the list view once

joseph-isaacs:


joseph-isaacs:
Hmm, maybe we want an exporter that slices and canonicalizes?

connortsui20:
Do we know why this fixes things? Shouldn't `new_array_exporter` take care of this?

Edit: Not sure if you're also doing this, but I just traced through the `test_vortex_scan_nested_fixed_size_list_utf8` test and I think the issue might be the `dict` exporter? That is the only difference between calling canonical up front versus within `new_array_exporter`.

connortsui20:
Added another simpler test in https://github.com/vortex-data/vortex/pull/4753 that hits the same assertion as before, but now instead of `FSL(FSL(UTF8))` it's just `FSL(UTF8)`, and it still hits the same assertion error (`child.GetVectorType() == VectorType::FLAT_VECTOR` fails)

joseph-isaacs:
I think the problem is that recursive datatypes require uncompressed/flat vectors, our exporters can return non-flat vectors, if we canon, then we will only use exporters which produce flat vectors since canon <==> flat

PR #4764, opened 2025-09-25
Check typos in CI and fix existing ones

AdamGS:
Seems like a nice idea given that this is publicly facing code and a bunch of places we don't look at often.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -1,5 +1,11 @@
 [default]
 extend-ignore-identifiers-re = ["FoR", "typ"]
+# We support a few common special comments to tell the checker to ignore sections of code
+extend-ignore-re = [
+    "(#|//)\\s*spellchecker:ignore-next-line\\n.*",                      # Ignore the next line
+    "(?Rm)^.*(#|//)\\s*spellchecker:disable-line$",                      # Ignore line that ends with this hint
+    "(?s)(#|//)\\s*spellchecker:off.*?\\n\\s*(#|//)\\s*spellchecker:on", # Ignore block between hints
+]
 
 [files]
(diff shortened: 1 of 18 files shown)


_typos.toml
```diff
@@ -0,0 +1,14 @@
+[default]
+extend-ignore-identifiers-re = [
+    "FoR",
+    "typ"
+]
+
+[files]
+extend-exclude = ["/bench-vortex/**", "/docs/references.bib"]
```
AdamGS, line 11: too many random strings in benchmarks coming from external files

lwwmanning (approved):

vortex-layout/src/layouts/flat/writer.rs
```diff
@@ -248,7 +248,7 @@ mod tests {
                 result.statistics().get_as::<String>(Stat::Min),
                 // The typo is correct, we need this to be truncated.
                 Some(Precision::Inexact(
-                    "Another string that's meant to be smaller than the previous valu".to_string()
+                    "Another string that's meant to be smaller than the previous value".to_string()
```
connortsui20: This test relies on the typo!!!

AdamGS: now I need to figure out how to exclude that 😂

lwwmanning: this truncation is purposeful but kind of stupid. it caught @connortsui20 a couple weeks ago because of his IDE's spell check

IMO, just replace "value" with "one" in the string, and then the truncated one will pass spell-checkers

AdamGS:
configured the type checker so you can tell it to ignore certain code using comments, which I think is useful for these cases.

PR #4765, opened 2025-09-25
fix[duckdb]: handle count(*) queries correctly using an EMPTY column

joseph-isaacs:
dont question the 🦆


vortex-duckdb/src/scan.rs
```diff
@@ -193,6 +190,11 @@ async fn open_file(url: Url, options: VortexOpenOptions) -> VortexResult<VortexF
     }
 }
 
+// taken from duckdb/common/constants.h COLUMN_IDENTIFIER_EMPTY
```
0ax1, line 196: Can we add more docs explaining why we need this.

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -108,8 +108,10 @@ impl ArrayExporter {
         }
 
         if self.fields.is_empty() {
-            // No fields can occur in e.g. count(*) queries. In these cases, we just need to
-            // set the length of the chunk and return.
+            // In the case of a projection pushdown with zero columns duckdb will ask us for the
```
0ax1, line 111: It'd be nice to describe in the comment what happened in the previous case.

0ax1 (approved):

vortex-duckdb/src/scan.rs
```diff
@@ -193,6 +193,13 @@ async fn open_file(url: Url, options: VortexOpenOptions) -> VortexResult<VortexF
...
+// This is used by duckdb whenever there is no projection id in a logical_get node.
+// For some reason we cannot return an empty DataChunk and duckdb will look for the virtual column
+// with this index and create a data chunk with a single vector of that type.
+static EMPTY_COLUMN_IDX: u64 = 18446744073709551614;
```
connortsui20, line 200:
```rust
static EMPTY_COLUMN_IDX: u64 = 18446744073709551614; // 2^64 - 1
```
Maybe?

joseph-isaacs: why?

joseph-isaacs: $2^{64} - 2$

joseph-isaacs: but why does it matter that the value is of that form?

0ax1: discussed offline

PR #4770, opened 2025-09-25
Try do some mutation in-place

AdamGS:


AdamGS:
this was a shot in the dark but just looks bad

Issue #4779, opened 2025-09-26
Offsets, indices, and sizes should be **unsigned only**


connortsui20:
I think it is also worth considering that we restrict most of our integer arrays (`offsets`, `sizes`, `patches_indices`, `take_indices`, `run_ends`) to be **unsigned integers only**.

I believe the only reason we allow signed integers is for FFI compatibility with Java and Arrow? This is not exactly a great pattern since there are only a handful of places where we correctly check that these values are all non-negative. Or put another way, the only thing that has "correct" semantics are dictionary codes.

The only downside @danking and I were able to come up with is that we have to check values at the FFI boundary. But the alternative (which is what we do now) is that we have to check that _everywhere else in the codebase_, which we only seldom do. The last downside would be that we can potentially create an `indices` array (or similar) that has values greater than `i32::MAX`, and we wouldn't be able to export that to Java or Arrow.

However, this is essentially just moving where the error occurs: right now we wouldn't be able to create that array in the first place within Vortex, and if we restrict to just unsigned types, we can simply check if the length of the corresponding `values` or `elements` array is less than `i32::MAX` (there are no cases that I can think of where we have an `offsets` or `indices` array where the values can be greater than the length of the array they are referencing).

Interested to hear what people think about this.

connortsui20:
Additionally, it doesn't really make sense that `IntegerPType` has a bound of `AsPrimitive<usize>` because the signed types can't guarantee that. So every single place where we rely on `IntegerPType` being bound by `AsPrimitive<usize>` is a place where we definitely should be using `UnsignedPType` instead.

connortsui20:
Since we've made the decision not to do this, I'm closing this.

PR #4780, opened 2025-09-26
Chore: change `OffsetPType` to `IntegerPType` and update `NativePType`

### LINKED ISSUE #4776: Move `OffsetPType` into `vortex-dtype` and use for any indices array
Right now I have this `OffsetPType` in `vortex-array/src/offset.rs`.

```rust
pub trait OffsetPType: NativePType + PrimInt + AsPrimitive<usize> + Into<Scalar> { ... }
```

I think it makes sense to move this into `vortex-dtype` instead and also make sure to use it anywhere we want indices, offsets, or sizes arrays. There are a lot of places in the codebase that use `NativePType` for indices, and I'm pretty sure that is just wrong because we should not be able to represent floats as indices or sizes.

The alternative is to call it an `IntegerPType` (which would have the same trait bounds as `OffsetPType`, and we could enforce that in the `match_each_integer_ptype!` macro.

connortsui20:
There were many places where we use `NativePType` when we only want to deal with integers (offsets, indices, codes, run ends, etc).

This change adds `IntegerPType` and `UnsignedPType`, which correspond to exactly the same branch arms as the `match_each_integer_ptype!` macro and the unsigned version. Note that `UnsignedPType` is only used for dictionary codes for now (see issue below), everything else is just making sure we don't use floats where we don't want them.

The relevant traits are located in `vortex-dtype/src/ptype.rs`.

Also, after making all these changes manually, I think that we should be restricting things even further and ensure that things like offsets and sizes are always unsigned. See issue https://github.com/vortex-data/vortex/issues/4779 for more details.

Closes https://github.com/vortex-data/vortex/issues/4776


vortex-scalar/src/bigint/bigcast.rs
```diff
@@ -6,7 +6,7 @@ use crate::i256;
 /// Checked conversion from one primitive type to another.
 ///
 /// This is meant to extend the `ToPrimitive` trait from `num-traits` with awareness of `i256`.
-pub trait ToPrimitive: num_traits::ToPrimitive {
+pub trait ToI256: num_traits::ToPrimitive {
```
joseph-isaacs: This is a strange name for our version of `ToPrimitive` with `I256` support?

encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
```diff
@@ -91,7 +93,7 @@ fn decimal_value_wrapper_to_primitive(
 
 fn decimal_value_to_primitive<P>(decimal_value: DecimalValue) -> Result<ScalarValue, Sign>
 where
-    P: NativePType + NumCast + Bounded + ToPrimitive,
+    P: IntegerPType + ScalarToPrimitive,
```
joseph-isaacs: `ScalarToPrimitive` is odd here?

joseph-isaacs (commented):
Nice change, But I think we should keep ToPrimitive?

vortex-scalar/src/bigint/bigcast.rs
connortsui20: Imo it is quite confusing to have 2 `ToPrimitive`s in our codebase that do basically the exact same thing but one has 1 extra method? And as far as I'm aware we don't actually use it in many places, so to me just having `ToPrimitive + ToI256` is much more readable.

joseph-isaacs: But yours extends `ToPrimitive` so you never need  `ToPrimitive + ToI256`

connortsui20: The issue is that we never actually use this version of `ToPrimitive`, and we just always use the num_traits version (likely because the LSP will default to that). Are you saying we should replace all instances of this with our own?

danking: FWIW, I don't hate it. If I can convert to i256, it seems sensible I can convert to any smaller type.

danking (approved):
I have likewise found the ergonomics around this "reflection", if you will, bad. I like this better.

vortex-scalar/src/bigint/bigcast.rs
connortsui20: Thoughts on `ToPrimitiveExt` instead?

connortsui20: Actually having a `ToPrimitiveExt` coming out of `vortex-scalar` is also super weird to me. In general, I would say that any version of a `ToPrimitive` trait coming out of `vortex-scalar` does not make sense, and I feel like we might as well just call it for what it is, something like `DecimalValueToScalar` or similarly specific.

connortsui20:
Could I get eyes on [3cee582](https://github.com/vortex-data/vortex/pull/4780/commits/3cee582b1c47effadffbb45a7f54e918052d4f44)?

vortex-scalar/src/pvalue.rs
```diff
@@ -385,6 +388,56 @@ impl From<usize> for PValue {
...
+/// The reason we have this (instead of a direct method on [`NativePType`]) is to avoid a circular
+/// dependency between `vortex-scalar` and `vortex-dtype` (`vortex-scalar` depends on
+/// `vortex-dtype`)
+pub trait NativeToPValue: NativePType {
+    /// Convert a native type value to PValue.
+    fn to_pvalue(&self) -> PValue {
+        // Use Any to determine the concrete type at runtime
+        let any_ref = self as &dyn Any;
+
+        if let Some(&val) = any_ref.downcast_ref::<u8>() {
```
danking: We should be able to resolve these conversions at compile-time, not run-time. Why not `pub trait NativeToPValue: NativePType + Into<PValue>`? You could still implement `to_pvalue` as `self.clone().into()`, if you want.

danking: This trait is effectively the primitive analogue of ScalarType, maybe we call it `ScalarPType`?

It seems like this is basically:
```
pub trait ScalarPType: NativePType + Into<PValue> {}
```
And we can even get things like 
```
fn silly<T: ScalarPType>(t: T) -> Scalar {
    Scalar::from(Some(vec![Some(t)]))
 }
```
to work correctly if we impl:
```
impl ScalarType for ScalarPType {
   fn dtype() -> DType { DType::Primitive(Self::PTYPE, NonNullable) }
}
```

danking (changes requested):

vortex-scalar/src/bigint/bigcast.rs
joseph-isaacs: I don't understand why `ToI256` extends ToPrimtive?

encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
```diff
@@ -91,7 +91,7 @@ fn decimal_value_wrapper_to_primitive(
 
 fn decimal_value_to_primitive<P>(decimal_value: DecimalValue) -> Result<ScalarValue, Sign>
 where
-    P: NativePType + NumCast + Bounded + ToPrimitive,
+    P: IntegerPType + ToI256,
```
joseph-isaacs, line 94: Should this be `DecimalValueType`?
```suggestion
    P: DecimalValueType,
```

vortex-scalar/src/pvalue.rs
connortsui20: I'm trying to understand this, are we relying on the `ScalarType` implementation? And are we implementing `ScalarPType` on all of the native ptypes manually?

Also while I agree we should be able to resolve this at compile-time, I think we are also limited by the trait solver, in that we have a bunch of manual implementations of `From<$T> for PValue` that will immediately conflict with a blanket implementation on types bounded by `NativePType`.

connortsui20: Yeah if we try to do this:

```rust
pub trait ScalarPType: NativePType + Into<PValue> {}
```

Then we must also implement one of these two:

```rust
impl<T: NativePType> Into<PValue> for T {}
impl<T: NativePType> From<T> for PValue {}
```

The first one isn't allowed because of the orphan rule, and the second conflicts with our other manual implementations of `From`. So unless I'm misunderstanding something I don't think this works?

encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
connortsui20: Do you mean we should rename the type? `DecimalValueType` already exists in `vortex-scalar` and it is an enum

robert3005 (approved):

vortex-scalar/src/pvalue.rs
connortsui20: Okay I just gave up and reimplemented the list builder `extend_from_array` method

vortex-scalar/src/bigint/bigcast.rs
connortsui20: Is now just a normal trait

encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
joseph-isaacs: I was thinking this should be named `DecimalValueType` (similar to NativePType), but that is taken

danking (approved):

vortex-scalar/src/pvalue.rs
danking: This seems better anyway. I guess we could do something like `impl<T: NativePType + Into<PValue>> ScalarPType for T`? Doesn't matter if we don't need it though.

encodings/decimal-byte-parts/src/decimal_byte_parts/compute/compare.rs
connortsui20: To get this over the line I made an issue that we can address later: https://github.com/vortex-data/vortex/issues/4808

PR #4782, opened 2025-09-28
More pipelined execution

gatesn:
This PR removes the FilterOperator and redefines the "execute" function to be a specific "project(mask)" function. We will soon add a "reduce(group_mask)" function too to support reduction operations like sum, min, max, etc.

What I want to do is model vortex arrays as equivalent to a RecordBatch in DataFusion. A fixed size thing. The novel thing, is that our batches have their own logical plan inside! It’s not a relational plan though.


joseph-isaacs:
Are you suggesting that all operators will have a `reduce(group_mask)` method too? I find that confusing?

Would you mind summarising (on the PR) why we removed the filter operator? I think will lead to a strange API when we add reduce. Could you sketch that out?

PR #4783, opened 2025-09-29
feat[fuzz]: support slow comparison using `make_comparator` for fallback arrow compare

joseph-isaacs:
Support a **very slow** fallback method for comparing recursive vortex/array datatypes.

This uses arrow comparisons logic and is_nullable if either side is.

### DIFF
--- a/vortex-dtype/src/dtype.rs
+++ b/vortex-dtype/src/dtype.rs
@@ -306,12 +306,12 @@ impl DType {
         matches!(self, Extension(_))
     }
 
-    /// Check if `self` is a recursive type, i.e. list, fixed size list, struct, or extension of a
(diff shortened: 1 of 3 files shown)


vortex-dtype/src/dtype.rs
```diff
@@ -306,6 +306,16 @@ impl DType {
         matches!(self, Extension(_))
     }
 
+    /// Check if `self` is a recursive type, i.e. list, fixed size list, struct, or extension of a
+    /// recursive type.
+    pub fn is_recursive(&self) -> bool {
```
AdamGS: maybe `nested` like in arrow?

joseph-isaacs: sure

AdamGS (approved):
LGTM

PR #4785, opened 2025-09-29
Initial version of Masked encoding

### LINKED ISSUE #4784: `take` on fsst explodes memory
I have 19 Mb vortex array and a sparse `take_array`:

[data.zip](https://github.com/user-attachments/files/22598984/data.zip)

When I run `take(data, take_array)` my array turns from 19 Mb into 2 Gb 🫨

AdamGS:
- Closes https://github.com/vortex-data/vortex/issues/4784#issuecomment-3347186341

Introduces a new `MaskedArray` encoding, which is an array that holds a child with no null values and an external validity array, allowing validity-only operations to be very cheap.

The greatest benefit is that together with constant array, it enables us to compress data more aggressively and keep it compact for longer.


vortex-array/src/arrays/masked/compute/compare.rs
```diff
@@ -0,0 +1,137 @@
...
+
+impl CompareKernel for MaskedVTable {
+    fn compare(
+        &self,
+        lhs: &MaskedArray,
+        rhs: &dyn Array,
+        operator: Operator,
+    ) -> VortexResult<Option<ArrayRef>> {
+        use crate::arrays::BoolArray;
+        use crate::canonical::ToCanonical;
```
robert3005: we can lift this

AdamGS: done

AdamGS:
Memory wise, @blaginin confirmed that it solved the issue he was seeing (now has a different one, but its a step forward).

blaginin:
yes! some commands are passing now

vortex-array/src/arrays/masked/compute/take.rs
```diff
@@ -0,0 +1,79 @@
...
+        // The child is guaranteed to have no nulls, so we can only take with non-null indices
+        let stripped_indices = if indices.dtype().is_nullable() {
+            // Create a non-nullable version of the indices by replacing nulls with 0
+            // This is safe because we'll mask out these positions in the validity
+            let prim = indices.to_primitive();
+            match_each_native_ptype!(prim.ptype(), |P| {
+                PrimitiveArray::new(prim.into_buffer::<P>(), Validity::NonNullable).into_array()
+            })
+        } else {
+            indices.to_owned()
```
robert3005: the fact that this does to_owned to later on just `&` is weird

vortex-array/src/arrays/masked/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    fn stats(array: &MaskedArray) -> StatsSetRef<'_> {
+        if array.all_valid() {
+            array.child.statistics()
+        } else {
+            let stats = array.child.statistics();
+            stats.retain(&[]);
+            stats
+        }
+    }
```
robert3005, line 70: This is very weird, especially given you always run it

vortex-array/src/arrays/masked/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+            child,
+            validity,
+            dtype: array.dtype.clone(),
+        }
+        .into_array()
+    }
+
+    fn scalar_at(array: &MaskedArray, index: usize) -> Scalar {
+        if array.validity.is_null(index) {
+            return Scalar::null(array.dtype().as_nullable());
```
robert3005: the dtype should be nullable?

vortex-array/src/arrays/masked/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn nullability(&self) -> Nullability {
+        match self.validity {
+            Validity::NonNullable => Nullability::NonNullable,
+            _ => Nullability::Nullable,
+        }
+    }
+}
+
+impl OperationsVTable<MaskedVTable> for MaskedVTable {
+    fn slice(array: &MaskedArray, range: std::ops::Range<usize>) -> ArrayRef {
```
robert3005: import this

vortex-btrblocks/src/float.rs
```diff
@@ -142,12 +142,19 @@ impl Scheme for ConstantScheme {
...
-            .source()
-            .as_constant()
-            .vortex_expect("must be constant");
-
-        Ok(ConstantArray::new(scalar, stats.source().len()).into_array())
+        let scalar_idx = (0..stats.source().len())
+            .position(|idx| stats.source().is_valid(idx))
+            .vortex_expect("Must have at least one valid value");
+        let scalar = stats.source().scalar_at(scalar_idx);
+        let scalar_is_valid = scalar.is_valid();
```
robert3005: This is redundant

vortex-btrblocks/src/integer.rs
```diff
@@ -185,13 +186,12 @@ impl Scheme for ConstantScheme {
             return Ok(0.0);
         }
 
-        // Only arrays with one distinct values can be constant compressed.
-        if stats.distinct_values_count != 1 {
+        if stats.src.all_invalid() {
             return Ok(0.0);
         }
```
a10y: don't we want to use constant compression for all-invalid?

vortex-array/src/arrays/masked/mod.rs
AdamGS: yeah you're right

vortex-btrblocks/src/integer.rs
AdamGS: seems like you're right

AdamGS: a3a8e6747e0282ef3910611419c3a42d811f0e2c

vortex-btrblocks/src/float.rs
AdamGS: a3a8e6747e0282ef3910611419c3a42d811f0e2c

vortex-array/src/arrays/masked/mod.rs
AdamGS: a3a8e6747e0282ef3910611419c3a42d811f0e2c

vortex-array/src/arrays/masked/mod.rs
AdamGS: a3a8e6747e0282ef3910611419c3a42d811f0e2c

vortex-mask/src/bitops.rs
```diff
@@ -43,6 +43,18 @@ impl BitOr for &Mask {
...
 
+impl Not for Mask {
+    type Output = Mask;
+
+    fn not(self) -> Self::Output {
+        match self.boolean_buffer() {
+            AllOr::All => Mask::new_false(self.len()),
+            AllOr::None => Mask::new_true(self.len()),
+            AllOr::Some(buffer) => Mask::from_buffer(!buffer),
+        }
```
robert3005:
```suggestion
        !&self
```

vortex-array/src/arrays/masked/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        let stats = if validity.all_valid(child.len()) {
+            child.statistics().to_owned()
+        } else {
+            let invalid_count = validity.to_mask(child.len()).false_count();
+            let mut stats = child.statistics().to_owned();
+            stats.retain_only(&[Stat::UncompressedSizeInBytes]);
+            stats.set(Stat::NullCount, Precision::exact(invalid_count));
+            stats
+        };
```
robert3005: this is untrue

vortex-array/src/arrays/masked/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub fn try_new(child: ArrayRef, validity: Validity) -> VortexResult<Self> {
+        if child.valid_count() != child.len() {
+            vortex_bail!("MaskedArray children must not have nulls");
+        }
+
+        if let Validity::Array(arr) = &validity
+            && arr.len() != child.len()
+        {
+            vortex_bail!("Validity must be the same length as a MaskedArray's child");
+        }
```
robert3005: validity.maybe_len()

vortex-array/src/arrays/masked/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    fn scalar_at(array: &MaskedArray, index: usize) -> Scalar {
+        if array.validity.is_null(index) {
+            return Scalar::null(array.dtype().clone());
+        }
+
+        let child_scalar = array.child.scalar_at(index);
+
+        if matches!(array.nullability(), Nullability::Nullable) {
```
robert3005: !matches!(&array.validity, Validity::NonNullable)

robert3005 (approved):
We should change this later to allow nullable children

vortex-array/src/arrays/masked/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+impl ValidityHelper for MaskedArray {
+    fn validity(&self) -> &Validity {
+        &self.validity
+    }
+}
+
+impl MaskedArray {
+    pub fn try_new(child: ArrayRef, validity: Validity) -> VortexResult<Self> {
```
joseph-isaacs, line 80: Why not return the child array is the validity is all true?

robert3005: I think it's a little surprising that you call this method and don't end up with a MaskedArray. The caller should know whether they want it which I think means we should assert here that Validity is not NonNullable

robert3005: https://github.com/vortex-data/vortex/pull/4810/files

PR #4787, opened 2025-09-29
Upgrade to DF 50.1.0

AdamGS:
Ran some benchmarks, now that its released just did `cargo update -p datafusion`


blaginin (approved):
lets wait for the benchmarks but IIRC there is no breaking changes we need to reflect

PR #4788, opened 2025-09-29
fix[duckdb]: handle dict and null correctly

joseph-isaacs:


vortex-duckdb/src/duckdb/value.rs
```diff
@@ -227,7 +231,8 @@ where
     fn from(value: Option<T>) -> Self {
         match value {
             Some(v) => v.into(),
-            None => Value::null(),
+            // TODO(joe): fixme
+            None => todo!(),
```
0ax1: What do we want to do here eventually? Construct a logical type to construct `Value::null` based on T?

vortex-duckdb/src/exporter/bool.rs
```diff
@@ -7,24 +7,31 @@ use vortex::error::VortexResult;
...
 struct BoolExporter {
     array: BoolArray,
-    validity: Mask,
+    validity_mask: Mask,
 }
 
 pub(crate) fn new_exporter(array: &BoolArray) -> VortexResult<Box<dyn ColumnExporter>> {
+    let validity_mask = array.validity_mask();
+    if validity_mask.all_false() {
+        return Ok(invalid::new_exporter(
```
0ax1: invalid or null exporter? It's not invalid to have a validity_mask all false right.

vortex-duckdb/src/exporter/dict.rs
```diff
@@ -27,12 +29,14 @@ struct DictExporter<I: NativePType> {
     value_id: usize,
 }
 
-pub(crate) fn new_exporter(
+pub(crate) fn new_exporter_with_flatten(
```
0ax1, line 32: I think I'd prefer to only keep on `new_exporter` func where we pass in the flatten bool explicitly at all times.

vortex-duckdb/src/exporter/dict.rs
```diff
@@ -27,12 +29,14 @@ struct DictExporter<I: NativePType> {
...
     array: &DictArray,
     cache: &ConversionCache,
+    mut flatten: bool,
```
0ax1, line 36: Add some docs why this argument exists and flattening is needed.

vortex-duckdb/src/exporter/cache.rs
```diff
@@ -16,6 +16,7 @@ use crate::duckdb::Vector;
 #[derive(Default)]
 pub struct ConversionCache {
     pub values_cache: DashMap<usize, (ArrayRef, Arc<Mutex<Vector>>)>,
+    pub canonical_cache: DashMap<usize, (ArrayRef, Canonical)>,
```
0ax1, line 20: Why do we actually store the ArrayRef in the values pair, besides Canonical?

0ax1: Is it to ensure no other array gets allocated for that address? If so let's add that to the docs.

vortex-duckdb/src/exporter/dict.rs
```diff
@@ -41,34 +45,63 @@ pub(crate) fn new_exporter(
         );
     }
 
+    let codes_mask = array.codes().validity_mask();
+
+    match codes_mask {
+        Mask::AllTrue(_) => {}
+        Mask::AllFalse(len) => return Ok(invalid::new_exporter(len, &values_type)),
+        Mask::Values(_) => {
+            // duckdb cannot have a dictionary with validity in the codes, so flatten the array.
```
0ax1: Might be nice to add where validity is applied down the line then.

vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -32,7 +32,13 @@ pub(crate) fn new_exporter(
     array: &FixedSizeListArray,
     cache: &ConversionCache,
 ) -> VortexResult<Box<dyn ColumnExporter>> {
-    let elements_exporter = new_array_exporter(array.elements(), cache)?;
+    let elements_exporter = new_array_exporter_with_flatten(array.elements(), cache, true)?;
```
0ax1, line 35: Same here, voting to have one explicit exporter new. Could consider an enum instead of bool to make this more readable at the call site.

vortex-duckdb/src/exporter/invalid.rs
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex::error::{VortexResult, vortex_ensure};
+
+use crate::duckdb::{LogicalType, Value, Vector};
+use crate::exporter::ColumnExporter;
+
+struct InvalidExporter {
```
0ax1: Invalid or Null?

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -187,40 +200,36 @@ fn new_array_exporter(
...
-    /// Since we're iterating these values anyway, then it's cheaper for us to check it inline.
-    ///
-    /// # Safety
-    ///
-    /// The caller must ensure that `len` is less than or equal to the capacity of this vector.
-    unsafe fn set_validity(&mut self, mask: &Mask, offset: usize, len: usize) -> bool;
-}
-
-impl VectorExt for Vector {
+// pub(crate) trait VectorExt {
```
0ax1: Remove instead of commenting out?

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -235,6 +244,16 @@ impl VectorExt for Vector {
             }
         }
     }
+
+    unsafe fn set_all_true_validity(&mut self, len: usize) {
+        if let Some(validity) = unsafe { self.validity_bitslice_mut(len) } {
```
0ax1, line 239: `assert` that `len` isn't larger than the validity bit slice?

vortex-duckdb/src/exporter/bool.rs
joseph-isaacs: We use all_invalid to mean an all false array, I guess I can change it to that?

vortex-duckdb/src/exporter/cache.rs
joseph-isaacs: There is already a comment above

vortex-duckdb/src/exporter/mod.rs
joseph-isaacs: I don't understand, you are saying check if the function does return the bitslice of the len I asked for?

vortex-duckdb/src/exporter/invalid.rs
joseph-isaacs: AllInvalid?

vortex-duckdb/src/exporter/bool.rs
0ax1: Let's go with all invalid then.

vortex-duckdb/src/exporter/mod.rs
0ax1: read this wrong => ignore

0ax1 (approved):

PR #4789, opened 2025-09-29
feat: wire in fastlanes rle to btrblocks

0ax1:
As part of this PR, validity is moved into the indices array instead of being a separate array.


vortex-btrblocks/src/integer.rs
```diff
@@ -100,6 +103,7 @@ const SPARSE_SCHEME: IntCode = IntCode(5);
 const DICT_SCHEME: IntCode = IntCode(6);
 const RUNEND_SCHEME: IntCode = IntCode(7);
 const SEQUENCE_SCHEME: IntCode = IntCode(8);
+const RLE_SCHEME: IntCode = IntCode(9);
```
joseph-isaacs: We have run end and rle? These mean the same thing. This is fast lanes-rle?

0ax1: On the other hand, this is not how we prefixed the other FastLanes schemes `FORScheme` or `BitPackingScheme` which are wired into BtrBlocks.

vortex-btrblocks/src/integer.rs
```diff
@@ -726,6 +736,53 @@ impl Scheme for SequenceScheme {
...
+
+    fn compress(
+        &self,
+        stats: &IntegerStats,
+        _is_sample: bool,
+        _allowed_cascading: usize,
+        _excludes: &[IntCode],
+    ) -> VortexResult<ArrayRef> {
+        let rle_encoded = RLEArray::encode(&stats.src)?;
+        Ok(rle_encoded.into_array())
```
robert3005: I think you want to recursively encode children here

encodings/fastlanes/src/rle/compute/cast.rs
```diff
@@ -0,0 +1,138 @@
...
+use vortex_array::vtable::ValidityHelper;
+use vortex_array::{ArrayRef, IntoArray, register_kernel};
+use vortex_dtype::DType;
+use vortex_error::VortexResult;
+
+use crate::rle::{RLEArray, RLEVTable};
+
+impl CastKernel for RLEVTable {
+    fn cast(&self, array: &RLEArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        if dtype.is_unsigned_int() {
```
robert3005: why can't we cast unsigned ints here?

encodings/fastlanes/src/rle/compute/cast.rs
```diff
@@ -0,0 +1,138 @@
...
+            return Ok(None);
+        }
+
+        // Cast RLE values. Indices and offsets stay the same.
+        let casted_values = cast(array.values(), dtype)?;
+
+        Ok(Some(
+            RLEArray::try_new(
+                casted_values,
+                array.indices().to_array(),
```
robert3005: Ah I missed this part, you want `indices()` to return `&ArrayRef` so this case is cheap instead of creating new array

robert3005: the methods returning rle children should also be pub (and not pub(crate)).

encodings/fastlanes/src/rle/compute/cast.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn cast(&self, array: &RLEArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        if dtype.is_unsigned_int() {
+            return Ok(None);
+        }
+
+        // Cast RLE values. Indices and offsets stay the same.
+        let casted_values = cast(array.values(), dtype)?;
+
+        Ok(Some(
+            RLEArray::try_new(
```
robert3005: we also want a variant of this method that's unsafe and doesn't perform validation.

robert3005 (changes requested):
There's some small changes required to the RLE array. We also need to compress children.

encodings/fastlanes/src/rle/compute/cast.rs
0ax1: If we try to cast to unsigned int this should be a no-op. Though I have to double check how that works out with us casting the RLE on encode and back on decode if the int is signed and non-negative.

encodings/fastlanes/src/rle/compute/cast.rs
0ax1: We have `new_unchecked` for RLE. Is that what you meant?

encodings/fastlanes/src/rle/compute/cast.rs
0ax1: This should not be None in any case!

0ax1: I think this should be along the lines what we do in FoR: 
```
        if !dtype.is_int() {
            return Ok(None);
        }
```
as the type can only be signed or unsigned int.

encodings/fastlanes/src/rle/compute/cast.rs
robert3005: yes, we should use it here

0ax1: done

encodings/fastlanes/src/rle/compute/cast.rs
0ax1: done

vortex-btrblocks/src/integer.rs
0ax1: done

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -157,6 +176,7 @@ impl RLEArray {
         dtype: DType,
         offset: usize,
         length: usize,
+        ptype: PType,
```
robert3005: if you're taking ptype you shouldn't be taking dtype

vortex-file/src/lib.rs
```diff
@@ -171,6 +171,7 @@ pub static DEFAULT_REGISTRY: LazyLock<Arc<ArrayRegistry>> = LazyLock::new(|| {
         EncodingRef::new_ref(FSSTEncoding.as_ref()),
         EncodingRef::new_ref(FoREncoding.as_ref()),
         EncodingRef::new_ref(PcoEncoding.as_ref()),
+        EncodingRef::new_ref(RLEEncoding.as_ref()),
```
joseph-isaacs, line 174: `RunLen`/`RunLength`

encodings/fastlanes/src/rle/mod.rs
robert3005: DType is the decompressed dtype and we don't need to store the ptype. The underlying array is always the unsigned variant of the dtype

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -49,6 +50,10 @@ impl VTable for RLEVTable {
 #[derive(Clone, Debug)]
 pub struct RLEArray {
```
joseph-isaacs, line 49: can you add a doc str for this array?

encodings/fastlanes/src/rle/compute/cast.rs
```diff
@@ -0,0 +1,142 @@
...
+
+impl CastKernel for RLEVTable {
+    fn cast(&self, array: &RLEArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        // RLE only supports integer types.
+        if !dtype.is_int() {
+            return Ok(None);
+        }
+
+        // Cast RLE values. Indices and offsets stay the same.
+        let casted_values = cast(array.values(), dtype)?;
```
robert3005, line 14: this should be the cast to the unsigned type

encodings/fastlanes/src/rle/mod.rs
joseph-isaacs: Can we explain what types we currently and can ever support

0ax1: The support matches what we check for in `validate` => any primitive type.

encodings/fastlanes/src/rle/mod.rs
0ax1: not an issue anymore, as rle now handles any ptype

vortex-file/src/lib.rs
0ax1: the rest of the repo currently also uses rle instead of runlength. i think i'd like to keep rle to keep a short name. i think its more that `RunEndEncoding` is an outlier, as others are abbreviated, e.g. `FoREncoding`.

encodings/fastlanes/src/rle/mod.rs
joseph-isaacs, line 54: Shall we put the validity in this indices?

vortex-btrblocks/src/integer.rs
```diff
@@ -732,6 +742,95 @@ impl Scheme for SequenceScheme {
...
+                compressed_values,
+                compressed_indices,
+                compressed_offsets,
+                rle_array.validity().clone(),
+                rle_array.dtype().clone(),
+                rle_array.offset(),
+                rle_array.len(),
+            )
+            .into_array())
+        }
```
joseph-isaacs: can we share this with the float compressor, extract a func and make it generic over the value compressor?

0ax1: done

encodings/fastlanes/src/rle/mod.rs
0ax1: let's keep as is.

joseph-isaacs: why?

0ax1: alright, alright

robert3005: Can you elaborate the conclusion of this discussion

0ax1: I fleshed this out in the meantime on a branch and actually concluded that wiring in validity into the indices is worse. Besides more code complexity, this also requires validity getting padded to 1024 chunks which is not the case if stored separately. 

see: https://github.com/vortex-data/vortex/commit/2ff060383768f0435ee9f9e468257ba23facde04

joseph-isaacs: I think the big reason to do this is that if there is a lot of null elements (indices) the underlying array need not store these elements. 

If you had a chunk of 1024 null values that whole bitpacked chunk could be omitted. Not that we do this now

vortex-btrblocks/src/integer.rs
joseph-isaacs, line 1: do we ever trigger this compressor?

encodings/fastlanes/src/rle/mod.rs
0ax1: wdyt @robert3005 ?

joseph-isaacs: Also since the indices are 1k aligned for the pipeline and GPU world it makes sense for the validity is too.

vortex-btrblocks/src/integer.rs
0ax1: Currently, it is picked for Clickbench `hits_0.vortex`.

encodings/fastlanes/src/rle/mod.rs
robert3005: I think alignment is a separate issue, we can always align the underlying buffers however we want. The question is whether we can reduce validity with indices compression. I’m not sure how this works in practice. Skipping invalid blocks can happen irrespective of where the validity lives and I think it would have to live at level of rle array anyway

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -90,18 +94,37 @@ where
...
+        Validity::Array(validity_array) => {
+            let padded_length = array.len().next_multiple_of(FL_CHUNK_SIZE);
+            let bool_array = validity_array.to_bool();
+            let bool_buffer = bool_array.boolean_buffer();
+            let padded_validity_iter = bool_buffer
+                .iter()
+                .take(array.len())
+                .chain(std::iter::repeat_n(false, padded_length - array.len()));
+
+            Validity::from_iter(padded_validity_iter)
```
robert3005: from_iter is cute for tests but we can't do that here. Allocate a new buffer, copy existing validity and zero the trailer

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -144,7 +168,7 @@ where
         range.finish();
     }
 
-    let offset_within_chunk = array.offset_in_chunk(array.offset);
+    let offset_within_chunk = array.offset();
     let primitive_array = builder.finish_into_primitive();
```
robert3005: If all you're doing with this builder is just set values and not validity you can simplify the logic by creating a BufferMut<T>, the builder lets you associate validity as well but in this case we need to mutate the built array to slice values at the end

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -162,59 +160,57 @@ impl RLEArray {
...
-    /// Start index offset in the values array for a given chunk.
+    /// Values index offset relative to the first chunk.
+    ///
+    /// Offsets in the `values_idx_offsets` arrays are absolute and need to be
+    /// shifted by the offset of the first chunk, respective the current slice,
+    /// in order to make them relative.
     #[allow(clippy::expect_used)]
-    pub(crate) fn value_chunk_offset(&self, chunk_idx: usize) -> usize {
-        self.value_chunk_offsets
+    pub(crate) fn values_idx_offset(&self, chunk_idx: usize) -> usize {
```
robert3005, line 197: I think this method is fine as a one off but in decompression you're using it in the loop. In that case you should canonicalize the value_idx_offsets instead of going through scalar at

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -238,53 +234,79 @@ impl CanonicalVTable<RLEVTable> for RLEVTable {
     }
 }
 
+impl ValidityVTable<RLEVTable> for RLEVTable {
+    fn is_valid(array: &RLEArray, index: usize) -> bool {
+        // Indices get padded to 1024 chunks. Ensure the index is within the array's length.
+        assert!(index < array.len());
```
robert3005: this is already asserted for you at the entry point

encodings/fastlanes/src/rle/mod.rs
```diff
@@ -115,25 +112,26 @@ impl RLEArray {
...
+    /// * `values_idx_offsets` - Start indices for each value chunk.
     /// * `validity` - Array validity
     /// * `length` - Array length
     pub fn try_new(
         values: ArrayRef,
         indices: ArrayRef,
-        value_chunk_offsets: ArrayRef,
-        validity: Validity,
+        values_idx_offsets: ArrayRef,
         length: usize,
```
robert3005, line 128: we need to validate that length is shorter than indices length and is less than 1024 away

encodings/fastlanes/src/rle/ops.rs
```diff
@@ -8,51 +8,66 @@ use vortex_array::{ArrayRef, IntoArray};
...
     fn scalar_at(array: &RLEArray, index: usize) -> Scalar {
-        // In case of `slice`, the validity is sliced when creating a new array.
-        // Therefore, we can use the slice local index when checking for the
-        // scalar's validity.
-        if !array.validity.is_valid(index) {
-            return Scalar::null(array.dtype.clone());
+        let offset_in_chunk = array.offset();
+
+        if !array.indices().is_valid(offset_in_chunk + index) {
+            return Scalar::null(array.dtype().clone());
```
robert3005: this is already handled for you in the entry point

vortex-btrblocks/src/rle.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        let compressed_values = (self.compress_values_fn)(
+            &rle_array.values().to_primitive(),
+            is_sample,
+            allowed_cascading - 1,
+            &new_excludes,
+        )?;
+
+        // For indices and offsets, we always use integer compression without dictionary encoding.
+        let compressed_indices = crate::integer::IntCompressor::compress_no_dict(
```
robert3005: can we import the IntCompressor?

vortex-btrblocks/src/rle.rs
```diff
@@ -0,0 +1,142 @@
...
+            &rle_array.indices().to_primitive(),
```
robert3005: call `downcast` here

vortex-btrblocks/src/rle.rs
```diff
@@ -0,0 +1,142 @@
...
+            is_sample,
+            allowed_cascading - 1,
+            &[],
+        )?;
+
+        let compressed_offsets = crate::integer::IntCompressor::compress_no_dict(
+            &rle_array.values_idx_offsets().to_primitive(),
```
robert3005: call downcast

vortex-btrblocks/src/rle.rs
```diff
@@ -0,0 +1,143 @@
...
+            code,
+            compress_values_fn,
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<S, C> Scheme for RLEScheme<S, C>
+where
+    S: RLEStats + crate::CompressorStats,
```
robert3005: import

vortex-btrblocks/src/integer.rs
```diff
@@ -861,4 +873,20 @@ mod tests {
         let decoded = compressed.to_primitive();
         assert_eq!(decoded.as_slice::<i32>(), values.as_slice());
     }
+
+    #[test]
+    fn test_rle_compression() {
+        let mut values = Vec::new();
+        values.extend(std::iter::repeat_n(42i32, 100));
```
robert3005, line 885: I would import change this to `iter::repeat_n`, i.e. import std::iter

vortex-btrblocks/src/float.rs
```diff
@@ -402,4 +412,19 @@ mod tests {
         let compressed = FloatCompressor::compress(&floats, false, MAX_CASCADE, &[]).unwrap();
         println!("compressed: {}", compressed.display_tree())
     }
+
+    #[test]
+    fn test_rle_compression() {
+        let mut values = Vec::new();
+        values.extend(std::iter::repeat_n(1.5f32, 100));
```
robert3005: import std::iter

encodings/fastlanes/src/rle/ops.rs
```diff
@@ -8,51 +8,66 @@ use vortex_array::{ArrayRef, IntoArray};
...
-                array.validity.slice(range),
+                sliced_values,
+                sliced_indices,
+                sliced_value_chunk_offsets,
                 array.dtype.clone(),
-                array.offset + start,
-                length,
+                // Keep the offset relative to the first chunk.
+                (array.offset + range.start) % FL_CHUNK_SIZE,
+                range.end - range.start,
```
robert3005: range.len()?

encodings/fastlanes/src/rle/compress.rs
```diff
@@ -1,48 +1,52 @@
...
     pub fn encode(array: &PrimitiveArray) -> VortexResult<Self> {
-        match_each_unsigned_integer_ptype!(array.ptype(), |T| { rle_encode_typed::<T>(array) })
+        match_each_native_ptype!(array.ptype(), |T| { rle_encode_typed::<T>(array) })
     }
 }
 
 /// Decompresses an RLE array back into a primitive array.
 pub fn rle_decompress(array: &RLEArray) -> PrimitiveArray {
-    match_each_unsigned_integer_ptype!(array.ptype(), |T| { rle_decode_typed::<T>(array) })
+    let decompressed = match_each_native_ptype!(array.values().dtype().as_ptype(), |T| {
```
robert3005: I would inline this

robert3005 (changes requested):
There's some style/performance changes. I think the most important are copying the validity in a reasonable way and iterating offsets

joseph-isaacs:
Its worth looking at the compression benchmark. To see if the files have changed size at all.

0ax1:
> Its worth looking at the compression benchmark. To see if the files have changed size at all.

There's an alignment issue right now to solve before we look into perf / size changes: https://github.com/vortex-data/vortex/actions/runs/18228843363/job/51906973673?pr=4789#step:9:102

robert3005:
When we added downcast to indices it means you have to serialize the indices ptype and can't assume it's u16

robert3005 (approved):

PR #4794, opened 2025-09-30
Try using profiling custom labels

AdamGS:


bench-vortex/src/benchmark_driver.rs
```diff
@@ -136,17 +138,26 @@ fn execute_queries<B: Benchmark>(
     let expected_row_counts = benchmark.expected_row_counts();
 
     for &(query_idx, ref query_string) in queries.iter() {
+        let mut label_set = Labelset::new();
```
0ax1, line 141: Where can I see the labels in the UI?

0ax1: As discussed offline, not yet. 🙂

AdamGS: Expanding a bit - I'm not 100% sure why. For datafusion it might be because of the threading model, but that seems easily solved with the [`JoinSetTracer`](https://docs.rs/datafusion/50.0.0/datafusion/common/runtime/trait.JoinSetTracer.html)

PR #4795, opened 2025-09-30
Update parca agent in CI

AdamGS:
1. Make sure we use the latest one
2. `off-cpu-threshold` is now an arbitrary fraction instead of a number between in [0, 1000], so this is just keeping it as the same value.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -71,7 +71,7 @@ jobs:
         with:
           polarsignals_cloud_token: ${{ secrets.POLAR_SIGNALS_API_KEY }}
           labels: "branch=${{ github.ref_name }};gh_run_id=${{ github.run_id }};benchmark=${{ matrix.benchmark.id }}"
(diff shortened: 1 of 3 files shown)


lwwmanning (approved):

.github/workflows/bench.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         with:
           polarsignals_cloud_token: ${{ secrets.POLAR_SIGNALS_API_KEY }}
           labels: "branch=${{ github.ref_name }};gh_run_id=${{ github.run_id }};benchmark=${{ matrix.benchmark.id }}"
-          parca_agent_version: "0.39.3"
+          parca_agent_version: "0.42.0"
           project_uuid: "e5d846e1-b54c-46e7-9174-8bf055a3af56"
           extra_args: "--off-cpu-threshold=1"  # Personally tuned by @brancz
```
robert3005, line 76: you also need to change this?

PR #4797, opened 2025-09-30
chore: zip struct per field

blaginin:
follow up on https://github.com/vortex-data/vortex/pull/4747


vortex-array/src/compute/zip.rs
```diff
@@ -193,7 +193,21 @@ pub(crate) fn zip_return_dtype(if_true: &dyn Array, if_false: &dyn Array) -> DTy
...
+
+    let builder = match return_type {
+        DType::Utf8(n) => Box::new(VarBinViewBuilder::with_buffer_deduplication(
+            DType::Utf8(n),
+            capacity,
+        )),
+        DType::Binary(n) => Box::new(VarBinViewBuilder::with_buffer_deduplication(
+            DType::Binary(n),
+            capacity,
+        )),
```
blaginin, line 215: @onursatici was right, we need to do it here otherwise we'll still get memory explosion in some edge cases (like with nested structs)

vortex-array/src/compute/zip.rs
```diff
@@ -193,7 +193,21 @@ pub(crate) fn zip_return_dtype(if_true: &dyn Array, if_false: &dyn Array) -> DTy
 
 fn zip_impl(if_true: &dyn Array, if_false: &dyn Array, mask: &Mask) -> VortexResult<ArrayRef> {
     // if_true.len() == if_false.len() from ComputeFn::invoke
```
joseph-isaacs: debug assert?

vortex-array/src/arrays/struct_/compute/zip.rs
```diff
@@ -0,0 +1,59 @@
...
+        &self,
+        if_true: &StructArray,
+        if_false: &dyn Array,
+        mask: &Mask,
+    ) -> VortexResult<Option<ArrayRef>> {
+        let Some(if_false) = if_false.as_opt::<StructVTable>() else {
+            return Ok(None);
+        };
+
+        assert_eq!(if_true.names(), if_false.names()); // this is properly checked above
```
joseph-isaacs: debug assert then? or I don't follow the comment

vortex-array/src/compute/zip.rs
joseph-isaacs, line 215: it worth explaining why you compress when you zip?

vortex-array/src/arrays/struct_/compute/zip.rs
```diff
@@ -0,0 +1,153 @@
...
+        ┌───────┐
+        │ field │
+        ├───────┤
+        │ null  │
+        ├───────┤
+        │ 20i32 │
+        ├───────┤
+        │ 3i32  │
+        ├───────┤
+        │ 40i32 │
```
blaginin, line 112: what do you think about check results this way?

vortex-array/src/arrays/struct_/compute/zip.rs
blaginin: i feel it should be a regular assert because it's cheap. If it's wrong it'll fail later anyway.. I rewrote the comment, what do you think?

vortex-array/src/compute/zip.rs
blaginin: sure, added!

vortex-array/src/arrays/struct_/compute/zip.rs
joseph-isaacs: I still don't understand the comment?

joseph-isaacs: Are you saying this is checked above?

blaginin: yes

onursatici (commented):
lgtm, deferring to @joseph-isaacs for approving in case he has more comments

vortex-array/src/arrays/struct_/compute/zip.rs
joseph-isaacs, line 26:
```suggestion
        assert_eq!(if_true.names(), if_false.names()); // this is checked in the entry point function
```
Where is this checked?

blaginin: i think you're seeing the outdated version? 

image

robert3005 (approved):

PR #4799, opened 2025-09-30
Implement filter for SequenceArray

AdamGS:
Like take, this potentially has much temporary memory footprint.


a10y (approved):
nice to avoid the intermediate allocation

PR #4804, opened 2025-10-01
perf[duckdb]: zero-copy primitive exporter

joseph-isaacs:


vortex-duckdb/src/e2e_test/vortex_scan_test.rs
```diff
@@ -441,7 +441,6 @@ fn test_vortex_scan_fixed_size_list_utf8() {
 }
 
 #[test]
-#[ignore = "Assertion fail in DuckDB (`child.GetVectorType() == VectorType::FLAT_VECTOR`) with debug"]
```
0ax1, line 444: 👍

0ax1 (approved):
Cursed but gives a good speedup.

PR #4810, opened 2025-10-01
chore: MaskedArray requires a nullable validity

robert3005:

### DIFF
--- a/vortex-array/src/arrays/masked/mod.rs
+++ b/vortex-array/src/arrays/masked/mod.rs
@@ -157,7 +157,6 @@ mod tests {
     use crate::{Array, IntoArray, ToCanonical as _};
 
     #[rstest]
-    #[case(Validity::NonNullable, Nullability::NonNullable)]
     #[case(Validity::AllValid, Nullability::Nullable)]
     #[case(Validity::from_iter([true, false, true]), Nullability::Nullable)]
     fn test_dtype_nullability(#[case] validity: Validity, #[case] expected: Nullability) {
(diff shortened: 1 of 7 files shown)


AdamGS:
where does it fail?

robert3005:
I thought it's weird from the api point of view that you'd produce a maskedarray with nonnullable validity... just why

vortex-array/src/arrays/masked/mod.rs
joseph-isaacs, line 1: Why not make this require a Mask validity?

robert3005: I think you have more flexibility by passing a Validity here. There's something weird where if you know you're going to do computation you want to convert to a Mask but otherwise you would want to keep a Validity

joseph-isaacs: I am saying that its only really worth creating one of these if validity variant is `Validity::Array`?

robert3005: But the current structure is that it's weird/hard to take Mask. You will convert validity to mask and then convert it back here to Validity. I think you're right though that there's something missing in our apis

joseph-isaacs: I think you might want to make a Boolean typed array (not a mask)

AdamGS (approved):
I don't feel strongly about this either way, seems very reasonable.

vortex-array/src/arrays/masked/mod.rs
robert3005: I have wrote something that separates these two concerns but it's hard to avoid changing type of the child. Logically it makes sense that MaskedArray can only be constructed with Array validity but we need some easy way to wrap an array with validity that constructs masked array when necessary. However, I think it's simpler to say that you delegate appropriately in MaskedArray instead of unwrapping. I am happy to revisit this in the future since the unwrapping behaviour might be generally useful

PR #4812, opened 2025-10-01
perf[duckdb]: zero-copy decimal exporter

joseph-isaacs:


vortex-duckdb/src/duckdb/logical_type.rs
joseph-isaacs, line 1: this was just in the wrong place

0ax1 (approved):

PR #4814, opened 2025-10-01
Fallback to ArrayBuilder in `ChunkedArray::take`

AdamGS:
The existing code assumes that the take indices are sorted, so it can create up to `indices` chunks which will worsen performance for any subsequent use of the array.

### DIFF
--- a/vortex-array/src/arrays/chunked/compute/take.rs
+++ b/vortex-array/src/arrays/chunked/compute/take.rs
@@ -7,9 +7,16 @@ use vortex_error::VortexResult;
 
 use crate::arrays::chunked::ChunkedArray;
 use crate::arrays::{ChunkedVTable, PrimitiveArray};
+use crate::builders::{ArrayBuilder, builder_with_capacity};
 use crate::compute::{TakeKernel, TakeKernelAdapter, cast, take};
 use crate::validity::Validity;
 use crate::{Array, ArrayRef, IntoArray, ToCanonical, register_kernel};
+
+/// The multiplier for determining the maximum number of resulting chunks before we fall back
+/// to using an ArrayBuilder. The threshold is calculated as initial_chunks * CHUNK_GROWTH_MULTIPLIER.
+/// When the indices are unsorted, the number of chunks can grow to be as large as the number
+/// of indices (worst case), which can lead to poor performance.
+const CHUNK_GROWTH_MULTIPLIER: usize = 2;
 
 impl TakeKernel for ChunkedVTable {
     fn take(&self, array: &ChunkedArray, indices: &dyn Array) -> VortexResult<ArrayRef> {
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/chunked/compute/take.rs
```diff
@@ -24,26 +31,46 @@ impl TakeKernel for ChunkedVTable {
...
             let idx = usize::try_from(*idx)?;
             let (chunk_idx, idx_in_chunk) = array.find_chunk_idx(idx);
 
             if chunk_idx != prev_chunk_idx {
-                // Start a new chunk
+                // Check if we're exceeding the threshold
+                if chunks.len() >= max_chunks_threshold && builder.is_none() {
+                    // Initialize builder and canonicalize all existing chunks
+                    let mut new_builder = builder_with_capacity(
+                        &array.dtype().clone().union_nullability(nullability),
```
joseph-isaacs, line 52: are you concerned about creating very long arrays. maybe we should try and keep the lengths of these compacted array a similar length to the original chunks.

joseph-isaacs (commented):
Originally this was not incorrect, just inefficient

vortex-array/src/arrays/chunked/compute/take.rs
AdamGS: Why are very long arrays necessarily bad?

joseph-isaacs: I don't have a good reason, just seems like an array would be chunked for a reason?

vortex-array/src/arrays/chunked/compute/take.rs
```diff
@@ -7,10 +7,17 @@ use vortex_error::VortexResult;
...
+use crate::builders::{ArrayBuilder, builder_with_capacity};
 use crate::compute::{TakeKernel, TakeKernelAdapter, cast, take};
 use crate::validity::Validity;
 use crate::{Array, ArrayRef, IntoArray, ToCanonical, register_kernel};
 
+/// The multiplier for determining the maximum number of resulting chunks before we fall back
+/// to using an ArrayBuilder. The threshold is calculated as initial_chunks * CHUNK_GROWTH_MULTIPLIER.
+/// When the indices are unsorted, the number of chunks can grow to be as large as the number
+/// of indices (worst case), which can lead to poor performance.
+const CHUNK_GROWTH_MULTIPLIER: usize = 2;
```
joseph-isaacs, line 19: not: It's not really about sortedness, it's about shuffling chunks.

joseph-isaacs, line 19: 4? or more. Two seems a little small. 

The double copy is expensive.

vortex-array/src/arrays/chunked/compute/take.rs
```diff
@@ -24,26 +31,46 @@ impl TakeKernel for ChunkedVTable {
...
                     Validity::from_mask(indices_mask.slice(start..stop), nullability),
                 );
-                chunks.push(take(
-                    array.chunk(prev_chunk_idx),
-                    indices_in_chunk_array.as_ref(),
-                )?);
+                let taken = take(array.chunk(prev_chunk_idx), indices_in_chunk_array.as_ref())?;
+
+                if let Some(ref mut b) = builder {
+                    b.extend_from_array(taken.as_ref());
```
joseph-isaacs, line 69: shame we don't have a take_into method.

robert3005: We deleted it at some point, we should add it back

blaginin:
Retested `take` [example](https://github.com/vortex-data/vortex/issues/4784#issuecomment-3352977814) on this branch 

in develop: 
`take` takes 0.47s and produces 9Gb (9097410576 bytes)

this branch:
`take` takes 3.76s and produces 32Gb (32730252921 bytes)

AdamGS:
This issue seem to be part of https://github.com/vortex-data/vortex/pull/4814#issuecomment-3360072564, might need some bigger rethinking about this usecase.

robert3005:
I thought this would be an improvement but since you don't actually decompress the data in your workflow this is just adding decompression overhead

PR #4818, opened 2025-10-01
Chore: `vortex-array` modules cleanup

connortsui20:
The way in which the modules are structured under `vortex-array/src/arrays` is quite inconsistent. This PR ensures that each kind of array consistently has the same structure, which is:

- `array.rs`: contains the array struct definition and implementation
- `compute`: contains any compute function implementations that the encoding implements in separate files
- `vtable`: contains the array vtable implementations, each in a separate file. The vtable itself is located in `vtable/mod.rs`

Probably the most controversial thing here is the addition of a `vtable` module, where each helper trait implementation is in its own file. I agree that it isn't ideal as it means we have a lot of files named the same thing, but at the same time I feel it is far worse to have every single module place their `vtable` implementations in slightly different files (it becomes _incredibly_ hard to find where things are, even with a language server).

If we really care that the `ArrayVTable` and the `OperationsVTable` should be located in the same file, then I would argue that those traits should be combined.

I also snuck in a change (first commit) where we validate in `new_unchecked` in debug mode only.

Note: All modules have files moved around in the exact same way with the exception of `primitive`, where I create an `array` module instead of an `array.rs` because there are so many extra things we implement specifically for `PrimitiveArray`.


connortsui20:
@claude review

danking (approved):
This is great. My only difference of opinion is that I dislike "tests.rs" files. I prefer the tests at the bottom of the file with the relevant functions. I guess when I'm reading a function I'm both interested in what it does but also why I should believe it does that correctly. If the tests are not in-line, the file doesn't serve as a boundary within which I search for evidence. Instead, the whole module or the whole repo must be searched for the relevant tests.

robert3005:
Agree with @danking on the test part. Otherwise we can merge this

connortsui20:
Discussed briefly offline, since the tests are not really specific to a single file (most tests will test several things that are split across the vtable trait, compute functions, and method implementations), for now we'll just keep it in `tests.rs`.

I will mention that I feel the tests _should_ be a bit more targeted to specific things (sort of how I structured the tests in `fixed_size_list` and `list_view`), but I think for now this is good enough(TM).

PR #4819, opened 2025-10-01
Feature: add listview rebuild + cleanup

connortsui20:

Adds a `rebuild` method on `ListViewArray` that allows us to rebuild the array to specific shapes that we want (via a `ListViewRebuildMode` enum).

This will make it easier to implement an optimization (that we are tentatively calling `IsZeroCopyToList`) that tells us the `ListViewArray` has no gaps nor overlaps from the views into the `elements` buffer, and that the views are all sorted and in order.

I also added an unimplemented `OverlapCompression` rebuild mode if we ever want to do that.

Also does some cleanup of other `ListView` functionality.


vortex-array/src/arrays/listview/compute/filter.rs
```diff
@@ -42,10 +53,18 @@ impl FilterKernel for ListViewVTable {
...
+        let mut new_array = unsafe {
             ListViewArray::new_unchecked(elements.clone(), new_offsets, new_sizes, new_validity)
+        };
+
+        // TODO(connor)[ListView]: IsZeroCopyToList optimization.
+
+        // TODO(connor)[ListView]: Figure out a better heuristic.
+        if selection_mask.density() < REBUILD_DENSITY_THRESHOLD {
+            new_array = new_array.rebuild(ListViewRebuildMode::RemoveGaps);
         }
```
joseph-isaacs: Can we create a follow up to benchmark this

vortex-array/src/arrays/listview/array.rs
```diff
@@ -0,0 +1,429 @@
...
+            size_max,
+            offset_ptype,
+            offset_max
+        );
+
+        // Validate the `offsets` and `sizes` arrays.
+        match_each_integer_ptype!(offset_ptype, |O| {
+            match_each_integer_ptype!(size_ptype, |S| {
+                let offsets_primitive = offsets.to_primitive();
+                let sizes_primitive = sizes.to_primitive();
```
joseph-isaacs: want to pull this out of the match_each

vortex-array/src/arrays/listview/array.rs
```diff
@@ -0,0 +1,429 @@
...
+            size_max <= offset_max,
+            "size type {:?} (max {}) must fit within offset type {:?} (max {})",
+            size_ptype,
+            size_max,
+            offset_ptype,
+            offset_max
+        );
+
+        // Validate the `offsets` and `sizes` arrays.
+        match_each_integer_ptype!(offset_ptype, |O| {
```
joseph-isaacs, line 225: can we add a todo to validate this without a canoncalize?

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -0,0 +1,398 @@
...
+            }
+        }
+
+        // Fast path: if all elements are referenced, no garbage collection needed.
+        if referenced_elements
+            .iter()
+            .all(|&is_referenced| is_referenced)
+        {
+            return self.clone();
+        }
```
joseph-isaacs, line 125: can you merge this with the previous loop?

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        let offset_ptype = self.offsets().dtype().as_ptype();
+        let size_ptype = self.sizes().dtype().as_ptype();
+
+        match_each_integer_ptype!(offset_ptype, |O| {
+            match_each_integer_ptype!(size_ptype, |S| { self.rebuild_remove_gaps_inner::<O, S>() })
+        })
+    }
```
joseph-isaacs, line 97: Did you benchmark this is seems very slow

vortex-array/src/arrays/listview/compute/filter.rs
```diff
@@ -42,10 +53,18 @@ impl FilterKernel for ListViewVTable {
...
-        Ok(unsafe {
+        let mut new_array = unsafe {
             ListViewArray::new_unchecked(elements.clone(), new_offsets, new_sizes, new_validity)
+        };
+
+        // TODO(connor)[ListView]: IsZeroCopyToList optimization.
+
+        // TODO(connor)[ListView]: Figure out a better heuristic.
+        if selection_mask.density() < REBUILD_DENSITY_THRESHOLD {
+            new_array = new_array.rebuild(ListViewRebuildMode::RemoveGaps);
```
joseph-isaacs: Do you think just adding everything to a builder (the other rebuild might be faster)?

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        // Copy only the referenced elements into a new compacted elements array.
+        let mut elements_builder = builder_with_capacity(self.elements().dtype(), cumulative_sum);
+        for (i, &is_referenced) in referenced_elements.iter().enumerate() {
+            if is_referenced {
+                elements_builder
+                    .append_scalar(&self.elements().scalar_at(i))
+                    .vortex_expect("append scalar");
+            }
+        }
```
joseph-isaacs, line 146: Can we merge these loops? Do you think the compiler would in-fact fuse?

joseph-isaacs (commented):
Nice approach

vortex-array/src/arrays/listview/compute/filter.rs
connortsui20: I'm going to remove this and we can do that after A) the `IsZeroCopyToList` optimization and B) listview is canonicalized

vortex-array/src/arrays/listview/array.rs
connortsui20: I think it is basically always more efficient to validate after canonicalizing? Or put another way, if we want to add `offsets + sizes` arrays then we basically always need to canonicalize the bitpacked arrays

vortex-array/src/arrays/listview/rebuild.rs
connortsui20: Don't think you can do that efficiently? Unless you used some sort of heap data structure...

vortex-array/src/arrays/listview/rebuild.rs
connortsui20: It probably is slow, but the issue is that the alternatives are generally data dependent or much more complex:

- We could track ranges in a segment-tree-like data structure, but that is only good when the gaps are large
- We could do a union find thing for ranges but that has the same problem as the previous one
- We could use a bitset, which would definitely be more efficient if we implement a complicated prefix summation of a bitvector, and I'm not actually sure it would be much faster (but definitely several orders of magnitude more complicated)
- Roaring bitmaps? Even more complicated...

vortex-array/src/arrays/listview/compute/filter.rs
connortsui20: Yeah that is the `MakeZeroCopyToList` mode. You're probably right (though for this PR I'm remove this)

vortex-array/src/arrays/listview/rebuild.rs
connortsui20: It wouldn't fuse because the group is dependent on `cumulative_sum` becoming the capacity of the builder. Maybe we could make the capacity of the builder 0 since it can keep growing? Personally I would just let the compiler figure this out.

vortex-array/src/arrays/listview/array.rs
joseph-isaacs: You could look at the specific encoding used (e.g. Delta) and do less work?

vortex-array/src/arrays/listview/rebuild.rs
joseph-isaacs: Yep, you are right

joseph-isaacs: Is it worth even checking this? I would assume its non-trival to compute and you are being asked to compact anyway?

vortex-array/src/arrays/listview/rebuild.rs
joseph-isaacs: You are correct

vortex-array/src/arrays/listview/rebuild.rs
joseph-isaacs, line 1: can we add benchmarks for these?

vortex-array/src/arrays/listview/compute/filter.rs
```diff
@@ -42,10 +53,17 @@ impl FilterKernel for ListViewVTable {
...
             ListViewArray::new_unchecked(elements.clone(), new_offsets, new_sizes, new_validity)
-        }
-        .into_array())
+        };
+
+        // TODO(connor)[ListView]: IsZeroCopyToList optimization.
+        // TODO(connor)[ListView]: Rebuild if the threshold is too low.
+        // if selection_mask.density() < REBUILD_DENSITY_THRESHOLD {
+        //     new_array = new_array.rebuild(ListViewRebuildMode::RemoveGaps);
+        // }
```
joseph-isaacs: cann you remove this

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -30,17 +44,43 @@ impl TakeKernel for ListViewVTable {
...
-        .into_array())
+        };
+
+        // TODO(connor)[ListView]: IsZeroCopyToList optimization.
+        // TODO(connor)[ListView]: Rebuild if the threshold is too low.
+        // #[allow(clippy::cast_possible_truncation)]
+        // let rebuild_threshold = (array.len() as f64 * REBUILD_DENSITY_THRESHOLD) as usize;
+        // if indices.len() <= rebuild_threshold {
+        //     new_array = new_array.rebuild(ListViewRebuildMode::RemoveGaps);
+        // }
```
joseph-isaacs: also this

joseph-isaacs (approved):

vortex-array/src/arrays/listview/rebuild.rs
connortsui20: I think I'm going to leave it to the next PR

PR #4820, opened 2025-10-02
Fastlanes GPU bit unpack kernel

robert3005:
Not a fan of the hack for kernel selection but we can solve this later with an entry point function. Loading local file definitely doesn't work for compiled library. We need to change it later to include the kernel logic in the binary


vortex-gpu/src/bit_unpack.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    let kernel_func = cuda_bit_unpack_kernel(
+        UnpackKernelId::new(
+            array.bit_width(),
+            u8::try_from(array.dtype().as_ptype().bit_width())
+                .vortex_expect("bit width must fit in u8"),
+        ),
+        ctx.clone(),
+    )?;
+    let num_chunks = u32::try_from(array.len() / 1024).vortex_expect("Too many grid elements");
```
joseph-isaacs:
```suggestion
    let num_chunks = u32::try_from(array.len().div_ceil(1024)).vortex_expect("Too many grid elements");
```

vortex-gpu/src/bit_unpack.rs
```diff
@@ -0,0 +1,159 @@
...
+        .map_err(|e| vortex_err!("Failed to load function: {e}"))?;
+    CUDA_KERNELS.write().insert(kernel_id, kernel_func.clone());
+    Ok(kernel_func)
+}
+
+pub fn cuda_bit_unpack(
+    array: &BitPackedArray,
+    ctx: Arc<CudaContext>,
+) -> VortexResult<PrimitiveArray> {
+    assert!(array.patches().is_none(), "Patches not supported");
```
joseph-isaacs, line 80: check if this is empty?

joseph-isaacs (approved):
🔥

PR #4821, opened 2025-10-02
varbinview compact micro bench

onursatici:
getting the benchmark in first to measure some changes I am doing on compacting

### DIFF
--- a/vortex-array/benches/varbinview_compact.rs
+++ b/vortex-array/benches/varbinview_compact.rs
@@ -41,6 +41,16 @@ fn compact_large(bencher: Bencher, args: (usize, usize)) {
     compact_impl(bencher, args);
 }
 
+#[divan::bench(args = SMALL_ARGS)]
+fn compact_sliced(bencher: Bencher, args: (usize, usize)) {
+    compact_sliced_impl(bencher, args);
+}
+
+#[divan::bench(args = LARGE_ARGS, sample_count = 10)]
+fn compact_sliced_large(bencher: Bencher, args: (usize, usize)) {
+    compact_sliced_impl(bencher, args);
+}
+
 fn compact_impl(bencher: Bencher, (output_size, utilization_pct): (usize, usize)) {
     let base_size = (output_size * 100) / utilization_pct;
     let base_array = build_varbinview_fixture(base_size);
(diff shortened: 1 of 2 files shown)


vortex-array/benches/varbinview_compact.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[divan::bench(args = SMALL_ARGS)]
+fn compact_sliced(bencher: Bencher, args: (usize, usize)) {
+    compact_sliced_impl(bencher, args);
+}
+
+#[divan::bench(args = LARGE_ARGS, sample_count = 10)]
+fn compact_sliced_large(bencher: Bencher, args: (usize, usize)) {
+    compact_sliced_impl(bencher, args);
+}
```
joseph-isaacs, line 51: OOI, why split these?

vortex-array/benches/varbinview_compact.rs
```diff
@@ -0,0 +1,104 @@
...
+    let base_size = (output_size * 100) / utilization_pct;
+    let base_array = build_varbinview_fixture(base_size);
+    let indices = random_indices(output_size, base_size);
+
+    bencher
+        .with_inputs(|| {
+            let taken = take(base_array.as_ref(), &indices).vortex_unwrap();
+            taken.to_varbinview()
+        })
+        .bench_values(|array| array.compact_buffers().vortex_unwrap())
```
joseph-isaacs, line 63: Do you think there is a way to avoid the memory allocation in the benchmark, it will make the results much more nosiy

joseph-isaacs (approved):

vortex-array/benches/varbinview_compact.rs
onursatici: with 100 iterations the large case takes a couple of seconds, seemed excessive for a divan benchmark. I want to have large arrays though because the perf difference becomes more pronounced

vortex-array/benches/varbinview_compact.rs
onursatici: yea it is not ideal, but I want to measure the overhead of allocations as well, I am experimenting with changes on compact to do less allocations but more cpu, and want to compare that approach with the current one

PR #4825, opened 2025-10-02
feat: teach Python about (part of) VortexWriteOptions

danking:
In particular, this allows Python users to write using the compact strategy.


vortex-file/src/writer.rs
```diff
@@ -90,6 +91,19 @@ impl VortexWriteOptions {
...
 impl VortexWriteOptions {
+    /// Prioritize small size over read-throughput and read-latency.
+    pub fn compact() -> Self {
+        Self {
+            strategy: WriteStrategyBuilder::new()
+                .with_compressor(CompactCompressor::default())
+                .build(),
+            exclude_dtype: false,
+            file_statistics: PRUNING_STATS.to_vec(),
+            max_variable_length_statistics_size: 64,
```
joseph-isaacs: Shall we make a constant for this, it seems random here

joseph-isaacs (approved):

vortex-file/src/writer.rs
danking: I opted to use `Self::default().with_compressor(...)`.

robert3005 (approved):

danking:
yolo

PR #4827, opened 2025-10-02
varbinview compact to zero copy whenever possible

onursatici:
This also adds a way to the varbinview builder to compact arrays as they get added to the builder.

The way this works is to calculate the overall utilisation (num bytes referred from views per buffer), and keep buffers from the array to be added if they already are highly utilised. 
Another optimisation this brings is to track the used ranges per buffer, and check if the byte usage within that range is above the utilisation threshold, and slice the buffer (zero copy) during compaction. This is useful when the varbinview array is constructed from our builders and then sliced, meaning some buffer has half of its bytes unreferenced, we take the other half without copying in those cases now.


vortex-array/src/builders/varbinview.rs
```diff
@@ -119,6 +138,11 @@ impl VarBinViewBuilder {
     // buffers adjusted.
     // The views must all point to sections of the buffers and the validity length must match
     // the view length.
+    //
+    /// ## Warning
+    /// This method does not check utilisation of the given buffers, callers must provide
+    /// buffers that are fully utilised by the given adjusted views.
+    ///
```
onursatici, line 139: I can check the utilisation and panic here but felt excessive, only fsst append_to_builder uses this currently and it does use it with a 100% utilised buffer

vortex-array/src/arrays/varbinview/compact.rs
```diff
@@ -77,32 +65,101 @@ impl VarBinViewArray {
...
+        }
+    }
+
+    fn add(&mut self, offset: u32, size: u32) {
+        self.used += size;
+        self.min_offset = self.min_offset.min(offset);
+        self.max_offset_end = self.max_offset_end.max(offset + size);
+    }
+
+    pub fn overall_utilisation(&self) -> f64 {
```
a10y: this is extremely nitty but i think in most of the repo we tend to use US English versions of these words, e.g. `utilization`

vortex-array/src/arrays/varbinview/compact.rs
```diff
@@ -77,32 +65,101 @@ impl VarBinViewArray {
...
+    ///
+    /// Buffers with utilization >= `buffer_utilisation_threshold` are kept as-is (zero-copy).
+    /// Buffers below the threshold are compacted into new buffers, reclaiming wasted space.
+    ///
+    /// # Arguments
+    ///
+    /// * `buffer_utilisation_threshold` - Threshold in range [0, 1]. Buffers with utilization
+    ///   below this value will be compacted. Use 0.0 for no compaction, 1.0 for aggressive
+    ///   compaction of any buffer with wasted space.
+    pub fn compact_buffers_selective(
```
a10y: `compact_with_threshold`

vortex-array/src/builders/varbinview.rs
```diff
@@ -449,6 +520,190 @@ impl BufferGrowthStrategy {
...
+
+        let buffers_with_offsets_iter = strategies.iter().zip(array.buffers().iter()).map(
+            |(strategy, buffer)| match strategy {
+                CompactionStrategy::KeepFull => (Some(buffer.clone()), 0u32),
+                CompactionStrategy::Slice { start, end } => {
+                    (Some(buffer.slice(*start as usize..*end as usize)), *start)
+                }
+                CompactionStrategy::Rewrite => (None, 0),
+            },
+        );
```
a10y: you can just move the `unzip` operation up here since it's done by both branches immediately

a10y (commented):
there's a lot happening so could use some more explanatory comments for the next person to come along and read it, but i think i get the gist

vortex-array/src/arrays/varbinview/compact.rs
onursatici: 🦅

vortex-array/src/builders/varbinview.rs
onursatici: removed this altogether to not allocate for offsets if they are all 0

vortex-array/src/builders/varbinview.rs
```diff
@@ -120,6 +140,11 @@ impl VarBinViewBuilder {
     // buffers adjusted.
     // The views must all point to sections of the buffers and the validity length must match
     // the view length.
+    //
```
robert3005: Can you make this a full doc instead of having a comment preceding a doc string?

vortex-array/src/builders/varbinview.rs
```diff
@@ -374,6 +439,13 @@ impl DeduplicatedBuffers {
         }
     }
 
+    fn extend_from_option_slice(&mut self, buffers: &[Option<ByteBuffer>]) -> Vec<Option<u32>> {
```
robert3005, line 445: Change this to take something that's `IntoIterator<Item=Option<ByteBuffer>>`

robert3005: This duplication lead me down the path of looking at adjustments, and isn't PrecomputedViewAdjustment::Lookup the same as RewritingViewAdjustment where the 2nd one handles rewrites being optional?

robert3005 (commented):
Can we simplify the adjustment strategies? It seems that one option leads to a lot of duplication down the line.

vortex-array/src/builders/varbinview.rs
onursatici: yea exactly, but that difference allows us to use `self.view_builder.extend_trusted` on cases where we can now beforehand where all buffers will end up at:

```rust
            ViewAdjustment::Precomputed(adjustment) => self.views_builder.extend_trusted(
                array
                    .views()
                    .iter()
                    .map(|view| adjustment.adjust_view(view)),
            ),
            ViewAdjustment::Rewriting(adjustment) => {
                for (idx, &view) in array.views().iter().enumerate() {
                    let new_view = if !array.is_valid(idx) {
                        BinaryView::empty_view()
                    } else if view.is_inlined() {
                        view
                    } else if let Some(adjusted) = adjustment.adjust_view(&view) {
                        adjusted
                    } else {
                        let bytes = array.bytes_at(idx);
                        let (new_buf_idx, new_offset) = self.append_value_to_buffer(&bytes);
                        BinaryView::make_view(bytes.as_slice(), new_buf_idx, new_offset)
                    };
                    self.views_builder.push(new_view)
                }
            }
```

robert3005 (approved):

PR #4830, opened 2025-10-02
Feature: `ListView` rebuilding strategies

connortsui20:

Adds `ListViewShape` and more `ListViewRebuildMode` variants and functionality. Additionally adds a `list_from_list_view` function and tests.

The main goal of these changes is to allow us to rebuild the `ListViewArray`s to have a similar shape as our `ListArray`s. There are no benchmarks here yet since we have yet to make `ListView` the canonical encoding for `List` (the next and probably last PR for `ListView`), but the hope here is that by adding this functionality now we can mimic the behavior of `ListArray` in benchmarks, meaning we won't have a regression.

### DIFF
--- a/vortex-array/src/arrays/listview/array.rs
+++ b/vortex-array/src/arrays/listview/array.rs
@@ -4,15 +4,13 @@
 use std::sync::Arc;
 
 use num_traits::AsPrimitive;
-use vortex_dtype::{DType, IntegerPType, Nullability, match_each_integer_ptype};
-use vortex_error::{VortexExpect, VortexResult, vortex_ensure, vortex_err};
+use vortex_dtype::{DType, IntegerPType, match_each_integer_ptype};
+use vortex_error::{VortexExpect, VortexResult, vortex_bail, vortex_ensure, vortex_err};
 
-use crate::arrays::{ListArray, PrimitiveVTable};
-use crate::builders::PrimitiveBuilder;
+use crate::arrays::{ListViewShape, PrimitiveArray, PrimitiveVTable};
 use crate::stats::ArrayStats;
 use crate::validity::Validity;
-use crate::vtable::ValidityHelper;
-use crate::{Array, ArrayRef, Canonical, IntoArray, ToCanonical};
+use crate::{Array, ArrayRef, ToCanonical};
 
 /// The canonical encoding for variable-length list arrays.
 ///
@@ -39,23 +37,27 @@ use crate::{Array, ArrayRef, Canonical, IntoArray, ToCanonical};
 /// # Examples
 ///
 /// ```
-/// use vortex_array::arrays::{ListViewArray, PrimitiveArray};
-/// use vortex_array::validity::Validity;
-/// use vortex_array::IntoArray;
-/// use vortex_buffer::buffer;
-/// use std::sync::Arc;
+/// # use vortex_array::arrays::{ListViewArray, ListViewShape, PrimitiveArray};
+/// # use vortex_array::validity::Validity;
+/// # use vortex_array::IntoArray;
+/// # use vortex_buffer::buffer;
+/// # use std::sync::Arc;
+/// #
+/// // Create a list view array representing [[3, 4], [1], [2, 3]].
+/// // Note: Unlike `ListArray`, offsets don't need to be monotonic.
 ///
-/// // Create a list view array representing [[3, 4], [1], [2, 5]]
-/// // Note: Unlike ListArray, offsets don't need to be monotonic
 /// let elements = buffer![1i32, 2, 3, 4, 5].into_array();
 /// let offsets = buffer![2u32, 0, 1].into_array();  // Out-of-order offsets
-/// let sizes = buffer![2u32, 1, 2].into_array();  // Corresponding sizes
+/// let sizes = buffer![2u32, 1, 2].into_array();  // The sizes cause overlaps
 ///
 /// let list_view = ListViewArray::try_new(
 ///     elements.into_array(),
 ///     offsets.into_array(),
 ///     sizes.into_array(),
 ///     Validity::NonNullable,
+///     ListViewShape::as_zero_copy_to_list()
+///         .with_sorted_offsets(false)
+///         .with_no_overlaps(false),
 /// ).unwrap();
 ///
 /// assert_eq!(list_view.len(), 3);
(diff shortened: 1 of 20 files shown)


vortex-array/src/arrays/listview/array.rs
```diff
@@ -172,16 +191,19 @@ impl ListViewArray {
             offsets,
             sizes,
             validity,
+            shape,
             stats_set: Default::default(),
         }
     }
 
     /// Validates the components that would be used to create a [`ListViewArray`].
+    #[allow(clippy::cognitive_complexity)]
```
connortsui20: Can probably remove this

vortex-array/src/arrays/listview/conversion.rs
```diff
@@ -0,0 +1,405 @@
...
+
+use crate::arrays::{ListArray, ListViewArray, ListViewShape};
+use crate::builders::{ArrayBuilder, ListBuilder, PrimitiveBuilder};
+use crate::vtable::ValidityHelper;
+use crate::{Canonical, IntoArray};
+
+/// Creates a [`ListViewArray`] from a [`ListArray`] by computing `sizes` from `offsets`.
+pub fn list_view_from_list(list: ListArray) -> ListViewArray {
+    // TODO(connor)[ListView]: Create a version of `Canonical::empty` for `ListView`. It might
+    // also be worth specializing that for all canonical encodings.
```
connortsui20: This will be fixed after `ListView` is canonicalized

vortex-array/src/arrays/listview/conversion.rs
```diff
@@ -0,0 +1,405 @@
...
+        let list_offsets = match_each_integer_ptype!(list_view.offsets().dtype().as_ptype(), |O| {
+            let mut offsets_builder =
+                PrimitiveBuilder::<O>::with_capacity(Nullability::NonNullable, len + 1);
+
+            // Copy the existing n offsets.
+            for i in 0..len {
+                let offset = list_view.offset_at(i);
+                offsets_builder
+                    .append_value(O::try_from(offset).vortex_expect("offset must fit in type"));
+            }
```
connortsui20: This should probably use `UninitRange` like above

vortex-array/src/arrays/listview/shape.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// This type keeps track of information that allows us to perform operations like slicing and
+/// rebuilding much more efficiently.
+///
+/// ### Zero-copy to [`ListArray`]
+///
+/// If all of the flags in this struct are set to `true`, then we know that the corresponding
+/// [`ListViewArray`] is "zero-copyable" to a [`ListArray`]. Note that technically it can never be
+/// truly zero-copyable since we must add a single `offset` to get the correct `n+1` offsets that
+/// [`ListArray`] needs, but the data is zero-copyable in spirit.
```
connortsui20, line 35: Since "zero-copyable" is actually not completely true, should I rename it to something else? I still don't really like the word "contiguous" here...

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -76,27 +105,94 @@ impl ListViewArray {
         })
     }
 
+    // TODO(connor)[ListView]: Write docs.
+    fn rebuild_trim_elements(&self) -> ListViewArray {
```
connortsui20: I'll get to this on Monday

robert3005 (commented):
I think there's something weird here. I wonder if there's a different way to figure this out. Maybe there's an already builtin property that we can leverage for the same behaviour. I think the difficult part for me is that we still have to make sure results are correct even if the hints are incorrect.

connortsui20:
@robert3005 could you clarify what you mean by "make sure results are correct even if the hints are incorrect"? I think I've taken care to make sure that the hints / invariants are always correct even through the different operations.

The downside is that when reading "untrusted" data the `validate` step will take slightly longer, but the benefit is that we can other operations more efficiently (namely take and filter) later. (Note that `ListView` version of take and filter + rebuild is roughly equivalent performance-wise to the normal `List` version of take and filter, not even considering any `reset_offsets` calls).

> Maybe there's an already builtin property that we can leverage for the same behaviour

Unless there's a something very subtle I'm missing, the whole point of `ListView` (from my understanding) is that we relax the "built-in properties" of `List` as a tradeoff for a different feature set. That's sort of why I built this semi-auxiliary structure and code to mimic those missing properties. Maybe we don't even need this for most use cases (as I have yet to properly canonicalize `ListView` so we can run benchmarks, I will make sure that will be done on Monday), but I think this is necessary to _always_ have similar performance to `List` in _all_ cases.

robert3005:
I didn't notice that the shape is being validated, I think that covers the correctness. I have been wondering if this is that big of an issue if we have more reasonably sized list element chunks? I think right now you end up with 100M elements in the child which is definitely bad. It could be that this is the right approach but I am always hesitant about adding more complexity

connortsui20:
@robert3005 to get this moving again, @joseph-isaacs is concerned that if in the future we want to do a chunked version of the elements (as the default version canonical `ListView`), we would probably want to use a little metadata struct array of these flags for **each** chunk (not global), and thus we would need to break the file format (see the `listview/vtable/serde.rs` file). 

We also ran the statpopgen benchmark with this optimization (on [this](https://github.com/vortex-data/vortex/pull/4853) branch) and it is still a regression, but much less than before, because the only slowdown here is from the much more expensive `validate` function. In fact, if we turn off `validate` for lists, ListView becomes about 15% faster than List.

I am not convinced that chunking the underlying elements array would solve this problem. What chunk size would we use for the elements? The writer has no idea what is a good chunk size because it has no idea what access pattern the reader might have (if the reader reads a single row versus a large amount of data, or if the reader reads in a sort of striped pattern, or completely random, or in a sort of pareto shape, etc). 

I also do not think that breaking the ListView format is such a big deal considering nobody is using ListView right now (because it's not done). And we can always revert this before #4853 eventually merges (if we even want to merge that at all, it doesn't seem like we are 100% confident that making ListView the default canonical version of List is the best idea).

I would like to merge this so I can make actual progress on #4853, which has been stale for almost 2 weeks now.

robert3005:
I don't want to change the array serialisation format. I want to change the layout for list arrays. What I think we should do is merge the change to have ListView be the canonical array, we shouldn't add the ListViewShape. Then we should work on a listviewlayout that wouldn't produce insanely large list chunks

connortsui20:
Dicussed offline, we can revisit this in the future.

PR #4831, opened 2025-10-03
Use DataFusion's FilePruner to prune files based on stats and partition values

AdamGS:
Just took it out of the Parquet datasource implementation. Some local experimentation shows that it should improve clickbench perf by a bit, which is exciting.


AdamGS:
1. Nice boost on some of the last clickbench queries
2. Seems like `datafusion:vortex-file-compressed` shows up twice?

a10y (approved):

PR #1082, opened 2024-10-17
German strings, attempt 3

a10y:
Implementing VarBinView as our canonical string type. Concretely, this involves

1. Changing `Canonical` variant from `VarBin` to `VarBinView` and updating all of those method names accordingly
2. `impl IntoCanonical for VarBinArray` that involves doing a single-pass construction of views
3. Changing the Arrow type from `Utf8/Binary` to `Utf8View/BinaryView` and propagating that in Datafusion and Python bindings
4. Changing how ChunkedArray canonicalize works to repack views instead of repacking arrays

Some caveats

* Currently in varbin -> varbinview, we reuse the `bytes` heap backing the varbin as the buffer for varbinview. This means that if the array contains > 2GiB of string data, we will fail currently. I can work on the rollover logic, it's just going to make the PR larger. Alternatively we can say that having chunks > 2GiB is ok, but then it fails converting to Arrow and we just say you can't have a RecordBatch of strings > 2GiB in size. But I don't think that's a great API.
* This change has little effect on TPC-H benchmarks. Most queries are the same, some are slightly faster and some are ~10-30% slower. Previously canonicalizing FSST was pretty trivial and didn't require looping over the values, now with the additional view construction it adds overhead to get all bytes and construct the views

image

PR #4832, opened 2025-10-03
binary view debug impl

onursatici:
I think this was changed way before at https://github.com/vortex-data/vortex/pull/1082 , I don't think we use this on any tests, and I needed to see what the references are during working on varbinviews recently


joseph-isaacs (approved):

joseph-isaacs:
Shall we have a fmt test?

PR #4834, opened 2025-10-03
revert divan codspeed compat to v3

onursatici:
seems like codspeed divan compat v4 broke on macos, to be able to run benches locally now we use divan itself on non codspeed runs


robert3005 (approved):
We should file an issue upstream

onursatici:
https://github.com/CodSpeedHQ/codspeed-rust/issues/130

PR #4835, opened 2025-10-03
chore: Simplify stats truncation, optimise take function stats inheritance

robert3005:

### DIFF
--- a/vortex-array/src/compute/take.rs
+++ b/vortex-array/src/compute/take.rs
@@ -8,9 +8,9 @@ use vortex_dtype::DType;
 use vortex_error::{VortexError, VortexResult, vortex_bail, vortex_err};
 use vortex_scalar::Scalar;
 
-use crate::arrays::{ConstantArray, ConstantVTable};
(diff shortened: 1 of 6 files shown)


vortex-array/src/compute/take.rs
```diff
@@ -79,19 +79,8 @@ impl ComputeFnVTable for Take {
...
-            // `StatsSetRef` so we don't have to incur a clone here in `.to_owned()`.
-            let mut stats = taken_array.statistics().to_owned();
-            stats.combine_sets(&derived_stats, array.dtype())?;
-
-            for (stat, val) in stats {
-                // Alternatively, use a monoidal pattern here to set `stat = val`, or if it already
-                // exists, combine the two stats (similar to how `combine_sets` does it).
-                taken_array.statistics().set(stat, val)
-            }
+        if !taken_array.is_constant() {
```
robert3005, line 82: this shows up as a regression because we do extra work here now but I think this is the logic we want. I know I went back and forth on this

AdamGS (approved):

Issue #4767, opened 2025-09-25
Support duckdb recursive types


joseph-isaacs:
Have a second mode in the duckdb column exporters which will export a vector array into a duckdb flat vector

connortsui20:
Otherwise:

```
called `Result::unwrap()` on an `Err` value: Failed to execute query: INTERNAL Error: Assertion triggered in file "/Users/connor/spiral/vortex-data/vortex-develop/target/duckdb-source-v1.4.0/duckdb-1.4.0/src/common/types/vector.cpp" on line 1689: child.GetVectorType() == VectorType::FLAT_VECTOR
```

PR #4843, opened 2025-10-03
Chore: Remove some TODOs

connortsui20:
see commit descriptions

For the flatten removal, that was fixed by some of the recent duckdb dictionary exporter changes recently

### DIFF
--- a/vortex-array/src/arrays/list/compute/is_sorted.rs
+++ b/vortex-array/src/arrays/list/compute/is_sorted.rs
@@ -3,21 +3,20 @@
 
 use vortex_error::VortexResult;
 
-use crate::arrays::{FixedSizeListArray, FixedSizeListVTable};
+use crate::arrays::{ListArray, ListVTable};
 use crate::compute::{IsSortedKernel, IsSortedKernelAdapter};
 use crate::register_kernel;
 
-/// IsSorted implementation for [`FixedSizeListArray`].
-impl IsSortedKernel for FixedSizeListVTable {
-    fn is_sorted(&self, _array: &FixedSizeListArray) -> VortexResult<Option<bool>> {
+impl IsSortedKernel for ListVTable {
+    fn is_sorted(&self, _array: &ListArray) -> VortexResult<Option<bool>> {
         // This would require comparing lists lexicographically.
         Ok(None)
     }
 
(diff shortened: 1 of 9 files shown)


vortex-duckdb/src/exporter/fixed_size_list.rs
```diff
@@ -75,12 +75,6 @@ impl ColumnExporter for FixedSizeListExporter {
         self.elements_exporter
             .export(offset * list_size, len * list_size, &mut elements_vector)?;
 
-        // TODO(connor): We must flatten the child vector to ensure any child dictionary views
-        // (namely UTF-8 string views in dictionaries) are materialized.
-        // See https://github.com/vortex-data/vortex/pull/4610#issuecomment-3286676825 for a
-        // detailed explanation on why we need this for now.
-        elements_vector.flatten((len * list_size) as u64);
-
```
danking, line 83: Which PR made this unnecessary?

danking (changes requested):

vortex-duckdb/src/exporter/fixed_size_list.rs
connortsui20: I believe it was this one? https://github.com/vortex-data/vortex/pull/4788/files

Not entirely sure, @joseph-isaacs is this correct?

joseph-isaacs: I don't think so. I think we need to have a flatten for rle and maybe constant?

Basically we need to check what vector types are allowed in the element buffer of a list vector.

connortsui20: In that case I'm just going to remove the flatten from this PR.

robert3005: Pretty certain that duckdb allows flat and constant children for lists

connortsui20: IIRC What Joe and I saw when we looked into the DuckDB code was that there was an assertion where they don't allow any non-flat vectors in nested types, so any ARRAY or LIST type is not allowed to have unflattened child type.

Edit: https://github.com/vortex-data/vortex/issues/4767

danking (approved):

Issue #4842, opened 2025-10-03
Implement `IsSortedKernel` and `MinMaxKernel` for `List`, `ListView`, and `FixedSizeList`


connortsui20:
Right now, neither `IsSortedKernel` nor `MinMaxKernel` have been implemented for any of `List`, `ListView`, and `FixedSizeList`.

This would require lexicographic comparisons between list scalars.

The implementations are all located under `vortex-array/src/arrays`, in the respective `compute` directories (i.e., `vortex-array/src/arrays/fixed_size_list/compute/is_sorted.rs`).

Right now they all return `Ok(None)`, so we want them to return `Ok(Some(true))` if they are in lexicographic order.

The naive way to do this is to make some `compare` function that takes two `ListScalar`s and manually iterate over each `Array` type with `scalar_at(i)`. That would also need to handle nulls.

I can't really think of a more efficient way to do this (unless we added another compute function for lexicographic comparisons between 2 array [slices]).

---

_I'm creating this issue just so it is not only discoverable through TODOs in the codebase, and I'm assigning it to people who had those TODOs there_

_If you're not a Vortex contributor and you would like to try this out, please let us know!_

robert3005:
I think the optimal solution here is to convert things to row based representation and then min/max, check is_sorted on that. This is implemented by arrow-ord and by foundationdb tuple libraries

connortsui20:
@joseph-isaacs is there an epic or tracking issue for the vortex row format?

PR #4847, opened 2025-10-04
feature: Implement IsSortedKernel and MinMaxKernel for List, ListView, and FixedSizeList

skil1sue:
Mentioned issue: vortex-data/vortex#4842

### DIFF
--- a/vortex-array/src/arrays/fixed_size_list/compute/is_sorted.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/is_sorted.rs
@@ -2,22 +2,129 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use vortex_error::VortexResult;
+use vortex_scalar::ListScalar;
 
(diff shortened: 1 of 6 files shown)


skil1sue:
@connortsui20 Any part of code need to be rewrite or any problems?

robert3005:
I'll bite on the explanation
1. The style doesn't look like rust code, you only ever need to invoke trait methods via trait in complicated situtations
2. The errors are ignored in multiple places
3. That issue doesn't explain why they're not implemented and is missing details about what a good implementation will have. We have left out these implementation because making them performant is more important than having them. 
4. This is not a good first issue to this project, we have left it unimplemented because it's not straightforward and you have to add some additional abstractions
5. Your account is 1 week old and has no other contributions.

skil1sue:
@robert3005 Understand, if it's a issue above first-good-issue level I can't handle with it. Also thanks for the code-style advices, and if possible could you pick any snippet in case, give any suggestions about how to improve my code to work in Rust well?

PR #4854, opened 2025-10-06
wip: shredded nested schemas

a10y:
Shredding nested schemas. This creates a nwe DType, but structurally is closer to what we want.

We'd also want a way to make sure that the field streams can be re-stitched together.

Note that this is only useful for nested structs, not lists or lists of structs.
We'd also need to change how expression partitioning works


robert3005:
I think all you really need to do is fix any bug that comes from nesting struct layouts, nesting struct layouts should be relatively cheap and we shouldn't require all nested structs to be flattened

PR #4862, opened 2025-10-07
pack expr bench

onursatici:


onursatici:
need this to measure some of my changes, also would be nice to not regress this because we use it in some places assuming it is cheap enough, like in the remove_merge transformer which gets the return_dtype's of each of its children.

robert3005 (approved):

PR #4863, opened 2025-10-07
faster StructFields::find

onursatici:


vortex-dtype/src/struct_.rs
```diff
@@ -215,11 +216,73 @@ impl Display for StructFields {
     }
 }
 
-#[derive(PartialEq, Eq, Hash, Default)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Default)]
 struct StructFieldsInner {
```
robert3005, line 220: ~I wonder if we should make this an enum since for narrow types it might still make sense to keep it as a list. I am speculating about how big the structFieldsInner gets with a hashmap in it~. Enum doesn't change the size

robert3005 (approved):

PR #4857, opened 2025-10-06
feat[gpu]: for kernel and framework for running pipeline

joseph-isaacs:
Looks like we need to JIT 

benchmark runs:

```
gpu_decompress_kernel_only/u32/1GB
                        time:   [5.6860 ms 5.6878 ms 5.6892 ms]
                        thrpt:  [175.77 GiB/s 175.81 GiB/s 175.87 GiB/s]

gpu_decompress_kernel_only/u32/10GB
                        time:   [56.376 ms 56.384 ms 56.392 ms]
                        thrpt:  [177.33 GiB/s 177.35 GiB/s 177.38 GiB/s]
          
gpu_for_decompress_kernel_only/u32/1GB
                        time:   [19.657 ms 19.674 ms 19.694 ms]
                        thrpt:  [50.778 GiB/s 50.827 GiB/s 50.871 GiB/s]

gpu_for_decompress_kernel_only/u32/10GB
                        time:   [203.45 ms 203.81 ms 204.21 ms]
                        thrpt:  [48.969 GiB/s 49.064 GiB/s 49.151 GiB/s]
```

PR #4864, opened 2025-10-07
RLE gpu decompress kernel

robert3005:
The minimal version of RLE kernel is the same as take, however, as we add
handling for masked kernels they will diverge


vortex-gpu/src/bit_unpack.rs
```diff
@@ -33,24 +31,18 @@ impl UnpackKernelId {
...
 }
 
-static CUDA_KERNELS: LazyLock<RwLock<HashMap<UnpackKernelId, CudaFunction>>> =
-    LazyLock::new(|| RwLock::new(HashMap::new()));
-
 fn cuda_bit_unpack_kernel(
     kernel_id: UnpackKernelId,
     ctx: Arc<CudaContext>,
 ) -> VortexResult<CudaFunction> {
-    if let Some(kernel) = CUDA_KERNELS.read().get(&kernel_id) {
```
0ax1, line 50: I assume we cached compiled kernels before. Just for my understanding, why do we drop this? Can kernels with the same id ever be different?

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,130 @@
...
+use vortex_fastlanes::RLEArray;
+
+use crate::take::cuda_take_kernel;
+
+pub fn cuda_rle_decompress(
+    rle: &RLEArray,
+    ctx: Arc<CudaContext>,
+) -> VortexResult<Option<ArrayRef>> {
+    match_each_native_ptype!(rle.values().dtype().as_ptype(), |V| {
+        // RLE indices are always u16 (or u8 if downcasted).
```
0ax1: I've seen this code somewhere before. 🙂

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,130 @@
...
+                rle.values().to_primitive().as_slice::<V>(),
+                ctx,
+            )
+            .map(Some),
+            PType::U16 => cuda_rle_decompress_typed(
+                rle.indices().to_primitive().as_slice::<u16>(),
+                rle.values().to_primitive().as_slice::<V>(),
+                ctx,
+            )
+            .map(Some),
```
0ax1: Why wrap the `ArrayRef` in an `Option`?

0ax1: Didn't we decide to make decompression infallible? Should this return a result in the error case or panic.

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,130 @@
...
+        .map_err(|e| vortex_err!("Failed to copy to device: {e}"))?;
+
+    let mut cu_out = unsafe {
+        stream
+            .alloc::<Values>(indices.len().next_multiple_of(1024))
+            .map_err(|e| vortex_err!("Failed to allocate stream: {e}"))?
+    };
+
+    let mut launch = stream.launch_builder(&kernel_func);
+    launch.arg(&indices);
```
0ax1: missing cu_ prefix

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,130 @@
...
+    let stream = ctx.default_stream();
+
+    let cu_values = stream
+        .memcpy_stod(values)
+        .map_err(|e| vortex_err!("Failed to copy to device: {e}"))?;
+    let cu_indices = stream
+        .memcpy_stod(indices)
+        .map_err(|e| vortex_err!("Failed to copy to device: {e}"))?;
+
+    let mut cu_out = unsafe {
```
0ax1: In the original RLE impl this returns the number of values in the `values` array, not the padded indices.

0ax1 (commented):
Couple of questions and remarks.

vortex-gpu/src/bit_unpack.rs
robert3005: We need to figure out a better way of doing this, I removed it so it's not sitting here randomly

vortex-gpu/src/rle_decompress.rs
robert3005: yes, it failed to compile because of this

vortex-gpu/src/rle_decompress.rs
robert3005: there will be as many entries as there's indices, then we need to adjust for offset and length but this has to be overallocated to the indices length to avoid out of bounds writes

vortex-gpu/src/rle_decompress.rs
robert3005: Need to unify these, the option is meant to say we don't have a kernel. I removed it for now and will add it back later.

java/testfiles/Cargo.lock
```diff
@@ -75,9 +75,9 @@ checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
 
 [[package]]
 name = "arrow-arith"
-version = "55.2.0"
```
0ax1, line 78: This seems unrelated.

0ax1 (approved):

vortex-gpu/src/rle_decompress.rs
joseph-isaacs, line 1: Think we can look at merging on the framework in https://github.com/vortex-data/vortex/pull/4857 ?

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    use crate::rle_decompress::cuda_rle_decompress;
+
+    #[test]
+    fn test_cuda_rle_decompress() {
+        let primitive_array = PrimitiveArray::new(
+            (0u32..4096)
+                .flat_map(|i| iter::repeat_n(i, i as usize % 10))
+                .collect::<Buffer<_>>(),
+            Validity::NonNullable,
```
joseph-isaacs: can test all possible types supported by any kernels?

robert3005:
RLE is not take... it's a very specific take where every 1024 indices are local to their value range so we need to slice values with offsets, that's why the result is incorrect.

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,131 @@
...
+        );
+        let array = RLEArray::encode(&primitive_array).vortex_unwrap();
```
0ax1, line 235: You need to factor in that indices in RLEArray are chunk local, as in relative to the chunk. So its a dictionary lookup within a 1024 chunk.

vortex-gpu/src/rle_decompress.rs
robert3005: we can rstest it

vortex-gpu/kernels/rle_decompress.cu
joseph-isaacs, line 1: any reason to avoid floats in the `ValueT`?

robert3005: there's not, I was working on tests to improve the coverage

vortex-gpu/kernels/rle_decompress.cu
```diff
@@ -0,0 +1,143 @@
...
+    auto out = values_out + block_offset;
+    auto values = values_array + offsets[blockIdx.x];
+
+    const int thread_ops = 32;
+
+    for (auto j = 0; j < thread_ops; j++) {
+        auto idx = i * thread_ops + j;
+        out[idx] = values[indices[idx]];
+    }
+}
```
joseph-isaacs, line 28: We will need a fl-transposed iterator order version?

joseph-isaacs: Would you mind adding a single fused kernel bp-rle. Where we fuse bitunpacking the indices and rle decoding. 

It can be in a follow up, but while you understand how the decode works we should merge something showing that.

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,241 @@
...
+                .vortex_expect("Failed to transmute")
+        }
+    }
+
+    fn len(&self) -> usize {
+        self.len
+    }
+}
+
+#[allow(clippy::cognitive_complexity)]
```
joseph-isaacs, line 79: I think we should follow this hint. Can we pull out the inner match below?

vortex-gpu/src/rle_decompress.rs
```diff
@@ -0,0 +1,241 @@
...
+    #[case::u16((0u16..2000).collect::<Buffer<u16>>())]
+    #[case::u32((0u32..2000).collect::<Buffer<u32>>())]
+    #[case::u64((0u64..2000).collect::<Buffer<u64>>())]
+    #[case::i8((-100i8..100).collect::<Buffer<i8>>())]
+    #[case::i16((-2000i16..2000).collect::<Buffer<i16>>())]
+    #[case::i32((-2000i32..2000).collect::<Buffer<i32>>())]
+    #[case::i64((-2000i64..2000).collect::<Buffer<i64>>())]
+    #[case::f32((-2000..2000).map(|i| i as f32).collect::<Buffer<f32>>())]
+    #[case::f64((-2000..2000).map(|i| i as f64).collect::<Buffer<f64>>())]
+    fn test_cuda_rle_decompress<T: NativePType>(#[case] values: Buffer<T>) {
```
joseph-isaacs, line 233: do we need to exercise the different offset types?

joseph-isaacs (approved):
I think its good to go, but it would be great to impl a fused bp-rle kernel so we can see how that might be done

joseph-isaacs:
I think its as simple as change teh indices iteration order to fl-transposed, but it might not be

PR #4866, opened 2025-10-07
Minor simplification of the serde dependency

AdamGS:
1. `vortex-array` doesn't need serde at all IMO.
2. `vortex-buffers` only needs `serde_core`, so that should help a bit in some cases.


robert3005 (approved):

a10y (approved):
does this actually affect code size or compile times? vs just importing serde without the derive feature

AdamGS:
seems like the answer is yes, because it decouples the compilation units - https://github.com/BurntSushi/rust-csv/pull/403#issue-3415258129

a10y:
Nice, yea I was a bit confused by the [crate doc](https://docs.rs/serde_core/latest/serde_core/)

> In crates that handwrite implementations of Serde traits, or only use them as trait bounds, depending on serde_core is permitted. But serde re-exports all of these traits and can be used for this use case too. If in doubt, disregard serde_core and always use serde.

But then it goes on to mention that it allows to parallelize the serde-derive build, so I guess for a downstream crate it will see benefits

PR #4871, opened 2025-10-07
rk/launchevents

robert3005:


vortex-gpu/src/for_.rs
```diff
@@ -64,17 +65,17 @@ fn cuda_for_kernel(ptype: PType, ctx: &Arc<CudaContext>) -> VortexResult<CudaFun
...
-        .map_err(|e| vortex_err!("Failed to load function: {e}"))?;
-    Ok(kernel_func)
+        .map_err(|e| vortex_err!("Failed to load function: {e}"))
 }
 
 impl<P: NativePType + DeviceRepr> GPUTask for ForTask<P> {
-    fn launch_task(&mut self) -> VortexResult<()> {
+    fn launch_task(&mut self) -> VortexResult<(CudaEvent, CudaEvent)> {
         let len = self.len();
         self.bp_task.launch_task()?;
```
joseph-isaacs, line 76: You need to return the start of this with the end of the other launch?

robert3005:
We likely shouldn't do this by default based on casual browse of nvidia issues where events might slow down the context

PR #4872, opened 2025-10-07
perf[gpu]: fused AoT `for` and `bitpacking` kernel

joseph-isaacs:
I have only implemented a PoC fused FoR-BP kernel, I don't want to implemented them all since there will be a lot of duplication. I think we likely need to compile these at runtime.

I have also fixed up the kernels build system.

Fused is fast

```
gpu_for_bp_fused_decompress_kernel_only/u32/1GB
                        time:   [5.5376 ms 5.5410 ms 5.5443 ms]
                        thrpt:  [180.37 GiB/s 180.47 GiB/s 180.58 GiB/s]
```


lwwmanning (approved):

vortex-gpu/src/for_bp.rs
```diff
@@ -0,0 +1,201 @@
...
+    Ok(kernel_func)
+}
+
+impl<P: NativePType + DeviceRepr> GPUTask for FoRBPTask<P> {
+    fn launch_task(&mut self) -> VortexResult<()> {
+        let mut launch = self.stream.launch_builder(&self.func);
+        launch.arg(&self.packed);
+        launch.arg(&self.unpacked);
+        launch.arg(&self.reference);
+        launch.record_kernel_launch(CU_EVENT_DEFAULT);
```
robert3005, line 96: you don't need this

robert3005 (approved):

PR #4882, opened 2025-10-07
Feature: Add Lance to our random access benchmarks

connortsui20:
Seems like adding lance adds a lot of dependencies. (Edit: for lance we make an exception and add it only to the `bench-vortex` dependencies, not the workspace)

Also, it looks like they are also on DF v50 which is nice!

Lance beats us on random access by about ~15% on average (on my macbook), which isn't too surprising. But also this specific benchmark is so fast and both are well over 100x faster than Parquet.

In the next few days I will do my best to add lance to all of those other benchmarks that are `unimplemented!`.

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -44,3 +44,8 @@ runs:
     - name: Export Path
       shell: bash
       run: echo "PATH=$PATH" >> $GITHUB_ENV
+
+    - name: Install Protoc (for lance-encoding build step)
+      uses: arduino/setup-protoc@v3
+      with:
+        version: "29.3"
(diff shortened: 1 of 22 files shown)


connortsui20:
Is there some issue with licenses that means we cannot add lance as a dependency?

robert3005:
there's a whitelist of licenses, you have to add it to the allowed list but also need to make sure its an opensource license

joseph-isaacs (approved):

danking (approved):

connortsui20:
We need to figure out how to add the secrets to avoid rate limiting: https://github.com/arduino/setup-protoc

Edit: done

PR #4890, opened 2025-10-08
Return filter from FileSource

AdamGS:
Literally just a function from the trait we missed.

### DIFF
--- a/vortex-datafusion/src/persistent/format.rs
+++ b/vortex-datafusion/src/persistent/format.rs
@@ -373,10 +373,6 @@ impl FileFormat for VortexFormat {
         _state: &dyn Session,
         file_scan_config: FileScanConfig,
     ) -> DFResult<Arc<dyn ExecutionPlan>> {
-        if !file_scan_config.output_ordering.is_empty() {
-            return not_impl_err!("Vortex doesn't support output ordering");
-        }
-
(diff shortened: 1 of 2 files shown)


AdamGS:
of course some random duckdb query improves by 50% 😕

vortex-datafusion/src/persistent/format.rs
```diff
@@ -373,10 +373,6 @@ impl FileFormat for VortexFormat {
         _state: &dyn Session,
         file_scan_config: FileScanConfig,
     ) -> DFResult<Arc<dyn ExecutionPlan>> {
-        if !file_scan_config.output_ordering.is_empty() {
-            return not_impl_err!("Vortex doesn't support output ordering");
-        }
-
```
AdamGS, line 379: This just doesn't do what I think it did, so it seems like its fine.

vortex-datafusion/src/persistent/source.rs
```diff
@@ -179,6 +179,10 @@ impl FileSource for VortexSource {
         Arc::new(source)
     }
 
+    fn filter(&self) -> Option<Arc<dyn PhysicalExpr>> {
+        self.vortex_predicate.clone()
+    }
```
AdamGS, line 184: This should make some difference during planning.

a10y (approved):

PR #4891, opened 2025-10-08
Perf: Create splits via indices instead of layout reader

connortsui20:
CC @joseph-isaacs 

Instead of computing splits via the layout reader every time we scan, if we see that selection indices are sparse then we can create the splits directly from the selection.

Also refactors the `vortex-scan/src/lib.rs` file into other files.


vortex-scan/src/splits.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::selection::Selection;
+
+/// The limit where we do not merge indices into a range if the gap is larger than this.
+const MAX_GAP_TO_MERGE: u64 = 1 << 5; // 32
+
+/// The maximum number of rows in a single range. This is somewhat arbitrarily chosen.
+const MAX_RANGE_SIZE: u64 = 1 << 10; // 1024
+
+/// The minimum gap between ranges. This is somewhat arbitrarily chosen.
+const MIN_GAP_BETWEEN_RANGES: u64 = 1 << 15; // 32768
```
joseph-isaacs: can you define these in terms of an ideal split size of 100k

vortex-scan/src/splits.rs
```diff
@@ -0,0 +1,94 @@
...
+        return Some(Vec::new());
+    }
+
+    debug_assert!(indices.is_sorted());
+
+    // We need to create ranges that will represent splits that cover our indices.
+    // We want to make sure that we do not create too many splits. We also want to make sure our
+    // splits do not cover too much as they would overlap column chunk boundaries.
+
+    let mut ranges = Vec::new();
```
joseph-isaacs: give this a capacity? Maybe of 30 and also we likely don't want more ranges than 30

connortsui20: I think it's fine to have an arbitrary number of ranges since we have no idea how large the file actually is. I gave it a heuristic capacity based on the minimum gaps.

vortex-scan/src/splits.rs
```diff
@@ -4,22 +4,34 @@
...
+
+/// The minimum gap between ranges. This is somewhat arbitrarily chosen.
+#[allow(clippy::cast_possible_truncation)]
+const MIN_GAP_BETWEEN_RANGES: u64 = (0.1 * IDEAL_SPLIT_SIZE as f64) as u64;
+
 /// The way in which we compute splits for a file.
 pub(super) enum Splits {
     /// Natural splits computed by the layout reader (e.g., computing splits across different-sized
     /// column chunks).
     Natural(BTreeSet<u64>),
```
joseph-isaacs, line 20: Registered?

joseph-isaacs (approved):

PR #4901, opened 2025-10-09
chore: vertical imports

a10y:
Switch to vertical imports.

Had a random 2 secs to try this after seeing the slack convo.

Uses more vertical space, but makes diffs easier to parse


connortsui20:
It's free real estate

connortsui20 (commented):
Nit: could you separate the rustfmt config change (I'm guessing thats what changed) into a separate commit so we can see what actually changed?

a10y:
@connortsui20 commits are split now

connortsui20 (approved):
I'm very much in favor of this, but this is such a large change that its probably good to get other people's opinions

a10y:
Yea, the only other one that I think might be be a good option is `HorizontalVertical`

image

connortsui20:
I'm in favor of all in over half measures, we can commit to preventing all merge conflicts here completely

PR #4902, opened 2025-10-09
feat: add concat expression to vortex-expr

blaginin:
Adds a new `concat` expression that concatenates multiple expressions of the same type into a single ChunkedArray. Each input expression becomes a chunk in the resulting ChunkedArray.

Features:
- Validates all child expressions have the same dtype
- Returns a ChunkedArray with each child as a chunk
- Comprehensive tests covering various usage patterns


vortex-expr/src/exprs/concat.rs
```diff
@@ -0,0 +1,294 @@
...
+use vortex_array::{ArrayRef, DeserializeMetadata, EmptyMetadata, IntoArray};
+use vortex_dtype::DType;
+use vortex_error::{VortexResult, vortex_bail};
+
+use crate::display::{DisplayAs, DisplayFormat};
+use crate::{AnalysisExpr, ExprEncodingRef, ExprId, ExprRef, IntoExpr, Scope, VTable, vtable};
+
+vtable!(Concat);
+
+/// Concatenate zero or more expressions into a single ChunkedArray.
```
joseph-isaacs: I don't think we need to require that a chunked array is returned, just an array containing all rows of the list of arrays?

vortex-expr/src/exprs/concat.rs
```diff
@@ -0,0 +1,294 @@
...
+/// use vortex_buffer::buffer;
+/// use vortex_expr::{ConcatExpr, Scope, lit};
+/// use vortex_scalar::Scalar;
+///
+/// let example = ConcatExpr::new(vec![
+///     lit(Scalar::from(100)),
+///     lit(Scalar::from(200)),
+///     lit(Scalar::from(300)),
+/// ]);
+/// let concatenated = example.evaluate(&Scope::new(buffer![0].into_array())).unwrap();
```
joseph-isaacs: `Scope::empty` not new

joseph-isaacs (approved):

PR #4905, opened 2025-10-09
chore[btrblocks]: move validity in `deltas` & enabled delta encoding

joseph-isaacs:
Think we need a fused RLE+Delta kernel?

We believe this Array was unused since it not never wired into the default compressor. If you find a problem reading a `DeltaArray` please reach out.

### DIFF
--- a/vortex-btrblocks/src/rle.rs
+++ b/vortex-btrblocks/src/rle.rs
@@ -150,8 +150,7 @@ pub fn try_compress_delta(
     excludes: &[IntCode],
 ) -> VortexResult<ArrayRef> {
     let (bases, deltas) = delta_compress(primitive_array)?;
-    let compressed_bases =
-        IntCompressor::compress(&bases, is_sample, allowed_cascading, excludes)?;
+    let compressed_bases = IntCompressor::compress(&bases, is_sample, allowed_cascading, excludes)?;
     let compressed_deltas =
         IntCompressor::compress_no_dict(&deltas, is_sample, allowed_cascading, excludes)?;
 
--- a/vortex-btrblocks/src/rle.rs
+++ b/vortex-btrblocks/src/rle.rs
@@ -112,6 +112,8 @@ where
             &new_excludes,
         )?;
 
+        // Delta in an unstable encoding, once we deem it stable we can switch over to this always.
+        #[cfg(feature = "unstable_encodings")]
         // For indices and offsets, we always use integer compression without dictionary encoding.
         let compressed_indices = try_compress_delta(
             &rle_array.indices().to_primitive().narrow()?,
@@ -120,8 +122,9 @@ where
             &[],
         )?;
 
+        #[cfg(not(feature = "unstable_encodings"))]
         let compressed_offsets = IntCompressor::compress_no_dict(
-            &rle_array.values_idx_offsets().to_primitive().narrow()?,
+            &rle_array.values_idx_offsets().to_primitive().downcast()?,
             is_sample,
             allowed_cascading - 1,
             &[],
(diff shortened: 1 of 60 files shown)


robert3005 (approved):

joseph-isaacs:
Gonna wait for @0ax1 before merging.

vortex-btrblocks/src/rle.rs
```diff
@@ -141,3 +142,23 @@ where
...
+    let (bases, deltas) = delta_compress(primitive_array)?;
+    let compressed_bases = IntCompressor::compress(&bases, is_sample, allowed_cascading, excludes)?;
+    let compressed_deltas =
+        IntCompressor::compress_no_dict(&deltas, is_sample, allowed_cascading, excludes)?;
+
+    DeltaArray::try_from_delta_compress_parts(
+        compressed_bases,
+        compressed_deltas,
+        // asserted that null_count == 0.
+        primitive_array.validity().clone(),
```
0ax1:
```
 // asserted that null_count == 0.
```
Where do we assert that? We know that RLE encode can never produce nullable indices. 

Looking at `delta_compress`: 
```
pub fn delta_compress(array: &PrimitiveArray) -> VortexResult<(PrimitiveArray, PrimitiveArray)> {
    // Compress the filled array
    let (bases, deltas) = match_each_unsigned_integer_ptype!(array.ptype(), |T| {
        const LANES: usize = T::LANES;
        let (bases, deltas) = compress_primitive::<T, LANES>(array.as_slice::<T>());
        let (base_validity, delta_validity) =
            if array.validity().nullability() != Nullability::NonNullable {
                (Validity::AllValid, Validity::AllValid)
            } else {
                (Validity::NonNullable, Validity::NonNullable)
            };
        (
            // To preserve nullability, we include Validity
            PrimitiveArray::new(bases, base_validity),
            PrimitiveArray::new(deltas, delta_validity),
        )
    });

    Ok((bases, deltas))
}
```

Should we actually bail if the validity null count is != 0?

```
if array.validity().null_count() > 0 {
    vortex_bail!("Delta compression with null values supported");
}
```

robert3005: We will produce nullable indices if we start with a nullable array, in deltas case the validity is stored on the array and the compression doesn't cover validity. This comment is incorrect

0ax1 (commented):
One remark but looks good to me. 

On a diff note, I'm still wondering whether we should try to eo the untranspose in-place rather than separating input and output buffers. In that case we could also decompress delta directly into the final output buffer, without a temp buffer.

vortex-btrblocks/src/rle.rs
0ax1: @joseph-isaacs tl;dr drop the comment

joseph-isaacs:
Am I right in saying we can change  the encoding delta since it was never used

vortex-btrblocks/src/rle.rs
joseph-isaacs: I'll look again at delta. The validity in the outer array confused me

0ax1:
> Am I right in saying we can change the encoding delta since it was never used

That's also my understanding.

0ax1 (approved):
🚢 it

robert3005:
Why did you end up adding a feature flag?

joseph-isaacs:
How confident are we in delta? @lwwmanning and me are not fully yet.

Also I think in general this might be a good approach for now encodings we add?

robert3005:
The thing I’m worried about is that we will never remove the feature flag. What is the condition that’s required to be met for the feature flag to be removed?

PR #4906, opened 2025-10-09
feat: Allow read from stream

m7kss1:


joseph-isaacs:
Mind signing the commits to make (DCO https://github.com/vortex-data/vortex/pull/4906/checks?check_run_id=52383374445) happy

joseph-isaacs:
We plan on in the future using `vortex-ipc` to stream serialized vortex arrays, however for now we don't have so I think adding a way to create a VortexFile using a byte array seems reasonable.

vortex-cxx/cpp/src/file.cpp
```diff
@@ -8,6 +8,15 @@
 
 namespace vortex {
 
+VortexFile VortexFile::Open(const uint8_t *data, size_t length) {
```
0ax1, line 11: `std::span` would be preferable here, though we currently compile with C++17 (span was introduced with C++20). @joseph-isaacs We should discuss bumping to C++20 or considering a library like abseil in order to use span. This is of course not in scope of this PR.

joseph-isaacs (approved):
Thanks for contribution

PR #4909, opened 2025-10-10
Propegate output_ordering to scan

AdamGS:
Fix a bug where datafusion scans the provide `output_ordering` assume the stream itself is provided in-order, and we explicitly don't guarantee the stream we return from files is in-order.

### DIFF
--- a/vortex-datafusion/src/convert/exprs.rs
+++ b/vortex-datafusion/src/convert/exprs.rs
@@ -372,12 +372,14 @@ mod tests {
 
         let result = ExprRef::try_from_df(&like_expr).unwrap();
 
-        assert_snapshot!(result.display_tree().to_string(), @r#"
-        Like
-        ├── child: GetItem(text_col)
-        │   └── Root
-        └── pattern: Literal(value: "test%", dtype: utf8)
-        "#);
+        insta::allow_duplicates! {
+            assert_snapshot!(result.display_tree().to_string(), @r#"
+            Like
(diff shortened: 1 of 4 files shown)


vortex-datafusion/src/convert/exprs.rs
```diff
@@ -372,12 +372,14 @@ mod tests {
 
         let result = ExprRef::try_from_df(&like_expr).unwrap();
 
-        assert_snapshot!(result.display_tree().to_string(), @r#"
-        Like
-        ├── child: GetItem(text_col)
-        │   └── Root
-        └── pattern: Literal(value: "test%", dtype: utf8)
-        "#);
+        insta::allow_duplicates! {
```
AdamGS, line 375: This makes `cargo test -p vortex-datafusion` work

vortex-datafusion/src/persistent/mod.rs
```diff
@@ -188,4 +191,80 @@ mod tests {
...
+        │  c1 ASC NULLS LASTlimit:  │
+        │             3             │
+        └─────────────┬─────────────┘
+        ┌─────────────┴─────────────┐
+        │       DataSourceExec      │
+        │    --------------------   │
+        │          files: 3         │
+        │       format: vortex      │
+        └───────────────────────────┘
+        ");
```
AdamGS, line 252: Shows plan is correct - it has no explicit `SortExec` node.

robert3005 (approved):

PR #4913, opened 2025-10-10
feat: constant time patches slicing

0ax1:
Follow up PRs will address `take`, `filter` and `mask`, as well as  chunk offset construction in ALP.

### DIFF
--- a/encodings/fastlanes/src/bitpacking/compress.rs
+++ b/encodings/fastlanes/src/bitpacking/compress.rs
@@ -260,15 +260,11 @@ where
     let mut chunk_offsets: BufferMut<u64> = BufferMut::with_capacity(total_chunks);
 
     let mut patch_idx = 0u64;
-    let mut current_chunk_filled = 0;
 
     for (idx, value) in data.iter().enumerate() {
-        let chunk = idx / 1024;
-
-        if current_chunk_filled <= chunk {
-            let count = chunk - current_chunk_filled + 1;
-            chunk_offsets.push_n(patch_idx, count);
-            current_chunk_filled = chunk + 1;
+        if (idx % 1024) == 0 {
+            // Record the patch index offset for each chunk.
+            chunk_offsets.push(patch_idx);
         }
 
         if (value.leading_zeros() as usize) < T::PTYPE.bit_width() - bit_width as usize
@@ -278,11 +274,6 @@ where
             values.push(*value);
             patch_idx += 1;
         }
-    }
-
-    let remaining = total_chunks - chunk_offsets.len();
-    if remaining > 0 {
-        chunk_offsets.push_n(patch_idx, remaining);
     }
 
     (!indices.is_empty()).then(|| {
--- a/encodings/fastlanes/src/bitpacking/compress.rs
+++ b/encodings/fastlanes/src/bitpacking/compress.rs
@@ -256,13 +256,33 @@ where
     let mut indices: BufferMut<P> = BufferMut::with_capacity(num_exceptions_hint);
     let mut values: BufferMut<T> = BufferMut::with_capacity(num_exceptions_hint);
 
-    for (i, v) in data.iter().enumerate() {
-        if (v.leading_zeros() as usize) < T::PTYPE.bit_width() - bit_width as usize
-            && validity_mask.value(i)
-        {
-            indices.push(P::from(i).vortex_expect("cast index from usize"));
-            values.push(*v);
+    let total_chunks = data.len().div_ceil(1024);
+    let mut chunk_offsets: BufferMut<u64> = BufferMut::with_capacity(total_chunks);
+
+    let mut patch_idx = 0u64;
+    let mut current_chunk_filled = 0;
+
+    for (idx, value) in data.iter().enumerate() {
+        let chunk = idx / 1024;
+
+        if current_chunk_filled <= chunk {
+            let count = chunk - current_chunk_filled + 1;
+            chunk_offsets.push_n(patch_idx, count);
+            current_chunk_filled = chunk + 1;
         }
+
+        if (value.leading_zeros() as usize) < T::PTYPE.bit_width() - bit_width as usize
+            && validity_mask.value(idx)
+        {
+            indices.push(P::from(idx).vortex_expect("cast index from usize"));
+            values.push(*value);
+            patch_idx += 1;
+        }
+    }
+
+    let remaining = total_chunks - chunk_offsets.len();
+    if remaining > 0 {
+        chunk_offsets.push_n(patch_idx, remaining);
     }
 
     (!indices.is_empty()).then(|| {
@@ -271,6 +291,7 @@ where
             0,
             indices.into_array(),
             PrimitiveArray::new(values, patch_validity).into_array(),
+            Some(chunk_offsets.into_array()),
         )
     })
 }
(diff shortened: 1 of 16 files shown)


encodings/fastlanes/src/bitpacking/serde.rs
```diff
@@ -78,7 +78,22 @@ impl SerdeVTable<BitPackedVTable> for BitPackedVTable {
             .map(|p| {
                 let indices = children.get(0, &p.indices_dtype(), p.len())?;
                 let values = children.get(1, dtype, p.len())?;
-                Ok::<_, VortexError>(Patches::new(len, p.offset(), indices, values))
+                let chunk_offsets = if children.len() >= 3 {
+                    let dtype = p
+                        .chunk_offsets_dtype()
+                        .vortex_expect("chunk offset dtype not found");
+                    Some(children.get(2, &dtype, p.chunk_offsets_len() as usize)?)
```
0ax1: tpcds fails on `chunk offset dtype not found`.

should this be
```
let chunk_offsets = p
    .chunk_offsets_dtype()
    .map(|dtype| children.get(2, &dtype, p.chunk_offsets_len() as usize))
    .transpose()?;
```
?

0ax1: => yes, bc validity might be the third entry

vortex-array/src/patches.rs
```diff
@@ -108,6 +155,8 @@ impl Patches {
             offset,
             indices,
             values,
+            chunk_offsets,
+            offset_within_chunk: Some(0),
```
0ax1, line 159: The assumption here is that `new` never creates a sliced Patches array (offset_within_chunk = 0).  

Also looking at the `offset` parameter, all call sites pass 0. Should we remove the `offset` parameter from `new` and pass `0` internally as a constant?

encodings/fastlanes/src/bitpacking/compress.rs
```diff
@@ -256,21 +256,38 @@ where
...
-    for (i, v) in data.iter().enumerate() {
-        if (v.leading_zeros() as usize) < T::PTYPE.bit_width() - bit_width as usize
-            && validity_mask.value(i)
+    let total_chunks = data.len().div_ceil(1024);
+    let mut chunk_offsets: BufferMut<u64> = BufferMut::with_capacity(total_chunks);
+
+    let mut patch_idx = 0u64;
+
+    for (idx, value) in data.iter().enumerate() {
+        if (idx % 1024) == 0 {
```
0ax1, line 263: The compiler will optimize this to `& (1024 - 1)`.

vortex-array/src/patches.rs
```diff
@@ -33,15 +39,31 @@ pub struct PatchesMetadata {
     offset: u64,
     #[prost(enumeration = "PType", tag = "3")]
     indices_ptype: i32,
+    #[prost(uint64, optional, tag = "4")]
+    chunk_offsets_len: Option<u64>,
+    #[prost(enumeration = "PType", optional, tag = "5")]
+    chunk_offsets_ptype: Option<i32>,
+    #[prost(uint64, optional, tag = "6")]
+    offset_within_chunk: Option<u64>,
```
joseph-isaacs, line 47: you can set a default prost value. Would that make sense vs this?

0ax1: Hmm, so it's either all of the chunk offset fields or none of them are set. We could rely on only checking one of them for `None` and use defaults for the 2 others. I do think it's nice to though to explicitly encode if these fields were not set.

encodings/fastlanes/src/bitpacking/serde.rs
joseph-isaacs: is this still true?

joseph-isaacs (commented):
I think we should write a comment explaining the layout of bitpacked array it has variable length children (on disk)

vortex-array/src/patches.rs
```diff
@@ -254,6 +326,83 @@ impl Patches {
         )
```
joseph-isaacs: want to share the subrange search with the below fn?

vortex-array/src/patches.rs
```diff
@@ -336,42 +485,56 @@ impl Patches {
...
+        let filtered_values = filter(&self.values, &filter_mask)?;
+
+        Ok(Some(Self {
+            array_len: self.array_len,
+            offset: self.offset,
+            indices: filtered_indices,
+            values: filtered_values,
+            // TODO(0ax1): chunk offsets
+            chunk_offsets: None,
+            offset_within_chunk: self.offset_within_chunk,
```
joseph-isaacs, line 509: you throw away chunk_offsets cause they are incorrect after filtering?

joseph-isaacs: could you expand on the TODO?

0ax1: > are incorrect after filtering?

Yes, exactly.

encodings/fastlanes/src/bitpacking/serde.rs
0ax1: ~~This needs to be adjusted.~~

vortex-array/src/patches.rs
0ax1: Not sure I understand what you mean. So we check whether `self.chunk_offsets` is set first and if so, we binary search within a single chunk, and early return. But if `self.chunk_offsets` is not set we need to fall back to the previous binary search in the full range. This for example happens if a Vortex file was written with the previous bitpack encoding.

vortex-array/src/patches.rs
```diff
@@ -336,42 +478,56 @@ impl Patches {
...
+        let chunk_offsets = self.chunk_offsets.as_ref().map(|chunk_offsets| {
+            let chunk_relative_offset = self.offset % PATCH_CHUNK_SIZE;
+            let chunk_start_idx = (chunk_relative_offset + range.start) / PATCH_CHUNK_SIZE;
+            let chunk_end_idx = (chunk_relative_offset + range.end).div_ceil(PATCH_CHUNK_SIZE);
+            chunk_offsets.slice(chunk_start_idx..chunk_end_idx)
+        });
+
+        let offset_within_chunk = chunk_offsets.as_ref().map(|chunk_offsets| {
+            let base_offset = chunk_offsets
+                .scalar_at(0)
```
0ax1, line 534: I'll make chunk offset lookup a fn on `Patches`, as this is used in diff places.

vortex-array/src/patches.rs
0ax1: The offset array gets created during encoding, not e.g. on deserialization in case it wasn't written to file before.

vortex-array/src/patches.rs
0ax1: Ideally, `Patches` would have a filter fn which whilst filtering adjusts the offsets.

vortex-array/src/patches.rs
joseph-isaacs: I am saying you can extract a common method between this and the below function to binary search within a random of an array (here the range is the whole buffer).

encodings/fastlanes/src/bitpacking/serde.rs
joseph-isaacs: How was this not found by CI?

joseph-isaacs: Can you verify this works with data generated without the PR (for clickbench, tpch and tpcds)?

0ax1: done

vortex-array/src/patches.rs
0ax1: Expanded on the TODOs.

0ax1:
> I think we should write a comment explaining the layout of bitpacked array it has variable length children (on disk)

done

vortex-array/src/patches.rs
```diff
@@ -204,12 +275,20 @@ impl Patches {
...
-            len: self.indices.len() as u64,
-            offset: self.offset as u64,
-            indices_ptype: PType::try_from(self.indices.dtype()).vortex_expect("primitive indices")
-                as i32,
-        })
+        let chunk_offsets_len = self.chunk_offsets.as_ref().map(|co| co.len());
+        let chunk_offsets_ptype = self
+            .chunk_offsets
+            .as_ref()
+            .and_then(|co| PType::try_from(co.dtype()).ok());
```
robert3005:
```suggestion
            .and_then(|co| co.dtype().as_ptype());
```

vortex-array/src/patches.rs
```diff
@@ -204,12 +275,20 @@ impl Patches {
...
+
+        Ok(PatchesMetadata::new(
+            self.indices.len(),
+            self.offset,
+            PType::try_from(self.indices.dtype()).vortex_expect("primitive indices"),
```
robert3005: I think we can improve this to do the same thing as above

vortex-array/src/patches.rs
```diff
@@ -248,10 +355,63 @@ impl Patches {
...
+    fn search_index_chunked(&self, index: usize) -> SearchResult {
+        let Some(chunk_offsets) = &self.chunk_offsets else {
+            vortex_panic!("chunk_offsets is required to be set")
+        };
+
+        let Some(offset_within_chunk) = self.offset_within_chunk else {
+            vortex_panic!("offset_within_chunk is required to be set")
+        };
+
+        let chunk_start_idx = (index + self.offset % PATCH_CHUNK_SIZE) / PATCH_CHUNK_SIZE;
```
robert3005: can you skip the mod here? I think the result of following is the same
```suggestion
        let chunk_start_idx = (index + self.offset) / PATCH_CHUNK_SIZE;
```

vortex-array/src/patches.rs
```diff
@@ -248,10 +355,63 @@ impl Patches {
...
+    ///
+    /// First determines which chunk the target index falls into, then performs
+    /// a binary search within that chunk's range.
+    ///
+    /// Returns a [`SearchResult`] indicating either the exact patch index if found,
+    /// or the insertion point if not found.
+    ///
+    /// # Panics
+    /// Panics if `chunk_offsets` or `offset_within_chunk` are not set.
+    fn search_index_chunked(&self, index: usize) -> SearchResult {
```
robert3005, line 370: Need to handle index > len

vortex-array/src/patches.rs
```diff
@@ -248,10 +355,63 @@ impl Patches {
...
+        let patches_start_idx = (self.chunk_offset_at(chunk_start_idx) - base_offset)
+            // Chunk offsets are only sliced off in case the slice is fully
+            // outside of the chunk range.
+            //
+            // Though the range for indices and values is sliced in terms of
+            // individual elements, not chunks. To account for that we do a
+            // saturating sub when adjusting the indices based on the chunk offset.
+            .saturating_sub(offset_within_chunk);
+
+        let patches_end_idx = if chunk_end_idx < chunk_offsets.len() {
```
robert3005: I think we can slightly rewrite this since the notion that we have chunk start and end is a bit misleading 
```suggestion
        let patches_end_idx = if chunk_start_idx < chunk_offsets.len() - 1 {
```
and rename chunk_start_idx to chunk_idx

vortex-array/src/patches.rs
```diff
@@ -336,42 +496,56 @@ impl Patches {
...
-        let indices = self.indices().slice(patch_start..patch_stop);
-
-        Some(Self::new(
-            range.len(),
-            range.start + self.offset(),
+        let values = self.values().slice(slice_start_idx..slice_end_idx);
+        let indices = self.indices().slice(slice_start_idx..slice_end_idx);
+
+        let chunk_offsets = self.chunk_offsets.as_ref().map(|chunk_offsets| {
+            let chunk_relative_offset = self.offset % PATCH_CHUNK_SIZE;
```
robert3005, line 526: I think you can skip the mod here

robert3005 (changes requested):
there's some style comments and one case that I think is not being handled

vortex-array/src/patches.rs
0ax1: No, `self.offset` is absolute and I want to find the offset within the chunk.

vortex-array/src/patches.rs
0ax1: Same here, `self.offset` is absolute, but I want to get the relative chunk offset.

vortex-array/src/patches.rs
robert3005: I see, forgot that everything apart from offset is sliced

vortex-array/src/patches.rs
robert3005: Can't do this since the offsets have already been sliced but chunk is absolute

vortex-array/src/patches.rs
0ax1: This case returns the insertion point that matches the indices length. 

```
     #[test]
    fn test_index_larger_than_length() {
        let chunk_offsets = buffer![0u64].into_array();
        let indices = buffer![1023u64].into_array();
        let values = buffer![42i32].into_array();
        let patches = Patches::new(1024, 0, indices, values, Some(chunk_offsets));
        assert_eq!(patches.search_index(2048), SearchResult::NotFound(1));
    }
```
I'll add this test.

vortex-array/src/patches.rs
0ax1: done

vortex-array/src/patches.rs
robert3005: Yes, but does this work the same way if you have offsets

0ax1: done

robert3005 (approved):

vortex-array/src/patches.rs
```diff
@@ -378,6 +376,10 @@ impl Patches {
             vortex_panic!("offset_within_chunk is required to be set")
         };
 
+        if index >= self.indices().len() {
```
robert3005: this is array_len not indices len

encodings/fastlanes/src/bitpacking/serde.rs
```diff
@@ -66,9 +71,13 @@ impl SerdeVTable<BitPackedVTable> for BitPackedVTable {
...
+        let validity = if let Some(patches_meta) = &metadata.patches {
+            let validity_idx = if patches_meta.chunk_offsets_dtype().is_some() {
+                3
+            } else {
+                2
+            };
+            load_validity(validity_idx)?
         } else {
             load_validity(0)?
         };
```
joseph-isaacs: can we clean this up a little
```suggestion
        let validity_idx = if let Some(patches_meta) = &metadata.patches {
            if patches_meta.chunk_offsets_dtype().is_some() {
                3
            } else {
                2
            }
} else {
0
};
let validity = load_validity(validity_idx)?
```

vortex-array/src/patches.rs
0ax1: done

encodings/fastlanes/src/bitpacking/serde.rs
0ax1: done

PR #4915, opened 2025-10-10
chore: allow `pack` to return a different array from scope len

blaginin:
https://spiraldb.slack.com/archives/C07BV3GKAJ2/p1760103400110299


joseph-isaacs:
We cannot do this.

blaginin:
we cannot do this!

PR #4920, opened 2025-10-10
feat[gpu]: jit kernels

joseph-isaacs:
Add support for jitting bitpacking, FoR and alp kernels to cuda.


vortex-gpu/src/jit/type_.rs
```diff
@@ -0,0 +1,33 @@
...
+            PType::U16 => "unsigned short",
+            PType::U32 => "unsigned int",
+            PType::U64 => "unsigned long long",
+            PType::I8 => "char",
+            PType::I16 => "short",
+            PType::I32 => "int",
+            PType::I64 => "long long",
+            PType::F32 => "float",
+            PType::F64 => "double",
+            PType::F16 => todo!(),
```
robert3005, line 30: this todo made me go down the rabbit hole https://stackoverflow.com/questions/32735292/can-anyone-provide-sample-code-demonstrating-the-use-of-16-bit-floating-point-in. You need an include and then you can use `half` but you really want to use `half2`, i.e. pack two halfs into 32 bits

robert3005:
Not quite sure how continuation passing style would look like differently? You are nesting the code emitting statements recursively at any given step. You could have a function that will in turn when called emit the code instead of directly emitting the code BUT I can't quite assess in abstract if its easier... maybe?

robert3005 (approved):

PR #4921, opened 2025-10-10
Chore: Clean up benchmarks website

connortsui20:
Because lance is so much slower in some of the queries than every other target, it makes a lot of charts harder to see everything.

EDIT: They are actually not showing up at all on the TPCH

EDIT AGAGIN: I'm going to use this PR to fix the two other problems. The first is that because we do not run lance on every PR the summaries will disappear. The second is that TPC-H queries are not showing up.


robert3005 (approved):

robert3005:
I am surprised we don't see some negative lines here. We rewrote the handling but old code should be removed? I doubt we just added 3k lines here

connortsui20:
Turns out it was just a lot of dead code

robert3005 (approved):

PR #4930, opened 2025-10-13
Chore: bump lance to version 2.0->2.1

connortsui20:
Following guidance from Weston, we can bump lance to v2.1, which has improvements in everything but might hit random access?


robert3005 (approved):

lwwmanning:
Thanks for the tips @westonpace

PR #4931, opened 2025-10-13
Release the GIL before block_on

### LINKED ISSUE #4895: vx.io.write gets stuck
### Describe the bug

I'm trying to convert a Parquet file into Vortex format, but vx.io.write is stuck, and I'm not sure how to debug.

vortex-data                        0.53.0

Python version is 3.12.3

pyarrow 21.0.0

I know for the toy example it's possible to read the entire Parquet file into memory, and then write to Vortex. But in production, there is not enough memory to hold an entire Parquet file.

### To Reproduce

cd /tmp && wget https://docs.vortex.dev/_static/example.parquet

```
import pyarrow.parquet as pq
import vortex as vx
import pyarrow as pa

def main():
    pf = pq.ParquetFile("/tmp/example.parquet")
    table = pf.read_row_group(0)
    schema = table.schema

    batches = pf.iter_batches(batch_size=1)
    batch_iter = pa.RecordBatchReader.from_batches(schema, batches)
    vx.io.write(batch_iter, "/tmp/file.vortex")

if __name__ == "__main__":
    main()
```

### Expected behavior

Script should work similar to the quick start example.

### Additional context

The script gets stuck at vx.io.write.

gatesn:
Fixes #4895

We were deadlocking when calling back into Python's parquet reader because we held the GIL within our Tokio runtime.
We should probably follow up with a pass to release the GIL in more places.

### DIFF
--- a/vortex-python/src/dataset.rs
+++ b/vortex-python/src/dataset.rs
@@ -11,7 +11,7 @@ use pyo3::prelude::*;
 use pyo3::types::PyString;
 use vortex::dtype::{FieldName, FieldNames};
 use vortex::error::VortexResult;
-use vortex::expr::{ExprRef, SelectExpr, root, select};
+use vortex::expr::{root, select, ExprRef, SelectExpr};
 use vortex::file::{VortexFile, VortexOpenOptions};
 use vortex::iter::ArrayIteratorExt;
 use vortex::scan::SplitBy;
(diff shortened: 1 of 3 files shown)


AdamGS:
I think it'll be nice to have some `block_on` function that does it and takes `Python`, forcing callers to do it correctly. You can also limit the exposure of the runtime which will make it more obvious.

gatesn:
Because we use into_array_iter and BlockingRuntime, it's not that easy. It can be quite abstract where the release is. We also want to release the GIL for non-block_on functions too, so not sure it's worth wrapping up

robert3005 (approved):

PR #4932, opened 2025-10-13
perf: annotate primitive array fns with `#[inline]`

0ax1:
Annotates small functions defined on `PrimitiveArray` with `#[inline]` in case they're good candidates for performance improvement.


0ax1:
Hard to get a read on whether this is an improvement due to the jitter. But closing as it is not a clear yes.

PR #4934, opened 2025-10-13
Fix: remove CMSProvider dataset for lance v2.1

connortsui20:
Issue filed directly on lance: https://github.com/lancedb/lance/issues/4943


connortsui20:
Instead we can just run Lance v2.0 for our compression benchmarks

PR #4938, opened 2025-10-13
Chore: feature flag lance

connortsui20:
Also modifies the actions so that we only run lance benchmarks in 2 places: _after_ a PR is merged and during nightly benchmarks.

Also let's hold off on merging this until tomorrow just in case it breaks something on the website.

Also note that the benchmark summaries on github (below) are going to be off since they no longer include lance

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,3 +1,5 @@
+# Runs after every commit to `develop` (or in other words, _after_ every pull request merges).
+
 name: Benchmarks
 
 on:
@@ -65,8 +67,9 @@ jobs:
         shell: bash
         env:
           RUSTFLAGS: "-C target-cpu=native -C force-frame-pointers=yes"
+        # The main difference between this and `bench-pr.yml` is that we add the `lance` feature.
         run: |
-          cargo build --bin ${{ matrix.benchmark.id }} --package bench-vortex --profile release_debug
+          cargo build --bin ${{ matrix.benchmark.id }} --package bench-vortex --profile release_debug --features lance
 
       - name: Setup Polar Signals
         uses: polarsignals/gh-actions-ps-profiling@v0.6.0
@@ -82,7 +85,7 @@ jobs:
         env:
           RUST_BACKTRACE: full
         run: |
-          target/release_debug/${{ matrix.benchmark.id }} -d gh-json -o ${{ matrix.benchmark.id }}.json
+          target/release_debug/${{ matrix.benchmark.id }} -d gh-json -o ${{ matrix.benchmark.id }}.json --formats parquet,lance,vortex
 
       - name: Setup AWS CLI
         uses: aws-actions/configure-aws-credentials@v5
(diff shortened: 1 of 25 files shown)


.github/workflows/bench.yml
connortsui20, line 1: The `benchmark_matrix` is copied and pasted from the default, but with `datafusion:lance` added in the first few benchmarks.

.github/workflows/bench.yml
```diff
@@ -99,3 +102,72 @@ jobs:
     secrets: inherit
     with:
       mode: "develop"
+      benchmark_matrix: |
```
0ax1, line 105: Maybe it's time to consider something along the lines of https://carvel.dev/ytt/ soon as our GH actions setup gets more complex.

bench-vortex/src/compress/bench.rs
```diff
@@ -2,16 +2,23 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use std::borrow::Cow;
+#[cfg(not(feature = "lance"))]
+use std::fmt;
+#[cfg(feature = "lance")]
 use std::path::PathBuf;
+#[cfg(feature = "lance")]
```
0ax1: Could consider grouping all lance dependent `use` statements:
```
#[cfg(feature = "lance")]
use {
    std::path::PathBuf,
    std::sync::Arc,
    std::{fmt, fs},
    arrow_array::RecordBatch,
    parking_lot::Mutex,
};
```

bench-vortex/src/compress/bench.rs
```diff
@@ -380,42 +394,45 @@ pub fn calculate_ratios(
...
-            name: format!("vortex:lance size/{bench_name}"),
-            format: Format::OnDiskVortex,
-            unit: Cow::from("ratio"),
-            value: *vortex_size as f64 / *lance_size as f64,
-        });
-    }
+    #[cfg(feature = "lance")]
+    {
+        // Size ratio: vortex vs lance.
+        if let (Some(vortex_size), Some(lance_size)) = (
```
0ax1: What do you think about making 2 functions out of this: `calculate_vortex_parquet_ratios`, `calculate_vortex_lance_ratios` or something along those lines?

bench-vortex/src/datasets/file.rs
```diff
@@ -11,12 +11,15 @@ use datafusion::datasource::listing::{
 };
 use datafusion::prelude::SessionContext;
 use glob::Pattern;
+#[cfg(feature = "lance")]
 use lance::datafusion::LanceTableProvider;
+#[cfg(feature = "lance")]
```
0ax1: Same comment as before, I think it's worth grouping these use statements and then use a single `#[cfg(feature = "lance")]`.

bench-vortex/src/engines/ddb/mod.rs
```diff
@@ -147,6 +147,7 @@ impl DuckDBCtx {
         let object = match file_format {
             Format::Parquet | Format::OnDiskVortex | Format::VortexCompact => DuckDBObject::View,
             Format::OnDiskDuckDB => DuckDBObject::Table,
+            #[cfg(feature = "lance")]
```
0ax1, line 150: I guess the enum variant doesn't need to be conditionally compiled. I think overall the current conditional compilation setup is quite noisy. If not strictly necessary it'd be nice if we don't have to sprinkle  `#[cfg`s.

.github/workflows/bench.yml
connortsui20: I filed an issue #4941

bench-vortex/src/engines/ddb/mod.rs
connortsui20: I think with the imports grouping it is less noisy? I actually like the verbosity here because it makes it extra clear what is not getting compiled, but I don't feel super strongly about this.

0ax1: Fair, but yeah that's the tradeoff here explicit vs. lower noise level.

bench-vortex/src/lib.rs
```diff
@@ -159,6 +159,7 @@ impl Format {
             Format::OnDiskVortex => "vortex",
             Format::VortexCompact => "vortex",
             Format::OnDiskDuckDB => "duckdb",
+            #[cfg(feature = "lance")]
```
0ax1, line 162: Actually looking again at the `Format::Lance` matches, I'm getting more biased towards not feature flagging the variant.

0ax1 (approved):
Looks good! I think though that it is worth considering not feature flagging the enum variant.

joseph-isaacs (approved):
Didn't look, but LGTM

PR #4939, opened 2025-10-14
fix[vortex-{scan,io,datafusion}]: propagate tracing spans

asubiotto:
This change correctly connects object storage spans with the expected parent spans. Previously, object store spans were orphaned.

### DIFF
--- a/vortex-datafusion/src/persistent/opener.rs
+++ b/vortex-datafusion/src/persistent/opener.rs
@@ -19,6 +19,7 @@ use datafusion_pruning::FilePruner;
 use futures::{FutureExt, StreamExt, TryStreamExt, stream};
 use object_store::ObjectStore;
 use object_store::path::Path;
+use tracing::Instrument;
 use vortex::dtype::FieldName;
 use vortex::error::VortexError;
 use vortex::expr::{root, select};
@@ -266,6 +267,7 @@ impl FileOpener for VortexOpener {
 
             Ok(stream)
         }
+        .in_current_span()
         .boxed())
     }
 }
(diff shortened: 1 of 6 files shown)


AdamGS (approved):
LGTM. The next major DataFusion release will expose its tracing API which will make the `vortex-datafusion` side of this use that, but as we already have `tracing` wired in a few places this probably an improvement.

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -266,6 +267,7 @@ impl FileOpener for VortexOpener {
 
             Ok(stream)
         }
+        .instrument(tracing::Span::current())
```
AdamGS: nit `.in_current_span()`

asubiotto: Done! Forgot about this one, thanks.

asubiotto:
> LGTM. The next major DataFusion release will expose its tracing API which will make the `vortex-datafusion` side of this use that, but as we already have `tracing` wired in a few places this probably an improvement.

TIL

AdamGS:
They hide it pretty well 😅, I would've loved if the whole ecosystem used `tracing` (or any other facade) but seems like we're not quite there yet

asubiotto:
> They hide it pretty well 😅, I would've loved if the whole ecosystem used `tracing` (or any other facade) but seems like we're not quite there yet

Maybe one day, but I wouldn't hold my breath. Tracing, regardless of language, always feels (unfortunately) like an afterthought.

PR #4940, opened 2025-10-14
Use BitBuffer instead of BooleanBuffer

robert3005:
Replace usages of BooleanBuffer with our own BitBuffer. This lets us have our own backing Buffer for BitBuffer and implement additional optimisations for use in vortex.

previously #2456 

This pr requires resolving outstanding performance issues before merging
1. Make iterators cheaper to construct, likely can't be owned
2. Improve append_packed_range performance. Arrow has a function that is faster than bitvec crate


0ax1 (commented):
Can we add a small PR description on why we're doing this change, and what the properties are of BitBuffer compared BooleanBuffer?

What's our take on the Codspeed results here?

robert3005:
This has been on the backburner since February. I will make some notes. We have separately merged just the core logic but merging this is blocked on fixing codspeed results.

gatesn:
FWIW, the long driving motivation for this was to have BitBufferMut present a nice interface for allocating, creating, and mutating boolean buffers.

BitBuffer vs BooleanBuffer is essentially no difference.

encodings/dict/src/builders/primitive.rs
```diff
@@ -64,7 +64,7 @@ where
         Self {
             lookup: HashMap::with_hasher(FxBuildHasher),
             values: BufferMut::<T>::empty(),
-            values_nulls: NullBufferBuilder::new(0),
+            values_nulls: BitBufferMut::empty(),
```
gatesn, line 67: TODO: use MaskMut from https://github.com/vortex-data/vortex/pull/4954 will have the same lazy allocation logic

gatesn (approved):

Issue #4889, opened 2025-10-08
Better data layout for nested/repeated schemas


a10y:
**Problem**: Currently, Vortex has poor support for nested data types.

Consider the following schema:

```
D describe events;
┌─────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────┬─────────┬─────────┬─────────┐
│ column_name │                                                                 column_type                                                                 │  null   │   key   │ default │  extra  │
│   varchar   │                                                                   varchar                                                                   │ varchar │ varchar │ varchar │ varchar │
├─────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┤
│ id          │ VARCHAR                                                                                                                                     │ YES     │ NULL    │ NULL    │ NULL    │
│ type        │ VARCHAR                                                                                                                                     │ YES     │ NULL    │ NULL    │ NULL    │
│ actor       │ STRUCT(id BIGINT, login VARCHAR, display_login VARCHAR, gravatar_id VARCHAR, url VARCHAR, avatar_url VARCHAR)                               │ YES     │ NULL    │ NULL    │ NULL    │
│ repo        │ STRUCT(id BIGINT, "name" VARCHAR, url VARCHAR)                                                                                              │ YES     │ NULL    │ NULL    │ NULL    │
│ payload     │ STRUCT("action" VARCHAR, issue STRUCT(url VARCHAR, repository_url VARCHAR, labels_url VARCHAR, comments_url VARCHAR, events_url VARCHAR, …  │ YES     │ NULL    │ NULL    │ NULL    │
│ public      │ BOOLEAN                                                                                                                                     │ YES     │ NULL    │ NULL    │ NULL    │
│ created_at  │ TIMESTAMP                                                                                                                                   │ YES     │ NULL    │ NULL    │ NULL    │
│ org         │ STRUCT(id BIGINT, login VARCHAR, gravatar_id VARCHAR, url VARCHAR, avatar_url VARCHAR)                                                      │ YES     │ NULL    │ NULL    │ NULL    │
└─────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────┴─────────┴─────────┴─────────┘
```

Currently the way our StructStrategy works, this is written with the following layout tree:

```
StructLayout
|__ "payload" child (Zoned) dtype=struct{id, name,}
[...]

PR #4942, opened 2025-10-14
feat: write nested struct layouts

a10y:
Part of #4889 

To better support writing nested data, here we update our builtin StructLayout in a backwards-compatible fashion.

- `StructStrategy` will shred struct fields into their own StructLayout recurseively
- StructLayout can support nullable structs now. It does this by writing a new child layout containing the validity buffer for nullable arrays

I use the `RealNest` dataset to evaluate, which contains a copy of ~200k github pull request webhook events. Nested struct layout reduces file size over the previous strategy by about ~10%, and also makes pushdown into the nested columns possible.

Some open questions

* The validity child requires some extra handling. It seems like the validity handling is very dependent on the expression being pushed down. For example if I'm doing a simple project of a child field, then adding the validity to the result is a simple masking operation. If I'm pushing down an `UNNEST` or something else that increases the result size, it is hard to map the validity buffer onto the projection_eval result
* I collect all validity chunks into a single buffer at write time. The idea being that it's better to access the struct validity as a single unit since it is much smaller than the data size. Assuming an 8MB target segment size, this lets us comfortably fit ~64mm rows into a single segment. Another alternative is to bring back roaring, or enable some other boolean compressors.


vortex-layout/src/layouts/struct_/reader.rs
```diff
@@ -249,7 +280,19 @@ impl LayoutReader for StructReader {
...
-        }
+        }?;
+
+        Ok(Box::pin(async move {
+            if let Some(validity_fut) = validity_fut {
+                let (validity, projection) = try_join!(validity_fut, projection_fut)?;
+                vortex_array::compute::mask(
+                    projection.as_ref(),
+                    &Mask::from_buffer(!validity.to_bool().boolean_buffer()),
+                )
```
a10y: this is only safe when we know that the projection has the same length as the struct itself. If we're projecting an `UNNEST(a)` that will not hold. We don't have UNNEST but i suspect we'll want to add it so we can pushdown into list elements

vortex-layout/src/layouts/collect.rs
```diff
@@ -0,0 +1,73 @@
...
+
+            let mut chunks = Vec::new();
+            let mut latest_sequence_id = None;
+            while let Some(chunk) = stream.next().await {
+                let (sequence_id, chunk) = chunk?;
+                latest_sequence_id = Some(sequence_id);
+                chunks.push(chunk);
+            }
+
+            let collected = ChunkedArray::try_new(chunks, _dtype)?.to_array();
```
robert3005, line 59: maybe we canonicalize here instead of writing a flat chunked array?

vortex-layout/src/layouts/struct_/writer.rs
```diff
@@ -78,21 +77,40 @@ impl LayoutStrategy for StructStrategy {
         }
 
         // stream<struct_chunk> -> stream<vec<column_chunk>>
-        let columns_vec_stream = stream.map(|chunk| {
+        let is_nullable = dtype.is_nullable();
+
+        // Write the entire stream as a single flat layout child
```
robert3005: this comment is wrong - you are partitioning the sequence pointer for each child here?

robert3005 (approved):
I think this makes sense. Some of the code comments are not correct though

vortex-layout/src/layouts/collect.rs
a10y: In practice that ends up happening b/c the `child` here is the CompressStrategy, so it will canonicalize + compress internally.

In general I tried to mirror this on the RepartitionStrategy which yields ChunkedArray as well. I think the LayoutStrategies are a bit weird tbh and there are a few unexpected things like this.

vortex-layout/src/layouts/struct_/writer.rs
a10y: Yep, this comment was from an old version where I did the CollectStrategy thing for validity inline

vortex-layout/src/layouts/collect.rs
robert3005: I thought repartition strategy will have a chunked strategy as a child so it will flatten the chunked array, in this case I think we go straight to flat.

vortex-layout/src/layouts/struct_/reader.rs
```diff
@@ -233,22 +266,65 @@ impl LayoutReader for StructReader {
...
+                        masked_fields,
+                        struct_array.len(),
+                        struct_array.validity().clone(),
+                    )?
+                    .into_array())
+                } else {
+                    // If the root expression was not a pack or merge, e.g. if it's something like
+                    // a get_item, then we apply the validity directly to the result
+                    vortex_array::compute::mask(array.as_ref(), &mask)
+                }
```
a10y, line 323: this was the simplest thing I could think of to get around the validity issues

a10y:
The last issue blocking correctness was support for pushing the validity into children based on the projection expression, which I think is fixed in the last commit.

Now, this brings us to a new issue: the gharchive dataset schema has 773 leaf fields, compared to just 8 top-level struct fields.

This leads to a majority of query execution time now being spent in the recursive `register_splits` phase:

image

The bulk seems to be inside of the ChunkReader initializing its lazy children, which goes through a OnceLock. There is a lot of time spent parked, indicating that there's a race from multiple threads to access the same children, which seems strange to me?

robert3005:
You have multiple threads reading different splits from layouts. Though register_splits should be single threaded. We had https://github.com/vortex-data/vortex/pull/3876 to change register_splits but at the time it wasn't a clear improvement

robert3005:
I think we should change the name string into something that doesn't allocate all the time, these strings are only used for debugging, we can have them be lazily built on display from a sequence/tree

a10y:
The latest few commits remove the contention in `register_splits` by pushing down the row range into split registration. So rather than having every partition register_splits for the entire file, it will only register its row range. That sped up query 0 by 83%, but didn't seem to make much of a difference on the others.

It's a bit bittersweet, but the reason the other queries seem to have gotten slower is because...we're compressing better.

We now have 773 leaf StructLayouts, most of which encode string data. Now that those are all being shredded out, we're compressing them with Dict+FSST much more frequently, which makes decoding slower:

image


image

So I think that the StructLayout is fine, it's actually the way we read back chunks from a DictLayout that may be problematic.

For Query 3 from gharchive: `select actor.login, count() as freq from events group by actor.login order by freq desc limit 10`, we end up reading with splitsize of 8192 rows. A single dictionary chunk in the `actor.login` column is nearly 300k codes, and 54,000 values. 

image

image


This means that to read just one single chunk with splits of 8192 rows, require 36 split tasks. Each split task returns a DictArray wrapping 8192 codes, and the entire 54,000 item FSST-encoded dictionary. Calling to_arrow on each of these DictArray chunks will in turn decode the entire FSST values dictionary.

So in effect, it's not that FSST is too slow, but rather that we're decoding the same FSST array 36 times.

a10y:
I suppose this is something that Operator work is supposed to help with? Though that really seems more focused on memoizing decodes in a single split task, rather than across split tasks?

robert3005:
If we only get 8192 codes we should only decode those codes. I know duckdb have a decompression cache on the fsst array to avoid any double decompression which is similar to us caching canonicalisations. I would still try first to avoid canonicalising values unconditionally and performing the take first

a10y:
This is about values not codes though. The issue is that every chunk goes through ToArrow, which canonicalizes. The entire values array gets decompressed for each chunk in that case.

As a test, when I switch the DictReader to instead eagerly canonicalize the values, Q3 goes from being 40% slower to nearly 2x faster than baseline.

I'm going to revert that commit for now, but I think this is a problem that is independent of nested struct splitting and is something we'll have to handle more fundamentally with DictLayouts.

robert3005 (approved):

PR #4945, opened 2025-10-14
fix: match_between rewriting

a10y:
Found this while fuzzing, the failure was

image

Was confused how the expression was getting reverse. The previous logic was just discarding information and picking the larger lit as the upper bound. This version should be more robust.

I've added a unit test for this case, it fails with the old version of `find_between` and succeeds now. I've also confirmed my fuzz artifact succeeds when I run with new code:

image

### DIFF
--- a/vortex-expr/src/transform/match_between.rs
+++ b/vortex-expr/src/transform/match_between.rs
@@ -5,7 +5,8 @@ use vortex_array::compute::{BetweenOptions, StrictComparison};
 
 use crate::forms::conjuncts;
 use crate::{
-    BetweenExpr, BinaryVTable, ExprRef, GetItemExpr, IntoExpr, LiteralExpr, Operator, and, lit,
+    BetweenExpr, BinaryExpr, BinaryVTable, ExprRef, GetItemVTable, IntoExpr, LiteralVTable,
+    Operator, and, lit,
 };
 
 /// This pass looks for expression of the form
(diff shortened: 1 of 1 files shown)


vortex-expr/src/transform/match_between.rs
```diff
@@ -53,90 +54,61 @@ fn maybe_match(lhs: &ExprRef, rhs: &ExprRef) -> Option<ExprRef> {
...
-    ) else {
-        return None;
+    // First, get both halves to have GetItem on the left
+    let lhs = match (
+        lhs.lhs().is::<GetItemVTable>(),
+        lhs.rhs().is::<GetItemVTable>(),
+    ) {
+        (true, false) => lhs.clone(),
+        (false, true) => BinaryExpr::new(lhs.rhs().clone(), lhs.op().swap()?, lhs.lhs().clone()),
+        _ => return None,
```
joseph-isaacs, line 64: I assume that a variable could be between two other variables?
`a <= x <= b`

a10y: This function only does the rewrite if it's variable between two lits

vortex-expr/src/transform/match_between.rs
```diff
@@ -53,90 +54,61 @@ fn maybe_match(lhs: &ExprRef, rhs: &ExprRef) -> Option<ExprRef> {
...
     };
 
-    let (left, left_op, right, right_op) = if left_lit.value() > right_lit.value() {
```
a10y, line 102: this was the bad piece

vortex-expr/src/transform/match_between.rs
joseph-isaacs: I see, I guess we can leave it like than, but there is not reason this needs to be the case

joseph-isaacs (approved):
LG

PR #4853, opened 2025-10-06
Feature: Make `ListView` canonical

connortsui20:
TODO

Issue #4844, opened 2025-10-03
Make all `_Scalar` types `Clone` and `Copy`


connortsui20:
Several of our `_Scalar` types (e.g., `ListScalar` and `BinaryScalar`) are not `Copy` because they store `Option<Arc<>>` instead of a `Option<&'a Arc<>>`. Given that all of these are supposed to be typed views into a `Scalar`, it doesn't make a lot of sense that these are not copyable like a reference.

```rust
#[derive(Debug)] // <- Can only derive `Clone` and cannot derive `Copy`
pub struct ListScalar<'a> {
    dtype: &'a DType,
    element_dtype: &'a Arc<DType>,
    elements: Option<Arc<[ScalarValue]>>,
    //               ^^^
}
```

I don't think that there is any reason that we cannot have all of our scalars be similar to `StructScalar` (which for some reason is not `Clone` or `Copy` but it can be):

```rust
#[derive(Debug)] // <- CAN derive `Clone` and `Copy`
pub struct StructScalar<'a> {
    dtype: &'a DType,
    fields: Option<&'a Arc<[ScalarValue]>>,
    //             ^^^^^^^
}
```

Unfortunately, making all of these `Copy` is not as simple as adding a `&'a` to every `Arc`, since the constructors require us passing in a reference to `ScalarValue` instead of an owned value. The constructors are quite inconsistent, and the methods on `InnerScalarValue` are also quite inconsistent:

```rust
impl InnerScalarValue {

    pub(crate) fn as_buffer(&self) -> VortexResult<Option<Arc<ByteBuffer>>> {}
    pub(crate) fn as_buffer_string(&self) -> VortexResult<Option<Arc<BufferString>>> {}
    pub(crate) fn as_list(&self) -> VortexResult<Option<&Arc<[ScalarValue]>>> {}
    //                                                  ^^^^
}
```

I think that all of our `_Scalar` types should be able to `#[derive(Clone, Copy)]`.

_I also still think that it doesn't make sense to have each scalar view have an internal null (with `Option` fields) and the fact that this is inconsistent is more evidence that we should really have `Option<ListScalar<'a>>` instead of an internal `Option<Arc<[ScalarValue]>>`, but that is an issue for another day_.

PR #4946, opened 2025-10-14
Feature: Make `ListView` the canonical encoding for `List`

connortsui20:

Revived from: https://github.com/vortex-data/vortex/pull/4853 and https://github.com/vortex-data/vortex/pull/4830

This PR makes `ListViewArray` the canonical encoding of `DType::List`, which used to be `ListArray`.

We will probably have a large regression in term of performance for now, but that can be improved in the future.

### DIFF
--- a/vortex-array/src/arrays/chunked/vtable/canonical.rs
+++ b/vortex-array/src/arrays/chunked/vtable/canonical.rs
@@ -82,28 +82,38 @@ fn swizzle_struct_chunks(
     unsafe { StructArray::new_unchecked(field_arrays, struct_dtype.clone(), len, validity) }
 }
 
+/// Packs list arrays together into a chunked `ListViewArray`.
+///
+/// There is guaranteed to be at least 2 chunks.
 fn pack_lists(chunks: &[ArrayRef], validity: Validity, elem_dtype: &DType) -> ListViewArray {
     let len: usize = chunks.iter().map(|c| c.len()).sum();
 
-    // Estimate total elements capacity by summing the element counts of all chunks.
-    let elements_capacity: usize = chunks
-        .iter()
-        .map(|c| c.to_listview().elements().len())
-        .sum();
+    assert_eq!(
+        chunks[0]
+            .dtype()
+            .as_list_element_opt()
+            .vortex_expect("DType was somehow not a list")
+            .as_ref(),
+        elem_dtype
+    );
 
-    // Note that since views can be out of order, it is easier to just rebuild the entire array than
-    // to recalculate views based on new offsets for each chunk.
-    let mut elements_builder = builder_with_capacity(elem_dtype, elements_capacity);
+    // Since each list array in `chunks` has offsets local to each array, we can reuse the existing
+    // array's child `elements` as the chunks and recompute offsets.
+    let mut list_elements_chunks = Vec::with_capacity(chunks.len());
+    let mut num_elements = 0;
 
     // TODO(connor)[ListView]: We could potentially choose a smaller type here, but that would make
     // this much more complicated.
-
-    // We (somewhat arbitrarily) choose `u64` for our offsets and sizes here.
+    // We (somewhat arbitrarily) choose `u64` for our offsets and sizes here. These can always be
+    // narrowed later by the compressor.
     let mut offsets = BufferMut::<u64>::with_capacity(len);
     let mut sizes = BufferMut::<u64>::with_capacity(len);
 
     for chunk in chunks {
         let chunk = chunk.to_listview();
+
+        // Add the `elements` of the current array as a new chunk.
+        list_elements_chunks.push(chunk.elements().clone());
 
         // Cast offsets and sizes to u64.
         let offsets_arr = cast(
@@ -123,18 +133,18 @@ fn pack_lists(chunks: &[ArrayRef], validity: Validity, elem_dtype: &DType) -> Li
         let offsets_slice = offsets_arr.as_slice::<u64>();
         let sizes_slice = sizes_arr.as_slice::<u64>();
 
-        // Track the current position in our combined elements array.
-        let current_elements_offset = elements_builder.len() as u64;
-
-        // Append all elements from this chunk.
-        elements_builder.extend_from_array(chunk.elements());
-
         // Append offsets and sizes, adjusting offsets to point into the combined array.
-        offsets.extend(offsets_slice.iter().map(|o| o + current_elements_offset));
+        offsets.extend(offsets_slice.iter().map(|o| o + num_elements));
         sizes.extend(sizes_slice);
+
+        num_elements += chunk.elements().len() as u64;
     }
 
-    let elements = elements_builder.finish();
+    // SAFETY: elements are sliced from valid `ListViewArray`s (from `to_listview()`).
+    let chunked_elements =
+        unsafe { ChunkedArray::new_unchecked(list_elements_chunks, elem_dtype.clone()) }
+            .into_array();
+
     let offsets = PrimitiveArray::new(offsets.freeze(), Validity::NonNullable).into_array();
     let sizes = PrimitiveArray::new(sizes.freeze(), Validity::NonNullable).into_array();
 
@@ -143,7 +153,7 @@ fn pack_lists(chunks: &[ArrayRef], validity: Validity, elem_dtype: &DType) -> Li
     // - Each `offset[i] + size[i]` list view is within bounds of elements array because it came
     //   from valid chunks
     // - Validity came from the outer chunked array so it must have the same length
-    unsafe { ListViewArray::new_unchecked(elements, offsets, sizes, validity) }
+    unsafe { ListViewArray::new_unchecked(chunked_elements, offsets, sizes, validity) }
 }
 
 #[cfg(test)]
--- a/vortex-array/src/arrays/chunked/vtable/canonical.rs
+++ b/vortex-array/src/arrays/chunked/vtable/canonical.rs
@@ -23,14 +23,14 @@ impl CanonicalVTable<ChunkedVTable> for ChunkedVTable {
 
         match array.dtype() {
             DType::Struct(struct_dtype, _) => {
-                let struct_array = swizzle_struct_chunks(
+                let struct_array = pack_struct_chunks(
                     array.chunks(),
                     Validity::copy_from_array(array.as_ref()),
                     struct_dtype,
                 );
                 Canonical::Struct(struct_array)
             }
-            DType::List(elem_dtype, _) => Canonical::List(pack_lists(
+            DType::List(elem_dtype, _) => Canonical::List(swizzle_list_chunks(
                 array.chunks(),
                 Validity::copy_from_array(array.as_ref()),
                 elem_dtype,
@@ -50,9 +50,11 @@ impl CanonicalVTable<ChunkedVTable> for ChunkedVTable {
     }
 }
 
-/// Swizzle the pointers within a ChunkedArray of StructArrays to instead be a single
-/// StructArray, where the Array for each Field is a ChunkedArray.
-fn swizzle_struct_chunks(
+/// Packs many [`StructArray`]s to instead be a single [`StructArray`], where the [`Array`] for each
+/// field is a [`ChunkedArray`].
+///
+/// The caller guarantees there are at least 2 chunks.
+fn pack_struct_chunks(
     chunks: &[ArrayRef],
     validity: Validity,
     struct_dtype: &StructFields,
@@ -71,6 +73,7 @@ fn swizzle_struct_chunks(
                     .to_array()
             })
             .collect::<Vec<_>>();
+
         // SAFETY: field_chunks are extracted from valid StructArrays with matching dtypes.
         // Each chunk's field array is guaranteed to be valid for field_dtype.
         let field_array = unsafe { ChunkedArray::new_unchecked(field_chunks, field_dtype.clone()) };
@@ -82,10 +85,16 @@ fn swizzle_struct_chunks(
     unsafe { StructArray::new_unchecked(field_arrays, struct_dtype.clone(), len, validity) }
 }
 
-/// Packs list arrays together into a chunked `ListViewArray`.
+/// Packs [`ListViewArray`]s together into a chunked `ListViewArray`.
 ///
-/// There is guaranteed to be at least 2 chunks.
-fn pack_lists(chunks: &[ArrayRef], validity: Validity, elem_dtype: &DType) -> ListViewArray {
+/// We use the existing arrays (chunks) to form a chunked array of `elements` (the child array).
+///
+/// The caller guarantees there are at least 2 chunks.
+fn swizzle_list_chunks(
+    chunks: &[ArrayRef],
+    validity: Validity,
+    elem_dtype: &DType,
+) -> ListViewArray {
     let len: usize = chunks.iter().map(|c| c.len()).sum();
 
     assert_eq!(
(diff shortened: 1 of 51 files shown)


vortex-array/src/arrays/listview/array.rs
```diff
@@ -345,74 +352,3 @@ where
 
     Ok(())
 }
-
-/// Create a [`ListViewArray`] from a [`ListArray`](crate::arrays::ListArray) by computing `sizes`
-/// from `offsets`.
-pub fn list_view_from_list(list: ListArray) -> ListViewArray {
```
connortsui20, line 352: Moved to `conversion.rs`

vortex-array/src/compute/list_contains.rs
connortsui20, line 1: I want to clean this file up in a separate PR

vortex-array/src/arrays/chunked/vtable/canonical.rs
```diff
@@ -82,56 +82,68 @@ fn swizzle_struct_chunks(
...
+        .sum();
+
+    // Note that since views can be out of order, it is easier to just rebuild the entire array than
+    // to recalculate views based on new offsets for each chunk.
+    let mut elements_builder = builder_with_capacity(elem_dtype, elements_capacity);
+
+    // TODO(connor)[ListView]: We could potentially choose a smaller type here, but that would make
+    // this much more complicated.
+
+    // We (somewhat arbitrarily) choose `u64` for our offsets and sizes here.
```
a10y: this seems fine to me, we can always .narrow() at the end

vortex-array/src/arrays/list/compute/take.rs
```diff
@@ -1,179 +1,30 @@
...
+use crate::compute::{self, TakeKernel, TakeKernelAdapter};
+use crate::{Array, ArrayRef, IntoArray, register_kernel};
 
 /// Take implementation for [`ListArray`].
 ///
-/// Unlike `ListView`, `ListArray` must rebuild the elements array to maintain its invariant
-/// that lists are stored contiguously and in-order (`offset[i+1] >= offset[i]`). Taking
-/// non-contiguous indices would violate this requirement.
+/// This implementation converts the [`ListArray`] to a [`ListViewArray`] and then delegates to its
+/// `take` implementation. This approach avoids the need to rebuild the `elements` array.
```
a10y, line 14: for very short indices arrays, it'd probably be more efficient to build the offsets/lengths incrementally. but this is probably fine for now

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -72,10 +74,11 @@ impl TakeKernel for ListViewVTable {
             ListViewArray::new_unchecked(elements.clone(), new_offsets, new_sizes, new_validity)
         };
 
-        // TODO(connor)[ListView]: IsZeroCopyToList optimization.
-        // TODO(connor)[ListView]: Rebuild if the threshold is too low.
+        // TODO(connor)[ListView]: Rebuild if the threshold is too low, not unconditionally.
 
-        Ok(new_array.into_array())
+        Ok(new_array
+            .rebuild(ListViewRebuildMode::MakeZeroCopyToList)
```
a10y, line 82: why are we doing this here? we should lazily compact the ListView not every time we do a take. there should be something like the `compact` method on VarBinViewArray. This is something we want todo at compression time but not during query processing probably?

vortex-array/src/canonical.rs
```diff
@@ -78,14 +78,14 @@ pub enum Canonical {
     Primitive(PrimitiveArray),
     Decimal(DecimalArray),
     VarBinView(VarBinViewArray),
-    // TODO(connor)[ListView]: Convert the canonical encoding of `List` to `ListViewArray`.
-    List(ListArray),
+    List(ListViewArray),
```
a10y, line 85: perhaps we should change the variant name here to ListView, similar to VarBinView

vortex-array/src/context.rs
```diff
@@ -25,21 +25,21 @@ impl ArrayRegistry {
...
-        // Register the canonical encodings
+        // Register the canonical encodings.
         this.register_many([
             EncodingRef::new_ref(NullEncoding.as_ref()) as EncodingRef,
             EncodingRef::new_ref(BoolEncoding.as_ref()),
             EncodingRef::new_ref(PrimitiveEncoding.as_ref()),
             EncodingRef::new_ref(DecimalEncoding.as_ref()),
             EncodingRef::new_ref(StructEncoding.as_ref()),
-            EncodingRef::new_ref(ListEncoding.as_ref()),
+            EncodingRef::new_ref(ListViewEncoding.as_ref()),
```
a10y, line 35: The comment is sadly wrong, this should contain all of the builtin encodings from the vortex-array crate. See below how we include VarBinEncoding. I think similarly, we should include ListEncoding here as well

vortex-dtype/src/arrow.rs
```diff
@@ -127,7 +127,10 @@ impl FromArrowType<(&DataType, Nullability)> for DType {
             | DataType::Timestamp(..) => DType::Extension(Arc::new(
                 make_temporal_ext_dtype(data_type).with_nullability(nullability),
             )),
-            DataType::List(e) | DataType::LargeList(e) => {
+            DataType::List(e)
+            | DataType::LargeList(e)
+            | DataType::ListView(e)
+            | DataType::LargeListView(e) => {
```
a10y, line 133: below in to_arrow_dtype needs to be updated to return DataType::ListView

vortex-scalar/src/binary.rs
```diff
@@ -15,7 +15,7 @@ use crate::{InnerScalarValue, Scalar, ScalarValue};
 ///
 /// This type provides a view into a binary scalar value, which can be either
 /// a valid byte buffer or null.
-#[derive(Debug, Hash)]
+#[derive(Debug, Clone, Hash)]
```
a10y, line 18: why did you make all of these clone?

vortex-array/src/arrays/listview/compute/filter.rs
```diff
@@ -57,10 +59,11 @@ impl FilterKernel for ListViewVTable {
             ListViewArray::new_unchecked(elements.clone(), new_offsets, new_sizes, new_validity)
         };
 
-        // TODO(connor)[ListView]: IsZeroCopyToList optimization.
-        // TODO(connor)[ListView]: Rebuild if the threshold is too low.
+        // TODO(connor)[ListView]: Rebuild if the threshold is too low, not unconditionally.
 
-        Ok(new_array.into_array())
+        Ok(new_array
+            .rebuild(ListViewRebuildMode::MakeZeroCopyToList)
```
a10y, line 67: same deal here: i don't think we should try and optimize/gc the elements when we run filter(ListView)

vortex-array/src/arrays/chunked/vtable/canonical.rs
```diff
@@ -82,56 +82,68 @@ fn swizzle_struct_chunks(
...
-                .map(|off| off + adjustment_from_previous - first_offset_value as u64),
-        );
+        let offsets_slice = offsets_arr.as_slice::<u64>();
+        let sizes_slice = sizes_arr.as_slice::<u64>();
+
+        // Track the current position in our combined elements array.
+        let current_elements_offset = elements_builder.len() as u64;
+
+        // Append all elements from this chunk.
+        elements_builder.extend_from_array(chunk.elements());
```
joseph-isaacs: I think we want to use a chunked array of the elements and update the offsets

vortex-array/src/arrays/chunked/vtable/canonical.rs
joseph-isaacs: You can find the max size and offset DType and use that? I think that might be worth it to hopefully avoid a  few casts.

vortex-array/src/arrays/constant/vtable/canonical.rs
```diff
@@ -204,58 +208,54 @@ fn canonical_byte_view(scalar_bytes: Option<&[u8]>, dtype: &DType, len: usize) -
...
+        scalar.dtype().nullability(),
+        elements.len() * len,
+        len,
+    );
+
+    // Simply append the same scalar to the `ListViewBuilder` `len` times.
+    for _ in 0..len {
+        listview_builder
+            .append_scalar(scalar)
+            .vortex_expect("somehow unable to append list scalar to listview builder");
```
joseph-isaacs: Here we can return a LIstView with a constant size, element and offset arrays.

vortex-array/src/arrays/listview/compute/filter.rs
joseph-isaacs: I think we should do this, otherwise when we later export we end up with a load of unused data we need to operate on (when we canoncalize)

vortex-scalar/src/binary.rs
connortsui20: https://github.com/vortex-data/vortex/issues/4844

I think I needed clone on list scalars for something in this change, and so I just added it to everything.

vortex-array/src/canonical.rs
joseph-isaacs: I think these would match where possible the DType?

vortex-btrblocks/src/lib.rs
```diff
@@ -399,18 +399,29 @@ impl BtrBlocksCompressor {
...
             Canonical::List(list_array) => {
-                // Compress the inner
+                // Compress the inner elements.
                 let compressed_elems = self.compress(list_array.elements())?;
+
+                // Compress the offsets.
                 let compressed_offsets = IntCompressor::compress_no_dict(
                     &list_array.offsets().to_primitive().narrow()?,
                     false,
                     MAX_CASCADE,
```
joseph-isaacs, line 412: Do we want `MAX_CASCADE` to reset for structural types?

vortex-duckdb/src/exporter/list.rs
```diff
@@ -15,74 +16,125 @@ use crate::cpp;
...
+        None => {
+            // We have no cached the vector yet, so create a new DuckDB vector for the elements.
+            let mut duckdb_elements =
+                Vector::with_capacity(elements.dtype().try_into()?, elements.len());
+            let elements_exporter = new_array_exporter_with_flatten(array.elements(), cache, true)?;
+
+            // TODO(connor)[4809]: THIS IS A HACK, we should be checking this everywhere.
+            if !elements.is_empty() {
+                elements_exporter.export(0, elements.len(), &mut duckdb_elements)?;
+            }
```
joseph-isaacs, line 60: can we just fix this, since I am not sure this exporter will handle empty size/offsets

vortex-layout/src/layouts/compact.rs
```diff
@@ -132,19 +133,19 @@ impl CompactCompressor {
                 .into_array()
             }
             Canonical::List(list_array) => {
-                // recurse
                 let compressed_elems = self.compress(list_array.elements())?;
                 let compressed_offsets = self.compress(list_array.offsets())?;
+                let compressed_sizes = self.compress(list_array.sizes())?;
```
joseph-isaacs: Does this narrow?

vortex-array/src/arrays/chunked/vtable/canonical.rs
connortsui20: Dan has a macro in `encodings/sparse/src/canonical.rs` called `match_smallest_offset_type` that I was thinking about extracting it out into `vortex-dtype` so we can use it everywhere. I left a TODO in there and I can look at it later.

vortex-array/src/arrays/list/compute/take.rs
connortsui20: I left a TODO in there

vortex-array/src/canonical.rs
connortsui20: We talked about it offline with Nick, these should have the same name as `DType`, with the exception of binary data. Arguably, it shouldn't even be called `VarBinView` and instead be called something like `Canonical::Binary` or `Canonical::String`? Regardless, I added some more documentation at the top to make it clearer.

vortex-array/src/context.rs
connortsui20: I moved some things around so we do canonical first then everything else below. It seems like we are missing `DatetimeEncoding` too?

vortex-dtype/src/arrow.rs
connortsui20: nice catch!

vortex-array/src/arrays/listview/compute/filter.rs
connortsui20: The reason I wrote it like this is because if we don't do this, we get an even larger regression whenever we try and export data to something (like DuckDB) because, until we have operators, we cannot decide a "good" point where we want to garbage collect / rebuild. Obviously a good point is just after we have run / pipelined a bunch of filters and takes in succession, but we can't really do that with the current architecture.

For now I think we should just keep the TODO in there and we can revisit this.

vortex-layout/src/layouts/compact.rs
connortsui20: I'm actually now questioning if we should be calling `narrow()` in `vortex-btrblocks`. Isn't it somewhat unexpected that we write one type and then get back a different type? I guess it's probably fine because the types of the offsets and sizes are not encoded in our `DType`. But yes both should be the same.

vortex-array/src/arrays/constant/vtable/canonical.rs
connortsui20: https://github.com/vortex-data/vortex/pull/4946/commits/1b7a179f68a76177114c796a94859fdc80e7f225

vortex-layout/src/layouts/compact.rs
connortsui20: Added narrowing in https://github.com/vortex-data/vortex/pull/4946/commits/f34587bc4e1f686cd2210763162032b4ab20113a

vortex-array/src/arrays/chunked/vtable/canonical.rs
connortsui20: Added this in https://github.com/vortex-data/vortex/pull/4946/commits/c18c40ec29e10bf4122d382c4da94cea649dcb14

vortex-array/src/context.rs
connortsui20: Oh datetime is an extension type, nevermind

vortex-btrblocks/src/lib.rs
connortsui20: Not sure what the alternative should be, `MAX_CASCADE - 1`?

vortex-duckdb/src/exporter/list.rs
connortsui20: I will leave it to this: https://github.com/vortex-data/vortex/issues/4809

If it was just list I would fix this, but because the dict exporter is broken in the exact same way (I forgot that I just `ignore` the tests for both list and dict) I'd rather fix them together.

vortex-array/src/arrays/listview/compute/take.rs
connortsui20: See the response in the other conversation, but I also added a TODO comment that was more descriptive

connortsui20:
So... we hit this in the parquet implementation 😭:

https://github.com/apache/arrow-rs/blob/751b0822a7f0b2647c1c662131131b35f268bfef/parquet/src/arrow/schema/mod.rs#L682-L684

connortsui20:
Now ready for re-review, I'm pretty sure I hit everything, and everything I didn't hit is either in a TODO or an issue

robert3005:
This parquet todo is fabulous

robert3005 (approved):
Hard to review such a big change but multiple people left comments throughout and I think at this point it's best to merge this since we need listview as canonical list and work on whatever we find in followups

vortex-array/src/canonical.rs
a10y: that's fine, i just like consistency

PR #4947, opened 2025-10-14
chore[fuzz]: fuzz vortex compact (file_io)

joseph-isaacs:
Do we want to add compact fuzz to the file_io fuzzer also? Might be converted by array_ops?


robert3005 (approved):

a10y:
Can we run these as two separate jobs?

PR #4952, opened 2025-10-15
Use arrow's bit iterators for BitBuffer

gatesn:


vortex-buffer/src/bit/buf.rs
```diff
@@ -266,12 +267,12 @@ impl BitBuffer {
...
     /// If the caller doesn't hold only reference to the underlying buffer, a copy is created.
     /// The second value of the tuple is a bit_offset of the first value in the first byte
-    pub fn into_mut(self) -> (BitBufferMut, usize) {
+    pub fn into_mut(self) -> BitBufferMut {
         let bit_offset = self.offset % 8;
         let len = self.len;
         // TODO(robert): if we are copying here we can strip offset bits
         let shrunk = self.into_inner().into_mut();
-        (BitBufferMut::from_buffer(shrunk, len), bit_offset)
+        BitBufferMut::from_buffer(shrunk, bit_offset, len)
```
robert3005, line 289: cheeky

vortex-buffer/Cargo.toml
```diff
@@ -17,14 +17,14 @@ version = { workspace = true }
 all-features = true
 
 [features]
-arrow = ["dep:arrow-buffer"]
+arrow = []
```
robert3005, line 20: remove this?

gatesn: It still feature-flags the buffer conversion code (separate from the mandatory use of bitbuffer iterators)

vortex-buffer/src/bit/buf_mut.rs
robert3005: I think you need to account for the offset here. I.e. set_len should add offset and self_slice should be self.offset + self.len

gatesn: Added a bunch of tests too

vortex-buffer/benches/vortex_bitbuffer.rs
```diff
@@ -33,7 +33,7 @@ fn from_iter<B: FromIterator<bool>>(n: usize) {
 }
 
 #[divan::bench(args = INPUT_SIZE)]
-fn append_vortex_buffer(bencher: Bencher, length: usize) {
+fn append_vortex(bencher: Bencher, length: usize) {
```
gatesn: I probably should not have renamed these... ruins the codspeed results! But we can see from the old comments

vortex-buffer/src/bit/macros.rs
```diff
@@ -0,0 +1,30 @@
...
+    () => (
+        $crate::BitBuffer::empty()
+    );
+    ($elem:expr; $n:expr) => (
+        $crate::BitBuffer::full($elem, $n)
+    );
+    ($($x:expr),+ $(,)?) => (
+        $crate::BitBuffer::from_iter([$($x),+])
+    );
+}
```
joseph-isaacs, line 16: bit strings would be nice here `bitbuffer![101]`

joseph-isaacs (approved):

PR #4954, opened 2025-10-15
Vortex Vectors

gatesn:
Vectors are the first part of the move to making Vortex arrays logical plans. They will eventually largely subsume the physical functionality of the canonical arrays (leaving behind logical functionality), as well as array builders.

They follow a similar pattern to Tokio Bytes and vortex-buffer with a Vector and VectorMut that can be converted between each other with zero-copy.


connortsui20:
Since I need nick to review this I'll open a new PR.

PR #4955, opened 2025-10-15
perf[fastlanes]: fused for bitpacking decode

joseph-isaacs:


joseph-isaacs:
Its annoying we need to duplicate the decompression logic

robert3005:
Looks like this is mostly flat? I guess FoR gets amortized?

joseph-isaacs:
I think its was only not applied in many places due to unimplemented offset + patches, which is now removed

encodings/fastlanes/src/for/compress.rs
```diff
@@ -42,9 +49,38 @@ fn compress_primitive<T: NativePType + WrappingSub + PrimInt>(
...
+            FoR::unchecked_unfor_pack(bit_width, chunk, self.reference, dst);
+        }
+    }
+}
+
 pub fn decompress(array: &FoRArray) -> PrimitiveArray {
     let ptype = array.ptype();
 
+    // try to do fused unpack
+    if array.dtype().is_unsigned_int()
```
robert3005, line 76: I thought this works fine irrespective of the dtype of the array?

Cargo.toml
```diff
@@ -111,7 +111,7 @@ divan = { package = "codspeed-divan-compat", version = "4.0.4" }
 dyn-hash = "0.2.0"
 enum-iterator = "2.0.0"
 erased-serde = "0.4"
-fastlanes = "0.4.0"
+fastlanes = { rev = "3820eeb33433458fa035bea5a834dee0039612f5", git = "https://github.com/spiraldb/fastlanes.git" }
```
robert3005, line 114: I think you need a release

robert3005 (approved):
I think this will become nicer once patches can also be fused into decoding

PR #4957, opened 2025-10-15
fix[btr-blocks]: `cfg(unstable_encodings)` imports

joseph-isaacs:


gatesn (approved):
Would be good to know why the linter isn't failing currently

robert3005:
This is only an issue if you build this project in isolation, so when we do clippy on each project independetly we don't disable warnings, i.e. we run `cargo hack clippy --no-default-features` instead of `cargo hack clippy --no-default-features -- -D warnings`

PR #4961, opened 2025-10-15
support DynamicFilterPhysicalExpr

adriangb:
cc @AdamGS

### DIFF
--- a/vortex-datafusion/src/persistent/mod.rs
+++ b/vortex-datafusion/src/persistent/mod.rs
@@ -41,7 +41,8 @@ mod tests {
         ListingOptions, ListingTable, ListingTableConfig, ListingTableUrl,
     };
     use datafusion::execution::SessionStateBuilder;
-    use datafusion::prelude::SessionContext;
+    use datafusion::prelude::{SessionConfig, SessionContext};
+    use datafusion_common::record_batch;
     use datafusion_datasource::file_format::format_as_file_type;
     use datafusion_expr::LogicalPlanBuilder;
     use datafusion_physical_plan::display::DisplayableExecutionPlan;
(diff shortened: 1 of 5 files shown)


vortex-datafusion/src/persistent/opener.rs
```diff
@@ -206,28 +201,38 @@ impl FileOpener for VortexOpener {
...
+            let predicate = predicate
                 .and_then(|f| {
                     let exprs = split_conjunction(&f)
                         .into_iter()
+                        .filter_map(|e| {
+                            if is_dynamic_physical_expr(&e) {
+                                e.snapshot().ok().flatten()
+                            } else {
+                                Some(Arc::clone(e))
+                            }
```
adriangb, line 213: Not sure if it's possible or worth it but you could do this more lazily, or wrap the DataFusion dynamic filter in a Vortex dynamic filter and call `DynamicFilterPhysicalExpr::current()` or `PhysicalExpr::snapshot()` on each RecordBatch (that would be more overhead but also will kick in sooner on scans that touch very large files).

vortex-datafusion/src/persistent/mod.rs
```diff
@@ -190,6 +191,103 @@ mod tests {
...
+
+        let data_source_line = plan
+            .lines()
+            .filter(|line| line.contains("DataSourceExec"))
+            .skip(1)
+            .next()
+            .ok_or_else(|| {
+                vortex_err!("Plan should have 2 DataSourceExec, the second is the probe side")
+            })?;
+        assert!(data_source_line.contains("DynamicFilterPhysicalExpr [ a@0 >= 1 AND a@0 <= 3 ]"));
```
adriangb, line 286: I'm not sure what other assertions we could make. I don't think Vortex gives any metrics on rows pruned, etc. And the Vortex expression isn't saved anywhere / is not in the plan.

adriangb: I put in a print statement and can see that an appropriate Vortex predicate is being created, but can't verify that it's doing what's expected

joseph-isaacs:
We cannot currently run pre commit benchmark on fork so I created this (https://github.com/vortex-data/vortex/pull/4986)

AdamGS:
@adriangb haven't forgotten this PR, just really busy week

AdamGS:
I can belive some of the perf impact, but I suspect AWS tonight is just extremely noisy, gotta love the 40% on clickbecnh q0.

adriangb:
> I can belive some of the perf impact, but I suspect AWS tonight is just extremely noisy, gotta love the 40% on clickbecnh q0.

Is there a benchmark or something I'm not seeing? If it's Polar Signal I don't seem to have access.

AdamGS:
They ran on [this](https://github.com/vortex-data/vortex/pull/4986) PR, seems like there's an issue to trigger benchmarks on this branch (@joseph-isaacs got something I can look at? Would love to be able to do that)

vortex-datafusion/src/persistent/mod.rs
AdamGS: I'm not sure I have a better idea here right now, I need to figure out how to improve the overall testability

vortex-datafusion/src/persistent/opener.rs
AdamGS: I roughly tried to do that in some old PR, but I don't remember why I stopped pushing it. This seems like a good starting point, and we can experiment with it in the future to see which one is better.

adriangb:
@AdamGS what's your read on the state of this PR / what can I do to help? It's not clear to me if there are issues with benchmarks, if @joseph-isaacs is planning on taking it over in another PR, if I should rewrite the commits with the licensing info, etc.

AdamGS:
You beat me to comment by a minute!
I've re-triggered the benchmarks over at https://github.com/vortex-data/vortex/pull/4986, if you can it'll be great if you can fix the lint failure (just run with nightly) and the DCO thing and we can merge this IMO.

AdamGS:
Something is going on with the label check, I'll figure that one out.

AdamGS (approved):
This is awesome 🥳

AdamGS:
Last benchmarks run - https://github.com/vortex-data/vortex/actions/runs/18684856431

joseph-isaacs (changes requested):
There are some big regressions on tpch which we need to address before merging

AdamGS:
Completely missed that, I'll take a deeper look.

adriangb:
I think you may have found some real performance issues. We are discussing similar findings on the DataFusion side:
https://github.com/apache/datafusion/pull/19639
https://github.com/apache/datafusion/issues/3463

TLDR is we think some of the expressions may be expensive to evaluate. Unclear if that means they're not worth having at all, or if the plan shape / parallelism needs to be tweaked, but it does seem real.

PR #4962, opened 2025-10-15
fix: Builder ensure_capacity is now reserve_exact to acquire additional and not total capacity

### LINKED ISSUE #4958: Use `reserve` and `reserve_exact` in our builders
At that point we might as well add both [`reserve`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve) and [`reserve_exact`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve_exact)?

_Originally posted by @connortsui20 in https://github.com/vortex-data/vortex/issues/4956#issuecomment-3408179750_

Honestly `ensure_capacity` is a bit strange and we should definitely be following `std::vec::Vec`

robert3005:
fix #4958 

### DIFF
--- a/vortex-array/src/arrays/list/compute/take.rs
+++ b/vortex-array/src/arrays/list/compute/take.rs
@@ -82,7 +82,7 @@ fn _take<I: IntegerPType, O: IntegerPType>(
             vortex_panic!("Failed to convert range length to usize: {}", stop - start)
         });
 
-        elements_to_take.reserve(additional);
+        elements_to_take.reserve_exact(additional);
         for i in 0..additional {
             elements_to_take.append_value(start + O::from_usize(i).vortex_expect("i < additional"));
         }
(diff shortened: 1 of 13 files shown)


0ax1 (commented):
Looks good. Giving it a benchmark run just now to see whether this leads to regressions.

robert3005:
The only practical difference is that we might be doing two capacity checks in some cases instead. We should rename this function to reserve exact

0ax1:
> The only practical difference is that we might be doing two capacity checks in some cases instead. We should rename this function to reserve exact

I was thinking about the case that calling `reserve` would allocate repeatedly whilst `ensure_capacity` conditionally.

```
fn ensure_capacity(&mut self, capacity: usize) {
    if capacity > self.inner.capacity() {
        self.inner.reserve(capacity - self.inner.capacity());
        self.nulls.ensure_capacity(capacity);
    }
}

fn reserve(&mut self, additional: usize) {
    self.inner.reserve(additional);
    self.nulls.reserve(additional);
}
```

robert3005:
Our buffer checks the capacity just like arrow Booleanbuffer

0ax1 (approved):

0ax1:
> clickbench_q23/duckdb:vortex-compact (2.591x)

Is annoyingly just benchmark noise.

PR #4965, opened 2025-10-16
feat: collect patch offsets per chunk during alp encode

0ax1:


encodings/alp/src/alp/mod.rs
```diff
@@ -121,18 +121,27 @@ pub trait ALPFloat: private::Sealed + Float + Display + NativePType {
...
+        let mut patch_indices = BufferMut::<u64>::with_capacity(values.len() / 256);
+        let mut patch_values = BufferMut::<Self>::with_capacity(values.len() / 256);
+
+        // There's exactly one offset per 1024 chunk.
+        let mut chunk_offsets = BufferMut::<u64>::with_capacity(values.len() / 1024);
         let mut fill_value: Option<Self::ALPInt> = None;
 
-        // this is intentionally branchless
-        // we batch this into 32KB of values at a time to make it more L1 cache friendly
-        let encode_chunk_size: usize = (32 << 10) / size_of::<Self::ALPInt>();
```
0ax1, line 134: Reducing the chunk iter size to `1024` made no diff in the alp micro-benchmarks locally.  Those at max though also only encode `10_000` elements.

robert3005 (approved):

PR #4967, opened 2025-10-16
chore: remove merge's explicit nullability

danking:
Merge's nullability should be defined in terms of its children. If you want to change that, use `cast`.

### DIFF
--- a/vortex-expr/src/exprs/merge.rs
+++ b/vortex-expr/src/exprs/merge.rs
@@ -25,7 +25,6 @@ vtable!(Merge);
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct MergeExpr {
     values: Vec<ExprRef>,
-    nullability: Nullability,
 }
 
 pub struct MergeExprEncoding;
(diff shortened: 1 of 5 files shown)


vortex-expr/src/exprs/merge.rs
```diff
@@ -127,11 +118,17 @@ impl VTable for MergeVTable {
         let mut field_names = Vec::new();
         let mut arrays = Vec::new();
 
+        let mut nullability = Nullability::NonNullable;
```
joseph-isaacs, line 121: why this?

joseph-isaacs (approved):

robert3005 (approved):

blaginin:
@danking it's a bit tricky to debug this change. Right now, it fails on `merge expects non-nullable input` when `infer_data_type` is called, but i don't know where the original expression is created 🥲

Maybe we can fail when expression is first created? in `pub fn merge` or `MergeExpr::new` (i guess it'll be `try_new`?)

@claude can you sketch this

upd ... oh, i guess we can't because we don't have scope 😔

PR #4972, opened 2025-10-16
chore: change 4 lines; eliminate countless Arc clones

danking:
If a consumer needs an owned version they can clone. There are like 50 spots in the code where we only need the slice, not an owned Arc.


robert3005 (approved):

onursatici (approved):

joseph-isaacs (approved):
😍

PR #4977, opened 2025-10-17
fix[vortex-expr]: implement stat_falsification for BetweenExpr

asubiotto:
This would otherwise cause chunks to not be pruned when they could be with range filters.

### DIFF
--- a/vortex-expr/src/exprs/between.rs
+++ b/vortex-expr/src/exprs/between.rs
@@ -22,7 +22,8 @@ use vortex_proto::expr as pb;
 
 use crate::display::{DisplayAs, DisplayFormat};
 use crate::{
-    AnalysisExpr, BinaryExpr, ExprEncodingRef, ExprId, ExprRef, IntoExpr, Scope, VTable, vtable,
+    AnalysisExpr, BinaryExpr, ExprEncodingRef, ExprId, ExprRef, IntoExpr, Scope, StatsCatalog,
+    VTable, vtable,
 };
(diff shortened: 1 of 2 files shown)


asubiotto:
I wonder if maybe we shouldn't have `stat_falsification` implicitly return `None`?

robert3005:
It would be nice to not have a default, we were a bit lazy and only wanted to consider things we were certain of. Thanks for finding this one!

vortex-expr/src/exprs/between.rs
```diff
@@ -219,7 +219,11 @@ impl DisplayAs for BetweenExpr {
     }
 }
 
-impl AnalysisExpr for BetweenExpr {}
+impl AnalysisExpr for BetweenExpr {
+    fn stat_falsification(&self, catalog: &mut dyn crate::StatsCatalog) -> Option<ExprRef> {
```
robert3005: you want to import this

robert3005 (approved):
Great find, thanks

vortex-expr/src/exprs/between.rs
asubiotto: oops, done!

asubiotto:
> It would be nice to not have a default, we were a bit lazy and only wanted to consider things we were certain of. Thanks for finding this one!

Took a quick look at this, but I think the majority of remaining expressions would just return None, so maybe it is saner to have a `None` default.

PR #4982, opened 2025-10-17
fix: Continue to prefer List over ListView by default in ToArrow

a10y:
Turns out, arrow-rs/DataFusion lacks complete support for ListView. Until that is resolved, for now the default to_arrow_preferred should prefer returning List over ListView.

See https://github.com/vortex-data/vortex/actions/runs/18591096514/job/53006051293?pr=4977 for example failure

### DIFF
--- a/vortex-dtype/src/arrow.rs
+++ b/vortex-dtype/src/arrow.rs
@@ -326,11 +326,11 @@ mod test {
         assert_ne!(arrow_list_non_nullable, arrow_list_nullable);
         assert_eq!(
             arrow_list_nullable,
-            DataType::ListView(Arc::new(Field::new_list_field(DataType::Int64, true))),
+            DataType::List(Arc::new(Field::new_list_field(DataType::Int64, true))),
         );
         assert_eq!(
             arrow_list_non_nullable,
-            DataType::ListView(Arc::new(Field::new_list_field(DataType::Int64, false))),
+            DataType::List(Arc::new(Field::new_list_field(DataType::Int64, false))),
         );
     }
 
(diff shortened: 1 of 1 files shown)


connortsui20:
I'm actually somewhat surprised that only 1 test failed since just changing the arrow dtype should also require that we need to make the conversions properly?

Edit: Oh right we match on the vortex and arrow type and use the correct conversion function

a10y:
This is just the preferred conversion. Anything that has an explicit conversion should still go to ListView

vortex-dtype/src/arrow.rs
```diff
@@ -215,7 +215,7 @@ impl DType {
             // There are four kinds of lists: List (32-bit offsets), Large List (64-bit), List View
             // (32-bit), Large List View (64-bit). We cannot both guarantee zero-copy and commit to an
             // Arrow dtype because we do not how large our offsets are.
-            DType::List(elem_dtype, _) => DataType::ListView(FieldRef::new(Field::new_list_field(
+            DType::List(elem_dtype, _) => DataType::List(FieldRef::new(Field::new_list_field(
```
connortsui20, line 218: I think this deserves a comment noting that we are using `List` over `ListView` because datafusion doesn't really support it yet?

connortsui20 (approved):

robert3005:
For posterity, it falls through here https://github.com/apache/arrow-rs/blob/main/arrow-select/src/filter.rs#L309 (where there could be an optimised impl for listviews) and then panics in https://github.com/apache/arrow-rs/blob/main/arrow-data/src/transform/mod.rs#L269

a10y:
Yep, I'll push a fix to arrow soon

a10y:
Fix to arrow: https://github.com/apache/arrow-rs/pull/8645

PR #4983, opened 2025-10-17
add blanket impl for futures::io::AsyncWrite on VortexWrite

a10y:


a10y:
I thought there might be an easier way to do this without a newtype wrapper but I guess there isn't, due to the complexity of having many different blanket impls where the compiler can't reason about the potential for the blanket impl to cover some external type from another crate

PR #4989, opened 2025-10-17
Arc<Box<_>> a struct array's children

AdamGS:
The thinking here is that `Vec<T>` -> `Arc<[T]>` requires copying the data, but `Vec<T>` -> `Box<[T]>` does not.

This PR also introduces a new struct named `StructArrays` (really need a different name) that wraps the children to make the API a bit more ergonomic and less dependent on the actual pointers it uses.

The benchmarks results are really confusing, but there's some positive trend (potentially just wishful thinking).

### DIFF
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -3,9 +3,10 @@
 
 use std::fmt::Debug;
 use std::iter::once;
+use std::ops::Index;
 use std::sync::Arc;
 
-use vortex_dtype::{DType, FieldName, FieldNames, StructFields};
+use vortex_dtype::{DType, FieldName, FieldNames, Nullability, StructFields};
 use vortex_error::{VortexExpect, VortexResult, vortex_bail, vortex_err};
 
 use crate::stats::ArrayStats;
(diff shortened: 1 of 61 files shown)


vortex-array/src/arrays/struct_/array.rs
```diff
@@ -192,7 +192,7 @@ impl StructArray {
     /// in [`StructArray::new_unchecked`].
     pub fn new(
         names: FieldNames,
-        fields: impl Into<Arc<[ArrayRef]>>,
+        fields: impl Into<Arc<Box<[ArrayRef]>>>,
```
gatesn: Either way, I would strongly prefer this to not be an `impl Into` and have the caller invoke `.into()`!

AdamGS: yeah if this shows any promise I'll create a type here, we should really stop having functions that take these complex `Into`s

AdamGS:
I really want to believe these benchmark results, but this is the second PR I saw in the past ~hour that shows a 40% improvement in clickbench q0

AdamGS:
seems like the varbinview change regressed performance? I'll try to better separate it out to make it easier to figure out.

gatesn (changes requested):
Discussed offline, we will try to find a lint that checks for unnecessary `vec![..].into()` when `[...].into()` would be sufficient.

PR #4991, opened 2025-10-18
fix[vortex-io]: register read_at intent in InstrumentedReadAt

asubiotto:
This VortexReadAt implementation was unintentionally not registering its read_at intent until polled, which breaks scan pipeline coalescing.

Just for informational purposes, this reduced our object store requests by 80% for an example file scan.


gatesn (commented):
This is such a good catch 👏🤦‍♂️

gatesn (approved):

PR #4992, opened 2025-10-19
Stop enabling `pyo3/extension-modules` by default

AdamGS:
See [this](https://pyo3.rs/main/building-and-distribution#the-extension-module-feature) section the pyo3's docs. We already enable it when building the whl with maturin.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -20,5 +20,4 @@ timeout = 3600
 
 # Environment variables for PyO3. Ensures reproducible builds and avoids spurious recompilations.
 [env]
-PYO3_PYTHON = { value = ".venv/bin/python", relative = true, force = true }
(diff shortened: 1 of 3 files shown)


.cargo/config.toml
```diff
@@ -20,5 +20,4 @@ timeout = 3600
 
 # Environment variables for PyO3. Ensures reproducible builds and avoids spurious recompilations.
 [env]
-PYO3_PYTHON = { value = ".venv/bin/python", relative = true, force = true }
```
AdamGS, line 23: We saw in Spiral that this behave nicely now, setting it globally makes a bunch of build processes have unintended dependency on this path and general setup, which is not always desirable.

robert3005 (approved):

PR #4996, opened 2025-10-20
Clear up difference between struct fields and columns

AdamGS:
We currently call a StructArray's children "columns" in some places and "fields" in others, but "fields" is used inconsistently, sometimes its just the types and sometimes its the child arrays. This PR tries to clear up two concepts, making the API hopefully more intuitive and consistent. After this PR the split should be:
1. Fields - essentially a struct dtype (sans its top level nullability). An array of `Field`, where each field has a name, and a dtype.
2. Columns - the array's (or scalar's) values, stored in arrays matching the fields.

### DIFF
--- a/bench-vortex/src/bin/random_access.rs
+++ b/bench-vortex/src/bin/random_access.rs
@@ -207,10 +207,10 @@ fn validate_vortex_array(array: ArrayRef) {
     let struct_ = array.to_struct();
     assert_eq!(struct_.len(), 6, "expected 6 rows");
     let pu_location_id = struct_
-        .field_by_name("PULocationID")
+        .column_by_name("PULocationID")
         .vortex_expect("could not get PULocationID");
     let do_location_id = struct_
-        .field_by_name("DOLocationID")
+        .column_by_name("DOLocationID")
         .vortex_expect("could not get DOLocationID");
     for (idx, loc) in [90i32, 249, 230, 79, 239, 236].iter().enumerate() {
         assert_eq!(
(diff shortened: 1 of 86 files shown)


joseph-isaacs:
I always thought that struct array have fields list of (name, type and array). I am not sure we should change the name to column. To be precise maybe field_array or field_value?

AdamGS:
`field_value` also seems confusing to me, we generally only use the word `value` to indicate a `ScalarValue` or something similar. A `column` also matches a bunch of the docs we have and the mental model for a struct array.

joseph-isaacs:
Happy to go with a majority, but to me struct contain fields (not columns).

AdamGS:
I don't really care about whether its fields or columns or anything else, but I do think that the word "fields" (or any other term) should mean one thing and not multiple things in multiple contexts.

joseph-isaacs:
`Field` and `FieldType`? 

We don't really name List and FixedList, but we use `ExtDType` so maybe it follows?

AdamGS:
Why not make it clearer? The `Field` prefix just gets confusing if we use it everywhere.

gatesn:
My understanding was that we have lots of struct-like objects. StructArray, StructFields, and StructScalar.

We _always_ call the "child" thing a field. 

So StructArray::field returns Array. StructScalar::field returns Scalar. And so on.

I like this model. I think we'd are bad at making all the other functions consistent, I.e field_by_index, names, etc.

One option would be to have a StructLike trait that forces this with an associated type for the "child"/"field" type.

AdamGS:
That's a good enough rational for me. I'll close this PR in favor of one that will things more consistent in that direction.

PR #4999, opened 2025-10-20
Move fastlanes back to released version

AdamGS:
dependencies must be actually released versions, otherwise we can't release ourselves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2868,7 +2868,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.59.0",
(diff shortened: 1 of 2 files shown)


joseph-isaacs:
Sorry this was me I forgot to update

joseph-isaacs (approved):

robert3005:
I assumed renovate would move us back to released version but it would just bump git commits

AdamGS:
I think that theoretically you could configure it to do that, it picks an "update strategy" for each dependency and I assume we can tell it to always prefer a real version over a digest for cargo ones.

PR #4988, opened 2025-10-17
chore: use assert_arrays_eq in more places

danking:
I'm just trying to climb the deletions leader-board.

PR #5000, opened 2025-10-20
chore: expose `assertions` in `test-harness`

blaginin:


blaginin:
oh, dan already did that in https://github.com/vortex-data/vortex/pull/4988/ 🥳

PR #5001, opened 2025-10-20
fix[ci]: allow fork `PR`s to run benchmarks

joseph-isaacs:


.github/actions/setup-rust/action.yml
```diff
@@ -4,7 +4,8 @@ description: "Toolchain setup and Initial compilation"
 inputs:
   repo-token:
     description: "GitHub token for accessing the repository (typically secrets.GITHUB_TOKEN)"
-    required: true
+    required: false
+    default: ""
```
AdamGS: why not default to `${{ secrets.GITHUB_TOKEN }}`?

joseph-isaacs: secrets are not defined in actions are they?

joseph-isaacs: Also this is just to avoid rate limiting.

AdamGS: so shouldn't we set it to that?

robert3005: Yes, why this isn't the default?

.github/workflows/bench-pr.yml
```diff
@@ -108,7 +111,7 @@ jobs:
             | jq -r '.workflow_runs[].head_sha' \
           )
 
-          aws s3 cp s3://vortex-benchmark-results-database/data.json.gz - \
+          aws s3 cp s3://vortex-benchmark-results-database/data.json.gz - --no-sign-request \
```
robert3005, line 114: why do we have to add `--no-sign-request`?

joseph-isaacs: If we don't then this need OIDC, which forked prs don't

.github/actions/setup-rust/action.yml
joseph-isaacs: Its not defined? https://github.com/vortex-data/vortex/pull/5001#discussion_r2444840420

.github/workflows/bench-pr.yml
0ax1: The bucket then must be public. Do we want that though in general? Had a chat with @gatesn about this lately in a similar context (downloading pre-built duckdb libs).

.github/actions/setup-rust/action.yml
0ax1: @joseph-isaacs Did you look at https://docs.github.com/en/actions/tutorials/authenticate-with-github_token?

0ax1: > An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action.

.github/workflows/bench-pr.yml
joseph-isaacs: It is, its used in the bench vortex website

.github/actions/setup-rust/action.yml
0ax1: In my understanding the token is per workflow, per run.

0ax1:
@joseph-isaacs Got a fork PR where I can see an actual run? Your fork could attempt to merge into this branch.

robert3005 (approved):

PR #5006, opened 2025-10-20
chore: avoid copy when building chunk_offsets in chunked layout reader

robert3005:


robert3005:
Small thing I noticed while working on gpu readers

joseph-isaacs (approved):
Think it's worth computing this as an accumulate iter?

robert3005:
None of the itearator options look better imho

PR #5009, opened 2025-10-20
Add ArrayHash and ArrayEq traits

gatesn:
This allows us to check for structural equality of array trees.


gatesn:
I have decided to make associated Metadata type non-optional, so now we can implement hash/eq using an array visitor rather than have everyone implement these new traits.

vortex-array/src/hash.rs
```diff
@@ -0,0 +1,213 @@
...
+use crate::{Array, ArrayRef};
+
+/// A hash trait for arrays that loosens the semantics to permit pointer-based hashing for
+/// data objects such as buffers.
+///
+/// Note that since this trait can use pointer hashing, the hash is only valid for the lifetime of
+/// the object.
+pub trait ArrayHash {
+    fn array_hash<H: Hasher>(&self, state: &mut H);
+}
```
joseph-isaacs: Can we refine what this is trying to encode? In relation to the expr transforms we want to do based on this eq

connortsui20:
@claude review

vortex-array/src/hash.rs
gatesn: It's a cheaper hash function that full data hash. 

The false-negative is when two arrays that have the same exact tree, with the same bytes in the leaf buffers, but different buffer allocations, will not hash to the same thing.

joseph-isaacs: But are you trying to find a maybe equal (as in two arrays) might have all the same values. Or are you wanting to use this to decide if two computation trees are equiv upto the buffer. These are different and the difference matters. 

We need to add a comment.

gatesn: Oh I see, yeah sure, obviously the latter; the former would require compute

joseph-isaacs: I am not sure its true the former would necessarily require compute.

connortsui20 (approved):
Obviously tests would be nice, but I think we are in a "move quickly and break things" phase right now

PR #5010, opened 2025-10-20
Feature: Add `vortex-vector`

connortsui20:

Roadmap: https://github.com/vortex-data/vortex/issues/4492

Continuation of https://github.com/vortex-data/vortex/pull/4954

This is mostly boilerplate code with a bit more documentation than normal.

There are still a few bugs in the implementation (documented by TODOs), but I would like this PR to just focus on the structure of the crate rather than the logic.

Also add logic in other crates that was missing before.

---

The general structure of the crate is as such:

```
vortex-vector
├── Cargo.toml
└── src
    ├── <type>
    │   ├── mod.rs
    │   ├── vector.rs
    │   └── vector_mut.rs
    ├── lib.rs
    ├── macros.rs
    ├── ops.rs
    ├── private.rs
    ├── vector.rs
    └── vector_mut.rs
```

The exceptions to this structure will be `primitive`, which needs a generic type as well (so there will be extra files), a `varbin` module that holds both `binary` and `string` modules, and probably the 2 list types might share a lot of code (we'll see once I get there).

Again, a lot of the logic here is not super clean and some of it is wrong, but I'd like to get this merged first before I fix those.


vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -25,12 +26,26 @@ use crate::{BitBuffer, BufferMut, ByteBufferMut, buffer_mut};
...
     len: usize,
 }
 
+impl PartialEq for BitBufferMut {
+    fn eq(&self, other: &Self) -> bool {
+        if self.len != other.len {
+            return false;
+        }
+
+        self.chunks()
```
gatesn, line 44: I think this is wrong, you'll miss off the final trailer chunk.

See the Arrow BooleanBuffer implementation:

```rust
        let lhs = self.bit_chunks().iter_padded();
        let rhs = other.bit_chunks().iter_padded();
        lhs.zip(rhs).all(|(a, b)| a == b)
```

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -362,6 +384,63 @@ impl BitBufferMut {
...
+            return Self {
+                buffer: new_buffer,
+                offset: new_offset,
+                len: new_len,
+            };
+        }
+
+        // Otherwise, we need to copy bits into a new buffer
+        let mut new_buffer = BitBufferMut::with_capacity(new_len);
+        for i in 0..new_len {
```
gatesn, line 416: Should we slice and build the other one from iter?

vortex-vector/src/bool/vector_mut.rs
```diff
@@ -0,0 +1,120 @@
...
+#[derive(Debug, Clone)]
+pub struct BoolVectorMut {
+    pub(super) bits: BitBufferMut,
+    pub(super) validity: Option<MaskMut>,
+}
+
+impl BoolVectorMut {
+    /// Creates a new mutable boolean vector with the given `capacity` and `nullability`.
+    pub fn with_capacity(capacity: usize, nullability: Nullability) -> Self {
+        let validity = match nullability {
```
gatesn: You may want to add `Nullability::then(|| )` similar to `bool::then` and `bool::then_some`

vortex-vector/src/bool/vector_mut.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        if let Some(v) = self.validity.as_mut() {
+            v.reserve(additional);
+        }
+    }
+
+    fn extend_from_vector(&mut self, other: &BoolVector) {
+        self.bits.append_buffer(&other.bits);
+
+        // TODO(connor): We must `other`'s nullability in relation to `self`.
```
gatesn: Do we want this function to "expand" nullability? I think we probably want to assert that it's the same. Then force the user to do the cast?

vortex-vector/src/vector/ops.rs
```diff
@@ -0,0 +1,178 @@
...
+    ///
+    /// # Errors
+    ///
+    /// If `self` is not unique, this will fail and return `self` back to the caller.
+    fn try_into_mut(self) -> Result<Self::Mutable, Self>
+    where
+        Self: Sized;
+}
+
+impl VectorOps for Vector {
```
gatesn: Seems weird that the impl is here and not next to the Vector struct

vortex-vector/src/ops.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Does nothing if capacity is already sufficient.
+    ///
+    /// Please let us know if you need `reserve_exact` functionality!
+    fn reserve(&mut self, additional: usize);
+
+    /// Extends the vector by appending elements from another vector.
+    ///
+    /// TODO(connor): Document semantics of what happens if `self` is non-nullable and `other` is
+    /// nullable (should panic?).
+    fn extend_from_vector(&mut self, other: &Self::Immutable);
```
gatesn, line 89: I think we should add an `append_nulls(n)` function

vortex-vector/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Mutable decompressed (canonical) vectors for Vortex.
+//!
+//! TODO(connor) More docs.
```
gatesn: Remove this from public docs. Can keep the comment docs below

vortex-buffer/src/bit/buf_mut.rs
connortsui20: I just copied and pasted the implementation from `buf.rs` above:

```rust
impl PartialEq for BitBuffer {
    fn eq(&self, other: &Self) -> bool {
        if self.len != other.len {
            return false;
        }

        self.chunks()
            .iter()
            .zip(other.chunks())
            .all(|(a, b)| a == b)
    }
}
```

Is this also wrong?

Edit: It is indeed also wrong

vortex-vector/src/bool/vector_mut.rs
connortsui20: I think that this needs to panic if we try to extend a non-nullable vector with a nullable vector

vortex-vector/src/ops.rs
connortsui20: I'm not so sure about this... Unlike our current builders we want to ensure that the type is always in a correct state. If we add an `append_nulls` that adds nulls such that the length of the bitmask and the elements is no longer the same, we cannot call `freeze` without validating the data + returning a result / panicking.

vortex-buffer/src/bit/buf_mut.rs
connortsui20: I'm not sure that would be cleaner? To be honest I didn't actually look through any of this code

gatesn: It's not about cleaner, the performance of this won't be great

vortex-vector/src/ops.rs
gatesn: It just forces each vector to support it (they should append dummy values to elements or whatever is required). And each vector should panic if you try to append nulls to a non-nullable vector.

vortex-buffer/src/bit/buf.rs
```diff
@@ -442,4 +450,51 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_partial_eq_regression_under_64_bits() {
```
gatesn: It's weird to write tests that reference old buggy behavior. Can you just write tests against the new thing? e.g. create two arrays, one with 64 trues, one with 32 trues and 32 false. Then slice them to first 32 bits and compare equal.

vortex-vector/src/null/vector_mut.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    fn split_off(&mut self, at: usize) -> Self {
+        assert!(
+            at <= self.capacity(),
+            "split_off out of bounds: {:?} <= {:?}",
+            at,
+            self.capacity(),
+        );
+
+        // TODO(connor): This is wrong (https://docs.rs/bytes/latest/src/bytes/bytes_mut.rs.html#320-335)
```
gatesn: Either add a `todo!()` that fails at runtime, or fix the bug. I don't think we should merge code that will silently do the wrong thing

vortex-vector/src/primitive/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+mod vector_mut;
+pub use vector_mut::PrimitiveVectorMut;
+
+/// Helper macros for working with the different variants of [`PrimitiveVector`] and
+/// [`PrimitiveVectorMut`].
+///
+/// All macros are exported at the crate level with `#[macro_use]`.
+mod macros;
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
```
gatesn: Not a fan of these comment "sections"

vortex-vector/src/primitive/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+mod generic_mut;
+pub use generic_mut::GenericPVectorMut;
+
+mod vector;
+pub use vector::PrimitiveVector;
+
+mod vector_mut;
+pub use vector_mut::PrimitiveVectorMut;
+
+/// Helper macros for working with the different variants of [`PrimitiveVector`] and
```
gatesn: Module docs should be `//!` at the top of the macros.rs file

vortex-vector/src/primitive/vector.rs
```diff
@@ -0,0 +1,127 @@
...
+    {
+        match_each_pvector!(self, |v| {
+            v.try_into_mut()
+                .map(PrimitiveVectorMut::from)
+                .map_err(PrimitiveVector::from)
+        })
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
```
gatesn: Same here

vortex-vector/src/primitive/vector_mut.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    fn unsplit(&mut self, other: Self) {
+        match_each_pvector_mut_pair!(self, other, |a, b| {
+            a.unsplit(b);
+        });
+    }
+}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
```
gatesn: And here

vortex-vector/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+mod vector;
+pub use vector::Vector;
+
+mod vector_mut;
+pub use vector_mut::VectorMut;
+
+/// Definition and implementation of [`VectorOps`] and [`VectorMutOps`] for [`Vector`] and
+/// [`VectorMut`], respectively.
+///
+/// [`VectorOps`]: ops::VectorOps
```
gatesn: You don't need this bit if the types are imported to this module

vortex-vector/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+mod bool;
+mod null;
+mod primitive;
+
+pub use bool::{BoolVector, BoolVectorMut};
+pub use null::{NullVector, NullVectorMut};
+pub use primitive::{GenericPVector, GenericPVectorMut, PrimitiveVector, PrimitiveVectorMut};
+
+/// Helper macros for working with the different variants of [`Vector`] and [`VectorMut`].
```
gatesn: Same here, move the docs into the module

vortex-vector/src/private.rs
```diff
@@ -0,0 +1,28 @@
...
+
+impl Sealed for NullVector {}
+impl Sealed for NullVectorMut {}
+
+impl Sealed for BoolVector {}
+impl Sealed for BoolVectorMut {}
+
+impl Sealed for PrimitiveVector {}
+impl Sealed for PrimitiveVectorMut {}
+impl<T> Sealed for GenericPVector<T> {}
```
gatesn: I'll be honest... I prefer the old names. They were more consistent with the rest of the codebase, e.g. vortex-scalar

gatesn (changes requested):

vortex-vector/src/lib.rs
connortsui20: Usually I do this when it's not immediately obvious what the module contains

vortex-vector/src/primitive/mod.rs
connortsui20: Personally I like it since it helps break up the file when reading, but I won't fight this

vortex-vector/src/ops.rs
connortsui20: Oh I see, so an `append_null_elements` rather than just append nulls to the mask (I always get confused with the builders as well)

vortex-vector/src/private.rs
connortsui20: Renamed to `PVector`

gatesn (approved):

PR #5011, opened 2025-10-20
pickling support

onursatici:
Implements `__reduce__` and `__reduce_ex__` methods to enable pickling of Vortex arrays in Python. Arrays are serialized using the Vortex IPC format.
For pickle protocol 5+ (Python 3.8+, PEP 574), uses PickleBuffer to keep array buffers separate from the main pickle stream rather than copying them inline. This enables us to use shared memory in the future to potentially zero-copy large arrays even across process boundaries. Protocol 4 and below serialise buffers inline as bytes.
Both protocols share the same deserialization path via `decode_ipc_array_buffers`, which reconstructs arrays from IPC-encoded buffer lists (or memoryviews).

### DIFF
--- a/vortex-python/Cargo.toml
+++ b/vortex-python/Cargo.toml
@@ -25,12 +25,15 @@ crate-type = ["rlib", "cdylib"]
 arrow-array = { workspace = true }
 arrow-data = { workspace = true }
 arrow-schema = { workspace = true }
+bytes = { workspace = true }
 itertools = { workspace = true }
 log = { workspace = true }
 object_store = { workspace = true, features = ["aws", "gcp", "azure", "http"] }
 parking_lot = { workspace = true }
 pyo3 = { workspace = true, features = ["abi3", "abi3-py311"] }
+pyo3-bytes = { workspace = true }
 pyo3-log = { workspace = true }
 tokio = { workspace = true, features = ["fs", "rt-multi-thread"] }
 url = { workspace = true }
 vortex = { workspace = true, features = ["object_store", "python", "tokio"] }
+vortex-ipc = { workspace = true }
--- a/vortex-python/Cargo.toml
+++ b/vortex-python/Cargo.toml
@@ -36,4 +36,3 @@ pyo3-log = { workspace = true }
 tokio = { workspace = true, features = ["fs", "rt-multi-thread"] }
 url = { workspace = true }
 vortex = { workspace = true, features = ["object_store", "python", "tokio"] }
-vortex-ipc = { workspace = true }
--- a/vortex-python/src/serde/mod.rs
+++ b/vortex-python/src/serde/mod.rs
@@ -8,11 +8,12 @@ use bytes::Bytes;
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 use pyo3::{Bound, Python};
-use vortex::ArrayRegistry;
+use vortex::ArraySessionExt;
 use vortex_ipc::messages::{DecoderMessage, MessageDecoder, PollRead};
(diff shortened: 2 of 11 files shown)


onursatici:
image

onursatici:
better
image

vortex-python/Cargo.toml
```diff
@@ -25,13 +25,16 @@ crate-type = ["rlib", "cdylib"]
...
 mimalloc = { workspace = true }
 object_store = { workspace = true, features = ["aws", "gcp", "azure", "http"] }
 parking_lot = { workspace = true }
 pyo3 = { workspace = true, features = ["abi3", "abi3-py311"] }
+pyo3-bytes = { workspace = true }
 pyo3-log = { workspace = true }
 tokio = { workspace = true, features = ["fs", "rt-multi-thread"] }
 url = { workspace = true }
 vortex = { workspace = true, features = ["object_store", "python", "tokio"] }
+vortex-ipc = { workspace = true }
```
gatesn: I think you should expose this via the main vortex crate

gatesn:
You're gonna need a PR description for this one 😅

gatesn:
I don't understand the two screenshots you sent? Also, is there any point implementing P4 at all? We're way beyond 3.8 as a min version requirement.

onursatici:
🤦‍♂️ right. I did initially add 4 to be a baseline to see if 5 would make a difference. Screenshots show the benchmarks ran comparing 4 vs 5. You are right we have min 3.11 required for vortex python so v4 would not be used, I will remove

onursatici:
@gatesn protocol 5 is introduced on py3.8, but made default just recently at py3.14, so we still need both

vortex-python/src/serde/mod.rs
```diff
@@ -19,6 +24,137 @@ pub(crate) fn init(py: Python, parent: &Bound<PyModule>) -> PyResult<()> {
...
+/// dtype_bytes : bytes
+///     The IPC-encoded dtype message
+///
+/// Returns
+/// -------
+/// Array
+///     The decoded Vortex array
+#[pyfunction]
+fn decode_ipc_array(array_bytes: Vec<u8>, dtype_bytes: Vec<u8>) -> PyResult<PyArrayRef> {
+    let registry = ArrayRegistry::canonical_only();
```
robert3005: I think you need a different registry here. Otherwise you will fail here deserialisation

onursatici: right, changed to using the default registry from vortex_file. We can also make this work with custom python arrays but that change feels a bit more involved, so supporting all default compressed encodings + canonical should be fine for now I guess

onursatici: well if we go with this, I need to add the `files` feature of vortex to vortex-python, or pull up the default registry somewhere else

onursatici: maybe not as this is not failing to be built on my laptop, fails only on CI. will have a look tomorrow

robert3005: On CI we do build projects in isolation since otherwise cargo does a union of all features so you don't know if youre dependencies are correct.

onursatici: yea, I think the failure I was seeing was some weird gh actions cache where it did give me a newer version of vortex-file where the default_registry was gone. In my branch it was there so it did compile fine. 
I was confused because in vortex-python we don't set the files feature but do workspace = true, and vortex-file comes from a default features. I will use the session equivalent for the default array registry

robert3005 (approved):

PR #5015, opened 2025-10-20
fix: include offset in fastlanes rle metadata

0ax1:
For backwards compat, the offset is default initialized with 0.


robert3005 (approved):
I think fortunately because we canonicalize in the layout writer but this is unfortunate

PR #5020, opened 2025-10-21
chore[duckdb]: set max threads

joseph-isaacs:
Set max threads on the rust side


0ax1 (approved):
Nice! I guess we still need to propagate the `-t` value as passed by the user or am I missing sth?

joseph-isaacs:
That is quite a bit more work, so will follow up

PR #5022, opened 2025-10-21
Array Operator API

gatesn:
# Open-ish Questions

These aren't really open questions, more just documenting our reasoning and trade-offs for arriving at the current API. We may well re-visit some of these assumptions.

## 1. Mutable output vectors

These are a little annoying to use, and in many places we end up allocating zero-capacity vectors to pass in, that ultimately never get used. This is because we allow implementations to return their own data zero-copy if they can.

This API allows us to execute chunks of a ChunkedArray directly into contiguous output memory. I cannot think of another use-case besides this. The question is how much do we hit this, and how much do we care?

Our current scanning logic currently chooses splits every time a column is partitioned, meaning we only ever see chunked arrays with at most one chunk (and hence we don't benefit from the optimization).

If we were to remove the output argument, we can also make `BatchKernelRef == BoxFuture<VortexResult<Vector>>`. Currently it is an async_trait, meaning we allocate to create the `Box<dyn KernelRef>`, then we allocate again to create the boxed future returned by the async trait. So we could save an allocation and pointer de-reference.

### Decision

* It's not worth the complexity or runtime cost of heap-allocating mutable output vectors just for them to potentially not be used.
* This also allows batch kernels to be pure `BoxFuture`, saving another heap allocation from what was formerly an async_trait.

## 2. To Arrow or not to Arrow

Instead of:
```rust
/// Trait for batch execution kernels.
#[async_trait]
pub trait BatchKernel: Send {
    /// Execute the batch kernel and produce a canonicalized vector.
    ///
    /// If the kernel can return data zero-copy from its own state, then it should prefer to do so.
    /// If the kernel is producing _new_ data, it should write this data into the provided `out`
    /// vector.
    async fn execute(self: Box<Self>, out: VectorMut) -> VortexResult<Vector>;
}
```

We would have
```rust
/// Trait for batch execution kernels.
#[async_trait]
pub trait BatchKernel: Send {
    async fn execute(self: Box<Self>) -> VortexResult<ArrowArrayRef>;
}
```

Note that we also lost the mutable output argument. That's because Arrow doesn't currently have mutable arrays. We could PR them if we really care about the ChunkedArray optimization, but as we say above, perhaps it doesn't matter.

This Arrow-focussed API would let us delegate a lot of array implementations almost directly to arrow-rs. Admittedly, the rust crate has some bugs we've been running into and fixing them upstream takes time, particularly if we want to stay in sync with DataFusion's arrow version. There are also many missing kernels, e.g. most compute functions of ListView arrays.

That said, it would make writing our plugin infrastructure significantly easier and more efficient if we could simply use Arrow for the vector part of our ABI. 

We would also see community benefits as we upstream more compute into arrow-rs.

This would also help us migrate (as we believe we inevitably will) to have physical types in Vortex. This would mean we are not bound by the Vortex canonical representation, and arrays could emit VarBin _or_ VarBinView depending on the desired output type (indicated using a PhysicalCastArray), and the actual input type found serialized on disk.

It's easy to grok what Vortex is in this world. Simply scalar expressions over Arrow arrays.

### Decision - not to Arrow

For now, we will continue to use Vortex Vectors. They will for now remain Arrow-compatible. Since vectors represent physical data, they actually have less metadata attached to them than arrow-rs arrays, for example structs do not carry field names. This also makes them much cheaper to convert directly to/from Arrow C ABI if we don't go via arrow-rs as we can avoid constructing the Arc'd tree of arrow arrays.

We may one day decide we are so close to Arrow that we should unwind and unify with vectors. But for now, we benefit from having independent APIs.

## 3. To async or not to async

This PR shows an async API for `BatchKernel`. This allows arrays to perform I/O during execution. Note that we will have an alternate API for pipelined compute over 1k elements at a time that is not async. Most scalar performance-sensitive kernels would implement this API in the future.

In a non-async world, Vortex would struggle to represent expressions such as "map these strings through an LLM". We may have to introduce interesting shortcuts to evaluate these expressions within a scan (which _is_ an async API).

Having native support for async expressions allows these to be naturally expressed and executed as part of a Vortex array evaluation, rather than delegating the evaluation to part of the scan, or to the query engine that is ultimately calling into Vortex.

Async also allows us to implement short-circuiting of I/O. For example, only load the VarBin buffers that are referenced by a filtered varbin array, instead of eagerly loading all of them.

One major downside is that our plugin ABI becomes quite a bit more complex. Foreign implementations would need to return an enum similar to the Rust Future API, and we'd need to poll across an FFI boundary. 

We could wrap up synchronous arrays as plugins, but there's no way to short-circuit. So we would need to evaluate all children eagerly up-front.

### Decision - to async

The batch execution mode for a vortex array will be async, allowing us to perform I/O and spill-to-disk buffer management. However, we also support "pipelined" CPU execution which will be a push-based blocking DAG that most high performance arrays should prefer to implement. Eventually, arrays will have the choice to implement either of these.


vortex-array/src/array/operator.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::{Array, ArrayAdapter, ArrayRef};
+
+/// Array functions as provided by the `OperatorVTable`.
+///
+/// Note: the public functions such as "execute" should move onto the main `Array` trait when
+/// operators is stabilized. The other functions should remain on a `pub(crate)` trait.
+#[async_trait]
+pub trait ArrayOperator: 'static + Send + Sync {
+    /// Execute the array producing a canonical vector using a [`SingleThreadRuntime`].
+    #[cfg(test)]
```
joseph-isaacs: `test-harness`?

gatesn: I think this disappears very quickly tbh in favor of async tests

vortex-array/src/compute/arrays/logical.rs
```diff
@@ -0,0 +1,253 @@
...
+use crate::vtable::{
+    ArrayVTable, NotSupported, OperatorVTable, SerdeVTable, VTable, VisitorVTable,
+};
+use crate::{
+    Array, ArrayBufferVisitor, ArrayChildVisitor, ArrayEq, ArrayHash, ArrayRef,
+    DeserializeMetadata, EmptyMetadata, EncodingId, EncodingRef, Precision, vtable,
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Enum)]
+pub enum LogicalOperator {
```
joseph-isaacs, line 30: can we have a doc str here

vortex-array/src/compute/arrays/logical.rs
```diff
@@ -0,0 +1,253 @@
...
+impl LogicalArray {
+    /// Create a new logical array.
+    pub fn new(lhs: ArrayRef, rhs: ArrayRef, operator: LogicalOperator) -> Self {
+        assert_eq!(
+            lhs.len(),
+            rhs.len(),
+            "Logical arrays require lhs and rhs to have the same length"
+        );
+
+        // TODO(ngates): should we automatically cast non-null to nullable if required?
```
joseph-isaacs, line 62: yes

vortex-array/src/compute/arrays/logical.rs
```diff
@@ -0,0 +1,253 @@
...
+    type OperatorVTable = Self;
+
+    fn id(encoding: &Self::Encoding) -> EncodingId {
+        match encoding.operator {
+            LogicalOperator::And => EncodingId::from("vortex.and"),
+            LogicalOperator::AndKleene => EncodingId::from("vortex.and_kleene"),
+            LogicalOperator::Or => EncodingId::from("vortex.or"),
+            LogicalOperator::OrKleene => EncodingId::from("vortex.or_kleene"),
+            LogicalOperator::AndNot => EncodingId::from("vortex.and_not"),
+        }
```
joseph-isaacs, line 115: think we should namespace this any further?

vortex-array/src/compute/arrays/logical.rs
```diff
@@ -0,0 +1,253 @@
...
+
+#[derive(Debug, Clone)]
+pub struct LogicalEncoding {
+    // We include the operator in the encoding so each operator is a different encoding ID.
+    // This makes it easier for plugins to construct expressions and perform pushdown
+    // optimizations.
+    operator: LogicalOperator,
+}
+
+#[allow(clippy::mem_forget)]
```
joseph-isaacs, line 88: ?

joseph-isaacs: in enummap?

vortex-array/src/compute/arrays/logical.rs
joseph-isaacs, line 1: arrays have a lot of boiler plate

vortex-array/src/operator/mod.rs
```diff
@@ -20,10 +20,13 @@
...
 //!
-//! We also introduce the idea of an executor that can evaluate an operator tree efficiently. It
+//! We also introduce the idea of an execution that can evaluate an operator tree efficiently. It
 //! supports common subtree elimination, as well as extracting sub-graphs for pipelined and GPU
-//! execution. The executor is also responsible for managing memory and scheduling work across
+//! execution. The execution is also responsible for managing memory and scheduling work across
 //! different execution resources.
+//!
+
+#![allow(dead_code)]
```
joseph-isaacs, line 29: do we need this?

vortex-array/src/vtable/operator.rs
```diff
@@ -1,19 +1,107 @@
...
+    /// Compute the array over all-constant inputs.
+    ///
+    /// Instead of every array implementing constant folding as part of `reduce_children`, the
+    /// optimizer will call into this function when all inputs are constant.
+    fn compute_constant(
+        _array: &V::Array,
+        _children: &[&Scalar],
+    ) -> VortexResult<Option<ArrayRef>> {
+        Ok(None)
+    }
```
joseph-isaacs: im not sure about this? I would prefer constant folding

joseph-isaacs: do you have an example of where constant folding would be hard?

joseph-isaacs (approved):
Going to approve but I do think the comments need addressing

vortex-array/src/compute/arrays/logical.rs
gatesn: I'm not sure whether we should mutate children in the constructor. Feels a bit gross. Maybe the caller should just be forced to do it? e.g. the function that converts expr -> array?

vortex-array/src/compute/arrays/logical.rs
gatesn: Could put `vortex.compute.and` ?

vortex-array/src/compute/arrays/logical.rs
gatesn: Yeah, it's a static thing anyway, and it calls mem::forget inside the macro annoyingly.

vortex-array/src/compute/arrays/logical.rs
gatesn: They do indeed. I want to consolidate the vtables eventually. Can do that in intermediate steps they're just big annoying refactor PRs.

vortex-array/src/operator/mod.rs
gatesn: Sadly yes. I don't want to kill this code yet because I want to bring back the pipeline stuff soon

vortex-array/src/vtable/operator.rs
gatesn: It's not hard, it's just this makes it easier to remember to do. We could even make it mandatory? This API is still in progress

vortex-array/src/compute/arrays/logical.rs
gatesn: Actually I don't think so. It's nice that compute arrays aren't special?

vortex-array/src/vtable/operator.rs
joseph-isaacs: I really don't think we should have this

PR #5023, opened 2025-10-21
fix[decimal]: add sum support for decimals to chunked and constant

joseph-isaacs:
The fuzzer found these.

### DIFF
--- a/vortex-array/src/stats/mod.rs
+++ b/vortex-array/src/stats/mod.rs
@@ -205,7 +205,7 @@ impl Stat {
                         }
                     },
                     DType::Extension(ext_dtype) => self.dtype(ext_dtype.storage_dtype())?,
-                    DType::Decimal(decimal_dtype, nullability) => {
+                    DType::Decimal(decimal_dtype, _) => {
                         // Both Spark and DataFusion use this heuristic.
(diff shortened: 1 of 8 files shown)


vortex-array/src/stats/mod.rs
```diff
@@ -205,15 +205,15 @@ impl Stat {
...
+                    DType::Decimal(decimal_dtype, _) => {
                         // Both Spark and DataFusion use this heuristic.
                         // - https://github.com/apache/spark/blob/fcf636d9eb8d645c24be3db2d599aba2d7e2955a/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/Sum.scala#L66
                         // - https://github.com/apache/datafusion/blob/4153adf2c0f6e317ef476febfdc834208bd46622/datafusion/functions-aggregate/src/sum.rs#L188
                         let precision =
                             u8::min(DECIMAL256_MAX_PRECISION, decimal_dtype.precision() + 10);
                         DType::Decimal(
                             DecimalDType::new(precision, decimal_dtype.scale()),
-                            *nullability,
+                            Nullable,
```
joseph-isaacs, line 216: This was incorrect

AdamGS (approved):

robert3005:
I think we should unify the behaviour - for decimal array sum we widen the type and use wrapping arithmetic but here we widen the type and use checked arithmetic. I think I convinced myself that with the widended type you will only hit the issue if you ever have an array bigger than 2^32 elements.

PR #5024, opened 2025-10-21
performance[vortex-datafusion]: support pushing down filters on struct columns

asubiotto:
The motivation for this PR is to allow pushing down filters of the type `my_struct.field = 'foo'`. There are some related fixes for things that broke when I added support for these, namely rewriting the filter on a physical struct schema that was a subset of the full table schema and right-hand side literals that are of the dictionary type. Please see commits for details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ datafusion-common-runtime = { version = "50" }
 datafusion-datasource = { version = "50", default-features = false }
 datafusion-execution = { version = "50" }
 datafusion-expr = { version = "50" }
+datafusion-functions = { version = "50" }
 datafusion-physical-expr = { version = "50" }
 datafusion-physical-expr-adapter = { version = "50" }
 datafusion-physical-expr-common = { version = "50" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8652,6 +8652,7 @@ dependencies = [
  "datafusion-datasource",
  "datafusion-execution",
  "datafusion-expr",
+ "datafusion-functions",
  "datafusion-physical-expr",
  "datafusion-physical-expr-adapter",
  "datafusion-physical-expr-common",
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -101,6 +101,7 @@ datafusion-common-runtime = { version = "50" }
 datafusion-datasource = { version = "50", default-features = false }
 datafusion-execution = { version = "50" }
 datafusion-expr = { version = "50" }
+datafusion-functions = { version = "50" }
```
asubiotto, line 104: The only reason to add this dependency was to be able to check that a scalar function was a `get_field` function. The other option is to check that `fn.name() == 'get_field'`, but that seemed brittle.

AdamGS: I'm all for delegating to DF in these cases

AdamGS:
Took a quick look and everything looks good, I'll take a deeper look tomorrow morning and probably merge it then

a10y:
Is this just for the `select get_field(x, 'y')` scalar function pushdown? `select x.y` should already pushdown I think?

See also #4942, which once that's merged will allow pruning on struct fields more completely

a10y (approved):

a10y:
Thank you!

asubiotto:
> Is this just for the `select get_field(x, 'y')` scalar function pushdown? `select x.y` should already pushdown I think?

No, I think `select x.y` is just syntactic sugar for `select get_field(x, 'y')`. We only use dot notation in our queries and the expression datafusion was calling `try_pushdown_filters` with was a `binaryexpr` with `get_field` and a `literal` which was rejected by the previous implementation.

> See also #4942, which once that's merged will allow pruning on struct fields more completely

Nice! I saw the referenced issue but didn't realize you had a PR out. We have very deeply nested schemas so this should help a lot. I also saw that stats are not collected for struct fields currently so we can't prune, I guess #4942 adds that ability?

Issue #4973, opened 2025-10-16
 ̶r̶l̶e̶ ̶ Delta triggers undefined behavior


a10y:
### Describe the bug

I'm running some tests using a small sample of the new Github Archive dataset and hitting undefined behavior check failures in RLE due to out of bounds indexing in the decode step:

```
warning: `bench-vortex` (lib) generated 1 warning (run `cargo fix --lib -p bench-vortex` to apply 1 suggestion)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
     Running `target/debug/query_bench gharchive --targets 'datafusion:vortex' --scale-factor=1`
2025-10-16T21:01:23.416551Z  INFO bench_vortex::realnest::gharchive: bench-vortex/src/realnest/gharchive.rs:200: gharchive data generated in /Volumes/Code/Reviews/vortex/bench-vortex/data/gharchive/vortex-file-compressed/events.vortex
2025-10-16T21:01:23.455999Z  INFO bench_vortex::realnest::gharchive: /Volumes/Code/Reviews/vortex/bench-vortex/src/realnest/gharchive.rs:241: registering table for GHARCHIVE: file:///Volumes/Code/Reviews/vortex/bench-vortex/data/gharchive/vortex-file-compressed/

thread 'tokio-runtime-worker' panicked at /Users/aduffy/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fastlanes-0.4.0/src/rle.rs:58:42:
unsafe precondition(s) violated: slice::get_unchecked requires that the index is within the slice

This indicates a bug in the program. This Undefined Behavior check is optional, and cannot be relied on for safety.
stack backtrace:
   0:        0x107d565ac - std::backtrace_rs::backtrace::libunwind::trace::h0c3f80ac1dc6e20d
                               at /rustc/a65b04d7c9a1eaf02b62667da1c203c8a07976d9/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1:        0x107d565ac - std::backtrace_rs::backtrace::trace_unsynchronized::hfc766a3dd1b4ce19
                               at /rustc/a65b04d7c9a1eaf02b62667da1c203c8a07976d9/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2:        0x107d565ac - std::sys::backtrace::_print_fmt::h742520ded1cc1feb
[...]

PR #5026, opened 2025-10-21
change gharchive query to reduce memory blowup, disable Delta

a10y:
Disable RLE scheme, change `select *` to `select count(*)`. That in combination with #5024 stops us from OOMing every time we run this benchmark.

### DIFF
--- a/vortex-btrblocks/src/integer.rs
+++ b/vortex-btrblocks/src/integer.rs
@@ -765,9 +765,7 @@ mod tests {
     use vortex_sparse::SparseEncoding;
     use vortex_utils::aliases::hash_set::HashSet;
 
-    use crate::integer::{
-        IntCompressor, IntegerStats, RLE_INTEGER_SCHEME, SequenceScheme, SparseScheme,
-    };
+    use crate::integer::{IntCompressor, IntegerStats, SequenceScheme, SparseScheme};
     use crate::{Compressor, CompressorStats, Scheme};
 
     #[test]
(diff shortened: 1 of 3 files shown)


AdamGS:
Does it seem like datafusion loads all columns in this case? I wonder if it wants the null counts for all arrays.
If that's the case, that could be really cool case for expression pushdown, we could load just the validity data without the actual arrays for any column that doesn't have a filter.

a10y:
Oh huh, we don't push the filter down at all apparently

image

I have been under the mistaken impression that we are able to. I see also #5024

a10y:
Merging @asubiotto's change helps a lot here. I'm going to review and try and merge that and then revisit this
image

robert3005:
Why are we disabling RLE?

0ax1:
> Why are we disabling RLE?

I'm pretty sure at this point the culprit is delta. 

```
     pub fn encode(array: &PrimitiveArray) -> VortexResult<Self> {
-        match_each_native_ptype!(array.ptype(), |T| { rle_encode_typed::<T>(array) })
+        match_each_native_ptype!(array.ptype(), |T| {
+            let rle_array = rle_encode_typed::<T>(array);
+            let _slice = array.as_slice::<T>();
+            std::hint::black_box(rle_decompress(&rle_array.as_ref().unwrap()));
+            rle_array
+        })
```

Applying this diff running in debug, did not trigger the issue for GHArchive. The difference being that the indices are not delta encoded here yet. If the issue would be within RLE, this should also trigger the out of bounds access in my understanding. I did this in encode to see the uncompressed pritimitive array in case the assertion fires.

a10y:
I can just disable Delta if that is preferable

vortex-btrblocks/src/integer.rs
```diff
@@ -881,19 +877,19 @@ mod tests {
...
-        let mut values = Vec::new();
-        values.extend(iter::repeat_n(42i32, 100));
-        values.extend(iter::repeat_n(123i32, 200));
-        values.extend(iter::repeat_n(987i32, 150));
-
-        let array = PrimitiveArray::new(Buffer::copy_from(&values), Validity::NonNullable);
-        let compressed = RLE_INTEGER_SCHEME
-            .compress(&IntegerStats::generate(&array), false, 3, &[])
-            .unwrap();
+    //#[test]
```
0ax1: Keep the tests if we keep RLE.

0ax1:
The new query succeeds by the way without triggering an out of bounds, even with Delta + RLE enabled. Just ran this locally.

a10y:
That's because the query is no longer decoding the array that needs to be decoded to trigger #4973 , but the data is still being written corrupted

0ax1:
> That's because the query is no longer decoding the array that needs to be decoded to trigger #4973 , but the data is still being written corrupted

Yeah makes sense

gatesn (approved):
Let's do it as temporary fix, then can bring the tests back

robert3005 (approved):

PR #5027, opened 2025-10-21
feat: merge supports nullable children

danking:
I think this is the right behavior, but these are admittedly weird:

```
merge(null({a: int, b:int})) = {a: [null], b: [null]}
```

```
merge(null({a: int, b: int}), null({b:int, c:int})) = {a: [null], b: [null], c: [null]}
```

### DIFF
--- a/vortex-expr/src/exprs/merge.rs
+++ b/vortex-expr/src/exprs/merge.rs
@@ -2,9 +2,11 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use std::hash::Hash;
+use std::ops::Not;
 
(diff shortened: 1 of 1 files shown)


gatesn:
I'm not quite sure I can interpret the semantics you're proposing. But I agree that merge with nullable struct arrays seems weird. I would even go so far as to say the structs must be non-nullable and we have a separate function that pushes nulls from a struct array into all fields?

joseph-isaacs:
Have different expr for this that pushes struct null into its children

danking:
See instead https://github.com/vortex-data/vortex/pull/5030

PR #5030, opened 2025-10-21
feat: push_nulls

danking:
In response to comments at https://github.com/vortex-data/vortex/pull/5027


danking:
We will attempt to ban nullable structs from the Spiral system instead.

Issue #5028, opened 2025-10-21
Tracking Issue: Vectors


connortsui20:
This is a tracking issue for adding vectors to Vortex.

Related: https://github.com/vortex-data/vortex/issues/4492
RFC: TODO LINK (https://github.com/vortex-data/vortex/pull/4726)

# Vectors

Since "vector" is a heavily overloaded term, we we define a vector in Vortex specifically as:

- A fully decompressed (which we call "Canonical" in Vortex) array of contiguous data
- A datatype that can hold itself as children (a vector can hold other child vectors)
- The data is _recursively_ **de**compressed, which means all child vectors are also fully decompressed (this is different from the current _canonical encodings_, which allow compressed children)
- Vectors are either borrowed or owned (`Vector` and `VectorMut`), mimicking the popular `Bytes` and `BytesMut` from the [`bytes`](https://docs.rs/bytes/latest/bytes/) crate.
- The owned version of vectors (`VectorMut`) is fully mutatable

See RFC 0001 (TODO LINK) for more information on why we need vectors. To summarize, instead of potentially compressed `Array`s being the intermediate representation (IR) between edges in our decompression plans (which is the model we use for our current compute functions), we want to enforce that fully decompressed arrays (which will be vectors) are the IR so that we can:

- Guarantee progress in decompression at every step (right now, compute functions can do whatever they want) 
- Redefine `Array`s to represent logical query plans to decompress data. This allows us to perform better optimizations (pushdown and common subexpression elimination) if we know exactly what data is flowing between nodes in our decompression plan

## Why not Arrow?

_Copied and pasted from the RFC_

There are a few reasons we are not directly using Arrow arrays:

* Vortex vectors are recursively built around variable-width primitives, allowing us to use the narrowest type possible
  (e.g. we can narrow dictionary codes, run-end offsets, list view sizes, etc.).
* Internally we use the `vortex-buffer` crate, meaning we preserve our capabilities around runtime alignment.
* We reserve the right to diverge from Arrow layouts in the future. For example, a `VarBinVector` could use in-memory
  pointers like DuckDB, rather than logical offsets. This makes the vectors cheaper to concatenate and slice, but
  incurs a conversion cost to/from Arrow.

We may re-visit this decision in the future prior to finalizing the Vortex extension API.

# Steps / History

- [x] https://github.com/vortex-data/vortex/pull/5010
- [x] https://github.com/vortex-data/vortex/pull/5056
[...]

PR #5031, opened 2025-10-21
Chore: Add more docs + `from_option_iter`

connortsui20:

Adds documentation examples to everything.

Also removes some of the macros that were only specific to a single method that are probably not super useful for users.

There is still an issue where `try_into_mut` on `Mask` doesn't really behave as it should (because we need to still change out arrow's `BooleanBuffer` for our `BitBuffer` inside `MaskValues`, but I have TODOs on those doctests.


vortex-vector/src/bool/vector_mut.rs
```diff
@@ -26,6 +76,47 @@ impl BoolVectorMut {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use vortex_vector::{BoolVectorMut, VectorMutOps};
+    ///
+    /// let mut vec = BoolVectorMut::from_option_iter([Some(true), None, Some(false)]);
+    /// assert_eq!(vec.len(), 3);
+    /// ```
+    pub fn from_option_iter<I>(iter: I) -> Self
```
gatesn: I think you want impl FromIterator<Option<...>>

gatesn (changes requested):

connortsui20:
I'm wondering if it makes sense to enforce that the immutable vectors are really just borrowed / frozen versions of the mutable vectors, and thus the constructors should only exist on the mutable vectors...

This might be too restrictive? But at the same time `freeze` needs to be constant time anyways and it means we don't have to duplicate so much code.

gatesn:
Yeah that's generally how these things work. Seems fine to me to only have the ctors on mutable then have the caller use freeze

vortex-vector/src/bool/vector_mut.rs
```diff
@@ -28,6 +78,49 @@ impl BoolVectorMut {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use vortex_vector::{BoolVectorMut, VectorMutOps};
+    ///
+    /// let mut vec = BoolVectorMut::from_iter([Some(true), None, Some(false)]);
+    /// assert_eq!(vec.len(), 3);
+    /// ```
+    fn from_iter<I>(iter: I) -> Self
```
gatesn, line 96: @robert3005 any ideas on what the canonical way to do from TrustedLen iter is?

vortex-vector/src/primitive/generic_mut.rs
```diff
@@ -31,6 +114,49 @@ impl<T: NativePType> PVectorMut<T> {
     }
 }
 
+impl<T: NativePType> FromIterator<Option<T>> for PVectorMut<T> {
```
gatesn, line 119: You probably want `FromIterator<T>` also

vortex-vector/src/bool/vector_mut.rs
robert3005: Basically you need a different method to take a trusted_len_iter because otherwise you need some flavour of specialisation. We have `pub fn from_trusted_len_iter<I: TrustedLen<Item = T>>(iter: I) -> Self` on Buffer

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -150,12 +150,11 @@ impl BitBufferMut {
 
     /// Reserve additional bit capacity for the buffer.
     pub fn reserve(&mut self, additional: usize) {
-        let required_capacity = (self.offset + self.len + additional).div_ceil(8);
-        let buffer_capacity = self.buffer.capacity();
-        if required_capacity > self.buffer.capacity() {
-            let additional = required_capacity - buffer_capacity;
```
connortsui20, line 156: the `required_capacity - buffer_capacity` was the incorrect part

vortex-vector/src/bool/vector_mut.rs
connortsui20: I'll do this in the next PR

gatesn: Also in next PR you may want to move these functions to `from_iter.rs` else this file gets quite long

gatesn (approved):

PR #5032, opened 2025-10-22
chore[array]: tree display no `/0`

joseph-isaacs:


vortex-array/src/array/display/tree.rs
```diff
@@ -32,13 +32,18 @@ impl<'a, 'b: 'a> TreeFormatter<'a, 'b> {
     fn format(&mut self, name: &str, array: ArrayRef) -> fmt::Result {
         let nbytes = array.nbytes();
         let total_size = self.total_size.unwrap_or(nbytes);
+        let percent = if total_size == 0 {
```
0ax1, line 35: Would have thought there's something like checked_div in the std lib for float but I guess this is the way to do it.

0ax1 (approved):

PR #5036, opened 2025-10-22
chore[fuzz]: add new compute function fuzzing `[fill_null, mask, max/min, sum]`

joseph-isaacs:
The baseline here is the canonical implementation, so this really check if the custom compute is correct using the canonical array as the baseline.


robert3005:
I think you need alternative implementations from the canonical ones, otherwise you don't actually fuzz our canonical implenentations. You can do it using Vec or arrow otherwise we claim to fuzz those functions but we really are not

joseph-isaacs:
This is what I said in the description. I don't think its not worth having this. Just seems like a next step to add non-canonical baselines

robert3005 (approved):
This is great. Would be nice to add fuzzing of canonical impls but will not block on it

PR #5038, opened 2025-10-22
Chore: Vectors even more cleanup

connortsui20:

These didn't make it in the the last one.

All todos are basically done and once I implement the `TrustedLen` iterator stuff I'll be able to take off on implementing the remaining vector types.

Note that the commit granularity will make it easier to review


vortex-vector/src/bool/from_iter.rs
```diff
@@ -0,0 +1,79 @@
...
+    /// assert_eq!(vec.len(), 3);
+    /// ```
+    fn from_iter<I>(iter: I) -> Self
+    where
+        I: IntoIterator<Item = Option<bool>>,
+    {
+        let iter = iter.into_iter();
+        let (lower_bound, _) = iter.size_hint();
+
+        let mut bits = Vec::with_capacity(lower_bound);
```
gatesn, line 31: Use BitBufferMut here

vortex-vector/src/primitive/from_iter.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// assert_eq!(vec.len(), 3);
+    /// ```
+    fn from_iter<I>(iter: I) -> Self
+    where
+        I: IntoIterator<Item = Option<T>>,
+    {
+        let iter = iter.into_iter();
+        let (lower_bound, _) = iter.size_hint();
+
+        let mut elements = Vec::with_capacity(lower_bound);
```
gatesn, line 32: Use BitBufferMut

gatesn (changes requested):

vortex-vector/src/bool/from_iter.rs
connortsui20: The change is in #5040

gatesn: Wanna port it over here while we're here?

connortsui20:
Closing in favor of #5040

PR #5039, opened 2025-10-22
chore: use offset_from_unsigned instead of byte_offset_from

robert3005:

### DIFF
--- a/vortex-buffer/src/buffer_mut.rs
+++ b/vortex-buffer/src/buffer_mut.rs
@@ -500,8 +500,7 @@ impl<T> BufferMut<T> {
             }
         }
 
-        // TODO(joe): replace with ptr_sub when stable
-        let length = self.len() + unsafe { dst.byte_offset_from(begin) as usize / size_of::<T>() };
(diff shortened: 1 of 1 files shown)


vortex-buffer/src/buffer_mut.rs
```diff
@@ -526,8 +525,8 @@ impl<T> BufferMut<T> {
                 dst = dst.add(1);
             }
         });
-        // TODO(joe): replace with ptr_sub when stable
-        let length = self.len() + unsafe { dst.byte_offset_from(begin) as usize / size_of::<T>() };
+
+        let length = self.len() + unsafe { dst.offset_from_unsigned(begin) };
```
joseph-isaacs, line 529: the point of the todo was to avoid the two ops, which this doesn't do but I guess its a pretty small win

joseph-isaacs (approved):

gatesn (approved):

vortex-buffer/src/buffer_mut.rs
robert3005: So what is the thing you want to do here? Do the sub from the head of the bytes?

PR #5040, opened 2025-10-22
Chore: clean up vectors more

connortsui20:
Cleans up some things in the crate (this is a purely cosmetic change)


vortex-buffer/src/lib.rs
```diff
@@ -73,7 +73,8 @@ mod memmap2;
 #[cfg(feature = "serde")]
 mod serde;
 mod string;
-mod trusted_len;
+
+pub mod trusted_len;
```
gatesn: `pub use trusted_len::*` instead

vortex-vector/src/bool/from_iter.rs
```diff
@@ -4,10 +4,81 @@
...
+        let upper_bound = iter
+            .size_hint()
+            .1
+            .vortex_expect("`TrustedLen` iterator did not have an upper bound on length");
+
+        let mut bits = BitBufferMut::with_capacity(upper_bound);
+        let mut validity = MaskMut::with_capacity(upper_bound);
+
+        // TODO(connor): This can potentially be made more efficient by adding a similar method to
+        // `BitBufferMut` directly.
```
gatesn: I think you need to do this first. This impl doesn't make use of the "trusted" information at all right now

vortex-vector/src/bool/from_iter.rs
```diff
@@ -4,10 +4,81 @@
...
+    /// use vortex_vector::{BoolVectorMut, VectorMutOps};
+    ///
+    /// let mut vec = BoolVectorMut::from_trusted_len_iter([true, false, false, true]);
+    /// assert_eq!(vec.len(), 4);
+    /// ```
+    pub fn from_trusted_len_iter<I>(iter: I) -> Self
+    where
+        I: TrustedLen<Item = bool>,
+    {
+        let buffer = BitBufferMut::from_iter(iter);
```
gatesn: Similarly, needs to be from_trusted

vortex-vector/src/primitive/from_iter.rs
```diff
@@ -4,11 +4,87 @@
...
+    ///
+    /// let mut vec = PVectorMut::<i32>::from_trusted_len_iter([1, 2, 3, 4]);
+    /// assert_eq!(vec.len(), 4);
+    /// ```
+    pub fn from_trusted_len_iter<I>(iter: I) -> Self
+    where
+        I: TrustedLen<Item = T>,
+    {
+        // We `from_iter` do the work of allocating the correct amount.
+        let buffer = BufferMut::from_iter(iter);
```
gatesn: Yeah, this also isn't "trusted"

gatesn (changes requested):

vortex-vector/src/bool/from_iter.rs
connortsui20: The `from_iter` here looks at the upper bound, and since `TrustedLen` basically just guarantees an upper bound this does the same thing

vortex-vector/src/bool/from_iter.rs
connortsui20: It takes the upper bound, which basically gets us all the way there? Doing the raw writes may or may not be faster since we have to match against each value and incur a branch

vortex-vector/src/bool/from_iter.rs
connortsui20: For reference in `vortex-buffer/src/bit/buf_mut.rs`:

```rust
impl FromIterator<bool> for BitBufferMut {
    fn from_iter<T: IntoIterator<Item = bool>>(iter: T) -> Self {
        let iter = iter.into_iter();
        let (low, high) = iter.size_hint();
        if let Some(len) = high {
            let mut buf = BitBufferMut::new_unset(len);
            for (i, v) in iter.enumerate() {
                if v {
                    // SAFETY: i is in bounds
                    unsafe { buf.set_unchecked(i) }
                }
            }
            buf
        } else {
```

So I think this is sufficient

vortex-vector/src/primitive/from_iter.rs
connortsui20: See above

connortsui20:
This PR no longer has anything to do with trusted length iterators

PR #5045, opened 2025-10-23
Additions to vector API

gatesn:


vortex-vector/src/bool/vector.rs
```diff
@@ -23,12 +23,32 @@ pub struct BoolVector {
     pub(super) validity: Mask,
 }
 
+impl BoolVector {
```
connortsui20, line 26: I think we want constructors only on the mutable vectors?

vortex-vector/src/lib.rs
connortsui20, line 1: Personally I like splitting up the modules as they were before because it makes it easy to distinguish module groupings, what is getting reexported, what is not, etc

connortsui20 (approved):

vortex-vector/src/bool/vector.rs
gatesn: How do I construct an immutable vector from cloned components? 

If you force everyone via mut, you force a copy

PR #5046, opened 2025-10-23
Simplify the serde implementation for StructFields

AdamGS:
Was just curious if its possible to this, and apparently it is. I don't think there is any compile time difference, it just saves on some code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8735,6 +8735,7 @@ dependencies = [
  "arrow-schema",
  "flatbuffers",
  "half",
+ "insta",
  "itertools 0.14.0",
  "jiff",
  "num-traits",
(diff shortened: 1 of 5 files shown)


joseph-isaacs (commented):
nice, did you run locally with old clickbench or gharchive files? I think its a good idea to check before merging

joseph-isaacs (approved):

AdamGS:
We don't use `serde` for any persistent data. As far as I can tell, its only use is that it allows serializing dtypes using `serde` which is probably useful.

AdamGS:
We don't use the `serde` feature for any persistent data. Also added a test to make sure its compatibly and tested it. We should probably improve some of this structure, but that's a breaking change.

PR #5048, opened 2025-10-23
bug[fastlanes]: delta encoding validity must be transposed

joseph-isaacs:


gatesn (changes requested):
No changes, just want to make sure I see the solution to this!

joseph-isaacs:
There are two options we store validity in this array or we transpose and push down.
I vote for number two to allow more freedom to the underlying encoding but ideally we wound fused this with the underlying array or some how signify fastlanes vs regular ordering?

PR #5049, opened 2025-10-23
Chore: TODO

connortsui20:
I got tripped up when reading this and misunderstood how this worked so I just decided to clean it up

Edit: now this PR is also going to attempt to remove `TrustedLen` iterator

### DIFF
--- a/vortex-buffer/src/bit/buf_mut.rs
+++ b/vortex-buffer/src/bit/buf_mut.rs
@@ -221,6 +221,16 @@ impl BitBufferMut {
         unsafe { unset_bit_unchecked(self.buffer.as_mut_ptr(), self.offset + index) }
     }
 
+    /// Foces the length of the `BitBufferMut` to `new_len`.
+    ///
+    /// # Safety
+    ///
+    /// - `new_len` must be less than or equal to [`capacity()`](Self::capacity)
+    /// - The elements at `old_len..new_len` must be initialized
+    pub unsafe fn set_len(&mut self, new_len: usize) {
+        self.len = new_len;
+    }
+
     /// Truncate the buffer to the given length.
     pub fn truncate(&mut self, len: usize) {
         if len > self.len {
@@ -549,24 +559,35 @@ impl From<Vec<bool>> for BitBufferMut {
 
 impl FromIterator<bool> for BitBufferMut {
     fn from_iter<T: IntoIterator<Item = bool>>(iter: T) -> Self {
-        let iter = iter.into_iter();
-        let (low, high) = iter.size_hint();
-        if let Some(len) = high {
-            let mut buf = BitBufferMut::new_unset(len);
-            for (i, v) in iter.enumerate() {
-                if v {
-                    // SAFETY: i is in bounds
-                    unsafe { buf.set_unchecked(i) }
-                }
+        let mut iter = iter.into_iter();
+
+        // Note that these hints might be incorrect.
+        let (lower_bound, upper_bound_opt) = iter.size_hint();
+        let capacity = upper_bound_opt.unwrap_or(lower_bound);
+
+        let mut buf = BitBufferMut::new_unset(capacity);
+
+        // Directly write within our known capacity.
+        for i in 0..capacity {
+            let Some(v) = iter.next() else {
+                // SAFETY: We are definitely under the capacity and all values are already
+                // initialized from `new_unset`.
+                unsafe { buf.set_len(i) };
+                return buf;
+            };
+
+            if v {
+                // SAFETY: We have ensured that we are within the capacity.
+                unsafe { buf.set_unchecked(i) }
             }
-            buf
-        } else {
-            let mut buf = BitBufferMut::with_capacity(low);
-            for v in iter {
-                buf.append(v);
-            }
-            buf
         }
+
+        // Append the remaining items (as we do not know how many more there are).
+        for v in iter {
+            buf.append(v);
+        }
+
+        buf
     }
 }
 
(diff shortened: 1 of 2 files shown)


vortex-buffer/src/buffer_mut.rs
```diff
@@ -477,59 +477,105 @@ impl<T> AsMut<[T]> for BufferMut<T> {
...
+            // derived from a valid reference to byte data.
+            unsafe { dst.write(item) };
+
+            // Note: We used to have `dst.add(iteration).write(item)`, here. However this was much
+            // slower than just incrementing `dst`.
+            // SAFETY: The offsets fits in `isize`, and because we were able to reserve the memory
+            // we know that `add` will not overflow.
+            unsafe { dst = dst.add(1) };
         });
-        // TODO(joe): replace with ptr_sub when stable
```
connortsui20, line 529: It was renamed to `offset_from_unsigned`

connortsui20:
Actually I'm realizing this is kind of stupid. We might as well just case on if we have an upper bound in the normal `extend_iter` implementation, that way we don't need to have an additional `extend_trusted` method

Edit: nvm Rust does not guarantee those hints are correct

a10y:
We need to be careful. We build on stable compiler, which means that we don't get `min_specialization` feature yet, so we need to call `extend_trusted` explicitly to save capacity checks in the hot loop of extending.

See https://github.com/vortex-data/vortex/pull/3881 for the PR that added our own TrustedLen and the `extend_trusted` method. That in turn is based off of a similar thing down in arrow-rs for their MutableBuffer type (IIRC).

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -221,6 +221,16 @@ impl BitBufferMut {
         unsafe { unset_bit_unchecked(self.buffer.as_mut_ptr(), self.offset + index) }
     }
 
+    /// Foces the length of the `BitBufferMut` to `new_len`.
+    ///
+    /// # Safety
+    ///
+    /// - `new_len` must be less than or equal to [`capacity()`](Self::capacity)
```
gatesn, line 228: I think you should still assert this, it's cheap enough, no one is calling this in a hot loop

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -241,18 +251,35 @@ impl BitBufferMut {
...
+    }
+
     /// Append a new true value to the buffer.
     pub fn append_true(&mut self) {
         let bit_pos = self.offset + self.len;
         let byte_pos = bit_pos / 8;
         let bit_in_byte = bit_pos % 8;
 
-        // Ensure buffer has enough bytes
+        // Ensure buffer has enough bytes.
```
gatesn, line 277: Lol, no

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -263,12 +290,61 @@ impl BitBufferMut {
...
         if byte_pos >= self.buffer.len() {
             self.buffer.push(0u8);
         }
 
-        // Bit is already 0 if we just pushed a new byte, otherwise ensure it's unset
+        // Bit is already 0 if we just pushed a new byte, otherwise ensure it's unset.
+        if bit_in_byte != 0 {
+            self.buffer.as_mut_slice()[byte_pos] &= !(1 << bit_in_byte);
```
gatesn, line 300: Use `crate::bit::set_bit` instead

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -263,12 +290,61 @@ impl BitBufferMut {
...
+        let bit_in_byte = bit_pos % 8;
+
+        // Ensure buffer has enough bytes.
+        if byte_pos >= self.buffer.len() {
+            // SAFETY: caller ensures sufficient capacity.
+            unsafe { self.buffer.push_unchecked(0u8) };
+        }
+
+        // Set the bit.
+        self.buffer.as_mut_slice()[byte_pos] |= 1 << bit_in_byte;
```
gatesn, line 325: Same here, use the set function

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -263,12 +290,61 @@ impl BitBufferMut {
...
+    ///
+    /// The caller must ensure there is sufficient capacity in the underlying byte buffer to
+    /// accommodate the new bit. If the bit position requires a new byte to be allocated, the buffer
+    /// must have capacity for at least one more byte.
+    pub unsafe fn append_true_unchecked(&mut self) {
+        let bit_pos = self.offset + self.len;
+        let byte_pos = bit_pos / 8;
+        let bit_in_byte = bit_pos % 8;
+
+        // Ensure buffer has enough bytes.
```
gatesn, line 318: I thought the caller asserted this already. So this statement should never be true, i.e. you can delete it.

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -263,12 +290,61 @@ impl BitBufferMut {
...
+    /// The caller must ensure there is sufficient capacity in the underlying byte buffer to
+    /// accommodate the new bit. If the bit position requires a new byte to be allocated, the buffer
+    /// must have capacity for at least one more byte.
+    pub unsafe fn append_false_unchecked(&mut self) {
+        let bit_pos = self.offset + self.len;
+        let byte_pos = bit_pos / 8;
+        let bit_in_byte = bit_pos % 8;
+
+        // Ensure buffer has enough bytes.
+        if byte_pos >= self.buffer.len() {
```
gatesn, line 342: Same here, delete it

vortex-buffer/src/buffer_mut.rs
```diff
@@ -477,59 +477,105 @@ impl<T> AsMut<[T]> for BufferMut<T> {
...
                 break;
-            }
+            };
+
+            // SAFETY: We have reserved enough capacity to hold this item, and `dst` is a pointer
+            // derived from a valid reference to byte data.
+            unsafe { dst.write(item) };
+
+            // Note: We used to have `dst.add(iteration).write(item)`, here. However this was much
+            // slower than just incrementing `dst`.
```
gatesn, line 508: Was it really slower? Still doing one addition per loop

gatesn (commented):
In general, your unsafe functions should have more debug_asserts. e.g. set_len

gatesn:
@a10y even with the trick we employ here? Basically reserve the lower bound of items, then consume the iterator up to the known capacity, after which, we loop over items using checked appends.

Seems like if the iterator was emitting a correct bound, then we'd never hit the slower loop portion? 

My understanding of the min-specialization uses is that e.g. extend(slice::iter) can be implemented using a memcpy, since we can specialize and know that the iterator is just backed by a slice.

vortex-buffer/src/bit/buf_mut.rs
connortsui20: What is the semantics of `with_capacity` though? Tracing through I can't figure out the answer to this question: are all bytes guaranteed to be zeroed? If not then I think we might need to do this

vortex-buffer/src/buffer_mut.rs
connortsui20: Not sure, that comment was there before.

robert3005: it was weirdly slower, somehow it was not optimised to the same thing

connortsui20:
Closing in favor of #5055 and #5056

PR #5051, opened 2025-10-23
fix[ci]: more cat-s3 retries and sleep a little between tries

joseph-isaacs:


AdamGS (approved):

scripts/cat-s3.sh
```diff
@@ -39,6 +39,8 @@ while (( n_failures < 2 )); do
     aws s3api put-object --bucket "$bucket" --key "$key" --body "$local_concatenated" --if-match "$current_etag" || {
         echo "ETag does not match during upload. Trying again."
         n_failures=$(( n_failures + 1 ))
+        # wait for before retrying
+        sleep 0.1
```
0ax1, line 43: Increasing wait time between retries, in an exponential backoff way, would be nice.

PR #5052, opened 2025-10-23
chore[ci]: benchmarks to take the median of runs not the min

joseph-isaacs:
I hope for this we reduce benchmark noise. Which is really bad at the moment

### DIFF
--- a/bench-vortex/src/measurements.rs
+++ b/bench-vortex/src/measurements.rs
@@ -203,19 +203,17 @@ pub struct QueryMeasurement {
 
 impl QueryMeasurement {
     pub fn median_run(&self) -> Duration {
-        let mut sorted_runs = self.runs.clone();
-        sorted_runs.sort();
-
-        let len = sorted_runs.len();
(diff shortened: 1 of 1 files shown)


AdamGS (approved):

bench-vortex/src/measurements.rs
```diff
@@ -202,8 +202,26 @@ pub struct QueryMeasurement {
...
+
+        let mut sorted_runs = self.runs.clone();
+        sorted_runs.sort();
+
+        if len % 2 == 1 {
+            sorted_runs[len / 2]
+        } else {
+            let mid1 = sorted_runs[len / 2 - 1];
+            let mid2 = sorted_runs[len / 2];
+            let avg_nanos = (mid1.as_micros() + mid2.as_nanos()) / 2;
```
lwwmanning, line 219: One of these is micros, the other nanos

AdamGS: I'll fix it

Issue #5044, opened 2025-10-23
Buffer::empty should not allocate


gatesn:
### Describe the bug

Currently Buffer::empty goes via BufferMut::empty, but this forces an allocation.

If we use Bytes::empty instead, then it refers to a static empty bytes slice. Alignment checks may fail.. which we'll need to figure out.

robert3005:
It does not anymore

PR #5055, opened 2025-10-23
Chore: Clean up `extend` impls for `BufferMut`

connortsui20:
_I'm moving this from #5049_ since things are becoming more complicated.

The only logical change here is at the top of `extend_iter`:

```rust
        // Since we do not know the length of the iterator, we can only guess how much memory we
        // need to reserve. Note that these hints may be inaccurate.
        let (lower_bound, upper_bound_opt) = iter.size_hint();
        self.reserve(upper_bound_opt.unwrap_or(lower_bound));
```

And the other is now using [`offset_from_unsigned`](https://github.com/rust-lang/rust/issues/95892) (used to be called `ptr_sub` on nightly)

### DIFF
--- a/vortex-buffer/src/buffer_mut.rs
+++ b/vortex-buffer/src/buffer_mut.rs
@@ -477,58 +477,108 @@ impl<T> AsMut<[T]> for BufferMut<T> {
 }
 
 impl<T> BufferMut<T> {
+    /// A helper method for the two [`Extend`] implementations.
+    ///
+    /// We use the lower bound hint on the iterator to manually write data, and then we continue to
+    /// push items normally past the lower bound.
     fn extend_iter(&mut self, mut iter: impl Iterator<Item = T>) {
-        // Attempt to reserve enough memory up-front, although this is only a lower bound.
-        let (lower, _) = iter.size_hint();
-        self.reserve(lower);
+        // Since we do not know the length of the iterator, we can only guess how much memory we
+        // need to reserve. Note that these hints may be inaccurate.
+        let (lower_bound, upper_bound_opt) = iter.size_hint();
+        self.reserve(upper_bound_opt.unwrap_or(lower_bound));
 
-        let remaining = self.capacity() - self.len();
+        let unwritten = self.capacity() - self.len();
 
+        // We store `begin` in the case that the lower bound hint is incorrect.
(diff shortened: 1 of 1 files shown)


vortex-buffer/src/buffer_mut.rs
```diff
@@ -477,59 +477,105 @@ impl<T> AsMut<[T]> for BufferMut<T> {
...
+    /// Internally, this calls [`extend_trusted()`](Self::extend_trusted).
+    pub fn from_trusted_len_iter<I>(iter: I) -> Self
+    where
+        I: TrustedLen<Item = T>,
+    {
+        // We allow the `extend_trusted` method to correctly allocate the required memory.
+        let mut buffer = Self::with_capacity(0);
+        buffer.extend_trusted(iter);
+
+        debug_assert_eq!(buffer.alignment(), Alignment::of::<T>());
```
gatesn: Not sure the point of this?

connortsui20: Not sure, I just copied and pasted it from the `from_iter` method just below

vortex-buffer/src/buffer_mut.rs
```diff
@@ -477,59 +477,105 @@ impl<T> AsMut<[T]> for BufferMut<T> {
...
+
+    /// Creates a `BufferMut` from an iterator with a trusted length.
+    ///
+    /// Internally, this calls [`extend_trusted()`](Self::extend_trusted).
+    pub fn from_trusted_len_iter<I>(iter: I) -> Self
+    where
+        I: TrustedLen<Item = T>,
+    {
+        // We allow the `extend_trusted` method to correctly allocate the required memory.
+        let mut buffer = Self::with_capacity(0);
```
gatesn: It's actually more efficient to pass the correct capacity here from the trusted iter, otherwise we do two allocations instead of one.

gatesn (approved):

vortex-buffer/src/buffer_mut.rs
connortsui20: I think this is related: https://github.com/vortex-data/vortex/issues/5044, putting this here for the link

robert3005 (approved):

PR #5056, opened 2025-10-23
Fix: incorrect `FromIterator<bool>` implementation

connortsui20:
It incorrectly assumed that the upper bound for iterators is true.

Also adds a `set_len` method.


connortsui20:
We get a small regression here (likely due to the extra branch) and it also regresses on my macbook, but AFAIK this is the cost of correctness here (and the benchmark is the idealized version where we know the exact length of the iterator anyways)

gatesn (approved):

robert3005 (approved):

PR #5060, opened 2025-10-23
doc: intro guide to vortex

### LINKED ISSUE #3996: Intro guide for Vortex
When doing https://github.com/vortex-data/vortex/pull/3906 I couldn't find a way in the documentation to put information about this method - even though debug printing feels like a very important tool for learning a new library

Moreover, as I was getting into the project, I couldn't find a simple cheat sheet on Vortex containing basic ops like:
- how to create a new array (for different types, for example that you want to use VarBin for string)
- how to print them 
- how to iterate them
- how to get individual element
- how to modify an array
- how to write it to the file
- and so on
- what's the difference from, e.g. `VarBinArray` from `VarBinView`

To be fair, some of those are covered in examples, but it's very slow to iterate over all of them checking if they have something you need. 

I think we should have this information as a single page in two ways: 
- pretty formatted and concise for humans, with some comments and explanation
-  in `vortex.dev/llm.txt` for LLMs. Example: `https://alloy.rs/llms-full.txt` and https://alloy.rs/introduction/prompting

ptravers:
closes: https://github.com/vortex-data/vortex/issues/3996

### DIFF
--- a/docs/guides/rust-cookbook.md
+++ b/docs/guides/rust-cookbook.md
@@ -1,0 +1,568 @@
+# Rust Cookbook
+
+This guide provides practical, copy-pasteable examples for common operations in Vortex. For more in-depth information, see the [Rust API documentation](https://docs.rs/vortex).
+
+## Topics Covered
+
+✓ **Creating arrays** - Different types including VarBinArray for strings ([Creating Arrays](#creating-arrays))
+✓ **Printing/debugging** - Display methods and inspection ([Inspecting and Debugging](#inspecting-and-debugging-arrays))
+✓ **Iterating** - Index-based and `with_iterator` patterns ([Iterating Over Arrays](#iterating-over-arrays))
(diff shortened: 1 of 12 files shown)


docs/guides/rust-cookbook.md
joseph-isaacs, line 1: The md rendering looks a little wrong here.

blaginin:
сс @connortsui20 as you worked on docs improvements recently too

docs/guides/rust-cookbook.md
ptravers: will take a look! thanks!

ptravers: ah sorry, I was using a feature not in vanilla markdown

robert3005:
Annoyingly the concepts have changed since this pr was created. We are more cognizant of writing documentation material as we update the codebase. Intro guide would still be useful but not sure what kind of structure it should take

PR #5061, opened 2025-10-24
chore[fuzz]: add `fill_null` baselinne

joseph-isaacs:


fuzz/src/array/fill_null.rs
```diff
@@ -1,16 +1,453 @@
...
+                        result
+                    }
+                }
+                _ => unreachable!(),
+            }
+        }
+    }
+}
+
+#[cfg(test)]
```
robert3005, line 240: I don't think these tests are ever run

robert3005 (approved):
I think if we want to add tests to the fuzzer we need to make sure they run, right now I don't think they do but I could be wrong

joseph-isaacs:
i'll follow up with that

PR #5065, opened 2025-10-24
perf: fuse alp decode with applying patches

0ax1:


joseph-isaacs:
I added some e2e decompression. https://github.com/vortex-data/vortex/pull/5067

Might be nice to see numbers go down

encodings/alp/src/alp/compress.rs
```diff
@@ -128,6 +144,56 @@ pub fn decompress(array: &ALPArray) -> PrimitiveArray {
     }
 }
 
+pub fn decompress_with_patches(
```
robert3005: I would really like this to somehow unify with existing logic. We already have patch method, ideally we can figure out how to fit into it

encodings/alp/src/alp/compress.rs
```diff
@@ -128,6 +144,56 @@ pub fn decompress(array: &ALPArray) -> PrimitiveArray {
...
+    match_each_alp_float_ptype!(array_ptype, |T| {
+        match_each_unsigned_integer_ptype!(patches_chunk_offsets.ptype(), |C| {
+            match_each_unsigned_integer_ptype!(patches_indices.ptype(), |I| {
+                let patches_indices = patches_indices.as_slice::<I>();
+                let patches_values = patches_values.as_slice::<T>();
+                let patches_chunk_offsets = patches_chunk_offsets.as_slice::<C>();
+
+                let alp_buffer = alp_encoded.into_buffer();
+                let array_len = array.len();
+                let mut decoded_values = BufferMut::<T>::with_capacity(array_len);
```
robert3005: I think even though it might be slower to decode in place we want the version that does it in place

0ax1: ~~Why is that? It's non-trivially slower.~~ It turned out we were decompressing twice for the inplace version, which was a setup issue in the benchmark.

encodings/alp/src/alp/compress.rs
0ax1: Iterated on this. I think this fn wasn't named well before either => `decompress_chunked`.  The main diff being that either the whole array is decompressed and patches applied after, or decoding and applying patches happening in 1024 chunks.

robert3005 (approved):

PR #5075, opened 2025-10-24
Chore: fix from_iter impls in vectors

connortsui20:
Didn't make it into the last one by accident

### DIFF
--- a/vortex-mask/src/lib.rs
+++ b/vortex-mask/src/lib.rs
@@ -20,7 +20,7 @@ use std::sync::{Arc, OnceLock};
 
 use itertools::Itertools;
 pub use mask_mut::*;
-use vortex_buffer::{BitBuffer, BitBufferMut};
+use vortex_buffer::{BitBuffer, BitBufferMut, set_bit_unchecked};
 use vortex_error::{VortexResult, vortex_panic};
 
 /// Represents a set of values that are all included, all excluded, or some mixture of both.
@@ -601,11 +601,13 @@ impl Mask {
                 let existing_buffer = mask_values.bit_buffer();
 
                 let mut new_buffer_builder = BitBufferMut::new_unset(mask_values.len());
+                debug_assert!(limit < mask_values.len());
 
+                let ptr = new_buffer_builder.as_mut_ptr();
                 for index in existing_buffer.set_indices().take(limit) {
-                    unsafe {
-                        new_buffer_builder.set_unchecked(index);
-                    }
+                    // SAFETY: We checked that `limit` was less than the mask values length,
+                    // therefore `index` must be within the bounds of the bit buffer.
+                    unsafe { set_bit_unchecked(ptr, index) }
                 }
 
                 Self::from(new_buffer_builder.freeze())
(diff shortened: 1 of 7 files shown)


vortex-vector/src/bool/from_iter.rs
```diff
@@ -26,28 +26,28 @@ impl FromIterator<Option<bool>> for BoolVectorMut {
         I: IntoIterator<Item = Option<bool>>,
     {
         let iter = iter.into_iter();
-        let (lower_bound, _) = iter.size_hint();
+        // Since we do not know the length of the iterator, we can only guess how much memory we
+        // need to reserve. Note that these hints may be inaccurate.
+        let (lower_bound, upper_bound_opt) = iter.size_hint();
+        let reserve_amount = upper_bound_opt.unwrap_or(lower_bound);
```
gatesn: I'm with Rob on this one, we should allocate lower bound not upper bound.

And fix all the others too.

Upper bound can be usize::MAX, in which case... it ain't pretty

connortsui20: I'm not super convinced that this is so terrible. For one, since we are reserving # of bits and not bytes the true upper bound is `usize::MAX / 8`. But even if it was `isize::MAX` I think it is still fine because the OS isn't actually going to allocate that amount. I forgot to put that comment in here like I did in https://github.com/vortex-data/vortex/pull/5055/files, let me do that...

connortsui20: Also note that iterators are perfectly allowed to set the lower bound to `usize::MAX`, which while probably rare is still possible.

The mental model I'm going with is that we should be thinking about what we want to allocate after a `.filter()` on an iterator. If I have an iterator of length 4096, then `.filter()` is going to return a `size_hint` of `(0, Some(4096))`.

To me, it feels wrong to not make use of that information and potentially trigger 10 reallocations just because of the adversarial case that might happen anyways if we only use the lower bound.

gatesn: I think we ought to just match the stdlib here. This is kind of the point of all of our TrustedLen stuff, which is when we _can_ make use of the upper bound

connortsui20:
So there is a huge regression on the microbenchmark when we switch to only using the lower bound. Not sure how much I trust microbenchmarks, but I do expect this given the amount of allocations we have to do for each of those sizes (up to 2^16 which means 15 allocations)

vortex-mask/src/lib.rs
```diff
@@ -601,11 +601,12 @@ impl Mask {
...
                 let mut new_buffer_builder = BitBufferMut::new_unset(mask_values.len());
+                debug_assert!(limit < mask_values.len());
 
                 for index in existing_buffer.set_indices().take(limit) {
-                    unsafe {
-                        new_buffer_builder.set_unchecked(index);
-                    }
+                    // SAFETY: We checked that `limit` was less than the mask values length,
+                    // therefore `index` must be within the bounds of the bit buffer.
+                    unsafe { set_bit_unchecked(new_buffer_builder.as_mut_ptr(), index) }
```
gatesn: You should pull out the mut_ptr and reuse its value in each loop

gatesn (approved):

PR #5076, opened 2025-10-24
chore[encodings]: use vx array assert eq macro

joseph-isaacs:


0ax1 (approved):
Nice, this is the better default.

PR #5078, opened 2025-10-24
Chore: rename `PVector` to `PVec`

connortsui20:
Since it will be quite similar to `Vec` from the standard library.


gatesn:
Everything in this crate is called Vector, I think we should be consistent with that rather than stdlib.

PR #5079, opened 2025-10-24
Feature: Add methods to `PVectorMut` similar to `std::vec::Vec`

connortsui20:

Adds methods to `PVec` that are similar in nature to the methods available on the standard library [`Vec`](https://doc.rust-lang.org/std/vec/struct.Vec.html#) implementation.

Note that none of these are strictly necessary right now, but I wanted to have this now so we can write more meaningful tests (instead of ones that just track metadata like length and capacity).

I want to add these methods in the future, but they should probably come in a separate PR:

- `into_iter`
- `retain` (that takes a `Fn(Option<&T>) -> bool` instead of `Fn(&T) -> bool`
- `retain_mut` (similar to above but `&mut T`
- `retain_indices` (we can use this to implement filter for primitive arrays)
- `extend_from_slice` (a nice to have because we know the exact length of the slice and don't have to do all the weird iterator size hint stuff)
- `copy_within`
- `fill_range`
- `drain` (maybe we don't need this but I can see it being useful)
- `splice` (same)
- `remove`
- `swap_remove`

There are others, but probably a lot less important


vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -162,6 +162,12 @@ impl BitBufferMut {
         self.buffer.reserve(additional_bytes);
     }
 
+    /// Clears the bit buffer (but keeps any allocated memory).
+    pub fn clear(&mut self) {
+        // Since there are no items we need to drop, we simply set the length to 0.
+        self.len = 0;
```
gatesn, line 171: You could also set the offset to zero?

vortex-vector/src/primitive/generic.rs
```diff
@@ -68,6 +68,76 @@ impl<T: NativePType> PVector<T> {
 
         Self { elements, validity }
     }
+
+    /// Gets a nullable element at the given index, with bounds checking.
+    ///
+    /// If the index is out of bounds, returns `None`. If the element at the given index is null,
+    /// returns `Some(None)`. Otherwise, returns `Some(Some(x))`, where `x: T`.
+    pub fn get_checked(&self, index: usize) -> Option<Option<T>> {
```
gatesn: Nah, kill this. The caller can implement bounds checking if they care

vortex-vector/src/primitive/generic.rs
```diff
@@ -68,6 +68,76 @@ impl<T: NativePType> PVector<T> {
...
+    pub fn get(&self, index: usize) -> Option<T> {
+        assert!(
+            index < self.len(),
+            "index out of bounds: the length is {} but the index is {index}",
+            self.len()
+        );
+
+        self.validity.value(index).then(|| {
+            self.elements
+                .get(index)
```
gatesn: Use `self.elements[index]` it performs the bounds checking for you

vortex-vector/src/primitive/generic.rs
```diff
@@ -68,6 +68,76 @@ impl<T: NativePType> PVector<T> {
...
+    /// Gets a nullable element at the given index.
+    ///
+    /// If the element at the given index is null, returns `None`. Otherwise, returns `Some(x)`,
+    /// where `x: T`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the index is out of bounds.
+    pub fn get(&self, index: usize) -> Option<T> {
+        assert!(
```
gatesn: No need, the code below will check bounds

vortex-vector/src/primitive/generic.rs
```diff
@@ -68,6 +68,76 @@ impl<T: NativePType> PVector<T> {
...
+    /// Gets a nullable element at the given index, without checking bounds and without checking
+    /// nullability.
+    ///
+    /// The caller should ensure that the element at the given index is not null (though doing so
+    /// will not cause undefined behavior).
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the index is within bounds.
+    pub unsafe fn get_unchecked(&self, index: usize) -> T {
```
gatesn: Do you even want this? We might just want `self.elements()` and allow the caller to do crazy stuff on the Buffer<T>

vortex-vector/src/primitive/generic.rs
```diff
@@ -68,6 +68,76 @@ impl<T: NativePType> PVector<T> {
...
+
+    /// Returns a slice over the internal buffer with elements of type `T`.
+    ///
+    /// Note that this slice may contain garbage data where the [`validity()`] mask states that an
+    /// element is invalid.
+    ///
+    /// The caller should check the [`validity()`] before performing any operations.
+    ///
+    /// [`validity()`]: Self::validity
+    pub fn as_slice(&self) -> &[T] {
```
gatesn: Inline always

vortex-vector/src/primitive/pvector_impl.rs
```diff
@@ -0,0 +1,509 @@
...
+    /// Gets a nullable element at the given index, without checking bounds and without checking
+    /// nullability.
+    ///
+    /// The caller should ensure that the element at the given index is not null (though doing so
+    /// will not cause undefined behavior).
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the index is within bounds.
+    pub unsafe fn get_unchecked(&self, index: usize) -> T {
```
gatesn: Same, kill this. Just provide `elements() -> &BufferMut<T>`

vortex-vector/src/primitive/pvector_impl.rs
```diff
@@ -0,0 +1,509 @@
...
+
+use crate::{PVectorMut, VectorMutOps};
+
+/// Point operations for [`PVectorMut`].
+impl<T: NativePType> PVectorMut<T> {
+    /// Returns the first element of the vector, or `None` if it is empty.
+    ///
+    /// If the first element is null, returns `Some(None)`. Otherwise, returns `Some(Some(x))`,
+    /// where `x: T`.
+    pub fn first(&self) -> Option<Option<T>> {
```
gatesn: Do we really need this?

vortex-vector/src/primitive/pvector_impl.rs
```diff
@@ -0,0 +1,509 @@
...
+        } else {
+            self.get_checked(self.len() - 1)
+        }
+    }
+
+    /// Gets a nullable element at the given index, with bounds checking.
+    ///
+    /// If the index is out of bounds, returns `None`. If the element at the given index is null,
+    /// returns `Some(None)`. Otherwise, returns `Some(Some(x))`, where `x: T`.
+    pub fn get_checked(&self, index: usize) -> Option<Option<T>> {
```
gatesn: Kill it

vortex-vector/src/primitive/pvector_impl.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    /// Returns a mutable slice over the internal mutable buffer with elements of type `T`.
+    ///
+    /// Note that this slice may contain garbage data where the [`validity()`] mask from the frozen
+    /// [`PVector`](crate::PVector) type states that an element is invalid.
+    ///
+    /// The caller should check the frozen [`validity()`] before performing any operations.
+    ///
+    /// [`validity()`]: crate::PVector::validity
+    pub fn as_mut_slice(&mut self) -> &mut [T] {
```
gatesn: Inline

vortex-vector/src/primitive/pvector_impl.rs
```diff
@@ -0,0 +1,509 @@
...
+impl<T: NativePType> PVectorMut<T> {
+    /// Returns an immutable slice over the internal mutable buffer with elements of type `T`.
+    ///
+    /// Note that this slice may contain garbage data where the [`validity()`] mask from the frozen
+    /// [`PVector`](crate::PVector) type states that an element is invalid.
+    ///
+    /// The caller should check the frozen [`validity()`] before performing any operations.
+    ///
+    /// [`validity()`]: crate::PVector::validity
+    pub fn as_slice(&self) -> &[T] {
```
gatesn: Inline

vortex-vector/src/primitive/pvector_impl.rs
gatesn: Tbh you may instead want `impl AsMut<[T]>` and `impl AsRef<[T]>` then you get nice `[]` syntax automatically

vortex-vector/src/primitive/pvector_impl.rs
```diff
@@ -0,0 +1,509 @@
...
+    /// Creates a new [`PVectorMut<T>`] from an iterator of `Option<T>` values.
+    ///
+    /// `None` values will be marked as invalid in the validity mask.
+    ///
+    /// Internally, this uses the [`Extend<Option<T>>`] trait implementation.
+    fn from_iter<I>(iter: I) -> Self
+    where
+        I: IntoIterator<Item = Option<T>>,
+    {
+        let mut vec = Self::with_capacity(0);
```
gatesn: `iter.size_hint().0` ?

vortex-vector/src/primitive/pvector_impl.rs
```diff
@@ -0,0 +1,509 @@
...
+    /// use vortex_vector::{PVectorMut, VectorMutOps};
+    ///
+    /// let mut vec = PVectorMut::from_iter([1i32, 2, 3, 4]);
+    /// assert_eq!(vec.len(), 4);
+    /// ```
+    fn from_iter<I>(iter: I) -> Self
+    where
+        I: IntoIterator<Item = T>,
+    {
+        let mut vec = Self::with_capacity(0);
```
gatesn: Same here, use size hint

gatesn (changes requested):

vortex-vector/src/primitive/pvector_impl.rs
connortsui20: It was just easy to have with `get_checked`, but since we're removing that I'll get rid of it

gatesn (approved):

PR #5080, opened 2025-10-25
fix: patch chunk offsets for alp encode

0ax1:
Patches get filtered after the `T::encode` call in case they are null. Therefore patches-chunk-offsets need to be adjusted as part of filtering. Subject to be cleaned up in a follow up.


0ax1:
Followed up by: https://github.com/vortex-data/vortex/pull/5065

robert3005 (approved):
I think we need to rethink the compression flow here. It’s pretty convoluted

PR #5088, opened 2025-10-27
Exact layout reader row count

gatesn:
This was an unused idea from a while back. Putting them back to exact.


robert3005 (approved):
Ah, I was wondering why this had a precision

Issue #5089, opened 2025-10-27
Async Expressions


gatesn:
This issue describes my current thinking around how to support async expressions in Vortex.

1. Introduce an AsyncExprLayoutReader.

This will partition the expression into an async node, its sync children, and its sync parent expression. Async expressions are determined as their ID resolving to a LayoutEncoding in the LayoutRegistry, vs resolving to an ArrayEncoding in the ArrayRegistry.

2. AsyncExprLayoutReader wraps its child layout in the layout loaded from the registry. The arrays returned from this layout are then wrapped in the parent sync expression.

We may wish to change the LayoutReader API to return streams, since some async expressions can blow up the in-memory size considerably and might want to return partial results.

robert3005:
I think the answer here is that the Evaluator can be async and expressions can ask for things to be resolved async

PR #5090, opened 2025-10-27
Move Array BatchKernel to sync API

gatesn:
* It's a little aggressive to require an async runtime simply to decompress in-memory data.
* Async expressions tend to require more powerful APIs anyway. For example, returning partial results since the output can be much larger than the input (not common for in-memory scalar compute). As such, I propose implementing async expressions as layout readers (with an improved API), see #5089


connortsui20 (approved):
🙏

PR #5092, opened 2025-10-27
Feature: Add conversion from vortex vectors to std `Vec`

connortsui20:
This is the last thing I need to write tests for everything more easily

Probably want this on the immutable vectors but I need this for testing first and then we can add those later


gatesn:
I don't think we want this

PR #5093, opened 2025-10-27
Feature: Add `IntoIter` implementation for vectors

connortsui20:

Adds `IntoIter` impls for primitive and bool vector.

Consolidates all the extend, from_iter, and (new) into_iter stuff into `iter.rs` files.

The non-cosmetic changes are at the bottom of the `iter.rs` files.

There is a world where we could use the internal `BufferIterator` types but this was way simpler.

### DIFF
--- a/vortex-vector/src/bool/iter.rs
+++ b/vortex-vector/src/bool/iter.rs
@@ -1,12 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-//! [`FromIterator`] and related implementations for [`BoolVectorMut`].
+//! Iterator implementations for [`BoolVectorMut`].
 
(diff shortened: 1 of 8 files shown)


vortex-vector/src/bool/iter.rs
```diff
@@ -78,3 +78,63 @@ impl FromIterator<bool> for BoolVectorMut {
...
     }
 }
+
+/// Iterator over a [`BoolVectorMut`] that yields [`Option<bool>`] values.
+///
+/// This iterator is created by calling [`IntoIterator::into_iter`] on a [`BoolVectorMut`].
+///
+/// It consumes the mutable vector and iterates over the elements, yielding `None` for null values
+/// and `Some(value)` for valid values.
+pub struct BoolVectorIntoIter {
```
gatesn: Nit: name this `BoolVectorIterator`

gatesn (approved):

PR #5094, opened 2025-10-27
Feature: Struct Vectors

connortsui20:

Adds `StructVector` and `StructVectorMut`.

Note: It would be really nice to have something like serde values that we can iterate and take as an iterator to construct these struct vectors, otherwise we are left with what I currently have on this PR, which imo is very unwieldy. This can come in a later PR though.

The second commit fixes some errors and also removes all of the `new` functions for vectors (as per an offline discussion with @gatesn).

### DIFF
--- a/vortex-array/src/arrays/bool/vtable/operator.rs
+++ b/vortex-array/src/arrays/bool/vtable/operator.rs
@@ -27,7 +27,7 @@ impl OperatorVTable<BoolVTable> for BoolVTable {
             // Note that validity already has the mask applied so we only need to apply it to bits.
             let bits = bits.filter(&mask);
 
-            Ok(BoolVector::new(bits, validity).into())
+            Ok(BoolVector::try_new(bits, validity)?.into())
         }))
     }
 }
--- a/vortex-array/src/arrays/primitive/vtable/operator.rs
+++ b/vortex-array/src/arrays/primitive/vtable/operator.rs
@@ -30,7 +30,7 @@ impl OperatorVTable<PrimitiveVTable> for PrimitiveVTable {
                 // the elements.
                 let elements = elements.filter(&mask);
 
-                Ok(PVector::new(elements, validity).into())
+                Ok(PVector::try_new(elements, validity)?.into())
             }))
         })
     }
--- a/vortex-compute/src/filter/bool.rs
+++ b/vortex-compute/src/filter/bool.rs
@@ -8,6 +8,12 @@ use crate::filter::Filter;
 
 impl Filter for BoolVector {
     fn filter(&self, mask: &Mask) -> Self {
-        Self::new(self.bits().filter(mask), self.validity().filter(mask))
+        let filtered_bits = self.bits().filter(mask);
+        let filtered_validity = self.validity().filter(mask);
+
+        // SAFETY: We filter the bits and validity with the same mask, and since they came from an
+        // existing and valid `BoolVector`, we know that the filtered output must have the same
+        // length.
+        unsafe { Self::new_unchecked(filtered_bits, filtered_validity) }
     }
 }
--- a/vortex-vector/src/bool/iter.rs
+++ b/vortex-vector/src/bool/iter.rs
@@ -85,6 +85,7 @@ impl FromIterator<bool> for BoolVectorMut {
 ///
 /// It consumes the mutable vector and iterates over the elements, yielding `None` for null values
 /// and `Some(value)` for valid values.
+#[derive(Debug)]
 pub struct BoolVectorMutIterator {
     /// The vector being iterated over.
     vector: BoolVectorMut,
--- a/vortex-vector/src/primitive/iter.rs
+++ b/vortex-vector/src/primitive/iter.rs
@@ -116,6 +116,7 @@ impl<T: NativePType> FromIterator<T> for PVectorMut<T> {
 ///
 /// It consumes the mutable vector and iterates over the elements, yielding `None` for null values
 /// and `Some(value)` for valid values.
+#[derive(Debug)]
 pub struct PVectorMutIterator<T: NativePType> {
     /// The vector being iterated over.
     vector: PVectorMut<T>,
--- a/vortex-vector/src/struct_/vector.rs
+++ b/vortex-vector/src/struct_/vector.rs
@@ -66,19 +66,7 @@ impl StructVector {
     /// - Any field vector has a length that does not match the length of other fields.
     /// - The validity mask length does not match the field length.
     pub fn try_new(fields: Arc<Box<[Vector]>>, validity: Mask) -> VortexResult<Self> {
-        let len = if fields.is_empty() {
-            validity.len()
-        } else {
-            fields[0].len()
-        };
-
-        // Validate that the validity mask has the correct length.
-        vortex_ensure!(
-            validity.len() == len,
-            "Validity mask length ({}) does not match expected length ({})",
-            validity.len(),
-            len
-        );
+        let len = validity.len();
 
         // Validate that all fields have the correct length.
         for (i, field) in fields.iter().enumerate() {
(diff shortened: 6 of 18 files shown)


vortex-vector/src/bool/iter.rs
```diff
@@ -85,6 +85,7 @@ impl FromIterator<bool> for BoolVectorMut {
 ///
 /// It consumes the mutable vector and iterates over the elements, yielding `None` for null values
 /// and `Some(value)` for valid values.
+#[derive(Debug)]
```
gatesn, line 88: Nah

vortex-vector/src/primitive/iter.rs
```diff
@@ -116,6 +116,7 @@ impl<T: NativePType> FromIterator<T> for PVectorMut<T> {
 ///
 /// It consumes the mutable vector and iterates over the elements, yielding `None` for null values
 /// and `Some(value)` for valid values.
+#[derive(Debug)]
```
gatesn, line 119: Also remove

vortex-vector/src/struct_/vector.rs
```diff
@@ -0,0 +1,199 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Definition and implementation of [`StructVector`].
+
+use vortex_error::{VortexExpect, VortexResult, vortex_ensure};
+use vortex_mask::Mask;
+
+use crate::{StructVectorMut, Vector, VectorMutOps, VectorOps};
+
+/// An immutable vector of boolean values.
```
gatesn: Comment is wrong

vortex-vector/src/struct_/vector.rs
```diff
@@ -0,0 +1,199 @@
...
+///
+/// `StructVector` can be considered a borrowed / frozen version of [`StructVectorMut`], which is
+/// created via the [`freeze`](crate::VectorMutOps::freeze) method.
+///
+/// See the documentation for [`StructVectorMut`] for more information.
+#[derive(Debug, Clone)]
+pub struct StructVector {
+    /// The fields of the `StructVector`, each stored column-wise as a [`Vector`].
+    pub(super) fields: Box<[Vector]>,
```
gatesn: I think we want an `Arc<[Vector]>`, just so cloning is cheaper?

vortex-vector/src/struct_/vector.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    /// The validity mask (where `true` represents an element is **not** null).
+    pub(super) validity: Mask,
+
+    /// The length of the vector (which is the same as all field vectors).
+    ///
+    /// This is stored here as a convenience, and also helps in the case that the `StructVector` has
+    /// no fields.
+    pub(super) len: usize,
```
gatesn, line 38: The length is actually stored in validity, so this isn't necessary

vortex-vector/src/struct_/vector.rs
```diff
@@ -0,0 +1,199 @@
...
+
+    /// Tries to create a new [`StructVector`] from the given fields and validity mask.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if:
+    ///
+    /// - Any field vector has a length that does not match the length of other fields.
+    /// - The validity mask length does not match the field length.
+    pub fn try_new(fields: Box<[Vector]>, validity: Mask) -> VortexResult<Self> {
```
gatesn: We don't do this anywhere else. For vectors specifically I think we should just continue to panic with assertions.

vortex-vector/src/struct_/vector.rs
```diff
@@ -0,0 +1,199 @@
...
+        Ok(())
+    }
+
+    /// Decomposes the struct vector into its constituent parts (fields, validity, and length).
+    pub fn into_parts(self) -> (Box<[Vector]>, Mask, usize) {
+        (self.fields, self.validity, self.len)
+    }
+
+    /// Returns the fields of the `StructVector`, each stored column-wise as a [`Vector`].
+    pub fn fields(&self) -> &[Vector] {
```
gatesn, line 120: This should then return `&Arc<[Vector]>` to allow cheap clones

vortex-compute/src/filter/bool.rs
```diff
@@ -1,13 +1,16 @@
...
+use vortex_error::VortexExpect;
 use vortex_mask::Mask;
 use vortex_vector::{BoolVector, VectorOps};
 
 use crate::filter::Filter;
 
 impl Filter for BoolVector {
     fn filter(&self, mask: &Mask) -> Self {
-        Self::new(self.bits().filter(mask), self.validity().filter(mask))
+        Self::try_new(self.bits().filter(mask), self.validity().filter(mask)).vortex_expect(
```
gatesn: new_unchecked

vortex-array/src/arrays/primitive/vtable/operator.rs
```diff
@@ -30,7 +30,11 @@ impl OperatorVTable<PrimitiveVTable> for PrimitiveVTable {
                 // the elements.
                 let elements = elements.filter(&mask);
 
-                Ok(PVector::new(elements, validity).into())
+                Ok(PVector::try_new(elements, validity)
```
gatesn: Raise the result I think

vortex-array/src/arrays/bool/vtable/operator.rs
```diff
@@ -27,7 +27,11 @@ impl OperatorVTable<BoolVTable> for BoolVTable {
             // Note that validity already has the mask applied so we only need to apply it to bits.
             let bits = bits.filter(&mask);
 
-            Ok(BoolVector::new(bits, validity).into())
+            Ok(BoolVector::try_new(bits, validity)
```
gatesn: Raise the result

gatesn (commented):
Ok fine, let's keep `new()`, but don't be specific in the vortex_except message, just say "failed to create vector" or similar, the backtrace from the result will contain the actual message.

vortex-vector/src/bool/iter.rs
connortsui20: These are nice to have for debugging. See https://doc.rust-lang.org/std/slice/struct.Iter.html for reference, we should technically also have `Default` and `Clone` (and the other iterator traits) which I can add for completeness.

vortex-vector/src/struct_/vector.rs
connortsui20: The allocation won't be reused from the `Vec` in `StructVectorMut` if we convert to `Arc`.

Also, I think we _do_ want clone to be a deep copy? This means that if you want to reuse the underlying bytes, you have to clone the `Vector`, not just `Arc` the vector and clone that.

I want to avoid doing the double reference-count thing that Arrow is already doing, so I've been avoiding `Arc` completely. I think it makes it even harder to reason about what allocations are happening.

vortex-vector/src/struct_/vector.rs
connortsui20: I think this is nice for a convenience. And also while this isn't exactly the same problem, storing length in the validity is what caused Arrow's `FixedSizeList` to be wrong, so I feel we might as well be explicit to be safe.

vortex-vector/src/struct_/vector.rs
connortsui20: See above

vortex-vector/src/primitive/iter.rs
connortsui20: See above

vortex-compute/src/filter/bool.rs
```diff
@@ -8,6 +8,14 @@ use crate::filter::Filter;
 
 impl Filter for BoolVector {
     fn filter(&self, mask: &Mask) -> Self {
-        Self::new(self.bits().filter(mask), self.validity().filter(mask))
+        let filtered_bits = self.bits().filter(mask);
+        let filtered_validity = self.validity().filter(mask);
+
+        debug_assert_eq!(filtered_bits.len(), filtered_validity.len());
```
gatesn: This happens inside new_unchecked anyway?

vortex-vector/src/struct_/vector.rs
connortsui20: Offline: We will use `Box<[VectorMut]>` for mutable struct and `Arc<[Vector]>` for immutable

gatesn (approved):

PR #5095, opened 2025-10-27
chore: deny let_underscore_drop

danking:
I'm curious for opinions on this.

I stubbed my toe recently on:

```
let _ = span.enter();
```

Which does not enter the span for the remainder of the function (it immediately drops the span). There is already a by-default-deny lint for `let _ = lock()` but no such lint for other guard-like things:

- [`tracing::Span::enter()`](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.enter)
- [`xshell::Shell::push_dir`](https://docs.rs/xshell/latest/xshell/struct.Shell.html#method.push_dir)

### DIFF
--- a/vortex-datafusion/src/persistent/sink.rs
+++ b/vortex-datafusion/src/persistent/sink.rs
@@ -392,7 +392,7 @@ mod tests {
         register_vortex_format_factory(factory, &mut session_state_builder);
         let session = SessionContext::new_with_state(session_state_builder.build());
 
-        let _ = session
+        let _unused = session
             .sql(&format!(
                 "CREATE EXTERNAL TABLE my_tbl \
                 (c1 VARCHAR NOT NULL, c2 INT NOT NULL) \
--- a/vortex-duckdb/build.rs
+++ b/vortex-duckdb/build.rs
@@ -44,7 +44,7 @@ fn download_duckdb_lib_archive() -> Result<PathBuf, Box<dyn std::error::Error>> 
     let archive_path = duckdb_dir.join(&archive_name);
 
     // Recreate the duckdb directory
-    let _ = fs::remove_dir_all(&duckdb_dir);
+    drop(fs::remove_dir_all(&duckdb_dir));
     fs::create_dir_all(&duckdb_dir)?;
 
     if !archive_path.exists() {
@@ -192,7 +192,7 @@ fn build_duckdb(duckdb_source_root: &Path) -> Result<PathBuf, Box<dyn std::error
     let target_dir = manifest_dir.parent().unwrap().join("target");
     let duckdb_library_dir = target_dir.join("duckdb-lib");
 
-    let _ = fs::remove_dir_all(&duckdb_library_dir);
+    drop(fs::remove_dir_all(&duckdb_library_dir));
     fs::create_dir_all(&duckdb_library_dir)?;
 
     // Copy .dylib and .so files (macOS and Linux).
@@ -223,7 +223,7 @@ fn main() {
     // Download, extract and symlink DuckDB source code.
     let zip_source_path = download_duckdb_source_archive().unwrap();
     let extracted_source_path = extract_duckdb_source(zip_source_path).unwrap();
-    let _ = fs::remove_dir_all(&duckdb_repo);
+    drop(fs::remove_dir_all(&duckdb_repo));
     std::os::unix::fs::symlink(&extracted_source_path, &duckdb_repo).unwrap();
 
     let library_path =
--- a/vortex-duckdb/build.rs
+++ b/vortex-duckdb/build.rs
@@ -3,6 +3,7 @@
 
 #![allow(clippy::unwrap_used)]
 #![allow(clippy::expect_used)]
+use std::io::ErrorKind;
 use std::path::{Path, PathBuf};
 use std::{env, fs};
 
@@ -44,7 +45,10 @@ fn download_duckdb_lib_archive() -> Result<PathBuf, Box<dyn std::error::Error>> 
     let archive_path = duckdb_dir.join(&archive_name);
 
     // Recreate the duckdb directory
-    drop(fs::remove_dir_all(&duckdb_dir));
+    match fs::remove_dir_all(&duckdb_dir) {
+        Err(err) if err.kind() == ErrorKind::NotFound => (),
+        otherwise => otherwise?,
+    }
     fs::create_dir_all(&duckdb_dir)?;
 
     if !archive_path.exists() {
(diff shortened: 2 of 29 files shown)


vortex-array/src/compute/conformance/cast.rs
```diff
@@ -111,7 +111,7 @@ fn test_cast_from_bool(array: &dyn Array, nullability: Nullability) {
     test_cast_nullability_changes(array, &DType::Bool(Nullability::Nullable));
     if nullability == Nullability::Nullable {
         // Try casting to non-nullable (may fail if nulls present)
-        let _ = cast(array, &DType::Bool(Nullability::NonNullable));
+        cast(array, &DType::Bool(Nullability::NonNullable)).vortex_unwrap();
```
danking: This file had several tests that did not actually run their tests because they didn't try to unwrap the result.

vortex-datafusion/src/persistent/sink.rs
```diff
@@ -385,7 +385,7 @@ mod tests {
         register_vortex_format_factory(factory, &mut session_state_builder);
         let session = SessionContext::new_with_state(session_state_builder.build());
 
-        let _ = session
+        let _unused = session
```
danking, line 395: This is a bit suspicious to me? Shouldn't we fail if the external table fails to create?

vortex-expr/src/exprs/cast.rs
```diff
@@ -156,7 +157,7 @@ mod tests {
     #[test]
     fn replace_children() {
         let expr = cast(root(), DType::Bool(Nullability::Nullable));
-        let _ = expr.with_children(vec![root()]);
+        expr.with_children(vec![root()]).vortex_unwrap();
```
danking, line 159: This did not test what it claimed to test.

vortex-expr/src/exprs/is_null.rs
```diff
@@ -151,7 +152,7 @@ mod tests {
     #[test]
     fn replace_children() {
         let expr = is_null(root());
-        let _ = expr.with_children(vec![root()]);
+        expr.with_children(vec![root()]).vortex_unwrap();
```
danking: This did not test what it claimed to test.

vortex-io/src/limit.rs
```diff
@@ -259,7 +259,7 @@ mod tests {
         assert!(result.is_err());
 
         // After consuming, should be able to push again
-        let _ = stream.next().await;
+        assert!(stream.next().await.is_some());
```
danking: This seems like a reasonable assertion to me. The stream ought to have that one element we put into it!

vortex-mask/src/bitops.rs
```diff
@@ -181,7 +181,7 @@ mod tests {
     fn test_bitand_different_lengths() {
         let mask1 = Mask::new_true(5);
         let mask2 = Mask::new_true(3);
-        let _ = &mask1 & &mask2;
+        let _unused = &mask1 & &mask2;
```
danking, line 184: ~~bitand has a lint which prevents me from completely ignoring the result a la `&mask1 & &mask2`.~~

Masks have non-trivial drops (I think, an Arc).

vortex-datafusion/src/persistent/sink.rs
a10y: isn't that handled by the `.await?` ?

vortex-io/src/limit.rs
a10y: can just do .unwrap() in that case?

vortex-datafusion/src/persistent/sink.rs
danking: Ah, yes, you're right. This is flagged because DataFrame has a non-trivial Drop (probably drops some Arcs).

vortex-io/src/limit.rs
danking: Done.

danking:
Blocked on fixing our conformance tests: https://github.com/vortex-data/vortex/pull/5101

a10y:
the only place I really have used `let _ = ..` in production code is in FFI implementations, usually when you do `let _ = Box::from_raw(..)` to allow the drop handler to run. But I guess that the same can be achieved by just explicitly calling `drop(Box::from_raw(..))`. And pretty much anywhere else we use `let _ = ..` and actually mean it?

vortex-duckdb/build.rs
```diff
@@ -223,7 +223,7 @@ fn main() {
     // Download, extract and symlink DuckDB source code.
     let zip_source_path = download_duckdb_source_archive().unwrap();
     let extracted_source_path = extract_duckdb_source(zip_source_path).unwrap();
-    let _ = fs::remove_dir_all(&duckdb_repo);
+    drop(fs::remove_dir_all(&duckdb_repo));
```
a10y, line 233: unwrap the result

vortex-duckdb/build.rs
```diff
@@ -192,7 +192,7 @@ fn build_duckdb(duckdb_source_root: &Path) -> Result<PathBuf, Box<dyn std::error
     let target_dir = manifest_dir.parent().unwrap().join("target");
     let duckdb_library_dir = target_dir.join("duckdb-lib");
 
-    let _ = fs::remove_dir_all(&duckdb_library_dir);
+    drop(fs::remove_dir_all(&duckdb_library_dir));
```
a10y: unwrap the result here too

vortex-io/src/file/read/mod.rs
```diff
@@ -124,7 +124,7 @@ impl Drop for ReadFuture {
     fn drop(&mut self) {
         // When the FileHandle is dropped, we can send a shutdown event to the I/O stream.
         // If the I/O stream has already been dropped, this will fail silently.
-        let _ = self.events.unbounded_send(ReadEvent::Dropped(self.id));
+        drop(self.events.unbounded_send(ReadEvent::Dropped(self.id)));
```
a10y, line 127: this should also unwrap the result  i think?

a10y: oh i see, we don't want to panic in the drop impl here.

vortex-duckdb/build.rs
danking: I went with the following because I think the intention is to not fail if the directory already exists (but I agree we should fail if there's, e.g. a permissions issue or some other weird problem).
```
    match fs::remove_dir_all(&duckdb_dir) {
        Err(err) if err.kind() == ErrorKind::NotFound => (),
        otherwise => otherwise?,
    }
```

danking:
> the only place I really have used `let _ = ..` in production code is in FFI implementations, usually when you do `let _ = Box::from_raw(..)` to allow the drop handler to run. But I guess that the same can be achieved by just explicitly calling `drop(Box::from_raw(..))`. And pretty much anywhere else we use `let _ = ..` and actually mean it?

Yeah this is my argument. `let _ = ...` is surprisingly different from `let _x = ...`. The surprisingly different behavior is exactly `drop(...)`, so we should just say that directly.

a10y (approved):

PR #5096, opened 2025-10-27
Add checked arithmetic compute

gatesn:
FLUPs:

* Arithmetic array to invert lhs/rhs for constant checks
* Special handling for vectors with many nulls (iterate only the sparse non-null indices)


vortex-array/src/array/operator.rs
```diff
@@ -33,6 +34,21 @@ pub trait ArrayOperator: 'static + Send + Sync {
...
+                    selection.dtype()
+                );
+            }
+            if selection.len() != self.len() {
+                vortex_bail!(
+                    "Selection array length {} does not match array length {}",
+                    selection.len(),
+                    self.len()
+                );
+            }
```
connortsui20, line 50: `vortex_ensure`?

vortex-array/src/compute/arrays/arithmetic.rs
```diff
@@ -0,0 +1,375 @@
...
+    pub fn new(lhs: ArrayRef, rhs: ArrayRef, operator: ArithmeticOperator) -> Self {
+        assert_eq!(
+            lhs.len(),
+            rhs.len(),
+            "Arithmetic arrays require lhs and rhs to have the same length"
+        );
+
+        // TODO(ngates): should we automatically cast non-null to nullable if required?
+        assert!(matches!(lhs.dtype(), DType::Primitive(..)));
+        assert_eq!(lhs.dtype(), rhs.dtype());
```
connortsui20, line 63: This is just a question: how do we handle add where values can be null currently? Do we even need to worry about this if we are just taking the `OR` of the masks?

vortex-array/src/compute/arrays/arithmetic.rs
```diff
@@ -0,0 +1,375 @@
...
+    // optimizations.
+    operator: ArithmeticOperator,
+}
+
+#[allow(clippy::mem_forget)]
+static ENCODINGS: LazyLock<EnumMap<ArithmeticOperator, EncodingRef>> = LazyLock::new(|| {
+    enum_map! {
+        operator => ArithmeticEncoding { operator }.to_encoding(),
+    }
+});
```
connortsui20, line 92: I'm not sure I understand the purpose of this? Could you instead just implement `From<ArithmeticOperator> for ArithmeticEncoding`?

vortex-array/src/compute/arrays/arithmetic.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+    fn stats(array: &ArithmeticArray) -> StatsSetRef<'_> {
+        array.stats.to_ref(array.as_ref())
+    }
+
+    fn array_hash<H: Hasher>(array: &ArithmeticArray, state: &mut H, precision: Precision) {
+        array.lhs.array_hash(state, precision);
+        array.rhs.array_hash(state, precision);
+    }
```
connortsui20, line 137: Also hash the encoding since it contains which operation?

vortex-array/src/compute/arrays/arithmetic.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        Ok(None)
+    }
+
+    fn bind(
+        array: &ArithmeticArray,
+        selection: Option<&ArrayRef>,
+        ctx: &mut dyn BindCtx,
+    ) -> VortexResult<BatchKernelRef> {
+        // Optimize for constant RHS
```
connortsui20, line 225: Is there a difference between RHS and LHS being constant?

vortex-array/src/compute/arrays/arithmetic.rs
```diff
@@ -0,0 +1,375 @@
...
+                    .typed_value::<T>()
+                    .vortex_expect("Already checked for null above");
+                Ok(match array.operator() {
+                    ArithmeticOperator::Add => arithmetic_scalar_kernel::<Add, _>(lhs, rhs_value),
+                    ArithmeticOperator::Sub => arithmetic_scalar_kernel::<Sub, _>(lhs, rhs_value),
+                    ArithmeticOperator::Mul => arithmetic_scalar_kernel::<Mul, _>(lhs, rhs_value),
+                    ArithmeticOperator::Div => arithmetic_scalar_kernel::<Div, _>(lhs, rhs_value),
+                })
+            });
+        }
```
connortsui20: Do we always run `reduce_children` before `bind` (or vice versa)? I'm still not super familiar with the details, but it seems strange that we would need to do optimizations both here in `bind` and also `reduce_children`? Or if we do need to do this in both places maybe this should be modularized?

vortex-array/src/compute/mod.rs
```diff
@@ -43,7 +43,7 @@ use crate::{Array, ArrayRef};
 
 #[cfg(feature = "arbitrary")]
 mod arbitrary;
-mod arrays;
+pub mod arrays;
```
connortsui20, line 46: Do we want `vortex::compute::arrays::ArithmeticArray` or `vortex::compute::ArithmeticArray` to mimic `vortex::arrays::PrimitiveArray`?

vortex-array/src/compute/arrays/arithmetic.rs
```diff
@@ -0,0 +1,375 @@
...
+#[derive(Debug, Clone)]
+pub struct ArithmeticArray {
+    encoding: EncodingRef,
+    lhs: ArrayRef,
+    rhs: ArrayRef,
+    stats: ArrayStats,
+}
+
+impl ArithmeticArray {
+    /// Create a new logical array.
```
connortsui20: Create a new arithmetic array?

vortex-compute/src/arithmetic/buffer.rs
```diff
@@ -0,0 +1,258 @@
...
+    T: Copy + num_traits::Zero,
+    BufferMut<T>: for<'a> Checked<Op, &'a Buffer<T>, Output = Buffer<T>>,
+    for<'a> &'a Buffer<T>: Checked<Op, &'a Buffer<T>, Output = Buffer<T>>,
+{
+    type Output = Buffer<T>;
+
+    fn checked_op(self, rhs: &Buffer<T>) -> Option<Self::Output> {
+        match self.try_into_mut() {
+            Ok(lhs) => lhs.checked_op(rhs),
+            Err(lhs) => (&lhs).checked_op(rhs),
```
connortsui20: Probably good to explain why we need this to be `(&lhs)` and not `lhs` to prevent infinite recursion.

But that does lead me to think about why we need to implement `Checked for<'a>`? Could we get away with just saying that `rhs` can always be a reference? See a few comments below.

vortex-compute/src/arithmetic/buffer.rs
```diff
@@ -0,0 +1,258 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_buffer::{Buffer, BufferMut};
+
+use crate::arithmetic::{Checked, CheckedOperator};
+
+/// Implementation that attempts to downcast to a mutable buffer and operates in-place.
+impl<Op, T> Checked<Op, &Buffer<T>> for Buffer<T>
+where
+    T: Copy + num_traits::Zero,
```
connortsui20: I don't think I would be opposed to just slapping `T: NativePType` here, I think we **do** want all of the native types to have checked arithmetic

vortex-compute/src/arithmetic/buffer.rs
```diff
@@ -0,0 +1,258 @@
...
+    fn checked_op(self, rhs: &Buffer<T>) -> Option<Self::Output> {
+        assert_eq!(self.len(), rhs.len());
+
+        let mut i = 0;
+        let mut overflow = false;
+        let buffer = self
+            .map_each(|a| {
+                // SAFETY: lengths are equal, so index is in bounds
+                let b = unsafe { *rhs.get_unchecked(i) };
+                i += 1;
```
connortsui20: We probably want an `enumerate` type thing here?

vortex-compute/src/arithmetic/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+mod pvector;
+
+/// Trait for checked arithmetic operations.
+///
+/// It may be easier to use the specific traits like [`CheckedAdd`], [`CheckedSub`], etc.
+pub trait Checked<Op, Rhs = Self> {
+    /// The result type after performing the operation.
+    type Output;
+
+    /// Perform the operation.
```
connortsui20, line 18: Can you add a comment saying what `None` means? I think it means that the operation overflowed / divided by zero but that wasn't clear until I read the code directly

vortex-compute/src/arithmetic/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    fn checked_op(self, rhs: Rhs) -> Option<Self::Output>;
```
connortsui20: Do we really need `rhs` to be owned? I think it might be creating more complexity than we need: it is always very easy to add a reference in front of something that is already owned.

vortex-compute/src/arithmetic/buffer.rs
```diff
@@ -0,0 +1,258 @@
...
+{
+    type Output = Buffer<T>;
+
+    fn checked_op(self, rhs: &Buffer<T>) -> Option<Self::Output> {
+        assert_eq!(self.len(), rhs.len());
+
+        let mut i = 0;
+        let mut overflow = false;
+        let buffer = self
+            .map_each(|a| {
```
connortsui20: I think `map_each_in_place` would be a better name, I got very confused by this for a few minutes

vortex-compute/src/arithmetic/buffer.rs
```diff
@@ -0,0 +1,258 @@
...
+    T: Copy + num_traits::Zero,
+    BufferMut<T>: for<'a> Checked<Op, &'a T, Output = Buffer<T>>,
+    for<'a> &'a Buffer<T>: Checked<Op, &'a T, Output = Buffer<T>>,
+{
+    type Output = Buffer<T>;
+
+    fn checked_op(self, rhs: &T) -> Option<Self::Output> {
+        match self.try_into_mut() {
+            Ok(lhs) => lhs.checked_op(rhs),
+            Err(lhs) => (&lhs).checked_op(rhs),
```
connortsui20: A comment here as well would be nice

vortex-compute/src/arithmetic/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+        impl CheckedOperator<$T> for Mul {
+            fn apply(lhs: &$T, rhs: &$T) -> Option<$T> {
+                Some(lhs * rhs)
+            }
+        }
+        impl CheckedOperator<$T> for Div {
+            fn apply(lhs: &$T, rhs: &$T) -> Option<$T> {
+                Some(lhs / rhs)
```
connortsui20: Divide by zero check?

connortsui20 (changes requested):

vortex-compute/src/arithmetic/buffer.rs
connortsui20, line 1: It took me a while to grok this file, I think I would appreciate more detailed doc comments?

vortex-compute/src/arithmetic/mod.rs
gatesn: It's more general, with no extra work? The whole reason we can't use num_traits::Checked operators is because they force taking the arguments by reference.

connortsui20: Answer: no this is just more flexible and we still need all those trait bounds

vortex-compute/src/arithmetic/mod.rs
connortsui20: Nevermind, float divide by 0 returns infinity which is valid

connortsui20 (approved):
LGTM

PR #5099, opened 2025-10-28
Fix: revert `List::take` compute function

connortsui20:
Related: https://spiraldb.slack.com/archives/C07BV3GKAJ2/p1761219012611469

Reverts the `List` `take` compute function to again return a `ListArray` instead of converting to `ListView` and relying on the other `take` compute function.

We will probably keep this as is until the performance of `ListView` improves / more people adopt `ListView` (Arrow and DF).

### DIFF
--- a/vortex-array/src/arrays/list/compute/take.rs
+++ b/vortex-array/src/arrays/list/compute/take.rs
@@ -2,31 +2,41 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use vortex_buffer::BitBufferMut;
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

connortsui20:
Seems like the FineWeb benchmark is actually slower with the old List take vs the already slow and naive ListView take...

danking (approved):
Parquet is also 20% slower on FineWeb S3. I don't believe that Vortex is meaningfully slower after this change.

PR #5100, opened 2025-10-28
Vector comparison compute

gatesn:


vortex-compute/src/comparison/bool.rs
```diff
@@ -0,0 +1,70 @@
...
+{
+    type Output = BoolVector;
+
+    fn compare(self, rhs: Self) -> Self::Output {
+        let validity = self.validity().bitand(rhs.validity());
+
+        let lhs = self.bits().chunks();
+        let rhs = rhs.bits().chunks();
+
+        let mut buffer = BufferMut::<u64>::with_capacity(lhs.chunk_len() + 1);
```
connortsui20, line 26: A comment about the +1 needed for the padding would be nice, tripped me up

vortex-compute/src/comparison/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+pub struct LessThan;
+/// A marker type for less-than-or-equal comparison operations.
+pub struct LessThanOrEqual;
+/// A marker type for greater-than comparison operations.
+pub struct GreaterThan;
+/// A marker type for greater-than-or-equal comparison operations.
+pub struct GreaterThanOrEqual;
+
+/// Marker trait for comparable items.
+pub trait ComparableItem {
```
connortsui20, line 41: Would making a method that returns an [`Ordering`](https://doc.rust-lang.org/std/cmp/enum.Ordering.html) make things cleaner? Not sure

vortex-vector/src/primitive/generic.rs
```diff
@@ -82,8 +82,8 @@ impl<T: NativePType> PVector<T> {
     /// # Panics
     ///
     /// Panics if the index is out of bounds.
-    pub fn get(&self, index: usize) -> Option<T> {
-        self.validity.value(index).then(|| self.elements[index])
+    pub fn get(&self, index: usize) -> Option<&T> {
+        self.validity.value(index).then(|| &self.elements[index])
```
connortsui20, line 83: Did we need this to be a reference somewhere?

vortex-compute/src/comparison/collection.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Marker trait for comparable collections.
+pub trait ComparableCollection {
+    /// The item type that can be compared.
+    type Item;
+
+    /// Get the length of the comparable collection.
+    fn len(&self) -> usize;
+
+    /// Get the item at the specified index without bounds checking.
+    unsafe fn item_unchecked(&self, index: usize) -> Self::Item;
```
connortsui20, line 38: Missing `# Safety` section

vortex-compute/src/comparison/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Check if two items are equal.
+    fn is_equal(lhs: &Self, rhs: &Self) -> bool;
+
+    /// Check if the `lhs` item is less than the `rhs` item.
+    fn is_less_than(lhs: &Self, rhs: &Self) -> bool;
```
connortsui20, line 46: Might be worth documenting that this is a total comparison (from the float implementation below)

vortex-compute/src/comparison/collection.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl<T: Copy> ComparableCollection for &[T] {
+    type Item = T;
+
+    fn len(&self) -> usize {
+        <[T]>::len(self)
+    }
+
+    unsafe fn item_unchecked(&self, index: usize) -> Self::Item {
+        unsafe { *self.get_unchecked(index) }
```
connortsui20, line 49: Missing `// SAFETY:` comment

vortex-compute/src/comparison/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+use vortex_dtype::half::f16;
+
+mod bool;
+mod collection;
+mod pvector;
+
+/// Trait for comparison operations.
+pub trait Compare<Op, Rhs = Self> {
+    /// The result type after performing the operation.
+    type Output;
```
connortsui20, line 15: Is this every not just a `BitBuffer` (or `BoolVector`, which is effectively the same thing)? The only alternative I can think of is this being some sort of scalar type, but that means we just compared 2 scalars and those should be comparable without having to implement this `Compare` trait that is local to this crate.

connortsui20 (changes requested):
Everything generally makes sense! I would also appreciate more comments but I know we are moving fast.

vortex-compute/src/comparison/mod.rs
gatesn: It's more expensive than returning bools, particularly when we start implementing this for VarBin BinaryView etc

vortex-vector/src/primitive/generic.rs
gatesn: It means we can relax the copy constraint. I imagine this "Generic" vector is renamed `SizedVector` when we come to add support for decimals and then we won't want to tie ourselves to NativePType

vortex-compute/src/comparison/collection.rs
gatesn: Unsafe is being propagated though?

vortex-compute/src/comparison/mod.rs
gatesn: Kind of implied that it doesn't return Option<bool> I think?

connortsui20 (approved):
LGTM

PR #5106, opened 2025-10-28
feat: ListLayout

a10y:
This PR implements a new layout for `List` typed data.

## Background

List columns are written like all other columns, as FlatLayouts. This means that offsets, elements and validity are written together into a single segment. This presents challenges when you have a deeply nested schema with multiple layers of lists/structs.

This is what a schema with list columns looks like:

image

## Solution

This PR adds a new ListLayout, which splits offsets, elements, and validity for the list into separate child layouts. Here's what Lists will look like now:

image

This increases the # of IOs for randomly reading a single list from 1 - > 2 (or 3, if validity). But, it has the added benefit that Lists with something like struct elements can greatly reduce the amount of IO needed when you're only loading a subset of the columns.

Notes:

* Data is always written as List, not ListView, as List is almost always more compact. In the future if we supported dictionary compression of ListViews, we could add a Layout for that.

## Open problems

* In the reader, we want to register splits at row boundaries. In our case, we want to read splits that roughly correspond to the elements child. Right now, we don't have a good way to do that. So I just use a static split size in the ListReader. I don't really like this but didn't have better ideas.
* Ideally, elements of the same list would never be split into separate segments, but we don't guarantee that, and in practice it's not something we can really express since we just let you set any `Arc<dyn LayoutStrategy>` for the elements child
* We don't implement any special pruning or filtering pushdown support. Mostly, because we don't have any expressions that operate over Lists. It's actually not that common for this to exist, so we need to implement some more of these things instead.
* I implemented two different read paths:
  * Dense: when the project_eval mask is dense (>50%), this loads the full offsets/elements for the split in parallel, then applies the mask afterward
  * Sparse: when the mask is sparse, we try and read as little of `elements` as possible. We first read all of the offsets for the split, intersect with the mask, and then build a new mask for reading the elements.

### DIFF
--- a/vortex-layout/src/layouts/list/reader/list.rs
+++ b/vortex-layout/src/layouts/list/reader/list.rs
@@ -7,20 +7,22 @@ use std::collections::BTreeSet;
 use std::ops::Range;
 use std::sync::Arc;
 
-use crate::ArrayFuture;
-use crate::LayoutReader;
-use crate::LayoutReaderRef;
-use crate::layouts::list::ListLayout;
 use async_trait::async_trait;
 use futures::future::try_join_all;
+use vortex_array::MaskFuture;
 use vortex_array::arrays::ListArray;
-use vortex_array::expr::{Expression, root};
+use vortex_array::expr::Expression;
+use vortex_array::expr::root;
 use vortex_array::validity::Validity;
-use vortex_array::{Array, MaskFuture};
 use vortex_dtype::DType;
 use vortex_dtype::FieldMask;
 use vortex_error::VortexResult;
 use vortex_mask::Mask;
+
+use crate::ArrayFuture;
+use crate::LayoutReader;
+use crate::LayoutReaderRef;
+use crate::layouts::list::ListLayout;
 
 /// `LayoutReader` for the list layout holding
 /// `List`-typed data.
@@ -133,7 +135,7 @@ impl LayoutReader for ListReader {
 
         let expr = expr.clone();
 
-        // TODO(aduffy): this is awful. We have to splat the mask to be in the elements space.
+        // TODO(aduffy): this is awful. But it's hard to splat the mask onto the elements space.
         Ok(Box::pin(async move {
             let offsets_fut =
                 offsets.projection_evaluation(&offsets_range, &root(), offsets_mask)?;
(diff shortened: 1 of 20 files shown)


vortex-layout/src/layouts/list/reader/list.rs
```diff
@@ -0,0 +1,518 @@
...
+        let start = row_range.start;
+        let end = start + self.layout.row_count;
+
+        // TODO(aduffy): what you really want here is use the elements (possibly nested fields)
+        //  to set the splits. But the elements row indices can't be mapped back to the table row
+        //  indices, and we can't read the List offsets here. For now, we use a naive fixed-size
+        //  split for scanning ListLayout.
+        for idx in (start..=end).step_by(LIST_STEP_SIZE).skip(1) {
+            splits.insert(idx);
+        }
```
a10y, line 98: @gatesn curious if you have thoughts here? It's hard to figure out the splits b/c you want to do it in elements-space and then map that back to row-space, but it's hard to do that here in register_splits

vortex-layout/src/layouts/list/reader/list.rs
```diff
@@ -0,0 +1,518 @@
...
+
+        let row_range = row_range.clone();
+        let offsets = self.offsets.clone();
+        let elements = self.elements.clone();
+        let validity = self.validity.clone();
+
+        Ok(Box::pin(async move {
+            let mask = mask.await?;
+
+            let list = if mask.density() <= DENSE_THRESHOLD {
```
a10y, line 143: this threshold was chosen somewhat arbitrarily

asubiotto:
Just a heads up that we do write ListViews since https://github.com/vortex-data/vortex/pull/6393 and we plan to use them heavily (including overlapping views) at PS

AdamGS:
I guess I was on PTO when this PR opened, @a10y do you wan to revive it? I can take the conflict resolution part here.

asubiotto:
Just wanted to note some things we talked about in the sync. I'm not familiar with the layout code in this PR so a lot of this might be obvious/addressed already:
- We write ListViews (with sizes) now, so we should maybe also introduce a listview layout? (i.e. encode sizes in a separate child layout?).
- I think the code already handles this, but we have complex types within lists (like dictionaries), so the list layout should support arbitrary element layouts (e.g. dict within struct). I wouldn't worry to much about chunking at list boundaries to start with.

PR #5108, opened 2025-10-29
Operator trait for sequence array

gatesn:


encodings/sequence/src/operator.rs
```diff
@@ -1,223 +1,291 @@
...
 
-    fn dtype(&self) -> &DType {
-        Array::dtype(self.as_ref())
-    }
+        let elements = match selection.indices() {
+            AllOr::All => PVectorMut::<T>::from_iter((0..selection.len()).map(|i| {
+                // This should never panic if the SequenceArray was constructed correctly
+                let offset = T::from_usize(i).vortex_expect("Overflow converting usize to ptype");
+                let scaled = self.multiplier * offset;
+                self.base + scaled
```
joseph-isaacs, line 87: It's more likely to overflow here.

encodings/sequence/src/operator.rs
```diff
@@ -1,223 +1,291 @@
...
+    fn execute(self: Box<Self>) -> VortexResult<Vector> {
+        let selection = self.selection.execute()?;
 
-    fn dtype(&self) -> &DType {
-        Array::dtype(self.as_ref())
-    }
+        let elements = match selection.indices() {
+            AllOr::All => PVectorMut::<T>::from_iter((0..selection.len()).map(|i| {
+                // This should never panic if the SequenceArray was constructed correctly
+                let offset = T::from_usize(i).vortex_expect("Overflow converting usize to ptype");
```
joseph-isaacs, line 85: Maybe we want to assert the first and last number are valid and use an unchecked into usize?

joseph-isaacs (approved):

encodings/sequence/src/operator.rs
gatesn: Is there an unchecked conversion? Note that these can come from floats too, which is why we go via num traits iirc

robert3005 (approved):

PR #5111, opened 2025-10-29
Vortex Session

gatesn:
# VortexSession

We have had a `VortexSession` object for a while, but never really made use of it, in a large part because none of the Vortex components could take a dependency without creating a cycle.

Since Vortex is incredibly extensible, the idea of the session is to hold the plugin registries for arrays, layouts, expressions and so on. While our existing solution does this, it created very fragile APIs where it was possible to construct file readers/writers _without_ using the main registry of plugins. Similarly, metrics registries were getting lost and some metrics ended up in isolated `VortexMetrics::default()` instances - never to see the light of day.

This PR introduces a top-level VortexSession create that essentially acts as a type map. Components of Vortex can register session state, and then read it back again later. This allows components to accept and hold a VortexSession internally, while our root `vortex` crate performs a default configuration.

# Runtime Handles

One of the most fragile bits of mis-configured API was the Vortex runtime handle. This is our abstraction over runtimes (in practice, only over Tokio or a custom smol-based CurrentThreadRuntime) and provides an API for components to spawn CPU, I/O, blocking and other types of work.

This PR demonstrates the session API by holding a runtime handle and plumbing this through APIs that formerly accepted a with_handle argument.

# Language Bindings

This PR doesn't not completely migrate language bindings onto a session-based API. Many bindings still use a global static session where a session instance would be preferred.


vortex-session/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// It is also the entry-point passed to dynamic libraries to initialize Vortex plugins.
+#[derive(Clone, Debug)]
+pub struct VortexSession(Arc<SessionVars>);
+
+impl VortexSession {
+    /// Creates an empty Vortex session.
+    ///
+    /// Do not call this function otherwise you will end up with an empty session!
+    pub fn _empty() -> Self {
```
a10y: any reason to make pub?

if we really don't want people calling it we could slap a `#[doc(hidden)]` on it

gatesn: I haven't yet decided between:
* User creates an empty session, in which case we can just impl Default and the functions on the session are get_or_create.
* Downstream crates must explicitly initialize their state into the session. Possibly using inventory?? This seems weirder to me though. But it does allow for non-default state on the session, for example I don't think the async runtime should be defaulted on first access?

vortex-session/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+            .map(|v| {
+                v.as_any_mut()
+                    .downcast_mut::<V>()
+                    .vortex_expect("Type mismatch - this is a bug")
+            })
+    }
+}
+
+/// A TypeMap based on `https://docs.rs/http/1.2.0/src/http/extensions.rs.html#41-266`.
+type SessionVars = DashMap<TypeId, Box<dyn SessionVar>, BuildHasherDefault<IdHasher>>;
```
a10y, line 114: is the idea that every VTableRegistry<T> would be a separate session_var, and then you'd have to do like `session.get_mut::<ArrayRegistry>().register_many(..)` ?

i wonder if it'd be cleaner to just let this be a multiset mapping instead

gatesn: More like, each crate stores its own session state. End users / code shouldn't call session.get_mut, they should call `ArraySessionExt::register_many(...)`

AdamGS:
another possible solution is to split the implementation and the trait - we implement some `DefaultVortexSession` in the top-level `Vortex` crate, and a `VortexSession` crate in `vortex-session` which can be a dependency of all the other crates, and they can take a `&dyn 
VortexSession` argument that gets populated or just passed around.

gatesn:
@AdamGS the trait definition wouldn't be able to reference types from any of the other vortex crates. If you remove those types, you end up with general get/set on the session, which is basically the design we end up with here.

AdamGS:
@gatesn Depending on how its used, the trait can potentially only know about the traits, right? in which case it could be placed really close to the root of the dependency tree. It might require some restructuring there, which might not be desirable.
Either way - seems like you're already pretty deep into this, and I don't think I fully grok the implications here.

vortex-file/src/lib.rs
```diff
@@ -155,11 +154,12 @@ mod forever_constant {
...
-/// A default registry containing the built-in Vortex encodings and layouts.
-pub static DEFAULT_REGISTRY: LazyLock<Arc<ArrayRegistry>> = LazyLock::new(|| {
-    // Register the compressed encodings that Vortex ships with.
-    let mut registry = ArrayRegistry::canonical_only();
-    registry.register_many([
+/// Register the default encodings use in Vortex files with the provided session.
+///
+/// NOTE: this function will be changed in the future to encapsulate logic for using different
+/// Vortex "Editions" that may support different sets of encodings.
+pub fn register_default_encodings(session: &VortexSession) {
```
lwwmanning, line 161: it's weird that this is not a mut ref?

lwwmanning (approved):

PR #5117, opened 2025-10-29
Feature: Implement `bind` for `BitPackedArray`

connortsui20:
Copied and pasted from the changes:

```
// TODO(connor): This function is implemented in a very roundabout way where we use the
// existing `BitPackedArray` `filter` implementation that gives us an array, and then we
// extract out the underlying buffer of the `PrimitiveArray` to create a `PrimitiveVector`.
//
// Ideally, we should take the underlying `ByteBuffer` of the `BitPackedArray` and unpack
// that directly into a `Buffer<T>` via a `BufferMut<T>`. This is a much more general
// solution that does not force everyone to use `PrimitiveBuilder`.
//
// However, the current decompression implementation for `BitPackedArray` is heavily tied
// to the `PrimitiveBuilder` and `UninitRange` API. What we really need to do is _replace_
// the `PrimitiveBuilder` with `PrimitiveVectorMut`, where instead of `UninitRange` we can
// write directly to a `PVectorMut`.
//
// For the sake of time to get this working, we have implemented this like so.
// When we eventually replace our builders with vectors, we can revisit this.
```

TLDR; implementing this in a good way would require ripping out the existing bitpack decompression implementation that uses `PrimitiveBuilder` and `UninitRange` and replacing it with `PVectorMut`. I think our time would be better spent merging this hacky solution and then revisiting this when we eventually do replace the builders.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8898,6 +8898,7 @@ dependencies = [
  "vortex-alp",
  "vortex-array",
  "vortex-buffer",
+ "vortex-compute",
  "vortex-dtype",
  "vortex-error",
  "vortex-expr",
@@ -8905,6 +8906,7 @@ dependencies = [
  "vortex-mask",
  "vortex-scalar",
  "vortex-utils",
+ "vortex-vector",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


connortsui20:
We'll work on this after [vectors get implemented](https://github.com/vortex-data/vortex/issues/5028) and we replace builders with vectors

PR #5119, opened 2025-10-30
feat: Support passing writer options through the datafusion integration

peasee:
## Description

* Adds a `VortexWriterOptionsFactory` to allow someone using Vortex via the DataFusion `VortexFormat` implementation to be able to pass down custom writer options (e.g. for determining layout structure, etc). I've created a factory without a `Handle`, so the values can be copied into a `VortexWriterOptions` whilst finding an active handle at the time the write occurs.
* Made some opinionated changes to `VortexFormat` and `VortexFormatFactory`. Instead of `::new_with_options()`, prefers a more fleshed out builder pattern with `::new().with_options().with_writer_options_factory()`. This is a breaking change (anyone using `::new_with_options` must now use `::new().with_options()`).


AdamGS (approved):

AdamGS:
LGTM, just needs fixing the clippy error and we can merge it

peasee:
Looks like this was made redundant by https://github.com/vortex-data/vortex/pull/5111

PR #5123, opened 2025-10-30
Chore: Add `match_vector_pair` macro

connortsui20:
Note that the formatting is ugly because of this: https://github.com/rust-lang/rust/pull/52234#issuecomment-903419099

Several of the match statements keep getting out of sync because we have to remember to add variants to match statements everywhere.

Also fixes some of the macro imports since we're `macro_export`ing everything again.

Also fixes an issue with the struct API (accidentally didn't return an `Arc`)

(Note that this is quite important for the implementation of `FixedSizeList` and `ListView` vectors) otherwise there is way too much boilerplate

### DIFF
--- a/vortex-vector/src/macros.rs
+++ b/vortex-vector/src/macros.rs
@@ -39,35 +39,15 @@
 macro_rules! match_each_vector {
     ($self:expr, | $vec:ident | $body:block) => {{
         match $self {
-            $crate::Vector::Null(v) => {
-                let $vec = v;
-                $body
-            }
-            $crate::Vector::Bool(v) => {
-                let $vec = v;
-                $body
-            }
-            $crate::Vector::Primitive(v) => {
-                let $vec = v;
-                $body
-            }
-            $crate::Vector::String(v) => {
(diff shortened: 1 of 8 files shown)


vortex-vector/src/macros.rs
```diff
@@ -132,4 +130,110 @@ macro_rules! match_each_vector_mut {
...
+        $left:expr,
+        $right:expr,
+        $enum_left:ident,
+        $enum_right:ident,
+        $a:ident,
+        $b:ident,
+        $body:expr
+    ) => {{
+        match ($left, $right) {
+            ($crate::$enum_left::Null(left), $crate::$enum_right::Null(right)) => {
```
gatesn: You don't need the `let` expressions, just do:
```
($crate::$enum_left::Null($a), $crate::$enum_right::Null($b)) => $body,
```

connortsui20:
Do you think that the syntax that I chose here makes sense? I could also add a `match` literal at the beginning front to make it super obvious

gatesn (approved):

PR #5127, opened 2025-10-30
Feature: `FixedSizeListVector`

connortsui20:

### DIFF
--- a/vortex-compute/src/arrow/fixed_size_list.rs
+++ b/vortex-compute/src/arrow/fixed_size_list.rs
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use std::sync::Arc;
+
+use arrow_array::{ArrayRef, FixedSizeListArray};
+use arrow_schema::Field;
+use vortex_error::VortexResult;
+use vortex_vector::FixedSizeListVector;
+
+use crate::arrow::IntoArrow;
+
+impl IntoArrow<ArrayRef> for FixedSizeListVector {
+    fn into_arrow(self) -> VortexResult<ArrayRef> {
+        let (elements, list_size, validity) = self.into_parts();
+
+        let converted_elements = elements.as_ref().clone().into_arrow()?;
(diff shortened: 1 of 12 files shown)


vortex-vector/src/fixed_size_list/vector.rs
```diff
@@ -0,0 +1,181 @@
...
+///
+/// `FixedSizeListVector` can be considered a borrowed / frozen version of
+/// [`FixedSizeListVectorMut`], which is created via the [`freeze`](crate::VectorMutOps::freeze)
+/// method.
+///
+/// See the documentation for [`FixedSizeListVectorMut`] for more information.
+#[derive(Debug, Clone)]
+pub struct FixedSizeListVector {
+    /// The child vector of elements.
+    pub(super) elements: Arc<Vector>,
```
gatesn, line 23: I don't think you should arc this

vortex-compute/src/arrow/fixed_size_list.rs
```diff
@@ -0,0 +1,30 @@
...
+    fn into_arrow(self) -> VortexResult<ArrayRef> {
+        let (elements, list_size, validity) = self.into_parts();
+
+        let converted_elements = elements.as_ref().clone().into_arrow()?;
+        let field = Arc::new(Field::new_list_field(
+            converted_elements.data_type().clone(),
+            true, // Vectors are always nullable.
+        ));
+
+        Ok(Arc::new(FixedSizeListArray::try_new(
```
gatesn, line 23: Maybe leave a todo with the Arrow bug linked?

connortsui20: This doesn't run into the same arrow bug since vectors are always nullable

vortex-vector/src/fixed_size_list/vector.rs
connortsui20: We need indirection otherwise infinite size

gatesn (approved):

vortex-vector/src/fixed_size_list/vector.rs
robert3005: Why wouldn't this be a Box though?

PR #5128, opened 2025-10-30
feat: into_struct_fields_*

danking:
I find myself wanting this in Spiral. Is it reasonable?

### DIFF
--- a/vortex-dtype/src/dtype.rs
+++ b/vortex-dtype/src/dtype.rs
@@ -390,6 +390,27 @@ impl DType {
         }
     }
 
+    /// Returns the [`StructFields`] from a struct [`DType`].
+    ///
+    /// # Panics
+    ///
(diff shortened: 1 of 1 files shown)


a10y:
why is the `as_struct_fields` version not good enough here?

danking:
We maybe shouldn't do this but there's cases where I get an owned DType and I want to convert to a Spiral Schema which is a StructFields without nullability (b/c its a SQL-style table: no top-level nulls).

danking:
I currently do `as_struct_fields(...).cloned()`

a10y:
I think this is fine, maybe we should add owned into_ variants for the other conversion methods too

danking:
I added an `into_` method for each of the per-variant `as_` methods.

danking:
all checks passing now.

robert3005 (approved):

PR #5129, opened 2025-10-30
Performance: `ListViewArray` zero-copyable to `ListArray` optimizations

### LINKED ISSUE #5053: ListView to List conversion is slow
### Describe the bug

Currently converting ListView to List goes through scalar operations. Ideally we would avoid double copy (to a scalar and to the target array) to produce list arrays

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

connortsui20:
Closes: https://github.com/vortex-data/vortex/issues/5053

This PR introduces a performance optimization for `ListViewArray` by adding an `is_zero_copy_to_list` flag that tracks whether the `ListViewArray` can be (near) constant-time converted to a `ListArray` without copying data.

When this flag is true (indicating sorted offsets with no gaps and no overlaps), conversions can bypass the very expensive rebuild process (which just calls `append_scalar` in a loop). 

Additionally, this PR refactors all `ListViewArray` constructor call sites and compute operations throughout the codebase to properly maintain this new invariant. The only time you can set this flag is via the `new_unchecked` constructor which is unsafe, which means the caller better be careful to know what they are doing.

Some things that are not included in this PR:
- Making conversion _actually_ zero-copy would require implementing logic to always allocate an additional `offsets` slot for an `n+1`th offset in `ListViewArray`. This is because `ListArray` has `n+1` offsets, and from a plain `ListViewArray` we have to reallocate the offsets to make space.
- `reset_offsets` right now is probably more expensive than it could be since it uses the `sub_scalar` compute function. We know in advance that `reset_offsets` should _never_ fail in the non-recursive case.
- Validation of this `is_zctl` flag is **SUPER** expensive, but it can be made a bit less expensive by using a `BitBuffer` to track references + a flag for overlaps.
- It would probably be nice to have an `check_is_zero_copy_to_list` function on `ListViewArray` that sets the flag if it detects it is zero-copyable to a `ListArray`. This would be nice when we are reading from disk or converting from an Arrow `GenericListView`.
- This PR **does not** change the on-disk format, which means we lose this metadata when writing to disk. I think that this flag is somewhat similar to the `is_sorted` statistic, but at the same time I feel like it is not correct to store this as a statistic? Not sure.

### DIFF
--- a/fuzz/src/array/mask.rs
+++ b/fuzz/src/array/mask.rs
@@ -57,13 +57,18 @@ pub fn mask_canonical_array(canonical: Canonical, mask: &Mask) -> VortexResult<A
         }
         Canonical::List(array) => {
             let new_validity = apply_mask_to_validity(array.validity(), mask);
-            ListViewArray::try_new(
-                array.elements().clone(),
-                array.offsets().clone(),
-                array.sizes().clone(),
-                new_validity,
-            )
-            .vortex_unwrap()
+
+            // SAFETY: Since we are only masking the validity and everything else comes from an
+            // already valid `ListViewArray`, all of the invariants are still upheld.
+            unsafe {
+                ListViewArray::new_unchecked(
+                    array.elements().clone(),
+                    array.offsets().clone(),
+                    array.sizes().clone(),
+                    new_validity,
+                    array.is_zero_copy_to_list(),
+                )
+            }
             .into_array()
         }
         Canonical::FixedSizeList(array) => {
@@ -243,9 +248,15 @@ mod tests {
         let elements = PrimitiveArray::from_iter([1i32, 2, 3, 4, 5, 6]).into_array();
         let offsets = PrimitiveArray::from_iter([0i32, 2, 4]).into_array();
         let sizes = PrimitiveArray::from_iter([2i32, 2, 2]).into_array();
-        let array =
-            ListViewArray::try_new(elements, offsets, sizes, Nullability::NonNullable.into())
-                .unwrap();
+        let array = unsafe {
+            ListViewArray::new_unchecked(
+                elements,
+                offsets,
+                sizes,
+                Nullability::NonNullable.into(),
+                true, // Is zero-copy to list.
+            )
+        };
 
         let mask = Mask::from_iter([false, true, false]);
 
--- a/fuzz/src/array/mask.rs
+++ b/fuzz/src/array/mask.rs
@@ -66,8 +66,8 @@ pub fn mask_canonical_array(canonical: Canonical, mask: &Mask) -> VortexResult<A
                     array.offsets().clone(),
                     array.sizes().clone(),
                     new_validity,
-                    array.is_zero_copy_to_list(),
                 )
+                .with_zero_copy_to_list(array.is_zero_copy_to_list())
             }
             .into_array()
         }
@@ -249,13 +249,8 @@ mod tests {
         let offsets = PrimitiveArray::from_iter([0i32, 2, 4]).into_array();
         let sizes = PrimitiveArray::from_iter([2i32, 2, 2]).into_array();
         let array = unsafe {
-            ListViewArray::new_unchecked(
-                elements,
-                offsets,
-                sizes,
-                Nullability::NonNullable.into(),
-                true, // Is zero-copy to list.
-            )
+            ListViewArray::new_unchecked(elements, offsets, sizes, Nullability::NonNullable.into())
+                .with_zero_copy_to_list(true)
         };
 
         let mask = Mask::from_iter([false, true, false]);
--- a/vortex-array/src/arrays/listview/array.rs
+++ b/vortex-array/src/arrays/listview/array.rs
@@ -5,9 +5,9 @@ use std::sync::Arc;
 
 use num_traits::AsPrimitive;
 use vortex_dtype::{DType, IntegerPType, match_each_integer_ptype};
-use vortex_error::{VortexExpect, VortexResult, vortex_ensure, vortex_err};
+use vortex_error::{VortexExpect, VortexResult, vortex_bail, vortex_ensure, vortex_err};
 
-use crate::arrays::PrimitiveVTable;
+use crate::arrays::{PrimitiveArray, PrimitiveVTable, bool};
 use crate::stats::ArrayStats;
 use crate::validity::Validity;
 use crate::{Array, ArrayRef, ToCanonical};
@@ -50,12 +50,12 @@ use crate::{Array, ArrayRef, ToCanonical};
 /// let offsets = buffer![2u32, 0, 1].into_array();  // Out-of-order offsets
 /// let sizes = buffer![2u32, 1, 2].into_array();  // The sizes cause overlaps
 ///
-/// let list_view = ListViewArray::try_new(
+/// let list_view = ListViewArray::new(
 ///     elements.into_array(),
 ///     offsets.into_array(),
 ///     sizes.into_array(),
 ///     Validity::NonNullable,
-/// ).unwrap();
+/// );
 ///
 /// assert_eq!(list_view.len(), 3);
 ///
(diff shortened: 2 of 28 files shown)


vortex-array/src/arrays/listview/compute/filter.rs
```diff
@@ -46,17 +46,24 @@ impl FilterKernel for ListViewVTable {
         );
 
         // Simply filter the offsets and sizes arrays.
-        // Filters keep scalars in order, and it cannot cause overlaps to form. However, filters
-        // **will** create gaps of unused elements.
         let new_offsets = compute::filter(offsets.as_ref(), selection_mask)?;
         let new_sizes = compute::filter(sizes.as_ref(), selection_mask)?;
 
+        // Filters will create gaps of unused elements.
+        let is_zctl = false;
```
a10y: can just inline this and the comment to the constructor call

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -64,23 +64,27 @@ impl TakeKernel for ListViewVTable {
             )?
         });
 
+        // Take can reorder offsets, create gaps, and may introduce overlaps if the `indices`
+        // contain duplicates.
+        let is_zctl = false;
```
a10y: also can inline into constructor call

vortex-array/src/arrays/listview/conversion.rs
```diff
@@ -12,27 +12,35 @@ use crate::vtable::ValidityHelper;
...
         return Canonical::empty(list.dtype()).into_listview();
     }
 
-    let len = list.len();
+    // We reset the offsets here because mostly for convenience, and also because callers of this
+    // function might not expect the output `ListViewArray` to have a bunch of leading and trailing
+    // garbage data when they turn it back into a `ListArray`.
+    let list = list
+        .reset_offsets(false)
+        .vortex_expect("TODO(connor)[ListView]: This can't fail");
```
a10y, line 30: you don't have to do this right now, but i think that `Canonical::compact` should be infallible. I don't think there's ever a valid case right now where it could return an error since we've made validity and other array ops infallilble. Then we can delete this expect

connortsui20: Yeah I think this should come later (see my updated PR description above) and it can be optimized better

joseph-isaacs:
What perf changes did you expect from the benchmarks?

connortsui20:
@joseph-isaacs none since none of our benchmarks really test the conversion between list and listview (we have a test file that didn't finish on develop but finishes in 3-4 sec on this branch). Maybe we can add that as a be regression benchmark?

fuzz/src/array/mask.rs
```diff
@@ -57,13 +57,18 @@ pub fn mask_canonical_array(canonical: Canonical, mask: &Mask) -> VortexResult<A
...
+
+            // SAFETY: Since we are only masking the validity and everything else comes from an
+            // already valid `ListViewArray`, all of the invariants are still upheld.
+            unsafe {
+                ListViewArray::new_unchecked(
+                    array.elements().clone(),
+                    array.offsets().clone(),
+                    array.sizes().clone(),
+                    new_validity,
+                    array.is_zero_copy_to_list(),
```
joseph-isaacs: This feels leeky here

vortex-array/src/arrays/listview/tests/operations.rs
```diff
@@ -156,9 +155,16 @@ fn test_slice_edge_cases(
...
-    let listview = ListViewArray::try_new(elements, offsets, sizes, Validity::NonNullable)
-        .unwrap()
-        .into_array();
+    let listview = unsafe {
+        ListViewArray::new_unchecked(
+            elements,
+            offsets,
+            sizes,
+            Validity::NonNullable,
+            true, // Is zero-copy to list.
```
joseph-isaacs: Seems like a `with_zero_copy` method would read better here

vortex-array/src/arrays/listview/array.rs
```diff
@@ -345,11 +388,87 @@ where
...
+    Ok(())
+}
+
+/// Helper function to validate if the [`ListViewArray`] components are actually zero-copyable to
+/// [`ListArray`](crate::arrays::ListArray).
+fn validate_zctl(
+    elements: &dyn Array,
+    offsets_primitive: PrimitiveArray,
+    sizes_primitive: PrimitiveArray,
+) -> VortexResult<()> {
```
joseph-isaacs, line 459: is this tested?

joseph-isaacs (commented):
Its nice we can avoid changing serialised array. Its telling that we always have a is_zctl comments everywhere

vortex-array/src/arrays/listview/array.rs
```diff
@@ -128,39 +137,61 @@ impl ListViewArray {
...
+    ///   (even if the corresponding view is defined as null by the validity array).
     /// - If validity is an array, its length must equal `offsets.len()`.
+    /// - If the `is_zctl` flag is true, then the [`ListViewArray`] is zero-copyable to a
+    ///   [`ListArray`].
     pub unsafe fn new_unchecked(
         elements: ArrayRef,
         offsets: ArrayRef,
         sizes: ArrayRef,
         validity: Validity,
+        is_zctl: bool,
```
gatesn: In general, boolean arguments to _any_ function are a no no.

But here, I think this logic should be wrapped up in a `From<ListArray> for ListView` impl, rather than the user ever being able to pass this flag. Even the serde `build` function can check the flag and go via ListArray.

vortex-array/src/arrays/listview/array.rs
connortsui20: Yes at every call site where we set is_zctl to true

vortex-array/src/arrays/listview/array.rs
connortsui20: I think the issue with that here is we need a way to specify something is zero-copyable without  going through a `ListArray` (see the call sites where we change one component and then reconstruct things, like in slice or mask).

I think I like @joseph-isaacs's idea of adding an unsafe builder-like `with_zctl` method, but at the same time I feel that also makes it even harder to reason about the unsafety (harder than it already was) as you now need to worry about two unsafe calls instead of 1. Maybe that's fine?

fuzz/src/array/mask.rs
connortsui20: Leaky in what way? This is basically identical to the `mask` compute function impl

vortex-array/src/arrays/listview/array.rs
connortsui20: I'm going to add an unsafe `with_zctl` method

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -53,121 +66,112 @@ impl ListViewArray {
...
-        // 2. The sizes array is unchanged, so each (new_offset, size) pair still describes a
-        //    valid range within the compacted elements array.
-        // 3. The validity is unchanged and still matches the array length.
+    /// Rebuilds a [`ListViewArray`] by trimming any unused / unreferenced leading and trailing
+    /// elements, which is defined as a contiguous run of values in the `elements` array that are
+    /// not referecened by any views in the corresponding [`ListViewArray`].
+    fn rebuild_trim_elements(&self) -> ListViewArray {
+        let start = if self.is_zero_copy_to_list() {
+            // If offsets are sorted, then the minimum offset is the first offset.
+            self.offset_at(0)
```
a10y, line 116: does this hold if the first item is null?

connortsui20: We have the constraint in `ListArray` where even if there are null elements defined by the validity mask, the offsets still must be sorted. So if the `ListViewArray` came from a `ListArray` that had a null first element, this is still valid.

Also when we validate offsets and sizes we completely ignore nulls since for `ListViewArray` we have the constraint that the views must be valid even if the validity mask specifies the view is null.

See https://arrow.apache.org/docs/format/Columnar.html#listview-layout, specifically:

> Every list-view value, including null values, has to guarantee the following invariants:

This does deserve a comment though

Edit: Added the comment in https://github.com/vortex-data/vortex/pull/5129/commits/258a35a5d8c1a8a4e5147a8a65adfa2af6b41243

a10y (approved):
Looks good, also ran it again with our broken test case file and perf on the last commit is back to acceptable. Ran the fuzzer for a few minutes and made sure it didn't catch issues either.

Let's ship it

PR #5133, opened 2025-10-31
feat: rest of BinaryViewVector{,Mut}

a10y:


vortex-vector/src/binaryview/types.rs
```diff
@@ -0,0 +1,120 @@
...
+impl BinaryViewType for StringType {
+    type Slice = str;
+
+    #[inline(always)]
+    fn validate(bytes: &[u8]) -> bool {
+        std::str::from_utf8(bytes).is_ok()
+    }
+
+    fn from_bytes(bytes: &[u8]) -> &Self::Slice {
+        // SAFETY: vectors should be checked at the boundary for upholding the UTF8 variant,
```
gatesn, line 61: Seems weird that you would assume the safety here. This function should probably itself be marked unsafe

vortex-vector/src/binaryview/vector.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// components.
+    ///
+    /// The caller must uphold all validation that would otherwise be validated by
+    /// the [safe constructor][Self::try_new].
+    pub unsafe fn new_unchecked(
+        views: Buffer<BinaryView>,
+        buffers: Arc<Box<[ByteBuffer]>>,
+        validity: Mask,
+    ) -> Self {
+        Self {
```
gatesn: We settled on a pattern of calling try_new in debug mode. Check the other vectors

vortex-vector/src/binaryview/vector.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// Get the `index` item from the vector as a native `Slice` type.
+    ///
+    /// This function will panic is `index` is out of range for the vector's length.
+    pub fn get(&self, index: usize) -> Option<&T::Slice> {
+        if !self.validity.value(index) {
+            return None;
+        }
+
+        let view = self.views[index];
```
gatesn: You probably want a reference here? I'm not sure we should mark the BinaryView as Copy since it's actually quite large... 

I wonder if there was some call site that required that bound that we could relax?

vortex-vector/src/binaryview/vector.rs
```diff
@@ -0,0 +1,198 @@
...
+        if view.is_inlined() {
+            // start = seek to the `index`-th BinaryView, plus 4 bytes for the leading u32 field
```
gatesn: If the above was a reference, not a copy, I think you'd be able to call `view.inlined().as_slice()` or whatever. Presumably view has some function that returns enum { Inlined , Ref }?

vortex-vector/src/binaryview/vector.rs
```diff
@@ -0,0 +1,198 @@
...
+                    buffers: self.buffers,
+                    _marker: std::marker::PhantomData,
+                });
+            }
+        };
+
+        let buffers_mut = match Arc::try_unwrap(self.buffers) {
+            Ok(buffers) => buffers.into_vec(),
+            Err(arc_buffers) => {
+                return Err(Self {
```
gatesn: I'm not sure it's worth bailing here? Since we're passing the buffers as immutable ByteBuffer to the BinaryViewMut _anyway_, we may as well just zero-copy them? Really this try_unwrap just avoids copying the Vec that holds the buffers. It's not a big deal if we have to though.

vortex-vector/src/binaryview/vector_mut.rs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the validity mask has the same length as the views.
+    pub unsafe fn new_unchecked(
+        views: BufferMut<BinaryView>,
+        validity: MaskMut,
+        buffers: Vec<ByteBuffer>,
+    ) -> Self {
+        Self {
```
gatesn: Same here for cfg!(debug)

vortex-vector/src/binaryview/vector_mut.rs
```diff
@@ -0,0 +1,340 @@
...
+        // Close any existing views into a new buffer
+        if let Some(open) = self.open_buffer.take() {
+            self.buffers.push(open.freeze());
+        }
+
+        // We build a lookup table to map BinaryView's from the `other` to have
+        // valid buffer indices in the current array.
+        let mut buf_index_lookup: Vec<u32> = Vec::with_capacity(other.buffers().len());
+        let mut new_buffers = Vec::new();
+        for buffer in other.buffers().iter() {
```
gatesn: If you assume the buffers are unique, then don't you just need to offset all the indices by `n = self.buffers.len()`?

vortex-vector/src/binaryview/vector_mut.rs
```diff
@@ -0,0 +1,340 @@
...
+            if new_index as usize == new_buffers.len() {
+                // We need to append the buffer
+                new_buffers.push(buffer.clone());
+            }
+
+            buf_index_lookup.push(new_index);
+        }
+
+        // rewrite the views using our lookup table
+        let new_views = rewrite_views(other.views().iter().copied(), &buf_index_lookup);
```
gatesn: Maybe name this `new_views_iter` to make it clear it's not an eager copy

vortex-vector/src/binaryview/view.rs
```diff
@@ -259,3 +265,73 @@ impl fmt::Debug for BinaryView {
...
     }
 }
+
+/// Validate that all views either
+///
+/// 1. Contain valid inline data that conforms to type constraints as defined by the `validator`
+/// 2. Points at a valid range of owned buffer memory, and the bytes stored there conform to
+///    the type constraints as defined by the `validator`.
+pub(super) fn validate_views<ValidateFn, IsValidFn>(
+    views: impl AsRef<[BinaryView]>,
```
gatesn: Why not just take the `&[BinaryView]`? Save on some code gen

vortex-vector/src/binaryview/vector.rs
a10y: I think Buffer forces T: Copy for a lot of methods, but you're right this can be a reference and it allows us to clean up the impl a bit

vortex-vector/src/binaryview/types.rs
a10y: done

vortex-vector/src/binaryview/vector.rs
a10y: fixed here and below

vortex-vector/src/binaryview/vector.rs
a10y: Yea I was a bit on the fence here. The documentation for this method says

> This method will only succeed if self is the only unique strong reference

So in that case, all allocations should be reusable and this Arc should unwrap into the underlying Box without issue. It would technically break the contract as-written if we don't return Err here, but you're right that it's probably quite cheap most of the time to clone into a new Vec. Happy to do whichever

vortex-vector/src/binaryview/vector_mut.rs
a10y: But after merging the buffer points might be out of order now, so we can't just do a blanket offset.

E.g. if self has buffers `[buf1, buf2]` and other has buffers `[buf3, buf1, buf4, buf2]`

vortex-vector/src/binaryview/view.rs
a10y: done

vortex-vector/src/binaryview/vector_mut.rs
gatesn: I guess I assumed we just wouldn't de-dupe? As in, it doesn't really matter if `other` has buf1 or buf2, can just treat them as new buffers.

vortex-vector/src/binaryview/vector.rs
gatesn: Yeah i think we should copy. Seems a shame not to reuse the views Buffer if we fail on this (although you're right, I'm not sure when one is true and the other isn't? I'm sure there's some weird sequence of events that gets us there though)

vortex-vector/src/binaryview/vector_mut.rs
a10y: The existing builders do have some logic to try and dedupe that I figured we'd want to carry over, but if not then that certainly makes things simpler.

vortex-vector/src/binaryview/vector_mut.rs
```diff
@@ -0,0 +1,344 @@
...
+use crate::binaryview::view::{BinaryView, validate_views};
+use crate::{VectorMutOps, VectorOps};
+
+/// Mutable variable-length binary vector.
+#[derive(Clone, Debug)]
+
+/// A mutable vector of binary view data.
+///
+/// The immutable equivalent of this type is [`BinaryViewVector`].
+pub struct BinaryViewVectorMut<T: BinaryViewType> {
```
connortsui20: This doesn't seem right? Didn't even know this could happen without a compile fail

vortex-vector/src/binaryview/vector.rs
a10y: done

vortex-vector/src/binaryview/vector_mut.rs
a10y: I removed the deduping logic in favor of trivial appending

gatesn (approved):

PR #5134, opened 2025-10-31
Worker pool for current thread runtimes

gatesn:
This allows current thread runtimes to have a dynamic number of background threads partaking in the work. 

Removes the Tokio dependency from vortex-duckdb

Other language bindings should migrate to this runtime and expose a set_worker_threads(n) function to the language.


vortex-duckdb/src/lib.rs
```diff
@@ -2,10 +2,12 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 #![allow(clippy::missing_safety_doc)]
+// **WARNING end
```
robert3005, line 5: What's this for?

vortex-duckdb/src/copy.rs
```diff
@@ -29,21 +29,19 @@ pub struct BindData {
...
 /// flushed to disk.
 pub struct GlobalState {
-    write_task: Option<JoinHandle<VortexResult<WriteSummary>>>,
+    write_task: Mutex<Option<Task<VortexResult<WriteSummary>>>>,
     sink: Option<Sender<VortexResult<ArrayRef>>>,
+    // Pool of background workers helping to drive the write task.
+    // Note that this is optional and without it, we would only drive the task when DuckDB calls
+    // into us, and we call `RUNTIME.block_on`.
+    #[allow(dead_code)]
+    worker_pool: CurrentThreadWorkerPool,
```
robert3005, line 44: why is this dead code?

robert3005 (approved):
Is this supposed to be used in this pr or just passed around?

vortex-duckdb/src/lib.rs
gatesn: Great question

vortex-duckdb/src/copy.rs
gatesn: We need to keep it alive as it holds onto the threads, but never need to do anything with it

vortex-io/src/runtime/pool.rs
```diff
@@ -0,0 +1,99 @@
...
+            for _ in current..n {
+                let shutdown = Arc::new(AtomicBool::new(false));
+                let executor = self.executor.clone();
+                let shutdown_clone = shutdown.clone();
+
+                std::thread::Builder::new()
+                    .name("vortex-current-thread-worker".to_string())
+                    .spawn(move || {
+                        // Run the executor with a sleeping future that checks for shutdown
+                        block_on(executor.run(async move {
```
robert3005, line 53: The thing that wasn't immediately obvious to me is that this is not `run this closure` but `run this executor until this future completes` which maybe makes more sense if you have more rust runtime building experience. I.e. I thought this was `spawn` but there's `spawn` on this executor as well. This is just a comment for my understanding

vortex-io/src/runtime/current.rs
```diff
@@ -7,51 +7,44 @@ use futures::stream::BoxStream;
...
+/// It's also possible to clone the runtime onto other threads, each of which can call `block_on`
+/// to drive work on that thread. Each thread shares the same underlying executor with the same
+/// set of tasks, allowing work to be driven in parallel.
+///
+/// For automatic driving of work, a [`CurrentThreadWorkerPool`] can be created from the runtime
+/// by calling [`new_pool`](CurrentThreadRuntime::new_pool). The returned pool can be configured
+/// with the desired number of worker threads that will drive work on behalf of the runtime.
 #[derive(Clone, Default)]
 pub struct CurrentThreadRuntime {
     executor: Arc<smol::Executor<'static>>,
```
robert3005, line 28: is this a deliberate choice why this is not imported at the top level?

vortex-io/src/runtime/current.rs
```diff
@@ -7,51 +7,44 @@ use futures::stream::BoxStream;
...
     }
 
+    /// Create a new worker pool for driving the runtime in the background.
+    ///
+    /// This pool can be used to offload work from the current thread to a set of worker threads
+    /// that will drive the runtime's executor.
+    ///
+    /// By default, the pool has no worker threads; the caller must set the desired number of
+    /// worker threads using the `set_workers` method on the returned pool.
+    pub fn new_pool(&self) -> CurrentThreadWorkerPool {
```
a10y, line 44: what is the semantic of calling this method twice and having two CurrentThreadWorkerPools pointed at the same CurrentThreadRuntime?

Why isn't managment of workers part of the CurrentThreadRuntime?

gatesn: Both pools can make progress on the same execute. Agree it's a bit odd. Maybe it's worth just having a pool inside the runtime?

PR #5143, opened 2025-11-03
fix[array]: invalid arbitrary decimal scalar value

joseph-isaacs:


vortex-dtype/src/decimal/smallest_type.rs
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use crate::{DecimalDType, DecimalType};
+
+/// Maps a decimal precision into the smallest type that can represent it.
+pub fn smallest_decimal_value_type(decimal_dtype: &DecimalDType) -> DecimalType {
```
AdamGS: seems not important enough to merit its own file

vortex-scalar/src/arbitrary.rs
```diff
@@ -82,19 +84,12 @@ fn random_pvalue(u: &mut Unstructured, ptype: &PType) -> Result<PValue> {
...
-        ))))
-    } else {
-        Ok(ScalarValue(InnerScalarValue::Decimal(DecimalValue::I256(
-            u.int_in_range(
-                i256::MIN_BY_PRECISION[precision as usize]
-                    ..=i256::MAX_BY_PRECISION[precision as usize],
-            )?,
-        ))))
-    }
+    let value = match_each_decimal_value_type!(smallest_decimal_value_type(decimal_type), |D| {
```
robert3005: ah yes, this code predates other decimal types

vortex-array/src/builders/decimal.rs
```diff
@@ -271,6 +284,8 @@ impl Default for DecimalBuffer {
 #[cfg(test)]
 mod tests {
     use vortex_dtype::DecimalDType;
+    
+
```
robert3005: why the padding?

vortex-scalar/src/arbitrary.rs
joseph-isaacs: I think the actual problem was the off-by-one in the Precision array.

vortex-dtype/src/decimal/smallest_type.rs
joseph-isaacs: Where do you think, I don't think it fits in another file

vortex-dtype/src/decimal/types.rs
```diff
@@ -61,8 +61,11 @@ pub trait NativeDecimalType:
     const MAX_SCALE: i8;
 
     /// The minimum value for each precision supported by this decimal type.
+    /// This is an array of length `MAX_PRECISION + 1` where the `i`th element is the minimum value
+    /// for a precision of `i` (including precision 0).
```
joseph-isaacs, line 86: note this?

vortex-dtype/src/decimal/smallest_type.rs
AdamGS: I think there are two better options here:
1. Move this function and `is_compatible_decimal_value_type` into here as a `utils` module or into `vortex-dtype/src/decimla/mod.rs`.
2. Make both of them `pub(crate)`, Its a theoretically nice thing to expose but the semantics are pretty clear and I don't think it should be part of the public API.

joseph-isaacs: This is used outside of the array crate even without that move, and I think this should be pub for that reason

joseph-isaacs: `DecimalType` being here is the annoying thing.

joseph-isaacs: Moved into a impl block

vortex-scalar/src/arbitrary.rs
robert3005: When I looked at this pr there were no chanegs to the MAX_PRECISION. Anyway, sure that seems like a bug

robert3005 (approved):

PR #5149, opened 2025-11-03
chore[fuzz]: run fuzz in rel with debug remove multiple jobs

joseph-isaacs:


robert3005:
What does it run otherwise if you don't specify these? I thought release is the default?

joseph-isaacs:
debug, I think, but if not we didn't run `--debug-assertions`

robert3005:
It does release by default. Anyway, we can make it explicit here.

robert3005 (approved):

PR #5154, opened 2025-11-03
chore[fuzzer]: fixup idx selection scalar_at

joseph-isaacs:
I think this would get stuck in a loop


fuzz/src/array/mod.rs
```diff
@@ -331,14 +332,14 @@ impl<'a> Arbitrary<'a> for FuzzArrayAction {
...
-
-                    while indices.len() < num_indices {
-                        let idx = u.choose_index(current_array.len())?;
-                        indices.insert(idx);
-                    }
-
-                    let indices_vec: Vec<usize> = indices.into_iter().collect();
+                    let indices_vec = (0..num_indices)
+                        .map(|_| {
+                            u.int_in_range(0usize..=(current_array.len() - 1))
```
robert3005: Any particular reason to not use choose_index?

robert3005 (approved):

PR #5156, opened 2025-11-03
chore[array]: fsl append null scalar

joseph-isaacs:


vortex-array/src/builders/fixed_size_list.rs
```diff
@@ -71,6 +71,10 @@ impl FixedSizeListBuilder {
     ///
     /// [`ListArray`]: crate::arrays::ListArray
     pub fn append_value(&mut self, value: ListScalar) -> VortexResult<()> {
+        if value.is_null() {
+            self.append_null();
+            return Ok(());
+        }
```
connortsui20: The null check is done below with the `let Some` statement so we don't need this (and we want to do that after the length check to ensure the dtype is the same).

Maybe you can move the comment inside the let statement up above the let statement to make it more obvious?

connortsui20 (changes requested):

vortex-array/src/builders/fixed_size_list.rs
joseph-isaacs: swapped the order

connortsui20 (approved):
oh right len returns 0 when null

PR #5158, opened 2025-11-03
bug[array]: min max bool incorrect

joseph-isaacs:
```
  left: Some(MinMaxResult { min: Scalar { dtype: Bool(Nullable), value: ScalarValue(Bool(false)) }, max: Scalar { dtype: Bool(Nullable), value: ScalarValue(Bool(true)) } })
 right: Some(MinMaxResult { min: Scalar { dtype: Bool(Nullable), value: ScalarValue(Bool(true)) }, max: Scalar { dtype: Bool(Nullable), value: ScalarValue(Bool(true)) } })

```


robert3005:
made #5163

PR #5159, opened 2025-11-03
Chore: use `try_new` for struct array `into_arrow`

connortsui20:
plus some comments


vortex-compute/src/arrow/struct_.rs
```diff
@@ -18,17 +18,24 @@ impl IntoArrow<ArrayRef> for StructVector {
...
+                        i.to_string(),
+                        arrow_fields[i].data_type().clone(),
+                        true, // Vectors are always nullable.
+                    )
+                })
                 .collect::<Vec<Field>>(),
         );
 
-        Ok(Arc::new(StructArray::new(
+        Ok(Arc::new(StructArray::try_new(
```
gatesn: Shouldn't this be new_unchecked?

gatesn (approved):

PR #5161, opened 2025-11-03
chore: rename and clarify our default vortex_io runtime

a10y:
This renames the CurrentThreadRuntime to DefaultRuntime, and documents its two distinct modes of operation:

* `DefaultRuntime::current_thread` -> tasks only get polled in `block_on` by the calling thread
* `DefaultRuntime::multithread` -> a target number of background workers is spawned, and tasks will be eagerly polled even without a call to `block_on`

Updates usages in cxx and duckdb crates


vortex-duckdb/src/copy.rs
```diff
@@ -121,11 +115,9 @@ impl CopyFunction for VortexCopyFunction {
                 .await
         });
 
-        let worker_pool = RUNTIME.new_pool();
-        worker_pool.set_workers_to_available_parallelism();
+        RUNTIME.set_workers_to_available_cores();
```
gatesn, line 118: This seems weirder to me than the previous thing? We're setting state on some global thing in the invocation of an instance function.

Before, we would have an isolated pool of workers alive only for the duration of the operation.

Both are weird tbh

vortex-io/src/runtime/default.rs
```diff
@@ -20,29 +24,134 @@ use crate::runtime::{BlockingRuntime, Executor, Handle};
...
+///     if count == 10 {
+///         break;
+///     }
+///     // yield so worker threads can do their thing.
+///     std::thread::yield_now();
+/// }
+/// ```
 #[derive(Clone, Default)]
-pub struct CurrentThreadRuntime {
+pub struct DefaultRuntime {
```
gatesn, line 71: I don't mind looking for a new name, but "Default" has less link to the actual meaning and implies some sort of preference over Tokio which is also not true?

gatesn (changes requested):

vortex-duckdb/src/copy.rs
a10y: The old thing was not isolated, every worker pool had a reference to the underlying Executor which is a global queue of tasks. If you created new pools they were not distinct in any way.

This just makes it explicit that the runtime is a single unit of isolation.

vortex-io/src/runtime/default.rs
a10y: `AdaptiveRuntime` maybe? I don't think the old name accurately depicted how this runtime worked though.

Everything maps onto a global shared smol Executor. That is current thread when there is no worker pool, but when there was one (or more) worker pools, it was just a normal work-stealing multithreaded scheduler.

PR #5162, opened 2025-11-03
feat: error instead of panic on too many buffers

danking:
Currently this panics in `write_flatbuffer`. Panics print a double stack trace and are a bit weird when they happen in a future.

### DIFF
--- a/vortex-array/src/serde.rs
+++ b/vortex-array/src/serde.rs
@@ -155,6 +155,13 @@ impl<'a> ArrayNodeFlatBuffer<'a> {
                 );
             }
         }
+        let n_buffers_recursive = array.nbuffers_recursive();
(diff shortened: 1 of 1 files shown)


vortex-array/src/serde.rs
```diff
@@ -155,6 +155,13 @@ impl<'a> ArrayNodeFlatBuffer<'a> {
                 );
             }
         }
+        let n_buffers_recursive = array.nbuffers_recursive();
+        if u16::try_from(n_buffers_recursive).is_err() {
```
robert3005: n_buffers_recursive > u16::MAX?

danking: Done.

robert3005 (approved):

PR #5163, opened 2025-11-03
fix: Handle nulls in bool array min/max

### LINKED ISSUE #5157: bug: min/max bool array incorrect
### Describe the bug

Incorrectly handles any null values. 

see ji/bool-max-min-bug

### To Reproduce

_No response_

### Expected behavior

Should pass linked test. Should be performant, once check for each u64 value in the bit buffer

### Additional context

_No response_

robert3005:
fix #5157


gatesn (approved):

robert3005:
~There's additional edge case that this doesn't handle - all null bool arrays. We will claim they have min/max of false~ I think entry point will actually cover it

PR #5165, opened 2025-11-03
Chore: Decimal vector cleanup

connortsui20:

Adds a bunch of missing methods plus reorganizes things to be in line with the other existing vectors.

There's also no tests for decimal vectors at all but that can come later...


vortex-vector/src/primitive/generic.rs
```diff
@@ -71,11 +71,15 @@ impl<T> PVector<T> {
         (self.elements, self.validity)
     }
 
-    /// Gets a nullable element at the given index.
+    /// Gets a nullable element at the given index, **WITHOUT** bounds checking.
```
gatesn: This isn't true? It panics on out of bounds?

connortsui20: What I really mean is that it does not follow the behavior of `slice::get` which returns an option on out of bounds... Not really sure of a better way to summarize this? I could also just remove that statement and force everyone to read the whole docs.

connortsui20: I reworded it to `panicking on out-of-bounds.`

gatesn (approved):

PR #5172, opened 2025-11-04
chore[vector]: run fmt

joseph-isaacs:


robert3005 (approved):

vortex-vector/src/vector_mut.rs
```diff
@@ -66,10 +66,9 @@ impl VectorMut {
             DType::Decimal(decimal_dtype, _) => {
                 DecimalVectorMut::with_capacity(decimal_dtype, capacity).into()
             }
-            DType::Utf8(_)
-            | DType::Binary(_)
-            | DType::List(..)
-            | DType::Extension(_) => vortex_panic!("Unsupported dtype for VectorMut"),
+            DType::Utf8(_) | DType::Binary(_) | DType::List(..) | DType::Extension(_) => {
```
0ax1, line 69: Don't we have a CI check for formatting, how did this land?

joseph-isaacs: No idea?

gatesn:
If you could fill in the blank that would be even better! Not sure how we missed this.. otherwise @connortsui20 can FLUP

vortex-vector/src/vector_mut.rs
robert3005: Only way things like that land is a semantic merge conflict

joseph-isaacs:
happy to follow up, but this is breaking CI

PR #5175, opened 2025-11-04
chore[array]: fill_null cleanup + decimal impl

joseph-isaacs:
clean up fill null
add decimal fill null

### DIFF
--- a/vortex-array/src/arrays/decimal/compute/fill_null.rs
+++ b/vortex-array/src/arrays/decimal/compute/fill_null.rs
@@ -3,12 +3,11 @@
 
 use std::ops::Not;
 
-use vortex_buffer::BufferMut;
 use vortex_error::{VortexExpect, VortexResult};
 use vortex_scalar::{Scalar, match_each_decimal_value_type};
 
+use crate::arrays::DecimalVTable;
 use crate::arrays::decimal::DecimalArray;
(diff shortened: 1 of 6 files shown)


vortex-array/src/arrays/decimal/compute/fill_null.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::compute::{FillNullKernel, FillNullKernelAdapter};
+use crate::validity::Validity;
+use crate::vtable::ValidityHelper;
+use crate::{ArrayRef, IntoArray, ToCanonical, register_kernel};
+
+impl FillNullKernel for DecimalVTable {
+    fn fill_null(&self, array: &DecimalArray, fill_value: &Scalar) -> VortexResult<ArrayRef> {
+        let result_validity = Validity::from(fill_value.dtype().nullability());
+
+        Ok(match array.validity() {
```
robert3005, line 20: This should operate on Mask instead of validity.

joseph-isaacs: Why we have a validity and would have to materialise a mask?

robert3005: My thinking is that the mask is the you do the compute over. However, since you immediately take the buffer I think this should be fine

robert3005 (approved):

PR #5179, opened 2025-11-04
fix[vortex-expr]: delegate AnalysisExpr to CastExpr child

asubiotto:
checked_pruning_expr would otherwise return None which would result in no pruning using expressions that contained a CastExpr somewhere. It's relatively common for e.g. struct fields to be cast to the right type for comparison with literals by datafusion, and this expression translates to get_item(cast(struct), "field") = "value". This cast within a get_item would return None when field_path was called before this change.


robert3005 (approved):
I think there's some extra pruning possible with cast expression but that's a more complicated rewrite

PR #5180, opened 2025-11-04
fix[array]: handle empty/all_invalid sum correctly

joseph-isaacs:
empty/all_invalid arrays have a sum of 0.

### DIFF
--- a/vortex-array/src/arrays/chunked/compute/sum.rs
+++ b/vortex-array/src/arrays/chunked/compute/sum.rs
@@ -42,37 +42,31 @@ register_kernel!(SumKernelAdapter(ChunkedVTable).lift());
 fn sum_int<T: NativePType + PrimInt + FromPrimitiveOrF16>(
     chunks: &[ArrayRef],
 ) -> VortexResult<Option<T>> {
-    let mut result: Option<T> = None;
+    let mut result: T = T::zero();
     for chunk in chunks {
         let chunk_sum = sum(chunk)?;
-
-        let Some(chunk_sum) = chunk_sum.as_primitive().as_::<T>() else {
-            // Skip missing null chunk
-            continue;
+        let Some(chunk_sum) = chunk_sum
+            .as_primitive()
+            .as_::<T>()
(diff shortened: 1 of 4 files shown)


vortex-array/src/arrays/chunked/compute/sum.rs
```diff
@@ -84,21 +78,19 @@ fn sum_decimal(chunks: &[ArrayRef], result_decimal_type: DecimalDType) -> Vortex
...
-        let Some(r) = result.checked_add(&chunk_value).filter(|sum_value| {
-            sum_value
-                .fits_in_precision(result_decimal_type)
-                .unwrap_or(false)
-        }) else {
-            // Overflow
+        let Some(r) = chunk_decimal
+            .decimal_value()
+            // TODO(joe): added a precision capped checked_add.
+            .and_then(|c_sum| result.checked_add(&c_sum))
```
robert3005, line 84: In decimal impl we widen the type and then let it wrap around on overflow which is what all the other engines do, we should likely follow the same pattern here

joseph-isaacs: Its already a I256

robert3005: right, then why are we doing a checked add here. This can wrap around

robert3005: My argument here is that the default should not be overflow checking and we can add a checked variant later

joseph-isaacs: We definitely want this to not wrap around? 
Otherwise the sum is meaningless
This is also the assumed behaviour for sum so I would want a larger PR to change this.

robert3005: Got pointe at https://github.com/vortex-data/vortex/blob/2a52b950e0bab6c456b95491dd7263caa2f50b47/vortex-array/src/compute/sum.rs#L30 offline, I think we need to change it but it's not important to this pr

robert3005 (approved):

PR #5181, opened 2025-11-04
feat[scalar]: zero_value

joseph-isaacs:


vortex-scalar/src/scalar.rs
```diff
@@ -189,6 +189,56 @@ impl Scalar {
...
+    /// This is similar to `default_value` except in its handling of nullability.
+    pub fn zero_value(dtype: DType) -> Self {
+        match dtype {
+            DType::Null => Self::null(dtype),
+            DType::Bool(nullability) => Self::bool(false, nullability),
+            DType::Primitive(pt, nullability) => {
+                Self::primitive_value(PValue::zero(pt), pt, nullability)
+            }
+            DType::Decimal(dt, nullability) => {
+                Self::decimal(DecimalValue::from(0), dt, nullability)
```
a10y: what is the decimalvaluetype getting inferred to here? do we care?

robert3005: It will be I32, can we make this explicitly i8?

joseph-isaacs: Why? There is no real space saving.

robert3005: Let’s at least make it explicit, so it doesn’t change by accident

joseph-isaacs: It is, but it really doesn't matter if it does, in it would be good if it did randomly as a test

robert3005 (approved):

PR #5185, opened 2025-11-04
Chore: add docs and tests for decimal vectors

connortsui20:

Note that I'm probably going to move where this is placed in the next [PR](https://github.com/vortex-data/vortex/pull/5186) but the content will be the same.

### DIFF
--- a/vortex-vector/src/decimal/generic_mut.rs
+++ b/vortex-vector/src/decimal/generic_mut.rs
@@ -10,7 +10,151 @@ use vortex_mask::MaskMut;
 
 use crate::{DVector, VectorMutOps, VectorOps};
 
-/// A specifically typed mutable decimal vector.
+/// A mutable vector of decimal values with fixed precision and scale.
+///
+/// `D` is bound by [`NativeDecimalType`], which can be one of the native integer types (`i8`,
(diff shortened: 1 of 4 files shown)


vortex-vector/src/decimal/generic_mut.rs
```diff
@@ -10,7 +10,151 @@ use vortex_mask::MaskMut;
...
+/// [`freeze()`]: crate::VectorMutOps::freeze
+///
+/// # Examples
+///
+/// ## Creating and building decimal vectors
+///
+/// ```
+/// use vortex_dtype::{DecimalDType, PrecisionScale};
+/// use vortex_vector::{DVectorMut, VectorMutOps};
+///
```
a10y, line 39: usually i try and `# ` out the import blocks in doc tests

a10y (approved):

vortex-vector/src/decimal/generic_mut.rs
connortsui20: For the longer examples like these I find that I am usually copy and pasting it somewhere else, and in these cases I actually get really annoyed when imports are hidden. For smaller examples though I agree

PR #5186, opened 2025-11-04
Chore: Reorganize documentation and modules for vectors

connortsui20:

Now that we have basically all of the vectors (`List`s incoming later today), the crate namespace is getting super crowded. This PR removes the re-exporting into the crate level and makes all vectors accessible via the vortex-vector submodules. Additionally, this moves the main examples for each vector type into their respective modules (rather than on the mutable vector type directly).

This is a **purely cosmetic change** that makes it easier to navigate the documentation of `vortex-vector`. Given that this is effectively going to be Vortex's IR, I think it is important to be more careful about what we put into the top-level namespace.

If people don't like having to type out an extra word when importing, then I would suggest we add a `prelude` rather than forcing everything at the top level.


gatesn (changes requested):
I don't know why we would change the namespacing? Yeah there's a bunch of things, but it's finite and as you say, complete.

connortsui20:
It's very difficult to understand what to look for when on the "landing" page docs. This way, it is obvious what is what (for example `Vector` is the top-level type and everything else is a variant of it, if you have 20+ types all in the same place there is no chance you will understand that)

gatesn (approved):

PR #5187, opened 2025-11-04
bench spawn, spawn_blocking, unblock

a10y:
i noticed we just use tokio::spawn for our spawn_cpu tasks for the builtin runtime. i threw together a simple microbench, i'm wondering if we should use either unblock/spawn_blocking instead?


vortex-io/benches/spawn.rs
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: Apache-2.0
```
a10y, line 1: Results on my mac:

```
     Running benches/spawn.rs (target/release/deps/spawn-7b7df3e321b4bf8c)
Timer precision: 41 ns
spawn                    fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ smol_unblock                        │               │               │               │         │
│  ├─ 1                  9.979 ms      │ 9.979 ms      │ 9.979 ms      │ 9.979 ms      │ 1       │ 1
│  ├─ 10                 31.36 ms      │ 31.36 ms      │ 31.36 ms      │ 31.36 ms      │ 1       │ 1
│  ╰─ 100                212.1 ms      │ 212.1 ms      │ 212.1 ms      │ 212.1 ms      │ 1       │ 1
├─ tokio_spawn                         │               │               │               │         │
│  ├─ 1                  81.24 ms      │ 81.24 ms      │ 81.24 ms      │ 81.24 ms      │ 1       │ 1
│  ├─ 10                 742.4 ms      │ 742.4 ms      │ 742.4 ms      │ 742.4 ms      │ 1       │ 1
│  ╰─ 100                6.503 s       │ 6.503 s       │ 6.503 s       │ 6.503 s       │ 1       │ 1
├─ tokio_spawn_blocking                │               │               │               │         │
│  ├─ 1                  15.8 ms       │ 15.8 ms       │ 15.8 ms       │ 15.8 ms       │ 1       │ 1
│  ├─ 10                 30.31 ms      │ 30.31 ms      │ 30.31 ms      │ 30.31 ms      │ 1       │ 1
│  ╰─ 100                213.7 ms      │ 213.7 ms      │ 213.7 ms      │ 213.7 ms      │ 1       │ 1
╰─ tokio_unblock                       │               │               │               │         │
   ├─ 1                  13.31 ms      │ 13.31 ms      │ 13.31 ms      │ 13.31 ms      │ 1       │ 1
   ├─ 10                 31.11 ms      │ 31.11 ms      │ 31.11 ms      │ 31.11 ms      │ 1       │ 1
   ╰─ 100                213.9 ms      │ 213.9 ms      │ 213.9 ms      │ 213.9 ms      │ 1       │ 1
```

gatesn:
We use spawn as this follows DataFusion's CPU scheduling logic and makes most sense.

spawn_blocking isn't really for CPU heavy workloads. It's for I/O bound but blocking workloads. In other words, it's ok to have 500 threads running these tasks.

a10y:
> spawn_blocking isn't really for CPU heavy workloads. It's for I/O bound but blocking workloads.

I thought it was explicitly for CPU work, and that's why the spawn_blocking interface takes a function and not a Future? If you spawn enough expensive work to block all the runtime threads you're going to start crushing your tail latencies won't you?

gatesn:
🤷 it's a bit of both?

> Tokio will spawn more blocking threads when they are requested through this function until the upper limit configured on the [Builder](https://dtantsur.github.io/rust-openstack/tokio/task/struct@crate::runtime::Builder) is reached. This limit is very large by default, because spawn_blocking is often used for various kinds of IO operations that cannot be performed asynchronously. When you run CPU-bound code using spawn_blocking, you should keep this large upper limit in mind; to run your CPU-bound computations on only a few threads, you should use a separate thread pool such as [rayon](https://docs.rs/rayon) rather than configuring the number of blocking threads.

That's why I kept them separate in the Vortex Handle interface, so we _could_ route them to different locations later. I wouldn't be against e.g. TokioRuntime holding an optional second handle to a different runtime / thread pool that we use for CPU tasks.

a10y:
I setup a simple test case, streaming `hits_83.parquet` from file through the writer into a Vortex file in tmpfs https://github.com/vortex-data/vortex/pull/5187/files#diff-8d942868d0453754f6247ebafffd3ea2c08c5c051dbe7c9536619c026cd9548d

There were 3 knobs:
* File system: S3 (~100ms avg latency from my laptop) vs local FS
* Use of `spawn` vs `unblock` for executing the CPU tasks
* Number of Tokio worker threads (1, 2, 4, 8)

The results for the S3 test:

image

In chart form

￼
image

Results for the Local FS test:
￼
image

In chart form:
￼
performance_plot_fs

For the smaller worker thread counts, the unblock version performs considerably better, which isn’t too surprising because that is the situation where you’re likely to get worker thread starvation from the CPU tasks.

With some println debugging, it appears that the CPU tasks spawned inside of the CompressingStrategy can be > 5ms to compress a chunk, which in the `spawn` mode is time spent blocking the executor, and we spawn each chunk for each column, so the number of blocking tasks grows based on the size of the schema.

The performance difference goes away with larger thread counts, but that doesn’t mean it’s not structurally present, just that we haven’t thrown a workload at it where however many worker thread Tokio spawns by default can’t handle it.

PR #5188, opened 2025-11-04
Chore: add immutable `validity` for mutable vectors

connortsui20:

### DIFF
--- a/vortex-vector/src/ops.rs
+++ b/vortex-vector/src/ops.rs
@@ -4,7 +4,7 @@
 //! Definition and implementation of [`VectorOps`] and [`VectorMutOps`] for [`Vector`] and
 //! [`VectorMut`], respectively.
 
-use vortex_mask::Mask;
+use vortex_mask::{Mask, MaskMut};
 
 use crate::{Vector, VectorMut, private};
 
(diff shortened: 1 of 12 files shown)


vortex-vector/src/ops.rs
```diff
@@ -56,6 +56,13 @@ pub trait VectorMutOps: private::Sealed + Into<VectorMut> {
         self.len() == 0
     }
 
+    /// Returns the validity mask of the vector, where `true` represents a _valid_ element and
+    /// `false` represents a `null` element.
+    ///
+    /// Note that while this returns an owned [`MaskMut`], the caller is only allowed to inspect it
```
a10y: it's not an owned MaskMut it's a ref to a MaskMut

connortsui20: oops I was trying to say that we are giving back something that is usually meant for mutating but the caller isn't allowed to

a10y (commented):
is it better to do this or to provide a `is_valid(index: usize)` accessor? why would you want to access the validity before freezing?

connortsui20:
I need it to access child validities for `ListVector` validation (need to check that there are no nulls in offsets and sizes), and tbh this seems like a nice thing to have in general

gatesn (approved):

PR #5190, opened 2025-11-04
Feature: `ListVector`

connortsui20:

This PR adds the `ListVector` type and the basic operations EXCEPT `split_off` and `unsplit` (as we probably should figure that out with `binaryview`).

More docs can come later once more methods are added (we probably also want a `get_list` kind of method).

### DIFF
--- a/vortex-vector/src/listview/vector.rs
+++ b/vortex-vector/src/listview/vector.rs
@@ -1,0 +1,212 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Definition and implementation of [`ListViewVector`].
+
+use std::sync::Arc;
+
+use vortex_error::{VortexExpect, VortexResult};
+use vortex_mask::Mask;
+
+use super::ListViewVectorMut;
+use crate::Vector;
+use crate::ops::{VectorMutOps, VectorOps};
+use crate::primitive::PrimitiveVector;
+
+/// A vector of variable-width lists.
+///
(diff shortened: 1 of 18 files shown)


vortex-vector/src/listview/vector.rs
```diff
@@ -0,0 +1,304 @@
...
+/// Checks that all views are `<= elements_len`.
+#[allow(clippy::cognitive_complexity, clippy::cast_possible_truncation)]
+fn validate_views_bound(
+    elements_len: usize,
+    offsets: &PrimitiveVector,
+    sizes: &PrimitiveVector,
+) -> VortexResult<()> {
+    let len = offsets.len();
+
+    match_each_integer_pvector!(&offsets, |offsets_vector| {
```
gatesn, line 290: You can actually just `match_each_native_ptype!(offsets.ptype(), |T| offsets.into::<PVector<T>>())` I think

gatesn (approved):

vortex-vector/src/listview/vector.rs
connortsui20: Should we keep the macros I wrote for completeness?

gatesn: If you don't need them, then no

connortsui20: Ok I banged my head for a while and basically if we want to do it this way then we need to have a downcast that doesnt take ownership. I think it might be worth adding `as_` methods to the primitive and decimal vectors

gatesn (approved):

PR #5191, opened 2025-11-05
Expression VTable

gatesn:
Experiment with new style of vtable. 

Overview:
* No more vtable macro.
* "Encoding" struct is now combined with the vtable struct (we had an empty vtable struct before for no reason).
* Instead of Arc<dyn Expr>, we have an Expression struct with vtable, opaque instance data, and children.
  * We may want this struct to be Arc'd also?
* Because private data can be anything, simple expressions do not need a separate instance struct (e.g. Binary simply stores Operator enum as private data)
* We retain benefits of vtables in that we can run pre/post logic around typed impl functions. We can also expose a smaller API surface on Expression than we do on VTable.


fuzz/fuzz_targets/file_io.rs
```diff
@@ -42,7 +42,7 @@ fuzz_target!(|fuzz: FuzzFileAction| -> Corpus {
         let filtered = filter(&array_data, &mask).vortex_unwrap();
         projection_expr
             .clone()
-            .unwrap_or_else(|| root())
+            .unwrap_or_else(root)
```
joseph-isaacs, line 45: This different from the std lib, can we revert?

vortex-expr/src/exprs/dynamic.rs
```diff
@@ -1,234 +1,224 @@
...
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.operator.hash(state);
+        Arc::as_ptr(&self.rhs).hash(state);
+        self.default.hash(state);
+    }
+}
+
+/// Hash and PartialEq are implemented based on the ptr of the value function, such that the
+/// internal value doesn't impact the hash of an expression tree.
+struct Rhs {
```
joseph-isaacs, line 199: can we choose a better name?

vortex-expr/src/expression.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Returns the expression's vtable.
+    pub fn vtable(&self) -> &ExprVTable {
+        &self.vtable
+    }
+
+    /// Returns the opaque data of the expression.
+    pub fn data(&self) -> &Arc<dyn Any + Send + Sync> {
+        &self.data
+    }
```
joseph-isaacs, line 103: Is this what we have need calling metadata?

vortex-expr/src/expression.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// _must not_ assume the array is empty _nor_ may you assume the array only contains non-null
+    /// values.
+    pub fn max(&self, catalog: &mut dyn StatsCatalog) -> Option<Expression> {
+        self.vtable.as_dyn().max(self, catalog)
+    }
+
+    /// An expression for the lower non-null bound of this expression, if available.
+    ///
+    /// See [AnalysisExpr::max] for important details.
+    pub fn min(&self, catalog: &mut dyn StatsCatalog) -> Option<Expression> {
```
joseph-isaacs: why add this to the expr, its will be a huge trait if we keep adding

fuzz/fuzz_targets/file_io.rs
gatesn: I think I just ran clippy?

vortex-expr/src/exprs/dynamic.rs
gatesn: This was here before, agree it's a bit weird but don't really want to change more than I need

vortex-expr/src/expression.rs
gatesn: Kind of...? Metadata sort of implied that it only held onto things that needed to be serialized. Instance data (as in, fat pointer of vtable + data), feels like it's more generic / familiar to the vtable structure.

vortex-expr/src/expression.rs
gatesn: Anything that needs to be recursively called within a vtable implementation needs to be exposed on the public Expression API.

Agree these names are a bit weird / non-descriptive. Pulling them into a separate trait I don't think fixes that. Maybe stat_max() / stat_min() ?

vortex-expr/src/expression.rs
joseph-isaacs: what do you expect to be in data?

fuzz/fuzz_targets/file_io.rs
joseph-isaacs: ahh I see.

vortex-expr/src/expression.rs
joseph-isaacs: its more about having a trait for analysis stuff that is not on the expr itself

joseph-isaacs (approved):

PR #5193, opened 2025-11-05
fix[pco]: `cast` array with validity

joseph-isaacs:
We currently cannot cast from a sliced PCO array with invalid values (outside of the sliced) range to a nonnullable pco array.
We likely want a cast: https://github.com/vortex-data/vortex/issues/5196


robert3005 (approved):

robert3005:
Maybe a small test?

robert3005 (approved):

PR #5194, opened 2025-11-05
chore: rename `PValue::as_primitive` to `cast`

joseph-isaacs:
`FromPrimitiveOrF16` to NativePType replacing `FromPrimitive`

Maybe we should add a NumCastOrF16?


robert3005 (dismissed):

vortex-dtype/src/f16.rs
```diff
@@ -0,0 +1,61 @@
...
+
+macro_rules! from_primitive_or_f16_for_unsigned {
+    ($T:ty) => {
+        impl FromPrimitiveOrF16 for $T {
+            fn from_f16(value: f16) -> Option<Self> {
+                value.to_u64().and_then(|v| FromPrimitive::from_u64(v))
+            }
+        }
+    };
+}
```
joseph-isaacs, line 33: This is how f32 and f64 are implemented as a lossy cast

robert3005 (approved):

PR #5198, opened 2025-11-05
Do not construct backtrace for opt

gatesn:


lwwmanning (approved):

a10y:
Why is this better?

gatesn:
Running with RUST_BACKTRACE=1 gets super expensive, as benchmarks show

PR #5201, opened 2025-11-05
Ban Result::ok

gatesn:
See how widespread this is...


joseph-isaacs:
Very I bet

AdamGS:
how do you feel about putting backtraces behind a feature?

gatesn:
It's already gated by an environment variable. The whole point is you can switch them on/off using an env var without recompiling.

robert3005:
I tried fixing this pr and removing usages of ok. It didn’t yield better code and mostly suppressions. While it can be a foot gun it’s too prevalent in valid use cases where all you get is a TryFrom impl

PR #5205, opened 2025-11-05
Take 3 on supporting dynamic expression in DF

AdamGS:
Inspired by @adriangb's attempt and some conversations

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4096,7 +4096,7 @@ dependencies = [
  "portable-atomic",
  "portable-atomic-util",
  "serde_core",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


adriangb (approved):
Looks good to me! Is it worth adding a test or two that shows dynamic filters being pushed down, maybe even that they actually result in pruning? I think I had one in my PR

AdamGS:
Was going to add tests, but than I ran into [this](https://github.com/apache/datafusion/issues/18513) issue which took me a while to repro.

adriangb:
@AdamGS anything I can do to help here? Is the holdup benchmarks?

AdamGS:
Yeah it's basically me wanting to dive into the benchmarks. There are some cases here that I expected it to make more of a difference, and obviously some regressions too. Just got busy with other stuff and taking some time off now but once I'm back at work I'm planning on picking this back up

adriangb:
Okay just let me know if I can help!

I agree the benchmark results are surprising but I'm not sure where it's coming from, how much could be benchmark noise, etc.

PR #5206, opened 2025-11-05
Feature: add scaffolding for `FixedSizeListArray::bind`

connortsui20:
We can't actually implement this right now without some sort of way to expand the mask (see the TODO comment)

### DIFF
--- a/vortex-array/src/arrays/fixed_size_list/vtable/operator.rs
+++ b/vortex-array/src/arrays/fixed_size_list/vtable/operator.rs
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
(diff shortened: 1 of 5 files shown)


vortex-array/src/arrays/fixed_size_list/vtable/operator.rs
```diff
@@ -0,0 +1,79 @@
...
+        }))
+    }
+}
+
+/// Takes a selection mask and "expands" it out by duplicating each bit `list_size` times.
+///
+/// If `selection` is not `None`, the output array is guaranteed to have
+/// `selection.len() * list_size` total bits.
+fn expand_selection(_selection: Option<&ArrayRef>, _list_size: u32) -> Option<ArrayRef> {
+    todo!(
```
gatesn, line 73: Instead of failing, I think you should just do the filter using vortex-compute.

PR #5208, opened 2025-11-05
Bind constant array

gatesn:


vortex-dtype/src/decimal/macros.rs
joseph-isaacs, line 1: vortex-dtype is the wrong place? Do you have a place of where to move this to at a later date?

joseph-isaacs (approved):

vortex-dtype/src/decimal/macros.rs
gatesn: NativePType, NativeDecimalType, PValue (PScalar), DecimalValue (DScalar), and all associated macros should really live in vortex-vector. They're all physical.

connortsui20: I made this: https://github.com/vortex-data/vortex/issues/5209

connortsui20 (approved):

PR #5212, opened 2025-11-05
feat: OperatorVTable::bind for VarBin/VarBinView

a10y:


connortsui20 (commented):
I haven't been able to really look at this module yet, but now that I'm looking at it it seems somewhat strange that we store immutable `ByteBuffer`s inside the `BinaryViewVectorMut`? Basically this means that we will only ever support append operations, and never support interior mutation, even though this mutable vector is supposed to be "owned".

This is probably showing my ignorance with how our varbin view arrays currently work but its also not clear to me how we correctly garbage collect these buffers? We are storing these views inline in a `Buffer`, but it isn't clear to me that there is any logic to free this memory since `BinaryView` does not implement reference counting. I might be misunderstanding something?

a10y:
The current vector is structured for append construction. Once a buffer is closed, it never gets written into again, so using `ByteBuffer` is right in this case.

As for GC: we currently don't do anything intelligent to size the buffers on construction, or to GC them after operations complete. You can look inside of the existing `VarBinViewArray::compact_buffers` impl for an example of how you can scan a vector and compact unused space out of it by building a new one

vortex-array/src/arrays/varbin/vtable/operator.rs
```diff
@@ -0,0 +1,184 @@
...
+                let bytes = &self.bytes[offset as usize..(offset + len) as usize];
+                let view = if len as usize <= BinaryView::MAX_INLINED_SIZE {
+                    BinaryView::new_inlined(bytes)
+                } else {
+                    BinaryView::make_view(bytes, 0, offset)
+                };
+                views.push(view);
+            }
+
+            let selection = self.selection.execute()?;
```
gatesn, line 98: Feels like we should do this first? Then do like a `selection.threshold_iter()` to iterate over indices and slices when constructing the views?

vortex-array/src/arrays/varbinview/vtable/operator.rs
```diff
@@ -0,0 +1,103 @@
...
+        let views = array.views().clone();
+        let buffers = array.buffers().clone();
+
+        Ok(kernel(move || {
+            let selection = mask.execute()?;
+            let validity = validity.execute()?;
+
+            // We only filter the views buffer
+            let views = views.filter(&selection);
+            let views = Buffer::<BinaryView>::from_byte_buffer(views.into_byte_buffer());
```
gatesn: What does this line do?

vortex-array/src/arrays/varbinview/vtable/operator.rs
```diff
@@ -0,0 +1,103 @@
...
+
+            match dtype {
+                // SAFETY: the incoming array has the same validation as the vector
```
gatesn, line 40: I couldn't decide whether building vectors from arrays should be unchecked? I would have thought this would be the place we _do_ run validation since in theory the buffers arrive from the disk unchecked.

But I guess we check during array construction now?

vortex-vector/src/binaryview/vector.rs
```diff
@@ -185,13 +185,7 @@ impl<T: BinaryViewType> VectorOps for BinaryViewVector<T> {
...
         };
 
-        let buffers_mut = match Arc::try_unwrap(self.buffers) {
-            Ok(buffers) => buffers.into_vec(),
-            Err(buffers) => {
-                // Backup: collect a new Vec with clones of each buffer
-                buffers.iter().cloned().collect()
-            }
-        };
+        let buffers_mut = self.buffers.to_vec();
```
gatesn: The reason we had `Arc<Box<[..]>>` was because `Box<[..]>::into_vec` is zero-copy. This now _always_ creates a new `Vec<..>`

Not sure it makes a huge difference. And admittedly wasn't made obvious in the code

vortex-array/src/arrays/varbinview/vtable/operator.rs
a10y: transmutes from a `Buffer<vortex_array::BinaryView>` to `Buffer<vortex_vector::BinaryView>`

vortex-array/src/arrays/varbinview/vtable/operator.rs
a10y: Maybe things will be different once we're done with all of this, but currently yea all arrays are checked on deserialization from file

vortex-array/src/arrays/varbinview/vtable/operator.rs
gatesn: Aha 🤦 

Should we just remove vortex_array::BinaryView? vortex_array already takes a dep on vortex-vector

vortex-vector/src/binaryview/vector.rs
a10y: Yea I was just annoyed at the cruft required to construct this type and thought the extra Vec allocation wouldn't be too harmful, but I can revert this

vortex-array/src/arrays/varbin/vtable/operator.rs
a10y: I think we just want iterating indices, since you need to touch every string to build the view so there's no way to make use of slices, but you're right better to avoid building all the views if we don't need to

gatesn (approved):

PR #5213, opened 2025-11-05
Vector scalars

gatesn:


vortex-buffer/src/bit/buf.rs
```diff
@@ -194,16 +194,24 @@ impl BitBuffer {
...
+            std::ops::Bound::Excluded(&s) => s + 1,
+            std::ops::Bound::Unbounded => 0,
+        };
+        let end = match range.end_bound() {
+            std::ops::Bound::Included(&e) => e + 1,
+            std::ops::Bound::Excluded(&e) => e,
+            std::ops::Bound::Unbounded => self.len,
+        };
+
+        assert!(start <= end);
```
connortsui20, line 209: The default behavior in the std library is to just have an empty slice here

vortex-vector/src/scalar.rs
connortsui20, line 1: Nit: Personally I would prefer to see a `scalar/` module with the 3 different files

gatesn:
So I think we _do_ want options inside the scalars. The reason is that scalars are primarily used for optimized compute (we will add a Datum enum and associated type-specific datums too), and since vectors are all possibly nullable, it's nice to have the analog where scalars are also all possibly nullable.

Having nullable scalars also means we can have an enum with the same set of variants as vector. If we didn't have nullable scalars, then we couldn't have a NullScalar, and the enums don't line up and all our macros get annoying.

vortex-vector/src/bool/vector.rs
```diff
@@ -81,6 +84,21 @@ impl VectorOps for BoolVector {
         &self.validity
     }
 
+    fn scalar_at(&self, index: usize) -> Scalar {
+        debug_assert!(index < self.len());
```
a10y: nit: some of these use debug_assert, some use assert

vortex-buffer/src/bit/buf.rs
a10y: It seems like it actually does panic?
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d77357e3848c7cb1b8927a1ee071afab

vortex-vector/src/scalar.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::listview::ListViewScalar;
+use crate::null::NullScalar;
+use crate::primitive::PrimitiveScalar;
+use crate::struct_::StructScalar;
+use crate::{ScalarOps, VectorMut, match_each_scalar};
+
+/// Represents a scalar value of any supported type.
+pub enum Scalar {
+    /// Null scalars are always null.
+    Null(NullScalar),
```
connortsui20, line 19: So one of the issues I have with pushing the `Option` down into `Scalar` is that it is not clear to me what `scalar.is_null()` should now return. Do we check if the variant is null or if the dependent scalar is null? Note that our current scalar `is_null()` is _incredibly_ confusing to me.

I think a much nicer type is where `Scalar` is never null, and to match our vector variant we simply have `Scalar::Null(Never)` to imply scalars can never be null.

connortsui20:
I think we all agree that the difference between having the option pushed down vs wrapping the `Scalar` is mostly an API thing rather than a logical difference (other than the `is_null()` issue I have above).

I think that as a library, we should be striving to have the most bulletproof possible interface, and I think this is the most ideal in terms of correctness:

```rust
// In an ideal world, we could "implement" `?` and all similar `Option` methods on this type...
pub enum Scalar {
    Null,
    Bool(BoolScalar), // Each "typed" scalar can never be null.
    Decimal(DecimalScalar),
    Primitive(PrimitiveScalar),
    String(StringScalar),
    Binary(BinaryScalar),
    List(ListViewScalar),
    FixedSizeList(FixedSizeListScalar),
    Struct(StructScalar),
}
```

The issue with this, however, is that by doing this we lose the nice properties of `Option` like the entire interface of [`Option`](https://doc.rust-lang.org/std/option/enum.Option.html) as well as the `?` operator and things like `collect()`.

So to me, having an `Option<Scalar>` is the best compromise of accurately describing our data and making it easy to work with.

vortex-vector/src/scalar.rs
joseph-isaacs: `is_null` is both right, the `Null` variant is an untyped (logical) null and the `Option<T> == None` is the typed scalar null.

vortex-mask/src/lib.rs
```diff
@@ -392,12 +392,26 @@ impl Mask {
     }
 
     /// Slice the mask.
-    #[inline]
-    pub fn slice(&self, range: Range<usize>) -> Self {
-        assert!(range.end <= self.len());
+    pub fn slice(&self, range: impl RangeBounds<usize>) -> Self {
+        let start = match range.start_bound() {
+            std::ops::Bound::Included(&s) => s,
```
joseph-isaacs: import please

joseph-isaacs:
I am worried this PR copies over Scalar impls and remove the test/correctness from the previous ones.

joseph-isaacs:
I think having the option inlined into the scalar is a good idea, mirroring Vector. 

I think we should have an OptionalVector and therefore 
OptionScalar variant. 

But if not we would match the Vector and Scalar

a10y (approved):

PR #5214, opened 2025-11-05
Feature: add `into_mut` methods on mask and buffer

connortsui20:


vortex-buffer/src/buffer.rs
```diff
@@ -423,6 +417,18 @@ impl<T> Buffer<T> {
             })
     }
 
+    /// Convert self into `BufferMut<T>`, cloning the data if there are multiple strong references.
+    pub fn into_mut(self) -> BufferMut<T> {
+        let new_bytes = self.bytes.try_into_mut().unwrap_or_else(BytesMut::from);
```
a10y: this still copies though, i don't see why this is different from what previous implementation

connortsui20: It's just less indirection because there is a temporary `Buffer` created in between

a10y: Are you saying that there's a temporary buffer created by 

```
self.try_into_mut()
            .unwrap_or_else(|buffer| BufferMut::<T>::copy_from(&buffer))
```

?

gatesn: You actually can't do this because you trigger a re-allocation of BytesMut without our special alignment logic. The new buffer may not be aligned to self.alignment at all

gatesn (dismissed):

vortex-buffer/src/buffer.rs
connortsui20: I was trying to be clever and avoid the extra `Buffer` struct that's created in the else case but the compiler will probably optimize that away anyways

a10y (approved):

PR #5221, opened 2025-11-06
Feature: add `into_mut` to vectors

connortsui20:
Revived from https://github.com/vortex-data/vortex/pull/5215 since that was supposed to merge _after_ https://github.com/vortex-data/vortex/pull/5214


vortex-buffer/src/bit/buf.rs
gatesn, line 19: I agree this should be true, I don't think it is right now...

We could make this cleaner with:
```
struct BitOffset(usize);

impl From<usize> for BitOffset {
   fn from(...) { assert <8 or panic }
}

impl Deref for BitOffset { Target = usize }
```

It's unrelated to this PR though, so feel free to follow up

gatesn (approved):

PR #5227, opened 2025-11-06
Update benchmarks in bench-vortex

### LINKED ISSUE #5066: Make random access benchmarks more stable
The random access benchmarks are somewhat unstable, so we definitely can run them with more iterations. We'll probably want to do a timed thing (like divan) since parquet is so much slower.

connortsui20:
Closes: https://github.com/vortex-data/vortex/issues/5066

~~I had the claude code web do this so I have no idea if this works yet~~

Should work...

### DIFF
--- a/bench-vortex/src/bench_run.rs
+++ b/bench-vortex/src/bench_run.rs
@@ -41,3 +41,42 @@ where
 
     fastest_result
 }
+
+/// Run a benchmark for a specified time limit, collecting all run durations.
+///
+/// At least one run is always guaranteed, even if it exceeds the time limit.
+pub fn run_timed_with_setup<I, O, S, R, F>(
+    runtime: &Runtime,
+    time_limit_secs: u64,
+    mut setup: S,
+    mut routine: R,
+) -> Vec<Duration>
+where
+    S: FnMut() -> I,
+    R: FnMut(I) -> F,
+    F: Future<Output = O>,
+{
+    let time_limit = Duration::from_secs(time_limit_secs);
+    let overall_start = Instant::now();
+    let mut runs = Vec::new();
+
+    // Looping like this ensures at least one run happens.
+    loop {
+        let state = black_box(setup());
+        let elapsed = runtime.block_on(async {
+            let start = Instant::now();
+            let output = routine(state).await;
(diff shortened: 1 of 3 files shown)


bench-vortex/src/bench_run.rs
```diff
@@ -41,3 +41,41 @@ where
...
+    let mut runs = Vec::new();
+
+    // Ensure at least one run
+    loop {
+        let state = black_box(setup());
+        let elapsed = runtime.block_on(async {
+            let start = Instant::now();
+            let output = routine(state).await;
+            let elapsed = start.elapsed();
+            drop(black_box(output));
```
AdamGS, line 70: can you explain the use of `black_box` here?

connortsui20: no idea... I'm going to test this locally first and I'll probably rewrite stuff

connortsui20: ok its just because the function above uses it

joseph-isaacs:
Any idea why we got a big tpc-h win?

bench-vortex/src/measurements.rs
```diff
@@ -161,7 +161,23 @@ pub struct TimingMeasurement {
...
+            vortex_panic!("cannot have no runs");
+        }
+
+        let total_nanos: u128 = self.runs.iter().map(|d| d.as_nanos()).sum();
+        let mean_nanos = total_nanos / len as u128;
+        Duration::new(
+            u64::try_from(mean_nanos / 1_000_000_000).vortex_unwrap(),
+            u32::try_from(mean_nanos % 1_000_000_000).vortex_unwrap(),
+        )
+    }
```
joseph-isaacs, line 180: can we do median here

joseph-isaacs (changes requested):
median please

bench-vortex/src/bench_run.rs
joseph-isaacs: We want bb in cause the compiler kills the call to routine() which does seem very unlikely

connortsui20:
Closing in favor of https://github.com/vortex-data/vortex/pull/5234

PR #5228, opened 2025-11-06
chore: Convert DuckDB vectors directly to vortex vectors

robert3005:


vortex-duckdb/src/duckdb/vector.rs
```diff
@@ -351,30 +352,24 @@ mod tests {
...
+        assert!(validity.is_valid(0));
+        assert!(validity.is_null(1));
+        assert!(validity.is_valid(2));
+        assert!(validity.is_null(3));
+        assert!(validity.is_valid(4));
+        assert!(validity.is_valid(5));
+        assert!(validity.is_valid(6));
+        assert!(validity.is_null(7));
+        assert!(validity.is_valid(8));
+        assert!(validity.is_valid(9));
```
joseph-isaacs: can we do a single mask assert?

vortex-duckdb/src/convert/vector.rs
```diff
@@ -62,293 +46,275 @@ impl<'a> DuckString<'a> {
...
-                vector.validity_ref(data.len()).to_null_buffer(),
+            let micros = data.iter().map(|duckdb_timestamp { micros }| *micros);
+            let arr = PrimitiveArray::new(
+                Buffer::from_trusted_len_iter(micros),
+                vector.validity_ref(data.len()).to_validity(),
+            )
+            .into_array();
+            Ok(
+                TemporalArray::new_timestamp(arr, TimeUnit::Microseconds, Some("UTC".to_string()))
+                    .into_array(),
```
joseph-isaacs, line 115: happy to not do this but you could impl this with a extraction function and a array creation function sharing the other logic

vortex-duckdb/src/convert/vector.rs
```diff
@@ -62,293 +46,275 @@ impl<'a> DuckString<'a> {
...
-                PrimitiveArray::<Float64Type>::from_iter_values_with_nulls(
-                    data.iter().copied(),
-                    vector.validity_ref(data.len()).to_null_buffer(),
-                ),
-            ))
+            Ok(PrimitiveArray::new(
+                Buffer::copy_from(data),
+                vector.validity_ref(data.len()).to_validity(),
+            )
+            .into_array())
```
joseph-isaacs: again we should share this impl with other PTypes?

vortex-duckdb/src/convert/vector.rs
```diff
@@ -62,293 +46,275 @@ impl<'a> DuckString<'a> {
...
-                PrimitiveArray::<UInt16Type>::from_iter_values_with_nulls(
-                    data.iter().copied(),
-                    vector.validity_ref(data.len()).to_null_buffer(),
-                ),
-            ))
+            Ok(PrimitiveArray::new(
+                Buffer::copy_from(data),
+                vector.validity_ref(data.len()).to_validity(),
+            )
+            .into_array())
```
joseph-isaacs: shall we group all the PTypes?

vortex-duckdb/src/convert/vector.rs
```diff
@@ -62,293 +46,275 @@ impl<'a> DuckString<'a> {
...
-            let decimal_array = Decimal128Array::from_iter_values_with_nulls(
-                decimal_values.into_iter(),
-                vector.validity_ref(len).to_null_buffer(),
-            )
-            .with_precision_and_scale(precision, scale as i8)?;
-
-            Ok(Arc::new(decimal_array))
+                _ => vortex_bail!("Unsupported decimal precision: {precision}"),
+            }
+            .map(|a| a.into_array())
```
joseph-isaacs, line 185: can you return a buffer from above?

vortex-duckdb/src/convert/vector.rs
joseph-isaacs, line 1: Feel free to ignore these, but it would be nice to make the match expr smaller

joseph-isaacs (approved):

joseph-isaacs:
I wonder if its possible for duckdb to hand us an null data ptr is the vector is len 0?

vortex-duckdb/src/convert/vector.rs
robert3005: You cannot since we want a Buffer<T> and now they're no longer getting widened to i128

joseph-isaacs (approved):

PR #5229, opened 2025-11-06
experiment: eagerly decode dict on read

a10y:
Full scan of Clickbench increases by 35%, from 5 seconds -> ~3.2 seconds

Profile from before: https://share.firefox.dev/3JK7h73

Profile from this commit: https://share.firefox.dev/47GHkNB


vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -76,6 +76,7 @@ impl DictReader {
                     )
                     .vortex_expect("must construct dict values array evaluation")
                     .map_err(Arc::new)
+                    .map_ok(|arr| arr.to_canonical().into_array())
```
a10y, line 79: "before" is just running the example with this commented out

joseph-isaacs (approved):
LGTM, we ca revisit in the operator world

PR #5233, opened 2025-11-06
Do not do eager formatting of error strings

gatesn:


robert3005 (approved):

a10y:
There's a clippy lint to prevent this https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call

But obviously it just doesn't apply to our custom vortex_expect

a10y:
I pushed a commit forcing the message argument of `vortex_expect` to have `&'static` lifetime which should prevent this from happening in the future

PR #5237, opened 2025-11-06
feat: use Sparse compression for floats, strings

a10y:
We add sparse compression for array chunks that are null-dominated (> 90%), mirroring the existing support in Ints.

### DIFF
--- a/vortex-btrblocks/src/float.rs
+++ b/vortex-btrblocks/src/float.rs
@@ -43,7 +43,7 @@ impl Compressor for FloatCompressor {
             &ALPScheme,
             &ALPRDScheme,
             &DictScheme,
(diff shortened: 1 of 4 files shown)


vortex-btrblocks/src/float.rs
```diff
@@ -373,14 +380,90 @@ impl Scheme for DictScheme {
     }
 }
 
+impl Scheme for SparseScheme {
```
joseph-isaacs: This is what we want. 

But it's a little odd that a compressor with the same name as int has different behaviour. This one can only choose null there as the int one can choose null or a dominate value?

Maybe name this SparseValue or something?

joseph-isaacs (approved):

PR #5243, opened 2025-11-07
Add conversions from Vectors into ArrayRef

gatesn:


joseph-isaacs (approved):
I assume that we later change the Canonical Arrays to own a vector?

gatesn:
No, canonical arrays stay since they're logical. For example, struct can have compressed children. Primitive can have compressed validity.

PR #5251, opened 2025-11-07
Pipeline execution

gatesn:


vortex-array/src/arrays/primitive/vtable/operator.rs
```diff
@@ -61,3 +71,117 @@ impl OperatorVTable<PrimitiveVTable> for PrimitiveVTable {
         Ok(None)
     }
 }
+
+impl PipelinedSource for PrimitiveArray {
+    fn bind_source(&self, ctx: &mut dyn BindContext) -> VortexResult<Box<dyn SourceKernel>> {
```
gatesn: It's a lot of code for just a primitive array!

I imagine when we actually run in a pipelined fashion, for canonical inputs we just pass a zero-copy view of them into the pipeline. vs here where we copy the elements of the array into the mutable output.

vortex-array/src/pipeline/bit_view.rs
```diff
@@ -0,0 +1,606 @@
...
+
+use std::borrow::Cow;
+use std::fmt::{Debug, Formatter};
+
+use vortex_buffer::BitBuffer;
+use vortex_error::VortexResult;
+
+use crate::pipeline::{N, N_BYTES, N_WORDS};
+
+/// A borrowed fixed-size bit vector of length `N` bits, represented as an array of usize words.
```
gatesn, line 12: This comment is out of date now

vortex-array/src/pipeline/source_driver.rs
```diff
@@ -0,0 +1,131 @@
...
+                    let selection_view = BitView::with_prefix(remaining);
+
+                    let prev_len = output.len();
+                    kernel.step(&kernel_ctx, &selection_view, &mut output)?;
+                    debug_assert_eq!(output.len(), prev_len + remaining);
+                    debug_assert_eq!(output.len(), selection.len());
+                }
+            }
+            Mask::AllFalse(_) => {
+                // Select nothing, return empty output!
```
gatesn, line 73: We should do this check before allocating the output vector's capacity

vortex-buffer/src/buffer_mut.rs
```diff
@@ -241,10 +241,15 @@ impl<T> BufferMut<T> {
         }
     }
 
+    /// Sets the length of the buffer.
+    ///
     /// # Safety
-    /// The caller must ensure that the buffer was properly initialized up to `len`.
+    ///
+    /// The caller must ensure that there is sufficient capacity in the buffer and that the values
+    /// are valid up to `len`.
```
a10y, line 249:
```suggestion
    /// are initialized up to `len`.
```

vortex-vector/src/primitive/generic_mut.rs
```diff
@@ -73,6 +73,38 @@ impl<T> PVectorMut<T> {
         }
     }
 
+    /// Set the length of the vector.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the new length does not exceed the capacity of the vector.
+    pub unsafe fn set_len(&mut self, new_len: usize) {
+        debug_assert!(new_len < self.elements.capacity());
```
a10y, line 82: I wonder if we can just assert this, I think the unsafeness is more about uninitialized memory

vortex-array/src/pipeline/bit_view.rs
```diff
@@ -0,0 +1,606 @@
...
+/// the bit-ordering is LSB0 (least-significant-bit first).
+///
+/// Note that [`BitView`] does not support an offset. Therefore, bits are assumed to start at
+/// index and end at index `N - 1`.
+pub struct BitView<'a> {
+    bits: Cow<'a, [u8; N_BYTES]>,
+    // TODO(ngates): we may want to expose this for optimizations.
+    // If set to Selection::Prefix, then all true bits are at the start of the array.
+    // selection: Selection,
+    true_count: usize,
```
joseph-isaacs, line 24: should this move over to a BitBuffer in the future?

vortex-vector/src/primitive/generic_mut.rs
gatesn: I really really want this to compile down to a single store instruction. This is on a super hot loop

connortsui20 (approved):
I just want the new traits 😄 

(we also talked offline and we're just going to move forwards and fix things later)

vortex-array/src/pipeline/bit_view.rs
gatesn: Should what?

PR #5255, opened 2025-11-08
chore: Fix fuzzers baseline nullability for compare and take

robert3005:

### DIFF
--- a/fuzz/src/array/compare.rs
+++ b/fuzz/src/array/compare.rs
@@ -131,8 +131,7 @@ pub fn compare_canonical_array(
     }
 }
 
-#[allow(clippy::unwrap_used)]
-fn compare_to<T: PartialOrd + PartialEq + Debug>(
+fn compare_to<T: PartialOrd>(
     values: impl Iterator<Item = Option<T>>,
     cmp_value: T,
     operator: Operator,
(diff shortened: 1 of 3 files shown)


fuzz/src/array/compare.rs
```diff
@@ -125,56 +131,48 @@ pub fn compare_canonical_array(
     }
 }
 
+#[allow(clippy::unwrap_used)]
```
joseph-isaacs: shall we put this at the mod level?

fuzz/src/array/compare.rs
```diff
@@ -125,56 +131,48 @@ pub fn compare_canonical_array(
...
-        })
-    }))
-    .into_array()
+    let eval_fn = |v| match operator {
+        Operator::Eq => v == cmp_value,
+        Operator::NotEq => v != cmp_value,
+        Operator::Gt => v > cmp_value,
+        Operator::Gte => v >= cmp_value,
+        Operator::Lt => v < cmp_value,
+        Operator::Lte => v <= cmp_value,
```
joseph-isaacs, line 142: think we can share this with the below?

joseph-isaacs (approved):

fuzz/src/array/compare.rs
robert3005: We would need to wrap the NativePType into a newtype and implement the std traits.

robert3005: Turns out we have NativeValue for this use case. I have added PartialOrd impl for it which then lets us use generic version for everything

PR #5258, opened 2025-11-09
Fix: casting Zstd array to non nullable requires decompression

robert3005:


lwwmanning (approved):

encodings/zstd/src/compute/cast.rs
```diff
@@ -10,6 +10,11 @@ use crate::{ZstdArray, ZstdVTable};
 
 impl CastKernel for ZstdVTable {
     fn cast(&self, array: &ZstdArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
+        if !dtype.is_nullable() || !array.all_valid() {
```
a10y, line 13: should this be && ?

a10y: if dtype is non-nullable and array has no nulls then the cast should be trivial?

PR #5264, opened 2025-11-10
[DataFusion] Don't pushdown empty filters

AdamGS:
Taken out of #5205, I think we might hit [this](https://github.com/apache/datafusion/issues/18513) issue otherwise, and this also just make sense generally?

### DIFF
--- a/vortex-datafusion/src/persistent/source.rs
+++ b/vortex-datafusion/src/persistent/source.rs
@@ -230,6 +230,12 @@ impl FileSource for VortexSource {
         filters: Vec<Arc<dyn PhysicalExpr>>,
         _config: &ConfigOptions,
     ) -> DFResult<FilterPushdownPropagation<Arc<dyn FileSource>>> {
+        if filters.is_empty() {
+            return Ok(FilterPushdownPropagation::with_parent_pushdown_result(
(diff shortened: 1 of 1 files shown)


vortex-datafusion/src/persistent/source.rs
```diff
@@ -230,6 +230,12 @@ impl FileSource for VortexSource {
         filters: Vec<Arc<dyn PhysicalExpr>>,
         _config: &ConfigOptions,
     ) -> DFResult<FilterPushdownPropagation<Arc<dyn FileSource>>> {
+        if filters.is_empty() {
+            return Ok(FilterPushdownPropagation::with_parent_pushdown_result(
+                vec![],
```
onursatici, line 235: does this have the same meaning with `vec![PushedDown::No; filters.len()]`?

AdamGS: yes, when `filters.len() == 0`.

AdamGS: but we currently always modify the node (AKA - return a value using `with_updated_node` which this avoids.

onursatici (approved):

PR #5267, opened 2025-11-10
chore: serialize for `Mask`

blaginin:


vortex-mask/src/lib.rs
```diff
@@ -123,6 +126,27 @@ pub struct MaskValues {
     density: f64,
 }
 
+impl MaskValues {
+    #[cfg(feature = "serde")]
+    pub(crate) fn from_buffer(buffer: BitBuffer) -> Self {
+        let count = buffer.true_count();
+        Self::from_buffer_with_counts(buffer, count)
```
joseph-isaacs: should `true_count` should be stored?

joseph-isaacs (approved):

PR #5269, opened 2025-11-10
chore[array/expr]: move `vortex-expr` into `vortex_array::expr`

joseph-isaacs:
We hope to re-introduce this once expr doesn't depend on array.


vortex/src/lib.rs
```diff
@@ -4,16 +4,16 @@
 // https://github.com/rust-lang/cargo/pull/11645#issuecomment-1536905941
 #![doc = include_str!(concat!("../", env!("CARGO_PKG_README")))]
 
-pub use vortex_array::*;
-use vortex_expr::session::ExprSession;
+use vortex_array::expr::session::ExprSession;
+pub use vortex_array::{expr, *};
```
gatesn: Why does this need to change? Isn't expr already exported from array?

joseph-isaacs: good spot

gatesn (approved):

PR #5270, opened 2025-11-10
feat: impl `expand` for `Buffer` & `BufferMut`

0ax1:


vortex-compute/src/expand/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+/// Function for expanding values of `self` to the true positions of a mask.
+pub trait Expand {
+    /// The result type after expansion.
+    type Output;
+
+    /// Expands `self` using the provided mask.
+    ///
+    ///
+    /// The result will have length equal to the mask. All values of `self` are
+    /// then scattered to the true positions of the mask.
```
gatesn: Maybe describe whether the other values have any guarantee? They could be zero'd, they could be garbage. Either is fine IMO

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,175 @@
...
+        return Buffer::empty();
+    }
+
+    let mut buf_mut = buf.into_mut();
+    let mask_len = mask_values.len();
+    buf_mut.reserve(mask_len - buf_len);
+
+    // Expand to the new buffer size which is equals the length of the mask.
+    unsafe {
+        buf_mut.set_len(mask_len);
```
gatesn: Nit, but move the semi-colon outside the braces and it collapses to one line

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    let buf_slice = buf_mut.as_mut_slice();
+    let mut element_idx = buf_len;
+
+    // Pick the first value as a default value. The buffer is not empty, and we
+    // know that the first value is guaranteed to be initialized. By doing this
+    // T does does not require to implement `Default`.
```
gatesn: It's probably fine to require Default

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,175 @@
...
+            mask.true_count(),
+            self.len(),
+            "Expand mask true count must equal the buffer length"
+        );
+
+        match mask {
+            Mask::AllTrue(_) => self.clone(),
+            Mask::AllFalse(_) => Buffer::empty(),
+            Mask::Values(mask_values) => expand_indices(self.clone(), mask_values),
+        }
```
joseph-isaacs, line 52: self.clone() and expand?

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+    fn expand(self, mask: &Mask) -> Self::Output {
+        assert_eq!(
+            mask.true_count(),
+            self.len(),
+            "Expand mask true count must equal the buffer length"
+        );
+
+        match mask {
+            Mask::AllTrue(_) => self,
+            Mask::AllFalse(_) => Buffer::empty(),
```
gatesn: You can also just return `self`? Since true_count == self.len()

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+        );
+
+        match mask {
+            Mask::AllTrue(_) => {}
+            Mask::AllFalse(_) => self.clear(),
+            Mask::Values(mask_values) => {
+                let buf_len = self.len();
+                let mask_len = mask_values.len();
+
+                if buf_len == 0 {
```
gatesn: This can never be hit. self.len == true_count, which is handled as Mask::AllFalse

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+                    return;
+                }
+
+                self.reserve(mask_len - buf_len);
+
+                // SAFETY: We just reserved enough space above.
+                unsafe {
+                    self.set_len(mask_len);
```
gatesn: Move semi-colon out of braces

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+    buf_slice: &mut [T],
+    src_len: usize,
+    mask_values: &vortex_mask::MaskValues,
+) {
+    let mask_len = buf_slice.len();
+
+    // Pick the first value as a default value. The buffer is not empty, and we
+    // know that the first value is guaranteed to be initialized. By doing this
+    // T does not require to implement `Default`.
+    let pseudo_default_value = buf_slice[0];
```
gatesn, line 97: Yeah I think it's better to use Default here. Taking first element makes it less deterministic

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    let mut element_idx = src_len;
+
+    // Iterate backwards through the mask to avoid overwriting unprocessed elements.
+    for mask_idx in (src_len..mask_len).rev() {
+        if mask_values.value(mask_idx) {
```
gatesn, line 104: This is kinda slow. Don't need to fix it for now, but just leave a NOTE stating that it's slow and can be improved.

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    // Pick the first value as a default value. The source buffer is not empty.
+    let pseudo_default_value = src[0];
+
+    let src_len = src.len();
+    let mut element_idx = src_len;
+
+    // Iterate backwards through the mask to avoid any issues.
+    for mask_idx in (src_len..mask_len).rev() {
+        if mask_values.value(mask_idx) {
```
gatesn, line 149: Likewise here is slow

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+/// # Arguments
+///
+/// * `src` - The source slice containing elements to scatter
+/// * `mask` - The mask indicating where elements should be placed
+///
+/// # Returns
+///
+/// A new `Buffer<T>` with length equal to `mask.len()`, with elements from `src` scattered
+/// to positions marked true in the mask. Positions marked false can have arbitrary values.
+fn expand_into_new_buffer<T: Copy>(src: &[T], mask: &Mask) -> Buffer<T> {
```
gatesn: Just take MaskValues here since you always match on it above

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+    let src_len = src.len();
+    let mask_len = mask.len();
+
+    match mask {
+        Mask::AllTrue(_) => Buffer::from_trusted_len_iter(src.iter().copied()),
+        Mask::AllFalse(_) => Buffer::empty(),
+        Mask::Values(mask_values) => {
+            if src_len == 0 {
```
gatesn: Never hit

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,413 @@
...
+
+            let mut buf_mut = BufferMut::<T>::with_capacity(mask_len);
+
+            // SAFETY: We're preallocating the full target capacity.
+            unsafe {
+                buf_mut.set_len(mask_len);
+            }
+
+            let buf_slice = buf_mut.as_mut_slice();
+            scatter_into_slice_from(src, buf_slice, mask_values);
```
gatesn: You can probably just inline the BufferMut::with_capacity, set_len, and calling this function?

vortex-compute/src/expand/buffer.rs
0ax1: I could return `self.clone` but not `self`.

vortex-compute/src/expand/buffer.rs
0ax1: In `expand` we'll have 2 loops. `0..buf_len_source` and `buf_len_source..new_len`. Even if we initialize false positions for `buf_len_source..new_len` to default, we still might not want to do the extra work work of setting false positions in `0..buf_len_source` to default, as they are already initialized for the in-place case.

vortex-compute/src/expand/buffer.rs
0ax1: See https://github.com/vortex-data/vortex/pull/5270#discussion_r2514215931. For the inplace-case we'll then end up with a mix of default initialized and whatever was prev written to the buffer, unless we want to overwrite false positions that are already initialized to some value with the default value.

vortex-compute/src/expand/buffer.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    // Pick the first value as a default value. The buffer is not empty, and we
+    // know that the first value is guaranteed to be initialized. By doing this
+    // T does not require to implement `Default`.
+    let pseudo_default_value = buf_slice[0];
+
+    let mut element_idx = buf_len;
+
+    // Iterate backwards through the mask to avoid overwriting unprocessed elements.
+    for mask_idx in (buf_len..mask_len).rev() {
```
joseph-isaacs, line 102: if we had an iterator this would be fast.

joseph-isaacs (approved):

PR #5271, opened 2025-11-10
fix: compare for struct with constant items

a10y:
subtle bug introduced in #5142 where the nested elements are constant, would change the len of `lhs` and cause the assertion to fail. Fix is to assert the length earlier.

### DIFF
--- a/vortex-layout/src/layouts/struct_/reader.rs
+++ b/vortex-layout/src/layouts/struct_/reader.rs
@@ -261,6 +261,7 @@ impl LayoutReader for StructReader {
         }
     }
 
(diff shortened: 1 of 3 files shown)


vortex-layout/src/layouts/struct_/reader.rs
```diff
@@ -261,6 +261,7 @@ impl LayoutReader for StructReader {
         }
     }
 
+    #[allow(clippy::unwrap_in_result)]
```
robert3005: don't need this anymore.

robert3005 (approved):
Remove clippy suppression

PR #5272, opened 2025-11-10
ALP Pipeline

gatesn:


encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,120 @@
...
+                    }))
+                })
+            }
+            Some(patches) => {
+                let patch_idxs = ctx.batch_input(0).into_primitive();
+                let patch_vals = ctx.batch_input(1).into_primitive();
+
+                match_each_alp_float_ptype!(self.ptype(), |A| {
+                    match_each_integer_ptype!(patches.indices_ptype(), |P| {
+                        let patch_indices: Buffer<P> = P::downcast(patch_idxs).into_buffer();
```
gatesn: The problem is, these patch indices reference row offsets into the entire array.

If a pipeline drives source nodes until it fills up 1k, then passes the 1k elements to this node, then we have no nice way of knowing which rows need patching.

Three options I can immediately think of:
1. We _don't_ step source nodes repeatedly to fill up a chunk. Instead, we run one step of the source and pass that vector through the entire pipeline, regardless of how sparse it is. e.g. it may only contain 1 element of the 1024. Each transform node would only process the elements that exist, but we may still have high per-chunk overheads. We would of course entirely skip empty chunks. But very sparse uniformly distributed selection masks would be the worst case.
2. We provide the selection mask to every node during the bind phase. They can push it down / store it / whatever. The kernel would then have to count how many elements it has seen, align this to true values of the mask, and use that to index into the patch indices.
3. We do not pass in a selection mask at all. Instead, we have a FilterArray and we allow nodes to push it down in the optimizer. We initially didn't like this because every source node would need a filtered/unfiltered variant.

I _think_ for our GPU compute, we would want to do option (1). Essentially, not do selection push-down at all (since the un-selected values are "free" in the GPU). 

Alternatively, we accept that we need our arrays to represent a logical plan, and then bind into separate physical plans for CPU vs GPU with separate optimization rules. I just worry this is _too much_ work to do on a per-batch basis during the scan.

robert3005: Chunk offsets let you relativise patch indices to 1024 element chunk

gatesn: I'm saying that's only useful for option (1) where we _preserve_ the 1024 element chunks. 

But on a CPU, if we push-down selection masks all the way. Then why not just emit dense 1024 element chunks. In which case we need our own book-keeping for where we are within the patches.

robert3005: I think even for option (2) you have your patches kernel use offsets to quickly apply selection mask to patches

encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,120 @@
...
+            .into_buffer();
+
+        let mut decoded = out.into_primitive().downcast::<A>();
+        decoded.extend(
+            encoded
+                .iter()
+                .map(|encoded| A::decode_single(*encoded, self.exponents)),
+        );
+
+        // Errrrrrr what patches do we apply?
```
robert3005, line 99: you call this https://github.com/vortex-data/vortex/blob/32d083480896f402f40888898ed060a0881ac559/vortex-array/src/arrays/primitive/array/patch.rs#L74?

PR #5273, opened 2025-11-10
chore: constrain the lifetime of SessionExt::get

danking:
Seems pretty reasonable that it should live at least as long as the Session itself, right?

### DIFF
--- a/vortex-array/src/expr/session.rs
+++ b/vortex-array/src/expr/session.rs
@@ -74,7 +74,7 @@ impl Default for ExprSession {
 /// Extension trait for accessing expression session data.
 pub trait ExprSessionExt: SessionExt {
     /// Returns the expression vtable registry.
-    fn expressions(&self) -> Ref<'_, ExprSession> {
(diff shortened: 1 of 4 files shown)


a10y:
What is the lifetime being inferred to before this change?

danking:
I do not know.

danking:
Maybe it's correct? This is primarily motivated by my looking at the API and trying to understand what to do with the returned object.

a10y:
Yea I'm pretty sure the lifetime is already being bound by the lifetime of `&self`, in this case? The lifetime can be elided to `'_` because it's unambiguous (there's only one reference in the parameter list)

gatesn:
The problem is the dashmap. So the lifetime is actually tied to the locked bucket of the dashmap, rather than just self.

I have a PR in progress that removes dashmap in favor of a session builder

danking:
Closing to move off my queue. We can reopen if there's renewed interest.

PR #5276, opened 2025-11-11
fix: Decimal sum is checked

robert3005:
Let's make it consistent


joseph-isaacs:
Moving this into a larger refactor, I will close this once that lands

joseph-isaacs:
I include this here: https://github.com/vortex-data/vortex/pull/5278

PR #5278, opened 2025-11-11
fix[array]: sum with initial value to fix op assoc

joseph-isaacs:


fuzz/fuzz_targets/array_ops.rs
```diff
@@ -89,6 +89,8 @@ fuzz_target!(|fuzz_action: FuzzArrayAction| -> Corpus {
                 current_array = cast_result;
             }
             Action::Sum => {
+                println!("sum {}", current_array.display_tree());
+                println!("sum {}", current_array.display_values());
```
a10y: do we want to keep these

a10y:
This is the thing the fuzzer is currently failing on right?

vortex-array/src/compute/sum.rs
```diff
@@ -1,19 +1,30 @@
...
 use vortex_scalar::Scalar;
 
 use crate::Array;
-use crate::compute::{ComputeFn, ComputeFnVTable, InvocationArgs, Kernel, Output, UnaryArgs};
+use crate::compute::{ComputeFn, ComputeFnVTable, InvocationArgs, Kernel, Options, Output};
 use crate::stats::{Precision, Stat, StatsProvider};
 use crate::vtable::VTable;
 
+#[derive(Debug, Clone)]
+pub struct SumOptions {
```
robert3005: This should be an input not an option

robert3005:
Yes, checked sum is not associative

vortex-array/src/arrays/chunked/compute/sum.rs
```diff
@@ -39,8 +38,11 @@ impl SumKernel for ChunkedVTable {
 
 register_kernel!(SumKernelAdapter(ChunkedVTable).lift());
 
-fn sum_int<T: NativePType + PrimInt>(chunks: &[ArrayRef]) -> VortexResult<Option<T>> {
-    let mut result: T = T::zero();
+fn sum_int<T: NativePType + CheckedAdd>(
+    chunks: &[ArrayRef],
+    initial_value: T,
+) -> VortexResult<Option<T>> {
+    let mut result: T = initial_value;
```
a10y: we don't want to push this into the chunk sums like we do for decimal?

joseph-isaacs:
@claude are the sum docs correct wrt the impl

joseph-isaacs:
@claude can you create an issue with this sum_with_initial doc change so I can follow up?

vortex-array/src/compute/sum.rs
```diff
@@ -107,31 +148,32 @@ impl<V: VTable + SumKernel> SumKernelAdapter<V> {
...
 /// Sum an array.
 ///
 /// If the sum overflows, a null scalar will be returned.
 /// If the sum is not supported for the array's dtype, an error will be raised.
-/// If the array is all-invalid, the sum will be zero.
+/// If the array is all-invalid, the sum will be the accumulator.
 pub fn sum_impl(
     array: &dyn Array,
-    sum_dtype: DType,
+    _sum_dtype: DType,
```
robert3005: should we remove the sum_dtype? It's contained in the accumulator

robert3005 (approved):

vortex-array/src/arrays/constant/compute/sum.rs
```diff
@@ -12,32 +13,44 @@ use crate::register_kernel;
...
+                .as_::<u64>()
+                .vortex_expect("cannot be null");
+            Ok(ScalarValue::from(accumulator.checked_add(count)))
+        }
+        DType::Primitive(ptype, _) => {
+            let result = match_each_native_ptype!(
+                ptype,
+                unsigned: |T| { sum_integral::<u64>(scalar.as_primitive(), len, accumulator)?.into() },
+                signed: |T| { sum_integral::<i64>(scalar.as_primitive(), len, accumulator)?.into() },
+                floating: |T| { sum_float(scalar.as_primitive(), len, accumulator)?.into() }
```
connortsui20, line 46: The `T` here is unused, is this intentional?

joseph-isaacs: I remember this being different, what is the return type of the sum_float?

connortsui20: code is below, its option f64. But more importantly there is a hardcoded u64 and i64 here

PR #5284, opened 2025-11-11
API and partial implementation of sparse pipelined exection

gatesn:


a10y (approved):
let's merge and iterate

vortex-array/src/pipeline/driver/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    kernels: Vec<Box<dyn Kernel>>,
+    exec_order: Vec<NodeId>,
+    output_targets: Vec<OutputTarget>,
+}
+
+impl Pipeline {
+    fn execute(&mut self, selection: &Mask) -> VortexResult<Vector> {
+        // Start by allocating the output vector.
+        let capacity = selection.true_count().next_multiple_of(N);
+        let mut output = VectorMut::with_capacity(&self.dtype, capacity);
```
connortsui20, line 241: I think we should be only allocating up to `N` now since we know that we are only writing that much. For dense vectors in the future, we can make this `2 * N` and shift everything back after a stage has been processed.

Otherwise I think we waste a lot of memory

PR #5286, opened 2025-11-11
Clarify calling convention for pipeline kernels

gatesn:
And implement a zero-copy input kernel

### DIFF
--- a/vortex-array/src/pipeline/mod.rs
+++ b/vortex-array/src/pipeline/mod.rs
@@ -59,21 +59,27 @@ pub trait BindContext {
     ///
     /// Note that this child index references the batch inputs only, not all children of the
     /// array.
-    fn batch_input(&self, batch_child_idx: usize) -> Vector;
+    fn batch_input(&mut self, batch_child_idx: usize) -> Vector;
 }
 
 /// A pipeline kernel is a stateful object that performs steps of a pipeline.
 ///
-/// Each step of the kernel takes and returns vectors (depending on the type of kernel) of `N`
-/// elements. Input vectors are provided via the [`KernelCtx`] and indicate the position of their
-/// elements as either [`PipelineVector::Sparse`] or [`PipelineVector::Compact`] based on whether
+/// Each step of the kernel processes zero or more input vectors, and writes output to a
+/// pre-allocated mutable output vector.
+///
+/// Input vectors are provided via the [`KernelCtx`] and indicate the position of their elements
+/// as either [`PipelineVector::InPlace`] or [`PipelineVector::Compact`] based on whether
 /// the selected elements are in their original positions or compacted at the start of the vector
 /// respectively.
 ///
 /// The provided mutable output vector is guaranteed to have at least `N` elements of capacity.
-/// The kernel **must** return a vector of exactly `N` elements in each step, and indicate with
-/// the returned [`ElementPosition`] enum whether the output elements are in their original
-/// positions or compacted at the start of the output vector.
+/// The kernel **must** append either [`BitView::true_count`] elements to the output vector (in
+/// which case the output elements are considered to be in the "Compact" position), or it must
+/// append `N` elements (in which case the output elements are considered to be in their "Flat"
+/// positions). The pipeline driver will assert these conditions after each step.
+///
+/// Note that the output vector may not be empty at the start of the step. The kernel must append
+/// its output to the existing contents of the output vector, rather than replacing it.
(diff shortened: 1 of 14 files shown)


vortex-array/src/pipeline/mod.rs
```diff
@@ -59,37 +59,35 @@ pub trait BindContext {
...
-    ) -> VortexResult<ElementPosition>;
-}
-
-/// Defines where the elements produced by a kernel are written in the output vector.
-pub enum ElementPosition {
-    /// Elements are written to the output vector in their original selected positions.
-    Sparse,
-    /// Elements are compacted at the start of the output vector.
-    Compact,
+    ) -> VortexResult<()>;
```
connortsui20, line 90: I think that forcing the kernel to return what they did is nice from an implementors perspective, it makes it obvious that the kernel is going to do something different based on what we return.

connortsui20 (changes requested):

vortex-array/src/pipeline/mod.rs
connortsui20: Additionally, the `step` function probably needs to be `unsafe` unless we are ok `assert!`ing that we have enough capacity in each kernel.

gatesn: I disagree. Partly because there's an overloaded case where true_count == N. In that case, did I return InPlace? Or Compact? Well.... both? So this API avoids the ambiguity and says that N implies InPlace always. I get that not being super explicit is maybe a bit weird though.

The step function does not need to be unsafe. The caller guarantees the capacity. set_len or any other unsafe things can also debug_assert the capacity anyway.

connortsui20: Then just return 3 variants? Could have a `NoChange` variant for example?

connortsui20 (approved):
we can update things later

vortex-array/src/pipeline/mod.rs
```diff
@@ -59,37 +59,35 @@ pub trait BindContext {
...
 /// the selected elements are in their original positions or compacted at the start of the vector
 /// respectively.
 ///
 /// The provided mutable output vector is guaranteed to have at least `N` elements of capacity.
-/// The kernel **must** return a vector of exactly `N` elements in each step, and indicate with
-/// the returned [`ElementPosition`] enum whether the output elements are in their original
-/// positions or compacted at the start of the output vector.
+/// The kernel **must** append either [`BitView::true_count`] elements to the output vector (in
+/// which case the output elements are considered to be in the "Compact" position), or it must
+/// append `N` elements (in which case the output elements are considered to be in their "Flat"
```
joseph-isaacs: Flat is confusing here. I can we use a different work, uncompact?

joseph-isaacs: Do we mean `InPlace`?

vortex-array/src/pipeline/mod.rs
```diff
@@ -59,37 +59,35 @@ pub trait BindContext {
...
+/// positions). The pipeline driver will assert these conditions after each step.
+///
+/// Note that the output vector may not be empty at the start of the step. The kernel must append
+/// its output to the existing contents of the output vector, rather than replacing it.
```
joseph-isaacs, line 82: Does this handle external vectors? It odd that I am not give a view?

gatesn: Yeah, the append is for the external vector.

We could if you prefer use split_off/unsplit to give a slice of the output vector, then zero-copy stitch it back together.

There's _some_ overhead to that, it's not clear yet how much. But it would mean everyone gets given a vector with length = 0 so it might be nicer.

vortex-array/src/pipeline/mod.rs
joseph-isaacs: I agree an enum would be nice here.

vortex-array/src/pipeline/mod.rs
gatesn: Yeah, InPlace. Or maybe even just Full.

To avoid UB, I think you kind of have to guarantee writing N valid values. Even if you filled the empty spots with T::default right?

vortex-array/src/pipeline/mod.rs
joseph-isaacs: I don't see why we cannot keep the two variants? Are you saying the enum allows the implementor to choose the suboptimial one?

PR #5287, opened 2025-11-11
Feature: add `spare_capacity_mut` to `PVectorMut`

connortsui20:
Tracking issue: https://github.com/vortex-data/vortex/issues/5028

### DIFF
--- a/vortex-vector/src/primitive/generic_mut_impl.rs
+++ b/vortex-vector/src/primitive/generic_mut_impl.rs
@@ -3,6 +3,8 @@
 
 //! Helper methods for [`PVectorMut<T>`] that mimic the behavior of [`std::vec::Vec`].
 
+use std::mem::MaybeUninit;
+
 use vortex_buffer::BufferMut;
 use vortex_dtype::NativePType;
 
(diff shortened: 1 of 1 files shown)


vortex-vector/src/primitive/generic_mut_impl.rs
```diff
@@ -131,6 +133,24 @@ impl<T: NativePType> PVectorMut<T> {
...
+    /// initialized using unsafe methods like [`set_len`].
+    ///
+    /// # Safety
+    ///
+    /// Note that this only provides access to the spare capacity of the **elements** buffer.
+    ///
+    /// After writing to the spare capacity and calling [`set_len`], the caller must also ensure the
+    /// validity mask is updated accordingly to maintain consistency.
+    ///
+    /// [`set_len`]: Self::set_len
```
gatesn: I really really dislike this pattern of docs. Can you link them in-place

vortex-vector/src/primitive/generic_mut_impl.rs
```diff
@@ -131,6 +133,24 @@ impl<T: NativePType> PVectorMut<T> {
...
+    #[inline]
+    pub unsafe fn spare_capacity_mut(&mut self) -> &mut [MaybeUninit<T>] {
```
gatesn: Do we need this? MaybeUninit is a really horrible API to work with, particularly for primitives where just calling reserve + set_len + as_mut seems to be the way most things go

vortex-vector/src/primitive/generic_mut_impl.rs
connortsui20: This follows the style of the standard library: https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3065

vortex-vector/src/primitive/generic_mut_impl.rs
connortsui20: The issue is that this pattern is wrong (or if it isn't wrong, it is prone to being wrong).

See: https://users.rust-lang.org/t/is-this-correct-way-to-write-uninitialized-bytes/90613/2 and all the comments after it.

To summarize, there is a consensus around how to work with uninitialized memory, but no guarantee. So people who work with uninitialized memory generally would expect a method like this to be available.

Our wide use of transmute is actually problematic because this _could_ change. It probably won't, but we don't know that.

connortsui20:
we can tidy up our unsafe code later

connortsui20:
now I actually need it for adapting the existing bit unpack code to vectors instead of builders

gatesn (approved):

PR #5288, opened 2025-11-12
fix[vortex-dict]: faster min_max for dict array

joseph-isaacs:


joseph-isaacs:
LGTM, but I cannot approve

0ax1 (approved):

PR #5289, opened 2025-11-12
move vortex-dict into vortex-array

onursatici:
I think this is the lowest lift change to enable arrow -> vortex conversions for dict arrays.
This PR only moves the vortex-dict crate into vortex-array, first commit only renames, second fixes imports.

with this vortex-array maybe grows 5% in size. looking at crates.io I don't think vortex-dict is used separately from vortex-array much, so I removed the vortex-dict crate instead of re-exporting from vortex-array.

I think the change is minimal enough to be reverted when we extract arrow out from vortex-array, happy to hear peoples thoughts. With this change we have all arrow supported encodings live in vortex-array


encodings/fsst/Cargo.toml
```diff
@@ -38,3 +42,8 @@ vortex-array = { workspace = true, features = ["test-harness"] }
 [[bench]]
 name = "fsst_compress"
 harness = false
+
+[[bench]]
+name = "chunked_dict_array_builder"
```
onursatici: I should split this into two benches, and move the non fsst dict builder benchmarks to vortex-array

gatesn (approved):
Sadly I think this is the right way forwards for now. Once compute is absorbed by the operators RFC, we won't have an Arrow dependency from vortex-array and we can pull this back out and have a separate vortex-arrow crate for _all_ the conversion logic

PR #5290, opened 2025-11-12
ALP pipeline kernel

gatesn:
A WIP just to experiment with the kernel APIs


encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl<A: ALPFloat> Kernel for UnpatchedALPKernel<A> {
+    fn step(
+        &mut self,
+        ctx: &KernelCtx,
+        selection: &BitView,
+        out: &mut VectorMut,
+    ) -> VortexResult<()> {
+        if selection.true_count() == 0 {
```
gatesn, line 61: I actually don't hate this. Because it allows kernels that _do_ have mutable internal state to keep track of how many rows they've seen.

encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,117 @@
...
+                    decoded.extend(
+                        encoded_buf
+                            .iter()
+                            .map(|e| A::decode_single(*e, self.exponents)),
+                    );
+                }
+            }
+            Position::Compact => {
+                // Loop over the compacted input elements
+                decoded.extend(
```
gatesn: Compact and "dense" InPlace both have the same code...

It hasn't really been documented, but not all input vectors have length == N. For those with len < N, we _know_ that the elements are compacted.

For those with len == N, we know that the elements are necessarily in-place.

So I wonder if we simply get rid of the Position enum entirely and we just look at input length + selection mask?

encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,117 @@
...
+        let encoded_buf = encoded_vec.elements().as_slice();
+
+        let decoded = out.as_primitive_mut().downcast::<A>();
+
+        match encoded.position() {
+            Position::InPlace => {
+                // TODO(ngates): tune the threshold
+                if selection.true_count() < (N / 8) {
+                    // Operate only over the selected elements, appending `true_count` elements
+                    unsafe {
```
gatesn: We could just use `decoded.append_nulls(selection.true_count())` here. It would remove a few lines of unsafe and we simply pay the cost to initialize the values buffer to zero.

encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,117 @@
...
+                        decoded
+                            .validity_mut()
+                            .append_n(true, selection.true_count())
+                    };
+                    unsafe { decoded.elements_mut().set_len(selection.true_count()) };
+                    let decoded_buf = unsafe { decoded.elements_mut() };
```
gatesn: This assumes that output.len() == 0 at the start of the kernel. That's currently not true. I think it should be true and we pay the book keeping cost of split/unsplit when appending pipeline outputs to large vectors.

encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,167 @@
...
+        if encoded.len() < N {
+            sparse_alp(decoded_vec, encoded.as_slice(), self.exponents, selection)
+        }
+
+        debug_assert_eq!(encoded.len(), N);
+        debug_assert_eq!(decoded_vec.len(), N);
+
+        // Otherwise, iterate the entire input.
+        decoded_vec.reserve(N.saturating_sub(decoded_vec.capacity()));
+        unsafe { decoded_vec.validity_mut().append_n(true, N) };
```
connortsui20: This seems wrong? Shouldnt this copy the input vector validity?

encodings/alp/src/alp/operator.rs
```diff
@@ -0,0 +1,167 @@
...
+) {
+    // Reserve capacity for the true_count elements.
+    decoded_vec.reserve(
+        selection
+            .true_count()
+            .saturating_sub(decoded_vec.capacity()),
+    );
+
+    // SAFETY: we set_len and append_validity ensuring elements len matches validity len.
+    unsafe { decoded_vec.validity_mut() }.append_n(true, selection.true_count());
```
connortsui20: Same here?

PR #5291, opened 2025-11-12
Pipeline Calling Convention 2

gatesn:
* Guarantees that output vectors have length == 0
* Does not guarantee capacity.
* Inputs are either N or n elements, and defines that in-place but non-selected elements _must_ be valid.


connortsui20 (approved):
To be honest, this feels like we are removing a lot of type information in return for less verbosity, which I usually try to avoid. But for the sake of making progress, we should just merge this.

PR #5295, opened 2025-11-12
fix[vortex-datafusion]: check field exists in get_field pushdown

asubiotto:
Previously, any get field expression was pushed down, which resulted in incorrectly pushing down get fields on non-existent fields, causing an error at execution time rather than plan time.

This was a mistake on my part that should've been included in #5024


vortex-datafusion/src/convert/exprs.rs
```diff
@@ -277,6 +276,53 @@ fn supported_data_types(dt: &DataType) -> bool {
...
+    let DataType::Struct(fields) = source_dt else {
+        tracing::debug!(
+            field_name = field_name,
+            schema = ?schema,
+            source_expr = ?source_expr,
+            "Failed to get source type as struct for GetField, not pushing down"
+        );
+        return false;
+    };
+    fields.find(field_name).is_some()
```
asubiotto, line 323: @AdamGS let me know if there's any other validation you think we should be doing

AdamGS: this seems pretty solid

AdamGS (approved):

PR #5299, opened 2025-11-12
Feature: add aligned bitpacked pipeline kernel

connortsui20:
Adds an `AlignedBitPackedKernel` that implements bit unpacking as a pipeline source.

Note that this completely ignores patches for now, since we probably want to implement patches as a transform node on top of this instead.


encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
```diff
@@ -0,0 +1,288 @@
...
+
+                // SAFETY: We just reserved enough capacity to push these values.
+                unsafe { output_vector.push_unchecked(unpacked_value) };
+            });
+        } else {
+            // Otherwise if the mask is dense, it is faster to fully unpack the entire 1024
+            // element lane with SIMD / FastLanes and let other nodes in the pipeline decide if they
+            // want to perform the selection filter themselves.
+            output_vector.reserve(N);
+            debug_assert!(N <= output_vector.capacity());
```
a10y, line 150: is this really necessary? this is just testing the correctness of `reserve`

a10y (approved):

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
connortsui20: Not really, I usually just like to have it to make reasoning about the safety contracts better

connortsui20:
benchmarks in the followup PR

PR #5301, opened 2025-11-12
Feature: implement filter for primitive vector (and mask)

connortsui20:
Implements filter for primitive vectors (which needed a filter implementation for `MaskMut`, and I was too lazy to figure out how to do it in place).

### DIFF
--- a/vortex-compute/src/filter/vector/primitive.rs
+++ b/vortex-compute/src/filter/vector/primitive.rs
@@ -2,15 +2,37 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use vortex_mask::Mask;
-use vortex_vector::match_each_pvector_mut;
-use vortex_vector::primitive::PrimitiveVectorMut;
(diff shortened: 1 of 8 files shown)


vortex-compute/src/filter/vector/primitive.rs
```diff
@@ -2,15 +2,37 @@
...
     fn filter(self, selection_mask: &Mask) {
         match_each_pvector_mut!(self, |v| { v.filter(selection_mask) })
     }
 }
+
+impl Filter<Mask> for PrimitiveVector {
+    type Output = Self;
+
+    fn filter(self, selection_mask: &Mask) -> Self {
+        assert_eq!(
```
gatesn: I don't think there's any point doing this assertion here? Presumably the PVector<T> will do it

connortsui20: forgot to delete

gatesn (approved):

a10y (changes requested):
https://github.com/vortex-data/vortex/pull/5333

PR #5303, opened 2025-11-12
chore: make ArrayAccessor infallible

a10y:
Long overdue follow up from #4177 / #4412

There was no reason why ArrayAccessor needed to return an error, so we just remove the dummy `Ok()` wrapping and update the types.

Turns out this was quite infectious and also allows us to cleanup a bunch of things with FSST and DictEncoder

### DIFF
--- a/vortex-array/src/accessor.rs
+++ b/vortex-array/src/accessor.rs
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use vortex_error::VortexResult;
-
 /// Trait for arrays that support iterative access to their elements.
 pub trait ArrayAccessor<Item: ?Sized> {
     /// Iterate over each element of the array, in-order.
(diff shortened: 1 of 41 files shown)


vortex-array/src/accessor.rs
```diff
@@ -1,16 +1,14 @@
...
-
 /// Trait for arrays that support iterative access to their elements.
 pub trait ArrayAccessor<Item: ?Sized> {
     /// Iterate over each element of the array, in-order.
     ///
     /// The function `f` will be passed an [`Iterator`], it can call [`next`][Iterator::next] on the
     /// iterator [`len`][crate::Array::len] times. Iterator elements are `Option` types,
     /// regardless of the nullability of the underlying array data.
-    fn with_iterator<F, R>(&self, f: F) -> VortexResult<R>
+    fn with_iterator<F, R>(&self, f: F) -> R
```
a10y, line 11: Note that the inner type `R` can _still_ be `VortexResult`, this outer result wrapping was from the time when Array Ops used to be fallible. But all arrays are checked at the boundaries and in practice none of the implementations ever returned Err, they just did dumb `Ok(...)` wrapping

AdamGS (commented):
awesome

AdamGS (approved):
Now with the actual approval

PR #5304, opened 2025-11-12
Implement missing FFI functions for array writing

### LINKED ISSUE #5275: FFI API Create array
According to the following it should be possible to create  a new vx_array with vx_array_primitive_new. However no function with that name exists. This might just be an error. Otherwise im curious about how im supposed to create new arrays when writing a vortex file using the FFI API. https://github.com/vortex-data/vortex/blob/32d083480896f402f40888898ed060a0881ac559/vortex-ffi/cinclude/vortex.h#L232-L243

a10y:
Implements FFI builder functions for creating various Vortex array types:

- **Primitive arrays**: Support for all primitive types (u8, u16, u32, u64, i8, i16, i32, i64, f16, f32, f64) with optional validity masks
- **Bool arrays**: Boolean array creation with validity support
- **Decimal arrays**: Decimal128 array creation with precision and scale
- **Null arrays**: Simple null array construction
- **UTF8/Binary arrays**: Builder-pattern API for variable-length string and binary data
- **Struct arrays**: Construction from field arrays and struct dtype

Adds `write-vortex.c` example demonstrating:
- Creating arrays of different types using the new builder APIs
- Building complex struct arrays with multiple fields
- Writing arrays to Vortex files using the sink API
- Proper resource management and error handling

Resolves #5275

### DIFF
--- a/vortex-ffi/examples/write-vortex.c
+++ b/vortex-ffi/examples/write-vortex.c
@@ -1,0 +1,276 @@
+// SPDX-License-Identifier: CC-BY-4.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#include "vortex.h"
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+int main(int argc, char *argv[]) {
+  if (argc < 2) {
+    printf("Usage: %s <OUTPUT_FILE>\n", argv[0]);
+    printf("  Creates a Vortex file with example data demonstrating all array "
+           "types.\n");
+    return 1;
+  }
+
+  vx_error *error = NULL;
+  vx_session *session = vx_session_new();
+  if (session == NULL) {
+    fprintf(stderr, "Failed to create Vortex session\n");
+    return -1;
+  }
+
+  char *output_path = argv[1];
+  printf("Creating Vortex file: %s\n\n", output_path);
+
+  // ==========================================================================
+  // Create example struct with various array types
+  // ==========================================================================
+
+  // 1. Create a primitive i32 array for "age" field
+  printf("Creating i32 array (ages)...\n");
+  int32_t ages_data[] = {25, 30, 35, 40, 45};
+  bool ages_validity[] = {true, true, false, true, true}; // 3rd value is null
+  size_t row_count = sizeof(ages_data) / sizeof(ages_data[0]);
+
+  const vx_array *ages_array = vx_array_primitive_new_i32(
+      ages_data, row_count, ages_validity, &error);
+  if (error != NULL) {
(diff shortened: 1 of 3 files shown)


vortex-ffi/examples/write-vortex.c
```diff
@@ -0,0 +1,276 @@
...
+
+  // ==========================================================================
+  // Create example struct with various array types
+  // ==========================================================================
+
+  // 1. Create a primitive i32 array for "age" field
+  printf("Creating i32 array (ages)...\n");
+  int32_t ages_data[] = {25, 30, 35, 40, 45};
+  bool ages_validity[] = {true, true, false, true, true}; // 3rd value is null
+  size_t row_count = sizeof(ages_data) / sizeof(ages_data[0]);
```
danking, line 36: I wasn't aware you could sizeof an expression! Does it have to be a valid index? does `sizeof ages_data[10]` work?

vortex-ffi/src/array_builder.rs
```diff
@@ -0,0 +1,468 @@
...
+/// The `validity` pointer, if not null, must point to a valid array of `len` booleans.
+/// If `validity` is null, the array is assumed to have no null values.
+///
+/// Returns a new array, or null on error.
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_array_bool_new(
+    data: *const bool,
+    len: usize,
+    validity: *const bool,
+    error_out: *mut *mut vx_error,
```
danking, line 93: the need for this has to be my least favorite part of C

vortex-ffi/src/array_builder.rs
```diff
@@ -0,0 +1,468 @@
...
+            Nullability::NonNullable
+        } else {
+            Nullability::Nullable
+        };
+
+        let mut builder = BoolBuilder::with_capacity(nullability, len);
+
+        for (i, &value) in data_slice.iter().enumerate() {
+            if !validity.is_null() {
+                let validity_slice = unsafe { slice::from_raw_parts(validity, len) };
```
danking, line 112: Is this free? It seems at least for legibility's sake, we should lift it maybe to line 101?

a10y:
This was fully Claude-coded and has not yet been reviewed by me, I'll invest some time in this today to hopefully get it ready to be a real PR

vortex-ffi/src/array_builder.rs
danking: ohhhhhhh ass, right, pointers. Can we switch on null-ness here and have two loops? That avoids the branch in the loop in the non-null case.

frederikja163:
This PR is still missing the header file declarations for cinlude.

PR #5330, opened 2025-11-13
Fix: handle validity in bitpack pipeline kernel

connortsui20:
Forgot to handle this before


encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
```diff
@@ -127,20 +138,26 @@ impl<BP: PhysicalPType<Physical: BitPacking>> Kernel for AlignedBitPackedKernel<
...
+                            self.packed_bit_width,
+                            not_yet_unpacked_values,
+                            idx,
+                        )
+                    };
+
+                    // SAFETY: We just reserved enough capacity to push these values.
+                    unsafe { output_vector.push_unchecked(unpacked_value) };
+                } else {
+                    output_vector.push_opt(None);
```
gatesn: This seems like a weird function to call. It's invoked inside the else statement so we know it's None. But presumably inside this function you have _another_ if/else statement to check the option.

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
```diff
@@ -149,16 +166,17 @@ impl<BP: PhysicalPType<Physical: BitPacking>> Kernel for AlignedBitPackedKernel<
             output_vector.reserve(N);
```
gatesn, line 168: This will always reserve an additional N, even if there's already N.

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
```diff
@@ -171,6 +189,14 @@ impl<BP: PhysicalPType<Physical: BitPacking>> Kernel for AlignedBitPackedKernel<
                     &mut output_vector.as_mut()[..FL_VECTOR_SIZE],
```
gatesn: You should add static_assertions crate and assert FL_VECTOR_SIZE == N

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
```diff
@@ -81,12 +83,19 @@ pub struct AlignedBitPackedKernel<BP: PhysicalPType<Physical: BitPacking>> {
     /// The buffer containing the bitpacked values.
     packed_buffer: Buffer<BP::Physical>,
 
+    /// The validity mask for the bitpacked array.
+    validity: Validity,
```
gatesn: This is a weird thing to hold on the kernel. You probably want a Mask.

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
connortsui20: that is already at the top

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
connortsui20: I thought the length of the vector is always 0? so this becomes a noop

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
connortsui20: I could add add a `push_unchecked_opt`?

gatesn: Isn't this just `append_nulls(1)`?

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
gatesn: Oh yeah.... my understanding of reserve is wrong!

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
gatesn: Then you don't need this slice?

connortsui20: ah right, this no longer handles arbitrary N

gatesn (approved):

PR #5333, opened 2025-11-13
chore: implement Filter for new MaskIndices type for all vectors

a10y:


vortex-compute/src/filter/vector/binaryview.rs
```diff
@@ -0,0 +1,39 @@
...
+                unsafe {
+                    self.views_mut().filter(selection);
+                    self.validity_mut().filter(selection);
+                }
+            }
+        }
+    };
+}
+
+delegate_filter_impl!(Mask);
```
gatesn: Surely we don't need macros here? Can you define a new mask-y trait and impl for that instead?

a10y:
yes i will fix the license headers

vortex-compute/src/filter/vector/binaryview.rs
a10y: switched over to existential types version

a10y: err, higher-ranked types

vortex-compute/src/filter/buffer.rs
```diff
@@ -32,7 +32,15 @@ impl<T: Copy> Filter for &Buffer<T> {
     }
 }
 
-impl<T: Copy> Filter for &mut BufferMut<T> {
+impl<T: Copy> Filter<MaskIndices<'_>> for &Buffer<T> {
+    type Output = Buffer<T>;
+
+    fn filter(self, indices: &MaskIndices) -> Buffer<T> {
+        filter_indices(self, indices)
```
connortsui20, line 39: assert that the lengths are the same?

vortex-compute/src/filter/mod.rs
```diff
@@ -22,5 +22,37 @@ pub trait Filter {
...
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the indices are strict-sorted, i.e. that they
+    /// are monotonic and unique.
+    ///
+    /// Users of the `Indices` type assume this and failure to uphold this guarantee
+    /// can result in UB downstream.
+    pub unsafe fn new_unchecked(indices: &'a [usize]) -> Self {
+        Self(indices)
```
connortsui20, line 48: You can add a `debug_assert!(indices.is_sorted())`

connortsui20 (approved):

connortsui20:
nice!

Maybe I'm weird but I actually found the generic version to be more readable because it tells me exactly what is getting filtered

vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -33,6 +33,14 @@ impl Filter for &BitBuffer {
     }
 }
 
+impl Filter<MaskIndices<'_>> for &BitBuffer {
+    type Output = BitBuffer;
+
+    fn filter(self, indices: &MaskIndices) -> BitBuffer {
+        filter_indices(self, indices)
```
connortsui20, line 40: here too

gatesn (approved):

vortex-compute/src/filter/buffer.rs
a10y: the lengths won't be the same b/c it's an indices

PR #5335, opened 2025-11-13
fix: unblock the fuzzers

a10y:
#5271 wasn't quite right actually for file_io

array_ops was broken on what I believe to be an unnecessary assertion

### DIFF
--- a/vortex-array/src/arrays/listview/array.rs
+++ b/vortex-array/src/arrays/listview/array.rs
@@ -221,17 +221,6 @@ impl ListViewArray {
         // Check that the size type can fit within the offset type to prevent overflows.
         let size_ptype = sizes.dtype().as_ptype();
         let offset_ptype = offsets.dtype().as_ptype();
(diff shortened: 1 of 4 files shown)


vortex-array/src/arrays/listview/array.rs
```diff
@@ -221,17 +221,6 @@ impl ListViewArray {
...
-        let offset_max = offsets.dtype().as_ptype().max_value_as_u64();
-
-        vortex_ensure!(
-            size_max <= offset_max,
-            "size type {:?} (max {}) must fit within offset type {:?} (max {})",
-            size_ptype,
-            size_max,
-            offset_ptype,
-            offset_max
-        );
```
a10y, line 234: array_ops fuzzer was failing here, i'm pretty sure this is not a necessary check? it's perfectly normal for offsets and sizes to be different ptypes. We already have a real check to validate the offsets and sizes only point at valid element ranges below

robert3005 (approved):

PR #5336, opened 2025-11-14
test: more `iter_bits` test cases

0ax1:


vortex-buffer/src/bit/buf.rs
joseph-isaacs, line 1: what made you add more tests?

joseph-isaacs (approved):

vortex-buffer/src/bit/buf.rs
0ax1: If you altenate only on a per bit basis `[true,false,true,false]` its not possible to spot if `iter_bits` starts on the correct byte. 

I wanted a test that fails if `let mut buffer_ptr = unsafe { self.buffer.as_ptr().add(self.offset / 8) };` would be set to `let mut buffer_ptr = unsafe { self.buffer.as_ptr().add(0) };`.

PR #5338, opened 2025-11-14
chore: drop duckdb install from gh actions

0ax1:
~~The binary is no longer used in our SQL benchmarks.~~ => still used in `compress`


0ax1:
The DuckDB binary is still used in compress.

PR #5342, opened 2025-11-14
chore: duckdb e2e roundtrip tests

0ax1:


vortex-duckdb/Cargo.toml
```diff
@@ -38,9 +38,12 @@ vortex = { workspace = true, features = ["files", "tokio", "object_store"] }
 vortex-utils = { workspace = true, features = ["dashmap"] }
 vortex-vector = { workspace = true }
 
+
 [dev-dependencies]
 jiff = { workspace = true }
 rstest = { workspace = true }
+vortex-runend = { path = "../encodings/runend" }
+vortex-sequence = { path = "../encodings/sequence" }
```
joseph-isaacs: where these not exposed from vector?

vortex-duckdb/src/e2e_test/vortex_scan_test.rs
```diff
@@ -703,3 +706,192 @@ fn test_vortex_scan_ultra_deep_nesting() {
     }
     assert_eq!(row_count, 1, "Should have retrieved 1 row");
 }
+
+async fn write_vortex_file_with_encodings() -> NamedTempFile {
```
joseph-isaacs, line 710: could this be 8 tests or you want all these columns?

0ax1: Strictly speaking we're also testing `StructArray` through that nesting.

vortex-duckdb/Cargo.toml
0ax1: My "assistant" might have error'd these in.  😅  But also, lemme double check.

0ax1: Looks like this is idiomatic.

vortex-duckdb/src/e2e_test/vortex_scan_test.rs
joseph-isaacs: I guess.

vortex-duckdb/Cargo.toml
joseph-isaacs: Really? dev deps inherit deps right?

0ax1: This is now 
```
vortex-runend = { workspace = true }
vortex-sequence = { workspace = true }
```

0ax1: The relative path was non-sense of course.

vortex-duckdb/Cargo.toml
```diff
@@ -41,6 +41,8 @@ vortex-vector = { workspace = true }
 [dev-dependencies]
 jiff = { workspace = true }
 rstest = { workspace = true }
+vortex-runend = { workspace = true }
+vortex-sequence = { workspace = true }
```
joseph-isaacs, line 46: there have to be relative, but can we not emit them or is seq not exposed?

joseph-isaacs (approved):

vortex-duckdb/Cargo.toml
0ax1: all other instances in the repo use `{ workspace = true }`

PR #5348, opened 2025-11-14
feat[vortex-expr]: add expr rewrite rules

joseph-isaacs:
# Features

Adds pluggable rewrite rule and use them in the optimizer.

# Breaking:

 - simplify_typed/simplify -> ExprOptimizer::{optimize/optimize_typed}, see https://github.com/vortex-data/vortex/blob/39004fc6dc9f21e18891f0b01f134ab2bf7f93d4/vortex-layout/src/layouts/struct_/reader.rs#L145

Using a vortex session in more places
 - simplify_typed/simplify
 - Layout::new_reader


joseph-isaacs:
@claude review please

vortex-array/src/expr/session/rewrite.rs
```diff
@@ -0,0 +1,269 @@
...
+/// Stores rewrite rules indexed by the expression ID they apply to.
+/// Typed and untyped rules are stored separately for better organization.
+#[derive(Default)]
+pub struct RewriteRuleRegistry {
+    /// Typed reduce rules (require TypedRewriteContext), indexed by expression ID
+    typed_reduce_rules: HashMap<ExprId, Vec<Arc<dyn DynTypedReduceRule>>>,
+    /// Untyped reduce rules (require only RewriteContext), indexed by expression ID
+    reduce_rules: HashMap<ExprId, Vec<Arc<dyn DynReduceRule>>>,
+    /// Parent reduce rules, indexed by expression ID
+    parent_rules: HashMap<ExprId, Vec<Arc<dyn DynParentReduceRule>>>,
```
joseph-isaacs: should we store both child and parent id?

joseph-isaacs:
@claude review please

vortex-array/src/expr/transform/partition.rs
```diff
@@ -62,7 +64,7 @@ where
             Nullability::NonNullable,
         );
 
-        let expr = simplify_typed(expr.clone(), scope)?;
+        let expr = simplify_typed(expr.clone(), scope, session)?;
```
gatesn: This seems like an odd place to run a full optimizer pass no? Isn't this itself run _within_ an array optimizer rule?

vortex-array/src/expr/transform/rules.rs
```diff
@@ -0,0 +1,135 @@
...
+        child_idx: usize,
+        ctx: C,
+    ) -> VortexResult<Option<Expression>>;
+}
+
+/// Base context for rewrite rules.
+pub trait RewriteContext {}
+
+// Blanket implementation: all references to RewriteContext implementors also implement RewriteContext
+impl<T: RewriteContext + ?Sized> RewriteContext for &T {}
```
gatesn: Why have this an not take `&C` in the traits above?

vortex-array/src/expr/transform/rules.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Context available to rewrite rules during expression optimization.
+/// Extends `RewriteContext` and provides access to dtype information.
+///
+/// Any `TypedRewriteContext` can be used as a `RewriteContext`, but not vice versa.
+pub trait TypedRewriteContext: RewriteContext {
+    fn dtype(&self) -> &DType;
```
gatesn: I think it's worth documenting exactly what this dtype represents

vortex-array/src/expr/transform/rules.rs
```diff
@@ -0,0 +1,135 @@
...
+impl<'a> RewriteContext for RootRewriteContext<'a> {}
+
+impl<'a> TypedRewriteContext for RootRewriteContext<'a> {
+    fn dtype(&self) -> &DType {
+        self.dtype
+    }
+}
+
+/// Type-erased wrappers that allows dynamic dispatch.
+pub(crate) trait DynReduceRule: Send + Sync {
```
gatesn, line 111: I haven't got there yet, but presumably there are some structs that wrap the generics version and impl the Dyn versions. I think they should be defined/moved here and called `ReduceRuleAdapter<T>(T)` etc to follow the expression/array pattern.

vortex-array/src/expr/transform/rules.rs
```diff
@@ -0,0 +1,135 @@
...
+        &self,
+        expr: &ExpressionView<V>,
+        parent: &Expression,
+        child_idx: usize,
+        ctx: C,
+    ) -> VortexResult<Option<Expression>>;
+}
+
+/// Base context for rewrite rules.
+pub trait RewriteContext {}
```
gatesn: I'm not sure we need full trait hierarchy here. Something simpler / I have seen more commonly is to just use concrete marker types, e.g.

```rust
pub trait RewriteContext: private::Sealed {}

pub struct Typed {
  dtype: DType
}
impl RewriteContext for Typed {}

pub struct Untyped
impl RewriteContext for Untyped {}

impl ReduceRule<Blah, Typed> for Blah {

}
```

vortex-array/src/expr/transform/simplify.rs
```diff
@@ -4,45 +4,171 @@
...
 /// Simplifies an expression into an equivalent expression which is faster and easier to analyze.
 ///
+/// This applies only untyped rewrite rules registered in the default session.
 /// If the scope dtype is known, see `simplify_typed` for a simplifier which uses dtype.
-pub fn simplify(e: Expression) -> VortexResult<Expression> {
-    let e = e
-        .transform_up(simplify_transformer)
-        .map(|e| e.into_inner())?;
-    Ok(find_between(e))
+pub fn simplify(e: Expression, session: &ExprSession) -> VortexResult<Expression> {
```
gatesn: I think it's worth wrapping this up as a `struct ExprOptimizer` with either a `pub fn new(rules: ...)` and/or a`ExprSessionExt { fn new_expr_optimizer() -> ExprOptimizer }`. 

That way we can pass an optimizer around where necessary, instead of passing a VortexSession all over the place. I'm not sure yet whether we want to just pass a session into basically everything, or try and pass more scoped things.

vortex-array/src/expr/transform/simplify.rs
```diff
@@ -4,45 +4,171 @@
...
+    let ctx = EmptyRewriteContext;
+
+    let e = apply_parent_rules(e, &ctx, session)?;
```
gatesn, line 19: I think we want to reduce child first no?

gatesn (commented):
Leaving some early comments - I have only gone through vortex-array atm

vortex-array/src/expr/transform/simplify.rs
joseph-isaacs: I don't know, I think both make sense, really we want to iterate

vortex-array/src/expr/transform/rules.rs
joseph-isaacs: I liked this, but you often prefer traits. I will do this

vortex-array/src/expr/transform/partition.rs
joseph-isaacs: Its useful to apply get_item + pack reductions. Shall I add a todo to revisit?

vortex-array/src/expr/transform/rules.rs
joseph-isaacs: they where called `RuleAdapter`

vortex-array/src/expr/transform/rules.rs
```diff
@@ -0,0 +1,138 @@
...
+
+/// A context for rewrite rules that don't need dtype information.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct RuleContext;
+
+impl private::Sealed for RuleContext {}
+impl RewriteContext for RuleContext {}
+
+impl private::Sealed for &RuleContext {}
+impl RewriteContext for &RuleContext {}
```
gatesn, line 108: You shouldn't need these

gatesn (approved):

PR #5349, opened 2025-11-14
feat: faster TakeKernel for VarBinArray

a10y:
also removes potential overflow issues found in https://github.com/vortex-data/vortex/issues/5347


a10y:
40-50% faster

Before:
image


After:

image

a10y:
Sadly, FSST requires that its child is VarBin encoded and expects that VarBin.take yields VarBin. Removing this assumption from FSSTArray is nontrivial so for now this is a non-starter :/

PR #5352, opened 2025-11-14
Faster Pipelines

gatesn:
Removes a bunch of the pipeline calling overhead.


vortex-array/src/pipeline/filter/buffer.rs
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use crate::pipeline::bit_view::BitView;
+use crate::pipeline::N;
+use vortex_compute::filter::Filter;
+
+impl<'a, T: Copy> Filter<BitView<'a>> for &'a mut [T] {
```
gatesn: It's actually very useful to have highly specialized filter kernels for N elements. So I think we keep these here, outside the vortex-compute package, possibly delegating to vortex-compute where possible.

fargito:
Hello, CodSpeed team member here.

We had some trouble processing the results in your latest run, due to the large number of uploaded benchmarks. We have some limitations on our side that we are currently working on. In the meantime, could you consider sharding the benchmarks into more than 2 jobs? It should fix the issue for now.

Here are [some docs on benchmark sharding](https://codspeed.io/docs/features/sharded-benchmarks/index#how-to-split-the-execution-of-benchmark-suites).

vortex-mask/src/lib.rs
```diff
@@ -297,6 +303,12 @@ impl Mask {
         Self::from_indices(len, intersection)
     }
 
+    /// Clear the mask, setting the length to zero.
+    #[inline(always)]
+    pub fn clear(&mut self) {
```
a10y: does this really do much? we don't save an allocation right

encodings/fastlanes/src/bitpacking/array/bitpack_pipeline.rs
```diff
@@ -151,72 +154,50 @@ impl<BP: PhysicalPType<Physical: BitPacking>> Kernel for AlignedBitPackedKernel<
...
                             self.packed_bit_width,
-                            not_yet_unpacked_values,
+                            packed_bytes,
                             idx,
                         )
                     };
 
                     // SAFETY: We just reserved enough capacity to push these values.
-                    unsafe { output_vector.push_unchecked(unpacked_value) };
+                    unsafe { output.push_unchecked(transmute_copy(&unpacked_value)) };
```
a10y, line 163: why do you need this now?

gatesn: Because our output is the (possibly) signed type of the bitpacked array, but the "physical" type is always unsigned. So fastlanes returns u8 and we write our an i8.

vortex-mask/src/lib.rs
gatesn: This might be leftover from before, I'm not sure it's called anywhere is it?

a10y (approved):

PR #5356, opened 2025-11-16
BitView SIMD filtering

gatesn:


vortex-compute/src/filter/slice/neon/neon_u16.rs
```diff
@@ -0,0 +1,75 @@
...
+/// Since the interleaved load is done in a single instruction, we then use the same u8 shuffle
+/// masks to shuffle both the lower and higher byte vectors separately. Finally, we use an
+/// interleaved store to write the compressed u16 elements back to memory.
+#[cfg(target_arch = "aarch64")]
+#[target_feature(enable = "neon")]
+pub(super) unsafe fn filter_neon_u16<const NB: usize>(data: *mut u16, mask: &BitView<NB>) {
+    let mut read_ptr = data;
+    let mut write_ptr = data;
+
+    for word in mask.iter_words() {
```
gatesn, line 28: This outer loop is duplicative, maybe lift it into parent

vortex-buffer/src/bit/view.rs
```diff
@@ -156,6 +156,17 @@ impl<'a, const NB: usize> BitView<'a, NB> {
         (0..Self::N_WORDS).map(move |idx| unsafe { ptr.add(idx).read_unaligned() })
     }
 
+    /// Iterate the [`BitView`] in fixed-size words.
+    ///
+    /// The words are loaded using unaligned loads to ensure correct bit ordering.
+    /// For example, bit 0 is located in `word & 1 << 0`, bit 63 is located in `word & 1 << 63`,
+    /// assuming the word size is 64 bits.
+    pub fn iter_sized<W: 'static>(&self) -> impl Iterator<Item = W> + '_ {
```
0ax1, line 164: `#[inline]`?

gatesn: Oh I meant to delete this function, but 👍

vortex-compute/src/filter/slice/neon/neon_u8.rs
```diff
@@ -0,0 +1,101 @@
...
+
+use vortex_buffer::BitView;
+
+/// For u8 types, we use NEON's tbl lookup instruction to perform a shuffle based on a pre-computed
+/// lookup table (LUT).
+///
+/// In theory, we could use vqtbl1q_u8 for 16-byte vectors, but that would require a 65KB LUT
+/// (256 entries × 16 bytes each), which is too large for practical use as it thrashes
+/// the CPU cache.
+#[cfg(target_arch = "aarch64")]
```
0ax1, line 17: What's the point of only adding neon intrinsics, opposed to also having x86. To explore whether this outperforms auto-vectorization?

gatesn: There is kind of no reasonable auto-vectorization for filtering that I've seen. 

@connortsui20 was working on x86 / AVX512 so I figured I'd poke at neon!

vortex-compute/src/filter/slice/neon/neon_u8.rs
```diff
@@ -0,0 +1,101 @@
...
+                        }
+                        0xFF => {
+                            // All bits set - fast path
+                            ptr::copy(read_ptr, write_ptr, 8);
+                            write_ptr = write_ptr.add(8);
+                        }
+                        _ => {
+                            // Finally, use the lookup table to compress selected elements
+                            // Load uint8x8 values and compress them using the lookup table.
+                            let values = vld1_u8(read_ptr);
```
0ax1, line 53: My assumption is that branching and branch mispredictions have a significant overhead here if we do it on a per byte level. I think it's worth benchmarking only keeping the ` _ =>` case. 

What might be reasonable is keeping these checks on a word level.
```
0u64 => {
    // Skip empty chunks
}
0xFF => {
    // All bits set - fast path
    ptr::copy(read_ptr, write_ptr, 8);
    write_ptr = write_ptr.add(8);
}
```
But same here: Would be interesting to look at different workloads and see whether introducing the branches on a word level is actually a win over running `_ => ` unconditionally.

vortex-buffer/src/bit/view.rs
```diff
@@ -156,6 +156,17 @@ impl<'a, const NB: usize> BitView<'a, NB> {
...
+        let ptr = self.bits.as_ptr().cast::<W>();
+        // We use constant N_WORDS to trigger loop unrolling.
+        (0..(NB / size_of::<W>())).map(move |idx| unsafe { ptr.add(idx).read_unaligned() })
+    }
+
     /// Runs the provided function `f` for each index of a `true` bit in the view.
     pub fn iter_ones<F>(&self, mut f: F)
```
0ax1, line 171:
```
 while raw != 0 {
                        let bit_pos = raw.trailing_zeros();
                        f(bit_idx + bit_pos as usize);
                        raw &= raw - 1; // Clear the bit at `bit_pos`
                    }
```
Not introduced as part of this PR, but iterating based on `while raw` seems to be bad for pipelining, as an iteration depends on the previous one.

connortsui20:
actually I think this deserves some cleaning up / monomorphizing before we merge this (I will make the scalar implementation better on my PR https://github.com/vortex-data/vortex/pull/5399)

robert3005:
We have deleted bitview from the codebase

PR #5360, opened 2025-11-17
feat: add VarBin compressor to btrblocks string compressor

joseph-isaacs:
## Summary

Adds `VarBinScheme` as a new compression option for string data in the btrblocks compressor. This scheme converts `VarBinViewArray` to the more compact `VarBinArray` representation using i32 offsets.

**Changes:**
- Added `VarBinScheme` struct and registered it with `StringCode(1)`
- Implemented `Scheme` trait with compression ratio estimation via sampling
- Used `ArrayAccessor` pattern for efficient VarBinView to VarBin conversion
- Updated all subsequent scheme codes sequentially (Dict: 1→2, FSST: 2→3, Constant: 3→4, Sparse: 4→5)

**Implementation Details:**
- The `compress` method uses `VarBinBuilder<i32>` for i32 offset arrays
- Leverages `ArrayAccessor::with_iterator` for zero-copy iteration when possible
- Properly handles nullable values through the iterator pattern
- Uses `estimate_compression_ratio_with_sampling` consistent with other schemes

## Test plan

- [x] Build succeeds: `cargo build -p vortex-btrblocks`
- [x] Clippy passes: `cargo clippy --all-targets --all-features -p vortex-btrblocks`
- [x] Code formatted: `cargo +nightly fmt --all`
- [x] Benchmarks to be run with `benchmark` tag

### DIFF
--- a/vortex-btrblocks/src/string.rs
+++ b/vortex-btrblocks/src/string.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use vortex_array::arrays::builder::VarBinBuilder;
 use vortex_array::arrays::{
     ConstantArray, DictArray, MaskedArray, VarBinArray, VarBinViewArray, VarBinViewVTable,
 };
@@ -93,6 +94,7 @@ impl Compressor for StringCompressor {
     fn schemes() -> &'static [&'static Self::SchemeType] {
         &[
             &UncompressedScheme,
+            &VarBinScheme,
             &DictScheme,
             &FSSTScheme,
             &ConstantScheme,
@@ -117,6 +119,9 @@ impl<T> StringScheme for T where T: Scheme<StatsType = StringStats, CodeType = S
 pub struct UncompressedScheme;
 
 #[derive(Debug, Copy, Clone)]
+pub struct VarBinScheme;
+
+#[derive(Debug, Copy, Clone)]
 pub struct DictScheme;
 
 #[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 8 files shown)


vortex-btrblocks/src/string.rs
```diff
@@ -132,11 +137,12 @@ pub struct NullDominated;
...
-const DICT_SCHEME: StringCode = StringCode(1);
-const FSST_SCHEME: StringCode = StringCode(2);
-const CONSTANT_SCHEME: StringCode = StringCode(3);
+const VARBIN_SCHEME: StringCode = StringCode(1);
+const DICT_SCHEME: StringCode = StringCode(2);
+const FSST_SCHEME: StringCode = StringCode(3);
+const CONSTANT_SCHEME: StringCode = StringCode(4);
 
-const SPARSE_SCHEME: StringCode = StringCode(4);
+const SPARSE_SCHEME: StringCode = StringCode(5);
```
a10y: can we avoid renumbering all the codes here?

vortex-btrblocks/src/string.rs
```diff
@@ -167,6 +173,52 @@ impl Scheme for UncompressedScheme {
...
+        use vortex_array::accessor::ArrayAccessor;
+
+        let src = stats.source();
+        let mut builder = VarBinBuilder::<i32>::with_capacity(src.len());
+
+        src.with_iterator(|iter| {
+            for value in iter {
+                builder.append(value);
+            }
+        });
```
a10y: string compression is already bottleneck on writing (mostly FSST), I'm curious how much this will cost us

vortex-btrblocks/src/string.rs
```diff
@@ -167,6 +173,53 @@ impl Scheme for UncompressedScheme {
...
+        allowed_cascading: usize,
+        excludes: &[StringCode],
+    ) -> VortexResult<f64> {
+        estimate_compression_ratio_with_sampling(
+            self,
+            stats,
+            is_sample,
+            allowed_cascading,
+            excludes,
+        )
```
a10y: I think we can actually estimate this pretty easily by iterating the views, counting total bytes + offsets. That saves us actually having to materialize

vortex-btrblocks/src/string.rs
joseph-isaacs: I should be able to calc without a conversion?

joseph-isaacs:
This will always kick in over uncompressed. We need to work out when we should include this.

PR #5361, opened 2025-11-17
fix: VarBin overflow on take

a10y:
Fix for https://github.com/vortex-data/vortex/issues/5347

Naively, on take we widen to either 32 or 64-bit offsets based on the input offsets PType. This is similar to what arrow-rs does.

I think it's hard to be more intelligent than this without doing the take in 2 passes.

### DIFF
--- a/vortex-array/src/arrays/varbin/compute/take.rs
+++ b/vortex-array/src/arrays/varbin/compute/take.rs
@@ -17,15 +17,16 @@ impl TakeKernel for VarBinVTable {
         let offsets = array.offsets().to_primitive();
         let data = array.bytes();
         let indices = indices.to_primitive();
+        let dtype = array
+            .dtype()
+            .clone()
+            .union_nullability(indices.dtype().nullability());
         let array = match_each_integer_ptype!(indices.ptype(), |I| {
             // On take, offsets get widened to either 32- or 64-bit based on the original type,
             // to avoid overflow issues.
             match offsets.ptype() {
                 PType::U8 => take::<I, u8, u32>(
-                    array
-                        .dtype()
-                        .clone()
-                        .union_nullability(indices.dtype().nullability()),
+                    dtype,
                     offsets.as_slice::<u8>(),
                     data.as_slice(),
                     indices.as_slice::<I>(),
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/varbin/compute/take.rs
```diff
@@ -17,37 +17,118 @@ impl TakeKernel for VarBinVTable {
...
+                        .dtype()
+                        .clone()
+                        .union_nullability(indices.dtype().nullability()),
+                    offsets.as_slice::<u16>(),
+                    data.as_slice(),
+                    indices.as_slice::<I>(),
+                    array.validity_mask(),
+                    indices.validity_mask(),
+                ),
+                PType::U32 => take::<I, u32, u32>(
```
joseph-isaacs, line 44: why not u64 here?

vortex-array/src/arrays/varbin/compute/take.rs
```diff
@@ -17,37 +17,118 @@ impl TakeKernel for VarBinVTable {
         let offsets = array.offsets().to_primitive();
         let data = array.bytes();
         let indices = indices.to_primitive();
-        match_each_integer_ptype!(offsets.ptype(), |O| {
-            match_each_integer_ptype!(indices.ptype(), |I| {
-                Ok(take(
+        let array = match_each_integer_ptype!(indices.ptype(), |I| {
+            // On take, offsets get widened to either 32- or 64-bit based on the original type,
+            // to avoid overflow issues.
```
joseph-isaacs, line 26: not in this PR, but I guess we could do `|Indices| * max_string_len`?

vortex-array/src/arrays/varbin/compute/take.rs
```diff
@@ -17,37 +17,118 @@ impl TakeKernel for VarBinVTable {
...
-                )?
-                .into_array())
-            })
-        })
+                ),
+                PType::U16 => take::<I, u16, u32>(
+                    array
+                        .dtype()
+                        .clone()
+                        .union_nullability(indices.dtype().nullability()),
```
joseph-isaacs: can we pull this out?

vortex-array/src/arrays/varbin/compute/take.rs
a10y: I sort of just stuck with arrow behavior: take on ListArray -> ListArray, take on LargeListArray -> LargeListArray

vortex-array/src/arrays/varbin/compute/take.rs
a10y: yes

vortex-array/src/arrays/varbin/compute/take.rs
a10y: possibly yea, I was just trying to avoid doing a second pass of the indices to implement this

robert3005 (approved):

PR #5362, opened 2025-11-17
chore: All of vortex-gpu requires compiler flag

robert3005:

### DIFF
--- a/vortex-gpu/src/lib.rs
+++ b/vortex-gpu/src/lib.rs
@@ -8,6 +8,7 @@
 //! ```shell
 //! RUSTFLAGS="--cfg gpu_unstable" cargo build -p ...
 //! ```
+#![cfg(gpu_unstable)]
 
 pub mod bit_unpack;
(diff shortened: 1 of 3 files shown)


joseph-isaacs (approved):

vortex-gpu/src/lib.rs
```diff
@@ -8,6 +8,7 @@
 //! ```shell
 //! RUSTFLAGS="--cfg gpu_unstable" cargo build -p ...
 //! ```
+#![cfg(gpu_unstable)]
```
0ax1, line 11: Are there no CI tasks where we now need to enable the cfg?

0ax1 (approved):

vortex-gpu/src/lib.rs
robert3005: We have previously enabled the flag in CI to validate that everything builds but this doesn't need to be enabled locally

PR #5366, opened 2025-11-17
Consolidate into stat_expression function

gatesn:
This removes the unused stat_field_path function, as well as improves coverage for propagating other statistics.

Interestingly it has highlighted the difference between pruning stats (min, max, is_null), and aggregate stats (sum, null_count, etc.). I have some ideas for generalizing these concepts, but that's for another day.


vortex-array/src/expr/exprs/get_item.rs
```diff
@@ -94,34 +94,20 @@ impl VTable for GetItem {
...
-            .stat_field_path()
-            .map(|fp| fp.push(expr.data().clone()))
+        // TODO(ngates): I think we can do better here and support stats over nested fields.
+        //  It would be nice if delegating to our child would return a struct of statistics
+        //  matching the nested DType such that we can write:
+        //    `get_item(expr.child(0).stat_expression(...), expr.data().field_name())`
+
+        // TODO(ngates): For now, we maintain the existing *broken* behavior of assuming that
+        //  getitem refers to a field on the root scope.
+        catalog.stats_ref(&FieldPath::from_name(expr.data().clone()), stat)
```
danking: can we verify child is root?

vortex-array/src/expr/exprs/literal.rs
```diff
@@ -72,41 +73,46 @@ impl VTable for Literal {
...
+                // For floating point types, the NaNCount is 1 for lit(NaN), and 0 otherwise.
+                let value = expr.data().as_primitive_opt()?;
+                if !value.ptype().is_float() {
+                    return None;
+                }
+
+                match_each_float_ptype!(value.ptype(), |T| {
+                    match value.typed_value::<T>() {
+                        None => Some(lit(0u64)),
+                        Some(value) if value.is_nan() => Some(lit(1u64)),
```
danking: this seems wrong.

danking (changes requested):
x

vortex-array/src/expr/exprs/literal.rs
gatesn: Huh... this existed before. And I guess proves again why it should be has_nulls / has_nans...? We're just assuming that the caller is going to check for >0

danking (approved):

vortex-array/src/expr/exprs/get_item.rs
gatesn: So a test now fails that was checking: `(cast($ as {a=utf8?, b=utf8?}).a = "value")`

The get_item on the cast doesn't now forward stats since the child isn't the root expression. This seems like an overly conservative check.... 

Will have a think. 

@robert3005 in case you have ideas

vortex-array/src/expr/exprs/cast.rs
```diff
@@ -86,32 +87,32 @@ impl VTable for Cast {
...
+            | Stat::UncompressedSizeInBytes => expr.child(0).stat_expression(stat, catalog),
+            Stat::NullCount => {
+                // if !expr.data().is_nullable() {
+                // NOTE(ngates): we should decide on the semantics here. In theory, the null
+                //  count of something cast to non-nullable will be zero. But if we return
+                //  that we know this to be zero, then a pruning predicate may eliminate data
+                //  that would otherwise have caused the cast to error.
+                // return Some(lit(0u64));
+                // }
+                None
```
joseph-isaacs, line 117: I think we should move error into the pruning predicate, can assert that child `null_count == 0`

vortex-array/src/expr/exprs/literal.rs
```diff
@@ -72,41 +73,47 @@ impl VTable for Literal {
...
+                // The NaNCount for a non-float literal is not defined.
+                // For floating point types, the NaNCount is 1 for lit(NaN), and 0 otherwise.
+                let value = expr.data().as_primitive_opt()?;
+                if !value.ptype().is_float() {
+                    return None;
+                }
+
+                match_each_float_ptype!(value.ptype(), |T| {
+                    if value.typed_value::<T>().is_some_and(|v| v.is_nan()) {
+                        Some(lit(1u64))
```
joseph-isaacs, line 100: This should be `array.len()` which we don't know?

joseph-isaacs (commented):
Its a shame to loose the &mut self?

vortex-array/src/expr/exprs/literal.rs
joseph-isaacs: I guess this could be u64::MAX?

vortex-array/src/expr/exprs/cast.rs
gatesn: But the pruning predicate operates over the expression we return from here. 

The example is if I have some data in one zone of the file with a null value, and I cast to non-null, then clearly that should error. If we return `0` from here, then that zone may get pruned without any error at all.

vortex-array/src/expr/exprs/literal.rs
gatesn: This is related to that discussion in the slack thread that pruning stats != pre-computed aggregations. This code was just moved from before, so I'm inclined to keep the existing behavior for now.

vortex-array/src/expr/exprs/get_item.rs
```diff
@@ -94,34 +96,23 @@ impl VTable for GetItem {
...
-        expr.children()[0]
-            .stat_field_path()
-            .map(|fp| fp.push(expr.data().clone()))
+        // TODO(ngates): I think we can do better here and support stats over nested fields.
+        //  It would be nice if delegating to our child would return a struct of statistics
+        //  matching the nested DType such that we can write:
+        //    `get_item(expr.child(0).stat_expression(...), expr.data().field_name())`
+
+        // For now, we check that the child is the root expression.
+        if is_root(expr.child(0)) {
```
gatesn: This breaks the test that checks that `(cast($ as {a=utf8?, b=utf8?}).a = "value")` can be pruned because the child of the getitem expr is a cast, not the root expression.

Without this check though, the expr `$.x.a` would incorrectly return stats for `$.a` if it happened to exist.

PR #5367, opened 2025-11-17
Mooooaarrrrr generics

gatesn:


vortex-compute/src/filter/vector/bool.rs
```diff
@@ -28,18 +28,15 @@ where
...
 impl<M> Filter<M> for &mut BoolVectorMut
 where
-    for<'a> &'a BitBuffer: Filter<M, Output = BitBuffer>,
+    for<'a> &'a mut BitBufferMut: Filter<M, Output = ()>,
     for<'a> &'a mut MaskMut: Filter<M, Output = ()>,
 {
     type Output = ();
 
     fn filter(self, selection: &M) -> Self::Output {
         // TODO(aduffy): how can we do this faster in-place?
```
a10y, line 37: I think we can kill this comment now

a10y (approved):

PR #5371, opened 2025-11-18
Use the specified concurrency value when reading from a ObjectStoreIoSource

nrc:
Previously, `ObjectStoreIoSource` always used the default value, even if it was overridden using `ObjectStoreReadSource::with_concurrency`.

Sorry, I have no idea how to nicely test this.


a10y (approved):
thank you!

PR #5376, opened 2025-11-18
chore[vortex-gpu]: remove vortex-gpu

joseph-isaacs:
This PoC will be replaced with a extensible implementation


.github/workflows/ci.yml
```diff
@@ -396,65 +389,16 @@ jobs:
...
           # Build with full debug info first (helps with caching)
-          cargo +nightly build --locked --workspace --exclude vortex-gpu --all-features --target x86_64-unknown-linux-gnu
+          cargo +nightly build --locked --workspace --all-features --target x86_64-unknown-linux-gnu
           # Run tests with sanitizers and debug output
           cargo +nightly nextest run \
             --locked \
             --workspace \
-            --exclude vortex-gpu \
-            --all-features \
-            --no-fail-fast \
```
robert3005, line 406: restore these two lines?

vortex-buffer/src/lib.rs
```diff
@@ -2,7 +2,6 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 #![deny(missing_docs)]
-// cudarc HostSlice has len and is_empty methods that duplicate BufferMut methods.
 #![allow(clippy::same_name_method)]
```
robert3005: you can remove this since you removed the comment

robert3005 (approved):
just the two comments

.github/workflows/ci.yml
joseph-isaacs: they are below?

robert3005: this is such a bad diff to read. You're right

PR #5387, opened 2025-11-18
chore[vortex-array]: remove SerdeVTable

joseph-isaacs:
`SerdeVTable` assumes that metadata just be serializable. This change allows metadata and metadata serialization to be split.

Breaks:

- Moved the SerdeVTable methods into the main array VTable.
- Added serialize and deserialize methods which subsume `SerializeMetadata` and `DeserializeMetadata`


joseph-isaacs:
@claude review  plz

robert3005 (approved):
I think I don't understand something. We don't need to require `metadata` to return a serialisable metadata, we can handle this in serialize method, however, I realise that we probably need the wrappers since that's how we have written everything so far. Though I don't know in which case this would be a simplifcation.

Minor thing is that (unrelated to this pr but noticed while reviewing metadata) is that we sometimes do `PType::try_from(DType)` and propagate the error and sometimes we panic.

joseph-isaacs:
Want to create an issue for the PType thing?

joseph-isaacs:
This is the first step in a series of array changes

robert3005:
there's two cases of the ptype unwrap and it's a valid unwrap, the constraint is validated on array construction

PR #5389, opened 2025-11-18
feat: cast expression

danking:
Without this, it is impossible to compare two columns with different bit-widths.


vortex-python/src/expr/mod.rs
```diff
@@ -323,3 +328,41 @@ pub fn and_(left: PyExpr, right: PyExpr) -> PyResult<PyExpr> {
...
+/// <vortex.Expr object at ...>
+/// ```
+///
+/// Cast to a wider floating-point type:
+///
+/// ```python
+/// >>> import vortex.expr as ve
+/// >>> import vortex as vx
+/// >>> ve.cast(ve.literal(vx.float_(16), 3.145), vx.float_(64))
+/// <vortex.Expr object at ...>
```
joseph-isaacs, line 361: any reason we omit the types?

joseph-isaacs (approved):

vortex-python/src/expr/mod.rs
danking: @joseph-isaacs what do you mean? I think this is just the default printed form for any expression.

PR #5396, opened 2025-11-19
fix: vx_struct_fields_field_dtype takes usize instead of u64

### LINKED ISSUE #5391: FFI inconsistency with field idx type
I would expect the idx parameter of these two functions to have the same size. Yet the size here is different.

```c

/**
 * Return a borrowed reference to the name of the field at the given index.
 *
 * The returned pointer is valid as long as the struct fields is valid.
 * Do NOT free the returned string pointer - it shares the lifetime of the struct fields.
 * Returns null if the index is out of bounds.
 */
const vx_string *vx_struct_fields_field_name(const vx_struct_fields *dtype, size_t idx);

/**
 * Returns an *owned* reference to the dtype of the field at the given index.
 *
 * The return type is owned since struct dtypes can be lazily parsed from a binary format, in
 * which case it's not possible to return a borrowed reference to the field dtype.
 *
 * Returns null if the index is out of bounds or if the field dtype cannot be parsed.
 */
const vx_dtype *vx_struct_fields_field_dtype(const vx_struct_fields *dtype, uint64_t idx);
```

robert3005:
fix #5391 

### DIFF
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -805,7 +805,7 @@ const vx_string *vx_struct_fields_field_name(const vx_struct_fields *dtype, size
  *
  * Returns null if the index is out of bounds or if the field dtype cannot be parsed.
  */
-const vx_dtype *vx_struct_fields_field_dtype(const vx_struct_fields *dtype, uint64_t idx);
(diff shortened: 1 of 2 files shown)


joseph-isaacs (approved):

a10y (approved):

robert3005:
thanks @a10y

PR #5399, opened 2025-11-19
add AVX512 support for filtering in place

connortsui20:
Reorganizes some of the filter implementations and adds AVX512 filter implementation and benchmarks


a10y:
Does GitHub give us CI runners with AVX512 support? My recollection from when I did the AVX2 take kernel is that they don't

connortsui20:
not really sure how to deal with this, I think sometimes they have it, sometimes not?

robert3005:
CI runs in AWS, you can get whatever machine you want, right now by default we use m7i/m7a which are sapphire rapids/zen4 which have AVX512. On benchmarks we run c6i which is ice-lake sp which also has avx512

0ax1:
Heads up that we can't micro-benchmark avx512 with codspeed, as valgrind/callgrind does not support avx512. Saying you'll get numbers for avx2 if it is implemented: 

```
      - name: Build benchmarks (shard 1)
        env:
          RUSTFLAGS: "-C target-feature=+avx2"
      ...
```

connortsui20:
Given that were pivoting to batch I'm going to put a pin in this.

gatesn:
Nooo I like this!

vortex-compute/src/filter/slice.rs
connortsui20, line 1: The items here are moved to the `out` module in `slice`

vortex-compute/benches/avx512.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Different probability values to benchmark.
+const PROBABILITIES: &[f64] = &[0.0, 0.1, 0.25, 0.5, 0.75, 0.9, 1.0];
+
+/// The number of samples per benchmark.
+const SAMPLE_SIZE: u32 = 64;
+
+#[divan::bench(sample_size = SAMPLE_SIZE, args = SIZES.iter().copied().cartesian_product(PROBABILITIES.iter().copied()))]
+fn in_place_scalar(bencher: divan::Bencher, (size, probability): (usize, f64)) {
+    let mask = create_random_mask(size, probability);
+    bencher
```
0ax1, line 51: Let's black-box the mask here. Could create 
```
(0..size as i32).collect::<Vec<_>>(
```
upfront, and then 
```
 .with_inputs(|| (data.clone(), &mask))
```

vortex-compute/benches/avx512.rs
```diff
@@ -0,0 +1,87 @@
...
+        .bench_refs(|data| filter_in_place_scalar(data, &mask))
+}
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+#[divan::bench(sample_size = SAMPLE_SIZE, args = SIZES.iter().copied().cartesian_product(PROBABILITIES.iter().copied()))]
+fn in_place_avx512(bencher: divan::Bencher, (size, probability): (usize, f64)) {
+    let mask = create_random_mask(size, probability);
+    bencher
+        .with_inputs(|| (0..size as i32).collect::<Vec<_>>())
+        .bench_refs(|data| unsafe { filter_in_place_avx512(data, &mask) })
```
0ax1: same here

vortex-compute/src/filter/slice/simd_compress.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+}
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+impl SimdCompress for u64 {
+    const WIDTH: usize = 64;
+    type MaskType = u8;
+
+    #[target_feature(enable = "avx512f")]
+    unsafe fn compress_vector(mask: Self::MaskType, vec: __m512i) -> __m512i {
```
a10y, line 155: may as well `#[inline]`

connortsui20:
@0ax1 is there a way for me to disable the benchmark just for codspeed?

vortex-compute/src/filter/slice/in_place/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    let data_len = data.len();
+
+    for read_pos in 0..data_len {
+        let byte_idx = read_pos / 8;
+        let bit_idx = read_pos % 8;
+
+        if (mask[byte_idx] >> bit_idx) & 1 == 1 {
+            data[write_pos] = data[read_pos];
+            write_pos += 1;
+        }
```
a10y, line 40: note: i think you can do this faster by always doing speculative write to data[write_pos], and then just incrementing write_pos by the mask bit on each loop iteration. FSST uses a similar trick in its compression loop.

maybe save for FLUP though so we can benchmark the difference

0ax1:
> @0ax1 is there a way for me to disable the benchmark just for codspeed?

~~You shouldn't need to. We compile for AVX2, so the AVX512 version shouldn't be picked up.~~

~~Ah I see that we're running into `failed to execute the benchmark process, exit code: 132` which I assume is picked based on runtime capabilities of the CPU?~~

Discussed to exclude the benchmark via: `#[cfg(not(codspeed))]`

connortsui20:
hmm it seems I've done something wrong here then: https://github.com/vortex-data/vortex/actions/runs/19644046057/job/56254751182?pr=5399

vortex-compute/src/filter/slice/in_place/mod.rs
connortsui20: yeah this entire function is super inefficient but I'm going to wait on optimizing all of these further until we actually start using it in operators and can get better / more realistic benchmarks

gatesn: We probably shouldn't merge SIMD code unless it is fully optimized. It's a bit misleading

connortsui20: I could just remove this then? This isn't SIMD code

connortsui20:
I'm going to close this until we start doing e2e tests and we can actually measure how much faster things are.

PR #5189, opened 2025-11-04
Upgrade DF to 51

AdamGS:
Testing the upgrade to DF 51.

- https://github.com/apache/datafusion/issues/17558#issuecomment-3487645051

PR #5406, opened 2025-11-19
fix: filter pushdown for nested fields

a10y:
In #5295, we accidentally broke nested filter pushdown. The issue is that FileSource::try_pushdown_filters seems like it's meant to evaluate using the whole file schema, rather than any projected schema. As an example, in the Github Archive benchmark dataset, we have the following query, which should trivially pushdown and be pruned, executing about 30ms or so:

```
SELECT COUNT(*) from events WHERE payload.ref = 'refs/head/main'
```

However, after this change, pushdown of this field was failing, pushing query time up 100x. The root cause is that the old logic attempted to apply the file schema to the source_expr directly.

Concretely, for the gharchive query, the whole expression is something like:

```text
BinaryExpr {
    lhs: GetField {
    	source_expr: Column { name: "payload", index: 0 },
	field_expr: Literal { value: "ref" }
    }
    rhs: Literal { value: "refs/head/main" }
    operator: Eq
}
```

The issue is that the column index 0 is wrong for the whole file. Instead, we need to recursively ensure that the source_expr is a valid sequence of Column and GetField expressions that resolve properly.

Note how we already were doing this for checking if a standalone Column expression can be pushed down:

```
    } else if let Some(col) = expr.downcast_ref::<df_expr::Column>() {
        schema
            .field_with_name(col.name())
            .ok()
            .is_some_and(|field| supported_data_types(field.data_type()))
```


a10y:
GH Archive query 0 before: 3 seconds

image


After: 30ms

image

a10y:
Actually, I'm not sure why this check was added. AFAICT, attempts to create a GetFieldFunction should fail at planning time, before we even try and pushdown any filter expressions to the source

image

a10y:
I'm actually fairly confident that we just need to validate the ScalarFunction being pushed down to us is `GetFieldFunc` and let DF do the rest.

robert3005:
The case I found while reading parquet logic for this is that there might be a constant synthetic column created and you might have a filter on it where you could get to see the column. There might be test cases in DataFusion parquet

asubiotto:
The reason I added this check was that I was seeing getfield fail at execution time unexpectedly because the table schema (merged schema over all files) does have the field so the df planning works fine (as it should), but a specific file does not. I do think we need to check field existence.

a10y:
> I was seeing getfield fail at execution time unexpectedly because the table schema (merged schema over all files) does have the field so the df planning works fine (as it should), but a specific file does not.

That makes sense, wasn't thinking about schema evolution.

The Source reports what can be pushed down, and it has access to the table schema but doesn't know the individual file schemas. So, I _think_ we should be adapting the predicate in the FileOpener instead of the FileSource.

I've also noticed that some of these APIs have changed in DF 51 so I can double-check that today.

asubiotto:
Ahh, that makes sense. We already do pushdown checks in the opener against the file schema, so `can_be_pushed_down` should return true in `try_pushdown_filters` at the source level against the table schema, but each individual opener should check field existence in the file schema. I think this should basically already work, `can_be_pushed_down` is called at both levels.

a10y:
https://github.com/vortex-data/vortex/blob/879a53baab7b2a4f72e928fba3cbf0fb92eb989a/vortex-datafusion/src/persistent/opener.rs#L297-L308

I think we should error here instead of dropping the predicate silently after we had told DF that we're going to handle it.

By my reading of DF, when we report `PushDownPredicate::supported` back up to DF, it assumes that the filter is wholly executed by the DataSource. Here's how the Filter node for reference: https://github.com/apache/datafusion/blob/f17cc09fb839431b469e7c707364c1cf99042650/datafusion/physical-plan/src/filter.rs#L514

I think once we add some protection there this should be gtg

vortex-datafusion/src/convert/ranges.rs
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use std::ops::Range;
```
a10y, line 4: this just got moved out of opener.rs since it was getting long, and having these tests there was distracting

vortex-datafusion/src/convert/exprs.rs
```diff
@@ -267,6 +272,7 @@ fn supported_data_types(dt: &DataType) -> bool {
                 | Timestamp(_, _)
                 | Time32(_)
                 | Time64(_)
+                | Struct(_)
```
a10y, line 277: I think we should probably expand this list further? I just added Struct to make one of the tests pass

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -294,17 +293,22 @@ impl FileOpener for VortexOpener {
...
 
-                    make_vortex_predicate(&exprs).transpose()
-                })
-                .transpose()
-                .map_err(|e| DataFusionError::External(e.into()))?;
+                    for expr in &exprs {
+                        if !can_be_pushed_down(expr, &predicate_file_schema) {
+                            internal_datafusion_err!("DataFusion predicate {expr} cannot be pushed down to Vortex file {} with schema {predicate_file_schema}",
+                            file_meta.object_meta.location);
+                        }
```
a10y: here be dragons when we have filters being applied to a column that doesn't exist in every file in the source.

if we have filters that touch columns which are not in the file's physical schema, we can't just skip it because the default value returned by the schema adapter might actually have failed the filter

vortex-datafusion/src/convert/exprs.rs
```diff
@@ -207,6 +207,8 @@ pub(crate) fn can_be_pushed_down(df_expr: &PhysicalExprRef, schema: &Schema) ->
         return false;
     }
 
+    let _expr_str = format!("{:?}", df_expr);
```
robert3005: I assume this is debugging? `dbg!` is useful in these cases

vortex-datafusion/src/convert/exprs.rs
robert3005: I think we are missing List/ListView/FixedSiedList and interval types. Map and Union as well

vortex-datafusion/src/persistent/opener.rs
asubiotto: I'm not sure about this behavior though, doesn't this break cases where we would like to push down an expression as much as possible *even* if it operates on a column that doesn't exist in certain files? Wouldn't an error here error the whole query out?

asubiotto:
I had a more general message on discord:

> One other thing I'd like to add is that IIRC, if the file source returns that it cannot push down a filter, there is currently no way to communicate that filter to the vortex opener. This is currently done after a successful try_pushdown_filters by setting source.vortex_predicate. Parquet has a with_predicate option so whoever's creating a data source can independently pass down a filter that will be best-effort applied to the scan layer with no guarantees to readers.

> Mis-applying filters on missing columns filled in by the schema adapter is a more general concern, so I think short-term we should just do what everyone else does. We just need to make sure that even if we return that we can't push down filters at the source level, we still apply these in the opener.

But related to this PR is the opener erroring out on the predicate. I think it should *not* error even if the expression is on a missing column.

vortex-datafusion/src/persistent/opener.rs
a10y: Yes you're right, the correct thing to do is to ignore any predicates over missing columns in the opener (`develop` behavior) and just rely on DF to post filter for us

a10y:
Here's GHArchive query `SELECT * FROM events WHERE payload.ref = 'main'` with current state of this branch: https://share.firefox.dev/481IcOj

Post-filtering the string match in DF is actually a trivial amount of the overall runtime (0.3%)

image

The bigger problem is that when we tell DF that we can't push the filter, it prompts us to return a projection of `payload`, so it forces us to read and decode the entire payload column and all of its nested fields. So instead of getting a 10x speedup it's more like a 35% speedup. boo!

vortex-datafusion/src/convert/exprs.rs
a10y: This was leftover from debugging, but the rationale was that I wanted a string variable when I stepped thru this in the debugger.

a10y:
Closing this. This has been fixed largely in upstream DF plus some pushdown fixes, which have been integrated into #5189

PR #5414, opened 2025-11-20
feat[vortex-array] add an array session rule registry

joseph-isaacs:


vortex-array/src/encoding.rs
```diff
@@ -87,6 +93,22 @@ impl<V: VTable> Encoding for EncodingAdapter<V> {
...
+    ) -> VortexResult<ArrayRef> {
+        V::build(
+            &self.0,
+            array.dtype(),
+            array.len(),
+            &V::metadata(array.as_::<V>())?,
+            array.buffers().as_slice(),
+            children,
+        )
+        .map(|a| a.to_array())
```
gatesn: `into_array`?

vortex-array/src/lib.rs
```diff
@@ -87,6 +97,49 @@ impl ArraySession {
...
+        &self,
+        child_encoding: &Child::Encoding,
+        parent_encoding: &Parent::Encoding,
+        rule: R,
+    ) where
+        Child: VTable,
+        Parent: VTable,
+        R: 'static + ArrayParentReduceRule<Child, Parent>,
+    {
+        self.rewrite_rules.register_parent_rule::<Child, Parent, R>(
```
gatesn, line 122: Is it worth doing the lifting here? Rather than duplicating the API with generics?

vortex-array/src/arrays/bool/vtable/operator.rs
```diff
@@ -30,23 +31,29 @@ impl OperatorVTable<BoolVTable> for BoolVTable {
...
-                .into_array(),
-            ));
-        }
-
-        Ok(None)
+        // Merge the parent's validity mask into the child's validity
+        Ok(Some(
+            BoolArray::from_bit_buffer(
+                array.bit_buffer().clone(),
+                array.validity().clone().and(parent.validity().clone()),
```
gatesn, line 55: Is the assumption that the `and` compute function will be lazy also? I think this code will eagerly do the intersection atm

gatesn (commented):
Nice

vortex-array/src/arrays/bool/vtable/operator.rs
joseph-isaacs: I just copied this over from the operator table, can add a todo?

gatesn (approved):

Issue #5412, opened 2025-11-20
panic: failed to create ListView views to the end of the elements array must have size 0


asubiotto:
### Describe the bug

Merging a couple of testing files with `0.56.0` + #5390  causes this panic:
```
thread 'bisect_opt::tests::test_minimal_repro' panicked at /Users/asubiotto/.cargo/git/checkouts/vortex-ea6777c0fea02158/3a714aa/vortex-error/src/lib.rs:346:33:
Failed to crate `ListViewArray`: views to the end of the elements array (length 3100) must have size 0
Backtrace:
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2: std::backtrace::Backtrace::create
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/backtrace.rs:331:13
   3: vortex_array::arrays::listview::array::validate_offsets_and_sizes
             at /Users/asubiotto/.cargo/git/checkouts/vortex-ea6777c0fea02158/3a714aa/vortex-array/src/arrays/listview/array.rs:426:13
   4: vortex_array::arrays::listview::array::ListViewArray::validate
             at /Users/asubiotto/.cargo/git/checkouts/vortex-ea6777c0fea02158/3a714aa/vortex-array/src/arrays/listview/array.rs:243:17
   5: vortex_array::arrays::listview::array::ListViewArray::new_unchecked
             at /Users/asubiotto/.cargo/git/checkouts/vortex-ea6777c0fea02158/3a714aa/vortex-array/src/arrays/listview/array.rs:179:13
   6: vortex_array::arrays::listview::rebuild::<impl vortex_array::arrays::listview::array::ListViewArray>::rebuild_trim_elements
             at /Users/asubiotto/.cargo/git/checkouts/vortex-ea6777c0fea02158/3a714aa/vortex-array/src/arrays/listview/rebuild.rs:205:13
   7: vortex_array::arrays::listview::rebuild::<impl vortex_array::arrays::listview::array::ListViewArray>::rebuild_make_exact
[...]

PR #5417, opened 2025-11-20
fix[vortex-array]: don't chunk elements in ListViewArray.naive_rebuild

asubiotto:
Opening this PR in case this change makes sense. I don't really know what I'm doing, so happy to close if not.

### DIFF
--- a/vortex-array/src/arrays/listview/rebuild.rs
+++ b/vortex-array/src/arrays/listview/rebuild.rs
@@ -7,7 +7,8 @@ use vortex_dtype::{IntegerPType, Nullability, match_each_integer_ptype};
 use vortex_error::VortexExpect;
 use vortex_scalar::Scalar;
 
-use crate::arrays::{ChunkedArray, ListViewArray};
+use crate::arrays::ListViewArray;
+use crate::builders::builder_with_capacity;
 use crate::vtable::ValidityHelper;
 use crate::{Array, IntoArray, ToCanonical, compute};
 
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -138,15 +139,17 @@ impl ListViewArray {
         let offsets = offsets.into_array();
         let sizes = sizes.into_array();
 
-        // SAFETY: all chunks were sliced from the same array so have same DType.
-        let elements =
-            unsafe { ChunkedArray::new_unchecked(chunks, element_dtype.as_ref().clone()) };
+        let mut builder = builder_with_capacity(element_dtype.as_ref(), n_elements.as_());
```
robert3005, line 142: I think what you want to do is create a builder instead of vec of chunks and keep appending slices to the builder

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -138,15 +139,17 @@ impl ListViewArray {
...
+        for chunk in &chunks {
+            chunk.append_to_builder(builder.as_mut());
```
a10y, line 144: could also do `builder.extend_from_array(chunk.as_ref())`

asubiotto:
TFTRs! Definitely makes sense to extend the builder directly. Will try this out first to see if it helps resolve the CPU and mem issue

connortsui20:
@asubiotto I am working on a fix to #5412 that changes a few things here, do you mind if I take these changes and put them on my own branch? I can address the above comments on the other branch :smile:

asubiotto:
Sure!

connortsui20:
Closing in favor of #5423

PR #5423, opened 2025-11-20
Fix: incorrect offset for nulls + don't chunk elements

connortsui20:
Potentially fixes? https://github.com/vortex-data/vortex/issues/5412

A continuation of https://github.com/vortex-data/vortex/pull/5417

# Notes

There was an issue where the offset being appended for null lists was the previous offset instead of the previous offset + previous size which broke the invariant of no overlaps. The `validate_zctl` failed to catch this edge case. This then broke `rebuild_trim_elements` since it assumes that everything is in order, including the last offset. This PR fixes that bug.

Additionally, this PR cleans up several things in the `ListView` implementation and improves the performance of a `naive_rebuild` and `list_from_list_view`.

On develop on my AMD 7950X:

```
Timer precision: 10 ns
listview_rebuild  fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ rebuild_naive  158.8 µs      │ 175.3 µs      │ 160.7 µs      │ 162.6 µs      │ 100     │ 100000
```

On this branch:

```
Timer precision: 10 ns
listview_rebuild  fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ rebuild_naive  110 µs        │ 149.1 µs      │ 110.7 µs      │ 111.8 µs      │ 100     │ 100000
```

### DIFF
--- a/encodings/zstd/benches/listview_rebuild.rs
+++ b/encodings/zstd/benches/listview_rebuild.rs
@@ -10,7 +10,7 @@ use vortex_array::validity::Validity;
 use vortex_buffer::Buffer;
 use vortex_zstd::ZstdArray;
 
-#[divan::bench]
+#[divan::bench(sample_size = 1000)]
 fn rebuild_naive(bencher: Bencher) {
     let dudes = VarBinViewArray::from_iter_str(["Washington", "Adams", "Jefferson", "Madison"])
         .into_array();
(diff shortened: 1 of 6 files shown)


a10y (approved):
this should fix the trailing NULLs thing, i'm not sure if it will have much of a toll on the memory issues.

I think we'd need to find a way to avoid canonicalizing the whole elements, but maybe that should be happening upstream.

PR #5426, opened 2025-11-20
feat[vortex-array] chunked array has owned chunk_offsets child

joseph-isaacs:
We should ensure that visitors are given ref to externally owned array

### DIFF
--- a/vortex-array/src/arrays/chunked/vtable/visitor.rs
+++ b/vortex-array/src/arrays/chunked/vtable/visitor.rs
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use crate::arrays::{ChunkedArray, ChunkedVTable, PrimitiveArray};
-use crate::validity::Validity;
+use crate::arrays::{ChunkedArray, ChunkedVTable};
 use crate::vtable::VisitorVTable;
 use crate::{ArrayBufferVisitor, ArrayChildVisitor};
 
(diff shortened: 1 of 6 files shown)


vortex-array/src/arrays/chunked/vtable/visitor.rs
```diff
@@ -1,18 +1,15 @@
...
+use crate::arrays::{ChunkedArray, ChunkedVTable};
 use crate::vtable::VisitorVTable;
 use crate::{ArrayBufferVisitor, ArrayChildVisitor};
 
 impl VisitorVTable<ChunkedVTable> for ChunkedVTable {
     fn visit_buffers(_array: &ChunkedArray, _visitor: &mut dyn ArrayBufferVisitor) {}
 
     fn visit_children(array: &ChunkedArray, visitor: &mut dyn ArrayChildVisitor) {
-        let chunk_offsets =
-            PrimitiveArray::new(array.chunk_offsets().clone(), Validity::NonNullable);
```
gatesn, line 14: But this doesn't allocate?

robert3005: I think allocation is not the right term, it's that we only ever visit things that are held by array and not temporaries that exist only in the visit call

joseph-isaacs: I want to be able to hold on to a ref to this outside of the visit.

joseph-isaacs: Also the visitors themself likely have to allocate to hold the child

robert3005 (approved):

PR #5432, opened 2025-11-20
Chore: add set functions to mask mut

connortsui20:


vortex-mask/src/mask_mut.rs
```diff
@@ -355,6 +355,104 @@ impl MaskMut {
...
+            match &mut self.0 {
+                Inner::Empty { .. } => {
+                    // In debug mode, we still want to catch this error.
+                    debug_assert!(false, "cannot set value in empty mask");
+                }
+                Inner::Constant {
+                    value: current_value,
+                    ..
+                } => {
+                    if *current_value != value {
```
gatesn, line 451: Do you want the debug_assert for out of bounds here?

gatesn (approved):

PR #5433, opened 2025-11-21
fix: FFI returns a string/binary instead of requiring the caller to guess the necessary memory

### LINKED ISSUE #5373: FFI unsafe API endpoints
### Describe the bug

```c
void vx_array_get_utf8(const vx_array *array, uint32_t index, void *dst, int *len);
void vx_array_get_binary(const vx_array *array, uint32_t index, void *dst, int *len);
void vx_dtype_time_zone(const DType *dtype, void *dst, int *len);
```

Both return the length written in the DST buffer after getting the value. However this is very unsafe as there is no way to tell Vortex what the size of the buffer is. Nor is there a way to get the size of the string before this call.

robert3005:
fix: #5373 

I think we could have these do RAII but we need something better than just scalar_at

### DIFF
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -274,6 +274,11 @@ typedef struct vx_array_iterator vx_array_iterator;
 typedef struct vx_array_sink vx_array_sink;
 
 /**
+ * Strings for use within Vortex.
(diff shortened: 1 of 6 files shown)


joseph-isaacs (approved):
Nnice

PR #5440, opened 2025-11-21
chore: Shard codspeed into 8 runs

robert3005:


0ax1 (approved):
Thanks for doing this. But sad state of affairs that this is necessary.

robert3005:
I was between removing it and letting claude do this rewrite, seems claude did mostly correct thing so let's keep it. I think if there are more issues I will just yank it.

0ax1:
> I was between removing it and letting claude do this rewrite, seems claude did mostly correct thing so let's keep it. I think if there are more issues I will just yank it.

I think the main issue lately has been that we stopped trusting the benchmarks which renders them useless.

PR #5443, opened 2025-11-21
perf[alp]: compare |- alp pushdown

joseph-isaacs:


vortex-file/src/lib.rs
```diff
@@ -179,4 +179,6 @@ pub fn register_default_encodings(session: &VortexSession) {
         #[cfg(feature = "zstd")]
         EncodingRef::new_ref(vortex_zstd::ZstdEncoding.as_ref()),
     ]);
+
+    register_alp_rules(&session.arrays())
```
gatesn: I think we should just have `vortex_alp::initialize(session)`

So there's a single entry-point for the crate to register all its things (including its encodings!)

encodings/alp/src/alp/compute/mod.rs
```diff
@@ -4,11 +4,15 @@
...
 mod cast;
 mod compare;
+mod compare_common;
+mod expr_pushdown;
 mod filter;
 mod mask;
 mod nan_count;
 mod take;
 
+pub use expr_pushdown::ALPExprPushdownRule;
```
gatesn, line 14: Why pub?

encodings/alp/src/alp/mod.rs
```diff
@@ -9,7 +9,7 @@ use num_traits::{CheckedSub, Float, PrimInt, ToPrimitive};
 
 mod array;
 mod compress;
-mod compute;
+pub mod compute;
```
gatesn, line 12: Why pub?

encodings/alp/src/alp/compute/expr_pushdown.rs
```diff
@@ -0,0 +1,492 @@
...
+                | vortex_array::expr::Operator::Lt
+                | vortex_array::expr::Operator::Lte
+                | vortex_array::expr::Operator::Gt
+                | vortex_array::expr::Operator::Gte
+        ) {
+            return Ok(None);
+        }
+
+        // Convert to compute operator
+        let compute_op = match operator {
```
gatesn: Feels like this does the same logic as the matches! macro on line 62?

encodings/alp/src/alp/compute/expr_pushdown.rs
```diff
@@ -0,0 +1,492 @@
...
+                (binary_view.lhs(), compute_op.swap())
+            } else {
+                return Ok(None);
+            };
+
+        // Get the literal scalar - literals evaluate to a constant array with one element
+        let literal_value = literal_expr.as_::<Literal>().data().clone();
+
+        // Don't optimize nullable comparisons
+        if literal_value.dtype().is_nullable() {
```
gatesn, line 82: I'm 90% sure this is covered by the null checks above

encodings/alp/src/alp/compute/expr_pushdown.rs
```diff
@@ -0,0 +1,492 @@
...
+            return Ok(None);
+        }
+
+        // Convert to primitive scalar
+        let Ok(pscalar) = PrimitiveScalar::try_from(&literal_value) else {
```
gatesn: There should be a conversion that uses Option rather than Result?

gatesn (commented):
This is... kinda nice??

I guess there's a lot more stuff to pull apart than just the eager compute function.
I wonder if after writing a few more of these there are certain patterns we have to extract sub-expressions out of an ExprArray and push them through another array?

encodings/alp/src/alp/compute/expr_pushdown.rs
joseph-isaacs: me too but also only 90

joseph-isaacs:
Yep, I agree I just wanted to see what it would be like.

joseph-isaacs:
The other approach for scalar arrays. is to convert this into an expression and create a reduce_parent expr rule for `alp($) > x`

encodings/alp/src/alp/compute/mod.rs
joseph-isaacs: vortex_alp::initialize(session)

gatesn: pub(crate) at most ?

joseph-isaacs:
TODO: convert this to reduce parent

PR #5444, opened 2025-11-21
Fix: remove unreachable cases in bitpack patches

connortsui20:
note that we probably shouldnt be working with `Mask` at all, but this code is going to be replaced by the vector rewrite and I've already fixed that.

### DIFF
--- a/encodings/fastlanes/src/bitpacking/array/bitpack_decompress.rs
+++ b/encodings/fastlanes/src/bitpacking/array/bitpack_decompress.rs
@@ -10,7 +10,7 @@ use vortex_array::patches::Patches;
 use vortex_dtype::{
     IntegerPType, NativePType, match_each_integer_ptype, match_each_unsigned_integer_ptype,
 };
-use vortex_error::VortexExpect;
+use vortex_error::{VortexExpect, vortex_panic};
 use vortex_mask::Mask;
 use vortex_scalar::Scalar;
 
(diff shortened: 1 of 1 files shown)


encodings/fastlanes/src/bitpacking/array/bitpack_decompress.rs
```diff
@@ -103,24 +103,12 @@ fn insert_values_and_validity_at_indices_to_uninit_range<
...
-                let out_index = index.as_() - indices_offset;
-                dst.set_value(out_index, f(value));
-                dst.set_validity_bit(out_index, vb.value(out_index));
-            }
-        }
+    let Mask::AllTrue(_) = values_validity else {
+        vortex_panic!("BitPackedArray somehow had nullable patch values");
+    };
+
+    for (index, &value) in indices.iter().zip_eq(values) {
```
robert3005, line 110: I think at this point it's easier to inline this to the caller and remove some arguments

connortsui20: actually having the function makes the trait bounds easier

robert3005: If you say so, I was hoping it would be opposite but I believe it

robert3005 (approved):

encodings/fastlanes/src/bitpacking/array/bitpack_decompress.rs
connortsui20: yeah unfortunate

PR #5446, opened 2025-11-21
Feature: implement bitpacking batch execute

connortsui20:
Adds unpacking directly into buffers instead of a builder.


connortsui20:
Ideally the patches should be implemented as a `PatchArray` that then can fuse into things like `BitPackedArray` and `ALPArray` as new arrays (could be generic wrappers over the existing arrays) that have their own execute implementations, but that can come after we migrate everything.

encodings/fastlanes/src/bitpacking/array/bitpack_decompress.rs
```diff
@@ -7,16 +7,118 @@ use vortex_array::ToCanonical;
...
+/// Unpacks a bit-packed array into a generic [`PVectorMut`].
+pub fn unpack_to_pvector<P: BitPacked>(array: &BitPackedArray) -> PVectorMut<P> {
+    if array.is_empty() {
+        return PVectorMut::with_capacity(0);
+    }
+
+    let len = array.len();
+    let mut elements = BufferMut::<P>::with_capacity(len);
+    let uninit_slice = &mut elements.spare_capacity_mut()[..len];
+    debug_assert_eq!(uninit_slice.len(), len);
```
gatesn: This seems a little excessive!

encodings/fastlanes/src/bitpacking/vtable/operator.rs
```diff
@@ -7,7 +7,8 @@ use vortex_array::vtable::OperatorVTable;
 use crate::{BitPackedArray, BitPackedVTable};
 
 impl OperatorVTable<BitPackedVTable> for BitPackedVTable {
```
gatesn: Just delete the whole file, we don't need it

connortsui20:
whoops accidentally added that last commit to this branch

gatesn (approved):

PR #5453, opened 2025-11-21
fix: handling of struct fields in Spark

a10y:
Fixes some issues I encountered trying to get some files with nested structs/lists to load.

I'm still hitting a very strange issue deep in C Arrow Data interface with the 3.5GB github archives dataset, but this fixes some other issues.

### DIFF
--- a/vortex-jni/src/errors.rs
+++ b/vortex-jni/src/errors.rs
@@ -76,7 +76,7 @@ impl JNIDefault for jobject {
     }
 }
 
-/// Run the provided function inside of the JNIEnv context. Throws an exception if the function returns an error.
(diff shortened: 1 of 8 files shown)


vortex-jni/src/errors.rs
```diff
@@ -87,12 +87,21 @@ where
     match function(env) {
         Ok(result) => result,
         Err(error) => {
+            // Propagate the exception instead of throwing our own.
+            if env
+                .exception_check()
+                .expect("checking exception should succeed")
+            {
+                return T::jni_default();
+            }
```
a10y, line 96: if the inner JNI code returned Err b/c an exception was thrown, we weren't properly propagating that

robert3005 (approved):

PR #5503, opened 2025-11-24
Feature: add a take compute trait

connortsui20:
In the new operator world, `take` as a compute function just becomes a `DictArray` (since arrays are lazy), so it is nice to have `take` as a standalone trait (like `filter`).


gatesn (commented):
You may want something similar to FilterMask which provides a `len` function on the arbitrary Indices object

gatesn (approved):

PR #5507, opened 2025-11-24
Expression execution to be passed input vectors

gatesn:
This execution model trades off short-circuiting in favor of CSE.

It also allows expressions to be executed independently of the expression tree, which may be useful depending on how we ultimately end up wrapping up scalar expressions into arrays.

@joseph-isaacs the alternative would be to pass a `trait ExecutionArgs` and then calling `child(usize)` could lazily compute the child, bringing back the flexibility of short-circuiting depending on how the expression is ultimately executed.

### DIFF
--- a/vortex-array/src/expr/vtable.rs
+++ b/vortex-array/src/expr/vtable.rs
@@ -78,13 +78,9 @@ pub trait VTable: 'static + Sized + Send + Sync {
     fn evaluate(&self, expr: &ExpressionView<Self>, scope: &ArrayRef) -> VortexResult<ArrayRef>;
 
     /// Execute the expression on the given vector with the given dtype.
-    fn execute(
-        &self,
-        _expr: &ExpressionView<Self>,
-        _vector: &Vector,
-        _dtype: &DType,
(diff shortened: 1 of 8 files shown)


vortex-array/src/expr/vtable.rs
```diff
@@ -237,33 +240,32 @@ impl<V: VTable> DynExprVTable for VTableAdapter<V> {
...
-                vector_matches_dtype(&result, &expected_dtype),
-                "Expression execution invalid for dtype {}",
-                expected_dtype
-            );
-        }
+        vortex_ensure!(
+            vector_matches_dtype(&result, &expected_dtype),
+            "Expression execution invalid for dtype {}",
+            expected_dtype
+        );
```
connortsui20, line 268: This seems kind of strange to me, might as well just debug assert here?

connortsui20 (approved):
Its not super clear to me how the execution arguments actually get constructed but this seems fine for now to make progress

connortsui20:
oh whoops auto merge was on

PR #5521, opened 2025-11-25
chore: apply imports_granularity = "Item" everywhere expect bench-vortex

joseph-isaacs:
requires some non-auto changes to to `bench-vortex`


0ax1 (approved):
🐘

PR #5528, opened 2025-11-25
feat[vortex-expr]: add a `is_null_sensitive` expression analysis

joseph-isaacs:


vortex-array/src/expr/analysis/labeling.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use crate::expr::Expression;
+use crate::expr::traversal::NodeExt;
+use crate::expr::traversal::NodeVisitor;
+use crate::expr::traversal::TraversalOrder;
+
+/// Label each node in an expression tree using a bottom-up traversal.
+///
+/// This function separates tree labeling into two distinct steps:
+/// 1. **Label Edge**: Compute a label for each node based only on the node itself
```
0ax1: Why is it called an edge if it is a label on the node, and not label for the connection between nodes?

0ax1 (commented):
Question about the wording/terminology.

vortex-array/src/expr/analysis/labeling.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl<'a, 'b, L: Clone, F, G> NodeVisitor<'a> for LabelingVisitor<'a, 'b, L, F, G>
+where
+    F: Fn(&Expression) -> L,
+    G: FnMut(L, &L) -> L,
+{
+    type NodeTy = Expression;
+
+    fn visit_down(&mut self, _node: &'a Self::NodeTy) -> VortexResult<TraversalOrder> {
```
0ax1: Do we always want to test through `label_tree` or should we have one test each extra for visit up/down directly?

0ax1 (approved):
lgtm. but left 2 remarks to consider.

vortex-array/src/expr/analysis/labeling.rs
joseph-isaacs: fixed this

vortex-array/src/expr/analysis/labeling.rs
joseph-isaacs: I am not sure what you mean here?

0ax1: Saying we never test visit_down/down directly. But I also don't think this is an issue.

PR #5530, opened 2025-11-25
chunked array zip

onursatici:
chunked array to push zip down to chunks

### DIFF
--- a/vortex-array/src/arrays/chunked/compute/zip.rs
+++ b/vortex-array/src/arrays/chunked/compute/zip.rs
@@ -13,6 +13,9 @@ use crate::compute::ZipKernelAdapter;
 use crate::compute::zip;
 use crate::register_kernel;
 
+// Push down the zip call to the chunks. Without this kernel
+// the default implementation canonicalises the chunked array
+// then zips once.
 impl ZipKernel for ChunkedVTable {
     fn zip(
         &self,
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/chunked/compute/zip.rs
joseph-isaacs, line 1: cann you explain the optimisation here

joseph-isaacs: not canoncalizing?

onursatici: push down the operation to chunks, so we don't canonicalise the chunked array in the default zip implementation

joseph-isaacs: please add this as a comment

vortex-array/src/arrays/chunked/compute/zip.rs
```diff
@@ -0,0 +1,125 @@
...
+
+            if lhs_offset == lhs_chunk.len() {
+                lhs_idx += 1;
+                lhs_offset = 0;
+            }
+            if rhs_offset == rhs_chunk.len() {
+                rhs_idx += 1;
+                rhs_offset = 0;
+            }
+        }
```
joseph-isaacs, line 67: is it worth making an intersection iterator? Do we think other binary ops will be pushed down?

joseph-isaacs (approved):
Add the explanation and me can merge

vortex-array/src/arrays/chunked/compute/zip.rs
onursatici: Yea it would be cool to have this iterator work on all elementwise compute functions. I am not sure what would be the best analogue to that in the operators world though, but might be worth doing that iterator in the new world

PR #5531, opened 2025-11-25
feat[vortex-expr]: `is_fallible` expression analysis

joseph-isaacs:
For now we put this on the VTable, but I think we should figure out a way to make this analysis extensible


vortex-array/src/expr/exprs/get_item/mod.rs
```diff
@@ -155,6 +155,11 @@ impl VTable for GetItem {
     fn is_null_sensitive(&self, _instance: &Self::Instance) -> bool {
         true
     }
+
+    fn is_fallible(&self, _instance: &Self::Instance) -> bool {
+        // if this type checks its infallible
```
0ax1: Let's do a proper sentence here with punctuation.

vortex-array/src/expr/exprs/select/mod.rs
```diff
@@ -201,6 +201,15 @@ impl VTable for Select {
             .collect();
         Ok(unsafe { StructVector::new_unchecked(Arc::new(new_fields), mask) }.into())
     }
+
+    fn is_null_sensitive(&self, _instance: &Self::Instance) -> bool {
+        true
+    }
+
+    fn is_fallible(&self, _instance: &Self::Instance) -> bool {
+        // if this type checks its infallible
```
0ax1: -//-

0ax1 (approved):

PR #5536, opened 2025-11-25
Feature: make `Cast` trait have arbitrary outputs

connortsui20:
We should allow `Cast` to operate over things other than vectors.

(I need this to implement some things for f16)


connortsui20:
actually I'm realizing this isn't sufficient for what I want to do:

```
// Doesnt compile because U is not bound
impl<T, U> Cast for Buffer<T> {
    type Output = Buffer<U>;

    fn cast(&self, dtype: ???) -> VortexResult<Buffer<U>> {
        todo!()
    }
}
```

PR #5537, opened 2025-11-25
Feature: add `take` compute for slices

connortsui20:
Ports over the `take` implementation on primitive arrays to `vortex-compute` for `take` on generic slices.


vortex-compute/src/take/slice/portable.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/// Takes elements from an array using SIMD indexing.
+///
+/// Performs a gather operation that takes values at specified indices and returns them in a new
+/// buffer. Uses SIMD instructions to process `LANE_COUNT` indices in parallel.
+///
+/// Returns a `Buffer<T>` where each element corresponds to `values[indices[i]]`.
+#[multiversion(targets("x86_64+avx2", "x86_64+avx", "aarch64+neon"))]
+fn take_portable_simd<T, I, const LANE_COUNT: usize>(values: &[T], indices: &[I]) -> Buffer<T>
```
0ax1: Can you clarify what the goal of this PR is? The current implementation of `compute/take/portable/take_portable_simd` is identical.

0ax1: `take_portable_simd` and `take_primitive_scalar` take slices as input parameters already.

connortsui20: yes this is just moving them into `vortex-compute` since we're eventually going to delete all of the compute functions

0ax1: Ah right, we're just not doing it as part of this PR.

0ax1: Is it maybe better to do that move in one big PR atomically?

vortex-vector/src/primitive/macros.rs
```diff
@@ -184,3 +220,39 @@ macro_rules! match_each_integer_pvector_mut {
...
+/// uniformly across all unsigned type variants (`U8`, `U16`, `U32`, `U64`).
+///
+/// See [`match_each_pvector_mut`] for similar usage.
+///
+/// [`PrimitiveVectorMut`]: crate::primitive::PrimitiveVectorMut
+///
+/// # Panics
+///
+/// Panics if the vector passed in to the macro is not an unsigned vector variant.
+#[macro_export]
```
0ax1, line 237: Why do we introduce this twice?

vortex-compute/src/take/slice/portable.rs
connortsui20: What I really should do is move that implementation here and have the primitive array call into this, but I want to wait until I add the avx2 impl plus the pvector take impl (which will deal with nullable indices)

vortex-vector/src/primitive/macros.rs
connortsui20: this one is for mutable, and the only reason I also added this is for symmetry

vortex-compute/src/take/slice/portable.rs
0ax1: Yeah - However we sequence this, it'd be good to not to have duplicated code intermediately.

connortsui20: I've changed it so that the existing portable compute function on the array calls into this new one in `vortex-compute`

a10y:
Are we going to handle nullability or is that handled a level up?

connortsui20:
@a10y yeah I think that it doesnt really make sense to handle nullable indices for a plain buffer, once we add take on things like vectors then we can take a primitive array or pvector as indices

joseph-isaacs:
Can we name this PR appropriately?

connortsui20:
@joseph-isaacs why is it not named appropriately? Do you mean it should be named a migration instead?

connortsui20:
ok I'm just going to port it over now to make things cleaner

cc @a10y since I believe you were the one who implemented the AVX2 impl? Everything is identical except that I pulled out the validity (which was just getting passed in so that it could construct a `PrimitiveArray`) and I changed a buffer transmute to use a new `cast_into` method I added on `Buffer`.

vortex-compute/src/take/slice/portable.rs
connortsui20: ok I've just ported the whole thing over

a10y (approved):
lgtm, avx2 tests are running in CI and still passing

PR #5538, opened 2025-11-25
Feature: add `take` for `BitBuffer`

connortsui20:
I tried to make the existing `BoolArray::take` function call into this (so we don't duplicate too much code) but because we have allowed signed integers as indices this fails to compile / fails tests because some of our tests use signed integers as indices


connortsui20:
closing in favor of https://github.com/vortex-data/vortex/pull/5540

PR #5540, opened 2025-11-25
Feature: add `take` for primitive and bool vectors

connortsui20:
Adds `take` implementations and tests for:

- `BitBuffer`
- `Mask`
- `BoolVector`
- `PrimitiveVector`

For `BitBuffer`, I don't think it makes sense to implement take with nullable indices since I would argue that there isn't a clear behavior to have (because it might not be obvious that `true` means not null and `false` means null). But for `Mask` it feels a bit different, we are using it for validity in all of our vectors, so I think it makes a bit more sense (and it will make the implementation of `take` for the rest of the vectors easier).

We represent indices in 3 different ways here (a generic indices slice of unsigned integers, generic unsigned pvector, and primitive vector). I feel quite strongly that we should not allow signed indices, as that opens the door to too many bugs. 

This is also based on top of https://github.com/vortex-data/vortex/pull/5537, so that needs to merge first.


vortex-compute/src/take/vector/bool.rs
```diff
@@ -0,0 +1,49 @@
...
+    // determine which elements are null.
+    let taken_bits = bvector.bits().take(indices.elements().as_slice());
+    let taken_validity = bvector.validity().take(indices);
+
+    debug_assert_eq!(taken_bits.len(), taken_validity.len());
+
+    // SAFETY: We used the same indices to take from both components, so they should still have the
+    // same length.
+    unsafe { BoolVector::new_unchecked(taken_bits, taken_validity) }
+}
```
joseph-isaacs, line 49: OOI, why not impl this over a PVector?

connortsui20: not sure I understand? I do have an implementation for this in `vortex-compute/src/take/vector/pvector.rs`

joseph-isaacs: I am saying its odd we have a trait and then break out of it

vortex-compute/src/take/vector/tests.rs
```diff
@@ -0,0 +1,283 @@
...
+    // - Index 0 -> data[0] = 10 (valid)
+    // - Index null -> null (null index produces null)
+    let PrimitiveVector::I32(result) = result else {
+        panic!("Expected I32 variant");
+    };
+    assert_eq!(result.get(0), Some(&50));
+    assert_eq!(result.get(1), None); // Null index.
+    assert_eq!(result.get(2), None); // data[2] is null.
+    assert_eq!(result.get(3), Some(&10));
+    assert_eq!(result.get(4), None); // Null index.
```
joseph-isaacs: can we assert that two vectors are equal instead of .get? Just much more concise

joseph-isaacs:
How do we support arrow dict without signed take?

joseph-isaacs:
Do we have benchmarking for old dict/take that would be move over and compare

vortex-compute/src/take/vector/bool.rs
connortsui20: I mean it's just so I don't have to inline it above? This will also be more consistent with other take implementations on the more complex vectors that will be much more verbose than this

vortex-compute/src/take/vector/tests.rs
connortsui20: We don't have vector equality implemented so I'll just do manual helper functions for now and we can change it later.

connortsui20:
> How do we support arrow dict without signed take?

We can just do a transmute. If Arrow decides to give us a negative index, then that is their problem, not ours (we will panic with out of bounds and it should be a very obvious error message).

Edit: Also, we've had unsigned codes in our dictionaries for a while and I don't think that has caused many issues?

> Do we have benchmarking for old dict/take that would be move over and compare

We have some take benchmarks for the existing take compute kernels and those aren't going anywhere. The existing impls also aren't going anywhere so we can still benchmark them (this PR isn't removing anything)

vortex-compute/src/take/vector/bool.rs
joseph-isaacs: Dont worry

joseph-isaacs:
I really want to add match the benchmarks from the old system when we add new capabilities to the new one otherwise we will not.

connortsui20:
I can add benchmarks comparing `take` on a primitive vector vs canonicalization of `DictArray -> PrimitiveArray`

vortex-compute/src/take/vector/tests.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    let values: Vec<bool> = (0..result.len()).map(|i| result.value(i)).collect();
+    assert_eq!(values, vec![false, true, true, true, false]);
+
+    // Large buffer (uses take_bool path, len > 4096).
+    let large: BitBuffer = (0..5000).map(|i| i % 3 == 0).collect();
+    let result = (&large).take(&[4999u32, 0, 1, 2, 3, 4998][..]);
+
+    let values: Vec<bool> = (0..result.len()).map(|i| result.value(i)).collect();
+    assert_eq!(values, vec![false, true, false, false, true, true]);
```
joseph-isaacs: would be nice to assert_eq over the Vector itself? Is that a long way off?

connortsui20: I can use the `BitIterator` here

vortex-array/benches/take_strings.rs
joseph-isaacs, line 1: we save not replaced these

joseph-isaacs: do you think we don't need them?

vortex-array/benches/take_primitive.rs
```diff
@@ -0,0 +1,101 @@
...
+fn pvector_take_uniform<const VECTOR_SIZE: usize>(bencher: Bencher, num_indices: usize) {
+    let data: Buffer<u32> = (0..VECTOR_SIZE as u32).collect();
+    let pvector = PVector::new(data, Mask::AllTrue(VECTOR_SIZE));
+
+    let rng = StdRng::seed_from_u64(0);
+    let range = Uniform::new(0u32, VECTOR_SIZE as u32).unwrap();
+    let indices: Vec<u32> = rng.sample_iter(range).take(num_indices).collect();
+
+    bencher
+        .with_inputs(|| (&pvector, indices.as_slice()))
```
joseph-isaacs, line 45: this should be with another PVector?

vortex-array/benches/take_strings.rs
connortsui20: tbh they look more like tests than benchmarks

vortex-array/benches/take_primitive.rs
connortsui20: it doesn't really matter because this benchmark doesn't deal with null indices

joseph-isaacs (approved):
Did we think about take in place at all?

PR #5541, opened 2025-11-26
Fix: Schema projection order mismatch from table/file schema

brancz:
Looks like I only solved a subset of the issue with #5436

The test I added failed before the fix.

### DIFF
--- a/vortex-datafusion/src/persistent/opener.rs
+++ b/vortex-datafusion/src/persistent/opener.rs
@@ -230,6 +230,11 @@ impl FileOpener for VortexOpener {
                 DataFusionError::Execution(format!("Failed to convert file schema to arrow: {e}"))
             })?);
 
+            // Compute the logical file schema by merging physical file types with logical table types.
+            // This schema has the same field names as logical_schema, but with physical types from the file.
+            let logical_file_schema =
+                compute_logical_file_schema(&physical_file_schema, &logical_schema);
+
             if let Some(expr_adapter_factory) = expr_adapter_factory {
                 let partition_values = partition_fields
                     .iter()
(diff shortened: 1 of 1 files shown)


brancz:
Coincidentally, this is now the same strategy as what the parquet opener does:

Map file schema to logical schema: https://github.com/apache/datafusion/blob/769f36732a284cec3d549bfcda50cb67684fb5dc/datafusion/datasource-parquet/src/opener.rs#L242
Project: https://github.com/apache/datafusion/blob/769f36732a284cec3d549bfcda50cb67684fb5dc/datafusion/datasource-parquet/src/opener.rs#L319

vortex-datafusion/src/persistent/opener.rs
asubiotto: Do we still need this step or can we now use the ignored schema mapping above?

brancz: Good catch, indeed, we no longer need it!

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -870,4 +841,81 @@ mod tests {
...
+            .open(
+                make_meta(file_path, data_size),
+                PartitionedFile::new(file_path.to_string(), data_size),
+            )?
+            .await?
+            .try_collect::<Vec<_>>()
+            .await?;
+
+        // Verify the columns are in the right order and have the right values
+        use datafusion::arrow::util::pretty::pretty_format_batches_with_options;
```
asubiotto, line 903: nit: I think this probably belongs as a test module import rather than inline

asubiotto (approved):
LGTM

asubiotto:
cc @a10y @AdamGS in case you want to take a look

robert3005 (approved):

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -260,56 +260,27 @@ impl FileOpener for VortexOpener {
...
-            // to select.
-            let (_schema_mapping, adapted_projections) =
-                schema_adapter.map_schema(&physical_file_schema)?;
-
-            // Build the Vortex projection expression using the adapted projections.
-            // This will reorder the fields to match the target order.
+            // Compute the logical file schema by merging physical file types with logical table types.
+            // This schema has the same field names as logical_schema, but with physical types from the file.
+            let logical_file_schema =
+                compute_logical_file_schema(&physical_file_schema, &logical_schema);
```
a10y: if we hoist this up somewhere above line 242 i can we just compute this once

brancz: Good catch! Done!

a10y (approved):

a10y:
@brancz I think there are still issues here, see this simple example: https://github.com/vortex-data/vortex/pull/5550

Hoping to fix all of these edge cases as part of #5406

PR #5548, opened 2025-11-26
feat: alp decompress to vector

0ax1:


encodings/alp/src/alp/array.rs
```diff
@@ -136,6 +141,12 @@ impl VTable for ALPVTable {
             patches,
         )
     }
+
+    fn execute(array: &ALPArray, _ctx: &mut dyn ExecutionCtx) -> VortexResult<Vector> {
+        match_each_alp_float_ptype!(array.dtype().as_ptype(), |T| {
+            Ok(decompress_to_pvector::<T>(array.clone()).freeze().into())
+        })
+    }
 }
```
joseph-isaacs: This should call execute on all children and then apply exponents over the vector itself

encodings/alp/src/alp/decompress.rs
```diff
@@ -0,0 +1,317 @@
...
+    reason = "complexity is from nested match_each_* macros"
+)]
+fn decompress_chunked(
+    array: ArrayRef,
+    exponents: Exponents,
+    patches: &Patches,
+    patches_chunk_offsets: &PrimitiveArray,
+    dtype: DType,
+) -> PrimitiveArray {
+    let encoded = array.to_primitive();
```
joseph-isaacs, line 128: This should be execute right?

encodings/alp/src/alp/array.rs
connortsui20: yes I think I was thinking of a similar thing where we call `execute` on the `encoded` array, downcast into a `PVector`, extract the parts (`Buffer` and `Mask`), and then do a map in place sort of thing after calling `into_mut` on the internal buffer.

Edit: forgot there's already a `decode_in_place`

encodings/alp/src/alp/decompress.rs
```diff
@@ -0,0 +1,317 @@
...
+    // primitive didn't create a copy. In that case both alp_encoded and array
+    // will hold a reference to the buffer we want to mutate.
+    drop(array);
+
+    let validity = encoded.validity().clone();
+    let ptype = dtype.as_ptype();
+
+    let decoded = match_each_alp_float_ptype!(ptype, |T| {
+        PrimitiveArray::new::<T>(
+            <T>::decode_buffer(encoded.into_buffer_mut(), exponents),
```
connortsui20: Similar thing as above here, and this can also be done in place (also note that we did measure in place vs out-of-place and in-place was always faster for arrays with length > 1024)

encodings/alp/src/alp/array.rs
0ax1: @gatesn thoughts?

connortsui20 (dismissed):
Other than Joe's comment, I am wondering if we should just do the "naive" thing here for patches: decompress the entire array and then patch everything instead of doing this in chunks. A few reasons for this:

- It feels a bit strange to have the batch execution use chunks that we will eventually use for pipelining
- It will be nice to see comparisons between batch and pipelined (+ fused patches) execution when we eventually do implement pipelining
- At some point (hopefully very soon 🤞) we want to have a `PatchArray` so that we don't duplicate all of this logic for the other arrays that need patches (bitpacked) for both batch and pipeline execution models

TLDR; Other than the existing review comments, I think this is fine since we need it for pipelining but also I think we still want a proper batched execution for patches for now

0ax1:
> Other than Joe's comment, I am wondering if we should just do the "naive" thing here for patches: decompress the entire array and then patch everything instead of doing this in chunks. A few reasons for this:
> 
> * It feels a bit strange to have the batch execution use chunks that we will eventually use for pipelining
> * It will be nice to see comparisons between batch and pipelined (+ fused patches) execution when we eventually do implement pipelining
> * At some point (hopefully very soon 🤞) we want to have a `PatchArray` so that we don't duplicate all of this logic for the other arrays that need patches (bitpacked) for both batch and pipeline execution models
> 
> TLDR; Other than the existing review comments, I think this is fine since we need it for pipelining but also I think we still want a proper batched execution for patches for now

Note that fused patching is much faster, see: https://github.com/vortex-data/vortex/pull/5065

connortsui20:
I'm aware it will be much faster, but my point is more that this falls under a different execution model that we are punting to later. I guess it is fine for now but we need to remember that we will eventually want batch execution to call `decompress_unchunked` instead and pipelining to call something similar to `decompress_chunked` with the patches logic abstracted away (either as a dedicated array, or if that is too slow, a wrapper type that can implement patching on chunks and entire vectors)

encodings/alp/src/alp/decompress.rs
```diff
@@ -0,0 +1,239 @@
...
+            patches_offset,
+            validity,
+        )
+    } else {
+        unreachable!("ALPFloat can only be f32 or f64")
+    }
+}
+
+fn decompress_from_buffer<T: ALPFloat, IntType>(
+    int_buffer: BufferMut<IntType>,
```
joseph-isaacs: remove this

joseph-isaacs:
```suggestion
    int_buffer: BufferMut<T::ALPInt>,
```

encodings/alp/src/alp/decompress.rs
```diff
@@ -0,0 +1,239 @@
...
+    } else if T::PTYPE == PType::F64 {
+        let int_buffer = encoded_mutable.into_i64().into_parts().0;
+        decompress_from_buffer::<T, i64>(
+            int_buffer,
+            exponents,
+            patches_vectors,
+            patches_offset,
+            validity,
+        )
+    } else {
```
joseph-isaacs: would be nice to have a into::<T> but maybe just a todo here?

encodings/alp/src/alp/decompress.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    // Convert to float buffer.
+    let mut decoded_buffer: BufferMut<T> = unsafe { transmute(alp_buffer) };
+
+    // Apply patches if they exist
+    if let Some((indices_vector, values_vector, _chunk_offsets_vector)) = patches_vectors {
+        let indices_primitive = indices_vector.into_primitive();
+        let values_primitive = values_vector.into_primitive();
+
+        match_each_unsigned_integer_ptype!(indices_primitive.ptype(), |I| {
```
joseph-isaacs: use `match_each_alp_float_ptype` and `I::ALPInt` is what you want?

0ax1: These are the patches indices and values. The patches index type is unrelated to `I::ALPInt`. I renamed the variables to make that clearer.

encodings/alp/src/alp/decompress.rs
```diff
@@ -0,0 +1,235 @@
...
+            let values_buffer = T::downcast(patches_values.into_mut()).into_parts().0;
+
+            let indices_slice = indices_buffer.as_slice();
+            let values_slice = values_buffer.as_slice();
+            let decoded_slice = decoded_buffer.as_mut_slice();
+
+            for (&idx, &value) in indices_slice.iter().zip(values_slice.iter()) {
+                decoded_slice[AsPrimitive::<usize>::as_(idx) - patches_offset] = value;
+            }
+        });
```
joseph-isaacs, line 105: cann you make this macro smaller?

0ax1: done

joseph-isaacs (approved):

PR #5550, opened 2025-11-26
add test to demonstrate schema evolution fail

a10y:


vortex-datafusion/tests/schema_evolution.rs
```diff
@@ -0,0 +1,149 @@
...
+        ])
+    );
+
+    // Filter the result to only the values from the first file
+    let result = df
+        .filter(col("a").is_not_null())
+        .unwrap()
+        .collect()
+        .await
+        .unwrap();
```
a10y, line 124: test fails here with

```
thread 'test_filter_with_schema_evolution' panicked at vortex-datafusion/tests/schema_evolution.rs:124:10:
called `Result::unwrap()` on an `Err` value: ArrowError(SchemaError("Unable to get field named \"a\". Valid fields: [\"b\"]"), Some(""))
```

PR #5551, opened 2025-11-26
fix[display]: escape strings

joseph-isaacs:


robert3005 (commented):
This seems weird. We should be able to escape string but maybe debug is the simplest option?

0ax1:
@claude is this the most elegant way to do this? what are alternatives

0ax1 (approved):
\\

PR #5555, opened 2025-11-27
fix: correctly handle schema evolution in DF

a10y:
This PR fixes a handful of broken things in the overlapping venn diagrams of 

1. vortex-datafusion
2. pushdown over nested fields
3. schema evolution

See https://github.com/vortex-data/vortex/pull/5406 and https://github.com/vortex-data/vortex/pull/5550 for context on recent issues.

This PR brings in a modified version of the `DefaultPhysicalExprRewriter` from DataFusion main, and makes a fix that I've filed https://github.com/apache/datafusion/issues/18957 to upstream

**Concretely**

* Fixes the bug where we weren't pushing down filters in VortexSource over nested struct fields
* Fixes issue found in #5550 where filtering over a table field that is not in every file would cause runtime error
* Make a couple of breaking changes to the `VortexSource` struct by removing the `pub` fields for overriding the SchemaAdapterFactory and PhysicalExprAdapterFactory, since we need to use our vendored version with the fix to ensure correctness.

I can also confirm that this fixes the gharchive benchmark, bringing it from 3 seconds back down to 30ms

image


vortex-datafusion/src/convert/exprs.rs
```diff
@@ -244,8 +244,8 @@ pub(crate) fn can_be_pushed_down(df_expr: &PhysicalExprRef, schema: &Schema) ->
     } else if let Some(in_list) = expr.downcast_ref::<df_expr::InListExpr>() {
         can_be_pushed_down(in_list.expr(), schema)
             && in_list.list().iter().all(|e| can_be_pushed_down(e, schema))
-    } else if let Some(scalar_fn) = expr.downcast_ref::<ScalarFunctionExpr>() {
-        can_scalar_fn_be_pushed_down(scalar_fn, schema)
```
a10y, line 248: as discussed in the other PR, this was never needed, and in fact is not possible to check in the source because we don't know which fields are present in the file or not.

When we've reached this point, DF planner has already determined that all column references are valid, we just need to ensure that the only scalarfunction that gets called is the GetFieldFunc

vortex-datafusion/src/persistent/adapter.rs
```diff
@@ -0,0 +1,339 @@
...
+//! Shared utilities and custom implementations for schema and expression adapters.
+//!
+//! Most of this code is taken from one of two files:
+//!
+//! `datafusion/physical-expr-adapter/src/schema_rewriter.rs` -> everything related to the
+//! DefaultPhysicalExprAdapter
+//!
+//! `datafusion/datasource/src/schema_adapter.rs` -> for can_cast_field (which is crate-private)
+//!
+//! See https://github.com/apache/datafusion/issues/18957
```
a10y: https://github.com/apache/datafusion/issues/18957

vortex-datafusion/src/persistent/adapter.rs
```diff
@@ -0,0 +1,339 @@
...
+        };
+
+        if logical_field.data_type() == physical_field.data_type() {
+            // If the data types match, we can use the column as is
+            return Ok(Transformed::yes(Arc::new(column)));
+        }
+
+        // This has been changed to replace the DF upstream version which just does can_cast_types,
+        // which ignores struct fields with compatible but differing columns.
+        let is_compatible = can_cast_field(physical_field, logical_field)?;
```
a10y, line 291: this is the only line of this file that I have changed from the original

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -48,11 +49,18 @@ use super::cache::VortexFileCache;
 use crate::convert::exprs::can_be_pushed_down;
 use crate::convert::exprs::make_vortex_predicate;
 
+static DEFAULT_EXPR_ADAPTER: LazyLock<Arc<dyn PhysicalExprAdapterFactory>> =
+    LazyLock::new(|| Arc::new(crate::adapter::DefaultPhysicalExprAdapterFactory));
+
+static DEFAULT_SCHEMA_ADAPTER: LazyLock<Arc<dyn SchemaAdapterFactory>> =
+    LazyLock::new(|| Arc::new(DefaultSchemaAdapterFactory));
```
a10y, line 56: we override to use the vendored adapter here

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -61,130 +69,59 @@ pub(crate) struct VortexOpener {
...
     /// statistics and partition values without opening them.
     pub file_pruning_predicate: Option<PhysicalExprRef>,
-    pub expr_adapter_factory: Option<Arc<dyn PhysicalExprAdapterFactory>>,
-    pub schema_adapter_factory: Arc<dyn SchemaAdapterFactory>,
     /// Hive-style partitioning columns
     pub partition_fields: Vec<Arc<Field>>,
     pub file_cache: VortexFileCache,
-    /// This is the table's schema without partition columns. It might be different than
-    /// the physical schema, and the stream's type will be a projection of it.
-    pub logical_schema: SchemaRef,
```
a10y, line 71: these are not breaking changes b/c VortexOpener is pub(crate)

AdamGS:
I'll try and review everything on Friday when I'll probably have some time to burn at the airport, but from the description, shouldn't we rely on a fix upstream?

vortex-datafusion/src/convert/exprs.rs
```diff
@@ -244,8 +244,8 @@ pub(crate) fn can_be_pushed_down(df_expr: &PhysicalExprRef, schema: &Schema) ->
...
+    } else if ScalarFunctionExpr::try_downcast_func::<GetFieldFunc>(df_expr.as_ref()).is_some() {
+        true
```
asubiotto, line 248: Will returning `true` here avoid the planning of a downstream filter? We've talked about edge cases before, i.e. is not null on a non-existent column in one file that is filled with nulls. I don't have a good answer for this and tbh care more about perf than correctness for these edge cases.

One thing we've mentioned is to pass in the list of file schemas to the source so it can check for these cases so we can get the best of both worlds, not sure how you view that.

I also am not very familiar with the `try_pushdown_filters` interface so this might not be a problem. Maybe the issue is that we can only return `PushedDown::Yes` or `PushedDown::No` and might need to upstream an equivalent of `Inexact` that table providers return for filters.

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -61,130 +69,58 @@ pub(crate) struct VortexOpener {
...
         let mut filter = self.filter.clone();
         let file_pruning_predicate = self.file_pruning_predicate.clone();
-        let expr_adapter_factory = self.expr_adapter_factory.clone();
+
+        // We create our custom expression and schema adapters. The schema adapter is just the
+        // DF default adapter. Our expression adapter is currently built to work around a bug
+        // in the upstream adapter.
+        // See: https://github.com/apache/datafusion/issues/17114
+        let expr_adapter_factory = DEFAULT_EXPR_ADAPTER.clone();
+        let schema_adapter_factory = DEFAULT_SCHEMA_ADAPTER.clone();
```
asubiotto: Should we return an error in https://github.com/vortex-data/vortex/blob/2f80934efa7e66595dafa365d6e32cf4e0b1d395/vortex-datafusion/src/persistent/source.rs#L304-L307

on the source? It might be confusing to use that and then figure out why the provided schema adapter factory is not being used. Or maybe in a more backwards-compatible way use the provided schema adapter factory but log a warning.

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -61,130 +69,58 @@ pub(crate) struct VortexOpener {
...
+
         let partition_fields = self.partition_fields.clone();
         let file_cache = self.file_cache.clone();
-        let logical_schema = self.logical_schema.clone();
+        let table_schema = self.table_schema.clone();
```
asubiotto, line 110: Nice, I like the nomenclature change

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -517,96 +439,8 @@ mod tests {
         }
     }
```
asubiotto, line 441: Could you add the test that was deleted from `source.rs` that checked the getfield pushdown on a valid table schema field doesn't error out if that field is missing from the file?

vortex-datafusion/tests/schema_evolution.rs
```diff
@@ -0,0 +1,173 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#![allow(
+    clippy::unwrap_in_result,
+    clippy::unwrap_used,
+    clippy::tests_outside_test_module
+)]
+
+//! Test that checks we can evolve schemas in a cmpatible way across files.
```
asubiotto:
```suggestion
//! Test that checks we can evolve schemas in a compatible way across files.
```

vortex-datafusion/tests/schema_evolution.rs
```diff
@@ -0,0 +1,173 @@
...
+    .await;
+
+    // file2 only contains field "b"
+    write_file(
+        &store,
+        "files/file2.vortex",
+        &record_batch!(("b", Utf8, vec![Some("four"), Some("five"), Some("six")])).unwrap(),
+    )
+    .await;
+
```
asubiotto, line 108: If you want to enhance this test a little you could also:
- Add file3 with ["b", "a"] and file4 with ["a", "b"] in that order or just one file with a different order of fields than the table schema.
- include a struct column in the table schema and have different orderings/missing fields there too.

brancz (approved):
Looks good! I'm hoping that the unit tests I added adequately tested the behavior we were seeing, if not I'll make sure to open follow-ups.

brancz:
btw this might be relevant in case you haven't seen it already: https://github.com/apache/arrow-rs/pull/8871

(I suppose it's only half of the equation though, iiuc you are looking to "cast" to a struct that has a new field that didn't exist before, so https://github.com/apache/arrow-rs/issues/7176 is more relevant, but there the recommendation was to make it the adapter responsibility haha .. you may also be interested in https://github.com/apache/arrow-rs/issues/8840)

AdamGS:
Seems like this might be potentially resolved by https://github.com/apache/datafusion/pull/18998, I'll focus on getting our DF upgrade through so we can test it, might also take tests from here to test on top of that.

robert3005:
This is done as part of df 52 upgrade

PR #5560, opened 2025-11-27
fix[dict]: min_max must not be null

joseph-isaacs:


0ax1:
@claude review

vortex-array/src/mask_future.rs
```diff
@@ -68,6 +68,27 @@ impl MaskFuture {
         let inner = self.inner.clone();
         Self::new(range.len(), async move { Ok(inner.await?.slice(range)) })
     }
+
+    pub fn inspect(
```
robert3005, line 72: do we want to keep this function?

0ax1, line 72: Why was this introduced as part of this PR? Can we add a test for this that checks for `inspect` firing in case of `mask.len() != len`.

robert3005 (approved):

connortsui20:
Does this have anything to do with this new fuzzer error? https://github.com/vortex-data/vortex/issues/6009

robert3005:
I don't think so, this was case of min-max being null in dictionary array but non null in values array of dictionary

PR #5561, opened 2025-11-27
Scalar functions

gatesn:
Introduces the traits required to define abstract scalar functions and wrap them up as either expressions or arrays.

This is one of the types of functions that we will split the generic ComputeFn trait into and allows us to lazily defer scalar compute in the array tree.

We don't add them in this PR, but there may be some properties that are useful for ScalarFns in some form:
```

    /// The identity element `e` where `f(e, x) = f(x, e) = x`.
    ///
    /// When an argument is the identity element, the function can be
    /// eliminated entirely, returning the other argument unchanged.
    ///
    /// # Examples
    /// - `AND`: `true` (AND(true, x) → x)
    /// - `OR`: `false` (OR(false, x) → x)
    /// - `+`: `0` (0 + x → x)
    /// - `*`: `1` (1 * x → x)
    /// - `COALESCE`: `NULL` (COALESCE(NULL, x) → x)
    fn identity_element(&self, options: &Self::Options) -> Option<Scalar> {
        _ = options;
        None
    }

    /// The absorbing element `a` where `f(a, x) = f(x, a) = a`.
    ///
    /// When any argument is the absorbing element, the function short-circuits
    /// immediately, returning that element without evaluating other arguments.
    /// Also known as the "annihilator" or "zero element".
    ///
    /// # Examples
    /// - `AND`: `false` (AND(false, x) → false)
    /// - `OR`: `true` (OR(true, x) → true)
    /// - `*`: `0` (0 * x → 0)
    fn absorbing_element(&self, options: &Self::Options) -> Option<Scalar> {
        _ = options;
        None
    }

    /// Whether argument order is irrelevant: `f(a, b) = f(b, a)`.
    ///
    /// Enables expression normalization (e.g., sorting arguments by column id)
    /// for better common subexpression elimination and pattern matching.
    ///
    /// # Examples
    /// - Commutative: `+`, `*`, `AND`, `OR`, `=`, `!=`, `MIN`, `MAX`
    /// - Non-commutative: `-`, `/`, `<`, `>`, `CONCAT`
    fn is_commutative(&self, options: &Self::Options) -> bool {
        _ = options;
        false
    }

    /// Whether `f(x, x) = x`.
    ///
    /// Enables simplification when the same expression appears multiple times
    /// as arguments to the function.
    ///
    /// # Examples
    /// - Idempotent: `AND`, `OR`, `MIN`, `MAX`
    /// - Non-idempotent: `+` (x + x = 2x), `*` (x * x = x²)
    fn is_idempotent(&self, options: &Self::Options) -> bool {
        _ = options;
        false
    }

    /// Whether `f(f(x)) = x` for unary functions.
    ///
    /// Enables cancellation of nested self-applications.
    ///
    /// # Examples
    /// - Involutions: `NOT`, `NEG` (for signed types), `REVERSE`
    /// - Non-involutions: `ABS`, `UPPER`, `LOWER`
    fn is_involution(&self, options: &Self::Options) -> bool {
        _ = options;
        false
    }

    /// How the function behaves when one or more arguments are NULL.
    ///
    /// Most functions propagate NULL (any NULL argument produces NULL output).
    /// Some functions have special NULL handling that can short-circuit
    /// evaluation or treat NULL as a meaningful value.
    ///
    /// Required for correct NULL semantics; may also enable optimizations
    /// when argument nullability is known from schema or statistics.
    fn null_handling(&self, options: &Self::Options) -> NullHandling {
        _ = options;
        NullHandling::default()
    }
    
    
/// How a function handles NULL arguments.
#[derive(Clone, Debug, Default, PartialEq, Eq)]
pub enum NullHandling {
    /// NULL in any argument produces NULL output.
    ///
    /// This is standard SQL behavior for most scalar functions.
    /// Enables simplification when any argument is known to be NULL.
    Propagate,

    /// NULL is short-circuited when paired with the absorbing element.
    ///
    /// This is a special case where the absorbing element "wins" over NULL.
    ///
    /// # Examples
    /// - `AND_KLEENE(false, NULL)` → `false` (false absorbs NULL)
    /// - `OR_KLEENE(true, NULL)` → `true` (true absorbs NULL)
    AbsorbsNull,

    /// The function has special NULL semantics that don't follow
    /// simple propagation rules.
    ///
    /// This prevents any simplifications based on NULL arguments.
    ///
    /// # Examples
    /// - `IS NULL`, `IS NOT NULL`: NULL → true/false
    /// - `COALESCE`: returns first non-NULL argument
    /// - `NULLIF`: conditionally produces NULL
    #[default]
    Custom,
}
```


vortex-array/src/arrays/scalar_fn/vtable/visitor.rs
```diff
@@ -0,0 +1,19 @@
...
+use crate::arrays::scalar_fn::vtable::ScalarFnVTable;
+use crate::vtable::VisitorVTable;
+
+impl VisitorVTable<ScalarFnVTable> for ScalarFnVTable {
+    fn visit_buffers(_array: &ScalarFnArray, _visitor: &mut dyn ArrayBufferVisitor) {}
+
+    fn visit_children(array: &ScalarFnArray, visitor: &mut dyn ArrayChildVisitor) {
+        for (idx, child) in array.children().iter().enumerate() {
+            let name = array.scalar_fn.signature().arg_name(idx);
+            visitor.visit_child(name.as_deref().unwrap_or("unnamed"), child.as_ref())
```
joseph-isaacs: do you want unnamed here?

vortex-array/src/expr/functions/scalar.rs
```diff
@@ -0,0 +1,198 @@
...
+    pub fn monotonicity(&self, arg_idx: usize) -> Monotonicity {
+        self.vtable.as_dyn().monotonicity(self.options, arg_idx)
+    }
+
+    pub fn null_handling(&self) -> NullHandling {
+        self.vtable.as_dyn().null_handling(self.options)
+    }
+
+    pub fn arg_name(&self, arg_idx: usize) -> Option<String> {
+        self.vtable.as_dyn().arg_name(self.options, arg_idx)
```
joseph-isaacs, line 139: do we need all of these can we add them once we need them?

vortex-array/src/expr/functions/vtable.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// simple propagation rules.
+    ///
+    /// This prevents any simplifications based on NULL arguments.
+    ///
+    /// # Examples
+    /// - `IS NULL`, `IS NOT NULL`: NULL → true/false
+    /// - `COALESCE`: returns first non-NULL argument
+    /// - `NULLIF`: conditionally produces NULL
+    #[default]
+    Custom,
```
joseph-isaacs, line 139: this is not correct

joseph-isaacs: `AND_KLEENE` might absorb and might not

vortex-array/src/arrays/scalar_fn/vtable/visitor.rs
gatesn: We could force a name from ScalarFn I guess?

vortex-array/src/expr/functions/vtable.rs
gatesn: It absorbs null `when paired with the absorbing element.`

joseph-isaacs:
Can we remove stuff from the PR that is not required to impl ScalarFunc

vortex-array/src/expr/functions/vtable.rs
joseph-isaacs: Can we add this in a different PR if we want this

vortex-array/src/arrays/scalar_fn/vtable/visitor.rs
joseph-isaacs: Exactly

vortex-array/src/arrays/scalar_fn/vtable/validity.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn all_invalid(array: &ScalarFnArray) -> bool {
+        match array.scalar_fn.signature().null_handling() {
+            NullHandling::Propagate => {
+                // All null if any child is all null
+                array.children().iter().any(|child| child.all_invalid())
+            }
+            NullHandling::AbsorbsNull | NullHandling::Custom => {
+                // We cannot guarantee that the array is all valid without evaluating the function
+                false
+            }
```
joseph-isaacs, line 40: This is approximate which I am not sure was the original behaviour

gatesn: Yup, but its only used for approximate short-circuits and we have to move the Array trait towards this definition anyway since we want to defer compute.

vortex-array/src/expr/functions/vtable.rs
joseph-isaacs, line 139: can we call this unknown?

gatesn: Hmmm, I prefer custom? It's not that it's unknown. It just isn't describable in a useful way

vortex-array/src/functions/vtable.rs
```diff
@@ -0,0 +1,397 @@
...
+        None
+    }
+
+    /// Whether argument order is irrelevant: `f(a, b) = f(b, a)`.
+    ///
+    /// Enables expression normalization (e.g., sorting arguments by column id)
+    /// for better common subexpression elimination and pattern matching.
+    ///
+    /// # Examples
+    /// - Commutative: `+`, `*`, `AND`, `OR`, `=`, `!=`, `MIN`, `MAX`
```
joseph-isaacs: This is likely a typed property?

joseph-isaacs (approved):

PR #5562, opened 2025-11-27
chore: Remove codspeed warning about deprecated mode

robert3005:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -586,7 +586,7 @@ jobs:
         with:
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
-          mode: "instrumentation"
+          mode: "simulation"
 
   license-check-and-audit-check:
     name: License Check and Audit Check


0ax1 (approved):
Curious though whether this changes the perf results: https://codspeed.io/docs/instruments/cpu/overview#what-is-cpu-simulation


> CodSpeed instruments your benchmarks to measure the performance of your code, simulating the CPU behavior. A benchmark will be run only once and the CPU behavior will be simulated. This ensures that the measurement is as accurate as possible, taking into account not only the instructions executed but also the cache and memory access patterns. The simulation gives us an equivalent of the CPU cycles that includes cache and memory access.

robert3005:
I assumed this was purely a config option rename

0ax1:
> I assumed this was purely a config option rename

I guess
> "Previously, this instrument was referred to as "instrumentation" or "instrumentation mode". This terminology is being phased out in favor of "CPU simulation" to better reflect what the instrument does: simulating CPU behavior. While the old `instrumentation` value is still accepted for backward compatibility..."

PR #5565, opened 2025-11-27
feat[gpu]: use BufferHandle to the array vtable build

joseph-isaacs:
We want to have array which are backed by either CPU or GPU/device buffers. 

The first stage is to create that enum and change the build array method.
Then add this to the segment source and array construction.
Then we can go through each array and support non-cpu buffers.

There is an open question if we want to have this logical-scan-plann have buffers or BufferIDs


robert3005 (approved):

gatesn (commented):
Nice. My only poke is whether the BufferHandle should live in vortex-buffer, if if that crate should remain CPU-only. But we can figure that out later

PR #5581, opened 2025-11-28
chore[ci]: bump query bench runner

joseph-isaacs:


AdamGS:
The problem with i7 instances is that they use AWS' Nitro SSDs which are crazy fast and good, and aren't really reproducible.

PR #5600, opened 2025-12-01
Feature: add total read size counter

harryscholes:
I found that I want to know the total number of bytes read when querying a file, not just the distribution of individual IOs. Implemented as a counter metric named `vortex.io.read.total_size`.


lwwmanning:
thanks for the contribution @harryscholes !

to fix the DCO check, I recommend running something like `git rebase -i origin/develop --signoff` (need to signoff every git commit in the PR, see CONTRIBUTING.md for details)

onursatici (approved):
Thanks for the contribution, yea `sizes` would have the reads histogram, all histograms have a count but not a sum.

PR #5603, opened 2025-12-01
Allow nullable children of MaskedArray

gatesn:


vortex-array/src/arrays/masked/array.rs
joseph-isaacs, line 45: should we rename this to way it doesn't include the mask?

vortex-array/src/arrays/masked/vtable/mod.rs
```diff
@@ -43,15 +44,24 @@ impl VisitorVTable<MaskedVTable> for MaskedVTable {
     }
 }
 
+#[derive(Clone, prost::Message)]
+pub struct MaskedMetadata {
+    /// If true, then the child array's DType is the same as the masked array.
+    /// If false, then the child array's DType is the masked array's DType.as_nonnullable() as was
+    /// the legacy behavior of this array.
+    #[prost(bool, tag = "1")]
+    pub(super) passthrough_child_dtype: bool,
```
joseph-isaacs, line 53: I see why this makes more sense for the build, but I would think that you want child  nullability?

PR #5607, opened 2025-12-02
feat[scalar_fn]: binary ops

joseph-isaacs:


vortex-compute/src/comparison/mod.rs
```diff
@@ -128,3 +130,56 @@ macro_rules! impl_float {
...
+///     Operator::Eq,
+///     Operator::NotEq,
+///     Operator::Lt,
+///     Operator::Lte,
+///     Operator::Gt,
+///     Operator::Gte
+/// );
+/// ```
+#[macro_export]
+macro_rules! compare_op {
```
gatesn, line 165: I'm not sure I understand why this is a macro?

PR #5609, opened 2025-12-02
Add GetItemFn

gatesn:


vortex-array/src/scalar_fns/get_item/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+            .as_struct_fields_opt()
+            .and_then(|st| st.field(field_name))
+            .ok_or_else(|| {
+                vortex_err!("Couldn't find the {} field in the input scope", field_name)
+            })?;
+
+        // Match here to avoid cloning the dtype if nullability doesn't need to change
+        if matches!(
+            (struct_dtype.nullability(), field_dtype.nullability()),
+            (Nullability::Nullable, Nullability::NonNullable)
```
connortsui20, line 85: I think we have `is_null`?

connortsui20 (approved):

vortex-array/src/scalar_fns/get_item/mod.rs
gatesn: Was just porting from existing expr

PR #5610, opened 2025-12-02
FilterArray

gatesn:


vortex-array/src/arrays/filter/vtable.rs
connortsui20, line 1: would prefer the vtables in different files but if were collapsing them all soon then its not that important

connortsui20 (approved):
lets go

PR #5612, opened 2025-12-02
Chore: allow searching subcrates in vortex rustdoc

connortsui20:


vortex/src/lib.rs
```diff
@@ -4,34 +4,83 @@
...
-pub use vortex_scalar as scalar;
-pub use vortex_scan as scan;
-pub use vortex_session as session;
 use vortex_session::VortexSession;
-pub use vortex_utils as utils;
+
+// We re-export like so in order to allow users to search inside subcrates when using the Rust docs.
+
+pub mod array {
+    pub use vortex_array::*;
```
danking, line 22: Does this only allow searching the top-level exports of vortex_array?

danking (approved):

vortex/src/lib.rs
connortsui20: it was more that you couldnt at all before

PR #5616, opened 2025-12-03
Combine ScalarFnVTable and ExprVTable

gatesn:
We proved the idea out with ScalarFns, but realised we can minimize the diff for end-users if we actually just re-purpose an expression to == a scalar function. It pretty much was already anyway.

This PR is a step on the way to having Expression -> Array be a constant-time operation, with actual compute deferred until we execute the array tree.

# Breaking Changes

* Expression VTable changes
  * `Instance` renamed to `Options` to make the purpose more obvious 
  *  Added an `Arity` function to replace `validate`, since the only possible validation in the absence of types is arity based anyway.
  * Added an optional `execute` function that will replace `evaluate` in the future.
  * Removed `fmt_data` in favor of requiring a Display bound on the associated Options type.
  * `return_dtype` takes a `ReturnDTypeCtx` to avoid eagerly computing child dtypes.
* Removed ExpressionView in favor of returning the options from `Expression::as_` and `Expression::as_opt`.

Other changes that I don't expect will impact users:
* Moved `is_null_sensitive` and `child_name` from `Expression`, to `ExpressionSignature`
* Moved `Expression::serialize_metadata()` to `Expression::options().serialize()`
* Removed ExprOptimizer, in favor of `VTable::simplify`. Any more complex optimizations can be implemented in future over the Array tree prior to execution.


joseph-isaacs:
Where do the reduce_parent rules go?

joseph-isaacs:
we are more than half way thought the migration to what I consider a better end state why stop?

PR #5617, opened 2025-12-03
feat[scalar_fn]: binary ops

joseph-isaacs:


vortex-vector/src/lib.rs
```diff
@@ -5,7 +5,6 @@
 // differences from Arrow (builders and arrays and scalars), etc.
 //! Immutable and mutable decompressed (canonical) vectors for Vortex.
 
-#![deny(missing_docs)]
```
gatesn, line 8: Only thing! Otherwise looks great

vortex-compute/src/comparison/datum.rs
```diff
@@ -0,0 +1,65 @@
...
+    BoolDatum: Compare<Op, Output = BoolDatum>,
+    PrimitiveDatum: Compare<Op, Output = BoolDatum>,
+{
+    type Output = BoolDatum;
+
+    fn compare(self, rhs: Self) -> Self::Output {
+        match (self.into_typed(), rhs.into_typed()) {
+            (TypedDatum::Bool(d1), TypedDatum::Bool(d2)) => d1.compare(d2),
+            (TypedDatum::Primitive(d1), TypedDatum::Primitive(d2)) => d1.compare(d2),
+            _ => unreachable!(""),
```
gatesn: This isn't unreachable

joseph-isaacs: replaced with a todo

gatesn (approved):

PR #5625, opened 2025-12-04
Lazy array expression evaluation

gatesn:
This long-lived feature branch contain changes required for the "operators RFC", "Lazy Arrays", whatever we're calling it.

The change is hidden behind the `VORTEX_OPERATORS=true` environment variable allowing us to run parallel benchmarks until we reach parity between legacy compute function dispatch and the new optimizer rules framework.

Summary of changes:
* Rename `execute` -> `execute_vector`, add similar `execute_mask` function. This pattern will be used when turning a Vortex Array into concrete physical data via whatever means, e.g. `execute_arrow`, `execute_duckdb`, and so on.
* Introduces `trait Kernel` to represent a physical tree of CPU-based vector execution
* `Array::batch_execute -> Array::bind_kernel`
* Implements ScalarFnArray execution using a ScalarFnKernel, binding constant children to scalar datum inputs for the ScalarFn.
* Reverts the change to pull out a `ScalarFnVTable`, in favor of allowing all `expr::VTable` to be used as scalar functions. This provides a smoother migration for end-users, at the expense of having to special-case the `root` expression during execution.
* Expression now holds ScalarFn instance (vtable + options), as well as children.
* Removes `ExpressionView`, easier to pass separate Expression + options into vtable.
* `expr::VTable::execute` takes arguments by ownership.
* Implements all `expr::VTable::execute` functions.
* Adds `Array::filter` to vtable, providing an O(1) function for masking arrays. Fallback is to wrap up the child in a FilterArray.
* Adds `Datum::ensure_vector` and fleshes out `Scalar::repeat` implementations.
* Fixes expression display to not use debug formatting of options.
* Removes untyped simplification, it was never called.
* Removes ExprOptimizer in favor of a simple reduce_child `Expr::simplify` function. Complex reduction rules can be implemented over the array tree only to avoid duplication.
* Introduces `Array::apply(&Expression)` to turn expressions into arrays. This function is the new API that will live in vortex-array and allow us to invert the expr -> array dependency to become array -> expr.


vortex-array/src/expr/transform/simplify.rs
joseph-isaacs, line 1: do we have not simplify untyped?

joseph-isaacs:
Can you explain the changes you have make in this branch

joseph-isaacs:
- Can we keep untyped optimise. we decoded we still want this.  
 - why we we call this simplify not reduce (no match array)? 
 - I am think we don't want reduce parent for exprs, I guess we can add later, but why remove?


Can we also explain how to migrate past the break

vortex-array/src/kernel/ready.rs
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_error::VortexResult;
+use vortex_vector::Vector;
+
+use crate::kernel::Kernel;
+use crate::kernel::KernelRef;
+
+/// Create a kernel that is already computed and ready.
+pub fn ready(vector: Vector) -> KernelRef {
```
joseph-isaacs, line 13: can we call this complete?

vortex-array/src/kernel/validate.rs
```diff
@@ -0,0 +1,57 @@
...
+use vortex_error::vortex_ensure;
+use vortex_mask::Mask;
+use vortex_vector::Vector;
+use vortex_vector::VectorOps;
+
+use crate::kernel::Kernel;
+use crate::kernel::KernelRef;
+
+#[derive(Debug)]
+pub struct ValidateKernel {
```
joseph-isaacs, line 16: when is this used?

vortex-array/src/kernel/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+use vortex_error::VortexResult;
+use vortex_mask::Mask;
+use vortex_session::VortexSession;
+use vortex_vector::Vector;
+
+/// A boxed reference to a kernel.
+pub type KernelRef = Box<dyn Kernel>;
+
+/// A trait representing the physical CPU execution of an array tree.
+pub trait Kernel: 'static + Send + Debug {
```
joseph-isaacs, line 26: Shall we call this CPUKernel?

joseph-isaacs:
@claude this is a HUGE PR ultrathink and review this using subagents to cover most of it

vortex-vector/src/datum.rs
```diff
@@ -53,6 +56,23 @@ impl From<Vector> for Datum {
 }
 
 impl Datum {
+    /// Ensure the datum is a vector by repeating the scalar value if necessary.
+    pub fn ensure_vector(self, len: usize) -> Vector {
```
joseph-isaacs, line 60: {convert}/to_vector?

joseph-isaacs: ensure implies validation not conversion

encodings/runend/src/rules.rs
```diff
@@ -0,0 +1,85 @@
...
+use vortex_error::VortexResult;
+
+use crate::RunEndArray;
+use crate::RunEndVTable;
+
+/// A rule to push down scalar functions through run-end encoding into the values array.
+///
+/// This only works if all other children of the scalar function array are constants.
+#[derive(Debug)]
+pub(crate) struct RunEndScalarFnRule;
```
joseph-isaacs, line 22: seems like something to test?

vortex-array/src/arrays/chunked/vtable/mod.rs
```diff
@@ -125,15 +132,35 @@ impl VTable for ChunkedVTable {
...
         Ok(vector.freeze())
     }
-}
 
-#[derive(Debug)]
-pub struct ChunkedVTable;
+    fn push_down_filter(self: Box<Self>, _selection: &Mask) -> VortexResult<PushDownResult> {
+        Ok(PushDownResult::NotPushed(self))
+    }
+}
```
joseph-isaacs, line 166: why do I do this vs a function?:

joseph-isaacs (approved):
Approving with the idea, that we will merge more breaks shortly.


The return_dtype input

PR #5626, opened 2025-12-04
fix[vortex-array]: fix take on varbinviews with NULL indices

asubiotto:
Index values at NULL positions were used. In the case that these were garbage, an OOB panic would occur.

### DIFF
--- a/vortex-array/src/arrays/varbinview/compute/take.rs
+++ b/vortex-array/src/arrays/varbinview/compute/take.rs
@@ -1,12 +1,15 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::iter;
 use std::ops::Deref;
 
 use num_traits::AsPrimitive;
 use vortex_buffer::Buffer;
 use vortex_dtype::match_each_integer_ptype;
 use vortex_error::VortexResult;
+use vortex_mask::AllOr;
+use vortex_mask::Mask;
 use vortex_vector::binaryview::BinaryView;
 
 use crate::Array;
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/varbinview/compute/take.rs
```diff
@@ -55,15 +58,36 @@ register_kernel!(TakeKernelAdapter(VarBinViewVTable).lift());
...
+        )),
+        AllOr::Some(buffer) => Buffer::<BinaryView>::from_trusted_len_iter(
+            buffer.iter().zip(indices.iter()).map(|(valid, idx)| {
+                if valid {
+                    views_ref[idx.as_()]
+                } else {
+                    BinaryView::default()
+                }
+            }),
+        ),
```
a10y, line 83: since taking views is a common operation, we might want to optimize this a bit further, e.g. check for nulls 8 at a time instead of forcing the branch on every iteration

joseph-isaacs: I am not sure that is always an optimisation. Likely around 1/8 > selectivity it is. 

Shall we merge this and create a follow up to benchmark?

vortex-array/src/arrays/varbinview/compute/take.rs
```diff
@@ -23,16 +26,16 @@ use crate::vtable::ValidityHelper;
...
-        // Compute the new validity
-
-        // This is valid since all elements (of all arrays) even null values must be inside
-        // min-max valid range.
+        // Compute the new validity.
         let validity = array.validity().take(indices)?;
         let indices = indices.to_primitive();
 
         let views_buffer = match_each_integer_ptype!(indices.ptype(), |I| {
-            // This is valid since all elements even null values are inside the min-max valid range.
```
joseph-isaacs, line 34: OOI do you know what violated this? We tried to ensure that all codes even null ones have the value point to a real value.

vortex-array/src/arrays/varbinview/compute/take.rs
a10y: idk when it would ever be slower than iterating each bit one by one but yea, we can flup

a10y (approved):

vortex-array/src/arrays/varbinview/compute/take.rs
asubiotto: It was the output of a datafusion sort written to vortex. I didn't really dig into the "why" of the OOB values in null codes, but there was a pattern where the the value was always == length.

PR #5636, opened 2025-12-05
fix: avoid ListBuilder

a10y:
Avoid reimplementing the slow ListBuilder pathway, use `list_from_list_view` to also get the ZCTL fast path when available

### DIFF
--- a/vortex-array/src/arrow/compute/to_arrow/canonical.rs
+++ b/vortex-array/src/arrow/compute/to_arrow/canonical.rs
@@ -71,12 +71,11 @@ use crate::arrays::NullArray;
 use crate::arrays::PrimitiveArray;
 use crate::arrays::StructArray;
 use crate::arrays::VarBinViewArray;
+use crate::arrays::list_from_list_view;
 use crate::arrow::IntoArrowArray;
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrow/compute/to_arrow/canonical.rs
```diff
@@ -530,28 +529,9 @@ fn to_arrow_list<O: IntegerPType + OffsetSizeTrait>(
...
-        array.len(),
-    );
-
-    // TODO(connor)[ListView]: We can potentially make a generic version of `list_from_list_view`
-    // over the offsets so we don't have to rewrite this.
-
-    list_builder.extend_from_array(&array.to_array());
-    let list_array = list_builder.finish_into_list();
+    // Convert listview -> list, via the fast path when possible.
+    let list_array = list_from_list_view(array);
```
joseph-isaacs, line 535: Want to add a TODO to extend rebuild to support a target offset + size dtype?

connortsui20 (approved):
Yep this looks good, I think I just forgot to update this when we did the zctl optimizations

PR #5637, opened 2025-12-06
fix: cargo-deny for rustls-pemfile

a10y:
https://rustsec.org/advisories/RUSTSEC-2025-0134.html

Removing the GCP feature flag prevents us bringing in the transitive dep, and I believe all of our benchmarking infra is off GCP now anyway.


a10y:
ok nvm we can't really eliminate it unless we drop it from the FFI/Java/Python bindings too, which is probably not ideal.

Guess we just wait until the object_store release is cut.

PR #5640, opened 2025-12-06
feature: TableStrategy

a10y:
Extracting this out of my hack week project. This adds a new `TableStrategy` that is like the StructStrategy, except that it allows users to override the write strategy at particular field paths in the schema.

For example, if you have a schema like this:

```text
request: struct?
|__ id: utf8
|__ duration_ms: i64
|__ body: struct?
	|__ bytes: binary?
    |__ json: utf8?
```

Before you'd be stuck with choosing a single strategy for all fields of the Struct.

TableStrategy allows you to override particular field paths anywhere in the field tree. For example, if you wanted to allow uncompressed struct validity, default BtrBlocks compression but ZSTD compression just for the `request.body.bytes` field, you'd do:

```rust
let btrblocks = Arc::new(CompressingStrategy::new_btrblocks(FlatLayoutStrategy::default()));
let zstd = Arc::new(CompressingStrategy::new_compact(FlatLayoutStrategy::default()));

let table_strategy = TableStrategy::new(
            FlatLayoutStrategy::default(), 
            btrblocks,
        )
        .with_leaf_strategy(field_path!(request.body.bytes), zstd);
```

I've replaced StructStrategy with this in the default WriteBuilder.

If we like this, we might want to consider marking `StructStrategy` as deprecated.

### DIFF
--- a/vortex-file/src/strategy.rs
+++ b/vortex-file/src/strategy.rs
@@ -31,7 +31,7 @@ const ONE_MEG: u64 = 1 << 20;
 pub struct WriteStrategyBuilder {
     compressor: Option<Arc<dyn CompressorPlugin>>,
     row_block_size: usize,
-    field_writers: Option<HashMap<FieldPath, Arc<dyn LayoutStrategy>>>,
+    field_writers: HashMap<FieldPath, Arc<dyn LayoutStrategy>>,
 }
 
 impl Default for WriteStrategyBuilder {
@@ -43,11 +43,11 @@ impl Default for WriteStrategyBuilder {
 impl WriteStrategyBuilder {
     /// Create a new empty builder. It can be further configured, and then finally built
     /// yielding the [`LayoutStrategy`].
-    pub const fn new() -> Self {
+    pub fn new() -> Self {
         Self {
             compressor: None,
             row_block_size: 8192,
-            field_writers: None,
+            field_writers: HashMap::new(),
         }
     }
 
@@ -73,9 +73,7 @@ impl WriteStrategyBuilder {
         field: impl Into<FieldPath>,
         writer: Arc<dyn LayoutStrategy>,
     ) -> Self {
-        let mut writers = self.field_writers.unwrap_or_default();
-        writers.insert(field.into(), writer);
-        self.field_writers = Some(writers);
+        self.field_writers.insert(field.into(), writer);
         self
     }
 
@@ -143,14 +141,9 @@ impl WriteStrategyBuilder {
         // 0. start with splitting columns
         let validity_strategy = CollectStrategy::new(compress_then_flat);
 
-        let table_strategy = PathStrategy::new(Arc::new(validity_strategy), Arc::new(repartition));
-
         // Take any field overrides from the builder and apply them to the final strategy.
-        let table_strategy = if let Some(overrides) = self.field_writers {
-            table_strategy.set_field_writers(overrides)
-        } else {
-            table_strategy
-        };
+        let table_strategy = PathStrategy::new(Arc::new(validity_strategy), Arc::new(repartition))
+            .with_field_writers(self.field_writers);
 
         Arc::new(table_strategy)
     }
--- a/vortex-file/src/strategy.rs
+++ b/vortex-file/src/strategy.rs
@@ -14,9 +14,9 @@ use vortex_layout::layouts::compressed::CompressingStrategy;
 use vortex_layout::layouts::compressed::CompressorPlugin;
 use vortex_layout::layouts::dict::writer::DictStrategy;
 use vortex_layout::layouts::flat::writer::FlatLayoutStrategy;
-use vortex_layout::layouts::path::PathStrategy;
 use vortex_layout::layouts::repartition::RepartitionStrategy;
 use vortex_layout::layouts::repartition::RepartitionWriterOptions;
+use vortex_layout::layouts::table::TableStrategy;
 use vortex_layout::layouts::zoned::writer::ZonedLayoutOptions;
 use vortex_layout::layouts::zoned::writer::ZonedStrategy;
 use vortex_utils::aliases::hash_map::HashMap;
@@ -142,7 +142,7 @@ impl WriteStrategyBuilder {
         let validity_strategy = CollectStrategy::new(compress_then_flat);
 
         // Take any field overrides from the builder and apply them to the final strategy.
-        let table_strategy = PathStrategy::new(Arc::new(validity_strategy), Arc::new(repartition))
+        let table_strategy = TableStrategy::new(Arc::new(validity_strategy), Arc::new(repartition))
             .with_field_writers(self.field_writers);
 
         Arc::new(table_strategy)
--- a/vortex-file/src/strategy.rs
+++ b/vortex-file/src/strategy.rs
@@ -5,6 +5,7 @@
 
 use std::sync::Arc;
 
+use vortex_dtype::FieldPath;
 use vortex_layout::LayoutStrategy;
 use vortex_layout::layouts::buffered::BufferedStrategy;
 use vortex_layout::layouts::chunked::writer::ChunkedLayoutStrategy;
@@ -13,11 +14,12 @@ use vortex_layout::layouts::compressed::CompressingStrategy;
 use vortex_layout::layouts::compressed::CompressorPlugin;
 use vortex_layout::layouts::dict::writer::DictStrategy;
 use vortex_layout::layouts::flat::writer::FlatLayoutStrategy;
+use vortex_layout::layouts::path::PathStrategy;
 use vortex_layout::layouts::repartition::RepartitionStrategy;
 use vortex_layout::layouts::repartition::RepartitionWriterOptions;
-use vortex_layout::layouts::struct_::writer::StructStrategy;
 use vortex_layout::layouts::zoned::writer::ZonedLayoutOptions;
 use vortex_layout::layouts::zoned::writer::ZonedStrategy;
+use vortex_utils::aliases::hash_map::HashMap;
 
 const ONE_MEG: u64 = 1 << 20;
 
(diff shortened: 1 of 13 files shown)


vortex-layout/src/layouts/path.rs
```diff
@@ -0,0 +1,324 @@
...
+    ///         field_path!(request.body.bytes),
+    ///         Arc::new(compress_compact),
+    ///     );
+    /// ```
+    pub fn set_field_writer(
+        mut self,
+        field_path: impl Into<FieldPath>,
+        writer: Arc<dyn LayoutStrategy>,
+    ) -> Self {
+        self.leaf_writers.insert(field_path.into(), writer);
```
gatesn: I think we should check that we don't have any overlapping prefixes?

vortex-layout/src/layouts/table.rs
```diff
@@ -0,0 +1,324 @@
...
+use crate::sequence::SequentialStreamAdapter;
+use crate::sequence::SequentialStreamExt;
+
+/// A configurable strategy for writing tables with nested field columns, allowing
+/// overrides for specific leaf columns.
+pub struct PathStrategy {
+    /// A set of leaf field overrides, e.g. to force one column to be compact-compressed.
+    leaf_writers: HashMap<FieldPath, Arc<dyn LayoutStrategy>>,
+    /// The writer for any validity arrays that may be present
+    validity: Arc<dyn LayoutStrategy>,
```
gatesn, line 51: Is this only for non-leaf struct validity? Or all validity? Do we just need a special type of FieldPath that allows us to index into the nulls etc?

e.g. `enum PathComponent { Field(FieldName), ListElements, Validity }` maybe? Not sure... might be gross.

vortex-layout/src/layouts/table.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// # use std::sync::Arc;
+    /// # use vortex_layout::layouts::flat::writer::FlatLayoutStrategy;
+    /// # use vortex_layout::layouts::path::PathStrategy;
+    ///
+    /// // Build a write strategy that does not compress validity or any leaf fields.
+    /// let flat = Arc::new(FlatLayoutStrategy::default());
+    ///
+    /// let strategy = PathStrategy::new(flat.clone(), flat.clone());
+    /// ```
+    pub fn new(validity: Arc<dyn LayoutStrategy>, fallback: Arc<dyn LayoutStrategy>) -> Self {
```
gatesn, line 86: I think validity should just default to flat? Or maybe default everything and use builder pattern to override?

vortex-file/src/strategy.rs
```diff
@@ -29,6 +31,7 @@ const ONE_MEG: u64 = 1 << 20;
 pub struct WriteStrategyBuilder {
     compressor: Option<Arc<dyn CompressorPlugin>>,
     row_block_size: usize,
+    field_writers: Option<HashMap<FieldPath, Arc<dyn LayoutStrategy>>>,
```
AdamGS: This has the size of a empty hashmap (which I think is basically ~0), I think we can just store it directly.

vortex-layout/src/layouts/table.rs
```diff
@@ -0,0 +1,324 @@
...
+        &self,
+        ctx: ArrayContext,
+        segment_sink: SegmentSinkRef,
+        stream: SendableSequentialStream,
+        mut eof: SequencePointer,
+        handle: Handle,
+    ) -> VortexResult<LayoutRef> {
+        let dtype = stream.dtype().clone();
+
+        // Fallback: if the array is not a struct, fallback to writing a single array.
```
gatesn, line 215: Yeah, I think you actually want to check the leaf_writers here for a root FieldPath, that would allow overriding intermediate struct strategies.

gatesn (dismissed):
Yeah really nice idea. Definitely should just immediately deprecate the StructStrategy.

vortex-file/tests/test_write_table.rs
```diff
@@ -0,0 +1,113 @@
...
+    // leave it uncompressed.
+    let default_strategy = Arc::new(CompressingStrategy::new_btrblocks(
+        FlatLayoutStrategy::default(),
+        false,
+    ));
+
+    let writer = Arc::new(
+        PathStrategy::new(Arc::new(FlatLayoutStrategy::default()), default_strategy)
+            .set_field_writer(field_path!(a.raw), Arc::new(FlatLayoutStrategy::default())),
+    );
```
AdamGS, line 79: nit - can the test override two different paths?

vortex-layout/src/layouts/table.rs
```diff
@@ -0,0 +1,324 @@
...
+            return self
+                .fallback
+                .write_stream(ctx, segment_sink, stream, eof, handle)
+                .await;
+        }
+
+        let struct_dtype = dtype.as_struct_fields();
+
+        // Check for unique field names at write time.
+        if HashSet::<_, DefaultHashBuilder>::from_iter(struct_dtype.names().iter()).len()
```
AdamGS, line 226: Don't we directly expose an alias for the better hashset?

vortex-layout/src/layouts/table.rs
```diff
@@ -0,0 +1,324 @@
...
+                .chain(struct_dtype.names().iter().cloned())
+                .collect()
+        } else {
+            struct_dtype.names().iter().cloned().collect()
+        };
+
+        let layout_futures: Vec<_> = column_dtypes
+            .into_iter()
+            .zip_eq(column_streams_rx)
+            .zip_eq(column_names)
```
AdamGS, line 320: ~nit - `itertools::izip`~ after reading some `itertools` issues, I suspect there's a better things we can do here, I'll try it out in a different branch.

vortex-layout/src/layouts/table.rs
```diff
@@ -0,0 +1,324 @@
...
+
+            columns.extend(
+                struct_chunk
+                    .fields()
+                    .iter()
+                    .map(|field| (sequence_pointer.advance(), field.to_array())),
+            );
+
+            Ok(columns)
+        });
```
AdamGS, line 266: ~Is all this code just the same as the current `StructStrategy`? Is there a reason to maintain both?~ Should've read the full PR description

vortex-layout/src/layouts/table.rs
AdamGS: trying something here - https://github.com/vortex-data/vortex/pull/5675

vortex-layout/src/layouts/table.rs
a10y: This is actually to maintain compatibility with behavior in StructStrategy, which lets you serialize non-struct flat arrays. Not sure why we let you do that but it's the current behavior and we have unit tests for it 🤷

vortex-file/src/strategy.rs
a10y: This was just b/c I wanted to keep WriteStrategyBuilder::new() as const

vortex-layout/src/layouts/table.rs
a10y: structs are the only case right now where validity is extracted out and compressed into its own child layout, all of the other strategies just store it alongside the array data

vortex-file/src/strategy.rs
a10y: actually we don't really gain much by only having the constructor be const so i just removed the constness and inlined the hashmap

vortex-layout/src/layouts/table.rs
a10y: I just copied this over from StructStrategy

https://github.com/vortex-data/vortex/blob/develop/vortex-layout/src/layouts/struct_/writer.rs#L72-L78

a10y: for w/e reason, doesn't seem like it can infer the hasher here 🤷 
image

vortex-layout/src/layouts/path.rs
a10y: should we panic on this or just ignore/maybe log warning?

a10y: my vote would be panic

a10y: i went with panic

vortex-layout/src/layouts/table.rs
AdamGS: 😢

vortex-layout/src/layouts/struct_/writer.rs
```diff
@@ -35,7 +37,13 @@ use crate::sequence::SequencePointer;
 use crate::sequence::SequentialStreamAdapter;
 use crate::sequence::SequentialStreamExt;
 
+/// A write strategy that shreds tabular data into columns and writes each column
+/// as its own distinct stream.
+///
+/// This is now deprecated, users are encouraged to instead use the
+/// [`TableStrategy`][crate::layouts::table::TableStrategy].
 #[derive(Clone)]
+#[deprecated(since = "0.57.0", note = "Use the `TableStrategy` instead.")]
```
AdamGS: nit - 58 is the latest release

vortex-layout/src/layouts/table.rs
AdamGS: @a10y do you want to do something here?

vortex-layout/src/layouts/table.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    /// Override the strategy for compressing struct validity at all levels of the schema tree.
+    pub fn with_validity_strategy(mut self, validity: Arc<dyn LayoutStrategy>) -> Self {
+        self.validity = validity;
+        self
+    }
+}
+
+impl TableStrategy {
+    // Descend into a subfield for the writer.
```
AdamGS: nit - rustdoc

AdamGS (approved):
LGTM, a big part of this is pulled from `StructLayout`

vortex-layout/src/layouts/table.rs
a10y: i should've closed this before, but yea i went with making Default set everything to Flat and added `with`ers

PR #5647, opened 2025-12-08
Arrow Executor

gatesn:
We eventually need to port all of our execution logic over to executing using Vectors + VortexSession. 
We can release this new API prior to deprecating the existing one.
We should figure out what this API should be.
Once the full operators change has finished, we should be able to pull this module out into vortex-arrow.


vortex-array/src/arrow/executor.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// If a [`DataType`] is given, the array will be converted to the desired Arrow type.
+    fn execute_arrow(
+        &self,
+        // TODO(ngates): do we even want optional data type? Or do we make it required and tell
+        //  users to call `DType::into_arrow` to get a default logical type? I'm inclined to think
+        //  the latter is preferable. Although there's a world where the user may want the minimal
+        //  conversion to Arrow without knowing what that conversion is. In which case, should
+        //  DictionaryArray and non-logical types be supported? Should the user provide a list of
+        //  supported Arrow arrays? I dunno...
+        data_type: Option<&DataType>,
```
joseph-isaacs: I would vote for non-optional. Then have an Ext trait impl?

vortex-array/src/arrow/executor.rs
```diff
@@ -0,0 +1,100 @@
...
+
+impl ArrowArrayExecutor for crate::ArrayRef {
+    fn execute_arrow(
+        &self,
+        data_type: &Option<DataType>,
+        session: &VortexSession,
+    ) -> VortexResult<ArrowArrayRef> {
+        match data_type {
+            None => {
+                // Special-case the Arrow-shaped encodings that are not part of our vector API
```
joseph-isaacs: this needs to extensible?

gatesn: I'm not sure it does? After the operators change, we pull out a vortex-arrow crate that depends on the main `vortex` crate. Arrow becomes a regular "integration", and it can depend on any of the encodings crates. At least that's one option. Same way we do DuckDB where we know about all the destination types, so we can just hard-code.

vortex-array/src/arrow/executor.rs
```diff
@@ -0,0 +1,100 @@
...
+                }
+                if let Some(_list) = self.as_opt::<ListVTable>() {
+                    // Convert directly to preferred Arrow List array.
+                }
+
+                let vector = self.execute(session)?;
+                vector.into_arrow()
+            }
+            // Once we know the target Arrow DataType, how do we get there? Should we allow crates
+            // to register Arrow conversion kernels? Should we wrap up the Vortex array in a
```
asubiotto: I think I would prefer to wrap the vortex array in a cast, since it would generally be cheaper to push this cast down to the encoding layer.

vortex-array/src/arrow/executor.rs
```diff
@@ -0,0 +1,100 @@
...
+use arrow_schema::DataType;
+use vortex_compute::arrow::IntoArrow;
+use vortex_error::VortexResult;
+use vortex_session::VortexSession;
+
+use crate::arrays::ListVTable;
+use crate::arrays::VarBinVTable;
+
+/// Trait for executing a Vortex array to produce an Arrow array.
+pub trait ArrowArrayExecutor {
```
asubiotto: Is the use of "executor" here 100% necessary? I think `ArrowConverter` or `ArrowConvertible` makes more sense to me and then the trait function would just be `convert`. More of a nit, so feel free to ignore.

gatesn: No not necessarily. I agree the name isn't great, but we want to move towards making it such that the only way to get data out of a Vortex array is via some form of "execution".

By default this turns an array into a vector, but we have MaskExecutor that checks for constant true/false (and should also check for run-end in order to construct slices), and obviously ArrowExecutor checks for lots of things.

Maybe exporter is a better name? Or maybe just into_arrow...

vortex-compute/src/cast/dvector.rs
```diff
@@ -19,26 +24,51 @@ use crate::cast::try_cast_vector_common;
...
-            // TODO(connor): cast to different precision/scale
-            DType::Decimal(..) => {
+            DType::Decimal(ddt, n) => {
+                // Check nullability compatibility
+                if !n.is_nullable() && !self.validity().all_true() {
+                    vortex_bail!(
+                        "Cannot cast nullable DVector to non-nullable {}",
+                        target_dtype
+                    );
+                }
```
connortsui20: Nit: Might as well do `let else` clause here. Also periods in comments

vortex-array/src/lib.rs
```diff
@@ -60,3 +65,15 @@ pub mod flatbuffers {
     //! Re-exported autogenerated code from the core Vortex flatbuffer definitions.
     pub use vortex_flatbuffers::array::*;
 }
+
+static USE_VORTEX_OPERATORS: LazyLock<bool> = LazyLock::new(|| {
+    std::env::var("VORTEX_OPERATORS")
+        .map(|v| v == "1" || v.to_lowercase() == "true")
+        .unwrap_or(false)
+});
```
connortsui20, line 73: I think you can just do this, so it gets turned on [as long as it is set](https://doc.rust-lang.org/std/env/fn.var.html):

```suggestion
static USE_VORTEX_OPERATORS: LazyLock<bool> = LazyLock::new(|| {
    std::env::var("VORTEX_OPERATORS").is_ok()
});
```

vortex-array/src/optimizer/mod.rs
connortsui20, line 18: Personally I don't think it makes sense to expose `try_optimize` methods as I don't think the caller should know if a transform rule was applied, but I don't feel super strongly about it.

vortex-vector/src/primitive/generic.rs
```diff
@@ -93,6 +93,19 @@ impl<T> PVector<T> {
         (self.elements, self.validity)
     }
 
+    /// Decomposes the primitive vector into its internal buffer, consuming the vector.
+    ///
+    /// # Panics
+    ///
+    /// Panics if there are any null values in the vector.
+    pub fn into_buffer(self) -> Buffer<T> {
```
connortsui20: Can we rename this to `into_nonnull_buffer`?

vortex-array/src/arrow/executor/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+            DataType::Null => to_arrow_null(self, session),
+            DataType::Boolean => to_arrow_bool(self, session),
+            DataType::Int8 => to_arrow_primitive::<UInt8Type>(self, session),
+            DataType::Int16 => to_arrow_primitive::<UInt16Type>(self, session),
+            DataType::Int32 => to_arrow_primitive::<UInt32Type>(self, session),
+            DataType::Int64 => to_arrow_primitive::<UInt64Type>(self, session),
+            DataType::UInt8 => to_arrow_primitive::<Int8Type>(self, session),
+            DataType::UInt16 => to_arrow_primitive::<Int16Type>(self, session),
+            DataType::UInt32 => to_arrow_primitive::<Int32Type>(self, session),
+            DataType::UInt64 => to_arrow_primitive::<Int64Type>(self, session),
```
connortsui20: Types seem to be wrong?

vortex-array/src/array/mod.rs
```diff
@@ -691,13 +722,6 @@ impl<V: VTable> Array for ArrayAdapter<V> {
     }
 
     fn reduce_parent(&self, parent: &ArrayRef, child_idx: usize) -> VortexResult<Option<ArrayRef>> {
-        #[cfg(debug_assertions)]
-        vortex_ensure!(
-            Arc::as_ptr(&parent.children()[child_idx]) == self,
-            "Parent array's child at index {} does not match self",
-            child_idx
-        );
```
connortsui20, line 699: This feels like it would still be good to have as a `debug_assert_eq`?

vortex-array/src/arrow/executor/fixed_size_list.rs
connortsui20, line 1: This should be called `fixed_size_list.rs` for consistency with the rest of the project

vortex-array/src/arrow/executor/decimal.rs
```diff
@@ -0,0 +1,61 @@
...
+pub(super) fn to_arrow_decimal<D: DecimalType, N: NativeDecimalType>(
+    array: ArrayRef,
+    precision: u8,
+    scale: i8,
+    session: &VortexSession,
+) -> VortexResult<ArrowArrayRef> {
+    // Since Vortex doesn't have physical types, the vector execution will try to use the
+    // narrowest type that fits the precision + scale.
+    //
+    // We therefore fake the required precision such that the correct native type is used.
```
connortsui20: I think this needs more explanation... I had no idea that this is true? Shouldn't vector execution should be decompressing into whatever `DecimalDType` is specified?

vortex-array/src/arrow/executor/decimal.rs
```diff
@@ -0,0 +1,61 @@
...
+        "Decimal array conversion produced unexpected decimal type: expected {:?}, got {:?}",
+        N::DECIMAL_TYPE,
+        vector.decimal_type()
+    );
+
+    let (_ps, buffer, validity) = N::downcast(vector).into_parts();
+    let nulls = to_null_buffer(validity);
+
+    // Again, because our i256 type is different from Arrow's, we need to special-case it.
+    let buffer = unsafe { std::mem::transmute::<Buffer<N>, Buffer<D::Native>>(buffer) };
```
connortsui20: This might be related to above but this is crazy to me without a safety comment explaining why this is fine since the function above (in isolation) never even checks that `N` is related to `D::Native` in any way.

vortex-array/src/arrow/executor/fixed_size_list.rs
```diff
@@ -0,0 +1,73 @@
...
+        return list_to_list(array, elements_field, list_size, session);
+    }
+
+    // Otherwise, we execute the array to become a FixedSizeListArray.
+    let vector = array
+        .execute_vector(session)?
+        .into_fixed_size_list_opt()
+        .ok_or_else(|| vortex_err!("Failed to convert array to FixedSizeListArray"))?;
+    vortex_ensure!(
+        Ok(list_size) == i32::try_from(vector.list_size()),
```
connortsui20, line 38:
```suggestion
        list_size == i32::try_from(vector.list_size())?,
```

vortex-array/src/arrow/executor/fixed_size_list.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+fn list_to_list(
+    array: &FixedSizeListArray,
+    elements_field: &FieldRef,
+    list_size: i32,
+    session: &VortexSession,
+) -> VortexResult<arrow_array::ArrayRef> {
+    vortex_ensure!(
+        Ok(list_size) == i32::try_from(array.list_size()),
```
connortsui20, line 54:
```suggestion
        list_size == i32::try_from(vector.list_size())?,
```

connortsui20 (commented):
Just a few things.

I think there's a lot of other stuff that can be cleaned up but since we're on time pressure let's get this over the line.

vortex-array/src/lib.rs
gatesn: I always find it confusing if MY_ENV=false turns the thing on 🤷‍♂️

vortex-array/src/array/mod.rs
gatesn: It wasn't actually correct. Trying to figure out if I can get the Arc::ptr_eq to work

vortex-array/src/arrow/executor/decimal.rs
gatesn: DecimalDType doesn't specify a binary width, only a precision and scale

vortex-array/src/arrow/executor/fixed_size_list.rs
gatesn: This will produce an opaque "cannot cast error" instead of the more obvious error message that I wrote

vortex-array/src/arrow/executor/decimal.rs
```diff
@@ -0,0 +1,75 @@
...
+    // We therefore create a fake precision that forces Vortex to use a native type that is at
+    // least as wide as the requested Arrow type. We cast the array into this type.
+    //
+    // NOTE(ngates): we assume that a cast operation will produce the narrowest possible type that
+    //  fits the requested precision.
+    let fake_precision = precision.max(N::MAX_PRECISION);
+    let array = array.cast(DType::Decimal(
+        DecimalDType::new(fake_precision, scale),
+        Nullability::Nullable,
+    ))?;
```
joseph-isaacs, line 45: We document the cast to do this

joseph-isaacs (approved):

PR #5650, opened 2025-12-08
vx layout display

joseph-isaacs:
```
vortex.chunked, dtype: i32, children: 2, rows: 10
├── [0]: vortex.flat, dtype: i32, metadata: 98 bytes, rows: 5, segment 0, buffers=[20B], total=20B
└── [1]: vortex.flat, dtype: i32, metadata: 98 bytes, rows: 5, segment 1, buffers=[20B], total=20B
```

### DIFF
--- a/vortex-array/src/serde.rs
+++ b/vortex-array/src/serde.rs
@@ -403,6 +403,43 @@ impl ArrayParts {
             })
     }
 
+    /// Returns the buffer lengths as stored in the flatbuffer metadata.
+    ///
+    /// This reads the buffer descriptors from the flatbuffer, which contain the
+    /// serialized length of each buffer. This is useful for displaying buffer sizes
+    /// without needing to access the actual buffer data.
+    pub fn buffer_lengths(&self) -> Vec<usize> {
+        let fb_array = root::<fba::Array>(self.flatbuffer.as_ref())
+            .vortex_expect("ArrayParts flatbuffer must be a valid Array");
+        fb_array
+            .buffers()
+            .map(|buffers| buffers.iter().map(|b| b.length() as usize).collect())
+            .unwrap_or_default()
+    }
+
+    /// Create an [`ArrayParts`] from a raw array tree flatbuffer (metadata only).
+    ///
+    /// This constructor creates an `ArrayParts` with no buffer data, useful for
+    /// inspecting the metadata when the actual buffer data is not needed
+    /// (e.g., displaying buffer sizes from inlined array tree metadata).
+    ///
+    /// Note: Calling `buffer()` on the returned `ArrayParts` will fail since
+    /// no actual buffer data is available.
+    pub fn from_array_tree(array_tree: impl Into<ByteBuffer>) -> VortexResult<Self> {
+        let fb_buffer = FlatBuffer::align_from(array_tree.into());
+        let fb_array = root::<fba::Array>(fb_buffer.as_ref())?;
+        let fb_root = fb_array
+            .root()
+            .ok_or_else(|| vortex_err!("Array must have a root node"))?;
+        let flatbuffer_loc = fb_root._tab.loc();
+
+        Ok(ArrayParts {
+            flatbuffer: fb_buffer,
+            flatbuffer_loc,
+            buffers: Arc::new([]),
+        })
+    }
+
     /// Returns the root ArrayNode flatbuffer.
     fn flatbuffer(&self) -> fba::ArrayNode<'_> {
         unsafe { fba::ArrayNode::follow(self.flatbuffer.as_ref(), self.flatbuffer_loc) }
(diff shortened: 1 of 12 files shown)


vortex-array/src/serde.rs
```diff
@@ -403,6 +403,43 @@ impl ArrayParts {
...
+    ///
+    /// Note: Calling `buffer()` on the returned `ArrayParts` will fail since
+    /// no actual buffer data is available.
+    pub fn from_array_tree(array_tree: impl Into<ByteBuffer>) -> VortexResult<Self> {
+        let fb_buffer = FlatBuffer::align_from(array_tree.into());
+        let fb_array = root::<fba::Array>(fb_buffer.as_ref())?;
+        let fb_root = fb_array
+            .root()
+            .ok_or_else(|| vortex_err!("Array must have a root node"))?;
+        let flatbuffer_loc = fb_root._tab.loc();
```
0ax1, line 434: `_tab` reads like accessing an internal field? I see though that we're using this pattern in other places too.

vortex-layout/src/layouts/flat/mod.rs
```diff
@@ -186,6 +186,65 @@ impl FlatLayout {
     }
 }
 
+/// Display wrapper for the array tree metadata stored in a FlatLayout.
+pub struct ArrayTreeDisplay {
+    array_tree: Option<ByteBuffer>,
+    segment_id: SegmentId,
```
0ax1: We only ever use the segment id if `array_tree` is not set. Should these be diff cases represented by an enum?

vortex-layout/src/layouts/flat/mod.rs
```diff
@@ -186,6 +186,65 @@ impl FlatLayout {
...
+        write!(f, "ArrayTree({}", encoding_name)?;
+
+        // Show children count
+        let nchildren = fb_root.children().map_or(0, |c| c.len());
+        if nchildren > 0 {
+            write!(f, ", children={}", nchildren)?;
+        }
+
+        // Show buffer info
+        if let Some(buffers) = fb_array.buffers() {
```
0ax1: How about 
```
if let Some(buffers) = fb_array.buffers() {
    let buffer_list = buffers.iter()
        .map(|buf| format!("{}B", buf.length()))
        .collect::<Vec<_>>()
        .join(", ");
    write!(f, ", buffers=[{}]", buffer_list)?;
}
```

vortex-layout/src/display.rs
```diff
@@ -1,98 +1,468 @@
...
+        if flat_layout.array_tree().is_none() {
+            segment_ids.push(flat_layout.segment_id());
+        }
+    } else {
+        // For other layouts, add all segment IDs
+        segment_ids.extend(layout.segment_ids());
+    }
+
+    // Recurse into children
+    for child in layout.children()? {
```
0ax1, line 68: Do we need to dedup segment ids?

vortex-utils/src/env.rs
```diff
@@ -0,0 +1,58 @@
...
+/// ```
+///
+/// # Safety
+///
+/// Environment variable modification is inherently unsafe in multi-threaded contexts.
+/// This guard is intended for use in tests that are run serially or where env var
+/// races are acceptable.
+pub struct EnvVarGuard(&'static str);
+
+impl EnvVarGuard {
```
0ax1, line 69: As discussed offline. Might be worthwhile to assert that this API is called from the main thread.

0ax1: Or add the respective concurrency guards.

0ax1:
@claude review

vortex-layout/src/display.rs
joseph-isaacs: I am not sure they can be. But Ill add anyway

vortex-layout/src/display.rs
```diff
@@ -1,98 +1,469 @@
...
+/// This function performs IO to fetch each segment's buffer. For layouts with
+/// many segments, this may result in significant IO overhead.
+pub(super) async fn display_tree_with_segment_sizes(
+    layout: LayoutRef,
+    segment_source: Arc<dyn SegmentSource>,
+) -> VortexResult<DisplayLayoutTree> {
+    // First, collect all segment IDs from the layout tree (excluding those with inline array_tree)
+    let mut segments_to_fetch = Vec::new();
+    collect_segments_to_fetch(&layout, &mut segments_to_fetch)?;
+    segments_to_fetch.dedup();
```
0ax1, line 31: dedup requires elements to be sorted, are they?

0ax1 (approved):

vortex-layout/src/display.rs
joseph-isaacs: nope, now they are

PR #5651, opened 2025-12-08
Remove Lance benchmarks

### LINKED ISSUE #5266: When upgrading DF, even the optional Lance dependency might cause issues
When we're upgrading DF, if there's also a major `arrow-rs` release we'll run into dependency issues with lance until they also upgrade their arrow dependency. See the build issues in https://github.com/vortex-data/vortex/pull/5189.

AdamGS:
Closes https://github.com/vortex-data/vortex/issues/5266.

Remove the Lance benchmarks to unblock arrow and datafusion upgrades. The alternative I see is to split lance benchmarks into their own crate, Vortex/Parquet into a second crate and have a shared benchmarks crate that just contains the reporting infra, but that seems like too much effort to run a benchmark we don't really inform any decisions by currently.


AdamGS:
Wanted to put this up, I'm working in parallel about trying to see if its possible to untangle some of the benchmarking infra in a way that will make it easier to keep the benchmarks.

connortsui20:
If we ever do this, I would much prefer to put it behind a feature flag that never is enabled or just have a bunch of dead code, as adding this all back is somewhat of a pain.

AdamGS:
Feature flag isn't enough here

connortsui20:
with a dependency rename it should be possible?

connortsui20:
ah right, I forgot about that

AdamGS:
ok maybe patching works here? need to try some stuff out to make sure

Issue #5652, opened 2025-12-08
Tracking Issue: Migrate `Array`s to use `batch_execute` instead of `into_canonical`


connortsui20:
This is a tracking issue for migrating all of our `Array` types to decompress into `Vector`s via the new `batch_execute` method on the new `VTable`. Note that we will keep around the `into_canonical` implementations until everything has been migrated.

This is part of the larger effort to rework `Array`s to be lazy, which we are calling the "operators" rework for Vortex.

Related RFC: https://github.com/vortex-data/vortex/pull/4726

Related Tracking Issue: https://github.com/vortex-data/vortex/issues/5028

# Steps / History

### `vortex-array` crate

Core array types:
- [x] `BoolArray`: https://github.com/vortex-data/vortex/pull/5424
- [x] `ChunkedArray`: https://github.com/vortex-data/vortex/pull/5424
- [x] `ConstantArray`: https://github.com/vortex-data/vortex/pull/5424
- [x] `DecimalArray`: https://github.com/vortex-data/vortex/pull/5424
- [ ] `DictArray`
  - [ ] Implement `take` for all vectors
    - [x] https://github.com/vortex-data/vortex/pull/5537
    - [x] https://github.com/vortex-data/vortex/pull/5540
    - [x] https://github.com/vortex-data/vortex/pull/5653
    - [ ] https://github.com/vortex-data/vortex/pull/5682
    - [ ] https://github.com/vortex-data/vortex/pull/5686
- [x] `ExtensionArray`: https://github.com/vortex-data/vortex/pull/5424
- [x] `FilterArray`: https://github.com/vortex-data/vortex/pull/5610
- [x] `FixedSizeListArray`: https://github.com/vortex-data/vortex/pull/5424
- [ ] `ListArray`
- [x] `ListViewArray`: https://github.com/vortex-data/vortex/pull/5424
- [x] `MaskedArray`: https://github.com/vortex-data/vortex/pull/5606
- [x] `NullArray`: https://github.com/vortex-data/vortex/pull/5424
- [x] `PrimitiveArray`: https://github.com/vortex-data/vortex/pull/5424
- [x] `ScalarFnArray`: https://github.com/vortex-data/vortex/pull/5561
- [x] `StructArray`: https://github.com/vortex-data/vortex/pull/5424
- [ ] `VarBinArray`
- [x] `VarBinViewArray`: https://github.com/vortex-data/vortex/pull/5424

[...]

PR #5653, opened 2025-12-08
Feature: Implement `take` on more vectors + stubs

connortsui20:
Tracking Issue: https://github.com/vortex-data/vortex/issues/5652

https://github.com/vortex-data/vortex/pull/5540 has to be merged first.

Implements `take` on the remaining vectors, except for `DecimalVector`, `BinaryViewVector`, and `ListViewVector`

For `DecimalVector` we cannot trivially reuse the existing `take` compute on slices and buffers because it has a bound of `NativePType`. We can relax this bound to `Copy` instead, but that will require some more work. Note that the current `DecimalArray` `take` implementation is fully scalar right now so we can definitely improve that in the new world.

For the view vectors I believe we want to think through how we "rebuild" those in the array plan (we probably want to place rebuild nodes at good places after just doing a `take` on the existing views).

So this PR implements:
- `FixedSizeListVector`
- `NullVector`
- `StructVector`

Edit: Now that https://github.com/vortex-data/vortex/pull/5666 has merged, this also implements `DecimalVector`.

### DIFF
--- a/vortex-array/benches/take_struct.rs
+++ b/vortex-array/benches/take_struct.rs
@@ -1,0 +1,111 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#![allow(clippy::unwrap_used)]
+
+use divan::Bencher;
+use rand::Rng;
+use rand::SeedableRng;
+use rand::distr::Uniform;
+use rand::rngs::StdRng;
+use vortex_array::IntoArray;
+use vortex_array::arrays::StructArray;
+use vortex_array::compute::take;
+use vortex_array::validity::Validity;
+use vortex_buffer::Buffer;
+use vortex_dtype::FieldNames;
+
+fn main() {
+    divan::main();
+}
+
+const ARRAY_SIZE: usize = 100_000;
+const TAKE_SIZE: usize = 1000;
+
+#[divan::bench]
+fn take_struct_simple(bencher: Bencher) {
+    let mut rng = StdRng::seed_from_u64(0);
+    let range = Uniform::new(0i64, 100_000_000).unwrap();
+
+    // Create single field for the struct
+    let field = (0..ARRAY_SIZE)
+        .map(|_| rng.sample(range))
+        .collect::<Buffer<i64>>()
+        .into_array();
+
+    let struct_array = StructArray::try_new(
+        FieldNames::from(["value"]),
+        vec![field],
+        ARRAY_SIZE,
+        Validity::NonNullable,
+    )
+    .unwrap();
+
+    let indices: Buffer<u64> = (0..TAKE_SIZE)
+        .map(|_| rng.random_range(0..ARRAY_SIZE) as u64)
+        .collect();
+    let indices_array = indices.into_array();
+
+    bencher
+        .with_inputs(|| (&struct_array, &indices_array))
+        .bench_refs(|(array, indices)| take(array.as_ref(), indices.as_ref()).unwrap());
+}
+
+#[divan::bench(args = [8])]
+fn take_struct_wide(bencher: Bencher, width: usize) {
+    let mut rng = StdRng::seed_from_u64(0);
+    let range = Uniform::new(0i64, 100_000_000).unwrap();
+
+    let fields: Vec<_> = (0..width)
+        .map(|_| {
+            (0..ARRAY_SIZE)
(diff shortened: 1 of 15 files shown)


vortex-compute/src/take/vector/struct_.rs
joseph-isaacs, line 1: can we add the struct take bench back?

vortex-compute/src/take/vector/binaryview.rs
```diff
@@ -0,0 +1,89 @@
...
+            take_nullable(self, indices)
+        }
+    }
+}
+
+impl<T: BinaryViewType, I: UnsignedPType> Take<[I]> for &BinaryViewVector<T> {
+    type Output = BinaryViewVector<T>;
+
+    fn take(self, _indices: &[I]) -> BinaryViewVector<T> {
+        todo!("TODO(connor): Implement `take` for `BinaryViewVector` and figure out rebuilding");
```
joseph-isaacs, line 33: why would we ever rebuild here?

joseph-isaacs: I would think that something above (an array or an exporter) would say it requires rebuild

vortex-compute/src/take/vector/struct_.rs
connortsui20: The one we removed in https://github.com/vortex-data/vortex/pull/5540?

vortex-compute/src/take/vector/binaryview.rs
connortsui20: I am not as familiar with binary views but I would assume they have the same problem as list views where we can do a bunch of takes and filters on the views and delay doing compaction / garbage collection until the end.

Are we saying that `VarBinViewArray` and `ListViewArray` filters and takes become decomposed into 2 arrays, where one can be pulled up? In that case we can do the simple thing here and delegate that logic to somewhere else but we need to make sure the optimizer is aware of this + that logic is actually implemented somewhere, which is why I'm being careful.

connortsui20: regardless we can talk about this in the followup PR

vortex-compute/src/take/vector/struct_.rs
joseph-isaacs: something similar yes

vortex-compute/src/take/bit_buffer.rs
```diff
@@ -58,3 +58,29 @@ fn take_bool<I: UnsignedPType>(bools: &BitBuffer, indices: &[I]) -> BitBuffer {
...
+        let values: Vec<bool> = (0..result.len()).map(|i| result.value(i)).collect();
+        assert_eq!(values, vec![false, true, true, true, false]);
+
+        // Large buffer (uses take_bool path, len > 4096).
+        let large: BitBuffer = (0..5000).map(|i| i % 3 == 0).collect();
+        let result = (&large).take(&[4999u32, 0, 1, 2, 3, 4998][..]);
+
+        let values: Vec<bool> = (0..result.len()).map(|i| result.value(i)).collect();
+        assert_eq!(values, vec![false, true, false, false, true, true]);
+    }
```
joseph-isaacs: since you are trying to test each impl, can you do that directly? Otherwise this test might drift in what it tests

vortex-compute/src/take/vector/struct_.rs
connortsui20: Ok I'll just bring it back then

vortex-compute/src/take/bit_buffer.rs
connortsui20: I don't really understand what this means... `BitBuffer` is a type we use in other places?

joseph-isaacs: test take_byte_bool and take_bool not take

joseph-isaacs (approved):

vortex-array/benches/take_struct.rs
joseph-isaacs, line 1: do yyou think this is a good benchmark, if not how can we improve

connortsui20: I don't really know. I think we should address all of our benchmarks at once when we start doing an old vs new world (operators) comparison

vortex-compute/src/take/bit_buffer.rs
connortsui20: oh I understand

PR #5654, opened 2025-12-08
Fix: portable SIMD `take` implementation

connortsui20:
It turns out we are not checking `RUSTFLAGS="--cfg vortex_nightly" cargo +nightly nextest run` on our CI. I'll add that to our actions in a followup PR.

Also adds some non-trivial tests.


a10y (approved):
nice

PR #5655, opened 2025-12-08
Fix: add `RUSTFLAGS="--cfg vortex_nightly"` to all nightly tests

connortsui20:
I think this is everything, I had claude search through all of our actions


a10y (approved):
assuming ci checks pass this seems fine

PR #5649, opened 2025-12-08
Enable CI when PR targets non-develop branch

gatesn:

PR #5659, opened 2025-12-09
Chore: optimize CI with concurrency groups

connortsui20:
These rules allow us to stop CI when changes get overwritten.

I was pretty careful to make sure each of the rules make sense, but I could have misunderstood something.

This also fixes the issue where we trigger CI twice on both push and pull request changes (https://github.com/vortex-data/vortex/pull/5649).

Also, if you look at https://github.com/vortex-data/vortex/pull/5660 you'll see that CI triggers when merging into this branch so just the plain `pull_request:` works fine.


.github/workflows/release-drafter.yml
connortsui20, line 1: Not entirely sure if this is OK for releases? I would assume this is a good thing but I'm not too familiar with the release process.

connortsui20:
Also note: if someone adds the `benchmark-sql` tag and then force pushes or adds a new commit, the benchmark runs get [cancelled](https://github.com/vortex-data/vortex/actions/workflows/sql-pr.yml). Personally, I think this makes more sense, but if anyone has a good reason to always keep running the outdated code then I can change it back.

.github/workflows/ci.yml
```diff
@@ -1,10 +1,19 @@
...
+on:
+  push:
+    branches: [develop]
+  pull_request: { }
+  workflow_dispatch: { }
+  # TODO(connor): Maybe add these path filters.
+  # paths-ignore:
+  #   - '**.md'
+  #   - 'docs/**'
+  #   - 'benchmarks-website/**'
```
AdamGS: 👍

.github/workflows/release-drafter.yml
AdamGS: I think that's fine here, IIRC release drafter always does the delta between the latest release and the branch's HEAD. Its also a really fast job so I don't expect it to matter a lot.

.github/workflows/fuzz.yml
```diff
@@ -1,9 +1,16 @@
...
+  group: fuzz
+  # This ensures each run builds on the previous run's corpus discoveries rather than losing them to
+  # failed compare-and-swap uploads.
+  cancel-in-progress: false
+
 on:
   schedule:
     - cron: "0 */4 * * *"  # every 4 hours
-  workflow_dispatch:
+  workflow_dispatch: { }
```
AdamGS, line 13: can we enforce this in yaml-lint?

.github/workflows/ci.yml
```diff
@@ -1,10 +1,19 @@
 name: CI
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
gatesn: What is this doing exactly? Doesn't this mean only one PR will run CI if multiple PRs target develop?

.github/workflows/ci.yml
a10y: How does this work when this workflow is required for merge though

AdamGS: good point

.github/workflows/ci.yml
connortsui20: In practice, nothing since I have `cancel-in-progress: ${{ github.ref != 'refs/heads/develop' }}` below. This was just a way to ensure new PR merges (commits) to develop do not cancel previous CI runs since we want CI to finish for every commit (benchmarks website) without the special cancel in progress clause. I can remove it but it also doesn't hurt to have this here.

.github/workflows/fuzz.yml
connortsui20: I think let's leave this to a separate PR because I just saw that enforcing this causes quite a lot of errors

connortsui20: https://github.com/vortex-data/vortex/issues/5663

.github/workflows/ci.yml
connortsui20: I left a comment above it

a10y (approved):

PR #5665, opened 2025-12-09
Touch up some benchmarking code and clean up dependencies

AdamGS:
As I'm struggling through figuring out how to untangle the benchmarks a bit, these things fell out and I figured they are worth their own PR.
1. Used `cargo machete` to clear up some unused dependencies.
2. Made some lance-only code in benchmarks more explicitly behind a feature


bench-vortex/src/utils/mod.rs
```diff
@@ -2,13 +2,10 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 pub mod constants;
-pub mod file_utils;
+pub mod file;
 pub mod logging;
-pub mod parquet_utils;
+#[cfg(feature = "lance")]
+pub mod parquet;
```
connortsui20, line 8: I'm surprised that these utils aren't used elsewhere? If that is the case then we can just rename this to lance utils or similar right?

connortsui20 (approved):
LGTM once CI passes

PR #5666, opened 2025-12-09
Temp fix: Make `take` bound by `Copy` instead of `NativePType`

connortsui20:
Just so we don't have to refactor the SIMD code to not rely on the `PTYPE` constant of `NativePType`.

This will unblock being able to implement `take` for `DecimalVector` and then `batch_execute` for `DictArray`.

### DIFF
--- a/vortex-compute/src/take/slice/mod.rs
+++ b/vortex-compute/src/take/slice/mod.rs
@@ -4,7 +4,6 @@
 //! Take function implementations on slices.
 
 use vortex_buffer::Buffer;
-use vortex_dtype::NativePType;
 use vortex_dtype::UnsignedPType;
 
 use crate::take::Take;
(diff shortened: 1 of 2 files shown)


connortsui20:
Interesting to see the perf improvement when using SIMD!

vortex-compute/src/take/slice/mod.rs
```diff
@@ -13,10 +12,13 @@ pub mod avx2;
 pub mod portable;
 
 /// Specialized implementation for non-nullable indices.
-impl<T: NativePType, I: UnsignedPType> Take<[I]> for &[T] {
+impl<T: Copy, I: UnsignedPType> Take<[I]> for &[T] {
     type Output = Buffer<T>;
 
     fn take(self, indices: &[I]) -> Buffer<T> {
+        // TODO(connor): Make the SIMD implementations bound by `Copy` instead of `NativePType`.
```
gatesn, line 19: What we ultimately want is to match on `size_of::<T>`, and transmute into unsigned types so we don't bloat code-gen. Then we will only SIMD dispatch on types <= word size, fall back to scalar for anything else.

gatesn (approved):

PR #5669, opened 2025-12-10
Chore: Add all logical bool implementations + refactor

connortsui20:
This PR consolidates and finalizes a lot of the boolean and Kleene logic, and also deduplicates a bunch of code.

Instead of having each of the operands in separate modules, I grouped them into logical operands, kleene operands, and `not` by itself since it is the only unary operator.

The convenience trait impls all have the same 3 impls: for `&BoolScalar`, `&BoolVector`, `&BoolDatum`. 

```rust
impl LogicalAndKleene for &BoolScalar { ... }

impl LogicalAndKleene for &BoolVector { ... }

impl LogicalAndKleene for &BoolDatum { ... }
```

Also adds more tests.


vortex-array/src/scalar_fns/binary/mod.rs
```diff
@@ -101,14 +101,16 @@ impl VTable for BinaryFn {
             Operator::Lte => Ok(Compare::<LessThanOrEqual>::compare(lhs, rhs).into()),
             Operator::Gt => Ok(Compare::<GreaterThan>::compare(lhs, rhs).into()),
             Operator::Gte => Ok(Compare::<GreaterThanOrEqual>::compare(lhs, rhs).into()),
-            Operator::And => Ok(<BoolDatum as LogicalOp<KleeneAnd>>::op(
-                lhs.into_bool(),
-                rhs.into_bool(),
+            Operator::And => Ok(<&BoolDatum as LogicalOp<KleeneAnd>>::op(
+                &lhs.into_bool(),
```
gatesn, line 105: Why do this take by reference? I thought we preferred to take LHS by ownership so we can try_into_mut and reuse the allocation?

joseph-isaacs:
What is the win here?

vortex-array/src/scalar_fns/binary/mod.rs
connortsui20: AFAIK we don't actually reuse any allocations? If you look at the `not` implementation we do a `try_into_mut` and then call `bits.not()` that takes ownership, but that makes an allocation anyways so we might as well take a reference. We would need to add bitwise in-place methods onto `BitBuffer` in order to get any benefit.

connortsui20:
@joseph-isaacs code deduplicatuon: this already consolidates a bunch of code and it meant that I didn't have to copy and paste a bunch of stuff to add the datum impls

vortex-array/src/scalar_fns/binary/mod.rs
joseph-isaacs: Can we try and reuse one of the datums

vortex-compute/src/logical/kleene.rs
```diff
@@ -0,0 +1,571 @@
...
+    }
+}
+
+impl KleeneBinaryOp for KleeneOr {
+    fn bit_op(lhs: &BitBuffer, rhs: &BitBuffer) -> BitBuffer {
+        lhs.bitor(rhs)
+    }
+
+    fn identity_bits(len: usize) -> BitBuffer {
+        BitBuffer::new_set(len) // All true.
```
joseph-isaacs: `T \/ a == T`, I don't think that is the identity element

vortex-compute/src/logical/kleene.rs
```diff
@@ -0,0 +1,571 @@
...
+    fn scalar_op(lhs: Option<bool>, rhs: Option<bool>) -> Option<bool>;
+}
+
+impl KleeneBinaryOp for KleeneAnd {
+    fn bit_op(lhs: &BitBuffer, rhs: &BitBuffer) -> BitBuffer {
+        lhs.bitand(rhs)
+    }
+
+    fn identity_bits(len: usize) -> BitBuffer {
+        BitBuffer::new_unset(len) // All false.
```
joseph-isaacs: also incorrect

vortex-compute/src/logical/kleene.rs
connortsui20: The function in the code (I'm pretty sure) is correct, I guess my naming is wrong?

vortex-array/src/scalar_fns/binary/mod.rs
connortsui20: @joseph-isaacs could you be more specific?

vortex-compute/src/logical/kleene.rs
connortsui20: `absorbing_bits_from_len` is probably a better description

vortex-array/src/scalar_fns/binary/mod.rs
joseph-isaacs: The idea is that either the left or right Vector (if either is a vector), can we mutated in-place

joseph-isaacs: https://github.com/vortex-data/vortex/blob/374882d57d9112235cfe2c99879284b8d882ca17/vortex-compute/src/arithmetic/pvector.rs#L151

connortsui20: Theoretically we could but as I said above we do not have any in-place binary operations implemented so we can't do that right now, and adding that is outside the scope of this PR.

Edit: @joseph-isaacs that code link you sent only works for `Buffer<T>`, we would need something similar for `BitBuffer` which is far less trivial because we will want to do operations in word batches (which would require more logic than just passing in a mapping fn).

joseph-isaacs: Sure, can you leave that as a TODO, but can we impl this on (Bool, &Bool) so its easy to use add in-place later

vortex-compute/src/logical/kleene.rs
```diff
@@ -0,0 +1,571 @@
...
+
+    match (lhs.validity(), rhs.validity()) {
+        // Everything is valid, so we can just do a simple logical AND over all bits.
+        (Mask::AllTrue(_), Mask::AllTrue(_)) => {
+            BoolVector::new(Op::bit_op(lhs.bits(), rhs.bits()), Mask::new_true(len))
+        }
+
+        // Everything is null, so the entire result vector is null.
+        (Mask::AllFalse(_), Mask::AllFalse(_)) => {
+            BoolVector::new(Op::absorbing_bits_from_len(len), Mask::new_false(len))
```
joseph-isaacs: Why do you need to set the bits to a specific value, they are undefined, can do AllFalse?

joseph-isaacs: use either the left or right bool vec?

vortex-compute/src/logical/kleene.rs
```diff
@@ -0,0 +1,571 @@
...
+            // The result vector is valid where the LHS has an absorbing value.
+            let result_bits = Op::absorbing_bits_from_len(len);
+            let validity = Op::absorbing_bits(lhs.bits());
+            BoolVector::new(result_bits, Mask::from(validity))
+        }
+
+        // LHS is all null, RHS is all valid.
+        (Mask::AllFalse(_), Mask::AllTrue(_)) => {
+            // The result vector is valid where the RHS has an absorbing value.
+            let result_bits = Op::absorbing_bits_from_len(len);
```
joseph-isaacs: what is wrong with rhs here?

connortsui20: not sure what you mean?

joseph-isaacs: I am saying you could return rhs not this?

vortex-compute/src/logical/kleene.rs
```diff
@@ -0,0 +1,571 @@
...
+/// - For OR: `TRUE` absorbs nulls (`TRUE OR NULL = TRUE`)
+pub trait KleeneBinaryOp {
+    /// Apply the operation to two [`BitBuffer`]s.
+    fn bit_op(lhs: &BitBuffer, rhs: &BitBuffer) -> BitBuffer;
+
+    /// Absorbing bits when one side is all-null.
+    ///
+    /// - AND: all-false, since `x AND null` is false or null.
+    /// - OR: all-true, since `x OR null` is true or null.
+    fn absorbing_bits_from_len(len: usize) -> BitBuffer;
```
joseph-isaacs: I don't think you need this you can use the input buffer?

vortex-array/src/scalar_fns/binary/mod.rs
connortsui20: I'll leave it as a TODO, but I'm going to hold off on adding a large amount of bad implementations on `(Bool, &Bool)` right now since we might just forget (and there won't be any difference right now)

vortex-compute/src/logical/kleene.rs
connortsui20: nice catch, I think you are right

connortsui20:
@joseph-isaacs I think that [a1ebc7a](https://github.com/vortex-data/vortex/pull/5669/commits/a1ebc7af4e321c1bad6faf556d70c8f2434b2041) simplifies things?

vortex-array/src/scalar_fns/binary/mod.rs
joseph-isaacs: why. we know we want this?

joseph-isaacs (approved):
should be nice to have this all in one go.

vortex-array/src/scalar_fns/binary/mod.rs
connortsui20: its just a lot of garbage code that likely is not going to get fixed if we add it now? I would prefer to have callers do `.clone()` so we see where we are actually losing performance over the next few weeks rather than hide it behind bad trait implementations which are impossible to see at the call site.

PR #5670, opened 2025-12-10
Centralize all logging to `tracing` instead of `log`

AdamGS:
`tracing` has more structured logging support (see the changes I made in `btrblocks`) which is more suitable for modern observability tooling. It also has tracing support which is useful and already used in a few places.

In python we use the tracing "log" feature which make it emit `log` events


gatesn (approved):
<3

AdamGS:
tracing has some microsecond scale overhead compared to `log`, so some of the small fast benchmarks which hit a logging codepath are slightly slower, I think that's acceptable.
We can turn them off by enabling the `tracing/release_max_level_off` feature, but that's just a bunch of boilerplate for very little benefit IMO.

PR #5675, opened 2025-12-10
Try zip over zip_eq

AdamGS:
Some comments over on `itertools` seem to indicate that people are suspicous of `zip_eq` perf, which I think is mostly due to the lack of specialization over the `std` function, figured I'll give it a try, and if its much better we can build our own specialized version that maybe has a different `debug` version to keep the assertions.


AdamGS:
I guess this doesn't make any difference for us

PR #5676, opened 2025-12-10
Benchmarks re-org

AdamGS:
This PR makes two big changes to our benchmarking setup:
1. It splits duckdb, datafusion and lance benchmarks into their own crates, so they don't depend on each other, minimizing the friction when bumping dependencies like Arrow or DataFusion.
2. Introduces a new benchmarking tool - `vx-bench` which allows comparing various runs, persisting results and takes care of building and running the individual binaries. `vx-bench` is written in Python using `typer`.

I've tried to not touch the CI setup too much, but it can be simplified and improved significantly to fit this structure. 

## Notes for reviewers
1. DF/Arrow results are worse because the data is chunked differently, we use an `MemTable` which is just a `Vec<RecordBatch>`, which is now longer.
2. ~I'm looking into the differences in Vortex results, not sure what's going on there.~ Turns out the `vortex-compact` results are slower because they actually run on data compressed with the compact compressor! Some of the tables are about half the size, but they are slower to decompress/process.
4. ~Seems like we've been dropping some compression benchmark results, so they dont have a baseline (or I changed the name) I'm trying to figure out what's going on there.~ I've changed the name of some compression benchmarks, ratios have the same name though.

## `vx-bench` demo

render1765993255912


AdamGS:
End of the week state:
1. public_bi is the only benchmark not in the new structure, because we don't actually run it.
2. Benchmarks persist data under `target/` which will give us criterion-like functionality and the ability to compare between all the tools
3. Some benchmarks already run in CI, a few don't and results look very different from develop, which is going to be fun to debug.

benchmarks/df-bench/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "df-bench"
```
gatesn: Use long names for this + ddb

connortsui20:
Any ideas on why the benchmark measurements are very different? Is it because this has a new version of DF?

AdamGS:
This PR doesn't have a new DF version.
For Arrow - the chunking is different, I don't think it matters.
For Vortex - I've changed the conversion logic, so maybe some files end up looking different or I messed something up, I'm going to spend some time looking into that.

bench-orchestrator/bench_orchestrator/utils.py
joseph-isaacs, line 1: ?

.github/workflows/sql-benchmarks.yml
```diff
@@ -21,69 +21,69 @@ on:
...
               "targets": "datafusion:arrow,datafusion:parquet,datafusion:vortex,datafusion:vortex-compact,duckdb:parquet,duckdb:vortex,duckdb:vortex-compact,duckdb:duckdb",
-              "scale_factor": "--scale-factor 10.0"
+              "scale_factor": "10.0"
             },
             {
               "id": "tpch-s3-10",
               "subcommand": "tpch",
               "name": "TPC-H SF=10 on S3",
-              "local_dir": "bench-vortex/data/tpch/10.0",
+              "local_dir": "vortex-bench/data/tpch/10.0",
```
joseph-isaacs, line 53: shall we have this in benchmarks?

bench-orchestrator/bench_orchestrator/utils.py
AdamGS: nice catch

joseph-isaacs:
I think we need to ensure that the way data is now generated doesn't really change the benchmarks

AdamGS:
@joseph-isaacs can you expand on what you mean? Right now for some benchmarks like TPC-DS we don't even have a basic correctness check.

gatesn (approved):

PR #1146, opened 2024-10-28
feat: specialized IntoCanonical for DictArray utf8/binary

a10y:
Fixes #1041 

Change the IntoCanonical implementation for DictArray to do canonicalize-then-take for stringy things, take-then-canonicalize for everything else.

It is always the case for strings that canonicalize-then-take will be faster, regardless of the compression of the values array.

Overrides #1136

PR #5686, opened 2025-12-10
Feature: implement `DictArray` batched execute

connortsui20:
Tracking Issue: https://github.com/vortex-data/vortex/issues/5652

Adds batched execution to `DictArray`, which is just implemented as `take` on the child `values` vector with the codes as indices.

~~Additionally adds an optimization for `BoolVector::take` which is similar to the optimization we have for `BoolArray`'s `canonicalize` function, but does an additional check for zero or one `false`s (instead of just zero or one `true`s). That code is located at https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/arrays/dict/vtable/canonical.rs.~~

~~The difference here is that I use a heuristic check on the default `take` implementation on `BoolVector` (instead of only use this optimization for dictionary decompression) because I don't think there is any reason not to utilize this in general.~~

~~I still need to add some benchmarks.~~

Also note that we don't need to specialize for the string types like in https://github.com/vortex-data/vortex/pull/1146 since we are _always_ decompressing the children immediately.

### DIFF
--- a/vortex-array/src/arrays/dict/vtable/mod.rs
+++ b/vortex-array/src/arrays/dict/vtable/mod.rs
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use vortex_buffer::BufferHandle;
+use vortex_compute::take::Take;
 use vortex_dtype::DType;
 use vortex_dtype::Nullability;
 use vortex_dtype::PType;
@@ -16,6 +17,9 @@ use crate::ArrayRef;
 use crate::DeserializeMetadata;
 use crate::ProstMetadata;
 use crate::SerializeMetadata;
+use crate::kernel::BindCtx;
+use crate::kernel::KernelRef;
+use crate::kernel::kernel;
 use crate::serde::ArrayChildren;
 use crate::vtable;
 use crate::vtable::ArrayId;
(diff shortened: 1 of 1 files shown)


vortex-compute/src/take/vector/bool.rs
```diff
@@ -24,26 +50,179 @@ impl<I: UnsignedPType> Take<[I]> for &BoolVector {
...
+/// - All of the values are `true`, so create a [`BoolVector`] with `n` `true`s.
+/// - All of the values are `false`, so create a [`BoolVector`] with `n` `false`s.
+/// - There is a single `true` value, so compare indices against that index.
+/// - There is a single `false` value, so compare indices against that index and negate.
+/// - Otherwise, there are multiple `true`s and `false`s in the `values` vector and we must do a
+///   normal `take` on it.
+///
+/// The `compute_validity` closure computes the output validity mask, allowing callers to handle
+/// nullable vs non-nullable indices differently.
+fn optimized_take<I: UnsignedPType>(
```
gatesn: I'm skeptical of functions like this that don't come with an associated benchmark. How do you know this is more optimal?

connortsui20: This mimics what we have on develop (otherwise I wouldn't have added this). I will add benchmarks soon.

vortex-array/src/arrays/dict/vtable/mod.rs
joseph-isaacs, line 1: can we merge only this, now

connortsui20: ill split it out

vortex-compute/src/take/vector/bool.rs
connortsui20: nvm we can do this later

joseph-isaacs:
How does this look when run with operators?

joseph-isaacs (approved):

connortsui20:
confirmed locally that performance is basically the same

PR #5704, opened 2025-12-11
fix: getting gharchive to run under operators

a10y:
Still OOMs but at least it's not panicking anymore

### DIFF
--- a/vortex-array/src/arrays/bool/vtable/operator.rs
+++ b/vortex-array/src/arrays/bool/vtable/operator.rs
@@ -33,8 +33,12 @@ impl ArrayParentReduceRule<Exact<BoolVTable>, Exact<MaskedVTable>> for BoolMaske
         &self,
         array: &BoolArray,
         parent: &MaskedArray,
-        _child_idx: usize,
+        child_idx: usize,
     ) -> VortexResult<Option<ArrayRef>> {
+        if child_idx > 0 {
+            return Ok(None);
+        }
+
         // Merge the parent's validity mask into the child's validity
         // TODO(joe): make this lazy
         Ok(Some(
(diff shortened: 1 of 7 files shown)


vortex-array/src/arrays/bool/vtable/operator.rs
```diff
@@ -33,8 +33,12 @@ impl ArrayParentReduceRule<Exact<BoolVTable>, Exact<MaskedVTable>> for BoolMaske
         &self,
         array: &BoolArray,
         parent: &MaskedArray,
-        _child_idx: usize,
+        child_idx: usize,
```
a10y, line 36: this uses the validity child of MaskedArray so the output of this reducer is always wrong

gatesn (approved):
Hahahaha forgot that the other child of a mask array is a boolean 🤦

vortex-vector/src/vector_mut.rs
```diff
@@ -87,7 +87,7 @@ impl VectorMut {
             DType::Utf8(..) => StringVectorMut::with_capacity(capacity).into(),
             DType::Binary(..) => BinaryVectorMut::with_capacity(capacity).into(),
             DType::Extension(ext) => VectorMut::with_capacity(ext.storage_dtype(), capacity),
-            DType::List(..) => ListViewVectorMut::with_capacity(dtype, capacity).into(),
+            DType::List(elem, ..) => ListViewVectorMut::with_capacity(elem.as_ref(), capacity).into(),
```
a10y: this was self-recursive before

vortex-layout/src/layouts/struct_/reader.rs
```diff
@@ -343,28 +342,26 @@ impl LayoutReader for StructReader {
...
 
-                    Ok(StructArray::try_new(
-                        struct_array.names().clone(),
-                        masked_fields,
-                        struct_array.len(),
-                        struct_array.validity().clone(),
+                    Ok(ScalarFnArray::try_new(
+                        array.scalar_fn().clone(),
+                        masked_children,
+                        array.len(),
```
a10y, line 358: am i doing this right?

a10y:
seems like most of this got cp'ed into #5711

PR #5705, opened 2025-12-11
Feature: add zero and null constructors for scalars

connortsui20:
I need this to help with implementing all of the `Cast` logic.

### DIFF
--- a/vortex-vector/src/binaryview/types.rs
+++ b/vortex-vector/src/binaryview/types.rs
@@ -7,6 +7,7 @@ use std::fmt::Debug;
 
 use vortex_buffer::BufferString;
 use vortex_buffer::ByteBuffer;
+use vortex_dtype::DType;
 
 use crate::Vector;
 use crate::VectorMut;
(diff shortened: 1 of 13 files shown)


vortex-vector/src/binaryview/types.rs
```diff
@@ -93,9 +100,17 @@ impl BinaryViewType for StringType {
     fn upcast<V: BinaryViewTypeUpcast>(input: V::Input<Self>) -> V {
         V::from_string(input)
     }
+
+    fn empty_scalar() -> Self::Scalar {
+        BufferString::from("")
```
a10y: nit: if you use String::new() instead you can save an allocation

a10y: or it might be worth adding an explicit BufferString::empty() constructor

gatesn (commented):
I think adding this to the trait is odd. `BoolScalar::new(&DType::Primitive)` is a weird API, even though it does panic.

Can you just add to the scalars themselves? `BoolScalar::zero()` and `BoolScalar::null()`?

Where is this being called?

connortsui20:
This will need to be called in the `NullScalar` cast impl, where we can cast a `NullScalar` into a null `Scalar` of any type.

gatesn (approved):
Nice

PR #5707, opened 2025-12-11
Feature: add `append_array_as_list` methods to the list arrays

connortsui20:
Adds `append_array_as_list` methods to the list arrays.

This makes it possible to circumvent having to use `ListScalar`s to build `ListArray`s (or any of the other list array kinds).

cc @nrc Let me know if this helps your usecase!

### DIFF
--- a/vortex-array/src/builders/listview.rs
+++ b/vortex-array/src/builders/listview.rs
@@ -110,6 +110,43 @@ impl<O: IntegerPType, S: IntegerPType> ListViewBuilder<O, S> {
         }
     }
 
+    /// Appends an array as a single non-null list entry to the builder.
+    ///
+    /// The input `array` must have the same dtype as the element dtype of this list builder.
+    ///
+    /// Note that the list entry will be non-null but the elements themselves are allowed to be null
+    /// (only if the elements [`DType`] is nullable, of course).
+    pub fn append_array_as_list(&mut self, array: &dyn Array) -> VortexResult<()> {
+        vortex_ensure!(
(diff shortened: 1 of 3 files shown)


nrc:
This seems to do exactly what I want, thanks!

vortex-array/src/builders/listview.rs
```diff
@@ -110,6 +110,43 @@ impl<O: IntegerPType, S: IntegerPType> ListViewBuilder<O, S> {
         }
     }
 
+    /// Appends an array as a single non-null list entry to the builder.
+    ///
+    /// The input `array` must have the same dtype as the element dtype of this list builder.
+    ///
+    /// Note that the list entry will be non-null but the elements themselves are allowed to be null
+    /// (only if the elements [`DType`] is nullable, of course).
+    pub fn append_array_as_list(&mut self, array: &dyn Array) -> VortexResult<()> {
```
a10y, line 119: is there a reason this returns a Result? It doesn't look like you ever return an err back

connortsui20: the `vortex_ensure` returns an error. Maybe that needs to be renamed...

a10y: this is a weird mix of err and panics, i'm wondering if the overflow check should also be an error

connortsui20: Maybe, though I usually associate overflow with panic so I thought it is fine. We would have to change other things in this file as well then

a10y (approved):

PR #5708, opened 2025-12-11
Feature: add some `Cast` infrastructure

connortsui20:
Adds the very basic implementations of `Cast` to `vortex-compute`.

The only things that are implemented are:
- Cast to `NullVector`/`NullScalar`
- Passthrough cast to the extension storage type
- Identity cast (a cast that is just cloning `self`)

All of the non-extremely-trivial cast implementations have not been added, but this PR makes it a lot easier to see what is missing and a lot easier to add those impls.

Note that I've also constrained the cast implementations so that we cannot cast nullable data to a non-nullable `DType`, and this is just checked by calling `self.validity().all_true()` on the vector which should rarely be expensive.

### DIFF
--- a/vortex-compute/src/cast/pvector.rs
+++ b/vortex-compute/src/cast/pvector.rs
@@ -5,30 +5,27 @@ use vortex_dtype::DType;
 use vortex_dtype::NativePType;
 use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
+use vortex_vector::Scalar;
+use vortex_vector::ScalarOps;
 use vortex_vector::Vector;
 use vortex_vector::VectorOps;
-use vortex_vector::match_each_pvector;
-use vortex_vector::null::NullVector;
+use vortex_vector::primitive::PScalar;
 use vortex_vector::primitive::PVector;
-use vortex_vector::primitive::PrimitiveVector;
 
 use crate::cast::Cast;
-
-impl Cast for PrimitiveVector {
-    type Output = Vector;
-
-    fn cast(&self, dtype: &DType) -> VortexResult<Vector> {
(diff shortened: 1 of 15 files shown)


vortex-compute/src/cast/pvector.rs
```diff
@@ -5,47 +5,74 @@ use vortex_dtype::DType;
...
+            DType::Primitive(target_ptype, n) if n.is_nullable() || self.validity().all_true() => {
+                vortex_bail!(
+                    "Casting PVector from PType {} to PType {} not yet implemented",
+                    T::PTYPE,
+                    target_ptype
+                );
+            }
+            _ => {
+                vortex_bail!("Cannot cast PVector<{}> to {}", T::PTYPE, target_dtype);
+            }
```
joseph-isaacs, line 45: Vectors don't have nullability?

connortsui20: The main purpose of vectors are that they are nullable arrays, so vectors are _always_ nullable

vortex-compute/src/filter/vector/mod.rs
```diff
@@ -30,6 +30,10 @@
 //! the recursive trait bounds (e.g. from [`StructVector`] requiring `Vector: Filter<M>` for its
 //! fields) by manually implementing [`Filter`] for [`Vector`] and [`VectorMut`] for each concrete
 //! mask type in this file.
+//!
+//! [`VectorOps::try_into_mut`]: vortex_vector::VectorOps::try_into_mut
```
gatesn, line 34: I really hate this style of doc comment

gatesn (approved):

PR #5710, opened 2025-12-12
Run optimizer on root array only on

gatesn:


vortex-array/src/optimizer/tests.rs
```diff
@@ -99,127 +94,20 @@ fn test_reduce_rules_traverse_whole_tree() -> VortexResult<()> {
     )?;
 
     println!("PRE: {}", outer_struct.display_tree());
```
joseph-isaacs, line 96: can we remove

vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -115,11 +115,14 @@ impl DictReader {
...
-                                .arrays()
-                                .optimizer()
-                                .optimize_array(&array?.apply(&expr)?)
+                            array?
+                                .apply(&expr, session.arrays().optimizer())
                                 .map_err(Arc::new)
+                            // session
+                            //     .arrays()
+                            //     .optimizer()
+                            //     .optimize_array(&array?.apply(&expr, ses)?)
```
joseph-isaacs, line 124: remove

joseph-isaacs (approved):

PR #5712, opened 2025-12-12
Move optimizer rules onto Array vtable

gatesn:


vortex-array/src/expr/vtable.rs
```diff
@@ -101,6 +101,24 @@ pub trait VTable: 'static + Sized + Send + Sync {
...
+    /// Return `Ok(None)` if no reduction is possible.
+    fn reduce(
+        &self,
+        options: &Self::Options,
+        node: &dyn ReduceNode,
+        ctx: &dyn ReduceCtx,
+    ) -> VortexResult<Option<Box<dyn ReduceNode>>> {
+        _ = options;
+        _ = node;
+        _ = ctx;
```
connortsui20, line 118: is this intentional?

gatesn: I prefer it since when implementing the trait function in a struct the IDE doesn't auto-complete the parameter names with underscores.

a10y (approved):

PR #5722, opened 2025-12-12
Perf: bring SIMD `take` back and generalize by `Copy`

connortsui20:
Brings back the `portable_simd` `take` implementation back, and instead of constraining by `NativePType`, this bounds by `T: Copy` and will cast to `u8` - `u64` depending on the size of the type.

This also adds a check for out-of-bounds indices that adds a single simd and bitwise instruction to the hot loop so we correctly panic at the end if there was an out of bounds.

It might be the case that separating out the `&=` and the `simd_lt` is so that the gather isn't depending on the `&=`, but if register pressure is high then we do not want to evict any data from the SIMD registers. I should probably benchmark that...


connortsui20:
it might be the case that the portable simd is actually faster than the avx2 impl? I think we need to do some more directed benchmarks...

PR #5724, opened 2025-12-14
ALP Kernel

gatesn:
We should move away from using ClosureKernel since it cannot be debug printed

### DIFF
--- a/encodings/alp/src/alp/kernel.rs
+++ b/encodings/alp/src/alp/kernel.rs
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use std::marker::PhantomData;
+
+use vortex_array::kernel::Kernel;
+use vortex_array::kernel::KernelRef;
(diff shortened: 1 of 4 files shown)


encodings/alp/src/alp/kernel.rs
```diff
@@ -0,0 +1,87 @@
...
+            encoded,
+            patches,
+            patches_offset,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(super) struct PatchKernels {
```
joseph-isaacs, line 45: `PatchesKernel`?

encodings/alp/src/alp/kernel.rs
```diff
@@ -0,0 +1,87 @@
...
+            encoded,
+            self.exponents,
+            patches_vectors,
+            self.patches_offset,
+        )
+    }
+
+    fn push_down_filter(self: Box<Self>, _selection: &Mask) -> VortexResult<PushDownResult> {
+        Ok(PushDownResult::NotPushed(self))
+    }
```
joseph-isaacs, line 86: push down if no patches?

PR #5730, opened 2025-12-15
fix: zstd frame step width for alignment

0ax1:
The beginning of each frame in zstd needs to align with the buffer alignment.

### DIFF
--- a/encodings/zstd/src/array.rs
+++ b/encodings/zstd/src/array.rs
@@ -330,6 +330,11 @@ impl ZstdArray {
                 .get(i + 1)
                 .copied()
                 .unwrap_or(value_bytes.len());
+
+            // Use slice_unaligned instead of slice to handle cases where frame boundaries
+            // don't align with the buffer's alignment requirements (e.g., 64-byte aligned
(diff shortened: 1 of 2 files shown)


encodings/zstd/src/test.rs
```diff
@@ -202,3 +203,16 @@ fn test_sliced_array_children() {
     let sliced = compressed.slice(0..4);
     sliced.children();
 }
+
+#[test]
+fn test_zstd_alignment_issue() {
+    // Scenario where a u8 array has a 64-byte alignment.
+    // See: https://github.com/vortex-data/vortex/issues/5715
```
joseph-isaacs: move this

encodings/zstd/src/array.rs
joseph-isaacs, line 294: why not make this 1 byte aligned?

0ax1: We don't want to split values across blocks / frames.

encodings/zstd/src/test.rs
0ax1: done

joseph-isaacs (approved):

PR #5736, opened 2025-12-15
Execute parent

gatesn:


encodings/decimal-byte-parts/src/decimal_byte_parts/rules.rs
```diff
@@ -0,0 +1,47 @@
...
+        //  For now, we only push down if there are no lower parts.
+        if !child._lower_parts.is_empty() {
+            return Ok(None);
+        }
+
+        let new_msp =
+            FilterArray::new(child.msp.clone(), parent.filter_mask().clone()).into_array();
+        let new_child =
+            DecimalBytePartsArray::try_new(new_msp, *child.decimal_dtype())?.into_array();
+        Ok(Some(new_child))
```
joseph-isaacs, line 45: do you think we need this vs just decompressing?

joseph-isaacs (approved):

encodings/decimal-byte-parts/src/decimal_byte_parts/rules.rs
gatesn: If we implemented DBP as a scalar function, it would happen via unary filter pushdown. It's kind of important

PR #5737, opened 2025-12-15
feat[array]: duckdb executor

joseph-isaacs:


gatesn (approved):

vortex-duckdb/src/exporter/struct_.rs
```diff
@@ -52,6 +55,35 @@ impl ColumnExporter for StructExporter {
     }
 }
 
+pub(crate) fn new_vector_exporter(
+    array: StructArray,
+    cache: &ConversionCache,
+    session: &VortexSession,
+) -> VortexResult<Box<dyn ColumnExporter>> {
+    let validity = array.validity_mask();
+    // DuckDB requires that the validity of the child be a subset of the parent struct so we mask out children with
```
gatesn, line 64: I think you should use get_item in that case?

vortex-duckdb/src/exporter/struct_.rs
```diff
@@ -52,6 +55,35 @@ impl ColumnExporter for StructExporter {
     }
 }
 
+pub(crate) fn new_vector_exporter(
+    array: StructArray,
+    cache: &ConversionCache,
+    session: &VortexSession,
+) -> VortexResult<Box<dyn ColumnExporter>> {
+    let validity = array.validity_mask();
```
gatesn, line 63: See this is weird.. validity_mask is deprecated since it's old "canonical", you should use Validity + execute_mask

PR #5739, opened 2025-12-15
fix[duckdb]: remove validity mask from vector export

joseph-isaacs:


vortex-duckdb/src/exporter/dict.rs
```diff
@@ -186,7 +190,10 @@ pub(crate) fn new_vector_exporter_with_flatten(
     if let Some(constant) = values.as_opt::<ConstantVTable>() {
         return constant::new_exporter_with_mask(
             &ConstantArray::new(constant.scalar().clone(), array.codes().len()),
-            array.codes().validity_mask(),
+            array
+                .codes()
+                .apply(&not(is_null(root())))?
```
gatesn: You should be able to do `array.codes().is_null().not()`. See `builtins.rs`

gatesn (approved):

PR #5740, opened 2025-12-15
Fix: incorrect child vector size in duckdb list array import

connortsui20:
Fixes https://github.com/vortex-data/vortex/discussions/5597

Since list views can be stored out of order and because there can be nulls at the end of the vector, we have to calculate the length of the child vector by looking at all views in the vector.

Note that this does not take validity into account, as I have no idea if DuckDB allows for complete garbage in their list views if they are null. At least in Arrow, that is not allowed.

Also adds 3 regression tests.

### DIFF
--- a/vortex-duckdb/src/convert/vector.rs
+++ b/vortex-duckdb/src/convert/vector.rs
@@ -111,13 +111,13 @@ fn vector_as_string_blob(vector: &mut Vector, len: usize, dtype: DType) -> Array
 
 /// Converts a valid [`duckdb_list_entry`] to `(offset, size)`, updating tracking state.
 ///
-/// Updates `child_max_length` with the maximum end offset seen so far, and `previous_end` with this
+/// Updates `child_min_length` with the maximum end offset seen so far, and `previous_end` with this
 /// entry's end offset (for use as the offset of subsequent null entries).
 ///
 /// Panics if the offset or size are negative or don't fit in the expected types.
 fn convert_valid_list_entry(
     entry: &duckdb_list_entry,
-    child_max_length: &mut usize,
+    child_min_length: &mut usize,
     previous_end: &mut i64,
 ) -> (i64, i64) {
     let offset = i64::try_from(entry.offset).vortex_expect("list offset must fit i64");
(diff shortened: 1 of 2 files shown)


vortex-duckdb/src/convert/vector.rs
```diff
@@ -222,26 +222,44 @@ pub fn flat_vector_to_vortex(vector: &mut Vector, len: usize) -> VortexResult<Ar
...
+            } in vector.as_slice_with_len::<duckdb_list_entry>(len)
             {
+                let offset = i64::try_from(*offset).vortex_expect("offset must fit i64");
+                assert!(offset >= 0, "offset must be non-negative");
+                let size = i64::try_from(*size).vortex_expect("length must fit i64");
+                assert!(size >= 0, "size must be non-negative");
+
+                // TODO(connor): Figure out if we need to worry about validity. In Arrow, null list
+                // views must still have a valid offset and size, but this might not be true is
+                // DuckDB? The public docs don't say anything about this...
```
joseph-isaacs: likely you can put anything there

joseph-isaacs: see: https://github.com/duckdb/duckdb/blob/235647c9358f23f7115352c2bc7c627b69bbd514/src/common/types/vector.cpp#L1812

joseph-isaacs: This seems to imply that they only valid the contents of valid rows

connortsui20:
decided to add specialization on the validity

vortex-duckdb/src/convert/vector.rs
```diff
@@ -221,33 +321,18 @@ pub fn flat_vector_to_vortex(vector: &mut Vector, len: usize) -> VortexResult<Ar
...
-                &mut vector.list_vector_get_child(),
-                usize::try_from(offsets[len - 1] + lengths[len - 1])
-                    .vortex_expect("list child data size must fit in usize"),
-            )?;
+            let validity = vector.validity_ref(len).to_validity();
+            let entries = vector.as_slice_with_len::<duckdb_list_entry>(len);
+
+            let (offsets, sizes, child_max_length) = process_duckdb_lists(entries, &validity);
+            let child_data =
+                flat_vector_to_vortex(&mut vector.list_vector_get_child(), child_max_length)?;
```
robert3005: doesn't this have to be the sum of the lengths not the max? You're exporting this many elements of children

connortsui20: Hmm I guess the name is a bit confusing? This is the maximum over all _ends_ of the list views into the child vector

connortsui20: Actually the correct name here is `child_min_length` which is the max over all view ends

robert3005 (approved):

PR #5748, opened 2025-12-16
fix: list contains

0ax1:
Fixes:
```
thread 'tokio-runtime-worker' panicked at vortex-array/src/expr/vtable.rs:484:13:
assertion `left == right` failed: Expression execution vortex.list.contains returned vector of length 1, but expected 12
  left: 1
````


gatesn (approved):

joseph-isaacs:
Can we add the scalar-scalar version tomorrow @0ax1

PR #5749, opened 2025-12-16
feat: build vortex-duckdb on windows

0ax1:


0ax1:
picked up at a later time, not priority right now

PR #5751, opened 2025-12-16
feat[fixed-size-list]: impl repeat and slice

joseph-isaacs:


vortex-vector/src/fixed_size_list/scalar.rs
```diff
@@ -84,9 +86,46 @@ impl ScalarOps for FixedSizeListScalar {
...
+                ))
+            };
+        }
+
+        let list_size = self.0.list_size();
+
+        // Get the elements from the scalar's inner length-1 vector and repeat them
+        // Clone the inner Vector from the Arc
+        let scalar_elements = self.0.elements();
+        let mut elements = scalar_elements.as_ref().clone().into_mut();
```
gatesn: Maybe we should deprecate into_mut and force try_into_mut? It's too easy to accidentally double copy

vortex-vector/src/fixed_size_list/scalar.rs
```diff
@@ -84,9 +86,46 @@ impl ScalarOps for FixedSizeListScalar {
...
+        elements.reserve((n - 1) * list_size as usize);
+
+        // Repeat the elements n-1 more times (we already have 1 copy)
+        for _ in 1..n {
+            elements.extend_from_vector(scalar_elements.as_ref());
+        }
+
+        // Build validity for n lists
+        let validity = MaskMut::new(n, self.is_valid());
```
gatesn: If is_valid, we should just append_n zeroes to the elements, faster than extend_from_vector

gatesn (changes requested):

gatesn (approved):

PR #5752, opened 2025-12-16
Feature: makes the decompressor and compressor methods public

connortsui20:
Completes: https://github.com/vortex-data/vortex/discussions/5709#discussioncomment-15263090


gatesn (approved):

encodings/fsst/src/array.rs
```diff
@@ -319,11 +319,12 @@ impl FSSTArray {
     /// this array.
     ///
     /// This is private to the crate to avoid leaking `fsst-rs` types as part of the public API.
```
robert3005, line 321: This comment is incorrect now

connortsui20: whoops

PR #5755, opened 2025-12-16
FSST Filter Kernel

gatesn:


encodings/fsst/src/kernel.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    fn execute_parent(
+        &self,
+        array: &FSSTArray,
+        parent: &FilterArray,
+        _child_idx: usize,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<Vector>> {
+        let mask_values = match parent.filter_mask() {
+            Mask::AllTrue(_) | Mask::AllFalse(_) => return Ok(None),
```
joseph-isaacs, line 62: should be log here, this should never be hit?

encodings/fsst/src/kernel.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        // We filter the uncompressed lengths
+        let uncompressed_lens = array
+            .uncompressed_lengths()
+            .filter(parent.filter_mask().clone())?
+            .execute(ctx)?
+            .into_primitive();
+
+        // Extract the filtered validity
+        let validity = array.codes().validity_mask().filter(parent.filter_mask());
```
joseph-isaacs: no validity_mask?

encodings/fsst/src/kernel.rs
gatesn: Not sure you can assume that you parent will be optimized away. That seems odd

joseph-isaacs: should all_false return none

joseph-isaacs (approved):

PR #5758, opened 2025-12-16
Fix: IPC `AsyncMessageReader` does not handle partial reads

connortsui20:
Fixes: https://github.com/vortex-data/vortex/discussions/5633

I doubt that the extra enums I added for readability will hurt performance in any meaningful way.


connortsui20:
actually I just realized this does not handle EOF in the middle of a partial read as an error

gatesn (dismissed):

connortsui20:
@gatesn can you take a look again I made some changes

robert3005 (approved):

PR #5761, opened 2025-12-16
Fix: hardcoded `DType` in dict layout

### LINKED ISSUE #5563: Fuzzing Crash: Sequential stream dtype mismatch in file_io
### Analysis

**Crash Location**: `vortex-layout/src/sequence.rs:313` (in `SequentialStreamAdapter::poll_next`)

**Error Message**:
```
assertion `left == right` failed: Sequential stream of u8 got chunk of u16.
  left: Primitive(U8, NonNullable)
 right: Primitive(U16, NonNullable)
```

**Stack Trace**:
```
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed::<&vortex_dtype::dtype::DType, &mut vortex_dtype::dtype::DType>
   4: poll_next<core::pin::Pin<alloc::boxed::Box<(dyn futures_core::stream::Stream<Item=core::result::Result<...>>>>>>
   5: poll_next<alloc::boxed::Box<(dyn vortex_layout::sequence::SequentialStream<...>)>>
   6: poll_next<core::pin::Pin<alloc::boxed::Box<(dyn vortex_layout::sequence::SequentialStream<...>)>>, ...>
   7: poll_next<futures_util::stream::stream::map::Map<...>>
```

**Root Cause**:

The fuzzer discovered a bug in the sequential stream processing where chunks of different dtypes are being produced by the same stream. The `SequentialStreamAdapter` validates that all chunks in a sequential stream have the same dtype (line 313-319 of `vortex-layout/src/sequence.rs`).

The crash occurs when:
1. A PrimitiveArray with dtype `Primitive(U8, Nullable)` is written to a file using the `file_io` fuzzer
2. The file uses the Compact compressor strategy
3. During streaming read operations, the sequential stream adapter expects all chunks to have the same dtype
4. However, the stream produces chunks with inconsistent dtypes: some chunks are `U8` and others are `U16`

This indicates a bug in either:
- The file writing logic that should ensure consistent dtypes across chunks
- The compression logic (Compact strategy) that may be changing dtypes during encoding
- The streaming read logic that may be incorrectly inferring or converting dtypes

**Note**: The assertion error message has the arguments reversed - it says "Sequential stream of u8 got chunk of u16" but the left/right values show it's actually expecting U8 but got U16 in the stream.

<details>
<summary>Debug Output</summary>

```
FuzzFileAction {
    array: PrimitiveArray {
        dtype: Primitive(
            U8,
            Nullable,
        ),
        buffer: Buffer<u8> {
            length: 57,
            alignment: Alignment(
                1,
            ),
            as_slice: [33, 34, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, ...],
        },
        validity: Array(
            BoolArray {
                dtype: Bool(
                    NonNullable,
                ),
                buffer: BitBuffer {
                    buffer: Buffer<u8> {
                        length: 8,
                        alignment: Alignment(
                            1,
                        ),
                        as_slice: [169, 255, 255, 255, 39, 213, 63, 0],
                    },
                    offset: 0,
                    len: 57,
                },
                validity: NonNullable,
                stats_set: ArrayStats {
                    inner: RwLock {
                        data: StatsSet {
                            values: [],
                        },
                    },
                },
            },
        ),
        stats_set: ArrayStats {
            inner: RwLock {
                data: StatsSet {
                    values: [],
                },
            },
        },
    },
    projection_expr: None,
    filter_expr: None,
    compressor_strategy: Compact,
}
```
</details>

### Summary

- **Target**: `file_io`
- **Crash File**: `crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2`
- **Branch**: develop
- **Commit**: c6ed47f68f630c4e578b179f91eb2ea3551a707d
- **Crash Artifact**: https://github.com/vortex-data/vortex/actions/runs/19742138914/artifacts/4699659701

### Reproduction

1. Download the crash artifact:
   - **Direct download**: https://github.com/vortex-data/vortex/actions/runs/19742138914/artifacts/4699659701
   - Or find `io-fuzzing-crash-artifacts` at: https://github.com/vortex-data/vortex/actions/runs/19742138914
   - Extract the zip file

2. Reproduce locally:
```bash
# The artifact contains file_io/crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2
cargo +nightly fuzz run --sanitizer=none file_io file_io/crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2
```

3. Get full backtrace:
```bash
RUST_BACKTRACE=full cargo +nightly fuzz run --sanitizer=none file_io file_io/crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2
```

---
*Auto-created by fuzzing workflow with Claude analysis*

### LINKED ISSUE #5591: Fuzzing Crash: Type mismatch in SequentialStreamAdapter (U8 vs U16)
### Analysis

**Crash Location**: `vortex-layout/src/sequence.rs:313:13` in `SequentialStreamAdapter::poll_next`

**Error Message**:
```
assertion `left == right` failed: Sequential stream of u8 got chunk of u16.
  left: Primitive(U8, NonNullable)
 right: Primitive(U16, NonNullable)
```

**Stack Trace**:
```
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed::<&vortex_dtype::dtype::DType, &mut vortex_dtype::dtype::DType>
   4: poll_next<core::pin::Pin<alloc::boxed::Box<(dyn futures_core::stream::Stream<Item=core::result::Result<...
   5: poll_next<alloc::boxed::Box<(dyn vortex_layout::sequence::SequentialStream<Item=core::result::Result<...
   6: poll_next<core::pin::Pin<alloc::boxed::Box<(dyn vortex_layout::sequence::SequentialStream<Item=core::result::Result<...
   7: poll_next<futures_util::stream::stream::map::Map<core::pin::Pin<alloc::boxed::Box<(dyn vortex_layout::sequence::SequentialStream<...
```

**Root Cause**: The `SequentialStreamAdapter` enforces that all chunks in a stream must have the same dtype. During file I/O fuzzing, a stream was created that produced chunks with inconsistent types: the stream expected `U8` but received a `U16` chunk. This indicates a bug in the file I/O pipeline where type consistency is not being properly maintained across stream chunks.

<details>
<summary>Debug Output</summary>

```
FuzzFileAction {
    array: PrimitiveArray {
        dtype: Primitive(
            U8,
            Nullable,
        ),
        buffer: Buffer<u8> {
            length: 57,
            alignment: Alignment(
                1,
            ),
            as_slice: [33, 34, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, 33, ...],
        },
        validity: Array(
            BoolArray {
                dtype: Bool(
                    NonNullable,
                ),
                buffer: BitBuffer {
                    buffer: Buffer<u8> {
                        length: 8,
                        alignment: Alignment(
                            1,
                        ),
                        as_slice: [169, 255, 255, 255, 39, 213, 63, 0],
                    },
                    offset: 0,
                    len: 57,
                },
                validity: NonNullable,
                stats_set: ArrayStats {
                    inner: RwLock {
                        data: StatsSet {
                            values: [],
                        },
                    },
                },
            },
        ),
        stats_set: ArrayStats {
            inner: RwLock {
                data: StatsSet {
                    values: [],
                },
            },
        },
    },
    projection_expr: None,
    filter_expr: None,
    compressor_strategy: Compact,
}
```
</details>

### Summary

- **Target**: `file_io`
- **Crash File**: `crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2`
- **Branch**: develop
- **Commit**: 91a62f35781371db6e06cdcc574c314c886cf126
- **Crash Artifact**: https://github.com/vortex-data/vortex/actions/runs/19798645935/artifacts/4716852651

### Reproduction

1. Download the crash artifact:
   - **Direct download**: https://github.com/vortex-data/vortex/actions/runs/19798645935/artifacts/4716852651
   - Or find `io-fuzzing-crash-artifacts` at: https://github.com/vortex-data/vortex/actions/runs/19798645935
   - Extract the zip file

2. Reproduce locally:
```bash
# The artifact contains file_io/crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2
cargo +nightly fuzz run --sanitizer=none file_io file_io/crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2
```

3. Get full backtrace:
```bash
RUST_BACKTRACE=full cargo +nightly fuzz run --sanitizer=none file_io file_io/crash-f34f3a6108ae41cb796a5e9d27f0f636135656f2
```

---
*Auto-created by fuzzing workflow with Claude analysis*

connortsui20:
Fixes https://github.com/vortex-data/vortex/issues/5591
Also fixes https://github.com/vortex-data/vortex/issues/5563

Still not super sure about why we limit the dict layout codes to be a max of `u16::MAX`...

### DIFF
--- a/vortex-layout/src/layouts/dict/writer.rs
+++ b/vortex-layout/src/layouts/dict/writer.rs
@@ -27,8 +27,10 @@ use vortex_array::builders::dict::DictEncoder;
 use vortex_array::builders::dict::dict_encoder;
 use vortex_btrblocks::BtrBlocksCompressor;
 use vortex_dtype::DType;
+use vortex_dtype::Nullability;
 use vortex_dtype::PType;
 use vortex_error::VortexError;
+use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_error::vortex_err;
 use vortex_io::kanal_ext::KanalExt;
@@ -61,9 +63,9 @@ pub struct DictLayoutConstraints {
     /// Maximum dictionary length. Limited to `u16` because dictionaries with more than 64k unique
     /// values provide diminishing compression returns given typical chunk sizes (~8k elements).
     ///
-    /// The codes dtype is chosen dynamically based on the actual dictionary size:
-    /// - [`PType::U8`] when the dictionary has at most 255 entries
-    /// - [`PType::U16`] when the dictionary has more than 255 entries
+    /// The codes dtype is determined upfront from this constraint:
+    /// - [`PType::U8`] when max_len <= 255
+    /// - [`PType::U16`] when max_len > 255
     pub max_len: u16,
 }
 
@@ -237,7 +239,11 @@ impl LayoutStrategy for DictStrategy {
 }
 
 enum DictionaryChunk {
-    Codes((SequenceId, ArrayRef)),
+    Codes {
+        seq_id: SequenceId,
+        codes: ArrayRef,
+        codes_ptype: PType,
+    },
     Values((SequenceId, ArrayRef)),
 }
 
@@ -299,26 +305,33 @@ impl DictStreamState {
             match self.encoder.take() {
                 None => match start_encoding(&self.constraints, &remaining) {
                     EncodingState::Continue((encoder, encoded)) => {
-                        res.push(labeler.codes(encoded));
+                        let ptype = encoder.codes_ptype();
+                        res.push(labeler.codes(encoded, ptype));
                         self.encoder = Some(encoder);
                     }
                     EncodingState::Done((values, encoded, unencoded)) => {
-                        res.push(labeler.codes(encoded));
+                        // Encoder was created and consumed within start_encoding
+                        let ptype = PType::try_from(encoded.dtype())
+                            .vortex_expect("codes should be primitive");
+                        res.push(labeler.codes(encoded, ptype));
                         res.push(labeler.values(values));
                         to_be_encoded = Some(unencoded);
                     }
                 },
-                Some(encoder) => match encode_chunk(encoder, &remaining) {
-                    EncodingState::Continue((encoder, encoded)) => {
-                        res.push(labeler.codes(encoded));
-                        self.encoder = Some(encoder);
+                Some(encoder) => {
+                    let ptype = encoder.codes_ptype();
+                    match encode_chunk(encoder, &remaining) {
+                        EncodingState::Continue((encoder, encoded)) => {
+                            res.push(labeler.codes(encoded, ptype));
+                            self.encoder = Some(encoder);
+                        }
+                        EncodingState::Done((values, encoded, unencoded)) => {
+                            res.push(labeler.codes(encoded, ptype));
+                            res.push(labeler.values(values));
+                            to_be_encoded = Some(unencoded);
+                        }
                     }
-                    EncodingState::Done((values, encoded, unencoded)) => {
-                        res.push(labeler.codes(encoded));
-                        res.push(labeler.values(values));
-                        to_be_encoded = Some(unencoded);
-                    }
-                },
+                }
             }
         }
         res
@@ -342,8 +355,12 @@ impl DictChunkLabeler {
         Self { sequence_pointer }
     }
 
-    fn codes(&mut self, chunk: ArrayRef) -> DictionaryChunk {
-        DictionaryChunk::Codes((self.sequence_pointer.advance(), chunk))
+    fn codes(&mut self, chunk: ArrayRef, ptype: PType) -> DictionaryChunk {
+        DictionaryChunk::Codes {
+            seq_id: self.sequence_pointer.advance(),
+            codes: chunk,
+            codes_ptype: ptype,
+        }
     }
 
     fn values(&mut self, chunk: ArrayRef) -> DictionaryChunk {
@@ -398,7 +415,11 @@ impl Stream for DictionaryTransformer {
             }
 
             match self.input.poll_next_unpin(cx) {
-                Poll::Ready(Some(Ok(DictionaryChunk::Codes((seq_id, codes))))) => {
+                Poll::Ready(Some(Ok(DictionaryChunk::Codes {
+                    seq_id,
+                    codes,
+                    codes_ptype,
+                }))) => {
                     if self.active_codes_tx.is_none() {
                         // Start a new group
                         let (codes_tx, codes_rx) = kanal::bounded_async::<SequencedChunk>(1);
@@ -407,7 +428,8 @@ impl Stream for DictionaryTransformer {
                         self.active_codes_tx = Some(codes_tx.clone());
                         self.active_values_tx = Some(values_tx);
 
-                        let codes_dtype = codes.dtype().clone();
+                        // Use passed codes_ptype instead of getting from array
+                        let codes_dtype = DType::Primitive(codes_ptype, Nullability::NonNullable);
 
                         // Send first codes.
                         self.pending_send =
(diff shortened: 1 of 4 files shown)


gatesn:
We decided that more than 64k unique elements in a DictLayout (where chunks can be ~8k elements), doesn't make much sense! I imagine this is done way with bad assumptions in a few places..

onursatici:
So this was done before the dict encoder was generic over the codes ptype, and was dynamically setting the codes dtype based on the magnitude of the max_len constraint. So it was always encoding to a wider int, then after the encoding is done we would downcast to a narrower int. In that world it is not possible to tell the codes ptype before encoding is fully finished.

also I don't think the codes ptype is selected dynamically, is should still be depending on the max_len argument, and if that is smaller than 256, only then we should get u8 codes:

https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/builders/dict/primitive.rs#L36-L42

I think what is happening here is that we are trying to dict encode u8 values, in that case this code would select u8, even though max_len is u16::max.

So probably the right fix is to expose the codes ptype from the dict builder?

connortsui20:
@onursatici I think that is what I am doing since I am extracting the dtype of the codes? Or are you saying that we need to forward the original dtype from somewhere else to the writer? Or are you saying we need to upcast the codes always?

onursatici:
I guess what I am saying is that this comment is not exactly right:

```
    /// The codes dtype is chosen dynamically based on the actual dictionary size:
    /// - [`PType::U8`] when the dictionary has at most 255 entries
    /// - [`PType::U16`] when the dictionary has more than 255 entries
```

the way we chose the ptype for codes is upfront, so it doesn't depend on the actual cardinality after encoding. It depends on two things, the width of the input primitive ptype, and the max_len constraint. Normally we have max_len set to u16::max because we don't want a higher cardinality in a chunk, so we always get u16 codes. In fuzzer we end up dict encoding u8 types, and because the input is narrower than the max_len constraint (u8 vs u16), dict encoder returns u8 codes.

for the fix, I think this will work and I am happy to merge as is, but I think the right way is to get the codes dtype from the dict encoder, because as soon as you construct one you can get a codes ptype, you don't need to encode a chunk to see what ptype it will end up having.

joseph-isaacs:
Shall we do this then @connortsui20 use the max_len instead of the first batch?

vortex-layout/src/layouts/dict/writer.rs
```diff
@@ -396,13 +407,17 @@ impl Stream for DictionaryTransformer {
                         self.active_codes_tx = Some(codes_tx.clone());
                         self.active_values_tx = Some(values_tx);
 
-                        // Send first codes
+                        let codes_dtype = codes.dtype().clone();
```
AdamGS: I agree with @onursatici and @joseph-isaacs, seems like we should take this from `DictLayoutConstraints::max_len` and not by the first batch.

onursatici: as discussed with Joe, I will make that change and push to this branch

vortex-layout/src/layouts/dict/writer.rs
```diff
@@ -49,10 +50,22 @@ use crate::sequence::SequentialStream;
 use crate::sequence::SequentialStreamAdapter;
 use crate::sequence::SequentialStreamExt;
 
+/// Constraints for dictionary layout encoding.
+///
+/// Note that [`max_len`](Self::max_len) is limited to `u16` (65,535 entries) by design. Since
+/// layout chunks are typically ~8k elements, having more than 64k unique values in a dictionary
+/// means dictionary encoding provides little compression benefit. If a column has very high
+/// cardinality, the fallback encoding strategy should be used instead.
```
joseph-isaacs, line 58: might be worth saying you can still use a DictArray?

joseph-isaacs (commented):
Shall we let the tests finish before we merge

joseph-isaacs (approved):

PR #5765, opened 2025-12-17
Map view types when converting arrow_expressions to vortex expressions

### LINKED ISSUE #5759: Arrow Expressions on Vortex Datasets raise ArrowNotImplementedError on string_views
### Discussed in https://github.com/vortex-data/vortex/discussions/5725

<sup>Originally posted by **paultiq** December 14, 2025</sup>
</div>

### Problem

A few facts conspire to be a problem for Vortex-DuckDB integration:

1. PyArrow does not implement string_view-string_view (in)equality.
2. When Vortex converts a PyArrow expression into a Vortex expression, it goes through substrait. Vortex provides the true schema (which has string_view fields) to pyarrow.compute.Expression.to_substrait.
3. pyarrow.compute.Expression.to_substrait will only produce a substrait expression if the PyArrow expression has an implementation in PyArrow.

Due to 1 and 3, PyArrow errors on equality involving a string_view column. I think @paultiq correctly identifies three options ("modify to_substrait", "implement string_view (in)equality in PyArrow", and "deceit"). There's another option which is to convert from Arrow expression directly to Vortex expression. I do not know how hard that is.

I think deceit is our best option here. We ignore the schema when we parse a Substrait expression (our column expressions are untyped, see substrait.py:94 in `field_reference`).

paultiq:
Fixes https://github.com/vortex-data/vortex/issues/5759

This PR maps string_views to strings and binary_views to binary so that `to_substrait` will no longer raise ArrowNotImplementedError when constructing the substrait. 

Vortex supports expressions over views and Arrow compute doesn't, but [to_substrait](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Expression.html#pyarrow.dataset.Expression.to_substrait) raises ArrowNotImplementedError based on Arrow compute kernels... regardless of the backend.


danking (approved):
This is great and we should merge! Can you follow up with a test that writes and reads through DuckDB a Vortex file containing strings? I think paultiq included one at one point in one of his GitHub issues.

paultiq:
@danking Do you want that test case in this PR or a separate? Happy to do either.

danking:
Separate is fine! Sorry I referred to you in the third person! GitHub mobile UI had me turned around haha

joseph-isaacs:
@paultiq think we can get this over the line

paultiq:
> @paultiq think we can get this over the line

AFAIK, this is ready to merge, was there something else for this PR?

joseph-isaacs:
The CI / Python lint is failing, we need to fix that

joseph-isaacs:
Why did you ignore all the warnings?

paultiq:
> Why did you ignore all the warnings?

The root reason is lack of type stubs for pyarrow, which is being addressed in https://github.com/apache/arrow/issues/32609

The ignores are consistent with the existing code: 
https://github.com/vortex-data/vortex/blob/41bb2bf14f1582d6a34a2493c03ee4de62f12a4a/vortex-python/python/vortex/arrow/expression.py#L32-L35

And existing test cases: 
https://github.com/vortex-data/vortex/blob/41bb2bf14f1582d6a34a2493c03ee4de62f12a4a/vortex-python/benchmark/test_aggregation.py#L23-L24

(I'll fix that DCO in a sec...)

paultiq:
The one question would be whether the direct test of `_schema_for_substrait` is OK (hence the reportPrivateUsage). 
reportPrivateUsage is used in a few places in the project, although not in any test cases, so that'd be a judgement call. 

test_expression.py tests both paths: the "private" _schema_for_substrait, as well as the "public" arrow_to_vortex path.

So, the choice would be either to allow the `ignore: reportPrivateUsage` or drop the direct test of _schema_for_substrait.

paultiq:
(sorry for the multiple DCO commits, just not in my normal habits)

The remaining DCO warning is for 3107e20814df357f36e08722eb88372eb6934550, @danking's commit.

danking:
Thank you @paultiq! Sorry for the friction in our process currently. I'll keep an eye on the pyarrow types issue, thanks for flagging.

PR #5770, opened 2025-12-17
deterministic encoding ordering

onursatici:
pre register all encodings known by the session, so they have the same encoding index from run to run. Layout encoding indices don't suffer from this because we get them after we have the full layout tree, and we traverse it dept first in a single thread to get the layout indices

### DIFF
--- a/docs/specs/file-format.md
+++ b/docs/specs/file-format.md
@@ -117,26 +117,3 @@ The plan is that at write-time, a minimum supported reader version is declared. 
 reader version can then be embedded into the file with WebAssembly decompression logic. Old readers are able to decompress new
 data (slower than native code, but still with SIMD acceleration) and read the file. New readers are able to make the best use of
 these encodings with native decompression logic and additional push-down compute functions (which also provides an incentive to upgrade).
-
-## File Determinism and Reproducibility
-
-### Encoding Order Indeterminism
-
-When writing Vortex files, each array segment references its encoding via an integer index into the footer's `array_specs`
-list. During serialization, encodings are registered in the order they are first encountered via calls to
-`ArrayContext::encoding_idx()`. With concurrent writes, this encounter order depends on thread scheduling and lock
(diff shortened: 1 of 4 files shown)


gatesn:
I think instead we should just pre-configure the writer with a list of permitted encodings. Any encoding not in the list should be forbidden (the compressor should respect it too). And we can pre-configure the full list if we want stable ordering?

The flags in this PR feel a bit specific

brancz:
I've plugged this into various mechanisms we have and I can no longer trigger non-determinism. So the result is what we desire, but I do find the flag a bit odd, is the saving really that great that this shouldn't just be how it always works?

gatesn:
Yeah that's also an option. Depends how far we go. Like in theory we could add say `geo-vortex` plugin with a dozen new encodings for geospatial data. So no we have a dozen more encoding IDs listed in the footer 🤷

brancz:
Can we not find all the actually used encodings first, deterministically write them to the footer and then write concurrently? That should be deterministic even in the event of eg. upgrades where the total possible set of encodings might change.

onursatici:
@brancz if the strategy ends up changing the input array encoding, like the btrblocks compressor we have, it is tricky to know what encodings they will end up having before being written

brancz:
Tactically, I'm happy with the approach in the PR (except I think I'd make it default, but also don't feel too strongly one way or another). Then, in the future, collect the encodings as they are being used in-memory and then persist at the end?

brancz:
For easier cherry-picking, could you squash into 1 commit?

onursatici:
Yea I agree, it would be a bit tricky to get it right though. We currently have a strict order on segments, so those would never change, but that means the only strict happens before relationship is between writing segments, which already have serialised buffers. These buffers of array data is serialised in parallel, and during this serialisation we lookup the encoding id from the array context. That lookup has to also have an ordering guarantee to get this work.

gatesn (approved):

PR #5778, opened 2025-12-17
perf[date-time-parts]: compressor handle constant and pushdown rules

joseph-isaacs:


encodings/datetime-parts/src/compute/rules.rs
```diff
@@ -0,0 +1,67 @@
...
+        Exact::from(&FilterVTable)
+    }
+
+    fn reduce_parent(
+        &self,
+        child: &DateTimePartsArray,
+        parent: &FilterArray,
+        child_idx: usize,
+    ) -> VortexResult<Option<ArrayRef>> {
+        if child_idx != 0 {
```
gatesn, line 41: Just ignore it. Filter only has one child

encodings/datetime-parts/src/compute/rules.rs
```diff
@@ -0,0 +1,67 @@
...
+            return Ok(None);
+        }
+
+        if !child.seconds().is::<ConstantVTable>() || !child.subseconds().is::<ConstantVTable>() {
+            return Ok(None);
+        }
+
+        DateTimePartsArray::try_new(
+            child.dtype().clone(),
+            FilterArray::new(child.days().clone(), parent.filter_mask().clone())
```
gatesn, line 51: Always use `child.days().filter(...)?` instead, since it runs optimization to push-down filtering more. I want to make `FilterArray::new` pub(crate) if not stricter.

vortex-duckdb/src/convert/scalar.rs
```diff
@@ -201,8 +201,11 @@ impl ToDuckDBScalar for ExtScalar<'_> {
                 _ => vortex_bail!("cannot have TimeUnit {unit}, so represent a day"),
             },
             TemporalMetadata::Timestamp(unit, tz) => {
-                if tz.is_some() {
-                    todo!("timezones to duckdb scalar")
+                if let Some(tz) = tz {
+                    if tz != "UTC" {
```
gatesn, line 205: I don't think you can do this. UTC != no timezone.

gatesn (changes requested):

vortex-duckdb/src/convert/scalar.rs
joseph-isaacs: This is how timestamp_tz is defined

joseph-isaacs: `//! Type used to represent TIMESTAMPTZ. timestamp_tz_t holds the microseconds since 1970-01-01 (UTC).
//! It is physically the same as timestamp_t, both hold microseconds since epoch.`

joseph-isaacs: https://github.com/duckdb/duckdb/blob/2cc24534a384551af3cf5508158a397e5af63810/src/include/duckdb/common/types/timestamp.hpp#L107

gatesn (approved):

PR #5779, opened 2025-12-17
Feature: Make `vortex-tui` a library

### LINKED ISSUE #5742: Expose Vortex TUI as a library crate
To support custom arrays and layouts, and in the absence of a stable ABI to load these via shared libraries, we should expose the Vortex TUI main function as a Rust library that accepts a Vortex session.

This allows users to easily compile their own TUI binaries with extensions built in.

Note I think we also exclude TUI from publishing atm.

connortsui20:
Closes: https://github.com/vortex-data/vortex/issues/5742

Also adds documentation everywhere and refactors a bunch of stuff to be more ergonomic since it's now "more" user facing.

### DIFF
--- a/vortex-tui/Cargo.toml
+++ b/vortex-tui/Cargo.toml
@@ -2,7 +2,7 @@
 name = "vortex-tui"
 authors = { workspace = true }
 categories = { workspace = true }
-description = "a small but might tool for working with Vortex files"
+description = "a small but mighty tool for working with Vortex files"
 edition = { workspace = true }
 homepage = { workspace = true }
 include = { workspace = true }
--- a/vortex-tui/src/lib.rs
+++ b/vortex-tui/src/lib.rs
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Vortex TUI library for interactively browsing and inspecting Vortex files.
+//!
+//! This crate provides both a CLI tool (`vx`) and a library API for working with Vortex files.
+//! Users can bring their own [`VortexSession`] to enable custom encodings and extensions.
+//!
+//! # Example
+//!
+//! ```ignore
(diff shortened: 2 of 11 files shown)


vortex-tui/src/main.rs
```diff
@@ -55,9 +47,6 @@ impl Commands {
     }
 }
 
-pub(crate) static SESSION: LazyLock<VortexSession> =
-    LazyLock::new(|| VortexSession::default().with_tokio());
-
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
```
gatesn, line 10: I think we basically want this to be:

```
vortex_tui::main(session).await
```

So move all this logic into the library too.

gatesn (changes requested):

vortex-tui/src/main.rs
AdamGS: nit - `vortex_tui::launch`?

vortex-tui/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "vortex-tui"
 authors = { workspace = true }
 categories = { workspace = true }
-description = "a small but might tool for working with Vortex files"
+description = "a small but mighty tool for working with Vortex files"
```
AdamGS, line 5: ❤️

vortex-tui/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+}
+
+/// Main entrypoint for `vx`.
+///
+/// # Errors
+///
+/// Raises any errors from subcommands.
+pub async fn launch() -> anyhow::Result<()> {
```
gatesn: You need to take session as an argument here!

connortsui20:
So since this is a new library crate we have to publish this. How do we get around the semver checks for just this specific run?

robert3005:
You have to manually override them once

gatesn (approved):

PR #5791, opened 2025-12-18
Dictionary execution

gatesn:


vortex-layout/src/layouts/flat/reader.rs
```diff
@@ -157,7 +147,7 @@ impl LayoutReader for FlatReader {
             }
 
             let array_mask = if *USE_VORTEX_OPERATORS {
-                if mask.density() < FILTER_OF_FILTER_THRESHOLD {
+                if mask.density() < EXPR_EVAL_THRESHOLD {
```
gatesn, line 150: Seemed odd that we would use a different threshold to develop. Let's try to repro the performance and improve later.

0ax1 (approved):
lgtm

PR #5792, opened 2025-12-18
perf: list contains scalar scalar

0ax1:


vortex-array/src/expr/exprs/list_contains.rs
```diff
@@ -330,6 +335,31 @@ fn constant_list_scalar_contains(list: ListViewScalar, values: Vector) -> Vortex
     Ok(result)
 }
 
+/// Used when both needle and list are scalars.
+fn list_contains_scalar_scalar(
+    list: &ListViewScalar,
+    needle: &vortex_vector::Scalar,
```
joseph-isaacs, line 332: should this method be in vortex-compute?

joseph-isaacs: can we move this to vortex-compute with a todo to add a list_contains trait like other compute fns and therefore we can specialise further

vortex-array/src/expr/exprs/list_contains.rs
```diff
@@ -330,6 +335,31 @@ fn constant_list_scalar_contains(list: ListViewScalar, values: Vector) -> Vortex
...
+            let slice: &[_] = pvec.as_ref();
+            let validity = pvec.validity();
+            slice
+                .iter()
+                .enumerate()
+                .any(|(i, &elem)| needle == &PScalar::new(Some(elem)).into() && validity.value(i))
+        })
+    } else {
+        // Fallback for non-primitive vectors
+        (0..elements.len()).any(|i| needle == &elements.scalar_at(i))
```
joseph-isaacs: can we use a array compare here with the elements and a constant array?

joseph-isaacs: shouldn't need to special case anything

vortex-array/src/expr/exprs/list_contains.rs
0ax1: For now can keep this consistent with the other `contain` functions.

vortex-array/src/expr/exprs/list_contains.rs
0ax1: Did that, and added a comment what the tradeoffs are perf wise.

vortex-array/src/expr/exprs/list_contains.rs
```diff
@@ -128,30 +127,34 @@ impl VTable for ListContains {
             .try_into()
             .map_err(|_| vortex_err!("Wrong number of arguments for ListContains expression"))?;
 
-        let matches = match (lhs.as_scalar().is_some(), rhs.as_scalar().is_some()) {
+        match (lhs.as_scalar().is_some(), rhs.as_scalar().is_some()) {
```
joseph-isaacs: shall we match on datum directly

vortex-array/src/expr/exprs/list_contains.rs
```diff
@@ -330,6 +333,35 @@ fn constant_list_scalar_contains(list: ListViewScalar, values: Vector) -> Vortex
...
+    // containment. `execute` allocates the returned vector on the heap. Further, the `eq`
+    // comparison does not short-circuit on the first match found.
+    let found = Binary
+        .bind(operators::Operator::Eq)
+        .execute(ExecutionArgs {
+            datums: vec![
+                Datum::Vector(elements.deref().clone()),
+                Datum::Scalar(needle.clone()),
+            ],
+            dtypes: vec![],
```
joseph-isaacs, line 346: do we not need dtypes?

0ax1: In my understanding they can be derived here.

vortex-array/src/expr/exprs/list_contains.rs
0ax1: done

joseph-isaacs (approved):

PR #5793, opened 2025-12-18
perf[dtp]: push down compare over dtp constant second subsecond

joseph-isaacs:
If compare(DTP) operators only on the day, you can replace with a compare over the day.

### DIFF
--- a/encodings/datetime-parts/src/compute/rules.rs
+++ b/encodings/datetime-parts/src/compute/rules.rs
@@ -100,10 +100,9 @@ impl ArrayParentReduceRule<DateTimePartsVTable> for DTPComparisonPushDownRule {
         child_idx: usize,
     ) -> VortexResult<Option<ArrayRef>> {
         // Only handle comparison operations (Binary comparisons or Between)
-        if !parent
+        if parent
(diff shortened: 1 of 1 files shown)


encodings/datetime-parts/src/compute/rules.rs
```diff
@@ -61,3 +72,296 @@ impl ArrayParentReduceRule<DateTimePartsVTable> for DTPFilterPushDownRule {
...
+            ScalarFnArray::try_new(parent.scalar_fn().clone(), new_children, parent.len())?
+                .into_array()
+                .optimize()?;
+
+        Ok(Some(result))
+    }
+}
+
+/// Check if the scalar function is a comparison operation we can push down.
+fn is_comparison_op(parent: &ScalarFnArray) -> bool {
```
gatesn: Blergh. Split out Binary expressions?!

gatesn (approved):

encodings/datetime-parts/src/compute/rules.rs
joseph-isaacs: Have a different rule for binary?

PR #5795, opened 2025-12-18
perf[sequence]: array constant compare execution kernel

joseph-isaacs:


encodings/sequence/Cargo.toml
```diff
@@ -22,13 +22,17 @@ vortex-dtype = { workspace = true }
...
 vortex-proto = { workspace = true }
+vortex-runend = { workspace = true }
 vortex-scalar = { workspace = true }
 vortex-vector = { workspace = true }
 
 [dev-dependencies]
 itertools = { workspace = true }
 rstest = { workspace = true }
 vortex-array = { path = "../../vortex-array", features = ["test-harness"] }
+vortex-file = { path = "../../vortex-file", features = ["tokio"] }
```
0ax1, line 33: These should probably all be: ~~`workspace = true`~~ relative paths.

encodings/sequence/src/kernel.rs
```diff
@@ -0,0 +1,536 @@
...
+    // For NotEq: match_val=false, default_val=true
+    let match_val = !negate;
+    let not_match_val = negate;
+
+    if let Some(idx) = set_idx {
+        // Single match_val at idx, rest is default_val
+        let idx = idx as u64;
+        let len = array.length as u64;
+        let (ends, values) = if idx == 0 {
+            let ends = buffer![1u64, len].into_array();
```
0ax1: What happens if length is zero? Or 1. Will this panic if the ends are not ascending?

encodings/sequence/src/kernel.rs
```diff
@@ -0,0 +1,536 @@
...
+/// When `negate` is false, returns true where sequence == constant.
+/// When `negate` is true, returns true where sequence != constant.
+fn compare_eq_neq(
+    array: &SequenceArray,
+    constant: PValue,
+    nullability: Nullability,
+    negate: bool,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<Option<Vector>> {
+    // Check if there exists an integer solution to const = base + idx * multiplier
```
0ax1: Early return on empty array?

joseph-isaacs: We will never the the kernel if the array is empty

encodings/sequence/src/kernel.rs
joseph-isaacs: True

0ax1 (approved):

PR #5796, opened 2025-12-18
Fix: support widening of `ListViewVector` offsets and sizes

connortsui20:
Since the list view "logical" type does not include the sizes and offsets as part of its type, we should allow resizing.

### DIFF
--- a/vortex-dtype/src/ptype.rs
+++ b/vortex-dtype/src/ptype.rs
@@ -767,6 +767,56 @@ impl PType {
             }
         }
     }
+
+    /// Returns the minimum unsigned integer [`PType`] that can represent the given value.
+    #[inline]
+    pub const fn min_unsigned_ptype_for_value(value: u64) -> Self {
+        if value <= u8::MAX as u64 {
+            Self::U8
+        } else if value <= u16::MAX as u64 {
+            Self::U16
+        } else if value <= u32::MAX as u64 {
+            Self::U32
+        } else {
+            Self::U64
+        }
+    }
+
+    /// Returns the minimum signed integer [`PType`] that can represent the given value.
+    #[inline]
+    pub const fn min_signed_ptype_for_value(value: i64) -> Self {
+        if value >= i8::MIN as i64 && value <= i8::MAX as i64 {
+            Self::I8
+        } else if value >= i16::MIN as i64 && value <= i16::MAX as i64 {
+            Self::I16
+        } else if value >= i32::MIN as i64 && value <= i32::MAX as i64 {
+            Self::I32
+        } else {
+            Self::I64
+        }
+    }
+
+    /// Returns the wider of two unsigned integer [`PType`]s based on byte width.
+    #[inline]
+    pub const fn max_unsigned_ptype(self, other: Self) -> Self {
+        debug_assert!(self.is_unsigned_int() && other.is_unsigned_int());
+        if self.byte_width() >= other.byte_width() {
+            self
+        } else {
+            other
+        }
+    }
+
+    /// Returns the wider of two signed integer [`PType`]s based on byte width.
+    #[inline]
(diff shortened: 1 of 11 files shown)


vortex-vector/src/primitive/cast.rs
joseph-isaacs, line 1: is annoying if here, but ye

vortex-dtype/src/ptype.rs
```diff
@@ -767,6 +767,56 @@ impl PType {
...
+        } else if value <= u32::MAX as u64 {
+            Self::U32
+        } else {
+            Self::U64
+        }
+    }
+
+    /// Returns the minimum signed integer [`PType`] that can represent the given value.
+    #[inline]
+    pub const fn min_signed_ptype_for_value(value: i64) -> Self {
```
joseph-isaacs, line 787: ? unused

vortex-vector/src/listview/vector_mut.rs
```diff
@@ -300,27 +301,23 @@ impl VectorMutOps for ListViewVectorMut {
         self.len = self.validity.len();
     }
 
-    /// This will also panic if we try to extend the `ListViewVector` beyond the maximum offset
-    /// representable by the type of the `offsets` primitive vector.
     fn extend_from_vector(&mut self, other: &ListViewVector) {
```
joseph-isaacs, line 309: I wonder if we want to be able to specify the max size to not cast from u8 -> u16 -> u32...

joseph-isaacs (approved):

vortex-vector/src/listview/vector_mut.rs
gatesn: We should do that in with_capacity

vortex-dtype/src/ptype.rs
connortsui20: I just thought it was weird that we would only have one for unsigned

connortsui20:
@joseph-isaacs I made a small change

vortex-vector/src/listview/vector_mut.rs
joseph-isaacs: How we don't know the offset of size max value? We can only guess based on avg list len

vortex-vector/src/listview/vector_mut.rs
```diff
@@ -185,11 +186,16 @@ impl ListViewVectorMut {
 
     /// Creates a new [`ListViewVectorMut`] with the specified capacity.
     pub fn with_capacity(element_dtype: &DType, capacity: usize) -> Self {
+        let offsets_ptype = PType::min_unsigned_ptype_for_value(capacity as u64);
```
joseph-isaacs: You need to multiple but a guess of each list size?

vortex-vector/src/listview/vector_mut.rs
connortsui20: Well we do know the max size, it cannot be larger than the number of elements (same as offset)

vortex-vector/src/listview/vector_mut.rs
joseph-isaacs: maybe you want another ctor with total list len?

connortsui20: Oh whoops I forgot capacity is different from the elements length

connortsui20: I did the same as our existing list view array builder and take it as a parameter. The default for the elements capacity is then 2 * capacity

vortex-vector/src/listview/vector_mut.rs
connortsui20: oops nvm I forgot capacity and elements capacity are not the same

PR #5799, opened 2025-12-18
feat[array]: stats + validity in display array

joseph-isaacs:


connortsui20 (approved):
LGTM

Looks like some of the display tests need updating though

vortex-array/src/display/tree.rs
```diff
@@ -39,13 +39,25 @@ impl<'a, 'b: 'a> TreeFormatter<'a, 'b> {
         } else {
             100_f64 * nbytes as f64 / total_size as f64
         };
+        // Compute validity status string
+        let validity_str = if !array.dtype().is_nullable() {
+            ""
+        } else if array.all_valid() {
```
gatesn: We can't do this - in the new world it would require execution

gatesn (changes requested):

vortex-array/src/display/tree.rs
joseph-isaacs: what do you want instead?

joseph-isaacs: I choose this since it would easily be stats based

gatesn (approved):

PR #5814, opened 2025-12-23
performance[vortex-array]: don't call is_valid to count bytes in varbinview

asubiotto:
I noticed that the better blocks compressor uses `count_referenced_bytes` which calls `is_valid` on each view and results in an expensive `scalar_at` call. This was 30% of our system-wide CPU usage over 12 hours (see PR for attached screenshot).

This commit iterates over the nullability mask directly to avoid these expensive calls.

image

### DIFF
--- a/vortex-array/src/arrays/varbinview/compact.rs
+++ b/vortex-array/src/arrays/varbinview/compact.rs
@@ -8,12 +8,12 @@ use std::ops::Range;
 
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
+use vortex_mask::Mask;
+use vortex_vector::binaryview::Ref;
 
 use crate::arrays::VarBinViewArray;
 use crate::builders::ArrayBuilder;
 use crate::builders::VarBinViewBuilder;
-use crate::validity::Validity;
-use crate::vtable::ValidityHelper;
 
 impl VarBinViewArray {
     /// Returns a compacted copy of the input array, where all wasted space has been cleaned up. This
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/varbinview/compact.rs
```diff
@@ -60,23 +60,51 @@ impl VarBinViewArray {
...
+                                view.len() as u64
+                            }
+                        })
+                        .sum(),
+                }
+            }
         }
     }
 
     pub(crate) fn buffer_utilizations(&self) -> Vec<BufferUtilization> {
```
asubiotto, line 92: I noticed the same pattern here, so thought I might as well fix it here too.

vortex-array/src/arrays/varbinview/compact.rs
```diff
@@ -60,23 +60,51 @@ impl VarBinViewArray {
...
                 })
                 .sum(),
+            Validity::Array(_) => {
+                let mask = self.validity().to_mask(self.len());
+                match mask.bit_buffer() {
+                    vortex_mask::AllOr::All => self
+                        .views()
+                        .iter()
+                        .map(|&view| {
+                            if view.is_inlined() {
```
asubiotto: There's some code repetition here and below. I can probably use an always inlined function instead if that's desirable.

AdamGS:
compression slowdown is interesting, I'll a look but overall seems like a win 🥳

AdamGS (approved):
LGTM

vortex-array/src/arrays/varbinview/compact.rs
```diff
@@ -85,17 +113,41 @@ impl VarBinViewArray {
...
 
-        if matches!(self.validity(), Validity::AllInvalid) {
-            return utilizations;
-        }
-
-        for (idx, &view) in self.views().iter().enumerate() {
-            if !self.is_valid(idx) || view.is_inlined() {
-                continue;
+        match self.validity() {
+            Validity::AllInvalid => {}
```
AdamGS: nit - maybe `if matches!`?

vortex-array/src/arrays/varbinview/compact.rs
```diff
@@ -60,23 +60,51 @@ impl VarBinViewArray {
...
                         0u64
                     } else {
                         view.len() as u64
                     }
                 })
                 .sum(),
+            Validity::Array(_) => {
+                let mask = self.validity().to_mask(self.len());
+                match mask.bit_buffer() {
+                    vortex_mask::AllOr::All => self
```
AdamGS: nit - import `AllOr` here

asubiotto:
It seems like it's due to now including `runend_encode`. This is used in the decision to compress, so it might be compressing when previously it wasn't? I'll double check since the behavior shouldn't have changed.

vortex-array/src/arrays/varbinview/compact.rs
asubiotto: Refactored

vortex-array/src/arrays/varbinview/compact.rs
asubiotto: Refactored

vortex-array/src/arrays/varbinview/compact.rs
asubiotto: I prefer an explicit match because you'll need an arm anyways. I simplified this code to use the mask directly anyways.

vortex-array/src/arrays/varbinview/compact.rs
asubiotto: Changed to use the validity mask directly, which is much simpler.

AdamGS:
Looking at a few codspeed diffs, seems like it can't figure out the diff correctly, so it thinks all the traces are new

asubiotto:
> Looking at a few codspeed diffs, seems like it can't figure out the diff correctly, so it thinks all the traces are new

Yes, I also think this is codspeed. I don't think the runend benchmarks (that the regression reported) use varbinviews. Also now codspeed seems to be happy again.

robert3005:
I am very confused how VarBinViewVTable::is_valid ends up calling scalar at, is it the case that you have runend compressed validity?

asubiotto:
`is_valid` seems to always call `scalar_at`: https://github.com/vortex-data/vortex/blob/31026ce588620a8276a5daf25cf0e0c79d437350/vortex-array/src/validity.rs#L135

But it does seem like the expensive part is specifically `scalar_at` on a `RunEndVTable` which implies validity is REE encoded. We use the default write strategy.

robert3005:
Ah, it was me not reading through the code deep enough.

gatesn:
Yeah this is why I want to make it really really obvious with the operators work what is "executing", vs operating on metadata only. I think even deprecating these scalar validity accessor functions, and instead just forcing the user to execute the array validity. (Still working on a name for "execute"!)

Great catch though

PR #5817, opened 2025-12-25
[WIP] Chore: leaky decimal fixes not implemented

connortsui20:
DO NOT MERGE YET

I just had these changes staged on my laptop so I don't want to lose them.

At some point we need to create an issue for the decimal logical and physical type separation issue (maybe I will get around to that soon).

Part of these changes should also fix all of the fuzzer issues being titled incorrectly (a bunch of null fuzzer issues that are really cast errors).


connortsui20:
related: https://github.com/vortex-data/vortex/issues/5820

Issue #1588, opened 2024-12-06
Add support for generating arbitrary Extension arrays to the fuzzer


robert3005:

connortsui20:
I don't think this is crazy difficult to add with a little help.

If you would like to contribute, please let us know if you are interested in this!

robert3005:
The difficult part is having to know what’s the valid domain of storage type for the extension type. Requires extension types to have validation function for a value

connortsui20:
I think the fuzzer implementation of the extensions arrays can easily "plug in" what a valid domain is and then the fuzzer can always check that the values stay in the domain

Pratham1812:
Hey I would like to pick this up.

Pratham1812:
Hi I have started working on this issue, while coding I found out there are places where the the code for extension arrays has not been added like for instance here-
`fuzz/src/array/sort.rs`
Can I get some insights? should I raise a draft PR. Like currently there are many instances of unreachable code which states that fuzzing is not supported for extension arrays.

connortsui20:
@Pratham1812 I would classify those things as part of this issue: implement as much as possible so that we can add extension types to the fuzzer in a meaningful way. Note that does not mean implement every possible thing (for example, we do not implement sort for the list types in the fuzzer or as compute functions), but we do need that for the other "primitive" types.

connortsui20:
I'll also mention that I set this as a good first issue not because it is a quick or easy change, but because it introduces you to a good percentage of the codebase and requires adding a small amount of code to lots of places (essentially a lot of pattern matching). So please don't hesitate to ask questions!

_Also, a lot of us are on holiday until next year so we might only be able to help you next year_

Pratham1812:
Yeah this issue made me dive deep into the code. Would ask more questions if I face issues. I will try not to disturb you guys during the holidays.

connortsui20:
@Pratham1812 feel free to ask questions! This project is open source, so it's not like we require people to ask questions during specific hours 😅.

Also, if you find something is challenging or difficult to understand, it would be really helpful if you tell us because it is likely something we can improve (in terms of documentation or code organization). Thanks for taking a look at this!

PR #5819, opened 2025-12-26
Feature: introduces fuzzing with extension arrays

Pratham1812:
This PR introduces fuzzing in extension arrays solving #1588 . I have tried running the fuzzer locally and have tried to solve the possible edge cases I could identify.

### DIFF
--- a/fuzz/src/array/compare.rs
+++ b/fuzz/src/array/compare.rs
@@ -123,8 +123,16 @@ pub fn compare_canonical_array(array: &dyn Array, value: &Scalar, operator: Oper
             )
             .into_array()
         }
-        d @ (DType::Null | DType::Extension(_)) => {
-            unreachable!("DType {d} not supported for fuzzing")
+        DType::Null => {
+            unreachable!("DType null not supported for fuzzing")
+        }
+        DType::Extension(..) => {
+            // Extension arrays delegate comparison to their storage type
+            compare_canonical_array(
+                array.to_extension().storage(),
+                &value.as_extension().storage(),
+                operator,
+            )
         }
     }
 }
--- a/fuzz/src/array/fill_null.rs
+++ b/fuzz/src/array/fill_null.rs
@@ -1,8 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
-
-use std::sync::Arc;
-
 use vortex_array::ArrayRef;
 use vortex_array::Canonical;
 use vortex_array::IntoArray;
@@ -19,13 +16,14 @@ use vortex_array::vtable::ValidityHelper;
 use vortex_buffer::Buffer;
 use vortex_buffer::BufferMut;
 use vortex_dtype::DType;
-use vortex_dtype::ExtDType;
 use vortex_dtype::Nullability;
 use vortex_dtype::match_each_decimal_value_type;
 use vortex_dtype::match_each_native_ptype;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_scalar::Scalar;
+
+use crate::array::clone_ext_dtype;
 /// Apply fill_null on the canonical form of the array to get a consistent baseline.
 /// This implementation manually fills null values for each canonical type
 /// without using the fill_null method, to serve as an independent baseline for testing.
@@ -43,10 +41,10 @@ pub fn fill_null_canonical_array(
         Canonical::VarBinView(array) => {
             fill_varbinview_array(&array, fill_value, result_nullability)
         }
-        Canonical::Extension(array) => fill_extension_array(&array, fill_value),
         Canonical::Struct(_) | Canonical::List(_) | Canonical::FixedSizeList(_) => {
             fill_null(canonical.as_ref(), fill_value)?
         }
+        Canonical::Extension(array) => fill_extension_array(&array, fill_value),
     })
 }
 
@@ -60,13 +58,15 @@ fn fill_extension_array(array: &ExtensionArray, fill_value: &Scalar) -> ArrayRef
     if filled_storage.dtype().nullability() == array.ext_dtype().storage_dtype().nullability() {
         ExtensionArray::new(array.ext_dtype().clone(), filled_storage).into_array()
     } else {
-        let new_ext_dtype = Arc::new(ExtDType::new(
-            array.ext_dtype().id().clone(),
-            Arc::new(filled_storage.dtype().clone()),
-            array.ext_dtype().metadata().cloned(),
-        ));
-
-        ExtensionArray::new(new_ext_dtype, filled_storage).into_array()
+        ExtensionArray::new(
+            clone_ext_dtype(
+                array.ext_dtype().id().clone(),
+                filled_storage.dtype().clone(),
+                array.ext_dtype().metadata().cloned(),
+            ),
+            filled_storage,
+        )
+        .into_array()
     }
 }
 
--- a/fuzz/src/array/fill_null.rs
+++ b/fuzz/src/array/fill_null.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::sync::Arc;
+
 use vortex_array::ArrayRef;
 use vortex_array::Canonical;
 use vortex_array::IntoArray;
@@ -8,6 +10,7 @@ use vortex_array::ToCanonical;
 use vortex_array::arrays::BoolArray;
 use vortex_array::arrays::ConstantArray;
 use vortex_array::arrays::DecimalArray;
+use vortex_array::arrays::ExtensionArray;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::arrays::VarBinViewArray;
 use vortex_array::compute::fill_null;
@@ -16,13 +19,13 @@ use vortex_array::vtable::ValidityHelper;
 use vortex_buffer::Buffer;
 use vortex_buffer::BufferMut;
 use vortex_dtype::DType;
+use vortex_dtype::ExtDType;
 use vortex_dtype::Nullability;
 use vortex_dtype::match_each_decimal_value_type;
 use vortex_dtype::match_each_native_ptype;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_scalar::Scalar;
-
 /// Apply fill_null on the canonical form of the array to get a consistent baseline.
 /// This implementation manually fills null values for each canonical type
 /// without using the fill_null method, to serve as an independent baseline for testing.
@@ -40,11 +43,31 @@ pub fn fill_null_canonical_array(
         Canonical::VarBinView(array) => {
             fill_varbinview_array(&array, fill_value, result_nullability)
         }
-        Canonical::Struct(_)
-        | Canonical::List(_)
-        | Canonical::FixedSizeList(_)
-        | Canonical::Extension(_) => fill_null(canonical.as_ref(), fill_value)?,
+        Canonical::Extension(array) => fill_extension_array(&array, fill_value),
+        Canonical::Struct(_) | Canonical::List(_) | Canonical::FixedSizeList(_) => {
+            fill_null(canonical.as_ref(), fill_value)?
+        }
     })
+}
+
+fn fill_extension_array(array: &ExtensionArray, fill_value: &Scalar) -> ArrayRef {
+    let filled_storage = fill_null_canonical_array(
+        array.storage().to_canonical(),
+        &fill_value.as_extension().storage(),
+    )
+    .vortex_expect("fill_null should succeed in canonical form");
+
+    if filled_storage.dtype().nullability() == array.ext_dtype().storage_dtype().nullability() {
+        ExtensionArray::new(array.ext_dtype().clone(), filled_storage).into_array()
+    } else {
+        let new_ext_dtype = Arc::new(ExtDType::new(
+            array.ext_dtype().id().clone(),
+            Arc::new(filled_storage.dtype().clone()),
+            array.ext_dtype().metadata().cloned(),
+        ));
+
+        ExtensionArray::new(new_ext_dtype, filled_storage).into_array()
+    }
 }
 
 fn fill_bool_array(
--- a/fuzz/src/array/filter.rs
+++ b/fuzz/src/array/filter.rs
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use std::sync::Arc;
-
 use vortex_array::Array;
 use vortex_array::ArrayRef;
 use vortex_array::IntoArray;
@@ -18,11 +16,11 @@ use vortex_array::validity::Validity;
 use vortex_buffer::BitBuffer;
 use vortex_buffer::Buffer;
 use vortex_dtype::DType;
-use vortex_dtype::ExtDType;
 use vortex_dtype::match_each_decimal_value_type;
 use vortex_dtype::match_each_native_ptype;
 use vortex_error::VortexResult;
 
+use crate::array::clone_ext_dtype;
 use crate::array::take_canonical_array_non_nullable_indices;
 
 pub fn filter_canonical_array(array: &dyn Array, filter: &[bool]) -> VortexResult<ArrayRef> {
@@ -120,18 +118,22 @@ pub fn filter_canonical_array(array: &dyn Array, filter: &[bool]) -> VortexResul
             take_canonical_array_non_nullable_indices(array, indices.as_slice())
         }
         DType::Extension(ext_dtype) => {
-            // Extension arrays delegate filter to their storage type
+            // Extension arrays delegate filter to their storage type.
             let filtered_storage = filter_canonical_array(array.to_extension().storage(), filter)?;
 
             if filtered_storage.dtype().nullability() == ext_dtype.storage_dtype().nullability() {
                 Ok(ExtensionArray::new(ext_dtype.clone(), filtered_storage).into_array())
             } else {
-                let new_ext_dtype = Arc::new(ExtDType::new(
-                    ext_dtype.id().clone(),
-                    Arc::new(filtered_storage.dtype().clone()),
-                    ext_dtype.metadata().cloned(),
-                ));
-                Ok(ExtensionArray::new(new_ext_dtype, filtered_storage).into_array())
+                // The storage dtype changed (i.e., became nullable due to filtering).
+                Ok(ExtensionArray::new(
+                    clone_ext_dtype(
+                        ext_dtype.id().clone(),
+                        filtered_storage.dtype().clone(),
+                        ext_dtype.metadata().cloned(),
+                    ),
+                    filtered_storage,
+                )
+                .into_array())
             }
         }
         DType::Null => {
--- a/fuzz/src/array/filter.rs
+++ b/fuzz/src/array/filter.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::sync::Arc;
+
 use vortex_array::Array;
 use vortex_array::ArrayRef;
 use vortex_array::IntoArray;
@@ -8,6 +10,7 @@ use vortex_array::ToCanonical;
 use vortex_array::accessor::ArrayAccessor;
 use vortex_array::arrays::BoolArray;
 use vortex_array::arrays::DecimalArray;
+use vortex_array::arrays::ExtensionArray;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::arrays::StructArray;
 use vortex_array::arrays::VarBinViewArray;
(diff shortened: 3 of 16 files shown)


Pratham1812:
@connortsui20 Please let me know your comments, I will try to resolve them at the earliest. 
Thanks.

connortsui20:
Running the fuzzer on this branch: https://github.com/vortex-data/vortex/actions/runs/20531444854

connortsui20 (changes requested):
Thanks for the PR, this is a definitely good start!

The main concern here is what @robert3005 said in the original issue:

> The difficult part is having to know what’s the valid domain of storage type for the extension type. Requires extension types to have validation function for a value

Basically, in addition to testing operations over the different kinds of extension arrays, we also want to validate that the array values are always in the domain of the given type. For example, we want to test that the temporal arrays like `vortex.date` or `vortex.time` and also make sure that their values do not overflow when calling `.to_jiff()`.

I was thinking of something along the lines of this (note this is claude-assisted code):

```rust
type DomainValidator = Box<dyn Fn(&Scalar) -> bool + Send + Sync>;

fn validator_for_ext_type(ext_dtype: &ExtDType) -> DomainValidator {
    if is_temporal_ext_type(ext_dtype.id()) {
        let metadata = TemporalMetadata::try_from(ext_dtype).unwrap();
        Box::new(move |scalar| {
            scalar.is_null()
                || metadata
                    .to_jiff(scalar.as_extension().storage().as_i64().unwrap())
                    .is_ok()
        })
    } else if {
        // Logic for handling extension types that are specific to the fuzzer,
        // for example maybe an extension type that only accepts even i32 integers?
    } else {
        // Unknown extension type, so just accept everything?
        Box::new(|_| true)
    }
}
```

And then the array generation becomes easy:

```rust
fn random_extension(
    u: &mut Unstructured,
    ext_dtype: &Arc<ExtDType>,
    chunk_len: Option<usize>,
) -> Result<(ArrayRef, DomainValidator)> {
    let validator = validator_for_ext_type(ext_dtype);
    let array_length = chunk_len.unwrap_or(u.int_in_range(0..=20)?);

    let mut builder = ExtensionBuilder::with_capacity(ext_dtype.clone(), array_length);

    for _ in 0..array_length {
        // Generate storage values, retry if invalid.
        // Just make sure the selectivity isn't too small otherwise this will take forever...
        loop {
            let storage_scalar = random_scalar(u, ext_dtype.storage_dtype())?;
            let ext_scalar = Scalar::extension(ext_dtype.clone(), storage_scalar);

            if validator.validate_scalar(&ext_scalar) {
                builder.append_scalar(&ext_scalar).unwrap();
                break;
            }
            // else: retry with new random value.
        }
    }

    Ok((builder.finish(), validator))
}
```

You would probably need to also add the `DomainValidator` type as an optional field to `FuzzArrayAction` and maybe some other places so the fuzzer can access the validation closure by calling the validation function at the end of `run_fuzz_action`.

---

I had Claude tell me a rough "plan of attack" that might be helpful. Obviously it might be missing stuff, but it should be a very good starting point. Please let us know if you have any questions!

```
  ### Goal

  Add a validation closure that is generated alongside each extension array and called after
  every fuzz operation to verify domain invariants are preserved.

  ### Implementation

  1. **Define the validator type** (in `fuzz/src/array/mod.rs` or similar):
     - `pub type DomainValidator = Arc<dyn Fn(&dyn Array) -> bool + Send + Sync>;`

  2. **Add validator to `FuzzArrayAction`**:
     - Add field `pub validator: Option<DomainValidator>` to the struct

  3. **Create `validator_for_ext_type(ext_dtype: &ExtDType) -> DomainValidator`**:
     - For temporal types (`vortex.date`, `vortex.time`, `vortex.timestamp`): validate that
  `TemporalMetadata::to_jiff()` succeeds for all non-null values
     - For unknown extension types: return a validator that accepts everything

  4. **Modify `random_extension()` in `vortex-array/src/arrays/arbitrary.rs`**:
     - Get validator via `validator_for_ext_type()`
     - When generating values, retry if `!validator.validate_scalar(&scalar)`
     - Return both the array and the validator

  5. **Modify `run_fuzz_action()`**:
     - After each operation that produces an array, call `validator(&result)` if present
     - Panic/error if validation fails

  6. **Use real temporal extension types in `random_ext_dtype()`**
  (`vortex-dtype/src/arbitrary/mod.rs`):
     - Generate `vortex.date`, `vortex.time`, `vortex.timestamp` with valid `(TimeUnit, PType)`
  combinations and proper metadata
     - Can also keep synthetic types for stress testing, but they need validators too
```

connortsui20:
Also, you cannot use "fake" extension types without registering them (see how the fuzzer failed [here](https://github.com/vortex-data/vortex/actions/runs/20531444854/job/58983246046)), so I would try and stick to the existing extension types (temporal types) before creating custom ones for the fuzzer.

```
thread '<unnamed>' (10539) panicked at vortex-error/src/lib.rs:334:33:
compare operation should succeed in fuzz test:
  Unsupported extension type "example.ipv4"
Backtrace:
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/fabece9e9491d0a3c3655dba488881968b7c5f7a/library/std/src/../../backtrace/src/backtrace/libunwind.rs:117:9
   1: std::backtrace_rs::backtrace::trace_unsynchronized::<<std::backtrace::Backtrace>::create::{closure#0}>
             at /rustc/fabece9e9491d0a3c3655dba488881968b7c5f7a/library/std/src/../../backtrace/src/backtrace/mod.rs:66:14
   2: <std::backtrace::Backtrace>::create
             at /rustc/fabece9e9491d0a3c3655dba488881968b7c5f7a/library/std/src/backtrace.rs:331:13
   3: to_arrow_dtype
             at ./vortex-dtype/src/arrow.rs:262:21
   4: to_arrow_dtype
             at ./vortex-dtype/src/arrow.rs:234:28
   5: {closure#0}
             at ./vortex-array/src/arrow/compute/to_arrow/canonical.rs:109:46
   6: unwrap_or_else<core::result::Result<arrow_schema::datatype::DataType, vortex_error::VortexError>, vortex_array::arrow::compute::to_arrow::canonical::{impl#0}::invoke::{closure_env#0}>
             at /rustc/fabece9e9491d0a3c3655dba488881968b7c5f7a/library/core/src/option.rs:1067:21
   7: invoke
             at ./vortex-array/src/arrow/compute/to_arrow/canonical.rs:109:14
   8: invoke
             at ./vortex-array/src/arrow/compute/to_arrow/mod.rs:122:42
   9: invoke
             at ./vortex-array/src/compute/mod.rs:149:34
  10: to_arrow_opts
             at ./vortex-array/src/arrow/compute/to_arrow/mod.rs:77:10
  11: into_arrow_preferred
             at ./vortex-array/src/arrow/mod.rs:41:9
  12: arrow_compare
             at ./vortex-array/src/compute/compare.rs:333:36
  13: invoke
             at ./vortex-array/src/compute/compare.rs:219:12
  14: invoke
             at ./vortex-array/src/compute/mod.rs:149:34
  15: compare
             at ./vortex-array/src/compute/compare.rs:59:10
```

Pratham1812:
Hi, got what you are saying. Should I also add a scope for retries incase the validator is unable to validate the scalar?

Pratham1812:
I have tried to introduce the validator logic, I was not sure which crate should be the right place to introduce it so the code might need some refactoring as well. Please let me know your comments.

connortsui20:
Thanks @Pratham1812, I'll have to take a look at this later today but overall looks closer to what I imagined. I think the validation function should go in the `fuzz/` crate, and I think it would be nice if you could add a custom extension type to the fuzzer (like an integer array that only has even numbers) and see if that works with the domain validator logic you added?

Pratham1812:
Hi regarding putting the validation function in `fuzz/` crate, I did not do it because the validator was also being used in `vortex-array/src/arrays/arbitrary.rs` so if I put it in the fuzz crate the vortex-array crate will as well depend upon the `fuzz/`, if there are no issues with that I will do that.

connortsui20:
Ah I missed that @Pratham1812, sorry about that!

We don't want to move the fuzz logic into `vortex-array`, but I think we could maybe make the validator a trait method and implement it on all extension arrays? Or maybe even add it as "required" metadata on the `ExtDType`...

Anyways that is probably out of scope for this PR (unless @robert3005 you have any thoughts on this). I'll run the fuzzer on this branch and we can see how it runs before I give a review again. Could you also fix the docs CI error and rebase / merge this branch on the latest `develop`?

Edit: Fuzzer run is https://github.com/vortex-data/vortex/actions/runs/20586558368 (looks like it hit a different known issue which is a good thing!)
Edit: New fuzzer run https://github.com/vortex-data/vortex/actions/runs/20586787724

Pratham1812:
I have introduced a custom extension type to check if the validator even works. I have fixed and updated the doctest as well. Please have a look and let me know your comments.

Edit: For the second fuzz run I am unable to understand if this error was introduced due to changes here `fuzz/src/error.rs`

fuzz/src/error.rs
```diff
@@ -121,6 +123,13 @@ impl Display for VortexFuzzError {
                     rhs.display_tree(),
                 )
             }
+            VortexFuzzError::DomainValidationFailed(expected, idx, lhs, step, backtrace) => {
```
Pratham1812: I am not sure if the second fuzz run is failing due to this error.

connortsui20:
@Pratham1812 this looks good! Don't worry about that fuzzer error, that is already a known issue.

I think you may have to register the extension type as well?

Fuzzer run on this branch: https://github.com/vortex-data/vortex/actions/runs/20608064690

Pratham1812:
Hi @connortsui20, so the issue is my custom extension type "vortex.even" is not supported because it does not have any implementation in arrow. I have created a doctest for this type which proves that validator shall work, but if you want me to have a custom implementation for this type as well let me do that, but since this is just for testing purposes we may as well remove it. I am fine with whatever you say.
image

connortsui20:
Interesting, I had no idea!

In that case, I think it's fine to just add the existing temporal types we have in Vortex to the fuzzer and leave it at that. @robert3005 do you have any thoughts on this? If not I'll just review this as is.

And @Pratham1812 can you remove the custom extension type? Thanks again for doing this long and unexpected process!

Pratham1812:
Hey I removed the custom type. Can you trigger the fuzzer and CI checks?

connortsui20:
@Pratham1812 Apologies for the delay, I triggered this run: https://github.com/vortex-data/vortex/actions/runs/20672458544

It should start after the current fuzzer finishes. I am now realizing that the concurrency group for this action should probably be branch dependent...

fuzz/src/array/compare.rs
```diff
@@ -123,8 +123,16 @@ pub fn compare_canonical_array(array: &dyn Array, value: &Scalar, operator: Oper
             )
             .into_array()
         }
-        d @ (DType::Null | DType::Extension(_)) => {
-            unreachable!("DType {d} not supported for fuzzing")
+        DType::Null => {
+            unreachable!("DType null not supported for fuzzing")
+        }
+        DType::Extension(..) => {
+            // Extension arrays delegate comparison to their storage type
```
connortsui20, line 130: Nit: periods at the end of sentences (this is true everywhere)

connortsui20:
It looks like the fuzzer run that I triggered on this branch failed, but it actually seems like this is a proper error with something we have not implemented? In this case this branch fuzzer is working!

https://github.com/vortex-data/vortex/actions/runs/20672458544/job/59358118001

I have a few nits but other than that I think this is good to go. Thank you @Pratham1812 for your work!

fuzz/src/array/fill_null.rs
```diff
@@ -40,13 +43,33 @@ pub fn fill_null_canonical_array(
         Canonical::VarBinView(array) => {
             fill_varbinview_array(&array, fill_value, result_nullability)
         }
-        Canonical::Struct(_)
-        | Canonical::List(_)
-        | Canonical::FixedSizeList(_)
-        | Canonical::Extension(_) => fill_null(canonical.as_ref(), fill_value)?,
+        Canonical::Extension(array) => fill_extension_array(&array, fill_value),
```
connortsui20: Nit: Move this match arm to the bottom of the match statement

fuzz/src/array/fill_null.rs
```diff
@@ -40,13 +43,33 @@ pub fn fill_null_canonical_array(
...
+    .vortex_expect("fill_null should succeed in canonical form");
+
+    if filled_storage.dtype().nullability() == array.ext_dtype().storage_dtype().nullability() {
+        ExtensionArray::new(array.ext_dtype().clone(), filled_storage).into_array()
+    } else {
+        let new_ext_dtype = Arc::new(ExtDType::new(
+            array.ext_dtype().id().clone(),
+            Arc::new(filled_storage.dtype().clone()),
+            array.ext_dtype().metadata().cloned(),
+        ));
```
connortsui20: A comment here would be nice to explain why we need to do this

fuzz/src/array/slice.rs
```diff
@@ -113,8 +116,24 @@ pub fn slice_canonical_array(
...
+            if sliced_storage.dtype().nullability() == ext_dtype.storage_dtype().nullability() {
+                Ok(ExtensionArray::new(ext_dtype.clone(), sliced_storage).into_array())
+            } else {
+                let new_ext_dtype = Arc::new(ExtDType::new(
+                    ext_dtype.id().clone(),
+                    Arc::new(sliced_storage.dtype().clone()),
+                    ext_dtype.metadata().cloned(),
+                ));
+                Ok(ExtensionArray::new(new_ext_dtype, sliced_storage).into_array())
+            }
```
connortsui20, line 136: Actually this seems to be duplicated in a few files, maybe we can extract this out into a helper method and put it in `fuzz/src/array/mod.rs`?

fuzz/src/array/mod.rs
```diff
@@ -662,6 +662,9 @@ pub fn assert_array_eq(
     rhs: &ArrayRef,
     step: usize,
 ) -> crate::error::VortexFuzzResult<()> {
+    use vortex_array::ToCanonical;
+    use vortex_array::arrays::validator_for_ext_type;
+
     use crate::error::Backtrace;
     use crate::error::VortexFuzzError;
```
connortsui20: This isn't technically your change but can you lift these imports to the top of the file?

fuzz/src/array/mod.rs
```diff
@@ -699,7 +702,22 @@ pub fn assert_array_eq(
                 Backtrace::capture(),
             ));
         }
+
+        // Also validate the expected array's domain constraints for extension types
+        if matches!(lhs.dtype(), DType::Extension(..)) {
+            let validator = validator_for_ext_type(lhs.to_extension().ext_dtype());
+            if !validator(&l) {
```
connortsui20, line 720: I think I missed this before, but is there a reason why the `DomainValidator` takes a `Scalar` and not an `ArrayRef`? I feel like it makes more sense for the validator to take the entire array and check the values directly instead of requiring the caller to do that for every scalar (note that our current `Scalar`s are _super_ inefficient and slow to get because they usually make one or even several allocations).

vortex-array/src/arrays/arbitrary.rs
```diff
@@ -164,12 +166,51 @@ fn random_array_chunk(
...
 }
 
+/// Creates a random extension array.
+/// If the `chunk_len` is specified, the length of the array will be equal to the chunk length.
+fn random_extension(
+    u: &mut Unstructured,
+    ext_dtype: &Arc<ExtDType>,
+    chunk_len: Option<usize>,
+) -> Result<ArrayRef> {
+    use crate::builders::ExtensionBuilder;
```
connortsui20: Can you put this import at the top?

vortex-array/src/arrays/arbitrary.rs
```diff
@@ -164,12 +166,51 @@ fn random_array_chunk(
...
+
+    // Determine array length
+    let array_length = chunk_len.unwrap_or(u.int_in_range(0..=20)?);
+
+    // Create builder for the extension array
+    let mut builder = ExtensionBuilder::with_capacity(ext_dtype.clone(), array_length);
+
+    // Generate random values, retrying if they don't pass validation
+    for _ in 0..array_length {
+        // Retry loop to generate valid values
```
connortsui20: Nit:

```
        // Retry loop in case we generate invalid values.
```

connortsui20 (changes requested):

fuzz/src/array/mod.rs
Pratham1812: Hey this was done because while generating a random_extension array each scalar had to be validated individually before inserting. I can make a helper function that returns a validator for the whole array if needed. `https://github.com/Pratham1812/vortex/blob/5b99fa6fd0c2d9505d2ce0bdf156b4d9c39360a6/vortex-array/src/arrays/arbitrary.rs#L201-L208`

robert3005:
The issue existed because in order to truly fix it you need to implement extension dtypes. I think now it's possible with vortex session but it's a considerable change. Having timestamp types implemented in fuzzer though is a good start

Pratham1812:
I think now I do understand the Code Base and probably I can help in the changes required. Maybe not in this PR or a separate PR for that matter. :)

fuzz/src/array/filter.rs
```diff
@@ -115,8 +117,27 @@ pub fn filter_canonical_array(array: &dyn Array, filter: &[bool]) -> VortexResul
...
+                        filtered_storage.dtype().clone(),
+                        ext_dtype.metadata().cloned(),
+                    ),
+                    filtered_storage,
+                )
+                .into_array())
+            }
+        }
+        DType::Null => {
+            unreachable!("Cannot search sorted on Null array")
```
joseph-isaacs, line 140: This is not true

vortex-array/src/arrays/extension/validate.rs
joseph-isaacs, line 1: I am not sure this is the structure we want for validation. Shall we just hardcode timestamp validation and have the fuzz only generated those

vortex-array/src/compute/fill_null.rs
```diff
@@ -133,6 +137,36 @@ impl ComputeFnVTable for FillNull {
...
+                let new_ext_dtype = Arc::new(ExtDType::new(
+                    array.to_extension().ext_dtype().id().clone(),
+                    Arc::new(filled_storage.dtype().clone()),
+                    array.to_extension().ext_dtype().metadata().cloned(),
+                ));
+                return Ok(ExtensionArray::new(new_ext_dtype, filled_storage)
+                    .into_array()
+                    .into());
+            }
+        }
```
joseph-isaacs, line 169: we cannot do this.

joseph-isaacs:
Shall we open another PR which just add fuzzing for temporal values. This is something we will accept

Pratham1812:
@joseph-isaacs should I create another PR? This PR should work for temporal types.

joseph-isaacs:
I am happy for you to update and reopen. But we cannot impl all the compute for any extension array just temporal

Pratham1812:
I mean if I can get a starting point, I can potentially do it. Also before I was trying with some custom types but since those types were not supported by "arrow" and the fuzzer failed as pointed in earlier comments. So I was not sure how can it be done.

connortsui20:
Personally I think it's fine to open a separate PR that only hardcodes the temporal array types into the fuzzer since this one has turned more into research and development, so a clean slate would be easier to review.

fuzz/src/array/fill_null.rs
```diff
@@ -40,13 +41,35 @@ pub fn fill_null_canonical_array(
...
+            clone_ext_dtype(
+                array.ext_dtype().id().clone(),
+                filled_storage.dtype().clone(),
+                array.ext_dtype().metadata().cloned(),
+            ),
+            filled_storage,
+        )
+        .into_array()
+    }
+}
```
joseph-isaacs, line 71: Here you should for now only implemented the fill_null for ext array that are also temporal arrays

fuzz/src/array/filter.rs
```diff
@@ -115,8 +117,27 @@ pub fn filter_canonical_array(array: &dyn Array, filter: &[bool]) -> VortexResul
...
-            unreachable!("DType {d} not supported for fuzzing")
+        DType::Extension(ext_dtype) => {
+            // Extension arrays delegate filter to their storage type.
+            let filtered_storage = filter_canonical_array(array.to_extension().storage(), filter)?;
+
+            if filtered_storage.dtype().nullability() == ext_dtype.storage_dtype().nullability() {
+                Ok(ExtensionArray::new(ext_dtype.clone(), filtered_storage).into_array())
+            } else {
+                // The storage dtype changed (i.e., became nullable due to filtering).
+                Ok(ExtensionArray::new(
```
joseph-isaacs, line 128: same for filter

vortex-array/src/arrays/arbitrary.rs
```diff
@@ -164,12 +167,48 @@ fn random_array_chunk(
...
+    // Generate random values, retrying if they don't pass validation.
+    for _ in 0..array_length {
+        // Retry loop in case we generate invalid values.
+        loop {
+            // Generate a random storage scalar.
+            let storage_scalar = random_scalar(u, ext_dtype.storage_dtype())?;
+
+            // Wrap it in an extension scalar.
+            let ext_scalar = Scalar::extension(ext_dtype.clone(), storage_scalar);
+
```
joseph-isaacs, line 199: here only create random temporal types see `TemporalArray`

joseph-isaacs (commented):
Thanks for doing this PR. I have outlines some of the change you need to only support temporal types. Please remove all instances of ext and replace with a temporal array instance (date, time and datetime).

We can review that one

Issue #3859, opened 2025-07-14
Add a method to push struct validity into children


blaginin:
Unlike Arrow's `RecordBatches`, `StructArray` can be "nullable" in two places: in the child arrays and also at the top level. This makes some struct operations tricky - for example, how do you zip two arrays with different top-level validity?

One solution could be to add a method that pushes top-level validity down to the children (assuming all fields are nullable). For example:

`{a: 1, b: 2}, NULL` would become `{a: 1, b: 2}, {a: NULL, b: NULL}`

robert3005:
We have https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/compute/mask.rs#L48, however maybe the faster way is to accept mask in some compute functions and let them decide how to apply it

blaginin:
> We have https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/compute/mask.rs#L48

Sure! But what do you think about adding a separate method for the struct? This feels like a super common usecase - have written this three times in spiral last week

blaginin:
Okay, so what needs to be done here: 
- We need to implement a method to push validity
- We need to add docs
- Method should be tested

I actually wrote an in-place method, feel free to use this as a reference / base on top of that: https://gist.github.com/blaginin/9927afd94a8d5bb272135d0c5dc99652

amorynan:
Is this issue on going ? and  can I  pick this up for diving deep into vortex ?

gatesn:
Yes!

Since this issue was filed, we now have a "Mask" expression that essentially performs intersection with validity.

So you could take a look at how you might take the validity from the struct array itself, and wrap up each child field in a mask expression (see builtins.rs), before constructing a new StructArray with or without the previous validity. These two are different behaviors.

Also please note most of the team are off for the holidays until early next year, so responses may be slow or delayed. Thanks for getting involved!

PR #5826, opened 2025-12-27
feat: Add push_validity_into_children methods to StructArray

amorynan:
For solving : https://github.com/vortex-data/vortex/issues/3859
Add methods to push struct-level validity into child fields:
- push_validity_into_children(preserve_struct_validity: bool)
- push_validity_into_children_default() - convenience method with preserve=false

The functionality propagates null information from struct level down to individual fields, with options to preserve or remove the struct-level validity.

Includes comprehensive tests covering all scenarios:
- preserve_struct_validity = true
- preserve_struct_validity = false (default)
- no nulls edge case

### DIFF
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -462,8 +462,8 @@ impl StructArray {
     /// # Parameters
     ///
     /// * `preserve_struct_validity` - If true, the new struct array retains the original struct-level
-    ///   validity. If false, the new struct array has `Validity::AllValid` since all null information
-    ///   is now contained within the individual fields.
+    ///   validity. If false, the new struct array has `Validity::NonNullable` since all null information
+    ///   is now contained within the individual fields and the struct instances themselves cannot be null.
     ///
     /// # Returns
     ///
@@ -477,33 +477,35 @@ impl StructArray {
     /// use vortex_array::IntoArray;
     /// use vortex_buffer::buffer;
     ///
-    /// // Create struct with top-level nulls
+    /// // Create struct with top-level nulls. Child fields are created as non-nullable
+    /// // but become nullable after pushing validity.
     /// let struct_array = StructArray::try_new(
     ///     ["a", "b"].into(),
     ///     vec![
-    ///         buffer![1i32, 2i32, 3i32].into_array(),
-    ///         buffer![10i32, 20i32, 30i32].into_array(),
+    ///         buffer![1i32, 2i32, 3i32].into_array(),  // non-nullable initially
+    ///         buffer![10i32, 20i32, 30i32].into_array(), // non-nullable initially
     ///     ],
     ///     3,
-    ///     Validity::from_iter([true, false, true]), // row 1 is null
+    ///     Validity::from_iter([true, false, true]), // row 1 is null at struct level
     /// ).unwrap();
     ///
     /// // Push validity into children, preserving struct validity
     /// let pushed = struct_array.push_validity_into_children(true).unwrap();
-    /// // pushed.fields()[0] now has nulls at position 1
-    /// // pushed.fields()[1] now has nulls at position 1
-    /// // pushed.validity still shows row 1 as null
+    /// // Result: struct has nulls at position 1, fields have nulls at position 1
+    /// // pushed.validity: [true, false, true] (preserved)
+    /// // pushed.fields()[0]: [1, null, 3] (nullable)
+    /// // pushed.fields()[1]: [10, null, 30] (nullable)
     ///
     /// // Push validity into children, removing struct validity
     /// let pushed_no_struct = struct_array.push_validity_into_children(false).unwrap();
-    /// // pushed_no_struct.fields()[0] now has nulls at position 1
-    /// // pushed_no_struct.fields()[1] now has nulls at position 1
-    /// // pushed_no_struct.validity is AllValid
+    /// // Result: struct instances cannot be null, all null info is in fields
+    /// // pushed_no_struct.validity: NonNullable (struct instances are never null)
+    /// // pushed_no_struct.fields()[0]: [1, null, 3] (nullable)
+    /// // pushed_no_struct.fields()[1]: [10, null, 30] (nullable)
     /// ```
     /// Push validity into children with default behavior (preserve_struct_validity = false).
 
     pub fn push_validity_into_children(&self, preserve_struct_validity: bool) -> VortexResult<Self> {
-        use crate::compute::mask;
 
         // Get the struct-level validity mask.
         let struct_validity_mask = self.validity_mask();
@@ -535,7 +537,7 @@ impl StructArray {
             .iter()
             .map(|field| {
                 // Use the mask function to apply null positions to each field.
-                mask(field.as_ref(), &null_mask)
+                crate::compute::mask(field.as_ref(), &null_mask)
             })
             .collect::<VortexResult<Vec<_>>>()?;
 
@@ -543,7 +545,7 @@ impl StructArray {
         let new_struct_validity = if preserve_struct_validity {
             self.validity.clone()
         } else {
-            Validity::AllValid
+            Validity::NonNullable
         };
 
         // Construct the new struct array.
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -501,25 +501,19 @@ impl StructArray {
     /// // pushed_no_struct.validity is AllValid
     /// ```
     /// Push validity into children with default behavior (preserve_struct_validity = false).
-    ///
-    /// This is equivalent to calling `push_validity_into_children(false)`.
-    pub fn push_validity_into_children_default(&self) -> VortexResult<Self> {
-        self.push_validity_into_children(false)
-    }
-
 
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,112 @@ impl StructArray {
...
+    /// // pushed_no_struct.fields()[0] now has nulls at position 1
+    /// // pushed_no_struct.fields()[1] now has nulls at position 1
+    /// // pushed_no_struct.validity is AllValid
+    /// ```
+    /// Push validity into children with default behavior (preserve_struct_validity = false).
+    ///
+    /// This is equivalent to calling `push_validity_into_children(false)`.
+    pub fn push_validity_into_children_default(&self) -> VortexResult<Self> {
+        self.push_validity_into_children(false)
+    }
```
connortsui20: I don't think we need to have this method, the one below should be sufficient since a the `bool` impl of `Default` is already `false`.

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,112 @@ impl StructArray {
...
+
+
+    pub fn push_validity_into_children(&self, preserve_struct_validity: bool) -> VortexResult<Self> {
+        use crate::compute::mask;
```
connortsui20: Can you move this import to the top?

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,112 @@ impl StructArray {
...
+
+        // Get the struct-level validity mask
```
connortsui20: Nit: periods at the ends of sentences.

connortsui20 (changes requested):
I think this is a good first step, though I'm trying to figure out how we want to implement this behavior for the new operator world we are migrating to.

@gatesn mentioned this in the original issue:

> Since this issue was filed, we now have a "Mask" expression that essentially performs intersection with validity.
>
> So you could take a look at how you might take the validity from the struct array itself, and wrap up each child field in a mask expression (see builtins.rs), before constructing a new StructArray with or without the previous validity. These two are different behaviors.

This PR implements this feature correctly for the old (current) world, but we will likely want to implement the behavior described above very very soon, so soon that it might not even be worth merging this right now? @gatesn do you have any thoughts?

vortex-array/src/arrays/struct_/array.rs
amorynan: OK, I had fix it . Please review again.

vortex-array/src/arrays/struct_/array.rs
amorynan: Yes , I remove the default Method . Please review it .

vortex-array/src/arrays/struct_/array.rs
connortsui20: I also prefer to see `compute::mask` as the function call instead of just `mask`, but that is just my personal preference

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,106 @@ impl StructArray {
...
+    /// // pushed.fields()[0] now has nulls at position 1
+    /// // pushed.fields()[1] now has nulls at position 1
+    /// // pushed.validity still shows row 1 as null
+    ///
+    /// // Push validity into children, removing struct validity
+    /// let pushed_no_struct = struct_array.push_validity_into_children(false).unwrap();
+    /// // pushed_no_struct.fields()[0] now has nulls at position 1
+    /// // pushed_no_struct.fields()[1] now has nulls at position 1
+    /// // pushed_no_struct.validity is AllValid
+    /// ```
```
connortsui20: I think it would be more helpful to comment what the struct array looks like instead of requiring the reader to parse what the created struct array becomes after `try_new`. And you might as well have assertions stating that specific positions are null rather than just comments.

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,106 @@ impl StructArray {
...
+    /// use vortex_array::validity::Validity;
+    /// use vortex_array::IntoArray;
+    /// use vortex_buffer::buffer;
+    ///
+    /// // Create struct with top-level nulls
+    /// let struct_array = StructArray::try_new(
+    ///     ["a", "b"].into(),
+    ///     vec![
+    ///         buffer![1i32, 2i32, 3i32].into_array(),
+    ///         buffer![10i32, 20i32, 30i32].into_array(),
```
connortsui20: It may not be immediately obvious to someone reading this that the child buffers are non-nullable

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,106 @@ impl StructArray {
...
+        if struct_validity_mask.all_true() {
+            return if preserve_struct_validity {
+                Ok(self.clone())
+            } else {
+                // Remove struct validity if requested.
+                Self::try_new(
+                    self.names().clone(),
+                    self.fields().clone(),
+                    self.len(),
+                    Validity::AllValid,
```
connortsui20, line 523: Hmmm I wonder if we should make the validity of the top-level struct array `Validity::NonNullable` instead of `AllValid` if the user requests to not preserve the struct validity. Do you have any thoughts?

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,106 @@ impl StructArray {
...
+        });
+
+        let masked_fields: Vec<ArrayRef> = self
+            .fields()
+            .iter()
+            .map(|field| {
+                // Use the mask function to apply null positions to each field.
+                mask(field.as_ref(), &null_mask)
+            })
+            .collect::<VortexResult<Vec<_>>>()?;
```
connortsui20, line 542: This inverted logic makes me think more that we should just move this to the new world now instead of merging this and then getting rid of it immediately. @gatesn Any thoughts about this?

vortex-array/src/arrays/struct_/array.rs
amorynan: done!

vortex-array/src/arrays/struct_/array.rs
amorynan: done!

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,108 @@ impl StructArray {
...
+                    self.len(),
+                    Validity::AllValid,
+                )
+            };
+        }
+
+        // Apply the struct validity mask to each child field.
+        // We want to set nulls where the struct is null (i.e., where struct_validity_mask is false).
+        // So we need to invert the mask: where struct is invalid, set child to invalid.
+        let null_mask = struct_validity_mask.iter_bools(|iter| {
```
joseph-isaacs, line 531: there is a  not method

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,106 @@ impl StructArray {
...
+    ///
+    /// This method takes the top-level validity of the struct array and applies it to each child field
+    /// using a mask operation. The resulting struct array will have the struct-level nulls propagated
+    /// down to the individual fields.
+    ///
+    /// # Parameters
+    ///
+    /// * `preserve_struct_validity` - If true, the new struct array retains the original struct-level
+    ///   validity. If false, the new struct array has `Validity::AllValid` since all null information
+    ///   is now contained within the individual fields.
```
joseph-isaacs, line 466: can we invert this and have a remove_struct_validity since the default (false) should be to keep it. Since that doesn't change the values in the struct whereas removing this does

vortex-array/src/arrays/struct_/array.rs
joseph-isaacs: create a two struct are compare using assert_array_eq!

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -451,4 +452,106 @@ impl StructArray {
...
+    /// // pushed.fields()[1] now has nulls at position 1
+    /// // pushed.validity still shows row 1 as null
+    ///
+    /// // Push validity into children, removing struct validity
+    /// let pushed_no_struct = struct_array.push_validity_into_children(false).unwrap();
+    /// // pushed_no_struct.fields()[0] now has nulls at position 1
+    /// // pushed_no_struct.fields()[1] now has nulls at position 1
+    /// // pushed_no_struct.validity is AllValid
+    /// ```
+    /// Push validity into children with default behavior (preserve_struct_validity = false).
```
joseph-isaacs, line 503: invert please

vortex-array/src/arrays/struct_/array.rs
joseph-isaacs: I think we should keep this for now

vortex-array/src/arrays/struct_/tests.rs
joseph-isaacs, line 1: can you use `assert_arrays_eq!` to check equality

joseph-isaacs (commented):
I wonder if we want to split this into two functions. 

One which pushes down the validity into all children and makes the struct non-nullable (`push_validity_into_children()`) and another that only push down validity (possibly duplicating it) (`compact_validity()`)

PR #5828, opened 2025-12-27
Fix: benchmark name for `bench.yml`

connortsui20:
Ok this will actually fix it after https://github.com/vortex-data/vortex/pull/5823


robert3005 (approved):

connortsui20:
I guess it didnt fix it

PR #5830, opened 2025-12-29
Fix: actually actually fix CI

connortsui20:
I think this will finally fix it

Note that there is still a datafusion not implemented thing here but I think that is intentional?


robert3005 (approved):

connortsui20:
Ok things are just not implemented, @AdamGS can you take a look at this when you get back?

connortsui20:
Nevermind this is fixed in https://github.com/vortex-data/vortex/pull/5830

PR #5836, opened 2025-12-29
Fix: CI SQL Remote

connortsui20:
Removes lance from the SQL remote benchmarks on S3 and cleaned up some commands.


connortsui20:
I also _just_ realized that the lance compression benchmark was not carried over from the [previous implementation](https://github.com/vortex-data/vortex/blob/155ca5b311e660f2be0f7d8e57c76919ae4676fb/bench-vortex/src/compress/lance.rs). If someone could take a look at this (bring back the old code and use it in the new benchmarking infra), I think that that will finally unblock all CI from crashing.

https://github.com/vortex-data/vortex/blob/dde9c0b0d6b9bcfd593fc741327bdc0004046a45/benchmarks/compress-bench/src/main.rs#L95

robert3005 (approved):

PR #5853, opened 2026-01-05
Pluggable expression conversion for DF

AdamGS:
This PR introduces an extendable interface for converting DataFusion physical expressions into Vortex ones, so that users can define their own behavior for UDFs or any other home-brewed physical expressions.

I think this is an OK way to expose the defaults for now, but would appreciate thoughts.

### DIFF
--- a/vortex-datafusion/src/persistent/source.rs
+++ b/vortex-datafusion/src/persistent/source.rs
@@ -37,6 +37,8 @@ use vortex_utils::aliases::dash_map::DashMap;
 use super::cache::VortexFileCache;
 use super::metrics::PARTITION_LABEL;
 use super::opener::VortexOpener;
+use crate::convert::exprs::DefaultExpressionConvertor;
+use crate::convert::exprs::ExpressionConvertor;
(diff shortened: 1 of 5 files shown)


vortex-datafusion/src/persistent/source.rs
```diff
@@ -79,8 +82,16 @@ impl VortexSource {
             expr_adapter_factory: None,
             _unused_df_metrics: Default::default(),
             layout_readers: Arc::new(DashMap::default()),
+            expression_convertor: Arc::new(DefaultExpressionConvertor::default()),
         }
     }
+
+    /// Provide a new [`ExpressionConvertor`] to control how Datafusion expression should be converted and pushed down.
+    pub fn with_expression_convertor(&self, expr_convertor: Arc<dyn ExpressionConvertor>) -> Self {
```
gatesn: Take `Self` here? The caller can clone if they need to, rather than forcing the clone

AdamGS: done

gatesn (approved):

PR #5854, opened 2026-01-05
fix: int compressor mul overflow bp

joseph-isaacs:


robert3005 (approved):

vortex-btrblocks/src/integer.rs
```diff
@@ -598,13 +598,13 @@ impl Scheme for DictScheme {
         // Assume codes are compressed RLE + BitPacking.
         let codes_bw = usize::BITS - stats.distinct_values_count.leading_zeros();
```
connortsui20, line 599: So I believe that this should be `u32::BITS` instead of usize? That would have likely prevented this bug in the first place?

I've fixed this in my PR: https://github.com/vortex-data/vortex/pull/7230

connortsui20: Fun fact, changing this to `u32` makes this compressed array tree:

```
root: vortex.dict(f64, len=1000) nbytes=424 B (100.00%)
  metadata: DictMetadata { values_len: 5, codes_ptype: U8, is_nullable_codes: Some(false), all_values_referenced: Some(true) }
  codes: fastlanes.bitpacked(u8, len=1000) nbytes=384 B (90.57%) [nulls=0, min=0u8, max=4u8]
    metadata: BitPackedMetadata { bit_width: 3, offset: 0, patches: None }
    buffer: packed host 384 B (align=1) (100.00%)
  values: vortex.primitive(f64, len=5) nbytes=40 B (9.43%) [nulls=0]
    metadata: EmptyMetadata
    buffer: values host 40 B (align=8) (100.00%)
```

turn into this instead (which I think is better for optimization / compute even though it is the same size)

```
root: vortex.alp(f64, len=1000) nbytes=424 B (100.00%)
  metadata: ALPMetadata { exp_e: 14, exp_f: 13, patches: None }
  encoded: vortex.dict(i64, len=1000) nbytes=424 B (100.00%)
    metadata: DictMetadata { values_len: 5, codes_ptype: U8, is_nullable_codes: Some(false), all_values_referenced: Some(true) }
    codes: fastlanes.bitpacked(u8, len=1000) nbytes=384 B (90.57%) [nulls=0, min=0u8, max=4u8]
      metadata: BitPackedMetadata { bit_width: 3, offset: 0, patches: None }
      buffer: packed host 384 B (align=1) (100.00%)
    values: vortex.primitive(i64, len=5) nbytes=40 B (9.43%) [nulls=0, min=11i64, max=55i64]
      metadata: EmptyMetadata
      buffer: values host 40 B (align=8) (100.00%)
```

PR #5875, opened 2026-01-06
fix: Move from macos-13 runners to macos-15-intel runners

robert3005:


AdamGS (approved):
maybe should just kill macos intel?

Issue #4677, opened 2025-09-17
`time_elapsed_opening` spends too much time


xudong963:
I have two tables, one with vortex format, another parquet.

Vortex:
```
dft -c "explain analyze select * from vortex_table where ticker = 'EAOK' order by market limit 10" --flightsql
Registering deltalake
+-------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| plan_type         | plan                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
+-------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Plan with Metrics | CooperativeExec, metrics=[]                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
|                   |   FileScanExec: table = vortex_table, filters=ticker@0 = EAOK, output_ordering=["market@2 ASC NULLS LAST", "start_date@39 ASC NULLS LAST", "last_updated_utc@16 ASC NULLS LAST"], scan_direction=Native, page_size=1024, partitions=[22], fetch=10, metrics=[output_rows=10, elapsed_compute=1ns, batches_splitted=0, file_open_errors=0, file_scan_errors=0, time_elapsed_opening=23.989515ms, time_elapsed_processing=33.761589ms, time_elapsed_scanning_total=15.77185ms, time_elapsed_scanning_until_data=15.77165ms] |
|                   |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
+-------------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

Parquet:
```
 dft -c "explain analyze select * from parquet_table where ticker = 'EAOK' order by market limit 10" --flightsql
Registering deltalake
[...]

PR #5879, opened 2026-01-07
perf: reduce time_elapsed_opening overhead

godnight10061:
This PR tries to reduce `time_elapsed_opening` for Vortex scans (observed vs Parquet).

Changes:
- `vortex-file`: shrink default footer initial read from 1MiB to `MAX_POSTSCRIPT_SIZE + EOF_SIZE` (~64KiB) and add a regression test.
- `vortex-scan`: make `ScanBuilder::into_stream()` lazy (defer `prepare()` / split registration until first poll) and add a unit test to ensure stream construction has no split-planning side effects.
- `vortex-datafusion`: expose the footer initial read size as a format option (`footer_initial_read_size_bytes`) and plumb it into `VortexOpenOptions::with_initial_read_size`.

Notes:
- Scan planning errors now surface on first poll instead of during `into_stream()` construction.
- If the footer/schema/layout don’t fit in the initial window, `read_footer` will issue additional reads as before.

Tests:
- `cargo +nightly fmt --all --check`
- `cargo clippy -p vortex-datafusion --all-targets --all-features -- -D warnings`
- `cargo test --locked -p vortex-file -p vortex-scan -p vortex-io`
- `cargo test --locked -p vortex-datafusion`

Related: vortex-data/vortex#4677


AdamGS:
I think that for many modern setups, 1Mb and 64KB are pretty close in terms of latency, maybe the takeaway here is to expose that as a config for the DataFusion integration?
The deferred scan stream looking very promising, I'll give it a deeper look tomorrow.

AdamGS:
Seems like running benchmarks from forks is still painful, I'm working on fixing it so we can get some numbers here.

godnight10061:
FYI: 'Rust (semver checks)' was failing due to cargo-semver-checks enabling all features, which pulls in the 	est-harness feature. That feature exports stest_reuse #[template] #[export]-generated hashed macro_rules! names (e.g. search_sorted_conformance_<hash>), and semver-checks flags them as removed/renamed vs the baseline crate.

This PR updates the semver job to use eature-group: default-features so the semver surface matches the supported public API and avoids those false positives.

Because this touches .github/workflows, GitHub marks the latest CI run as action_required until a maintainer approves running workflows for this fork PR.

AdamGS:
we're working on the semver check elsewhere, it's not a required check anyway, more of a general indication

godnight10061:
Thanks! I'll turn the initial read size into a config as suggested. Really appreciate you handling the benchmark infra for forks—looking forward to your feedback tomorrow!

.github/workflows/ci.yml
```diff
@@ -364,6 +364,10 @@ jobs:
 
       - name: Check semver
         uses: obi1kenobi/cargo-semver-checks-action@v2
+        with:
+          # Avoid enabling test-only feature flags (e.g. `test-harness`) that export unstable
+          # procedural-macro-generated items and create false-positive semver diffs.
+          feature-group: default-features
```
AdamGS: Lets remove this from this PR, we might refactor that whole macro or use an upstream fix

AdamGS:
You can see the benchmark results in the run summary, seems to make more of a difference for DuckDB but I'll take that, I expect it to also make a difference for DataFusion or at least for that metric, I'll try it out soon.

AdamGS:
A small local test I ran showed very nice improvements to `time_elapsed_opening` which is cool.

AdamGS (approved):
LGTM!

connortsui20:
There's a noticeable improvement in several of our [benchmarks](https://bench.vortex.dev/) because of this which is nice! Though I did notice regressions in Clickbench Q6 and Q23 that someone should probably investigate?

PR #5881, opened 2026-01-07
Fix: struct cast for arrays with duplicate fields if cast doesn't change column names

robert3005:
This has started to fail after DF 51 upgrade where we added additional logic to cast to support reordering

### DIFF
--- a/vortex-array/src/arrays/struct_/compute/cast.rs
+++ b/vortex-array/src/arrays/struct_/compute/cast.rs
@@ -26,11 +26,12 @@ impl CastKernel for StructVTable {
 
         let source_sdtype = array.struct_fields();
 
-        let fields_match_order = target_sdtype
-            .names()
-            .iter()
-            .zip_eq(source_sdtype.names().iter())
-            .all(|(f1, f2)| f1 == f2);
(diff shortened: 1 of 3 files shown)


vortex-array/src/arrays/struct_/compute/cast.rs
joseph-isaacs, line 1: tests please

joseph-isaacs: can we also document the cast being purely nominal and allow shuffle, duplication??, remove/filter and maybe null expand?

vortex-array/src/compute/cast.rs
```diff
@@ -34,6 +34,8 @@ pub(crate) fn warm_up_vtable() -> usize {
 /// Attempt to cast an array to a desired DType.
 ///
 /// Some array support the ability to narrow or upcast.
+/// Struct DType casting is nominal and will resolve fields by their name and not their position.
```
joseph-isaacs, line 37: (5) subset, superset with null, reorder, cast struct nullability (?) and also cast inner fields

joseph-isaacs (approved):

PR #5882, opened 2026-01-07
[CI] Don't remove label if running from fork

AdamGS:
Removing a label requires `contents: write` permissions, which we don't want to grant to forks.


joseph-isaacs (approved):
add fail_on_error:

PR #5893, opened 2026-01-08
Add benchmark compare to action summary

AdamGS:
Taken from the [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary), also only try to comment on the PR if its not coming from a fork.


AdamGS:
running benchmarks here to see that the PR comments still work + that the summery looks good

joseph-isaacs (approved):
can we turn off sscache

PR #5895, opened 2026-01-08
feat[array]: use execute to return `Canonical` array not `Vector`

joseph-isaacs:
Use `execute_canonical` instead of `execute` [likely no one will need to fix anything since the API is very new


vortex-array/src/arrays/dict/execute.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::arrays::StructVTable;
+use crate::arrays::VarBinViewArray;
+use crate::arrays::VarBinViewVTable;
+use crate::compute::TakeKernel;
+
+/// Take from a canonical array using indices (codes), returning a new canonical array.
+///
+/// This is the core operation for dictionary decoding - it expands the dictionary
+/// by looking up each code in the values array.
+pub fn take_canonical(values: Canonical, codes: &PrimitiveArray) -> Canonical {
```
joseph-isaacs: we don't need this, but I think we will want this once take compute fn is gone

vortex-array/src/arrays/filter/execute.rs
```diff
@@ -0,0 +1,169 @@
...
+            for (start, end) in slices.iter().copied() {
+                output.extend_from_slice(&values[start..end]);
+            }
+            output.freeze()
+        }
+    }
+}
+
+/// Filter a canonical array by a mask, returning a new canonical array.
+pub fn filter_canonical(canonical: Canonical, mask: &Mask) -> Canonical {
```
joseph-isaacs: same as take and mask

0ax1:
@claude analyze this pr deeply

vortex-array/src/arrays/dict/execute.rs
```diff
@@ -0,0 +1,117 @@
...
+        Canonical::Decimal(a) => Canonical::Decimal(take_decimal(&a, codes)),
+        Canonical::VarBinView(a) => Canonical::VarBinView(take_varbinview(&a, codes)),
+        Canonical::List(a) => Canonical::List(take_listview(&a, codes)),
+        Canonical::FixedSizeList(a) => Canonical::FixedSizeList(take_fixed_size_list(&a, codes)),
+        Canonical::Struct(a) => Canonical::Struct(take_struct(&a, codes)),
+        Canonical::Extension(a) => Canonical::Extension(take_extension(&a, codes)),
+    }
+}
+
+fn take_null(_array: &NullArray, codes: &PrimitiveArray) -> NullArray {
```
0ax1: For the sake of keeping PRs small, can we separate diff changes like that into diff PRs in the future?

0ax1 (commented):
I can't spot anything obvious that stands out as wrong but this PR is quite large to review due to its size. I'd really like us to split up PRs into smaller chunks in the future. Probably good if also @gatesn has a peek here.

joseph-isaacs:
How would you want this split up?

encodings/fastlanes/src/bitpacking/array/bitpack_decompress.rs
```diff
@@ -537,7 +535,10 @@ mod tests {
             let unpacked_array = unpack_array(&bitpacked);
 
             // Method 3: Using the execute() method (this is what would be used in production).
-            let executed = bitpacked.into_array().execute_vector(&SESSION).unwrap();
+            let executed = {
+                let mut ctx = SESSION.create_execution_ctx();
```
gatesn, line 539: Should we add `SESSION.execute(bitpacked.into_array())` as an option?

vortex-array/src/arrays/dict/vtable/mod.rs
```diff
@@ -131,10 +131,9 @@ impl VTable for DictVTable {
         Ok(())
     }
 
-    fn execute(array: &Self::Array, ctx: &mut ExecutionCtx) -> VortexResult<Vector> {
-        let values = array.values().execute(ctx)?;
-        let codes = array.codes().execute(ctx)?.into_primitive();
-        Ok(values.take(&codes))
+    fn execute(array: &Self::Array, ctx: &mut ExecutionCtx) -> VortexResult<Canonical> {
+        // TODO(joe): remove take compute fn
```
gatesn, line 135: I think this is better worded as "inline Canonical::take functionality alongside DictArray"

vortex-array/src/arrays/list/vtable/kernel/filter.rs
```diff
@@ -54,7 +56,8 @@ impl ExecuteParentKernel<ListVTable> for ListFilterKernel {
         // TODO(ngates): for ultra-sparse masks, we don't need to optimize the entire offsets.
         let offsets = array
             .offsets()
-            .execute_vector(ctx.session())?
+            .execute(ctx)?
```
gatesn, line 59: This mix of execute / to_vector seems weird to me.

vortex-array/src/arrow/executor/mod.rs
```diff
@@ -1,16 +1,16 @@
...
-mod byte_view;
+pub mod byte_view;
 mod decimal;
 mod dictionary;
 mod fixed_size_list;
 mod list;
 mod list_view;
-mod null;
-mod primitive;
+pub mod null;
```
gatesn, line 12: Why are these now pub?

vortex-array/src/canonical_to_vector.rs
```diff
@@ -0,0 +1,135 @@
...
+impl Canonical {
+    /// Convert a Canonical array to a Vector.
+    ///
+    /// This is the reverse of `VectorIntoArray` - it takes a fully materialized
+    /// canonical array and converts it into the corresponding vector type.
+    pub fn to_vector(self, ctx: &mut ExecutionCtx) -> VortexResult<Vector> {
+        Ok(match self {
+            Canonical::Null(a) => Vector::Null(NullVector::new(a.len())),
+            Canonical::Bool(a) => {
+                Vector::Bool(BoolVector::new(a.bit_buffer().clone(), a.validity_mask()))
```
gatesn, line 37: `.validity_mask()` performs compute!

gatesn (approved):
Lots of follow ups, notably I think we want the API to be `execute::<E>(ctx)`, but we have to interleave that with some other fixes

encodings/fastlanes/src/bitpacking/array/bitpack_decompress.rs
joseph-isaacs: I think think so, I think the explicitness is okay

vortex-array/src/arrow/executor/mod.rs
joseph-isaacs: To share logic with the old too arrow

PR #5898, opened 2026-01-08
feat: display tree with no metadata

danking:


robert3005 (approved):

vortex-array/src/display/mod.rs
```diff
@@ -164,6 +187,32 @@ impl dyn Array + '_ {
         DisplayArrayAs(self, options)
     }
 
+    /// Display the tree of array encodings and lengths without metadata or buffers.
```
joseph-isaacs: and stats?

vortex-array/src/display/mod.rs
```diff
@@ -164,6 +187,32 @@ impl dyn Array + '_ {
...
+    ///     ("y", buffer![3, 4].into_array()),
+    /// ]).unwrap().into_array();
+    /// let expected = "root: vortex.struct({x=i32, y=i32}, len=2)
+    ///   x: vortex.primitive(i32, len=2)
+    ///   y: vortex.primitive(i32, len=2)
+    /// ";
+    /// assert_eq!(format!("{}", array.display_tree_no_metadata()), expected);
+    /// ```
+    pub fn display_tree_no_metadata(&self) -> impl Display {
+        DisplayArrayAs(self, DisplayOptions::TreeNoMetadataDisplay)
```
joseph-isaacs: do you think we want to make these option compose so you can take any subset?

joseph-isaacs (commented):
I am not against this, just not sure we want to open up str eq as the best way to do array encoding equality

vortex-array/src/display/mod.rs
iakshay: Maybe we just add option here to do without metadata?

danking:
@joseph-isaacs @iakshay alright, now you can specify any subset of buffers, metadata, or stats. I included an example for each of the eight possible cases.

PR #5905, opened 2026-01-09
feat[array]: use execute to return Canonical array not Vector pt2

joseph-isaacs:


encodings/alp/src/alp/array.rs
```diff
@@ -183,9 +184,11 @@ impl VTable for ALPVTable {
         Ok(())
     }
 
-    fn execute(array: &Self::Array, _ctx: &mut ExecutionCtx) -> VortexResult<Canonical> {
-        // TODO(joe): take by value
-        Ok(decompress_into_array(array.clone()).to_canonical())
+    fn execute(array: &Self::Array, ctx: &mut ExecutionCtx) -> VortexResult<Canonical> {
+        Ok(Canonical::Primitive(execute_decompress(
```
0ax1, line 189: Is `execute_decompress` a good fn name, given that `execute` has a particular meaning how it fits into the bigger picture? `execute_decompress` is just a one off inner fn. Not sure what I'm looking for here but `execute_decompress` suggests to me that this is a fn that is part of a pattern that is also used in other locations, which it isn't.

encodings/alp/src/alp/decompress.rs
```diff
@@ -108,24 +140,16 @@ fn decompress_chunked(
...
     exponents: Exponents,
     patches: Option<Patches>,
     dtype: DType,
 ) -> PrimitiveArray {
-    let encoded = array.to_primitive();
-
-    // We need to drop ALPArray here in case converting encoded buffer into
-    // primitive didn't create a copy. In that case both alp_encoded and array
-    // will hold a reference to the buffer we want to mutate.
-    drop(array);
```
0ax1, line 127: no more drop?

0ax1 (commented):
2 remarks

encodings/alp/src/alp/decompress.rs
joseph-isaacs: you dont need ?

joseph-isaacs: we did but not here

vortex-array/src/arrays/chunked/vtable/mod.rs
```diff
@@ -176,4 +178,33 @@ impl VTable for ChunkedVTable {
     ) -> VortexResult<Option<ArrayRef>> {
         PARENT_RULES.evaluate(array, parent, child_idx)
     }
+
+    fn execute(array: &Self::Array, ctx: &mut ExecutionCtx) -> VortexResult<Canonical> {
+        if array.nchunks() == 0 {
```
gatesn: Seems weird that this isn't just the current canonicalize? We do funky stuff around swizzling struct/chunked arrays

joseph-isaacs: oops this was not meant to be here

0ax1 (approved):

vortex-array/src/arrays/filter/execute.rs
```diff
@@ -0,0 +1,162 @@
...
+                output.extend_from_slice(&values[start..end]);
+            }
+            output.freeze()
+        }
+    }
+}
+
+fn filter_decimal(array: &DecimalArray, mask: &Mask) -> DecimalArray {
+    DecimalVTable
+        .filter(array, mask)
```
gatesn, line 119: Do we not want to inline this logic here? I guess it can wait

gatesn (approved):

vortex-array/src/arrays/filter/execute.rs
joseph-isaacs: Yep, I setup the skeleton so we can do these carefully

PR #5907, opened 2026-01-09
Setup Trino plugin

gatesn:
Empty Trino plugin. 

TODO:
- [ ] Setup e2e tests


a10y (commented):
Just need to draw rest of the owl then LGTM

PR #5918, opened 2026-01-12
chore: add projection read/write example to quickstart tests

caicancai:
Gives users a minimal, end-to-end example of:
  - writing a multi-column struct,
  - reading it back with a projection expression,
  - understanding how `with_projection(select([...], root()))` behaves.

### DIFF
--- a/vortex/src/lib.rs
+++ b/vortex/src/lib.rs
@@ -175,18 +175,23 @@ impl VortexSessionDefault for VortexSession {
 /// get too verbose if we include _everything_.
 #[cfg(test)]
 mod test {
+    use std::path::PathBuf;
+
     use itertools::Itertools;
     use vortex_array::ArrayRef;
     use vortex_array::IntoArray;
     use vortex_array::ToCanonical;
     use vortex_array::arrays::PrimitiveArray;
+    use vortex_array::arrays::StructArray;
     use vortex_array::expr::gt;
     use vortex_array::expr::lit;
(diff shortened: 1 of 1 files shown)


caicancai:
@joseph-isaacs I'm not sure if this test is appropriate; if it isn't, I can close the pr.

vortex/src/lib.rs
```diff
@@ -330,4 +333,48 @@ mod test {
...
+            5,
+            Validity::NonNullable,
+        )?
+        .into_array();
+
+        // Write a Vortex file containing both columns.
+        session
+            .write_options()
+            .write(
+                &mut tokio::fs::File::create("example_projection.vortex").await?,
```
joseph-isaacs: should put this in a temp dir

caicancai: This implementation is mainly based on the implementation of other tests. The file will be deleted after the test passes. If it's to be placed in a temporary directory, should other tests also be modified accordingly?

joseph-isaacs: you want to prefix the path with cargo manifest, see: https://github.com/vortex-data/vortex/blob/ef33c3637c938594fea8a94495c9aa7d306fa205/vortex-ffi/examples/hello_vortex.rs#L61

caicancai: done, thanks. I also modified other tests to make them more standardized.

joseph-isaacs (approved):

PR #5923, opened 2026-01-12
feat[array]: pushdown struct validity on write

joseph-isaacs:
Add a new validity_pushed_down field to struct 
```
  /// If true, child validity is a superset of struct validity (validity was pushed down).
    /// For nullable children, their validity already includes struct nulls. For non-nullable
    /// children, we apply struct validity on field read. If false (default), no guarantee
    /// about relationship - must intersect validities on read.
```

Intersect struct validity with nullable children allowing us to skip validity intersection in read.

This is tracked with metadata on struct array/layout.

### DIFF
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -578,6 +578,7 @@ impl StructArray {
     /// When `validity_pushed_down` is true, for every nullable child field,
     /// the child's validity must be a superset of the struct's validity.
     /// That is, wherever the struct is invalid (null), the child must also be invalid.
+    #[cfg(debug_assertions)]
     fn validate_validity_pushed_down(&self) -> VortexResult<()> {
         if !self.validity_pushed_down {
             return Ok(());
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -32,7 +32,7 @@ pub struct StructMetadata {
     /// For nullable children, their validity already includes struct nulls. For non-nullable
     /// children, we apply struct validity on field read. If false (default), no guarantee
     /// about relationship - must intersect validities on read.
-    #[prost(bool, tag = "1")]
+    #[prost(bool, tag = "1", default = false)]
     pub(super) validity_pushed_down: bool,
 }
 
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -616,6 +616,11 @@ impl StructArray {
     ///
     /// For nullable structs, setting this to true indicates that child validity
     /// is a superset of struct validity (children include struct's nulls).
+    ///
+    /// # Safety
+    ///
+    /// If set all non-nullable field must have their nullability be a superset of the struct
+    /// validity
     pub unsafe fn with_validity_pushed_down(mut self, validity_pushed_down: bool) -> Self {
         // For non-nullable structs, the flag is meaningless - keep it false
         if !self.dtype.is_nullable() {
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -3,6 +3,7 @@
 
 use std::fmt::Debug;
 use std::iter::once;
+use std::ops::Not;
 use std::sync::Arc;
 
 use vortex_dtype::DType;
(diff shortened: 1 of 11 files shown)


vortex-array/src/arrays/struct_/array.rs
```diff
@@ -18,10 +19,23 @@ use vortex_error::vortex_err;
...
+use crate::compute::mask;
 use crate::stats::ArrayStats;
 use crate::validity::Validity;
 use crate::vtable::ValidityHelper;
 
+/// Metadata for StructArray serialization.
+#[derive(Clone, prost::Message)]
+pub struct StructMetadata {
+    /// If true, child validity is a superset of struct validity (validity was pushed down).
+    /// For nullable children, their validity already includes struct nulls. For non-nullable
```
a10y, line 32: why only push this into null children why not all ?

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -157,12 +174,74 @@ pub struct StructArrayParts {
...
+        let name = name.as_ref();
+        self.field_by_name_opt(name)?.ok_or_else(|| {
+            vortex_err!(
+                "Field {name} not found in struct array with names {:?}",
+                self.names()
+            )
+        })
+    }
+
+    pub fn field_by_name_opt(&self, name: impl AsRef<str>) -> VortexResult<Option<ArrayRef>> {
```
a10y, line 226: doc comment that these are masked fields?

vortex-array/src/arrays/struct_/array.rs
joseph-isaacs: It is unneeded (free to add on read) and changes the dtype of the child.

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -473,4 +554,128 @@ impl StructArray {
...
+        }
+        self.validity_pushed_down
+    }
+
+    /// Checks that the validity_pushed_down invariant holds.
+    ///
+    /// When `validity_pushed_down` is true, for every nullable child field,
+    /// the child's validity must be a superset of the struct's validity.
+    /// That is, wherever the struct is invalid (null), the child must also be invalid.
+    fn check_validity_pushed_down(&self) -> VortexResult<()> {
```
a10y: can we call this `validate_validity_pushed_down` instead?

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -473,4 +554,128 @@ impl StructArray {
...
+    }
+
+    /// Set the validity_pushed_down flag.
+    ///
+    /// For non-nullable structs, this is a no-op (flag stays false) since there's no validity to
+    /// push down
+    ///
+    /// For nullable structs, setting this to true indicates that child validity
+    /// is a superset of struct validity (children include struct's nulls).
+    pub fn with_validity_pushed_down(mut self, validity_pushed_down: bool) -> Self {
```
a10y: this should be `unsafe` because the caller enforces this and if they're wrong it leads to incorrect results? similar to `with_zero_copy_to_list` in ListView

joseph-isaacs: I will do this, but its annoying we mix up real memory unsoundness with result invalidity

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -157,12 +174,78 @@ pub struct StructArrayParts {
...
+                if !self.dtype.is_nullable() {
+                    return Ok(field);
+                }
+                // Non-nullable field: always apply struct validity (even with validity_pushed_down,
+                // since we can't push validity to non-nullable fields)
+                if !field.dtype().is_nullable() {
+                    return field.mask(self.validity().to_array(self.len));
+                }
+                // Nullable field: return as-is (validity is either in the field or was pushed down)
+                Ok(field)
```
connortsui20, line 242: I feel like this is actually a larger semantic change than it seems, because we would then not have this logic for the other nested types (list, list view, and fixed size list). For list and fsl maybe this makes sense. But the scariest example would be list view, what does pushing down validity even mean for list view given there can be overlaps in the views?

I won't block this since it seems like we need it for other things(?), but it's something we should at least think about.

connortsui20: At the very least, I don't think it should be pushed down by default, only when you request a field with validity pushed down

joseph-isaacs: I would suggest we do this everywhere, since you can never read back the value.

PR #5925, opened 2026-01-12
feat[array]: use execute in canonical

joseph-isaacs:
Remove the `VORTEX_OPERATORS` remove the old `to_canonical` method and using `execute::<E>` in its place.

This cause CPU compute to return canonical (Not `ArrayRef` as before). see https://github.com/vortex-data/vortex/pull/5895

### DIFF
--- a/vortex-array/src/lib.rs
+++ b/vortex-array/src/lib.rs
@@ -69,12 +69,6 @@ pub mod flatbuffers {
     pub use vortex_flatbuffers::array::*;
 }
 
-static USE_VORTEX_OPERATORS: LazyLock<bool> = LazyLock::new(|| {
-    std::env::var("VORTEX_OPERATORS")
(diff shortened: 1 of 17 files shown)


joseph-isaacs:
These codspeed regressions are comparisons with a constant. I will investigate. Likely we need a execute_parent here

gatesn (approved):

vortex-array/src/lib.rs
```diff
@@ -69,12 +69,6 @@ pub mod flatbuffers {
     pub use vortex_flatbuffers::array::*;
 }
 
-static USE_VORTEX_OPERATORS: LazyLock<bool> = LazyLock::new(|| {
```
0ax1, line 72: 👍

0ax1 (approved):
🎉

PR #5928, opened 2026-01-12
fix[duckdb]: do not panic on operations that cannot be pushed-down

### LINKED ISSUE #5927: Vortex should not panic when an operator is not able to be pushed down.
### Discussed in https://github.com/vortex-data/vortex/discussions/5911

The following code should not panic. Vortex should indicate it does not support the NOT IN operator and allow DuckDB to evaluate it.

```
INSTALL vortex;
CALL dsdgen(sf = 1);
LOAD vortex;

COPY (SELECT * FROM web_sales) TO example.vortex (FORMAT vortex);
SELECT * FROM read_vortex('example.vortex') WHERE ws_sold_date_sk NOT IN (2452230);
```
```
thread '<unnamed>' (725647) panicked at vortex-duckdb/src/convert/expr.rs:135:18:
not yet implemented: operator DUCKDB_VX_EXPR_TYPE_COMPARE_NOT_IN
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5, aborting
```

danking:
Resolves #5927

### DIFF
--- a/vortex-duckdb/src/convert/expr.rs
+++ b/vortex-duckdb/src/convert/expr.rs
@@ -4,6 +4,7 @@
 use std::sync::Arc;
 
 use itertools::Itertools;
+use tracing::debug;
 use vortex::compute::BetweenOptions;
 use vortex::compute::LikeOptions;
 use vortex::compute::StrictComparison;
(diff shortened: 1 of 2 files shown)


vortex-duckdb/src/convert/expr.rs
a10y, line 188: do we need the same thing here?

a10y: actually this is probably unreachable, error seems fine. maybe even panic

0ax1 (approved):

PR #5929, opened 2026-01-12
switch Buffer -> BufferHandle in PrimitiveArray

a10y:
Fleshes out the BufferHandle type more.

I hid the inner enum and introduced some extra methods to build/unwrap BufferHandles.

This removes the `PrimitiveArray::as_slice` method, replacing it with an `into_buffer::<T>` method which _may_ allocate and copy a new buffer (if the handle points to device memory), or else provides a new Buffer pointing to the existing host memory.

This PR is large but a lot of it is just adding `&`'s and updating test code to use the `buffer!` macro

### DIFF
--- a/vortex-array/src/arrays/primitive/array/conversion.rs
+++ b/vortex-array/src/arrays/primitive/array/conversion.rs
@@ -19,7 +19,6 @@ use crate::ArrayRef;
 use crate::IntoArray;
 use crate::arrays::PrimitiveArray;
 use crate::validity::Validity;
-use crate::vtable::ValidityHelper;
 
 impl PrimitiveArray {
     /// Attempts to create a `PrimitiveArray` from a [`PrimitiveVector`] given a [`Nullability`].
@@ -72,7 +71,10 @@ impl PrimitiveArray {
         Self::new(values.freeze(), Validity::from(validity.freeze()))
     }
 
-    pub fn buffer<T: NativePType>(&self) -> Buffer<T> {
+    /// Get a buffer in host memory holding all the values.
+    ///
+    /// NOTE: some values may be nonsense if the validity buffer indicates that the value is null.
+    pub fn to_buffer<T: NativePType>(&self) -> Buffer<T> {
         if T::PTYPE != self.ptype() {
             vortex_panic!(
                 "Attempted to get buffer of type {} from array of type {}",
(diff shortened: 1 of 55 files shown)


joseph-isaacs:
I wonder if we can remove at lot of these buffer calls from tests?

a10y:
I don't think there's another way to get the values out now

vortex-array/src/arrays/primitive/vtable/mod.rs
```diff
@@ -95,12 +92,6 @@ impl VTable for PrimitiveVTable {
 
         let ptype = PType::try_from(dtype)?;
 
-        if !buffer.is_aligned(Alignment::new(ptype.byte_width())) {
-            vortex_bail!(
-                "Buffer is not aligned to {}-byte boundary",
-                ptype.byte_width()
-            );
-        }
```
joseph-isaacs, line 103: can we keep this check for host buffers only?

vortex-array/src/arrays/primitive/vtable/operations.rs
```diff
@@ -15,18 +15,21 @@ use crate::vtable::ValidityHelper;
...
-                array.validity().slice(range),
-            )
-            .into_array()
-        })
+        let stride = array.ptype().byte_width();
+        let start = range.start * stride;
+        let end = range.end * stride;
+
+        PrimitiveArray::from_buffer_handle(
+            array.buffer.slice(start..end),
```
joseph-isaacs: should we have a slice typed method on the buffer handle, that does the calculation?

vortex-array/src/arrays/primitive/array/conversion.rs
```diff
@@ -72,56 +71,44 @@ impl PrimitiveArray {
...
     /// Extract a mutable buffer from the PrimitiveArray. Attempts to do this with zero-copy
     /// if the buffer is uniquely owned, otherwise will make a copy.
     pub fn into_buffer_mut<T: NativePType>(self) -> BufferMut<T> {
-        if T::PTYPE != self.ptype() {
-            vortex_panic!(
-                "Attempted to get buffer_mut of type {} from array of type {}",
-                T::PTYPE,
-                self.ptype()
-            )
-        }
```
joseph-isaacs, line 106: why remove this?

robert3005: Looks like it was happening twice

joseph-isaacs (approved):

gatesn (dismissed):
I don't think this is quite right

PR #5935, opened 2026-01-13
Attempting to simplify the IO read traits

AdamGS:
We currently have 3 traits (`IntoReadSource`, `ReadSource` and `VortexReadAt`), each implemented for different types, but with a large overlap between them. At least for me, it isn't obvious how to extend it for a custom type or how they are relate to each other.

This PR is an attempt at trying to refactor the world of IO traits and types, in order to make it easier for users to provide their own specialized IO objects for observability/caching or any other reason.

I've tried to reduce the number of things a user needs to care and know about, ideally leaving us with a smaller API surface.


vortex-file/src/open.rs
```diff
@@ -139,10 +138,19 @@ impl VortexOpenOptions {
...
+    pub async fn open(self, source: Arc<dyn VortexRead>) -> VortexResult<VortexFile> {
         let handle = self.session.handle();
         let metrics = self.metrics.clone();
-        self.open_read_at(handle.open_read(source, metrics)?).await
+        self.open_read(handle.open_read(source, metrics)).await
+    }
+
+    /// Open a Vortex file from a filesystem path.
+    #[cfg(not(target_arch = "wasm32"))]
+    pub async fn open_path(self, path: impl AsRef<std::path::Path>) -> VortexResult<VortexFile> {
```
AdamGS, line 147: I'm not super happy with this API, but I'm not sure how to explain why.

vortex-io/src/file/object_store.rs
```diff
@@ -99,79 +103,66 @@ impl ReadSource for ObjectStoreIoSource {
...
+                        range
+                    );
+
+                    buffer
                 }
-                .in_current_span();
-
-                async move { req.resolve(Compat::new(read).await) }
-            })
-            .map(move |f| self2.handle.spawn(f))
```
danking, line 172: This is lost, right? I'm not entirely sure why we needed it, but the new drive implementation doesn't spawn the read future, right?

danking (changes requested):
Count me in for a single I/O source trait!

vortex-io/src/file/object_store.rs
AdamGS: Yeah I'm not convinced its necessary. `reqwest` already spawns some work, and `object_store` is tightly coupled with `tokio` and has its own abstraction (`SpawnedReqwestConnector`) for spawning work on a dedicated runtime.

danking: 👍

danking (approved):
They can revoke my approval privileges if they don't like my yolo approvals

vortex-io/src/read.rs
```diff
@@ -16,95 +18,144 @@ use vortex_metrics::Histogram;
...
+
+    /// Coalescing configuration. Returns `None` to disable coalescing.
+    fn coalesce_config(&self) -> Option<CoalesceConfig> {
+        None
+    }
+
+    /// Concurrency hint for the default driver implementation.
+    fn concurrency(&self) -> usize {
+        16
+    }
```
joseph-isaacs: this is the number of concurrent `read_at` calls per VortexRead impl instance?

vortex-io/src/read.rs
```diff
@@ -16,95 +18,144 @@ use vortex_metrics::Histogram;
...
+pub trait VortexRead: Send + Sync + 'static {
+    /// URI for debugging/logging. Returns `None` for anonymous sources.
+    fn uri(&self) -> Option<&Arc<str>> {
+        None
+    }
+
+    /// Coalescing configuration. Returns `None` to disable coalescing.
+    fn coalesce_config(&self) -> Option<CoalesceConfig> {
+        None
+    }
```
joseph-isaacs, line 57: I am interested why you think we should put this here. I would this the is a property of the underlying file system not the API to read from it. ebs vs local disk would be a good example.

vortex-io/src/read.rs
AdamGS: yeah, its the buffering of the `drive` stream. IDK if it needs a default value, this is just a random value.

vortex-io/src/read.rs
AdamGS: My assumption is that most systems will have their own `VortexRead` implementation for some adapter type (maybe we should provide some?) or some other custom type, and they can tune it to their expected hardware.

vortex-io/src/read.rs
joseph-isaacs: Can we update the comment to reflect this. 

Also maybe we can remove the default?

vortex-io/src/read.rs
joseph-isaacs: I see you want this method to be modified by an adapter. I wonder if it makes sense to move from out of the trait and use composition?

AdamGS: More traits really increase the complexity IMO, I would really love to keep this in one trait.

vortex-io/src/read.rs
AdamGS: done

vortex-file/src/open.rs
```diff
@@ -323,8 +337,8 @@ mod tests {
             self.inner.read_at(offset, length, alignment)
         }
 
-        fn size(&self) -> BoxFuture<'static, VortexResult<u64>> {
-            self.inner.size()
+        fn concurrency(&self) -> usize {
+            16
```
gatesn: This is wrong?

gatesn (changes requested):
Give me some time to grok this..

AdamGS:
After talking to @gatesn offline, I moved a lot of the file-specific logic from `vortex-io` to `vortex-file`, got rid of `FileRead` and moved that logic into the `FileSegmentSource`, which is the most the most file-specific abstraction.

AdamGS:
Benchmarks results now seem fine, potentially some minor differences but everything seems to be much closer to the noise threshold.

vortex-file/src/open.rs
```diff
@@ -139,10 +138,17 @@ impl VortexOpenOptions {
     /// out-of-the-box performance. The underlying I/O system will continue to be optimised for
     /// different file systems and object stores so we encourage users to use this method
     /// whenever possible and file issues if they encounter problems.
-    pub async fn open<S: IntoReadSource>(self, source: S) -> VortexResult<VortexFile> {
+    pub async fn open(self, source: Arc<dyn VortexReadAt>) -> VortexResult<VortexFile> {
```
gatesn, line 141: Are you sure you don't want `open<R: TryInto<VortexReadAt>>(read: R)` and `open_read_at(Arc<dyn VortexReadAt>)`?

Should help reduce the public API diff

vortex-file/src/read/driver.rs
```diff
@@ -112,7 +116,23 @@ struct State {
...
+    metrics: StateMetrics,
+}
+
+struct StateMetrics {
+    individual_requests: Arc<Counter>,
+    coalesced_requests: Arc<Counter>,
+    num_requests_coalesced: Arc<Histogram>,
+}
+
+impl StateMetrics {
```
AdamGS, line 128: this is a fly-by change, no need to keep reaching for the registry that's behind a lock on every request.

vortex-file/src/open.rs
AdamGS: Its going to be a while before we release again, mind if I do it in a followup PR? There are a few things I want to try here but would love to merge this so I can move on to the python thing that came up.

AdamGS: It also shouldn't be `TryInto`, if it has some issue getting started it'll just fail.

gatesn (approved):

PR #5936, opened 2026-01-13
feat: cuda gpu skeleton

0ax1:
Provides the initial skeleton to register and run CUDA kernels. 

Note that `cudarc` does not need to be compiled conditionally as we defer symbol loading via `dynamic-loading` such that `vortex-cuda` can also be compiled on macOS.


vortex-cuda/src/session.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use crate::executor::CudaSupport;
+
+/// CUDA session for GPU accelerated execution.
+///
+/// Maintains a registry of CUDA kernel implementations for array encodings.
+#[derive(Debug, Default)]
+pub struct CudaSession {
+    executors: DashMap<ArrayId, &'static dyn CudaSupport>,
+}
```
joseph-isaacs, line 26: how does this connect to a vortex session?

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,230 @@
...
+            .synchronize()
+            .map_err(|e| vortex_err!("Failed to synchronize device: {}", e))
+    }
+}
+
+/// Support trait for CUDA-accelerated execution of arrays.
+///
+/// Implementations provide CUDA-specific execution for array encodings.
+#[async_trait]
+pub trait CudaSupport: 'static + Send + Sync + Debug {
```
joseph-isaacs: Shall we call this CudaExecute?

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,230 @@
...
+use vortex_error::vortex_err;
+
+use crate::session::CudaSession;
+
+/// CUDA execution context.
+///
+/// Provides access to the CUDA context and stream for kernel execution.
+/// Handles memory allocation and data transfers between host and device.
+#[derive(Clone)]
+pub struct ExecutionCtx {
```
joseph-isaacs: this should have a stream?

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    /// Synchronizes the stream
+    ///
+    /// On `synchronize` the host waits for all pending operations of the stream to complete.
+    pub fn synchronize(&self) -> VortexResult<()> {
+        self.context
+            .default_stream()
+            .synchronize()
+            .map_err(|e| vortex_err!("Failed to synchronize device: {}", e))
+    }
```
joseph-isaacs, line 108: we should prefer synch(event_id)

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,230 @@
...
+        if self.is_canonical() {
+            return Ok(self.to_canonical());
+        }
+
+        let Some(support) = ctx.session.get_executor(&self.encoding_id()) else {
+            tracing::debug!(
+                encoding = %self.encoding().id(),
+                "No CUDA support registered for encoding, falling back to CPU execution"
+            );
+            return Ok(self.to_canonical());
```
joseph-isaacs: use execute

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,230 @@
...
+    ///
+    /// * `array` - The array to execute
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if execution fails on the GPU.
+    pub async fn execute_canonical(&self, array: ArrayRef) -> VortexResult<Canonical> {
+        let ctx = self.create_context();
+        array.execute_cuda(&ctx).await
+    }
```
joseph-isaacs: no PLZ

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        tracing::info!(device_id = device_id, "CUDA executor initialized");
+
+        Ok(Self { context, session })
+    }
+
+    /// Creates a new execution context for this executor.
+    pub fn create_context(&self) -> ExecutionCtx {
+        ExecutionCtx::new(self.context.clone(), self.session.clone())
+    }
```
joseph-isaacs: create_execution_ctx

vortex-cuda/src/executor.rs
joseph-isaacs: This clashes with vortex_array::ExecutionCtx

vortex-cuda/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+pub use session::CudaSession;
+
+/// Registers CUDA kernels for a given CUDA session.
+pub fn initialize_cuda(session: &CudaSession) {
+    tracing::info!("Initializing CUDA support");
+
+    // Register CUDA kernel implementations for supported array encodings.
+    //
+    // session.register(BitPackedVTable::ID, &bitpacking::CUDA_SUPPORT);
+    // session.register(RLEVTable::ID, &rle::CUDA_SUPPORT);
```
joseph-isaacs: can we have for?

vortex-cuda/src/executor.rs
joseph-isaacs: `CudaExecutable`

vortex-cuda/src/lib.rs
0ax1: yaeh

vortex-cuda/src/executor.rs
0ax1: offline discussion: whilst convenient for initial development, in production we rarely want to do a blocking wait on cpu. therefore we now compile this fn conditinally for tests.

vortex-cuda/src/executor.rs
0ax1: Either is fine for me.

vortex-cuda/src/executor.rs
0ax1: ~~The context holds the CUDA streams.~~ Just the default stream. Other streams created from the context need to be maintained by the caller.

0ax1: added a mechanism to create new streams, and access them by index.

0ax1: Added `vortex_array::ExecutionCtx` as a field, and renamed it to CudaExection context.

vortex-cuda/src/session.rs
0ax1: `impl<S: SessionExt> CudaSessionExt for S {}`

vortex-cuda/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+pub use session::CudaSession;
+pub use session::CudaSessionExt;
+
+/// Registers CUDA kernels for a given CUDA session.
+pub fn initialize_cuda(session: &CudaSession) {
+    tracing::info!("Initializing CUDA support");
+
+    // Register CUDA kernel implementations for supported array encodings.
+    //
+    // session.register(BitPackedVTable::ID, &bitpacking::CUDA_SUPPORT);
```
gatesn, line 30: Just a note, this crate would only register cuda support for the arrays in vortex-array. Other support should live in encodings/* crates

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,165 @@
...
+    context: Arc<CudaContext>,
+    session: Arc<CudaSession>,
+    array_ctx: ExecutionCtx,
+    stream_counter: Arc<AtomicU64>,
+    streams: Arc<DashMap<u64, Arc<CudaStream>>>,
+}
+
+impl CudaExecutionCtx {
+    /// Creates a new CUDA execution context.
+    pub fn new(
```
gatesn: Do not make this pub. No one should be able to construct one of these by hand

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Allocates a typed buffer on the GPU.
+    pub fn alloc<T: DeviceRepr + ValidAsZeroBits>(&self, len: usize) -> VortexResult<CudaSlice<T>> {
+        self.context
+            .default_stream()
+            .alloc_zeros::<T>(len)
+            .map_err(|e| vortex_err!("Failed to allocate device memory: {}", e))
+    }
+
+    /// Copies data from host to device.
+    pub fn to_device<T: DeviceRepr>(&self, data: &[T]) -> VortexResult<CudaSlice<T>> {
```
gatesn, line 64: Should this be async?

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.context
+            .default_stream()
+            .clone_dtoh(buffer)
+            .map_err(|e| vortex_err!("Failed to copy from device: {}", e))
+    }
+
+    /// Creates a new CUDA stream with a unique index.
+    ///
+    /// Returns both the stream and its assigned index.
+    pub fn new_stream(&self) -> VortexResult<(u64, Arc<CudaStream>)> {
```
gatesn: Not sure this should be exposed.

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,165 @@
...
+            .context
+            .new_stream()
+            .map_err(|e| vortex_err!("Failed to create CUDA stream: {}", e))?;
+        self.streams.insert(idx, stream.clone());
+        Ok((idx, stream))
+    }
+
+    /// Returns a reference to the CUDA context.
+    pub fn context(&self) -> Arc<CudaContext> {
+        self.context.clone()
```
gatesn: Return `&Arc<CudaContext>`

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /// Retrieves a previously created stream by its index.
+    pub fn stream(&self, idx: u64) -> Option<Arc<CudaStream>> {
```
gatesn: I think we said each CudaExecutionCtx would hold a single stream, constructed during `CudaSession::new_execution_ctx()` ?

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.streams.get(&idx).map(|entry| entry.clone())
+    }
+
+    /// Synchronizes the stream
+    ///
+    /// On `synchronize` the host waits for all pending operations of the stream to complete.
+    pub fn synchronize(&self) -> VortexResult<()> {
```
gatesn, line 104: Perhaps heavy-handed? Would need to see how it's used

vortex-cuda/src/executor.rs
0ax1: done

vortex-cuda/src/executor.rs
0ax1: pub crate?

vortex-cuda/src/executor.rs
0ax1: Had gated this behind cfg(test) in the meantime.

vortex-cuda/src/executor.rs
0ax1: done

vortex-cuda/src/executor.rs
0ax1: What if we want to run multiple streams in parallel in that context?

vortex-cuda/src/executor.rs
0ax1: This is a bit of a weird one bc it might be already, depending on the capabilities of the device/driver.
```
    pub unsafe fn alloc<T: DeviceRepr>(
        self: &Arc<Self>,
        len: usize,
    ) -> Result<CudaSlice<T>, DriverError> {
        self.ctx.bind_to_thread()?;
        let cu_device_ptr = if self.ctx.has_async_alloc {
            result::malloc_async(self.cu_stream, len * std::mem::size_of::<T>())?
```

So whether or not the allocation happens async already (though without async keyword), depends on
```
      let has_async_alloc = unsafe {
            let memory_pools_supported = result::device::get_attribute(
                cu_device,
                sys::CUdevice_attribute_enum::CU_DEVICE_ATTRIBUTE_MEMORY_POOLS_SUPPORTED,
            )?;
            memory_pools_supported > 0
        };
```

vortex-cuda/src/lib.rs
0ax1: I'll add a doc note.

vortex-cuda/src/executor.rs
gatesn: I actually think this shouldn't exist on a Ctx at all. I don't want kernels to be able to create new streams. It can be pub(crate) on the CudaSession?

vortex-cuda/src/executor.rs
gatesn: We said we don't. Parallelism comes at the scan level (for now...). In the same way that CPU/Host execution is entirely single-threaded.

0ax1: Settled on single stream per ctx, but not the default stream, as this would be shared across ctxs.

vortex-cuda/src/executor.rs
0ax1: removed, as we only need one stream per executionctx

vortex-cuda/src/executor.rs
0ax1: We don't need to make this on the Rust side therefore.

0ax1: I added:
```
                // Note that alloc is async in case the device and driver support this.
                //
                // The condition for alloc to be async is support for memory pools:
                // `CU_DEVICE_ATTRIBUTE_MEMORY_POOLS_SUPPORTED`. Any kernel
                // submitted to the stream after alloc can safely use the memory,
                // as operations on the stream are ordered sequentially.
```

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,148 @@
...
+/// CUDA execution context.
+///
+/// Provides access to the CUDA context and stream for kernel execution.
+/// Handles memory allocation and data transfers between host and device.
+pub struct CudaExecutionCtx {
+    #[allow(dead_code)] // TODO(0ax1)
+    context: Arc<CudaContext>,
+    session: Arc<CudaSession>,
+    array_ctx: vortex_array::ExecutionCtx,
+    pub stream: Arc<CudaStream>,
```
gatesn: Nooo

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,148 @@
...
+
+impl CudaExecutionCtx {
+    /// Creates a new CUDA execution context.
+    #[allow(dead_code)] // TODO(0ax1)
+    fn new(
+        context: Arc<CudaContext>,
+        session: Arc<CudaSession>,
+        array_ctx: vortex_array::ExecutionCtx,
+    ) -> VortexResult<Self> {
+        let stream = context
```
gatesn: `new` should do no work. The CudaSession should hold the CudaContext, and `CudaSession::new_execution_ctx()` should create the stream

vortex-cuda/src/kernel.rs
```diff
@@ -0,0 +1,79 @@
...
+
+//! CUDA kernel configuration.
+
+use std::fmt::Debug;
+
+use vortex_array::vtable::ArrayId;
+use vortex_utils::aliases::dash_map::DashMap;
+
+/// Configuration for a CUDA kernel.
+#[derive(Debug, Clone)]
```
0ax1, line 12: Might drop this file in a follow up but wanted to capture somewhere with which parameters we want to run a kernel.

vortex-cuda/src/executor.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    /// Copies data from host to device.
+    pub fn to_device<T: DeviceRepr>(&self, data: &[T]) -> VortexResult<CudaSlice<T>> {
+        self.stream
+            .clone_htod(data)
+            .map_err(|e| vortex_err!("Failed to copy to device: {}", e))
+    }
+
+    /// Copies data from device to host.
+    pub fn to_host<T: DeviceRepr>(&self, buffer: &CudaSlice<T>) -> VortexResult<Vec<T>> {
```
gatesn: Take an `alignment` and return a `Buffer<T>`

0ax1: done

vortex-cuda/src/lib.rs
0ax1: lemme put the first kernel in a follow up to keep the pr compact.

gatesn (approved):

vortex-cuda/src/session.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::executor::CudaExecutionCtx;
+
+/// CUDA session for GPU accelerated execution.
+///
+/// Maintains a registry of CUDA kernel implementations for array encodings.
+/// Holds the CUDA context for all GPU operations.
+#[derive(Clone, Debug)]
+pub struct CudaSession {
+    context: Arc<CudaContext>,
+    kernels: Arc<DashMap<ArrayId, &'static dyn CudaExecute>>,
```
a10y, line 25: i don't think this needs to be arc'ed

0ax1: @a10y Ah missed this one. I'll double check.

0ax1: I actually want this to make it clonable.

PR #5941, opened 2026-01-13
chore: use `FilterArray` instead of `compute::filter`

joseph-isaacs:


vortex-array/src/arrays/filter/array.rs
```diff
@@ -34,4 +67,9 @@ impl FilterArray {
     pub fn filter_mask(&self) -> &Mask {
```
gatesn, line 67: I really think you need to hide this / name it better / something. It's too much of a foot-gun to expose the unsliced mask

PR #5945, opened 2026-01-13
Add scan concurrency config for datafusion source

AdamGS:
This PR adds a new configuration to the `FileSource` based API, allowing users to control the intra-partition scan concurrency.

### DIFF
--- a/vortex-datafusion/src/persistent/format.rs
+++ b/vortex-datafusion/src/persistent/format.rs
@@ -154,9 +154,9 @@ impl VortexFormatFactory {
     ///
     /// For example:
     /// ```rust
-    /// use vortex_datafusion::{VortexFormatFactory, VortexOptions};
+    /// use vortex_datafusion::{VortexFormatFactory, VortexTableOptions};
     ///
-    /// let factory = VortexFormatFactory::new().with_options(VortexOptions::default());
+    /// let factory = VortexFormatFactory::new().with_options(VortexTableOptions::default());
     /// ```
     pub fn with_options(mut self, options: VortexTableOptions) -> Self {
         self.options = Some(options);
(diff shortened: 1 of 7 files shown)


vortex-datafusion/src/persistent/format.rs
```diff
@@ -97,6 +97,8 @@ config_namespace! {
         /// Values smaller than `MAX_POSTSCRIPT_SIZE + EOF_SIZE` will be clamped to that minimum
         /// during footer parsing.
         pub footer_initial_read_size_bytes: usize, default = DEFAULT_FOOTER_INITIAL_READ_SIZE_BYTES
+        /// The per-file Vortex scan concurrency.
+        pub scan_concurrency: Option<usize>, default = None
```
a10y, line 108: doesn't DF let you set this for an entire sessioncontext? do we want to override this on a per-source basis?

AdamGS: I think that's the copy of `VortexOptions` that is part of `VortexFormat`, which propagates it downstream from there through either `VortexFormatFactory::create` or `VortexFormat::file_source`

a10y: I was wondering if it makes sense to use the target_partitions config off the environment, but I realize that's different.

Maybe we can make it clear in the doc comment that this is the intra-partition concurrency

AdamGS: intra-partition was the term I was looking for!

a10y:
I really forgot about this. Think we still want this @AdamGS ?

AdamGS:
yeah I think so, I'll try and unconflict it later today 😢

a10y (approved):
pre-approving

PR #5949, opened 2026-01-14
Dataset API

gatesn:
# Dataset API

A unified API to sit between query engines and data sources that preserves support for late materialization, deferred decompression, and alternate device buffers.

# Open Questions

* What does capabilities negotiation look like? e.g. for unsupported encodings?
* What about shared dictionaries?
* Is deserialize_split in the right place? Without serde, splits can share state in the DataSourceScan. With serde, there's no where for shared state to live. Perhaps we should reconstruct a datasource on the worker? Should the datasource be serializable?


vortex-dataset/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The API passes data in the form of Vortex ArrayStreams, which are zero-copy, columnar, compressed, and even support
+passing device buffers. This allows query engines to efficiently scan data into their internal execution formats with
+minimal overhead and (very) late materialization.
+
+Known implementations of the Vortex Dataset API are:
+
+* `vortex-iceberg` - Expose Iceberg tables as a Vortex Dataset
+* `vortex-python` - Expose PyArrow Datasets as a Vortex Dataset
+* `vortex-layout` - Expose a Vortex Layout as a Vortex Dataset
```
danking: we should definitely write a vortex-parquet

vortex-scan/src/api.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Returns an estimate of the row count of the source.
+    fn row_count_estimate(&self) -> Estimate<u64>;
+
+    /// Returns a scan over the source.
+    async fn scan(&self, scan_request: ScanRequest) -> VortexResult<DataSourceScanRef>;
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct ScanRequest {
+    pub projection: Option<Expression>,
```
danking, line 79: I don't love the fact that, in theory, someone could write Spark SQL which gets compiled to vortex Expression which gets compiled to PyArrow Expression.

Hunting a semantic-bug across those two compilers gives me the heebie jeebies but I don't have an alternative solution! It's better to hunt bugs in O(N+M) integrations than O(NM), assuming I care about all the integrations.

vortex-scan/src/v2/source.rs
```diff
@@ -0,0 +1,98 @@
...
+pub trait DataSourceScan: 'static + Send + Sync {
+    /// The returned dtype of the scan.
+    fn dtype(&self) -> &DType;
+
+    /// An estimate of the remaining splits.
+    fn remaining_splits_estimate(&self) -> Estimate<usize>;
+
+    /// Returns the next batch of splits to be processed.
+    ///
+    /// This should not return _more_ than the max_batch_size splits, but may return fewer.
```
danking: max_batch_size isn't defined anywhere; should that be in ScanRequest?

vortex-scan/src/v2/source.rs
```diff
@@ -0,0 +1,98 @@
...
+use vortex_error::VortexResult;
+
+/// Create a Vortex source from serialized configuration.
+///
+/// Providers can be registered with Vortex under a specific
+#[async_trait(?Send)]
+pub trait DataSourceProvider: 'static {
+    /// URI schemes handled by this source provider.
+    ///
+    /// TODO(ngates): this might not be the right way to plugin sources.
```
danking: Yeah, hmm, how do I read a PyArrow dataset? That's something sitting in memory in Python.

I'm not sure [how duckdb does it](https://duckdb.org/docs/stable/guides/python/sql_on_arrow#apache-arrow-datasets), but I assume it just tries every other source and then, failing that, starts rummaging around in the Python variable environment?

Suppose I'm writing neoduckdb and I want to support the same magic, but I'm delegating to Vortex for *all* my scanning. I feel like `init_source` should return `VortexResult<Option<DataSourceRef>>` so that my engine can check if a data source exists? We don't usually use VortexResult for recoverable errors.

vortex-scan/src/v2/source.rs
```diff
@@ -0,0 +1,98 @@
...
+    fn schemes(&self) -> &[&str];
+
+    /// Initialize a new source.
+    async fn init_source(&self, uri: String) -> VortexResult<DataSourceRef>;
+
+    /// Serialize a source split to bytes.
+    async fn serialize_split(&self, split: &dyn Split) -> VortexResult<Vec<u8>>;
```
danking: I guess the Provider just assumes the split is a type it knows and downcasts it? Kinda seems like Providers should have a `Split` associated type and their DataSources & Scans are parameterized by that type?

danking: Or maybe splits can just serialize themselves? What's the case where a Split lacks sufficient information to serialize itself?

danking: I see you raised this above in your PR comment.

Yeah I kinda think the source and the split should be serializable. The source can play the role of Spark's broadcast whereas the split can play the role of Spark's Partition.

vortex-scan/src/datasource.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// URI schemes handled by this source provider.
+    ///
+    /// TODO(ngates): this might not be the right way to plugin sources.
+    fn schemes(&self) -> &[&str];
+
+    /// Initialize a new source.
+    async fn init_source(&self, uri: String) -> VortexResult<DataSourceRef>;
+
+    /// Serialize a source split to bytes.
+    async fn serialize_split(&self, split: &dyn Split) -> VortexResult<Vec<u8>>;
```
a10y: this is the plan not the split data right?

gatesn: Correct

joseph-isaacs (approved):

PR #5952, opened 2026-01-14
Run codspeed memory benchmarks

AdamGS:
This adds less then a minute to CI and seems super useful, see results https://codspeed.io/vortex-data/vortex/branches/adamg%2Frun-memory-benchmarks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -632,13 +632,17 @@ jobs:
       - name: Build benchmarks
         env:
           RUSTFLAGS: "-C target-feature=+avx2 -C debug-assertions=yes"
-        run: cargo codspeed build ${{ matrix.features }} $(printf -- '-p %s ' ${{ matrix.packages }}) --profile bench
+        run: cargo codspeed build -m simulation ${{ matrix.features }} $(printf -- '-p %s ' ${{ matrix.packages }}) --profile bench
       - name: Run benchmarks
         uses: CodSpeedHQ/action@99c1668ae56776024208bba66011679f7f844a7e
         with:
           run: cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
           mode: "simulation"
+      - name: Build memory benchmarks
+        env:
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

adriencaccia:
Hey @AdamGS, glad to have using the new memory instrument 😉 

However there is an issue with your current setup, at the moment all the CPU simulation benchmarks are skipped.
This is because in this PR you ran the CodSpeedHQ/action with `mode: memory` in the same job of the one with `mode: simulation`. We do not support that at the moment, we need to have different jobs for different instruments.

The fix will be to have `mode: [simulation, memory]` in your matrix definition and use that value in the build and run steps.

AdamGS:
Thank you for the heads up! I'll revert this change and do it right when I have the bandwidth

adriencaccia:
> Thank you for the heads up! I'll revert this change and do it right when I have the bandwidth

Hey @AdamGS, we just released a new feature that supports consecutive runs, so you can re-enable memory benchmarks if you want!
Another thing, the builds for `simulation` and `memory` are the same, so you do not need to do it twice.

The following should do the trick:

```yaml
      - name: Build benchmarks
        env:
          RUSTFLAGS: "-C target-feature=+avx2 -C debug-assertions=yes"
        run: cargo codspeed build ${{ matrix.features }} $(printf -- '-p %s ' ${{ matrix.packages }}) --profile bench
      - name: Run benchmarks
        uses: CodSpeedHQ/action@e736f0d2aeb36da38e9f08eca4dff7967408d154 # 4.8.2
        with:
          run: cargo codspeed run
          token: ${{ secrets.CODSPEED_TOKEN }}
          mode: "simulation"
      - name: Run memory benchmarks
        uses: CodSpeedHQ/action@e736f0d2aeb36da38e9f08eca4dff7967408d154 # 4.8.2
        with:
          run: cargo codspeed run
          token: ${{ secrets.CODSPEED_TOKEN }}
          mode: "memory"
```

PR #5957, opened 2026-01-14
chore[ci]: don't build lance always

joseph-isaacs:


benchmarks/lance-bench/Cargo.lock
robert3005, line 1: Does cargo generate this file when you change default-members?

joseph-isaacs: I guess so?

.github/workflows/bench.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         env:
           RUSTFLAGS: "-C target-cpu=native -C force-frame-pointers=yes"
         run: |
-          cargo build --bin ${{ matrix.benchmark.id }} --profile release_debug --features lance
```
AdamGS, line 70: this is the nightly flow where we do want to benchmark against lance right? if you remove it here, we're basically not benchmarking lance anywhere.

.github/workflows/bench.yml
```diff
@@ -107,16 +107,14 @@ jobs:
             "id": "clickbench-nvme",
             "subcommand": "clickbench",
             "name": "Clickbench on NVME",
-            "targets": "datafusion:parquet,datafusion:vortex,datafusion:vortex-compact,datafusion:lance,duckdb:parquet,duckdb:vortex,duckdb:vortex-compact,duckdb:duckdb",
```
AdamGS, line 110: same as the point above

PR #5958, opened 2026-01-14
feat[arary]: add a slice array to replace the slice op.

joseph-isaacs:
A follow up is to remove the slice method from the `OperationsVTable`


encodings/fastlanes/src/bitpacking/vtable/mod.rs
```diff
@@ -251,6 +257,40 @@ impl VTable for BitPackedVTable {
...
+    fn reduce_parent(
+        array: &Self::Array,
+        parent: &ArrayRef,
+        child_idx: usize,
+    ) -> VortexResult<Option<ArrayRef>> {
+        RULES.evaluate(array, parent, child_idx)
+    }
+
+    fn slice(array: &Self::Array, range: Range<usize>) -> VortexResult<Option<ArrayRef>> {
+        let offset_start = range.start + array.offset() as usize;
```
gatesn, line 271: I would have thought for FastLanes that slice we round the start/end to 1024, hard-slice the encoded buffer, then leave the remaining start/end as a SliceArray on top. So remove the array offset. 

Happy for that to happen later, but does that make sense?

encodings/fsst/src/array.rs
```diff
@@ -189,6 +191,26 @@ impl VTable for FSSTVTable {
...
+        Ok(Some(
+            unsafe {
+                FSSTArray::new_unchecked(
+                    array.dtype().clone(),
+                    array.symbols().clone(),
+                    array.symbol_lengths().clone(),
+                    array
+                        .codes()
+                        .slice(range.clone())
+                        .as_::<VarBinVTable>()
```
gatesn, line 206: This seems sketchy. I think we may have to store the offsets/lens separately rather than holding a VarBinArray. Unclear. Or maybe we can downcast, then slice, then wrap up again to know we still have a VarBinArray.

encodings/runend/src/array.rs
```diff
@@ -81,6 +85,30 @@ impl VTable for RunEndVTable {
         RunEndVTable.as_vtable()
     }
 
+    fn slice(array: &Self::Array, range: Range<usize>) -> VortexResult<Option<ArrayRef>> {
+        let new_length = range.len();
+
+        let slice_begin = array.find_physical_index(range.start);
```
gatesn: This seems wrong? I don't quite understand the scope of this PR? Are we supposed to be unwinding the slicing doing compute?

encodings/sparse/src/lib.rs
```diff
@@ -164,6 +165,28 @@ impl VTable for SparseVTable {
...
+            ));
+        };
+
+        // If the number of values in the sparse array matches the array length, then all
+        // values are in fact patches, since patches are sorted this is the correct values.
+        if new_patches.array_len() == new_patches.values().len() {
+            return Ok(Some(new_patches.into_values()));
+        }
+
+        // SAFETY:
```
gatesn, line 184: Safety what ??

vortex-array/src/array/mod.rs
```diff
@@ -499,7 +500,10 @@ impl<V: VTable> Array for ArrayAdapter<V> {
             return Canonical::empty(self.dtype()).into_array();
         }
 
-        let sliced = <V::OperationsVTable as OperationsVTable<V>>::slice(&self.0, range);
+        let sliced = SliceArray::new(self.0.to_array(), range)
```
gatesn: Is it worth calling the vtable here anyway? To avoid allocating a SliceArray just to optimize it away?

vortex-array/src/vtable/mod.rs
```diff
@@ -186,6 +187,22 @@ pub trait VTable: 'static + Sized + Send + Sync + Debug {
         _ = (array, parent, child_idx);
         Ok(None)
     }
+
+    /// Perform a constant-time slice of the array.
+    ///
+    /// If an encoding cannot perform this slice in constant time, it should instead
+    /// wrap itself in the `SliceArray`.
```
gatesn: This isn't what the array implementing this function should do. It should return `Ok(None)` right?

gatesn (changes requested):

encodings/fastlanes/src/bitpacking/vtable/mod.rs
joseph-isaacs: added a todo

encodings/fsst/src/array.rs
joseph-isaacs: Wdyt here. I downcasted before slicing

gatesn (approved):

PR #5961, opened 2026-01-14
fix: canonicalize VarBin/FSST with >2GB buffers

a10y:
Fixes a potential bug where decoding a large FSST and VarBin arrays results in an invalid VarBinViewArray.

When you have a large buffer that is, currently we generate a new VBV with the single buffer plus some views built against it. There will be trouble if the buffer is > 2GiB though.

This PR splits out a separate `build_views` function that takes a `max_buffer_len` parameter and as it generates views, it splits (zero-copy) the underlying buffer into segments of no more than `max_buffer_len`.

### DIFF
--- a/encodings/fsst/src/canonical.rs
+++ b/encodings/fsst/src/canonical.rs
@@ -3,17 +3,16 @@
 
 use std::sync::Arc;
 
-use num_traits::AsPrimitive;
 use vortex_array::Canonical;
 use vortex_array::ExecutionCtx;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::arrays::VarBinViewArray;
+use vortex_array::arrays::build_views::MAX_BUFFER_LEN;
+use vortex_array::arrays::build_views::build_views;
 use vortex_array::vtable::ValidityHelper;
 use vortex_buffer::Buffer;
-use vortex_buffer::BufferMut;
 use vortex_buffer::ByteBuffer;
 use vortex_buffer::ByteBufferMut;
-use vortex_dtype::NativePType;
 use vortex_dtype::match_each_integer_ptype;
 use vortex_error::VortexResult;
 use vortex_vector::binaryview::BinaryView;
@@ -37,9 +36,6 @@ pub(super) fn canonicalize_fsst(
     })
 }
 
-/// Maximum number of buffer bytes that can be referenced by a single `BinaryView`
-const MAX_BUFFER_LEN: usize = i32::MAX as usize;
-
 pub(crate) fn fsst_decode_views(
     fsst_array: &FSSTArray,
     start_buf_index: u32,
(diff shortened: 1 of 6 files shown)


encodings/fsst/src/canonical.rs
```diff
@@ -84,24 +88,57 @@ fn fsst_decode_views(fsst_array: &FSSTArray, buf_index: u32) -> (ByteBuffer, Buf
...
+
+    let mut buffers = Vec::new();
+    let mut buf_index = start_buf_index;
+
+    let mut offset = 0;
+    for &len in uncompressed_lens {
+        let len = len.as_();
+        assert!(len <= max_buffer_len, "values cannot exceed max_buffer_len");
+
+        if (offset + len) > max_buffer_len {
```
a10y: This is actually very conservative, and doesn't account for the presence of inlined strings.

There isn't really a way to get around this without lots of data copying

a10y: (FWIW this is already a problem on develop as well, except there everything lives in one giant buffer which sometimes is too large for VarBinView)

encodings/fsst/src/canonical.rs
```diff
@@ -44,17 +46,19 @@ impl CanonicalVTable<FSSTVTable> for FSSTVTable {
...
     }
 }
 
-#[expect(
-    clippy::cast_possible_truncation,
-    reason = "truncation is intentional for buffer index"
-)]
-fn fsst_decode_views(fsst_array: &FSSTArray, buf_index: u32) -> (ByteBuffer, Buffer<BinaryView>) {
+/// Maximum number of buffer bytes that can be referenced by a single `BinaryView`
+const MAX_BUFFER_LEN: usize = i32::MAX as usize;
```
AdamGS: is this a Java heresy?

a10y: Yea, there's some ambiguity here with the spec and arrow-rs diverges from arrow-cpp

Arrow-rs uses u32 for all of these fields.

But, the spec states that all offsets should be treated as signed. That's how the Java implementation works, and also how the arrow-cpp implementation works https://github.com/apache/arrow/blob/7820f672edbbd516661740db9c355f2bc42bf602/cpp/src/arrow/util/binary_view_util.h#L52-L58

So just to be on the safe side I use i32 here instead of u32

vortex-array/src/arrays/varbinview/build_views.rs
```diff
@@ -0,0 +1,109 @@
...
+#[cfg(test)]
+mod tests {
+    use vortex_buffer::ByteBuffer;
+    use vortex_buffer::ByteBufferMut;
+    use vortex_vector::binaryview::BinaryView;
+
+    use crate::arrays::build_views::build_views;
+
+    #[test]
+    fn test_to_canonical_large() {
```
joseph-isaacs, line 75: is this the test that takes a while to run?

joseph-isaacs (approved):

vortex-array/src/arrays/varbinview/build_views.rs
a10y: no this one is very small b/c i artificially constrain the max buffer len

a10y: but now that we're hitting the same codepath for varbin/fsst it's ok

PR #5966, opened 2026-01-15
feat: CUDA compilation infra & FoR kernel

0ax1:


vortex-cuda/src/for_.rs
```diff
@@ -0,0 +1,146 @@
...
+    use vortex_error::VortexExpect;
+    use vortex_fastlanes::FoRArray;
+
+    use super::*;
+    use crate::has_nvcc;
+    use crate::session::CudaSession;
+
+    #[tokio::test]
+    async fn test_cuda_for_decompression() {
+        if !has_nvcc() {
```
0ax1, line 115: I'll feature flag this down the road but for dev it's nice to runtime guard.

vortex-cuda/src/for_.rs
```diff
@@ -0,0 +1,146 @@
...
+            return;
+        }
+
+        let vortex_session = vortex_session::VortexSession::empty()
+            .with::<ArraySession>()
+            .with::<CudaSession>();
```
0ax1: Strictly speaking we don't need the CUDA session here, as we just call into `execute_for` directly. 

In general, where should/will
```
pub fn initialize_cuda(session: &CudaSession) {
    tracing::info!("Registering CUDA kernels");
    session.register_kernel("fastlanes.for".into(), &ForExecutor);
    // TODO(0ax1): Register additional executors
}
```
be called actually? 

Should that down the line live in 
```
    /// Creates a new CUDA execution context.
    pub fn new_ctx(
        vortex_session: vortex_session::VortexSession,
    ) -> VortexResult<CudaExecutionCtx> {
        let stream = vortex_session
            .cuda_session()
            .context
            .new_stream()
            .map_err(|e| vortex_err!("Failed to create CUDA stream: {}", e))?;
        Ok(CudaExecutionCtx::new(stream, vortex_session))
    }
```

vortex-cuda/src/for_.rs
```diff
@@ -0,0 +1,146 @@
...
+    // Recursively decompresss the child.
+    let unpacked_canonical = encoded.execute_cuda(ctx).await?;
+
+    let unpacked_array = unpacked_canonical.as_primitive();
+    let unpacked_slice = unpacked_array.as_slice::<P>();
+
+    // TODO(0ax1): Check whether buffer is already on device.
+    let device_data = ctx.to_device(unpacked_slice)?;
+
+    let array_len = array.len() as u64;
```
joseph-isaacs, line 74: what is this doesn't divide 1024?

vortex-cuda/src/for_.rs
```diff
@@ -0,0 +1,146 @@
...
+    launch_cuda_kernel!(
+        execution_ctx: ctx,
+        module: "for",
+        ptypes: &[array.ptype()],
+        launch_args: [device_data, reference, array_len],
+        array_len: array.len()
+    );
```
joseph-isaacs, line 81: can we make this a function and use a macro for launch args?

vortex-cuda/kernels/for.cu
```diff
@@ -0,0 +1,46 @@
...
+    uint64_t array_len
+) {
+    // Each block handles 1024 elements. Each thread handles 1024 / 32 contiguous elements.
+    const uint8_t elements_per_thread = 32;
+    const uint32_t block_start = blockIdx.x * 1024;
+    const uint32_t thread_start = block_start + threadIdx.x * elements_per_thread;
+
+    for (uint8_t j = 0; j < elements_per_thread; ++j) {
+        const uint32_t idx = thread_start + j;
+        if (idx < array_len) {
```
joseph-isaacs: we cannot do this?

vortex-cuda/src/for_.rs
0ax1: no

vortex-cuda/kernels/for.cu
0ax1: you mean the comparison? yeah for perf we probably want to not do that.

vortex-cuda/src/for_.rs
0ax1: feel free to give it try on your end

vortex-cuda/src/for_.rs
0ax1: we're doing this consistently in v1 for all kernels. if we add a kernel that doesn't align with this we'll just move it up.

vortex-cuda/kernels/for.cu
0ax1: Swapped for
```
    const uint32_t start = thread_start;
    const uint32_t end = std::min(start + elements_per_thread, array_len);

    for (uint8_t idx = start; idx < end; ++idx) {
        values_in_out_array[idx] = values_in_out_array[idx] + reference;
    }
```

vortex-cuda/src/for_.rs
0ax1: as discussed offline, clean up later.

vortex-cuda/src/executor.rs
```diff
@@ -5,42 +5,93 @@ use std::fmt::Debug;
...
+        $(
+            launch_builder.arg(&$arg);
+        )*
+
+        let num_chunks = u32::try_from($len.div_ceil(1024))
+            .vortex_expect("Too many elements for grid");
+
+        let config = cudarc::driver::LaunchConfig {
+            grid_dim: (num_chunks, 1, 1),
+            block_dim: (32, 1, 1),
```
robert3005: What I have realised previously is that we want bigger blocks. You have 2048 threads per SM but only 32 blocks so you can only saturate threads if your block is at least 64 threads large

vortex-cuda/kernels/for.cu
```diff
@@ -0,0 +1,45 @@
...
+    uint64_t array_len
+) {
+    // Each block handles 1024 elements. Each thread handles 1024 / 32 contiguous elements.
+    const uint8_t elements_per_thread = 32;
+    const uint32_t block_start = blockIdx.x * 1024;
+
+    const uint64_t start = block_start + threadIdx.x * elements_per_thread;
+    const uint64_t end = (start + elements_per_thread < array_len) ? (start + elements_per_thread) : array_len;
+
+    for (uint64_t idx = start; idx < end; ++idx) {
```
robert3005: should this go via shared memory or is there no difference for FOR?

joseph-isaacs: I would expect if your threads write one batch each it would not matter?

joseph-isaacs: So not this, but each thread write on each iteration writing continguos memory?

vortex-cuda/src/for_.rs
joseph-isaacs: why not?

vortex-cuda/kernels/for.cu
0ax1: @joseph-isaacs 
> writing continguos memory?

Yes, we write to contiguous memory.

@robert3005 

Can block shared memory even be faster here, given that we for now have distinct kernels when combining them? We create a buf with global memory on the host. If we would want to operate on local memory we would have to copy to it first and then back to global, in order make the results visible for subsequent kernels?

vortex-cuda/src/executor.rs
0ax1: In my understanding, one also gets better latency hiding by having more threads.

0ax1: Done

vortex-cuda/kernels/for.cu
robert3005: I think we can benchmark it later. I’m mostly making sure we don’t lose things from v1

vortex-cuda/src/for_.rs
0ax1: i'll iterate in a follow up.

vortex-cuda/kernels/for.cu
0ax1: Turns out the coalesced access is 2x faster for FoR => pushed.

0ax1: Handwavy bench which does pick up on diff once the buffers get really big: https://github.com/vortex-data/vortex/compare/ad/cuda-for-bench . Otherwise, this basically measure the kernel launch duration.

robert3005 (approved):

PR #5967, opened 2026-01-15
fix[vortex-scalar]: correctly preserve nulls in from_vector_scalar

asubiotto:
Each type had a null check within its match arm, but more complex types were missing null checks, so nullability was lost after conversion.

This commit simplifies converting nulls by introducing a top-level check which should handle all types.

### DIFF
--- a/vortex-scalar/src/vectors.rs
+++ b/vortex-scalar/src/vectors.rs
@@ -19,6 +19,7 @@ use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_error::vortex_ensure;
 use vortex_mask::Mask;
+use vortex_vector::ScalarOps;
 use vortex_vector::VectorMut;
 use vortex_vector::VectorMutOps;
 use vortex_vector::VectorOps;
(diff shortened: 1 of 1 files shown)


vortex-scalar/src/vectors.rs
```diff
@@ -169,31 +170,32 @@ impl Scalar {
...
+        if !scalar.is_valid() {
+            vortex_ensure!(
+                dtype.is_nullable(),
+                "Cannot create null scalar for non-nullable dtype"
+            );
+            return Ok(Scalar::null(dtype.clone()));
+        }
+
         Ok(match dtype {
             DType::Null => Scalar::null(DType::Null),
```
asubiotto: Not sure whether I should add an `unreachable!` to this arm

AdamGS: personally, all for asserting stuff like that

AdamGS (approved):

vortex-scalar/src/vectors.rs
asubiotto: Done.

PR #5976, opened 2026-01-15
Correct physical schema inference for DF files with non-logical types.

AdamGS:
This PR fixes how a file or a scan's `DType` is mapped into an arrow types in cases where the table has non-logical types that don't round trip between the types systems.

This change both simplifies how we push projection expressions and allows us to save on unnecessary work converting arrays between various logically-equivalent types.

### DIFF
--- a/vortex-datafusion/src/persistent/opener.rs
+++ b/vortex-datafusion/src/persistent/opener.rs
@@ -5,9 +5,6 @@ use std::ops::Range;
 use std::sync::Arc;
 use std::sync::Weak;
 
-use arrow_schema::DataType;
-use arrow_schema::Field;
-use arrow_schema::Schema;
 use datafusion_common::DataFusionError;
 use datafusion_common::Result as DFResult;
 use datafusion_common::ScalarValue;
@@ -37,7 +34,6 @@ use tracing::Instrument;
 use vortex::array::ArrayRef;
 use vortex::array::VortexSessionExecute;
 use vortex::array::arrow::ArrowArrayExecutor;
-use vortex::dtype::DType;
 use vortex::error::VortexError;
 use vortex::layout::LayoutReader;
 use vortex::metrics::VortexMetrics;
@@ -51,6 +47,7 @@ use crate::VortexAccessPlan;
 use crate::convert::exprs::ExpressionConvertor;
 use crate::convert::exprs::ProcessedProjection;
 use crate::convert::exprs::make_vortex_predicate;
+use crate::convert::schema::calculate_physical_schema;
 use crate::persistent::stream::PrunableStream;
 
 #[derive(Clone)]
@@ -102,8 +99,7 @@ impl FileOpener for VortexOpener {
         let expr_adapter_factory = self.expr_adapter_factory.clone();
 
         let file_cache = self.file_cache.clone();
-        let table_schema = self.table_schema.clone();
-        let logical_file_schema = table_schema.file_schema().clone();
+        let unified_file_schema = self.table_schema.file_schema().clone();
         let batch_size = self.batch_size;
         let limit = self.limit;
         let metrics = self.metrics.clone();
@@ -114,7 +110,8 @@ impl FileOpener for VortexOpener {
 
         // Replace column access for partition columns with literals
         #[allow(clippy::disallowed_types)]
-        let literal_value_cols = table_schema
+        let literal_value_cols = self
+            .table_schema
             .table_partition_cols()
             .iter()
             .map(|f| f.name())
@@ -146,7 +143,7 @@ impl FileOpener for VortexOpener {
                 .and_then(|predicate| {
                     FilePruner::try_new(
                         predicate.clone(),
-                        &logical_file_schema,
+                        &unified_file_schema,
                         &file,
                         Count::default(),
                     )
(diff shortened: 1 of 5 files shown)


vortex-datafusion/src/persistent/opener.rs
```diff
@@ -350,6 +355,175 @@ impl FileOpener for VortexOpener {
...
+///
+/// Some Arrow types don't roundtrip cleanly through Vortex's DType system:
+/// - Dictionary types lose their encoding (become the value type)
+/// - Utf8/LargeUtf8 become Utf8View
+/// - Binary/LargeBinary become BinaryView
+/// - RunEndEncoded loses its encoding
+///
+/// For these types, we use the logical schema's type instead of the DType's natural Arrow
+/// conversion, since Vortex's Arrow executor can produce these types when requested.
+fn calculate_physical_file_schema(
```
asubiotto: I think we should remove this. See my comments on the callsites.

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -161,9 +165,10 @@ impl FileOpener for VortexOpener {
                 .await
                 .map_err(|e| exec_datafusion_err!("Failed to open Vortex file {e}"))?;
 
-            let physical_file_schema = Arc::new(vxf.dtype().to_arrow_schema().map_err(|e| {
-                exec_datafusion_err!("Failed to convert file schema to arrow: {e}")
-            })?);
+            let physical_file_schema = Arc::new(calculate_physical_file_schema(
```
asubiotto: I think the nomenclature here is a bit confusing because datafusion's classic interpretation of logical (i.e. table types) vs physical (i.e. file types) doesn't really make sense in a vortex world, where the output arrow types can be dynamic, not static. I think `s/physical_file_schema/this_file_schema` and `s/logical_file_schema/unified_file_schema` would make more sense. What do you think?

Also, given the vortex dynamic output schema nature, calculating `this_file_schema` should just be the subset of `unified_file_schema` that the file actually stores, no extra type calculations necessary.

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -196,12 +201,12 @@ impl FileOpener for VortexOpener {
...
             // The schema of the stream returned from the vortex scan.
+            // We use the physical_file_schema as reference for types that don't roundtrip.
             let scan_dtype = scan_projection.return_dtype(vxf.dtype()).map_err(|_e| {
                 exec_datafusion_err!("Couldn't get the dtype for the underlying Vortex scan")
             })?;
-            let stream_schema = scan_dtype.to_arrow_schema().map_err(|_e| {
-                exec_datafusion_err!("Couldn't get the schema for the underlying Vortex scan")
-            })?;
+            let stream_schema =
+                calculate_physical_file_schema(&scan_dtype, &projected_physical_schema)?;
```
asubiotto: I think this will be brittle because it's a lookup by name which can be changed with aliases etc... What about retaining the physical arrow type information when converting expressions in split_projection? You could then get a post-scan-projection arrow schema by calling something on the ProcessedProjection.

vortex-datafusion/tests/schema_evolution.rs
```diff
@@ -402,3 +406,123 @@ async fn test_filter_schema_evolution_struct_fields() {
...
+        get_field(col("my_struct"), "a")
+            .eq(lit("x1"))
+            .or(get_field(col("my_struct"), "a").eq(lit("x2")));
+    // run a filter that touches both the payload.uptime AND the payload.instance nested fields
+    let df = ctx.read_table(table.clone())?;
+    let filtered_scan = df.filter(filter)?.collect().await?;
+
+    assert_eq!(filtered_scan[0].schema(), read_schema);
+
+    assert_batches_eq!(
```
asubiotto: I think this needs to be assert_batches_sorted_eq! as well

vortex-datafusion/src/persistent/opener.rs
AdamGS: I agree that its brittle, but it should also be able to maintain the names as both (should) be projected in the same way. I'm thinking through the comments, would definitely love to make this more robust.

AdamGS: That would be a very similar flow right? Just only available after splitting the projection which should make sure all the names match.

vortex-datafusion/src/persistent/opener.rs
a10y: The output Arrow types that Vortex reports are always going to be the simple types, e.g. `Dictionary` and `RunEnd` will never be reported as a column type even if Vortex encodes it that way.

So I think we need to keep the calculation?

Also I agree that I find DF's nomenclature confusing here. In my head, I always call this the "reader schema" (logical file schema) and the "writer schema" (physical file schema).

If we don't do the calculation here, then by default we never pushdown the cast to Dictionary/RunEnd and always end up doing it in Arrow after.

a10y:
We also would need to add `ToArrowKernel`s for Dictionary and RunEnd in Vortex right? Otherwise we're going through canonical first.

vortex-datafusion/src/persistent/opener.rs
asubiotto: Yes, the flow would be pretty much exactly the same, just that every DF expression you successfully convert to vortex would also keep the DF result type, i.e. you calculate the physical arrow schema in the `split_projection` recursion. I'll defer to you on what to do, but I think we can avoid the complicated `calculate_physical_file_schema` this way.

asubiotto:
> We also would need to add `ToArrowKernel`s for Dictionary and RunEnd in Vortex right? Otherwise we're going through canonical first.

I don't think so, direct conversion is taken care of by `execute_arrow`: https://github.com/vortex-data/vortex/blob/cc81639bd5599149854bc084413e16ae9a3be081/vortex-array/src/arrow/executor/mod.rs#L122

vortex-datafusion/src/persistent/opener.rs
asubiotto: Right, but what I am suggesting is removing any reference to `vxf.dtype()` here. The “calculation” would become just finding the subset of columns in `logical_file_schema` (physical `arrow` types) that this specific file stores. This should all be done by name.

The complexity of the current calculation that chooses the physical arrow type from one schema or the other based on the type itself is the part that is not necessary I think especially since the physical arrow types from the `logical_file_schema` are what is “desired” by datafusion and what you need to end up outputting (minus projections) any way.

AdamGS: this conversation split up a bit, but I do think its important to keep pulling from `vxf.dtype()` and use the names, that's how the expression adapter rewrites the expression to take advantage of the data's concrete structure (in the case of filters like `non_exist_col > 5`)

asubiotto: Ah yes, you're right. We still need something to tell us what the actual columns are in the file, but we can always choose the `logical_file_schema[column_name]` type

AdamGS: As I suspected, Vortex does care that expressions match type-wise, we don't inject any casting or coercion, and for primitive types they do have to match. Turning catch-all branch to just `logical_type.clone()` breaks downstream.

I do think the current logic is necessary, I'll add some more comments and maybe rename some of the variables, but I'm not sure that something simpler will work now.

asubiotto: Sounds good! Would be curious to see what breaks, but don't want to take up any more of your time so might play around with this when I have a moment.

asubiotto (approved):

PR #5982, opened 2026-01-15
Add divan benchmarks for Mask::rank

joseph-isaacs:
## Summary

Optimize `Mask::rank()` by adding an efficient `BitBuffer::select()` operation that finds the nth set bit without materializing the full indices vector.

### Key changes:

**`BitBuffer::select()`** - O(n/64) algorithm that:
- Uses **BMI2 `pdep` instruction** on x86_64 with bmi2 for O(1) select-in-word
- Falls back to **hybrid loop/binary-search** on other platforms
- Includes **fast path for all-ones words** (skips dense chunks in O(1))

**`BitBuffer::select_via_slices()`** - Alternative for highly correlated data:
- Iterates over contiguous runs of set bits
- O(num_runs) complexity - better when data has long runs of 1s/0s

**`Mask::rank()` optimization:**
- Uses cached indices for O(1) lookup when available
- Falls back to `BitBuffer::select()` when uncached (avoids allocating indices vector)
- This also speeds up `intersect_by_rank()` which calls `rank()` internally

### Files changed:
- `vortex-buffer/src/bit/buf.rs` - Add select implementations
- `vortex-buffer/benches/select.rs` - Benchmarks for different approaches
- `vortex-mask/src/lib.rs` - Update rank() to use select()


joseph-isaacs:
@claude compare this PR to https://docs.rs/bitm/latest/bitm/fn.select64.html

PR #5983, opened 2026-01-15
Add benchmarks for intersect_by_rank method

joseph-isaacs:
## Summary

Optimizes `intersect_by_rank` in `vortex-mask` with 2-3x performance improvement through a hybrid approach that combines indices lookup with direct u64 output building.

## Performance Improvements

**Main `intersect_by_rank` @ 1M elements:**

| Benchmark | Before | After | Speedup |
|-----------|--------|-------|---------|
| realistic_low_selectivity | 53.67 µs | 17.7 µs | **3.0x** |
| realistic_medium_selectivity | 78.89 µs | 43.74 µs | **1.8x** |
| high_density_rank_90% | 244.5 µs | 104.2 µs | **2.3x** |

**Additional `intersect_by_rank_runs` for correlated data:**

| Data Pattern | Original | Runs | Speedup |
|--------------|----------|------|---------|
| Long runs (64 on/64 off) | 1.32 ms | 53 µs | **25x** |

### Main Optimization (Hybrid Approach)

The main `intersect_by_rank` now uses a hybrid approach that avoids allocation overhead:

**Before:**
```rust
// Creates an indices vector output
mask_indices.iter().map(|idx| self_indices[*idx]).collect()

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10575,6 +10575,7 @@ name = "vortex-mask"
 version = "0.1.0"
 dependencies = [
  "arrow-buffer 57.2.0",
+ "codspeed-divan-compat",
  "itertools 0.14.0",
  "rstest",
  "serde",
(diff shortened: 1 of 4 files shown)


robert3005:
done in #7744

PR #5988, opened 2026-01-15
chore: Swap ByteBuffer for BufferHandle more

a10y:


a10y:
It looks like BoolArray is not preferred for GPU, see e.g. cudf

image

But we have ByteBool here

gatesn:
Arrow also has a 1-byte boolean extension. I did wonder whether byte bools would be preferred on GPU since bit unpacking would be madness. We had the same conclusion for pipelined CPU code too in fact. 🤷

joseph-isaacs (approved):

PR #5993, opened 2026-01-16
chore: return CUDA events from kernel launch

0ax1:
This PR changes the launch kernel API in vortex-cuda to propagate the pair of CUDA events returned from `cudarc::LaunchArgs:launch`.

When launching a kernel, CUDAs events are automatically submitted before and after. 
The events can be used to wait for a particular kernel to finish or to benchmark a CUDA kernel.

### DIFF
--- a/vortex-cuda/src/executor.rs
+++ b/vortex-cuda/src/executor.rs
@@ -28,6 +28,15 @@ use vortex_session::VortexSession;
 use crate::CudaSession;
 use crate::session::CudaSessionExt;
 
+/// CUDA kernel events recorded before and after kernel launch.
+#[derive(Debug)]
+pub struct CudaKernelEvents {
+    /// Event recorded before kernel launch.
+    pub before_launch: CudaEvent,
+    /// Event recorded after kernel launch.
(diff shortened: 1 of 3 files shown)


vortex-cuda/src/executor.rs
```diff
@@ -68,10 +78,18 @@ macro_rules! launch_cuda_kernel {
...
+/// A pair of CUDA events submitted before and after the kernel.
+/// Depending on `CUevent_flags` these events can contain timestamps. Use
+/// `CU_EVENT_DISABLE_TIMING` for minimal overhead and `CU_EVENT_DEFAULT` to
+/// enable timestamps.
 pub fn launch_cuda_kernel_impl(
     launch_builder: &mut LaunchArgs,
+    event_flags: CUevent_flags,
     array_len: usize,
-) -> VortexResult<()> {
+) -> VortexResult<(CudaEvent, CudaEvent)> {
```
gatesn: Create a struct for this

0ax1: done

gatesn (approved):

PR #5996, opened 2026-01-16
Use static field names for min_max return value and dtype

AdamGS:


AdamGS:
the codspeed improvements seem possible, min/max is used in a bunch of places in the compressor

robert3005 (approved):

PR #5997, opened 2026-01-16
chore[arrow]: remove execute::<Vector>() from arrow, use canonical

joseph-isaacs:


docs/guides/python-integrations.md
```diff
@@ -98,7 +98,7 @@ If you have a struct array, use {meth}`.Array.to_arrow_table` to construct an Ar
 >>> struct_arr.to_arrow_table()
 pyarrow.Table
 age: int64
-name: string
+name: string_view
```
gatesn: This feels like a behavior change we really don't want to sneak into this!

joseph-isaacs:
see: https://github.com/vortex-data/vortex/issues/6038

docs/guides/python-integrations.md
joseph-isaacs: fixed

gatesn (approved):

PR #6000, opened 2026-01-16
Array VTable

gatesn:
Having started to implement ExpressionArray::deserialize, I realised the best thing to do is pass a session to the deserialize function.

If we do that.... then there's no reason for Array VTables to hold state. They're just ZSTs. 

So this should avoid a whole bunch of heap allocations.

We should follow up with other vtables and eventually converge on a pattern.


vortex-session/src/registry.rs
```diff
@@ -34,25 +44,93 @@ impl<T: Clone + Display + Eq> Registry<T> {
...
-            self.0.insert(item.to_string(), item);
+    /// Register a new item, replacing any existing item with the same ID, and return self for
+    pub fn with(self, id: impl Into<Id>, item: impl Into<T>) -> Self {
+        self.register(id, item.into());
+        self
+    }
+}
+
+#[derive(Debug)]
+pub struct Context<T> {
```
joseph-isaacs, line 79: can we add a doc str?

joseph-isaacs (approved):

PR #6002, opened 2026-01-16
Only check for ConstantVTable

gatesn:


encodings/fastlanes/src/for/array/for_compress.rs
```diff
@@ -97,8 +98,8 @@ mod test {
         assert!(compressed.reference_scalar().dtype().is_signed_int());
         assert!(compressed.encoded().dtype().is_signed_int());
 
-        let constant = compressed.encoded().as_constant().unwrap();
-        assert_eq!(constant, Scalar::from(0i32));
+        let encoded = compressed.encoded().scalar_at(0);
+        assert_eq!(encoded, Scalar::from(0i32));
```
joseph-isaacs, line 102: this doesn't do the same thing?

joseph-isaacs (approved):

PR #6004, opened 2026-01-16
Remove unnecessary work from `ListArray::validate`

AdamGS:
This PR removes some unnecessary work, relying on the previous assertion that the offsets array is sorted.

### DIFF
--- a/vortex-array/src/arrays/list/array.rs
+++ b/vortex-array/src/arrays/list/array.rs
@@ -180,36 +180,34 @@ impl ListArray {
         // the elements array.
         if let Some(min_max) = min_max(offsets)? {
             match_each_integer_ptype!(offsets_ptype, |P| {
-                let max_offset = P::try_from(offsets.scalar_at(offsets.len() - 1))
-                    .vortex_expect("Offsets type must fit offsets values");
-
                 #[allow(clippy::absurd_extreme_comparisons, unused_comparisons)]
                 {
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/list/array.rs
```diff
@@ -187,17 +187,10 @@ impl ListArray {
                 {
                     if let Some(min) = min_max.min.as_primitive().as_::<P>() {
                         vortex_ensure!(
-                            min >= 0 && min <= max_offset,
+                            min >= 0,
```
AdamGS: at this point we know min is `lt_eq` the last element, because the offsets array is sorted

vortex-array/src/arrays/list/array.rs
```diff
@@ -187,17 +187,10 @@ impl ListArray {
...
                             "offsets minimum {min} outside valid range [0, {max_offset}]"
                         );
                     }
-
-                    if let Some(max) = min_max.max.as_primitive().as_::<P>() {
```
AdamGS, line 195: max == max_offset, and if min is `>=`, so should it be for the same reasons

robert3005:
Because we check is_sorted this is indeed unnecessary

vortex-array/src/arrays/list/array.rs
joseph-isaacs: dont do

AdamGS: done!

joseph-isaacs (approved):

PR #6014, opened 2026-01-17
chore: validate the error type during testing

caicancai:
The current test only checks for errors, but doesn't check the error type. Perhaps this could be improved.

example
```
    fn test_primitive_array_validation_failure_length_mismatch() {
        // Invalid case: validity length doesn't match buffer length.
        let buffer = Buffer::from_iter([1i32, 2, 3]);
        let validity = Validity::from_iter([true, false]); // Length 2, buffer is length 3.
        let result = PrimitiveArray::try_new(buffer, validity);

        assert!(matches!(result, Err(VortexError::InvalidArgument(_, _))));
        assert!(result.is_err());
    }
```

### DIFF
--- a/vortex-array/src/arrays/validation_tests.rs
+++ b/vortex-array/src/arrays/validation_tests.rs
@@ -16,6 +16,7 @@ mod tests {
     use vortex_dtype::DType;
     use vortex_dtype::Nullability;
     use vortex_dtype::PType;
+    use vortex_error::VortexError;
     use vortex_vector::binaryview::BinaryView;
 
     use crate::IntoArray;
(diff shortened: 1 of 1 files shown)


caicancai:
@joseph-isaacs Sorry to bother you, but I'll take a look at this small change when I have time and see if it's appropriate.

joseph-isaacs (approved):
Thanks, but please do come with a larger scale PR if want to update tests next time

caicancai:
> Thanks, but please do come with a larger scale PR if want to update tests next time

Thank you for the reminder. I probably won't be updating the test results separately anymore, as I've already started delving into and learning about the Vortex functionality.

PR #6022, opened 2026-01-19
feat[python]: support full read/write from object storage

a10y:
This adds the ability to write in PyVortex directly to object storage, along with more ergonomic configuration of object store credentials for the major PyVortex read/write paths: `vx.open`, `vx.io.read`, and `vx.io.write`, as well as `VortexWriteOptions.write`.

This is using the https://docs.rs/pyo3-object_store/latest/pyo3_object_store/ crate, which is what runs the [obstore](https://developmentseed.org/obstore/latest/) Python library.

Notably, we can't reuse `obstore` and all of its types directly. This is b/c of the way it works, you're required to actually link in pyo3-object_store and then expose it as a new module **within your Python bundle**.

There are some other examples of this pattern:

* [nutpie](https://github.com/pymc-devs/nutpie/blob/adb83e24235773d7ff97f86e1858ee9d31db0936/src/wrapper.rs#L1624-L1625)
* [geoarrow](https://github.com/geoarrow/geoarrow-rs/blob/main/python/geoarrow-io/src/lib.rs#L46-L48)

Sadly there's no nicely distributed types package, so I need to copy all of the pyi files here with their original MIT license.

I've added a Python unit test to demonstrate using the new object store builders with simple local file system object store

Resolves https://github.com/vortex-data/vortex/discussions/5673

### DIFF
--- a/docs/api/python/store.rst
+++ b/docs/api/python/store.rst
@@ -1,0 +1,17 @@
+Object Store support
+====================
+
+Vortex arrays support reading and writing to object storage systems such as, S3, Google Cloud Storage, and
+Azure Blob Storage.
+
+.. autosummary::
+   :nosignatures:
+
+.. raw:: html
+
+   <hr>
+
+.. automodule:: vortex.store
+    :members:
+    :imported-members:
+
--- a/vortex-python/src/object_store/registry.rs
+++ b/vortex-python/src/object_store/registry.rs
@@ -1,0 +1,192 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Apache Software Foundation (ASF)
+
+//! This file is an adapted version of the `DefaultObjectStoreRegistry` from the object_store crate,
+//! but modified to resolve configurations out of environment variables case-insensitively. This
+//! is similar to how all the `Store::from_env` builders work for the various object stores.
+//!
+//! See also <https://github.com/apache/arrow-rs-object-store/issues/529>
+
+#![allow(clippy::disallowed_types)]
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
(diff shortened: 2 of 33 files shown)


a10y:
Just pushed a commit with a modified version of the ObjectStoreRegistry that resolves configurations out of the environment in a more consistent way.

This is related to https://github.com/apache/arrow-rs-object-store/issues/529 which we stumbled upon last week

vortex-python/src/object_store/registry.rs
```diff
@@ -0,0 +1,192 @@
...
+                .resolve(&Url::parse("s3://my-bucket/test").unwrap())
+                .unwrap();
+
+            // NOTE(aduffy): object_store doesn't let us downcast stores, the only way to verify
+            //  that a configuration was added was to validate that it ends up in the Debug
+            //  output :/
+            let mut debug_str = String::new();
+            write!(&mut debug_str, "{store:?}").unwrap();
+
+            assert!(debug_str.contains("us-east-3"));
```
a10y, line 189: this is pretty sad but i don't have a better idea

docs/api/python/store.rst
```diff
@@ -0,0 +1,17 @@
+Object Store support
```
AdamGS, line 1: this file renders in a a weird way (see image). I would expect the config classes to appear here at least.

Image

AdamGS (approved):
Looks great to me, aside from the docs rendering thing (do we also want to put some example there too?)

docs/api/python/store.rst
a10y: Yea I'm so bad with Python/Sphinx tooling. I might just try and get Claude to clean this up in a FLUP if that's ok

PR #6024, opened 2026-01-19
propagate is_sorted for runend ends child

onursatici:
I am testing locally on a clickbench file, after these we get 100% of the runend -> end arrays have is_sorted set, before it was 0%

Also a note for checking this again in future, with runend validation on we will compute is_sorted when decoding the array, so to get the stored stats we have to disable runend validation


joseph-isaacs (approved):
It seems like there must be a more systematic thing we can do?

onursatici:
yea feels like stat propagation should be defined in the encoding level, how child stats propagate to the parent

PR #6025, opened 2026-01-19
do not validate is_sorted on runend

onursatici:
assert no invalid memory access will happen on decoding time instead.

I think with https://github.com/vortex-data/vortex/pull/6024 we won't hit this with new files, but there is nothing preventing this to come back when we add a new encoding into the mix, and this will also speed up decoding old files that didn't have the stat propagated

### DIFF
--- a/encodings/runend/src/compress.rs
+++ b/encodings/runend/src/compress.rs
@@ -263,12 +263,6 @@ pub fn runend_decode_typed_bool(
         Mask::AllTrue(_) => {
             let mut decoded = BitBufferMut::with_capacity(length);
             for (end, value) in run_ends.zip_eq(values.iter()) {
-                assert!(
-                    end >= decoded.len(),
-                    "Runend ends must be monotonic, got {end} after {}",
-                    decoded.len()
-                );
-                assert!(end <= length, "Runend end must be less than overall length");
                 decoded.append_n(value, end - decoded.len());
             }
             BoolArray::from_bit_buffer(decoded.freeze(), values_nullability.into())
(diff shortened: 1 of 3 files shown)


encodings/runend/src/compress.rs
```diff
@@ -253,6 +263,12 @@ pub fn runend_decode_typed_bool(
         Mask::AllTrue(_) => {
             let mut decoded = BitBufferMut::with_capacity(length);
             for (end, value) in run_ends.zip_eq(values.iter()) {
+                assert!(
+                    end >= decoded.len(),
+                    "Runend ends must be monotonic, got {end} after {}",
+                    decoded.len()
+                );
+                assert!(end <= length, "Runend end must be less than overall length");
```
joseph-isaacs: do we need this?

encodings/runend/src/iter.rs
```diff
@@ -29,7 +29,12 @@ pub fn trimmed_ends_iter<E: IntegerPType>(
     run_ends
         .iter()
         .copied()
-        .map(move |v| v - offset_e)
+        .map(move |v| {
+            if v < offset_e {
+                vortex_panic!("run end {v} must be >= offset {offset}");
+            }
```
joseph-isaacs, line 35: checked `-` and expect?

onursatici: yea I tried checked sub but that also has the same branch so not much perf improvement, but I can do that here would be cleaner

joseph-isaacs: Just think it read better

joseph-isaacs (approved):

PR #6031, opened 2026-01-19
feat: add configurable in-memory read concurrency

caicancai:
The `VortexReadAt::concurrency()` function of `ByteBuffer` has a fixed concurrency of 16, which cannot be overridden by the caller. Users' in-memory data sources may require lower/higher concurrency to control scheduling overhead.

We could provide a default value of 16 while allowing users to explicitly set the concurrency.


caicancai:
@joseph-isaacs Sorry to bother you, could you please take a look at this PR when you have time?

PR #6032, opened 2026-01-19
Pass missing cmake sccache-related flags

AdamGS:
See [docs](https://github.com/mozilla-actions/sccache-action/tree/v0.0.9/?tab=readme-ov-file#cc-code). This change takes the cxx build from ~5 minutes to just over 1 minute.


0ax1 (approved):
i think we should double check whether we want to nuke vortex-cxx. but otherwise, happy if ci is happy and picks up the cache.

PR #6033, opened 2026-01-19
performance[vortex-array]: avoid is_valid call in varbinview builder

asubiotto:
Similar to #5814, this creates a lot of unnecessary CPU usage:

image


AdamGS (approved):
LGTM, nice find

gatesn (approved):

PR #6035, opened 2026-01-19
Add delta encoding to btrblocks

palaska:


a10y:
There's a reason we removed this from the compressor, I just forget why, @robert3005 ?

robert3005:
I talked to @palaska offline. There's the outstanding issue that validity in delta encoding is not transposed (while the values are)

joseph-isaacs:
The decompress is unsound

PR #6042, opened 2026-01-19
ci: Cache clippy runs in different key

AdamGS:
There's a bigger effort here to potentially untangle some of the caching keys, but for clippy it definitely needs a different key than steps that build/test.


AdamGS:
nvm, underlying cache uses job id in key

PR #6044, opened 2026-01-19
Don't cache the coverage report itself

AdamGS:
The size of the coverage data is between 200MB to **1GB**, this PR makes sure we don't cache it as the cache actions caches everything in `/target`


.github/workflows/ci.yml
```diff
@@ -431,6 +431,7 @@ jobs:
         run: |
           cargo run --bin datafusion-bench -- tpch -i2 --formats "vortex,vortex-compact" --opt scale-factor=0.1
           cargo run --bin duckdb-bench -- tpch -i2 --formats "vortex,vortex-compact" --opt scale-factor=0.1
+          rm -rf vortex-bench/data/
```
AdamGS, line 434: This might be an unnecessarily cautious thing to do, but why not

joseph-isaacs (approved):

PR #6045, opened 2026-01-19
chore[array]: remove CanonicalVTable and move append_to_builder to main VTable

joseph-isaacs:
Internal Array API break, just use execute


joseph-isaacs:
https://github.com/Amanieu/parking_lot/issues/477

gatesn (approved):

PR #6049, opened 2026-01-20
feat: stats pruning for LIKE filters

### LINKED ISSUE #6128: Zone pruning for LIKE filters
### Discussed in https://github.com/vortex-data/vortex/discussions/5943

<sup>Originally posted by **a10y** January 13, 2026</sup>
In the ZonedReader, we don't currently prune `LIKE` expressions since they don't report any stats falsification, but we should be able to prune using min/max stats for exact queries or prefix queries.</div>

a10y:
Fixes #6128 

There are two types of `LIKE` filters that we can prune using our zone maps:

| Filter | Zone Map Predicate |
|--------|--------|
|`col LIKE 'exact'` | `col.min > 'exact' OR col.max < 'exact'` |
| `col LIKE 'prefix%'` | `col.min >= 'prefiy' OR col.max < 'prefix'` | 

I extracted out the existing logic from Utf8Scalar::upper_bound into a trait so we can make sure logic is consistent across stats gen/pruning

### DIFF
--- a/vortex-array/src/expr/exprs/like.rs
+++ b/vortex-array/src/expr/exprs/like.rs
@@ -9,6 +9,7 @@ use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
 use vortex_error::vortex_err;
 use vortex_proto::expr as pb;
+use vortex_scalar::StringLike;
 
 use crate::ArrayRef;
 use crate::compute::LikeOptions;
@@ -19,9 +20,16 @@ use crate::expr::ExecutionArgs;
 use crate::expr::ExecutionResult;
 use crate::expr::ExprId;
 use crate::expr::Expression;
+use crate::expr::Literal;
+use crate::expr::StatsCatalog;
 use crate::expr::VTable;
 use crate::expr::VTableExt;
 use crate::expr::and;
+use crate::expr::gt;
+use crate::expr::gt_eq;
+use crate::expr::lit;
+use crate::expr::lt;
+use crate::expr::or;
 
 /// Expression that performs SQL LIKE pattern matching.
 pub struct Like;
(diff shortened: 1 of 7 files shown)


robert3005:
FWIW the stats truncation logic has the increment utf8 behaviour you want. It's implemented on a scalar https://github.com/vortex-data/vortex/blob/develop/vortex-scalar/src/binary.rs#L101 so it can mutate it in place. I wonder if they just have to stay separate but maybe upper bound being in place for binary scalar is just a cute optimistaion

a10y:
That function is missing some of the UTF-8 special character handling in the DF implementation, so now I'm worried that one of us is doing the wrong thing...

robert3005:
I gave you the binary version of this function, the utf8 version is here https://github.com/vortex-data/vortex/blob/develop/vortex-scalar/src/utf8.rs#L98 apologies

a10y:
I was looking at the utf8 version, I guess I don't know enough about utf8 to know if this covers the same edge cases that the datafusion implementation does

robert3005:
I derived it from the same code so we should make sure it’s the same so stats behave the same as regular filters

vortex-array/src/expr/exprs/like.rs
```diff
@@ -127,6 +135,67 @@ impl VTable for Like {
...
+    fn stat_falsification(
+        &self,
+        like_opts: &LikeOptions,
+        expr: &Expression,
+        catalog: &dyn StatsCatalog,
+    ) -> Option<Expression> {
+        // Attempt to do min/max pruning for LIKE 'exact' or LIKE 'prefix%'
+
+        // Don't attempt to handle ilike or negated like
+        if like_opts.negated || like_opts.case_insensitive {
```
a10y, line 148: In theory we could handle negated like if we know that the block ONLY contains things which match the prefix, but we'd need to add a `starts_with` expression and I didn't want to do that here

robert3005 (approved):

vortex-array/src/expr/exprs/like.rs
a10y: actually, we can do this without a starts_with, if we see that min >= prefix AND min < succ AND max >= prefix AND max < succ AND null_count = 0

PR #6058, opened 2026-01-20
cuda pinned buffer pool

onursatici:
everything in this PR will probably change


vortex-cuda/src/device_buffer.rs
```diff
@@ -1,106 +1,126 @@
...
-    cuda_slice: CudaSlice<T>,
+use crate::PooledPinnedBuffer;
+
+/// A device buffer backed by CUDA device memory.
+pub struct CudaDeviceBuffer {
+    data: Arc<CudaSlice<u8>>,
+    offset: usize,
+    len: usize,
+    stream: Arc<CudaStream>,
+    completion: Arc<CudaEvent>,
```
0ax1, line 29: you shouldn't need an extra event to check for completion, read/write events are set on the cudaslice when submitted as part of a kernel launch.

0ax1:
```
#[derive(Debug)]
pub struct CudaSlice<T> {
    pub(crate) cu_device_ptr: sys::CUdeviceptr,
    pub(crate) len: usize,
    pub(crate) read: Option<CudaEvent>,
    pub(crate) write: Option<CudaEvent>,
    pub(crate) stream: Arc<CudaStream>,
    pub(crate) marker: PhantomData<*const T>,
}
```

vortex-cuda/src/device_buffer.rs
```diff
@@ -1,106 +1,126 @@
...
+        self.len
     }
-}
 
-impl<T: 'static> Hash for CudaDeviceBuffer<T> {
-    /// Hash the buffer pointer address.
-    fn hash<H: Hasher>(&self, state: &mut H) {
-        (&raw const self.cuda_slice).hash(state);
+    fn copy_to_host(&self) -> VortexResult<ByteBuffer> {
+        let mut host = ByteBufferMut::with_capacity_aligned(self.len, Alignment::of::<u8>());
```
0ax1, line 63: This might end up underaligned?

PR #6060, opened 2026-01-20
feat: introduce `CudaDeviceBuffer`

0ax1:
As part of this, FoR kernel execution is now async. GPU tests still pass and the benchmarks show unchanged performance.

### DIFF
--- a/vortex-array/src/buffer.rs
+++ b/vortex-array/src/buffer.rs
@@ -41,8 +41,11 @@ enum Inner {
     Device(Arc<dyn DeviceBuffer>),
 }
 
-/// A buffer that is stored on a device (e.g. GPU).
+/// A buffer that is stored on the GPU.
 pub trait DeviceBuffer: 'static + Send + Sync + Debug + DynEq + DynHash {
+    /// Returns a reference as `Any` to enable downcasting.
+    fn as_any(&self) -> &dyn Any;
+
     /// Returns the length of the buffer in bytes.
     fn len(&self) -> usize;
 
@@ -61,9 +64,6 @@ pub trait DeviceBuffer: 'static + Send + Sync + Debug + DynEq + DynHash {
     /// Create a new buffer that references a subrange of this buffer at the given
     /// slice indices.
     fn slice(&self, range: Range<usize>) -> Arc<dyn DeviceBuffer>;
-
-    /// Returns a reference as `Any` to enable downcasting.
-    fn as_any(&self) -> &dyn Any;
 }
 
 impl Hash for dyn DeviceBuffer {
--- a/vortex-array/src/buffer.rs
+++ b/vortex-array/src/buffer.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::any::Any;
 use std::fmt::Debug;
 use std::hash::Hash;
 use std::hash::Hasher;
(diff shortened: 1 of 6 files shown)


vortex-array/src/buffer.rs
```diff
@@ -60,6 +61,9 @@ pub trait DeviceBuffer: 'static + Send + Sync + Debug + DynEq + DynHash {
     /// Create a new buffer that references a subrange of this buffer at the given
     /// slice indices.
     fn slice(&self, range: Range<usize>) -> Arc<dyn DeviceBuffer>;
+
+    /// Returns a reference as `Any` to enable downcasting.
+    fn as_any(&self) -> &dyn Any;
```
gatesn: Nit: this is typically the first function in a trait

vortex-cuda/src/device_buffer.rs
```diff
@@ -2,95 +2,131 @@
...
-    pub fn cuda_slice(&self) -> &CudaSlice<T> {
-        &self.cuda_slice
+    /// Returns a [`CudaView`] to the CUDA device buffer.
+    pub fn as_view(&self) -> CudaView<'_, T> {
+        self.inner.slice(self.offset..self.offset + self.len)
+    }
+
+    /// Returns the device pointer for this buffer.
+    fn device_ptr(&self) -> CUdeviceptr {
+        self.inner.device_ptr(self.inner.stream()).0
```
gatesn: I don't think you can do this... you drop the SyncOnDrop thing.

I wonder if instead just want to `impl DevicePtr for CudaDeviceBuffer` so we share the same trait?

0ax1: _~~I understand you want me to do the unsafe cast here to the first field and bet on that the compiler does not reorder~~_

vortex-array/src/buffer.rs
0ax1: done

vortex-cuda/src/device_buffer.rs
0ax1:
```
/// Abstraction over [CudaSlice]/[CudaView]
pub trait DevicePtr<T>: DeviceSlice<T> {
    /// Retrieve the device pointer with the intent to read the device memory
    /// associated with it.
    ///
    /// Implementations of this method should ensure `stream` waits for any previous
    /// writes of this memory before continuing (do not need to wait for any previous reads).
    ///
    /// The [SyncOnDrop] item of the return tuple should be dropped **after** the read of
    /// the [sys::CUdeviceptr] is scheduled.
    ///
    /// In most cases you can use like:
    /// ```ignore
    /// let (src, _record_src) = src.device_ptr(&stream);
    /// ```
    /// Which will drop the [SyncOnDrop] at the end of the scope.
    fn device_ptr<'a>(&'a self, stream: &'a CudaStream) -> (sys::CUdeviceptr, SyncOnDrop<'a>);
}

impl<T> DevicePtr<T> for CudaSlice<T> {
    fn device_ptr<'a>(&'a self, stream: &'a CudaStream) -> (sys::CUdeviceptr, SyncOnDrop<'a>) {
        if self.stream.context().is_managing_stream_synchronization() {
            if let Some(write) = self.write.as_ref() {
                stream.ctx.record_err(stream.wait(write));
            }
        }
        (
            self.cu_device_ptr,
            SyncOnDrop::record_event(&self.read, stream),
        )
    }
}
```
This is fine, we only record a read event which does not lead to a synchronize. The purpose of these events is multi-stream synchronization which is not an issue we have. The device ptr doesn't change over the lifetime of an owned `CudaSlice`. I'll cache the retrieval of the pointer in `new` such that we don't need to call through `device_ptr` of cudarc multiple times.

gatesn (approved):

PR #6061, opened 2026-01-20
chore[array]: replace to_canonical with to execute.

joseph-isaacs:
# Breaks

Adds an execution context argument to append_to_builder.


vortex-array/src/arrays/varbin/vtable/canonical.rs
```diff
@@ -11,19 +11,23 @@ use vortex_error::VortexResult;
...
+pub(crate) fn varbin_to_canonical(
+    array: &VarBinArray,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<Canonical> {
     // Zero the offsets first to ensure the bytes buffer starts at 0
     let array = array.clone().zero_offsets();
     let (dtype, bytes, offsets, validity) = array.into_parts();
 
-    let offsets = offsets.to_primitive();
+    let offsets = offsets.execute::<PrimitiveArray>(ctx)?;
```
gatesn, line 30: I wonder if you want `offsets.execute::<Buffer<O>>` and it verifies all non-null?

joseph-isaacs: good idea

joseph-isaacs: this sides a to_host?

joseph-isaacs: Also this assumes that the array is primitive not decimal of the type storage type?

gatesn (approved):

PR #6064, opened 2026-01-20
Remove the python feature from vortex-error

AdamGS:
Instead of using a feature just for an error type, we introduce a new error type that helps us localize the functionality into `vortex-python`, similarly to the `PyVortex` type.

Like #6059, this should also help with CI and simplify our internal dependency matrix.


vortex-python/src/arrays/builtins/struct_.rs
```diff
@@ -24,7 +25,7 @@ impl EncodingSubclass for PyStructArray {
 #[pymethods]
 impl PyStructArray {
     /// Returns the given field of the struct array.
-    pub fn field(self_: PyRef<'_, Self>, name: &str) -> PyResult<PyArrayRef> {
+    pub fn field(self_: PyRef<'_, Self>, name: &str) -> Result<PyArrayRef, PyVortexError> {
```
gatesn: Can you type alias a `PyVortexResult`

AdamGS:
done

robert3005 (approved):

gatesn (approved):

PR #6066, opened 2026-01-20
Don't run clippy on each crate with no default features

AdamGS:
Opened a discussion on the community slack, will require and admin to make it not mandatory.


AdamGS:
This is not a bottle neck for now, closing it for now, might revisit in the future.

PR #6067, opened 2026-01-20
fix: Handle empty constant arrays in ScalarFnConstantRule

robert3005:
fix #6065 


gatesn (commented):
Seems like a weird place to live? Should optimizer.rs just replace length 0 arrays with Canonical::empty?

gatesn (approved):

robert3005:
I am not sure if blanket replacement is good. There's probably a different short cirtcuit you can perform depending on the rule

PR #6068, opened 2026-01-20
Expression::execute

gatesn:
Port Expression::execute to return ExecutionResult enum of Canonical or ConstantArray


joseph-isaacs:
see: https://github.com/vortex-data/vortex/pull/6076

PR #6079, opened 2026-01-21
chore: guard IoBuf::slice_owned against pointer overflow

caicancai:
Add a debug-only check to IoBuf::slice_owned so we fail fast if range.start is so large that base_ptr.add(range.start) would overflow

### DIFF
--- a/vortex-io/src/io_buf.rs
+++ b/vortex-io/src/io_buf.rs
@@ -43,6 +43,18 @@ pub unsafe trait IoBuf: Unpin + 'static {
             range.end,
             self.bytes_init()
         );
+        #[cfg(debug_assertions)]
+        {
+            let max_offset = isize::MAX as usize;
+            // Guard against pointer arithmetic overflow when we call `base_ptr.add(begin)`
+            // in `OwnedSlice::read_ptr`. `range.end` is already bounded by `bytes_init`,
+            // so only `range.start` needs this extra guard.
+            assert!(
+                range.start <= max_offset,
+                "Range start ({}) too large for pointer arithmetic",
+                range.start
+            );
+        }
 
         OwnedSlice {
             buf: self,


vortex-io/src/io_buf.rs
joseph-isaacs, line 152: see?

vortex-io/src/io_buf.rs
```diff
@@ -43,6 +43,15 @@ pub unsafe trait IoBuf: Unpin + 'static {
...
         );
+        #[cfg(debug_assertions)]
+        {
+            let max_offset = isize::MAX as usize;
+            assert!(
+                range.start <= max_offset,
+                "Range start ({}) too large for pointer arithmetic",
+                range.start
+            );
+        }
```
joseph-isaacs, line 57: can you add a comment explaining this? and why do we not need this on range.end?

vortex-io/src/io_buf.rs
caicancai: Added a debug-only check inside `IoBuf::slice_owned` to guard against range.start values so large that calling base_ptr.add(begin) would overflow, mirroring the existing safeguard in `OwnedSlice::read_ptr`. This provides a “fail fast” signal at the slice construction time, making it easier to catch illegal ranges before they reach the unsafe pointer arithmetic and preventing undefined behavior in release builds.

vortex-io/src/io_buf.rs
caicancai: done

vortex-io/src/io_buf.rs
caicancai: Of course, I also struggled with this change for a long time, mainly depending on community feedback.🤔

joseph-isaacs:
We don't need this. since start < end < bytes size

caicancai:
thanks, closing

PR #6080, opened 2026-01-21
Disable incremental compilation in CI

AdamGS:
Sccache doesn't cache incrementally compiled crates, and it takes more work to generate the artifacts that support it. It is enabled by default for workspace crates in `dev` profile.


AdamGS:
I think its the right thing to do, but I can't say I can measure a real difference, and I would rather have less load bearing pieces in CI

PR #6083, opened 2026-01-21
Remove cargo hack partitioning for no-features lint step

AdamGS:
Simplifies CI and only adds maybe 10s to this step compared to the last `develop` run.


robert3005 (approved):

AdamGS:
re running this to make sure its really runtime wise

PR #6090, opened 2026-01-22
fix[vortex-array]: lookup target fields by name in StructCastPushdownRule

asubiotto:
The previous code was assuming that target fields were in the same order as source fields and all target fields existing in the source schema were the first k fields. This would cause incorrect query results.

### DIFF
--- a/vortex-array/src/arrays/struct_/vtable/rules.rs
+++ b/vortex-array/src/arrays/struct_/vtable/rules.rs
@@ -50,38 +50,27 @@ impl ArrayParentReduceRule<StructVTable> for StructCastPushDownRule {
         _child_idx: usize,
     ) -> VortexResult<Option<ArrayRef>> {
         let target_fields = parent.options.as_struct_fields();
-        let source_field_count = array.fields.len();
-        let target_field_count = target_fields.nfields();
+        let mut new_fields = Vec::with_capacity(target_fields.nfields());
 
-        // Target must have at least as many fields as source
-        vortex_ensure!(
-            target_field_count >= source_field_count,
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/struct_/vtable/rules.rs
```diff
@@ -50,38 +50,27 @@ impl ArrayParentReduceRule<StructVTable> for StructCastPushDownRule {
         _child_idx: usize,
     ) -> VortexResult<Option<ArrayRef>> {
         let target_fields = parent.options.as_struct_fields();
-        let source_field_count = array.fields.len();
-        let target_field_count = target_fields.nfields();
+        let mut new_fields = Vec::with_capacity(target_fields.nfields());
 
-        // Target must have at least as many fields as source
```
asubiotto, line 56: I removed this because I'm not sure we want this restriction. Why disallow this?

vortex-array/src/arrays/struct_/vtable/rules.rs
joseph-isaacs, line 1: think we can add test local to this rule?

joseph-isaacs (approved):
Otherwise good to go

vortex-array/src/arrays/struct_/vtable/rules.rs
asubiotto: Done.

AdamGS:
(windows failure was unrelated)

joseph-isaacs:
@asubiotto do you have merge perms?

asubiotto:
> @asubiotto do you have merge perms?

Yes

PR #6091, opened 2026-01-22
Update GitHub labels

gatesn:


0ax1 (commented):
should the bench labels then also have the prefix?
```

jobs:
  label_trigger:
    runs-on: ubuntu-latest
    timeout-minutes: 120
    if: ${{ contains(github.event.head_commit.message, '[benchmark]') || github.event.label.name == 'action/benchmark' && github.event_name == 'pull_request' }}
    steps:
      # We remove the benchmark label first so that the workflow can be re-triggered.
      - uses: actions-ecosystem/action-remove-labels@v1
        if: ${{ github.event.pull_request.head.repo.full_name == 'vortex-data/vortex' }}
        with:
          labels: benchmark <= action/benchmark?
          fail_on_error: true

  bench:
    needs: label_trigger
```

0ax1 (approved):

PR #6094, opened 2026-01-22
chore[fuzz]: add fuzzer for arb-compressed arrays

joseph-isaacs:
Add a new fuzzer pass for checking compressed array decompress. Use arb to generic the compressed array so we can sample that space more effectively


fuzz/src/compress.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// Run the compressed encoding canonicalization fuzzer.
+///
+/// Returns:
+/// - `Ok(true)` - keep in corpus
+/// - `Ok(false)` - reject from corpus
+/// - `Err(_)` - a bug was found
+#[allow(clippy::result_large_err)]
+pub fn run_compress_roundtrip(fuzz: FuzzCompressRoundtrip) -> crate::error::VortexFuzzResult<bool> {
```
robert3005, line 64: Add a use statement

robert3005 (approved):
Really need a clippy lint for imports

PR #6100, opened 2026-01-22
chore: expose arrow and scalar methods

blaginin:


robert3005 (approved):

robert3005:
You need to adjust visibility of some of the types inside as well

vortex-python/src/lib.rs
```diff
@@ -19,7 +19,7 @@ mod iter;
 mod object_store_urls;
 mod python_repr;
 mod registry;
-pub(crate) mod scalar;
+pub mod scalar;
```
joseph-isaacs, line 22: shall we export only the ones we need?

PR #6102, opened 2026-01-22
fallible scalar_at and validity

joseph-isaacs:
This PR makes several validity and scalar-related methods fallible (returning VortexResult<T> instead of T).                                                                       
                                                                                                                                                                                     
  ---                                                                                                                                                                                
  Breaking Changes                                                                                                                                                                   
                                                                                                                                                                                     
  Validity methods (in vortex-array)                                                                                                                                                 
```  ┌──────────────────────────────────────┬─────────┬───────────────────────┐                                                                                                         
  │                Method                │ Before  │         After         │                                                                                                         
  ├──────────────────────────────────────┼─────────┼───────────────────────┤                                                                                                         
  │ is_valid(index)                      │ -> bool │ -> VortexResult<bool> │                                                                                                         
  ├──────────────────────────────────────┼─────────┼───────────────────────┤                                                                                                         
  │ all_valid(len)                       │ -> bool │ -> VortexResult<bool> │                                                                                                         
  ├──────────────────────────────────────┼─────────┼───────────────────────┤                                                                                                         
  │ all_invalid(len)                     │ -> bool │ -> VortexResult<bool> │                                                                                                         
  ├──────────────────────────────────────┼─────────┼───────────────────────┤                                                                                                         
  │ slice(range)                         │ -> Self │ -> VortexResult<Self> │                                                                                                         
  ├──────────────────────────────────────┼─────────┼───────────────────────┤                                                                                                         
  │ patch(len, offset, indices, patches) │ -> Self │ -> VortexResult<Self> │                                                                                                         
  └──────────────────────────────────────┴─────────┴───────────────────────┘                                                                                                         
  Array patch methods                                                                                                                                                                
  ┌─────────────────────────┬─────────┬───────────────────────┐                                                                                                                      
  │          Type           │ Before  │         After         │                                                                                                                      
  ├─────────────────────────┼─────────┼───────────────────────┤                                                                                                                      
  │ BoolArray::patch()      │ -> Self │ -> VortexResult<Self> │                                                                                                                      
  ├─────────────────────────┼─────────┼───────────────────────┤                                                                                                                      
  │ PrimitiveArray::patch() │ -> Self │ -> VortexResult<Self> │                                                                                                                      
  ├─────────────────────────┼─────────┼───────────────────────┤                                                                                                                      
  │ DecimalArray::patch()   │ -> Self │ -> VortexResult<Self> │   
  ```

### DIFF
--- a/vortex-array/src/array/mod.rs
+++ b/vortex-array/src/array/mod.rs
@@ -99,7 +99,7 @@ pub trait Array:
     fn encoding_id(&self) -> ArrayId;
 
     /// Performs a constant-time slice of the array.
-    fn slice(&self, range: Range<usize>) -> ArrayRef;
+    fn slice(&self, range: Range<usize>) -> VortexResult<ArrayRef>;
 
     /// Wraps the array in a [`FilterArray`] such that it is logically filtered by the given mask.
     fn filter(&self, mask: Mask) -> VortexResult<ArrayRef>;
@@ -216,7 +216,7 @@ impl Array for Arc<dyn Array> {
     }
 
     #[inline]
-    fn slice(&self, range: Range<usize>) -> ArrayRef {
+    fn slice(&self, range: Range<usize>) -> VortexResult<ArrayRef> {
         self.as_ref().slice(range)
     }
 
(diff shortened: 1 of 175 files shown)


vortex-array/src/array/mod.rs
gatesn: I think you may have left a lot of these things around. And annoyingly, if we used `.expect` not `.vortex_expect` then clippy would have told us.

a10y (approved):

gatesn (dismissed):
Let's just pause on this one until we know whether optimizer ruels should be fallible. If not, I think Array::filter / Array::slice don't need to be either.

PR #6109, opened 2026-01-22
feat[gpu]: scalar encodings

joseph-isaacs:


vortex-cuda-macros/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+///
+/// When used inside the `vortex-cuda` crate itself, use:
+///
+/// ```ignore
+/// #[vortex_cuda::test(crate = crate)]
+/// async fn test_my_cuda_kernel() {
+///     // test body
+/// }
+/// ```
+#[proc_macro_attribute]
```
0ax1: Not sure we want this. Just figured out yesterday that we can set cfg features at `build.rs` time. (see nvcomp). We should probably have one CUDA feat that gets set there. Thing is that the tests will still show up if we check for nvcc at runtime.

vortex-cuda/kernels/scalar_kernel.cuh
```diff
@@ -0,0 +1,60 @@
...
+    uint64_t array_len,
+    Op op
+) {
+    const uint32_t elements_per_block = 2048;
+    const uint64_t block_start = static_cast<uint64_t>(blockIdx.x) * elements_per_block;
+    const uint64_t block_end = (block_start + elements_per_block < array_len)
+        ? (block_start + elements_per_block)
+        : array_len;
+
+    // Vectorized loop - process 16 bytes per iteration for better memory throughput.
```
0ax1, line 30: Did I leave that comment. In any case the ops here are not vectorized.

joseph-isaacs: From FoR

0ax1: Yeah this isn't true. I thought this would happen on some archs. But CUDA only can do vec ops on loads and stores. It really is only the unrolling doing the trick here.

vortex-cuda/src/kernel/encodings/alp.rs
```diff
@@ -0,0 +1,179 @@
...
+    // Build launch args: input, output, f, e, length
+    launch_builder.arg(&input_view);
+    launch_builder.arg(&output_view);
+    launch_builder.arg(&f);
+    launch_builder.arg(&e);
+    launch_builder.arg(&array_len_u64);
+
+    // Launch kernel
+    let _cuda_events =
+        launch_cuda_kernel_impl(&mut launch_builder, CU_EVENT_DISABLE_TIMING, array_len)?;
```
0ax1, line 103: if we rework the launcher logic, it'd be nice to not record events by default for each launch.

0ax1 (approved):
nice!

vortex-cuda/src/kernel/encodings/alp.rs
joseph-isaacs: agreed

vortex-cuda/kernels/scalar_kernel.cuh
joseph-isaacs: Ill remove in the next one

PR #6112, opened 2026-01-22
chore: custom kernel launcher API to remove macro

0ax1:
@joseph-isaacs claude draft, wdyt? I'll do a clean up if we think this is the right direction.

### DIFF
--- a/vortex-cuda/src/kernel/launcher.rs
+++ b/vortex-cuda/src/kernel/launcher.rs
@@ -1,0 +1,281 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Type-safe CUDA kernel launcher that replaces the `launch_cuda_kernel!` macro.
+//!
+//! This module provides a builder-based API for launching CUDA kernels without
+//! requiring a macro. The key challenge is that CUDA's `cuLaunchKernel` requires
+//! a `void**` (array of pointers to arguments), so we need to store argument values
+//! and provide stable pointers to them.
+//!
+//! # Example
+//!
+//! ```ignore
+//! use vortex_cuda::kernel::KernelLauncher;
+//! use cudarc::driver::sys::CUevent_flags::CU_EVENT_DISABLE_TIMING;
+//!
+//! // Instead of:
+//! // launch_cuda_kernel!(
+//! //     execution_ctx: ctx,
+//! //     module: "for",
+//! //     ptypes: &[array.ptype()],
+//! //     launch_args: [cuda_view, reference, array_len],
+//! //     event_recording: CU_EVENT_DISABLE_TIMING,
+//! //     array_len: array.len()
+//! // );
+//!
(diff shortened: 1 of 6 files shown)


vortex-cuda/src/kernel/launcher.rs
```diff
@@ -0,0 +1,281 @@
...
+//! //     array_len: array.len()
+//! // );
+//!
+//! // Use:
+//! let events = KernelLauncher::new(ctx, "for", &[array.ptype()])?
+//!     .arg_view(&cuda_view)
+//!     .arg(&reference)
+//!     .arg(&array_len)
+//!     .event_flags(CU_EVENT_DISABLE_TIMING)
+//!     .launch(array.len())?;
```
joseph-isaacs, line 33: this is great

vortex-cuda/src/kernel/launcher.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// * `module_name` - Name of the PTX module (e.g., "for")
+    /// * `ptypes` - Primitive types that determine the kernel variant (e.g., `&[PType::U32]`)
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the kernel function cannot be loaded.
+    pub fn new(
+        ctx: &'a CudaExecutionCtx,
+        module_name: &str,
+        ptypes: &[PType],
```
joseph-isaacs, line 90: This is not enough for non prim?

vortex-cuda/src/kernel/launcher.rs
```diff
@@ -0,0 +1,281 @@
...
+            storage: Vec::new(),
+            event_flags: None,
+        })
+    }
+
+    /// Adds a scalar argument to the kernel launch.
+    ///
+    /// Supports any type that implements `DeviceRepr` and `Copy` and fits in 8 bytes:
+    /// - Integers: u8, u16, u32, u64, i8, i16, i32, i64
+    /// - Floats: f32, f64
```
joseph-isaacs, line 106: i128/i256?

joseph-isaacs: What about str literal?

vortex-cuda/src/kernel/launcher.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// ```ignore
+    /// let cuda_view = buffer_handle.cuda_view::<f32>()?;
+    /// launcher.arg_view(&cuda_view);
+    /// ```
+    pub fn arg_cuda_view<T: DeviceRepr>(mut self, view: &CudaView<'_, T>) -> Self {
+        // Get the device pointer value (CUdeviceptr is a u64)
+        // The _sync guard is dropped immediately, but that's fine since we're just
+        // reading the pointer value, not scheduling any work yet.
+        let (device_ptr, _sync) = view.device_ptr(self.stream);
+        self.storage.push(device_ptr);
```
joseph-isaacs, line 159: what keeps the view alive?

PR #6113, opened 2026-01-23
fix[vortex-array]: fix filter of sliced list

asubiotto:
Previously when filtering a list array, we could create a mask longer than the sliced elements range, causing a "Filter mask length mismatch" panic.


a10y (approved):
Great find!

PR #6117, opened 2026-01-23
chore[array]: remove vortex-vector dependency

joseph-isaacs:
We should now be able to remove vortex-vector and -compute if needed


joseph-isaacs:
some big regressions, I will explore

PR #6119, opened 2026-01-23
fix[list]: filter list execute parent

blaginin:
Fixed a bad element mask generated in a filter op


joseph-isaacs (approved):
Well done US

PR #6122, opened 2026-01-23
chore: clean up cond cuda compilation

0ax1:
Always include CUDA code in tests and benchmarks if the environment has a CUDA installation. This does not require the user to pass a feature flag but is determined at build.rs time.

### DIFF
--- a/vortex-cuda/benches/for_cuda.rs
+++ b/vortex-cuda/benches/for_cuda.rs
@@ -1,397 +1,412 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-#![allow(clippy::unwrap_used)]
-#![allow(clippy::cast_possible_truncation)]
+//! CUDA benchmarks for FoR decompression.
+//!
+//! This module is only compiled when CUDA/nvcc is available.
+//! The `cuda_available` cfg is automatically set by build.rs.
 
-use std::mem::size_of;
-use std::time::Duration;
+// When CUDA is not available, provide a stub main.
+#[cfg(not(cuda_available))]
+fn main() {}
 
-use criterion::BenchmarkId;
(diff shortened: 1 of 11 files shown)


vortex-cuda/benches/for_cuda.rs
```diff
@@ -1,397 +1,409 @@
...
 
-#![allow(clippy::unwrap_used)]
-#![allow(clippy::cast_possible_truncation)]
+//! CUDA benchmarks for FoR decompression.
+
+// When CUDA is not available, provide a stub main.
+#[cfg(not(cuda_available))]
+fn main() {}
+
+#[cfg(cuda_available)]
```
joseph-isaacs, line 10: 😢

vortex-cuda/benches/for_cuda.rs
joseph-isaacs, line 1: can we not cfg in the  the toml

vortex-cuda/nvcomp/src/lib.rs
joseph-isaacs, line 20: ?

joseph-isaacs, line 21: ?

vortex-cuda/benches/for_cuda.rs
0ax1: no, to fully disable the compilation we'd need to use a feature. but the nice part about the setup is that all the non test/benchmark code still gets compiled against the cuda c api on any system.

vortex-cuda/nvcomp/src/lib.rs
0ax1: no clippy guarantees for the rust-bindgen code from cuda

0ax1: this is for rust bindgen code

joseph-isaacs (approved):

PR #6132, opened 2026-01-23
[DO NOT MERGE] Rewrite the Vortex benchmarks website

connortsui20:
A **MINIMUM** of 2 weeks to do this if someone (me) decides to do this, simply due to the amount of data migration, benchmarks + CI refactoring, and of course the website itself (actually getting it to work + making it look nice).

Maybe this is overkill but to be honest anything less than this that fulfills the same goals we want will still take forever so the opportunity cost is not actually that high imo.

You don't have to read the whole thing to get the idea, probably just the first 300 lines and then skim the rest of the document.


BENCHMARK_WEBSITE_REDESIGN.md
```diff
@@ -0,0 +1,1669 @@
...
+    PRIMARY KEY (commit_hash, chart)
+);
+
+-- Index for filtering by chart
+CREATE INDEX idx_tpch_sf1_chart ON tpch_sf1(chart);
+```
+
+```sql
+-- clickbench table
+CREATE TABLE clickbench (
```
joseph-isaacs, line 273: would be nice to build in the cpu/io used. e.g.  ec2 type and s3 or nvme.

connortsui20: That would need to be a separate chart(s) right? Otherwise I don't know how you would display all the information

connortsui20:
Im going to close this for now since we have a much faster benchmark site now, but I still believe that it is worth having a well structured relational DBMS backing everything since it would allow us to easily add new benchmark groups and benchmarks in the future without ever having to worry about breaking things.

For example, it would be nice to see our microbenchmarks on this website as well, maybe not for outsiders who are interested in Vortex, but specifically for the developers (us) to track very slow regressions over time.

PR #6136, opened 2026-01-25
feat: Zstd CUDA nvCOMP decompression

0ax1:


vortex-cuda/benches/dict_cuda.rs
```diff
@@ -1,364 +1,378 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-#[cfg(not(cuda_available))]
```
0ax1, line 4: Changed this for both benchmarks to only cfg the main fn. So the benchmarks get always compiled but only run on CUDA systems.

vortex-cuda/src/kernel/arrays/dict.rs
```diff
@@ -105,13 +105,13 @@ async fn execute_dict_prim_typed<V: DeviceRepr + NativePType, I: DeviceRepr + Na
...
     } else {
-        ctx.copy_buffer_to_device_async::<V>(values_buffer)?.await?
+        ctx.move_to_device::<V>(values_buffer)?.await?
     };
 
     let codes_device = if codes_buffer.is_on_device() {
         codes_buffer
     } else {
-        ctx.copy_buffer_to_device_async::<I>(codes_buffer)?.await?
+        ctx.move_to_device::<I>(codes_buffer)?.await?
```
joseph-isaacs, line 114: onto/into device?

encodings/zstd/src/array.rs
joseph-isaacs, line 1: explain these changes

vortex-cuda/src/kernel/encodings/zstd.rs
```diff
@@ -0,0 +1,406 @@
...
+            stream.cu_stream().cast(),
+        )
+        .map_err(|e| vortex_err!("nvcomp decompress_async failed: {}", e))?;
+    }
+
+    // Unconditionally copy back to the host as Zstd arrays are fully
+    // self-contained. They neither have any parent or child encodings.
+    let host_buffer = CudaDeviceBuffer::new(exec.device_output)
+        .copy_to_host(Alignment::new(1))?
+        .await?;
```
joseph-isaacs, line 274:
```suggestion
    // Unconditionally copy back to the host as Zstd arrays are fully
    // self-contained. They neither have any parent or child encodings.
    // TODO(..): don't copy to host once VarBinView supports BufferHandle
    let host_buffer = CudaDeviceBuffer::new(exec.device_output)
        .copy_to_host(Alignment::new(1))?
        .await?;
```

encodings/zstd/src/array.rs
joseph-isaacs: Also explain the compact compressor the dict problem!!

0ax1: added a comment to the `without_dict` fns

joseph-isaacs (approved):
🔥

PR #6141, opened 2026-01-26
fix: fastlanes canonicalize benchmarks

0ax1:


robert3005:
What's the fix here? Why do we need to recreate the array?

joseph-isaacs (approved):

0ax1:
> What's the fix here? Why do we need to recreate the array?

Make the input more opaque, but more importantly don't allocate in the bench closure.

PR #6143, opened 2026-01-26
fix: make StructReader lazy

a10y:
Instead of applying mask eagerly it constructs a MaskArray and lets the application happen in the execution.

### DIFF
--- a/vortex-layout/src/layouts/struct_/reader.rs
+++ b/vortex-layout/src/layouts/struct_/reader.rs
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use std::collections::BTreeSet;
+use std::ops::Not;
 use std::ops::Range;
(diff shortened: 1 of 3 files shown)


vortex-layout/src/layouts/struct_/reader.rs
```diff
@@ -343,15 +343,14 @@ impl LayoutReader for StructReader {
         Ok(Box::pin(async move {
             if let Some(validity_fut) = validity_fut {
                 let (array, validity) = try_join!(projected, validity_fut)?;
-                let mask = Mask::from_buffer(validity.to_bool().bit_buffer().not());
 
                 // If root expression was a pack, then we apply the validity to each child field
                 if is_pack_merge {
                     let struct_array = array.to_struct();
```
joseph-isaacs, line 351: this does compute? maybe?

a10y:
i assume this has been covered by something else by now

PR #6145, opened 2026-01-26
feature: Add fastlanes bit unpacking cuda kernels

robert3005:
I'm happy to big bang the whole thing in one pr or we can merge this as an intermediary step that generates the cuda kernels


joseph-isaacs:
shall we have this in -cuda for now?

robert3005:
We should not, we should move things from vortex-cuda, there's too many things there

joseph-isaacs:
Can we have a test running a kernel?

vortex-cuda/kernels/.gitignore
joseph-isaacs, line 1: why not commit these?

vortex-cuda/src/kernel/encodings/bitpacked.rs
```diff
@@ -0,0 +1,383 @@
...
+    A::Physical: DeviceRepr + Send + Sync + 'static,
+{
+    let array_len = array.len();
+    assert!(array_len > 0);
+
+    // Get the exponent factors from the lookup tables.
+    let bit_width = array.bit_width();
+    let buffer = array.packed().clone();
+    let validity = array.validity()?;
+    let offset = array.offset() as usize;
```
joseph-isaacs: does this handle non-zero offset?

joseph-isaacs: and non 1024 aligned len

vortex-cuda/src/kernel/encodings/bitpacked.rs
```diff
@@ -0,0 +1,383 @@
...
+    #[case::bw_55(55)]
+    #[case::bw_56(56)]
+    #[case::bw_57(57)]
+    #[case::bw_58(58)]
+    #[case::bw_59(59)]
+    #[case::bw_60(60)]
+    #[case::bw_61(61)]
+    #[case::bw_62(62)]
+    #[case::bw_63(63)]
+    fn test_cuda_bitunpack_u64(#[case] bit_width: u8) -> VortexResult<()> {
```
joseph-isaacs, line 359: check with different offset + len

vortex-cuda/src/kernel/encodings/bitpacked.rs
joseph-isaacs: into_parts?

joseph-isaacs (approved):

PR #6146, opened 2026-01-26
Feature: Use `array.filter()` instead of `filter(array)` compute

connortsui20:
I also removed some benchmarks (3) that now no longer benchmark anything useful. We'll soon want a way to benchmark the execution of arbitrary arrays, and they should all share similar benchmarking logic.

In a follow up PR (https://github.com/vortex-data/vortex/pull/6147) I will remove every occurrence of the old `filter` compute function! mwa ha ha


vortex-btrblocks/src/string.rs
```diff
@@ -493,7 +492,8 @@ mod tests {
         let strings = strings.finish_into_varbinview();
 
         let compressed = StringCompressor::compress(&strings, false, MAX_CASCADE, &[])?;
```
joseph-isaacs, line 498: also check that if can actually be serialised. (you don't have any filter/slice nodes) that we might write to disk

vortex-array/src/arrays/assertions.rs
```diff
@@ -26,6 +26,26 @@ macro_rules! assert_nth_scalar {
...
+/// Asserts that the scalar at position `$n` in array `$arr` is null.
+///
+/// # Example
+///
+/// ```ignore
+/// let arr = PrimitiveArray::from_option_iter([Some(1), None, Some(3)]);
+/// assert_nth_scalar_null!(arr, 1);
+/// ```
+#[macro_export]
+macro_rules! assert_nth_scalar_is_null {
```
joseph-isaacs, line 38: didn't we already have this?

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
```diff
@@ -27,43 +27,31 @@ use crate::validity::Validity;
...
     vec![false, false, false],
-    0,
-    false
-)]
-#[case::degenerate_all_null(
-    4,
-    Validity::AllInvalid,
-    vec![true, true, false, true],
-    3,
-    true
```
joseph-isaacs, line 52: why is it gone?

vortex-btrblocks/src/string.rs
joseph-isaacs: maybe even assert on the display tree?

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
connortsui20: this was a bs test

vortex-array/src/arrays/assertions.rs
connortsui20: not that I can see anywhere?

vortex-btrblocks/src/string.rs
connortsui20: wdym by assert on the display tree? like the string itself?

connortsui20:
Note: it seems like only the bitpacking early filter benchmark suffers, everything else is an improvement

vortex-btrblocks/src/string.rs
joseph-isaacs: yes. just make sure that no compute arrays would be created here.

connortsui20: im not really sure how I would programatically assert this... Here is the display tree in this test:

```
compression tree: root: vortex.varbinview(utf8?, len=100) nbytes=1.63 kB (100.00%) [nulls=99]
  metadata: EmptyMetadata
  buffer (align=1): 17 B (1.04%)
  buffer (align=16): 1.60 kB (98.16%)
  validity: vortex.bool(bool, len=100) nbytes=13 B (0.80%) [sum=1u64]
    metadata: BoolMetadata { offset: 0 }
    buffer (align=1): 13 B (100.00%)
```

connortsui20: is it literally "filter" and "slice" not in compressed?

connortsui20: I just did this:

```rs
        let display = compressed.display_tree().to_string().to_lowercase();
        assert!(
            !display.contains("filter"),
            "display_tree should not contain 'filter'"
        );
        assert!(
            !display.contains("slice"),
            "display_tree should not contain 'slice'"
        );
```

encodings/pco/benches/pco.rs
```diff
@@ -1,62 +0,0 @@
...
-    let pco_array = PcoArray::from_primitive(&values, 3, 0).unwrap();
-    let mask = (0..size)
-        .map(|_| rng.random_bool(selectivity))
-        .collect::<BitBuffer>();
-
-    bencher
-        // Be sure to reconstruct the mask to avoid cached set_indices
-        .with_inputs(|| (&pco_array, Mask::from_buffer(mask.clone())))
-        .bench_refs(|(pco_array, mask)| {
-            filter(pco_array.to_canonical().unwrap().as_ref(), mask).unwrap()
```
joseph-isaacs, line 60: why don't we test filter then canonical?

encodings/runend/benches/run_end_filter.rs
joseph-isaacs, line 1: please justify

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
```diff
@@ -74,20 +62,15 @@ fn test_filter_degenerate_list_size_zero(
...
-    } else {
-        let filtered_fsl = filtered.as_::<FixedSizeListVTable>();
-        assert_eq!(filtered_fsl.list_size(), 0, "list_size should remain 0");
-        assert_eq!(
-            filtered_fsl.elements().len(),
-            0,
-            "no elements expected for list_size=0"
-        );
+    // For degenerate list_size=0, verify each element is an empty list (not null).
+    for i in 0..expected_len {
```
joseph-isaacs: you can use assert_arrays_eq here too?

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
```diff
@@ -296,29 +253,29 @@ fn test_filter_all_null_various_list_sizes() {
...
     let fsl10 = FixedSizeListArray::new(elements10, 10, Validity::AllInvalid, 5);
     let mask10 = Mask::AllTrue(5);
     let filtered10 = filter(fsl10.as_ref(), &mask10).unwrap();
     assert_eq!(filtered10.len(), 5);
-    // Check that all elements are null
-    assert!(filtered10.scalar_at(0).unwrap().is_null());
-    assert!(filtered10.scalar_at(4).unwrap().is_null());
+    // Check that all elements are null.
+    assert_nth_scalar_is_null!(filtered10, 0);
+    assert_nth_scalar_is_null!(filtered10, 4);
```
joseph-isaacs, line 283: and here

vortex-btrblocks/src/string.rs
joseph-isaacs: I want thinking either array the tree is exactly that above or just try to seralized into bytes and see if doesn't fail (maybe both)?

connortsui20: That feels like a very finicky test to be honest, and even right now imo it's not a useful test and I would rather just remove this

encodings/pco/benches/pco.rs
connortsui20: Like I said in the PR description I think we want a consolidated way to measure execution speed of any arbitrary array, which imo is more complicated than just setting up a simple execution context and running execute a bunch of times, we want more fine-grained detail.

We want to know how long it takes for optimization, what rules get applied, how long each step of execution takes, etc. I'd rather deal with this separately and get rid of this now so we don't have a bunch of codspeed regressions/improvements that don't actually mean anything.

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
connortsui20: I don't think that is more succinct / readable than this?

encodings/runend/benches/run_end_filter.rs
connortsui20: see above

vortex-btrblocks/src/string.rs
connortsui20: I made this the last commit and I'm thinking about just removing this check

joseph-isaacs: please can you add a serialisation check

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
```diff
@@ -130,21 +116,10 @@ fn test_filter_all_null_array() {
...
-        "Expected null at index 0"
-    );
-    assert!(
-        filtered_const.scalar_at(1).unwrap().is_null(),
-        "Expected null at index 1"
-    );
+    // Verify the result is an array of nulls.
+    assert_eq!(filtered.len(), 2, "All-null FSL should produce 2 elements");
+    assert_nth_scalar_is_null!(filtered, 0);
+    assert_nth_scalar_is_null!(filtered, 1);
```
joseph-isaacs, line 122: compare to a fsl of nulls?

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
connortsui20:
```rs
    let expected0 = FixedSizeListArray::new(
        PrimitiveArray::empty::<i32>(Nullability::NonNullable).into_array(),
        0,
        Validity::AllInvalid,
        2,
    );
    assert_arrays_eq!(filtered0, expected0);
```

I don't think this is better

vortex-array/src/arrays/fixed_size_list/tests/filter.rs
connortsui20: Same as above, it is much harder to read

vortex-btrblocks/src/string.rs
connortsui20: done

vortex-btrblocks/src/float.rs
```diff
@@ -622,7 +622,8 @@ mod scheme_selection_tests {
         builder.append_nulls(95);
         let array = builder.finish_into_primitive();
         let compressed = FloatCompressor::compress(&array, false, 3, &[])?;
-        assert!(compressed.is::<SparseVTable>());
+        // Verify the compressed array preserves values.
+        assert_eq!(compressed.len(), 100);
```
joseph-isaacs, line 638: here too?

joseph-isaacs (approved):

connortsui20:
Whoops looks like this broke the compression throughput and clickbench benchmarks, unfortunately theres no warning

PR #6147, opened 2026-01-26
Chore: Replace `filter(array)` with `array.filter()` in most places

connortsui20:
This is based on top of https://github.com/vortex-data/vortex/pull/6146/, that needs to merge first.

Simply changes the syntax so that we use the new `FilterArray`, so this is purely a cosmetic change after #6146.

Note that I only made the change when this gets called inside one of the old compute functions, and the code for all of the old filter compute are going to be moved around soon so this is fine.

There are just a few places (around 8) where we call the old `filter` _outside_ of the old compute kernels. I've left those as is so it's easy to search for them later. We still have to figure out what to do with them.

### DIFF
--- a/vortex-array/src/compute/filter.rs
+++ b/vortex-array/src/compute/filter.rs
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+// TODO(connor): REMOVE THIS FILE!
+
 use std::sync::LazyLock;
 
 use arcref::ArcRef;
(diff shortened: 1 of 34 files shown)


vortex-array/src/compute/filter.rs
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+// TODO(connor): REMOVE THIS FILE!
```
connortsui20, line 4: Eventually we will be able to remove this file completely, but only after we figure out how to deal with the 8 or so stray filter arrays

connortsui20:
Note that develop is currently broken because of an issue with `FilterArray` in validities, and this PR will make it easier to solve said problem.

danking (approved):
yolo

PR #6151, opened 2026-01-26
fix: sample multiples of 1024

danking:
Vortex now samples vectors in multiples of 1024 elements so that fastlanes bitpacking does not see artificially bad compression ratios due to padding bytes.

See: https://spiraldb.slack.com/archives/C07BV3GKAJ2/p1769463416246819

### DIFF
--- a/vortex-btrblocks/src/lib.rs
+++ b/vortex-btrblocks/src/lib.rs
@@ -93,7 +93,15 @@ impl Default for GenerateStatsOptions {
     }
 }
 
+/// The size of each sampled run.
 const SAMPLE_SIZE: u32 = 64;
+/// The number of sampled runs.
+///
+/// # Warning
+///
+/// The product of SAMPLE_SIZE and SAMPLE_COUNT should be (roughly) a multiple of 1024 so that
+/// fastlanes bitpacking of sampled vectors does not introduce (large amounts of) padding.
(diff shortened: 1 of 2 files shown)


vortex-btrblocks/src/lib.rs
```diff
@@ -186,25 +198,25 @@ fn estimate_compression_ratio_with_sampling<T: Scheme + ?Sized>(
...
-        // We want to sample about 1% of data, while keeping a minimal sample of 640 values.
-        let sample_count = usize::max(
-            (source_len / 100)
-                / usize::try_from(SAMPLE_SIZE).vortex_expect("SAMPLE_SIZE must fit in usize"),
-            10,
+        // We want to sample about 1% of data, while keeping a minimal sample of 1024 values.
+        let approximately_one_percent = (source_len / 100)
+            / usize::try_from(SAMPLE_SIZE).vortex_expect("SAMPLE_SIZE must fit in usize");
+        let sample_count = u32::max(
+            nearest_multiple_of_16(
```
robert3005: you could use https://doc.rust-lang.org/std/primitive.u32.html#method.next_multiple_of

danking: oh nice!

robert3005 (approved):

PR #6155, opened 2026-01-27
fix[fuzz]: single sesssion for the fuzzer

joseph-isaacs:


fuzz/src/gpu/mod.rs
```diff
@@ -118,11 +122,8 @@ pub async fn run_compress_gpu(fuzz: FuzzCompressGpu) -> VortexFuzzResult<bool> {
         }
     };
 
-    // 2. Create CUDA execution context
```
0ax1, line 121: 🥲

0ax1 (approved):

Issue #6078, opened 2026-01-21
VortexMetrics Memory Growth: Children Accumulate Indefinitely in Long-Lived Sessions Resulting in OOM


gatesn:
### Discussed in https://github.com/vortex-data/vortex/discussions/5946

<div type='discussions-op-text'>

<sup>Originally posted by **sgrebnov** January 13, 2026</sup>
### Issue Description

`VortexMetrics` accumulates child metrics indefinitely when `VortexSession` is reused across multiple queries, causing memory leaks in production workloads. The `child_with_tags()` method stores strong references to children that are never cleared, leading to memory growth in high-throughput scenarios.

### Expected Behavior

No memory growth when `VortexSession` is re-used across multiple queries

### Actual Behavior

Memory growth when `VortexSession` is re-used across multiple queries

Current behavior:
<img width="452" height="241" alt="image" src="https://github.com/user-attachments/assets/fd3158b6-b90b-4c7e-8678-791a50bbfdd3" />

With temporary disabled metrics (memory is higher due to larger data)

<img width="865" height="220" alt="image" src="https://github.com/user-attachments/assets/f93095d2-2160-4989-9e6f-f1826902c0ea" />
 

```
pub fn child_with_tags(&self, additional_tags: impl Into<DefaultTags>) -> Self {
        let child = Self::new_with_tags(self.inner.default_tags.merge(&additional_tags.into()));
        // TEMPORARY: Disabled child tracking to prevent memory leak during queries
        // self.inner.children.write().push(child.clone());
        child

```

### Reproduction Steps

**Ownership chain:**
```
VortexSession (Arc<SessionVars>)
  └─> DashMap<TypeId, Box<dyn SessionVar>>
      └─> Box<VortexMetrics> (root)
          └─> Arc<Inner>
              └─> children: RwLock<Vec<VortexMetrics>>  ← NEVER CLEARED
```

```rust
pub fn child_with_tags(&self, tags: impl IntoIterator<Item = (String, String)>) -> Self {
    let child = Self { inner: Arc::new(...) };
    self.inner.children.write().push(child.clone());  // ← Strong reference, never removed
    child
}
```

Each query creates many children (one per partition × files), but they're never cleared:
1. `VortexSession` is designed to be long-lived and reused
2. Children stay alive as long as root `VortexMetrics` exists
3. Root exists as long as any `VortexSession` clone exists

**Questions / Notes**
1. VortexSession to be per-query (new session each time) or long-lived?
[...]

PR #6156, opened 2026-01-27
Make metrics more local to prevent unbounded memory usage

AdamGS:
Part of #6078.

I have a bigger branch I'm experimenting with in the background, but I figured it makes to start breaking it into smaller parts while I figure out bigger issues.

This PR is a step towards simplifying the metrics API, starting by trying to make it scoped to scan so holding instances of `VortexFile` won't make memory grow in an unbounded way.

## Changes in this PR
1. I've also made some code less reliant on `Arc<dyn VortexReadAt>` in favor of a statically dispatched `R: VortexReadAt + Clone`, otherwise we end up with an `Arc<Arc<Arc<dyn VortexReadAt>>>` if not worse.
2. Made IO-level metrics opt-in, they are enabled by default for vortex-datafusion and `InstrumentedReadAt` is still part of the public API, maybe we should make it even more visible?
3. Make `VortexOpenOptions::open_read` public. I don't want to break the public API here, but it seems more useful than `VortexOpenOptions::read` that requires an `Arc` and just delegates to it anyway.


vortex-metrics/src/macros.rs
```diff
@@ -1,27 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: This is not used, if any downstream user wants this its easy enough to replicate and it delegates to a 3rd part crate.

vortex-metrics/src/session.rs
```diff
@@ -1,17 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: We don't want global metrics, at least not the way they work now.

onursatici (approved):

PR #6157, opened 2026-01-27
fix[gpu]: synchronous copy to host

0ax1:
The previous copy to host operations was synchronous but did not wait for other previous operations on the stream to complete. `cuMemcpyDtoH_v2` runs on the default stream and therefore raced with the buffer handle associated stream.


vortex-cuda/src/device_buffer.rs
```diff
@@ -111,6 +111,8 @@ impl<T: DeviceRepr + Send + Sync + 'static> DeviceBuffer for CudaDeviceBuffer<T>
 
     /// Synchronous copy of CUDA device to host memory.
     ///
+    /// The copy is not started before other operations on the streams are completed.
```
joseph-isaacs, line 114: can you update this doc str more?

joseph-isaacs (approved):

PR #6158, opened 2026-01-27
coalesced request align all segments

onursatici:
We align the coalesced range start down to the max alignment of all requests. That makes `coalesced_start % max_alignment == 0`, just like the file start. Because segment offsets are aligned in the file, their offsets relative to the coalesced start remain aligned, so each buffer stays properly aligned without per‑buffer realignment.

example:
bytes in the file:
[x, x, x, x, x, x, A, A, A, A, A, x, B]
A aligned to 2
B aligned to 4

coalesced read both to a 2 aligned buffer:
[ A, A, A, A, A, x, B]
A aligned to 2
B is not aligned to 4

with this change we read this to the coalesced buffer:
[x, x, A, A, A, A, A, x, B]
A aligned to 2
B aligned to 4

### DIFF
--- a/vortex-file/src/read/driver.rs
+++ b/vortex-file/src/read/driver.rs
@@ -213,7 +213,7 @@ impl State {
         let mut requests = vec![first_req];
         let mut current_start = requests[0].offset;
         let mut current_end = requests[0].offset + requests[0].length as u64;
-        let alignment = requests[0].alignment;
+        let mut max_alignment = requests[0].alignment;
 
         let mut keys_to_remove = Vec::new();
         let mut found_new_requests = true;
@@ -256,12 +256,16 @@ impl State {
                     // Calculate what the new range would be if we include this request
                     let new_start = current_start.min(req_offset);
                     let new_end = current_end.max(req_end);
-                    let new_total_size = new_end - new_start;
+                    let new_alignment = max_alignment.max(req.alignment);
+                    let align = *new_alignment as u64;
+                    let aligned_start = new_start - (new_start % align);
+                    let new_total_size = new_end - aligned_start;
 
                     // Check if the coalesced request would exceed max_size
                     if new_total_size <= window.max_size {
                         current_start = new_start;
                         current_end = new_end;
+                        max_alignment = new_alignment;
 
                         let req = self
                             .polled_requests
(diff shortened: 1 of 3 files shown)


vortex-file/src/read/driver.rs
```diff
@@ -256,12 +256,16 @@ impl State {
...
                     let new_end = current_end.max(req_end);
-                    let new_total_size = new_end - new_start;
+                    let new_alignment = max_alignment.max(req.alignment);
+                    let align = *new_alignment as u64;
+                    let aligned_start = new_start - (new_start % align);
+                    let new_total_size = new_end - aligned_start;
 
                     // Check if the coalesced request would exceed max_size
                     if new_total_size <= window.max_size {
                         current_start = new_start;
```
joseph-isaacs: shall we negate and continue?

vortex-file/src/read/driver.rs
```diff
@@ -256,12 +256,16 @@ impl State {
                     // Calculate what the new range would be if we include this request
                     let new_start = current_start.min(req_offset);
                     let new_end = current_end.max(req_end);
-                    let new_total_size = new_end - new_start;
+                    let new_alignment = max_alignment.max(req.alignment);
+                    let align = *new_alignment as u64;
+                    let aligned_start = new_start - (new_start % align);
```
joseph-isaacs, line 272: `rem_euclid` (https://doc.rust-lang.org/std/primitive.i32.html#method.rem_euclid)

vortex-file/src/read/driver.rs
joseph-isaacs, line 1: shall we test this?

joseph-isaacs, line 222: doc str?

vortex-file/src/read/driver.rs
onursatici: checked that it has the same behaviour with % for usize

vortex-file/src/read/driver.rs
```diff
@@ -206,14 +206,23 @@ impl State {
         None
     }
 
+    /// Coalesce nearby requests into a single range while aligning the range start down to the
+    /// maximum alignment of all included requests.
```
joseph-isaacs:
```suggestion
    /// Coalesce nearby requests into a single range request aligned to widest alignment of all coalesced requests
```

vortex-file/src/read/driver.rs
```diff
@@ -206,14 +206,23 @@ impl State {
...
+    ///
+    /// Example (file offsets):
+    /// [x, x, x, x, x, x, A, A, A, A, A, x, B]
+    /// A aligned to 2, B aligned to 4
+    /// Coalesced range starts at 4, so the buffer is:
+    /// [x, x, A, A, A, A, A, x, B]
```
joseph-isaacs, line 220: why does this have 9 values in?

onursatici: you want to read both A and B, and B is only one byte in this example. It has the leading 2 byte padding to keep B 4-aligned

joseph-isaacs (approved):

PR #6161, opened 2026-01-27
Fix some metrics propegation in DF

AdamGS:
Fixes some issues I've introduced #6156


AdamGS:
closing this change in favor of #6164

PR #6164, opened 2026-01-27
Use built-in DF cache for file metadata and fix metrics wiring

AdamGS:
Instead of using our own custom file cache, we delegate to DataFusion's file metadata cache to stash footers we've already read, allowing users to re-use other existing DF code and extensions while reducing the amount of code we have to care about.

It also makes wiring metrics simpler, as we don't cache the layout readers created during `infer_schema` or `infer_stats`, which is desired because `FileFormat` is long living while `FileSource` isn't, so we don't want it to hold a metrics instance.

### DIFF
--- a/benchmarks/datafusion-bench/src/main.rs
+++ b/benchmarks/datafusion-bench/src/main.rs
@@ -17,6 +17,7 @@ use datafusion::prelude::SessionContext;
 use datafusion_bench::format_to_df_format;
 use datafusion_bench::metrics::MetricsSetExt;
 use datafusion_physical_plan::ExecutionPlan;
+use datafusion_physical_plan::collect;
 use futures::StreamExt;
 use parking_lot::Mutex;
 use tokio::fs::File;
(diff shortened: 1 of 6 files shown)


benchmarks/datafusion-bench/src/main.rs
```diff
@@ -300,10 +301,11 @@ pub async fn execute_query(
...
     let df = ctx.sql(query).await?;
 
-    let physical_plan = df.clone().create_physical_plan().await?;
-    let result = df.collect().await?;
+    let task_ctx = Arc::new(df.task_ctx());
+    let plan = df.create_physical_plan().await?;
+    let result = collect(plan.clone(), task_ctx).await?;
 
-    Ok((result, physical_plan))
+    Ok((result, plan))
```
AdamGS, line 308: because metrics aren't global anymore, this is the way to get the actual metrics generated during the query.

vortex-datafusion/src/persistent/cache.rs
```diff
@@ -1,174 +1,68 @@
...
                 .sum::<usize>()
         })
-        .sum::<usize>();
+        .unwrap_or(0);
 
     let root_layout = footer.layout();
-    let layout_size = size_of::<DType>()
+    let layout_size = size_of_val(footer.dtype())
         + root_layout.metadata().len()
         + root_layout.segment_ids().len() * size_of::<SegmentId>();
```
a10y, line 173: i think this might be an under-estimate since it doesn't account for the size of the child layouts? the old code i think always was wrong in this way

a10y (approved):
Nice to delete some code

vortex-datafusion/src/persistent/cache.rs
AdamGS: thanks for the fix!

PR #6170, opened 2026-01-27
fix: cuda ptx folder

joseph-isaacs:
This is VERY hacky


vortex-cuda/build.rs
```diff
@@ -10,13 +10,24 @@ use std::path::Path;
...
-    if !is_cuda_available() {
-        return;
-    }
+    // Declare the cfg so rustc doesn't warn about unexpected cfg.
+    println!("cargo::rustc-check-cfg=cfg(cuda_available)");
 
     let manifest_dir = env::var("CARGO_MANIFEST_DIR").expect("Failed to get manifest dir");
     let kernels_dir = Path::new(&manifest_dir).join("kernels");
 
+    // Always emit the kernels directory path as a compile-time env var so any binary
```
0ax1, line 16: This is probably only visible within the crate right?

vortex-cuda/src/kernel/mod.rs
```diff
@@ -196,16 +201,14 @@ impl KernelLoader {
...
     /// The full path to the PTX file
-    fn ptx_path_for_module(module_name: &str) -> VortexResult<PathBuf> {
-        let manifest_dir = env::var("CARGO_MANIFEST_DIR")
-            .map_err(|e| vortex_err!("Failed to get manifest dir: {}", e))?;
-        Ok(Path::new(&manifest_dir)
-            .join("kernels")
-            .join(format!("{}.ptx", module_name)))
+    fn ptx_path_for_module(module_name: &str) -> PathBuf {
+        // VORTEX_CUDA_KERNELS_DIR is set by build.rs at compile time
+        Path::new(env!("VORTEX_CUDA_KERNELS_DIR")).join(format!("{}.ptx", module_name))
```
0ax1: Surprised compile time env vars work.

0ax1 (approved):

vortex-cuda/build.rs
joseph-isaacs: Its used in the fuzzer

0ax1: that's not my point. the env var is still only visible to the crate at build time. it's not like that env var propagates to clients

0ax1: the abs path is compiled into vortex cuda at build time

PR #6173, opened 2026-01-27
Feature: Add more `Filter<_>` implementations

connortsui20:
Allows us to filter by a `MaskValues` if we want to, as well as by indices or slices directly.

This is just a nice to have for using the code in `vortex-compute` on top of the incoming changes in https://github.com/vortex-data/vortex/pull/6152

### DIFF
--- a/vortex-compute/src/filter/bitbuffer.rs
+++ b/vortex-compute/src/filter/bitbuffer.rs
@@ -9,6 +9,7 @@ use vortex_buffer::get_bit_unchecked;
 use vortex_buffer::set_bit_unchecked;
 use vortex_buffer::unset_bit_unchecked;
 use vortex_mask::Mask;
+use vortex_mask::MaskValues;
 
 use crate::filter::Filter;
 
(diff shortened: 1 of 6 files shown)


vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -9,55 +9,128 @@ use vortex_buffer::get_bit_unchecked;
...
 use vortex_mask::Mask;
+use vortex_mask::MaskValues;
 
 use crate::filter::Filter;
 
 impl Filter<Mask> for &BitBuffer {
     type Output = BitBuffer;
 
     fn filter(self, selection_mask: &Mask) -> BitBuffer {
+        match selection_mask {
```
lwwmanning, line 23: check length

vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -9,55 +9,128 @@ use vortex_buffer::get_bit_unchecked;
...
+impl Filter<[usize]> for &BitBuffer {
+    type Output = BitBuffer;
+
+    fn filter(self, indices: &[usize]) -> BitBuffer {
+        let bools = self.inner().as_ref();
+        let bit_offset = self.offset();
+
+        // FIXME(ngates): this is slower than it could be!
+        BitBufferMut::collect_bool(indices.len(), |idx| {
+            let idx = *unsafe { indices.get_unchecked(idx) };
```
lwwmanning, line 64: do we trust these indices?

vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -9,55 +9,128 @@ use vortex_buffer::get_bit_unchecked;
...
+
+    fn filter(self, slices: &[(usize, usize)]) -> BitBuffer {
+        let bools = self.inner().as_ref();
+        let bit_offset = self.offset();
+        let output_len: usize = slices.iter().map(|(start, end)| end - start).sum();
+
+        let mut out = BitBufferMut::with_capacity(output_len);
+        for &(start, end) in slices {
+            for idx in start..end {
+                out.append(get_bit(bools, bit_offset + idx));
```
lwwmanning: this seems like it would be slow?

vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -9,55 +9,128 @@ use vortex_buffer::get_bit_unchecked;
...
         }
+        out.freeze()
     }
 }
 
 impl Filter<Mask> for &mut BitBufferMut {
     type Output = ();
 
     fn filter(self, selection_mask: &Mask) {
+        match selection_mask {
```
lwwmanning, line 109: check lengths?

vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -9,55 +9,128 @@ use vortex_buffer::get_bit_unchecked;
...
-    // FIXME(ngates): this is slower than it could be!
-    BitBufferMut::collect_bool(indices.len(), |idx| {
-        let idx = *unsafe { indices.get_unchecked(idx) };
-        get_bit(bools, bit_offset + idx)
-    })
+impl Filter<[usize]> for &mut BitBufferMut {
+    type Output = ();
+
+    fn filter(self, indices: &[usize]) {
+        let bools = self.inner().as_slice();
```
lwwmanning, line 146: check lengths

vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -9,55 +9,128 @@ use vortex_buffer::get_bit_unchecked;
...
+        let bit_offset = self.offset();
+
+        // FIXME(ngates): this is slower than it could be!
+        *self = BitBufferMut::collect_bool(indices.len(), |idx| {
+            let idx = *unsafe { indices.get_unchecked(idx) };
```
lwwmanning, line 151: do we actually trust the indices to be valid?

vortex-compute/src/filter/buffer.rs
```diff
@@ -36,25 +33,38 @@ impl<T: Copy> Filter<Mask> for &Buffer<T> {
     type Output = Buffer<T>;
 
     fn filter(self, selection_mask: &Mask) -> Buffer<T> {
-        assert_eq!(
```
lwwmanning, line 39: 👀

vortex-compute/src/filter/bitbuffer.rs
connortsui20: We don't need to check the length since it gets delegated to the impl below, and I think it's fine if we don't check length for the all true and all false case since it doesn't matter what the length is

vortex-compute/src/filter/bitbuffer.rs
connortsui20: That `idx` is into the length of `indices`, so yes because it must be less than `indices.len()`

vortex-compute/src/filter/bitbuffer.rs
connortsui20: Same as above

vortex-compute/src/filter/bitbuffer.rs
connortsui20: yes probably, I think nick forgot to add a fixme comment here, Ill add it

vortex-compute/src/filter/bitbuffer.rs
connortsui20: Don't need to check lengths for filtering by indices

vortex-compute/src/filter/bitbuffer.rs
connortsui20: we don't, but get_bit will panic on out of bounds

vortex-compute/src/filter/buffer.rs
connortsui20: similar reasons as above

joseph-isaacs:
Do these have tests?

connortsui20:
@joseph-isaacs It's the same tests as before, this just makes it so that the previously private impls are now public

joseph-isaacs (approved):

PR #6180, opened 2026-01-27
feat: reduce size of segments containing fixed-size-list arrays

danking:
I also verified this on some fixed-size-list data I have locally and, indeed, I see more reasonable segment sizes (~1MiB).

### DIFF
--- a/vortex-dtype/src/dtype.rs
+++ b/vortex-dtype/src/dtype.rs
@@ -358,7 +358,7 @@ impl DType {
                         .checked_add(element_size)
                         .vortex_expect("sum of field sizes is bigger than usize");
                 }
-                Some(sum);
+                Some(sum)
             }
             Extension(ext) => ext.storage_dtype().element_size(),
         }
--- a/vortex-dtype/src/dtype.rs
+++ b/vortex-dtype/src/dtype.rs
@@ -10,6 +10,7 @@ use std::sync::Arc;
 use DType::*;
 use itertools::Itertools;
 use static_assertions::const_assert_eq;
+use vortex_error::VortexExpect;
 use vortex_error::vortex_panic;
 
 use crate::ExtDType;
@@ -345,13 +346,21 @@ impl DType {
             Decimal(decimal, _) => {
                 Some(DecimalType::smallest_decimal_value_type(decimal).byte_width())
             }
+            Utf8(_) | Binary(_) | List(..) => None,
             FixedSizeList(elem_dtype, list_size, _) => {
                 elem_dtype.element_size().map(|s| s * *list_size as usize)
             }
+            Struct(fields, ..) => {
+                let mut sum = 0_usize;
+                for f in fields.fields() {
+                    let element_size = f.element_size()?;
+                    sum = sum
+                        .checked_add(element_size)
+                        .vortex_expect("sum of field sizes is bigger than usize");
+                }
+                Some(sum);
+            }
             Extension(ext) => ext.storage_dtype().element_size(),
-            // We could sum the elment_size of the fields of the struct, but it is not clear this is
-            // a useful number to know.
-            Struct(..) | Utf8(_) | Binary(_) | List(..) => None,
         }
     }
 
--- a/vortex-dtype/src/dtype.rs
+++ b/vortex-dtype/src/dtype.rs
@@ -333,9 +333,9 @@ impl DType {
         }
     }
 
-    /// Returns the number of bytes occupied by a single scalar of this fixed-width non-struct type.
+    /// Returns the number of bytes occupied by a single scalar of this fixed-width type.
     ///
-    /// For non-fixed-width or struct types, return None.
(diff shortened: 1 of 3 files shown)


vortex-dtype/src/dtype.rs
```diff
@@ -331,6 +332,29 @@ impl DType {
         }
     }
 
+    /// Returns the number of bytes occupied by a single scalar of this fixed-width non-struct type.
+    ///
+    /// For non-fixed-width or struct types, return None.
```
connortsui20: Do we want to calculate approximate size for struct types if the fields are all fixed size?

vortex-file/src/strategy.rs
```diff
@@ -97,6 +97,7 @@ impl WriteStrategyBuilder {
             RepartitionWriterOptions {
                 block_size_minimum: ONE_MEG,
                 block_len_multiple: self.row_block_size,
+                block_size_target: Some(ONE_MEG),
```
connortsui20, line 106: It would be nice to have a comment on this for why this is a nice number (maybe on the const definition itself?)

vortex-layout/src/layouts/repartition.rs
```diff
@@ -31,9 +32,43 @@ pub struct RepartitionWriterOptions {
...
+    ///
+    /// When set, for fixed-width types (where [`DType::element_size`] is `Some`), the effective
+    /// `block_len_multiple` is reduced so that each block does not exceed this byte target.
+    /// This prevents oversized segments for types with large per-element sizes (e.g.,
+    /// `FixedSizeList(f64, 1000)` which is ~8000 bytes per element).
+    ///
+    /// When `None`, `block_len_multiple` is used as-is regardless of element size.
+    ///
+    /// [`DType::element_size`]: vortex_dtype::DType::element_size
+    pub block_size_target: Option<u64>,
```
connortsui20, line 47: Not super familiar with the repartition strategies but do we ever need to account for the little bit of extra space the data might take up? If no then a little mention about that here would be nice

vortex-layout/src/layouts/repartition.rs
```diff
@@ -31,9 +32,43 @@ pub struct RepartitionWriterOptions {
...
+    fn effective_block_len(&self, dtype: &DType) -> usize {
+        let Some(block_size_target) = self.block_size_target else {
+            return self.block_len_multiple;
+        };
+        match dtype.element_size() {
+            Some(elem_size) if elem_size > 0 => {
+                // `div_ceil` ensures we overshoot the block_size_target; therefore preventing
+                // `write_stream` from combining adjacent 0.9 MiB chunks into one 1.8 MiB chunk.
+                let max_rows = usize::try_from(block_size_target.div_ceil(elem_size as u64))
+                    .unwrap_or(usize::MAX);
```
connortsui20, line 66: For example if it perfectly divides the target might the extra data cause a tiny amount of overflow? I don't actually know it this is bad or not so maybe this isn't an issue.

vortex-layout/src/layouts/repartition.rs
```diff
@@ -31,9 +32,43 @@ pub struct RepartitionWriterOptions {
...
+                self.block_len_multiple.min(max_rows).max(1)
```
lwwmanning, line 67: this should help with large binary values too, no?

vortex-dtype/src/dtype.rs
danking: There's a comment in the match statement. I'm not sure how meaningful that number is. Like, do we assume the values are packed as tightly as possible? Or is each field aligned correctly? I dunno.

connortsui20: Well this is supposed to be an upper bound right? So it could be as simple as taking the field with the max field size and multiply that by the number of fields to mimic "padding". That's at least better than nothing right?

danking: I think of it as an exact size. The reason bools are 1 byte is because a Scalar of type bool has to hold a single byte of data (there's no smaller addressable type in Rust).

I suppose defining it allows us to choose good segment sizes for `FSL<Struct<...>>`. I'll just define it as the sum of the types. AFAIK, an `FSL<struct<a: u8, b: u32>>` is serialized in, roughly, N + 4N bytes, not 8N bytes.

vortex-file/src/strategy.rs
danking: I added a comment to the const definition that represents my understanding of the choice. Does it address your concerns?

vortex-layout/src/layouts/repartition.rs
danking: I changed this comment. Does it address your concerns?

vortex-layout/src/layouts/repartition.rs
danking: See other comment.

connortsui20 (approved):
Very nice

PR #6181, opened 2026-01-27
Fix: Comment on a failed benchmark

connortsui20:
Let's see if this works...


danking (commented):
this seems right to me. Can we test by breaking SQL benchmarks and running them on this PR?

connortsui20:
There are 2 that are already broken on develop 😢 

Clickbench and Compression Throughput

connortsui20:
oh I have to do it for both actions

danking (approved):

connortsui20:
Seems like this didn't work for the compression benchmark?

PR #6182, opened 2026-01-27
feat[cuda]: basic filter kernel for primitive

a10y:
We add a filter kernel for the Primitive types.

It seems that in the GPGPU space, what we call `filter` is generally called either "stream compaction", "scanning", or "selection".

Stream compaction is non-trivial because unlike most scalar operations, a thread needs to coordinate with other threads so it knows where in the output buffer to start writing.

The best algorithm for this was published in 2009 and seems to be the basis of how NVIDIA's `cub::DeviceSelect::Flagged()` kernel works. I've ported a very minimal version of this here that we will need to improve on.

## Algorithm

Most GPU-based stream reduction algorithms work in two passes:

1. Every thread block is assigned a fixed range of the input mask. The threads count the number of set bits in it section of the mask, and puts that partial sum into a slot in some shared memory
2. A parallel prefix sum algorithm is run to turn the partial sums into prefix sums. The prefix sums are the starting positions in the output buffer where each thread is supposed to write its data

This is a fairly complex algorithm and I've only implemented a pretty simple variant of it here. In particular

* We hard-peg to 1 thread block, 512 worker threads. This puts a really hard upper bound on the amount of parallelism we get. The reason mostly is that exchange across thread blocks is pretty gnarly and I haven't worked it out yet. Seems like it requires some spin locking?
* We're doing serial prefix sum rn on one worker.

# More info

* [`cub::DeviceSelect::Flagged` docs](https://nvidia.github.io/cccl/cub/api/structcub_1_1DeviceSelect.html#_CPPv4I0000EN3cub12DeviceSelect7FlaggedE11cudaError_tPvR6size_t14InputIteratorT12FlagIterator15OutputIteratorT20NumSelectedIteratorTN4cudaSt7int64_tE12cudaStream_t)
* [Slides](https://www.highperformancegraphics.org/previous/www_2009/presentations/billeter-efficient.pdf)
* [Paper](https://www.cse.chalmers.se/~uffe/streamcompaction.pdf)
* [Very old blog post](https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda)

### DIFF
--- a/vortex-cuda/kernels/filter_primitive.cu
+++ b/vortex-cuda/kernels/filter_primitive.cu
@@ -1,0 +1,160 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#include <cuda_runtime.h>
+#include <stdint.h>
+
+#include "config.cuh"
+
+#define MAX(a, b) (((a) > (b)) ? (a) : (b))
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+
+// Execute a filter kernel on the inputs, scattering them to the output nodes.
+// We need to perform a prefix sum so we know where to write all of the sized things.
+// We need to know how many bits come before so we can create indices instead.
+template<typename T>
+__device__ void filter_primitive(
+    const T *const __restrict input,
+    const uint8_t *const __restrict mask,
+    T *const __restrict output,
+    const uint32_t maskOffset,
+    const uint32_t maskLen
+) {
+
+    // ASSUMPTIONS:
+    //  1. We only have a single thread block active. Attempting to launch with > 1 thread block will fail and instead
+    //     we just fall back to using a single block.
+    //  2. Each thread in the block is responsible for a range of input mask values. If there are T threads in the block and N values,
+    //     this means that every thread is responsible for counting N / T values.
+    //  3. The input is provided as a bitset packed into a uint8_t array, possibly with an offset applied.
+    //
+    //
+    // ALGORITHM
+    // =========
+    //
+    // The algorithm is an implementation of a parallel prefix scan operation, similar to the DeviceSelect::Flagged() routing from CUB.
+    //
+    // It operates in two phases, or "sweeps" as you'll see them called in other documentation.
+    //
+    // The first phase occurs across all threads in the thread blocks. Each thread counts the number of true bits in its portion of the mask,
+    // and writes that into a shared memory cache.
+    //
+    // The next phase computes the prefix sum over the values in the block, allowing each thread to know which range of the output buffer
+    // it owns for the writing phase. (TODO: do this in parallel)
+    //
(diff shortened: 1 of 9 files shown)


vortex-cuda/src/kernel/arrays/filter.rs
```diff
@@ -0,0 +1,227 @@
...
+    async fn execute(
+        &self,
+        array: ArrayRef,
+        ctx: &mut CudaExecutionCtx,
+    ) -> VortexResult<Canonical> {
+        let Ok(filter_array) = array.try_into::<FilterVTable>() else {
+            vortex_bail!("FilterExecutor expected FilterArray input");
+        };
+
+        match filter_array.dtype() {
```
0ax1, line 56: guess we could go with match_simd_ptype! here

joseph-isaacs: we can also easily support f16, with a reinterpret cast to u16/i16?

vortex-cuda/src/kernel/arrays/filter.rs
```diff
@@ -0,0 +1,227 @@
...
+        let baseline = filter_array.to_canonical()?;
+
+        // Execute on CUDA
+        let cuda_result = FilterExecutor
+            .execute(filter_array.to_array(), &mut cuda_ctx)
+            .await
+            .vortex_expect("GPU decompression failed")
+            .into_primitive();
+        cuda_ctx.synchronize_stream()?;
+        let cuda_result = cuda_primitive_to_host(cuda_result)?;
```
joseph-isaacs, line 237: we have a into_host method and dont need to sync any more

vortex-cuda/kernels/filter_primitive.cu
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        // Handle final byte. Only take remainder
+        if (byteIdx == maskBytes) {
+            uint32_t remainder = (maskOffset + maskLen) % 8;
+            // Only consider the first `remainder` bits of the last byte.
+            byte = byte & ((1 << remainder) - 1);
+        }
+
+        validCount += __popc(static_cast<uint32_t>(byte));
```
joseph-isaacs, line 96: is there no u64?

a10y: there is, i just did everything with u32 b/c i figured it'd be large enough. i can switch everything to u64's though

vortex-cuda/src/kernel/arrays/filter.rs
a10y: yea it's not tricky i just needed to get something compiling

a10y: done

a10y:
#6188 is the successor, it actually uses the CUB routines directly which I think is the path to success here

PR #6183, opened 2026-01-27
feat: "hedging": issue two concurrent requests for HEAD and GET

danking:
There's a reasonable argument that this should either (a) be implemented via an object_store::HttpService or (b) implemented as a reqwest middleware. Either of those approaches can be upstreamed or made a separate open source library.

From the AnyBlob paper, Durner et al. "Exploiting Cloud Object Storage for High-Performance Analytics". VLDB Vol 16 Iss 11:

> # 2.7 Tail Latency & Request Hedging
>
> Hedging against slow responses. Missing or slow responses from storage servers are a challenge for
> users of cloud object stores. In our latency experiments, we see requests that have a considerable
> tail latency. Some requests get lost without any notice. To mitigate these issues, cloud vendors
> suggest restarting unresponsive requests,known as request hedging [ 10 , 34 ]. For example, the
> typical 16 MiB request duration is below 600ms for AWS. However, less than 5% of objects are not
> downloaded after 600ms. Missing responses can also be found by checking the first byte
> latency. Similarly to the duration, less than 5% have a first byte latency above 200ms. Hedging
> these requests does not introduce significant cost overhead.

### DIFF
--- a/vortex-io/src/file/object_store.rs
+++ b/vortex-io/src/file/object_store.rs
@@ -3,6 +3,7 @@
 
 use std::io;
 use std::sync::Arc;
+use std::time::Duration;
 
 use async_compat::Compat;
 use futures::FutureExt;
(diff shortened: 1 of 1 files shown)


AdamGS:
happy to play with it here, but I'm in favor of making it its own crate so its easier to share around, happy to pair on this once I wrap up some other stuff.

danking:
I think we should enable this, but I think we first need to make smart use of our network. Until then, this is just lipstick on a pig.

PR #6187, opened 2026-01-28
vortex-metrics v2

### LINKED ISSUE #6078: VortexMetrics Memory Growth: Children Accumulate Indefinitely in Long-Lived Sessions Resulting in OOM

AdamGS:
Closes https://github.com/vortex-data/vortex/issues/6078.

This PR reworks our metrics crate to simplify it and makes it much harder to make it use unbounded amounts of memory. The implementation is inspired by our current use of `witchcraft-metrics`, the `metrics` crate, DataFusion's metrics and some reading of the OTel spec.

It includes several changes:
1. `VortexMetrics` is replaced by the new `MetricsRegistry` trait, which allows user to integrate metrics to whatever metrics backend they have.
1. Removes metrics from the session, making it the responsibility of the user to wire it correctly between various components.
1. No more hierarchical metrics instances, just wire labels if they exist.
1. Removes the `witchcraft-metrics` dependency. It does many things we don't care about or that are opinionated in a way that I don't think is intuitive or useful for most users. It also makes our external API less dependent on other crates we have no control over.
1. Histograms are now f64 and don't use exponential decay, Timers only speak in `Duration`, and counters are u64s. It also introduces a `Gauge` type which I don't think we actually use and I might just delete.

I've tested it by running our DF benchmarks with `--show-metrics` and that seems to not change, if anyone else has made use of the metrics, I would love to help test and debug them in another context.


wasm-test/.gitignore
```diff
@@ -0,0 +1 @@
+Cargo.lock
```
AdamGS, line 1: this is just some housekeeping I ran into

vortex-metrics/src/lib.rs
```diff
@@ -6,300 +6,254 @@
...
+    pub fn gauge(self, name: impl Into<Cow<'static, str>>) -> Gauge {
+        self.registry.register_gauge(name.into(), self.labels)
+    }
 }
 
-impl<'a> Iterator for VortexMetricsIter<'a> {
-    type Item = (MetricId, &'a Metric);
+/// Default metrics registry, stores all state in-memory.
+#[derive(Default, Clone)]
+pub struct DefaultMetricsRegistry {
```
robert3005, line 145: Debating whether we want a RwLock or Mutex here

vortex-metrics/src/timer.rs
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#![allow(clippy::expect_used)]
+#![allow(clippy::unwrap_in_result)]
```
robert3005: can we remove these?

vortex-metrics/src/histogram.rs
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#![allow(clippy::expect_used)]
+#![allow(clippy::unwrap_in_result)]
```
robert3005: remove these

vortex-metrics/src/gauge.rs
```diff
@@ -0,0 +1,65 @@
...
+                .is_ok()
+            {
+                break;
+            }
+        }
+    }
+
+    /// Sets the gauge to a specific value.
+    pub fn set(&self, value: f64) {
+        _ = self.0.swap(value.to_bits(), Ordering::AcqRel);
```
robert3005, line 59: this is `store` instead

robert3005 (commented):
I made couple small comments. The thing that I am not convinced by is double arcing. But I can see how it ends up being necessary

vortex-metrics/src/lib.rs
AdamGS: I don't think we ever take a `read` lock here unless its after the scan is finished and there's no contention

vortex-metrics/src/gauge.rs
AdamGS: my understanding is that using `swap` here makes the memory ordering such that I can do `gauge.get()` and then a `gague.set(..)` and get a consistent ordering.

vortex-metrics/src/timer.rs
AdamGS: moved them to the smallest scope possible

vortex-metrics/src/gauge.rs
robert3005: Can we leave a comment here? On the surface it makes sense to use store but as you point out swap makes the store wait for a get if there’s one ongoing so you get consistent ordering

robert3005 (approved):
Feel free to merge it once you add a comment about swap in gauge

PR #6191, opened 2026-01-28
chore[visitor]: add a visit buffer handle with name

joseph-isaacs:
buffer display: `buffer: values host 10 B (align=2) (100.00%)`


a10y (approved):
👍

PR #6194, opened 2026-01-28
Fix: dont do lazy filters on write

connortsui20:
This should fix the CI failures right now.

More detailed discussion that I'll write incoming...

Basically we shouldn't be eagerly decompressing validity when filtering it, but for now to make sure things aren't broken this will have to do.

Edit: This should fix the break on spiral as well @delta003, as reported by @danking


danking:
This fixes @delta003 's issue on Spiral 4046.

danking (approved):

PR #6196, opened 2026-01-28
feat[cuda]: filter for Decimal/VarBinView

a10y:
* Wire in BufferHandles for VarBinView
* Wire up cub filtering logic for decimals/strings


vortex-array/src/arrow/executor/byte_view.rs
```diff
@@ -25,11 +25,11 @@ pub fn canonical_varbinview_to_arrow<T: ByteViewType>(
     array: &VarBinViewArray,
 ) -> VortexResult<ArrowArrayRef> {
     let views =
-        ScalarBuffer::<u128>::from(array.views().clone().into_byte_buffer().into_arrow_buffer());
+        ScalarBuffer::<u128>::from(array.views_handle().as_host().clone().into_arrow_buffer());
```
0ax1, line 28: Why is it safe to assume here that the data is on the host?

vortex-array/src/arrow/executor/byte_view.rs
```diff
@@ -44,11 +44,11 @@ pub fn execute_varbinview_to_arrow<T: ByteViewType>(
     ctx: &mut ExecutionCtx,
 ) -> VortexResult<ArrowArrayRef> {
     let views =
-        ScalarBuffer::<u128>::from(array.views().clone().into_byte_buffer().into_arrow_buffer());
```
0ax1, line 47: Why is it safe to assume here that the data is on the host?

0ax1: Should this be `to_host`?

vortex-cuda/src/kernel/filter/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+            d_output_ptr,
+            d_num_selected_ptr,
+            len,
+            stream_ptr,
+        )
+        .map_err(|e| vortex_err!("CUB filter_bitmask failed: {}", e))?;
+    }
+
+    // Wait for completion
+    await_stream_callback(stream).await?;
```
0ax1: I forget to remove this in my PR. We don't want to do a blocking wait but just return the handle. The caller can opt-into waiting for prev operations on the stream to complete, by registering its own compiler driver callback or by calling ` await_stream_callback(stream).await?;`.

vortex-cuda/src/canonical.rs
```diff
@@ -24,6 +36,29 @@ pub trait CanonicalCudaExt {
...
+                let len = struct_array.len();
+                let StructArrayParts {
+                    fields,
+                    struct_fields,
+                    validity,
+                    ..
+                } = struct_array.into_parts();
+
+                // TODO(aduffy): try_join_all
+                let mut host_fields = vec![];
```
0ax1: Could collect directly into host_fields, then it doesn't need to be mut.

vortex-cuda/src/canonical.rs
```diff
@@ -24,6 +36,29 @@ pub trait CanonicalCudaExt {
...
+                // Children should all be canonical now
+                let len = struct_array.len();
+                let StructArrayParts {
+                    fields,
+                    struct_fields,
+                    validity,
+                    ..
+                } = struct_array.into_parts();
+
+                // TODO(aduffy): try_join_all
```
0ax1: Strictly speaking, the `try_join_all` will make no difference rn. The cuda stream can only handle one copy at a time (and not multiple async). The async part is here the C/C++ fn returning immediately.

On the other hand, we already use `try_join_all` in other places. So should the underlying impl details change how we run things, it might opt into concurrent async later. Might as well just add it.

0ax1 (commented):
Looks great! Couple of questions / remarks.

vortex-array/src/arrow/executor/byte_view.rs
a10y: It's not really safe to. The tactic we've taken so far while implementing the other encodings has been to unwrap bufferhandles as host data, panicking on failure. In practice, if you're only doing CPU operations this will never panic.

If you're doing GPU operations and you hit a panic, that just means there's an operation you forgot to implement

vortex-array/src/arrow/executor/byte_view.rs
a10y: see above comment and let me know what you think

vortex-file/src/tests.rs
```diff
@@ -1641,8 +1642,9 @@ async fn main_test() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-#[cuda_tests]
-#[cfg(not(target_arch = "wasm32"))]
+
+
+#[cfg_attr(all(unix, not(target_family = "wasm")), cuda_tests)]
```
0ax1: 😅

vortex-cuda/src/kernel/filter/mod.rs
```diff
@@ -154,9 +153,6 @@ async fn filter_sized<T: DeviceRepr + CubFilterable + Send + Sync + 'static>(
         .map_err(|e| vortex_err!("CUB filter_bitmask failed: {}", e))?;
     }
 
-    // Wait for completion
-    await_stream_callback(stream).await?;
```
0ax1: ❤️

vortex-cuda/src/canonical.rs
```diff
@@ -46,11 +47,16 @@ impl CanonicalCudaExt for Canonical {
...
-                let mut host_fields = vec![];
-                for field in fields.iter() {
-                    host_fields.push(field.to_canonical()?.into_host().await?.into_array());
-                }
+                let futures = fields
+                    .iter()
+                    .cloned()
+                    .map(|field| field.to_canonical().map(|c| c.into_host()))
+                    .collect::<VortexResult<Vec<_>>>()?;
+                let host_fields = try_join_all(futures).await?;
```
0ax1, line 53: 👏

0ax1 (approved):

Issue #5767, opened 2025-12-17
DuckDB extension to use DuckDB FileSystem


gatesn:
Vortex DuckDB extension should use DuckDB's file system abstraction to list, read, and write files. This means we hook into pre-configured object store credentials, HTTP file systems, caches, etc.

gatesn:
Closed by #6198

myrrc:
Closing temporarily due to https://github.com/vortex-data/vortex/issues/6706

PR #6198, opened 2026-01-29
feat: implement DuckDB filesystem integration for Vortex file handling

iceboundrock:
* This PR addresses https://github.com/vortex-data/vortex/issues/5767
* This PR utilizes AI (codex from OpenAI) to generate code.


vortex-duckdb/src/scan.rs
```diff
@@ -203,26 +205,50 @@ fn extract_table_filter_expr(
...
+async fn open_file(
+    client_ctx: SendableClientCtx,
+    url: Url,
+    options: VortexOpenOptions,
+) -> VortexResult<VortexFile> {
+    match url.scheme() {
+        "http" | "https" | "s3" => {
+            let reader = open_duckdb_reader(client_ctx, &url);
+
+            // Fallback to the legacy object_store path for s3 if DuckDB fs isn't configured.
```
AdamGS: when will that happen? I would really love to get rid of `object_store` here, it has some underlying dependencies on tokio that can cause some weird issues

vortex-duckdb/src/copy.rs
```diff
@@ -126,9 +134,16 @@ impl CopyFunction for VortexCopyFunction {
         let array_stream = ArrayStreamAdapter::new(bind_data.dtype.clone(), rx.into_stream());
 
         let handle = SESSION.handle();
+        let ctx_ptr = SendableClientCtx(client_context.as_ptr() as usize);
         let write_task = handle.spawn(async move {
-            let mut file = async_fs::File::create(file_path).await?;
-            SESSION.write_options().write(&mut file, array_stream).await
+            // Prefer DuckDB FS (httpfs/s3/etc.), fallback to local async fs if unavailable.
+            let ctx_raw = ctx_ptr.0 as cpp::duckdb_vx_client_context;
+            if let Ok(writer) = unsafe { duckdb_fs_create_writer(ctx_raw, &file_path) } {
```
0ax1: We should probably only use the DuckDB FS and not fallback to our prev logic implicitly but return an error if the new solution fails to init. We can handle this though in a follow up on our end.

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+};
+
+void SetError(duckdb_vx_error *error_out, const std::string &message) {
+    if (!error_out) {
+        return;
+    }
+    *error_out = duckdb_vx_error_create(message.data(), message.size());
+}
+
+duckdb_state HandleException(duckdb_vx_error *error_out) {
```
0ax1: This snippet does not handle an exception really, right? We take the error, call `throw` unconditionally ourselves and then return a `DuckDBError`.

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    unique_ptr<FileHandle> handle;
+};
+
+void SetError(duckdb_vx_error *error_out, const std::string &message) {
+    if (!error_out) {
+        return;
+    }
+    *error_out = duckdb_vx_error_create(message.data(), message.size());
```
0ax1: Where does this error get freed?

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include <duckdb/common/exception.hpp>
+#include <duckdb/common/file_system.hpp>
+#include <duckdb/main/client_context.hpp>
+
+#include <cstdlib>
+#include <cstring>
+
+using namespace duckdb;
+
+namespace {
```
0ax1: Why the anonymous namespace?

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    try {
+        auto *client_context = reinterpret_cast<ClientContext *>(ctx);
+        auto &fs = FileSystem::GetFileSystem(*client_context);
+        auto handle = fs.OpenFile(path, FileFlags::FILE_FLAGS_WRITE | FileFlags::FILE_FLAGS_FILE_CREATE);
+        handle->Truncate(0);
+        return reinterpret_cast<duckdb_vx_file_handle>(new FileHandleWrapper(std::move(handle)));
+    } catch (...) {
+        HandleException(error_out);
```
0ax1: Did you mean to forward the exception being passed to `catch` to `HandleException`?

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+        SetError(error_out, "Invalid arguments to fs_get_size");
+        return DuckDBError;
+    }
+
+    try {
+        auto *wrapper = reinterpret_cast<FileHandleWrapper *>(handle);
+        *size_out = wrapper->handle->GetFileSize();
+        return DuckDBSuccess;
+    } catch (...) {
+        return HandleException(error_out);
```
0ax1: Forward the exception to `HandleException`.

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+namespace {
+struct FileHandleWrapper {
+    explicit FileHandleWrapper(unique_ptr<FileHandle> handle_p) : handle(std::move(handle_p)) {
+    }
+
+    unique_ptr<FileHandle> handle;
+};
+
+void SetError(duckdb_vx_error *error_out, const std::string &message) {
```
0ax1: I don't think we really need a wrapper for `SetError` but can just assign from `duckdb_vx_error_create`. 
This should never happen if set error is called: 
```
 if (!error_out) {
        return;
 }
```

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+    return DuckDBError;
+}
+
+} // namespace
+
+extern "C" duckdb_vx_file_handle duckdb_vx_fs_open(duckdb_vx_client_context ctx, const char *path,
+                                                     duckdb_vx_error *error_out) {
+    if (!ctx || !path) {
+        SetError(error_out, "Invalid filesystem open arguments");
```
0ax1, line 32: For all `SetError` instances assign from `duckdb_vx_error_create` directly.

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,203 @@
...
+        auto *client_context = reinterpret_cast<ClientContext *>(ctx);
+        auto &fs = FileSystem::GetFileSystem(*client_context);
+        auto matches = fs.Glob(pattern);
+
+        if (matches.empty()) {
+            return result;
+        }
+
+        result.count = matches.size();
+        result.entries = static_cast<const char **>(malloc(sizeof(char *) * matches.size()));
```
0ax1: Replace all instances of `malloc/free` with either new/delete or `duckdb_malloc/duckdb_free`(which wraps `new` and `delete`).

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,296 @@
...
+pub unsafe fn duckdb_fs_create_writer(
+    ctx: cpp::duckdb_vx_client_context,
+    path: &str,
+) -> VortexResult<DuckDbFsWriter> {
+    unsafe { DuckDbFsWriter::create(ctx, path) }
+}
+
+/// A VortexReadAt implementation backed by DuckDB's filesystem (e.g., httpfs/s3).
+pub struct DuckDbFsReadAt {
+    handle: Arc<Mutex<FsFileHandle>>,
```
0ax1: Why does this need a `Mutex` given that `FsFileHandle` is `Send` and `Sync`? If it is send and sync `Arc<FsFileHandle>` should be sufficient.

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,296 @@
...
+use crate::cpp;
+use crate::duckdb::ClientContext;
+use crate::lifetime_wrapper;
+
+const DEFAULT_COALESCE: CoalesceConfig = CoalesceConfig {
+    distance: 1024 * 1024,     // 1 MB
+    max_size: 8 * 1024 * 1024, // 8 MB
+};
+
+const DEFAULT_CONCURRENCY: usize = 64;
```
0ax1, line 38: Can you give more context, add a comment how you came up with this number?

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,296 @@
...
+            }
+
+            let runtime = RUNTIME.handle();
+            let size = runtime
+                .spawn_blocking(move || {
+                    let mut err: cpp::duckdb_vx_error = ptr::null_mut();
+                    let mut size_out: cpp::idx_t = 0;
+                    let status = unsafe {
+                        cpp::duckdb_vx_fs_get_size(
+                            handle.lock().as_ptr(),
```
0ax1: We only lock the handle for the duration of retrieving the pointer but not for the duration of `duckdb_vx_fs_get_size`. But as questioned before, isn't the file handle thread safe (send + sync) anyway?

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,296 @@
...
+            runtime
+                .spawn_blocking(move || {
+                    let mut buffer = ByteBufferMut::with_capacity_aligned(length, alignment);
+                    unsafe { buffer.set_len(length) };
+
+                    let mut err: cpp::duckdb_vx_error = ptr::null_mut();
+                    let mut out_len: cpp::idx_t = 0;
+                    let status = unsafe {
+                        cpp::duckdb_vx_fs_read(
+                            handle.lock().as_ptr(),
```
0ax1: Same here, if we really need a lock, it would need to held for duration of `duckdb_vx_fs_read`.

vortex-duckdb/src/copy.rs
```diff
@@ -29,13 +29,20 @@ use crate::RUNTIME;
...
 use crate::duckdb::CopyFunction;
 use crate::duckdb::DataChunk;
 use crate::duckdb::LogicalType;
+use crate::duckdb::duckdb_fs_create_writer;
 
 #[derive(Debug)]
 pub struct VortexCopyFunction;
 
+#[derive(Clone, Copy)]
+struct SendableClientCtx(usize);
```
0ax1: Why do we encode the pointer as a usize?

0ax1 (commented):
Thanks for taking the time to look into this. There's a couple of issues around thread safety / locking, error handling and memory leaks.

If you used AI to assist in writing the code for your PR please mention this in your PR description (see: https://github.com/vortex-data/vortex/blob/develop/CONTRIBUTING.md).

vortex-duckdb/src/copy.rs
iceboundrock: Thanks for all your comments, I will address them ASAP.

vortex-duckdb/src/scan.rs
iceboundrock: HTTP/HTTPS/S3 now route solely through DuckDB FS; legacy object_store fallback removed, and failures now bubble as errors if `httpfs` ext of DuckDB is not initialized.

vortex-duckdb/cpp/file_system.cpp
iceboundrock: added `HandleException` that rethrows the captured `std::exception_ptr`, assigns a `duckdb_vx_error`, and returns `DuckDBError`; all catch-all blocks now pass `std::current_exception()` to it.

vortex-duckdb/cpp/file_system.cpp
iceboundrock: C++ allocates errors via `duckdb_vx_error_create`; Rust side consistently frees with `duckdb_vx_error_free` in `fs_error`.

vortex-duckdb/cpp/file_system.cpp
iceboundrock: removed the anonymous namespace

vortex-duckdb/cpp/file_system.cpp
iceboundrock: retained `SetError` for central null-guarded assignment; it now directly uses `duckdb_vx_error_create`, and all call sites route through it.

vortex-duckdb/cpp/file_system.cpp
iceboundrock: follows the existing FFI pattern: `try` around core logic, `catch (...)` converts to `duckdb_vx_error` via `duckdb_vx_error_create` and returns an error/null. This mirrors `duckdb_vx_client_context_*` helpers (`vortex-duckdb/cpp/client_context.cpp`), `duckdb_data_chunk_to_string` (`vortex-duckdb/cpp/data_chunk.cpp`), and the shared error helpers in `vortex-duckdb/cpp/error.cpp`.

vortex-duckdb/cpp/file_system.cpp
iceboundrock: addressed

vortex-duckdb/cpp/file_system.cpp
iceboundrock: addressed

vortex-duckdb/cpp/file_system.cpp
iceboundrock: moved to `duckdb_malloc/duckdb_free`; error helper retained and used uniformly per existing patterns.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: set `DEFAULT_CONCURRENCY = 64` as a conservative ceiling that keeps range reads parallel while staying under common per-host limits.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: handle is now `Arc<FsFileHandle>` without additional locking; Send/Sync markers retained.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed

vortex-duckdb/src/copy.rs
iceboundrock: replaced `usize` casting with `NonNull<duckdb_vx_client_context_>` (`SendableClientCtx`) and pass the raw pointer via `as_ptr()`. This avoids truncation/aliasing issues from integer casts and preserves provenance; all FFI touchpoints now use the typed wrapper.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: handle is now `Arc<FsFileHandle>` without additional locking

vortex-duckdb/src/copy.rs
```diff
@@ -29,13 +30,26 @@ use crate::RUNTIME;
...
 use crate::duckdb::CopyFunction;
 use crate::duckdb::DataChunk;
 use crate::duckdb::LogicalType;
+use crate::duckdb::duckdb_fs_create_writer;
 
 #[derive(Debug)]
 pub struct VortexCopyFunction;
 
+#[derive(Clone, Copy)]
+struct SendableClientCtx(NonNull<cpp::duckdb_vx_client_context_>);
```
0ax1: There's 2 definitions of `SendableClientCtx` and there's an existing `ClientContext` definition in `vortex-duckdb/src/duckdb/client_context.rs`. Implement the `SendableClientCtx` functionality on the existing `ClientContext`.

vortex-duckdb/src/utils/glob.rs
```diff
@@ -1,141 +1,48 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use futures::StreamExt;
+#[cfg(test)]
```
0ax1: All of `glob.rs` is unused at this point. We can drop the whole file.

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,297 @@
...
+};
+
+// Local cap to keep remote reads parallel without overwhelming typical per-host connection limits.
+// The DuckDB httpfs extension does not expose a fixed default concurrency in the vendored sources;
+// 64 is a conservative ceiling that stays well below common cloud limits while keeping range reads
+// busy.
+const DEFAULT_CONCURRENCY: usize = 64;
+
+lifetime_wrapper!(FsFileHandle, cpp::duckdb_vx_file_handle, cpp::duckdb_vx_fs_close, [owned, ref]);
+unsafe impl Send for FsFileHandle {}
```
0ax1, line 41: I looked up in the meantime whether DuckDB file handles are thread safe. There's no synchronization happening in `FileHandle`: https://github.com/duckdb/duckdb/blob/e5fb0a7eab6bc6fd7f26bd5b8cd74c3441f3895b/src/include/duckdb/common/file_system.hpp#L71

It is though possible to use parallel reads/writes. The requirement is that all operations are position based `Read(buffer, size, location)` `Write(buffer, size, location)` and not cursor based and therefore stateful.

The write therefore needs to change away from the Cursor based approach.
```
        wrapper->handle->Seek(offset);
        wrapper->handle->Write(const_cast<uint8_t *>(buffer), len);
```

`FILE_FLAGS_PARALLEL_ACCESS`: https://github.com/duckdb/duckdb/blob/e5fb0a7eab6bc6fd7f26bd5b8cd74c3441f3895b/src/include/duckdb/common/file_open_flags.hpp#L117. 

Please double check all operations on `FSFileHandle` whether they statefully change the handle. And make sure to use `FILE_FLAGS_PARALLEL_ACCESS`.

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,297 @@
...
+
+pub unsafe fn duckdb_fs_create_writer(
+    ctx: cpp::duckdb_vx_client_context,
+    path: &str,
+) -> VortexResult<DuckDbFsWriter> {
+    unsafe { DuckDbFsWriter::create(ctx, path) }
+}
+
+/// A VortexReadAt implementation backed by DuckDB's filesystem (e.g., httpfs/s3).
+pub struct DuckDbFsReadAt {
```
0ax1: How about renaming this to `DuckDbFsReader`?

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,297 @@
...
+                    Ok::<_, VortexError>(BufferHandle::new_host(frozen))
+                })
+                .await;
+            result
+        }
+        .boxed()
+    }
+}
+
+// SAFETY: DuckDB file handles are thread-safe for reads; writes are done via DuckDbFsWriter.
```
0ax1: We need to adjust these SAFETY comments to reflect under which conditions the handle is safe to use across threads. Please double check my previous statements in ref with the DuckDB source code: https://github.com/duckdb/duckdb/blob/e5fb0a7eab6bc6fd7f26bd5b8cd74c3441f3895b/src/include/duckdb/common/file_system.hpp#L71.

vortex-duckdb/src/copy.rs
0ax1: When you add send + sync to `ClientContext` in `client_context.rs` add an explanation to the SAFETY comments why it is thread safe https://github.com/duckdb/duckdb/blob/e5fb0a7eab6bc6fd7f26bd5b8cd74c3441f3895b/src/include/duckdb/main/client_context.hpp#L67.

vortex-duckdb/cpp/file_system.cpp
0ax1: I'd suggest moving
```
static void SetError(duckdb_vx_error *error_out, const std::string &message) {
    if (!error_out) {
        return;
    }
    *error_out = duckdb_vx_error_create(message.data(), message.size());
}

static duckdb_state HandleException(std::exception_ptr ex, duckdb_vx_error *error_out) {
    if (!ex) {
        SetError(error_out, "Unknown error");
        return DuckDBError;
    }

    try {
        std::rethrow_exception(ex);
    } catch (const Exception &caught) {
        SetError(error_out, caught.what());
    } catch (const std::exception &caught) {
        SetError(error_out, caught.what());
    } catch (...) {
        SetError(error_out, "Unknown error");
    }
    return DuckDBError;
}
```
to `error.cpp` as they're not specific to file system logic.

vortex-duckdb/cpp/file_system.cpp
0ax1: assert that the error is set:  `assert(error_out && "SetError called with null error_out");`

vortex-duckdb/src/copy.rs
0ax1: Can't we use `duckdb_vx_client_context`? We typedef pointers to make them typesafe: `typedef struct duckdb_vx_client_context_ *duckdb_vx_client_context;`.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: Ok, would you like me to rename the `VortexReadAt` to `VortexReader` as well?

vortex-duckdb/cpp/file_system.cpp
iceboundrock: Ok

vortex-duckdb/cpp/file_system.cpp
iceboundrock: Ok

vortex-duckdb/src/copy.rs
iceboundrock: added `SendableClientContext` on `ClientContext` and switched copy/scan to use the typed `duckdb_vx_client_context` pointer (`vortex-duckdb/src/duckdb/client_context.rs`, `vortex-duckdb/src/copy.rs`, `vortex-duckdb/src/scan.rs`).

vortex-duckdb/src/utils/glob.rs
iceboundrock: removed the glob.rs and inlined local globbing in `vortex-duckdb/src/scan.rs` while keeping DuckDB FS glob for remote schemes.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: C++ bridge opens handles with `FILE_FLAGS_PARALLEL_ACCESS` and writes use `Write(..., location)`; Rust SAFETY comment updated to reflect position-based access (`vortex-duckdb/cpp/file_system.cpp`, `vortex-duckdb/src/duckdb/file_system.rs`).

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: clarified that parallel access is enabled in the C++ bridge and writes are position-based with no shared cursor state.

vortex-duckdb/src/copy.rs
iceboundrock: addressed

iceboundrock: migrated `SendableClientCtx` to `vortex-duckdb/src/duckdb/client_context.rs` as comments suggested.

vortex-duckdb/cpp/error.cpp
```diff
@@ -24,9 +28,35 @@ namespace vortex {
...
+        return {};
     }
     return *reinterpret_cast<std::string *>(error);
 }
 
+void SetError(duckdb_vx_error *error_out, const std::string &message) {
+    assert(error_out && "SetError called with null error_out");
+    if (!error_out) {
+        assert(error_out && "SetError called with null error_out");
+    }
```
0ax1:
```suggestion
assert(error_out == nullptr && "SetError called with null error_out");
```

vortex-duckdb/src/duckdb/client_context.rs
```diff
@@ -17,7 +17,16 @@ wrapper!(
     |_| {}
 );
 
+#[derive(Clone, Copy)]
```
0ax1, line 20: Why do we need a separate type (`SendableClientContext`) given that Send + Sync is set on `ClientContext` directly now?

vortex-duckdb/src/duckdb/file_system.rs
0ax1: Great, thanks for adding that!

vortex-duckdb/src/scan.rs
```diff
@@ -276,20 +281,47 @@ impl TableFunction for VortexTableFunction {
...
+                || s.starts_with("https://")
+                || s.starts_with("http://") =>
+            {
+                duckdb_fs_glob(ctx, s)?
+            }
+            s if s.starts_with("gs://") => {
+                vortex_bail!("GCS glob expansion not yet implemented")
+            }
+            _ => {
+                let paths = glob(file_glob.as_ref())
```
0ax1: Shouldn't we also hook into the  `duckdb_fs_glob` for local paths? Or asked differently, should we make any assumptions at all here about globbing details  of  `duckdb_fs_glob` but rather unconditionally delegate to duckdb? Whether or not gcs is supported is now a Duckdb implementation detail.

vortex-duckdb/src/duckdb/file_system.rs
0ax1: No `VortexReadAt` is a trait whereas `DuckDbFsReader` is the type implementing it. Keep as is.

0ax1 (commented):
Almost there, thanks for taking the time to address the comments!

vortex-duckdb/src/scan.rs
0ax1: My suggestion is to just do this `let file_urls = duckdb_fs_glob(..)`.

vortex-duckdb/src/duckdb/client_context.rs
iceboundrock: Because it separates ownership/lifetime semantics from thread-safe handle passing.

`ClientContext` is a `wrapper!` type with owned state and `Drop` logic. It is not `Copy` and should not be casually moved across async boundaries or spawned tasks.

On the other hand, `SendableClientContext` is a tiny, `Copy` wrapper around the raw `duckdb_vx_client_context` pointer with no destructor. It’s **just a handle**, which is ideal for passing into `async` or `spawn_blocking` code.

0ax1: Fair enough. Let's add a SAFETY comment on the wrapper stating that the same guarantees apply as for `ClientContext`.

vortex-duckdb/src/scan.rs
```diff
@@ -469,7 +502,7 @@ impl TableFunction for VortexTableFunction {
 
     fn pushdown_complex_filter(
         bind_data: &mut Self::BindData,
-        expr: &duckdb::Expression,
+        expr: &crate::duckdb::Expression,
```
0ax1: keep `duckdb::Expression`

vortex-duckdb/cpp/include/duckdb_vx/file_system.h
```diff
@@ -0,0 +1,56 @@
...
+#ifdef __cplusplus /* If compiled as C++, use C ABI */
+extern "C" {
+#endif
+
+typedef struct duckdb_vx_file_handle_ *duckdb_vx_file_handle;
+
+typedef struct {
+    const char **entries;
+    size_t count;
+} duckdb_vx_string_list;
```
0ax1: We only use this as a URI list, and this currently lives in filesystem.hpp. Let's rename this to `duckdb_vx_uri_list` to be explicit.

vortex-duckdb/src/scan.rs
iceboundrock: Ok, I’ll rename it to just `Expression` and add `use crate::duckdb::Expression;` at the top of the file.

vortex-duckdb/src/scan.rs
iceboundrock: ok, I will move the local file handling logic into `duckdb_fs_glob`

vortex-duckdb/src/duckdb/client_context.rs
iceboundrock: addressed

vortex-duckdb/cpp/include/duckdb_vx/file_system.h
iceboundrock: addressed

iceboundrock:
> Almost there, thanks for taking the time to address the comments!

Hi @0ax1 , any other comments?

0ax1:
> > Almost there, thanks for taking the time to address the comments!
> 
> Hi @0ax1 , any other comments?

Currently swamped with other work but will get back to it when I find a free moment.

vortex-duckdb/src/scan.rs
```diff
@@ -176,55 +176,60 @@ fn extract_projection_expr(init: &TableInitInput<VortexTableFunction>) -> Expres
...
+    unsafe { DuckDbFsReader::open_url(client_ctx.as_ptr(), url) }
+}
+
+async fn open_file(
+    client_ctx: SendableClientContext,
+    url: Url,
+    options: VortexOpenOptions,
+) -> VortexResult<VortexFile> {
+    match url.scheme() {
+        "http" | "https" | "s3" => {
```
gatesn: Is there any reason to match only on these schemes? Should we just give _everything_ to duckdb to handle?

gatesn:
@iceboundrock - quick question on this. Do you know if it's possible to grab the config back out from DuckDB's filesystems? 

I don't want to block this PR, don't worry(!), but we do have plans to implement a high performance I/O layer at some point and it would probably make more sense to try and configure our own I/O using duckdb configs, rather than attempt to optimize calling back into DuckDB's I/O layer.

vortex-duckdb/src/scan.rs
0ax1: +1

iceboundrock: make sense, I will update it

iceboundrock:
> @iceboundrock - quick question on this. Do you know if it's possible to grab the config back out from DuckDB's filesystems?
> 
> I don't want to block this PR, don't worry(!), but we do have plans to implement a high performance I/O layer at some point and it would probably make more sense to try and configure our own I/O using duckdb configs, rather than attempt to optimize calling back into DuckDB's I/O layer.

Could you please elaborate on what `config` refers to? If it’s S3 credentials, HTTP Proxy configuration, etc., ~~I don’t believe we can read them from DuckDB. They are stored in [`Secrets Manager`](https://duckdb.org/docs/stable/configuration/secrets_manager) and the `read` API [redacts sensitive information](https://duckdb.org/docs/stable/sql/meta/duckdb_table_functions#duckdb_secrets).~~

We probably can do the same as the httpfs extension does. Code link: 
* [1](https://github.com/duckdb/duckdb-httpfs/blob/6afe39d5be688b032146baa97e49601b50c9456e/src/httpfs_extension.cpp#L73-L105) 
* [2](https://github.com/duckdb/duckdb-httpfs/blob/main/src/create_secret_functions.cpp)

vortex-duckdb/src/scan.rs
iceboundrock: addressed

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -0,0 +1,187 @@
...
+                                            const uint8_t *buffer, idx_t *out_len,
+                                            duckdb_vx_error *error_out) {
+    if (!handle || !buffer || !out_len) {
+        SetError(error_out, "Invalid arguments to fs_write");
+        return DuckDBError;
+    }
+
+    try {
+        auto *wrapper = reinterpret_cast<FileHandleWrapper *>(handle);
+        wrapper->handle->Write(QueryContext(), const_cast<uint8_t *>(buffer), len, offset);
```
0ax1: Drop the `const_cast` and adjust the signature accordingly.

vortex-duckdb/src/duckdb/client_context.rs
```diff
@@ -43,3 +55,16 @@ impl ClientContext {
...
     }
 }
+
+impl SendableClientContext {
+    pub(crate) fn as_ptr(self) -> cpp::duckdb_vx_client_context {
+        self.0
+    }
+}
+
+// SAFETY: ClientContext is an opaque pointer owned by DuckDB and remains valid for the lifetime of
```
0ax1: These are actually not needed, as we pass around `SendableClientContext`. 

Drop:
```
unsafe impl Send for ClientContext {}
unsafe impl Sync for ClientContext {}
```

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+impl VortexWrite for DuckDbFsWriter {
+    async fn write_all<B: IoBuf>(&mut self, buffer: B) -> std::io::Result<B> {
+        let len = buffer.bytes_init();
+        let mut err: cpp::duckdb_vx_error = ptr::null_mut();
+        let mut out_len: cpp::idx_t = 0;
+        let offset = self.pos;
+
+        let status = unsafe {
```
0ax1: Use spawn blocking, instead of a blocking write.

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,307 @@
...
+        return vortex_err!("DuckDB filesystem error (unknown)");
+    }
+    let message = unsafe { CStr::from_ptr(cpp::duckdb_vx_error_value(err)) }
+        .to_string_lossy()
+        .to_string();
+    unsafe { cpp::duckdb_vx_error_free(err) };
+    vortex_err!("{message}")
+}
+
+pub fn duckdb_fs_glob(ctx: &ClientContext, pattern: &str) -> VortexResult<Vec<url::Url>> {
```
0ax1: `pub(crate)`

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,307 @@
...
+
+pub unsafe fn duckdb_fs_create_writer(
+    ctx: cpp::duckdb_vx_client_context,
+    path: &str,
+) -> VortexResult<DuckDbFsWriter> {
+    unsafe { DuckDbFsWriter::create(ctx, path) }
+}
+
+/// A VortexReadAt implementation backed by DuckDB's filesystem (e.g., httpfs/s3).
+pub struct DuckDbFsReader {
```
0ax1: `pub(crate)`

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,307 @@
...
+    handle: Arc<FsFileHandle>,
+    uri: Arc<str>,
+    size: Arc<OnceLock<u64>>,
+}
+
+impl DuckDbFsReader {
+    pub unsafe fn open_url(
```
0ax1: `pub(crate)`

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+}
+
+// SAFETY: DuckDB file handles can be used across threads when operations are position-based. The
+// C++ bridge opens handles with FILE_FLAGS_PARALLEL_ACCESS, and writes use explicit offsets, so
+// there is no shared cursor state.
+unsafe impl Send for DuckDbFsReader {}
+unsafe impl Sync for DuckDbFsReader {}
+
+pub struct DuckDbFsWriter {
```
0ax1: `pub(crate)`

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,307 @@
...
+    handle: FsFileHandle,
+    pos: u64,
+}
+
+impl DuckDbFsWriter {
+    pub unsafe fn create(ctx: cpp::duckdb_vx_client_context, path: &str) -> VortexResult<Self> {
```
0ax1: `pub(crate)`

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -0,0 +1,307 @@
...
+        };
+        urls.push(url);
+    }
+
+    unsafe { cpp::duckdb_vx_uri_list_free(&raw mut list) };
+
+    Ok(urls)
+}
+
+pub unsafe fn duckdb_fs_create_writer(
```
0ax1: `pub(crate)`

0ax1:
> > @iceboundrock - quick question on this. Do you know if it's possible to grab the config back out from DuckDB's filesystems?
> > I don't want to block this PR, don't worry(!), but we do have plans to implement a high performance I/O layer at some point and it would probably make more sense to try and configure our own I/O using duckdb configs, rather than attempt to optimize calling back into DuckDB's I/O layer.
> 
> Could you please elaborate on what `config` refers to? If it’s S3 credentials, HTTP Proxy configuration, etc., ~I don’t believe we can read them from DuckDB. They are stored in [`Secrets Manager`](https://duckdb.org/docs/stable/configuration/secrets_manager) and the `read` API [redacts sensitive information](https://duckdb.org/docs/stable/sql/meta/duckdb_table_functions#duckdb_secrets).~
> 
> We probably can do the same as the httpfs extension does. Code link:
> 
>     * [1](https://github.com/duckdb/duckdb-httpfs/blob/6afe39d5be688b032146baa97e49601b50c9456e/src/httpfs_extension.cpp#L73-L105)
> 
>     * [2](https://github.com/duckdb/duckdb-httpfs/blob/main/src/create_secret_functions.cpp)

Let's handle this in a follow up to limit the scope of this PR.

0ax1 (commented):
PR looks great and CI (including all benchmarks) pass. 

Last bit of fine-tuning and this is ready to land.

vortex-duckdb/cpp/file_system.cpp
iceboundrock: Your suggested version is cleaner, I will fix it.

vortex-duckdb/src/duckdb/client_context.rs
iceboundrock: make sense, I will fix it.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: The write path is driven by `SESSION.write_options().write(writer, array_stream).await`, which processes array chunks sequentially. The blocking write isn't starving other work because the write task is the end of a pipeline. But I agree that in an async context, blocking the executor thread is generally bad practice. **I will fix it.**

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: I will address it.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: I will address it.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: I will address it.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: I will address it.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: I will address it.

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: I will address it.

vortex-duckdb/cpp/file_system.cpp
iceboundrock: I just realized that if we really want to drop the `const_cast`, the call site would need to add `.cast_mut()` to `buffer.read_ptr()`. Is that okay?

```
cpp::duckdb_vx_fs_write(
        self.handle.as_ptr(),
        offset as cpp::idx_t,
        len as cpp::idx_t,
        buffer.read_ptr().cast_mut(),
        &raw mut out_len,
        &raw mut err,
)
```

0ax1: Yes

iceboundrock: addressed

vortex-duckdb/src/duckdb/client_context.rs
iceboundrock: addressed

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed

vortex-duckdb/src/duckdb/file_system.rs
iceboundrock: addressed, but  since `IoBuf` is not bounded by `Send`, so it cannot be moved into a `spawn_blocking` closure. I have to pass the pointer as `usize` (which is Send) and keep `buffer` alive in this async fn until the blocking call completes, which looks **ugly**. Suggestions? @0ax1

0ax1: This requires more design on that trait from our end. For now we can keep it as is.

0ax1 (approved):
Thank you for the contribution! @iceboundrock  🎉

gatesn:
Looks like this is materially worse for S3 performance:

Screenshot 2026-02-17 at 09 19 44
https://bench.vortex.dev/

[edit] looks like the coalescing configs are different from what ObjectStore uses. I think we just use the same values for now until we can properly investigate how best to tune these things.

0ax1:
> Looks like this is materially worse for S3 performance: https://bench.vortex.dev/
> [edit] looks like the coalescing configs are different from what ObjectStore uses. I think we just use the same values for now until we can properly investigate how best to tune these things.

Will follow up on that!

iceboundrock:
> Looks like this is materially worse for S3 performance:
> 
> Screenshot 2026-02-17 at 09 19 44 https://bench.vortex.dev/
> [edit] looks like the coalescing configs are different from what ObjectStore uses. I think we just use the same values for now until we can properly investigate how best to tune these things.

Sorry, would you mind letting me know what are `coalescing configs`?

And I think it may be because the new version needs to install & load `httpfs` extension first. If you agree, I can make a PR to add the commands to the [benchmark script](https://github.com/vortex-data/tpchgen-rs/blob/main/benchmarks/parquet_duckdb.sh).

gatesn:
Oh that's an interesting point. Definitely worth us pre-loading the extension, although it ships with core DuckDB so there wouldn't be any download. 

Coalescing configs determine how to combine small individual requests to S3 into fewer larger requests to avoid the overhead of each request. We have made a PR to match them up to the old ones for one.

The odd thing for me is that the benchmarks didn't regress deterministically, they bounce up and down.

0ax1:
> Sorry, would you mind letting me know what are `coalescing configs`?
> 
> And I think it may be because the new version needs to install & load `httpfs` extension first. If you agree, I can make a PR to add the commands to the [benchmark script](https://github.com/vortex-data/tpchgen-rs/blob/main/benchmarks/parquet_duckdb.sh).

The install and load was addressed here: https://github.com/vortex-data/vortex/pull/6565. And the coalescing was fixed here: https://github.com/vortex-data/vortex/pull/6555. Must be sth else in the mix why the S3 perf regressed.

iceboundrock:
> Oh that's an interesting point. Definitely worth us pre-loading the extension, although it ships with core DuckDB so there wouldn't be any download.
> 
> Coalescing configs determine how to combine small individual requests to S3 into fewer larger requests to avoid the overhead of each request. We have made a PR to match them up to the old ones for one.
> 
> The odd thing for me is that the benchmarks didn't regress deterministically, they bounce up and down.

Thanks @gatesn . 

Based on my experience, even DuckDB doesn’t download the extensions, it still adds **2~4 seconds** to my AWS Lambda (ARM 64-bit and 512MB memory) cold start time, just to install and load the `parquet` and `httpfs` extensions from my Lambda package. To address this issue, I created a [project](https://github.com/Lychee-Technology/duckdb-static) to build the static library of DuckDB with the `parquet` and `httpfs` extensions bundled in.

PR #6202, opened 2026-01-29
chore[cuda]: clean up build & dict decompress float values

joseph-isaacs:


vortex-cuda/build.rs
```diff
@@ -20,36 +20,57 @@ pub mod cuda_kernel_generator;
...
+                .and_then(|n| n.to_str())
+                .is_some_and(|n| n.starts_with("bit_unpack_"));
+
             match path.extension().and_then(|e| e.to_str()) {
                 Some("cuh") => println!("cargo:rerun-if-changed={}", path.display()),
                 Some("cu") => {
-                    println!("cargo:rerun-if-changed={}", path.display());
-                    // Compile .cu files to PTX
-                    nvcc_compile_ptx(&kernels_dir, &path)
+                    // Only watch hand-written .cu files, not generated ones
```
0ax1, line 67: I do wonder whether generated .cu files should live in a separate dir.  Would could have:
- src (handwritten)
- gen(generated)
- ptx(ptx files)

Different question, should we rename cuda_kernel_generator? Right now there's nothing indicating that this is specific to bitpacking.

0ax1 (approved):
Let's land as is for now. But we should consider to iterate on this further.

robert3005:
We should rename it. But even better if we move it to the bitpacking crate

vortex-cuda/kernels/src/bit_unpack_16.cu
```diff
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
```
robert3005: I added them to REUSE.toml so we don't have to append the header here. If we are doing this we should at least remove them from there

PR #6203, opened 2026-01-29
Stop IO event stream once there are no more senders

AdamGS:
Implement a stream that wraps the underlying event channels, that stop once all senders are dropped, which I understand the flow correctly only happens once all read futures and the `FileSegmentSource` are dropped.
When running benchmarks its very easy to trigger that case, in which case the background task will keep processing work that no one consumes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10595,6 +10595,7 @@ dependencies = [
 name = "vortex-file"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "async-trait",
  "bytes",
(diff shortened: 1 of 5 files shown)


robert3005 (approved):

robert3005:
We need this to apply to https://github.com/vortex-data/vortex/blob/a4a2f0d7b6cfe852b91d70885dd45f15455ad11c/vortex-file/src/read/driver.rs#L77 without it I think this pr actually does nothing

PR #6207, opened 2026-01-29
feat[btrblocks]: add a dynamic btr blocks compressor

joseph-isaacs:


vortex-btrblocks/src/compressor/float/mod.rs
```diff
@@ -341,56 +414,38 @@ impl Scheme for DictScheme {
...
-            allowed_cascading - 1,
-            &[integer::DictScheme.code()],
+        let dict = dictionary_encode(stats);
+        let has_all_values_referenced = dict.has_all_values_referenced();
+        let DictArrayParts { codes, values, .. } = dict.into_parts();
+
+        let compressed_codes = compressor.compress_canonical(
+            Canonical::Primitive(codes.to_primitive()),
+            ctx.descend(),
+            Excludes::int_only(&[IntCode::Dict, IntCode::Sequence]),
```
joseph-isaacs, line 430: I think we could remove this? Since Seq can check it might succeed (which we didn't before)

vortex-btrblocks/src/ctx.rs
```diff
@@ -0,0 +1,136 @@
...
+use crate::IntCode;
+use crate::MAX_CASCADE;
+use crate::StringCode;
+
+/// Holds references to exclude lists for each compression code type.
+///
+/// This struct is passed through recursive compression calls to specify
+/// which schemes should be excluded at each level.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Excludes<'a> {
```
a10y, line 16: it might be nice to make this owned and then this is what gets passed into the schemes? needing to build new_excludes is a bit messy

a10y:
approach here looks fine overall

vortex-btrblocks/src/ctx.rs
joseph-isaacs: Wanted to avoid the move/alloc? What would you do instead?

a10y: yea that's fair. maybe we can flup but i think having a stack-allocated set might clean things up here

a10y (approved):
we probably want to make it possible to plugin new int/float/string schemes in the future?

joseph-isaacs:
Yep, should be easier to do that in a follow up

vortex-btrblocks/src/ctx.rs
joseph-isaacs: what do you mean? This allows that?

PR #6208, opened 2026-01-29
fix[cuda]: Fix gpu_scan test

a10y:
- Slice not being applied lazily
- execute_cuda for struct needs to pushdown to fields
- CudaDeviceBuffer applying slicing at item instead of byte granularity
- VarBinView had latent GPU bug leftover from BufferHandles conversion
- BitPackedExecutor using wrong slice indices
- `CudaDeviceBuffer` to type-erase the underlying allocation so it matches`BufferHandle` more closely

### DIFF
--- a/encodings/alp/src/alp/array.rs
+++ b/encodings/alp/src/alp/array.rs
@@ -3,7 +3,6 @@
 
 use std::fmt::Debug;
 use std::hash::Hash;
-use std::ops::Range;
 
 use vortex_array::Array;
 use vortex_array::ArrayBufferVisitor;
@@ -14,7 +13,6 @@ use vortex_array::ArrayRef;
 use vortex_array::Canonical;
 use vortex_array::DeserializeMetadata;
 use vortex_array::ExecutionCtx;
-use vortex_array::IntoArray;
 use vortex_array::Precision;
 use vortex_array::ProstMetadata;
 use vortex_array::SerializeMetadata;
@@ -172,21 +170,6 @@ impl VTable for ALPVTable {
             array.clone(),
             ctx,
         )?))
-    }
-
-    fn slice(array: &Self::Array, range: Range<usize>) -> VortexResult<Option<ArrayRef>> {
-        Ok(Some(
-            ALPArray::new(
-                array.encoded().slice(range.clone())?,
-                array.exponents(),
-                array
-                    .patches()
-                    .map(|p| p.slice(range))
-                    .transpose()?
-                    .flatten(),
-            )
-            .into_array(),
-        ))
     }
 }
 
--- a/encodings/alp/src/alp/array.rs
+++ b/encodings/alp/src/alp/array.rs
@@ -13,9 +13,11 @@ use vortex_array::ArrayRef;
 use vortex_array::Canonical;
 use vortex_array::DeserializeMetadata;
 use vortex_array::ExecutionCtx;
+use vortex_array::IntoArray;
 use vortex_array::Precision;
 use vortex_array::ProstMetadata;
 use vortex_array::SerializeMetadata;
+use vortex_array::arrays::SliceVTable;
 use vortex_array::buffer::BufferHandle;
 use vortex_array::patches::Patches;
 use vortex_array::patches::PatchesMetadata;
@@ -170,6 +172,33 @@ impl VTable for ALPVTable {
             array.clone(),
             ctx,
         )?))
+    }
+
+    fn execute_parent(
+        array: &Self::Array,
+        parent: &ArrayRef,
+        _child_idx: usize,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<Canonical>> {
+        // CPU-only: if parent is SliceArray, perform slicing of the buffer and any patches
+        // Note that this triggers compute (binary searching Patches) which we cannot do when the
+        // buffers live in GPU memory.
+        if let Some(slice_array) = parent.as_opt::<SliceVTable>() {
+            let range = slice_array.slice_range().clone();
+            let sliced_alp = ALPArray::new(
+                array.encoded().slice(range.clone())?,
+                array.exponents(),
+                array
+                    .patches()
+                    .map(|p| p.slice(range))
+                    .transpose()?
+                    .flatten(),
+            )
+            .into_array();
+            return Ok(Some(sliced_alp.execute::<Canonical>(ctx)?));
+        }
+
+        Ok(None)
     }
 }
 
--- a/vortex-array/src/arrays/varbinview/vtable/array.rs
+++ b/vortex-array/src/arrays/varbinview/vtable/array.rs
@@ -4,6 +4,7 @@
 use std::hash::Hash;
 
 use vortex_dtype::DType;
+use vortex_vector::binaryview::BinaryView;
 
 use crate::Precision;
 use crate::arrays::varbinview::VarBinViewArray;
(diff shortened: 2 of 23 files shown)


.github/workflows/ci.yml
joseph-isaacs, line 1: ?

a10y: note the "REVERT THIS"

encodings/alp/src/alp/array.rs
```diff
@@ -174,20 +172,20 @@ impl VTable for ALPVTable {
...
     }
 
-    fn slice(array: &Self::Array, range: Range<usize>) -> VortexResult<Option<ArrayRef>> {
-        Ok(Some(
-            ALPArray::new(
-                array.encoded().slice(range.clone())?,
-                array.exponents(),
-                array
-                    .patches()
-                    .map(|p| p.slice(range))
```
a10y, line 184: slicing patches is eager compute, b/c we need to binary search the patch indices to build the new patches. this doesn't work if you're direct scanning into GPU memory

encodings/alp/src/alp/array.rs
```diff
@@ -173,21 +171,6 @@ impl VTable for ALPVTable {
             ctx,
         )?))
     }
-
-    fn slice(array: &Self::Array, range: Range<usize>) -> VortexResult<Option<ArrayRef>> {
```
joseph-isaacs, line 177: can you add a execute_parent here so we don't CPU regress?

vortex-array/src/arrays/varbinview/vtable/array.rs
```diff
@@ -15,7 +16,7 @@ use crate::vtable::BaseArrayVTable;
 
 impl BaseArrayVTable<VarBinViewVTable> for VarBinViewVTable {
     fn len(array: &VarBinViewArray) -> usize {
-        array.views().len()
+        array.views_handle().len() / size_of::<BinaryView>()
```
joseph-isaacs, line 19: What is a little annoying

vortex-array/src/buffer.rs
```diff
@@ -83,6 +83,8 @@ pub trait DeviceBuffer: 'static + Send + Sync + Debug + DynEq + DynHash {
 
     /// Create a new buffer that references a subrange of this buffer at the given
     /// slice indices.
+    ///
+    /// Note that slice indices are in byte units.
     fn slice(&self, range: Range<usize>) -> Arc<dyn DeviceBuffer>;
```
joseph-isaacs, line 88: can we not make this of T?

vortex-array/src/patches.rs
```diff
@@ -162,34 +161,34 @@ impl Patches {
...
-            "Patch indices {max:?}, offset {offset} are longer than the array length {array_len}"
-        );
-
-        debug_assert!(
-            is_sorted(indices.as_ref())
-                .unwrap_or(Some(false))
-                .unwrap_or(false),
-            "Patch indices must be sorted"
-        );
+        // vortex_ensure!(
```
joseph-isaacs: can we do this on the CPU?

vortex-array/src/buffer.rs
joseph-isaacs: Did this change from the CPU version?

vortex-cuda/src/device_buffer.rs
```diff
@@ -192,33 +254,33 @@ impl<T: DeviceRepr + Send + Sync + 'static> DeviceBuffer for CudaDeviceBuffer<T>
...
         let exponent =
             u8::try_from(min(15, trailing)).vortex_expect("min(15, x) always fits in u8");
         let slice_align = Alignment::from_exponent(exponent);
-        let alignment = Alignment::of::<T>();
 
         assert!(
-            slice_align.is_aligned_to(alignment),
-            "slice must respect minimum alignment byte {}, min {}",
+            slice_align.is_aligned_to(self.allocation.alignment()),
+            "slice must respect minimum alignment {}, min {}",
```
a10y, line 277: i'm actually kind of skeptical that we need this, since the `CudaSlice<T>` is not actually aligned to `T` but always aligned to 256 bytes. And for most cases, `T` is going to be `u8` except for the output of some of the kernels, so this isn't really checking all that much.

But, I kept it for consistency.

vortex-array/src/buffer.rs
a10y: you can't have generics on a dyn trait

vortex-array/src/patches.rs
a10y: possibly, there isn't a straightforward way to check if an ArrayRef is fully on the CPU

0ax1: Joe and me had the same conversation earlier today. We can always do this for the host, and do this for the GPU version (by copying) to host in debug mode.

vortex-array/src/buffer.rs
a10y: BufferHandle has a slice_typed but the `Arc<dyn DeviceBuffer>` needs to use bytes

vortex-array/src/arrays/varbinview/vtable/array.rs
a10y: Yea

vortex-array/src/patches.rs
a10y: we can traverse the array tree and make sure all the bufferhandles `is_on_host`

vortex-cuda/src/device_buffer.rs
joseph-isaacs: Just want this to work for something else different default alignment

vortex-array/src/patches.rs
```diff
@@ -162,34 +161,34 @@ impl Patches {
...
-        vortex_ensure!(
-            max - offset < array_len,
-            "Patch indices {max:?}, offset {offset} are longer than the array length {array_len}"
-        );
-
-        debug_assert!(
-            is_sorted(indices.as_ref())
-                .unwrap_or(Some(false))
-                .unwrap_or(false),
-            "Patch indices must be sorted"
```
joseph-isaacs, line 191: maybe just remove these two

joseph-isaacs: but comment and later add a is_host method to a array? maybe

vortex-array/src/buffer.rs
joseph-isaacs: You can do a ext trait

encodings/alp/src/alp/array.rs
a10y: done

vortex-array/src/patches.rs
a10y: done

vortex-array/src/patches.rs
a10y: which two are you referencing? this and the other comment are on deleted code blocks

vortex-array/src/buffer.rs
a10y: i see

joseph-isaacs: dont need to now

vortex-array/src/patches.rs
joseph-isaacs: all other checks can be done without a copy?

joseph-isaacs (approved):
Please add back as many checks as possible and we can merge

vortex-array/src/buffer.rs
a10y: it's done

PR #6209, opened 2026-01-29
Revert "Stop IO event stream once there are no more senders"

gatesn:
Reverts vortex-data/vortex#6203

I can believe we may in some cases process in-flight requests after the senders have been dropped, but it's not entirely clear to me that this PR actually implements that behavior since the coalescing driver eagerly pulls everything off this event stream, meaning any check would have to be inside the driver's poll_next function.

That said, it's still not obvious to me that the sender being dropped means that there's no one waiting for the result of the I/O request. I can open a file, submit a request, hold onto the request future and drop the file object.


robert3005 (approved):

AdamGS:
Just putting what we've discussed here for other people.

You can drop the file and hold on to the scan or other pieces, but the IoStream stays alive as long as the `SegmentSource` which is part of the layout reader. If we revert this change, that background thread will keep processing requests (anecdotally - ran tpch with one iteration, and in total there were 652 futures in the channel **after** all the work was done, and that's while we still cache layout readers).

There definitely seems to be more work on the `IoRequestStream` side, but I do think this is a step in the right direction.

PR #6212, opened 2026-01-29
Chore: Remove the old filter compute function

connortsui20:
As part of the work to make filters lazy, we can now remove the old standalone `filter` function.

Note that the existing implementation code still have to move out of the kernels (which requires us to copy paste move a lot of code around), but that can be in a separate PR.

There is also still the issue where validity is eagerly filtered (which we don't want but we can't change yet until we make a relatively small change to the way we canonicalize before running through the compressor), and that will come in the next followup PR.

### DIFF
--- a/encodings/pco/src/array.rs
+++ b/encodings/pco/src/array.rs
@@ -26,11 +26,9 @@ use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
 use vortex_array::Precision;
 use vortex_array::ProstMetadata;
-use vortex_array::ToCanonical;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::arrays::PrimitiveVTable;
 use vortex_array::buffer::BufferHandle;
-use vortex_array::compute::filter;
 use vortex_array::serde::ArrayChildren;
 use vortex_array::stats::ArrayStats;
 use vortex_array::stats::StatsSetRef;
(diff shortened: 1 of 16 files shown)


vortex-array/src/lib.rs
```diff
@@ -73,5 +73,5 @@ pub mod flatbuffers {
 // TODO(ngates): canonicalize doesn't currently take a session, therefore we cannot invoke execute
 //  from the new array encodings to support back-compat for legacy encodings. So we hold a session
 //  here...
-static LEGACY_SESSION: LazyLock<VortexSession> =
+pub static LEGACY_SESSION: LazyLock<VortexSession> =
```
connortsui20, line 76: Is this ok?

encodings/sparse/src/lib.rs
gatesn: Take an ExecutionCtx here

encodings/pco/src/array.rs
```diff
@@ -200,7 +200,11 @@ pub(crate) fn number_type_from_dtype(dtype: &DType) -> NumberType {
 
 fn collect_valid(parray: &PrimitiveArray) -> VortexResult<PrimitiveArray> {
```
gatesn: Take an ExecutionCtx

gatesn (changes requested):
I think most of these shouldn't do eager cmpute

PR #6213, opened 2026-01-29
Normalize an Array

gatesn:
This will be used prior to serialization to remove any unwanted arrays from the tree.

Should it actually just be fused with serialization?

### DIFF
--- a/vortex-array/src/normalize.rs
+++ b/vortex-array/src/normalize.rs
@@ -7,7 +7,6 @@ use vortex_error::vortex_bail;
 use crate::Array;
 use crate::ArrayEq;
 use crate::ArrayRef;
-use crate::ArrayVisitor;
 use crate::Canonical;
 use crate::ExecutionCtx;
 use crate::IntoArray;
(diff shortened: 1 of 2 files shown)


vortex-array/src/normalize.rs
```diff
@@ -0,0 +1,209 @@
...
+        self.allowed.find(id).is_some()
+    }
+}
+
+impl dyn Array + '_ {
+    /// Normalize the array according to given options.
+    ///
+    /// This operation performs a recursive traversal of the array. Any non-allowed encoding is
+    /// normalized per the configured operation.
+    pub fn normalize(&self, options: &mut NormalizeOptions) -> VortexResult<ArrayRef> {
```
joseph-isaacs, line 60: When would we use this?

joseph-isaacs:
I am concerned that the is more of a bandaid than a fix. It will make it very easy to compress an array (include a accidental filter) and then undo it just before we serialise. I think we actually want to throw an error instead and stop the compressor adding the unwanted array in the first place

PR #6216, opened 2026-01-30
chore: support for `substrait>=0.27`

blaginin:
https://github.com/substrait-io/substrait-python/releases

> code under substrait.gen.proto is now found under substrait


joseph-isaacs (approved):
noting this include supports for further substrait versions

PR #6219, opened 2026-01-30
fix: skip buffer validation if on device

0ax1:


robert3005 (approved):

vortex-array/src/arrays/dict/array.rs
joseph-isaacs, line 1: this is wrong

encodings/runend/src/array.rs
```diff
@@ -212,6 +212,12 @@ impl RunEndArray {
             return Ok(());
         }
 
+        // Skip host-only validation when buffers are on the GPU.
```
joseph-isaacs, line 215: we should validate anything on the CPU and maybe in debug move to CPU and check?

vortex-array/src/arrays/varbin/array.rs
```diff
@@ -205,17 +205,22 @@ impl VarBinArray {
...
-            .as_::<usize>()
-            .ok_or_else(|| vortex_err!("Last offset must be convertible to usize"))?;
-        vortex_ensure!(
-            last_offset <= bytes.len(),
-            "Last offset {} exceeds bytes length {}",
-            last_offset,
-            bytes.len()
-        );
+        // Skip host-only validation when buffers are on the GPU.
+        let offsets_on_device = offsets.buffer_handles().iter().any(|h| h.is_on_device());
```
joseph-isaacs, line 209: move to CPU and validate on CPU in debug mode?

PR #6222, opened 2026-01-30
performance: Use push_unchecked when building indices for element take in fixedsizedlist take

robert3005:

### DIFF
--- a/vortex-array/src/arrays/fixed_size_list/compute/take.rs
+++ b/vortex-array/src/arrays/fixed_size_list/compute/take.rs
@@ -5,7 +5,6 @@ use vortex_buffer::BitBufferMut;
 use vortex_buffer::BufferMut;
 use vortex_dtype::IntegerPType;
 use vortex_dtype::match_each_integer_ptype;
-use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_error::vortex_panic;
 
@@ -104,8 +103,9 @@ fn take_non_nullable_fsl<I: IntegerPType>(
         // Expand the list into individual element indices.
         for i in list_start..list_end {
             // SAFETY: We've allocated enough space for enough indices for all `new_len` lists (that each consist of `list_size = list_end - list_start` elements), so we know we have enough capacity.
+            #[allow(clippy::unwrap_used)]
             unsafe {
-                elements_indices.push_unchecked(I::from_usize(i).vortex_expect("i < list_end"))
+                elements_indices.push_unchecked(I::from_usize(i).unwrap())
             };
         }
     }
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -106,14 +103,17 @@ fn take_non_nullable_fsl<I: IntegerPType>(
 
         // Expand the list into individual element indices.
         for i in list_start..list_end {
-            elements_indices.append_value(I::from_usize(i).vortex_expect("i < list_end"))
+            unsafe {
```
gatesn, line 107: I don't think you can do this? You're trusting the data in the indices array to not overflow here

gatesn (changes requested):

vortex-array/src/arrays/fixed_size_list/compute/take.rs
robert3005: we have created this array on line 93. I don't understand how we can overflow here

gatesn: No I think I'm wrong... this looks ok?

Needs a detailed SAFETY comment to explain why though.

connortsui20:
This is nice! I'm now convinced that if we migrated all of our builders to vectors we would have huge perf improvements everywhere since it makes these kinds of optimizations so much easier.

vortex-array/src/arrays/fixed_size_list/compute/take.rs
connortsui20:
```
// SAFETY: We've allocated enough space for enough indices for all `new_len` lists (that each consist of `list_size = list_end - list_start` elements), so we know we have enough capacity.
```

robert3005: added the comment

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -106,14 +103,18 @@ fn take_non_nullable_fsl<I: IntegerPType>(
 
         // Expand the list into individual element indices.
         for i in list_start..list_end {
-            elements_indices.append_value(I::from_usize(i).vortex_expect("i < list_end"))
+            // SAFETY: We've allocated enough space for enough indices for all `new_len` lists (that each consist of `list_size = list_end - list_start` elements), so we know we have enough capacity.
+            unsafe {
+                elements_indices.push_unchecked(I::from_usize(i).vortex_expect("i < list_end"))
```
joseph-isaacs: can we remove the vortex_expect this will increase perf

connortsui20: We would need a `I::from_usize_unchecked` or something similar in that case?

robert3005: I made it just unwrap the option. Maybe the complier knows that this shouldn't happen

connortsui20: I dont think that is faster? And it is definitely less readable

connortsui20: Yeah I just checked the internal std code and its exactly the same

robert3005: the numbers look the same

gatesn (approved):

PR #6228, opened 2026-01-30
Perf: Stream conversion from Parquet to Vortex to prevent OOM

connortsui20:
Updates the Parquet to Vortex conversion in `vortex-bench/src/conversions.rs` to stream the writer instead of collecting everything in memory.

Without this, the Clickbench benchmark on single files (where the input Parquet compressed file is 14GB) OOM on an EC2 instance with only 32 (effective 29) GB of memory.

I've confirmed that this change fixes that problem and also makes it a lot faster to convert.

Also renames some functions to be more descriptive.

I think there might be an argument to making these functions available in the `vortex` top-level crate too, this seems to be a generally useful thing to have if you have a bunch of Parquet files.

### DIFF
--- a/benchmarks/compress-bench/src/vortex.rs
+++ b/benchmarks/compress-bench/src/vortex.rs
@@ -17,7 +17,7 @@ use vortex::file::WriteOptionsSessionExt;
 use vortex_bench::Format;
 use vortex_bench::SESSION;
 use vortex_bench::compress::Compressor;
-use vortex_bench::conversions::parquet_to_vortex;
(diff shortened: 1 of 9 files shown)


AdamGS:
part of `vx` does that, IDK how well, but we expose the functionality that way.

AdamGS:
triggered benchmarks to make sure it doesn't change something unexpected about the data, otherwise this looks great.

connortsui20:
cancelled the SQL benchmarks run since I force pushed over it, running them again now

AdamGS (approved):
:shipit:

PR #6229, opened 2026-01-30
perf[runend]: run_end_bool decompress

joseph-isaacs:
speed up runend bool decompress


robert3005 (commented):
Love me some prefilling

encodings/runend/src/decompress_bool.rs
```diff
@@ -0,0 +1,358 @@
...
+            }
+        }
+        BoolArray::new(decoded.freeze(), Validity::from(decoded_validity.freeze()))
+    }
+}
+
+/// Fills bits in range [start, end) to true using byte-level operations.
+/// Assumes the buffer is pre-initialized to all zeros.
+#[inline(always)]
+fn fill_bits_true(slice: &mut [u8], start: usize, end: usize) {
```
robert3005: we already have `append_n` logic in BitBufferMut which is very similar to this logic. This logic is even the same between this function and the other logic. Maybe we have BitBuffer expose some `set_range`, `unset_range`

encodings/runend/src/decompress_bool.rs
```diff
@@ -0,0 +1,469 @@
...
+            decoded_validity.append_n(false, run_len);
+            decoded.append_n(false, run_len);
+        }
+    }
+
+    BoolArray::new(decoded.freeze(), Validity::from(decoded_validity.freeze()))
+}
+
+#[inline(always)]
+fn prefill_buffer(set: bool, length: usize) -> BitBufferMut {
```
robert3005: We have `BitBufferMut::full` which is exactly this function

encodings/runend/PERF_NOTES.md
```diff
@@ -0,0 +1,521 @@
+# Run-End Boolean Decoding Performance Notes
```
robert3005: I think this file is not useful?

encodings/runend/src/compress.rs
```diff
@@ -188,23 +188,7 @@ pub fn runend_decode_primitive(
...
-        runend_decode_typed_bool(
-            trimmed_ends_iter(ends.as_slice::<E>(), offset, length),
-            &values.to_bit_buffer(),
-            validity_mask,
-            values.dtype().nullability(),
-            length,
-        )
-    }))
-}
+pub use crate::decompress_bool::runend_decode_bools;
```
robert3005: I think you don't need this reexport here

encodings/runend/benches/run_end_decode.rs
```diff
@@ -0,0 +1,435 @@
...
+            decoded.append_n(false, end - prev_end);
+        }
+        prev_end = end;
+    }
+
+    BoolArray::new(decoded.freeze(), Validity::from(decoded_validity.freeze()))
+}
+
+#[divan::bench(args = NULLABLE_BOOL_ARGS)]
+fn decode_bool_nullable_develop(bencher: Bencher, args: NullableBoolBenchArgs) {
```
robert3005: what are we benchmarking here?

encodings/runend/benches/run_end_decode.rs
```diff
@@ -0,0 +1,435 @@
...
+        create_nullable_bool_test_data(total_length, avg_run_length, distribution, validity);
+    bencher
+        .with_inputs(|| (ends.clone(), values.clone()))
+        .bench_refs(|(ends, values)| {
+            runend_decode_bools(ends.clone(), values.clone(), 0, total_length)
+        });
+}
+
+/// Baseline using develop branch's append_n approach
+fn decode_bool_nullable_baseline(
```
robert3005: I think we can remove the baseline now?

encodings/runend/src/decompress_bool.rs
robert3005: can we avoid duplication and either expose methods from the buffer crate for setting bit range or just make these methods on bitbuffermut?

robert3005 (changes requested):
I think there's a bit too much llm leftover code here

encodings/runend/PERF_NOTES.md
joseph-isaacs: very diplomatic of you 💩

encodings/runend/src/compress.rs
robert3005: this function should also go

robert3005 (approved):

PR #6231, opened 2026-01-30
feat[cuda]: patches kernel

a10y:
Apply patches in-place for BP and ALP.

Added unit tests, and also added patches to the existing BP/ALP tests to verify it works


vortex-cuda/src/kernel/encodings/bitpacked.rs
```diff
@@ -97,27 +100,46 @@ where
     let thread_count = if bits == 64 { 16 } else { 32 };
     let suffixes: [&str; _] = [&format!("{bit_width}bw"), &format!("{thread_count}t")];
     let cuda_function = ctx.load_function(&format!("bit_unpack_{}", bits), &suffixes)?;
-    let mut launch_builder = ctx.launch_builder(&cuda_function);
 
-    // Build launch args: input, output, f, e, length
-    launch_builder.arg(&input_view);
-    launch_builder.arg(&output_view);
+    {
```
a10y, line 104: i had to move all the launch builder stuff into scope so it wasn't held across the await point for `execute_patches`

vortex-cuda/src/kernel/arrays/dict.rs
```diff
@@ -129,7 +129,7 @@ async fn execute_dict_prim_typed<V: DeviceRepr + NativePType, I: DeviceRepr + Na
     let _cuda_events = crate::launch_cuda_kernel!(
```
joseph-isaacs, line 129: Where did we revert the launch_cuda_kernel ptypes?

vortex-cuda/kernels/src/patches.cu
```diff
@@ -0,0 +1,55 @@
...
+
+// Apply patches to a source array
+template<typename ValueT, typename IndexT>
+__device__ void patches(
+    ValueT *const values,
+    const IndexT *const patchIndices,
+    const ValueT *const patchValues,
+    uint64_t patchesLen
+) {
+    const uint32_t idx = blockIdx.x * blockDim.x + threadIdx.x;
```
joseph-isaacs: we do likely want to have each kernel instances patch more than a single value?

joseph-isaacs (approved):
Would be nice to address the perf, also maybe TODO saying this is a pretty naive approach

vortex-cuda/src/kernel/arrays/dict.rs
a10y: it's always been like this afaict

a10y: the macro anyway

vortex-cuda/kernels/src/patches.cu
a10y: done

PR #6232, opened 2026-01-30
feat[cuda]: SequenceArray

a10y:


vortex-cuda/kernels/src/sequence.cu
```diff
@@ -0,0 +1,45 @@
...
+    const uint64_t worker = blockIdx.x * blockDim.x + threadIdx.x;
+
+    const uint64_t elemStart = MIN(worker * ELEMENTS_PER_THREAD, len);
+    const uint64_t elemEnd = MIN(elemStart + ELEMENTS_PER_THREAD, len);
+
+    for (uint64_t idx = elemStart; idx < elemEnd; idx++) {
+        output[idx] = static_cast<ValueT>(idx) * multiplier + base;
+    }
+}
+
```
joseph-isaacs, line 22: please can we be consistent with the others doing 32 units of work per thread, but set with a constant so its very easy to change moving forward

a10y: I included the config.cuh and use the `ELEMENTS_PER_THREAD` constant it exports. Is there a better approach?

joseph-isaacs: I think that is best for now, until we decide on the size these things in one go

joseph-isaacs (approved):

PR #6241, opened 2026-02-02
feat[array]: validate array encodings on write

joseph-isaacs:
based on https://github.com/vortex-data/vortex/pull/6213


vortex-array/src/normalize.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::Array;
+use crate::ArrayRef;
+use crate::ArrayVisitor;
+use crate::session::ArrayRegistry;
+
+/// Options for normalizing an array.
+pub struct NormalizeOptions<'a> {
+    /// The set of allowed array encodings (in addition to the canonical ones) that are permitted
+    /// in the normalized array.
+    pub allowed: &'a ArrayRegistry,
```
gatesn, line 18: Should this just be Vec<ArrayId>?

gatesn (approved):

PR #6242, opened 2026-02-02
chore: list contains over scalars

blaginin:


vortex-array/src/expr/exprs/list_contains.rs
```diff
@@ -158,6 +166,23 @@ impl VTable for ListContains {
...
+    if list.is_null() || needle.is_null() {
+        return Ok(Scalar::null(DType::Bool(nullability)));
+    }
+
+    let list_scalar = list.as_list();
+    let elements = list_scalar
+        .elements()
+        .ok_or_else(|| vortex_err!("Expected non-null list"))?;
+
+    let contains = elements.iter().any(|elem| elem == needle);
```
joseph-isaacs, line 182: do this with a compare and an any true. maybe keep that you you have for len less than some value?

joseph-isaacs:
this is a perf pr

vortex-array/src/expr/exprs/list_contains.rs
blaginin: i think we'll always falback to https://github.com/apache/arrow-rs/blob/181007df053e9004f7a211f7de086c4bbbd0a9e9/arrow-ord/src/cmp.rs#L395 - don't think it has any performance wins over that implementation

joseph-isaacs: this will use scalar equality not rust primitive type equality and not vectorize

joseph-isaacs (approved):
We have horrible scalars so this is best we can do

PR #6249, opened 2026-02-02
Check read-only cache first

gatesn:
Annoyingly the DashMap is a fragile API that can easily require a write-lock when one isn't required. There's no good clippy lint to forbid here either since the fallback is Entry::or_insert_with


AdamGS (approved):

AdamGS:
🥳

robert3005 (approved):

PR #6253, opened 2026-02-02
feat[cuda]: export arrays to ArrowDeviceArray

a10y:
This PR adds scaffolding for the first part of implementing export of Vortex Arrays directly to Arrow C Data Device Interface arrays, and implements the actual export for StructArray and PrimitiveArray

## Setup

Arrow defines a standard for exchanging data in-process to native code using the C Data Interface, which converts an Arrow array into a struct of pointers.

There is an (experimental, but mostly stable) extension called the C Data Device Interface which includes extra information such as the GPU device type and device ID.

cudf knows how to import data exported this way using the [`from_arrow_device_column` method](https://docs.rapids.ai/api/cudf/latest/libcudf_docs/api_docs/interop_arrow/#_CPPv424from_arrow_device_columnPK11ArrowSchemaPK16ArrowDeviceArrayN3rmm16cuda_stream_viewEN3rmm25device_async_resource_refE) for column_views.

We replicate the `ArrowArray` and `ArrowDeviceArray` structs from the spec.

These are very similar to the existing FFI_ArrowArray and FFI_ArrowSchema structs from the `arrow-data` crate, except that those use arrow Bytes and real pointers and we need to use CUDA device pointers and BufferHandles.

## Testing

Rather than polluting the repo/CI with fetching and building cudf, I made a new repo https://github.com/vortex-data/cudf-test-harness/

This will load a cdylib, call its `export_array()` function, imports it into cudf and calls some simple functions, then frees it by calling the embedded `release` function pointer. This step is hooked up to CI now and only adds ~1m to the CUDA tests jobs.


vortex-cuda/cudf-test/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+        let device_array = block_on(canonical.execute(primitive_array.clone(), &mut cuda_ctx))
+            .map_err(|e| CudfError {
+                message: e.to_string(),
+            })?;
+
+        // Synchronize the CUDA stream to ensure the data is ready
+        cuda_ctx.synchronize_stream().map_err(|e| CudfError {
+            message: e.to_string(),
+        })?;
+
```
a10y: should actually drop this and rely on cudf respecting the `sync_hook` inside of the ArrowDeviceArray

.github/workflows/ci.yml
```diff
@@ -537,6 +537,13 @@ jobs:
             --no-fail-fast \
             --target x86_64-unknown-linux-gnu \
             --verbose
+      - name: Build cudf-test library
+        run: cargo +nightly build --locked -p vortex-cudf-test --target x86_64-unknown-linux-gnu
+      - name: Download and run cudf-test-harness
+        run: |
+          curl -fsSL https://github.com/vortex-data/cudf-test-harness/releases/latest/download/cudf-test-harness-x86_64.tar.gz | tar -xz
+          cd cudf-test-harness-x86_64
```
0ax1, line 561: nit: I basically try to never `cd` in shell scripts, considering the current pwd as state. obs not super important, naturally keep as is if you like.

.github/workflows/ci.yml
```diff
@@ -537,6 +537,13 @@ jobs:
...
+          ./cudf-test-harness check $GITHUB_WORKSPACE/target/x86_64-unknown-linux-gnu/debug/libvortex_cudf_test.so
```
0ax1: Should we leave a comment here to briefly hint at what is happening here with the test harness and why it lives in a separate repo?

vortex-cuda/cudf-test/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+        .with::<VortexMetrics>()
+        .with::<ArraySession>()
+        .with::<LayoutSession>()
+        .with::<ExprSession>()
+        .with::<RuntimeSession>()
+        .with::<CudaSession>()
+});
+
+/// External array
+#[unsafe(no_mangle)]
```
0ax1: nit: Could throw in more docs here. What does external mean here exactly, and is `export_array` part of the arrow device ABI etc?

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -0,0 +1,239 @@
...
+    ctx: &mut CudaExecutionCtx,
+) -> VortexResult<(ArrowArray, SyncEvent)> {
+    let len = array.len();
+    let StructArrayParts {
+        validity, fields, ..
+    } = array.into_parts();
+
+    let null_count = match validity {
+        Validity::NonNullable | Validity::AllValid => 0,
+        Validity::AllInvalid => len,
```
0ax1: Early return all invalid case?

0ax1: E.g we're doing
```
        if matches!(values.validity()?, Validity::AllInvalid) {
            return ConstantArray::new(Scalar::null(values.dtype().clone()), output_len)
                .to_canonical();
        }
```
in the runend case.

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -0,0 +1,239 @@
...
+        Validity::Array(_) => {
+            vortex_bail!("Exporting PrimitiveArray with non-trivial validity not supported yet")
+        }
+    };
+
+    // We need the children to be held across await points.
+    let mut children = Vec::with_capacity(fields.len());
+
+    for field in fields.iter() {
```
0ax1, line 163: How about sth along the lines of:
```
let futures = fields
    .iter()
    .map(|field| {
        let field = field.clone();
        async move {
            let cuda_field = field.execute_cuda(ctx).await?;
            export_canonical(cuda_field, ctx).await
        }
    })
    .collect::<Vec<_>>();

let results = join_all(futures).await;
```
If fields live on different streams, they run in parallel.

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -0,0 +1,239 @@
...
+) -> VortexResult<(ArrowArray, SyncEvent)> {
+    let len = array.len();
+    let PrimitiveArrayParts {
+        buffer, validity, ..
+    } = array.into_parts();
+
+    let buffer = if buffer.is_on_device() {
+        buffer
+    } else {
+        // TODO(aduffy): I don't think this type parameter does anything
```
0ax1: I think we need `T` here to know what type is returned from cuda alloc.
`let mut cuda_slice: CudaSlice<T> = self.device_alloc(host_slice.len())?;`

Should we put a phantom data marker on `BufferHandle` instead?

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -0,0 +1,239 @@
...
+        let private_data_ptr =
+            std::ptr::replace(&raw mut (*array).private_data, std::ptr::null_mut());
+
+        if !private_data_ptr.is_null() {
+            let mut private_data = Box::from_raw(private_data_ptr.cast::<PrivateData>());
+            let children = std::mem::take(&mut private_data.children);
+            for child in children {
+                release_array(child);
+            }
+            drop(private_data);
```
0ax1, line 259: Why the explicit drop at the end of the scope?

vortex-cuda/src/arrow/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    device_id: i64,
+    device_type: DeviceType,
+    sync_event: SyncEvent,
+
+    // unused space reserved for future fields
+    _reserved: [i64; 3],
+}
+
+unsafe impl Send for ArrowDeviceArray {}
+unsafe impl Sync for ArrowDeviceArray {}
```
0ax1: We can drop the `Sync` impl I think.

vortex-cuda/src/arrow/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+use cudarc::driver::sys;
+use cudarc::runtime::sys::cudaEvent_t;
+use vortex_array::Array;
+use vortex_array::ArrayRef;
+use vortex_array::buffer::BufferHandle;
+use vortex_error::VortexResult;
+
+use crate::CudaExecutionCtx;
+
+#[derive(Debug, Copy, Clone)]
```
0ax1, line 33: Should we gen this with `bindgen`?

0ax1 (commented):
Couple of questions / remarks

vortex-cuda/src/arrow/canonical.rs
a10y: are fields going to live on different streams? if so, things might get a little weird with the SyncEvent stuff

vortex-cuda/src/arrow/canonical.rs
a10y: i guess not strictly necessary but i like being explicit here. the private_data owns all of the resources allocated on the rust/cudarc side so dropping it signals that the array is actually freed here

vortex-cuda/src/arrow/mod.rs
a10y: i'd prefer to leave bindgen out of this one, since ArrowArray and ArrowDeviceArray cannot be trivially bindgen'ed since they rely on some cudarc types.

I based these definitions off of the ones in https://github.com/apache/arrow-rs/blob/main/arrow-data/src/ffi.rs#L27, with the addition of device stuff

vortex-cuda/src/arrow/canonical.rs
a10y: This was actually an oversight and I think to start with, I actually think we should avoid handling any nulls whatsoever for now

vortex-cuda/cudf-test/src/lib.rs
a10y: this is all just test code. I can comment to make that clearer

vortex-cuda/src/arrow/canonical.rs
a10y: Specifically: the contract for the returned `ArrowDeviceArray` is that its `sync_event` field, if not NULL, contains a pointer to a `cudaStream_t` which AFAIU can only be linked to one stream.

Similarly the CudaExecutionCtx is linked to only one global stream rn. So assuming that `StructArray::execute_cuda(ctx)` gets called, all of its fields should be getting executed with the same context, and thus the same stream.

vortex-cuda/cudf-test/src/lib.rs
joseph-isaacs, line 1: put this inn vortex-test-e2e-cuda?

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -0,0 +1,239 @@
...
+use crate::executor::CudaArrayExt;
+
+/// An implementation of `ExportDeviceArray` that exports Vortex arrays to `ArrowDeviceArray` by
+/// first decoding the array on the GPU and then converting the canonical type to the nearest
+/// Arrow equivalent.
+#[derive(Debug)]
+pub(crate) struct CanonicalDeviceArrayExport;
+
+#[async_trait]
+impl ExportDeviceArray for CanonicalDeviceArrayExport {
```
joseph-isaacs, line 38: would this logic every differ from export arrow with a different execute function?

vortex-cuda/src/arrow/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+/// A (potentially null) pointer to a `cudaEvent_t`.
+pub type SyncEvent = Option<NonNull<cudaEvent_t>>;
+
+/// The C Device data interface representation of an Arrow array.
+///
+/// This array contains on-device pointers to Arrow array data, along with a synchronization
+/// event that the client must wait on.
+#[repr(C)]
+#[derive(Debug)]
+pub struct ArrowDeviceArray {
```
joseph-isaacs, line 61: can we use a regular arrow array instead of this?

joseph-isaacs:
Could we share logic between this exporter and the arrow export if we could pass around a different executor?

vortex-cuda/src/arrow/canonical.rs
robert3005: You can make a stream wait for another one but that’s a complexity we should figure out how to avoid.

vortex-cuda/src/arrow/mod.rs
a10y: we cannot no

a10y:
I'm not sure what you mean by passing around a different exporter, but no I don't think there's much shared logic between this and just `execute_arrow`

vortex-cuda/cudf-test/src/lib.rs
a10y: i moved this to e2e-cuda per Joe's comment, but added a doc comment at the top of the file

vortex-cuda/cudf-test/src/lib.rs
a10y: done

vortex-cuda/src/arrow/canonical.rs
a10y: I'm not sure I understand the question

vortex-cuda/src/arrow/canonical.rs
0ax1: I think we have at most one stream per file during a scan rn (looking at `vortex-file/src/tests gpu_scan`)?  Where the stream is retrieved from the stream pool and capped at max 4 (`DEFAULT_STREAM_POOL_CAPACITY`). 

> that’s a complexity we should figure out how to avoid.

We should manifest this in writing in code docs somewhere. The scenario where this can bite us is if we want to saturate a GPU with one large file, as there'd be no parallelism on a stream level in that case.

vortex-cuda/src/executor.rs
```diff
@@ -149,17 +150,22 @@ impl CudaExecutionCtx {
     }
 
     /// See `VortexCudaStream::move_to_device`.
-    pub fn move_to_device<T: DeviceRepr + Debug + Send + Sync + 'static>(
```
0ax1, line 152: 👌

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -0,0 +1,332 @@
...
+    // We need the children to be held across await points.
+    let mut children = Vec::with_capacity(fields.len());
+
+    for field in fields.iter() {
+        let cuda_field = field.clone().execute_cuda(ctx).await?;
+        let (arrow_field, _) = export_canonical(cuda_field, ctx).await?;
+        children.push(arrow_field);
+    }
+
+    let mut private_data = PrivateData::new(vec![None], children, ctx)?;
```
0ax1, line 169: If we await for all fields in the lines above to complete, why do we need the additional sync event? This should only be relevant if we want to hand out an arrowdevicearray where canonicalization might not be completed yet, and is still processing (c style) async right?

0ax1 (approved):

a10y:
Alright, it looks like cudf doesn't support importing VarBinView directly right now.

I have an issue filed upstream to track: https://github.com/rapidsai/cudf/issues/21352

I'm going to rip that out of the PR and get it merge-ready.

vortex-cuda/src/arrow/canonical.rs
robert3005: We should hold on to the design that execution is on a single stream and then the scan can spawn multiple executions. For a given slice of the file everything will execute on a given stream

PR #6257, opened 2026-02-02
Execute parent to return ArrayRef

gatesn:
This brings it in line with how powerful compute dispatch was prior to the operators change. The major benefits are:

1) It's obvious in this PR there are places we were canonicalizing for no good reason.
2) The existing compute kernels can be reused as parent execution kernels without changing any code.

### DIFF
--- a/encodings/fsst/src/kernel.rs
+++ b/encodings/fsst/src/kernel.rs
@@ -8,6 +8,7 @@ use vortex_array::IntoArray;
 use vortex_array::arrays::FilterArray;
 use vortex_array::arrays::FilterVTable;
 use vortex_array::arrays::VarBinVTable;
+use vortex_array::compute::FilterKernel;
 use vortex_array::kernel::ExecuteParentKernel;
 use vortex_array::kernel::ParentKernelSet;
 use vortex_array::matchers::Exact;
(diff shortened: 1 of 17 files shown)


encodings/fsst/src/kernel.rs
```diff
@@ -53,143 +35,29 @@ impl ExecuteParentKernel<FSSTVTable> for FSSTFilterKernel {
...
-            Canonical::VarBinView(VarBinViewArray::new_unchecked(
-                views,
-                Arc::from(vec![buffer]),
+        _ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<ArrayRef>> {
+        // TODO(ngates): pass execution context when we update the FilterKernel trait.
+        let filtered_codes =
+            FilterKernel::filter(&VarBinVTable, array.codes(), parent.filter_mask())?
+                .as_::<VarBinVTable>()
+                .clone();
```
robert3005, line 44: This is failing right now. I have a feeling FilterKernel is returning canonical

robert3005: Ignore me, it was the previous filter

robert3005 (approved):
I think this makes a lot more sense

encodings/fsst/src/kernel.rs
```diff
@@ -53,143 +35,29 @@ impl ExecuteParentKernel<FSSTVTable> for FSSTFilterKernel {
...
-
-        // SAFETY: FSST already validates the bytes for binary/UTF-8.
-        let canonical = unsafe {
-            Canonical::VarBinView(VarBinViewArray::new_unchecked(
-                views,
-                Arc::from(vec![buffer]),
+        _ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<ArrayRef>> {
+        // TODO(ngates): pass execution context when we update the FilterKernel trait.
+        let filtered_codes =
```
robert3005, line 41: this is causing fuzzer failures. FilterKernel::filter doesn't handle all the cases and assumes it's invoked by top level filter function

PR #6260, opened 2026-02-02
`select` preserve validity of fields

### LINKED ISSUE #6255: (Non-)nullability should be preserved when `select`ing struct fields
Right now, we have a simplification rule for `select` expressions that turns a `select` of a struct into a `pack` of `get_item`s of field names.

The issue with this is that `get_item` performs a validity intersection: Suppose the top-level struct is nullable, but the fields of the struct are non-nullable. `get_item` will return the intersection of the nullability, which means that a previously non-nullable field becomes nullable.

This means that evaluating `select` over a nullable struct with non-nullable fields will _always_ return a struct with nullable fields, which is not correct. What we want is to essentially "mask out" the fields we do not want without changing the validity of anything.

The first fix is to fix the simplification rule to only rewrite the `select` as a `pack` when the child is already a `pack`. There might be other things that go wrong is the struct layout reader after that change.

CC @gatesn @joseph-isaacs @danking

connortsui20:
Should fix https://github.com/vortex-data/vortex/issues/6255? Not sure.

This change fixes the incorrect simplification rule from `select` to `pack(get_item(...))`. That fix broke a few other things that made the (incorrect) assumption that `select` would always turn into `get_item`s.

### DIFF
--- a/vortex-array/src/expr/analysis/immediate_access.rs
+++ b/vortex-array/src/expr/analysis/immediate_access.rs
@@ -16,24 +16,31 @@ use crate::expr::exprs::select::Select;
 
 pub type FieldAccesses<'a> = Annotations<'a, FieldName>;
 
-/// Creates an [`AnnotationFn`] that annotates each expression node with its "free fields".
+/// Returns the "free fields" for this expression node.
 ///
-/// A "free field" is a top-level field from the root scope that an expression references. This is
-/// useful for column pruning, where we only read the fields that an expression actually needs.
+/// A "free field" is a top-level field from the root scope that this expression references—not
+/// nested fields within those top-level fields. For example, `root().a.b` has free field `{a}`,
+/// not `{b}`, because `a` is the top-level field being accessed from root.
+///
+/// The term "free" is borrowed from PL theory's "free variables"—variables that reference an
+/// outer scope rather than being introduced locally.
+///
+/// This is useful for column pruning, where we only need to read the top-level fields that an
+/// expression actually touches.
 ///
 /// # Annotation Rules
 ///
-/// - **[`GetItem`] on [`Root`]**: Returns `[field_name]`.
+/// - **[`Select`]**: Returns the included field names if the child is [`Root`].
+/// - **[`GetItem`] on [`Root`]**: Returns `[field_name]` if the child is [`Root`].
 /// - **[`Root`]**: Returns all field names from `scope` (conservative over-approximation).
-/// - **[`Select`]**: Returns the included field names.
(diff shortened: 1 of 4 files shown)


vortex-array/src/expr/analysis/immediate_access.rs
```diff
@@ -16,15 +16,37 @@ use crate::expr::exprs::select::Select;
...
+/// - **[`Select`]**: Returns the included field names.
+/// - **Everything else**: Returns empty (annotations aggregate from children automatically).
+///
+/// # Example
+///
+/// Given `scope = {a: {b: .., c: ..}, d: ..}` and `expr = root().a.b + root().d`:
+/// - `root().a` has free fields `{a}`
+/// - `root().d` has free fields `{d}`
+/// - The full expression has free fields `{a, d}` (not `b`, only top-level fields are tracked)
+pub fn make_free_field_annotator(
```
connortsui20, line 44: I can rename this back but this makes more sense to me

vortex-array/src/expr/transform/partition.rs
```diff
@@ -311,60 +311,19 @@ mod tests {
...
-    #[rstest]
-    fn test_expr_partition_many_occurrences_of_field(dtype: DType) {
-        let fields = dtype.as_struct_fields_opt().unwrap();
-
-        let expr = and(
-            get_item("y", get_item("a", root())),
-            select(["a", "b"], root()),
-        );
-        let expr = expr.optimize_recursive(&dtype).unwrap();
-        let partitioned = partition(expr, &dtype, annotate_scope_access(fields)).unwrap();
```
connortsui20, line 330: This test doesn't make sense since we should not be able to `select` over a `partition`.

vortex-layout/src/layouts/struct_/reader.rs
```diff
@@ -712,35 +713,43 @@ mod tests {
...
-            Scalar::primitive(4, Nullability::Nullable)
-        );
-        assert!(
-            result
-                .scalar_at(1)
-                .unwrap()
-                .as_struct()
-                .field_by_idx(0)
-                .unwrap()
-                .is_null(),
```
connortsui20, line 734: This test was wrong since the struct at row 1 should be null.

vortex-array/src/expr/analysis/immediate_access.rs
robert3005, line 44: This should be on the expression vtable

robert3005 (commented):
It should move to the vtable

robert3005 (approved):

connortsui20:
discussed offline, we wont move this to the expression vtable just yet

PR #6265, opened 2026-02-03
perf[vortex-array]: don't materialize const arrays in to_arrow_dictionary

asubiotto:
Executing a Constant array into an arrow dictionary would materialize the full array and then perform the cast. Since we only have one value, we can just cast that value and create the codes array. This results in a 15% performance improvement on our real-world queries.

### DIFF
--- a/vortex-array/src/arrow/executor/dictionary.rs
+++ b/vortex-array/src/arrow/executor/dictionary.rs
@@ -5,7 +5,9 @@ use std::sync::Arc;
 
 use arrow_array::ArrayRef as ArrowArrayRef;
 use arrow_array::DictionaryArray;
+use arrow_array::PrimitiveArray;
 use arrow_array::cast::AsArray;
+use arrow_array::new_null_array;
 use arrow_array::types::*;
 use arrow_schema::DataType;
 use vortex_error::VortexError;
@@ -14,6 +16,9 @@ use vortex_error::vortex_bail;
 
 use crate::ArrayRef;
 use crate::ExecutionCtx;
+use crate::IntoArray;
+use crate::arrays::ConstantArray;
+use crate::arrays::ConstantVTable;
 use crate::arrays::DictArray;
 use crate::arrays::DictArrayParts;
 use crate::arrays::DictVTable;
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrow/executor/dictionary.rs
joseph-isaacs, line 1: Can we add a test that 1 its correct (equal to the expected value) and 2 that the values count of the array dict is 1

vortex-array/src/arrow/executor/dictionary.rs
```diff
@@ -25,10 +30,13 @@ pub(super) fn to_arrow_dictionary(
...
     ctx: &mut ExecutionCtx,
 ) -> VortexResult<ArrowArrayRef> {
-    // Check if we have a Vortex dictionary array
     let array = match array.try_into::<DictVTable>() {
-        Ok(array) => return dict_to_dict(array, codes_type, values_type, ctx),
-        Err(a) => a,
+        Ok(dict) => return dict_to_dict(dict, codes_type, values_type, ctx),
+        Err(array) => array,
+    };
+    let array = match array.try_into::<ConstantVTable>() {
```
asubiotto, line 37: Is there a better way to match on vtables? The other thing I though of doing is an `if/else` chain on `array.is`

joseph-isaacs: Not now, but we should add this.

joseph-isaacs (approved):
happy to merge once you have those tests

asubiotto:
TFTR! Will work on adding some tests

vortex-array/src/arrow/executor/dictionary.rs
```diff
@@ -80,3 +133,110 @@ fn dict_to_dict(
...
+        assert_eq!(arrow.null_count(), 0);
+        let dict = arrow.as_dictionary::<UInt8Type>();
+        assert_eq!(dict.keys().values().as_ref(), &[0u8, 1, 0]);
+        let values = dict.values().as_string::<i32>();
+        assert_eq!(values.value(0), "a");
+        assert_eq!(values.value(1), "b");
+        Ok(())
+    }
+
+    #[test]
```
joseph-isaacs: Mind writing these test in terms of assert_eq(expected, actual)
```
#[test]
fn dict_to_dict_with_null_codes() -> VortexResult<()> {
    let dict = DictArray::try_new(
        PrimitiveArray::from_option_iter(vec![Some(0u8), None, Some(1)]).into_array(),
        VarBinViewArray::from_iter_str(["a", "b"]).into_array(),
    )?;
    let actual = execute(dict.into_array(), &dict_type(DataType::UInt8, DataType::Utf8))?;

    let expected: DictionaryArray<UInt8Type> =
        vec![Some("a"), None, Some("b")].into_iter().collect();

    assert_eq!(expected, actual.as_dictionary::<UInt8Type>().clone());
    Ok(())
}
```

joseph-isaacs: Its less code and tests more stuff

asubiotto: Good point, will also use rstest

PR #6274, opened 2026-02-03
Use liblzma instead of xz2

AdamGS:
xz2 is unmaintained, and this also causes build issues with datafusion


0ax1 (approved):
🏗️

0ax1:
for ref: this is transferring to https://github.com/portable-network-archive/liblzma-rs

PR #6276, opened 2026-02-03
feat: `alternate` for `StructFields`

blaginin:


vortex-dtype/src/struct_.rs
```diff
@@ -537,9 +572,33 @@ mod test {
...
+        {
+          name=utf8,
+          age=i32?,
+          address={
+            street=utf8,
+            city=utf8?
+          }?
+        }
+        ");
+    }
```
joseph-isaacs, line 640: can you show a larger print with a list and struct

joseph-isaacs (approved):

PR #6278, opened 2026-02-03
chore: avoid double-boxing in FileSegmentSource::request

caicancai:
This PR optimizes the implementation of FileSegmentSource::request(): it returns the error branch earlier, removes the redundant internal .boxed(), and ensures that each segment request only generates one BoxFuture allocation. The logic/behavior remains unchanged; only the allocation overhead is reduced.

### DIFF
--- a/vortex-file/src/segments/source.rs
+++ b/vortex-file/src/segments/source.rs
@@ -161,7 +161,7 @@ impl SegmentSource for FileSegmentSource {
         };
 
         // One allocation: we only box the returned SegmentFuture, not the inner ReadFuture.
-        async move { fut.await }.boxed()
+        fut.boxed()
     }
 }
 
--- a/vortex-file/src/segments/source.rs
+++ b/vortex-file/src/segments/source.rs
@@ -123,46 +123,45 @@ impl FileSegmentSource {
 
 impl SegmentSource for FileSegmentSource {
     fn request(&self, id: SegmentId) -> SegmentFuture {
-        // We eagerly create the read future here assuming the behaviour of [`FileRead`], where
+        // We eagerly register the read request here assuming the behaviour of [`FileRead`], where
         // coalescing becomes effective prior to the future being polled.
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

caicancai:
@robert3005 Sorry, I forgot to run `cargo clippy`, I've fixed it now.

vortex-file/src/segments/source.rs
```diff
@@ -123,46 +123,45 @@ impl FileSegmentSource {
...
-            });
-
-            // If we fail to submit the event, we create a future that has failed.
-            if let Err(e) = self.events.unbounded_send(event) {
-                return async move { Err(vortex_err!("Failed to submit read request: {e}")) }
-                    .boxed();
+        let spec = match self.segments.get(*id as usize).cloned() {
+            Some(spec) => spec,
+            None => {
+                return async move { Err(vortex_err!("Missing segment: {}", id)) }.boxed();
```
AdamGS: instead of an async closure, can you return a [`futures::future::ready`](https://docs.rs/futures/latest/futures/future/fn.ready.html)?

vortex-file/src/segments/source.rs
```diff
@@ -123,46 +123,45 @@ impl FileSegmentSource {
...
+            callback: send,
         });
 
-        async move {
-            maybe_fut
-                .ok_or_else(|| vortex_err!("Missing segment: {}", id))?
-                .await
+        // If we fail to submit the event, we create a future that has failed.
+        if let Err(e) = self.events.unbounded_send(event) {
+            return async move { Err(vortex_err!("Failed to submit read request: {e}")) }.boxed();
```
AdamGS: same as above

vortex-file/src/segments/source.rs
caicancai: thanks, done

vortex-file/src/segments/source.rs
caicancai: thanks, done

PR #6280, opened 2026-02-03
Use custom-labels to instrument SQL benchmarks

AdamGS:
This gives us much more debuggable benchmark data based on polar-signals, without requiring too many invasive changes.

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -80,7 +80,7 @@ jobs:
           labels: "branch=${{ github.ref_name }};gh_run_id=${{ github.run_id }};benchmark=${{ matrix.benchmark.id }}"
           project_uuid: "e5d846e1-b54c-46e7-9174-8bf055a3af56"
           profiling_frequency: 150
-          extra_args: "--off-cpu-threshold=0.01"  # Personally tuned by @brancz
+          extra_args: "--off-cpu-threshold=0.03"  # Personally tuned by @brancz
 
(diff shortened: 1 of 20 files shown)


joseph-isaacs (approved):

AdamGS:
@asubiotto if you guys happen to use this feature with DF any a similar way, mind taking a look here? I suspect I'm missing some stuff but I have no idea where from/why.

asubiotto:
> @asubiotto if you guys happen to use this feature with DF any a similar way, mind taking a look here? I suspect I'm missing some stuff but I have no idea where from/why.

@thorfour added this to DF queries on our side, so I believe this should work. Let us know if there are any issues and we'd be happy to look into them.

AdamGS:
I suspect tokio is just spawning some threads in places/ways I didn't account for, I might refactor the benchmarks to have a runtime per-query so I can make sure every thread has the labels set

asubiotto:
Hmm, I think there might be a datafusion API we need to hook into to propagate custom labels across future boundaries. I believe https://github.com/datafusion-contrib/datafusion-tracing had the same issues.

AdamGS:
I've implemented that (its `JoinSetTracer`), I just have general anxiety about missing this sort of stuff

asubiotto:
Me too 😂 This makes me think that maybe vortex needs to do provide a similar api for its scan futures?

AdamGS:
yeah for sure, its been in the back of my mind for a while, I'll try and get to it this week.

PR #6284, opened 2026-02-03
Rename execute to canonicalize

gatesn:
This helps distinguish between the behaviour of execute (return ArrayRef) and canonicalize (return Canonical)

### DIFF
--- a/encodings/alp/src/alp/array.rs
+++ b/encodings/alp/src/alp/array.rs
@@ -166,7 +166,7 @@ impl VTable for ALPVTable {
         Ok(())
     }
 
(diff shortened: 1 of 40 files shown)


robert3005 (approved):
why was this called execute before? I should spend more time reviewing your prs

joseph-isaacs (commented):
thoughts on decompress?

lwwmanning:
Do we still have something called `execute`? Because IMO that's a more legible name than `canonicalize`

joseph-isaacs:
Its called Canonical::execute(Array) or a.execute::<Canoncial>()

PR #6295, opened 2026-02-04
Implement IsSortedKernel and MinMaxKernel for List, ListView, and FixedSizeList

DXist:
This PR relates to issue #4842 .

It generated by "grok-code-fast-1" model without manual code changes.


robert3005:
Please see the comments on the issue, the solution here requires more work. I would expect this pr needs to be reworked quite a lot.

PR #6298, opened 2026-02-04
perf[arrays]: move over all filter kernels to exec/reduce rules

joseph-isaacs:


vortex-array/src/arrays/filter/kernel.rs
```diff
@@ -0,0 +1,118 @@
...
+    ///
+    /// # Preconditions
+    ///
+    /// The mask is guaranteed to have the same length as the array.
+    ///
+    /// Additionally, the mask is guaranteed to be a `Mask::Values` variant (i.e., neither
+    /// `Mask::AllTrue` nor `Mask::AllFalse`).
+    fn filter(
+        array: &Self::Array,
+        mask: &Mask,
```
onursatici, line 48: shall we accept a `&MaskValues` here?

joseph-isaacs: I think we should keep Mask to reduce code change for now. I also wonder if we might later want to have other ways to represent mask other than values

onursatici (approved):

PR #6302, opened 2026-02-04
feat[array]: slice reduce/kernel rules instead of VTable method

joseph-isaacs:


encodings/fastlanes/src/rle/array/mod.rs
```diff
@@ -480,8 +480,13 @@ mod tests {
         let sliced = rle_array.slice(100..200).unwrap();
         assert_eq!(sliced.len(), 100);
 
+        // With SliceKernel, the slice optimization happens during execution/canonicalization.
+        // Convert to canonical form before serialization.
+        let canonical = sliced.to_canonical().unwrap();
+        let executed = canonical.into_array();
```
onursatici: are we now serialising the primitive array in this test, whereas we were serialising the sliced rle array directly before?

vortex-array/src/arrays/slice/rules.rs
onursatici: so this would return None now because we moved every implementation to kernels right? Shall we assert that here you think?

vortex-array/src/compute/filter.rs
```diff
@@ -204,6 +204,20 @@ impl<'a> TryFrom<&InvocationArgs<'a>> for FilterArgs<'a> {
 pub struct FilterKernelRef(pub ArcRef<dyn Kernel>);
 inventory::collect!(FilterKernelRef);
 
+pub trait FilterReduce: VTable {
```
onursatici: This is currently unused right?

encodings/fastlanes/src/bitpacking/vtable/kernels/slice.rs
```diff
@@ -2,40 +2,24 @@
...
-        _child_idx: usize,
+        range: Range<usize>,
         _ctx: &mut ExecutionCtx,
     ) -> VortexResult<Option<ArrayRef>> {
-        // TODO(joe): fix me https://github.com/vortex-data/vortex/pull/5958#discussion_r2696436008
-        // If buffers are on device, we cannot eagerly slice because Patches::slice
-        // requires binary search on the indices which needs host memory for now
-        if !array.is_host() {
-            return Ok(None);
-        }
```
onursatici, line 36: shouldn't we keep this so we have the execute then slice fallback for device arrays instead of panicking?

vortex-array/src/arrays/varbinview/compute/rules.rs
```diff
@@ -0,0 +1,8 @@
+use crate::arrays::SliceReduceAdaptor;
+use crate::arrays::VarBinViewVTable;
+use crate::optimizer::rules::ParentRuleSet;
+
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
```
onursatici: hoist to top of file

onursatici (approved):

PR #6304, opened 2026-02-04
perf[vortex-array]: avoid expensive list view to list in arrow execution

asubiotto:
FilterVTable<ListVTable> produced a ListViewArray that did not have zctl set. This should be trivially zctl but was likely overlooked.

Additionally, after execution into a ListViewArray, the zctl flag can be checked again to use the optimized zctl path.

### DIFF
--- a/vortex-array/src/arrays/list/vtable/kernel/filter.rs
+++ b/vortex-array/src/arrays/list/vtable/kernel/filter.rs
@@ -1,25 +1,24 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use num_traits::Zero;
 use vortex_buffer::Buffer;
 use vortex_buffer::BufferMut;
-use vortex_dtype::Nullability;
 use vortex_dtype::match_each_integer_ptype;
 use vortex_error::VortexResult;
 use vortex_mask::Mask;
 
 use crate::Array;
 use crate::ArrayRef;
+use crate::Canonical;
 use crate::ExecutionCtx;
 use crate::IntoArray;
+use crate::arrays::ConstantArray;
 use crate::arrays::FilterArray;
 use crate::arrays::FilterVTable;
 use crate::arrays::ListArray;
 use crate::arrays::ListVTable;
-use crate::arrays::ListViewArray;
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/list/vtable/kernel/filter.rs
```diff
@@ -92,11 +92,10 @@ impl ExecuteParentKernel<ListVTable> for ListFilterKernel {
...
-            unsafe {
-                ListViewArray::new_unchecked(new_elements, new_offsets, new_sizes, new_validity)
-            }
-            .into_array(),
-        ))
+        Ok(Some(unsafe {
+            ListViewArray::new_unchecked(new_elements, new_offsets, new_sizes, new_validity)
+                .with_zero_copy_to_list(true)
+                .into_array()
+        }))
```
joseph-isaacs, line 87: can this not just return a ListArray?

asubiotto: I'm guessing the intention is to return a canonical representation? Not sure.

gatesn: It was until.... ~yesterday?! 

We realized we need to stay in ArrayRef space instead of pushing Canonical around else we lose a whole bunch of optimizations like this.

joseph-isaacs: it was, now we do not require this.

asubiotto: Modified to return a ListArray, but now I wonder if maybe we should execute to a ListArray rather than ListView in the to_arrow list executor since we can avoid a ListView size buffer allocation?

asubiotto: Nevermind, we can't since ListArray is not canonical. Maybe this is part of what you're mentioning around pushing Canonical around. In any case, I think this PR should be good to go and we can work on other improvements separately.

joseph-isaacs (approved):
Going to run CI benchmark and if they are good we can merge

asubiotto:
Results look good to me. Feel free to merge if you agree

PR #6307, opened 2026-02-04
Array Execution

gatesn:
This PR refines the concept of Array execution to mean "making progress towards being a canonical array".

It enables _all_ of the optimizations we have discussed but have been unable to implement in the current model. For example, `dict { codes: slice(runend), values: primitive }` would currently never leverage the dict-rle kernel. In this new model, the slice array is executed into runend and then the dict-rle kernel becomes applicable.

Follow Ups:
- [ ] Restrict execute functions to a single child execution before having to return. We will do this by passing ExecutionCtx by ownership, meaning it can only be used once.
- [ ] Canonical arrays should error, instead of returning themselves. This will help prevent endless execution loops.
- [ ] Much bigger fix, but move arrays onto the new vtable format so Array::children is cheap. This is a huge portion of execution given we swap each child out and re-run. It would also make with_children much cheaper.


vortex-array/src/executor.rs
```diff
@@ -1,250 +1,204 @@
...
 
-/// Execute the array to canonical form and unwrap as a [`ListViewArray`].
-///
-/// This will panic if the array's dtype is not list.
-impl Executable for ListViewArray {
-    fn execute(array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Self> {
-        match array.try_into::<ListViewVTable>() {
-            Ok(list) => Ok(list),
-            Err(array) => Ok(array.execute::<Canonical>(ctx)?.into_listview()),
+        // 2. reduce (metadata-only rewrites)
```
joseph-isaacs: I actually wonder if should be reduce first? do we have examples either way?

vortex-array/src/executor.rs
```diff
@@ -1,250 +1,204 @@
...
-/// This will panic if the array's dtype is not struct.
-impl Executable for StructArray {
-    fn execute(array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Self> {
-        match array.try_into::<StructVTable>() {
-            Ok(struct_array) => Ok(struct_array),
-            Err(array) => Ok(array.execute::<Canonical>(ctx)?.into_struct()),
-        }
+        // 4. execute (optimized execution)
+        // TODO(ngates): move over to calling Array::execute
+        ctx.log(format_args!("canonicalize {}", array));
```
joseph-isaacs: canonicalize?

vortex-array/src/arrays/varbin/vtable/canonical.rs
joseph-isaacs, line 1: canonical?

vortex-array/src/columnar.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Executing into a [`Columnar`] is implemented by repeatedly executing the array until we
+/// converge on either a constant or canonical.
+///
+/// For safety, we will error when the number of execution iterations reaches 128. We may want this
+/// to be configurable in the future in case of highly complex array trees, but in practice we
+/// don't expect to ever reach this limit.
+impl Executable for Columnar {
+    fn execute(mut array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Self> {
+        for _ in 0..128 {
```
joseph-isaacs: make this a constant or a envvar you can modify if you need?

vortex-array/src/columnar.rs
```diff
@@ -0,0 +1,123 @@
...
+                ctx.log(format_args!("-> constant({})", constant.scalar()));
+                return Ok(Columnar::Constant(constant.clone()));
+            }
+            if let Some(canonical) = array.as_opt::<AnyCanonical>() {
+                ctx.log(format_args!("-> canonical {}", array));
+                return Ok(Columnar::Canonical(canonical.into()));
+            }
+
+            // Otherwise execute the array one step
+            array = array.execute(ctx)?;
```
joseph-isaacs, line 105: would be nice if you could debug_assert the changed

vortex-array/src/executor.rs
```diff
@@ -1,250 +1,203 @@
...
-                }
-            }
-            break;
+    /// Log an execution step at the current depth.
+    ///
+    /// Steps are accumulated and dumped as a single trace on Drop at DEBUG level.
+    /// Individual steps are also logged at TRACE level for real-time following.
+    ///
+    /// Use the [`format_args!`] macro to create the `msg` argument.
+    pub fn log(&mut self, msg: fmt::Arguments<'_>) {
```
AdamGS, line 90: this is very cool, wdyt about making it a macro like `tracing`'s `instrument`? if it can just be at part of the function definition it hides it from the many more callsites.

vortex-array/src/columnar.rs
joseph-isaacs: so it's not possible you think?

joseph-isaacs (approved):

PR #6309, opened 2026-02-04
Fix `Scalar` de/serialization with proto and outline nulls from `ScalarValue`

connortsui20:
This PR is a major refactoring of the `vortex-scalar` crate and all of its dependents.

Currently, we store an opaque `InnerScalarValue` inside a `ScalarValue`, and the `InnerScalarValue` is allowed to be `Null`.

This PR both removes / outlines the `InnerScalarValue::Null` variant, where a nullable scalar is now an `Option<ScalarValue>` (instead of just a `ScalarValue`. This also completely removes `InnerScalarValue` in favor of a public `ScalarValue` enum:

```rust
// Before:
pub struct ScalarValue(pub(crate) InnerScalarValue);

// After:
pub enum ScalarValue {
    // No `Null` variant!

    /// A boolean value.
    Bool(bool),
    /// A primitive numeric value.
    Primitive(PValue),
    /// A decimal value.
    Decimal(DecimalValue),
    /// A UTF-8 encoded string value.
    Utf8(BufferString),
    /// A binary (byte array) value.
    Binary(ByteBuffer),
    /// A list of potentially null scalar values.
    /// TODO?
    // Extension(ExtScalarRef), // ?
}
```

(**IMPORTANT CHANGE**) Additionally, all `Scalar`s are verified to be sound on construction by checking that the `DType` of the `Scalar` `is_compatible` with the `Option<&ScalarValue>` that is passed.

The stricter construction changes then mean that we need to change how deserialization of scalars work. The protobuf format is not exactly 1-1 with our `Scalar`s (notably, it only supports 64-bit integers). This means that we might write 8-bit integers and the round trip returns a 64-bit integer. So this PR also changes some APIs to allow us to pass a `DType` to the statistics deserializers. TBD on if this needs to happen in more places (`FileStatistics`?).

For reviewers: try to look over all of the diffs since a large majority is **not** just renaming variables, they are semantic changes that I am not super confidant is right.

## Breaks

Breaks the old `file_stats` method on `VortexFile` to return `FileStatistics` instead of the array of `StatsSet`. We needed to do this in order to correctly read statistics from DataFusion, specifically in the case of schema evolution.

## TODO

Some benchmarks are failing, and also I still need to review everything myself to make sure everything is correct. I also want to add more tests in certain places where I'm very scared things are wrong.


vortex-scalar/src/convert/primitive.rs
```diff
@@ -0,0 +1,138 @@
...
+use vortex_dtype::half::f16;
+use vortex_error::VortexError;
+use vortex_error::vortex_err;
+
+use crate::PValue;
+use crate::PrimitiveScalar;
+use crate::Scalar;
+use crate::ScalarValue;
+
+// TODO(connor): Who actually needs this???
```
gatesn: Delete it

vortex-scalar/src/lib.rs
```diff
@@ -39,5 +41,5 @@ pub use scalar_value::*;
 pub use struct_::*;
 pub use utf8::*;
 
-#[cfg(test)]
-mod tests;
+// #[cfg(test)]
```
gatesn: Bring back?

vortex-scalar/src/scalar.rs
```diff
@@ -3,348 +3,335 @@
...
-/// including nullability. When the DType does match, ordering is nulls first (lowest), then the
-/// natural ordering of the scalar value.
-#[derive(Debug, Clone)]
+/// Scalars represent a single value with an associated [`DType`]. The value can be null, in which
+/// case the [`value`][Scalar::value] method returns `None`.
+#[derive(Clone, Debug, Eq)]
 pub struct Scalar {
     /// The type of the scalar.
-    dtype: DType,
+    pub(crate) dtype: DType,
```
gatesn: Would prefer not to expose these struct fields. Allows this crate to unsafely construct scalars, rather than going through checked constructors.

vortex-scalar/src/scalar.rs
```diff
@@ -3,348 +3,335 @@
...
-    /// Returns an error if the cast is not supported or if the value cannot be represented
-    /// in the target type.
-    pub fn cast(&self, target: &DType) -> VortexResult<Self> {
-        if let DType::Extension(ext_dtype) = target {
-            let storage_scalar = self.cast_to_non_extension(ext_dtype.storage_dtype())?;
-            Ok(Scalar::extension_ref(ext_dtype.clone(), storage_scalar))
-        } else {
-            self.cast_to_non_extension(target)
-        }
+    /// # Panics
```
gatesn, line 88: Hmmm, I think in this case we should just have `try_new` and `new_unchecked`. A new that panics always feels like a footgun.

vortex-scalar/src/scalar.rs
```diff
@@ -3,348 +3,335 @@
...
-            value: self.value,
-        }
+    /// Creates a new [`Scalar`] with the given [`DType`] and potentially null [`ScalarValue`]
+    /// without checking compatibility.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the given [`DType`] and [`ScalarValue`] are compatible per the
+    /// rules defined in [`Self::is_compatible`].
+    pub unsafe fn new_unchecked(dtype: DType, value: Option<ScalarValue>) -> Self {
```
gatesn, line 121: Can you switch in cfg!(debug) and call try_new in debug builds?

vortex-scalar/src/scalar.rs
```diff
@@ -3,348 +3,335 @@
...
-            DType::Extension(_) => self.as_extension().storage().nbytes(),
-        }
+    // Constructors for non-null scalars.
+
+    /// Creates a new [`Scalar`] with the given [`DType`] and [`ScalarValue`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the given [`DType`] and [`ScalarValue`] are incompatible.
+    pub fn new_value(dtype: DType, value: ScalarValue) -> Self {
```
gatesn: This feels unnecessary? You're saving the caller from wrapping the value in an option?

gatesn (commented):
A couple of comments on the main Scalar API.

vortex-scalar/src/scalar.rs
connortsui20: It is a lot more common than calling `new`

vortex-scalar/src/convert/primitive.rs
connortsui20: oh it looks like we use this for checked binary arithmetic

vortex-scalar/src/scalar.rs
connortsui20: I will make the new functions `cfg(test)`

vortex-scalar/src/scalar.rs
connortsui20: i (claude) will replace them all

encodings/fastlanes/src/for/vtable/mod.rs
connortsui20, line 1: I believe some changes are needed here

robert3005:
Any idea why this makes FixedSizedList take so much better? From what I see is we managed to inline more aggressively but not sure if that's the reason

encodings/fastlanes/src/for/vtable/mod.rs
```diff
@@ -41,7 +38,9 @@ vtable!(FoR);
 impl VTable for FoRVTable {
     type Array = FoRArray;
 
-    type Metadata = ScalarValueMetadata;
+    // TODO(connor): This should really be a `Scalar` but we need to deprecate `deserialize` for the
+    // `build` method.
```
robert3005: I don't understand this comment

connortsui20:
@robert3005 the perf swing is from #6310, not this one. Afaict this change doesn't affect performance in any meaningful manner

joseph-isaacs:
Can you justify the outlining of null, not fully against it just wanted to see your thoughts

encodings/runend/src/array.rs
```diff
@@ -222,13 +222,13 @@ impl RunEndArray {
 
         // Validate the offset and length are valid for the given ends and values
         if offset != 0 && length != 0 {
-            let first_run_end: usize = ends.scalar_at(0)?.as_ref().try_into()?;
+            let first_run_end: usize = usize::try_from(&ends.scalar_at(0)?)?;
```
robert3005: if you're doing this then you can drop the usize from lhs

encodings/runend/src/array.rs
```diff
@@ -222,13 +222,13 @@ impl RunEndArray {
...
             if first_run_end <= offset {
                 vortex_bail!("First run end {first_run_end} must be bigger than offset {offset}");
             }
         }
 
-        let last_run_end: usize = ends.scalar_at(ends.len() - 1)?.as_ref().try_into()?;
+        let last_run_end: usize = usize::try_from(&ends.scalar_at(ends.len() - 1)?)?;
```
robert3005: same as above

encodings/sparse/src/compute/filter.rs
```diff
@@ -60,7 +60,7 @@ mod tests {
             buffer![2u64, 9, 15].into_array(),
             PrimitiveArray::new(buffer![33_i32, 44, 55], Validity::AllValid).into_array(),
             20,
-            Scalar::null_typed::<i32>(),
+            Scalar::null_native::<i32>(),
```
robert3005, line 63: Names are hard. I did a double take on this name. At this point I don't like either of them. Unless we come up with better name we should keep the old one

encodings/sparse/src/canonical.rs
```diff
@@ -97,12 +97,12 @@ pub(super) fn execute_sparse(array: &SparseArray) -> VortexResult<ArrayRef> {
             })
         }
         dtype @ DType::Utf8(..) => {
-            let fill_value = array.fill_scalar().as_utf8().value();
+            let fill_value = array.fill_scalar().as_utf8().value().cloned();
```
robert3005, line 100: why is this different from the binary path? Shouldn't you call `to_value()` here as well?

vortex-array/src/arrays/constant/compute/take.rs
```diff
@@ -20,13 +20,13 @@ impl TakeReduce for ConstantVTable {
...
             AllOr::All => {
-                let scalar = Scalar::new(
+                let scalar = Scalar::try_new(
                     array
                         .scalar()
                         .dtype()
                         .union_nullability(indices.dtype().nullability()),
-                    array.scalar().value().clone(),
-                );
+                    array.scalar().value().cloned(),
```
robert3005, line 28: `to_value`?

vortex-array/src/arrays/decimal/compute/sum.rs
```diff
@@ -228,17 +232,18 @@ mod tests {
...
+        )
+        .unwrap();
 
         assert_eq!(result, expected);
     }
 
     #[test]
     fn test_sum_overflow_detection() {
-        use vortex_scalar::i256;
+        use vortex_dtype::i256;
```
robert3005, line 245: why is this in vortex_dtype?

vortex-array/src/arrays/dict/compute/fill_null.rs
```diff
@@ -21,37 +22,47 @@ use crate::register_kernel;
...
-            array.codes(),
-            &Scalar::new(
-                array
-                    .codes()
-                    .dtype()
-                    .with_nullability(fill_value.dtype().nullability()),
-                ScalarValue::from(first_fill_value),
-            ),
+            codes,
+            &Scalar::try_new(codes.dtype().as_nonnullable(), Some(fill_scalar_value))?,
```
robert3005, line 62: I think you're missing fill_value nullability here

vortex-array/src/arrays/masked/compute/take.rs
```diff
@@ -21,12 +21,14 @@ impl TakeExecute for MaskedVTable {
...
     ) -> VortexResult<Option<ArrayRef>> {
         let taken_child = if !indices.all_valid()? {
-            // This is safe because we'll mask out these positions in the validity
-            let filled_take = fill_null(
-                indices,
-                &Scalar::default_value(indices.dtype().clone().as_nonnullable()),
-            )?;
-            array.child.take(filled_take)?.to_canonical()?.into_array()
+            // This is safe because we'll mask out these positions in the validity.
+            let fill_scalar = Scalar::zero_value(indices.dtype());
```
robert3005, line 25: is this nullability correct?

vortex-array/src/arrays/struct_/compute/take.rs
```diff
@@ -34,11 +33,15 @@ impl TakeExecute for StructVTable {
             .map(StructArray::into_array)
             .map(Some);
         }
-        // The validity is applied to the struct validity,
-        let inner_indices = &compute::fill_null(
-            indices,
-            &Scalar::default_value(indices.dtype().with_nullability(Nullability::NonNullable)),
-        )?;
+
+        // TODO(connor): This could be bad for cache locality...
```
robert3005, line 37: How is this pr changing it?

vortex-array/src/builders/tests.rs
```diff
@@ -198,7 +198,7 @@ fn test_append_defaults_behavior(#[case] dtype: DType, #[case] should_be_null: b
                 i
             );
             // For non-nullable, it should match the default value.
-            let expected = Scalar::default_value(dtype.clone());
+            let expected = Scalar::default_value(&dtype);
```
robert3005, line 201: If these take reference then `null` should take a refernce as well

vortex-array/src/compute/sum.rs
```diff
@@ -113,16 +113,17 @@ impl ComputeFnVTable for Sum {
...
-            // For floats only use stats if accumulator is zero. otherwise we might have numerical stability issues.
-            match sum_dtype {
+        if let Some(Precision::Exact(sum_scalar)) = array.statistics().get(Stat::Sum) {
+            // For floats only use stats if accumulator is zero. otherwise we might have numerical
+            // stability issues.
+            match &sum_dtype {
                 DType::Primitive(p, _) => {
-                    if p.is_float() && accumulator.is_zero() {
-                        return Ok(sum.into());
+                    if p.is_float() && accumulator.is_null() {
```
robert3005: why change here?

vortex-array/src/compute/sum.rs
```diff
@@ -147,30 +148,29 @@ impl ComputeFnVTable for Sum {
         // Update the statistics with the computed sum. Stored statistic shouldn't include the accumulator.
         match sum_dtype {
             DType::Primitive(p, _) => {
-                if p.is_float() && accumulator.is_zero() {
+                if p.is_float()
+                    && accumulator.value().is_some_and(|value| value.is_zero())
```
robert3005: why this change?

vortex-error/src/lib.rs
```diff
@@ -483,6 +483,13 @@ impl From<std::array::TryFromSliceError> for VortexError {
     }
 }
 
+// impl From<arbitrary::Error> for VortexError {
```
robert3005: does this stay in or out?

vortex-datafusion/src/persistent/format.rs
```diff
@@ -367,81 +365,86 @@ impl FileFormat for VortexFormat {
...
+                column_statistics.push(ColumnStatistics {
+                    null_count: null_count.to_df(),
+                    min_value: min.to_df(),
+                    max_value: max.to_df(),
+                    sum_value: Precision::Absent,
+                    distinct_count: stats_set
+                        .get_as::<bool>(Stat::IsConstant, &DType::Bool(Nullability::NonNullable))
+                        .and_then(|is_constant| is_constant.as_exact().map(|_| Precision::Exact(1)))
+                        .unwrap_or(Precision::Absent),
+                    // TODO(connor): Is this correct?
```
robert3005, line 449: So this should be uncompressed size which should map to our UncompressedSizeInBytes

vortex-python/src/scalar/into_py.rs
```diff
@@ -64,8 +64,19 @@ impl<'py> IntoPyObject<'py> for PyVortex<&'_ Scalar> {
                     Some(value) => decimal_value_to_py(py, decimal_type.scale(), value),
                 }
             }
-            DType::Utf8(_) => self.0.as_utf8().value().map(PyVortex).into_pyobject(py),
-            DType::Binary(_) => self.0.as_binary().value().map(PyVortex).into_pyobject(py),
+            DType::Utf8(_) => self
+                .0
+                .as_utf8()
+                .value()
+                .cloned()
```
robert3005, line 71: to_value?

vortex-scalar/src/convert/decimal.rs
```diff
@@ -0,0 +1,132 @@
...
+use vortex_error::VortexError;
+use vortex_error::vortex_err;
+
+use crate::DecimalScalar;
+use crate::DecimalValue;
+use crate::Scalar;
+use crate::ScalarValue;
+
+// TODO(connor): Figure out if we can remove this.
+impl From<DecimalScalar<'_>> for Scalar {
```
robert3005, line 16: delete?

vortex-scalar/src/convert/from_scalar.rs
```diff
@@ -0,0 +1,261 @@
...
+use crate::BinaryScalar;
+use crate::BoolScalar;
+use crate::DecimalScalar;
+use crate::ListScalar;
+use crate::PrimitiveScalar;
+use crate::Scalar;
+use crate::StructScalar;
+use crate::Utf8Scalar;
+
+// TODO(connor): REMOVE ALL OF THESE!!! Should just use the downcasting methods on `Scalar`.
```
robert3005: These are nice if you need to parameterise something that takes a scalar

vortex-scalar/src/typed_view/decimal/tests.rs
```diff
@@ -734,7 +734,7 @@ fn test_decimal_cast_u64_boundary() {
     // Note: The cast logic checks the float value against U64::MAX
     let decimal = Scalar::decimal(
         DecimalValue::I128(i128::MAX), // Much larger than U64::MAX
-        DecimalDType::new(38, 0),
```
robert3005, line 737: was there a bug here?

vortex-scalar/src/typed_view/binary.rs
```diff
@@ -85,64 +82,45 @@ impl<'a> BinaryScalar<'a> {
...
-                    value: Some(value),
-                }
-            }
-        } else {
-            self
+    /// Construct a [`Scalar`] at most `max_length` in size that's less than or equal to
+    /// ourselves.
+    ///
+    /// Returns a null [`Scalar`] if the value is null.
+    pub fn lower_bound(&self, max_length: usize) -> Scalar {
```
robert3005, line 118: I was trying to avoid always creating new scalar here. It's likely that majority of these will be noops, that's why this took ownership and returned it back

vortex-scalar/src/typed_view/list.rs
```diff
@@ -215,7 +219,7 @@ impl Scalar {
     ) -> Self {
         let element_dtype = element_dtype.into();
 
-        let children: Arc<[ScalarValue]> = children
```
robert3005, line 218: Why don't we like `Arc<[...]>`?

vortex-scalar/src/typed_view/utf8.rs
```diff
@@ -147,72 +143,46 @@ impl<'a> Utf8Scalar<'a> {
...
-        if let Some(value) = self.value {
-            if value.len() > max_length {
-                let utf8_split_pos = (max_length.saturating_sub(3)..=max_length)
-                    .rfind(|p| value.is_char_boundary(*p))
-                    .vortex_expect("Failed to find utf8 character boundary");
-
-                let sliced = value.inner().slice(..utf8_split_pos);
-                drop(value);
+    /// Returns `None` if the value is null or if constructing a greater value would overflow.
+    pub fn upper_bound(&self, max_length: usize) -> Option<Scalar> {
```
robert3005, line 93: again, can we avoid doing anything if self.len() < max_length

robert3005 (commented):
Definitely didn't review everything. Most things are small inconsistencies.

connortsui20:
Most importantly, I believe it will help with how the new extension scalars will work (that is the next PR). CC @gatesn 

Personally, I think this is just a more expected API for people dealing with nulls: `Scalar` is the nullable and typed version, `ScalarValue` is the non-nullable deserialized version. When I first started working on Vortex, this was one of the most confusing things.

vortex-scalar/src/typed_view/decimal/tests.rs
connortsui20: yes unfortunately

vortex-scalar/src/convert/from_scalar.rs
connortsui20: We don't actually use these that often, and the main reason is that this logic should be on the downcasting `as_` methods on `Scalar`, not here.

If we think that we really need these, then the logic should be inverted regardless.

connortsui20: oh it looks like we need it for the Python bindings, so Ill keep it but have the logic contained on the methods instead.

encodings/sparse/src/compute/filter.rs
connortsui20: @gatesn prefers `null_native`, I have no preference but apparently this method shouldn't even exist...

encodings/sparse/src/canonical.rs
connortsui20: I have now got rid of `to_value` and `.value()` which returns a reference instead of implicitly cloning is the only method on both the binary and utf8 type

vortex-array/src/arrays/constant/compute/take.rs
connortsui20: see above

vortex-array/src/arrays/decimal/compute/sum.rs
connortsui20: The type was a re-export from `vortex-dtype` (I am pretty sure that was by accident, if not that needed to be documented), so I removed that re-export and fixed where this happened.

vortex-array/src/arrays/dict/compute/fill_null.rs
connortsui20: This is the definition of the trait:

```rust
pub trait FillNullKernel: VTable {
    /// Kernel for replacing null values in an array with a fill value.
    ///
    /// TODO(connor): Actually enforce these constraints (so that casts do not fail).
    ///
    /// Implementations can assume that:
    /// - The array has at least one null value (not all valid, not all invalid).
    /// - The fill value is non-null.
    /// - For decimal arrays, the fill value can be successfully cast to the array's storage type.
    fn fill_null(&self, array: &Self::Array, fill_value: &Scalar) -> VortexResult<ArrayRef>;
}
```

I think it does not matter? But at the same time I am not so confident that I trust this documentation.

Regardless, I can make a comment justifying this better.

vortex-array/src/arrays/masked/compute/take.rs
connortsui20: see above, I think `fill_null` assumes non-null

vortex-array/src/arrays/struct_/compute/take.rs
connortsui20: Its not changing that, I just left this here because I think it can be improved.

If it wasn't clear, if we do end up calling take on the child array, the underlying code might actually end up taking a bunch of stuff from the start of the array even if the "non-null" values all happen to be at the end of the array. So maybe there could be a better default value in this case? This might not matter at all though...

vortex-array/src/builders/tests.rs
connortsui20: arghhhh... thats a much bigger change. I do agree though that it should take a reference.

I can do this in a followup PR?

vortex-array/src/compute/sum.rs
connortsui20: I think this was claude hallucinating, but the real issue is that `is_zero` doesn't exist anymore on `Scalar` since that doesn't really make any sense when the scalar is null. Though if it returns an optional bool that makes more sense.

I'll add `is_zero` to the decimal scalar, primitive scalar, and also Scalar and ScalarValue.

vortex-array/src/compute/sum.rs
connortsui20: see above, we do not have a `is_zero` that returns a bool, now it returns `Option<bool>`

vortex-error/src/lib.rs
connortsui20: This is something I would have liked to add, but I'll remove it since this PR is way too large

vortex-python/src/scalar/into_py.rs
connortsui20: see above

encodings/fastlanes/src/for/vtable/mod.rs
```diff
@@ -67,14 +66,16 @@ impl VTable for FoRVTable {
         Ok(())
     }
 
+    // TODO(connor): DON'T TOUCH THIS UNLESS YOU KNOW WHAT YOU ARE DOING!!!
```
gatesn: lol you can't leave a comment like this. You can indeed now just store the reference Scalar (not ScalarVlaue) in the Metadata, and serialize only the value or whatever is required for back-compat.

vortex-scalar/src/typed_view/list.rs
connortsui20: I think the idea was that we want each scalar value to have full ownership? But tbh I just took the enum that @gatesn gave me and went ham. Maybe he has a reason?

vortex-datafusion/src/persistent/format.rs
connortsui20: This is calculated above (in the same function, probably should split it up) as such:

```rust
                // Update the total size in bytes.
                let column_size = stats_set
                    .get_as::<usize>(Stat::UncompressedSizeInBytes, &PType::U64.into())
                    .unwrap_or_else(|| stats::Precision::inexact(0_usize));
```

So I think it is correct?

vortex-datafusion/src/persistent/format.rs
connortsui20, line 1: @AdamGS could you take a look at this change to see if it makes sense with the rest of `vortex-datafusion`? Notably I am not sure if the comments are all correct. Not at all familiar with this crate

vortex-layout/src/layouts/zoned/builder.rs
connortsui20, line 1: Honestly I can't remember why I changed this, but I think I had a good reason... This in turn caused me to update that `lower_bound` and `upper_bound` code.

I'll revert this change tomorrow and maybe that will remind me why I changed it in the first place...

vortex-layout/src/layouts/compact.rs
connortsui20, line 1: this accidentally got into the changes, this needs to be removed

AdamGS:
For context - what's the motivation behind this PR? IIRC the current design was intended to hide the variant from external users, which made it easier to reason about the difference between the logical scalar and its physical representation.

encodings/fastlanes/src/for/vtable/mod.rs
```diff
@@ -41,7 +38,9 @@ vtable!(FoR);
 impl VTable for FoRVTable {
     type Array = FoRArray;
 
-    type Metadata = ScalarValueMetadata;
+    // TODO(connor): This should really be a `Scalar` but we need to deprecate `deserialize` for the
+    // `build` method.
+    type Metadata = Vec<u8>;
```
AdamGS: why not keep the proto serialized version here?

vortex-array/src/expr/stats/precision.rs
```diff
@@ -20,26 +21,14 @@ use crate::expr::stats::precision::Precision::Inexact;
 /// This is statistic specific, for max this will be an upper bound. Meaning that the actual max
 /// in an array is guaranteed to be less than or equal to the inexact value, but equal to the exact
 /// value.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
```
AdamGS, line 24: `Copy` seems wrong here

vortex-datafusion/src/persistent/cache.rs
```diff
@@ -58,6 +58,7 @@ fn estimate_footer_size(footer: &Footer) -> usize {
         .statistics()
         .map(|stats| {
             stats
+                .stats()
```
AdamGS: very nitty comment but this is a weird API

vortex-datafusion/src/persistent/format.rs
```diff
@@ -367,81 +365,86 @@ impl FileFormat for VortexFormat {
...
-                    let min = stats_set.get(Stat::Min).and_then(|n| {
-                        n.map(|n| {
-                            Scalar::new(
+            let mut sum_of_column_byte_sizes = stats::Precision::exact(0_usize);
+            let mut column_statistics = Vec::with_capacity(table_schema.fields().len());
+
+            for field in table_schema.fields().iter() {
+                // TODO(connor): Is this actually true?
+                // If the column does not exist, continue. This can happen if the schema has evolved
+                // but we have not yet updated the Vortex file.
```
AdamGS: Yes this is correct, as far as I can tell the table schema might contains fields that don't exist in the specific file.

vortex-datafusion/src/persistent/format.rs
AdamGS: I don't think this statistic is currently used anywhere in the upstream so its hard for me to tell how they intended to use it.
The current rustdoc is:
```rust
    /// Estimated size of this column's data in bytes for the output.
    ///
    /// Note that this is not the same as the total bytes that may be scanned,
    /// processed, etc.
    ///
    /// E.g. we may read 1GB of data from a Parquet file but the Arrow data
    /// the node produces may be 2GB; it's this 2GB that is tracked here.
    ///
    /// Currently this is accurately calculated for primitive types only.
    /// For complex types (like Utf8, List, Struct, etc), this value may be
    /// absent or inexact (e.g. estimated from the size of the data in the source Parquet files).
    ///
    /// This value is automatically scaled when operations like limits or
    /// filters reduce the number of rows (see [`Statistics::with_fetch`]).
```
Personally happy to leave it absent or inexact until we have more clarity about that.

encodings/fastlanes/src/for/vtable/mod.rs
connortsui20: Changed in the latest commits, this is now a `Scalar`.

vortex-array/src/expr/stats/precision.rs
connortsui20: `Copy` is impl whenever `T: Copy` so this should be fine

vortex-datafusion/src/persistent/cache.rs
connortsui20: I think that API itself is fine but the naming is bad here

vortex-scalar/src/typed_view/decimal/dvalue.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+impl DecimalValue {
+    /// Cast `self` to T using the respective `ToPrimitive` method.
+    /// If the value cannot be represented by `T`, `None` is returned.
+    pub fn cast<T: NativeDecimalType>(&self) -> Option<T> {
+        match_each_decimal_value!(self, |value| { T::from(*value) })
+    }
+
+    /// Returns a reasonable precision and scale as a [`DecimalDType`] for the given
```
AdamGS, line 49: reasonable here == max? I think we have consts for those somewhere

vortex-scalar/src/typed_view/decimal/dvalue.rs
```diff
@@ -0,0 +1,213 @@
...
+        match self {
+            DecimalValue::I8(v8) => write!(f, "decimal8({v8})"),
+            DecimalValue::I16(v16) => write!(f, "decimal16({v16})"),
+            DecimalValue::I32(v32) => write!(f, "decimal32({v32})"),
+            DecimalValue::I64(v64) => write!(f, "decimal64({v64})"),
+            DecimalValue::I128(v128) => write!(f, "decimal128({v128})"),
+            DecimalValue::I256(v256) => write!(f, "decimal256({v256})"),
+        }
+    }
+}
```
AdamGS, line 229: This seems like a debug implementation

vortex-scalar/src/typed_view/decimal/dvalue.rs
connortsui20: I think I copied those docs from somewhere else? If they are in fact the max then I can document that

vortex-scalar/src/typed_view/primitive/numeric_operator.rs
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! [`NumericOperator`] enum for arithmetic operations on primitive scalars.
+
+use std::fmt;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Binary element-wise operations on two arrays or two scalars.
+pub enum NumericOperator {
```
AdamGS, line 10: We should really avoid this at any cost.

vortex-scalar/src/typed_view/primitive/pvalue.rs
```diff
@@ -136,8 +139,25 @@ macro_rules! as_primitive {
...
+            self,
+            PValue::U8(0)
+                | PValue::U16(0)
+                | PValue::U32(0)
+                | PValue::U64(0)
+                | PValue::I8(0)
+                | PValue::I16(0)
+                | PValue::I32(0)
+                | PValue::I64(0)
+        ) || matches!(self, PValue::F16(f) if f.to_f32() == Some(0.0))
```
AdamGS: should use `is_zero` on the float types, as there is more than one zero.

vortex-scalar/src/typed_view/primitive/numeric_operator.rs
connortsui20: As in we shouldn't use the `checked_numeric` functions?

vortex-scalar/src/typed_view/decimal/dvalue.rs
AdamGS: it vaguely looks like the max, but I might be wrong

AdamGS:
Github's web UI broke while reviewing this PR, I'll pick it up later today/early tomorrow

vortex-scalar/src/typed_view/primitive/numeric_operator.rs
AdamGS: we have a very enum like this elsewhere (at least one), I'll try and grok the wider context here later

vortex-scalar/src/typed_view/decimal/dvalue.rs
connortsui20: it's slightly more concise: but also this is just carried over from the existing code

vortex-datafusion/src/persistent/cache.rs
connortsui20: changed this to `file_statistics.stats_sets()`

joseph-isaacs:
🤞 https://github.com/vortex-data/vortex/actions/runs/21871821358

connortsui20:
looks like a semantic merge conflict with #6388

vortex-scalar/src/typed_view/primitive/pvalue.rs
connortsui20: Is `use num_traits::Zero;` the correct trait to import here?

vortex-scalar/src/typed_view/primitive/numeric_operator.rs
connortsui20: I think any change to this will have to come in a later PR as we want to merge this sooner than later

gatesn (approved):

vortex-scalar/src/typed_view/primitive/pvalue.rs
AdamGS: Potentially, I think the individual types should also just have that function

connortsui20:
yolo

If anyone runs into issues with scalars, please ping me and Ill take a look at it ASAP. There's known inconsistencies in this PR (though I think I ironed out most of the bugs)

vortex-scalar/src/typed_view/primitive/pvalue.rs
connortsui20: I think that was correct (also the rust core lib doesn't seem to have `is_zero`)

vortex-array/src/builders/tests.rs
connortsui20: https://github.com/vortex-data/vortex/issues/6411

vortex-datafusion/src/persistent/format.rs
connortsui20: https://github.com/vortex-data/vortex/issues/6412

vortex-scalar/src/typed_view/primitive/pvalue.rs
AdamGS: yeah after reading the docs this seems reasonable, I guess it depends on the context.

PR #6310, opened 2026-02-05
chore[array]: move take kernel to execute/reduce rules

joseph-isaacs:


vortex-python/src/arrays/mod.rs
```diff
@@ -603,7 +602,7 @@ impl PyArray {
     /// >>> a = vx.array(['a', 'b', 'c', 'd'])
     /// >>> indices = vx.array([0, 2])
     /// >>> a.take(indices).to_arrow_array()
-    /// <pyarrow.lib.StringArray object at ...>
+    /// <pyarrow.lib.StringViewArray object at ...>
```
joseph-isaacs, line 605: sad

gatesn (approved):

vortex-python/src/arrays/mod.rs
joseph-isaacs: I guess we can impl take for varbin to return a varbin but its extra work

robert3005: take on varbin only ever returned varbin because of FSST type. Now that FSST is changed this is strictly faster IIRC

joseph-isaacs: Its only annoying for people that want to export to arrow (stringarray), they need to compact

gatesn: You can pass DataType into to_arrow_array right?

gatesn (dismissed):
Any ideas why the bench regressions?

joseph-isaacs:
which ones?

gatesn (approved):

vortex-array/src/patches.rs
```diff
@@ -1044,9 +1044,11 @@ where
...
             AllOr::Some(buf) => !buf.value(idx_in_take),
         };
-        if include_nulls && is_null {
-            new_sparse_indices.push(idx_in_take as u64);
-            value_indices.push(0);
+        if is_null {
+            if include_nulls {
+                new_sparse_indices.push(idx_in_take as u64);
+                value_indices.push(0);
+            }
```
joseph-isaacs, line 1051: this is a fit in the PR

connortsui20:
There are some pretty wild regressions/improvements from this change, was that expected?

PR #6312, opened 2026-02-05
chore[array]: move over filtering to `{reduce/execute}_parent`

joseph-isaacs:


encodings/alp/src/alp/compute/filter.rs
```diff
@@ -2,38 +2,44 @@
...
-    fn filter(&self, array: &ALPArray, mask: &Mask) -> VortexResult<ArrayRef> {
+    fn filter(
+        array: &ALPArray,
+        mask: &MaskValues,
+        _ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<ArrayRef>> {
         let patches = array
             .patches()
-            .map(|p| p.filter(mask))
+            .map(|p| p.filter(&Mask::from(mask.bit_buffer().clone())))
```
gatesn: This throws away all the caches inside MaskValues. 

type MaskValueRef = Arc<MaskValues>

gatesn (changes requested):

encodings/alp/src/alp/compute/filter.rs
joseph-isaacs: Can do this as a follow up if we want to

joseph-isaacs:
I think we should keep the precondition until we know those masks are opt away

encodings/alp/src/alp/compute/slice.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::ALPArray;
+use crate::ALPVTable;
+
+impl SliceKernel for ALPVTable {
+    fn slice(
+        array: &Self::Array,
+        range: Range<usize>,
+        _ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<ArrayRef>> {
+        let sliced_alp = ALPArray::new(
```
gatesn, line 21: Do we care that this is execution and not reduction?

encodings/fastlanes/src/bitpacking/compute/filter.rs
```diff
@@ -1,121 +1,115 @@
...
-        let ptype = array.ptype().to_unsigned();
-
-        match_each_unsigned_integer_ptype!(ptype, |U| {
-            // Note that the `filter_primitive` function will reinterpret cast the array back to the
-            // correct `PType`, even if it was changed in `to_unsigned` above.
-            Ok(filter_primitive::<U>(array, mask)?.into_array())
-        })
+/// The threshold over which it is faster to fully unpack the entire [`BitPackedArray`] and then
+/// filter the result than to unpack only specific bitpacked values into the output buffer.
+pub const fn unpack_then_filter_threshold(ptype: PType) -> f64 {
```
gatesn, line 31: Why pub? Pub(crate) at most?

gatesn:
For the super-common rules/kernels against builtin arrays like slice and filter, I wonder whether it does make sense to make it super obvious for plugin authors that they should think about implementing these. e.g.

```
impl VTable for FoRArray {
  type SliceKernel = Self;
  type FilterKernel = NotSupported;

  ...
}
```

encodings/alp/src/alp/compute/slice.rs
joseph-isaacs: it is cause of the patches

joseph-isaacs:
I think we could have a Optimized Array trait the forces impls for slice and filter.

joseph-isaacs:
But you still need to attach the rules

gatesn (approved):

PR #6314, opened 2026-02-05
fix: Empty select

gatesn:
We were not rewriting empty select expressions for struct partitioning.


connortsui20 (approved):

connortsui20:
oh whoops auto merge was on, lets just hope everything passes...

PR #6316, opened 2026-02-05
chore: run CUDA unit tests with compute sanitizers

0ax1:


joseph-isaacs (commented):
Whats the runtime of this?

joseph-isaacs (approved):

0ax1:
> Whats the runtime of this?

like 15s

a10y:
I think there's a way to do this with CARGO_RUNNER env var that might be less brittle than grepping the json

PR #6320, opened 2026-02-05
chore[vortex-bench]: add polarsignals benchmark

asubiotto:
Results vs our currently checked out commit:

```
 ┌───────┬─────────────────────────┬─────────────────────────┬────────────────────┐
 │ Query │      PS (current)       │          HEAD           │ Vortex improvement │
 ├───────┼─────────────────────────┼─────────────────────────┼────────────────────┤
 │ Q0    │ 392 ms                  │ 167 ms                  │ 2.3x faster        │
 ├───────┼─────────────────────────┼─────────────────────────┼────────────────────┤
 │ Q1    │ 503 ms                  │ 412 ms                  │ 1.2x faster        │
 ├───────┼─────────────────────────┼─────────────────────────┼────────────────────┤
 │ Q2    │ 13 ms                   │ 12 ms                   │ ~same              │
 └───────┴─────────────────────────┴─────────────────────────┴────────────────────┘
```

A good data point that this benchmark is representative is that the benchmark OOMs without #6304

The benchmark queries are relatively simple to focus on the scan layer. I might add some more queries in another PR.

This benchmark currently generates data in parallel. Another option is to store a pre-generated file somewhere. Let me know what you prefer. On one hand pre-generated is quicker, on the other hand we won't test writer changes (which have caused problems in the past e.g. with listview->list).

A lot of the boilerplate was written by claude, so feel free to point out anything that doesn't make sense since it is probably wrong.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -175,4 +175,11 @@ jobs:
             "remote_storage": "s3://vortex-bench-dev-eu/${{github.ref_name}}/${{github.run_id}}/fineweb/",
             "targets": "datafusion:parquet,datafusion:vortex,datafusion:vortex-compact,duckdb:parquet,duckdb:vortex,duckdb:vortex-compact"
           },
+          {
+            "id": "polarsignals",
+            "subcommand": "polarsignals",
+            "name": "PolarSignals Profiling",
+            "targets": "datafusion:vortex",
+            "scale_factor": "1"
+          },
         ]
--- a/vortex-bench/polarsignals.sql
+++ b/vortex-bench/polarsignals.sql
@@ -1,0 +1,106 @@
+-- Q0: Narrow time range with filters.
+SELECT locations, value, period, time_nanos, labels.comm FROM stacktraces
+ WHERE time_nanos >= arrow_cast(0, 'Timestamp(Nanosecond, Some("UTC"))')
+   AND time_nanos <= arrow_cast(50000000000, 'Timestamp(Nanosecond, Some("UTC"))')
+   AND producer = 'parca_agent'
+   AND sample_type = 'samples'
+   AND sample_unit = 'count'
+   AND period_type = 'cpu'
+   AND period_unit = 'nanoseconds'
+   AND temporality = 'delta';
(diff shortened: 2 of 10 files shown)


asubiotto:
It doesn't seem like the polarsignals benchmark was picked up even if it is in the matrix. Anything I'm missing?

.github/workflows/bench.yml
joseph-isaacs, line 1: can you add this to sql-benchmarks.yml

asubiotto:
Ah, I think I need to add it to some other places.

.github/workflows/bench.yml
asubiotto: That was it, thanks!

joseph-isaacs:
How does the output look relative to what you expect? Also if we break an optimisation you care about do you know if this benchmark regresses?

vortex-bench/polarsignals.sql
joseph-isaacs, line 1: do we want a wide read too?

asubiotto:
> How does the output look relative to what you expect?

Good! In practice most of our overhead is in object storage (which I think will be hopefully much improved after better scan planning coming out of the GPU work, see #4993) so as long as the CPU-bound work completes in a reasonable amount of time I'm happy. The issue is when the CPU bound work is broken for some reason (list/listview).

> Also if we break an optimisation you care about do you know if this benchmark regresses?

Yes. As I mentioned, this benchmark OOMed without #6304. The output schema is also specified in terms of dictionaries to stress the to_arrow_dict conversion which should catch the other regressions I can think of. Another big one was when listview was introduced as the canonical encoding which this benchmark would've also caught. I'm happy with this coverage and won't hesitate to add more things to the benchmark as we learn more.

vortex-bench/polarsignals.sql
asubiotto: Maybe a little wider, but a select * isn't that interesting to us. I'll add some more queries.

asubiotto: Done!

joseph-isaacs (approved):
thanks again 🥇

PR #6322, opened 2026-02-05
gpu compatible write strategy, move compact strategy to use btrblocks with zstd and pco

onursatici:


vortex-file/src/strategy.rs
```diff
@@ -106,6 +120,47 @@ pub static ALLOWED_ENCODINGS: LazyLock<ArrayRegistry> = LazyLock::new(|| {
     registry
 });
 
+/// Static registry of array encodings compatible with CUDA kernel execution.
+///
+/// This includes all canonical encodings needed to represent nested array trees, plus
+/// compressed encodings that have CUDA kernel implementations in `vortex-cuda`.
+#[cfg(feature = "zstd")]
+pub static GPU_ALLOWED_ENCODINGS: LazyLock<ArrayRegistry> = LazyLock::new(|| {
```
joseph-isaacs: what is this for?

vortex-file/src/strategy_gpu_zstd.rs
```diff
@@ -0,0 +1,100 @@
...
+        match canonical {
+            // Use nvcomp-compatible zstd (without dictionary) for string/binary leaves.
+            Canonical::VarBinView(vbv) => {
+                let zstd = ZstdArray::from_var_bin_view_without_dict(
+                    &vbv,
+                    GPU_ZSTD_LEVEL,
+                    GPU_ZSTD_VALUES_PER_PAGE,
+                )?
+                .into_array();
+                if zstd.nbytes() < vbv.nbytes() {
```
joseph-isaacs: is this hard to add as a schema to the other compressor?

vortex-file/src/strategy.rs
```diff
@@ -164,6 +219,31 @@ impl WriteStrategyBuilder {
...
+                StringCode::Fsst,
+                StringCode::Constant,
+                StringCode::Sparse,
+            ])
+            .build();
+
+        self.compressor = Some(Arc::new(GpuCompatibleCompressor::new(btrblocks)));
+        self.allow_encodings = Some((*GPU_ALLOWED_ENCODINGS).clone());
+        self
+    }
```
joseph-isaacs, line 207: this should be done by the caller?

onursatici: it can, but now compact also doesn't use a custom compressor so it is useful to keep these as convenience imo

vortex-btrblocks/src/compressor/integer/mod.rs
```diff
@@ -62,6 +62,8 @@ pub const ALL_INT_SCHEMES: &[&dyn IntegerScheme] = &[
     &RunEndScheme,
     &SequenceScheme,
     &RLE_INTEGER_SCHEME,
+    #[cfg(feature = "pco")]
+    &PcoScheme,
```
joseph-isaacs, line 66: not must not be in the default

joseph-isaacs (changes requested):
make sure teh default strategy doesn't have PCO or zstd

onursatici:
default is filtering out pco and zstd

vortex-btrblocks/src/builder.rs
```diff
@@ -51,9 +51,21 @@ pub struct BtrBlocksCompressorBuilder {
 impl Default for BtrBlocksCompressorBuilder {
     fn default() -> Self {
         Self {
-            int_schemes: ALL_INT_SCHEMES.iter().copied().collect(),
-            float_schemes: ALL_FLOAT_SCHEMES.iter().copied().collect(),
-            string_schemes: ALL_STRING_SCHEMES.iter().copied().collect(),
+            int_schemes: ALL_INT_SCHEMES
+                .iter()
+                .copied()
+                .filter(|s| s.code() != IntCode::Pco)
```
onursatici, line 57: here is currently where we exclude pco and zstd from the default

joseph-isaacs, line 57: shall we have ALL_DEFAULT_SCEHEMS? next to all schemes?

joseph-isaacs: its just people will forget to omit other schemes

vortex-file/src/strategy.rs
joseph-isaacs: its the cfg zstd new (is that the cuda zstd) or  regular

joseph-isaacs (approved):
Add the default schemes and we can merge

joseph-isaacs (changes requested):
need to fix the python test, this shouldn't change

```
File "../vortex-python/python/vortex/io.py", line ?, in default
Failed example:
    os.path.getsize('tiny.vortex')
Expected:
    55116
Got:
    55316
```

joseph-isaacs (approved):

onursatici:
the test failure is because the tiny vortex file in the test doesn't accommodate for the fastlanes 1024 byte padding that would be included. We should be accommodating that on larger samples because we sample with windows that are multiples of 1024, but for small arrays FoR compression estimation would overshoot, causing the compact compressor to pick FoR instead of Pco. That is why the test file is 200 byte larger, it did pick FoR instead of Pco for the zones child, and with the padding FoR ended up being larger that the primitive array so we fell back to primitive

mwlon:
Just noticed this change because of a conflict - I think this design is a bit suboptimal since Pco+Zstd will ~always be the strongest encodings for their respective data types, so here we're 1. adding a small risk that sampling will randomly choose something weaker based a small portion of the data (e.g. I think @onursatici's last comment is exactly this problem) and 2. increasing compression time somewhat.

I think instead of simply adding encodings, we should have different sets depending on the compression strength the user wants.

PR #6323, opened 2026-02-05
chore[array]: remove slice from vtable

joseph-isaacs:


vortex-array/src/array/mod.rs
```diff
@@ -471,8 +471,8 @@ impl<V: VTable> Array for ArrayAdapter<V> {
             return Ok(Canonical::empty(self.dtype()).into_array());
         }
 
-        let sliced = V::slice(&self.0, range.clone())?
-            .unwrap_or_else(|| SliceArray::new(self.to_array(), range).to_array())
+        let sliced = SliceArray::new(self.to_array(), range)
```
gatesn: Shouldn't all the checks above live in SliceArray::try_new? And presumably the checks below are enforced by the array also?

vortex-array/src/arrays/slice/mod.rs
```diff
@@ -54,14 +55,45 @@ pub trait SliceKernel: VTable {
     ) -> VortexResult<Option<ArrayRef>>;
 }
 
-pub fn precondition<V: VTable>(array: &V::Array, range: &Range<usize>) -> Option<ArrayRef> {
+pub fn precondition<V: VTable>(
```
gatesn: Why are these public functions

gatesn (changes requested):

vortex-array/src/array/mod.rs
joseph-isaacs: what about after optimize, this only checks the slice creation, not application

gatesn: All the reduce/execute wrappers enforce invariants

joseph-isaacs: ?

gatesn (approved):

PR #6325, opened 2026-02-05
Update FSST to store components of compressed codes

### LINKED ISSUE #6293: Arrays cannot hold concretely typed children
### What happened?

At least FSSTArray, possibly others, hold concretely typed child arrays. In this case, FSST holds a VarBinArray for its codes.

This makes the with_children function error when swapping the child out for something that isn't a VarBinArray.

Unfortunately, we're a little stuck. We could split this up into offsets / data children, but that's a breaking change to the serde format. For now we may have to just convert the new child array back to VarBinArray, at whatever cost.

A fix would be to deprecate FSST, and implement a reduce function to replace itself with FSST2. But we should really work out how we to support GPU decoding before doing this.

### Steps to reproduce

FSSTArray::with_children

### Environment

Develop

### Additional context

_No response_

gatesn:
Fixes #6293


vortex-array/src/serde.rs
```diff
@@ -269,6 +269,19 @@ pub trait ArrayChildren {
     }
 }
 
+impl ArrayChildren for &[ArrayRef] {
```
a10y, line 272: do we need this?

a10y (approved):

PR #6332, opened 2026-02-06
feat: clamp zero I/O concurrency

caicancai:
This PR adds a small safety guard in FileSegmentSource::open(): if VortexReadAt::concurrency() returns 0, we log a warning and clamp it to 1. This avoids potential stalls or panics from calling buffer_unordered(0) while keeping normal behavior unchanged for valid concurrency values.

### DIFF
--- a/vortex-file/src/segments/source.rs
+++ b/vortex-file/src/segments/source.rs
@@ -17,6 +17,7 @@ use vortex_array::buffer::BufferHandle;
 use vortex_buffer::Alignment;
 use vortex_error::VortexResult;
 use vortex_error::vortex_err;
+use vortex_error::vortex_panic;
 use vortex_io::VortexReadAt;
 use vortex_io::runtime::Handle;
 use vortex_layout::segments::SegmentFuture;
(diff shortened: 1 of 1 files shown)


vortex-file/src/segments/source.rs
```diff
@@ -86,7 +86,14 @@ impl FileSegmentSource {
...
         });
-        let concurrency = reader.concurrency();
+        let raw_concurrency = reader.concurrency();
+        if raw_concurrency == 0 {
+            tracing::warn!(
+                uri = ?reader.uri(),
+                "VortexReadAt::concurrency returned 0; clamping to 1 to avoid stalling I/O"
+            );
+        }
+        let concurrency = raw_concurrency.max(1);
```
joseph-isaacs: lets just panic here

caicancai: Thank you for the review, done.

vortex-file/src/segments/source.rs
```diff
@@ -87,6 +87,12 @@ impl FileSegmentSource {
             config
         });
         let concurrency = reader.concurrency();
+        if concurrency == 0 {
+            panic!(
```
joseph-isaacs: vortex_panic! please

caicancai: I learned something, thank you.

robert3005 (approved):

PR #6337, opened 2026-02-06
Improve visitor performance by reducing `Vec` and String allocations 

AdamGS:
Using `dhat`, identified these two code paths as heavy allocators, running the full tpch benchmark.

This PR fills out many default implementations in `VisitorVTable`, making them much more efficient for many cases, both by reducing recursive function calls and by not formatting names. As far as I can tell - many of the improvements shown by codspeed are completely real here.

1. ~Using the existing iterator when filtering instead of allocating a Vec saves about 2GB of memory allocation (around 1.6% of all memory allocated), which happens in a tiny amount of allocations (0.02%).~ Moving to a separate PR, I want to try and make a bigger change and be able to measure it.
2. The `nth_child` (should just be `child`?) was just another offender that was easy to experiment with, it doesn't allocate a lot of data but it does save ~0.6% of all allocations.

### DIFF
--- a/vortex-compute/src/filter/slice.rs
+++ b/vortex-compute/src/filter/slice.rs
@@ -16,7 +16,6 @@ use vortex_buffer::BitView;
 use vortex_buffer::Buffer;
 use vortex_buffer::BufferMut;
 use vortex_mask::Mask;
-use vortex_mask::MaskIter;
 use vortex_mask::MaskValues;
 
 use crate::filter::Filter;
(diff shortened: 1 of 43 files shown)


vortex-compute/src/filter/slice.rs
```diff
@@ -46,9 +45,16 @@ impl<T: Copy> Filter<MaskValues> for &[T] {
             "Selection mask length must equal the buffer length"
         );
 
-        match mask_values.threshold_iter(FILTER_SLICES_SELECTIVITY_THRESHOLD) {
-            MaskIter::Indices(indices) => self.filter(indices),
-            MaskIter::Slices(slices) => self.filter(slices),
+        if mask_values.density() >= FILTER_SLICES_SELECTIVITY_THRESHOLD {
```
robert3005: what's wrong with the threshold_iter method? Should we change it or remove it?

vortex-compute/src/filter/slice.rs
```diff
@@ -46,9 +45,16 @@ impl<T: Copy> Filter<MaskValues> for &[T] {
...
+            // High density: use slices (contiguous ranges)
+            self.filter(mask_values.slices())
+        } else {
+            // Low density: stream indices directly from bitmap without allocating
```
robert3005: I see, you want to avoid intermediary here

vortex-compute/src/filter/slice.rs
AdamGS: from profiling, seems like we just allocate a ton of `Vec`s there to cache them, but we don't seem to reuse it? or the benefit isn't very big.

Personally would love to just simplify this thing, state makes things more complex and I have no sense of the perf impact.

AdamGS:
moving the filter piece into a different PR so I can try and measure a bigger thing

AdamGS:
The codspeed improvement here is real 🥳, like 20% of those benchmarks was just formatting the names of the chunks.

AdamGS:
@gatesn following our conversation on Friday, are you going to break these APIs and I should just close this PR?

gatesn:
I will take a look. Very keen to take these performance wins, just want to make sure it moves in the direction we were going to avoid extra work.

joseph-isaacs (approved):
This looks good and is very easy to unwind if needed

PR #6342, opened 2026-02-06
DF: Only calculate the size of the metadata cache entry once

AdamGS:
Realized the function is called many times in DF, so it makes sense to store it.


a10y (approved):
nice

robert3005 (approved):

PR #6348, opened 2026-02-06
Public api lockfiles

gatesn:
This should help us work towards reducing our public API surface and track changes.

To re-generate, run `./scripts/public-api.sh`

It's not super fast... so it might be a bit annoying until the API stabilizes.


a10y (approved):
i think cargo-public-api already sorts? so this should be ok

https://github.com/cargo-public-api/cargo-public-api/blob/b230e1971cba1eb89be7648b991ba85f10e0e94d/public-api/src/lib.rs#L103-L115

a10y:
Also a good forcing function to stop automatically making new items `pub`

.github/workflows/ci.yml
```diff
@@ -351,7 +351,7 @@ jobs:
           git diff --name-only
           git diff --exit-code '**/public-api.lock'
 
-  rust-coverage:
+  ust-coverage:
```
a10y, line 354: doh

PR #6353, opened 2026-02-08
chore:optimize next_coalesced() by tracking removed ids in a HashSet

caicancai:
This PR speeds up State::next_coalesced() by avoiding an O(k) linear scan for every candidate when checking whether a request was already marked for removal. Previously, scanning a window of n candidates with k already-marked ids could cost O(n·k) (worst-case O(n²)). We keep the Vec<(offset, id)> for removals, but add a HashSet<RequestId> for O(1) average membership checks, bringing the scan closer to O(n). Coalescing behavior is unchanged.


caicancai:
@joseph-isaacs Sorry, the issue of cargo clippy failing has been fixed.

robert3005 (approved):

PR #6354, opened 2026-02-08
Upgrade Pco to 1.0.1

mwlon:
Bumped version, including minor changes to Pco API. The most important changes are here: https://github.com/pcodec/pcodec/releases/tag/v1.0.0

The most beneficial one to vortex is a big reduction in the overhead per page, especially on decompression, so I'm also reducing the # of values per page. Let me know if you have a benchmark you'd like to run for this (or would rather have these as two separate PRs).


mwlon:
Looks like there's a Python test failing that asserted exact size. I'll fix that later.

robert3005 (approved):

PR #6358, opened 2026-02-09
varbinview on device build skips validation

onursatici:


vortex-array/src/arrays/varbinview/vtable/mod.rs
```diff
@@ -95,15 +113,6 @@ impl VTable for VarBinViewVTable {
             vortex_bail!("Expected {} views, got {}", len, views.len());
         }
```
joseph-isaacs: pretty sure we could do this on the device too since it just a byte len assert?

vortex-array/src/arrays/varbinview/array.rs
```diff
@@ -164,6 +164,13 @@ impl VarBinViewArray {
         dtype: DType,
         validity: Validity,
     ) -> VortexResult<Self> {
+        let views_nbytes = views.len();
+        vortex_ensure!(
+            views_nbytes.is_multiple_of(size_of::<BinaryView>()),
+            "Expected views buffer length ({views_nbytes}) to be a multiple of {}",
+            size_of::<BinaryView>()
```
joseph-isaacs, line 171: nice

vortex-array/src/arrays/varbinview/vtable/mod.rs
```diff
@@ -104,7 +92,35 @@ impl VTable for VarBinViewVTable {
...
+
+        // If any buffer is on device, skip host validation and use try_new_handle.
+        if buffers.iter().any(|b| b.is_on_device()) {
+            return VarBinViewArray::try_new_handle(
+                views_handle.clone(),
+                Arc::from(data_handles.to_vec()),
+                dtype.clone(),
+                validity,
+            );
+        }
```
joseph-isaacs, line 115: can we not always use this path?

onursatici: `try_new` does validation reading buffer contents

joseph-isaacs (approved):

PR #6363, opened 2026-02-09
constant array metadata for small scalars

onursatici:
scalar value would be stored in the metadata for small scalars as well as in the buffer, this would help decoding a constant array in the gpu without forcing the buffer value to host, and gpu kernels can still use the constant value from the buffer


vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -53,35 +60,47 @@ impl VTable for ConstantVTable {
...
-        let sv = ScalarValue::from_protobytes(&buffer)?;
+        // Prefer reading the scalar from inlined metadata to avoid device-to-host copies.
+        let sv = if let Some(ref proto_bytes) = metadata.scalar_value {
+            ScalarValue::from_protobytes(proto_bytes)?
+        } else {
+            if buffers.len() != 1 {
+                vortex_bail!("Expected 1 buffer, got {}", buffers.len());
+            }
+            let buffer = buffers[0].clone().try_to_host_sync()?;
+            ScalarValue::from_protobytes(&buffer)?
```
joseph-isaacs, line 102: so we duplicate them?

onursatici: for small constants yea, to keep forward compat

onursatici: also if not and we need the scalar in the device we would need to issue a h2d copy anyway, these values are small enough to just duplicate I think

joseph-isaacs: does this ever actually get triggered?

onursatici: I don't think it does for clickbench or tpch

onursatici: but we do have constant arrays and they can't be read to the gpu without an extra host roundtrip, so constantarray::build is definitely triggered on a clickbench scan

joseph-isaacs: can we find out what scalar this is

joseph-isaacs (approved):

PR #6368, opened 2026-02-09
chore[array]: inline take compute function

joseph-isaacs:
follow up remove the from tests


robert3005 (approved):
I will make a pr to fix the benchmarks so they actually do work

PR #6371, opened 2026-02-09
feat[bench]: add more random access

joseph-isaacs:


vortex-bench/src/datasets/feature_vectors.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// Dimensionality of each feature vector.
+const VECTOR_DIM: i32 = 1024;
+
+/// Batch size for data generation.
+const BATCH_SIZE: usize = 100_000;
+
+/// Generate a synthetic feature vectors parquet file.
+///
+/// Schema: `id: Int64, embedding: FixedSizeList<Float32, 128>`.
```
onursatici: I think the embedding is 1024 elements, not 128

benchmarks/random-access-bench/src/main.rs
```diff
@@ -1,27 +1,269 @@
...
+                }
+            }
+            indices.sort_unstable();
+            indices
+        }
+        AccessPattern::Uniform => {
+            // Poisson process: exponential inter-arrival times with rate chosen to yield
+            // ~POISSON_EXPECTED_COUNT indices across the dataset.
+            let rate = POISSON_EXPECTED_COUNT as f64 / row_count as f64;
+            // SAFETY: rate is always positive (POISSON_EXPECTED_COUNT > 0, row_count > 0).
```
onursatici, line 93: nit: why `SAFETY` there is nothing unsafe?

vortex-bench/src/random_access/take.rs
```diff
@@ -121,90 +166,88 @@ impl RandomAccessor for ParquetRandomAccessor {
...
-                .map(RowGroupMetaData::num_rows),
+        let sorted_row_group_keys = row_groups.keys().copied().sorted().collect_vec();
+        let row_group_indices = sorted_row_group_keys
+            .iter()
+            .map(|i| row_groups[i].clone())
+            .collect_vec();
+
+        // Re-open the file for reading (Parquet builder consumes the file handle).
+        let file = tokio::fs::File::open(&meta.path).await?;
+        let builder = ParquetRecordBatchStreamBuilder::new_with_options(
```
onursatici: this feels unfair to parquet, we should be caching the metadata on `open` so here we don't do it every time. For vortex and lance we cache stuff with the session on open

benchmarks/random-access-bench/src/main.rs
joseph-isaacs: an unbounded loop?

benchmarks/random-access-bench/src/main.rs
```diff
@@ -362,14 +262,15 @@ async fn benchmark_random_access(
 /// For other datasets, includes dataset and pattern:
 /// `random-access/{dataset}/{pattern}/{format}-tokio-local-disk`.
 fn measurement_name(dataset: &str, pattern: Option<AccessPattern>, format: Format) -> String {
+    let fmt = format.ext();
```
onursatici, line 265: can conflict if we ever run vortex-compact with vortex

onursatici (approved):

PR #6373, opened 2026-02-09
Make projection pushdown in DF an opt-in feature

AdamGS:
The implementation of expression pushdown is missing some features, and it might not work for all use cases and query patterns. This PR makes it possible to disable it so existing code can keep working.

I think this change is sensible following a few bug reports by people testing the tip of `develop`


AdamGS:
heads up @asubiotto

AdamGS:
I need to figure out the issue with the benchmark, we should keep supporting that.

joseph-isaacs (approved):

PR #6377, opened 2026-02-09
chore[fuzzer]: gen issues new

joseph-isaacs:
Add a more reliable fuzzer issue generator.

This uses both scripts + template with a gen-AI summary attached.

The fully gen-AI solution would create bad issues all the time.


AdamGS (dismissed):
break into multiple PRs

AdamGS (approved):

PR #6380, opened 2026-02-09
Add execution target that recursively canonicalises as well as canonical with canonical validity

robert3005:
Should there be an explicit option to execute validity just by one step?

### DIFF
--- a/encodings/runend/benches/run_end_compress.rs
+++ b/encodings/runend/benches/run_end_compress.rs
@@ -97,15 +97,19 @@ fn take_indices(bencher: Bencher, (length, run_step): (usize, usize)) {
         .unwrap()
         .to_array();
 
-    let mut execution_ctx = LEGACY_SESSION.create_execution_ctx();
-
     bencher
-        .with_inputs(|| (&source_array, &runend_array))
-        .bench_refs(|(array, indices)| {
+        .with_inputs(|| {
+            (
+                &source_array,
+                &runend_array,
+                LEGACY_SESSION.create_execution_ctx(),
+            )
+        })
+        .bench_refs(|(array, indices, execution_ctx)| {
             array
                 .take(indices.to_array())
                 .unwrap()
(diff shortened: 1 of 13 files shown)


encodings/runend/benches/run_end_compress.rs
```diff
@@ -96,5 +99,11 @@ fn take_indices(bencher: Bencher, (length, run_step): (usize, usize)) {
 
     bencher
         .with_inputs(|| (&source_array, &runend_array))
-        .bench_refs(|(array, indices)| array.take(indices.to_array()).unwrap());
+        .bench_refs(|(array, indices)| {
+            array
+                .take(indices.to_array())
+                .unwrap()
+                .execute::<RecursiveCanonical>(&mut LEGACY_SESSION.create_execution_ctx())
```
joseph-isaacs: would be better if the ctx was created outside of the bench

vortex-array/src/canonical.rs
```diff
@@ -498,6 +506,242 @@ impl Executable for Canonical {
...
+                    validity,
+                } = st.into_parts();
+                Ok(RecursiveCanonical(Canonical::Struct(unsafe {
+                    StructArray::new_unchecked(
+                        fields
+                            .iter()
+                            .map(|f| {
+                                Ok(f.clone().execute::<RecursiveCanonical>(ctx)?.0.into_array())
+                            })
+                            .collect::<VortexResult<Arc<[_]>>>()?,
```
joseph-isaacs: cna we do this without an alloc?

joseph-isaacs (approved):
Looks good, just one Q

vortex-array/src/canonical.rs
robert3005: sadly no, we could store fields though as `Arc<Box<[ArrayRef]>>`. Otherwise this only works with unsized locals since after unwrap you get `[ArrayRef]` and compiler doesn't know what to do with it

PR #6381, opened 2026-02-09
CLI for measuring execute_cuda encoding perf

a10y:
## Overview of changes

**ergonomics/API focused changes**

* Introduced a new `LaunchStrategy` on the execution context. This by default will launch kernels and not track any timing information, but it is pluggable. For example in benchmarks we replace this with a `TimedLaunchedStrategy` which executes the kernels in blocking mode and logs their execution time.
* Centralized the entrypoint for launching all kernels. They are now forced to be dispatched off of the execution context using the `ctx.launch_kernel()` method, which accepts a closure that is used to populate kernel arguments

A lot of test and benchmark code needed to be updated to use the new launch methods.

~**Fused FOR + BP**~

This has been shelved for a FLUP since this was too big

~* I've updated the BP kernel generator to generate bp as FFOR, i.e. fused bitpacking with FOR. In practice, this is just adding a `const T reference` param. By default the execution for BitPackedArray passes `zero`, but there is a specialization in the `ForArray` execution tree where if it detects one of its descendants is BP, it fuses itself with the bit unpacking~

**GPU tracing tool**

There's a new binary in `vortex-test-e2e-cuda-scan` which takes as input a Vortex file.

It will recompress the file using only GPU-supported encodings, scan it back, and collect timings for how long each column scan took. The results are printed as either pretty text, or as JSON to stdout, which can be piped into duckdb or similar for analysis

Example usage:

```
FLAT_LAYOUT_INLINE_ARRAY_NODE=true RUST_LOG=vortex_cuda=trace,info cargo run --release --bin vortex-test-e2e-cuda-scan -- ./vortex-bench/data/tpch/1.0/vortex-file-compressed/lineitem_0.vortex
```


vortex-cuda/gpu-scan-cli/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+            FloatCode::Uncompressed,
+            FloatCode::Constant,
+            FloatCode::Alp,
+            FloatCode::AlpRd,
+            FloatCode::RunEnd,
+        ])
+        // Don't compress strings, this is b/c we don't have any BtrBlocks encodings that support
+        // strings.
+        .include_string([StringCode::Uncompressed])
+        .build();
```
joseph-isaacs, line 160: I @onursatici added something like this?

onursatici: yea @a10y you should be able to just use this for the strategy: https://github.com/vortex-data/vortex/blob/develop/vortex-file/src/strategy.rs#L184

vortex-cuda/benches/for_cuda.rs
```diff
@@ -36,20 +38,28 @@ const BENCH_ARGS: &[(usize, &str)] = &[(10_000_000, "10M")];
...
 /// Creates a FoR array with the specified type and length.
-fn make_for_array_typed<T>(len: usize) -> FoRArray
+fn make_for_array_typed<T>(len: usize, bp: bool) -> FoRArray
 where
     T: NativePType + From<u8> + Add<Output = T>,
     Scalar: From<T>,
 {
     let reference = <T as From<u8>>::from(REFERENCE_VALUE);
     let data: Vec<T> = (0..len)
-        .map(|i| <T as From<u8>>::from((i % 256) as u8) + reference)
```
a10y, line 46: this was overflowing before?

vortex-cuda/benches/dict_cuda.rs
```diff
@@ -132,26 +90,22 @@ where
...
                 b.iter_custom(|iters| {
-                    let mut cuda_ctx = CudaSession::create_execution_ctx(&VortexSession::empty())
-                        .vortex_expect("failed to create execution context");
+                    let timed = TimedLaunchStrategy::default();
+                    let timer = Arc::clone(timed.get());
 
-                    let mut total_time = Duration::ZERO;
+                    let mut cuda_ctx = CudaSession::create_execution_ctx(&VortexSession::empty())
+                        .vortex_expect("failed to create execution context")
+                        .with_launch_strategy(Arc::new(timed));
```
a10y, line 100: see here: instead of replicating the full launch setup in benchmark code, we can just stub in a launcher that collects timing information across runs

vortex-cuda/src/kernel/mod.rs
```diff
@@ -34,49 +34,54 @@ pub use filter::FilterExecutor;
...
-
-        $(
-            launch_builder.arg(&$arg);
-        )*
-
-        $crate::launch_cuda_kernel_impl(&mut launch_builder, $event_recording, $len)?
-    }};
+/// Implementations can add tracing, async callbacks, or other behavior
+/// around kernel launches.
+pub trait LaunchStrategy: Debug + Send + Sync + 'static {
```
a10y, line 46: this is where LaunchStrategy is defined and impled

vortex-cuda/src/kernel/encodings/runend.rs
```diff
@@ -32,10 +31,10 @@ use crate::CudaDeviceBuffer;
 use crate::executor::CudaArrayExt;
 use crate::executor::CudaExecute;
 use crate::executor::CudaExecutionCtx;
-use crate::launch_cuda_kernel_impl;
 
 /// CUDA executor for run-end encoded arrays.
 #[derive(Debug)]
+#[doc(hidden)]
```
joseph-isaacs: why are we adding these?

vortex-cuda/src/executor.rs
```diff
@@ -67,9 +75,68 @@ impl CudaExecutionCtx {
...
+    where
+        F: FnOnce(&mut LaunchArgs<'a>),
+    {
+        let mut launcher = self.launch_builder(function);
+        build_args(&mut launcher);
+
+        let events = launch_cuda_kernel_impl(&mut launcher, self.strategy.event_flags(), len)?;
+        self.strategy.on_complete(&events, len)?;
+
+        drop(events);
```
joseph-isaacs, line 127: nice

vortex-cuda/gpu-scan-cli/src/main.rs
joseph-isaacs, line 1: this is an odd place for a binary?

a10y: perhaps

vortex-cuda/src/kernel/encodings/runend.rs
a10y: I wanted to make them `pub` for benchmarks, but actually we can just go through execute_cuda instead

vortex-cuda/gpu-scan-cli/src/main.rs
a10y: moved it

vortex-cuda/src/kernel/encodings/runend.rs
a10y: fixed

vortex-cuda/benches/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+impl LaunchStrategy for TimedLaunchStrategy {
+    fn event_flags(&self) -> CUevent_flags {
+        // using blocking_sync to make sure all events flush before we complete.
+        CU_EVENT_BLOCKING_SYNC
+    }
+
+    fn on_complete(&self, events: &CudaKernelEvents, _len: usize) -> VortexResult<()> {
+        // NOTE: as long as the duration < 584 years this cast is safe.
```
0ax1, line 26: wth 😅

vortex-cuda/benches/common/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+use vortex_cuda::LaunchStrategy;
+use vortex_error::VortexResult;
+
+#[derive(Debug, Default)]
+pub struct TimedLaunchStrategy {
+    total_time_ns: Arc<AtomicU64>,
+}
+
+impl TimedLaunchStrategy {
+    pub fn get(&self) -> &Arc<AtomicU64> {
```
0ax1: As we're returning an arc ref here, why not just make the field public?

vortex-cuda/benches/for_cuda.rs
0ax1: yeh, but isn't this well defined?

vortex-cuda/gpu-scan-cli/src/main.rs
```diff
@@ -0,0 +1,193 @@
...
+#[cfg(feature = "cuda")]
+use vortex_cuda::CudaSession;
+#[cfg(feature = "cuda")]
+use vortex_cuda::TracingLaunchStrategy;
+#[cfg(feature = "cuda")]
+use vortex_cuda::VortexCudaStreamPool;
+#[cfg(feature = "cuda")]
+use vortex_cuda::executor::CudaArrayExt;
+
+#[cfg(not(feature = "cuda"))]
```
0ax1: cuda_available or cuda feature?

vortex-cuda/src/kernel/encodings/alp.rs
```diff
@@ -90,20 +92,14 @@ where
...
-        launch_builder.arg(&f);
-        launch_builder.arg(&e);
-        launch_builder.arg(&array_len_u64);
-
-        // Launch kernel
-        let _cuda_events =
-            launch_cuda_kernel_impl(&mut launch_builder, CU_EVENT_DISABLE_TIMING, array_len)?;
-    }
+
+    ctx.launch_kernel(&cuda_function, array_len, |args| {
```
0ax1, line 91: 👌

vortex-cuda/src/kernel/encodings/zstd.rs
```diff
@@ -250,6 +255,49 @@ async fn decode_zstd(array: ZstdArray, ctx: &mut CudaExecutionCtx) -> VortexResu
...
 
+    // NOTE(aduffy): we need to use the explicit tracing/not(tracing) blocks here because we go
+    //  through nvcomp instead of delegating through the LaunchBuilder.
+    //  We should find a way to bridge the two.
+    #[cfg(feature = "tracing")]
+    {
+        let before = stream
+            .record_event(Some(cudarc::driver::sys::CUevent_flags::CU_EVENT_DEFAULT))
+            .map_err(|e| vortex_err!("recording event: {e}"))?;
+        unsafe {
```
0ax1, line 258: Should we have 2 ` #[cfg(feature = "tracing")]` blocks and run `nvcomp_zstd::decompress_async` in the middle unconditionally?

vortex-cuda/src/macros.rs
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#[macro_export]
+macro_rules! warn {
+    ($($tts:tt)*) => {
+        #[cfg(feature = "tracing")]
```
0ax1: Is it really worth wrapping these? What is the cost of having no subscriber and calling into `trace!/debug` etc.?

0ax1: The other part being, that some tracing calls as part of this PR are again wrapped by: ` #[cfg(feature = "tracing")]`.

vortex-cuda/src/kernel/slice/mod.rs
```diff
@@ -19,6 +19,10 @@ pub struct SliceExecutor;
 
 #[async_trait]
 impl CudaExecute for SliceExecutor {
+    #[cfg_attr(
```
0ax1: I guess that's the way to opt into tracing logic?! At least at first glance not the most straightforward chunk to read.

vortex-cuda/src/macros.rs
```diff
@@ -0,0 +1,42 @@
...
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! trace {
+    ($($tts:tt)*) => {
+        #[cfg(feature = "tracing")]
+        {
+            tracing::info!($($tts)*);
```
0ax1: Should be `tracing::trace!`

vortex-cuda/src/stream.rs
```diff
@@ -155,10 +156,11 @@ fn register_stream_callback(stream: &CudaStream) -> VortexResult<kanal::AsyncRec
         let tx = unsafe { Box::from_raw(user_data as *mut Sender<()>) };
 
         // Blocking send as we're in a callback invoked by the CUDA driver.
-        #[expect(clippy::expect_used)]
-        tx.send(())
-            // A send should never fail. Panic otherwise.
-            .expect("CUDA callback receiver dropped unexpectedly");
+        // NOTE: send can fail if the CudaEvent is dropped by the caller, in which case the reeciver
```
0ax1: reeciver => receiver

vortex-cuda/src/kernel/encodings/zstd.rs
```diff
@@ -250,6 +255,49 @@ async fn decode_zstd(array: ZstdArray, ctx: &mut CudaExecutionCtx) -> VortexResu
 
     let stream = ctx.stream();
 
+    // NOTE(aduffy): we need to use the explicit tracing/not(tracing) blocks here because we go
+    //  through nvcomp instead of delegating through the LaunchBuilder.
+    //  We should find a way to bridge the two.
+    #[cfg(feature = "tracing")]
+    {
+        let before = stream
+            .record_event(Some(cudarc::driver::sys::CUevent_flags::CU_EVENT_DEFAULT))
```
0ax1: Should we use `CU_EVENT_DEFAULT` everywhere or `CU_EVENT_BLOCKING_SYNC`? 

Claude states:
```
CU_EVENT_DEFAULT` | ✅ all prior stream ops complete | **Spin-waits** (busy loop, burns CPU) |
| `CU_EVENT_BLOCKING_SYNC` | ✅ all prior stream ops complete | **Blocks** (yields thread, sleeps) |
| `CU_EVENT_DISABLE_TIMING` | ✅ all prior stream ops complete | Spin-waits, but no timing support
```

So the diff between blocking sync and default being busy wait and yielding.

0ax1 (commented):
Couple of questions.

vortex-cuda/benches/for_cuda.rs
a10y: Yea, I just mean the new one is closer to the purpose of the test I think

vortex-cuda/benches/common/mod.rs
a10y: https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_nanos

vortex-cuda/benches/common/mod.rs
a10y: we can, idk this is just benchmark code

vortex-cuda/gpu-scan-cli/src/main.rs
a10y: switched to cuda_available

vortex-cuda/src/macros.rs
a10y: it seems like it just introduces a runtime check for the log level. which is probably not too bad, i was just being conservative. i can make this unconditional

vortex-cuda/src/kernel/slice/mod.rs
a10y: i undid this as parting of making tracing non-optional

vortex-cuda/src/kernel/encodings/zstd.rs
a10y: i've switched this to blocking_sync to match the others

vortex-cuda/src/macros.rs
a10y: done

vortex-cuda/src/kernel/encodings/zstd.rs
a10y: the problem is then `before` is in a separate scope, which means it goes out of scope and can't be used in the second block

vortex-cuda/src/stream.rs
```diff
@@ -155,10 +156,11 @@ fn register_stream_callback(stream: &CudaStream) -> VortexResult<kanal::AsyncRec
         let tx = unsafe { Box::from_raw(user_data as *mut Sender<()>) };
 
         // Blocking send as we're in a callback invoked by the CUDA driver.
-        #[expect(clippy::expect_used)]
-        tx.send(())
-            // A send should never fail. Panic otherwise.
-            .expect("CUDA callback receiver dropped unexpectedly");
+        // NOTE: send can fail if the CudaEvent is dropped by the caller, in which case the receiver
```
0ax1, line 159: Why did we change this? This should never fail I think, as the `send` can only error if the channel does not have sufficient capacity which should never be the case.

a10y: I was getting panics during testing bc the last receiver had been dropped. I forgot which operation this was for but making this not panic fixed it

a10y: https://docs.rs/kanal/latest/kanal/enum.SendError.html

0ax1 (approved):

PR #6391, opened 2026-02-10
Use Scan API

gatesn:
Experiment to use the Scan API from DuckDB and DataFusion integrations.

This PR introduces:
* `vortex::file::MultiFileDataSource` - a data source builder that takes a listing of files, (in the future) unifies schemas, adds support for hive-style partitioning columns, etc, etc. 
* `vortex::scan::api` - improvements to the Scan API as have become apparent while implementing support in two engines. I'd like to implement more engines before we decide on a path to stability.


.github/workflows/sql-benchmarks.yml
```diff
@@ -187,6 +187,7 @@ jobs:
         if: matrix.remote_storage == null || github.event.pull_request.head.repo.fork == true
         shell: bash
         env:
+          VORTEX_USE_SCAN_API: "1"
```
joseph-isaacs: this should be 0 or remove the old one?

vortex-scan/src/layout.rs
```diff
@@ -31,12 +29,46 @@ use crate::api::SplitRef;
 pub struct LayoutReaderDataSource {
     reader: LayoutReaderRef,
     session: VortexSession,
+    split_size: u64,
```
joseph-isaacs: target_split_size

joseph-isaacs (commented):
look reasonable

vortex-scan/src/multi.rs
```diff
@@ -0,0 +1,386 @@
...
+
+                    let mut child_request = s.request.clone();
+                    child_request.limit = s.remaining_limit;
+                    let child_scan = match source.scan(child_request) {
+                        Ok(scan) => scan,
+                        Err(e) => return Some((Err(e), (None, None))),
+                    };
+
+                    current_stream = Some(child_scan.splits());
+                }
```
connortsui20: This can be extracted into a helper method that takes `&mut current_stream`, then you can do:

```rust
impl DataSourceScan for MultiDataSourceScan {
    fn splits(self: Box<Self>) -> SplitStream {
        stream::unfold(
            (*self, None::<SplitStream>),
            |(mut scan, mut current_stream)| async move {
                let result = scan.next_split(&mut current_stream).await?;
                Some((result, (scan, current_stream)))
            },
        )
        .boxed()
    }
}
```

Doesn't seem like you need it to be `Some(*self)`

vortex-datafusion/src/v2/table.rs
```diff
@@ -0,0 +1,276 @@
...
+    fn statistics(&self) -> Option<Statistics> {
+        let row_count_est = self.data_source.row_count_estimate();
+        let num_rows = match row_count_est.upper {
+            Some(upper) if row_count_est.lower == upper => usize::try_from(upper)
+                .map(Precision::Exact)
+                .unwrap_or_default(),
+            _ => Precision::Absent,
+        };
+
+        let byte_size_est = self.data_source.row_count_estimate();
```
connortsui20: This is probably wrong?

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,507 @@
...
+        // FIXME(ngates): this should be adjusted based on filters. See DuckDB for heuristics,
+        //  and in the future, store the selectivity stats in the session.
+        let num_rows = estimate_to_df_precision(&self.data_source.row_count_estimate());
+
+        // FIXME(ngates): byte size should be adjusted for the initial projection...
+        let total_byte_size = estimate_to_df_precision(&self.data_source.byte_size_estimate());
+
+        // FIXME(ngates): are column statistics after the projection expression? Or before?
+        // let column_statistics =
+        //     vec![ColumnStatistics::new_unknown(); self.final_schema.fields().len()];
```
AdamGS: as far as I can tell, after. See `FileScanConfig::partition_statistics`

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,507 @@
...
+        // Check if all expressions can be pushed down. If any cannot, bail out entirely
+        // since DataSource::try_swapping_with_projection replaces the ProjectionExec,
+        // requiring the output schema to match the projection output exactly.
+        for proj_expr in projection {
+            if !convertor.can_be_pushed_down(&proj_expr.expr, input_schema)
+                || has_decimal_binary(&proj_expr.expr, input_schema)
+            {
+                return Ok(None);
+            }
+        }
```
AdamGS: that's the projection splitter or however we ended up calling it.

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,507 @@
...
+        let scan_output_schema = Arc::new(Schema::new(scan_fields));
+
+        let mut this = self.clone();
+        this.final_projection = scan_projection;
+        this.final_schema = scan_output_schema;
+        Ok(Some(Arc::new(this)))
+    }
+
+    // TODO(ngates): do these filters apply to the initial_schema, or the final_schema?
+    fn try_pushdown_filters(
```
AdamGS, line 481: DF say the filter refers to the `DataSource` output schema

vortex-datafusion/src/v2/table.rs
```diff
@@ -0,0 +1,132 @@
...
+        _limit: Option<usize>,
+    ) -> DFResult<Arc<dyn ExecutionPlan>> {
+        let data_source = VortexDataSource::builder(self.data_source.clone(), self.session.clone())
+            .with_arrow_schema(self.arrow_schema.clone())
+            .with_some_projection(projection.cloned())
+            .build()
+            .await
+            .map_err(|e| DataFusionError::External(Box::new(e)))?;
+
+        Ok(DataSourceExec::from_data_source(data_source))
```
AdamGS, line 101: you can pass the limit here with `with_fetch`

vortex-io/src/filesystem/glob.rs
```diff
@@ -0,0 +1,142 @@
...
+use futures::TryStreamExt;
+use futures::stream::BoxStream;
+use vortex_error::VortexResult;
+use vortex_error::vortex_bail;
+use vortex_error::vortex_err;
+
+use crate::filesystem::FileListing;
+use crate::filesystem::FileSystem;
+
+impl dyn FileSystem + '_ {
```
AdamGS, line 14: are we doing this?

vortex-datafusion/src/persistent/cache.rs
```diff
@@ -1,33 +1,24 @@
...
 
 impl CachedVortexMetadata {
     /// Create a new cached metadata entry from a VortexFile.
     pub fn new(vortex_file: &VortexFile) -> Self {
-        let footer = vortex_file.footer();
-        let memory_size = estimate_footer_size(footer);
-        Self {
-            footer: footer.clone(),
-            memory_size,
-        }
```
AdamGS, line 19: why this change?

vortex-datafusion/src/v2/table.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// Returns statistics for the full table, before any projection.
+    /// To keep this reasonably cheap, we just return cardinality and byte size estimates.
+    /// We provide full statistics from the physical plan.
+    fn statistics(&self) -> Option<Statistics> {
```
AdamGS, line 111: I think the API/intended flow here is that you calculate it once and store it once, but looking at datafusion I'm not even sure this is used anywhere, can keep the default impl that returns `None`

vortex-io/src/filesystem/glob.rs
AdamGS: also why do we need `impl dyn FileSystem`? can't these be default functions on the trait/on an ext trait?

vortex-datafusion/src/persistent/cache.rs
gatesn: For one of the short queries, the develop profile showed like 40% of time was in walking the full layout tree to estimate the size of the footer. Now we just hold that information since we typically have it anyway from the initial read.

benchmarks/datafusion-bench/src/main.rs
```diff
@@ -220,13 +222,20 @@ async fn main() -> anyhow::Result<()> {
     Ok(())
 }
 
+fn use_scan_api() -> bool {
+    std::env::var("VORTEX_USE_SCAN_API").is_ok_and(|v| v == "1")
```
joseph-isaacs, line 226: default is off?

vortex-file/src/v2/file_stats_reader.rs
joseph-isaacs, line 1: this is v2 only correct?

joseph-isaacs:
Are any changes made to the original scan?

gatesn:
Nope, just extracted some common logic in the DuckDB scan.rs to reuse in scan-api.rs

joseph-isaacs:
Where did you run the benchmarks for v2?

gatesn:
Previous edit of each comment should hold the results.

joseph-isaacs:
Shall we make a second PR with the CI flag switched

gatesn:
Yeah I have lots of follow ups before we even thinking about cutting over. Some queries are 15x slower due to lack of stats for example

vortex-datafusion/src/v2/source.rs
AdamGS: any reason not to use the existing conversion/splitting logic we already have?

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,498 @@
...
+        &self,
+        target_partitions: usize,
+        _repartition_file_min_size: usize,
+        _output_ordering: Option<LexOrdering>,
+    ) -> DFResult<Option<Arc<dyn DataSource>>> {
+        // Vortex handles parallelism internally — always use a single partition.
+        let mut this = self.clone();
+        this.num_partitions = NonZero::new(target_partitions)
+            .ok_or_else(|| DataFusionError::Internal("non-zero partitions".to_string()))?;
+        Ok(Some(Arc::new(this)))
```
gatesn, line 381: I want to store the updated num_partitions and use that as a concurrency hint for ourselves

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,498 @@
...
+
+/// Check if an expression tree contains decimal binary arithmetic that Vortex cannot handle.
+///
+/// DataFusion assumes different decimal types can be coerced, but Vortex expects exact type
+/// matches for binary operations. We avoid pushing these down.
+fn has_decimal_binary(expr: &Arc<dyn PhysicalExpr>, schema: &Schema) -> bool {
+    use datafusion_common::tree_node::TreeNode;
+
+    let mut found = false;
+    drop(expr.apply(|node| {
```
AdamGS: why the drop?

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,498 @@
...
+        dt,
+        DataType::Decimal32(_, _)
+            | DataType::Decimal64(_, _)
+            | DataType::Decimal128(_, _)
+            | DataType::Decimal256(_, _)
+    )
+}
+
+/// Convert a Vortex [`Option<Precision>`] to a DataFusion [`Precision`](DFPrecision).
+fn estimate_to_df_precision(est: &Option<Precision<u64>>) -> DFPrecision<usize> {
```
AdamGS, line 546: we have the `PrecisionExt` trait for this

vortex-datafusion/src/v2/table.rs
```diff
@@ -0,0 +1,132 @@
...
+        TableType::Base
+    }
+
+    async fn scan(
+        &self,
+        _state: &dyn Session,
+        // Unlike filters and limit, we _do_ apply the projection at this stage since DataFusion's
+        // physical projection expression push-down is still in its early stages. In theory, we
+        // could also wait to apply the projection until we can push down over the physical plan.
+        projection: Option<&Vec<usize>>,
```
AdamGS, line 79: you can, not sure what you mean here

vortex-file/src/multi/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+        .unwrap_or_else(|| file.path.clone());
+
+    // Build open options. The DashMap Ref from multi_file() must not live across an await,
+    // so we scope the cache lookup in a block.
+    let options = {
+        let mut options = open_options_fn(session.open_options());
+        if let Some(size) = file.size {
+            options = options.with_file_size(size);
+        }
+        if let Some(footer) = session.multi_file().get_footer(&cache_key) {
```
AdamGS, line 193: ~storing this on the session will cause the same issue we had with metrics, where memory usage can grow indefinitely~ I see its capped

vortex-file/src/multi/session.rs
```diff
@@ -0,0 +1,63 @@
...
+/// Consider generalizing this cache into [`VortexOpenOptions`](crate::VortexOpenOptions) so
+/// that single-file opens also benefit from session-level footer caching.
+pub(super) struct MultiFileSession {
+    footer_cache: moka::sync::Cache<String, Footer>,
+}
+
+impl Default for MultiFileSession {
+    fn default() -> Self {
+        Self {
+            footer_cache: moka::sync::Cache::builder().max_capacity(10_000).build(),
```
AdamGS: 10k footers is ~640MB, why not use the approximated size?

vortex-scan/src/multi.rs
```diff
@@ -0,0 +1,414 @@
...
+//!
+//! # Future Work
+//!
+//! - **Schema union**: Allow missing columns (filled with nulls) and compatible type upcasts
+//!   across sources instead of requiring exact dtype matches.
+//! - **Hive-style partitioning**: Extract partition values from file paths (e.g. `year=2024/month=01/`)
+//!   and expose them as virtual columns.
+//! - **Virtual columns**: `filename`, `file_row_number`, `file_index`.
+//! - **Per-file statistics**: Merge column statistics across sources for planner hints.
+//! - **Error resilience**: Skip failed sources instead of aborting the entire scan.
```
AdamGS, line 25: nit - seems like this should be an issue and not docs?

vortex-scan/src/multi.rs
```diff
@@ -0,0 +1,414 @@
...
+        let mut deferred = VecDeque::new();
+
+        let children = self.children.lock();
+        for (i, child) in children.iter().enumerate() {
+            match child {
+                MultiChild::Opened(ds) => ready.push_back(ds.clone()),
+                MultiChild::Deferred(factory) => deferred.push_back((i, factory.clone())),
+            }
+        }
+        drop(children);
```
AdamGS: can't you `std::mem::take` the children and then `children.into_iter()`?

benchmarks/datafusion-bench/src/main.rs
```diff
@@ -265,6 +274,54 @@ async fn register_benchmark_tables<B: Benchmark + ?Sized>(
...
+async fn register_v2_tables<B: Benchmark + ?Sized>(
+    session: &SessionContext,
+    benchmark: &B,
+    format: Format,
+) -> anyhow::Result<()> {
+    use vortex::file::multi::MultiFileDataSource;
+    use vortex::io::object_store::ObjectStoreFileSystem;
+    use vortex::io::session::RuntimeSessionExt;
+    use vortex::scan::api::DataSource as _;
+    use vortex_datafusion::v2::VortexTable;
```
AdamGS, line 287: move to top? also - lets log here so its visible which API is used

vortex-array/src/expr/stats/precision.rs
```diff
@@ -21,6 +21,8 @@ use crate::scalar::ScalarValue;
 /// This is statistic specific, for max this will be an upper bound. Meaning that the actual max
 /// in an array is guaranteed to be less than or equal to the inexact value, but equal to the exact
 /// value.
+///
+// TODO(ngates): should we model Unknown as a variant of Precision? Or have Option<Precision<T>>?
```
AdamGS, line 25: 👍

vortex-datafusion/src/persistent/cache.rs
```diff
@@ -42,42 +33,15 @@ impl FileMetadata for CachedVortexMetadata {
...
     }
 
     #[allow(clippy::disallowed_types)]
     fn extra_info(&self) -> std::collections::HashMap<String, String> {
         Default::default()
     }
 }
-
-/// Approximate the in-memory size of a footer.
-fn estimate_footer_size(footer: &Footer) -> usize {
```
AdamGS: why not use this logic?

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,498 @@
...
+    pub fn with_projection(mut self, indices: Vec<usize>) -> Self {
+        self.projection = Some(indices);
+        self
+    }
+
+    /// Configure an initial projection using top-level field indices.
+    pub fn with_some_projection(mut self, indices: Option<Vec<usize>>) -> Self {
+        self.projection = indices;
+        self
+    }
```
AdamGS, line 96: having both seems redundant

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,498 @@
...
+            write!(f, ", limit={limit}")?;
+        }
+        Ok(())
+    }
+
+    fn repartitioned(
+        &self,
+        target_partitions: usize,
+        _repartition_file_min_size: usize,
+        _output_ordering: Option<LexOrdering>,
```
AdamGS: if output ordering is provided, we might need to do something else (potentially, keep a single overall partition?).

vortex-datafusion/src/v2/table.rs
AdamGS: you can also store stats here from table creation time, and then pass them on to the scan, which is roughly what `ListingTable` and `FileScanConfig` do

vortex-file/src/multi/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub async fn build(mut self) -> VortexResult<MultiDataSource> {
+        let glob = self
+            .glob
+            .take()
+            .ok_or_else(|| vortex_err!("MultiFileDataSource requires a glob URL"))?;
+
+        let fs = self
+            .fs
+            .map(Ok)
+            .unwrap_or_else(|| create_local_filesystem(&self.session))?;
```
AdamGS:
```suggestion
        let fs = match self.fs {
            Some(fs) => fs,
            None => create_local_filesystem(&self.session)?,
        };
```

vortex-file/src/v2/file_stats_reader.rs
```diff
@@ -0,0 +1,331 @@
...
+        expr: &Expression,
+        mask: MaskFuture,
+    ) -> VortexResult<ArrayFuture> {
+        self.child.projection_evaluation(row_range, expr, mask)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #![allow(clippy::unwrap_used)]
```
AdamGS: thought its already allowed in tests?

vortex-scan/src/api.rs
```diff
@@ -20,56 +20,77 @@
...
-    async fn scan(&self, scan_request: ScanRequest) -> VortexResult<DataSourceScanRef>;
+    /// Returns an estimate of the byte size of the un-filtered source.
+    fn byte_size_estimate(&self) -> Option<Precision<u64>> {
+        None
+    }
 
-    /// Serialize a split from this data source.
-    fn serialize_split(&self, split: &dyn Split) -> VortexResult<Vec<u8>>;
+    /// Serialize the [`DataSource`] to pass to a remote worker.
+    fn serialize(&self) -> VortexResult<Option<Vec<u8>>> {
```
AdamGS, line 89: do we need to include serialization in this PR? Is it used anywhere?

vortex-scan/src/api.rs
```diff
@@ -20,56 +20,77 @@
...
+    /// Returns an estimate of the row count of the un-filtered source.
+    fn row_count_estimate(&self) -> Option<Precision<u64>> {
+        None
+    }
 
-    /// Returns a scan over the source.
-    async fn scan(&self, scan_request: ScanRequest) -> VortexResult<DataSourceScanRef>;
+    /// Returns an estimate of the byte size of the un-filtered source.
+    fn byte_size_estimate(&self) -> Option<Precision<u64>> {
+        None
```
AdamGS: nit - estimate in the name is redundant as the type itself communicates that

vortex-scan/src/multi.rs
```diff
@@ -0,0 +1,414 @@
...
+            children: Arc::new(Mutex::new(children)),
+            handle: session.handle(),
+            prefetch: DEFAULT_PREFETCH,
+        }
+    }
+
+    /// Sets the number of deferred sources to open concurrently during scanning.
+    ///
+    /// Higher values overlap more file-opening I/O with split execution but use more memory
+    /// for in-flight metadata. Defaults to 8.
```
AdamGS: nit - make the const public and point to it here, so docs don't go stale.

vortex-scan/src/multi.rs
```diff
@@ -0,0 +1,414 @@
...
+///
+/// Children may be pre-opened or deferred via [`DataSourceFactory`]. During scanning,
+/// deferred children are opened in the background using spawned tasks on the session's runtime,
+/// keeping the I/O pipeline full while the engine processes splits from already-open sources.
+///
+/// Once a deferred child is successfully opened, it is stored so that subsequent scans reuse the
+/// opened source without re-opening.
+pub struct MultiDataSource {
+    dtype: DType,
+    children: Arc<Mutex<Vec<MultiChild>>>,
```
AdamGS: I don't think the `Mutex` is need, if you remove it here and make `MultiChild` `Clone` everything works, cloning the vec a couple of times seems really minor and way less dangerous that running into an exclusive lock

vortex-datafusion/src/v2/source.rs
```diff
@@ -0,0 +1,498 @@
...
+        {
+            found = true;
+            return Ok(TreeNodeRecursion::Stop);
+        }
+        Ok(TreeNodeRecursion::Continue)
+    }));
+    found
+}
+
+fn is_decimal(dt: &DataType) -> bool {
```
AdamGS: nit - this is not needed anymore 🥳 there should be `DataType::is_decimal`

robert3005:
I don't quite know how to review this. My preference is starting with something fast that then could have adjusted api instead of starting with a desired api that then cannot be made fast

vortex-datafusion/src/v2/table.rs
```diff
@@ -0,0 +1,132 @@
...
+        // We ignore push-down of logical filters since Vortex requires a physical
+        //  expression (i.e. we require that coercion semantics have already been performed by the
+        //  engine). Instead, DataFusion will push down filters through the physical plan via
+        //  the VortexScanSource DataSource.
+        _filters: &[Expr],
```
AdamGS, line 80: we can extract column references from the filter here, which will tell us which statistics are worth fetching

vortex-datafusion/src/v2/table.rs
```diff
@@ -0,0 +1,132 @@
...
+        // Similarly for limit, we wait until we can push down over the physical plan.
+        _limit: Option<usize>,
```
AdamGS: I think we can use limit if there are no expressions

vortex-scan/src/multi.rs
```diff
@@ -0,0 +1,414 @@
...
+/// An async factory that produces a [`DataSource`].
+///
+/// Implementations handle engine-specific concerns like file opening, caching, and
+/// statistics-based pruning. Returns `None` if the source should be skipped (e.g., pruned
+/// based on file-level statistics).
+#[async_trait]
+pub trait DataSourceFactory: 'static + Send + Sync {
+    /// Opens the data source, or returns `None` if it should be skipped.
+    async fn open(&self) -> VortexResult<Option<DataSourceRef>>;
+}
```
AdamGS: I might be missing some detail here, but `DataSourceFactory` and `DataSourceOpener` seems really similar, not sure why we want both

vortex-scan/src/multi.rs
```diff
@@ -0,0 +1,414 @@
...
+            return current_estimate;
+        }
+
+        // With remaining sources whose split counts are unknown, return inexact.
+        let current_count = current_estimate.map(|p| p.into_inner()).unwrap_or(0);
+        Some(Precision::inexact(current_count))
+    }
+
+    fn splits(self: Box<Self>) -> SplitStream {
+        stream::unfold(
```
AdamGS: unfold is so useful but I really hate it 😢

robert3005:
Is there maybe a way to set these up as a additional benchmark variant for one of the benchmarks so we can see the constant comparison?

vortex-datafusion/src/persistent/cache.rs
AdamGS: that only happens on table creation opening though, which is << lifetime of the table/stats

vortex-cuda/kernels/src/bit_unpack_16.cu
```diff
@@ -4,11 +4,10 @@
 #include <stdint.h>
 #include "fastlanes_common.cuh"
 
-__device__ void
-_bit_unpack_16_0bw_lane(const uint16_t *__restrict in, uint16_t *__restrict out, unsigned int lane) {
+__device__ void _bit_unpack_16_0bw_lane(const uint16_t *__restrict in, uint16_t *__restrict out, unsigned int lane) {
```
robert3005: @0ax1 do we not validate formatting in ci?

0ax1: We do.

0ax1: https://github.com/vortex-data/vortex/commit/1eac4ed9a32162593c9d8a86eeb751df316cbdaf

AdamGS:
+1 for @robert3005 idea to benchmark both variants

vortex-datafusion/src/persistent/cache.rs
gatesn: But we're still basically walking a flatbuffer in an attempt to estimate the size of the byte buffer that sits behind that very same flatbuffer?

vortex-datafusion/src/v2/source.rs
gatesn: We have this pattern all over the place, makes fluent APIs nicer

vortex-datafusion/src/persistent/cache.rs
robert3005: separated this out https://github.com/vortex-data/vortex/pull/6601

vortex-datafusion/src/v2/table.rs
gatesn: No point really, doesn't let us prune anything

vortex-datafusion/src/v2/source.rs
gatesn: Yep, will do

vortex-datafusion/src/v2/table.rs
gatesn: But we also want statistics for aggregate push down, e.g. `SELECT MIN(col) FROM ...`

vortex-datafusion/src/v2/table.rs
gatesn: We can, but it doesn't change anything. A filter may be pushed later. So we can't prune splits here or do anything useful at all. Seems like a bug in the API to me

vortex-scan/src/multi.rs
gatesn: Somewhat incidental. In fact, this multi scan should change to operate over LayoutReaders rather than opaque child DataSource. We can do a much better job of scheduling across scans. 

Someone can implement a dumb UnionAllDataSource if they so wish, but we don't need one.

vortex-datafusion/src/v2/source.rs
gatesn: yah, DF internals just use this to say "preserve output ordering"

vortex-datafusion/src/v2/source.rs
gatesn: It's not quite the same

vortex-file/src/multi/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            fs: None,
+            glob: None,
+            open_options_fn: Arc::new(|opts| opts),
+        }
+    }
+
+    /// Set the path glob for file discovery.
+    ///
+    /// This path should be relative to the filesystem's base URL.
+    pub fn with_glob(mut self, glob: impl Into<String>) -> Self {
```
lwwmanning, line 73: should figure out a way to pass a list of explicit files rather than a glob (listing files is slow on object storage)

lwwmanning (approved):

vortex-file/src/multi/mod.rs
0ax1: Maybe we could lazily expand globs? Something along the lines of https://github.com/duckdb/duckdb/pull/20619

PR #6393, opened 2026-02-10
vortex-btrblocks: write listviews when elements expand > 50%

asubiotto:
Previously, listview arrays were always written as lists. This involves materializing the views (i.e. expanding the elements and reorganizing the offsets).

This commit chooses to write listviews on compression if the elements would expand > 50% when converting to lists to save on CPU cost and storage space. ZCTL listviews and listviews with low duplication are still encoded as lists, which saves writing the sizes.

### DIFF
--- a/vortex-btrblocks/src/canonical_compressor.rs
+++ b/vortex-btrblocks/src/canonical_compressor.rs
@@ -12,12 +12,14 @@ use vortex_array::arrays::ConstantArray;
 use vortex_array::arrays::ExtensionArray;
 use vortex_array::arrays::FixedSizeListArray;
 use vortex_array::arrays::ListArray;
+use vortex_array::arrays::ListViewArray;
 use vortex_array::arrays::StructArray;
 use vortex_array::arrays::TemporalArray;
 use vortex_array::arrays::list_from_list_view;
 use vortex_array::compute::Cost;
 use vortex_array::compute::IsConstantOpts;
 use vortex_array::compute::is_constant_opts;
+use vortex_array::compute::sum;
 use vortex_array::vtable::ValidityHelper;
 use vortex_dtype::DType;
 use vortex_dtype::Nullability;
@@ -37,6 +39,13 @@ use crate::compressor::float::FloatScheme;
 use crate::compressor::integer::IntegerScheme;
 use crate::compressor::string::StringScheme;
 use crate::compressor::temporal::compress_temporal;
+use crate::sample::sample;
+use crate::sample::sample_count_approx_one_percent;
+use crate::stats::SAMPLE_SIZE;
+
+/// Maximum ratio of expanded (List) element count to shared (ListView) element count
+/// below which we prefer List encoding over ListView.
+const MAX_LIST_EXPANSION_RATIO: f64 = 1.5;
 
 /// Trait for compressors that can compress canonical arrays.
 ///
@@ -137,6 +146,63 @@ impl BtrBlocksCompressor {
             btr_blocks_compressor: self,
         }
     }
+
+    /// Compresses a [`ListArray`] by narrowing offsets and recursively compressing elements.
+    fn compress_list_array(
+        &self,
+        list_array: ListArray,
+        ctx: CompressorContext,
+    ) -> VortexResult<ArrayRef> {
+        // Reset the offsets to remove garbage data that might prevent us from narrowing our
+        // offsets (there could be a large amount of trailing garbage data that the current
+        // views do not reference at all).
+        let list_array = list_array.reset_offsets(true)?;
+
+        let compressed_elems = self.compress(list_array.elements())?;
+
+        // Note that since the type of our offsets are not encoded in our `DType`, and since
+        // we guarantee above that all elements are referenced by offsets, we may narrow the
+        // widths.
+        let compressed_offsets = self.compress_canonical(
+            Canonical::Primitive(list_array.offsets().to_primitive().narrow()?),
+            ctx,
+            Excludes::from(&[IntCode::Dict]),
+        )?;
+
+        Ok(ListArray::try_new(
+            compressed_elems,
+            compressed_offsets,
+            list_array.validity().clone(),
+        )?
+        .into_array())
+    }
+
+    /// Compresses a [`ListViewArray`] by narrowing offsets/sizes and recursively compressing
+    /// elements.
+    fn compress_list_view_array(
+        &self,
+        list_view: ListViewArray,
+        ctx: CompressorContext,
+    ) -> VortexResult<ArrayRef> {
+        let compressed_elems = self.compress(list_view.elements())?;
+        let compressed_offsets = self.compress_canonical(
+            Canonical::Primitive(list_view.offsets().to_primitive().narrow()?),
+            ctx,
+            Excludes::none(),
+        )?;
+        let compressed_sizes = self.compress_canonical(
+            Canonical::Primitive(list_view.sizes().to_primitive().narrow()?),
+            ctx,
+            Excludes::none(),
+        )?;
+        Ok(ListViewArray::try_new(
+            compressed_elems,
+            compressed_offsets,
+            compressed_sizes,
+            list_view.validity().clone(),
+        )?
+        .into_array())
+    }
 }
 
 impl CanonicalCompressor for BtrBlocksCompressor {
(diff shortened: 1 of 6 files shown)


asubiotto:
I haven't given much though to the heuristic and just picked 50% so open to discussion here. We could also introduce this slowly via a write option maybe?

The motivation for this is that we're starting to do the work of moving our lists to ListViews and would like to keep the "dict-like encoding" + zero copy back to list view.

vortex-btrblocks/src/canonical_compressor.rs
```diff
@@ -179,33 +239,28 @@ impl CanonicalCompressor for BtrBlocksCompressor {
...
+                } else {
+                    // If this list view is not zctl, we sum the sizes to get an
+                    // idea of what the elements would expand to. If they would
+                    // expand to < 50%, we choose List over ListView.
+                    let trimmed = list_view_array.rebuild(ListViewRebuildMode::TrimElements)?;
+                    let elements_len = trimmed.elements().len();
+                    let potential_expanded_len = sum(trimmed.sizes())?
+                        .as_primitive()
+                        .as_::<usize>()
+                        .unwrap_or(0);
```
joseph-isaacs: you could try a sampling compression here?

vortex-btrblocks/src/canonical_compressor.rs
```diff
@@ -179,33 +239,28 @@ impl CanonicalCompressor for BtrBlocksCompressor {
...
-                    Excludes::from(&[IntCode::Dict]),
-                )?;
+                if list_view_array.is_zero_copy_to_list() {
+                    // We can avoid the sizes array.
+                    let list_array = list_from_list_view(list_view_array)?;
+                    self.compress_list_array(list_array, ctx)
+                } else {
+                    // If this list view is not zctl, we sum the sizes to get an
+                    // idea of what the elements would expand to. If they would
+                    // expand to < 50%, we choose List over ListView.
```
joseph-isaacs: I don't follow this 50% can you clarify

joseph-isaacs: I think the sign if the wrong way around. But I would say that its about being 150% of compressed size?

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: Good idea, I forgot about that. Will take a look

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: When going from listview->list, the elements are expanded if any views overlap. This 50% just means that if after expansion, the elements len is not 50% larger than before (1.5x), we still choose list. I agree it's confusing. Any suggestions on how to word it better?

joseph-isaacs:
```suggestion
                    // If this `ListView` is not zero-copy-to-list, we sum the duplicated element size. 
                    // If the List element size would be <150% of the ListView, keep the list.
```

vortex-btrblocks/src/canonical_compressor.rs
```diff
@@ -179,33 +239,29 @@ impl CanonicalCompressor for BtrBlocksCompressor {
...
+                        .as_::<usize>()
+                        .unwrap_or(0);
 
-                Ok(ListArray::try_new(
-                    compressed_elems,
-                    compressed_offsets,
-                    list_array.validity().clone(),
-                )?
-                .into_array())
+                    if elements_len == 0 || potential_expanded_len <= elements_len * 3 / 2 {
```
joseph-isaacs: can we also make this a constant

joseph-isaacs: any reason you choose 50% I would think you want a little less. But can defer this?

asubiotto: Nope, as I said this was completely random. Just wanted to choose something so we could choose something better together 😄

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: Do you mean something like (disregard SAMPLE_SIZE, that needs to be max(1%, SAMPLE_SIZE) and still need to verify the need to trim/rebuild)?
```

                    let sampled = sample(list_view_array.as_ref(), SAMPLE_SIZE, SAMPLE_COUNT);
                    let sample_ctx = ctx.as_sample();

                    let chunks: Vec<ArrayRef> = match sampled.as_opt::<ChunkedVTable
>() {
                        Some(chunked) => chunked.chunks().to_vec(),
                        None => vec![sampled],
                    };

                    let (mut list_nbytes, mut listview_nbytes) = (0u64, 0u64);
                    for chunk in chunks {
                        let Canonical::List(chunk_lv) = chunk.to_canonical()? else {
                            unreachable!("sampled ListView chunk should canonicalize to List");
                        };
                        // Trim elements per chunk so the comparison is fair (slices
                        // share the full elements buffer from the original array).
                        let chunk_lv = chunk_lv.rebuild(ListViewRebuildMode::TrimElements)?;

                        let as_list = list_from_list_view(chunk_lv.clone())?;
                        list_nbytes += self.compress_list_array(as_list, sample_ctx)?.nbytes();
                        listview_nbytes += self
                            .compress_list_view_array(chunk_lv, sample_ctx)?
                            .nbytes();
                    }
```

joseph-isaacs: I miss read this at first. Maybe we want to estimate size of both list and listview and compare?

asubiotto: Yeah, that might work although sizes are not uniformly distributed which is a bit of a concern.

joseph-isaacs: we could look for overlap in the ListView and take samples around there, but that might be costly

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: What are you thinking? `1.25`?. Actually if you draw a parallel with dictionary encoding, we only choose dict encoding if distinct values < 50%, which implies an equivalent ratio of `2`.

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: Updated

vortex-btrblocks/src/canonical_compressor.rs
joseph-isaacs: this is fine for now

joseph-isaacs:
lets runn benchmarks and we can merge if its all good

vortex-btrblocks/src/canonical_compressor.rs
```diff
@@ -179,33 +245,35 @@ impl CanonicalCompressor for BtrBlocksCompressor {
...
+                    let estimated_expanded = sampled_sum * list_view_array.len() / sampled_len;
+                    let elements_len = list_view_array.elements().len();
 
-                Ok(ListArray::try_new(
-                    compressed_elems,
-                    compressed_offsets,
-                    list_array.validity().clone(),
-                )?
-                .into_array())
+                    if elements_len == 0
```
connortsui20: I think you can move the `elements_len == 0` check to the top?

vortex-btrblocks/src/canonical_compressor.rs
```diff
@@ -179,33 +245,35 @@ impl CanonicalCompressor for BtrBlocksCompressor {
...
+                    let list_array = list_from_list_view(list_view_array)?;
+                    self.compress_list_array(list_array, ctx)
+                } else {
+                    // Sample the sizes to estimate the total expanded element
+                    // count, then decide List vs ListView with the expansion
+                    // threshold.
+                    let sample_count = sample_count_approx_one_percent(list_view_array.len());
+                    let sampled_sizes = sample(list_view_array.sizes(), SAMPLE_SIZE, sample_count);
+                    let sampled_len = sampled_sizes.len();
+                    let sampled_sum = sum(&*sampled_sizes)?
```
connortsui20: Nit: Could you call this `sampled_expanded_elements_len` or something similar? It was somewhat hard to grok this on a single pass

vortex-btrblocks/src/canonical_compressor.rs
```diff
@@ -137,6 +146,63 @@ impl BtrBlocksCompressor {
...
+        let compressed_offsets = self.compress_canonical(
+            Canonical::Primitive(list_view.offsets().to_primitive().narrow()?),
+            ctx,
+            Excludes::from(&[IntCode::Dict]),
+        )?;
+        let compressed_sizes = self.compress_canonical(
+            Canonical::Primitive(list_view.sizes().to_primitive().narrow()?),
+            ctx,
+            Excludes::none(),
+        )?;
```
connortsui20, line 197: I wonder if we are actually ok with dictionary compressing offsets in list views? Especially since we are now saying that we only use list views in files when they **do not** look like list arrays.

connortsui20 (commented):
Generally looks good to me! Left a few nits

connortsui20:
Is the TPC-H regression expected? It seems like quite a lot of queries slow down on the file-compressed benchmark.

To be honest, I don't really believe it at all

asubiotto:
No rush to merge this (I would actually probably prefer to not have this in the release in case there are unexpected regressions). I can look into the benchmark issues.

connortsui20:
@asubiotto I think the benchmark issues were just bad runs, so no need to worry

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: 👍🏼 resolving since the ratio has been extracted into a constant

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: Yep, done. Thanks for the suggestion, simplified more of the code flow as well.

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: Renamed to `estimated_expanded_elements_len`.

vortex-btrblocks/src/canonical_compressor.rs
asubiotto: I was wondering about this as well. I think it's fine to delegate that decision and we can exclude if it's an issue.

asubiotto:
Updated and addressed the latest comments. Rerunning the benchmarks

asubiotto:
Not sure what happened to the bench comments

robert3005:
the comments get updated, last edit seems to be from 1h ago which aligns with the time when labels were added last?

asubiotto:
Ah, you're right. For some reason I assumed they were new comments every time.

vortex-btrblocks/src/canonical_compressor.rs
```diff
@@ -179,33 +245,42 @@ impl CanonicalCompressor for BtrBlocksCompressor {
...
             Canonical::List(list_view_array) => {
-                // TODO(joe): We might want to write list views in the future and chose between
-                // list and list view.
-                let list_array = list_from_list_view(list_view_array)?;
-
-                // Reset the offsets to remove garbage data that might prevent us from narrowing our
-                // offsets (there could be a large amount of trailing garbage data that the current
-                // views do not reference at all).
-                let list_array = list_array.reset_offsets(true)?;
+                if list_view_array.is_zero_copy_to_list() {
```
connortsui20: Very nitpicky but you can do this all in 1 if instead of 2, and also (I think?) you can use `.is_empty()` (I'm not at a computer right now so can't double check)

asubiotto: Ah yes, thanks 🤦🏼 `elements_len` is used below so `is_empty()` doesn't make sense here

robert3005 (approved):

PR #6394, opened 2026-02-10
zstd buffers cuda kernel

onursatici:


encodings/zstd/src/zstd_buffers.rs
```diff
@@ -64,7 +66,82 @@ pub struct ZstdBuffersArray {
...
+
+    pub fn output_size_max(&self) -> usize {
+        self.output_size_max
+    }
+
+    pub fn num_frames(&self) -> usize {
+        self.compressed_buffers.len()
+    }
+
+    pub fn split_output_handle(
```
0ax1, line 111: Let's add a doc describing what this fn does.

encodings/zstd/src/zstd_buffers.rs
```diff
@@ -176,11 +254,70 @@ impl ZstdBuffersArray {
...
+            buffer_handles,
             &children,
             session,
         )
     }
+
+    pub fn decode_plan(&self) -> VortexResult<ZstdBuffersDecodePlan> {
+        // if invariants are somehow broken the device decompression could have UB, so ensure
+        // they still hold
+        self.validate()
```
0ax1, line 270: Panic or return error?

encodings/zstd/src/zstd_buffers.rs
```diff
@@ -150,22 +226,24 @@ impl ZstdBuffersArray {
...
         Ok(result)
     }
 
-    fn rebuild_inner(
+    fn decompress_and_build_inner(&self, session: &VortexSession) -> VortexResult<ArrayRef> {
+        let decompressed_buffers = self.decompress_buffers()?;
+        self.build_inner(&decompressed_buffers, session)
+    }
+
+    // this is exposed to help non cpu executors pass uncompressed buffer handles to build
```
0ax1: For all comments in this pr, double check capitalization and punctuation.

encodings/zstd/src/zstd_buffers.rs
```diff
@@ -140,7 +215,8 @@ impl ZstdBuffersArray {
             // `set_len(size)` after zstd reports how many bytes were written.
             let dst =
                 unsafe { std::slice::from_raw_parts_mut(spare.as_mut_ptr().cast::<u8>(), size) };
-            let written = decompressor.decompress_to_buffer(buf.as_slice(), dst)?;
+            let compressed = buf.clone().try_to_host_sync()?;
```
0ax1, line 222: As we're returning BufferHandle's now, why sync back to the host eagerly here? There's basically 3 modes for running cuda kernels in the grand scheme of things. Sync, rust async, or never wait and let a client synchronize based on a recorded cuda event (think sirius integration for arrow device buffers.)

onursatici: my intention was to have this as the cpu execution path, the only place this is called should be the scalar_at, validity and `execute`

encodings/zstd/src/zstd_buffers.rs
onursatici: I should be panicking on the next line

0ax1: Just wanted to poke. Agreed that panicking is prob the right thing to do.

encodings/zstd/src/zstd_buffers.rs
0ax1: Duh, my head is too wired into GPU kernels rn.

0ax1: Let's add a comment that documents that assumption.

onursatici: yeap sure will do

0ax1 (approved):
lgtm! let's get the docs in shape before we land this.

PR #6398, opened 2026-02-10
RFC: Add a pull request template

AdamGS:
I think the change is pretty self explanatory, I don't care about the details of the template here as much as I think we have too many PRs that are effectively opaque, with the details hiding in some discussion that isn't link, slack or being discussed offline.


0ax1 (commented):
lgtm!

joseph-isaacs (approved):

robert3005 (approved):

PR #6403, opened 2026-02-10
cuda strategy to use zstd buffers for strings

onursatici:


vortex-btrblocks/src/compressor/string.rs
```diff
@@ -231,6 +238,8 @@ pub enum StringCode {
     Sparse,
     /// Zstd compression without dictionaries.
     Zstd,
+    /// Zstd buffer-level compression preserving array layout.
+    ZstdBuffers,
```
joseph-isaacs, line 242: cfg or can you not?

joseph-isaacs (approved):

vortex-btrblocks/src/compressor/string.rs
onursatici: I think it is possible, but the impl itself is feature flagged and if I feature flag here then all matches would need that feature flag as well.
Currently if someone uses this without the feature flag it will be a no-op anyway

PR #6405, opened 2026-02-10
fix[ci]: make less gh requests per ci run

joseph-isaacs:
ci keeps timing out, maybe we can make less reqs to github?

### DIFF
--- a/.github/actions/setup-protoc/action.yml
+++ b/.github/actions/setup-protoc/action.yml
@@ -27,7 +27,6 @@ runs:
 
         PROTOC_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${{ inputs.protoc_version }}/protoc-${{ inputs.protoc_version }}-${PROTOC_ARCH}.zip"
         wget -q -O /tmp/protoc.zip "$PROTOC_URL"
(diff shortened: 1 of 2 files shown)


0ax1:
Is it time for strace: https://strace.io/ ?

.github/actions/setup-protoc/action.yml
```diff
@@ -0,0 +1,33 @@
+name: "Setup protoc"
+description: "Download and install protoc binary directly from GitHub releases CDN (no API calls)"
+inputs:
+  protoc_version:
+    description: "Version of the protoc binary"
+    default: "29.3"
```
robert3005, line 6: make this be latest? The thing that the action does is fetch latest version from github

joseph-isaacs: do we not want to pin?

robert3005: yes, we do, just make it the current latest version?

robert3005: oh I see the action doesn't actually update it

robert3005: I can pr upgrade separately

robert3005 (approved):

PR #6410, opened 2026-02-10
Introduce `VortexError::Other` and simplify `Display` implementation

AdamGS:

This is a long running low burning effort to improve errors.

1. Removes `VortexError::Url`, VortexError::SerdeJson and VortexError::Invalid state - they are all not used.
2. Removes `VortexError::try_from_slice` - this is an invariant and we always panic, no reason to treat it as a useful error.
3. When backtraces are disabled, we don't print them.
4. Splits the `Generic` error variant into `External` and `Other`, where `External` is meant to wrap external errors, and `Other` is for any error that doesn't match an existing variant. `Other` is now also the default of `vortex_err` and `vortex_bail`. They are used widely and often have very different meaning, I think its better to convey vague intent than the wrong one. 
5. I've also split out the individual components of the error `Display` form to make it clearer what gets printed and how its constructed.

Error display is a bit more compact when downstream users don't enable backtraces, and I believe this makes maintaining `vortex-error` a bit easier.

## How is this change tested?

In addition to existing tests that verify the result type, I've added a few tests that 

One less variant, and `vortex_err` and `vortex_bail` return a different type.

### DIFF
--- a/vortex-array/src/scalar/proto.rs
+++ b/vortex-array/src/scalar/proto.rs
@@ -183,14 +183,14 @@ impl Scalar {
             value
                 .dtype
                 .as_ref()
-                .ok_or_else(|| vortex_err!(InvalidSerde: "Scalar missing dtype"))?,
+                .ok_or_else(|| vortex_err!(Serde: "Scalar missing dtype"))?,
             session,
         )?;
 
         let pb_scalar_value: &pb::ScalarValue = value
             .value
             .as_ref()
-            .ok_or_else(|| vortex_err!(InvalidSerde: "Scalar missing value"))?;
+            .ok_or_else(|| vortex_err!(Serde: "Scalar missing value"))?;
 
         let value: Option<ScalarValue> = ScalarValue::from_proto(pb_scalar_value, &dtype)?;
 
@@ -224,7 +224,7 @@ impl ScalarValue {
         let kind = value
             .kind
             .as_ref()
-            .ok_or_else(|| vortex_err!(InvalidSerde: "Scalar value missing kind"))?;
+            .ok_or_else(|| vortex_err!(Serde: "Scalar value missing kind"))?;
 
         // `DType::Extension` store their serialized values using the storage `DType`.
         let dtype = match dtype {
@@ -251,7 +251,7 @@ impl ScalarValue {
 fn bool_from_proto(v: bool, dtype: &DType) -> VortexResult<ScalarValue> {
     vortex_ensure!(
         dtype.is_boolean(),
-        InvalidSerde: "expected Bool dtype for BoolValue, got {dtype}"
+        Serde: "expected Bool dtype for BoolValue, got {dtype}"
     );
 
     Ok(ScalarValue::Bool(v))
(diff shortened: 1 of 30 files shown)


vortex-error/src/lib.rs
```diff
@@ -371,6 +437,12 @@ macro_rules! vortex_err {
             $crate::VortexError::Context($msg.into(), Box::new($err))
         )
     }};
+    (External: $err:expr) => {{
```
AdamGS, line 411: In general - using macros is better than functions here as they don't frames

vortex-error/src/lib.rs
```diff
@@ -136,129 +138,186 @@ impl VortexError {
...
+            NotImplemented(..) => "Not implemented error: ",
+            MismatchedTypes(..) => "Mismatched types error: ",
+            AssertionFailed(..) => "Assertion failed error: ",
+            Context(..) | Shared(..) => "", // basically delegate to the underlying one
+            Arrow(..) => "Arrow error: ",
+            #[cfg(feature = "flatbuffers")]
+            FlatBuffers(..) => "Flat buffers error: ",
+            Fmt(..) => "Fmt: ",
+            Io(..) => "Io: ",
+            TryFromSlice(..) => "Try from slice error: ",
```
gatesn: This one also seems weirdly vague to me? Kill it?

vortex-error/src/lib.rs
```diff
@@ -136,129 +138,186 @@ impl VortexError {
...
+        match self {
+            Other(..) => "Other error: ",
+            External(..) => "External error: ",
+            OutOfBounds(..) => "Out of bounds error: ",
+            Compute(..) => "Compute error: ",
+            InvalidArgument(..) => "Invalid argument error: ",
+            InvalidState(..) => "Invalid state error: ",
+            Serde(..) => "Serde error: ",
+            NotImplemented(..) => "Not implemented error: ",
+            MismatchedTypes(..) => "Mismatched types error: ",
```
gatesn, line 145: Merge into Compute?

AdamGS: which one/s?

vortex-error/src/lib.rs
AdamGS: done

vortex-array/src/scalar/proto.rs
```diff
@@ -365,24 +364,49 @@ fn bytes_from_proto(bytes: &[u8], dtype: &DType) -> VortexResult<ScalarValue> {
...
             1 => DecimalValue::I8(bytes[0] as i8),
-            2 => DecimalValue::I16(i16::from_le_bytes(bytes.try_into()?)),
-            4 => DecimalValue::I32(i32::from_le_bytes(bytes.try_into()?)),
-            8 => DecimalValue::I64(i64::from_le_bytes(bytes.try_into()?)),
-            16 => DecimalValue::I128(i128::from_le_bytes(bytes.try_into()?)),
-            32 => DecimalValue::I256(i256::from_le_bytes(bytes.try_into()?)),
-            l => vortex_bail!(InvalidSerde: "invalid decimal byte length: {l}"),
+            2 => DecimalValue::I16(i16::from_le_bytes(
+                bytes
+                    .try_into()
```
AdamGS, line 369: This is ugly, I think there are a few variants of `VortexError` we can get rid of, I'll do that in a followup PR because it might be a bigger change.

robert3005:
any idea why `vortex-error::fmt test_basic_display` would panic? Ah `backtrace should include test function`

robert3005 (approved):
Let's merge this after Windows test pass, wouldn't want them failing for silly reasons

AdamGS:
going to try and debug it for a bit, if I can't figure it out and it just some weird windows thing, I'll `#[cfg]` that check away.

vortex-array/src/scalar/proto.rs
AdamGS: see #6554

PR #6414, opened 2026-02-10
Scalar truncation is implemented over scalar values instead of scalars and can operate in place

### LINKED ISSUE #6413: Lower and upper bound scalars
I was trying to avoid always creating new scalar here. It's likely that majority of these will be noops, that's why this took ownership and returned it back

_Originally posted by @robert3005 in https://github.com/vortex-data/vortex/pull/6309#discussion_r2782775224_

I should investigate if this change was necessary

robert3005:
As a followup to the massive scalar refactor I have refactored the truncation logic to not require typed scalars. Since we now guarantee that values will be coerced to appropriate scalar value we can implement truncation logic over the underlying primitives

closes #6413


vortex-scalar/src/scalar_value.rs
```diff
@@ -197,6 +197,58 @@ impl ScalarValue {
...
 }
 
+impl ScalarValue {
+    /// Returns the boolean value, panicking if the value is not a [`Bool`][ScalarValue::Bool].
+    pub fn into_bool(self) -> bool {
+        match self {
+            ScalarValue::Bool(b) => b,
+            _ => vortex_panic!("ScalarValue is not a Bool"),
+        }
+    }
```
connortsui20: I feel like this should live above with the as_ methods, but no strong feelings about it.

In fact, now that we have this maybe it should all go into `downcast.rs`?

robert3005: I can move this around

vortex-layout/src/layouts/flat/writer.rs
```diff
@@ -69,6 +71,23 @@ impl FlatLayoutStrategy {
...
 
+fn truncate_scalar_stat<F: Fn(Scalar) -> (Option<Scalar>, bool)>(
+    statistics: StatsSetRef<'_>,
+    stat: Stat,
+    truncation: F,
+) {
+    if let Some(sv) = statistics.get(stat) {
+        let (value, truncated) = truncation(sv.into_inner());
+        if let Some(upper_bound) = value {
+            if truncated && let Some(v) = upper_bound.into_value() {
```
connortsui20: rename `upper_bound` to `bound` or `truncated_value`?

Also a nit: using `let Some() = _ else` here makes things easier to read imo

vortex-layout/src/layouts/zoned/builder.rs
```diff
@@ -174,68 +176,99 @@ impl<T: ScalarTruncation> TruncatedMinBinaryStatsBuilder<T> {
...
-impl ScalarTruncation for Utf8Scalar<'_> {
-    fn from_scalar(value: &Scalar) -> VortexResult<impl ScalarTruncation> {
+impl ScalarTruncation for BufferString {
+    fn from_scalar(value: Scalar) -> VortexResult<Self> {
         value
-            .as_utf8_opt()
-            .ok_or_else(|| vortex_err!("Expected utf8 scalar, found {}", value.dtype()))
+            .into_value()
+            .map(|b| b.into_utf8())
+            .ok_or_else(|| vortex_err!("Expected binary scalar"))
```
connortsui20:
```suggestion
            .ok_or_else(|| vortex_err!("Expected utf8 scalar"))
```

vortex-layout/src/layouts/zoned/builder.rs
```diff
@@ -300,18 +356,131 @@ impl<T: ScalarTruncation> StatsArrayBuilder for TruncatedMinBinaryStatsBuilder<T
...
+            if v.len() > max_length {
+                (
+                    v.upper_bound(max_length)
+                        .map(|v| v.into_scalar(nullability)),
+                    true,
+                )
+            } else {
+                (Some(v.into_scalar(nullability)), false)
+            }
+        }
```
connortsui20: I would appreciate a doc comment on these, even after reading them it's hard for me to understand why this need to return a `(Option<Scalar>, bool)` and not a `Option<(Scalar, bool)>`, for example (which I think would make that `truncate_scalar_stat` function above easier to write with `?`s.

vortex-layout/src/layouts/zoned/builder.rs
```diff
@@ -174,68 +176,99 @@ impl<T: ScalarTruncation> TruncatedMinBinaryStatsBuilder<T> {
...
+        let sliced =
+            unsafe { BufferString::new_unchecked(self.into_inner().slice(..utf8_split_pos)) };
+        sliced.increment().ok()
     }
 
-    fn lower_bound(self, max_length: usize) -> Scalar {
-        Utf8Scalar::lower_bound(&self, max_length)
+    /// Construct a [`BufferString`] at most `max_length` in size that's less than or equal to
+    /// ourselves.
+    ///
```
connortsui20: nit: remove

vortex-scalar/src/typed_view/utf8.rs
```diff
@@ -232,26 +195,6 @@ mod tests {
     use crate::Scalar;
     use crate::Utf8Scalar;
 
-    #[test]
-    fn lower_bound() {
-        let utf8 = Scalar::utf8("snowman⛄️snowman", Nullability::NonNullable);
```
AdamGS, line 237: can't believe we're losing this amazing test

AdamGS: oh it just moved 😍

robert3005: indeed. text like this is the whole reason truncating strings is hard

vortex-scalar/src/downcast.rs
```diff
@@ -142,3 +148,119 @@ impl Scalar {
...
+            _ => vortex_panic!("ScalarValue is not a List"),
+        }
+    }
+
+    // pub fn as_extension(&self) -> &ExtScalarRef {
+    //     match self {
+    //         ScalarValue::Extension(e) => e,
+    //         _ => vortex_panic!("ScalarValue is not an Extension"),
+    //     }
+    // }
```
joseph-isaacs: ?

joseph-isaacs (approved):

PR #6417, opened 2026-02-11
fix[fuzz]: fuzzer failure alerts

joseph-isaacs:
Add alertings to CI!


0ax1 (approved):
Can we start adding more descriptive PR desc? What does this PR fix?

PR #6421, opened 2026-02-11
One duckdb string type to rule them all

AdamGS:

- Closes #.

Removes the `duckdb::string::String` type in favor of using `DDBString` everywhere.

Aside from the problematic naming, `duckdb::string::String` is just a subset of `DDBString`'s owned functionality.

## How is this change tested?

Existing tests

No, potentially only for other DuckDB extension authors.


vortex-duckdb/src/duckdb/ddb_string.rs
```diff
@@ -0,0 +1,54 @@
+use std::ffi::CStr;
```
AdamGS, line 4: figured it deserves its own file, `logical_type.rs` is already pretty long and doesn't have any other type definition in it.

0ax1 (approved):
🦆

robert3005 (approved):

PR #6423, opened 2026-02-11
lazy dict layout values eval

onursatici:

- Closes #.

make values child evaluation of an expression in the projection evaluation of a dict layout lazy. We want this because the current execute is assuming cpu only, with shared array we defer to the actual execute call to choose which execution it wants

### DIFF
--- a/vortex-layout/src/layouts/dict/reader.rs
+++ b/vortex-layout/src/layouts/dict/reader.rs
@@ -13,8 +13,6 @@ use futures::future::BoxFuture;
 use futures::try_join;
 use vortex_array::Array;
 use vortex_array::ArrayRef;
-use vortex_array::Canonical;
-use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
 use vortex_array::MaskFuture;
 use vortex_array::VortexSessionExecute;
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -119,16 +117,13 @@ impl DictReader {
...
         self.values_evals
             .entry(expr.clone())
             .or_insert_with(|| {
                 self.values_array()
                     .map(move |array| {
                         let array = array?.apply(&expr)?;
-                        // We execute the array to avoid re-evaluating for every split.
-                        let mut ctx = ExecutionCtx::new(session);
-                        Ok(array.execute::<Canonical>(&mut ctx)?.into_array())
+                        Ok(SharedArray::new(array).into_array())
```
joseph-isaacs, line 126: this has double shared arrays ?

onursatici: yes, the first one to cache the values as is, and this caches the result of the expression, so it only is executed in the first split

onursatici: if the expr is root we would have shared(shared(values)) which is redundant but should be fine

PR #6424, opened 2026-02-11
fix[fuzz]: correct stack frame compare

joseph-isaacs:
This fixes the fuzzer comparison between fuzz failures and issues


robert3005:
it's cute that it generates the tests as well

robert3005 (approved):

PR #6435, opened 2026-02-11
Implement ReduceNode for ScalarFnArray to avoid boxing

### LINKED ISSUE #6436: `ScalarFnAbstractReduceRule` is very slow
The `ScalarFnAbstractReduceRule` matches all scalar_fns and requires a `Arc`ing even if the rule clearly doesn't match. https://github.com/vortex-data/vortex/blob/d2d0fb883f7817dce96ab57d70c9b8039ae29b17/vortex-array/src/arrays/scalar_fn/rules.rs#L95

robert3005:
fix #6436

### DIFF
--- a/vortex-array/src/arrays/scalar_fn/rules.rs
+++ b/vortex-array/src/arrays/scalar_fn/rules.rs
@@ -90,11 +90,10 @@ impl ArrayReduceRule<ScalarFnVTable> for ScalarFnConstantRule {
 struct ScalarFnAbstractReduceRule;
 impl ArrayReduceRule<ScalarFnVTable> for ScalarFnAbstractReduceRule {
     fn reduce(&self, array: &ScalarFnArray) -> VortexResult<Option<ArrayRef>> {
-        if let Some(reduced) = array.scalar_fn.reduce(
-            // Blergh, re-boxing
-            &array.to_array(),
-            &ArrayReduceCtx { len: array.len },
-        )? {
+        if let Some(reduced) = array
+            .scalar_fn
+            .reduce(array, &ArrayReduceCtx { len: array.len })?
+        {
             return Ok(Some(
                 reduced
                     .as_any()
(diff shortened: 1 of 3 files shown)


vortex-array/src/arrays/scalar_fn/rules.rs
```diff
@@ -107,6 +106,29 @@ impl ArrayReduceRule<ScalarFnVTable> for ScalarFnAbstractReduceRule {
...
+    fn child(&self, idx: usize) -> ReduceNodeRef {
+        Arc::new(self.children()[idx].clone())
+    }
+
+    fn child_count(&self) -> usize {
+        self.children.len()
+    }
+}
+
 impl ReduceNode for ArrayRef {
```
joseph-isaacs, line 132: do we need this?

vortex-array/src/expr/vtable.rs
```diff
@@ -243,6 +243,7 @@ pub trait ReduceNode {
     fn child_count(&self) -> usize;
 
     /// Returns the children of this node.
+    #[allow(clippy::same_name_method)]
```
joseph-isaacs, line 246: where is the dup?

joseph-isaacs (approved):
There is a horrible naming in the trait

vortex-array/src/expr/vtable.rs
robert3005: I was struggling to make clippy happy. I had to put it on both of the functions

vortex-array/src/arrays/scalar_fn/rules.rs
robert3005: I have a followup that maybe lets us remove this

robert3005: We do need this just like we need `impl Array for Arc<dyn Array>`

gatesn: When do we pull the trigger on Array vtable that lets us access &self within the Arc.

It should come with a _lot_ of performance

PR #6439, opened 2026-02-11
Use `Scalar` directly for constant array metadata inlining

connortsui20:
Changes the metadata for `ConstantArray` to `Option<Scalar>`.

#6363 added functionality for inlining the constant scalar if it was small enough in the array metadata. This change adds to that by using a scalar directly for this. It is optional since we do not always want to do this.

Tests incoming!

Also, I think it is worth considering if we should _not_ write the scalars to the buffers if we make this optimization. As is, this is doing double work.

### DIFF
--- a/vortex-array/src/arrays/constant/vtable/mod.rs
+++ b/vortex-array/src/arrays/constant/vtable/mod.rs
@@ -39,7 +39,7 @@ impl ConstantVTable {
 
 /// Maximum size (in bytes) of a protobuf-encoded scalar value that will be inlined
 /// into the array metadata. Values larger than this are stored only in the buffer.
-const CONSTANT_INLINE_THRESHOLD: usize = 1024;
+pub(crate) const CONSTANT_INLINE_THRESHOLD: usize = 1024;
 
 impl VTable for ConstantVTable {
     type Array = ConstantArray;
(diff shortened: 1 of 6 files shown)


vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -49,7 +44,15 @@ const CONSTANT_INLINE_THRESHOLD: usize = 1024;
...
 
-    type Metadata = ProstMetadata<ConstantMetadata>;
+    /// Optional inlined scalar constant.
+    ///
+    /// When the scalar value is small enough (<= `CONSTANT_INLINE_THRESHOLD` bytes), it is stored
+    /// directly in the metadata to avoid an extra buffer allocation and potential
+    /// device-to-host copy during deserialization.
+    ///
+    /// Currently, scalars are **always** stored in a separate buffer, regardless of if we inline a
+    /// small scalar into the metadata.
```
connortsui20, line 54: See PR description above also.

We should be able to modify the visitor to just check if the scalar size is too big? But that could potentially be error-prone?

robert3005:
The problem is that you need this scalar in two places almost always. We were debating having a constant pool in the file to support batching all the constant array buffers

gatesn (approved):

PR #6446, opened 2026-02-12
feat[array]: refactor invert compute kernel to lazy `not`

joseph-isaacs:

- Closes #.

We want lazy compute. This PR moves invert compute to use the lazy compute model with array.not() building a scalar_fn with expr=not

```rust
#[deprecated(note = "use array.not() via ArrayBuiltins instead")]
pub fn invert(array: &dyn Array) -> VortexResult<ArrayRef> {
    array.to_array().not()
}
```


encodings/runend/src/compute/invert.rs
joseph-isaacs, line 1: scalar_fn pushdown handles these

robert3005 (approved):

PR #6447, opened 2026-02-12
refactor[array]: move boolean ops to lazy binary expr

joseph-isaacs:

- Closes #.

Move over boolean compute to lazy expr comptue

Want lazy compute for all compute functions, move over boolean here

from array/compute/binary deprecated 
```rust
#[deprecated(note = "Use and_kleene instead. Non-Kleene boolean ops cannot be lazily evaluated.")]
pub fn and(lhs: &dyn Array, rhs: &dyn Array)

#[deprecated(note = "Use or_kleene instead. Non-Kleene boolean ops cannot be lazily evaluated.")]
pub fn or(lhs: &dyn Array, rhs: &dyn Array) -> VortexResult<ArrayRef> {
```

We will soon deprecate all functions in this mod


vortex-array/src/compute/boolean.rs
```diff
@@ -78,131 +52,7 @@ pub fn or_kleene(lhs: &dyn Array, rhs: &dyn Array) -> VortexResult<ArrayRef> {
...
-            )
-        }
-
-        Ok(lhs.len())
-    }
-
-    fn is_elementwise(&self) -> bool {
-        true
-    }
+    arrow_boolean(lhs.to_array(), rhs.to_array(), op)
```
robert3005: This should build the binary expression

vortex-array/src/expr/exprs/operators.rs
```diff
@@ -213,3 +213,17 @@ impl TryInto<compute::Operator> for Operator {
...
     }
 }
+
+impl TryFrom<compute::BooleanOperator> for Operator {
+    type Error = VortexError;
+
+    fn try_from(value: compute::BooleanOperator) -> VortexResult<Self> {
+        match value {
+            compute::BooleanOperator::AndKleene => Ok(Operator::And),
+            compute::BooleanOperator::OrKleene => Ok(Operator::Or),
```
robert3005, line 225: I know this is not part of this pr but do we differentiate between `And` and `AndKleene` in the expressions?

robert3005 (approved):

vortex-array/src/expr/exprs/operators.rs
joseph-isaacs: All use kleene, I think we can rename and add the others with different enum int values

joseph-isaacs:
I will fixup between shortly

PR #6448, opened 2026-02-12
fix[ci]: pin nightly for miri

joseph-isaacs:

- Closes #.


.github/workflows/ci.yml
AdamGS: I think it should also go here

joseph-isaacs: both right?

robert3005 (approved):

PR #6450, opened 2026-02-12
Try and enable sccache on windows builds

AdamGS:

- Closes #.

runs-on doesn't point windows runs to sccache for some reason, so trying to see if it works.


AdamGS:
For future reference - runs-on doesn't enable sccache on windows for some reason, but also it doesn't seem to make much of a difference? Maybe under 1 minute, which is not much for this pretty slow run.

PR #6452, opened 2026-02-12
VarBinBuilder: eagerly set IsSorted stat on offsets

dimitarvdimitrov:

no

## What this PR does?

VarBinBuilder guarantees monotonically increasing offsets, but finish() wasn't setting the IsSorted stat. This caused an O(n) recomputation every time a deserialized VarBinArray was validated.

## How is this change tested?

unit tests

VarBin arrays now persist stats for their offsets child array. Existing persisted arrays will still be verified as sorted upon reading from disk.

### DIFF
--- a/encodings/runend/src/array.rs
+++ b/encodings/runend/src/array.rs
@@ -217,10 +217,20 @@ impl RunEndArray {
             return Ok(());
         }
 
-        // Run ends must be strictly sorted for binary search to work correctly.
-        if let Some(is_strict_sorted) = ends.statistics().compute_is_strict_sorted() {
-            vortex_ensure!(is_strict_sorted, "run ends must be strictly sorted");
-        }
+        debug_assert!({
+            // Run ends must be strictly sorted for binary search to work correctly.
+            let pre_validation = ends.statistics().to_owned().clone();
+
+            let is_sorted = ends
+                .statistics()
+                .compute_is_strict_sorted()
+                .unwrap_or(false);
+
+            // Preserve the original statistics since compute_is_strict_sorted may have mutated them.
+            // We don't want to run with different stats in debug mode and outside.
+            ends.statistics().inherit(pre_validation.iter());
+            is_sorted
+        });
 
         // Skip host-only validation when ends are not host-resident.
         if !ends.is_host() {
--- a/encodings/runend/src/array.rs
+++ b/encodings/runend/src/array.rs
@@ -16,8 +16,6 @@ use vortex_array::ProstMetadata;
 use vortex_array::SerializeMetadata;
 use vortex_array::arrays::PrimitiveVTable;
 use vortex_array::buffer::BufferHandle;
-use vortex_array::expr::stats::Precision as StatPrecision;
-use vortex_array::expr::stats::Stat;
 use vortex_array::search_sorted::SearchSorted;
 use vortex_array::search_sorted::SearchSortedSide;
 use vortex_array::serde::ArrayChildren;
@@ -336,13 +334,6 @@ impl RunEndArray {
     ) -> VortexResult<Self> {
         Self::validate(&ends, &values, offset, length)?;
 
-        // Run ends are always strictly sorted (and therefore sorted) by invariant.
-        // Cache this so downstream consumers don't need to recompute it.
-        ends.statistics()
-            .set(Stat::IsStrictSorted, StatPrecision::Exact(true.into()));
-        ends.statistics()
-            .set(Stat::IsSorted, StatPrecision::Exact(true.into()));
-
         Ok(Self {
             ends,
             values,
@@ -509,14 +500,10 @@ pub(super) fn run_end_canonicalize(
 mod tests {
     use vortex_array::IntoArray;
     use vortex_array::assert_arrays_eq;
-    use vortex_array::expr::stats::Precision as StatPrecision;
-    use vortex_array::expr::stats::Stat;
-    use vortex_array::expr::stats::StatsProviderExt;
     use vortex_buffer::buffer;
     use vortex_dtype::DType;
     use vortex_dtype::Nullability;
     use vortex_dtype::PType;
-    use vortex_error::VortexResult;
 
     use crate::RunEndArray;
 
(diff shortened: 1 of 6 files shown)


vortex-array/src/arrays/varbin/builder.rs
```diff
@@ -94,6 +96,13 @@ impl<O: IntegerPType> VarBinBuilder<O> {
 
         let validity = Validity::from_bit_buffer(nulls, dtype.nullability());
 
+        // The builder guarantees offsets are monotonically increasing, so we can set
+        // this stat eagerly. This avoids an O(n) recomputation when the array is
+        // deserialized and VarBinArray::validate checks sortedness.
+        offsets
+            .statistics()
+            .set(Stat::IsSorted, Precision::Exact(true.into()));
```
joseph-isaacs, line 108: can we add a debug_assert that these are sorted (so do a .compute in debug)

joseph-isaacs:
Can also do this for runend varbinview if not

dimitarvdimitrov:
> Can also do this for runend varbinview if not

i couldn't find where in the VarBinViewArrayBuilder we create runend offsets for VarBin(View) arrays. Can you point me pls?

joseph-isaacs:
varbinview: https://github.com/vortex-data/vortex/blob/8e92de5321acd4de44ac0746f5857bb128f08ec0/vortex-array/src/builders/varbinview.rs#L195

runend: https://github.com/vortex-data/vortex/blob/1cd554b000bbc2d89a197e9f13313e38051c2214/encodings/runend/src/compress.rs

also maybe sequence array:
https://github.com/vortex-data/vortex/tree/007ef8c8738cdcad0c8c3702ecc5222b35b97818/encodings/sequence but this might have a default stats set, since it always sorted (It might be descending does sorted allow that)?

dimitarvdimitrov:
I added validation for runend and sequence in 591c77a and d0f2dac and stored stats for the child arrays of those. but i couldn't figure out what we can do on VarBinView arrays

encodings/sequence/src/array.rs
```diff
@@ -127,13 +130,27 @@ impl SequenceArray {
...
+        // and strictly sorted iff multiplier > 0.
+        let (is_sorted, is_strict_sorted) = match_each_native_ptype!(ptype, |P| {
+            let m = multiplier.cast::<P>();
+            (m >= zero::<P>(), m > zero::<P>())
+        });
+
+        let stats_set = ArrayStats::default();
+        stats_set.set(Stat::IsSorted, StatPrecision::Exact(is_sorted.into()));
+        stats_set.set(
+            Stat::IsStrictSorted,
```
joseph-isaacs: think we can get away with a statically alloced CoW here so we don't need to alloc a stats set here?

Not sure. 

Maybe also include min and max and other stats?

encodings/sequence/src/array.rs
```diff
@@ -127,13 +130,27 @@ impl SequenceArray {
         length: usize,
     ) -> Self {
         let dtype = DType::Primitive(ptype, nullability);
+
+        // A sequence A[i] = base + i * multiplier is sorted iff multiplier >= 0,
+        // and strictly sorted iff multiplier > 0.
+        let (is_sorted, is_strict_sorted) = match_each_native_ptype!(ptype, |P| {
+            let m = multiplier.cast::<P>();
+            (m >= zero::<P>(), m > zero::<P>())
```
joseph-isaacs: not sure we allow a zero multiplier. if no we should not allow?

encodings/runend/src/array.rs
```diff
@@ -319,6 +326,13 @@ impl RunEndArray {
     ) -> VortexResult<Self> {
         Self::validate(&ends, &values, offset, length)?;
 
+        // Run ends are always strictly sorted (and therefore sorted) by invariant.
+        // Cache this so downstream consumers don't need to recompute it.
+        ends.statistics()
+            .set(Stat::IsStrictSorted, StatPrecision::Exact(true.into()));
```
joseph-isaacs: does this run in release? it's very slow. 

Can we set this on compression?

encodings/runend/src/array.rs
```diff
@@ -215,6 +217,11 @@ impl RunEndArray {
             return Ok(());
         }
 
+        // Run ends must be strictly sorted for binary search to work correctly.
+        if let Some(is_strict_sorted) = ends.statistics().compute_is_strict_sorted() {
```
joseph-isaacs: this is too expensive

joseph-isaacs (commented):
we need to be careful on validation cost on the release path

encodings/sequence/src/array.rs
dimitarvdimitrov: apparently there are test cases for 0 as a multiplier

https://github.com/vortex-data/vortex/blob/e0de842ae9a7788988774a12db047079bc725cab/encodings/sequence/src/compute/take.rs#L145

https://github.com/vortex-data/vortex/blob/e0de842ae9a7788988774a12db047079bc725cab/encodings/sequence/src/compute/mod.rs#L57

https://github.com/vortex-data/vortex/blob/e0de842ae9a7788988774a12db047079bc725cab/encodings/sequence/src/compute/cast.rs#L199

encodings/runend/src/array.rs
```diff
@@ -319,6 +336,13 @@ impl RunEndArray {
     ) -> VortexResult<Self> {
         Self::validate(&ends, &values, offset, length)?;
 
+        // Run ends are always strictly sorted (and therefore sorted) by invariant.
+        // Cache this so downstream consumers don't need to recompute it.
+        ends.statistics()
+            .set(Stat::IsStrictSorted, StatPrecision::Exact(true.into()));
+        ends.statistics()
+            .set(Stat::IsSorted, StatPrecision::Exact(true.into()));
```
joseph-isaacs: why is this now true?

encodings/runend/src/array.rs
```diff
@@ -319,6 +336,13 @@ impl RunEndArray {
     ) -> VortexResult<Self> {
         Self::validate(&ends, &values, offset, length)?;
 
+        // Run ends are always strictly sorted (and therefore sorted) by invariant.
+        // Cache this so downstream consumers don't need to recompute it.
+        ends.statistics()
+            .set(Stat::IsStrictSorted, StatPrecision::Exact(true.into()));
```
joseph-isaacs: This has to be, but I wouldn't write the add it to the stats without verifying that

joseph-isaacs (commented):
Did you need to add sortedness to the compressor

encodings/runend/src/array.rs
dimitarvdimitrov: mb, i didn't go through the rest of the code after we spoke about it. deleted now, along with the tests

dimitarvdimitrov:
> Did you need to add sortedness to the compressor

it already does it - https://github.com/vortex-data/vortex/blob/1cd554b000bbc2d89a197e9f13313e38051c2214/encodings/runend/src/compress.rs#L76-L77

dimitarvdimitrov:
any idea why the benchmarks are failing? i'll rebase to try and fix it

Screenshot 2026-02-13 at 15 28 31

dimitarvdimitrov:
@joseph-isaacs I think we addressed everything, can we merge this?

joseph-isaacs (approved):
LG

PR #6457, opened 2026-02-12
fetch nanoarrow with git instead of Apache

a10y:
Noticed some test flakes related to apache endpoint being unavailable.


robert3005:
apparently dist.apache.org is having a hard time recently

robert3005 (approved):

a10y:
I'm taking the risk that Github is more available than Apache.org, which is...a risk lol

but at least the fate is shared. if github action is running then git checkouts are working

PR #6462, opened 2026-02-12
Experiment with reduced allocations in MaskValues iterators

AdamGS:

- Closes #.

This change was taken out of (and expanded upon) work that was in #6337 originally. Deeper profiling of memory allocations showed that over 1% of ALL memory allocations when running TPC-H with Vortex and DataFusion is allocated for index-based iterators, in a relatively small number of allocations (just 0.02%).

These allocations happen within `Mask::indices`, which allocates a vec of the indices which is then used to iterate the set indices of the mask.

### DIFF
--- a/vortex-compute/src/filter/bitbuffer.rs
+++ b/vortex-compute/src/filter/bitbuffer.rs
@@ -38,7 +38,16 @@ impl Filter<MaskValues> for &BitBuffer {
             "Selection mask length must equal the mask length"
         );
 
-        self.filter(mask_values.indices())
+        let bools = self.inner().as_slice();
+        let bit_offset = self.offset();
+
+        BitBufferMut::from_iter(
+            mask_values
+                .bit_buffer()
+                .set_indices()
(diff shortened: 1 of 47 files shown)


vortex-compute/src/filter/bitbuffer.rs
```diff
@@ -38,7 +38,16 @@ impl Filter<MaskValues> for &BitBuffer {
...
 
-        self.filter(mask_values.indices())
+        let bools = self.inner().as_slice();
+        let bit_offset = self.offset();
+
+        BitBufferMut::from_iter(
+            mask_values
+                .bit_buffer()
+                .set_indices()
+                .map(|idx| get_bit(bools, bit_offset + idx)),
```
gatesn: Isn't this where collect_bool is supposed to be much faster?

robert3005:
It would be useful to get this merged, maybe the simplification is boolarray should have a cached true count?

AdamGS:
the size of the conflict at this point is ~ the size of the change 😅 I'll try and do it again including the cached true count at some point

PR #6464, opened 2026-02-12
feat(expr): implement Kleene logic boolean operators

### LINKED ISSUE #6458: `expr::Operator::{And/Or/Not}` is actually `Kleene{And/Or/No}`
We should add both operation variants see: https://github.com/vortex-data/vortex/blob/b288aac3f20e8be23fafff46c5c26bb807083e56/vortex-array/src/expr/exprs/operators.rs#L38

romeoscript:

- Closes #6458 .

### DIFF
--- a/vortex-array/src/expr/exprs/binary.rs
+++ b/vortex-array/src/expr/exprs/binary.rs
@@ -13,13 +13,6 @@ use vortex_session::VortexSession;
 
 use crate::ArrayRef;
 use crate::compute;
-use crate::compute::add;
-use crate::compute::and_kleene;
-use crate::compute::compare;
-use crate::compute::div;
-use crate::compute::mul;
-use crate::compute::or_kleene;
-use crate::compute::sub;
 use crate::expr::Arity;
 use crate::expr::ChildName;
 use crate::expr::ExecutionArgs;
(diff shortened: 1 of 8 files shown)


vortex-array/src/expr/mod.rs
```diff
@@ -76,7 +76,7 @@ pub fn split_conjunction(expr: &Expression) -> Vec<Expression> {
 
 fn split_inner(expr: &Expression, exprs: &mut Vec<Expression>) {
     match expr.as_opt::<Binary>() {
-        Some(operator) if *operator == Operator::And => {
+        Some(operator) if *operator == Operator::KleeneAnd => {
```
joseph-isaacs, line 79: What is the reason for changing this?

vortex-array/src/expr/exprs/not/mod.rs
```diff
@@ -114,6 +114,11 @@ impl VTable for Not {
     }
 }
 
+/// Creates an expression that logically inverts boolean values using Kleene logic.
+pub fn not_kleene(operand: Expression) -> Expression {
+    Not.new_expr(EmptyOptions, vec![operand])
+}
```
joseph-isaacs, line 120: this is not correct?

joseph-isaacs:
Mind explain the change you made?

romeoscript:
> Mind explain the change you made?

 I'm just separating standard boolean logic from Kleene logic to make things explicit. And/Or now follow standard null propagation while KleeneAnd/KleeneOr handle the Kleene-style logic we were using implicitly before.

vortex-array/src/expr/mod.rs
romeoscript: The reason is that Operator::And now represents Standard logic , but Vortex’s query engine was built assuming Kleene logic

vortex-array/src/expr/exprs/not/mod.rs
romeoscript: aparently this is redundant

vortex-array/src/expr/mod.rs
joseph-isaacs: What about regular and?

vortex-array/src/expr/exprs/not/mod.rs
joseph-isaacs: Then why include it?

vortex-array/src/expr/exprs/binary.rs
```diff
@@ -215,12 +210,12 @@ impl VTable for Binary {
...
                 Some(with_nan_predicate(lhs, rhs, min_max_check, catalog))
             }
-            Operator::And => lhs
+            Operator::And | Operator::KleeneAnd => lhs
                 .stat_falsification(catalog)
                 .into_iter()
                 .chain(rhs.stat_falsification(catalog))
-                .reduce(or),
-            Operator::Or => Some(and(
+                .reduce(or_kleene),
```
joseph-isaacs, line 217: Why does or map to or kleene here?

romeoscript: Yeah, since or() now defaults to strict null handling, we have to use  or_kleene() here to keep the current behavior. Plus, Kleene is actually better for stats since it's more aggressive with pruning when there are nulls

PR #6468, opened 2026-02-12
vortex-cuda format in benchmarks

onursatici:

- Closes #.

allows benchmarks to create and ran against vortex-cuda files, which are vortex files that only include cuda compatible encodings


robert3005:
If we want to use this we should also add GPU runners to benchmarks. I think it's too early to focus on benchmarking these

joseph-isaacs (approved):
Can you clean this up once its gone

onursatici:
If we have different strategies we expose, we should be able to generate benchmark data with all of those. This one is named cuda compatible but it is just a subset of encodings like the compact encodings, and I should be able to generate those and run my existing benchmark code on it. If I choose to run the query benchmark as is, it will run but will be executed on the cpu. If I have a gpu benchmark in the future, I still can use the data-gen binary to generate the files and now run on the gpu.

That being said, I am happy to keep this on my branch

PR #6469, opened 2026-02-12
feat[array]: lazy numeric

joseph-isaacs:
Part of the series of work to move over compute functions to be lazily executed expressions.

We remove all pushdown since it free from scalar_fns if we want to make add this We should add back a NumericReduce/NumericKernel trait with adaptors.


vortex-array/src/compute/numeric.rs
```diff
@@ -101,142 +77,7 @@ pub fn div_scalar(lhs: &dyn Array, rhs: Scalar) -> VortexResult<ArrayRef> {
...
-            .array()
-            .ok_or_else(|| vortex_err!("RHS is not an array"))?;
-        let operator = *args
-            .options
-            .as_any()
-            .downcast_ref::<NumericOperator>()
-            .ok_or_else(|| vortex_err!("Operator is not a numeric operator"))?;
-        Ok(Self { lhs, rhs, operator })
-    }
+    arrow_numeric(lhs, rhs, op)
```
robert3005, line 80: why do we remove the 3 specialised impls we had?

robert3005 (approved):

PR #6474, opened 2026-02-12
Change `ConstantArray::Metadata` to `Scalar`

connortsui20:
Based on top of https://github.com/vortex-data/vortex/pull/6473

This is also blocking the extension scalar work we want to do.


connortsui20:
Closing for now while we figure out if we want to have two discrete kinds of `ConstantArray` where one inlines the scalar into the metadata and the other keeps it in buffers.

PR #6477, opened 2026-02-12
Extension Scalars

connortsui20:
TODO

Right now this doesn't compile because it is blocked on a few other things. The tests in vortex-scalar do pass though (when it does compile), so not everything is terrible


vortex-scalar/src/extension/scalar_value.rs
```diff
@@ -0,0 +1,377 @@
...
+            })
+            .vortex_expect("Failed to downcast ExtScalar")
+    }
+
+    /// Attempts to downcast the vtable to a concrete [`ExtScalarVTable`] type by reference.
+    ///
+    /// Unlike [`try_downcast`], this borrows rather than consuming `self`.
+    ///
+    /// [`try_downcast`]: ExtScalarValueRef::try_downcast
+    pub fn try_get_vtable<V: ExtScalarVTable>(&self) -> Option<&V> {
```
gatesn, line 192: Would be keen to try and remove functions like this, they're probably superfluous / not called?

vortex-scalar/src/extension/scalar_value.rs
```diff
@@ -0,0 +1,377 @@
...
+    pub fn storage_value(&self) -> &ScalarValue {
+        self.0.storage_value()
+    }
+
+    /// Formats the extension scalar using the provided [`ExtDTypeRef`] for metadata context.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying [`fmt::Write`] operation fails.
+    pub fn fmt_ext_scalar(
```
gatesn: Why do we have this function on the public API?

vortex-scalar/src/extension/scalar_value.rs
```diff
@@ -0,0 +1,377 @@
...
+        self.0.id().hash(state);
+        self.0.storage_value().hash(state);
+    }
+}
+
+/// Private module to seal [`ExtScalarValueAdapterImpl`].
+///
+/// Note that this is not strictly necessary since [`ExtScalarValueAdapter`] and
+/// [`ExtScalarValueAdapterImpl`] are both private to this module, this is just for hygiene.
+mod sealed {
```
gatesn, line 334: You added sealed for scalars, but removed it for dtypes?

vortex-scalar/src/extension/scalar_value.rs
```diff
@@ -0,0 +1,377 @@
...
+}
+
+/// An object-safe, sealed trait encapsulating the behavior for extension scalar values.
+///
+/// This mirrors [`ExtDTypeImpl`] in `vortex-dtype`: it provides type-erased access to the
+/// extension scalar's identity, storage value, and display formatting. The only implementor is
+/// [`ExtScalarValueAdapter`].
+///
+/// [`ExtDTypeImpl`]: vortex_dtype::extension
+trait ExtScalarValueAdapterImpl: sealed::Sealed + 'static + Send + Sync {
```
gatesn: I think `DynExtScalarValue` is a bit neater than `AdapterImpl`

vortex-scalar/src/extension/vtable.rs
```diff
@@ -0,0 +1,91 @@
...
+        storage: ScalarValue,
+    ) -> VortexResult<ExtScalarValueRef>;
+}
+
+impl<V: ExtScalarVTable> DynExtScalarVTable for V {
+    fn id(&self) -> ExtID {
+        ExtDTypeVTable::id(self)
+    }
+
+    fn create_ext_scalar_value_ref(
```
gatesn: `fn build` is what we call it in array land?

vortex-scalar/src/typed_view/primitive/scalar.rs
gatesn, line 85: I really think we should kill this function if possible?

vortex-scalar/src/typed_view/ext.rs
```diff
@@ -0,0 +1,144 @@
...
+/// Extension types allow wrapping a storage type with custom semantics.
+#[derive(Debug, Clone)]
+pub struct ExtScalar<'a> {
+    /// The dtype reference TODO better docs.
+    dtype: &'a DType,
+
+    /// The extension data type reference.
+    ext_dtype: &'a ExtDTypeRef,
+
+    /// The underlying downcasted extension, or [`None`] if null.
```
gatesn: This is erased, not downcasted

vortex-scalar/src/typed_view/ext.rs
```diff
@@ -0,0 +1,144 @@
...
+
+/// A scalar value representing an extension type.
+///
+/// Extension types allow wrapping a storage type with custom semantics.
+#[derive(Debug, Clone)]
+pub struct ExtScalar<'a> {
+    /// The dtype reference TODO better docs.
+    dtype: &'a DType,
+
+    /// The extension data type reference.
```
gatesn: `data type` -> `dtype`

vortex-scalar/src/typed_view/ext.rs
```diff
@@ -0,0 +1,144 @@
...
+use crate::ScalarValue;
+use crate::extension::ExtScalarValueRef;
+use crate::extension::Matcher;
+
+/// A scalar value representing an extension type.
+///
+/// Extension types allow wrapping a storage type with custom semantics.
+#[derive(Debug, Clone)]
+pub struct ExtScalar<'a> {
+    /// The dtype reference TODO better docs.
```
gatesn: Remove todo

vortex-scalar/src/typed_view/ext.rs
```diff
@@ -0,0 +1,144 @@
...
+    ext_value: Option<&'a ExtScalarValueRef>,
+}
+
+impl<'a> ExtScalar<'a> {
+    /// TODO docs.
+    ///
+    /// # Errors
+    ///
+    /// TODO errors.
+    pub fn try_new(dtype: &'a DType, value: Option<&'a ScalarValue>) -> VortexResult<Self> {
```
gatesn, line 42: I don't think any of these "views" should be directly constructable, let alone with a pub interface

vortex-scalar/src/typed_view/mod.rs
```diff
@@ -18,7 +18,7 @@
 mod binary;
 mod bool;
 mod decimal;
-mod extension;
+mod ext;
```
gatesn, line 21: Why the name change? Needless API churn

vortex-scalar/src/arrow.rs
```diff
@@ -443,45 +440,6 @@ mod tests {
         Arc::<dyn Datum>::try_from(&list_scalar).unwrap();
     }
 
-    #[test]
-    #[should_panic(expected = "Cannot convert extension scalar")]
```
gatesn, line 447: This should still panic shouldn't it?

vortex-scalar/src/constructor.rs
```diff
@@ -170,27 +176,57 @@ impl Scalar {
...
-    /// Panics if the storage dtype of `ext_dtype` does not match `value`'s dtype.
-    pub fn extension_ref(ext_dtype: ExtDTypeRef, value: Scalar) -> Self {
-        assert_eq!(ext_dtype.storage_dtype(), value.dtype());
-        Self::try_new(DType::Extension(ext_dtype), value.into_value())
+    /// Panics if the storage dtype is incompatible with the extension type, or if the storage
+    /// value fails validation.
+    pub fn extension<V: ExtScalarVTable + Default>(metadata: V::Metadata, value: Scalar) -> Self {
+        let ext_dtype = ExtDType::<V>::try_new(metadata, value.dtype().clone())
+            .vortex_expect("Failed to create extension dtype");
+        let storage_value = value.into_value();
```
gatesn: You can `value.into_parts()` to avoid the dtype clone above.
Also rename `value` -> `storage`

vortex-scalar/src/scalar.rs
```diff
@@ -155,16 +152,27 @@ impl Scalar {
...
     // Other methods.
 
-    /// Check if the given [`ScalarValue`] is compatible with the given [`DType`].
-    pub fn is_compatible(dtype: &DType, value: Option<&ScalarValue>) -> bool {
+    /// Validate that the given [`ScalarValue`] is compatible with the given [`DType`].
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the given [`DType`] and [`ScalarValue`] are incompatible.
+    pub fn validate(dtype: &DType, value: Option<&ScalarValue>) -> VortexResult<()> {
```
gatesn, line 164: Curious if this actually needs to be public

vortex-scalar/src/scalar.rs
```diff
@@ -174,61 +182,101 @@ impl Scalar {
...
+                    );
+                }
             }
         }
+
+        Ok(())
+    }
+
+    /// Check if the given [`ScalarValue`] is compatible with the given [`DType`].
+    pub fn is_compatible(dtype: &DType, value: Option<&ScalarValue>) -> bool {
```
gatesn: Do we need this function?

gatesn (commented):
I'll have a think about how we can attach logic to the ExtDTypeVTable such that we don't need to provide the session if we already have a DType

vortex-scalar/src/extension/scalar_value.rs
connortsui20: In this case this is now a helper function for the matcher. I'll leave it in for now and we can remove it if it makes more sense

vortex-scalar/src/extension/scalar_value.rs
connortsui20: Imo we should probably seal the dtype one as well

vortex-scalar/src/typed_view/primitive/scalar.rs
connortsui20: The `try_new`? Are you saying that _everything_ has to be called via `new_unchecked`?

vortex-scalar/src/typed_view/mod.rs
connortsui20: It's not public (everything is re-exported) so there's no change. I wanted to make a distinction between this and the `extension` module that contains all the vtable logic.

vortex-scalar/src/arrow.rs
connortsui20: This now just doesn't even compile so unless we want to put it in a doctest (which I think it kind of pointless) we can just delete it.

vortex-scalar/src/constructor.rs
connortsui20: The dtype is the extension dtype, `into_parts` would give the storage dtype

vortex-scalar/src/scalar.rs
connortsui20: nope

vortex-scalar/src/scalar.rs
connortsui20: good question 🤔

PR #6483, opened 2026-02-13
feat[gpu]: CUDA dynamic dispatch kernel

0ax1:

- Dynamic dispatch GPU kernel for fused decoding operations 
- Support for source operations (right now only bitunpack) and scalar transforms (alp, for, zigzag)
- Documentation and usage guide

Fusing kernels avoids intermediate memory copies to global memory by chaining operations together in a single dynamic dispatch kernel.

## How is this change tested?

- Tests covering dynamic kernel dispatch with bitunpack, FoR, ALP

- New `DynamicDispatchPlan` API for composing decoding operations
- User guide in `dynamic_dispatch.md`


0ax1:
benchmark on gh200: 
```
bitunpack_for_alp/dynamic_dispatch_u32/100M
                        time:   [163.98 µs 165.58 µs 169.77 µs]
                        thrpt:  [2194.3 GiB/s 2249.8 GiB/s 2271.8 GiB/s]
```

2600GiB/s being the ref, running the bit-unpack kernel (6 bit) without dynamic dispatch in isolation.

vortex-cuda/cub/kernels/filter.cu
```diff
@@ -20,7 +20,7 @@ struct BitExtractor {
     const uint8_t* packed;
     uint64_t bit_offset;
 
-    __host__ __device__ __forceinline__
+    __host__ __device__ inline
```
joseph-isaacs, line 23: I think we do want force, but not gonna block

vortex-cuda/kernels/src/dynamic_dispatch.cu
joseph-isaacs, line 1: I think this would be more efficient to use registers!

vortex-cuda/src/dynamic_dispatch.rs
```diff
@@ -0,0 +1,373 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Host interface for dynamic CUDA kernel dispatch.
+//!
+//! Provides an API to dynamically dispatch a sequence of decoding operations
+//! defined in a plan, to execute them in a single GPU kernel.
+//!
+//! Each plan has a single source operation to unpack the compressed data into
+//! shared memory (e.g., bitunpack). After that scalar operations transform
+//! values in registers (e.g., FoR, zigzag, ALP).
```
joseph-isaacs, line 11: can we say that we could support register only ops, but we do this now

joseph-isaacs (approved):
LG, whats the plan to hook this into the executor?

vortex-cuda/kernels/src/dynamic_dispatch.cu
0ax1: Maybe? Let's explore this down the road.

PR #6484, opened 2026-02-13
Improve substrait support to include more types and expressions

AdamGS:

- Closes #.

1. Support for more types - including new time related ones and decimal.
2. Support for basic arithmetic operations (+,-,/,*).
3. Support for the `is_null` and `is_not_null` substrate expression, which also allows for querying vortex-backed datasets **with filters** through duckdb!
4. Move for `expr.evalute(..)` API to `array.apply(...)`, to align the python API better with the Rust one.

Both the arrow dataset API and substrait open up more ways to use Vortex through existing tools without maintaining dedicated integrations.

## How is this change tested?

Adds a few additional tests covering new code paths, and it also unlocks an existing test that was expected to fail.

Extends the public API surface in Python, doesn't break any existing code.


vortex-python/python/vortex/_lib/expr.pyi
```diff
@@ -16,6 +16,7 @@ IntoExpr: TypeAlias = Expr | int | str | date | datetime | None
 
 @final
 class Expr:
+    def evaluate(self, array: IntoArray) -> Array: ...
```
gatesn: I think we should use the "new" inverted call of `Array::apply(expr)`

gatesn (commented):
One comment, otherwise looks great thank you!

vortex-python/python/vortex/_lib/expr.pyi
AdamGS: I just moved it up here because it made more sense visually to me, but I'll add that change

vortex-python/src/expr/mod.rs
```diff
@@ -173,6 +174,34 @@ impl PyExpr {
...
+        right: &Bound<'py, PyAny>,
+    ) -> PyResult<Bound<'py, PyExpr>> {
+        py_binary_operator(self_, Operator::Mul, coerce_expr(right)?)
+    }
+
+    fn __truediv__<'py>(
+        self_: PyRef<'py, Self>,
+        right: &Bound<'py, PyAny>,
+    ) -> PyResult<Bound<'py, PyExpr>> {
+        py_binary_operator(self_, Operator::Div, coerce_expr(right)?)
```
danking, line 199: I think operator.__truediv__ should always return float: https://docs.python.org/3/library/operator.html#operator.truediv.

I think the only sensible way to implement this is:
1. truediv is Operator::Div on floating-point types; otherwise, type error.
2. floordiv is Operator::Div on non-floating-point types; otherwise, type error.

vortex-python/src/scalar/factory.rs
```diff
@@ -75,7 +75,16 @@ fn scalar_helper_inner(value: &Bound<'_, PyAny>, dtype: Option<&DType>) -> PyRes
...
+    // decimal
+    if let Some(dtype) = dtype.and_then(|d| d.as_decimal_opt()) {
+        return Ok(Scalar::decimal(
+            vortex::scalar::DecimalValue::I128(value.extract::<i128>()?),
+            *dtype,
+            Nullability::NonNullable,
+        ));
+    }
+
+    // int and decimal
```
danking: This comment seems wrong? I the type is unspecified, you'll never get a decimal value, right? I wonder if we should have a Python-side function like `decimal` or something to create decimals. I don't see anything in PyArrow.

Like, decimal should be able to exactly represent `3.33` but I'm not sure how to correctly communicate that from Python to Rust right now. Maybe `vx.decimal(3, 33)`?

vortex-python/test/test_dataset.py
```diff
@@ -108,14 +108,35 @@ def test_to_record_batch_reader_with_polars(ds: pd.Dataset):
...
+    tbl = ds.to_table(filter=((pc.field("index") * 2) < 10))
+    assert len(tbl) == 5
+
+    tbl = ds.to_table(filter=((pc.field("index") / 2) < 10))
+    assert len(tbl) == 20
+
+
 def test_duckdb(ds: vx.dataset.VortexDataset):
     assert ds  # pyright cannot determine that ds is used by duckdb.execute
-    # This would be a nice test but we do not support IsNotNull which duckdb uses
```
danking, line 113: ❤️

vortex-python/src/scalar/factory.rs
```diff
@@ -75,7 +75,16 @@ fn scalar_helper_inner(value: &Bound<'_, PyAny>, dtype: Option<&DType>) -> PyRes
         ));
     }
 
-    // int
+    // decimal
+    if let Some(dtype) = dtype.and_then(|d| d.as_decimal_opt()) {
+        return Ok(Scalar::decimal(
+            vortex::scalar::DecimalValue::I128(value.extract::<i128>()?),
```
danking: Does this mean I have to specify the I128 representation as a Python int to compare to a decimal column? Seems a bit weird

danking (dismissed):

vortex-python/src/scalar/factory.rs
AdamGS: need to think this through, maybe there's no way to avoid doing the full match here

vortex-python/src/scalar/factory.rs
AdamGS: yeah comment is wrong

vortex-python/python/vortex/_lib/expr.pyi
AdamGS: done

vortex-python/src/expr/mod.rs
gatesn: I would say for a DSL it's ok to abuse the operators a little bit? Particularly as expressions are untyped.

vortex-python/src/scalar/factory.rs
AdamGS: ended up doing it

gatesn (approved):

PR #6485, opened 2026-02-13
break[array]: mask compute to lazy mask execution

joseph-isaacs:
Make masking lazy, as part of general lazy compute migration
## breaks
The masking compute function is lazy


vortex-array/src/arrays/extension/compute/mask.rs
```diff
@@ -2,32 +2,35 @@
...
-                .ext_dtype()
-                .with_nullability(masked_storage.dtype().nullability()),
-            masked_storage,
-        )
-        .into_array())
+impl MaskReduce for ExtensionVTable {
+    fn mask(array: &ExtensionArray, validity: &Validity) -> VortexResult<Option<ArrayRef>> {
+        let Validity::Array(keep_mask) = validity else {
+            return Ok(None);
+        };
```
robert3005: this seems weird

joseph-isaacs: this handles it in the outer fn, I guess we could panic here?

robert3005: I would panic

robert3005:
I think you also need to change benchmarks to recursively canonicalize

robert3005 (approved):

PR #6489, opened 2026-02-13
Better error messages for lockfile conflicts

connortsui20:
A bit more verbose error messages for when the public API changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -348,8 +348,19 @@ jobs:
         run: bash scripts/public-api.sh
       - name: Verify lock files are up to date
         run: |
-          git diff --name-only
-          git diff --exit-code '**/public-api.lock'
+          if ! git diff --quiet '**/public-api.lock'; then
+            git diff -U0 '**/public-api.lock'
+            echo ""
+            echo ""
+            echo "Changed files:"
+            git diff --name-only '**/public-api.lock'
+            echo ""
+            echo ""
+            echo "Public API lock files are out of date."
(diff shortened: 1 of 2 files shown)


scripts/public-api.sh
```diff
@@ -4,6 +4,27 @@
...
+
+if ! cargo +nightly public-api --version &>/dev/null; then
+    echo "Error: `cargo-public-api` is not installed."
+    echo ""
+    echo "Install it with:"
+    echo "  cargo +stable install cargo-public-api --locked"
+    echo ""
+    echo "See https://crates.io/crates/cargo-public-api"
+    exit 1
+fi
```
joseph-isaacs, line 26: ❤️

.github/workflows/ci.yml
```diff
@@ -348,8 +348,19 @@ jobs:
...
       - name: Verify lock files are up to date
         run: |
-          git diff --name-only
-          git diff --exit-code '**/public-api.lock'
+          if ! git diff --quiet '**/public-api.lock'; then
+            git diff '**/public-api.lock'
+            echo ""
+            echo ""
+            echo "Changed files:"
+            git diff --name-only '**/public-api.lock'
```
joseph-isaacs, line 356: shall we remove this and just say there is an error regen?

connortsui20: I just (latest commit) made it so that this is not very verbose so unless the public API is changing in hundreds of places this will not be hard to read

a10y (approved):

PR #6492, opened 2026-02-13
Add take-based rebuild strategy for ListView with small lists

palaska:
- Split `naive_rebuild` into two strategies: `rebuild_with_take` (bulk take) and `rebuild_list_by_list` (per-list slice + builder)
 - Add a simple heuristic (`should_use_take`) that picks take when average list size < 128, defaulting to list-by-list otherwise
 - Take avoids per-list builder overhead (slice + extend) and is 10-100× faster for small lists
 - Arrow exporter uses the same codepath


connortsui20 (approved):
very nice! LGTM

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -159,14 +145,14 @@ impl ListViewArray {
...
-            new_elements_builder.extend_from_array(&elements_canonical.slice(start..stop)?);
+            take_indices.extend(start as u64..stop as u64);
 
             n_elements += num_traits::cast(size).vortex_expect("Cast failed");
         }
 
         let offsets = new_offsets.into_array();
         let sizes = new_sizes.into_array();
-        let elements = new_elements_builder.finish();
+        let elements = self.elements().take(take_indices.into_array())?;
```
joseph-isaacs: this builds a lazy take plan, do you want this?

joseph-isaacs: I think you want a execute here?

joseph-isaacs (changes requested):

joseph-isaacs:
Its faster if you don't do the work 🙈

connortsui20:
oh right all of our compute is lazy now haha

But I'll be surprised if this is not significantly better than the old naive version for primitive types

Edit: Actually are we sure we want to execute here? I don't think it is absolutely necessary to do so, especially if we want to filter/take _after_ a rebuild has occurred (in which case we can optimize that plan)

palaska:
Yeah my bad, didn't know take was lazy. It's still a lot faster than previous impl though

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -159,14 +145,18 @@ impl ListViewArray {
...
         }
 
         let offsets = new_offsets.into_array();
         let sizes = new_sizes.into_array();
-        let elements = new_elements_builder.finish();
+        let elements = self
+            .elements()
+            .take(take_indices.into_array())?
+            .to_canonical()?
+            .into_array();
```
connortsui20: In the case that we read `ListViewArray` from disk and we want to export to datafusion (basically if we want to export as a `ListArray` to Arrow), I believe that _not_ executing here might be more beneficial. 

If we want to push down a filter/take into a `ListView` from Datafusion/arrow, then we definitely do not want to execute here since it would be better to combine the filter/take with this take here.

I also think that this is the more common (perhaps only?) path that this `naive_rebuild` exists in.

cc @joseph-isaacs

connortsui20 (changes requested):

vortex-array/src/arrays/listview/rebuild.rs
palaska: Yeah, not sure what the caller expects to get here. I think it makes sense to keep it lazy and leave canonicalization up to the caller, wdyt @joseph-isaacs ?

joseph-isaacs: I think it depends if you clean up a large amount of element by being eager.

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -103,12 +104,147 @@ impl ListViewArray {
...
+
+        if use_per_list_copy {
+            self.naive_rebuild_per_list_copy::<O, NewOffset, S>()
+        } else {
+            self.naive_rebuild_bulk_take::<O, NewOffset, S>()
+        }
+    }
+
+    /// Decides whether the per-list-copy strategy should be used over bulk-take.
+    ///
```
connortsui20, line 130: I think this is good, would appreciate a small high-level comment here explaining the general gist of this, something like:

```
Rebuilding list by list can introduce unnecessary overhead if there are a lot of small lists. In those cases, it is better to model this rebuild as a take over indices.
Beyond that, the size of the elements can also determine which is faster. Generally, if the sizes of the lists are large, we want to do list-by-list `memcpy`s.
```

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -103,12 +104,147 @@ impl ListViewArray {
...
+            .as_slice::<S>()
+            .iter()
+            .map(|s| (*s).as_() as u64)
+            .sum();
+        let use_per_list_copy = Self::should_use_per_list_copy(element_dtype, total, self.len());
+
+        if use_per_list_copy {
+            self.naive_rebuild_per_list_copy::<O, NewOffset, S>()
+        } else {
+            self.naive_rebuild_bulk_take::<O, NewOffset, S>()
```
connortsui20: Could this be renamed to `rebuild_list_by_list` and `rebuild_with_take`? Or something along those lines, I personally feel that it is clearer than bulk take.

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -103,12 +104,147 @@ impl ListViewArray {
...
+            // FixedSizeList<T, N>: bulk-take expands each outer index into N inner
+            // sub-indices, making it ~2x more expensive per element than flat types.
+            // We account for this by dividing the base threshold by clamp(N, 1, 2):
+            //   - N >= 2: threshold halved (e.g. FSL<i32,4> → (8+4)*32 = 384)
+            //   - N == 1: threshold unchanged (same as flat types)
+            // Uses the *inner* element size, not the total FSL size, because per-list-copy
+            // extends the inner flat buffer directly.
+            DType::FixedSizeList(inner, n, ..) => inner
+                .element_size()
+                .is_some_and(|e| avg >= (8 + e as u64) * 64 / (*n as u64).clamp(1, 2)),
```
connortsui20: I'm not sure I understand the clamping here, and also for the example in the comment if we have an `FSL<i32, 4>`, then the element size is `sizeof(i32) * 4 = 4 * 4 = 16` bytes, not 4.

connortsui20 (changes requested):
I think this is probably fine, though because this is very heuristic-y I think we need some comments giving a general idea of why we are doing this.

Also, we still need to figure out if we want to do the eager or lazy `take` (do we canonicalize or let someone else do that so that we can potentially optimize the take?)

vortex-array/src/arrays/listview/rebuild.rs
connortsui20: Not sure why this would clean up more than if we wait? It's not like a join where we might balloon the data out, later takes and filters can only make this smaller right?

vortex-array/src/arrays/listview/rebuild.rs
palaska: The e here is the inner element size (4 bytes for i32), not the total FSL size (16 bytes for FSL<i32, 4>). This is because `rebuild_list_by_list` operates on the inner flat buffer directly — it extends the underlying i32 buffer, not a buffer of 4-element arrays.

The clamp was there to just cap the divisor because based on initial observation, there wasn't any difference between N=2 vs N=4. I ran a sweep with different N values (up to 4096) and Claude came up with a better formula. At high N's (256+), both strategies converge in cost

connortsui20:
Ok so it seems like there's no difference in our SQL benchmarks, can you make the change to the lazy take and see how that affects things? You can just add the action/benchmark-sql tag again and we can see if that causes anything

palaska:
is `statpopgen_q01/duckdb:vortex-compact` a flake?

connortsui20:
@palaska might be good to check?

Here is the query from `vortex-bench/statpopgen.sql`:

```
-- 1.
--
-- Count the number of samples (e.g. human beings) in this dataset. The GT arrays are all the same
-- length, even though they're not stored as a fixed-list type.
SELECT array_length("GT", 1) FROM statpopgen LIMIT 1;
```

I will check how long they actually are

connortsui20:
I'm just going to run this again and lets see if it happens again.

connortsui20:
Hmm so it is still slower:

```
statpopgen_q01/duckdb:vortex-compact | 115260258 | 8.20781e+07 | 1.40428 | ns | 🚨
```

I'm not sure we care about this too much though. So given that this change is driven more by Spiral we can always change this back. I think this is fine to merge.

connortsui20 (approved):

connortsui20:
Other than the merge conflict is there something blocking this?

palaska:
> Other than the merge conflict is there something blocking this?

I simplified the heuristic, I think it was unnecessarily overfitting. I will post a separate PR for the benchmarks alone in a sec

danking:
@joseph-isaacs do you still request changes?

joseph-isaacs (approved):
Lets run the benchmarks before we merge

joseph-isaacs (dismissed):
We need to justify/fix the codspeed benchmarks

robert3005:
Looks like the benchmarks are better because of lazy execution, we now wrap the elements in the take. I think for these benchmarks you need to execute into RecursiveCanonical to compare before and after

gatesn:
If we had a good representation for masks with long runs of values, would slice + extend ever be faster? 

By the way, this is the same logic we have in the Arrow exporter. Maybe we can unify?

palaska:
> If we had a good representation for masks with long runs of values, would slice + extend ever be faster?

I don't think so. It'd be nice to take with a range-based mask (would work nicely since our indices are always contiguous here), which would just memcpy each range, right?

> By the way, this is the same logic we have in the Arrow exporter. Maybe we can unify?

Hmm, https://github.com/apache/arrow-rs/pull/8735 is merged and it also builds the flat indices vec and does `take` under the hood. I think it makes sense to reuse ours here as well since slice + extend outperforms this on large lists. I will unify

palaska:
> Looks like the benchmarks are better because of lazy execution, we now wrap the elements in the take. I think for these benchmarks you need to execute into RecursiveCanonical to compare before and after

Benchmarks on develop are forcing canonicalization now so the improvements on the codspeed comment are real.

robert3005 (approved):

PR #6493, opened 2026-02-13
switch to lockfiles CLI

a10y:
The deps here don't change often which means that you should mostly just compile this once, and run `cargo -p lockfiles` to fix all lockfiles.

Uses rayon to parallelize


a10y:
could also hoist this out and binstall

robert3005:
I think because cargo doc doesn't support multithreading you will not be any faster.

a10y:
You mean within a workspace it doesn't?

robert3005:
yes, that's what I am guessing, cargo takes a lock on the workspace when it runs

PR #6495, opened 2026-02-13
break[array]: lazy compare function

joseph-isaacs:
lazy compare, see: https://github.com/vortex-data/vortex/discussions/5921

##break

vortex_array compute/compare is not lazy.


vortex-array/src/arrays/constant/compute/compare.rs
joseph-isaacs, line 1: directly in the expr

vortex-array/src/arrays/masked/vtable/kernel.rs
```diff
@@ -5,5 +5,6 @@ use crate::arrays::MaskedVTable;
 use crate::arrays::TakeExecuteAdaptor;
 use crate::kernel::ParentKernelSet;
 
+// TODO(joe): add CompareExecuteAdaptor to push comparisons through the mask without canonicalizing.
```
robert3005, line 8: nice todo you got here

robert3005:
You want to change encodings/fsst/benches/fsst_compress.rs to use RecursiveCanonical, right now you have 50x improvement because functions are constant

robert3005 (approved):

PR #6498, opened 2026-02-13
feat[xtask]: faster public-api

joseph-isaacs:
much faster public api run


xtask/src/public_api.rs
joseph-isaacs, line 1: this is the whole changes

xtask/Cargo.toml
```diff
@@ -19,6 +19,7 @@ version = { workspace = true }
 anyhow = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
 prost-build = { workspace = true }
+public-api = "0.50"
```
robert3005: can you extract it to workspace level?

robert3005 (approved):

PR #6499, opened 2026-02-13
Add more SQL tests comparing DF and DuckDB

AdamGS:
More tests covering a bunch of basic SQL flows using both SQL engines we integrate with.

This is mostly about improving coverage and making sure we behave as expected on both, or that we can explain any difference.


joseph-isaacs (approved):
Look at duckdb-vortex repo and add those tests too?

PR #6501, opened 2026-02-13
fix[xtask]: faster public-api

joseph-isaacs:
fixup the lock file builder


vortex-tui/Cargo.toml
```diff
@@ -12,6 +12,7 @@ readme = "README.md"
 repository = { workspace = true }
 rust-version = { workspace = true }
 version = { workspace = true }
+publish = false
```
a10y: we actually do want to publish this one #5742

joseph-isaacs: I did think, but do we want to track public apis?

a10y: if it's a library then problaby?

a10y (approved):

PR #6504, opened 2026-02-13
fix[ci]: faster cuda build

joseph-isaacs:
This causes a single build for all tests instead of one build per step


0ax1:
Please add a desc why this makes the build faster.

0ax1 (approved):

PR #6505, opened 2026-02-13
fix[ci]: cuda no nightly

joseph-isaacs:
Dont run cuda with nightly rustc in CI


0ax1:
Worth scanning all GH actions for `toolchain: ${{ env.NIGHTLY_TOOLCHAIN }}` and removing them if not needed. This looks more like Claude spotted this as a pattern and then also added it for CUDA.

joseph-isaacs:
Ill follow up

0ax1 (approved):

PR #6503, opened 2026-02-13
fix handling of fill_null precondition cast

a10y:
## Does this PR closes an open issue or discussion?

Closes #6496

## What changes are included in this PR?

fill_null scalar fn execution will execute() its precondition result. In this case, the precondition was returning an unexecuted `CastArray` which breaks the assumption that ScalarFn::execute() always makes forward progress toward `Columnar`

PR #6506, opened 2026-02-13
add TODO

a10y:
Forgot to address https://github.com/vortex-data/vortex/pull/6503#discussion_r2805542176

### DIFF
--- a/vortex-array/src/expr/exprs/fill_null/mod.rs
+++ b/vortex-array/src/expr/exprs/fill_null/mod.rs
@@ -159,6 +159,8 @@ fn fill_null_canonical(
     if let Some(result) = precondition(canonical.as_ref(), fill_value)? {
         // The result of precondition may return another ScalarFn, in which case we should
         // apply it immediately.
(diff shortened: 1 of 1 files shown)


vortex-array/src/expr/exprs/fill_null/mod.rs
```diff
@@ -159,7 +160,8 @@ fn fill_null_canonical(
     if let Some(result) = precondition(canonical.as_ref(), fill_value)? {
         // The result of precondition may return another ScalarFn, in which case we should
         // apply it immediately.
-        return result.execute::<ArrayRef>(ctx);
+        // TODO(aduffy): Remove this once we have better driver check
+        return result.optimize()?.execute::<ArrayRef>(ctx);
```
joseph-isaacs: in fact check if the cast applies one itself?

a10y: updated

joseph-isaacs (approved):

PR #6507, opened 2026-02-13
chore: bump to rust 1.91

joseph-isaacs:
This is required by packages we use


a10y (changes requested):
I think the real fix is to disable default features: https://github.com/vortex-data/vortex/pull/6511

Note that lance has MSRV 1.88.0

PR #6511, opened 2026-02-13
disable default features for lance

a10y:
CI is failing currently due to some aws related deps that are brought in transitively into our workspace through lance.

Lance default features bring in a bunch of deps for talking to aws/gcp/azure/Datafusion integration that we don't care about for the compression bench.

### DIFF
--- a/vortex-duckdb/src/scan.rs
+++ b/vortex-duckdb/src/scan.rs
@@ -465,9 +465,9 @@ impl TableFunction for VortexTableFunction {
         unsafe {
             use custom_labels::sys;
 
-            if sys::labelset_current().is_null() {
-                let ls = sys::labelset_new(0);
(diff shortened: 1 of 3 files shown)


robert3005 (approved):

a10y:
What the heck is `customlabels`

vortex-duckdb/src/scan.rs
```diff
@@ -465,9 +465,9 @@ impl TableFunction for VortexTableFunction {
         unsafe {
             use custom_labels::sys;
 
-            if sys::labelset_current().is_null() {
```
robert3005, line 468: this seems wrong, customlabels is a polarsignals things for adding labels to your binaries

robert3005: ah I see they renamed things

PR #6512, opened 2026-02-14
Configure DuckDB native threads for benchmarking

gatesn:
Benchmarking takes a "threads" flag. We used to only set this on the Vortex table function, not the rest of DuckDB. This was different to how the same flag behaves in DataFusion which configures the entire Tokio runtime with n threads.

I can confirm it runs with a single thread when passed

Screenshot 2026-02-13 at 20 21 27

### DIFF
--- a/benchmarks/duckdb-bench/src/lib.rs
+++ b/benchmarks/duckdb-bench/src/lib.rs
@@ -17,7 +17,6 @@ use vortex_bench::generate_duckdb_registration_sql;
 use vortex_duckdb::duckdb::Config;
 use vortex_duckdb::duckdb::Connection;
 use vortex_duckdb::duckdb::Database;
-use vortex_duckdb::register_extension_options;
 
 /// DuckDB context for benchmarks.
 pub struct DuckClient {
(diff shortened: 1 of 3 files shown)


vortex-duckdb/src/scan.rs
```diff
@@ -389,13 +366,13 @@ impl TableFunction for VortexTableFunction {
...
         );
 
-        // Use the max_threads from bind_data (read from vortex_max_threads setting)
-        #[expect(clippy::cast_possible_truncation, reason = "max_threads fits in usize")]
-        let num_workers = bind_data.max_threads as usize;
-
         let client_context = init_input.client_context()?;
         let object_cache = client_context.object_cache();
 
+        let num_workers = std::thread::available_parallelism()
```
0ax1, line 372: For the extension, max_threads means up to how many threads are suppported by the extension, not how many should be used. That's why extensions use u64::MAX to signal that any thread count is supported.

So should DuckDB decide to use more threads than logical cores, they would mismatch.

benchmarks/duckdb-bench/src/lib.rs
```diff
@@ -67,10 +66,12 @@ impl DuckClient {
...
         threads: Option<usize>,
     ) -> Result<(Database, Connection)> {
-        let config = Config::new().vortex_expect("failed to create duckdb config");
+        let mut config = Config::new().vortex_expect("failed to create duckdb config");
 
-        // Register Vortex extension options before creating connection
-        register_extension_options(&config);
+        // Set DuckDB thread count if specified
+        if let Some(thread_count) = threads {
+            config.set("threads", &format!("{}", thread_count))?;
```
0ax1: Does this actually work?! Surprised we didn't do this before.

0ax1 (approved):

PR #6520, opened 2026-02-16
fix[layout]: snapshot nbytes at push time in ChunksBuffer to avoid underflow

brainhart:

- N/A

- Changed ChunksBuffer.data from VecDeque<ArrayRef> to VecDeque<(ArrayRef, u64)>, snapshotting each chunk's nbytes() at push time.
- pop_front now subtracts the stored snapshot instead of re-querying chunk.nbytes().
- Added a regression test that reproduces the underflow via SharedArray state transition.

SharedArray slices sharing an Arc<Mutex<SharedState>> can transition from Source to Cached when any one of them is canonicalized. This is interior mutation — it happens without the ChunksBuffer knowing. Because nbytes() traverses the current array tree, a chunk that reported a small nbytes() at push time (compressed/source-era) can report a much larger value at pop time (decompressed/cached-era). The running nbytes total was accumulated with the smaller values, so subtracting the larger value causes an integer underflow panic (attempt to subtract with overflow).

By snapshotting nbytes() once at push time and storing it alongside the chunk, the accounting stays self-consistent regardless of any interior state changes between push and pop.

## How is this change tested?

A new unit test chunks_buffer_pop_front_no_panic_after_shared_execution reproduces the exact failure scenario:

1. Creates a ConstantArray (tiny nbytes) wrapped in a SharedArray.
2. Slices it into two chunks and pushes both into a ChunksBuffer.
3. Pops the first chunk (succeeds, no state change yet).
4. Triggers canonicalization via shared_handle.get_or_compute(), which transitions all slices to Cached state with much larger nbytes.
5. Pops the second chunk — before this fix, this panicked with underflow.

No. This is an internal bug fix in the repartition writer. No public APIs are changed.

### DIFF
--- a/vortex-layout/src/layouts/repartition.rs
+++ b/vortex-layout/src/layouts/repartition.rs
@@ -155,7 +155,7 @@ impl LayoutStrategy for RepartitionStrategy {
                 }
                 if canonical_stream.as_mut().peek().await.is_none() {
                     let to_flush = ChunkedArray::try_new(
-                        chunks.data.drain(..).collect(),
+                        chunks.data.drain(..).map(|(arr, _)| arr).collect(),
                         dtype_clone.clone(),
                     )?;
                     if !to_flush.is_empty() {
@@ -189,7 +189,10 @@ impl LayoutStrategy for RepartitionStrategy {
 }
 
 struct ChunksBuffer {
-    data: VecDeque<ArrayRef>,
+    /// Each entry stores the chunk and the `nbytes()` snapshot taken at push time.
+    /// This avoids accounting mismatches when interior-mutable arrays (e.g. `SharedArray`)
+    /// change their reported size after being pushed.
+    data: VecDeque<(ArrayRef, u64)>,
     row_count: usize,
     nbytes: u64,
     block_size_minimum: u64,
@@ -216,7 +219,7 @@ impl ChunksBuffer {
         let mut res = Vec::with_capacity(self.data.len());
         let mut remaining = nblocks * self.block_len_multiple;
         while remaining > 0 {
-            let chunk = self
+            let (chunk, _) = self
                 .pop_front()
                 .vortex_expect("must have at least one chunk");
             let len = chunk.len();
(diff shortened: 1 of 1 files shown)


vortex-layout/src/layouts/repartition.rs
```diff
@@ -443,4 +449,43 @@ mod tests {
...
+    /// Regression test: `SharedArray` slices sharing an `Arc<Mutex<SharedState>>` can
+    /// transition from Source to Cached when any one of them is canonicalized. This caused
+    /// `pop_front` to panic with `attempt to subtract with overflow` because the buffer's
+    /// running `nbytes` total was accumulated with the smaller Source-era values while
+    /// `pop_front` subtracted the larger Cached-era values.
+    #[test]
+    fn chunks_buffer_pop_front_no_panic_after_shared_execution() -> VortexResult<()> {
+        use vortex_array::Array;
+        use vortex_array::arrays::ConstantArray;
+        use vortex_array::arrays::SharedArray;
```
connortsui20: Nit: can you move this to where the rest of the imports are?

connortsui20 (approved):
LGTM, and can confirm the regression test fails on develop and passes here. I think it would be good if someone else takes a look at this though since I am not too familiar with this code.

vortex-layout/src/layouts/repartition.rs
```diff
@@ -236,22 +239,25 @@ impl ChunksBuffer {
...
-        self.data.push_front(chunk);
+        self.nbytes += nb;
+        self.data.push_front((chunk, nb));
     }
 
-    fn pop_front(&mut self) -> Option<ArrayRef> {
+    fn pop_front(&mut self) -> Option<(ArrayRef, u64)> {
         let res = self.data.pop_front();
-        if let Some(chunk) = res.as_ref() {
+        // if let Some((chunk, nb)) = res.as_ref() {
```
onursatici: nit: remove

onursatici (approved):

PR #6521, opened 2026-02-16
cuda flat layout

onursatici:

- Closes #.

Introduces a cuda flat layout that picks the constant array buffers and hoists them into the layout metadata, so reading the footer would give us the scalars directly on the cpu


joseph-isaacs (approved):

vortex-layout/src/strategy.rs
```diff
@@ -64,4 +64,24 @@ pub trait LayoutStrategy: 'static + Send + Sync {
         0
     }
 }
+
+#[async_trait]
+impl LayoutStrategy for std::sync::Arc<dyn LayoutStrategy> {
```
robert3005, line 69: we should have imported this here

onursatici: ah yea, missed that sorry

PR #6523, opened 2026-02-16
Move `vortex-scalar` into `vortex-array`

connortsui20:
Part of addressing #6500, as we need this as a prerequisite to unifying the the extension vtable we want to add.

After this, we still need to move `vortex-dtype` into `vortex-array`.

---

I had claude generate a script to do basically _all_ of the work. Some manual changes were done after review but quite small.

<details>

```sh
#!/usr/bin/env bash
set -euo pipefail

# Merge vortex-scalar into vortex-array, then delete vortex-scalar entirely.
#
# Run from the repo root:
#   bash scripts/merge-scalar-into-array.sh

# ─── Step 1: Move source files into vortex-array/src/scalar/ ────────────────
mkdir -p vortex-array/src/scalar
cp -r vortex-scalar/src/* vortex-array/src/scalar/
mv vortex-array/src/scalar/lib.rs vortex-array/src/scalar/mod.rs

# ─── Step 2: Strip crate-level lint attrs from scalar/mod.rs ────────────────
# These #![...] inner attributes are only valid at the crate root. Drop them.
sed -i '' '/^#!\[deny/d; /^#!\[warn/d' vortex-array/src/scalar/mod.rs

# ─── Step 3: Add `pub mod scalar;` to vortex-array/src/lib.rs ──────────────
sed -i '' 's/^pub mod search_sorted;$/pub mod scalar;\npub mod search_sorted;/' \
  vortex-array/src/lib.rs

# Also fix the doc comment that references vortex_scalar::Scalar.
sed -i '' 's/\[scalars\](vortex_scalar::Scalar)/[scalars](crate::scalar::Scalar)/' \
  vortex-array/src/lib.rs

# ─── Step 4: Fix imports in moved scalar files ─────────────────────────────
# 4a: crate:: → crate::scalar:: (all internal references, ~138 occurrences).
fd -e rs . vortex-array/src/scalar -x sed -i '' 's/crate::/crate::scalar::/g'

# 4b: vortex_scalar:: → vortex_array::scalar:: (doc examples, 3 occurrences).
fd -e rs . vortex-array/src/scalar -x sed -i '' 's/vortex_scalar::/vortex_array::scalar::/g'

# ─── Step 5: Fix imports in existing vortex-array files ─────────────────────
# vortex_scalar:: → crate::scalar:: (~229 occurrences in ~144 files).
# Exclude the scalar/ subdirectory (already handled above).
fd -e rs . vortex-array/src --exclude scalar -x sed -i '' 's/vortex_scalar::/crate::scalar::/g'

# ─── Step 6: Fix imports in all other crates ────────────────────────────────
# vortex_scalar:: → vortex_array::scalar:: across the entire workspace,
# excluding vortex-array (handled above) and vortex-scalar (being deleted).
fd -e rs . \
  --exclude vortex-array \
  --exclude vortex-scalar \
  -x sed -i '' 's/vortex_scalar::/vortex_array::scalar::/g'

# ─── Step 7: Update vortex-array/Cargo.toml ─────────────────────────────────
# 7a: Add bytes dependency (after async-lock, alphabetically).
sed -i '' '/^async-lock = { workspace = true }/a\
bytes = { workspace = true }\
' vortex-array/Cargo.toml

# 7b: Add "scalar" feature to vortex-proto.
sed -i '' 's/vortex-proto = { workspace = true, features = \["expr"\] }/vortex-proto = { workspace = true, features = ["expr", "scalar"] }/' \
  vortex-array/Cargo.toml

# 7c: Remove vortex-scalar dependency.
sed -i '' '/^vortex-scalar = /d' vortex-array/Cargo.toml

# 7d: Remove vortex-scalar/arbitrary from arbitrary feature list.
sed -i '' '/"vortex-scalar\/arbitrary",/d' vortex-array/Cargo.toml

# ─── Step 8: Remove vortex-scalar dep from all other Cargo.toml files ───────
fd -g Cargo.toml . \
  --exclude vortex-array \
  --exclude vortex-scalar \
  -x sed -i '' '/^vortex-scalar = /d'

# ─── Step 9: Update vortex/src/lib.rs ───────────────────────────────────────
# Point the scalar re-export at vortex-array instead of vortex-scalar.
sed -i '' 's/pub use vortex_scalar::\*;/pub use vortex_array::scalar::*;/' \
  vortex/src/lib.rs

# ─── Step 10: Remove vortex-scalar from workspace root Cargo.toml ───────────
# Remove from members list.
sed -i '' '/"vortex-scalar",/d' Cargo.toml

# Remove workspace dependency definition.
sed -i '' '/^vortex-scalar = .*path = /d' Cargo.toml

# ─── Step 11: Delete the vortex-scalar crate entirely ───────────────────────
rm -rf vortex-scalar

# ─── Step 12: Format files ──────────────────────────────────────────────────
cargo +nightly fmt --all
taplo fmt

# ─── Step 13: Update public API lockfile ────────────────────────────────────
bash ./scripts/public-api.sh
cargo update --manifest-path java/testfiles/Cargo.toml

echo ""
echo "Done! Run these to verify:"
echo "  cargo +nightly fmt --all"
echo "  taplo fmt"
echo "  cargo clippy --all-targets --all-features"
echo "  cargo nextest run --all-features --no-fail-fast"
```

</details>


encodings/decimal-byte-parts/public-api.lock
```diff
@@ -126,7 +126,7 @@ pub fn vortex_decimal_byte_parts::DecimalBytePartsVTable::stats(array: &vortex_d
 
 impl vortex_array::vtable::operations::OperationsVTable<vortex_decimal_byte_parts::DecimalBytePartsVTable> for vortex_decimal_byte_parts::DecimalBytePartsVTable
 
-pub fn vortex_decimal_byte_parts::DecimalBytePartsVTable::scalar_at(array: &vortex_decimal_byte_parts::DecimalBytePartsArray, index: usize) -> vortex_error::VortexResult<vortex_scalar::scalar::Scalar>
+pub fn vortex_decimal_byte_parts::DecimalBytePartsVTable::scalar_at(array: &vortex_decimal_byte_parts::DecimalBytePartsArray, index: usize) -> vortex_error::VortexResult<vortex_array::scalar::scalar::Scalar>
```
robert3005: there's something bad here. Why is the type in `vortex_array::scalar::scalar::Scalar`?

vortex-scalar/src/lib.rs
```diff
@@ -9,12 +9,6 @@
 //!
 //! Note that the implementations of `Scalar` are split into several different modules.
 
-#![deny(missing_docs)]
-#![warn(clippy::missing_docs_in_private_items)]
-#![warn(clippy::missing_errors_doc)]
-#![warn(clippy::missing_panics_doc)]
-#![warn(clippy::missing_safety_doc)]
```
robert3005, line 16: Hmmm, this is annoying

connortsui20: I had that in there since I changed the entire crate, but now that it's no longer small I just removed it

encodings/decimal-byte-parts/public-api.lock
connortsui20: My understanding of the lockfiles is that they do the absolute path, not re-exported paths. So because we re-export `vortex-scalar` via `vortex_array::scalar`, we get this.

I could remove all public modules in the old `vortex_scalar` and make sure we _only_ export directly from `vortex_array::scalar`?

vortex-array/src/scalar/mod.rs
```diff
@@ -26,6 +20,11 @@ mod display;
 mod downcast;
 mod proto;
 
+#[expect(
+    clippy::module_inception,
+    reason = "This is a remnant of a past time where we had a dedicated `vortex-scalar` crate. \
+              See #6500 for more details on why we did this."
+)]
```
robert3005: can we somehow avoid this? Maybe inline scalar file into this one?

encodings/decimal-byte-parts/public-api.lock
robert3005: I think you want to do this. Module inception is weird to have to suppress

robert3005 (approved):

PR #6527, opened 2026-02-16
chore: replace macros with fns

0ax1:

Replaces macros used in CUDA code with constexpr functions.

### DIFF
--- a/vortex-cuda/kernels/src/config.cuh
+++ b/vortex-cuda/kernels/src/config.cuh
@@ -13,7 +13,13 @@
 //   grid_dim = ceil(array_len / 2048)
 constexpr uint32_t ELEMENTS_PER_THREAD = 32;
 
-#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+// We use `::min` from CUDA's `crt/math_functions.hpp` (declared `__host__ __device__`)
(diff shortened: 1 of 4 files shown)


vortex-cuda/kernels/src/config.cuh
```diff
@@ -13,7 +13,13 @@
 //   grid_dim = ceil(array_len / 2048)
 constexpr uint32_t ELEMENTS_PER_THREAD = 32;
 
-#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+// We use `::min` (CUDA device built-in) rather than
+// `std::min` as `std::min` is a host-only function.
 
-#define START_ELEM(idx, len) MIN((idx) * ELEMENTS_PER_THREAD, (len))
-#define STOP_ELEM(idx, len)  MIN(START_ELEM(idx, len) + ELEMENTS_PER_THREAD, (len))
+__device__ constexpr uint32_t start_elem(uint32_t idx, uint32_t len) {
```
a10y: should this be u64? or templates?

i think this is getting implicit cast down at usage sites

0ax1: Yeah this should just be T

robert3005 (approved):

PR #6528, opened 2026-02-16
break[array]: lazy between

joseph-isaacs:
Part of the lazy compute migration: https://github.com/vortex-data/vortex/discussions/5921.

## breaks
 - lazy between compute 
 - deprecate the compute/between method, use the `ArrayBuiltIns::between`


joseph-isaacs:
None of the codspeed regressions worry me, I think we need to do a pass of the execution system and try to optimize

robert3005 (approved):

PR #6529, opened 2026-02-16
Remove `vortex-vector`

connortsui20:
😢

We need to remove this now because we want to merge `vortex-dtype` into `vortex-array`, and if this still existed it would create a dependency cycle.

```
vortex-array --(depends on)--> vortex-vector --(depends on)--> vortex-dtype

vortex-array <--(depends on)--> vortex-dtype
```

😭


robert3005 (approved):

a10y (approved):
destroy it

connortsui20:
I should point out that I still think it is worth keeping this in our back pocket in case we want to replace our builders with these vectors.

With `vortex-scalar` being merged into `vortex-array` soon as well, we could actually have the completely type safe builders from vectors here!

connortsui20:
_A moment of silence for `vortex-vector` (2025-2026)_

PR #6530, opened 2026-02-16
perf[vortex-btrblocks]: remove listview expansion ratio

asubiotto:
The initial thinking here was to calculate the "cost" of expanding the listview elements when converting to a list. Specifically, we would choose listview when the elements expand to > 150%. However, we didn't give much importance to the fact that price of *not* converting to list is just writing the sizes (32KB per 8192 rows).

In practice, we really want to understand not the *count* of expanded elements, but their byte sizes. As long they expand to <32KB (if 8192 rows), list should be preferred in an ideal world. So no fixed threshold is correct. It's too high for long lists with complex types and too low for short lists with small elements.

Additionally, I ran into this problem running listview compression on a real-world workflow and the cost of non-zctl listview -> list is extremely high.

Given all this, I think the pragmatic way forward is to simplify the list vs listview decision and only write list if the listview is zctl.

No

See commit message/PR description

Removing an inexact calculation that could cause costly side-effects

## How is this change tested?
Existing tests

WRT to the latest release, this change will write ListView to disk when List was previously written. WRT to the previous commit, ListView will be written *more* often.

### DIFF
--- a/vortex-btrblocks/Cargo.toml
+++ b/vortex-btrblocks/Cargo.toml
@@ -62,6 +62,11 @@ harness = false
 test = false
 
 [[bench]]
+name = "compress_listview"
+harness = false
+test = false
+
+[[bench]]
 name = "dict_encode"
 harness = false
 test = false
--- a/vortex-btrblocks/benches/compress_listview.rs
+++ b/vortex-btrblocks/benches/compress_listview.rs
@@ -1,0 +1,190 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#![allow(clippy::unwrap_used)]
+#![allow(clippy::cast_possible_truncation)]
+#![allow(unexpected_cfgs)]
+
+#[cfg(not(codspeed))]
+mod benchmarks {
+    use divan::Bencher;
+    use divan::counter::BytesCount;
+    use divan::counter::ItemsCount;
+    use rand::Rng;
+    use rand::SeedableRng;
+    use rand::prelude::StdRng;
+    use vortex_array::ArrayRef;
+    use vortex_array::IntoArray;
+    use vortex_array::arrays::ListViewArray;
+    use vortex_array::arrays::StructArray;
+    use vortex_array::arrays::VarBinViewArray;
+    use vortex_array::validity::Validity;
+    use vortex_btrblocks::BtrBlocksCompressor;
+    use vortex_buffer::buffer_mut;
+    use vortex_dtype::FieldNames;
+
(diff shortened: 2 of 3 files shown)


joseph-isaacs (approved):
This seems okay, maybe we could follow up with a benchmark in crate comparing this?

joseph-isaacs:
I think we should make this pluggable so not PR are needed here too

asubiotto:
Yes, happy to add a benchmark before merging

asubiotto:
Done. I added a benchmark with a relatively complex nested schema and some overlap. Before:
```
Timer precision: 41 ns
compress_listview        fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ benchmarks                          │               │               │               │         │
   ╰─ compress_listview                │               │               │               │         │
      ├─ Overlapping     328.5 ms      │ 446.5 ms      │ 346.1 ms      │ 350.6 ms      │ 100     │ 100
      │                  37.57 MB/s    │ 27.64 MB/s    │ 35.66 MB/s    │ 35.2 MB/s     │         │
      │                  24.93 Kitem/s │ 18.34 Kitem/s │ 23.66 Kitem/s │ 23.36 Kitem/s │         │
      ╰─ Zctl            11.34 ms      │ 15.46 ms      │ 11.87 ms      │ 12.06 ms      │ 100     │ 100
                         1.087 GB/s    │ 798.2 MB/s    │ 1.039 GB/s    │ 1.022 GB/s    │         │
                         721.8 Kitem/s │ 529.7 Kitem/s │ 689.6 Kitem/s │ 678.7 Kitem/s │         │

```
After:
```
Timer precision: 41 ns
compress_listview        fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ benchmarks                          │               │               │               │         │
   ╰─ compress_listview                │               │               │               │         │
      ├─ Overlapping     11.86 ms      │ 16.57 ms      │ 12.6 ms       │ 12.64 ms      │ 100     │ 100
      │                  1.04 GB/s     │ 744.8 MB/s    │ 979.4 MB/s    │ 976.6 MB/s    │         │
      │                  690.5 Kitem/s │ 494.2 Kitem/s │ 649.9 Kitem/s │ 648 Kitem/s   │         │
      ╰─ Zctl            11.31 ms      │ 13.58 ms      │ 11.94 ms      │ 11.96 ms      │ 100     │ 100
                         1.09 GB/s     │ 908.4 MB/s    │ 1.033 GB/s    │ 1.031 GB/s    │         │
                         723.8 Kitem/s │ 602.8 Kitem/s │ 685.8 Kitem/s │ 684.8 Kitem/s │         │
```

PR #6532, opened 2026-02-16
pvalue::cast returns VortexResult

myrrc:

No

- PValue::cast returns VortexResult instead of Option.
- find_intersection* and a couple of other functions also return VortexResult instead of Option.

First issue

## How is this change tested?

Existing tests

Public API method signature changed

### DIFF
--- a/encodings/sequence/src/array.rs
+++ b/encodings/sequence/src/array.rs
@@ -133,7 +133,9 @@ impl SequenceArray {
 
         // A sequence A[i] = base + i * multiplier is sorted iff multiplier >= 0,
         // and strictly sorted iff multiplier > 0.
-        let m_int = multiplier.cast::<i64>();
+        let m_int = multiplier
+            .cast::<i64>()
+            .vortex_expect("must be able to cast");
         let is_sorted = m_int >= 0;
         let is_strict_sorted = m_int > 0;
 
@@ -179,9 +181,8 @@ impl SequenceArray {
             let len_t = <P>::from_usize(length - 1)
                 .ok_or_else(|| vortex_err!("cannot convert length {} into {}", length, ptype))?;
 
-            let base = base.cast::<P>();
-            let multiplier = multiplier.cast::<P>();
-
+            let base = base.cast::<P>()?;
+            let multiplier = multiplier.cast::<P>()?;
             let last = len_t
                 .checked_mul(multiplier)
                 .and_then(|offset| offset.checked_add(base))
@@ -194,8 +195,11 @@ impl SequenceArray {
         assert!(idx < self.len, "index_value({idx}): index out of bounds");
 
         match_each_native_ptype!(self.ptype(), |P| {
-            let base = self.base.cast::<P>();
-            let multiplier = self.multiplier.cast::<P>();
+            let base = self.base.cast::<P>().vortex_expect("must be able to cast");
+            let multiplier = self
+                .multiplier
+                .cast::<P>()
+                .vortex_expect("must be able to cast");
             let value = base + (multiplier * <P>::from_usize(idx).vortex_expect("must fit"));
 
             PValue::from(value)
@@ -310,8 +314,8 @@ impl VTable for SequenceVTable {
 
     fn execute(array: &Self::Array, _ctx: &mut ExecutionCtx) -> VortexResult<ArrayRef> {
         let prim = match_each_native_ptype!(array.ptype(), |P| {
-            let base = array.base().cast::<P>();
-            let multiplier = array.multiplier().cast::<P>();
+            let base = array.base().cast::<P>()?;
+            let multiplier = array.multiplier().cast::<P>()?;
             let values = BufferMut::from_iter(
                 (0..array.len())
                     .map(|i| base + <P>::from_usize(i).vortex_expect("must fit") * multiplier),
--- a/encodings/sequence/src/compute/compare.rs
+++ b/encodings/sequence/src/compute/compare.rs
@@ -14,6 +14,8 @@ use vortex_dtype::Nullability;
 use vortex_dtype::match_each_integer_ptype;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
+use vortex_error::vortex_bail;
+use vortex_error::vortex_err;
 use vortex_scalar::PValue;
 use vortex_scalar::Scalar;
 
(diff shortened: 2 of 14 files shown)


vortex-array/src/builders/primitive.rs
```diff
@@ -155,9 +156,13 @@ impl<T: NativePType> ArrayBuilder for PrimitiveBuilder<T> {
             scalar.dtype()
         );
 
-        match scalar.as_primitive().pvalue() {
-            Some(pv) => self.append_value(pv.cast::<T>()),
-            None => self.append_null(),
+        if let Some(pv) = scalar.as_primitive().pvalue() {
+            let pv = pv.cast_opt::<T>().ok_or_else(|| {
+                vortex_err!("cannot cast {} to {}", pv, std::any::type_name::<T>())
```
connortsui20: Instead of the type name you could also do `pv.ptype()` which gives us back a `PType`, which should implement display

vortex-cuda/src/kernel/encodings/sequence.rs
```diff
@@ -48,9 +48,12 @@ impl CudaExecute for SequenceExecutor {
         } = array.into_parts();
 
         match_each_native_ptype!(ptype, |P| {
-            let base = base.cast::<P>();
-            let multiplier = multiplier.cast::<P>();
-
+            let base = base
+                .cast_opt::<P>()
+                .ok_or_else(|| vortex_err!("cannot cast {} to {}", base, ptype))?;
```
connortsui20: you can inline these I think?

connortsui20 (commented):
LGTM! just some nits

myrrc:
> you could also do pv.ptype() 

Nice, thanks

> you can inline these I think

But then we're back to cast-like function?

vortex-cuda/src/kernel/encodings/sequence.rs
connortsui20: Oh sorry I meant:

```
.ok_or_else(|| vortex_err!("cannot cast {base} to {ptype}"))?;
```

connortsui20:
Following offline discussion with @joseph-isaacs I think we actually want this to return a `VortexResult`, and at that point we can just rename to `cast -> VortexResult<T>`

myrrc:
Ok, makes sense, I'll update the PR.

encodings/sequence/src/array.rs
```diff
@@ -310,8 +314,14 @@ impl VTable for SequenceVTable {
...
-            let base = array.base().cast::<P>();
-            let multiplier = array.multiplier().cast::<P>();
+            let base = array
+                .base()
+                .cast::<P>()
+                .vortex_expect("must be able to cast");
+            let multiplier = array
+                .multiplier()
+                .cast::<P>()
+                .vortex_expect("must be able to cast");
```
connortsui20: These should use `?` since the caller returns a result

encodings/sequence/src/kernel.rs
```diff
@@ -217,13 +217,10 @@ fn find_transition_point(
...
-            base.cast::<P>(),
-            multiplier.cast::<P>(),
-            len,
-            constant.cast::<P>(),
-            operator,
-        )
+        let base = base.cast::<P>().vortex_expect("must be able to cast");
+        let multiplier = multiplier.cast::<P>().vortex_expect("must be able to cast");
+        let constant = constant.cast::<P>().vortex_expect("must be able to cast");
+        find_transition_point_typed::<P>(base, multiplier, len, constant, operator)
```
connortsui20: This whole function should probably just return a `VortexResult` (since the only caller also returns a `VortexResult`)

vortex-scalar/src/typed_view/primitive/pvalue.rs
```diff
@@ -221,7 +214,9 @@ impl PValue {
             PValue::F16(f) => T::from_f16(f),
             PValue::F32(f) => T::from_f32(f),
             PValue::F64(f) => T::from_f64(f),
-        }
+        };
+        let ptype = self.ptype();
+        res.ok_or_else(|| vortex_err!("cannot cast {self} to {ptype}"))
```
connortsui20: Very very nitpicky but perhaps do:

```rust
    pub fn cast<T: NativePType>(&self) -> VortexResult<T> {
        match *self {
            PValue::U8(u) => T::from_u8(u),
            PValue::U16(u) => T::from_u16(u),
            PValue::U32(u) => T::from_u32(u),
            PValue::U64(u) => T::from_u64(u),
            PValue::I8(i) => T::from_i8(i),
            PValue::I16(i) => T::from_i16(i),
            PValue::I32(i) => T::from_i32(i),
            PValue::I64(i) => T::from_i64(i),
            PValue::F16(f) => T::from_f16(f),
            PValue::F32(f) => T::from_f32(f),
            PValue::F64(f) => T::from_f64(f),
        }
        .ok_or_else(|| vortex_err!("cannot cast {self} to {}", T::PTYPE))
    }
```

Also I just realized that `self.ptype` was incorrect since that is the type of self, not the type that we are casting to.

connortsui20:
Can you update the `cast` function to my suggestion above? It is generally helpful to know what the exact value is that failed to be casted since there might be many different casts happening in 1 benchmark run that could fail. Otherwise this looks good to me!

vortex-scalar/src/typed_view/primitive/pvalue.rs
```diff
@@ -361,10 +355,12 @@ impl TryFrom<PValue> for usize {
...
-        value
-            .cast_opt::<u64>()
-            .and_then(|v| v.to_usize())
-            .ok_or_else(|| vortex_err!("Cannot read primitive value {:?} as usize", value))
+        match value.cast::<u64>() {
+            Ok(v) => v
+                .to_usize()
+                .ok_or_else(|| vortex_err!("Cannot read primitive value {:?} as usize", value)),
+            Err(e) => Err(e),
+        }
```
connortsui20:
```suggestion
        value
            .cast::<u64>()?
            .to_usize()
            .ok_or_else(|| vortex_err!("Cannot read primitive value {:?} as usize", value))
```

encodings/sequence/src/compute/compare.rs
```diff
@@ -77,9 +77,9 @@ pub(crate) fn find_intersection_scalar(
     intercept: PValue,
 ) -> Option<usize> {
     match_each_integer_ptype!(base.ptype(), |P| {
-        let intercept = intercept.cast::<P>();
-        let base = base.cast::<P>();
-        let multiplier = multiplier.cast::<P>();
+        let intercept = intercept.cast::<P>().ok()?;
+        let base = base.cast::<P>().ok()?;
+        let multiplier = multiplier.cast::<P>().ok()?;
```
joseph-isaacs: This should pass up a Result

connortsui20 (approved):
LGTM!

PR #6549, opened 2026-02-17
Fix e2e test result with only stable encodings

AdamGS:
Just a small fix, seems like the value with no features was wrong, at least when I run it locally.

### DIFF
--- a/vortex-test/e2e/src/lib.rs
+++ b/vortex-test/e2e/src/lib.rs
@@ -47,7 +47,10 @@ mod tests {
 
         let sizes = futures::future::try_join_all(futures).await?;
         for (i, size) in sizes.iter().enumerate() {
-            assert_eq!(*size, EXPECTED_SIZE, "Run {i} produced size {size}");
+            assert_eq!(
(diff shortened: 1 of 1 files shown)


vortex-test/e2e/src/lib.rs
```diff
@@ -24,11 +24,7 @@ mod tests {
         let array =
             PrimitiveArray::new(Buffer::from_iter(values), Validity::NonNullable).into_array();
 
-        // Write in parallel and verify all sizes match expected
-        #[cfg(feature = "unstable_encodings")]
-        const EXPECTED_SIZE: usize = 216188;
-        #[cfg(not(feature = "unstable_encodings"))]
-        const EXPECTED_SIZE: usize = 216188;
+        // Write concurrently and verify all sizes match expected
```
AdamGS, line 27: this is all running in a single single threaded tokio runtime and then `try_join_all`, we can take this further but I think just clarifying the comment here is enough.

onursatici (approved):

joseph-isaacs (approved):

PR #6553, opened 2026-02-17
chore[ci]: explicitly use stable or night for all cargo commands

joseph-isaacs:
We want to use stable for all ci checks unless nightly is required. This PR does this


.github/workflows/ci.yml
```diff
@@ -284,18 +284,19 @@ jobs:
...
         with:
-          toolchain: ${{ env.NIGHTLY_TOOLCHAIN }}
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install nightly for fmt
+        run: rustup toolchain install $NIGHTLY_TOOLCHAIN --component rustfmt
       - name: Rust Lint - Format
         run: cargo +$NIGHTLY_TOOLCHAIN fmt --all --check
       - name: Rustc check
-        run: RUSTFLAGS="-D warnings" cargo check --locked --all-features --all-targets
+        run: RUSTFLAGS="-D warnings" cargo +stable check --locked --all-features --all-targets
```
0ax1: I like this being explicit _but_, I'd assume explicitly defining what toolchain to use would override what we define in the `rust-toolchain.toml?`

.github/workflows/publish.yml
```diff
@@ -39,7 +36,7 @@ jobs:
 
       - name: Release
         run: |
-          cargo +$NIGHTLY_TOOLCHAIN publish -Zpublish-timeout --no-verify --allow-dirty --workspace
```
0ax1, line 44: 🤦

0ax1: Or do we need nightly to publish @AdamGS ?

AdamGS: I think we do? let's wait for @a10y to weigh in

robert3005: We need rust 1.90.0

robert3005: I completely missed publish-timeout. This one is not in any stable release. While 1.90 stabilized `cargo publish --workspace` we will still need nightly here

0ax1 (approved):

robert3005 (changes requested):
Requesting changes so we don't break publish

robert3005 (approved):

PR #6555, opened 2026-02-17
fix: duckdb object store read coalescing

0ax1:
## Summary

Unifies coalescing configs for object store reads. In particular, this includes the coalescing config used for DuckDB.


AdamGS (approved):

vortex-duckdb/src/duckdb/file_system.rs
```diff
@@ -124,11 +114,11 @@ impl VortexReadAt for DuckDbFsReader {
     }
 
     fn coalesce_config(&self) -> Option<CoalesceConfig> {
-        Some(DEFAULT_COALESCE)
+        Some(CoalesceConfig::object_storage())
```
0ax1: We need to distinguish nvme and object store here.

Issue #6547, opened 2026-02-16
Tracking Issue: Refactors for Extension DTypes, Scalars, and Arrays


a10y:
### Discussed in https://github.com/vortex-data/vortex/discussions/6500

<div type='discussions-op-text'>

<sup>Originally posted by **connortsui20** February 13, 2026</sup>
# More Robust Extension Data Types In Vortex

We would like to build a more robust system for extension data types (or `DType`s).

https://github.com/vortex-data/vortex/pull/6081 introduced vtables for extension DTypes. Each extension type (e.g. `Timestamp`) now implements `ExtDTypeVTable`, which handles validation, serialization, and metadata. The type-erased `ExtDTypeRef` carries this vtable with it inside `DType::Extension`.

The natural next steps would be to add analogous vtables for `vortex-scalar` (e.g., custom `Display` and casting) and `vortex-array` (custom compute kernels). This would give us three traits:

```
ExtDTypeVTable   (vortex-dtype)
ExtScalarVTable  (vortex-scalar)
ExtArrayVTable   (vortex-array)
```

## Issues

There is a problem with this. This has come up a few times in the effort to add the scalar extension vtable to scalar extension values: https://github.com/vortex-data/vortex/pull/6477.

Once an `ExtDType` is type-erased to `ExtDTypeRef`, the only thing it carries is the dtype vtable (ONLY the dtype, not the scalar or array vtables). Suppose you have an `ExtensionArray` and want to call scalar or array logic: you then need to look up the other vtables by `ExtID` in a session registry. This means threading `&VortexSession` through every code path that touches extension types (this would be things like compute kernels, builders, canonicalization, display, etc.)

This is kind of torturous because the dtype vtable is literally _right there_ inside the `DType`, but the scalar/array vtables require a registry lookup to find. If the vtables were combined we would not have this issue. So to fix this, sessions need to be plumbed through APIs that otherwise have no reason to take one (in other words, many constructors would need to take a session if they could _potentially_ create an extension type).

What we probably want is a single `ExtVTable` per extension dtype that covers all three layers, so that when you have an `ExtDTypeRef` you already have everything you need.

## Crate Dependency Graph!

The crate dependency graph is:

```
vortex-array --(depends on)--> vortex-scalar --(depends on)--> vortex-dtype
```

A unified vtable trait would need to reference types from all three crates, which is impossible when the trait lives in `vortex-dtype`, which can't depend on `vortex-scalar` or `vortex-array`.

---

[...]

PR #6557, opened 2026-02-17
Remove `vortex-dtype` dependency from `vortex-compute`

connortsui20:
## Summary

Tracking Issue: https://github.com/vortex-data/vortex/issues/6547

Removes all instances of `vortex-dtype` imports for `vortex-compute`. This was not super hard since all of the existing imports from `vortex-dtype` are really just re-exports of other third-party crates.

Technically there are API changes but these are all generalizations / now the trait bounds on some of the compute is _less_ restrictive than before. So I wouldn't really consider this a breaking change.

The one concern would the SIMD code that was already commented out / not used before. At some point we want to do that (see my failed attempt here: https://github.com/vortex-data/vortex/pull/5722) in general, but I guess today is not that day :(

## Testing

N/A


robert3005:
At this point all of vortex-compute is dead code. Should we just delete this? Git always has history

robert3005:
ok, no, filter is being used. Should we move Filter into the vortex-array and delete the rest?

connortsui20:
We really should be using more from `vortex-compute`. For example there is a better `take_scalar` in there that we aren't using. That's the most obvious one, but there are other things that we were able to microbenchmark more easily in there

vortex-compute/src/comparison/mod.rs
```diff
@@ -3,8 +3,8 @@
 
 //! Comparison operations for Vortex vectors.
 
-use vortex_dtype::half::f16;
-use vortex_dtype::i256;
+use arrow_buffer::i256;
+use half::f16;
```
a10y: i think we want to keep the wrapper types

connortsui20: Should I just remove this comparison implementation for this then? We don't actually use it anywhere, and otherwise it would create a circular dependency when we merge `vortex-dtype` into `vortex-array`

connortsui20:
Ok arguably we should completely delete this crate, but I believe until we can say that all of the optimized code we have has been migrated to `vortex-array` properly we should keep it around.

vortex-compute/src/comparison/mod.rs
connortsui20: Yeah I'm just going to remove it

a10y: maybe we should keep the traits here but keep the impls in vortex-array? do we ever use vortex-compute directly instead of going thru vortex-array?

a10y: or we can keep impls for primitive types here but we need to keep the impls for f16/i256 in vortex-array? idk, i'm also in the camp of not sure why we keep this crate around at all

robert3005:
Was there a reason why we didn't use this code earlier? Git has the history even if you delete the code

vortex-compute/src/comparison/mod.rs
connortsui20: as of right now, only `vortex-array` uses `vortex-compute`. I think the original idea was that we wanted to have something other stuff (vectors) could call into but now there's probably no reason to keep it.

connortsui20:
This was when we thought that we were moving towards vectors instead of arrays, so we didn't bother migrating the code to `vortex-array` since `vortex-vector` was already using `vortex-compute`.

connortsui20:
Discussed offline: we're just going to remove this and I'll create an issue to bring some of these optimizations back.

PR #6560, opened 2026-02-17
fix is_sorted check failing for large multiplier values

a10y:
Fix for the issue found by fuzzer in [latest run](https://github.com/vortex-data/vortex/actions/runs/22098142659/job/63860475812)

image

Seems like this snuck in on https://github.com/vortex-data/vortex/pull/6452

### DIFF
--- a/encodings/sequence/src/array.rs
+++ b/encodings/sequence/src/array.rs
@@ -17,6 +17,7 @@ use vortex_array::arrays::PrimitiveArray;
 use vortex_array::buffer::BufferHandle;
 use vortex_array::expr::stats::Precision as StatPrecision;
 use vortex_array::expr::stats::Stat;
+use vortex_array::match_each_pvalue;
 use vortex_array::scalar::PValue;
 use vortex_array::scalar::Scalar;
 use vortex_array::scalar::ScalarValue;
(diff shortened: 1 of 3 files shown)


vortex-array/src/scalar/typed_view/primitive/pvalue.rs
```diff
@@ -23,6 +23,68 @@ use vortex_error::vortex_bail;
...
+macro_rules! match_each_pvalue {
+    (
+        $this:expr,uint: |
+        $vuint:ident |
+        $buint:block,int: |
+        $vint:ident |
+        $bint:block,float: |
+        $vfloat:ident |
+        $bfloat:block
+    ) => {{
```
a10y, line 38: i hate that this is how rustfmt formats all of these higher-order macros but dunno how to fix it.

encodings/sequence/src/array.rs
```diff
@@ -133,11 +134,13 @@ impl SequenceArray {
...
-            .cast::<i64>()
-            .vortex_expect("must be able to cast");
-        let is_sorted = m_int >= 0;
-        let is_strict_sorted = m_int > 0;
+
+        let (is_sorted, is_strict_sorted) = match_each_pvalue!(
+            multiplier,
+            uint: |v| { (true, v> 0) },
+            int: |v| { (v >= 0, v > 0) },
+            float: |v| { (!v.is_sign_negative(), v.is_sign_positive()) }
```
a10y: we probably shouldn't even allow SequenceArray of floats should we

robert3005: do we check that?

robert3005 (approved):

dimitarvdimitrov (approved):
LGTM, IIRC @joseph-isaacs and I went with `cast` instead of the macro because a macro is more expensive (to compile?). But if that's causing correctness problems, then the macro it is

can't tell about the sequence array of floats

encodings/sequence/src/array.rs
joseph-isaacs: I really thought it was banned in the ctor?

joseph-isaacs (approved):

joseph-isaacs:
I would just use a i128 cast?

encodings/sequence/src/array.rs
a10y: i think it is. should i just put `unreachable!()` here then?

a10y: apparently we do

a10y:
i don't think the macro is more expensive, the `cast` pathway also expands the enum variants so I think it shakes out to the same thing?

PR #6561, opened 2026-02-17
Remove `vortex-compute`

connortsui20:
## Summary

Tracking Issue: https://github.com/vortex-data/vortex/issues/6547

Removes `vortex-compute` and migrates some of the code that was used / should have been used from `vortex-compute` in `vortex-array` inline.

This is part of the effort to merge `vortex-dtype` into `vortex-array`. `vortex-compute` was just in the way because it depends on the `vortex-dtype` crate but `vortex-array` depends on `vortex-compute`.

## API Changes

Removes all of `vortex-compute`.

## Testing

N/A


vortex-mask/public-api.lock
```diff
@@ -124,10 +124,6 @@ impl core::convert::From<vortex_buffer::bit::buf::BitBuffer> for vortex_mask::Ma
 
 pub fn vortex_mask::Mask::from(value: vortex_buffer::bit::buf::BitBuffer) -> Self
 
-impl core::convert::From<vortex_mask::Mask> for core::option::Option<arrow_buffer::buffer::null::NullBuffer>
-
-pub fn core::option::Option<arrow_buffer::buffer::null::NullBuffer>::from(value: vortex_mask::Mask) -> Self
-
```
connortsui20, line 130: Don't really know why this was changed...

vortex-array/src/arrays/filter/execute/primitive.rs
```diff
@@ -18,9 +17,7 @@ pub fn filter_primitive(array: &PrimitiveArray, mask: &Arc<MaskValues>) -> Primi
     let filtered_validity = filter_validity(validity, mask);
 
     match_each_native_ptype!(array.ptype(), |T| {
-        // TODO(connor): Ideally we should filter directly on the `BufferHandle` once we implement
-        // `Filter` for that.
-        let filtered_buffer = array.to_buffer::<T>().filter(mask.as_ref());
+        let filtered_buffer = super::buffer::filter_buffer(array.to_buffer::<T>(), mask.as_ref());
```
robert3005: I think you still want to import this

connortsui20: import as in move the path to the top of the file?

a10y (commented):
seems good to me, i'll let someone else give final +1 but it's mainly moving stuff around, if it passes existing test suite i'm fine with it

robert3005 (approved):

PR #6564, opened 2026-02-17
Add FileSystem trait and make DuckDB configurable

gatesn:
We switched over to using DuckDB's own file system for I/O. For some reason we have really binary performance benchmarks, either the same performance as before, or really really slow. We are going to investigate, but in the meantime this change abstracts the I/O behind a file system and allows us to configure whether to use Vortex I/O or DuckDB I/O. This abstraction was planned and built in #6391 anyway, so just making use of it a little sooner.

As part of this change, we have a `FileSystem::list -> Stream` API. We used to use the async-compat crate to provide Tokio support within DuckDB, but that crate doesn't support streams. This PR inlines the core piece of logic from that crate and provides wrappers for I/O.

To summarise:

1. Pulls out the FileSystem trait as used in #6391 
1. Uses the trait to make fs configurable in DuckDB. Defaults to Vortex I/O as per before #5767 , with extension option to enable DuckDB native file system.
1. Clean up of the vortex-io crate for consistent structure + naming, as well as providing a "compat" module with compatibility wrappers.


vortex-duckdb/cpp/include/duckdb_vx/client_context.h
```diff
@@ -9,17 +9,14 @@
 extern "C" {
 #endif
 
-typedef struct duckdb_vx_client_context_ *duckdb_vx_client_context;
```
gatesn, line 12: The C API now exposes client_context, so we use DuckDB's instead of our own wrapper

vortex-duckdb/cpp/include/duckdb_vx/copy_function.h
```diff
@@ -72,8 +72,8 @@ typedef struct {
 duckdb_vx_copy_func_vtab_t *get_vtab_one();
 
 // A single function for configuring the DuckDB table function vtable.
-duckdb_state duckdb_vx_copy_func_register_vtab_one(duckdb_connection ffi_conn);
+duckdb_state duckdb_vx_copy_func_register_vtab_one(duckdb_database ffi_db);
```
gatesn, line 75: We now register against a database, rather than a specific connection (from which we extracted the DB anyways...)

vortex-duckdb/src/copy.rs
```diff
@@ -129,7 +129,7 @@ impl CopyFunction for VortexCopyFunction {
         let array_stream = ArrayStreamAdapter::new(bind_data.dtype.clone(), rx.into_stream());
 
         let handle = SESSION.handle();
-        let ctx_ptr = client_context.as_sendable();
+        let ctx_ptr = client_context;
```
AdamGS: seems redundant?

vortex-duckdb/src/duckdb/client_context.rs
```diff
@@ -13,27 +13,32 @@ use crate::wrapper;
...
+    cpp::duckdb_client_context,
+    |_| {
+        // No cleanup is necessary since the client context is owned by the connection and will
+        // be valid for the connection's lifetime.
+    }
 );
 
-#[derive(Clone, Copy)]
-pub(crate) struct SendableClientContext(cpp::duckdb_vx_client_context);
 // SAFETY: SendableClientContext carries the same opaque pointer as ClientContext. It is safe to
```
0ax1: Gotta adjust this comment as `SendableClientContext` does not exist anymore.

vortex-duckdb/cpp/file_system.cpp
```diff
@@ -132,51 +131,27 @@ extern "C" duckdb_state duckdb_vx_fs_write(duckdb_vx_file_handle handle,
...
-    }
-}
 
-extern "C" void duckdb_vx_uri_list_free(duckdb_vx_uri_list *list) {
-    if (!list || !list->entries) {
-        return;
-    }
-    for (size_t i = 0; i < list->count; i++) {
-        duckdb_free(const_cast<char *>(list->entries[i]));
+        fs.ListFiles(directory,
```
0ax1, line 148: 👍

vortex-duckdb/src/filesystem.rs
```diff
@@ -0,0 +1,327 @@
...
+use crate::duckdb::FsFileHandle;
+use crate::duckdb::duckdb_fs_list_dir;
+use crate::duckdb::fs_error;
+
+pub(super) fn resolve_filesystem(
+    base_url: &Url,
+    ctx: &ClientContext,
+) -> VortexResult<FileSystemRef> {
+    let fs_config = ctx
+        .try_get_current_setting(c"vortex_filesystem")
```
0ax1, line 49: Voting against making this an option. There should be one fs the extension uses which should be fast. Otherwise, we should include both fs variants in our benchmarks.

vortex-duckdb/src/lib.rs
```diff
@@ -45,10 +45,16 @@ static SESSION: LazyLock<VortexSession> =
 /// Initialize the Vortex extension by registering the extension functions.
 /// Note: This also registers extension options. If you want to register options
 /// separately (e.g., before creating connections), call `register_extension_options` first.
-pub fn register_table_functions(conn: &Connection) -> VortexResult<()> {
-    conn.register_table_function::<VortexTableFunction>(c"vortex_scan")?;
-    conn.register_table_function::<VortexTableFunction>(c"read_vortex")?;
-    conn.register_copy_function::<VortexCopyFunction>(c"vortex", c"vortex")
+pub fn initialize(db: &Database) -> VortexResult<()> {
+    db.config().add_extension_options(
+        "vortex_filesystem",
```
0ax1, line 50: See https://github.com/vortex-data/vortex/pull/6564/changes#r2821619770

vortex-duckdb/src/scan.rs
```diff
@@ -248,10 +234,41 @@ impl TableFunction for VortexTableFunction {
...
             .ok_or_else(|| vortex_err!("Missing file glob parameter"))?;
 
+        // Parse the URL and separate the base URL (keep scheme, host, etc.) from the path.
+        let glob_url_str = glob_url_parameter.as_ref().as_string();
+        let glob_url = match Url::parse(glob_url_str.as_str()) {
+            Ok(url) => url,
+            Err(_) => {
+                // Otherwise, we assume it's a file path.
+                let path = if !glob_url_str.as_str().starts_with("/") {
+                    // We cannot use Path::canonicalize to resolve relative paths since it requires the file to exist, and the glob may contain wildcards. Instead, we resolve relative paths against the current working directory.
```
0ax1: nit: Line break the comment.

vortex-duckdb/src/scan.rs
```diff
@@ -270,20 +287,28 @@ impl TableFunction for VortexTableFunction {
 
         tracing::trace!("running scan with max_threads {max_threads}");
 
-        let file_glob = file_glob_string.as_ref().as_string();
-        let file_urls = duckdb_fs_glob(ctx, file_glob.as_ref())?;
```
0ax1, line 274: Why switch back to our glob rather than the duckdb-fs one?

vortex-io/src/compat/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use futures::Stream;
+use pin_project_lite::pin_project;
+use vortex_error::VortexExpect;
+
+/// Get a handle to the current Tokio runtime, or create a new global single-thread runtime if one
+/// is not found.
+///
+/// From
+fn get_runtime_handle() -> tokio::runtime::Handle {
```
0ax1: Can we strip the Java-esque `get`-prefix? 🙂

vortex-duckdb/src/filesystem.rs
gatesn: The reason for now is that DuckDB FS supports way more object store backends. We only support S3. So you can enable this to use GCS / Azure... for now...

vortex-duckdb/src/scan.rs
gatesn: For consistency across all Vortex integrations

0ax1: This breaks the integration with httpfs?

0ax1 (approved):

PR #6565, opened 2026-02-18
fix: install and load httpfs in setup phase

0ax1:
## Summary

Change the DuckDB benchmark setup to install and load `httpfs`  to prevent this from implicitly happening during a benchmark run.


AdamGS (approved):

0ax1:
cc: @gatesn

PR #6566, opened 2026-02-18
minor c++ duckdb integration changes

myrrc:
## Summary

Minor C++ Duckdb integration API cleanup:

1. Slightly more modern cmake build commands
2. config.cpp: use duckdb_malloc instead of malloc for traced allocations
3. copy_functions.cpp: pre-reserve vector sizes
4. data_chunk.cpp: zero error pointer value in case of no errors (not the pointer itself).
5. error.cpp: construct SetError from string_view to avoid double allocations
   i.e. const char -> const string& -> (inside SetError) string
   error.cpp: don't catch extra duckdb::Exception (it's derived from std::exception)
6. object_cache.cpp: Don't reimplement unique_ptr with deleter, use it
7. table_filter: fix UB in case index is greater than filters present, set
   out error pointer correctly.
   check data_wrapper->wrapper before trying to lock one of its members

Also, replace raw new() allocations with make_unique + release if constructor may
throw to avoid memory leaks.


vortex-cxx/README.md
```diff
@@ -14,22 +14,21 @@ This directory contains C++ bindings for Vortex using the [cxx](https://cxx.rs/)
...
+cmake --build build -j
 ```
 
 ### Running Tests
 
 ```bash
 # Enable tests in CMake
-cmake -DVORTEX_ENABLE_TESTING=ON ..
-make -j$(nproc)
+cmake -Bbuild -DVORTEX_ENABLE_TESTING=ON
```
0ax1: Nice! I do wonder though whether we should show how to use `ninja` with CMake by default in the example.

vortex-duckdb/cpp/include/duckdb_vx/error.hpp
```diff
@@ -11,6 +11,6 @@
 
 namespace vortex {
 std::string IntoErrString(duckdb_vx_error error);
-void SetError(duckdb_vx_error *error_out, const std::string &message);
+void SetError(duckdb_vx_error *error_out, std::string_view message);
```
0ax1, line 14: 👍

vortex-duckdb/cpp/replacement_scan.cpp
```diff
@@ -19,8 +19,9 @@ VortexScanReplacement(duckdb::ClientContext &context,
         return nullptr;
     }
     auto table_function = duckdb::make_uniq<duckdb::TableFunctionRef>();
-    duckdb::vector<duckdb::unique_ptr<duckdb::ParsedExpression>> children;
-    children.push_back(duckdb::make_uniq<duckdb::ConstantExpression>(duckdb::Value(table_name)));
+
+    duckdb::vector<duckdb::unique_ptr<duckdb::ParsedExpression>> children(1);
```
0ax1, line 23: Can we do braced init, rather than the separate assignment?

vortex-duckdb/cpp/table_filter.cpp
```diff
@@ -111,16 +109,19 @@ extern "C" duckdb_value duckdb_vx_dynamic_filter_data_get_value(duckdb_vx_dynami
...
 
-    if (!data_wrapper->data || !data_wrapper->data->filter || !data_wrapper->data->initialized) {
+    if (!data_wrapper->data->filter || !data_wrapper->data->initialized) {
         return nullptr;
     }
 
     // Return a heap allocated copy of the value.
-    return reinterpret_cast<duckdb_value>(new Value(data_wrapper->data->filter->constant));
+    auto value = duckdb::make_uniq<Value>(data_wrapper->data->filter->constant);
+    return reinterpret_cast<duckdb_value>(value.release());
```
0ax1: In this particular case, I'm not fully convinced this needs to be unique_ptr if we release right after. 🙂

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -233,7 +232,8 @@ extern "C" duckdb_value duckdb_vx_tfunc_bind_input_get_parameter(duckdb_vx_tfunc
         return nullptr;
     }
     const auto info = reinterpret_cast<TableFunctionBindInput *>(ffi_input);
-    return reinterpret_cast<duckdb_value>(new Value(info->inputs[index]));
+    auto value = duckdb::make_uniq<Value>(info->inputs[index]);
+    return reinterpret_cast<duckdb_value>(value.release());
```
0ax1: -//-

0ax1 (commented):
Couple of thoughts.

0ax1:
Oh, and def worth fully investigating the todos you brought up!

myrrc:
> show how to use `ninja` with CMake by default in the example.

Agreed

>Can we do braced init, rather than the separate assignment?

I've tried, but the issue is, if you you braces/std::initialized_list,
you get compilation errors as list item is not same as vector's item
(base vs. derived), and if you cast to base type, it gets ugly.

>not fully convinced this needs to be unique_ptr

Ok, agreed.

0ax1 (approved):
lgtm

PR #6568, opened 2026-02-18
chore: make auto `clang-format` consistent

0ax1:
## Summary

All calls to clang-format in diff `build.rs` should follow the same pattern.


myrrc (approved):
LGTM

PR #6569, opened 2026-02-18
chore: drop sync event from arrow device array

0ax1:
## Summary

The current take is that we can't rely on the consumer of the device array taking the sync event into account and therefore have to wait for decompression / copies to complete on our end.

### DIFF
--- a/vortex-cuda/src/arrow/canonical.rs
+++ b/vortex-cuda/src/arrow/canonical.rs
@@ -26,7 +26,6 @@ use crate::arrow::ArrowDeviceArray;
 use crate::arrow::DeviceType;
 use crate::arrow::ExportDeviceArray;
 use crate::arrow::PrivateData;
-use crate::arrow::SyncEvent;
 use crate::arrow::check_validity_empty;
 use crate::arrow::varbinview::BinaryParts;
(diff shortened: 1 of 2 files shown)


robert3005:
Can we have a sync event and also await it or is this because its confusing what to do with the event?

0ax1:
@a10y Your call obviously what we do with this. Where I was getting at here is if no consumer respects this we shouldn't rely on it. And if we wait anyway, it shouldn't be there.

robert3005 (approved):

a10y:
cudf does wait it

https://github.com/rapidsai/cudf/blob/main/cpp/src/interop/from_arrow_device.cu#L409-L412

It is part of the spec that if it's not null you're required to block on it. Is there another client we care about that doesn't wait properly?

0ax1:
> cudf does wait it
> 
> https://github.com/rapidsai/cudf/blob/main/cpp/src/interop/from_arrow_device.cu#L409-L412
> 
> It is part of the spec that if it's not null you're required to block on it. Is there another client we care about that doesn't wait properly?

Alright, let's leave it for now. We should make this configurable for the client whether we never wait and they rely on a sync event or whether we should wait.

PR #6570, opened 2026-02-18
chore: drop flaky duckdb http e2e test

0ax1:


robert3005 (approved):

joseph-isaacs:
Boooo, would be better to fix this!

PR #6573, opened 2026-02-18
chore: Fix java publish

robert3005:

### DIFF
--- a/.github/actions/setup-protoc/action.yml
+++ b/.github/actions/setup-protoc/action.yml
@@ -28,7 +28,7 @@ runs:
         PROTOC_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${{ inputs.protoc_version }}/protoc-${{ inputs.protoc_version }}-${PROTOC_ARCH}.zip"
         wget -q -O /tmp/protoc.zip "$PROTOC_URL"
         unzip -o /tmp/protoc.zip -d /tmp/protoc
-        mv /tmp/protoc/bin/protoc /usr/local/bin/
-        cp -r /tmp/protoc/include/* /usr/local/include/
+        sudo mv /tmp/protoc/bin/protoc /usr/local/bin/
+        sudo cp -r /tmp/protoc/include/* /usr/local/include/
         rm -rf /tmp/protoc.zip /tmp/protoc
         protoc --version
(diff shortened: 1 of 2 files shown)


AdamGS (approved):

.github/workflows/package.yml
```diff
@@ -109,8 +109,8 @@ jobs:
       fail-fast: false
       matrix:
         target:
-          - { os: ubuntu, runs-on: "ubuntu-24.04-arm", target: aarch64-unknown-linux-gnu }
-          - { os: ubuntu, runs-on: "ubuntu-24.04", target: x86_64-unknown-linux-gnu }
+          - { os: ubuntu, runs-on: "ubuntu-latest", target: aarch64-unknown-linux-gnu }
```
a10y: why are we switching from the ARM runner?

a10y (changes requested):

.github/workflows/package.yml
robert3005: oh, I didn't understand what it was doing then

a10y: also, we need to build on 24.04 as that's the latest LTS, if we build on ubuntu-latest we'll link a newer version of glibc i believe?

robert3005: `container` is a bit too magical

robert3005: we have container 20.04 so that we get old glibc, otherwise we need some NixOs to link a different one

a10y (approved):

.github/actions/setup-protoc/action.yml
```diff
@@ -28,7 +28,16 @@ runs:
         PROTOC_URL="https://github.com/protocolbuffers/protobuf/releases/download/v${{ inputs.protoc_version }}/protoc-${{ inputs.protoc_version }}-${PROTOC_ARCH}.zip"
         wget -q -O /tmp/protoc.zip "$PROTOC_URL"
         unzip -o /tmp/protoc.zip -d /tmp/protoc
-        sudo mv /tmp/protoc/bin/protoc /usr/local/bin/
-        sudo cp -r /tmp/protoc/include/* /usr/local/include/
+
+        # This sometimes runs in a container where user is root and there's no sudo.
```
a10y, line 32: this is insane

robert3005: I am so sorry but if you look at the failure of 0.59.0 that's what it boils down to. ubuntu containers are weird. I think you generally don't want root as a docker user but default images don't care

PR #6574, opened 2026-02-18
Fix some things that don't have warnings for some reason

connortsui20:
## Summary

I am in the process of migrating `vortex-dtype` into `vortex-array` as part of #6547, and I have run into a strange issue where after I move stuff, a bunch of warnings pop up that I didn't see before. These are some of them.

## Testing

N/A for now, these are mostly cosmetic

### DIFF
--- a/vortex-array/src/arrays/constant/compute/sum.rs
+++ b/vortex-array/src/arrays/constant/compute/sum.rs
@@ -60,6 +60,7 @@ fn sum_scalar(
                 .map(|v| ScalarValue::Primitive(v.into())))
         }
         DType::Primitive(ptype, _) => {
+            // TODO(connor): T is not used here
             let result = match_each_native_ptype!(
                 ptype,
                 unsigned: |T| { sum_integral::<u64>(scalar.as_primitive(), len, accumulator)?.map(|v| ScalarValue::Primitive(v.into())) },
--- a/vortex-array/src/arrays/primitive/compute/rules.rs
+++ b/vortex-array/src/arrays/primitive/compute/rules.rs
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use vortex_dtype::match_each_native_ptype;
 use vortex_error::VortexResult;
 
 use crate::ArrayRef;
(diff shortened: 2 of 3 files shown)


vortex-array/src/arrays/constant/compute/sum.rs
joseph-isaacs, line 68: why does clippy not warn is there a lint?

vortex-array/src/arrays/constant/compute/sum.rs
```diff
@@ -60,6 +60,7 @@ fn sum_scalar(
                 .map(|v| ScalarValue::Primitive(v.into())))
         }
         DType::Primitive(ptype, _) => {
+            // TODO(connor): T is not used here
             let result = match_each_native_ptype!(
                 ptype,
                 unsigned: |T| { sum_integral::<u64>(scalar.as_primitive(), len, accumulator)?.map(|v| ScalarValue::Primitive(v.into())) },
```
connortsui20, line 66: Should probably figure this out

vortex-array/src/arrays/primitive/compute/rules.rs
```diff
@@ -39,19 +38,18 @@ impl ArrayParentReduceRule<PrimitiveVTable> for PrimitiveMaskedValidityRule {
         _child_idx: usize,
     ) -> VortexResult<Option<ArrayRef>> {
         // Merge the parent's validity mask into the child's validity
+        let new_validity = array.validity().clone().and(parent.validity().clone())?;
+
         // TODO(joe): make this lazy
```
joseph-isaacs, line 43: move this up. I also think you can remove this

vortex-array/src/arrays/constant/compute/sum.rs
connortsui20: no clue

vortex-array/src/arrays/constant/compute/sum.rs
robert3005: I think this is the part where we don't have any other way to match things. In this case you can have a macro that splits things into 3 but doesn't have any further generics

vortex-array/src/arrays/primitive/compute/rules.rs
connortsui20: not sure what you mean?

vortex-array/src/arrays/constant/compute/sum.rs
connortsui20: Sorry, my concern here is that we hardcode `u64` and `i64`, and we don't really explain it at all

vortex-array/src/arrays/primitive/compute/rules.rs
joseph-isaacs: the todo applies to the validity

connortsui20:
I'm going to move forward with my other change so I can make progress, just let me know what you think here and I'll make the changes when clippy inevitably does start complaining

vortex-array/src/arrays/constant/compute/sum.rs
joseph-isaacs: Add a comment explaining this is how we do stats? of different types

vortex-array/src/arrays/constant/compute/sum.rs
joseph-isaacs: is a macro variable so there is no T after the macro runs

vortex-array/src/arrays/constant/compute/sum.rs
connortsui20: Is it? I don't know since it isn't clear where this is documented

vortex-array/src/arrays/constant/compute/sum.rs
connortsui20: Sorry, I mean that I do not know why on current develop there is no warning, but on my branch https://github.com/vortex-data/vortex/pull/6582 I have to do this otherwise clippy gets mad:

```rust
        DType::Primitive(ptype, _) => {
            #[expect(dead_code, reason = "TODO(connor): good question")]
            let result = match_each_native_ptype!(
                ptype,
                unsigned: |T| { sum_integral::<u64>(scalar.as_primitive(), len, accumulator)?.map(|v| ScalarValue::Primitive(v.into())) },
                signed: |T| { sum_integral::<i64>(scalar.as_primitive(), len, accumulator)?.map(|v| ScalarValue::Primitive(v.into())) },
                floating: |T| { sum_float(scalar.as_primitive(), len, accumulator)?.map(|v| ScalarValue::Primitive(v.into())) }
            );
            Ok(result)
        }
```

PR #6575, opened 2026-02-18
clippy disable std::result::Result, eliminate VortexExpect

a10y:
## Summary

Closes: #000


a10y:
This almost works, the biggest blocker I hit is I can't find a way to combine `allow(disallowed_types)` with `derive(serde::Serialize)`, since the error is in the expansion. I think the only way to do it is to make a module-level `allow` which is kind of shitty

a10y:
I don't think there's an easy workaround for the derive macros without forking clippy or using dylint or something

PR #6579, opened 2026-02-18
lower your expectations

a10y:
Just removing a few ambient `.vortex_expect()` calls I found in PCo and ALP that I don't think are correct or necessary

### DIFF
--- a/encodings/pco/src/array.rs
+++ b/encodings/pco/src/array.rs
@@ -390,7 +390,7 @@ impl PcoArray {
         let mut page_value_start = 0;
         let mut n_skipped_values = 0;
         for (chunk_info, chunk_meta) in self.metadata.chunks.iter().zip(&self.chunk_metas) {
-            let mut cd: Option<ChunkDecompressor<T>> = None;
+            let mut chunk_decompressor: Option<ChunkDecompressor<T>> = None;
             for page_info in &chunk_info.pages {
                 let page_n_values = page_info.n_values as usize;
                 let page_value_stop = page_value_start + page_n_values;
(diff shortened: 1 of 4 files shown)


encodings/pco/src/array.rs
```diff
@@ -384,9 +383,8 @@ impl PcoArray {
 
         // Then we decompress those pages into a buffer. Note that these values
         // may exceed the bounds of the slice, so we need to slice later.
-        let (fd, _) = FileDecompressor::new(self.metadata.header.as_slice())
-            .map_err(vortex_err_from_pco)
-            .vortex_expect("FileDecompressor::new should succeed with valid header");
+        let (fd, _) =
+            FileDecompressor::new(self.metadata.header.as_slice()).map_err(vortex_err_from_pco)?;
```
a10y, line 387: at this point, it's possible that decoding the pages fails b/c of file corruption or other things out of our control. We shouldn't panic b/c of user data issues

encodings/zstd/src/array.rs
```diff
@@ -648,11 +648,10 @@ impl ZstdArray {
 
         // then we actually decompress those frames
         let mut decompressor = if let Some(dictionary) = &self.dictionary {
-            zstd::bulk::Decompressor::with_dictionary(dictionary)
+            zstd::bulk::Decompressor::with_dictionary(dictionary)?
```
a10y, line 651: dictionary is a ByteBuffer which comes from the file, and may be corrupted/malformed so we shouldn't panic here

encodings/zstd/src/zstd_buffers.rs
```diff
@@ -267,8 +267,7 @@ impl ZstdBuffersArray {
     pub fn decode_plan(&self) -> VortexResult<ZstdBuffersDecodePlan> {
         // If invariants are somehow broken, device decompression could have UB, so ensure
         // they still hold.
-        self.validate()
-            .vortex_expect("zstd_buffers invariant violated before decode_plan");
+        self.validate()?;
```
a10y, line 269: in all our other arrays we don't panic on validation failures

joseph-isaacs (approved):

AdamGS (approved):

robert3005 (approved):

PR #6584, opened 2026-02-18
chore: drop `async` from CUDA traits

0ax1:
## Summary

Removes async from CUDA Vortex traits.

#### Execution model

For a given execution context, all work (with the exception to host buf copies) is enqueued onto a single CUDA stream and executes in FIFO order. Kernel launches are synchronous fire-and-forget: they enqueue work and return immediately. The returned `Canonical` may reference device buffers with in-flight writes; correctness is guaranteed by CUDA stream ordering, not by any CPU-side synchronization.

The only operations that require waiting for GPU work to complete are device-to-host copies, where the CPU must block until data lands in host memory.

To insert an explicit sync point, use `await_stream_callback`, which completes when all preceding stream work has finished.

### DIFF
--- a/vortex-cuda/src/stream.rs
+++ b/vortex-cuda/src/stream.rs
@@ -12,7 +12,6 @@ use cudarc::driver::DevicePtrMut;
 use cudarc::driver::DeviceRepr;
 use cudarc::driver::result::memcpy_htod_async;
 use cudarc::driver::result::stream;
-use futures::future::BoxFuture;
 use kanal::Sender;
 use tracing::warn;
 use vortex_array::buffer::BufferHandle;
@@ -22,7 +21,7 @@ use vortex_error::vortex_err;
 use crate::CudaDeviceBuffer;
 
 #[derive(Clone)]
-pub struct VortexCudaStream(pub Arc<CudaStream>);
+pub struct VortexCudaStream(pub(crate) Arc<CudaStream>);
 
 impl VortexCudaStream {
     /// Allocates a typed buffer on the GPU.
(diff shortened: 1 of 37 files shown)


vortex-cuda/src/stream.rs
joseph-isaacs, line 61: looks sus

0ax1:
Turns out, memory that is already allocated can be registered as pinned with `cuMemHostRegister_v2`, such that the copy to device will be async. See: https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDA__MEM_gf0a9fe11544326dabd743b7aa6b54223.html.

robert3005:
FWIW when I read the docs they strongly encourage to use their allocator as that's faster than cuMemHostRegister/cuMemHostUnregister but in places where we can't do that we should definitely fallback to that method

0ax1:
> FWIW when I read the docs they strongly encourage to use their allocator as that's faster than cuMemHostRegister/cuMemHostUnregister but in places where we can't do that we should definitely fallback to that method

That was my read too. Maybe we can think about this as a stop gap in the future before we wire in a "CUDA allocator".

vortex-cuda/src/stream.rs
0ax1: Which part exactly? Any `T` is `DeviceRepr`.

0ax1:
Closing for now, until we're 100% we don't need async.

joseph-isaacs:
I want to keep async for now at least (while knowing it increases complexity). 
We may want to await kernel launches in the executor which is much easier with async traits.

PR #6585, opened 2026-02-18
oncelock shared array future on cuda flat layout

onursatici:
## Summary

shared array future is recreated on each call to `array_future`. Oncelock that to share the same array for filter and projection evaluations

Closes: #000

### DIFF
--- a/vortex-cuda/src/layout.rs
+++ b/vortex-cuda/src/layout.rs
@@ -7,6 +7,7 @@ use std::collections::BTreeSet;
 use std::ops::BitAnd;
 use std::ops::Range;
 use std::sync::Arc;
+use std::sync::OnceLock;
 
 use async_trait::async_trait;
 use futures::FutureExt;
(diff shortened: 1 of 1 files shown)


robert3005:
@gatesn we are trully coming full circle

onursatici:
I like this point in the circle, it is fast

vortex-cuda/src/layout.rs
```diff
@@ -234,36 +236,44 @@ pub struct CudaFlatReader {
     name: Arc<str>,
     segment_source: Arc<dyn SegmentSource>,
     session: VortexSession,
+    array: OnceLock<SharedArrayFuture>,
```
gatesn, line 234: This reason we stopped doing this was because we don't have clear lifetimes for our layout readers. So I think you end up holding onto this memory for the duration of the entire scan.

The Layout V2 PR addresses this by storing cached data on a stream, allowing us to scope how long it's alive for: https://github.com/vortex-data/vortex/pull/6069

Check if we have a LayoutReader cache in DataFusion still. If so, this gets even worse!

onursatici: right, so maybe we can enable this just for the cuda flat layout until layout v2 lands?

gatesn (approved):

PR #6588, opened 2026-02-18
cuda FFOR + perfetto logging

a10y:
This PR adds Perfetto tracing generation to the CUDA scan-cli. Traces are emitted to `trace.pb` which you can drop directly into https://ui.perfetto.dev/ and open, for example:

image

All span + log arguments are preserved.

This PR also migrates our BP kernel to a fused FOR+BP operation. Pure BP will use reference of 0, and FOR execution will collapse the tree and fuse. This gives a noticeable boost in some of the per-batch traces of maybe ~10-20%.


vortex-array/src/executor.rs
```diff
@@ -94,18 +85,6 @@ impl ExecutionCtx {
...
     }
-
-    /// Log an execution entry point. On the first call into this context, the full
-    /// `display_tree` of the array is included so the starting state is visible.
-    fn log_entry(&mut self, array: &dyn Array, msg: fmt::Arguments<'_>) {
-        if tracing::enabled!(tracing::Level::DEBUG) {
-            if self.ops.is_empty() {
-                self.log(format_args!("{msg}\n{}", array.display_tree()));
-            } else {
-                self.log(msg);
```
a10y, line 105: this was taking >50% of execution time when i had logging enabled for the `vortex-cuda` crate, even if I didn't have it enabled for vortex-array. Mainly calculating the display_tree().

Scrapping it made the traces easier to read

a10y:
This is helpful, but >90% of scan time is spent in ZSTD still

image

vortex-cuda/gpu-scan-cli/src/main.rs
```diff
@@ -34,20 +49,30 @@ async fn main() -> VortexResult<()> {
     let args: Vec<String> = args().collect();
     let json_output = args.iter().any(|arg| arg == "--json");
 
+    let perfetto_file = File::create("trace.pb")?;
```
0ax1, line 52: Should this just live in current work dir?

vortex-cuda/src/kernel/encodings/for_.rs
```diff
@@ -47,10 +51,32 @@ impl CudaExecute for FoRExecutor {
     ) -> VortexResult<Canonical> {
         let array = Self::try_specialize(array).ok_or_else(|| vortex_err!("Expected FoRArray"))?;
 
+        // Fuse FOR + BP => FFOR
+        if let Some(bitpacked) = array.encoded().as_opt::<BitPackedVTable>() {
+            match_each_integer_ptype!(bitpacked.ptype(), |P| {
+                let reference: P = array.reference_scalar().try_into()?;
+                return decode_bitpacked(bitpacked.clone(), reference, ctx).await;
```
0ax1, line 58: Is the bitpacking + FoR unit tested? As in with a ref value != 0.

vortex-cuda/src/kernel/patches/mod.rs
```diff
@@ -18,6 +19,7 @@ use crate::CudaExecutionCtx;
 use crate::executor::CudaArrayExt;
 
 /// Apply a set of patches in-place onto a [`CudaDeviceBuffer`] holding `ValuesT`.
+#[instrument(skip_all)]
```
0ax1, line 22: Why do we need this?

0ax1 (commented):
Is pushing for perfetto an artefact of us using async `cuda_execute` which would be otherwise harder to perf profile?

joseph-isaacs:
Why did we choose perfetto over nsys or polar signals?

vortex-cuda/cuda_kernel_generator/mod.rs
```diff
@@ -20,23 +20,22 @@ fn generate_lane_decoder<T: FastLanes, W: Write>(
...
-                writeln!(output, "out[INDEX({row}, lane)] = zero;")?;
+                writeln!(output, "out[INDEX({row}, lane)] = reference;")?;
             }
         } else if bit_width == bits {
             writeln!(output)?;
             for row in 0..bits {
                 writeln!(
                     output,
-                    "out[INDEX({row}, lane)] = in[LANE_COUNT * {row} + lane];",
+                    "out[INDEX({row}, lane)] = in[LANE_COUNT * {row} + lane] + reference;",
```
joseph-isaacs, line 38: what happens if the ref is signed?

vortex-cuda/src/kernel/encodings/for_.rs
```diff
@@ -47,10 +51,32 @@ impl CudaExecute for FoRExecutor {
...
+            })
+        }
+
+        // Fuse FOR + SLICE + BP => SLICE + FFOR
+        if let Some(slice_array) = array.encoded().as_opt::<SliceVTable>()
+            && let Some(bitpacked) = slice_array.child().as_opt::<BitPackedVTable>()
+        {
+            let slice_range = slice_array.slice_range().clone();
+            let unpacked = match_each_integer_ptype!(bitpacked.ptype(), |P| {
+                let reference: P = array.reference_scalar().try_into()?;
```
joseph-isaacs, line 68: does this work with signed and unsigned values?

a10y:
Perfetto is drag-and-drop file just lets you visualize tracing spans with all of the parameters very easily. Can I use Polar Signals for that? I don't think nsys would show us Rust spans but I didn't try.

vortex-cuda/src/kernel/patches/mod.rs
a10y: i wanted to see the span. skip_all just means don't try and format!() the params, just show the function name, which is all I wanted

vortex-cuda/src/kernel/encodings/for_.rs
a10y: oof yea this is almost certainly wrong

a10y: actually no i think it works, let me add a test

a10y: for kernel launch CUDA just cares about the size so it knows how many bytes to copy into device memory, arithmetic will happen in unsigned world and just get cast back to signed at the end, is what i expect

vortex-cuda/src/kernel/encodings/for_.rs
a10y: added a unit test

vortex-cuda/src/kernel/encodings/for_.rs
a10y: added a unit test for signed ffor with negative reference value

robert3005:
you can use nsight for this as well. It gives you pretty much the same thing with gpu streams

robert3005:
basically nsys will give you CPU profile and gpu profile

0ax1 (approved):

PR #6589, opened 2026-02-18
perf[array]: chunked take with bounded compute graph 

### LINKED ISSUE #6407: dict on chunk explodes memory
### What happened?

I have this vortex array: 

[display_tree.txt](https://github.com/user-attachments/files/25216310/display_tree.txt)

`Take` is reshuffle of the array - all elements are used, just on a different position

When I try to write this array as a file, it explodes my memory (using more than 200 Gb even though Array takes just 500 Mb)

### Steps to reproduce

```rust
use std::fs::File;

use arrow_ipc::reader::FileReader;
use rand::rng;
use rand::seq::SliceRandom;
use vortex::VortexSessionDefault;
use vortex::array::Array;
use vortex::array::ArrayRef;
use vortex::array::arrays::PrimitiveArray;
use vortex::array::arrow::FromArrowArray;
use vortex::array::iter::{ArrayIteratorAdapter, ArrayIteratorExt};
use vortex::array::stream::ArrayStreamExt;
use vortex::array::validity::Validity;
use vortex::buffer::Buffer;
use vortex::dtype::DType;
use vortex::dtype::arrow::FromArrowType;
use vortex::file::WriteOptionsSessionExt;
use vortex::io::session::RuntimeSessionExt;
use vortex::session::VortexSession;

#[tokio::main]
    let session = VortexSession::default().with_tokio();

    let file = File::open("gharchive_2023010100.arrow")?;
    let reader = FileReader::try_new(file, None)?;
    let dtype = DType::from_arrow(reader.schema());

    let iter = ArrayIteratorAdapter::new(
        dtype,
        reader.map(|batch| Ok(ArrayRef::from_arrow(&batch?, false)?)),
    );

    let array = iter.into_array_stream().read_all().await?;

    let len = array.len();
    let mut indices: Vec<u64> = (0..len as u64).collect();
    indices.shuffle(&mut rng());
    let indices_array = PrimitiveArray::new(Buffer::from(indices), Validity::NonNullable);
    let array = array.take(indices_array.to_array())?;
    let stream = array.to_array_stream();

    session
        .write_options()
        .write(
            &mut async_fs::File::create(
                "/Users/blaginin/Documents/playground/gharchive_2023010100.vortex",
            )
            .await?,
            stream,
        )
        .await?;

    Ok(())
}
```

### Environment

Latest develop

### Additional context

_No response_

joseph-isaacs:
Currently take on a chunked array will create dict arrays for each contiguous indices over a chunk. Instead sort the indices to a filter on each chunk and then materalize and take the whole array 

closes: https://github.com/vortex-data/vortex/issues/6407


vortex/src/lib.rs
```diff
@@ -377,4 +388,66 @@ mod test {
...
+
+        println!("result {}", result.display_tree());
+
+        let result = array.execute::<ArrayRef>(&mut LEGACY_SESSION.create_execution_ctx())?;
+        println!("result {}", result.display_tree());
+
+        // Write the result to a Vortex file on disk.
+        let session = VortexSession::default();
+        let home = std::env::var("HOME")?;
+        let out_path = PathBuf::from(format!("{home}/Downloads/gharchive_shuffled.vortex"));
```
blaginin: if you actually want this test, i can write a script that generates that arrow

vortex/src/lib.rs
```diff
@@ -377,4 +388,66 @@ mod test {
...
+        session
+            .write_options()
+            .write(
+                &mut tokio::fs::File::create(&out_path).await?,
+                result.to_array_stream(),
+            )
+            .await?;
+        eprintln!("wrote {}", out_path.display());
+
+        Ok(())
```
robert3005: This should change

vortex/src/lib.rs
joseph-isaacs: I removed for now. Might be a benchmark we should run, maybe make a issue?

blaginin (approved):

PR #6591, opened 2026-02-18
break: clean up compare

joseph-isaacs:
Part of the lazy compute migration. We add a ArrayBuiltIn `a.binary(b, Op)` method to create a lazily evaluated array tree.

We remove RSub and RDiv since we don't need to swap args any more. Since we have `execute_parent`.

Note to review: Start looking at `vortex-array/{compute,exprs}`.

##
compute method have move to exprs
use `a.binary(b, Op)` from ArrayBuiltIn over compare and other compute methods


robert3005 (approved):
you have a comment in between that you want to return lazy compare and now you can

PR #6592, opened 2026-02-18
Published packages cannot depend on unpublished packages

robert3005:
We need to write a CI check for this but for now this fixes the immediate problem


joseph-isaacs (approved):

onursatici (approved):

a10y:
You actually can do this with dev-dependencies, but you can't use `workspace = true` you need to explicitly do `{ path = "../vortex-cuda" }`

robert3005:
yeah, you're right here. I will make a pr to add a publish dry-run, then restore this test

PR #6593, opened 2026-02-18
clean up numeric compute

joseph-isaacs:
Clean up the compute/numeric.rs public functions

## Breaks
Removed `compute/numeric.rs:numeric` use `ArrayBuiltIns::binary`


encodings/fsst/src/compute/filter.rs
```diff
@@ -22,8 +22,9 @@ impl FilterKernel for FSSTVTable {
         // Directly invoke VarBin's FilterKernel to get a concrete VarBinArray back.
         let filtered_codes = <VarBinVTable as FilterKernel>::filter(array.codes(), mask, ctx)?
             .vortex_expect("VarBin filter kernel always returns Some")
-            .as_::<VarBinVTable>()
-            .clone();
+            .try_into::<VarBinVTable>()
+            .ok()
```
robert3005, line 26: you don't need the ok here

robert3005 (approved):
only one nit comment, feel free to cleanup in follow up

PR #6595, opened 2026-02-18
expose Executor, AbortHandle, AbortHandleRef, and Handle

danking:
## Summary

Allow third parties to implement their own Executors.

The first three are necessary to implement an Executor. The last one, Handle, is necessary to attach an Executor to a Vortex Session.

## API Changes

This adds the Executor trait to the public API.

## Testing

I did not change any behavior.


a10y:
You're probably gonna need to update lockfiles:

```
cargo xtask public-api
```

a10y (approved):

gatesn (approved):

PR #6596, opened 2026-02-19
Expose table scan progress to DuckDB

myrrc:
Add table_scan_progress in duckdb's table function vtable. Return sum of read
(exported) chunks' length divided by sum of all chunks' length.

Unrelated changes:
- panic in duckdb-bench if no output formats were provided
- add a note how to set up duckdb-vortex's full testing with vortex-duckdb.

Testing:
- Generate .vortex files: cargo run --bin duckdb-bench  -- tpch -i1 --opt scale-factor=10.0 --formats vortex
2. Build duckdb vortex extension from this PR
3. Verify these queries show progress bar

PRAGMA enable_progress_bar;
SET progress_bar_time = 10;
select * from read_vortex("/path/to/vortex/vortex-bench/data/tpch/10.0/vortex-file-compressed/l*.vortex");

### DIFF
--- a/benchmarks/duckdb-bench/src/main.rs
+++ b/benchmarks/duckdb-bench/src/main.rs
@@ -84,6 +84,10 @@ fn main() -> anyhow::Result<()> {
         args.exclude_queries.as_ref(),
     );
 
+    if args.formats.is_empty() {
+        anyhow::bail!("provide a format with --formats");
+    }
+
     // Generate Vortex files from Parquet for any Vortex formats requested
     if benchmark.data_url().scheme() == "file" {
         // This is ugly, but otherwise some complicated async interaction might result in a deadlock
--- a/vortex-duckdb/cpp/include/duckdb_vx/table_function.h
+++ b/vortex-duckdb/cpp/include/duckdb_vx/table_function.h
@@ -133,7 +133,9 @@ typedef struct {
     void *pushdown_expression;
     duckdb_vx_string_map (*to_string)(void *bind_data);
     // void *dynamic_to_string;
-    void *table_scan_progress;
+
+    double (*table_scan_progress)(duckdb_client_context ctx, void *bind_data, void *global_state);
+
     idx_t (*get_partition_data)(const void *bind_data,
                                 void *init_global_data,
                                 void *init_local_data,
--- a/vortex-duckdb/src/duckdb/table_function/mod.rs
+++ b/vortex-duckdb/src/duckdb/table_function/mod.rs
@@ -14,6 +14,7 @@ mod cardinality;
 mod init;
 mod partition;
 mod pushdown_complex_filter;
+mod table_scan_progress;
 mod virtual_columns;
 
 pub use bind::*;
@@ -30,6 +31,7 @@ use crate::duckdb::expr::Expression;
 use crate::duckdb::table_function::cardinality::cardinality_callback;
 use crate::duckdb::table_function::partition::get_partition_data_callback;
 use crate::duckdb::table_function::pushdown_complex_filter::pushdown_complex_filter_callback;
+use crate::duckdb::table_function::table_scan_progress::table_scan_progress_callback;
 use crate::duckdb::table_function::virtual_columns::get_virtual_columns_callback;
 use crate::duckdb_try;
 
@@ -102,6 +104,13 @@ pub trait TableFunction: Sized + Debug {
         init: &TableInitInput<Self>,
         global: &mut Self::GlobalState,
     ) -> VortexResult<Self::LocalState>;
+
+    /// Return table scanning progress from 0. to 100.
+    fn table_scan_progress(
+        client_context: &ClientContext,
+        bind_data: &mut Self::BindData,
+        global_state: &mut Self::GlobalState,
+    ) -> f64;
 
     /// Pushes down a filter expression to the table function.
     ///
@@ -181,7 +190,7 @@ impl Database {
             pushdown_expression: ptr::null_mut::<c_void>(),
             get_virtual_columns: Some(get_virtual_columns_callback::<T>),
             to_string: Some(to_string_callback::<T>),
-            table_scan_progress: ptr::null_mut::<c_void>(),
+            table_scan_progress: Some(table_scan_progress_callback::<T>),
             get_partition_data: Some(get_partition_data_callback::<T>),
             projection_pushdown: T::PROJECTION_PUSHDOWN,
             filter_pushdown: T::FILTER_PUSHDOWN,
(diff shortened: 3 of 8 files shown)


vortex-duckdb/src/scan.rs
```diff
@@ -508,6 +520,15 @@ impl TableFunction for VortexTableFunction {
         })
     }
 
+    fn table_scan_progress(
+        _client_context: &ClientContext,
+        _bind_data: &mut Self::BindData,
+        global_state: &mut Self::GlobalState,
+    ) -> f64 {
+        global_state.chunk_len_read.load(Ordering::Relaxed) as f64
```
0ax1: We could implement:
```
impl VortexGlobalData {
    pub fn progress(&self) -> f64 {
        self.chunk_len_read.load(Ordering::Relaxed) as f64
            / self.chunk_len_total.load(Ordering::Relaxed) as f64
    }
}
```

and test 
```
        #[test]
        fn test_progress_zero_of_total() {
            let state = global_state_with_progress(100, 0);
            assert_eq!(state.progress(), 0.0);
        }

        #[test]
        fn test_progress_complete() {
            let state = global_state_with_progress(100, 100);
            assert_eq!(state.progress(), 1.0);
        }

        #[test]
        fn test_progress_halfway() {
            let state = global_state_with_progress(200, 100);
            assert!((state.progress() - 0.5).abs() < f64::EPSILON);
        }
```
This does not test though if the integration is wired in correctly.

vortex-duckdb/src/duckdb/table_function/table_scan_progress.rs
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use crate::duckdb::TableFunction;
+use vortex::error::VortexExpect;
```
0ax1, line 4: `use crate::cpp`
`use crate::duckdb`
`use std::os`

vortex-duckdb/src/scan.rs
```diff
@@ -109,10 +109,13 @@ impl Debug for VortexBindData {
     }
 }
 
+#[repr(align(64))] // put atomics on different cache lines
```
0ax1: Unless this is really relevant for performance I'd say let's drop this. This suggests to the reader that the struct being 64 byte aligned is relevant for perf. And the atomics here could still end up in the same cache line right?
```
   chunk_len_total: AtomicU64,
   chunk_len_read: AtomicU64,
```

benchmarks/duckdb-bench/src/main.rs
```diff
@@ -63,7 +63,7 @@ struct Args {
     #[arg(long, default_value_t = false)]
     hide_progress_bar: bool,
 
-    #[arg(long, value_delimiter = ',', value_parser = value_parser!(Format))]
+    #[arg(long, value_delimiter = ',', value_parser = value_parser!(Format), default_value = "vortex")]
```
joseph-isaacs: can we avoid default

vortex-duckdb/cpp/include/duckdb_vx/table_function.h
```diff
@@ -133,7 +133,12 @@ typedef struct {
     void *pushdown_expression;
     duckdb_vx_string_map (*to_string)(void *bind_data);
     // void *dynamic_to_string;
-    void *table_scan_progress;
+
+    // duckdb's api is
+    // double (*table_scan_progress)(duckdb_client_context ctx, void *bind_data, void *global_state);
+    // However, we don't use anything but global_state
+    double (*table_scan_progress)(void *global_state);
```
joseph-isaacs: the idea of the crate is that it could be outside of vortex-duckdb, so it would be nice to match the duckdb API

vortex-duckdb/src/duckdb/table_function/mod.rs
```diff
@@ -103,6 +105,11 @@ pub trait TableFunction: Sized + Debug {
         global: &mut Self::GlobalState,
     ) -> VortexResult<Self::LocalState>;
 
+    /// Returns table scanning progress
+    fn table_scan_progress(_global_state: &mut Self::GlobalState) -> f64 {
+        0.0
```
joseph-isaacs: why this default?

joseph-isaacs (approved):

vortex-duckdb/cpp/include/duckdb_vx/table_function.h
0ax1: > the idea of the crate 

you mean the idea of the "_header_"?

0ax1: But yeah maybe it's best to keep the APIs aligned.

vortex-duckdb/src/duckdb/table_function/mod.rs
0ax1: this can just be: `fn table_scan_progress(_global_state: &mut Self::GlobalState) -> f64;` without providing a fn body

vortex-duckdb/src/scan.rs
```diff
@@ -113,6 +114,16 @@ pub struct VortexGlobalData {
...
     ctx: ExecutionCtx,
+    bytes_total: Arc<AtomicUsize>,
+    bytes_read: AtomicUsize,
+}
+
+impl VortexGlobalData {
+    pub fn progress(&self) -> f64 {
+        let read = self.bytes_read.load(Ordering::Relaxed) as f64;
+        let total = self.bytes_total.load(Ordering::Relaxed) as f64;
+        (read + 1.) / (total + 1.) * 100. // return 1 when nothing is read
```
0ax1: This currently doesn't match what the test expects, and can never be 0. I assume you wanted to prevent division by 0?

myrrc:
> But yeah maybe it's best to keep the APIs aligned.

I believe if we can use one pointer we shouldn't pass three,
and as we reimplement table function's vtable anyway, we can cut
some corners, but this is minor, so I've reverted the code.

vortex-duckdb/src/scan.rs
```diff
@@ -380,6 +390,9 @@ impl TableFunction for VortexTableFunction {
                 .vortex_expect("error: exporter missing");
 
             let has_more_data = exporter.export(chunk)?;
+            global_state
+                .bytes_read
+                .fetch_add(chunk.len(), Ordering::Relaxed);
```
gatesn, line 395: This is wrong right? The variable is called bytes_read, but we're adding the row count of the chunk.

We're also adding this _after_ pulling the chunk? So presumably the progress just hovers around or just below 100% the entire time?

I think we need to figure out a better measure of progress using the Scan API after #6652

myrrc: It should be called "row_count", yes. The progress suprisingly is quite fine, not 100% all the time.

PR #6597, opened 2026-02-19
Bump Rust MSRV to 1.90

robert3005:
The main goal is to get stable workspace publishing so we can reduce our usage of nightly rust in CI


0ax1:
~~Can you state in the PR description why we bump it.~~

joseph-isaacs (approved):

0ax1 (approved):

PR #6601, opened 2026-02-19
Use approx footer size for caching instead of walking the layout tree

robert3005:
Taken from #6598 to minimise changes there. This should help metadata only queries


joseph-isaacs (approved):

joseph-isaacs:
Do you think this might have made random access regress

PR #6604, opened 2026-02-19
Fix scalar deserialisation where value was unsigned/signed for signed/unsigned type

### LINKED ISSUE #6600: Can't read old vortex file. Stats expected signed integer ptype for Int64Value, got u64
### What happened?

`Failed to deserialize statistics for encoding=vortex.primitive, dtype=u64?: Serde error: expected signed integer ptype for Int64Value, got u64`

[public-177331_tables_table_yqNmIBdsb8EnVjJWyBnQ9_m_hk_hkFRWEDrKmNqMazVsIhP6.vortex.zip](https://github.com/user-attachments/files/25418693/public-177331_tables_table_yqNmIBdsb8EnVjJWyBnQ9_m_hk_hkFRWEDrKmNqMazVsIhP6.vortex.zip)

### Steps to reproduce

```rust
use vortex::array::Array;
use vortex::array::stream::ArrayStreamExt;
use vortex::file::OpenOptionsSessionExt;
use vortex::session::VortexSession;
use vortex::VortexSessionDefault;

#[tokio::main]
    let session = VortexSession::default();
    let path = "/Users/blaginin/Downloads/public-177331_tables_table_yqNmIBdsb8EnVjJWyBnQ9_m_hk_hkFRWEDrKmNqMazVsIhP6.vortex";

    let file = session.open_options().open_path(path).await?;

    println!("dtype: {}", file.dtype());
    println!("row_count: {}", file.row_count());

    let result = file
        .scan()?
        .into_array_stream()?
        .read_all()
        .await?;

    
    Ok(())
}
```

### Environment

latest develop

### Additional context

_No response_

robert3005:
It looks like it was possible to write unsinged array stats with signed values... this was demonstrated in #6600. Since this used to be possible and while this is a bug we need to support reading those files. This is a minimal change that allows reading unsigned stats from signed values.

fix #6600

### DIFF
--- a/vortex-array/src/scalar/proto.rs
+++ b/vortex-array/src/scalar/proto.rs
@@ -272,6 +272,11 @@ fn int64_from_proto(v: i64, dtype: &DType) -> VortexResult<ScalarValue> {
         PType::I16 => v.to_i16().map(PValue::I16),
         PType::I32 => v.to_i32().map(PValue::I32),
         PType::I64 => Some(PValue::I64(v)),
+        // It was previously possible for unsigned types to get their stats serialised as signed
+        PType::U8 => v.to_u8().map(PValue::U8),
+        PType::U16 => v.to_u16().map(PValue::U16),
+        PType::U32 => v.to_u32().map(PValue::U32),
+        PType::U64 => v.to_u64().map(PValue::U64),
         ptype => vortex_bail!(
             Serde: "expected signed integer ptype for Int64Value, got {ptype}"
         ),
@@ -297,6 +302,11 @@ fn uint64_from_proto(v: u64, dtype: &DType) -> VortexResult<ScalarValue> {
         PType::U16 => v.to_u16().map(PValue::U16),
         PType::U32 => v.to_u32().map(PValue::U32),
         PType::U64 => Some(PValue::U64(v)),
+        // It was previously possible for signed types to get their stats serialised as unsigned
+        PType::I8 => v.to_i8().map(PValue::I8),
+        PType::I16 => v.to_i16().map(PValue::I16),
+        PType::I32 => v.to_i32().map(PValue::I32),
+        PType::I64 => v.to_i64().map(PValue::I64),
         // Backwards compatibility: f16 values were previously serialized as u64.
         PType::F16 => v.to_u16().map(f16::from_bits).map(PValue::F16),
         ptype => vortex_bail!(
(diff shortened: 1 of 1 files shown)


connortsui20:
☠️ :sob:

vortex-array/src/scalar/proto.rs
```diff
@@ -272,6 +272,11 @@ fn int64_from_proto(v: i64, dtype: &DType) -> VortexResult<ScalarValue> {
         PType::I16 => v.to_i16().map(PValue::I16),
         PType::I32 => v.to_i32().map(PValue::I32),
         PType::I64 => Some(PValue::I64(v)),
+        // It was possible for unsigned types to get their stats serialised as signed
```
connortsui20:
```suggestion
        // It was previously possible for unsigned types to get their stats serialised as signed,
        // so we must allow reading unsigned integers as signed for backwards compatibility.
```

vortex-array/src/scalar/proto.rs
```diff
@@ -272,6 +272,11 @@ fn int64_from_proto(v: i64, dtype: &DType) -> VortexResult<ScalarValue> {
...
+        PType::U8 => v.to_u8().map(PValue::U8),
+        PType::U16 => v.to_u16().map(PValue::U16),
+        PType::U32 => v.to_u32().map(PValue::U32),
+        PType::U64 => v.to_u64().map(PValue::U64),
         ptype => vortex_bail!(
             Serde: "expected signed integer ptype for Int64Value, got {ptype}"
```
connortsui20, line 281: Can we change this error message to be specifically for floats now?

connortsui20 (changes requested):
Actually I think we should be careful here...

This sort of goes against the initial point of the change where we wanted to verify that all scalars are valid for their dtype. If we add this, that invariant is broken and there might be a bunch of code on top of this that breaks because it can no longer assume that.

I feel like it would be much better to just cast the type correctly. Since we are preserving bit width, technically all that would be "unexpected" is very high or very negative values, which would more likely be caught somewhere up the stack. That is opposed to allowing scalar dtype and value to continue to have mismatches.

We do a similar thing for the u64 -> f16 where we convert on construction:

https://github.com/vortex-data/vortex/blob/6206993879565098437782795e34e2a46ffd5dba/vortex-array/src/scalar/proto.rs#L289-L301

Also as a side note, the fact that we don't have tests for other incorrect types being read with the wrong dtype is worrying, after we decide what to do I can add more correctness tests

blaginin:
good news: i can now read the file from the issue✈️
bad news: in spiral, it still [fails](https://github.com/spiraldb/spiraldb/actions/runs/22189472849/job/64172730143?pr=4089#step:9:76)

joseph-isaacs:
We could drop these stats instead of casting?

connortsui20:
I think dropping is not ideal unless we write them back (also how would you know how to drop the stats? Match in the error?).

a10y:
I actually think I agree with Joe, if a stat is garbage we should just drop it on read

lwwmanning:
Better to drop than get incorrect value

blaginin:
latest commit fixes it for spiral 🙏

joseph-isaacs (approved):

connortsui20:
~~This change doesn't seem to match either of the threads of conversation we had on Slack nor here, is there an update in what we think should happen?~~

Nevermind, I just read the code wrong

connortsui20 (approved):
Some of the comments could be clearer but I'll just make the change myself in a followup PR

PR #6608, opened 2026-02-19
perf: remove double execute in exprs

joseph-isaacs:
Sometime we double execute array in expression eval


robert3005 (approved):

danking (approved):
hmm this is odd?

> clickbench_q23/duckdb:vortex-compact | 8284317229 | 4.34038e+09 | 1.90866 | ns | 🚨


is tpc-ds always this unreliable? 

> Significant (>10%): 214↑ 2↓


otherwise lgtm

joseph-isaacs:
I will be doing an opt pass of the compute going forwards

PR #6612, opened 2026-02-20
Trying Shipfox

connortsui20:
DO NOT MERGE

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,18 +48,9 @@ jobs:
 
   python-lint:
     name: "Python (lint)"
-    runs-on:
-      - runs-on=${{ github.run_id }}
-      - family=m7i+m7i-flex+m7a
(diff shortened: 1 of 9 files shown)


connortsui20:
So that was the cold run: https://github.com/vortex-data/vortex/actions/runs/22227024571

Let's try to hot run with the cache: https://github.com/vortex-data/vortex/actions/runs/22231022356

joseph-isaacs:
What's your analysis @connortsui20 ?

connortsui20:
@joseph-isaacs some of the shipfox people have offered to help me set it up (I'll look into this more next week), I just did the bare minimum and it seems like it is faster in some cases, slower in other cases, but note that I disabled the cache so that's likely a big part of it

robert3005:
fwiw this is still 10x more expensive than runson. I think if the perf improvement is not 10x this is not worth it.

PR #6614, opened 2026-02-20
Fix compression benchmarks build

AdamGS:
Some other crate stopped enabling this feature, which changed how the dependency is resolved, so now it doesn't build.


robert3005 (approved):

gatesn (approved):
Do we want to just enable in the top level Cargo.toml?

AdamGS:
@gatesn I don't think so. doesn't seem like we derive a `Serialize/Desirialize` implementation anywhere else

AdamGS:
(this build failure is unrelated, will rerun)

Issue #6618, opened 2026-02-20
Tracking Issue: Extension Types (old)


connortsui20:
This is a continuation of the tracking issue https://github.com/vortex-data/vortex/issues/6547.

# More Robust Extension Data Types In Vortex

We would like to build a more robust system for extension data types (or DTypes).

https://github.com/vortex-data/vortex/pull/6081 introduced vtables (virtaul tables, or Rust unit structs with methods) for extension `DType`s. Each extension type (e.g. `Timestamp`) now implements `ExtDTypeVTable`, which handles validation, serialization, and metadata. The type-erased `ExtDTypeRef` carries this vtable with it inside `DType::Extension`.

There were a few blockers (detailed in the previous tracking issue https://github.com/vortex-data/vortex/issues/6547), but now that those have been resolved we can move forward with our plans.

## One `VTable` to rule them all

Now that `vortex-scalar` and `vortex-dtype` have been merged into `vortex-array`, we can now place all extension logic (for types, scalars, and arrays) onto an `ExtVTable`. It will look something like so:

```rust
// Naming should be considered VERY unstable / not set!

pub trait ExtVTable: 'static + Send + Sync + ... {
    type Metadata: ...;

    // `DType`

    fn id(&self) -> ExtID;
    fn validate(&self, metadata: &Self::Metadata, storage: &DType) -> VortexResult<()>;
    fn serialize(&self, metadata: &Self::Metadata) -> VortexResult<Vec<u8>>;
    fn deserialize(&self, data: &[u8]) -> VortexResult<Self::Metadata>;

    // `Scalar`

    // (This is not how it actually would look, but close enough)
    fn display(&self, metadata: &Self::Metadata, value: &ScalarValue, f: &mut fmt::Formatter) -> fmt::Result { ... }
    fn cast(&self, ...) -> VortexResult<Scalar> { ... }

    // `ArrayRef`

    fn compute_thing???(&self, ...) -> VortexResult<ArrayRef> { ... }
    // <-- Probably a lot more than this -->
}
```

@joseph-isaacs has mentioned that it might be better to call this `CanonicalExtVTable`. I would propose `ExtCanonicalVTable` or maybe even `StorageExtVTable`?

This sort of stems from the fact that the `ExtensionArray` in `Canonical::Extension(ExtensionArray)` should really imply that we are extending canonical, not extending arrays with completely arbitrary data.

[...]

PR #6624, opened 2026-02-20
Add an `extension` module to `vortex`

connortsui20:
## Summary

Since `vortex-dtype` and `vortex-scalar` have merged into `vortex-array`, and since planned extension logic now covers all 3 of dtype, scalar, and array, we need to pull out the existing extension dtype logic so that all 3 of these modules can access it.

This PR adds a `vortex::extension` submodule to `vortex` that contains the `ExtVTable` (which was previously the `ExtDTypeVTable`), and a sub-submodule `dtype` that contains things like `ExtDTypeRef`.

We should hope to add other sub-submodules `scalar` and `array` that would have `ExtScalarValueRef` and `ExtArrayRef`.

I am wondering if we need to figure out the naming (please read the discussion in #6618) right now. Do we care that much right now or can we just change it later?

## API Changes

Adds `vortex::extension` (which is pulled from `vortex_array::extension`), and moves some logic around extension dtypes and extension vtable into this new submodule.

AFAIK nobody is actually using the new extension dtype so I think it should be fine to break?

There is still a concern though if we release this version and someone decides to start implementing `ExtVTable`: we need to add a lot more methods to that

## Testing

N/A since this is just moving code around.


joseph-isaacs:
> vortex::extension submodule

vortex_array ?

gatesn:
It should be exposed as `vortex::ext` because in theory we should ship this as a `vortex-ext` crate. Problem is we haven't pulled out `vortex-arrow` yet so we cannot pull this crate out.

connortsui20:
ok addressed comments made offline. I am heading out now so won't be able to work on this till next week.

vortex-array/src/dtype/extension/mod.rs
```diff
@@ -1,340 +1,32 @@
...
+pub use vtable::DynExtVTable;
+pub use vtable::ExtVTable;
 
-use crate::dtype::DType;
-use crate::dtype::Nullability;
+mod dtype;
+pub use dtype::*;
 
 /// A unique identifier for an extension type
-pub type ExtID = ArcRef<str>;
```
gatesn, line 29: Seems kinda pointless to move this under dtype::*, but 🤷

vortex-array/public-api.lock
```diff
@@ -6254,361 +6254,47 @@ impl vortex_array::dtype::arrow::TryFromArrowType<&arrow_schema::datatype::DataT
 
 pub fn vortex_array::dtype::PType::try_from_arrow(value: &arrow_schema::datatype::DataType) -> vortex_error::VortexResult<Self>
 
-pub mod vortex_array::dtype::datetime
-
-pub enum vortex_array::dtype::datetime::TemporalJiff
```
gatesn, line 6259: Niceeeeeee

gatesn (approved):

PR #6625, opened 2026-02-20
Make the tui an default optional dependecy of vortex-python

AdamGS:
## Summary

When using vortex-python as a library, adding the tui isn't necessarily useful and pulls in many significant dependencies (Like DF, parquet)


blaginin (approved):
🙏🏻🙏🏻🙏🏻

PR #6629, opened 2026-02-20
More minimal listview rebuild benches, force canonicalize

palaska:


joseph-isaacs:
Do we want to remove them?

0ax1:
@claude review

palaska:
> Do we want to remove them?

I updated them, neither take > 1ms now so should be fine to keep?

0ax1 (approved):

PR #6631, opened 2026-02-20
only a read lock for current_array_ref

danking:
## Summary

SharedArray uses a Mutex as of #6453. This interacts particularly badly with `scalar_at` by way of `is_valid`. In particular, all calls to `scalar_at` on the given array are serialized. It seems reasonable to allow multiple threads to simultaneously read the validity. This PR permits that by using an RwLock.

## Testing

I have a Vortex-based system which recently pulled in a version of Vortex that includes #6453. With that version, my profile looks like this:

https://share.firefox.dev/4tO94Ki
Screenshot 2026-02-20 at 4 26 28 PM

With this PR applied (i.e. with an RwLock in place of the Mutex), it looks like this:

https://share.firefox.dev/4aNPZzt
Screenshot 2026-02-20 at 4 27 25 PM

CPU utilization is high, and runtime is ~7s. By comparison, with a Mutex, CPU utilization is low and runtime is ~18s.

### DIFF
--- a/vortex-array/src/arrays/shared/array.rs
+++ b/vortex-array/src/arrays/shared/array.rs
@@ -4,8 +4,9 @@
 use std::future::Future;
 use std::sync::Arc;
 
-use async_lock::Mutex;
-use async_lock::MutexGuard;
+use async_lock::RwLock;
+use async_lock::RwLockReadGuard;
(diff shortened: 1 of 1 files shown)


a10y (approved):
async locks always make me shiver a bit, but i think this is fine

PR #6634, opened 2026-02-20
Convert &mut to & for DuckDB GlobalData

### LINKED ISSUE #6632: DuckDB GlobalState should be passed by immutable reference
We do not have unique ownership of the global state, it should be passed by immutable reference and handle its own interior thread-safety / mutability.

To fix this, you will run into the problem of requiring a mutable ExecutionCtx. This shouldn't be global state anyway and we should just create one for each array exporter.

gatesn:
Fixes #6632

Previously we were casting the DuckDB global data pointer to a `&mut`, which in Rust implies exclusive ownership. But this is not true since these callbacks were thread-local functions being given a pointer to the shared global data.

This change meant fixing the scope of our use of ExecutionCtx to one per ArrayRef chunk (same scope as the ArrayExporter)


0ax1:
@claude anything suspicious in this pr

0ax1 (approved):
🦆

vortex-duckdb/src/duckdb/copy_function/mod.rs
joseph-isaacs, line 45: This one should stay as mut and we would document they this and the above are different

vortex-duckdb/src/scan.rs
```diff
@@ -342,11 +339,12 @@ impl TableFunction for VortexTableFunction {
         _client_context: &ClientContextRef,
         _bind_data: &Self::BindData,
         local_state: &mut Self::LocalState,
-        global_state: &mut Self::GlobalState,
+        global_state: &Self::GlobalState,
         chunk: &mut DataChunkRef,
     ) -> VortexResult<()> {
         loop {
             if local_state.exporter.is_none() {
+                let mut ctx = SESSION.create_execution_ctx();
```
joseph-isaacs: One per split?

vortex-duckdb/src/scan.rs
```diff
@@ -365,15 +363,13 @@ impl TableFunction for VortexTableFunction {
                         pack_options.nullability.into(),
                     )
                 } else {
-                    array_result
-                        .execute::<Canonical>(&mut global_state.ctx)?
-                        .into_struct()
+                    array_result.execute::<Canonical>(&mut ctx)?.into_struct()
```
joseph-isaacs: StructArray?

vortex-duckdb/src/exporter/all_invalid.rs
```diff
@@ -22,7 +23,13 @@ pub(crate) fn new_exporter(len: usize, logical_type: &LogicalTypeRef) -> Box<dyn
 }
 
 impl ColumnExporter for AllInvalidExporter {
-    fn export(&self, offset: usize, len: usize, vector: &mut VectorRef) -> VortexResult<()> {
+    fn export(
+        &self,
+        offset: usize,
+        len: usize,
+        vector: &mut VectorRef,
+        _ctx: &mut ExecutionCtx,
```
joseph-isaacs, line 31: A bonus feature

myrrc (approved):

0ax1:
@gatesn Is this good to go?

PR #6635, opened 2026-02-20
Dk/buffer pool

danking:
## Summary

Closes: #000


AdamGS:
I know this is still WIP, but should it be optionally extendable? Arrow-cpp has a [`MemoryPool`](https://arrow.apache.org/docs/cpp/memory.html#memory-pools) extension point, and some logic for picking a default one (in the rust case - probably fallback to the default global allocator one?)

PR #6637, opened 2026-02-22
docs: Remove unused CLI argument (code always uses temporary path)

SYaoJun:
## Summary

Closes: #000

### DIFF
--- a/vortex-cxx/examples/README.md
+++ b/vortex-cxx/examples/README.md
@@ -7,5 +7,5 @@ mkdir -p build
 cd build
 cmake ..
 make -j$(nproc)
-./hello-vortex ../goldenfiles/example.vortex
+./hello-vortex
 ```


joseph-isaacs:
Mind explaining why this is unneeded?

SYaoJun:
> Mind explaining why this is unneeded?

This example lacks support for command-line arguments. Specifically, the main() function is not implemented to accept the conventional `int argc, char *argv[]` parameters—instead, it hardcodes `temp_dir` as the default path and provides no mechanism for users to specify a custom path via CLI input.


[hello-vortex.cpp: 71-74](https://github.com/vortex-data/vortex/blob/develop/vortex-cxx/examples/hello-vortex.cpp#L71-L74)

```cpp
int main() {
// Create a temporary file path
std::filesystem::path temp_dir = std::filesystem::temp_directory_path();
std::string vortex_file = (temp_dir / "hello_vortex_example.vortex").string();
```
cc @joseph-isaacs

0ax1 (approved):

PR #6645, opened 2026-02-23
allocators

onursatici:
## Summary

Add buffer allocator trait and a simple cpu buffer pool. `VortexReadAt` now can own a buffer pool through a lower level trait `ReadInto`, and AllocatingReader will bridge them. 

This PR adds the types, but doesn't wire the buffer pool in yet. The existing ReadAt implementors currently use the default allocator, which allocates a fresh new BufferMut per call

### DIFF
--- a/vortex-io/src/write_target.rs
+++ b/vortex-io/src/write_target.rs
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use futures::FutureExt;
+use futures::future::BoxFuture;
+use vortex_array::buffer::BufferHandle;
+use vortex_buffer::ByteBufferMut;
+use vortex_error::VortexResult;
+
+/// A destination for I/O reads that can be finalized into a [`BufferHandle`].
+pub trait WriteTarget: Send + 'static {
+    /// Returns the buffer as a mutable slice.
+    fn as_mut_slice(&mut self) -> &mut [u8];
+
+    /// Returns the length of the buffer in bytes.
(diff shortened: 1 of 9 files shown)


gatesn (commented):
This seems to draw some weird boundaries? 

What's the goal here? To move Vortex towards a buffer pool? Or to allow I/O to read into both host and device buffers? 

Feels like the sort of thing we should have a discussion / rfc for! But don't want to stall your experiments on this

onursatici:
yes we could use a discussion here, the goal is to have control on how the buffer handles are created during a scan. I want to use a vortex read at implementation, but I also want to control the allocations

onursatici:
https://github.com/vortex-data/vortex/discussions/6647

0ax1:
How would for example CUDA host mem malloc fit into this?

vortex-io/src/write_target.rs
```diff
@@ -0,0 +1,39 @@
...
+use futures::FutureExt;
+use futures::future::BoxFuture;
+use vortex_array::buffer::BufferHandle;
+use vortex_buffer::ByteBufferMut;
+use vortex_error::VortexResult;
+
+/// A destination for I/O reads that can be finalized into a [`BufferHandle`].
+pub trait WriteTarget: Send + 'static {
+    /// Returns the buffer as a mutable slice.
+    fn as_mut_slice(&mut self) -> &mut [u8];
```
a10y, line 13: would it make more sense to return an uninit slice? that way we can avoid zero-initializing all the memory. maybe that's too much of a microoptimization

onursatici:
decided to do cuda only here: https://github.com/vortex-data/vortex/pull/6675

PR #6648, opened 2026-02-23
Slice(ChunkedArray) canonicalizes all chunks instead of only relevant ones

palaska:
SliceVTable was using the default append_to_builder, which calls SliceVTable::execute directly. This bypasses the execution loop, so execute_parent optimizations like SliceExecuteAdaptor(ChunkedVTable) never fire.

For Slice(ChunkedArray), this means the entire chunked array gets canonicalized before slicing — instead of using find_chunk_idx to select only the relevant chunks first.

### DIFF
--- a/vortex-array/public-api.lock
+++ b/vortex-array/public-api.lock
@@ -3152,6 +3152,8 @@ pub type vortex_array::arrays::SliceVTable::ValidityVTable = vortex_array::array
 
 pub type vortex_array::arrays::SliceVTable::VisitorVTable = vortex_array::arrays::SliceVTable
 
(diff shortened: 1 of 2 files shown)


AdamGS:
this seems reasonable, running benchmarks to make sure there are no surprises

AdamGS (approved):
I think this makes perfect sense, I'm not sure there's a way to test it though

palaska:
I noticed this was also missing at Filter as well (https://github.com/vortex-data/vortex/pull/6650). Holding off until we decide whether we want to do this by default (https://github.com/vortex-data/vortex/pull/6655) . It looks like it degrades perf a little but not sure how important that is. A wasted execute_parent call should be just a couple of vtable dispatches and a type check, no?

palaska:
Closing this one, updated the default implementation here: https://github.com/vortex-data/vortex/pull/6655

PR #6650, opened 2026-02-23
FilterVTable append_to_builder bypasses execute_parent optimizations

palaska:
Same fix as https://github.com/vortex-data/vortex/pull/6648

`FilterVTable::execute` calls `child.execute(ctx)` directly, bypassing the full execution loop — so `execute_parent` optimizations like `FilterExecuteAdaptor(ChunkedVTable)` never fire

For `Filter(ChunkedArray)`, this canonicalizes all chunks then filters, instead of pushing the filter into each chunk

### DIFF
--- a/vortex-array/src/arrays/filter/vtable.rs
+++ b/vortex-array/src/arrays/filter/vtable.rs
@@ -18,6 +18,7 @@ use crate::ArrayChildVisitor;
 use crate::ArrayEq;
 use crate::ArrayHash;
 use crate::ArrayRef;
+use crate::Canonical;
 use crate::IntoArray;
 use crate::Precision;
 use crate::arrays::filter::array::FilterArray;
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/filter/vtable.rs
```diff
@@ -109,6 +111,18 @@ impl VTable for FilterVTable {
...
     }
 
+    fn append_to_builder(
+        array: &Self::Array,
+        builder: &mut dyn ArrayBuilder,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<()> {
+        // Execute through the full loop so that execute_parent optimizations
+        // can fire before we canonicalize.
+        let canonical = array.to_array().execute::<Canonical>(ctx)?;
```
gatesn, line 121: It's really annoying that we have to think about this everywhere we implement append_to_builder :/

palaska: would it make sense to make this the default?

palaska: I guess we should, I raised https://github.com/vortex-data/vortex/pull/6655

PR #6652, opened 2026-02-24
DuckDB to use the Scan API

gatesn:
This PR migrates the DuckDB implementation extension over to only using the Scan API.

We create a generic DuckDB table function that operates over any DataSource, and then a specific implementation that takes a path glob and runs a multi-file scan as is the existing behavior.

In the future we can easily add alternate table functions such as creating one from a DataSourceProvider URI, or by passing in a C pointer to a DataSource implementation.


vortex-duckdb/src/datasource.rs
```diff
@@ -0,0 +1,533 @@
...
+use crate::duckdb::TableInitInput;
+use crate::duckdb::VirtualColumnsResultRef;
+use crate::exporter::ArrayExporter;
+use crate::exporter::ConversionCache;
+
+// taken from duckdb/common/constants.h COLUMN_IDENTIFIER_EMPTY
+// This is used by duckdb whenever there is no projection id in a logical_get node.
+// For some reason we cannot return an empty DataChunk and duckdb will look for the virtual column
+// with this index and create a data chunk with a single vector of that type.
+static EMPTY_COLUMN_IDX: u64 = 18446744073709551614;
```
lwwmanning, line 70: fun fact: if you cast this to `i64`, it's `-1`

lwwmanning (approved):

joseph-isaacs:
@gatesn why did we remove the footer cache?

gatesn:
We want a multi-file version of a Vortex scan that is entirely implemented in Vortex, doesn't rely on query engine features. This is so we can provide good support to e.g. PyVortex where users might just write a scan directly. So the MultiFileDataSource handles this internally now.

PR #6655, opened 2026-02-24
Route default append_to_builder through full execution loop

palaska:
The default was calling `Self::execute` directly, bypassing the execution loop in `executor.rs`. This meant `reduce_parent` and `execute_parent` steps never ran, so child-driven optimizations couldn't fire before canonicalization.


robert3005:
@gatesn any reason we shouldn't do this? I don't quite understand why it was just `Self::execute` before

robert3005 (approved):

PR #6658, opened 2026-02-24
Support building duckdb from a hash, including CI

myrrc:
This is preliminary work to enable testing pre-release and hash-bashed
versions of DuckDB with Vortex locally and in CI.

- Enable release builds in CI from a hash to run benchmarks
- Build httpfs statically for hash-based builds, otherwise
  DuckDB will try to download it and fail.
- Enable curl development headers installation for httpfs
  if build.rs for vortex-duckb changed
- Test both httpfs backends in e2e scan test.
- Fix open_read's implementation incorrectly adding a slash between host and filename, which lead to a double-slash error with CURL backend https://github.com/duckdb/duckdb-httpfs/pull/261

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -54,6 +54,16 @@ jobs:
       - uses: actions/checkout@v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+      - uses: dorny/paths-filter@v3
+        id: duckdb-buildfile-changed
+        with:
+          filters: |
+            changed:
+              - 'vortex-duckdb/build.rs'
+      - name: Install curl dev headers for duckdb source build
+        if: steps.duckdb-buildfile-changed.outputs.changed == 'true'
+        run: |
+            sudo apt-get update -y -qq && sudo apt-get install -y -qq libcurl4-openssl-dev
       - uses: ./.github/actions/setup-rust
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,6 +188,29 @@ jobs:
       - uses: ./.github/actions/setup-rust
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      # DuckDB build from duckdb/, if building httpfs statically, fails unless
+      # you have cURL dev headers present. Ideally we'd see these loaded during
+      # DuckDB setup itself, but this is a tricky change, see
+      # https://github.com/duckdb/duckdb/issues/21073
+      #
+      # This does NOT check for DUCKDB_VERSION changes, and its use in CI
+      # is generally discouraged. Please change build.rs.
+      #
+      # Ideally I'd use a awalsh128/cache-apt-pkgs-action@v1.5.2,
+      # as apt-get takes around 22s, but this specific action is broken, see
+      # https://github.com/awalsh128/cache-apt-pkgs-action/issues/181
+      #
+      # TODO(myrrc): this should all go away once we have AMI in CI.
+      - uses: dorny/paths-filter@v3
+        id: duckdb-buildfile-changed
+        with:
+          filters: |
+            changed:
+              - 'vortex-duckdb/build.rs'
+      - name: Install curl dev headers for duckdb source build
+        if: steps.duckdb-buildfile-changed.outputs.changed == 'true'
+        run: |
+            sudo apt-get update -y -qq && sudo apt-get install -y -qq libcurl4-openssl-dev
       - name: Docs
         run: |
           RUSTDOCFLAGS="-D warnings" cargo doc --no-deps
@@ -287,6 +310,16 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Install nightly for fmt
         run: rustup toolchain install $NIGHTLY_TOOLCHAIN --component rustfmt
+      - uses: dorny/paths-filter@v3
+        id: duckdb-buildfile-changed
+        with:
+          filters: |
+            changed:
+              - 'vortex-duckdb/build.rs'
+      - name: Install curl dev headers for duckdb source build
+        if: steps.duckdb-buildfile-changed.outputs.changed == 'true'
+        run: |
+            sudo apt-get update -y -qq && sudo apt-get install -y -qq libcurl4-openssl-dev
       - name: Rust Lint - Format
         run: cargo +$NIGHTLY_TOOLCHAIN fmt --all --check
       - name: Rustc check
@@ -411,6 +444,16 @@ jobs:
         uses: taiki-e/install-action@v2
         with:
           tool: nextest
+      - uses: dorny/paths-filter@v3
+        id: duckdb-buildfile-changed
+        with:
+          filters: |
+            changed:
+              - 'vortex-duckdb/build.rs'
+      - name: Install curl dev headers for duckdb source build
+        if: steps.duckdb-buildfile-changed.outputs.changed == 'true'
+        run: |
+            sudo apt-get update -y -qq && sudo apt-get install -y -qq libcurl4-openssl-dev
       - name: Rust Tests
         if: ${{ matrix.suite == 'tests' }}
         run: |
@@ -477,8 +520,7 @@ jobs:
           components: "rust-src, rustfmt, clippy, llvm-tools-preview"
       - name: Install build dependencies
         run: |
-          sudo apt-get update
-          sudo apt-get install -y ninja-build cmake
+            sudo apt-get update -y -qq && sudo apt-get install -y -qq ninja-build cmake libcurl4-openssl-dev
       - name: Install nextest
         uses: taiki-e/install-action@v2
         with:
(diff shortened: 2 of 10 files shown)


.github/workflows/bench-pr.yml
```diff
@@ -54,6 +54,17 @@ jobs:
       - uses: actions/checkout@v6
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+      - uses: dorny/paths-filter@v3
```
0ax1, line 57: This should become a repo local GH action that we re-use: `.github/actions/setup-duckdb-deps/action.yml` rather than duplicating the CI snippet.

.github/workflows/sql-benchmarks.yml
```diff
@@ -138,6 +151,16 @@ jobs:
...
           cargo build $packages --profile release_debug
 
+          # If we've changed build.rs and we're on a hash, previous step would build
+          # vortex-duckdb, which in turn builds duckdb with httpfs. We then copy
+          # duckdb to our PWD, overriding the 1.4.2 binary.
+          # However, if we changed build.rs but we're NOT on a hash, then
+          # there will be no duckdb here, so cp may fail. Suppress the failure
+          # and let us use stock duckdb
+          if [ "${{ steps.duckdb-buildfile-changed.outputs.changed }}" = "true" ]; then
+            cp target/duckdb-source-*/duckdb-*/build/release/duckdb . || :
```
0ax1: This seems a bit brittle in terms of where we expect the binary to exist. Will this early exit loudly if the path doesn't exist?

vortex-duckdb/src/e2e_test/vortex_scan_test.rs
```diff
@@ -361,8 +361,10 @@ fn test_vortex_scan_over_http() {
     let listener = TcpListener::bind("127.0.0.1:0").unwrap();
     let addr = listener.local_addr().unwrap();
 
+    // Spawn 10 threads because DuckDB does HEAD and GET requests with retries,
+    // thus 2 threads, one for each implementation, aren't enough
     std::thread::spawn(move || {
-        for _ in 0..2 {
+        for _ in 0..10 {
             if let Ok((mut stream, _)) = listener.accept() {
```
0ax1, line 368: Not sure I fully follow this one. Do we just need one to succeed, and if so should we break after the first success?

.github/workflows/sql-benchmarks.yml
myrrc: We have two routes:
1. We're not on a hash => use stock duckdb which we've downloaded
2. We're on a hash => build duckdb and copy it from this weird path.
This explicitly doesn't fail if the path doesn't exist, and we use stock duckdb

vortex-duckdb/src/e2e_test/vortex_scan_test.rs
myrrc: No, because duckdb sends a HEAD followed a GET, all with retries, there may be more than two requests inflight. We need around ten because this is a one-shot serve, and we have at least two options (httplib, curl) plus multiple requests.

.github/workflows/bench-pr.yml
myrrc: It's used in 3 ways:
1. Just download curl if we've changed build.rs
2. Download curl and cmake and ninja ultimately
3. Download curl and also use "build.rs changed" trigger for other stuff (sql benchmarks).

I don't think generalizing would make it more pretty.

vortex-duckdb/src/filesystem.rs
```diff
@@ -159,7 +159,7 @@ impl FileSystem for DuckDbFileSystem {
 
     async fn open_read(&self, path: &str) -> VortexResult<Arc<dyn VortexReadAt>> {
         let mut url = self.base_url.clone();
-        url.set_path(&format!("{}/{}", url.path(), path));
+        url.set_path(path);
```
0ax1, line 162: Should we do the same thing for `list`?

Currently:
```
        if !prefix.is_empty() {
            directory_url.set_path(&format!("{}/{}", directory_url.path(), prefix));
        }
```

.github/workflows/ci.yml
```diff
@@ -287,6 +301,17 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Install nightly for fmt
         run: rustup toolchain install $NIGHTLY_TOOLCHAIN --component rustfmt
+      - uses: dorny/paths-filter@v3
```
0ax1, line 313: If I have a CI task where I set the `DUCKDB_VERSION` via env var, libcurl won't install and is going to break right. 

And what happens if I have 2 successive builds on CI where build.rs doesn't change and I want to build from source for both of them?

myrrc: > libcurl won't install

Yes, so I suggest using only build.rs changes for a CI build.

> I have 2 successive builds on CI 

The job checks PR changes, so multiple commits will be fine as long as the PR itself has the change

.github/workflows/sql-benchmarks.yml
0ax1: I don't think we actually need to do this copy at all. I think the only reason here that we need ddb is for tpc-ds generation: 
```
    let result = Command::new("duckdb")
        .arg("-c")
        .arg(&sql_script)
```
That should never require the ddb from hash.

.github/workflows/ci.yml
```diff
@@ -770,11 +817,21 @@ jobs:
...
+        id: duckdb-buildfile-changed
+        with:
+          filters: |
+            changed:
+              - 'vortex-duckdb/build.rs'
+      - name: install curl dev headers for duckdb source build
+        if: steps.duckdb-buildfile-changed.outputs.changed == 'true'
+        run: |
+            sudo apt update
+            sudo apt install -y libcurl4-gnutls-dev
```
gatesn: Does DuckDB not manage fetching these itself... seems odd since it does for all other dependencies?

.github/workflows/ci.yml
0ax1: How would this work for an upcoming DuckDB version which has no release yet. We need to set the commit hash in the build.rs and "build every time" (ignore that we'll fetch cached artefacts for now). So successive PRs with unchanged build.rs would need libcurl to be available.

> Yes, so I suggest using only build.rs changes for a CI build.

The question is how can we guard against that breaking though? If we go this route this should be documented somewhere.

.github/workflows/ci.yml
```diff
@@ -411,6 +436,17 @@ jobs:
         uses: taiki-e/install-action@v2
         with:
           tool: nextest
+      - uses: dorny/paths-filter@v3
+        id: duckdb-buildfile-changed
+        with:
+          filters: |
+            changed:
+              - 'vortex-duckdb/build.rs'
+      - name: install curl dev headers for duckdb source build
```
0ax1: Convention for names is uppercasing the first letter => ` Install cURL dev headers for DuckDB source build`

.github/workflows/ci.yml
```diff
@@ -475,10 +511,10 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           toolchain: ${{ env.NIGHTLY_TOOLCHAIN }}
           components: "rust-src, rustfmt, clippy, llvm-tools-preview"
-      - name: Install build dependencies
+      - name: install build dependencies
```
0ax1: nit: install => Install

0ax1 (commented):
My intiution is the following here: Relying on `build.rs` changing is a bit brittle. I agree that we don't want to spend the extra time on CI installing libcurl but I think the way to fully solve this is setting up custom AMIs in the mid run where libcurl would be preinstalled. Until then, I think it's fine installing libcurl unconditionally.

.github/workflows/ci.yml
0ax1: No annoyingly its not self-contained wrt libcurl. So there's no option to fetch libcurl as part of the ddb build.

.github/workflows/sql-benchmarks.yml
myrrc: Ok, makes sense.

vortex-duckdb/src/filesystem.rs
myrrc: Yes!

.github/workflows/ci.yml
myrrc: > Successive PRs with unchanged build.rs would need libcurl to be available

My opinion is that we should never depend on a pre-release version in stable Vortex, thus such CIs would be only "check new stuff, fix code that breaks beforehand", we can gate this under a failing check i.e. if you changed the hash, you can't merge up intil you revert it

.github/workflows/ci.yml
myrrc: I think they just didn't manage to do this yet. httpfs links curl statically (it's present in vcpkg.json), but if you build duckdb wih static httpfs, httpfs's vcpkg.json is ignored. I've opened the issue https://github.com/duckdb/duckdb/issues/21073, and we may adapt this if they decide to resolve

.github/workflows/ci.yml
myrrc: Still, checks about changes in file are also ugly so I suggest leaving this as for now.

0ax1:
@robert3005 Thoughts on the build.rs check approach?

.github/workflows/ci.yml
0ax1: > Still, checks about changes in file are also ugly

Def.

robert3005:
I would keep this simple. The thing you care about is whether the overall runtime of ci or benchmarks then we can make it better. If `apt` is slow you can get the libcurl source archive directly and unpack it in the right spot.

0ax1 (approved):
Let's land this so we can progress. But create a ticket for removing the build.rs state change check.

myrrc:
Regression https://github.com/vortex-data/vortex/actions/runs/22399016499/job/64840123013, reverting

PR #6664, opened 2026-02-24
feat: teach ALPRDArray how to append_to_builder

danking:
The PrimitiveBuilder API is a bit awkward for this. I needed two key things:

1. I want to treat a trailing slice of the builder's values as a different type. In particular, ALPRD decodes u32s (respectively u64s) and then converts them to f32s (resp. f64s).

2. I want to separately update the nulls and the values. In particular, ALPRD stores the nulls on the left parts whereas the right parts hold the values.

All of this is copacetic *when done on the UninitRange*, but I do it on the whole PrimitiveBuilder. I kinda think PrimitiveBuilder should be able to split itself into the initialized and uninitialized parts. In my opinion, it seems reasonable to transmute the builder of uninitialized parts since no one else is looking at that data anyway.


danking:
We should do this, but it segfaults currently and I don't have the time.

PR #6667, opened 2026-02-25
read future to skip polled and dropped events for completed requests

onursatici:
## Summary

if a request is complete as part of a coalesced request, we were emitting polled and dropped events for those requests as well. This is not too bad but does unnecessary bookkeeping, especially with aggressive coalescing, we end up fetching one request for 100s of segments


robert3005 (approved):
this whole io system needs a bit more test coverage

PR #6673, opened 2026-02-25
keep DropOnSync alive until after dispatching for cuda slices

onursatici:
## Summary

cuda device pointers come with `SyncOnDrop` that is used to synchronise reads and writes to the underlying cuda buffer. We should keep those alive until after dispatching the read or write work on them. For our single stream case this is fine, but if we were to have multiple streams accessing the same buffer these would be a problem

### DIFF
--- a/vortex-cuda/benches/dynamic_dispatch_cuda.rs
+++ b/vortex-cuda/benches/dynamic_dispatch_cuda.rs
@@ -51,14 +51,16 @@ const BENCH_ARGS: &[(usize, &str)] = &[
 /// Launch the dynamic_dispatch kernel and return GPU-timed duration.
 fn run_timed(
     cuda_ctx: &mut CudaExecutionCtx,
-    output_ptr: u64,
     array_len: usize,
+    output_buf: &CudaDeviceBuffer,
     device_plan: &Arc<cudarc::driver::CudaSlice<DynamicDispatchPlan>>,
     shared_mem_bytes: u32,
 ) -> VortexResult<Duration> {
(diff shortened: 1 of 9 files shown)


a10y:
I thought the ones returned from `device_ptr` don't sync and just record events?

onursatici:
yes they only record events, but those events are internal to the cuda view and they are used for synchronising other read or write operations on the same slice

robert3005:
So is there a practical thing to change here or is this something that is useful when combined with another change?

0ax1:
@onursatici I think this lgtm, but would love to review this tomorrow morning with a fresh mind.

onursatici:
I believe we don't use the same cuda slice from multiple cuda streams yet so this won't fix anything yet, but it does future proof us for that world

0ax1 (approved):

0ax1:
Unrelated to the changes in your PR I think there's sth odd about adding a sync/record event on drop for each `device_ptr` call. The drop also doesn't enforce that record event is set right after the launch. The other thing here being that launch kernel can record events as part of the launch.

0ax1:
The other thing to note is that cudarc sets events for each cudaview / cudaslice as part of the launch. So as long we enable recording events for launches we can check the events on each cudaview/cudaslice without recording additional events I think. Maybe it's worth considering going the opposite route relying on launch events, and provide a device ptr retrieval wrapper that bypasses recording events. Or I guess we'd need one event per split, once we start parallelizing splits across streams.

PR #6675, opened 2026-02-25
vortex read at for pooled cuda reads

onursatici:
## Summary

Adds two new `VortexReadAt` implementations, for files and object store, that uses a buffer pool that is page locked for fast h2d transfers


vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,379 @@
...
+    /// Returns the buffer as a mutable slice.
+    pub fn as_mut_slice(&mut self) -> VortexResult<&mut [u8]> {
+        self.inner
+            .as_mut_slice()
+            .map(|slice| &mut slice[..self.logical_len])
+            .map_err(|e| vortex_err!("failed to access pinned host buffer: {e}"))
+    }
+
+    /// Returns a raw pointer to the buffer.
+    pub fn as_ptr(&self) -> VortexResult<*const u8> {
```
0ax1: For symmetry, `as_mut_ptr`?

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,379 @@
...
+    ) -> VortexResult<Self> {
+        let inner = unsafe {
+            ctx.alloc_pinned::<u8>(capacity)
+                .map_err(|e| vortex_err!("failed to allocate pinned host buffer: {e}"))?
+        };
+        Ok(Self { inner, logical_len })
+    }
+
+    /// Returns the length of the buffer in bytes.
+    pub fn len(&self) -> usize {
```
0ax1: For all APIs introduced in this PR, maybe worth double checking whether they should be pub or pub(crate).

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,379 @@
...
+
+use crate::CudaDeviceBuffer;
+use crate::stream::VortexCudaStream;
+
+/// A page-locked host buffer allocated by CUDA.
+///
+/// This is intended as a staging buffer for H2D transfers. Contents are uninitialized after
+/// allocation.
+pub struct PinnedByteBuffer {
+    inner: PinnedHostSlice<u8>,
```
0ax1, line 27: We only wrap `PinnedHostSlice` bc of adding a logical length right?

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,379 @@
...
+            allocs: self.allocs.load(std::sync::atomic::Ordering::Relaxed),
+            puts: self.puts.load(std::sync::atomic::Ordering::Relaxed),
+        }
+    }
+
+    fn size_class_len(&self, len: usize) -> usize {
+        if len == 0 { 0 } else { len.next_power_of_two() }
+    }
+
+    fn reclaim_completed(&self) -> VortexResult<()> {
```
0ax1, line 188: This one deserves a doc comment.

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,379 @@
...
+        let mut inflight = self.inflight.lock();
+        if inflight.is_empty() {
+            return Ok(());
+        }
+        self.ctx
+            .bind_to_thread()
+            .map_err(|e| vortex_err!("Failed to bind CUDA context: {e}"))?;
+        let mut idx = 0usize;
+        while idx < inflight.len() {
```
0ax1, line 197: Could be `0..len().filter()`, rather than mut idx with while loop. Maybe weird though doing 
```
 let completed = inflight.swap_remove(idx);
 self.put(completed.buffer);
```
in an iterator.

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,379 @@
...
+            self.put(completed.buffer);
+        }
+        Ok(())
+    }
+
+    fn get_inner(&self, len: usize) -> VortexResult<PinnedByteBuffer> {
+        self.reclaim_completed()?;
+        let key_len = self.size_class_len(len);
+        {
+            let mut buckets = self.buckets.lock();
```
0ax1, line 212: I guess we don't have so much contention that we would care about lock free?

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,379 @@
...
+        inner
+            .as_mut_slice()
+            .unwrap_or_else(|e| vortex_panic!("failed to access pinned host buffer: {e}"))
+    }
+
+    /// Submits a non-blocking H2D DMA transfer and returns a device buffer.
+    ///
+    /// The pinned buffer is placed in the pool's inflight queue, gated on a `CudaEvent` marking
+    /// the transfer completion. The pool reclaims it once the event fires.
+    pub fn transfer_to_device(
```
0ax1, line 310: I think it'd be good to unit test this. In particular respective: `self.pool.put_inflight(event, inner)?;`.

vortex-cuda/src/pooled_read_at.rs
```diff
@@ -0,0 +1,345 @@
...
+use vortex::error::vortex_ensure;
+use vortex::error::vortex_err;
+use vortex::io::CoalesceConfig;
+use vortex::io::VortexReadAt;
+use vortex::io::runtime::Handle;
+
+use crate::pinned::PinnedByteBufferPool;
+use crate::stream::VortexCudaStream;
+
+const FILE_COALESCING_CONFIG: CoalesceConfig = CoalesceConfig {
```
0ax1: Why do these diverge from https://github.com/vortex-data/vortex/commit/74c0bcf615bc3c1717014ec7b45f2211b4d7c2c8? Or asked diff can we add a rationale why we picked these numbers?

0ax1 (commented):
Couple of remarks inline. Main blocker being that there are no tests for the new logic yet.

vortex-cuda/src/pinned.rs
0ax1: I assume there's no way to force_set a len on `PinnedHostSlice` without changing capacity.

vortex-cuda/src/stream.rs
```diff
@@ -86,6 +87,20 @@ impl VortexCudaStream {
...
+    pub(crate) fn memcpy_htod<H: HostSlice<u8>>(
+        &self,
+        src: &H,
+        dst: &mut CudaSlice<u8>,
+    ) -> Result<(), cudarc::driver::DriverError> {
+        self.0.memcpy_htod(src, dst)
+    }
+
+    /// Records a CUDA event on this stream.
+    pub(crate) fn record_event(&self) -> Result<CudaEvent, cudarc::driver::DriverError> {
```
0ax1: Should the AP allowI to pass different event record flags rather than always going with `CU_EVENT_DISABLE_TIMING`?
```
sys::CUevent_flags
```

vortex-cuda/src/pinned.rs
onursatici: I think `swap_remove` makes that a bit tricky to get right, if we hit that path, we don't want to increment the index because the previously last element now is at that current index, and we want to check that in the next iteration

vortex-cuda/src/pooled_read_at.rs
```diff
@@ -0,0 +1,345 @@
...
+                    },
+                )
+                .await?;
+
+            match response.payload {
+                #[cfg(not(target_arch = "wasm32"))]
+                GetResultPayload::File(file, _) => {
+                    target = handle
+                        .spawn_blocking(move || {
+                            read_exact_at(&file, target.as_mut_slice(), range.start)?;
```
0ax1, line 242: This ends up in `PinnedHostSlice` and will call sync on each `as_mut_slice` call.

```
impl<T: ValidAsZeroBits> PinnedHostSlice<T> {
    pub fn as_mut_slice(&mut self) -> Result<&mut [T], DriverError> {
        self.event.synchronize()?;
        Ok(unsafe { std::slice::from_raw_parts_mut(self.ptr, self.len) })
    }
```

vortex-cuda/src/pinned.rs
onursatici: yea I think the transfer cost should outweigh the lock overhead here by a lot, so I was thinking it should be fine to just lock

vortex-cuda/src/pooled_read_at.rs
```diff
@@ -0,0 +1,345 @@
...
+        &self,
+        offset: u64,
+        length: usize,
+        _alignment: Alignment,
+    ) -> BoxFuture<'static, VortexResult<BufferHandle>> {
+        let store = Arc::clone(&self.store);
+        let path = self.path.clone();
+        let handle = self.handle.clone();
+        let stream = self.stream.clone();
+        let mut target = match self.pool.get(length) {
```
0ax1: Should `pool.get` be eager like now or make it lazy somehow moving it into the closure? 

```
        let mut target = match self.pool.get(length) {
            Ok(target) => target,
            Err(err) => return async move { Err(err) }.boxed(),
        };
```

vortex-cuda/src/pinned.rs
onursatici: yea exactly

vortex-cuda/src/pooled_read_at.rs
0ax1: Oh but it's not a wait on the stream but on the event associated with `PinnedHostSlice`.

vortex-cuda/src/pooled_read_at.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+/// Read exactly `buffer.len()` bytes from `file` starting at `offset`.
+#[cfg(not(target_arch = "wasm32"))]
+fn read_exact_at(file: &File, buffer: &mut [u8], offset: u64) -> io::Result<()> {
+    #[cfg(unix)]
+    {
+        file.read_exact_at(buffer, offset)
+    }
+    #[cfg(windows)]
```
0ax1: 😢

vortex-cuda/src/pooled_read_at.rs
onursatici: ah now we use a separate config for nvmes, this 1MB and 4MB is the same as the latest std-file read_at impl we have. That constant there is not exposed from the crate. I agreed though I think I will add a `CoalesceConfig::file` so both of these can use that

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,385 @@
...
+    ///
+    /// # Safety
+    /// The returned buffer's contents are uninitialized. The caller must initialize before read.
+    pub(crate) unsafe fn uninit_with_capacity(
+        ctx: &Arc<CudaContext>,
+        capacity: usize,
+        logical_len: usize,
+    ) -> VortexResult<Self> {
+        let inner = unsafe {
+            ctx.alloc_pinned::<u8>(capacity)
```
0ax1, line 45: Maybe interesting to add that the memory gets allocated with `           sys::CU_MEMHOSTALLOC_WRITECOMBINED` which is fast to transfer but much slower to read from the host.

vortex-cuda/src/pinned.rs
```diff
@@ -0,0 +1,385 @@
...
+            .memcpy_htod(&inner, &mut cuda_slice)
+            .map_err(|e| vortex_err!("Failed to schedule H2D copy: {}", e))?;
+
+        let event = Arc::new(
+            stream
+                .record_event()
+                .map_err(|e| vortex_err!("Failed to record CUDA event: {}", e))?,
+        );
+
+        self.pool.put_inflight(event, inner)?;
```
0ax1, line 340: Do we leak anything in the error case (and do we care)?

vortex-cuda/src/pooled_read_at.rs
onursatici: gotta scan vortex into gpu from those gaming rigs

vortex-cuda/src/pooled_read_at.rs
```diff
@@ -0,0 +1,345 @@
...
+            let cuda_buf = target.transfer_to_device(&stream)?;
+            Ok(BufferHandle::new_device(Arc::new(cuda_buf)))
+        }
+        .boxed()
+    }
+}
+
+/// Read exactly `buffer.len()` bytes from `file` starting at `offset`.
+#[cfg(not(target_arch = "wasm32"))]
+fn read_exact_at(file: &File, buffer: &mut [u8], offset: u64) -> io::Result<()> {
```
0ax1: We have an identical impl in vortex-io. We'd only need to make it pub to use it here.

vortex-cuda/src/pinned.rs
0ax1: Where much slower being 10-100x.

vortex-cuda/src/pooled_read_at.rs
onursatici: yea I didn't want to add the vortex-io dep to vortex-cuda just for sharing that

vortex-cuda/src/pooled_read_at.rs
0ax1: addressed by: https://github.com/vortex-data/vortex/pull/6690

vortex-cuda/src/pooled_read_at.rs
onursatici: I guess I can it through vortex facade

vortex-cuda/src/pooled_read_at.rs
0ax1: 🌈

0ax1 (approved):

vortex-cuda/src/pooled_read_at.rs
onursatici: yes! missed this, thanks

vortex-cuda/src/pinned.rs
onursatici: yes good call. We would leak the slot in the pool, fixed

PR #6679, opened 2026-02-25
ScalarFn VTable

gatesn:
The vtable itself describes the behavior of a scalar function. I envisage us turning Expression into an enum in the future where ScalarFn is but one variant. 

This PR addresses the naming confusion (and prepares us to add AggregateFn VTable in the near future), as well as unifies the vtable names as per #6093


gatesn:
@claude review this PR

vortex-array/src/expr/exprs/binary/mod.rs
```diff
@@ -282,295 +289,25 @@ impl VTable for Binary {
...
-///         .into_array()
-///         .into_arrow_preferred()
-///         .unwrap()
-/// );
-/// ```
-pub fn checked_add(lhs: Expression, rhs: Expression) -> Expression {
-    Binary
-        .try_new_expr(Operator::Add, [lhs, rhs])
-        .vortex_expect("Failed to create Add binary expression")
-}
```
robert3005, line 559: what's the replacement for these?

gatesn: I kept them in the exprs module since they create expressions:
https://github.com/vortex-data/vortex/pull/6679/changes#diff-d1b532c524303a975ccf98376ed880c0b6e4bab2fce91e99157aac49f9c30e66

robert3005: ooo, thanks for pointing it out

robert3005 (approved):

PR #6680, opened 2026-02-25
Extension Scalar VTable

connortsui20:
## Summary

Adds methods related to extension scalars to the `ExtVTable`.

This then required updating existing implementors of `ExtVTable` (only datetime and mock extension test types).

I think this is a nice step before the much more intrusive change to `ScalarValue` where we add the `Extension` variant.

## API Changes

This is a breaking change to `ExtVTable`.

## Testing

TODO

### DIFF
--- a/vortex-array/src/arrays/extension/compute/rules.rs
+++ b/vortex-array/src/arrays/extension/compute/rules.rs
@@ -81,7 +81,7 @@ mod tests {
     struct TestExt;
     impl ExtVTable for TestExt {
         type Metadata = EmptyMetadata;
-        type Value<'a> = &'a str;
+        type NativeValue<'a> = &'a str;
 
         fn id(&self) -> ExtId {
             ExtId::new_ref("test_ext")
@@ -112,12 +112,12 @@ mod tests {
             Ok(())
         }
 
-        fn unpack<'a>(
+        fn unpack_native<'a>(
             &self,
             _metadata: &'a Self::Metadata,
             _storage_dtype: &'a DType,
             _storage_value: &'a ScalarValue,
-        ) -> Self::Value<'a> {
+        ) -> Self::NativeValue<'a> {
             ""
         }
     }
@@ -192,7 +192,7 @@ mod tests {
         struct TestExt2;
         impl ExtVTable for TestExt2 {
             type Metadata = EmptyMetadata;
-            type Value<'a> = &'a str;
+            type NativeValue<'a> = &'a str;
 
             fn id(&self) -> ExtId {
                 ExtId::new_ref("test_ext_2")
@@ -223,12 +223,12 @@ mod tests {
                 Ok(())
             }
 
-            fn unpack<'a>(
+            fn unpack_native<'a>(
                 &self,
                 _metadata: &'a Self::Metadata,
                 _storage_dtype: &'a DType,
                 _storage_value: &'a ScalarValue,
-            ) -> Self::Value<'a> {
+            ) -> Self::NativeValue<'a> {
                 ""
             }
         }
--- a/vortex-array/src/arrays/extension/compute/rules.rs
+++ b/vortex-array/src/arrays/extension/compute/rules.rs
@@ -103,22 +103,13 @@ mod tests {
             Ok(())
         }
 
-        fn validate_scalar_value(
-            &self,
-            _metadata: &Self::Metadata,
-            _storage_dtype: &DType,
-            _storage_value: &ScalarValue,
-        ) -> VortexResult<()> {
-            Ok(())
-        }
-
         fn unpack_native<'a>(
             &self,
             _metadata: &'a Self::Metadata,
             _storage_dtype: &'a DType,
             _storage_value: &'a ScalarValue,
-        ) -> Self::NativeValue<'a> {
-            ""
+        ) -> VortexResult<Self::NativeValue<'a>> {
+            Ok("")
         }
     }
 
@@ -214,22 +205,13 @@ mod tests {
                 Ok(())
             }
 
-            fn validate_scalar_value(
-                &self,
-                _metadata: &Self::Metadata,
-                _storage_dtype: &DType,
-                _storage_value: &ScalarValue,
-            ) -> VortexResult<()> {
-                Ok(())
-            }
-
             fn unpack_native<'a>(
                 &self,
                 _metadata: &'a Self::Metadata,
                 _storage_dtype: &'a DType,
                 _storage_value: &'a ScalarValue,
-            ) -> Self::NativeValue<'a> {
-                ""
+            ) -> VortexResult<Self::NativeValue<'a>> {
+                Ok("")
             }
         }
 
--- a/vortex-array/src/arrays/extension/compute/rules.rs
+++ b/vortex-array/src/arrays/extension/compute/rules.rs
@@ -87,11 +87,11 @@ mod tests {
             ExtId::new_ref("test_ext")
         }
 
-        fn serialize(&self, _metadata: &Self::Metadata) -> VortexResult<Vec<u8>> {
+        fn serialize_metadata(&self, _metadata: &Self::Metadata) -> VortexResult<Vec<u8>> {
             Ok(vec![])
         }
 
-        fn deserialize(&self, _data: &[u8]) -> VortexResult<Self::Metadata> {
+        fn deserialize_metadata(&self, _data: &[u8]) -> VortexResult<Self::Metadata> {
             Ok(EmptyMetadata)
         }
 
@@ -189,11 +189,11 @@ mod tests {
                 ExtId::new_ref("test_ext_2")
             }
 
-            fn serialize(&self, _metadata: &Self::Metadata) -> VortexResult<Vec<u8>> {
+            fn serialize_metadata(&self, _metadata: &Self::Metadata) -> VortexResult<Vec<u8>> {
                 Ok(vec![])
             }
 
-            fn deserialize(&self, _data: &[u8]) -> VortexResult<Self::Metadata> {
+            fn deserialize_metadata(&self, _data: &[u8]) -> VortexResult<Self::Metadata> {
                 Ok(EmptyMetadata)
             }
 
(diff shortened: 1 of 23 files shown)


vortex-array/src/arrays/extension/compute/rules.rs
```diff
@@ -75,23 +75,51 @@ mod tests {
     use crate::extension::EmptyMetadata;
     use crate::optimizer::ArrayOptimizer;
     use crate::scalar::Scalar;
+    use crate::scalar::ScalarValue;
 
     #[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
     struct TestExt;
     impl ExtVTable for TestExt {
         type Metadata = EmptyMetadata;
+        type Value<'a> = &'a str;
```
gatesn: I would maybe rename this `NativeValue` so it's obvious it's the in-memory representation

vortex-array/src/scalar/extension/typed.rs
```diff
@@ -0,0 +1,185 @@
...
+        self.0.vtable.id()
+    }
+
+    /// Returns a reference to the [`ExtVTable`] for this extension scalar.
+    pub fn vtable(&self) -> &V {
+        &self.0.vtable
+    }
+
+    /// Returns a reference to the underlying storage [`ScalarValue`].
+    pub fn storage_value(&self) -> &ScalarValue {
```
gatesn, line 158: I think you should also add `pub fn native_value(&self) -> V::Native` - just because we can!

vortex-array/src/scalar/tests/casting.rs
```diff
@@ -28,18 +29,45 @@ mod tests {
...
+        fn validate_scalar_value(
+            &self,
+            _metadata: &Self::Metadata,
+            _storage_dtype: &DType,
+            _storage_value: &ScalarValue,
+        ) -> VortexResult<()> {
+            Ok(())
+        }
+
+        fn unpack<'a>(
```
gatesn: `unpack_native`

vortex-array/src/scalar/tests/casting.rs
```diff
@@ -28,18 +29,45 @@ mod tests {
...
+
         fn validate_dtype(
             &self,
             _options: &Self::Metadata,
             _storage_dtype: &DType,
         ) -> VortexResult<()> {
             Ok(())
         }
+
+        fn validate_scalar_value(
```
gatesn: Is this not the same as making `unpack_native` return a VortexResult? Seems like it's easier to merge validation and parsing into the same function, instead of unpack having to use vortex_expect on all the parsing errors.

vortex-array/src/scalar/typed_view/extension/tests.rs
```diff
@@ -90,18 +118,45 @@ fn test_ext_scalar_partial_ord_different_types() {
     struct TestExt2;
     impl ExtVTable for TestExt2 {
         type Metadata = EmptyMetadata;
+        type Value<'a> = &'a str;
```
gatesn: I think the "default" / lazy version of this is for `type Value<'a> = &'a ScalarValue`, that way you still expose the right information but you do so without any validation.

gatesn (commented):
Looking nice

connortsui20:
I added some TODOs and concerns, could you take a look at the most recent commit @gatesn?

vortex-array/src/scalar/tests/casting.rs
connortsui20: I agree, and right now it makes things a lot harder to reason about

vortex-array/src/scalar/typed_view/extension/tests.rs
connortsui20: unfortunately assoc type defaults are unstable https://github.com/rust-lang/rust/issues/29661, and I would guess that GAT defaults are even further away.

As for this test, this is just the easiest way to get something to compile since it doesn't actually test the scalar logic. I will add some tests specific to extension scalars once I clean this PR up

vortex-array/src/scalar/tests/casting.rs
connortsui20: Actually I think a counterargument is that a lot of the internal things want to validate the scalar value without creating a native value

vortex-array/src/scalar/typed_view/extension/tests.rs
gatesn: I don't mean actually default it, I just mean in your tests you seem to return an empty `""` from unpack. You should just return the ScalarValue instead.

vortex-array/src/scalar/tests/casting.rs
gatesn: I agree, but typically creating the native value is such small overhead vs validation it doesn't matter. In the future if it _does_ matter, then we can add a separate validate function that defaults to unpack and drops the result. So I'd say fuse them for now

vortex-array/src/dtype/extension/vtable.rs
```diff
@@ -6,40 +6,67 @@ use std::fmt::Display;
...
+    ///
+    /// The value only represents non-null values. We denote nullable values as `Option<Value>`.
+    type NativeValue<'a>: Display;
+
     /// Returns the ID for this extension type.
     fn id(&self) -> ExtId;
 
+    // Methods related to the extension `DType`.
+
+    // TODO(connor): Should probably be called `serialize_metadata`.
```
gatesn: Yah go for it

vortex-array/src/extension/datetime/time.rs
```diff
@@ -63,6 +97,61 @@ impl ExtVTable for Time {
...
+
+    fn unpack_native(
+        &self,
+        metadata: &Self::Metadata,
+        _storage_dtype: &DType,
+        storage_value: &ScalarValue,
+    ) -> Self::NativeValue<'_> {
+        match metadata {
+            TimeUnit::Seconds => {
+                TimeValue::Seconds(storage_value.as_primitive().cast::<i32>().vortex_expect(
```
gatesn: You see what I mean that all these vortex_expects are == the errors in the validate function

vortex-array/src/scalar/extension/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+//! [`ScalarValue`]: crate::scalar::ScalarValue
+//! [`DType`]: crate::dtype::DType
+//! [`ExtDTypeRef`]: crate::dtype::extension::ExtDTypeRef
+//!
+//! ## File layout convention
+//!
+//! Note that there is a single unified vtable for working with extension types
+//! [`ExtVTable`](crate::dtype::extension::ExtVTable).
+
+// TODO(connor): Docs on file structure.
```
gatesn: Eh, I think the internal docs on vtables covers it

vortex-array/src/scalar/extension/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+
+// TODO(connor): Figure out what this should look like.
+// mod plugin;
+// pub use plugin::ExtScalarValuePlugin;
```
gatesn: I think in this case it's weird, we only need one Plugin for the ExtVTable, not one for DType and one for Scalar.

connortsui20: I still dont actually understand what the plugin is for even after reading the doc comment

connortsui20:
So I decided to keep the `validate_scalar_value`, but now it just calls `unpack_native.map(|_| ())`. This way if we see a perf problem later it is way easier to just implement this if we want to.

vortex-array/src/scalar/typed_view/extension/tests.rs
connortsui20: to be honest this seems kind of unnecessary for these tests? I will do that for the tests I'm planning to add now

connortsui20:
I am still concerned about some of the trait impls on `ExtScalarValueRef`, right now it basically just delegates things like Eq, Ord, and Hash to the storage value. I don't think we want this? We probably want the implementor to implement this manually right?

gatesn:
> I am still concerned about some of the trait impls on ExtScalarValueRef, right now it basically just delegates things like Eq, Ord, and Hash to the storage value. I don't think we want this? We probably want the implementor to implement this manually right?

Your instinct is correct! But we can do that in the future. We should do it alongside arrays so we can define the vectorized version of this at the same time, otherwise scalar and array semantics would differ.

vortex-array/src/scalar/extension/mod.rs
gatesn: The "plugin" is the thing that lives on the session. It's what allows us to deserialize things into in-memory objects. We lookup the plugin by ID, then call its deserialize function. 

It basically encapsulates extension dtype behavior that doesn't have an instance object to operate on.

vortex-array/src/scalar/extension/erased.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub fn storage_value(&self) -> &ScalarValue {
+        self.0.storage_value()
+    }
+
+    /// Formats the extension scalar using the provided [`ExtDTypeRef`] for metadata context.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying [`fmt::Write`] operation fails.
+    pub fn fmt_ext_scalar(
```
gatesn: I don't know how much of this we want on the public API, vs being pub(super) and just calling into this from the ExtensionScalar<'a> struct.

vortex-array/src/scalar/extension/erased.rs
```diff
@@ -0,0 +1,173 @@
...
+            })
+            .vortex_expect("Failed to downcast ExtScalar")
+    }
+
+    /// Attempts to downcast the vtable to a concrete [`ExtVTable`] type by reference.
+    ///
+    /// Unlike [`try_downcast`], this borrows rather than consuming `self`.
+    ///
+    /// [`try_downcast`]: ExtScalarValueRef::try_downcast
+    pub fn try_get_vtable<V: ExtVTable>(&self) -> Option<&V> {
```
gatesn: Are these vtable accessors used anywhere? Remove any public functions from the API that aren't yet used by anything

vortex-array/src/scalar/extension/erased.rs
connortsui20: I guess the getters for the vtable should probably also be removed?

connortsui20: oh saw the other comment

gatesn (approved):

PR #6681, opened 2026-02-26
cuda ctx bind to thread before unsafe

onursatici:
set the cuda context thread locals before unsafe cudarc calls. These lower level calls do not ensure the thread local cuda context some calls expect. cudarc's safe methods do this internally.

This PR also wraps the zstd buffers to use launch_external, matching other kernels

### DIFF
--- a/vortex-cuda/src/kernel/encodings/zstd_buffers.rs
+++ b/vortex-cuda/src/kernel/encodings/zstd_buffers.rs
@@ -106,7 +106,6 @@ async fn decode_zstd_buffers(
     let mut device_statuses: CudaSlice<nvcompStatus_t> = ctx.device_alloc(plan.num_frames())?;
     let mut nvcomp_temp_buffer: CudaSlice<u8> = ctx.device_alloc(nvcomp_temp_buffer_size)?;
     let stream = ctx.stream();
-
     let frame_ptrs_view = frame_ptrs_handle.cuda_view::<u64>()?;
     let frame_sizes_view = frame_sizes_handle.cuda_view::<usize>()?;
     let output_sizes_view = output_sizes_handle.cuda_view::<usize>()?;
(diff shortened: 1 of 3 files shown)


robert3005 (approved):

vortex-cuda/src/kernel/encodings/zstd_buffers.rs
```diff
@@ -105,51 +101,31 @@ async fn decode_zstd_buffers(
...
-        record_temp,
-        record_statuses,
-    ));
+    let frame_ptrs_ptr = device_ptr!(frame_ptrs_handle, u64);
+    let frame_sizes_ptr = device_ptr!(frame_sizes_handle, usize);
+    let output_sizes_ptr = device_ptr!(output_sizes_handle, usize);
+    let output_ptrs_ptr = device_ptr!(output_ptrs_handle, u64);
+
+    let stream = ctx.stream();
+    ctx.launch_external(plan.output_size_total(), || {
```
0ax1, line 126: I think we can strip having to pass a len parameter to `launch_external`? @a10y Is that only for tracing rn?

vortex-cuda/src/stream.rs
```diff
@@ -66,13 +66,17 @@ impl VortexCudaStream {
     {
         let host_slice: &[T] = data.as_ref();
         let mut cuda_slice: CudaSlice<T> = self.device_alloc(host_slice.len())?;
-        let (device_ptr, record_write) = cuda_slice.device_ptr_mut(&self.0);
+        let device_ptr = cuda_slice.device_ptr_mut(&self.0).0;
+
+        self.0
```
0ax1, line 76: Can we add a comment why this is needed exactly? Should we bind the context by default and not repeatedly in diff places?

onursatici: I think I need to call this before each public method that we have an unsafe cuda call in. But setting it if it is already set is really fast, it is a thread local lookup and pointer comparison.

but yea in copy_to_device we would be doing this twice, one before the memcpy and the other before `launch_host_function`, but that is needed because the latter is also can be called with the pub(crate) `await_stream_callback` method

0ax1: Should we add a todo to wrap these fns such that we always set this, given the that the cost of doing this repeatedly is low?

0ax1 (approved):

vortex-cuda/src/kernel/encodings/zstd_buffers.rs
a10y: Yea that's right

PR #6683, opened 2026-02-26
Add explain feature to DF benchmarks

AdamGS:
## Summary

I've added this in an ad-hoc ways multiple times during development, and now I feel called out by @gatesn, so added it to both engines.

When using `--explain`, it prints the query plan instead of running the query


connortsui20:
What does this flag do? I didn't even know this was a thing before...

joseph-isaacs:
This is great did you test on all engines?

joseph-isaacs (approved):

PR #6684, opened 2026-02-26
Remove code for old benchmarks website

AdamGS:
## Summary

Just not sure why its still there? Removed the old code and renamed the new website


gatesn (approved):

palaska:
> ## Summary
> Just not sure why its still there? Removed the old code and renamed the new website

👍 I left it so we can switch to the old website with one DNS record change but I guess we have given enough soak time at this point.

PR #6686, opened 2026-02-26
Remove unnecessary to_canonical in DictArray::take

palaska:
No need to force a full decompression I think when calling take. FilterReduce correctly skips this step.


robert3005 (approved):

joseph-isaacs (commented):
I am not sure this is always better. Might be that we want to gc the values.

robert3005:
I think we need a heuristic to decide if we need gc. Right now I am pro removing all to_canonical and we can add them back later

PR #6691, opened 2026-02-26
Remove info log in execution path

AdamGS:
Just a small touch up I noticed in benchmarks, doesn't seem very useful

### DIFF
--- a/vortex-array/src/arrays/scalar_fn/vtable/operations.rs
+++ b/vortex-array/src/arrays/scalar_fn/vtable/operations.rs
@@ -33,7 +33,7 @@ impl OperationsVTable<ScalarFnVTable> for ScalarFnVTable {
 
         let scalar = match result.execute::<Columnar>(&mut ctx)? {
             Columnar::Canonical(arr) => {
-                tracing::info!(
+                tracing::trace!(
                     "Scalar function {} returned non-constant array from execution over all scalar inputs",
                     array.scalar_fn,
                 );


robert3005:
how would this happen? I think we have a suboptimal implementation?

AdamGS:
see https://github.com/vortex-data/vortex/actions/runs/22445976721/job/65000428354?pr=6683

```
Scalar function vortex.between(lower_strict: <=, upper_strict: <=) returned non-constant array from execution over all scalar inputs
```

AdamGS:
there's also an annoying filesystem-related log in duckdb, but I can live with it

vortex-array/src/arrays/scalar_fn/vtable/operations.rs
```diff
@@ -33,7 +33,7 @@ impl OperationsVTable<ScalarFnVTable> for ScalarFnVTable {
 
         let scalar = match result.execute::<Columnar>(&mut ctx)? {
```
a10y, line 34: do we need to keep executing() until we get back a Constant here? i've lost track of what the contract for this is now

a10y: ok nvm this should already be iterating

robert3005 (changes requested):
Please fix that function instead of changing this log

robert3005:
The problem is that we don't have between for ConstantArray, we should fix it

a10y:
I opened a new PR at https://github.com/vortex-data/vortex/pull/6693

PR #6693, opened 2026-02-26
fix constant dispatch for Between scalarfn

a10y:
Between should have a fast path for when its inputs are all constant, similar to other ScalarFns.

NOTE: I poked at several of the ScalarFn::execute methods and there doesn't seem to be any consistent style with how they're implemented.

Between already had special-casing support for Primitive and Decimal instead of looking up the appropriate `Between` kernel for the array. So I replicated that here.


vortex-array/public-api.lock
```diff
@@ -798,6 +798,10 @@ impl vortex_array::compute::SumKernel for vortex_array::arrays::ConstantVTable
 
 pub fn vortex_array::arrays::ConstantVTable::sum(&self, array: &vortex_array::arrays::ConstantArray, accumulator: &vortex_array::scalar::Scalar) -> vortex_error::VortexResult<vortex_array::scalar::Scalar>
 
+impl vortex_array::scalar_fn::fns::between::BetweenKernel for vortex_array::arrays::ConstantVTable
```
a10y, line 801: It is strange to me that these lines appear twice?

joseph-isaacs:
I'll write some doc for this tomorrow.

What was the reason for adding this?

I would want this to be added to the expr execute method directly, the pattern is roughly -- canonical + constant execute there and everything else as kernels like this

a10y:
The benchmark logs were spamming at INFO level complaining that Between on constant inputs was emitting non-constant result.

> I would want this to be added to the expr execute method directly, the pattern is roughly -- canonical + constant execute there and everything else as kernels like this

Yea I actually went with this approach first, but was concerned that

1. There might be other entrypoints that call the `BetweenKernel` that I wasn't aware of
2. Between dispatch already explicitly dispatches the Primitive/Decimal kernels, so this seemed to map onto that

vortex-array/src/scalar_fn/fns/between/mod.rs
```diff
@@ -300,7 +301,20 @@ impl ScalarFnVTable for Between {
             .try_into()
             .map_err(|_| vortex_err!("Expected 3 arguments for Between expression",))?;
 
-        // canonicalize the arr and we might be able to run a between kernels over that.
+        // Fast path for constant
+        if let Some(constant) = arr.as_opt::<ConstantVTable>()
```
robert3005, line 305: Ah, we need to dispatch this but constant is a special enough case

robert3005 (approved):

joseph-isaacs (changes requested):
This should be a reduce rule. Then it should be hooked in.

a10y:
Ok I'll go ahead and close this then, mind taking it?

joseph-isaacs:
I'll write the docs and then show you the current pattern, but happy to revise it

PR #6694, opened 2026-02-26
Minor benchmarks changes

AdamGS:
1. I might just be getting old, but making the font a bit bigger in the chart is nice
2. Highlight which dataset you're pointing at in the tooltip:
Screenshot 2026-02-26 at 17 08 08


palaska:
@claude review this PR and in addition, answer the following questions by analysing the changes and relevant code: Do we calculate the closest data point in anywhere else? Can we reuse the logic? Should we update the chart in the modal as well?

lwwmanning (approved):

PR #6695, opened 2026-02-26
Implement BetweenReduceRule for ConstantArray

robert3005:
This seems to be missing after migration and it's obviously valuable. Avoids
having to canonicalise result of between on constant array.

### DIFF
--- a/vortex-array/src/arrays/constant/compute/between.rs
+++ b/vortex-array/src/arrays/constant/compute/between.rs
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_error::VortexResult;
+
+use crate::Array;
+use crate::ArrayRef;
+use crate::IntoArray;
+use crate::arrays::ConstantArray;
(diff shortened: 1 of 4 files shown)


vortex-array/src/arrays/constant/compute/between.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::Array;
+use crate::ArrayRef;
+use crate::IntoArray;
+use crate::arrays::ConstantArray;
+use crate::arrays::ConstantVTable;
+use crate::scalar::Scalar;
+use crate::scalar_fn::fns::between::BetweenOptions;
+use crate::scalar_fn::fns::between::BetweenReduce;
+
+impl BetweenReduce for ConstantVTable {
```
a10y, line 15: i thought recently there was a change/going to be a change where ConstantArray would save the scalar into a buffer. so wouldn't this break the contract of reduce rules where they don't read buffers?

robert3005: Constants will need some more special handling, for a lot of optimisations we will need the value in both cpu and gpu so we can't JUST store it in the buffer

a10y (approved):

PR #6696, opened 2026-02-26
Implement extension scalars correct logic

connortsui20:
## Summary

This is the final step to having fully functional extension scalars.

Implements the correct logic (before it was just delegated to the storage values) for validating and displaying extension scalars.

Additionally, this implements the correct constructors for all the necessary extension scalar and scalar value types.

Theres also a small change here that turns `is_compatible` into `validate` which returns a `VortexResult` instead of a `bool`, I thought I had changed that before somewhere but it got lost.

## Testing

WIP


gatesn (approved):

connortsui20:
whoops, we dont actually need this because the vtable logic is on the dtype!

Issue #6544, opened 2026-02-16
Unify VTables across all concepts


a10y:
All of arrays/layout/scalarfn VTables should be unified (?) in some way. TBD.

gatesn:
# VTable Unification: Array Migration Plan

## Target Types

```rust
pub trait ArrayVTable: 'static + Sized + Send + Sync {
    type State: 'static + Send + Sync + Clone + Debug;
}

struct ArrayInner<V: ArrayVTable> {
    vtable: V,
    dtype: DType,
    len: usize,
    state: V::State,
    children: Vec<Option<ArrayRef>>,
    stats: ArrayStats,
}

pub struct Array<V: ArrayVTable>(Arc<ArrayInner<V>>);

pub struct ArrayRef(Arc<dyn DynArray>);
```

`ArrayInner<V>: DynArray`. Upcast from `Array<V>` to `ArrayRef` is free.

Constructors live on the vtable ZST: `Primitive::new(...) -> Array<Primitive>`.
No type aliases. Encoding-specific accessors use extension traits on `Array<V>`.

## Phases

### Phase 0: Rename `Array` trait → `DynArray`

Mechanical rename. Frees the `Array` name for the typed struct.

### Phase 1: Consolidate State

Define a `State` type per encoding holding encoding-specific in-memory
data (buffers, offsets, etc.) — everything not covered by `ArrayInner`'s
common fields or child slots. Done encoding-by-encoding.

### Phase 2: Introduce `ArrayInner<V>` + `Array<V>`, migrate encodings

Per encoding:

1. Rename vtable ZST (`PrimitiveVTable` → `Primitive`).
2. Replace bespoke array struct with `ArrayInner<V>`.
3. Define fixed child slot layout.
4. Add extension trait for typed accessors.
5. Move constructors to vtable ZST.
6. Update all call sites (clean break, no aliases).

### Phase 3: Update sub-vtable signatures

`ValidityVTable`, `OperationsVTable` methods change from `&V::Array` to
`&Array<V>`.

### Phase 4: Migrate erased layer

1. `ArrayInner<V>` implements sealed `DynArray` trait.
2. `ArrayRef` becomes concrete struct wrapping `Arc<dyn DynArray>`.
3. Move `impl dyn DynArray` methods to `impl ArrayRef`.
4. Remove old `DynArray` trait, `ArrayAdapter`, `vtable!` macro.
   Introduce `ArrayPlugin` for ID-based deserialization.

[...]

PR #6700, opened 2026-02-26
Inline visitor vtable to ArrayVTable

gatesn:
This PR moves the functions from VisitorVTable to ArrayVTable

See #6544


connortsui20 (approved):
also yolo

connortsui20:
can you also link a tracking issue to these? If it doesnt exist already?

PR #6705, opened 2026-02-27
feat[gpu]: slice support for CUDA dyn dispatch

0ax1:
Introduce slice support for the CUDA dyn dispatch kernel. As part of that, The bitpacking `SliceKernel`  is replaced with `SliceReduce`.

### DIFF
--- a/vortex-cuda/src/device_buffer.rs
+++ b/vortex-cuda/src/device_buffer.rs
@@ -81,8 +81,6 @@ mod private {
     }
 }
 
-// Get it back out as a View of u8
-
 impl CudaDeviceBuffer {
     /// Creates a new CUDA device buffer from a [`CudaSlice<T>`].
     ///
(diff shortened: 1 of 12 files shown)


0ax1:
(On that note: I think all remaining `SliceKernel` in vortex codebase can with `SliceReduce` => separate PR)

vortex-cuda/src/device_buffer.rs
```diff
@@ -281,7 +289,7 @@ impl DeviceBuffer for CudaDeviceBuffer {
 
     /// Slices the CUDA device buffer to a subrange.
     ///
-    /// **IMPORTANT**: this is a byte range, not elements range, due to the DeviceBuffer interface.
+    /// This is a byte range, not elements range, due to the DeviceBuffer interface.
     fn slice(&self, range: Range<usize>) -> Arc<dyn DeviceBuffer> {
```
0ax1, line 293: In some way I think this fn shouldn't exist on the API. Device pointers should always be passed sliced to GPU, with the offset being applied to the device ptr. Encodings that don't support element wise slicing like bitpacking use an extra offset parameter to locate them in the FL block.

vortex-cuda/src/device_buffer.rs
```diff
@@ -101,6 +99,16 @@ impl CudaDeviceBuffer {
         }
     }
 
+    /// Returns the byte offset within the allocated buffer.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+
+    /// Returns the adjusted device pointer accounting for the offset.
+    pub fn offset_ptr(&self) -> sys::CUdeviceptr {
```
0ax1, line 108: `device_ptr` is private, there's no public API anymore to get a device pointer without an offset.

0ax1:
implemented in: https://github.com/vortex-data/vortex/pull/6862

PR #6708, opened 2026-02-27
data parallel patching for FastLanes

a10y:
## Summary

This PR adds G-ALP style patches that allow for data parallel access. This allows removing the additional `execute_patches` kernel launch, and instead inserts patch values inside of the unpacking kernels, at the step when we dispatch from shared -> global memory.

The benchmarks added in #6712 show a significant speedup with this change, and performance is pretty much invariant to patch count.

image

### Background

The [G-ALP](https://dl.acm.org/doi/10.1145/3736227.3736242) paper's contribution is modifying the standard layout of "exceptions" (what we call Patches in Vortex) to allow for fully data parallel access. Their target is an f32 ALP decoding kernel, but it is equally applicable for our unpacking kernels.

The core insight is that storing patches in sorted order, which is great for single-threaded execution on a super-scalar CPU, results in very poor GPU performance. 

Instead, we need to shuffle them so that they can be accessed ordered by `(chunk, lane)`. The chunk is the normal FastLanes vector chunk size (1024). The lane depends on the width of the type. Doing this means that  we get O(1) access to the patches within our kernel.

Replicating Figure 2 from the paper:

image

This diff looks large, but in reality the big pieces

* Adds a new `patches.h` header file with shared definitions between the CUDA kernel and the Rust code.
* Adds a `patches.cuh` with some C++ code to make it easier to seek/iterate over a range of patch values
* Updates the unpacking kernels to accept an optional set of patches

This does not change the memory format for the `Patches` type, rather this just does a  D2H -> transpose on CPU -> H2D transformation.

## Testing

Uses the existing test suite which includes a lot of bit-unpacking with patches.


vortex-cuda/kernels/src/bit_unpack_16.cu
```diff
@@ -3,8 +3,9 @@
 #include <cuda_runtime.h>
 #include <stdint.h>
 #include "fastlanes_common.cuh"
+#include "patches.cuh"
 
-__device__ void _bit_unpack_16_0bw_lane(const uint16_t *__restrict in, uint16_t *__restrict out, uint16_t reference, unsigned int lane) {
+__device__ void _bit_unpack_16_0bw_lane(const uint16_t *__restrict in, uint16_t *__restrict out, uint16_t reference, unsigned int lane, GPUPatches patches) {
```
0ax1: We pass `GPUPatches patches` to fns where it is never used, like here.

vortex-cuda/kernels/src/bit_unpack_16.cu
```diff
@@ -862,315 +863,673 @@ __device__ inline void bit_unpack_16_lane(
...
-    for (int i = 0; i < 32; i++) {
-        auto idx = i * 32 + thread_idx;
-        out[idx] = shared_out[idx];
-    }
+    _bit_unpack_16_8bw_lane(in, shared_out, reference, thread_idx * 2 + 0, patches);
+    _bit_unpack_16_8bw_lane(in, shared_out, reference, thread_idx * 2 + 1, patches);
+        // Setup the patches cursor so we can seek patches
+        PatchesCursor cursor(patches);
+        // Each thread block in the unpack kernel unpacks 1 chunk of 1024 values
+        auto chunk = blockIdx.x;
```
0ax1: Don't we need to call `__syncthreads` after unpacking and before applying the patches?

0ax1 (commented):
Quick first glance.

vortex-cuda/kernels/src/bit_unpack_16.cu
a10y: ah you're right, i moved the patching to happen in the layer above when we read from shared memory, good catch

vortex-cuda/kernels/src/bit_unpack_16.cu
a10y: yea this is a good point, the previous code did not have syncthreads though and was working fine?

I had a hunch that this is b/c we size the threadblock to equal 1 warp, so all of the shared mem writes become immediately visible. Here is what Claude has to say:


Image

a10y: i can inject a `__syncwarp()` call in there

vortex-cuda/kernels/src/bit_unpack_16.cu
a10y: done

vortex-cuda/kernels/src/bit_unpack_16.cu
0ax1: Hmm, we gotta add an exception for 64bit types, as we use 16 threads (2 warps) there:
```
const fn bitpacked_thread_count(output_width: usize) -> u32 {
    if output_width == 64 { 16 } else { 32 }
}
```

vortex-cuda/kernels/src/bit_unpack_16.cu
```diff
@@ -885,292 +886,683 @@ __device__ inline void bit_unpack_16_lane(
     }
 }
 
-__device__ void _bit_unpack_16_0bw_32t(const uint16_t *__restrict in, uint16_t *__restrict out, uint16_t reference, int thread_idx) {
+__device__ void _bit_unpack_16_0bw_32t(const uint16_t *__restrict in, uint16_t *__restrict out, uint16_t reference, int thread_idx, GPUPatches patches) {
```
0ax1: We should pass `GPUPatches` by value (copy) to global entry point kernels but pass by ref `&` to all fns inside of that to avoid copying `GPUPatches`.

vortex-cuda/kernels/src/bit_unpack_16.cu
a10y: it's actually half a warp

a10y: this is another problem that i surfaced when playing around with nsight compute: all of the 64-bit kernels are launching warps with half of the slots empty. this means that we're kind of putting ourselves at a disadvantage for maxing out memory bw when unpacking to 64 bit types

vortex-cuda/cuda_kernel_generator/mod.rs
```diff
@@ -141,12 +141,37 @@ fn generate_device_kernel_for_width<T: FastLanes, W: Write>(
             writeln!(output, "_bit_unpack_{bits}_{bit_width}bw_lane(in, shared_out, reference, thread_idx * {per_thread_loop_count} + {thread_lane});")?;
         }
 
-        writeln!(output, "for (int i = 0; i < {shared_copy_ncount}; i++) {{")?;
         output.indent(|output| {
-            writeln!(output, "auto idx = i * {thread_count} + thread_idx;")?;
-            writeln!(output, "out[idx] = shared_out[idx];")
-        })?;
-        writeln!(output, "}}")
+            writeln!(output, "// Make sure all threads in the block/warp have completed writing")?;
```
0ax1: This pattern wasn't introduced by this PR but we could write out the generated code via a raw string rather than line by line.

E.g.:
```
let patch_block = format!(r#"
__syncwarp();
PatchesCursor cursor(patches);
auto chunk = blockIdx.x;
cursor.seek(chunk, thread_idx);
uint16_t next_patch_index = cursor.n_patches > 0 ? cursor.get_index() : 1024;
uint{bits}_t next_patch_value = cursor.n_patches > 0 ? cursor.get_value<uint{bits}_t>() : 0;
for (int i = 0; i < {shared_copy_ncount}; i++) {{
    auto idx = i * {thread_count} + thread_idx;
    if (idx == next_patch_index) {{
        out[idx] = next_patch_value;
        if (cursor.next()) {{
            next_patch_index = cursor.get_index();
            next_patch_value = cursor.get_value<uint{bits}_t>();
        }} else {{
            next_patch_index = 1024;
            next_patch_value = 0;
        }}
    }} else {{
        out[idx] = shared_out[idx];
    }}
}}"#);
```

a10y: yea i was thinking about doing that as well, makes it a lot easier to read/modify in the future 👍

vortex-cuda/kernels/src/patches.cuh
```diff
@@ -0,0 +1,65 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#pragma once
+
+#include "patches.h"
+#include "fastlanes_common.cuh"
+
+// Cursor that makes it easy to statefully iterate over a set of patches.
+struct PatchesCursor {
```
0ax1: The convention in C++ is that data only types are structs, and a class otherwise. A side effect of this is that fns and fields are by default now `private` and not `public`. So the public part now needs a 

```
public:
```
section.

vortex-cuda/kernels/src/bit_unpack_16.cu
0ax1: I guess a thing to try here could doubling the lane size for the GPU. Hitting the resolve button as this is not a sync issue.

vortex-cuda/kernels/src/patches.cuh
```diff
@@ -0,0 +1,65 @@
...
+    // The set of patches
+    GPUPatches patches;
+
+    uint32_t chunk_index;
```
0ax1: `Patches` and `PatchesCursor` should be as small as possible as each thread has its own copy that lives in registers. Spilling past the stack is very expensive in CUDA.

vortex-cuda/kernels/src/patches.cuh
```diff
@@ -0,0 +1,65 @@
...
+    // this is the offset into the indices.
+    uint32_t start_offset;
+    uint32_t offset;
+
+    __device__ PatchesCursor(GPUPatches p)
+        : patches(p), chunk_index(0), lane(0), n_patches(0), start_offset(0), offset(0) {
+    }
+
+    // Reset the cursor to the start of the patches range given at this.
+    __device__ void seek(uint32_t chunk, uint32_t theLane) {
```
0ax1: How about doing this as part of the constructor? We do this once per thread before applying the patches for the respective lane, right?

vortex-cuda/kernels/src/patches.cuh
```diff
@@ -0,0 +1,65 @@
...
+        chunk_index = chunk;
```
0ax1: `chunk_index` & `lane` do not need to be a field (or member in C++ lingo) as they are only used as a temp variable here but not to track state when calling `next`.

vortex-cuda/kernels/src/patches.cuh
```diff
@@ -0,0 +1,65 @@
...
+// Cursor that makes it easy to statefully iterate over a set of patches.
+struct PatchesCursor {
+    // The set of patches
+    GPUPatches patches;
+
+    uint32_t chunk_index;
+    uint32_t lane;
+    // this is number of patches for the currently configured chunk_index
+    // and lane and is set whenever seek() is called
+    uint32_t n_patches;
```
0ax1: This can be a u8, as one lane can have at max 64 patches.

vortex-cuda/kernels/src/patches.cuh
```diff
@@ -0,0 +1,65 @@
...
+
+    // this is the offset into the indices.
+    uint32_t start_offset;
+    uint32_t offset;
```
0ax1: `uint32_t offset;` We should always be able to fit this offset into an u8 right?

vortex-cuda/src/kernel/patches/types.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    fn len(&self) -> usize {
+        self.indices.len()
+    }
+}
+
+/// Transpose a set of patches from the default sorted layout into the data parallel layout.
+#[allow(clippy::cognitive_complexity)]
+pub async fn transpose_patches(
```
0ax1, line 144: Should this live in vortex-cuda? I guess this would also be useful for our CPU fastlanes encodings?

vortex-cuda/src/kernel/patches/types.rs
```diff
@@ -0,0 +1,294 @@
...
+
+#[cfg(test)]
+mod tests {
+    use vortex::buffer::BufferMut;
+    use vortex::buffer::buffer;
+
+    use crate::kernel::patches::types::transpose;
+
+    #[test]
+    fn test_transpose_patches() {
```
0ax1, line 239: How about throwing in a bunch more test scenarios via rstest?

vortex-cuda/kernels/src/patches.cuh
```diff
@@ -0,0 +1,65 @@
...
+
+    // Advance to the next patch in the patching group.
+    __device__ bool next() {
+        return ++offset < n_patches;
+    }
+
+    template <typename T>
+    __device__ T get_value() {
+        T *values = reinterpret_cast<T *>(patches.values);
+        return values[start_offset + offset];
```
0ax1: In terms of cache hits, how about bundling pairs of (value, key) rather than having 2 separate buffers?

0ax1 (commented):
Couple of thoughts around minimizing stack memory given that spilling registers is very expensive on the GPU.

Drafted a version which tries to minimize that: https://github.com/vortex-data/vortex/commit/6fd4e8091b994525c53d554ee469e343e1db7a39

Curious what you think if this is valuable to integrate as part of this PR.

a10y:
Yes that seems better for sure, I've cherrypicked that commit onto the branch now

vortex-cuda/kernels/src/bit_unpack_16.cu
a10y: done

vortex-cuda/kernels/src/patches.cuh
a10y: in this case i think it needs to stay as u16, as the index is in 0..1024 space, not row index

vortex-cuda/src/kernel/patches/types.rs
a10y: We'll move this out of vortex-cuda as part of creating PatchedArray. But for now it's only used in vortex-cuda so just keeping it here

0ax1 (approved):
Very neat, really nice to have this building block now!

Last remaining nit is the rstest stuff but approving from my end already.

PR #6685, opened 2026-02-26
use curl backend and duckdbfs for benchmarks

myrrc:
Another guess for a regression in #5767: try benchmarks with duckdb fs but with curl backend.

#5889

PR #6709, opened 2026-02-27
Fix S3 globbing for duckdbfs with httplib

myrrc:
Using vortex_filesystem=duckdb, both of httpfs backends are broken
when using globs on S3 files:

1. httplib was broken because duckdb folder list API returns full paths for
   S3 files, and we assumed relative paths. This PR fixes this behavior
2. curl was broken because of https://github.com/duckdb/duckdb-httpfs/pull/265
   in 1.4. Up until upstream fixes this, globbing won't work.
   This issue is not present in 1.5 anymore

Because of this bug, our benchmarks fail to run with curl backend, see https://github.com/vortex-data/vortex/pull/6685


0ax1:
@claude review

0ax1 (approved):

PR #6710, opened 2026-02-27
chore: fix naming and function style for vortex-cxx

SYaoJun:
I have followed the warnings from .clang-tidy and fixed the naming of functions and class members to conform with the project’s coding style.

```yaml
  - key:             readability-identifier-naming.FunctionCase
    value:           CamelCase
  - key:             readability-identifier-naming.MemberCase
    value:           lower_case
``` 

## Summary

Closes: #000

### DIFF
--- a/vortex-cxx/cpp/include/vortex/dtype.hpp
+++ b/vortex-cxx/cpp/include/vortex/dtype.hpp
@@ -27,7 +27,7 @@ namespace dtype {
     class DType {
     public:
         DType() = delete;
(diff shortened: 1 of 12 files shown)


myrrc:
We have existing users for our C++ api so I don't think we should break compatibility. I'll add this as a TODO for myself to fix formatting as I'd be soon changing C++ api.

myrrc (changes requested):
This is a major C++ api change which breaks compatibility.

SYaoJun:
> This is a major C++ api change which breaks compatibility.

Happy to receive your feedback.

PR #6714, opened 2026-02-27
gpu scan bench

onursatici:
## Summary

Add a gpu scan binary for benchmarks, this is not wired in to CI yet

### DIFF
--- a/vortex-cuda/gpu-scan-bench/src/main.rs
+++ b/vortex-cuda/gpu-scan-bench/src/main.rs
@@ -1,208 +1,206 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-#![allow(unused_imports)]
-
-use std::fs::File;
-use std::path::PathBuf;
-use std::sync::Arc;
-use std::time::Instant;
-
-use clap::Parser;
-use futures::StreamExt;
-use object_store::aws::AmazonS3Builder;
-use object_store::path::Path as ObjectPath;
-use tracing::Instrument;
-use tracing_perfetto::PerfettoLayer;
-use tracing_subscriber::EnvFilter;
-use tracing_subscriber::Layer;
-use tracing_subscriber::fmt::format::FmtSpan;
-use tracing_subscriber::layer::SubscriberExt;
-use tracing_subscriber::util::SubscriberInitExt;
-use url::Url;
-use vortex::VortexSessionDefault;
-use vortex::error::VortexResult;
-use vortex::error::vortex_bail;
-use vortex::file::OpenOptionsSessionExt;
-use vortex::io::session::RuntimeSessionExt;
-use vortex::session::VortexSession;
-use vortex_cuda::CudaSession;
-use vortex_cuda::PinnedByteBufferPool;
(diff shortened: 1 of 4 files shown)


vortex-cuda/gpu-scan-bench/src/main.rs
```diff
@@ -0,0 +1,208 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#![allow(unused_imports)]
```
0ax1: Can we scope this more narrow or use `cuda_available`/`cuda_not_available`? We could put all of the CUDA logic in a separate module/file and include that mod with `cuda_available`. wdyt?

vortex-cuda/gpu-scan-bench/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+        let perfetto_file = File::create(perfetto_path)?;
+        Some(PerfettoLayer::new(perfetto_file).with_debug_annotations(true))
+    } else {
+        None
+    };
+
+    if cli.json {
+        let log_layer = tracing_subscriber::fmt::layer()
+            .json()
+            .with_span_events(FmtSpan::NONE)
```
0ax1: Should we hoist out shared construction of the `log_layer`?

vortex-cuda/gpu-scan-bench/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+    register_cuda_layout(&session);
+
+    let mut cuda_ctx = CudaSession::create_execution_ctx(&session)?
+        .with_launch_strategy(Arc::new(TracingLaunchStrategy));
+
+    let pool = Arc::new(PinnedByteBufferPool::new(Arc::clone(
+        cuda_ctx.stream().context(),
+    )));
+    let cuda_stream =
+        VortexCudaStreamPool::new(Arc::clone(cuda_ctx.stream().context()), 1).get_stream()?;
```
0ax1: This needs to adjust to latest develop: `get_stream` => `stream`

vortex-cuda/gpu-scan-bench/src/main.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    for iteration in 0..cli.iterations {
+        let start = Instant::now();
+
+        let gpu_file = session.open_options().open(Arc::clone(&reader)).await?;
+
+        let mut batches = gpu_file.scan()?.into_array_stream()?;
+
+        let mut chunk = 0;
+        while let Some(next) = batches.next().await.transpose()? {
```
0ax1: What batch size is this here, 1MB? And we'll call multiple kernels per batch here. That'll be very expensive.

vortex-cuda/gpu-scan-bench/src/main.rs
0ax1: Side note: I'll change the CI tasks to at least build all CUDA code on each PR.

vortex-cuda/gpu-scan-bench/src/main.rs
robert3005: let it rip

PR #6716, opened 2026-02-27
Some adjustments to methods on `ExtDTypeRef`

connortsui20:
## Summary

Makes some adjustments to code that slipped into the a previous PR. Mostly related to display logic. Ideally we would make it so that all of them don't have this `impl Display` logic but we can do that another time.

## API Changes

Some of these methods shouldn't be public, and this reverts it.

## Testing

N/A

### DIFF
--- a/vortex-array/src/dtype/extension/erased.rs
+++ b/vortex-array/src/dtype/extension/erased.rs
@@ -67,6 +67,16 @@ impl ExtDTypeRef {
         }
     }
 
+    /// Compute equality ignoring nullability.
+    pub fn eq_ignore_nullability(&self, other: &Self) -> bool {
+        self.id() == other.id()
+            && self.0.metadata_eq(other.0.metadata_any())
+            && self
+                .storage_dtype()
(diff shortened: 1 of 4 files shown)


vortex-array/src/dtype/extension/erased.rs
```diff
@@ -193,28 +173,54 @@ impl Hash for ExtDTypeRef {
...
-impl fmt::Display for StorageValueDisplay<'_> {
+struct MetadataDisplay<'a>(&'a dyn DynExtDType);
+impl fmt::Display for MetadataDisplay<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.0.value_display(f, self.1)
+        self.0.metadata_display(f)
     }
 }
+
+// struct PythonDisplay<'a>(&'a dyn DynExtDType);
```
gatesn, line 211: Kill this?

gatesn (approved):

PR #6717, opened 2026-02-27
`ScalarValue::Array`

connortsui20:
## Summary

Adds an `Array` variant to `ScalarValue` that holds an `ArrayRef`. For now, this `Array` variant will only represent list scalars (for `List` and `FixedSizeList` arrays). 

To maintain backwards compatibility, we take old list scalars (which is just a `Vec<pb::ScalarValue>` and push them into an array builder after deserializing to create an `Array`.

To make things less confusing now, this PR also renames the old `ListValue` to `StructValue`. We will only write these for struct scalars from now on.

This change also requires passing a `VortexSession` in a lot more places (this is why we needed to do #6751, #6759, and #6761).

## API Changes

The main change is the addition of the `Array` variant to `ScalarValue`. Additionally, the names of several things has changed to reduce confusion (`ListValue` to `StructValue`, and some constructors, etc).

## Testing

The existing tests should probably be sufficient but I will add some more. TODO


vortex-array/src/scalar/scalar_value.rs
```diff
@@ -4,36 +4,63 @@
...
+    /// scalars (and any other list scalars written to files in the past for backcompat).
     List(Vec<Option<ScalarValue>>),
+
+    /// An Array.
+    ///
+    /// We serialize this by using [`Array::serialize()`] into protobuf bytes. Note that because we
+    /// require passing the length of the array to deserialize with [`ArrayParts::decode()`], we
+    /// store the array length in the first 8 bytes.
+    ///
+    /// [`Array::serialize()`]: crate::Array::serialize
```
gatesn: Don't do this!

vortex-array/src/serde.rs
```diff
@@ -639,6 +639,30 @@ impl ArrayChildren for ArrayPartsChildren<'_> {
     }
 }
 
+// TODO(connor): There is probably performance that is left on the table here.
+
+impl TryFrom<&[u8]> for ArrayParts {
```
gatesn: Delete this

vortex-array/src/scalar/proto.rs
```diff
@@ -365,64 +409,139 @@ fn string_from_proto(s: &str, dtype: &DType) -> VortexResult<ScalarValue> {
...
+            .ok()
+            .vortex_expect("we just checked that there are at least 8 bytes"),
+    );
+    let len = usize::try_from(len)
+        .map_err(|_| vortex_err!(Serde: "array length {len} exceeds usize::MAX"))?;
+
+    let parts = ArrayParts::try_from(&bytes[8..])?;
+
+    // Deserialize the entire array.
+    let ctx = ArrayContext::empty();
```
gatesn: This doesn't work. The ArrayContext needs to be the same one that was used at write-time since it contains a mapping of u16 -> EncodingId.

I think we're better off using the IPC crate which already adds a header for framing that includes the context + the length

gatesn (changes requested):

vortex-array/src/scalar/proto.rs
connortsui20: We can't depend on `vortex-ipc` since it already has a dependency on `vortex-array`. I can mimic it in a similar way though.

vortex-array/src/arrays/fixed_size_list/vtable/operations.rs
```diff
@@ -6,22 +6,14 @@ use vortex_error::VortexResult;
...
-
-        debug_assert_eq!(children_elements.len(), array.list_size() as usize);
-
-        Ok(Scalar::fixed_size_list(
-            list.dtype().clone(),
-            children_elements,
-            array.dtype().nullability(),
-        ))
+        let scalar_value = ScalarValue::Array(list);
+        Scalar::try_new(array.dtype().clone(), Some(scalar_value))
```
connortsui20, line 23: Is it worth making a constructor for this?

vortex-array/src/scalar/proto.rs
```diff
@@ -377,51 +417,110 @@ fn string_from_proto(s: &str, dtype: &DType) -> VortexResult<ScalarValue> {
...
+    let parts = ArrayParts::try_from(array_value.data.as_slice())?;
+
+    let encoding_ids = array_value
+        .encoding_ids
+        .iter()
+        .map(|id| ArrayId::new_arc(Arc::from(id.clone())))
+        .collect::<Vec<_>>();
+    let ctx = ArrayContext::new(encoding_ids);
+
+    parts.decode(elem_dtype, array_len, &ctx, session)
```
connortsui20, line 568: I have no idea if this is correct

vortex-array/src/scalar/scalar_impl.rs
```diff
@@ -186,9 +186,11 @@ impl Scalar {
             DType::Decimal(..) => value.as_decimal().is_zero(),
             DType::Utf8(_) => value.as_utf8().is_empty(),
             DType::Binary(_) => value.as_binary().is_empty(),
-            DType::List(..) => value.as_list().is_empty(),
-            DType::FixedSizeList(_, list_size, _) => value.as_list().len() == *list_size as usize,
-            DType::Struct(struct_fields, _) => value.as_list().len() == struct_fields.nfields(),
+            DType::List(..) => self.as_list().is_empty(),
+            DType::FixedSizeList(_, list_size, _) => self.as_list().len() == *list_size as usize,
```
connortsui20, line 190: Is this strange? (value vs self)

vortex-array/src/serde.rs
```diff
@@ -639,6 +639,32 @@ impl ArrayChildren for ArrayPartsChildren<'_> {
     }
 }
 
+// TODO(connor): Do we need this?
+
+impl TryFrom<&[u8]> for ArrayParts {
+    type Error = VortexError;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
```
connortsui20, line 644: Should we just reimplement this for slices instead of byte buffer? It is not clear to me why alignment is important.

vortex-array/src/scalar/scalar_value.rs
```diff
@@ -163,6 +191,91 @@ impl Display for ScalarValue {
...
+            (ScalarValue::Array(_), ScalarValue::List(_))
+            | (ScalarValue::List(_), ScalarValue::Array(_)) => {
+                let lhs = self.as_list_expanded().vortex_expect(
+                    "something went wrong when expanding lhs scalar value for comparison",
+                );
+                let rhs = other.as_list_expanded().vortex_expect(
+                    "something went wrong when expanding rhs scalar value for comparison",
+                );
+                lhs == rhs
+            }
```
connortsui20: This whole thing feels ugly...

gatesn (commented):
I think we can convert to ScalarValue::Array during deserialization, then we don't need all the cross-list x Array checks?

vortex-proto/proto/scalar.proto
```diff
@@ -28,9 +28,16 @@ message ScalarValue {
...
 }
 
 message ListValue {
   repeated ScalarValue values = 1;
 }
+
+message ArrayValue {
+  uint64 row_count = 1;
+  repeated string encoding_ids = 2;
+  bytes data = 3;
```
joseph-isaacs, line 84: can we add a doc comment to this?

joseph-isaacs: or something on the file explaining where to look

vortex-array/src/arrays/list/vtable/operations.rs
```diff
@@ -1,27 +1,19 @@
...
 
 use crate::arrays::ListArray;
 use crate::arrays::ListVTable;
 use crate::scalar::Scalar;
+use crate::scalar::ScalarValue;
 use crate::vtable::OperationsVTable;
 
 impl OperationsVTable<ListVTable> for ListVTable {
     fn scalar_at(array: &ListArray, index: usize) -> VortexResult<Scalar> {
         // By the preconditions we know that the list scalar is not null.
```
joseph-isaacs, line 20: I think these should be debug asserts?

vortex-array/src/builders/fixed_size_list.rs
```diff
@@ -123,9 +122,16 @@ impl FixedSizeListBuilder {
             );
         }
 
+        // Fast-path: if the list scalar is backed by an array, append the entire thing directly.
+        if let Some(array) = value.as_array() {
+            return self.append_array_as_list(array);
+        }
```
joseph-isaacs: how would this not happen now?

joseph-isaacs:
I think I have a few questions about this: 
 1. Do we have anything written about how these Array scalar compose with others arrays, I guess they may contain a ConstantArray to get back into scalar land?
 2. Since we have array backed scalar what is the value of the non-Array backed ones?

vortex-proto/proto/scalar.proto
connortsui20: yep

vortex-array/src/builders/fixed_size_list.rs
connortsui20: dont understand this question?

connortsui20:
> Do we have anything written about how these Array scalar compose with others arrays, I guess they may contain a ConstantArray to get back into scalar land?

It just represents a list scalar for now, so it can have any kind of array it wants.

> Since we have array backed scalar what is the value of the non-Array backed ones?

Convenience. I don't really want to have to _always_ work with length-1 arrays for things like integers. And note that this variant (for now) only works with list scalars.

vortex-array/src/arrays/list/vtable/operations.rs
connortsui20: I mean... should we just add those everywhere then?

connortsui20: Or maybe the `list_elements_at` function should be unsafe

vortex-array/src/builders/fixed_size_list.rs
connortsui20: funnily enough in the most recent changes we now dont need to do this, am fixing now

joseph-isaacs:
> It just represents a list scalar for now, so it can have any kind of array it wants.

Of course you can store this here, but what is intended. Do we validate that these arrays only have read allowed encodings.

https://github.com/vortex-data/vortex/blob/8ebd2b88b98dd26f94b0383da791314620e272f0/vortex-layout/src/layouts/flat/writer.rs#L153

joseph-isaacs (changes requested):
I just want to validate that we actually validate that only allow-listed array can be written in this scalar variant (https://github.com/vortex-data/vortex/blob/8ebd2b88b98dd26f94b0383da791314620e272f0/vortex-layout/src/layouts/flat/writer.rs#L153)

connortsui20:
@joseph-isaacs Are you saying that we should not allow certain arrays to be serialized, even if it's just a scalar?

In that case we might have to execute the array before serialization...

joseph-isaacs:
I don't want to have a filter in that be serialised to disk.

connortsui20:
Since we want to add basic vector similarity search (test a query vector against a bunch of db vectors) we are going it want this soon, amin some form or another)

connortsui20:
im going to close this but keep the branch since we will inevitably come back to this in the future

PR #6720, opened 2026-02-28
rwlock instead of mutex around encoding ids

danking:
Is this reasonable? I noticed it while staring at Spiral profiles.


robert3005:
I think we should really have mutable session ie a builder and immutable session that doesn’t have any locks

gatesn:
@robert3005 - I'm not sure I agree

Also, this has nothing to do with the session. Session holds Registries, not Contexts.

Context represents the interned/dictionary-encoded set of encodings present in a file. It's dual-use for reading/writing which is maybe the weird thing here.

But RwLock should get us all the lift we need

gatesn (approved):

robert3005:
Right so if you avoid dual use you end up with immutable context for read and mutable one for writing

PR #6726, opened 2026-03-01
test: Enhance tests for RLE and PCO casting: add assertions for dtype…

SYaoJun:
… preservation and validity after nullability changes

## Summary
Two tests were only asserting on the output dtype after a cast operation,
without verifying that the actual values were correctly preserved. This
means the tests would pass even if the cast silently corrupted data.

Closes: #000

## Testing

tests passed


connortsui20 (approved):

connortsui20:
Thanks for the PR!

connortsui20:
@SYaoJun Could you fix the lint? Then we should be good to go. Don't worry about the perf regression, it's just wrong :/

SYaoJun:
> @SYaoJun Could you fix the lint? Then we should be good to go. Don't worry about the perf regression, it's just wrong :/

I have fixed the lint.

encodings/fastlanes/src/rle/compute/cast.rs
```diff
@@ -64,14 +65,15 @@ mod tests {
...
             .to_array()
-            .cast(DType::Primitive(PType::U16, Nullability::NonNullable));
-        assert!(res.is_ok());
+            .cast(DType::Primitive(PType::U16, Nullability::NonNullable))
+            .unwrap();
         assert_eq!(
-            res.unwrap().dtype(),
+            casted.dtype(),
             &DType::Primitive(PType::U16, Nullability::NonNullable)
         );
```
joseph-isaacs: This is checked below, and so can be removed

encodings/pco/src/compute/cast.rs
```diff
@@ -103,6 +103,14 @@ mod tests {
             casted.dtype(),
             &DType::Primitive(PType::U32, Nullability::Nullable)
         );
```
joseph-isaacs: please remove this

encodings/pco/src/compute/cast.rs
```diff
@@ -103,6 +103,14 @@ mod tests {
...
+
+        assert_arrays_eq!(
+            casted,
+            PrimitiveArray::new(
+                Buffer::copy_from(vec![10u32, 20, 30, 40]),
```
joseph-isaacs: this is not the pattern use buffer!

SYaoJun: > this is not the pattern use buffer!

Hi @joseph-isaacs, do you mean using Buffer::from_iter()?

encodings/pco/src/compute/cast.rs
SYaoJun: done

encodings/fastlanes/src/rle/compute/cast.rs
SYaoJun: done

connortsui20: @joseph-isaacs personally I'm fine (would honestly lean towards prefer) having these extra checks since it makes it obvious what this test is testing (rather than assume the reader understands nullability semantics of `from_iter` and `Validity::AllValid`, which I would argue is not obvious).

@SYaoJun I'll leave it to you to decide what makes more sense.

connortsui20:
@SYaoJun Just a heads up that we run `cargo +nightly fmt --all` to format our code, and the CI will check that it is correctly formatted this way!

SYaoJun:
> @SYaoJun Just a heads up that we run `cargo +nightly fmt --all` to format our code, and the CI will check that it is correctly formatted this way!

Ok, my mistake—I forgot to check this. I have now updated the PR.

connortsui20 (approved):
LGTM! Thanks again for the PR.

PR #6727, opened 2026-03-01
chore: Convert `SliceKernel` impls to `SliceReduce`

0ax1:
Resolving `SliceArray` nodes at optimization time rather than execution time.

Changed for: `ALPVTable`, `ALPRDVTable`, `BitPackedVTable`, `RLEVTable`, `SparseVTable`, `ChunkedVTable`.

### DIFF
--- a/encodings/alp/src/alp/compute/slice.rs
+++ b/encodings/alp/src/alp/compute/slice.rs
@@ -4,20 +4,15 @@
 use std::ops::Range;
 
 use vortex_array::ArrayRef;
-use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
-use vortex_array::arrays::SliceKernel;
+use vortex_array::arrays::SliceReduce;
 use vortex_error::VortexResult;
 
 use crate::ALPArray;
 use crate::ALPVTable;
 
-impl SliceKernel for ALPVTable {
-    fn slice(
-        array: &Self::Array,
-        range: Range<usize>,
-        _ctx: &mut ExecutionCtx,
-    ) -> VortexResult<Option<ArrayRef>> {
+impl SliceReduce for ALPVTable {
+    fn slice(array: &Self::Array, range: Range<usize>) -> VortexResult<Option<ArrayRef>> {
         let sliced_alp = ALPArray::new(
             array.encoded().slice(range.clone())?,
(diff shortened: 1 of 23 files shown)


gatesn:
Where they are constant time we should indeed be reduce rules.

RunEnd cannot be a reduce rule since it needs access to the data.

In either case, it should have no impact on execution time or available optimizations _if_ the other compute kernels were written to not recurse incorrectly.

@joseph-isaacs and I are planning an API change to make it harder to do the wrong thing here

0ax1:
> In either case, it should have no impact on execution time or available optimizations _if_ the other compute kernels were written to not recurse incorrectly.

Thanks for clarifying!

0ax1:
> RunEnd cannot be a reduce rule since it needs access to the data.

Dropped this change for runend.

encodings/alp/src/alp/compute/slice.rs
joseph-isaacs, line 23: This is a uses buffers

joseph-isaacs (changes requested):
This was intentionally marked as a kernel, see need to change the way we handle patches to move over

joseph-isaacs:
To do this we have to handle patches correctly, maybe using a custom array.

encodings/alp/src/alp/compute/slice.rs
a10y: Yea, if we added a PatchesArray I think this could work by wrapping with SliceArray

robert3005: with galp style patches all patches slicing is metadata only

robert3005:
Do we still want this pr?

a10y:
i have a separate pr #6815 in progress which adds the new array for patches and implements the slicerule

gatesn:
I'm still not sure we want a patches array. At least, not yet.

Unless we have some way of passing down an "is_defined" mask into execution, there's no way for the child array to know whether some checked operation should fail or not.

e.g. the child of a Patches array may overflow on some arithmetic, but the patch would have succeeded. We need a way to tell the child to ignore the overflow for the invalid positions.

0ax1:
Closing this but if anyone wants to take over in case we're sure we want this feel free to go ahead.

PR #6735, opened 2026-03-02
VTable Unification Docs

gatesn:
Updates the VTable plan to remove the `FooInner<V>` struct. So now `Foo<V>` is not internally Arc'd. 

Downsides:
* We lose the ability to go from `&Foo<V> -> FooRef` with only an Arc::clone. Turns out we were never doing this anyway, and where we _do_ need return self.clone() it happens to be on the type-erased API which remains internally arc'd.

Upsides:
* Borrowed downcasts `as_::<V>() -> &Foo<V>` can now return the full `Foo` wrapper struct, instead of just the associated type (`-> &V::Metadata`). This is a nicer API for users wanting to access typed fields instead of having to separately access shared properties from the erased object.


connortsui20 (approved):
LGTM

PR #6736, opened 2026-03-02
feat: add clickhouse-bench with auto-downloaded ClickHouse binary

fastio:
Introduce a new clickhouse-bench benchmark crate that runs ClickBench queries against Parquet data via clickhouse-local, providing a baseline for comparing Vortex performance against ClickHouse.

Key design decisions:
- build.rs auto-downloads the full ClickHouse binary (with Parquet support) into target/clickhouse-local/, similar to how vortex-duckdb downloads the DuckDB library. This eliminates manual install steps and avoids issues with slim/homebrew builds lacking Parquet support.
- The binary path is baked in via CLICKHOUSE_BINARY env at compile time; CLICKHOUSE_LOCAL env var allows runtime override.
- ClickHouse-dialect SQL queries are maintained in a separate clickbench_clickhouse_queries.sql file (43 queries).
- CI workflows updated to include clickhouse:parquet target in ClickBench benchmarks and conditionally build clickhouse-bench.

#6425

### DIFF
--- a/.github/scripts/run-sql-bench.sh
+++ b/.github/scripts/run-sql-bench.sh
@@ -2,8 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright the Vortex contributors
 #
-# Runs SQL benchmarks (datafusion-bench, duckdb-bench, lance-bench) for the given targets.
-# This script is used by the sql-benchmarks.yml workflow.
+# Runs SQL benchmarks (datafusion-bench, duckdb-bench, lance-bench, clickhouse-bench)
+# for the given targets. This script is used by the sql-benchmarks.yml workflow.
 #
 # Usage:
 #   run-sql-bench.sh <subcommand> <targets> [options]
@@ -11,13 +11,13 @@
 # Arguments:
 #   subcommand   The benchmark subcommand (e.g., tpch, clickbench, tpcds)
 #   targets      Comma-separated list of engine:format pairs
-#                (e.g., "datafusion:parquet,datafusion:vortex,duckdb:parquet")
+#                (e.g., "datafusion:parquet,datafusion:vortex,duckdb:parquet,clickhouse:parquet")
 #
 # Options:
 #   --scale-factor <sf>       Scale factor for the benchmark (e.g., 1.0, 10.0)
 #   --iterations <n>          Number of iterations to pass to each benchmark binary
 #   --remote-storage <url>    Remote storage URL (e.g., s3://bucket/path/)
-#                             If provided, runs in remote mode (no lance support).
+#                             If provided, runs in remote mode (no lance/clickhouse support).
 #   --benchmark-id <id>       Benchmark ID for error messages (e.g., tpch-s3)
 
 set -Eeu -o pipefail
@@ -71,6 +71,13 @@ if $is_remote && echo "$targets" | grep -q 'lance'; then
     exit 1
 fi
 
+# ClickHouse on remote storage is not supported. clickhouse-local reads local files only.
+if $is_remote && echo "$targets" | grep -q 'clickhouse:'; then
+    echo "ERROR: ClickHouse benchmarks are not supported for remote storage."
+    echo "Remove 'clickhouse:' targets for benchmark '${benchmark_id:-unknown}'."
+    exit 1
+fi
+
 # Extract formats for each engine from the targets string.
 # Example input: "datafusion:parquet,datafusion:vortex,datafusion:lance,duckdb:parquet"
 #
@@ -84,6 +91,7 @@ fi
 df_formats=$(echo "$targets" | tr ',' '\n' | (grep '^datafusion:' | grep -v ':lance$' || true) | sed 's/datafusion://' | tr '\n' ',' | sed 's/,$//')
 ddb_formats=$(echo "$targets" | tr ',' '\n' | (grep '^duckdb:' || true) | sed 's/duckdb://' | tr '\n' ',' | sed 's/,$//')
 has_lance=$(echo "$targets" | grep -q 'datafusion:lance' && echo "true" || echo "false")
+has_clickhouse=$(echo "$targets" | grep -q 'clickhouse:' && echo "true" || echo "false")
 
 # Build options string.
 opts=""
@@ -135,4 +143,15 @@ if ! $is_remote && [[ "$has_lance" == "true" ]] && [[ -f "target/release_debug/l
         -o lance-results.json
 
     cat lance-results.json >> results.json
+fi
+
+# ClickHouse-bench only runs for local benchmarks (clickhouse-local reads local files).
+if ! $is_remote && [[ "$has_clickhouse" == "true" ]] && [[ -f "target/release_debug/clickhouse-bench" ]]; then
+    # shellcheck disable=SC2086
+    target/release_debug/clickhouse-bench "$subcommand" \
+        -d gh-json \
+        $opts \
+        -o ch-results.json
+
+    cat ch-results.json >> results.json
 fi
--- a/.github/workflows/sql-benchmarks.yml
+++ b/.github/workflows/sql-benchmarks.yml
@@ -21,7 +21,7 @@ on:
               "id": "clickbench-nvme",
               "subcommand": "clickbench",
               "name": "Clickbench on NVME",
-              "targets": "datafusion:parquet,datafusion:vortex,duckdb:parquet,duckdb:vortex,duckdb:duckdb"
+              "targets": "datafusion:parquet,datafusion:vortex,duckdb:parquet,duckdb:vortex,duckdb:duckdb,clickhouse:parquet"
             },
             {
               "id": "tpch-nvme",
@@ -135,6 +135,16 @@ jobs:
 
       - uses: ./.github/actions/system-info
 
+      - name: Install ClickHouse
+        if: contains(matrix.targets, 'clickhouse:')
+        env:
+          CLICKHOUSE_VERSION: "25.8.18.1"
+        run: |
+          wget -qO- "https://github.com/ClickHouse/ClickHouse/releases/download/v${CLICKHOUSE_VERSION}-lts/clickhouse-common-static-${CLICKHOUSE_VERSION}-amd64.tgz" | tar xz
+          cp clickhouse-common-static-${CLICKHOUSE_VERSION}/usr/bin/clickhouse .
+          chmod +x clickhouse
+          echo "CLICKHOUSE_BINARY=$PWD/clickhouse" >> $GITHUB_ENV
+
       - name: Build binaries
         shell: bash
         env:
@@ -143,6 +153,9 @@ jobs:
           packages="--bin data-gen --bin datafusion-bench --bin duckdb-bench"
           if [ "${{ matrix.build_lance }}" = "true" ]; then
             packages="$packages --bin lance-bench"
+          fi
+          if echo "${{ matrix.targets }}" | grep -q 'clickhouse:'; then
+            packages="$packages --bin clickhouse-bench"
           fi
           cargo build $packages --profile release_debug
 
--- a/benchmarks/clickhouse-bench/Cargo.toml
+++ b/benchmarks/clickhouse-bench/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "clickhouse-bench"
+description = "ClickHouse (clickhouse-local) benchmark runner for Vortex"
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+readme.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+publish = false
+
+[dependencies]
+anyhow = { workspace = true }
+clap = { workspace = true, features = ["derive"] }
+parking_lot = { workspace = true }
+tokio = { workspace = true }
+tracing = { workspace = true }
+vortex-bench = { workspace = true }
+
+[lints]
+workspace = true
--- a/benchmarks/clickhouse-bench/src/lib.rs
+++ b/benchmarks/clickhouse-bench/src/lib.rs
@@ -1,0 +1,240 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! ClickHouse Local context for benchmarks.
+//!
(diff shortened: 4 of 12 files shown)


vortex-bench/clickbench_clickhouse_queries.sql
joseph-isaacs, line 1: why do we need this file is it difference to the already included one?

fastio: Good catch! I have removed the duplicate clickbench_clickhouse_queries.sql and validated with cargo check -p vortex-bench.

myrrc (commented):
I don't think downloading untrusted binaries from internet via a build script is a good idea. We want first-class integration with duckdb thus we need to download its sources (although I'd not do it in build script as well), but we don't need such integration with Clickhouse yet.

My idea is to use clickhouse binary in CI (as it runs on Linux only) and require users to download it by hand if they want a local run. Benchmarking on MacOS doesn't make much sense anyway as vectorized instrustion set is different.

fastio:
> I don't think downloading untrusted binaries from internet via a build script is a good idea. We want first-class integration with duckdb thus we need to download its sources (although I'd not do it in build script as well), but we don't need such integration with Clickhouse yet.
> 
> My idea is to use clickhouse binary in CI (as it runs on Linux only) and require users to download it by hand if they want a local run. Benchmarking on MacOS doesn't make much sense anyway as vectorized instrustion set is different.

Agreed — removed the binary download from build.rs entirely. The clickhouse binary is now resolved at runtime: via CLICKHOUSE_BINARY env var or from $PATH. CI installs it via the official installer before building. Local users need to install it manually. No more untrusted binary downloads in the build script.

.github/workflows/sql-benchmarks.yml
```diff
@@ -127,6 +127,13 @@ jobs:
 
       - uses: ./.github/actions/system-info
 
+      - name: Install ClickHouse
+        if: contains(matrix.targets, 'clickhouse:')
+        run: |
+          curl https://clickhouse.com/ | sh
```
myrrc: Why not download the latest release file for our architecture from Github releases? We then don't need any installation and curl in general.

fastio: Good call — updated CI to download the static binary directly from GitHub Releases (pined ClickHouse to LTS release v25.8.18.1 from GitHub Releases), no curl | sh or installation needed.

vortex-bench/src/clickbench/benchmark.rs
```diff
@@ -34,17 +35,46 @@ impl ClickBenchBenchmark {
...
+    }
+
+    /// Strips double quotes only from simple SQL identifiers for engines like
+    /// ClickHouse that don't require quoted column names.
+    fn normalize_query(&self, query: &str) -> String {
+        if !self.uses_unquoted_identifiers() {
+            return query.to_string();
+        }
+
+        strip_simple_identifier_quotes(query)
```
myrrc: Clickhouse does handle quoted identifiers correctly so I think we can pass them through to reduce this PR's diff.

myrrc:
The changes look good to me conceptually, let's see what the CI run says.

connortsui20:
@fastio Feel free to ping us in the public [slack channel](https://vortex.dev/slack) if you want us to run CI for you! (Feel free to ping me here as well)

Edit: To fix the CI issues right now, could you update the lockfile with `cargo check`?

fastio:
> @fastio Feel free to ping us in the public [slack channel](https://vortex.dev/slack) if you want us to run CI for you! (Feel free to ping me here as well)
> 
> Edit: To fix the CI issues right now, could you update the lockfile with `cargo check`?

@connortsui20 Thanks! I've updated the lockfile by running cargo check. The Cargo.lock should now be in sync. Let me know if CI still has issues after this.

myrrc (approved):

connortsui20:
@fastio you can click on the github action that is failing (in this case it is the "CI / Rust (list) (pull_request)" action) and see the problem. Let me know if you need help with this!

0ax1 (dismissed):
It's unclear yet, whether we actually want to maintain this integration. Will loop back with the team.

benchmarks/clickhouse-bench/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+        // Write SQL to stdin
+        {
+            let stdin = child
+                .stdin
+                .as_mut()
+                .context("Failed to open clickhouse-local stdin")?;
+            stdin
+                .write_all(full_sql.as_bytes())
+                .context("Failed to write SQL to clickhouse-local stdin")?;
+        }
```
connortsui20, line 207: I don't think this is right. If you look at the other benchmark engine setup, we do not open an entire new process and write to stdin, which means this engine is going to have wildly different characteristics to the rest of the engines.

I also don't know of a good way to solve this since I have not worked much with Clickhouse before. Do you have any ideas how to make this more consistent with the other engines?

benchmarks/clickhouse-bench/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// 4. Parse stdout to count result rows
+    pub fn execute_query(&self, query: &str) -> Result<(usize, Option<Duration>)> {
+        trace!("execute clickhouse query: {query}");
+
+        // Build the full SQL: setup views + the actual query
+        let mut full_sql = String::new();
+        for stmt in &self.setup_sql {
+            full_sql.push_str(stmt);
+            full_sql.push('\n');
+        }
```
connortsui20: This also seems strange to me, if this is setup work then why are we timing it in our measurement?

.github/scripts/run-sql-bench.sh
```diff
@@ -127,3 +128,14 @@ if ! $is_remote && [[ "$has_lance" == "true" ]] && [[ -f "target/release_debug/l
 
     cat lance-results.json >> results.json
 fi
+
+# ClickHouse-bench only runs for local benchmarks (clickhouse-local reads local files).
+if ! $is_remote && [[ "$has_clickhouse" == "true" ]] && [[ -f "target/release_debug/clickhouse-bench" ]]; then
```
connortsui20, line 149: If you look above at the lance setup code, we have a better guard against benching something on remote that is not supposed to be, can you mimic that?

benchmarks/clickhouse-bench/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+readme.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+publish = false
+
+[dependencies]
+anyhow = { workspace = true }
+clap = { workspace = true, features = ["derive"] }
+tokio = { workspace = true, features = ["full"] }
```
connortsui20: I doubt you need "full" features here

vortex-bench/src/clickbench/benchmark.rs
```diff
@@ -37,14 +35,21 @@ impl ClickBenchBenchmark {
...
     }
 
+    /// Returns the path to the queries file.
+    fn queries_file_path(&self) -> PathBuf {
+        if let Some(file) = &self.queries_file {
+            return file.into();
+        }
+        let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        manifest_dir.join("clickbench_queries.sql")
+    }
```
connortsui20, line 45: This refactoring seems somewhat unnecessary

connortsui20 (changes requested):
Before addressing some of these comments, can you fix the clippy lint? We should actually run these benchmarks to see if it even works.

benchmarks/clickhouse-bench/src/lib.rs
fastio: The setup SQL (CREATE VIEW statements) is indeed being timed along with the actual query. I'll separate the setup phase from the measurement phase to be consistent with how DuckDB and DataFusion
  handle this.

fastio:
> It's unclear yet, whether we actually want to maintain this integration. Will loop back with the team.


  Hi @0ax1, this PR is Phase 1 of the plan discussed in https://github.com/vortex-data/vortex/discussions/6425 with @gatesn. The idea was to first integrate ClickHouse into the ClickBench benchmarking framework
  (Parquet baseline only) before submitting the full vortex-clickhouse crate.

  As @gatesn put it: "I think we should also look to at Clickbench to the benchmarking framework first, even without support for Vortex. That will give us some sanity checks on the PR that we are doing things
  sensibly regarding both performance and correctness."

  This PR intentionally has a narrow scope — it only adds ClickHouse as a benchmark engine running against Parquet via clickhouse-local. The maintenance burden is minimal: one benchmark crate with no custom
  library bindings. Happy to discuss further if the team has concerns!

benchmarks/clickhouse-bench/src/lib.rs
gatesn: No but we essentially do this. DuckDB we close and re-open the database. DataFusion we start a new session. etc.

gatesn:
Just to confirm, I'm happy to go ahead and merge this. Sounds like there's some open questions about where to put the benchmark timings.

Also it's worth making sure the caching behavior is consistent with our other benchmarks. We haven't actually documented precisely what this is yet, but typically means no caching at all, even things like footers.

benchmarks/clickhouse-bench/src/lib.rs
connortsui20: I don't believe this is true (or not true anymore?). In our `benchmarks/datafusion-bench/src/main.rs` and `benchmarks/duckdb-bench/src/lib.rs` we create the connections and only time the query

fastio:
Thanks for confirming the merge! On the two points:

**Benchmark timings**: Could you clarify what you have in mind? Are you referring to where the timing results should be reported (e.g., stdout, a results file, CI artifacts), or how the timing logic is organized in code? Happy to follow whatever convention you prefer.

**Caching behavior**: Since clickhouse-local spawns a fresh process per query, there's no cross-query caching at all — no metadata cache, no footer cache, no warm-up effects. This is actually stricter than DuckDB's current setup (which has `parquet_metadata_cache = true`). If we want to formalize the "no caching" policy across all engines, I can open a follow-up issue to track that.

benchmarks/clickhouse-bench/src/lib.rs
fastio: That's a fair point — the current timing does include process startup overhead. This is a constraint of clickhouse-local's non-interactive mode, which reads all of stdin before executing any queries, so I can't keep a persistent connection the way DuckDB and DataFusion do.

That said, the process startup cost is relatively small compared to query execution time (especially for ClickBench queries). If we want to isolate it, I could run a no-op query first and subtract the baseline, or we could accept the small overhead as part of the "no caching" trade-off. Let me know what you'd prefer.

fastio: You're right that DuckDB and DataFusion create connections once and time only the query execution. The per-query process spawning is an architectural constraint of `clickhouse-local` — in piped stdin mode, it reads **all** of stdin before executing anything, which makes a persistent-process model impossible (I've documented this in the module-level docs).

That said, the overhead is minimal: process spawn is ~1ms, and the CREATE VIEW statements are lightweight references to Parquet files via `file()`. For ClickBench queries running 100ms+, this adds <1% overhead. Happy to explore alternatives like `--queries-file` if you think it's worth it, but I believe the current approach is pragmatically fine.

fastio:
@gatesn Thanks for confirming! On caching: the per-process model is actually the most cache-free approach of all our engines. Each query spawns a fresh `clickhouse-local` process with zero state — no metadata cache, no footer cache, no in-memory buffers. The `CREATE VIEW` statements just create references via `file()` without pre-reading.

The only shared layer is the OS page cache, which affects all engines equally. So the caching behavior is at least as strict as our other benchmarks (arguably stricter than DuckDB, which enables `parquet_metadata_cache`).

On timing placement — currently results go to `ch-results.json` and get appended to `results.json` alongside other engines. Let me know if you'd prefer a different approach.

vortex-bench/src/clickbench/benchmark.rs
fastio: Could you clarify what you're referring to here? This file (`vortex-bench/src/clickbench/benchmark.rs`) has no changes in the current PR — the `queries_file_path()` method and `queries_file` field already exist on `develop`. This comment may have been targeting an earlier revision that was rebased out. Let me know if there's something specific you'd like addressed!

benchmarks/clickhouse-bench/Cargo.toml
fastio: Good catch — removed `features = ["full"]` from tokio. The benchmark only uses `tokio::runtime::Runtime` for one async call (`generate_base_data`), so workspace-level features are sufficient.

.github/scripts/run-sql-bench.sh
fastio: Nice catch — the `^clickhouse:` anchor only matched when clickhouse was the first target in the comma-separated string. Dropped the `^` from both the remote guard and the `has_clickhouse` detection to match the lance pattern.

connortsui20:
Could you rebase this onto develop? There seems to be a lot of weird git artifacts here, and also the Cargo lockfile has been messed up again.

Also, with your permission, I can make some of these fixups for you, and that should improve the iteration cycles.

fastio:
> Could you rebase this onto develop? There seems to be a lot of weird git artifacts here, and also the Cargo lockfile has been messed up again.
> 
> Also, with your permission, I can make some of these fixups for you, and that should improve the iteration cycles.

Hi @connortsui20, Sure, go ahead! You're welcome to work on my branch directly. Looking forward to faster iterations — thanks for the help!

connortsui20:
Ok I will clean this up and we can review it faster

Actually @fastio I think you need to give me permissions to make changes to your branch. You can do that, or you can squash and rebase on top of `develop` (squash because otherwise you will run into merge conflicts)

fastio:
> Ok I will clean this up and we can review it faster
> 
> Actually @fastio I think you need to give me permissions to make changes to your branch. You can do that, or you can squash and rebase on top of `develop` (squash because otherwise you will run into merge conflicts)

Hi @connortsui20,  I've enabled "Allow edits from maintainers" on the PR. I'll also do a squash rebase onto the latest develop to clean up the git history and fix the lockfile. Will push shortly.

fastio:
It looks like the approach I chose was wrong — the C FFI should be created on the ClickHouse side rather than adding a new C FFI on the Vortex side.

PR #6737, opened 2026-03-02
fix[util]: add balanced reduce iterator

joseph-isaacs:
We want expr/array tree to be wide not deep.

We use a balanced reduce added in this PR and use in list_contains and or/and collect from exprs

### DIFF
--- a/vortex-utils/src/iter.rs
+++ b/vortex-utils/src/iter.rs
@@ -19,7 +19,7 @@
 /// |\
 /// a b
 /// ```
-pub trait ReduceBalanced: Iterator + Sized {
+pub trait ReduceBalancedIterExt: Iterator + Sized {
     /// Like [`Iterator::reduce`], but builds a balanced binary tree instead of a linear chain.
     ///
     /// `[a, b, c, d]` becomes `combine(combine(a, b), combine(c, d))`.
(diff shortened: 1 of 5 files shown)


vortex-utils/src/iter.rs
```diff
@@ -0,0 +1,206 @@
...
+/// reduce:          reduce_balanced:
+///     f                  f
+///    / \                / \
+///   f   d              f   f
+///  / \                / \ / \
+/// f   c              a  b c  d
+/// |\
+/// a b
+/// ```
+pub trait ReduceBalanced: Iterator + Sized {
```
AdamGS: ReduceBalancedIterExt?

vortex-utils/src/iter.rs
```diff
@@ -0,0 +1,206 @@
...
+                // Merge the odd element into the last paired element so it stays inside the tree.
+                let lhs = items[(len / 2) - 1].clone();
+                let rhs = items[len - 1].clone();
+                items[len / 2 - 1] = combine(lhs, rhs);
+            }
+
+            items.truncate(len / 2);
+        }
+
+        items.pop()
```
AdamGS, line 76: `debug_assert` length here? and at the end of the other function too

AdamGS (approved):
This is cool, just some minor comments

vortex-utils/src/iter.rs
```diff
@@ -0,0 +1,208 @@
...
+/// reduce:          reduce_balanced:
+///     f                  f
+///    / \                / \
+///   f   d              f   f
+///  / \                / \ / \
+/// f   c              a  b c  d
+/// |\
+/// a b
+/// ```
+pub trait ReduceBalancedIterExt: Iterator + Sized {
```
AdamGS: trait bounds for these are always weird IMO, I think what `Itertools` does is that the trait only requires `Iterator` and each function has specific additional bounds (Like `Self: Sized`)

PR #6741, opened 2026-03-02
unsafe Scalar::struct_unchecked in scalar_at vtable

palaska:
`Scalar::struct_` collects field scalars into a `Vec<Scalar>` before dtype-checking them, then collects again into the final value list — two allocations per row.

Add `Scalar::struct_unchecked` (accepts `impl IntoIterator`) that skips the checks and does a single allocation. The `StructVTable::scalar_at` vtable already guarantees index is in-bounds and non-null, and each field's `scalar_at` returns the correct dtype, so the unchecked path is safe there.

### DIFF
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
+use vortex_error::vortex_ensure;
 use vortex_error::vortex_err;
 
 use crate::Array;
@@ -186,6 +187,7 @@ impl StructArray {
         &self,
         index: usize,
     ) -> VortexResult<Option<impl Iterator<Item = VortexResult<Scalar>> + '_>> {
+        vortex_ensure!(index < self.len(), OutOfBounds: index, 0, self.len());
         if !self.validity.is_valid(index)? {
             return Ok(None);
         }
--- a/vortex-array/src/arrays/struct_/array.rs
+++ b/vortex-array/src/arrays/struct_/array.rs
@@ -17,6 +17,7 @@ use crate::dtype::DType;
 use crate::dtype::FieldName;
 use crate::dtype::FieldNames;
 use crate::dtype::StructFields;
+use crate::scalar::Scalar;
 use crate::stats::ArrayStats;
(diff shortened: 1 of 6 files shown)


vortex-array/src/arrays/struct_/array.rs
```diff
@@ -177,6 +178,17 @@ impl StructArray {
...
+    /// Return an iterator over the scalar value of each field at the given row index
+    pub fn scalar_at_fields(
+        &self,
+        index: usize,
+    ) -> VortexResult<impl Iterator<Item = VortexResult<Scalar>> + '_> {
+        Ok(self
+            .unmasked_fields()
+            .iter()
+            .map(move |f| f.scalar_at(index)))
+    }
```
joseph-isaacs: This is in correct and likely to the original one it since it works with unmasked fields

robert3005: in order to fix this comment you will end up doing effectively https://github.com/vortex-data/vortex/pull/6717

joseph-isaacs: Kind, also solved by https://github.com/vortex-data/vortex/pull/5923

joseph-isaacs: In fact both of us are wrong we need to check validity of the struct at index and return None

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -177,6 +178,17 @@ impl StructArray {
         self.struct_fields().find(name).map(|idx| &self.fields[idx])
     }
 
+    /// Return an iterator over the scalar value of each field at the given row index
+    pub fn scalar_at_fields(
+        &self,
+        index: usize,
+    ) -> VortexResult<impl Iterator<Item = VortexResult<Scalar>> + '_> {
```
joseph-isaacs: Should really be `Result<Option<impl ...>>`

vortex-array/src/arrays/struct_/array.rs
palaska: Yeah my bad, didn't notice wrapping scalar_at that does the validity check. I will change the return type following Joe's suggestion.

connortsui20 (approved):
LGTM!

joseph-isaacs:
Lets run benchmarks and merge

joseph-isaacs (approved):

PR #6745, opened 2026-03-02
Configure benchmark boxes to try and reduce noise

AdamGS:
This PR introduces a script to try and cleanup some noise from linux machines, disabling background services and trying to configuring memory behavior better.

### DIFF
--- a/scripts/setup-benchmark.sh
+++ b/scripts/setup-benchmark.sh
@@ -20,14 +20,11 @@ sudo sysctl kernel.randomize_va_space
 dmesg -n 1
 
 # Disable some unused services and features
-sudo systemctl stop apparmor snapd unattended-upgrades multipathd ModemManager
-sudo systemctl disable apparmor snapd unattended-upgrades multipathd ModemManager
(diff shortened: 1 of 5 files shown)


scripts/setup-benchmark.sh
```diff
@@ -0,0 +1,30 @@
...
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+set -Eeu -o pipefail -x
+
+if [ "$EUID" -ne 0 ]; then
+    echo "Environment setup script for benchmarks should run as root."
+    exit 0
+fi
+
+# Really discourage swapping to disk
```
0ax1, line 18: How about turning off frequency scaling?

0ax1 (approved):
Nice! One q inline.

scripts/setup-benchmark.sh
AdamGS: done

AdamGS:
script seems to work, I hope it'll actually show a difference over time

PR #6746, opened 2026-03-02
Parallelize slow CUDA tests

joseph-isaacs:
Split up the patches tests to run in parallel

### DIFF
--- a/vortex-cuda/src/kernel/patches/mod.rs
+++ b/vortex-cuda/src/kernel/patches/mod.rs
@@ -115,30 +115,30 @@ mod tests {
     use crate::CudaSession;
     use crate::kernel::patches::execute_patches;
 
-    #[tokio::test]
-    async fn test_patches() {
-        tokio::join!(
-            test_case::<u8>(),
-            test_case::<u16>(),
-            test_case::<u32>(),
-            test_case::<u64>(),
-            test_case::<i8>(),
-            test_case::<i16>(),
-            test_case::<i32>(),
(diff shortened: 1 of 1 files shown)


vortex-cuda/src/kernel/patches/mod.rs
```diff
@@ -118,25 +117,27 @@ mod tests {
...
-        test_case::<u64>().await;
-
-        test_case::<i8>().await;
-        test_case::<i16>().await;
-        test_case::<i32>().await;
-        test_case::<i64>().await;
-
-        test_case::<f32>().await;
-        test_case::<f64>().await;
+        tokio::join!(
```
AdamGS: if you separate these to individual #[test] cases you'll get more parallelism, as nextest will spawn each in its own process

AdamGS: you might be able to do that with rstest

a10y: yea, i thought the tokio test runtime is single-threaded anyway so idk if this helps us much

a10y: https://docs.rs/tokio/latest/tokio/attr.test.html#current-thread-runtime

AdamGS: or https://github.com/mzabaluev/generic-tests

0ax1 (approved):

PR #6748, opened 2026-03-02
Try and estimate benchmark and run noise when displaying benchmarking data

AdamGS:
## Summary

  This makes benchmark PR comments easier to understand when a whole run is noisy or shifted.

## Changes

  - Replace per-row CV% columns with a simpler noise label: low, medium, high, or unknown
  - Add a run-level drift summary to show when the entire run is broadly faster or slower than the baseline
  - Call out large common-mode drift explicitly when the baseline looks unreliable
  - Render table timings in ms instead of ns to keep the GitHub table readable

## Why

  - Raw CV numbers required statistical interpretation
  - Some runs are globally shifted, so per-query deltas can look meaningful when they are mostly baseline noise
  - The new output is easier to scan and makes unreliable baselines more obvious

This PR was done with the help of Codex


AdamGS:
Closed in favor of #6899

PR #6751, opened 2026-03-02
Make the constant metadata a `Scalar`

connortsui20:
## Summary

Changes the `Metadata` for `ConstantArray` to be a `Scalar`. Note that this is not a breaking change since the `Metadata` is purely an in-memory construct.

## Testing

All of the existing tests should be sufficient.

### DIFF
--- a/vortex-array/src/arrays/constant/vtable/mod.rs
+++ b/vortex-array/src/arrays/constant/vtable/mod.rs
@@ -11,7 +11,6 @@ use vortex_error::vortex_panic;
 use vortex_session::VortexSession;
 
 use crate::ArrayRef;
-use crate::EmptyMetadata;
 use crate::ExecutionCtx;
 use crate::IntoArray;
 use crate::Precision;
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -145,7 +134,17 @@ impl VTable for ConstantVTable {
         let scalar_value = ScalarValue::from_proto_bytes(bytes, dtype)?;
         let scalar = Scalar::try_new(dtype.clone(), scalar_value)?;
```
joseph-isaacs, line 137: This could be on the GPU and there still very slow to copy

connortsui20: I believe we talked about this before, I think we don't care right now? cc @gatesn @onursatici

gatesn: I don't think we care right now. 

We need to figure out the right flat layout logic for GPU + host behavior. Small constant arrays should store their values in the metadata. Larger constant arrays and list arrays should store their values in an array.

This would be using ScalarValue::Array.

The thing we need to figure out is a way for ConstantArray to store this array value in a child array, rather than serialized into its own metadata. If we do this, then list-constants and large constant scalars will remain on the GPU all the way through deserialization.

connortsui20: Note that `ScalarValue::Array` does not exist yet (we want to have it), and this is blocking that

gatesn (approved):

PR #6752, opened 2026-03-02
break: remove all compute public functions

joseph-isaacs:
This PR remove deprecated public functions

# Breaks
We have moved all array::compute function into ArrayBuiltIns


vortex-array/src/compute/mod.rs
```diff
@@ -49,22 +36,12 @@ use crate::scalar::Scalar;
 
 #[cfg(feature = "arbitrary")]
 mod arbitrary;
-mod boolean;
-mod cast;
-#[cfg(feature = "_test-harness")]
```
robert3005, line 54: you probably want to keep this line

robert3005 (approved):

PR #6758, opened 2026-03-02
let claude make PRs

lwwmanning:
## Summary

Closes: #000

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      pull-requests: read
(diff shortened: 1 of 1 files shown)


gatesn (approved):

a10y (changes requested):
couple q's

a10y:
1. Only committers can trigger Claude correct?
1. How do we prevent ourselves from tagging Claude to open a PR, then approving Claude's PR, thus cutting the entire rest of the committers out of making potentially large unreviewed codebase changes?

I guess there's a nice thing here which is that fully vibe-coded things are clearly marked as such b/c Claude is opening the PRs. But I also think that it might make it too easy to get lazy with code reviews.

PR #6761, opened 2026-03-02
Remove sparse metadata prost wrapper

connortsui20:
## Summary

Similar in nature to https://github.com/vortex-data/vortex/pull/6751 and https://github.com/vortex-data/vortex/pull/6759. We want to be able to work with in-memory metadata more easily, and we also want to make sure all scalar deserialization happens is `deserialize` instead of `build`. This is the last of the changes to metadata needed to unblock array-valued scalars in metadata.

## Testing

Existing tests should be sufficient.

### DIFF
--- a/encodings/sparse/src/lib.rs
+++ b/encodings/sparse/src/lib.rs
@@ -10,6 +10,7 @@ use vortex_array::Array;
 use vortex_array::ArrayEq;
 use vortex_array::ArrayHash;
 use vortex_array::ArrayRef;
+use vortex_array::DeserializeMetadata;
 use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
 use vortex_array::Precision;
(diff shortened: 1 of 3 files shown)


encodings/sparse/src/lib.rs
```diff
@@ -134,65 +133,82 @@ impl VTable for SparseVTable {
...
         bytes: &[u8],
-        _dtype: &DType,
+        dtype: &DType,
         _len: usize,
-        _buffers: &[BufferHandle],
+        buffers: &[BufferHandle],
         _session: &VortexSession,
     ) -> VortexResult<Self::Metadata> {
-        Ok(ProstMetadata(SparseMetadata::decode(bytes)?))
+        let patches = PatchesMetadata::decode(bytes)?;
```
gatesn: This has changed from `SparseMetadata::decode` to `PatchesMetadata::decode`

gatesn (changes requested):

connortsui20:
Honestly I think a test that would have caught my mistake would be nice but I'm not really sure of a good way to write one.

encodings/sparse/src/lib.rs
```diff
@@ -60,17 +62,23 @@ mod slice;
 
 vtable!(Sparse);
 
+#[derive(Debug)]
+pub struct SparseMetadata {
+    prost_patches: ProstPatchesMetadata,
```
gatesn: I think this should just be "Patches"

gatesn (approved):

PR #6767, opened 2026-03-03
Fast path append_to_builder ConstantArray

palaska:
Adds a `ConstantVTable::append_to_builder` override that bypasses the default canonicalize-then-copy path for all common leaf dtypes (Null, Bool, Primitive, Decimal, Utf8, Binary).

The default path allocates a full canonical array of n copies and copies it into the builder. For a ConstantArray, this is pure waste. The fast paths instead write the scalar value once and fill n slots directly.

For the newly added micro bench
Screenshot 2026-03-03 at 17 47 06

### DIFF
--- a/vortex-array/benches/chunk_array_builder.rs
+++ b/vortex-array/benches/chunk_array_builder.rs
@@ -13,6 +13,7 @@ use vortex_array::IntoArray;
 use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::BoolArray;
 use vortex_array::arrays::ChunkedArray;
+use vortex_array::arrays::ConstantArray;
 use vortex_array::builders::ArrayBuilder;
 use vortex_array::builders::VarBinViewBuilder;
 use vortex_array::builders::builder_with_capacity;
@@ -118,6 +119,27 @@ fn chunked_varbinview_opt_into_canonical(bencher: Bencher, (len, chunk_count): (
     bencher
         .with_inputs(|| &chunks)
         .bench_refs(|chunk| chunk.to_canonical())
+}
+
+#[divan::bench(args = BENCH_ARGS)]
+fn chunked_constant_i32_append_to_builder(bencher: Bencher, (len, chunk_count): (usize, usize)) {
+    let chunk = make_constant_i32_chunks(len, chunk_count);
+
+    bencher.with_inputs(|| &chunk).bench_refs(|chunk| {
+        let mut builder = builder_with_capacity(chunk.dtype(), len * chunk_count);
+        chunk
+            .append_to_builder(builder.as_mut(), &mut SESSION.create_execution_ctx())
+            .vortex_expect("append failed");
+        builder.finish()
+    })
+}
+
+fn make_constant_i32_chunks(len: usize, chunk_count: usize) -> ArrayRef {
+    // Each chunk is a ConstantArray of i32; dtype is I32/NonNullable via From<i32> for Scalar.
+    (0..chunk_count)
+        .map(|i| ConstantArray::new(i as i32, len).into_array())
+        .collect::<ChunkedArray>()
+        .into_array()
 }
 
 fn make_opt_bool_chunks(len: usize, chunk_count: usize) -> ArrayRef {
--- a/vortex-array/benches/chunk_array_builder.rs
+++ b/vortex-array/benches/chunk_array_builder.rs
@@ -137,7 +137,7 @@ fn chunked_constant_i32_append_to_builder(bencher: Bencher, (len, chunk_count): 
 fn make_constant_i32_chunks(len: usize, chunk_count: usize) -> ArrayRef {
     // Each chunk is a ConstantArray of i32; dtype is I32/NonNullable via From<i32> for Scalar.
     (0..chunk_count)
-        .map(|i| ConstantArray::new(i as i32, len).into_array())
+        .map(|_| ConstantArray::new(42i32, len).into_array())
         .collect::<ChunkedArray>()
         .into_array()
 }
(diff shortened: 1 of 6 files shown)


vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -169,4 +173,37 @@ impl VTable for ConstantVTable {
...
         Ok(constant_canonicalize(array)?.into_array())
     }
+
+    fn append_to_builder(
+        array: &ConstantArray,
+        builder: &mut dyn ArrayBuilder,
+        _ctx: &mut ExecutionCtx,
+    ) -> VortexResult<()> {
+        match array.dtype() {
+            DType::Primitive(ptype, _) => {
```
robert3005, line 205: how about we implement other dtypes?

vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -169,4 +173,37 @@ impl VTable for ConstantVTable {
...
+                    if array.scalar().is_null() {
+                        // SAFETY: append_nulls_unchecked requires a nullable builder.
+                        // A null scalar implies a nullable dtype, so the builder created
+                        // for this array's dtype is necessarily nullable.
+                        unsafe { pbuilder.append_nulls_unchecked(array.len()) }
+                    } else {
+                        let value = P::try_from(array.scalar())
+                            .vortex_expect("Couldn't unwrap constant scalar to primitive");
+                        pbuilder.append_value_n(value, array.len());
+                    }
```
joseph-isaacs: can you pull this out into a function?

vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -169,4 +173,37 @@ impl VTable for ConstantVTable {
...
+                });
+                Ok(())
+            }
+            _ => {
+                let canonical = constant_canonicalize(array)?.into_array();
+                builder.extend_from_array(&canonical);
```
joseph-isaacs: can you add a todo do add others

vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -169,4 +173,37 @@ impl VTable for ConstantVTable {
...
+                    } else {
+                        let value = P::try_from(array.scalar())
+                            .vortex_expect("Couldn't unwrap constant scalar to primitive");
+                        pbuilder.append_value_n(value, array.len());
+                    }
+                });
+                Ok(())
+            }
+            _ => {
+                let canonical = constant_canonicalize(array)?.into_array();
```
joseph-isaacs: Yan should use array.execute::<Canonical>(ctx) here

palaska: execute seemed to be calling `constant_canonicalize` in this case + copying the statistics over, which seemed a bit unnecessary in this case (since we are writing into an existing builder here). I guess it doesn't make a huge difference in terms of perf, I can change

vortex-array/src/arrays/constant/vtable/mod.rs
```diff
@@ -169,4 +173,37 @@ impl VTable for ConstantVTable {
...
+                match_each_native_ptype!(ptype, |P| {
+                    let pbuilder = builder
+                        .as_any_mut()
+                        .downcast_mut::<PrimitiveBuilder<P>>()
+                        .vortex_expect("Expected PrimitiveBuilder for primitive ConstantArray");
+                    if array.scalar().is_null() {
+                        // SAFETY: append_nulls_unchecked requires a nullable builder.
+                        // A null scalar implies a nullable dtype, so the builder created
+                        // for this array's dtype is necessarily nullable.
+                        unsafe { pbuilder.append_nulls_unchecked(array.len()) }
```
joseph-isaacs: Can you use the checked version?

palaska: 👍  I was trying to avoid the overhead of the assertion but it should be negligible.

joseph-isaacs: In fact I think this is okay ignore me

vortex-array/src/arrays/constant/vtable/mod.rs
joseph-isaacs: It will get called by execute

vortex-array/src/builders/primitive.rs
```diff
@@ -51,6 +51,12 @@ impl<T: NativePType> PrimitiveBuilder<T> {
         self.nulls.append_non_null();
     }
 
+    /// Appends `n` copies of `value` as non-null entries, directly writing into the buffer.
+    pub fn append_value_n(&mut self, value: T, n: usize) {
```
robert3005: Having common prefix is cute but reads horribly
```suggestion
    pub fn append_n_values(&mut self, value: T, n: usize) {
```

vortex-array/src/arrays/constant/vtable/mod.rs
palaska: added some more dtypes that are more common in constant arrays. I think others were a bit more complicated so I left them for now.

danking (approved):
LGTM

vortex-array/benches/chunk_array_builder.rs
```diff
@@ -120,6 +121,61 @@ fn chunked_varbinview_opt_into_canonical(bencher: Bencher, (len, chunk_count): (
...
 }
 
+#[divan::bench(args = BENCH_ARGS)]
+fn chunked_constant_i32_append_to_builder(bencher: Bencher, (len, chunk_count): (usize, usize)) {
+    let chunk = make_constant_i32_chunks(len, chunk_count);
+
+    bencher.with_inputs(|| &chunk).bench_refs(|chunk| {
+        let mut builder = builder_with_capacity(chunk.dtype(), len * chunk_count);
+        chunk
+            .append_to_builder(builder.as_mut(), &mut SESSION.create_execution_ctx())
```
joseph-isaacs, line 131: We should pass in to the context to each func

PR #6627, opened 2026-02-20
Duckdb 1.5-variegata

myrrc:
1.5 API changes:

- Add support for reusable dictionaries. Remove duckdb_vx_set_dictionary_vector_id in favor of reusable dictionaries.
- Add estimated_size to duckdb_vx_object_cache_put
- Replace set_variables with set_variable_defaults in database options.
- Replace DUCKDB_STANDARD_VECTOR_SIZE with C++ API constant.
- Add error on Bloom filter pushdown request, otherwise we get a duckdb assertion and a leak https://github.com/vortex-data/vortex/pull/6627#issuecomment-3991614737 when we try to get an ExpressionFilter in duckdb_vx_table_filter_get_expression but it's a Bloom filter
- Build httpfs statically if we're on a commit

https://github.com/vortex-data/vortex/issues/5767

Continuation of https://github.com/vortex-data/vortex/pull/5901

PR #6769, opened 2026-03-03
Self hashjoin SQL test for duckdb-vortex

myrrc:
Testcase which currently breaks https://github.com/vortex-data/vortex/pull/6627


0ax1 (commented):
lgtm, one q. was the file supposed to be named `hasjoin` or `hashjoin`?

0ax1 (approved):

PR #6774, opened 2026-03-04
Vortex TUI web

gatesn:
This PR wraps up Vortex TUI in a web UI

Screenshot 2026-03-03 at 20 40 06

Screenshot 2026-03-03 at 22 45 45

As part of this it rewrites the render loop to be fully async and select over the I/O future and the key inputs.

We'll need a follow up to deploy this to CloudFlare Pages or similar.


robert3005 (approved):
tbh not sure what a good paradigm is for building uis like this, changes look reasonable

PR #6777, opened 2026-03-04
feat: aligned chunked pair iter

joseph-isaacs:
Define pairwise chunked array iterator and use it in zip


vortex-array/src/arrays/chunked/paired_chunks.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+}
+
+impl Iterator for PairedChunks<'_> {
+    type Item = VortexResult<AlignedPair>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Skip empty chunks on either side.
+        while self.lhs_idx < self.left.nchunks() && self.left.chunk(self.lhs_idx).is_empty() {
+            self.lhs_idx += 1;
```
myrrc: Can't we skip to first non-empty chunk in constant time instead of busy looping?

vortex-array/src/arrays/chunked/paired_chunks.rs
```diff
@@ -0,0 +1,258 @@
...
+        let lhs_chunk = self.left.chunk(self.lhs_idx);
+        let rhs_chunk = self.right.chunk(self.rhs_idx);
+
+        let lhs_rem = lhs_chunk.len() - self.lhs_offset;
+        let rhs_rem = rhs_chunk.len() - self.rhs_offset;
+        let take = lhs_rem.min(rhs_rem);
+
+        let lhs_slice = match lhs_chunk.slice(self.lhs_offset..self.lhs_offset + take) {
+            Ok(s) => s,
+            Err(e) => return Some(Err(e)),
```
myrrc: Maybe we should just return None?
`Option<Result<Pair>>` seems like an overkill.

vortex-array/src/arrays/chunked/paired_chunks.rs
```diff
@@ -0,0 +1,258 @@
...
+        let right = ChunkedArray::try_new(vec![], i32_dtype())?;
+
+        let pairs = collect_pairs(&left, &right)?;
+        assert!(pairs.is_empty());
+        Ok(())
+    }
+
+    #[test]
+    #[should_panic(expected = "paired_chunks requires arrays of equal length")]
+    fn test_length_mismatch_panics() {
```
myrrc, line 258: If you choose to retain `Option<Result<Pair>>`, add a test where slicing one of chunks fails

vortex-array/src/arrays/chunked/paired_chunks.rs
joseph-isaacs: You cannot throw away an error

myrrc (approved):

PR #6781, opened 2026-03-04
chore: simplify bit-unpack CUDA generator

0ax1:
Makes the CUDA bitunpack code gen easier to read by using raw strings and generates less CUDA code by using constant template parameters and `pragma unroll`. The performance of all bit-unpacking kernels is unchanged (tested on a GH200).


robert3005 (approved):

vortex-cuda/kernels/src/bit_unpack_16.cu
```diff
@@ -5,32 +5,22 @@
...
-    out[INDEX(12, lane)] = reference;
-    out[INDEX(13, lane)] = reference;
-    out[INDEX(14, lane)] = reference;
-    out[INDEX(15, lane)] = reference;
+template <int BW>
+__device__ void _bit_unpack_16_lane(const uint16_t *__restrict in, uint16_t *__restrict out, uint16_t reference, unsigned int lane);
+
+template <>
+__device__ void _bit_unpack_16_lane<0>(const uint16_t *__restrict in, uint16_t *__restrict out, uint16_t reference, unsigned int lane) {
+    #pragma unroll
```
a10y, line 13: i'm not super familiar with how good compilers are at respecting this. Does nvcc/g++ do a good job generally?

a10y: it looks like there's also variant where you can say how many iterations you want unrolled instead of letting the compiler choose

PR #6785, opened 2026-03-04
fix[vortex-array]: fix panic caused by take on listview during compression

asubiotto:
If a ListView is a child of a Dictionary which is canonicalized during compression, Take is executed on the ListView and subsequently its validity slice. This could result in a lazy FillNull array wrapping the validity which was never executed, causing a panic on serialization "Array vortex.fill_null does not support serialization". This PR executes the validity on rebuild, similar to how offsets/sizes are already executed implicitly via to_primitive.

## Summary

Fixes a panic writing complex nested types -> dict<struct<list>>

## Testing

A high-level regression test was added

### DIFF
--- a/vortex-array/src/arrays/listview/rebuild.rs
+++ b/vortex-array/src/arrays/listview/rebuild.rs
@@ -8,7 +8,9 @@ use vortex_error::VortexResult;
 
 use crate::DynArray;
 use crate::IntoArray;
+use crate::LEGACY_SESSION;
 use crate::ToCanonical;
+use crate::VortexSessionExecute;
 use crate::arrays::ConstantArray;
 use crate::arrays::ListViewArray;
 use crate::builders::builder_with_capacity;
@@ -19,7 +21,6 @@ use crate::dtype::Nullability;
 use crate::match_each_integer_ptype;
 use crate::scalar::Scalar;
 use crate::scalar_fn::fns::operators::Operator;
-use crate::vtable::ValidityHelper;
 
 /// Modes for rebuilding a [`ListViewArray`].
 pub enum ListViewRebuildMode {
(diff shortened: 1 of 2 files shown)


asubiotto:
I still need to clean this PR up a little, but wanted to see if the direction was appropriate. Is there a more ergonomic way to execute the validity? I see `validity.execute()` but I think that requires `LEGACY_SESSION.create_execution_ctx()` or something. Is that preferable?

joseph-isaacs:
Ideally we would fix this at the problematic call sight in the compressor. Since we don't want to force evaluation of the mask in the scan path

vortex-array/src/validity.rs
```diff
@@ -188,8 +188,13 @@ impl Validity {
             Self::AllInvalid => Ok(Self::AllInvalid),
             Self::Array(is_valid) => {
                 let maybe_is_valid = is_valid.take(indices.to_array())?;
-                // Null indices invalidate that position.
-                let is_valid = maybe_is_valid.fill_null(Scalar::from(false))?;
+                // Null indices invalidate that position. Skip the no-op fill_null
+                // when the result is already non-nullable.
+                let is_valid = if maybe_is_valid.dtype().is_nullable() {
+                    maybe_is_valid.fill_null(Scalar::from(false))?
```
joseph-isaacs: This is checked in func def

asubiotto: 👍🏼 I think I didn't have that commit. Removed.

asubiotto:
> Ideally we would fix this at the problematic call sight in the compressor. Since we don't want to force evaluation of the mask in the scan path

I think I might need a pointer. This fill null is a result of a `take` caused by canonicalization of an outer dict in the compressor. I believe we want this fill null to not be eagerly evaluated in a `take` to not affect the general path. AFAIU, we already evaluate offsets/sizes in rebuild and were just not doing so for validity, so in my mind `rebuild` is the place to evaluate this. Do we `rebuild` on scans even if the type matches? That seems expensive for other reasons if so.

joseph-isaacs:
This is the fix: https://github.com/vortex-data/vortex/pull/6797/changes

asubiotto:
> This is the fix: https://github.com/vortex-data/vortex/pull/6797/changes

Ok, thanks! Much cleaner

asubiotto:
Superseded by #6797

PR #6788, opened 2026-03-04
feat[vortex-runend]: handle canonicalizing Ree<Utf8>

asubiotto:
This allows writing these kinds of types. I'm unaware of the motivation behind restricting ree values to just primitive and bool prior to this commit.

## Summary

The motivation for this PR is that I want to write RunEndArrays with these value types.

## Testing

Unit tests for each new canonicalization path.

### DIFF
--- a/encodings/runend/benches/run_end_compress.rs
+++ b/encodings/runend/benches/run_end_compress.rs
@@ -5,12 +5,12 @@
 
 use divan::Bencher;
 use itertools::repeat_n;
-use vortex_array::DynArray;
 use vortex_array::IntoArray;
 use vortex_array::LEGACY_SESSION;
 use vortex_array::RecursiveCanonical;
 use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::PrimitiveArray;
+use vortex_array::arrays::VarBinViewArray;
 use vortex_array::compute::warm_up_vtables;
 use vortex_array::dtype::IntegerPType;
 use vortex_array::validity::Validity;
@@ -76,8 +76,12 @@ fn decompress<T: IntegerPType>(bencher: Bencher, (length, run_step): (usize, usi
     let array = run_end_array.into_array();
 
     bencher
-        .with_inputs(|| &array)
-        .bench_refs(|array| array.to_canonical());
+        .with_inputs(|| (array.clone(), LEGACY_SESSION.create_execution_ctx()))
+        .bench_values(|(array, mut execution_ctx)| {
+            array
+                .execute::<RecursiveCanonical>(&mut execution_ctx)
+                .unwrap()
+        });
 }
 
 #[divan::bench(args = BENCH_ARGS)]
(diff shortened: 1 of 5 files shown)


asubiotto:
Removed the dict arm since it was unnecessary and inverting the order of canonicalization (causing the benchmark regressions). The existing code already works for dicts

encodings/runend/src/array.rs
```diff
@@ -519,13 +493,27 @@ pub(super) fn run_end_canonicalize(
...
+            let num_runs = array.values().len();
+            let indices_values = PrimitiveArray::from_iter(0u64..num_runs as u64);
+            let flat_indices =
+                runend_decode_primitive(pends, indices_values, array.offset(), array.len())?;
+            let values = array
+                .values()
+                .clone()
+                .execute_as::<VarBinViewArray>("values", ctx)?;
+            VarBinViewVTable::take(&values, &flat_indices.into_array(), ctx)?
+                .vortex_expect("take varbinview should not return None")
```
joseph-isaacs: I would think you want a kernel like (`runend_decode_primitive`) instead of a take (which will be much slower). Since you have a `VarBinViewArray` you need do exactly primitive decode on a u128/i128 (the view size).

joseph-isaacs:
Can we have a benchmark similar to the https://github.com/vortex-data/vortex/blob/2ef5750704dd4cfb99871129fc52dec217eb28e5/encodings/runend/benches/run_end_compress.rs#L63

encodings/runend/src/array.rs
joseph-isaacs: I think you might get away with `runend_decode_typed_primitive`. Apply this to VarBinViewArray::views

asubiotto: Done. I had to abstract to `runend_decode_slice` first.

asubiotto:
Thanks for the great suggestions. Implemented.

encodings/runend/benches/run_end_compress.rs
```diff
@@ -113,3 +114,22 @@ fn take_indices(bencher: Bencher, (length, run_step): (usize, usize)) {
...
+
+    let values = VarBinViewArray::from_iter_str((0..num_runs).map(|i| format!("run_value_{i}")))
+        .into_array();
+
+    let run_end_array = RunEndArray::new(ends, values);
+    let array = run_end_array.into_array();
+
+    bencher
+        .with_inputs(|| &array)
+        .bench_refs(|array| array.to_canonical());
```
joseph-isaacs: please use execute::<RecursiveCanonical>().unwrap()

joseph-isaacs: just like above

encodings/runend/src/compress.rs
```diff
@@ -304,6 +326,36 @@ pub fn runend_decode_typed_bool(
     }
 }
 
+/// Decode a run-end encoded VarBinView array by expanding views directly.
+pub fn runend_decode_varbinview(
```
joseph-isaacs, line 330: perfect

joseph-isaacs (approved):
Just one thing about the benchmark

encodings/runend/benches/run_end_compress.rs
asubiotto: Done.

asubiotto:
Updated benchmarks to use execute

robert3005:
I rebased it and fix the one lint

PR #6792, opened 2026-03-05
fix[layout]: add required ALLOWED_ENCODINGS to `WriteStrategyBuilder`

joseph-isaacs:
This is incorrectly set to none allowing all encodings

### DIFF
--- a/vortex-file/src/strategy.rs
+++ b/vortex-file/src/strategy.rs
@@ -139,7 +139,7 @@ impl Default for WriteStrategyBuilder {
             compressor: None,
             row_block_size: 8192,
             field_writers: HashMap::new(),
-            allow_encodings: None,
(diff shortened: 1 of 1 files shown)


AdamGS:
shouldn't the change be that if the builder is `None`, when calling `build` it will use all encodings? What does it even mean to have no allowed encodings?

joseph-isaacs:
None implies Any encoding. I am not sure we want to bake into the compressor an allow list that cannot be updated

gatesn:
No but I think Adam might be right. We should default to all _registered_ encodings (i.e. in the ArrayRegistry), rather than all encodings (which included arrays constructed in-memory but whose encoding was never registered). In theory we could then just not register compute arrays in the session

joseph-isaacs:
Can we merge the fix and follow up

gatesn (approved):

PR #6793, opened 2026-03-05
Branchless decimal sum

AdamGS:
## Summary

Just noticed this code path while refactoring something else, and figured this is better.

I think the trait bounds could be made nicer, but I'll leave that to a bigger refactor.

### DIFF
--- a/vortex-array/src/arrays/decimal/compute/sum.rs
+++ b/vortex-array/src/arrays/decimal/compute/sum.rs
@@ -4,6 +4,7 @@
 use itertools::Itertools;
 use num_traits::AsPrimitive;
 use num_traits::CheckedAdd;
+use num_traits::NumOps;
 use vortex_buffer::BitBuffer;
 use vortex_buffer::Buffer;
 use vortex_error::VortexExpect;
@@ -81,7 +82,8 @@ fn sum_to_scalar<T, O>(
 ) -> Scalar
 where
     T: AsPrimitive<O>,
-    O: Copy + CheckedAdd + Into<DecimalValue> + 'static,
+    O: CheckedAdd + NumOps + Into<DecimalValue> + Copy + 'static,
+    bool: AsPrimitive<O>,
 {
     let raw_sum = match validity {
         Some(v) => sum_decimal_with_validity(values, v, initial),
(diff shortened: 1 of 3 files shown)


vortex-array/src/arrays/decimal/compute/sum.rs
```diff
@@ -109,17 +111,17 @@ fn sum_decimal<T: AsPrimitive<I>, I: Copy + CheckedAdd + 'static>(
...
+    I: NumOps + CheckedAdd + Copy + 'static,
+    bool: AsPrimitive<I>,
+{
     let mut sum = initial;
     for (v, valid) in values.iter().zip_eq(validity) {
-        if valid {
-            let v: I = v.as_();
-            sum = CheckedAdd::checked_add(&sum, &v)?;
-        }
+        let v: I = v.as_() * valid.as_();
```
joseph-isaacs, line 122: do `&`?

AdamGS: wdym?

joseph-isaacs: Instead of `*` by `0` use bitwise and with a bool or T?

AdamGS: it just seems like more work, because it needs to be with T::MAX or 0 instead of 0/1.

AdamGS: not even MAX, just all 1

joseph-isaacs: `&` is much cheeper than `*`. I think the compiler will work it our since there will be a zext u1 to uX, so it know the value is 0/1

vortex-array/src/arrays/decimal/compute/sum.rs
```diff
@@ -109,17 +111,17 @@ fn sum_decimal<T: AsPrimitive<I>, I: Copy + CheckedAdd + 'static>(
...
+
+        sum = CheckedAdd::checked_add(&sum, &v)?;
```
joseph-isaacs, line 124: There is still a branch here tho?

joseph-isaacs (approved):

vortex-array/src/arrays/decimal/compute/sum.rs
AdamGS: there's no BitAnd<bool> for uX

vortex-array/src/arrays/decimal/compute/sum.rs
AdamGS: one branch we can't avoid as long as we only have checked arithmetic

PR #6795, opened 2026-03-05
perf: ensure that `into_array` is used for owned specific arrays

joseph-isaacs:
The pattern `ConstantArray::new(..).to_array()` hides a clone, one should use `into_array`.

We cannot remove this form the DynArray since its required to create a ArrayRef.
This PR shadows the auto-deref  `to_array` on `DynArray` with a deprecated method in the vtable macro.

### DIFF
--- a/vortex-array/src/vtable/mod.rs
+++ b/vortex-array/src/vtable/mod.rs
@@ -339,6 +339,14 @@ macro_rules! vtable {
                     value.into_array()
                 }
             }
+
+            impl [<$V Array>] {
+                #[deprecated(note = "use `.into_array()` (owned) or `.clone().into_array()` (ref) to make clones explicit")]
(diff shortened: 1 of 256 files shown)


vortex-array/src/vtable/mod.rs
```diff
@@ -339,6 +339,14 @@ macro_rules! vtable {
...
                 }
             }
+
+            impl [<$V Array>] {
+                #[deprecated(note = "use `.into_array()` (owned) or `.clone().into_array()` (ref) to make clones explicit")]
+                pub fn to_array(&self) -> $crate::ArrayRef {
+                    use $crate::IntoArray;
+                    self.clone().into_array()
+                }
+            }
```
joseph-isaacs, line 349: see

AdamGS (approved):

PR #6796, opened 2026-03-05
ignore CUDA tests when cuda not available

AdamGS:
Currently the code for CUDA tests is macros away on systems that don't have CUDA installed, which means it can't be linted, making development much more painful with slow cycles trying to make clippy happy in CI.

### DIFF
--- a/vortex-cuda/macros/src/lib.rs
+++ b/vortex-cuda/macros/src/lib.rs
@@ -30,7 +30,6 @@ use std::sync::LazyLock;
 
 use proc_macro::TokenStream;
 use quote::quote;
-use syn::Item;
 use syn::parse_macro_input;
 
 /// Cached result of nvcc availability check.
--- a/vortex-cuda/macros/src/lib.rs
+++ b/vortex-cuda/macros/src/lib.rs
@@ -19,7 +19,7 @@
 //! fn fallback_function() { /* ... */ }
 //!
 //! // Ignore tests when CUDA is not available
-//! #[vortex_cuda_macros::test]
+//! #[crate::test]
 //! async fn my_test() {
 //! ...
 //! }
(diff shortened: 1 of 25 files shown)


vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -553,7 +550,7 @@ mod tests {
         Ok(())
     }
 
-    #[test]
+    #[vortex_cuda_macros::test]
```
0ax1: Can we alias this to `#[vortex_cuda::test]`?

vortex-cuda/macros/src/lib.rs
```diff
@@ -61,17 +60,39 @@ pub fn cuda_not_available(_attr: TokenStream, item: TokenStream) -> TokenStream
...
+pub fn test(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let item = parse_macro_input!(item as syn::ItemFn);
     if *NVCC_AVAILABLE {
-        let item = parse_macro_input!(item as Item);
-        quote! {
-            #[cfg(test)]
-            #item
+        if item.sig.asyncness.is_some() {
+            quote! {
+                #[tokio::test]
```
0ax1, line 72: Do we really want to hard-code `#[tokio::test]` in the marco?

0ax1 (commented):
Seems reasonable. Couple of thoughts inline.

vortex-cuda/src/dynamic_dispatch/mod.rs
AdamGS: "re-exported" it internally, so now its `#[crate::test]`, LMK what you think

vortex-cuda/macros/src/lib.rs
AdamGS: is there another crate that give an async test attribute that just works? that's the only reason here

0ax1 (approved):

PR #6802, opened 2026-03-05
chore[ci]: Enable SPOT for cuda jobs

joseph-isaacs:
Enable SPOT for cuda jobs


0ax1 (commented):
Aren't cuda instances more scarce which might imply long wait times?

joseph-isaacs:
We are working on increase our quotas I believe these are different quotas so it better to have both for now

0ax1 (approved):

a10y (approved):

robert3005:
We also moved the ci to different region which should have more gpu instances available

PR #6804, opened 2026-03-05
CaseWhen uses forward pass with a remaining mask

palaska:
Each condition is ANDed with remaining (unmatched rows), producing disjoint branch masks by construction. This enables early exit when all rows are claimed and skips evaluating THEN expressions for branches with no remaining matches.

Matched branches are merged in a single builder pass instead of N intermediate arrays.

Inspired by [Datafusion's CASE WHEN optimization blog post](https://datafusion.apache.org/blog/2026/02/02/datafusion_case/)

Future work
  - Hash table lookup — detecting WHEN col = const THEN const patterns and compiling them into a vectorized dictionary lookup (I think this needs to live in `simplify`?)
  - Column projection — narrowing the input batch to only columns referenced by a branch's condition before evaluating it (not sure if we need this)

Also fixes a nullability bug in `zip_impl_with_builder` where `AllOr::All`/`AllOr::None` short-circuits bypassed the builder's declared return type, producing a wrong dtype when if_true and if_false had different nullability.

Benches
Before
Screenshot 2026-03-05 at 01 38 14
After
Screenshot 2026-03-05 at 01 33 24


vortex-array/src/scalar_fn/fns/zip/mod.rs
```diff
@@ -204,8 +199,14 @@ fn zip_impl_with_builder(
     mut builder: Box<dyn ArrayBuilder>,
 ) -> VortexResult<ArrayRef> {
     match mask.slices() {
-        AllOr::All => Ok(if_true.to_array()),
-        AllOr::None => Ok(if_false.to_array()),
+        AllOr::All => {
+            builder.extend_from_array(if_true);
```
palaska: this is an unrelated bug fix

joseph-isaacs: what is the fix?

vortex-array/src/scalar_fn/fns/zip/mod.rs
```diff
@@ -296,6 +297,24 @@ mod tests {
         assert_eq!(result.dtype(), if_false.dtype())
     }
 
+    /// When the mask is all-false and `if_true` is Nullable, the result dtype must be Nullable
```
joseph-isaacs: this is odd? The mask doesn't matter here

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -893,6 +1013,89 @@ mod tests {
...
+            vec![
+                (gt(get_item("value", root()), lit(0i32)), lit(100i32)),
+                // Never evaluated due to early exit; 999 must never appear in output.
+                (gt(get_item("value", root()), lit(0i32)), lit(999i32)),
+            ],
+            Some(lit(0i32)),
+        );
+
+        let result = evaluate_expr(&expr, &test_array).to_primitive();
+        assert_eq!(result.as_slice::<i32>(), &[100, 100, 100]);
```
joseph-isaacs: use assert_arrays_eq

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -893,6 +1013,89 @@ mod tests {
...
+        );
+        assert_eq!(
+            result.scalar_at(2)?,
+            Scalar::utf8("high", Nullability::NonNullable)
+        );
+        assert_eq!(
+            result.scalar_at(3)?,
+            Scalar::utf8("high", Nullability::NonNullable)
+        );
+        Ok(())
```
joseph-isaacs, line 1140: can we simplify these tests

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -191,37 +194,45 @@ impl ScalarFnVTable for CaseWhen {
...
+            remaining = &remaining & &(!&cond_mask);
+            branches.push((effective_mask, then_value));
+        }
 
-            result = zip_impl(&then_value, &result, &mask)?;
+        let else_value: ArrayRef = if options.has_else {
+            args.get(num_pairs * 2)?
+        } else {
+            let then_dtype = args.get(1)?.dtype().as_nullable();
+            ConstantArray::new(Scalar::null(then_dtype), row_count).into_array()
```
joseph-isaacs, line 241: Do we need this?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -191,37 +194,45 @@ impl ScalarFnVTable for CaseWhen {
...
+            let condition = args.get(i * 2)?;
             let cond_bool = condition.execute::<BoolArray>(ctx)?;
-            let mask = cond_bool.to_mask_fill_null_false();
+            let cond_mask = cond_bool.to_mask_fill_null_false();
+            let effective_mask = &remaining & &cond_mask;
 
-            if mask.all_true() {
-                result = then_value;
+            if effective_mask.all_false() {
                 continue;
```
joseph-isaacs, line 229: break?

joseph-isaacs:
Cool

joseph-isaacs:
mind if we get this on in first: https://github.com/vortex-data/vortex/pull/6806

vortex-array/src/scalar_fn/fns/zip/mod.rs
palaska: see the test I added (`test_zip_all_false_widens_nullability`)

eg: `zip(if_true: i32?, if_false: i32, mask)` should return a nullable when the mask is all-false, right?

vortex-array/src/scalar_fn/fns/zip/mod.rs
palaska: see my other comment, what is the contract of zip?

https://github.com/vortex-data/vortex/pull/6804#discussion_r2891217592

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: I think so. We need something for the else slot because in `merge_case_branches` it uses `else_value.dtype()` as the base type when computing the return dtype.

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: effective_mask.all_false means we won't pick any rows from this branch and we should keep iterating over the rest of the branches. `continue` is correct here

vortex-array/src/scalar_fn/fns/zip/mod.rs
joseph-isaacs: union of true false. The type cannot be value dependent

vortex-array/src/scalar_fn/fns/zip/mod.rs
joseph-isaacs: I think you should use a cast?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -191,37 +194,45 @@ impl ScalarFnVTable for CaseWhen {
...
             }
 
-            if mask.all_false() {
-                continue;
-            }
+            let then_value = args.get(i * 2 + 1)?;
+            remaining = &remaining & &(!&cond_mask);
```
joseph-isaacs: can we simplify this?

palaska: added `Mask::andnot`, I think its cleaner

vortex-array/src/scalar_fn/fns/zip/mod.rs
palaska: moved the check to `zip_impl` and used `cast()`

vortex-array/src/scalar_fn/fns/zip/mod.rs
joseph-isaacs: can you make this a function of `&MaskValues`?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +258,55 @@ impl ScalarFnVTable for CaseWhen {
...
+    for (start, end, branch_idx) in &events {
+        if builder.len() < *start {
+            builder.extend_from_array(&else_value.slice(builder.len()..*start)?);
+        }
+        builder.extend_from_array(&branches[*branch_idx].1.slice(*start..*end)?);
+    }
+    if builder.len() < row_count {
+        builder.extend_from_array(&else_value.slice(builder.len()..row_count)?);
+    }
+
```
joseph-isaacs: this might be very expensive if the slices are small

joseph-isaacs (changes requested):

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: based on my benchmark sweep (again 😄 ), 4 as the average length of the runs seems to be a good cutoff point to choose between strategies. I know you wanted this optimization to live in zip but as we talked disjointness is not guaranteed there so I feel like it's fine to have it here?

It'd be great if you can double check if my [merge_row_by_row](https://github.com/vortex-data/vortex/pull/6804/changes#diff-97b76364bb1b334b463214cc65df229f823622f9680592b02176d571462710feR307-R340) implementation is optimal.

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+    branches: Vec<(Mask, ArrayRef)>,
+    else_value: ArrayRef,
+) -> VortexResult<ArrayRef> {
+    if branches.len() == 1 {
+        let (mask, then_value) = &branches[0];
+        return zip_impl(then_value, &else_value, mask);
+    }
+
+    let row_count = else_value.len();
+    let mut spans: Vec<(usize, usize, usize)> = Vec::new();
```
joseph-isaacs, line 284: you know the cap?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+            }
+        }
+    }
+    spans.sort_unstable_by_key(|&(start, ..)| start);
+
+    let output_dtype = branches
+        .iter()
+        .fold(else_value.dtype().clone(), |acc, (_, arr)| {
+            acc.union_nullability(arr.dtype().nullability())
+        });
```
joseph-isaacs: do this with a bool and union after lets save a load of Arc allocs vs a bit or

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+        } else {
+            else_value
+        };
+        let scalar = src.scalar_at(row)?;
+        let scalar = if needs_cast && scalar.dtype() != &builder_dtype {
+            scalar.cast(&builder_dtype)?
+        } else {
+            scalar
+        };
+        builder.append_scalar(&scalar)?;
```
joseph-isaacs: Why did you choose to cast each value vs casting the whole array first?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+        let src = if cursor < spans.len() && spans[cursor].0 <= row {
+            &branches[spans[cursor].2].1
+        } else {
+            else_value
+        };
+        let scalar = src.scalar_at(row)?;
+        let scalar = if needs_cast && scalar.dtype() != &builder_dtype {
+            scalar.cast(&builder_dtype)?
+        } else {
+            scalar
```
joseph-isaacs: this condition is loop invariant!

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+    if fragmented {
+        merge_row_by_row(&branches, &else_value, &spans, row_count, builder)
+    } else {
+        merge_run_by_run(&branches, &else_value, &spans, row_count, builder)
+    }
+}
+
+/// Walks rows with a span cursor, emitting one `scalar_at` per row.
+/// Zero per-run allocations; preferred for fragmented masks (avg run < [`SLICE_CROSSOVER_RUN_LEN`]).
+fn merge_row_by_row(
```
joseph-isaacs, line 314: Is this benchmarked?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -198,43 +202,52 @@ impl ScalarFnVTable for CaseWhen {
...
+            let then_dtype = args.get(1)?.dtype().as_nullable();
+            ConstantArray::new(Scalar::null(then_dtype), row_count).into_array()
+        };
+
+        if branches.is_empty() {
+            return Ok(else_value);
         }
 
-        Ok(result)
+        merge_case_branches(branches, else_value)
```
joseph-isaacs, line 248: This really feels like we want an expr that is n-way merge, but that is future work I think

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+    if branches.len() == 1 {
+        let (mask, then_value) = &branches[0];
+        return zip_impl(then_value, &else_value, mask);
+    }
+
+    let row_count = else_value.len();
+    let mut spans: Vec<(usize, usize, usize)> = Vec::new();
+    for (branch_idx, (mask, _)) in branches.iter().enumerate() {
+        match mask.slices() {
+            AllOr::All => spans.push((0, row_count, branch_idx)),
```
joseph-isaacs: Is this is the case can you break and return a single row?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+            AllOr::None => {}
+            AllOr::Some(slices) => {
+                for &(start, end) in slices {
+                    spans.push((start, end, branch_idx));
+                }
+            }
+        }
+    }
+    spans.sort_unstable_by_key(|&(start, ..)| start);
```
joseph-isaacs, line 296: This works since each range is globally disjoint?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +256,111 @@ impl ScalarFnVTable for CaseWhen {
...
+    let needs_cast = branches
+        .iter()
+        .any(|(_, arr)| arr.dtype() != &builder_dtype)
+        || else_value.dtype() != &builder_dtype;
+
+    let mut cursor = 0;
+    for row in 0..row_count {
+        while cursor < spans.len() && spans[cursor].1 <= row {
+            cursor += 1;
+        }
```
joseph-isaacs: This seems very inefficient? Didn't you just sort this?

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: i was dealing with a dtype mismatch error and I added this as a fix but you're right, we should cast each source array once

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: Yepp

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: can we do something else other than `append_scalar(branch_arr.scalar_at(...))` once per row? I guess you are saying inefficient because of the inner while loop which just increments a counter. I refactored it so it's better structured now but it's still does n scalar_at calls

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: Yes but only for primitive arrays (avg length of 4 is calibrated using primitive arrays). I guess the cost will differ based on the underlying encodings of the arrays. Probably a lower cutoff will work better when `scalar_at` is more expensive. Not sure what the best move here is.

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +257,113 @@ impl ScalarFnVTable for CaseWhen {
...
+    }
+}
+
+/// Iterates spans directly, emitting one `scalar_at` per row.
+/// Zero per-run allocations; preferred for fragmented masks (avg run < [`SLICE_CROSSOVER_RUN_LEN`]).
+fn merge_row_by_row(
+    branch_arrays: &[ArrayRef],
+    else_value: &ArrayRef,
+    spans: &[(usize, usize, usize)],
+    row_count: usize,
```
joseph-isaacs: else_value.len()?

vortex-array/src/scalar_fn/fns/zip/mod.rs
```diff
@@ -189,34 +185,37 @@ pub(crate) fn zip_impl(
         .dtype()
         .clone()
         .union_nullability(if_false.dtype().nullability());
+
+    if mask.all_true() {
+        return if_true.cast(return_type);
+    }
+    if mask.all_false() {
+        return if_false.cast(return_type);
+    }
```
joseph-isaacs, line 194: This was already checkout in the zip expr and since this is a private func it shouldn't be changed. Can add a wrapper method of check in the call site?

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -243,6 +257,110 @@ impl ScalarFnVTable for CaseWhen {
...
+    if spans.is_empty() {
+        return else_value.cast(output_dtype);
+    }
+
+    let builder = builder_with_capacity(&output_dtype, else_value.len());
+
+    let branch_arrays: Vec<ArrayRef> = branches
+        .iter()
+        .map(|(_, arr)| arr.cast(output_dtype.clone()))
+        .collect::<VortexResult<_>>()?;
```
joseph-isaacs: I think you want to apply this cast here for the scalar_at part? I am not sure if the lazy cast matters with scalar at

joseph-isaacs (commented):
One last one with the zip impl

vortex-array/src/scalar_fn/fns/case_when.rs
palaska: Cast doesn't seem to have a scalar_at override so it falls through the default so scalar_at on a lazy cast array executes. I think it makes sense to cast after scalar_at calls

vortex-array/src/scalar_fn/fns/zip/mod.rs
palaska: Should I put back the check to `zip_impl_with_builder` then or do we just assert that passed mask can never be all_true or all_false?

joseph-isaacs: No is okay

joseph-isaacs (approved):

joseph-isaacs (changes requested):
The benchmarks are too large they take over a few ms

joseph-isaacs (approved):

PR #6806, opened 2026-03-05
fix: don't commit to case when binary format

joseph-isaacs:
We want to bad the serialisation of expr incase we which to change the format

### DIFF
--- a/vortex-array/src/scalar_fn/fns/case_when.rs
+++ b/vortex-array/src/scalar_fn/fns/case_when.rs
@@ -1,7 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-//! N-ary CASE WHEN expression for conditional value selection.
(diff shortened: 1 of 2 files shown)


robert3005 (approved):

vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -70,8 +77,10 @@ impl ScalarFnVTable for CaseWhen {
     }
 
     fn serialize(&self, options: &Self::Options) -> VortexResult<Option<Vec<u8>>> {
-        let num_children = options.num_when_then_pairs * 2 + u32::from(options.has_else);
-        Ok(Some(pb::CaseWhenOpts { num_children }.encode_to_vec()))
+        // let num_children = options.num_when_then_pairs * 2 + u32::from(options.has_else);
+        // Ok(Some(pb::CaseWhenOpts { num_children }.encode_to_vec()))
+        // stablize the expr
```
palaska:
```suggestion
        // stabilize the expr
```

palaska:
@joseph-isaacs is this good to go?

PR #6808, opened 2026-03-05
Layout readers don't need a mutable context

robert3005:
When reading the set of array ids is initialised from the file and then is
immutable for whole scan duration. We don't need to share the mutable version
with immutable one.


encodings/fastlanes/src/rle/array/mod.rs
```diff
@@ -229,6 +229,7 @@ mod tests {
     use vortex_array::validity::Validity;
     use vortex_buffer::Buffer;
     use vortex_buffer::ByteBufferMut;
+    use vortex_session::registry::ReadContext;
```
robert3005, line 232: I'm open to a better name

AdamGS (approved):

PR #6810, opened 2026-03-05
Remove reuse lint check

connortsui20:
## Summary

For some reason, `reuse lint` has slowly gotten slower and slower to run on my machine over the past few months. To the point where I can't be bothered to run it anymore pre commit.

## Testing

N/A


a10y:
i didn't even realize we had a pre-commit script in the repo. Does it need to live in here?

connortsui20:
I believe I am the only person who actually uses this, but I thought might as well share it here since it doesn't hurt anyone.

robert3005 (approved):

PR #6812, opened 2026-03-05
Vortex Fixed-Shape Tensor

connortsui20:
## Summary

Adds an experimental fixed-shape tensor extension type in a new `vortex-tensor` crate.

See https://vortex-data.github.io/rfcs/rfc/0024.html for info about the design of this tensor type.

Additionally adds a `CosineSimilarity` expression that takes 2 tensor arrays and computes the cosine similarity of tensors in the arrays (resulting in a `PrimitiveArray`).

## Testing

Adds some very basic tests for cosine similarity and tensor metadata operations.

## Future Work

I think this was a good way to see if our `ExtVTable` is not completely wrong, but at the same time this tells us nothing about what we might want to add for extension arrays on the `ExtVTable` because we as long as the storage `DType` is correct, any storage array is valid.

The more interesting expressions have not been implemented here. Those would include:

- Cast
- Index / Slice (and lazily get back another tensor array, potentially non-contiguous)
- Maybe others?

Additional work includes exporting to Arrow, NumPy, and PyTorch. Arrow will require a cheap translation from logical to physical shape, but other than that those conversions should be easy.

### DIFF
--- a/vortex-tensor/src/proto.rs
+++ b/vortex-tensor/src/proto.rs
@@ -11,12 +11,26 @@ use vortex::error::vortex_err;
 use crate::FixedShapeTensorMetadata;
 
 /// Protobuf representation of [`FixedShapeTensorMetadata`].
+///
+/// Protobuf does not distinguish between an absent repeated field and an empty one (both will
+/// deserialize as an empty `Vec`). This is fine because the semantic meaning is unambiguous:
+///
+/// - `logical_shape` empty: 0-dimensional (scalar) tensor.
+/// - `dim_names` empty: no dimension names (`None`).
+/// - `permutation` empty: no permutation, i.e., identity layout (`None`).
 #[derive(Clone, PartialEq, Message)]
 struct FixedShapeTensorMetadataProto {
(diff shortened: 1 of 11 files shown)


vortex-tensor/src/proto.rs
```diff
@@ -0,0 +1,71 @@
...
+    let proto = FixedShapeTensorMetadataProto::decode(bytes).map_err(|e| vortex_err!("{e}"))?;
+
+    let logical_shape = proto
+        .logical_shape
+        .into_iter()
+        .map(|d| d as usize)
+        .collect();
+    let mut m = FixedShapeTensorMetadata::new(logical_shape);
+
+    if !proto.dim_names.is_empty() {
```
gatesn, line 81: `if proto.dim_names.len() != logical_shape.len`

You need to allow present dim_names if logical_shape has zero dimensions

gatesn (commented):
Looks great! Maybe move cosine similarity into a `scalar_fns` module since we're going to add more.

You should also get public-api to work somehow?

gatesn (approved):

PR #6815, opened 2026-03-06
[wip] PatchedArray

a10y:
## Summary

- [x] new PatchedArray
- [x] On read, BitPackedArray w/Patches -> PatchedArray(BitPackedArray)
- [x] Slice
- [x] Filter
- [x] Compare
- [ ] Take
- [ ] ALP
- [ ] ALP-RD (?)

Closes: #000

### DIFF
--- a/encodings/alp/src/alp/array.rs
+++ b/encodings/alp/src/alp/array.rs
@@ -158,7 +158,7 @@ impl VTable for ALPVTable {
         metadata: &Self::Metadata,
         _buffers: &[BufferHandle],
         children: &dyn ArrayChildren,
-    ) -> VortexResult<ALPArray> {
+    ) -> VortexResult<ArrayRef> {
         let encoded_ptype = match &dtype {
             DType::Primitive(PType::F32, n) => DType::Primitive(PType::I32, *n),
             DType::Primitive(PType::F64, n) => DType::Primitive(PType::I64, *n),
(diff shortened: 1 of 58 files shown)


gatesn:
Moving my comment from another PR:

Unless we have some way of passing down an "is_defined" mask into execution, there's no way for the child array to know whether some checked operation should fail or not.

e.g. the child of a Patches array may overflow on some arithmetic, but the patch would have succeeded. We need a way to tell the child to ignore the overflow for the invalid positions.

a10y:
do you think that's something we could ferry in the executionctx?

a10y:
Alternatively you can just not pushdown scalarfns to the child and force it to be done after patching? It's clear to me why you'd wanna push down cardinality reducing ops (e.g. Slice) but idk why it matters if fallible scalar fns like Sum or Binary need the patched result

gatesn:
The push-down makes sense, but it seems a shame that we cannot have a specialized kernel for ALP::compare because there's a patch array in the middle. I guess compare _can_ be pushed down? Because it isn't a checked operation?

In fact, I think we already have this defined as `ScalarFnVTable::is_fallible`

gatesn:
Other option ofc is `ArrayVTable::execute(selection: Mask, defined: Mask)` - but 🤷

a10y:
Oh interesting, I didn't realize we already reported fallibility. Then in that case yea, think we can push infallible expressions through to the child

PR #6824, opened 2026-03-06
feat: Iterative Execution

joseph-isaacs:
Changes the model of execution from internal iterator to external.

We go from a recursive execution to an iterative execution with a explicit stack and VTable requesting execution on a child.

This can help with stack overflow on high-depth array trees.

## Breaks
Changed the execute method. Likely unused externally. 

Follow ups:

- Will remove the usage of recursive calls from all arrays.
- Will change the execution method to take a Session so its easier to see internal execution.


gatesn (approved):
Please include details in the PR description and fix the title!

robert3005 (approved):
I think we need a lighter way of replacing children, with_child feels like a costly method

PR #6827, opened 2026-03-06
No longer need to install visual studio for windows machines

robert3005:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -649,17 +649,10 @@ jobs:
         if: github.repository == 'vortex-data/vortex'
         with:
           sccache: s3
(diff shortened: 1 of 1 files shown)


a10y (approved):
❤️

robert3005:
seems that we are ahead of the curve and this is not built yet, the pr only got merged like 40 mins ago

joseph-isaacs (changes requested):
It didn't work

robert3005:
we need a new release here https://github.com/runs-on/runner-images-for-aws/releases

joseph-isaacs:
can we do that?

robert3005:
We no longer need this

PR #6829, opened 2026-03-06
gpu: transpose patches faster

a10y:
Eliminate all of the wasteful allocating and deallocating, replace it with a 3-pass algorithm that reuses the buffers.

Drops the CPU transpose time by a factor of 5-10 in my benchmarks on the other branch

Tested using existing test suite

### DIFF
--- a/vortex-cuda/src/kernel/patches/types.rs
+++ b/vortex-cuda/src/kernel/patches/types.rs
@@ -311,8 +311,7 @@ mod tests {
         test_case(512, 512, &[512, 513, 514], &[10i8, 20, 30])?;
         test_case(10_000, 100, &[500, 1_000, 1_001, 1_002], &[1i16, 2, 3, 4])?;
 
-        // Try every size from 0..4096 and a range of indices.
-        for len in (1..4096).step_by(20) {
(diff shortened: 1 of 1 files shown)


vortex-cuda/src/kernel/patches/types.rs
```diff
@@ -193,30 +162,56 @@ fn transpose<I: IntegerPType, V: NativePType>(
...
+        lane_offsets[chunk * n_lanes + lane + 1] += 1;
+    }
 
+    // Prefix-sum sizes -> offsets
+    for index in 1..lane_offsets.len() {
+        lane_offsets[index] += lane_offsets[index - 1];
+    }
+
+    // Loop over patches, writing them to final positions
+    let indices_out = indices_buffer.spare_capacity_mut();
```
0ax1, line 202: Should we add a broader test coverage for `transpose`?
- with non-zero offset
- different value types, in particular 64bit as that changes the lane count to 16
- array len % 1024 != 0

0ax1 (approved):
Looks great! Added one comment about adding more tests inline.

a10y:
I just pushed a test case generator, will merge once I verify it doesn't hurt CI times

PR #6830, opened 2026-03-06
Add TPC-H to sqllogictests

AdamGS:
This PR adds TPC-H to to our SLT suite, it requires generating data (script provided), and currently doesn't work due to subtle formatting differences between the crates.

One reason for those differences is that we actually run the queries on different schemas as they each infer a different schema.

I've also added some duckdb functionality as I tried to get it working, and I figure its worth merging it anyway.


.github/workflows/ci.yml
```diff
@@ -821,8 +821,13 @@ jobs:
         uses: ./.github/actions/setup-rust
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install uv
```
myrrc, line 824: We set up uv to upload coverage results, worth removing it from there to avoid duplicate installation

vortex-duckdb/src/duckdb/value.rs
```diff
@@ -264,6 +272,11 @@ pub fn i128_from_parts(high: i64, low: u64) -> i128 {
     ((high as i128) << 64) | (low as i128)
 }
 
+#[inline]
```
myrrc, line 275: Maybe worth defining in #[tests]?

vortex-sqllogictest/slt/tpch/generate_data.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
myrrc, line 1: Not a blocker, but I'd avoid using bash if we can go with sh.
I see only &> (can be replaced with 1>/dev/null 2>&1) and
BASH_SOURCE. Replacing the second one is more tricky but I think something like $_
would work.

myrrc (approved):
LGTM sans minor changes.

vortex-duckdb/src/convert/scalar.rs
```diff
@@ -253,6 +253,9 @@ impl<'a> TryFrom<&'a ValueRef> for Scalar {
             ExtractedValue::HugeInt(_) => {
                 vortex_bail!("DuckDB HugeInt is not yet supported in Vortex");
             }
+            ExtractedValue::UHugeInt(_) => {
+                vortex_bail!("DuckDB UHugeInt is not yet supported in Vortex");
```
joseph-isaacs, line 257: is this u128 int? We might add this soon.  

Good to know duckdb has it

joseph-isaacs (approved):

vortex-duckdb/src/convert/scalar.rs
AdamGS: yeah duckdb seems to have both

.github/workflows/ci.yml
AdamGS: that's a different job IMO, runs on a different machine

vortex-duckdb/src/duckdb/value.rs
AdamGS: We use it in `ExtractedValue`

PR #6831, opened 2026-03-06
use full column set on empty projection list in duckdb

myrrc:
- Use full column list if input post filter projection list is empty as
  requested by DuckDB. Before we used the empty list which resulted in
  uninitialized read (columns requested but not populated).
- Reset DataChunk in ArrayExporter as done by Parquet reader.
  Not needed right now, but must be done once we start estimating cardinalities
- generate comments for Rust bindings in bindgen
- In duckdb_vx_node_statistics, reorder fields to reduce size 32->24.
- Explain why and how we use EMPTY_COLUMN_ID in comments
- Add tests for projection pushdown, including zero projections


vortex-duckdb/cpp/include/duckdb_vx/table_function.h
```diff
@@ -62,22 +62,41 @@ void duckdb_vx_string_map_free(duckdb_vx_string_map map);
...
+     * Indices in this list reference values from projection_ids. I.e. if
+     * projection_ids=[1,5,6], post_filter_projection_ids=[1], output column
+     * should be projection_ids[1] = 5
+     *
+     * In DuckDB API it's called just "projection_ids", renamed to resolve ambiguity.
+     *
+     * Example usage:
+     * https://github.com/duckdb/duckdb/blob/dc11eadd8f0a7c600f0034810706605ebe10d5b9/src/include/duckdb/function/table_function.hpp#L147
+     */
+    const idx_t *post_filter_projection_ids;
```
joseph-isaacs: i really think we should keep the naming here and explain the semantics. 

It nice to keep the duckdb name

vortex-duckdb/src/datasource.rs
```diff
@@ -193,27 +205,29 @@ impl<T: DataSourceTableFunction> TableFunction for T {
...
-            "Global init Vortex scan SELECT {} WHERE {}",
-            &projection_expr,
-            filter_expr
-                .as_ref()
-                .map_or_else(|| "true".to_string(), |f| f.to_string())
-        );
+        let filter_expr_str = filter_expr
+            .as_ref()
+            .map_or_else(|| "true".to_string(), |f| f.to_string());
+        debug!("Global init Vortex scan SELECT {projection_expr} WHERE {filter_expr_str}");
```
joseph-isaacs, line 227: will this not alloc the str on even if debug ignores it?

vortex-duckdb/src/datasource.rs
```diff
@@ -455,35 +470,45 @@ fn extract_schema_from_dtype(dtype: &DType) -> VortexResult<(Vec<String>, Vec<Lo
...
+    select(names, root())
 }
 
 /// Creates a table filter expression from the table filter set, column metadata, additional
 /// filter expressions, and the top-level DType.
 fn extract_table_filter_expr(
     table_filter_set: Option<&TableFilterSetRef>,
-    column_ids: &[u64],
-    column_names: &[String],
+    projection_ids: &[u64],
```
joseph-isaacs: i find this naming more confusing, column seems to make more sense here. i think

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -74,26 +77,30 @@ impl ArrayExporter {
     ///
     /// Returns `true` if a chunk was exported, `false` if all rows have been exported.
     pub fn export(&mut self, chunk: &mut DataChunkRef) -> VortexResult<bool> {
+        chunk.reset();
```
joseph-isaacs, line 80: are you sure we need this?

why if we estimate card

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -44,6 +44,8 @@ use crate::duckdb::VectorRef;
 
 pub struct ArrayExporter {
     ctx: ExecutionCtx,
+    /// Columns DuckDB requested to read from file. If empty, it's a zero-column
+    /// projection and should be handled accordingly, see ArrayExporter::export.
     fields: Vec<Box<dyn ColumnExporter>>,
```
joseph-isaacs, line 49: Could Model this with an enum? It might be easier to understand?

vortex-duckdb/src/datasource.rs
```diff
@@ -455,35 +470,45 @@ fn extract_schema_from_dtype(dtype: &DType) -> VortexResult<(Vec<String>, Vec<Lo
...
-                    .vortex_expect("prune idx in column names")
-            })
-            .map(|s| Arc::from(s.as_str()))
-            .collect::<FieldNames>(),
-        root(),
-    )
+    // Post filter projection ids may be empty, in which case
+    // you need to use projection_ids
+    // https://github.com/duckdb/duckdb/blob/6e211da91657a94803c465fd0ce585f4c6754b54/src/planner/operator/logical_get.cpp#L168
+    let (post_filter_projection_ids, has_projection_ids) = match post_filter_projection_ids {
```
joseph-isaacs: so this is a projection_id then?

joseph-isaacs (commented):
So if we (where to) always remove all filter column then we never need to differentiate between column and projection? right 


I am still a little unclear about the email of columns_ids and projection_ids

vortex-duckdb/src/exporter/mod.rs
myrrc: I've thought about this, but then we need something like NonzeroVec, and I believe it's an overkill.

myrrc:
> I really think we should keep the naming here and explain the semantics.

Ok, I had a second thought, and we can keep the previous names.

vortex-duckdb/src/datasource.rs
myrrc: I'm sure compiler will move this to the debug call, but IMHO this looks better

vortex-duckdb/src/exporter/mod.rs
myrrc: Existing table function implementations like Parquet do it for some reason, so I think this won't harm

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -74,26 +77,30 @@ impl ArrayExporter {
...
-            // uninitialized and just return a DataChunk with the correct length.
-            // One place no fields can occur is in count(*) queries.
-            chunk.set_len(self.remaining);
-            self.remaining = 0;
-
-            return Ok(true);
+        let expected_cols = self.fields.len();
+        let chunk_cols = chunk.column_count();
+        let zero_projection = expected_cols == 0;
+        if !zero_projection && chunk_cols != expected_cols {
```
joseph-isaacs, line 88: Why is this the validation condition?

joseph-isaacs (approved):
LG

vortex-duckdb/src/exporter/mod.rs
myrrc: To guard against bugs in our reader vs. data chunk columns. However, if we're requested a zero projection, we can't validate whether it's our bug or we don't need to populate the output data chunk.

PR #6832, opened 2026-03-06
Add a UUID extension type

connortsui20:
## Summary

Tracking Issue: https://github.com/vortex-data/vortex/issues/6854

Adds a UUID extension type to Vortex over a `FixedSizeList<u8, 16>`. There isn't much functionality implemented on top of this, and there is no exporter to Arrow's [UUID canonical extension type](https://arrow.apache.org/docs/format/CanonicalExtensions.html#uuid). All of the logic is delegated to the Rust `uuid` crate.

There is still a question of if we want to bring in `FixedSizeBinary` instead of doing this.

## Testing

Some small basic tests. They serve mostly as example code on how to work with Uuid extension arrays in general.

## Open Questions

- Do we actually need `FixedSizeBinary` instead here? Why does this not suffice?
- Is there actually an alignment issue here?
- The performance of `unpack_native` is terrible because we still do not have `ScalarValue::Array` (see https://github.com/vortex-data/vortex/pull/6717)
- Should we name this something other than `vortex.uuid` since we know we might change the storage type in the future?
- `UuidFromString` should really be implemented under the `Cast` expression, but we do not have a system yet for pluggable expressions so it is not entirely clear how we would do that.
- What other expressions should we have on Uuid? `UuidToString` is an easy next one, what else is there?


vortex-array/src/extension/uuid/metadata.rs
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use std::fmt;
+
+/// Metadata for the UUID extension type, which is empty.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+pub struct UuidMetadata;
```
gatesn: I think we have `EmptyMetadata` struct. The type ID is already printed, so this display is duplicative.

gatesn: Alternatively, you may want to store an `Option<Version>` for the UUID version, if specified

vortex-array/src/extension/uuid/vtable.rs
```diff
@@ -0,0 +1,205 @@
...
+        Ok(UuidMetadata)
+    }
+
+    fn validate_dtype(
+        &self,
+        _metadata: &Self::Metadata,
+        storage_dtype: &DType,
+    ) -> VortexResult<()> {
+        let DType::FixedSizeList(element_dtype, list_size, _nullability) = storage_dtype else {
+            vortex_bail!("UUID storage dtype must be a FixedSizeList, got {storage_dtype}");
```
gatesn, line 55: You could do all this in a single `matches!` - worse error messages though

vortex-array/src/extension/uuid/vtable.rs
```diff
@@ -0,0 +1,205 @@
...
+    ) -> VortexResult<Self::NativeValue<'a>> {
+        let elements = storage_value.as_list();
+        vortex_ensure_eq!(
+            elements.len(),
+            UUID_BYTE_LEN,
+            "UUID scalar must have exactly {UUID_BYTE_LEN} bytes, got {}",
+            elements.len()
+        );
+
+        let mut bytes = [0u8; UUID_BYTE_LEN];
```
gatesn, line 95: 🤮  - can't wait for #6717

vortex-array/src/scalar_fn/fns/uuid_from_string.rs
```diff
@@ -0,0 +1,330 @@
...
+use crate::scalar_fn::ExecutionArgs;
+use crate::scalar_fn::ScalarFnId;
+use crate::scalar_fn::ScalarFnVTable;
+
+/// Parses a UTF-8 string column into a [`Uuid`] extension array.
+///
+/// Accepts any standard UUID string format (hyphenated, simple, braced, URN). Invalid strings
+/// cause an error.
+#[derive(Clone)]
+pub struct UuidFromString;
```
gatesn: Hmm, I really do think we should just go directly to pluggable casting. CC @joseph-isaacs

vortex-array/src/extension/uuid/metadata.rs
connortsui20: this follows the arrow extension type, and given that uuid has a self describing versioning I think we don't need to do this?

connortsui20: oh whoops I forgot that we have 2 different `EmptyMetadata`

vortex-array/src/extension/uuid/vtable.rs
connortsui20: yes I want to have good error messages in here

vortex-array/src/extension/uuid/vtable.rs
connortsui20: 😭

vortex-array/src/extension/uuid/metadata.rs
connortsui20: actually if we have that info we can potentially compress better (which arrow doesn't care about), so we probably want to have that metadata

connortsui20:
Ok so I'm going to remove the `UuidFromString` scalar fn and add a `Option<Version>` using https://docs.rs/uuid/latest/uuid/enum.Version.html as metadata

vortex-array/src/extension/uuid/metadata.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Creates metadata with no version constraint.
+    pub fn any() -> Self {
+        Self { version: None }
+    }
+}
+
+impl fmt::Display for UuidMetadata {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.version {
+            None => write!(f, "UUID"),
```
gatesn: Get rid of UUID, just print the version number

vortex-array/src/extension/uuid/metadata.rs
```diff
@@ -0,0 +1,70 @@
...
+/// sort-random). When `None`, the column may contain any mix of versions.
+#[derive(Clone, Debug)]
+pub struct UuidMetadata {
+    /// The UUID version, if known.
+    pub version: Option<Version>,
+}
+
+impl UuidMetadata {
+    /// Creates metadata with no version constraint.
+    pub fn any() -> Self {
```
gatesn: Remove this, just derive Default on the metadata struct

gatesn (changes requested):

asubiotto:
> there is no exporter to Arrow's [UUID canonical extension type](https://arrow.apache.org/docs/format/CanonicalExtensions.html#uuid).

Just to confirm, is the reason just to keep this PR contained?

gatesn (approved):

PR #6834, opened 2026-03-06
Public array modules

gatesn:
We were running into naming conflicts by using `pub use primitive::*` etc for all arrays. This PR instead exports the modules as pub, and to minimize the user-facing break continues to explicitly re-export the FooArray and FooVTable for each array.


connortsui20 (approved):
yolo

PR #6840, opened 2026-03-07
refactor(layout): add shared mask/row_range debug assertions

caicancai:
This PR adds a shared debug-only assertion that mask.len() matches row_range length, and applies it across dict, chunked, and flat layout readers.

Background: these readers assume a 1:1 mapping between mask positions and rows in the evaluated range.

Why: when that contract is violated upstream, errors can surface later and be hard to debug. This makes the invariant explicit and catches mismatches at the reader boundary, with no release-runtime cost.

### DIFF
--- a/vortex-layout/src/reader.rs
+++ b/vortex-layout/src/reader.rs
@@ -15,6 +15,7 @@ use vortex_array::builtins::ArrayBuiltins;
 use vortex_array::dtype::DType;
 use vortex_array::dtype::FieldMask;
 use vortex_array::expr::Expression;
+use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
 use vortex_mask::Mask;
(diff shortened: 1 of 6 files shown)


vortex-layout/src/reader.rs
```diff
@@ -91,6 +92,33 @@ pub trait LayoutReader: 'static + Send + Sync {
...
+        .vortex_expect("row range length must fit usize");
+        debug_assert_eq!(
+            mask_len, row_range_len,
+            "{context}: mask length must match row range length"
+        );
+    }
+
+    #[cfg(not(debug_assertions))]
+    {
+        let _ = (mask_len, row_range, context);
```
joseph-isaacs: This is still keep around the &str no? Or do you think it will be `DCE`ed

caicancai: @joseph-isaacs Good catch — this is just an unused-arg sink for non-debug builds and should DCE. I’ll update it to a cfg-split/underscore-arg form for clarity.

caicancai:
@joseph-isaacs @robert3005 I'm sorry to bother you again, but it seems this PR has no further developments. If this PR isn't suitable, I can close it.

robert3005:
Things would fail a lot earlier if mask and row range weren't aligned. I think asserting in these methods doesn't help. What we could do is remove one of these arguments but would have to check how exactly they are used. There's also an option to have sparse masks which would make these assertions exactly opposite of what's required.

PR #6842, opened 2026-03-08
Improve URL parsing in VortexMultiFileScan to handle file paths more …

### LINKED ISSUE #6841: Relative paths cause `Neither URL nor path error` because `Url::from_file_path` requires absolute paths
## Description

When using a relative path with the Vortex DuckDB integration, the system returns the error:

`Neither URL nor path ...`

This appears to be caused by the following code:

https://github.com/vortex-data/vortex/blob/5ec8d2f2492753d415a888f67857eac288f7625c/vortex-duckdb/src/multi_file.rs#L44

The implementation uses:

`Url::from_file_path(...)`

However, [Url::from_file_path](https://docs.rs/url/latest/url/struct.Url.html#method.from_file_path) only accepts absolute paths. When a relative path is provided, the function returns `Err`, which leads to the observed error.

## Reproduction

Example usage with a relative path with Vortex 0.61.0:

`SELECT * FROM read_vortex('./data/file.vortex');`

### Result:

`query failed: Binder Error: Other error: Neither URL nor path: 'data/file.vortex'`
Using an absolute path works correctly.

## Root Cause

`Url::from_file_path` requires an absolute filesystem path, but the code does not normalize relative paths before passing them to it.

## Proposed Fix

Convert relative paths to absolute paths before calling Url::from_file_path, for example using:

```rust
let glob_url = Url::parse(glob_url_str).or_else(|_| {
            let path = Path::new(glob_url_str);

           let path = if glob_url_str.contains(['*', '?', '[']) {           
                path.to_path_buf()
            } else {
                std::fs::canonicalize(path)
                    .map_err(|_| vortex_err!("Neither URL nor path: '{}'", glob_url_str))?
            };

            Url::from_file_path(path)
                .map_err(|_| vortex_err!("Neither URL nor path: '{}'", glob_url_str))
        })?;
```

While `std::fs::canonicalize` requires the file to exist, this should not be a problem in this context because the file must exist in order to be read anyway.

Related discussion:
https://github.com/vortex-data/vortex/issues/6599

## Summary

The error occurs because relative paths are passed directly to Url::from_file_path.
Normalizing the path to an absolute path (e.g. via std::fs::canonicalize) should resolve the issue.

### Steps to reproduce

Cannot reproduce it with current DuckDB extension since the bug has not been introduced to it, I have a repo to build Vortex 0.61.0 DuckDB extension.  https://github.com/Lychee-Technology/duckdb-static

### Environment

- Vortex 0.61.0
- Linux

### Additional context

_No response_

iceboundrock:
## Summary

This PR fixes the issue that relative paths cause Neither URL nor path error

Closes: #6841

This PR utilizes AI (codex from OpenAI) to generate code.

## API Changes

None

## Testing

`cargo test -p vortex-duckdb --lib`


vortex-duckdb/src/multi_file.rs
```diff
@@ -19,6 +19,60 @@ use crate::duckdb::ClientContextRef;
...
+}
+
+/// Canonicalize the non-glob prefix of `path_str` and return the resulting absolute path.
+///
+/// For paths without any glob characters the whole path is canonicalized. For paths that contain
+/// glob characters (`*`, `?`, `[`) the directory portion that precedes the first glob character
+/// is canonicalized and the remaining glob suffix is appended unchanged. This allows relative glob
+/// patterns such as `./data/*.vortex` to be resolved correctly.
+fn canonicalize_path_prefix(path_str: &str) -> VortexResult<PathBuf> {
+    let first_glob_char_pos = path_str.find(['*', '?', '[']);
```
AdamGS: can we use the `glob` crate to detect these pattern? this feel brittle

myrrc (changes requested):
Duckdb resolves symlinks in its filesystem module, so we do a double job here with canonicalization. Same for manual glob checks. The error was that we needed to pass Url an absolute path which I've fixed in https://github.com/vortex-data/vortex/pull/6851. We do want to merge tests from this PR though, increased code coverage is always a good thing.

Can you please keep the tests but replace canonicalization/globbing checks with the solution present upstream?

iceboundrock:
> Duckdb resolves symlinks in its filesystem module, so we do a double job here with canonicalization. Same for manual glob checks. The error was that we needed to pass Url an absolute path which I've fixed in #6851. We do want to merge tests from this PR though, increased code coverage is always a good thing.
> 
> Can you please keep the tests but replace canonicalization/globbing checks with the solution present upstream?

Done

vortex-duckdb/src/multi_file.rs
iceboundrock: I will use it in the future. For this PR, I will follow @myrrc 's comment to remove the implementation and just keep tests.

vortex-duckdb/src/multi_file.rs
```diff
@@ -62,3 +71,85 @@ impl DataSourceTableFunction for VortexMultiFileScan {
...
+        let url = parse_glob_url(filename)?;
+        assert_eq!(url.scheme(), "file");
+        // The relative name must have been resolved to an absolute path.
+        assert!(url.path().ends_with(filename));
+        assert!(url.path().starts_with('/'));
+        Ok(())
+    }
+
+    #[test]
+    fn test_parse_glob_url_relative_glob_path() -> VortexResult<()> {
```
myrrc, line 154: Thanks, I believe the PR is nearly ready.

Can you also add a test for relative paths like "../test.vortex", please?
I've tested and it fails, likely a flaw in my implementation.

```
$ duckdb -c 'SELECT * FROM "../vortex/vortex-bench/data/tpch/1.0/vortex-file-compressed/customer_0.vortex"'
Binder Error:
Object store error: Object at location /home/myrrc/duckdb-vortex/%2E%2E/vortex/vortex-bench/data/tpch/1.0/vortex-file-compressed/customer_0.vortex not found: No such file or directory (os error 2)
```

iceboundrock: ### Root cause

`std::path::absolute()` makes a path absolute but does not normalize `..` and/or `.` components. So `../test.vortex` from `/home/user/vortex/` becomes `/home/user/vortex/../test.vortex`, and `Url::from_file_path` then percent-encodes the dots as `%2E%2E`, causing DuckDB to look up a wrong/encoded path.

## Fix

Added a `normalize_path` helper function that walks the path components and resolves `.` and `..` without touching the filesystem. This is called after `absolute()` before creating the Url.

## Tests

* added: `test_parse_glob_url_parent_relative_path` — creates a temp file in `..`, constructs a `../filename` relative path, and asserts the resulting URL is absolute, has no `%2E` encoding, and ends with the correct filename.
* added `test_parse_glob_url_dot_normalization`, just test the `normalize_path` algorithm without creating any files.

vortex-duckdb/src/multi_file.rs
```diff
@@ -62,3 +89,130 @@ impl DataSourceTableFunction for VortexMultiFileScan {
...
+
+    // Use absolute paths so the expected result is cwd-independent.
+    #[rstest]
+    #[case("/a/./b", "/a/b")]
+    #[case("/a/b/./c", "/a/b/c")]
+    #[case("/a/../b", "/b")]
+    #[case("/a/b/../c", "/a/c")]
+    #[case("/a/b/../../c", "/c")]
+    #[case("/a/./b/.././c", "/a/c")]
+    #[case("/a/b/../..", "/")]
```
myrrc, line 204: What about `../../` and `/../` testcases? It will fail on current implementation since you try to pop from an empty PathBuf. Seems like we're reinventing canonicalization.

There are two distinct cases we should cover:
- remote path. Don't do any canonicalization/normalization whatsoever.
- Local path, either relative or absolute. We need to ensure we have an absolute path without .. or . as Url parsing will fail otherwise. However, we can't check whether the path exists on disk, because there may be globs, so fs::canonicalize/realpath won't work.

I think we can just take https://docs.rs/crate/normalize-path/latest/source/src/lib.rs#47 instead of normalize_path and use it on non-remote paths, and leave remote paths as is.

iceboundrock: * `popping from an empty PathBuf` simply [does nothing and returns `false`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.pop), without throwing an error. I’ll add more test cases.
* The term “remote path” refers to a UNC path on Windows, such as `//my.smb.server/foo/bar.vortex`? Are there any other use cases? I’ll first fix the UNC path.
* The `normalize_path` functions don’t perform any file system calls; they simply manipulate the `PathBuf`, it is same with `normalize-path::normalize` except it assumes the path is absolute . Since we call `absolute` before calling it. I can rename it to `normalize_absolute_path` if it is more clear.
* Currently don’t depend on the `normalize-path` crate, do we really need to import a library for just one function?

myrrc: I suggested using the function without copying, but yes, I"ve overlooked pop() (muscle memory from C++).
In this case we're fine.

myrrc (approved):

vortex-duckdb/src/multi_file.rs
myrrc: Let's fix CI issues and we're good to go.

iceboundrock: it seems all CI checks are passed or skipped, anything else you'd like me to change?

PR #6740, opened 2026-03-02
Upgrade to DataFusion 53

AdamGS:
Just a routine upgrade to DataFusion 53.

- Tracking issue for the release: https://github.com/apache/datafusion/issues/19692

This PR includes a perf regression that is fixed by https://github.com/vortex-data/vortex/pull/7176, keeping them as separate PRs so it doesn't get lost in all the noise here.

~Things left here~ All done!
- [x] Either wait for https://github.com/clflushopt/tpchgen-rs/pull/246 to merge and release, or change how we generate data in benchmarks (either by removing the dependency on `tpchgen-arrow` or using the CLI).
- [x] Go through the 53 release changelog and see if there's anything new we should take advantage of or adjust.
- [x] Finish the `object_store` upgrade

PR #6849, opened 2026-03-09
DataFusion 53 + morselized file scan

AdamGS:
Basically two changes smashed together:
- https://github.com/vortex-data/vortex/pull/6740
- https://github.com/apache/datafusion/pull/20481

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2084,9 +2084,8 @@ dependencies = [
 
 [[package]]
 name = "datafusion"
-version = "52.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "503f1f4a9060ae6e650d3dff5dc7a21266fea1302d890768d45b4b28586e830f"
+version = "52.1.0"
+source = "git+https://github.com/Dandandan/arrow-datafusion.git?rev=ed061e794b95dbf932bcc06f93ba9bc1252e1db5#ed061e794b95dbf932bcc06f93ba9bc1252e1db5"
 dependencies = [
  "arrow 58.0.0",
  "arrow-schema 58.0.0",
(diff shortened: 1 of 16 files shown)


alamb:
FWIW the morsels in https://github.com/apache/datafusion/pull/20481 are very much an IO scheduler for Parquet scans...

AdamGS:
@alamb I haven't read through the PR in a few days and it went through a lot of change since, this is mostly claude-driven because I was curious if there's any "free" perf here, I now see that the work stealing queue is enabled by `FileScanConfig`, and that's probably the piece I'm most interested in here.

AdamGS:
I guess if we can figure out some morsel size, there's also benefit in trying approach right?

alamb:
> I guess if we can figure out some morsel size, there's also benefit in trying approach right?

Absolutely -- I am also trying to sketch out a slightly different API (see details here https://github.com/apache/datafusion/issues/20529#issuecomment-4023988876) for morsels. The idea of using this in Vortex is a great usecase -- I'll ping you on the PR for your feedback

Dandandan:
> I guess if we can figure out some morsel size, there's also benefit in trying approach right?

Based on some iterative benchmarking on Parquet, a "projected compressed size" seems a relatively good proxy to avoid splitting really narrow / small (e.g. all primitive) row groups of a number of MB (e.g. ~2-16MB).

100K rows (which is commonly used) seems to create lot's of small scans / too much overhead of e.g. one-column scans on primitive fields

Dandandan:
Something that works quite good for Parquet on a couple of benchmarks it seems (on local file system):

* Coalesce small row groups until at minimum 1MB projected/compressed (some queries have only a few super small columns to scan) 
* Only introduce extra splitting once the queue depth is < #threads (e.g. to only parallelize once extra parallelism is necessary / reduce the latency at the tail and queries having only few row groups to scan)
* Cap the splitting to be minimal 1MB and max 2 * #threads (to avoid creating lot's of 1MB ones if the scans are large)

This is with splitting both IO and decode a row group level (i.e. it does a bunch of IO a number of times) but seems not to hurt those benchmarks really.

AdamGS:
defiantly some wins here now, probably can get some more if I tune it a bit more

AdamGS:
the regression in clickbench 23 is interesting, requires a deeper look

AdamGS:
clickbench q23 used to be 12x slower than baseline, now its 3x

PR #6851, opened 2026-03-09
    duckdb: read from local path, correct err on invalid path

### LINKED ISSUE #6841: Relative paths cause `Neither URL nor path error` because `Url::from_file_path` requires absolute paths

myrrc:
Fix two bugs for duckdb integration:
1. Allow relative path reads as in SELECT * FROM 'test.vortex'.
   Threw "neither url not a path" previously.
2. Show correct error when trying to write to a non-existent path.
   Threw "send error cannot send because receiver is gone".

### DIFF
--- a/vortex-duckdb/src/copy.rs
+++ b/vortex-duckdb/src/copy.rs
@@ -95,7 +95,7 @@ impl CopyFunction for VortexCopyFunction {
             .clone();
         RUNTIME
             .block_on(sink.send(chunk))
-            .map_err(|e| vortex_err!("send error {}", e.to_string()))?;
+            .map_err(|e| vortex_err!("send error {e}"))?;
         Ok(())
     }
 
@@ -130,13 +130,13 @@ impl CopyFunction for VortexCopyFunction {
         // SAFETY: The ClientContext is owned by the Connection and lives for the duration of
         // query execution. DuckDB keeps the connection alive while this copy function runs.
         let ctx = unsafe { client_context.erase_lifetime() };
-        let write_task = handle.spawn(async move {
-            // Use DuckDB FS exclusively to match the DuckDB client context configuration.
-            let writer = DuckDbFsWriter::new(ctx, &file_path).map_err(|e| {
-                vortex_err!("Failed to create DuckDB FS writer for {file_path}: {e}")
-            })?;
-            SESSION.write_options().write(writer, array_stream).await
-        });
+
+        // Use DuckDB FS exclusively to match the DuckDB client context configuration.
+        let writer = DuckDbFsWriter::new(ctx, &file_path)
+            .map_err(|e| vortex_err!("Failed to create DuckDB FS writer for {file_path}: {e}"))?;
+
+        let write_task =
+            handle.spawn(async move { SESSION.write_options().write(writer, array_stream).await });
 
         let worker_pool = RUNTIME.new_pool();
         worker_pool.set_workers_to_available_parallelism();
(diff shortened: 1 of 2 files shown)


vortex-duckdb/src/copy.rs
```diff
@@ -130,13 +130,13 @@ impl CopyFunction for VortexCopyFunction {
...
-        let write_task = handle.spawn(async move {
-            // Use DuckDB FS exclusively to match the DuckDB client context configuration.
-            let writer = DuckDbFsWriter::new(ctx, &file_path).map_err(|e| {
-                vortex_err!("Failed to create DuckDB FS writer for {file_path}: {e}")
-            })?;
-            SESSION.write_options().write(writer, array_stream).await
-        });
+
+        // Use DuckDB FS exclusively to match the DuckDB client context configuration.
+        let writer = DuckDbFsWriter::new(ctx, &file_path)
```
0ax1, line 135: Do we have a good grip on whether there's a write perf difference between the two? Is it weird to not use DDB FS by default for reading but for writing files?

0ax1 (commented):
2 questions

vortex-duckdb/src/copy.rs
myrrc: I didn't examine this one, the only difference I made is move the file open handle from the async task. What has happened before is that although we threw on non-existent file, duckdb still spawned the executor to write to it and failed again.

0ax1 (approved):

vortex-duckdb/src/copy.rs
myrrc: There is a distinction: we read and write files using duckdb if vortex_filesystem is updated, but we ultimately use duckdb API for opening and closing files.

PR #6853, opened 2026-03-09
WIP: Variant Type

AdamGS:
Just stashing results here to make the work more visible:
1. Variant DType and Array
2. ParquetVariant encoding
3. VariantGet expression


vortex-array/src/dtype/mod.rs
a10y, line 54: do you think we should make this `#[non_exhaustive]` ?

AdamGS: I think adding a new dtype should be very rare, but it is pretty painful to do right now.

AdamGS:
Impl moved away from this branch, split it into multiple things and moved it elsewhere.

PR #6857, opened 2026-03-09
move stuff around in `vortex-tensor`

connortsui20:
## Summary

Moves some stuff around in the `vortex-tensor` crate since we are about to move a lot more than just a fixed shape tensor into there.

## Testing

N/A


danking (approved):
looks right to me.

PR #6858, opened 2026-03-09
monoio / io-uring / clickbench

0ax1:


vortex-io/src/monoio_read_at.rs
```diff
@@ -0,0 +1,281 @@
...
+        let file_size = file.metadata()?.len();
+
+        let (request_tx, request_rx) = kanal::unbounded::<ReadRequest>();
+
+        // We need to pass the std File to the monoio thread where it will be converted
+        // into a monoio File. We use into_raw_fd + from_raw_fd to move ownership across
+        // the thread boundary safely.
+        let raw_fd = file.into_raw_fd();
+
+        thread::Builder::new()
```
gatesn, line 66: Do you know how much improvement is due to monoio, vs just having an extra thread running?

We currently have the problem that I/O only gets polled from DuckDB whenever it asks us for the next chunk. If it's busy further up the pipeline, e.g. doing aggregations, then we never drive the runtime.

0ax1: Might be the extra thread or monoio ~~but actually my guess here is, it's just bc of using o_direct~~. I'd need to dissect why some of the queries are actually faster. I guess o_direct shouldn't actually give any gains for warm runs. :shrug:

PR #6860, opened 2026-03-10
chore[array]: iterative execute method p1

joseph-isaacs:
Added a `require_child` macro to impl iterative execution for children:

 - alp_rd
 - dict


robert3005:
Any idea why there's an extra copy now? Looks like `PrimitiveArray::into_buffer_mut` needs to copy while previously it didn't show up

robert3005:
ah, the reason is that executed arrays are put back into the parent array while previously they were local to the function executing them

robert3005 (approved):

PR #6862, opened 2026-03-10
feat[gpu]: slice support for dyn dispatch

0ax1:
Slice offsets are either applied during the dyn dispatch build plan construction or in case of bitpacking within the dyn dispatch kernel, as sub-byte offsets can't be expressed by pointer arithmetic on the device ptr.

Calling `output_tile_len` as part of the final `execute_stage` loop has no perf overhead compared to `develop`.

### DIFF
--- a/vortex-cuda/kernels/src/dynamic_dispatch.cu
+++ b/vortex-cuda/kernels/src/dynamic_dispatch.cu
@@ -277,17 +277,15 @@ __device__ void execute_stage(const struct Stage &stage,
     __syncthreads();
 }
 
-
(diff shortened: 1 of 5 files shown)


robert3005 (approved):

a10y:
Why adding a new method vs fixing `cuda_device_ptr()`? We use cuda_device_ptr in a number of places where we probably mean `offset_ptr()`

a10y:
Oh I see, cuda_device_ptr becomes an alias for offset_ptr now

PR #6864, opened 2026-03-10
fix cuda_device_ptr() method

a10y:
## Summary

Previously we'd return the pointer to the front of the allocation.

We _should_ be returning the pointer to the beginning of the pointed range, e.g. for sliced allocations like what happens when you try and load the patches out of a CudaFlatLayout.

I encountered this after wasting a lot of time trying to unpack why `patches.indices.buffer_handle()` and `patches.values.buffer_handle()` were the same pointer.

## Testing

Added a simple unit test + existing test suite.


robert3005 (approved):

a10y:
Github is throwing another one of its tantrums it seems

robert3005:
I think #6862 might already fix it

PR #6866, opened 2026-03-10
Validate extension dtype in extension scalar validation

connortsui20:
## Summary

Tracking Issue: TODO

Right now, the `ExtVTable` has a leaky abstraction with respect to the `unpack_native` method. We thought that because the only important place we call it is within the `Scalar` logic, which by definition must have a valid `ExtDType` and `ScalarValue` combination, we technically do not need to validate the dtype inside `unpack_native`.

However, since the `ExtVTable` is public and the type that implements it is also public, there is nothing preventing someone from just calling this method directly. This means that it is very possible for someone to call this method without first calling `validate_dtype`.

So either we can make this method `unsafe` to make sure the caller knows what they are doing, or we can just say that implementors should call `validate_dtype`. This change proposes the latter.
 
## Testing

This change further validates correctness in exchange for (what is now a small amount of) performance

### DIFF
--- a/vortex-array/src/extension/datetime/timestamp.rs
+++ b/vortex-array/src/extension/datetime/timestamp.rs
@@ -184,9 +184,11 @@ impl ExtVTable for Timestamp {
     fn unpack_native<'a>(
         &self,
         metadata: &'a Self::Metadata,
-        _storage_dtype: &'a DType,
+        storage_dtype: &'a DType,
         storage_value: &'a ScalarValue,
     ) -> VortexResult<Self::NativeValue<'a>> {
+        self.validate_dtype(metadata, storage_dtype)?;
+
         let ts_value = storage_value.as_primitive().cast::<i64>()?;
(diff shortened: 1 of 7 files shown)


vortex-array/src/extension/datetime/timestamp.rs
```diff
@@ -184,9 +184,11 @@ impl ExtVTable for Timestamp {
     fn unpack_native<'a>(
         &self,
         metadata: &'a Self::Metadata,
-        _storage_dtype: &'a DType,
+        storage_dtype: &'a DType,
         storage_value: &'a ScalarValue,
     ) -> VortexResult<Self::NativeValue<'a>> {
+        self.validate_dtype(metadata, storage_dtype)?;
```
gatesn, line 190: Don't validate, just panic

gatesn (changes requested):
We should not require impls to _also_ apply precondition checks. Else there's no point in the vtable structure at all

PR #6867, opened 2026-03-10
feat[array]: change VTable execution to pass owned arrays

joseph-isaacs:
We want to be able to mutate buffers in-place this is done trying to unwrap Arcs, requiring owned Arc<Array> inputs.

This PR add owned execution.

```rust
fn execute(&self, array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<ExecutionResult>;
```

TODO: allow owned extraction of children from an owned array, this will be done using slots (https://github.com/vortex-data/vortex/pull/6870).


gatesn (approved):
This is the best we can do until we move over to vtables

joseph-isaacs:
Do you expect the execute interface to change?

fuzz/src/array/mod.rs
```diff
@@ -398,7 +398,7 @@ impl<'a> Arbitrary<'a> for FuzzArrayAction {
                         current_array
                             .to_canonical()
                             .vortex_expect("to_canonical should succeed in fuzz test"),
-                        &Mask::from_iter(mask.clone()),
+                        &!Mask::from_iter(mask.clone()),
```
robert3005, line 401: NOOOOOOOOO

PR #6868, opened 2026-03-10
[experimental] metal execution

a10y:
just parking this here. performance is not that great, not coming anywhere close to saturating the memory bw on M4 Max


a10y:
Fun experiment but don't think we want this for now

PR #6869, opened 2026-03-10
Extension Array Cleanup

connortsui20:
## Summary

Adds some missing constructors to `ExtensionArray` that we will want soon, as well as renames `storage` to `storage_array`.

## API Changes

The API changes are minimal, and nobody should really be using them anyways right now.

## Testing

N/A

### DIFF
--- a/vortex-array/src/arrays/datetime/mod.rs
+++ b/vortex-array/src/arrays/datetime/mod.rs
@@ -125,7 +125,7 @@ impl TemporalArray {
     /// These values are to be interpreted based on the time unit and optional time-zone stored
     /// in the TemporalMetadata.
     pub fn temporal_values(&self) -> &ArrayRef {
-        self.ext.storage()
+        self.ext.storage_array()
     }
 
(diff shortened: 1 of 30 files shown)


vortex-array/src/arrays/datetime/mod.rs
```diff
@@ -125,7 +125,7 @@ impl TemporalArray {
     /// These values are to be interpreted based on the time unit and optional time-zone stored
     /// in the TemporalMetadata.
     pub fn temporal_values(&self) -> &ArrayRef {
-        self.ext.storage()
+        self.ext.storage_array()
```
gatesn, line 128: I'd rather not just break the API without some reason other than naming

connortsui20: Since we now have storage dtypes, storage scalars, and storage arrays, we need to specify what it is (especially on the `ExtensionArray` where this could easily mean both the storage dtype or the storage array

gatesn (approved):

PR #6871, opened 2026-03-10
Reducing lock contention around expression partition caches

AdamGS:
## Summary

This PR includes two small changes:
1. Introduces the same read-only check around the partition cache that RowIdxLayoutReader has to StructReader.
2. Hold a `OnceLock` in the cache, so that the the write lock in the dashmap doesn't have to be held for a long time, which might cause a lot of unnecessary contention as it locks the whole shard. Ran into this issue while experimenting with #6849, where one layout reader might be shared between many small scans.


AdamGS:
statspopgen getting better in the same query for different engines seems promising

connortsui20 (approved):
looks very nice!

its a shame that the diff was hard to read...

Issue #6872, opened 2026-03-10
Tracking Issue: Extension Types


connortsui20:
This is a further continuation of https://github.com/vortex-data/vortex/issues/6618

# Extension Types

Now that things have more or less been figured out, we can actually start implementing extension arrays. This tracking issue will simply link together other tracking issues and PRs related to extension types.

See the Extension Types [RFC](https://vortex-data.github.io/rfcs/rfc/0005.html) for more details.

## (Potential) Future Extension Types

- `DateTimeParts` (`Primitive`)
- Vector (`FixedSizeList`)
- Matrix (`FixedSizeList`)
- Tensor (`FixedSizeList`)
- UUID (Do we need to add `FixedSizeBinary` as a canonical type?)
- JSON (`UTF8` and also maybe `Variant`?)
- PDX: https://arxiv.org/pdf/2503.04422v1 (`FixedSizeList`)
- Tags: https://github.com/vortex-data/vortex/discussions/5772#discussioncomment-15279892 (`ListView<Utf8>`)
- `Struct` but with protobuf-style field numbers (`Struct`)
- Probably lots more!

## Steps / History

- [ ] https://github.com/vortex-data/vortex/issues/6865
- [ ] https://github.com/vortex-data/vortex/issues/6854
- [ ] https://github.com/vortex-data/vortex/pull/7018

## Unresolved Questions

TODO

connortsui20:
Closing in favor of https://github.com/vortex-data/vortex/issues/7683

PR #6873, opened 2026-03-10
Validate Extension Arrays

connortsui20:
## Summary

Tracking Issue: https://github.com/vortex-data/vortex/issues/6872

Adds a `validate_array` and `array_validate` method to `ExtVTable` and `ExtDTypeRef`, respectively.

```rust
    /// Validates that the given storage array is compatible with this extension type and type
    /// medatada.
    ///
    /// Note that [`ExtVTable::validate_dtype()`] is always called first on the dtype of the storage
    /// array to validate the storage [`DType`].
    fn validate_array(
        &self,
        metadata: &Self::Metadata,
        storage_array: &dyn DynArray,
    ) -> VortexResult<()>;
```

This addition required adding implementations for all of the existing extension types (which is currently just the datetime types, UUID, and Tensor).

Also fixes some bugs in the implementation of the existing types, as well as some other tests that used to pass because we didn't validate anything before!

## API Changes

Adds some new methods, so this doesn't break anything.

## Testing

Adds a few tests via `ExtensionArray::try_new`.

### DIFF
--- a/vortex-array/src/arrays/extension/array.rs
+++ b/vortex-array/src/arrays/extension/array.rs
@@ -75,13 +75,7 @@ impl ExtensionArray {
     ///
     /// Returns an error if the storage array in not compatible with the extension dtype.
     pub fn try_new(ext_dtype: ExtDTypeRef, storage_array: ArrayRef) -> VortexResult<Self> {
-        // TODO(connor): Replace these statements once we add `validate_storage_array`.
-        // ext_dtype.validate_storage_array(&storage_array)?;
-        assert_eq!(
-            ext_dtype.storage_dtype(),
-            storage_array.dtype(),
-            "ExtensionArray: storage_dtype must match storage array DType",
-        );
+        ext_dtype.validate_storage_array(&storage_array)?;
 
         // SAFETY: we validate that the inputs are valid above.
         Ok(unsafe { Self::new_unchecked(ext_dtype, storage_array) })
@@ -95,15 +89,16 @@ impl ExtensionArray {
     /// other words, they must know that `ext_dtype.validate_storage_array(&storage_array)` has been
     /// called successfully on this storage array.
     pub unsafe fn new_unchecked(ext_dtype: ExtDTypeRef, storage_array: ArrayRef) -> Self {
-        // TODO(connor): Replace these statements once we add `validate_storage_array`.
-        // #[cfg(debug_assertions)]
-        // ext_dtype
-        //     .validate_storage_array(&storage_array)
-        //     .vortex_expect("[Debug Assertion]: Invalid storage array for `ExtensionArray`");
-        debug_assert_eq!(
+        #[cfg(debug_assertions)]
+        ext_dtype
+            .validate_storage_array(&storage_array)
+            .vortex_expect("[Debug Assertion]: Invalid storage array for `ExtensionArray`");
+
+        // The dtype check is much cheaper than the storage validation, so we might as well do this.
+        assert_eq!(
             ext_dtype.storage_dtype(),
             storage_array.dtype(),
-            "ExtensionArray: storage_dtype must match storage array DType",
+            "tried to construct an extension array with a storage array that has the wrong dtype"
         );
 
         Self {
--- a/vortex-array/src/arrays/extension/array.rs
+++ b/vortex-array/src/arrays/extension/array.rs
@@ -75,7 +75,11 @@ impl ExtensionArray {
     ///
     /// Returns an error if the storage array in not compatible with the extension dtype.
     pub fn try_new(ext_dtype: ExtDTypeRef, storage_array: ArrayRef) -> VortexResult<Self> {
-        ext_dtype.validate_storage_array(&storage_array)?;
+        if storage_array.is_host() {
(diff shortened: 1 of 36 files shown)


gatesn:
The codspeed regression is real imo. 

I think we need to avoid validation in almost all cases tbh....

gatesn (changes requested):
Let's discuss _when_ we should run validation. Scalars are already slow, so it makes sense. Arrays it's less obvious...

connortsui20:
there are also failures on cuda due to `as_slice` must be called on host buffer, also probably an issue.

@gatesn Note that we do have a `new_unchecked` function that will _not_ run any validation, but are you referring to a deeper problem?

vortex-array/src/arrays/extension/array.rs
```diff
@@ -95,16 +89,10 @@ impl ExtensionArray {
     /// other words, they must know that `ext_dtype.validate_storage_array(&storage_array)` has been
     /// called successfully on this storage array.
     pub unsafe fn new_unchecked(ext_dtype: ExtDTypeRef, storage_array: ArrayRef) -> Self {
```
gatesn, line 95: I like new_unchecked, but it should mean no value validation. We should still assert ext_dtype.storage_dtype() == storage_array.dtype()

gatesn (changes requested):
I think you need to go through all call sites of ExtensionArray::new and decide which should be unchecked, e.g. the compressor should be new_unchecked

connortsui20:
so I'm running into cuda test failures where `.as_slice()` must be called on the host buffer, does that mean I am not allowed to validate an array that might not be on the host? Or do I just need to copy it back to the host?

robert3005:
`as_slice()` is inherently a host method. If you have a cuda slice and you want to validate it you have to either write a cuda version of the validation or you have to copy or you skip. Our current bias was to skip validation on cuda slices

connortsui20:
@robert3005 so the error is triggered during stats collection (of min max), is there a way to tell stats to copy the buffer to host? Is that even something that makes sense?

robert3005:
BufferHandle has `copy_to_host/copy_to_host_async` methods. The current practice we follow is do validation if the buffer is on the host and do the validation for cuda buffer only in debug mode (you will have to copy it)

connortsui20:
So for me to do this, I have to figure out how to make this code:

```rust
        let stats = storage_array.statistics();
        if let Some(min) = stats.compute_min::<i64>() {
```

copy into a host buffer in order to compute the min stat. Do we need to fix that for the rest of the statistics, or is this just a bug? Shouldn't this compute function handle this internally?

robert3005:
@onursatici @0ax1 @joseph-isaacs whats the solution to Connor’s problem? I think we have some api to validate if whole array is on the host or not. We can use that to skip but not sure how to do optional validation. Long term we want to have compute function deal with it but exact behaviour of it is still undecided

gatesn:
I think we just document that validate will only be called for all-host arrays?

Or you say that validation implementations should allow for non-host buffers and skip validation

gatesn:
In theory, constructing an array from disk should be new_unchecked because we shouldnt perform validation while reading a file.

Similarly, any transformation / scalar function should assume correct behavior and not revalidate the array.

So this is only really on the write path, which is all in-memory anyway.

robert3005:
I think what you suggest is fine, ie only validate all host arrays. We were adding debug assertions so you can run a slow gpu mode that still validate all the arrays for debugging

connortsui20:
Note that this issue is not limited to reading arrays from disk, for example if we decide to run an expression on an extension array (where really we run it on the storage array), that change might have invalidated the storage array. So I feel like this issue is inherent to stats computation on GPU buffers, not when and where we validate arrays.

gatesn:
I don't really understand this? I'm saying validation _shouldn't_ be run on read

And stats are _all_ considered to be optional at the moment

connortsui20:
so basically this is a stats bug, not a validation issue (stats computation should not panic). Just so we can move forward since this is blocking some other things, should I just disable those specific tests?

Also, something that I just learned is that if stats computation ERRORs, it flattens and gives back `None`, which is bad since `None` can also mean the array is full of nulls or is empty. That's not good!

robert3005:
You have `is_host` on Array that will validate that all buffers are on the host. For now just run validation on extension array only if that function returns true. I think @gatesn had some bigger changes in mind.

I think computation error being None is consistent... Not sure what the caller would do if you gave them something else.

connortsui20:
> I think computation error being None is consistent... Not sure what the caller would do if you gave them something else.

Specifically for min and max, None has a double meaning:

- Something went wrong (compute error)
- There are no non-null values in the array so we can't compute the min or max (empty or all null)

I have no way to distinguish between these 2 things (without inspecting the array directly), I feel that this is concerning

gatesn:
This PR still feels weird to me. I think it boils down to the fact that validation can do compute. Therefore it should take an ExecutionCtx.... therefore ExtensionArray::try_new takes an ExecutionCtx? Surely not.

Maybe there is _only_ a new_unchecked constructor? The ExtensionArrayBuilder can validate. And other call sites must provide their own guarantees.

PR #6874, opened 2026-03-10
Remove ExtDTypeInner

gatesn:


robert3005:
What's the justification?

gatesn:
The vtable docs says so 😀

Real reason, is that this allows both the type erased and typed versions to defef to the same &Foo<V>, which make a lot of APIs a lot simpler.

The cost is that &Foo<V> is more expensive to clone since there is no arc. However any function that could require returning "self" should be implemented over the type erased version FooRef and the typed vtable should return Option<T>. Therefore the clone is always performed over the erased version which contains an Arc.

robert3005 (approved):

PR #6877, opened 2026-03-11
Fix fuzzer incorrect mask application

myrrc:
#6515

- Apply validity & mask in fuzzer target generation.
  Before, validity & !mask was generated.
- Add tracing logging to fuzzer
- Add debug output for fuzzer mask, info output for mistmatches,
  initial array, and fuzz actions

### DIFF
--- a/fuzz/src/array/mask.rs
+++ b/fuzz/src/array/mask.rs
@@ -23,28 +23,31 @@ use vortex_error::VortexResult;
 use vortex_mask::AllOr;
 use vortex_mask::Mask;
 
-/// Set to false any entries for which the mask is true.
-///
+/// Apply a logical AND of a validity and a mask.
+/// This needs to be coherent with applications of Mask.
 /// The result is always nullable. The result has the same length as self.
 #[inline]
 pub fn mask_validity(validity: &Validity, mask: &Mask) -> Validity {
-    match mask.bit_buffer() {
-        AllOr::All => Validity::AllInvalid,
-        AllOr::None => validity.clone().into_nullable(),
-        AllOr::Some(make_invalid) => match validity {
-            Validity::NonNullable | Validity::AllValid => {
-                Validity::from_bit_buffer(!make_invalid, Nullability::Nullable)
-            }
+    let out = match mask.bit_buffer() {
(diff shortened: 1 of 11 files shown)


vortex-buffer/src/bit/buf.rs
```diff
@@ -35,6 +35,19 @@ pub struct BitBuffer {
     len: usize,
 }
 
+impl std::fmt::Debug for BitBuffer {
```
robert3005: import this

vortex-buffer/src/bit/buf.rs
```diff
@@ -35,6 +35,19 @@ pub struct BitBuffer {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut buf: Vec<bool> = Vec::new();
+        for i in 0..self.len {
+            buf.push(self.value(i));
+        }
```
joseph-isaacs: surely there is a better way

fuzz/src/array/mask.rs
```diff
@@ -23,28 +23,31 @@ use vortex_error::VortexResult;
...
+                    is_valid.to_bit_buffer() & make_valid,
                     Nullability::Nullable,
                 )
             }
         },
-    }
+    };
+
+    tracing::debug!(validity = ?validity, mask = ?mask, out = ?out, "generated fuzzer mask");
+    out
```
joseph-isaacs, line 50: can we add a test?

vortex-buffer/src/bit/buf.rs
```diff
@@ -35,6 +35,19 @@ pub struct BitBuffer {
...
+        f.debug_struct("BitBuffer")
```
robert3005, line 46: you want offset as well

myrrc: self.value() includes offset already

robert3005: There's slight conflict here where if your buffer is wrong you will just get nonsensical values. There's imho a difference between debug and display and for debug you would want to ideally see raw values. But I agree that it's useful to see the logical values

robert3005 (approved):

PR #6880, opened 2026-03-11
Use unaligned bit buffer iterators where possible

### LINKED ISSUE #6836: Use unaligned bitwise iterators for unary and for binary operations where possible
At some point before we deleted our own iterators we had this optimisation, we lost it again where we picked up arrow boolean buffer iterators. Now those optimisations have hit arrow we should leverage them in our implementation

robert3005:
Use unalingned bit iterators where possible, they are faster as they avoid
unaligned reads. They're always fine to use for unary operations and can be used
for n-ary operations if all buffers are aligned

## Summary

We used to have this optimisation when we added BitBuffer but we reverted to
arrow implementation of bit iterators and it got lost in the process.

Closes: #6836

## Testing
Existing tests

### DIFF
--- a/vortex-buffer/src/bit/buf.rs
+++ b/vortex-buffer/src/bit/buf.rs
@@ -101,11 +101,7 @@ impl BitBuffer {
         // Slice the buffer to ensure the offset is within the first byte
         let byte_offset = offset / 8;
         let offset = offset % 8;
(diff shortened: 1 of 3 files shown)


AdamGS (approved):

robert3005:
I have to revert the last commit before merging

PR #6888, opened 2026-03-11
More detailed fuzzer report template

connortsui20:
## Summary

The fuzzer template is confusing, even though I have worked with it a solid number of times at this point I always forget how to set it up correctly.

Hopefully with these instructions it will lower the amount of time spent remembering what to do.

## Testing

N/A

### DIFF
--- a/.github/scripts/fuzz_report/templates/new_issue.md
+++ b/.github/scripts/fuzz_report/templates/new_issue.md
@@ -35,6 +35,8 @@
 
 ### Reproduction
 
+<details>
+
 1. Download the crash artifact:
    - **Direct download**: {{ARTIFACT_URL}}
    - Extract the zip file (`unzip`)
@@ -58,10 +60,14 @@ RUST_BACKTRACE=1 cargo +nightly fuzz run -D --sanitizer=none {{FUZZ_TARGET}} /pa
 RUST_BACKTRACE=full cargo +nightly fuzz run -D --sanitizer=none {{FUZZ_TARGET}} /path/to/crash_file -- -rss_limit_mb=0
 ```
 
+</details>
+
 ### Workflow Example
 
 Assuming you download the zipfile to `~/Downloads`, and your working directory is the repository
 root, you can follow these steps:
+
+<details>
 
 ```bash
 # Create the artifacts directory if you haven't already.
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

joseph-isaacs (changes requested):
Can we make this a link to follow?

joseph-isaacs:
I really want to avoid a noisy issue, A first time issue manual would be better?

connortsui20:
I don't see how this is a noisy issue, the only thing that is technically noisy is the commands, and the whole point of this is that I do not want to have to manually construct these super long commands. I do not want to have to think about this when there's something that is broken.

joseph-isaacs:
https://github.com/vortex-data/vortex/actions/runs/22959766498

But we might need to close an issue so it creates a new one

.github/scripts/fuzz_report/templates/new_issue.md
```diff
@@ -35,19 +37,64 @@
...
 2. Reproduce locally:
+
+```bash
+cargo +nightly fuzz run -D --sanitizer=none {{FUZZ_TARGET}} /path/to/crash_file -- -rss_limit_mb=0
+```
+
+3. Get a backtrace:
+
 ```bash
-cargo +nightly fuzz run -D --sanitizer=none {{FUZZ_TARGET}} {{FUZZ_TARGET}}/{{CRASH_FILE}} -- -rss_limit_mb=0
```
myrrc, line 42: I like how it was before, you could just copy paste the command and it worked, now you need to manually fill the path

connortsui20: That is the point of the section below, the first part explains how it works, the workflow example lets you copy and paste everything

joseph-isaacs:
see: https://github.com/vortex-data/vortex/issues/6896 maybe we want hide in a section the longer info?

connortsui20:
I really don't see the point in trying to minimize these kinds of issues. If we have run into a problem and I am tasked to investigate, I arguably want **as much information as possible** so I can solve it quickly.

I can add a details section drop-down if that works?

joseph-isaacs:
that is what I meant

joseph-isaacs (approved):

PR #6898, opened 2026-03-11
ci: use pre-built AMI for arm64 CI and skip redundant setup

joseph-isaacs:
## Summary

Use the custom `ubuntu24-full-arm64-pre` AMI image for the arm64 test runner, which has the Rust toolchain and nextest pre-installed.

Changes:
- Add `image=ubuntu24-full-arm64-pre` to the arm64 runner config in CI
- Skip `setup-rust` and `nextest` install steps for arm64 (already in the AMI)
- Remove `fallback_runner` for arm64 (not needed since forks don't use runs-on runners)

## Testing

CI will validate the arm64 tests run correctly with the pre-built AMI.


0ax1 (approved):
🚤

PR #6901, opened 2026-03-11
Trying some more benchmarking noise reduction things

AdamGS:
Pins workloads to specific cores, disables more things, turns off more features.

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -93,7 +93,7 @@ jobs:
         env:
           RUST_BACKTRACE: full
         run: |
-          target/release_debug/${{ matrix.benchmark.id }} -d gh-json -o results.json
+          bash scripts/bench-taskset.sh target/release_debug/${{ matrix.benchmark.id }} -d gh-json -o results.json
 
       - name: Setup AWS CLI
         if: github.event.pull_request.head.repo.fork == false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,9 +110,12 @@ jobs:
           uv run --all-packages pytest --benchmark-disable -n auto test/
         working-directory: vortex-python/
 
+      - name: Setup benchmark environment
+        run: sudo bash scripts/setup-benchmark.sh
+
       - name: Pytest Benchmarks - Vortex
         run: |
-          uv run --all-packages pytest --benchmark-only benchmark/
+          bash ../scripts/bench-taskset.sh uv run --all-packages pytest --benchmark-only benchmark/
         working-directory: vortex-python/
 
       - name: Doctest - PyVortex
@@ -743,7 +746,7 @@ jobs:
       - name: Run benchmarks
         uses: CodSpeedHQ/action@281164b0f014a4e7badd2c02cecad9b595b70537
         with:
-          run: cargo codspeed run
+          run: bash scripts/bench-taskset.sh cargo codspeed run
           token: ${{ secrets.CODSPEED_TOKEN }}
           mode: "simulation"
 
--- a/scripts/bench-taskset.sh
+++ b/scripts/bench-taskset.sh
@@ -15,13 +15,11 @@ if [[ -f /tmp/vortex-benchmark.env ]]; then
     source /tmp/vortex-benchmark.env
 fi
 
+
+
 if [[ -z "${BENCH_CPUS:-}" ]]; then
     cpu_count="$(nproc)"
-    if (( cpu_count <= 2 )); then
-        BENCH_CPUS="0-$((cpu_count - 1))"
-    else
-        BENCH_CPUS="2-$((cpu_count - 1))"
-    fi
+    BENCH_CPUS="2-$((cpu_count - 1))"
 fi
 
 if command -v numactl >/dev/null 2>&1; then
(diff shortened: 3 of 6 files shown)


.github/workflows/bench-pr.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         env:
           RUST_BACKTRACE: full
         run: |
-          target/release_debug/${{ matrix.benchmark.id }} -d gh-json -o results.json
+          bash scripts/bench-taskset.sh target/release_debug/${{ matrix.benchmark.id }} -d gh-json -o results.json
```
0ax1, line 96: Unrelated to this PR, should we consider using the cargo runner env, like we do for CUDA sanitizers?
```
- name: CUDA - initcheck
  env:
    CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER: compute-sanitizer --tool initcheck --error-exitcode 1
  run: cargo test --locked -p vortex-cuda --all-features --target x86_64-unknown-linux-gnu
```

.github/workflows/ci.yml
```diff
@@ -110,9 +110,12 @@ jobs:
           uv run --all-packages pytest --benchmark-disable -n auto test/
         working-directory: vortex-python/
 
+      - name: Setup benchmark environment
+        run: sudo bash scripts/setup-benchmark.sh
+
       - name: Pytest Benchmarks - Vortex
         run: |
-          uv run --all-packages pytest --benchmark-only benchmark/
+          bash ../scripts/bench-taskset.sh uv run --all-packages pytest --benchmark-only benchmark/
```
0ax1, line 118: Generally not loving relative paths in scripts but 🤷

scripts/bench-taskset.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [[ -f /tmp/vortex-benchmark.env ]]; then
+    # shellcheck disable=SC1091
+    source /tmp/vortex-benchmark.env
+fi
+
+if [[ -z "${BENCH_CPUS:-}" ]]; then
+    cpu_count="$(nproc)"
+    if (( cpu_count <= 2 )); then
+        BENCH_CPUS="0-$((cpu_count - 1))"
```
0ax1: Should we even bother making this 2 cases. I'd think its safe to assume we will always run on CI machines with more than 2 logical cores.

scripts/bench-taskset.sh
```diff
@@ -0,0 +1,31 @@
...
+    else
+        BENCH_CPUS="2-$((cpu_count - 1))"
+    fi
+fi
+
+if command -v numactl >/dev/null 2>&1; then
+    exec numactl --physcpubind="$BENCH_CPUS" --localalloc "$@"
```
0ax1, line 26: Can you elaborate / add a comment why we replace the process here and for `taskset`? Is that the common pattern to use them?

scripts/setup-benchmark.sh
```diff
@@ -15,12 +15,23 @@ for gov in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do
     echo performance > "$gov" 2>/dev/null || true
 done
 
+[[ -w /sys/devices/system/cpu/intel_pstate/no_turbo ]] && echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo || true
```
0ax1, line 14: Let's add comments why we do this.

scripts/setup-benchmark.sh
```diff
@@ -15,12 +15,23 @@ for gov in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do
...
+[[ -w /sys/devices/system/cpu/cpufreq/boost ]] && echo 0 > /sys/devices/system/cpu/cpufreq/boost || true
+
 # Really discourage swapping to disk
 sysctl vm.swappiness=0
+swapoff -a || true
```
0ax1, line 19: We ignore a lot of errors in this script via `|| true`. In some way I'd like us to only run things that actually apply and otherwhise fail. We can comment out commands that don't apply to the current runner.

scripts/setup-benchmark.sh
```diff
@@ -33,3 +44,42 @@ systemctl mask ModemManager
...
+  snapd
+do
+  systemctl disable --now "$unit" 2>/dev/null || true
+done
+systemctl mask irqbalance 2>/dev/null || true
+
+CPU_COUNT="$(nproc)"
+if (( CPU_COUNT <= 2 )); then
+  HOUSEKEEPING_CPUS="0"
+  BENCH_CPUS="0-$((CPU_COUNT - 1))"
```
0ax1: Same here, I'd think we'll always have more cores.

scripts/setup-benchmark.sh
```diff
@@ -33,3 +44,42 @@ systemctl mask ModemManager
...
+else
+  HOUSEKEEPING_CPUS="0-1"
+  BENCH_CPUS="2-$((CPU_COUNT - 1))"
+fi
+
+# Pin all IRQs to housekeeping CPUs
+for f in /proc/irq/[0-9]*/smp_affinity_list; do
+  [[ -w "$f" ]] && echo "$HOUSEKEEPING_CPUS" > "$f" || true
```
0ax1: Same here I'd prefer to fail loudly. Is that tedious work with over time, wdyt?

0ax1 (commented):
Couple of thoughts

0ax1 (approved):

PR #6907, opened 2026-03-11
Rename Array VTables

gatesn:
We are about to move more functions onto the VTable struct. To make this API nicer for users we're renaming FooVTable to Foo. Since users don't really interact with these structs at the moment, it should have a relatively small blast radius.

That said, anyone who has code like this should update it

```diff
- if let Some(foo) = array.as_opt::<FooVTable>() { ... }
+ if let Some(foo) = array.as_opt::<Foo>() { ... }
```

### DIFF
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -233,6 +233,8 @@ typedef struct DType DType;
  */
 typedef struct Nullability Nullability;
 
+typedef struct Primitive Primitive;
+
 /**
  * Base type for all Vortex arrays.
  *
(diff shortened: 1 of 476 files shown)


vortex-ffi/cinclude/vortex.h
```diff
@@ -233,6 +233,8 @@ typedef struct DType DType;
  */
 typedef struct Nullability Nullability;
 
+typedef struct Primitive Primitive;
```
joseph-isaacs, line 236: ?

joseph-isaacs:
Look like a few other commits got it

vortex-ffi/cinclude/vortex.h
gatesn: I poked around for like 2 hours last night and honestly I have no idea... but didn't feel problematic enough to care about.

robert3005 (approved):

Issue #5152, opened 2025-11-03
`sum` statistic should not include `NaN`s


joseph-isaacs:
This can be derived from the NaN count.

lwwmanning:
I think this is clearly fine for the statistic, since we do the same for Min/Max. It's less clear that it's valid for the compute function, which should arguably be IEEE-compliant (and can short-circuit to NaN if nan_count stat > 0)

PR #6910, opened 2026-03-12
Cut over sum aggregate function

gatesn:
Move sum compute function over to the Sum impl of AggregatFnVTable

Note that we remove the ability to sum extension types since we were incorrectly just summing the storage DType. This doesn't make sense, e.g. summing timestamps is invalid, but summing time deltas is allowed.


robert3005:
We should make sure we port any exhisting behaviour we have in the current implementations

gatesn:
Like what?

joseph-isaacs:
NaN handlings at least. Its work fuzzing this on the side is possible

robert3005:
I left a comment on the pr that added sum with the code location since that code isn’t touched in this pr. If we want to change decimal sum to always work in i256 we should make this in a separate pr

robert3005:
FWIW NaNs are a long standing issue https://github.com/vortex-data/vortex/issues/5152. I think fixing NaN handling is what's actually required for this stat to be universally useful but it didn't change in this pr

gatesn:
Added bonus, benchmarks are 10-15% faster!

vortex-array/src/aggregate_fn/fns/sum/mod.rs
```diff
@@ -0,0 +1,785 @@
...
+    type Partial = SumPartial;
+
+    fn id(&self) -> AggregateFnId {
+        AggregateFnId::new_ref("vortex.sum")
+    }
+
+    fn return_dtype(&self, _options: &Self::Options, input_dtype: &DType) -> Option<DType> {
+        // When a sum overflows, we return a sum _value_ of null. Therefore, we all return dtypes
+        // are nullable.
+        use Nullability::Nullable;
```
joseph-isaacs, line 80: ?

vortex-array/src/aggregate_fn/fns/sum/mod.rs
```diff
@@ -0,0 +1,785 @@
...
+            .ok_or_else(|| vortex_err!("Unsupported sum dtype: {}", input_dtype))?;
+        let initial = make_zero_state(&return_dtype);
+
+        Ok(SumPartial {
+            return_dtype,
+            current: Some(initial),
+        })
+    }
+
+    fn combine_partials(&self, partial: &mut Self::Partial, other: Scalar) -> VortexResult<()> {
```
joseph-isaacs, line 131: this code is the same as constant array?

vortex-array/src/aggregate_fn/fns/sum/constant.rs
```diff
@@ -0,0 +1,224 @@
...
+use super::checked_add_i64;
+use super::checked_add_u64;
+use crate::arrays::ConstantArray;
+use crate::dtype::DType;
+use crate::scalar::DecimalValue;
+
+/// Accumulate a constant array into the sum state.
+/// Computes `scalar * len` and adds to the accumulator.
+/// Returns Ok(true) if saturated (overflow), Ok(false) if not.
+pub(super) fn accumulate_constant(inner: &mut SumState, c: &ConstantArray) -> VortexResult<bool> {
```
joseph-isaacs: extract the scalar and use combine partiaL?

vortex-array/src/aggregate_fn/fns/sum/primitive.rs
```diff
@@ -0,0 +1,240 @@
...
+        ),
+        SumState::Float(acc) => match_each_native_ptype!(p.ptype(),
+            unsigned: |_T| { vortex_panic!("float sum state with unsigned input") },
+            signed: |_T| { vortex_panic!("float sum state with signed input") },
+            floating: |T| {
+                for (&v, valid) in p.as_slice::<T>().iter().zip_eq(validity.iter()) {
+                    if valid {
+                        *acc += ToPrimitive::to_f64(&v).vortex_expect("float to f64");
+                    }
+                }
```
joseph-isaacs, line 104: would be nice to handle the NaN thing now

vortex-array/src/aggregate_fn/fns/sum/mod.rs
gatesn: What?

vortex-array/src/aggregate_fn/fns/sum/primitive.rs
gatesn: I think any behavior change should be isolated

robert3005 (approved):

vortex-array/src/aggregate_fn/fns/sum/mod.rs
joseph-isaacs: Why is this here?

PR #6911, opened 2026-03-12
ignore[ci]: remove compact vortex from clickbench (for speed)

joseph-isaacs:
## Summary

Closes: #000

### DIFF
--- a/.github/workflows/sql-benchmarks.yml
+++ b/.github/workflows/sql-benchmarks.yml
@@ -21,7 +21,7 @@ on:
               "id": "clickbench-nvme",
               "subcommand": "clickbench",
               "name": "Clickbench on NVME",
-              "targets": "datafusion:parquet,datafusion:vortex,datafusion:vortex-compact,duckdb:parquet,duckdb:vortex,duckdb:vortex-compact,duckdb:duckdb"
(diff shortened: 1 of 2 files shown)


.github/workflows/sql-benchmarks.yml
```diff
@@ -21,7 +21,7 @@ on:
               "id": "clickbench-nvme",
               "subcommand": "clickbench",
               "name": "Clickbench on NVME",
-              "targets": "datafusion:parquet,datafusion:vortex,datafusion:vortex-compact,duckdb:parquet,duckdb:vortex,duckdb:vortex-compact,duckdb:duckdb"
+              "targets": "datafusion:parquet,datafusion:vortex,duckdb:parquet,duckdb:vortex,duckdb:duckdb"
```
AdamGS, line 24: here you can keep it right? it'll just run on merges

AdamGS (approved):
LGTM, one question so I disabled auto merge

.github/workflows/sql-benchmarks.yml
robert3005: I would kill the benchmarks wholesale, no need to spend time running them on develop as well

joseph-isaacs: Compact will only run nightly

PR #6912, opened 2026-03-12
Variant DType

AdamGS:
This PR includes initial support for the Variant DType, as described in the [Variant RFC](https://github.com/vortex-data/rfcs/blob/develop/accepted/0015-variant-type.md). It includes most of the required boilerplate and initial structure for this new dtype.

It includes the following changes:
1. New dtype
2. serialization for the dtype
3. Scalar variant and new scalar value
4. A lot of code paths that aren't supported yet and error accordingly.


fuzz/src/array/mod.rs
```diff
@@ -488,6 +488,8 @@ fn actions_for_dtype(dtype: &DType) -> HashSet<ActionType> {
             // Extension types delegate to storage dtype, support most operations
             ActionType::iter().collect()
         }
+        // Currently, no support at all
+        DType::Variant => Default::default(),
```
joseph-isaacs: should be unreachable?

vortex-proto/proto/scalar.proto
joseph-isaacs, line 1: some docs for this or a ptr would be nice

vortex-array/src/arrays/constant/vtable/canonical.rs
```diff
@@ -164,6 +165,9 @@ pub(crate) fn constant_canonicalize(array: &ConstantArray) -> VortexResult<Canon
             let storage_self = ConstantArray::new(storage_scalar, array.len()).into_array();
             Canonical::Extension(ExtensionArray::new(ext_dtype.clone(), storage_self))
         }
+        DType::Variant => {
+            vortex_bail!("Variant don't currently support canonicalization");
```
connortsui20: Are we defining the semantics of this now? If not I think this should be left as `unimplemented` with a clear TODO marked so we can easily search for this.

vortex-array/src/dtype/serde/flatbuffers.rs
```diff
@@ -231,9 +231,11 @@ impl TryFrom<ViewedDType> for DType {
...
             }
-            // This is here to fail to compile if another variant is included.
-            #[allow(clippy::wildcard_in_or_patterns)]
-            fb::Type(11) => Err(vortex_err!("Unknown DType variant")),
+            fb::Type::Variant => {
+                fb.type__as_variant()
+                    .ok_or_else(|| vortex_err!("failed to parse variant from flatbuffer"))?;
+                Ok(Self::Variant)
+            }
             _ => Err(vortex_err!("Unknown DType variant")),
```
connortsui20, line 240: This should probably keep the type guard

vortex-array/src/dtype/dtype_impl.rs
```diff
@@ -54,7 +54,7 @@ impl DType {
     #[inline]
     pub fn is_nullable(&self) -> bool {
         match self {
-            Null => true,
+            Null | Variant => true,
```
connortsui20: I missed the nullability section in the RFC, do we really want to do this? Based on a tiny bit of research on other systems with variant it seems like they do support NOT NULL variant.

AFAICT this was not really discussed in the original RFC PR, and I feel like this is something worth reconsidering. Should I open a new PR on the RFC page?

vortex-array/src/scalar/display.rs
```diff
@@ -21,6 +21,10 @@ impl Display for Scalar {
             DType::Struct(..) => write!(f, "{}", self.as_struct()),
             DType::List(..) | DType::FixedSizeList(..) => write!(f, "{}", self.as_list()),
             DType::Extension(_) => write!(f, "{}", self.as_extension()),
+            DType::Variant => match self.value() {
+                None => write!(f, "null"),
+                Some(value) => write!(f, "{value}"),
+            },
```
connortsui20: This should be `self.as_variant()` (or if we want to do that in a future PR then it should just be unimplemented)

vortex-array/src/dtype/dtype_impl.rs
AdamGS: I think this is a question of how you define nullability, in my mind its "can I get a null value out of this array?", which is basically always true for any variant type I'm familiar with, whether its an explicit `Variant::null`, top-level null or a path that doesn't exist in a specific row

vortex-array/src/scalar/downcast.rs
```diff
@@ -153,6 +154,22 @@ impl Scalar {
...
+    }
+
+    /// Returns the semantic variant value if the scalar has `DType::Variant` and is non-null.
+    pub fn as_variant_opt(&self) -> Option<&VariantValue> {
+        self.dtype()
+            .is_variant()
+            .then(|| self.value())
+            .flatten()
+            .map(ScalarValue::as_variant)
+    }
```
connortsui20: I think we might as well add the typed view of a variant scalar now? I believe it would just need to hold a scalar value. If it's not in the scope of this PR then I think it would be better to leave this out and keep it as `unimplemented`

vortex-array/src/dtype/serde/flatbuffers.rs
AdamGS: What does it do? It didn't cause any build issues to me

vortex-array/src/scalar/proto.rs
```diff
@@ -238,6 +245,16 @@ impl ScalarValue {
             .as_ref()
             .ok_or_else(|| vortex_err!(Serde: "Scalar value missing kind"))?;
 
+        if matches!(dtype, DType::Variant) {
```
connortsui20: Nit: I think `if dtype == DType::Variant {` is more readable

vortex-array/src/scalar/proto.rs
```diff
@@ -448,6 +468,177 @@ fn list_from_proto(
     Ok(ScalarValue::List(values))
 }
 
+fn variant_to_proto(value: &VariantValue) -> pb::VariantValue {
```
connortsui20: could you move these function up near the other into proto functions? Either that or maybe proto deserves its own directory of modules at this point.

vortex-array/src/scalar/scalar_value.rs
```diff
@@ -64,6 +67,7 @@ impl ScalarValue {
                 // zero storage value and try to make an extension scalar from that.
                 Self::zero_value(ext_dtype.storage_dtype())
             }
+            DType::Variant => Self::Variant(VariantValue::Null),
```
connortsui20: I'm worried that this might break our abstractions (maybe its fine though), see below

vortex-array/src/scalar/variant_value.rs
```diff
@@ -0,0 +1,74 @@
...
+pub enum VariantValue {
+    Null,
+    Bool(bool),
+    Primitive(PValue),
+    Decimal(DecimalValue),
+    Utf8(BufferString),
+    Binary(ByteBuffer),
+    List(Vec<VariantValue>),
+    Object(Vec<(BufferString, VariantValue)>),
+}
```
connortsui20: I'm worried that this will break our abstraction. This is eerily similar to what `ScalarValue` (and the old `InnerScalarValue`) looks like. We also very recently made a change to extract out nulls from scalar value such that a nullable scalar has an `Option<ScalarValue>` (so `ScalarValue` is always non-nullable).

Is it fine that this is basically scalar value with a few extra variants (heh)? Couldn't we just extend the existing scalar value enum to support object and then everything falls out of that? Is there something here that requires we store these exact parameters?

connortsui20 (dismissed):
Nice first steps! I am a bit worried about the format for variant scalars, and I think I would like to see more TODOs marked in places that we know will change, but other than that just a few nits.

vortex-array/src/dtype/serde/flatbuffers.rs
connortsui20: it just means that if someone else adds a new type here it will cause a compile error instead of falling through (this was helpful when I was new to the codebase)

vortex-array/src/dtype/dtype_impl.rs
connortsui20: see below for my comment on the null inside the variant scalar value, I guess my question is if a variant null is semantically different from nulls currently in Vortex? In my mind it does not seem any different so this decision is a bit confusing to me.

vortex-array/src/scalar/variant_value.rs
AdamGS: But there's no other Object type in our scalar system. This is also about differentiating between the various ways variants can be nulls, you can have a JSON that looks like `{"value": null}`, which is different than an array with a nullability mask that marks that one row as null

vortex-array/src/scalar/proto.rs
gatesn: I thought variants were arbitrary nested vortex types? So isn't the pb::ScalarValue::Variant just a pb::Scalar?

AdamGS: not exactly, they also have an "object" variant and list where every item can be a variant

vortex-proto/proto/scalar.proto
```diff
@@ -28,9 +28,63 @@ message ScalarValue {
...
     uint64 f16_value = 10;
+    VariantValue variant_value = 11;
   }
 }
 
 message ListValue {
   repeated ScalarValue values = 1;
 }
+
+message VariantValue {
```
gatesn: Again, this == Scalar?

gatesn (changes requested):
I think we can simplify scalar values by just storing scalars

vortex-proto/proto/scalar.proto
AdamGS: https://github.com/vortex-data/vortex/pull/6912#discussion_r2925658678

vortex-array/src/scalar/proto.rs
connortsui20: We can just add an "object" variant to the `ScalarValue` enum though? It already has a `List` variant that we use for `Struct` scalars (we also use them for `List` scalars but we would like to add an `Array` variant too, see https://github.com/vortex-data/vortex/pull/6717).

fuzz/src/array/mod.rs
AdamGS: not sure what the contract here is, but sure.

vortex-array/src/scalar/proto.rs
AdamGS: (lets keep this conversation in this thread) We can, but there's also the issue of differentiating between the various nulls. I don't know if its a thing in every possible variant encoding, but it is in `parquet-variant`, which is the main reference point we have.

connortsui20: sorry is this a nullability concern or adding a variant that doesnt match the model concern?

```rust
    /// A list of potentially null scalar values. (This is what we have now)

    /// We would add this as a variant (or maybe as a HashMap):
    Object(Vec<(BufferString, Option<ScalarValue>)>),
```

If we say that variant is any scalar value then I don't see why this is an issue? The null just gets outlined.

AdamGS: Its both? Sure we can extend ScalarValue, I have no problem with that, but Variant (or at least - ParquetVariant) has different nullability behavior that isn't just outlining the null with an `Option`, see [this](https://github.com/apache/parquet-format/blob/master/VariantShredding.md#value-shredding) section from the spec, specifically the table regarding null representation (unfortunately can't quote it here), but the key paragraph is:
> If a Variant is missing in a context where a value is required, readers must return a Variant null (00): basic type 0 (primitive) and physical type 0 (null). For example, if a Variant is required (like measurement above) and both value and typed_value are null, the returned value must be 00 (Variant null).

I read this as it requires being able to differentiate between being marked as invalid by a validity buffer (Scalar { Variant, None }) and having a valid value that is null (`Scalar { Variant, VariantValue::Null }`)

connortsui20: So I went on a little rabbit chase and I think you are right, but perhaps for the wrong reason. Variant shredding is an implementation detail, so it has nothing to do with how we model scalars.

Rather, most systems seem define a "VARIANT NULL" that is _distinct_ from a "SQL null" (which is equivalent to Vortex nulls). The semantics of this null type is different:

- Parquet Variant Binary Encoding spec (null as primitive type 0): https://parquet.apache.org/docs/file-format/types/variantencoding/#encoding-types
- Databricks "Query variant data" (defines SQL NULL vs Variant NULL): https://docs.databricks.com/aws/en/semi-structured/variant#variant-null-rules
- Databricks `is_variant_null` (shows `is_variant_null` vs `IS NULL` difference): https://docs.databricks.com/aws/en/sql/language-manual/functions/is_variant_null
- Databricks `isnotnull` (variant-encoded null returns true for `isnotnull`): https://docs.databricks.com/en/sql/language-manual/functions/isnotnull.html
- Spark `is_variant_null` function docs: https://spark.apache.org/docs/4.2.0-preview2/api/sql/variant-functions/

Crazy stuff.

So I would not be opposed to adding a `VariantNull` and `Object` variant to `ScalarValue` to support this. I feel like this is just slightly cleaner than defining a new `VariantValue` type that is effectively the same with 2 different variants. And if we eventually add the `Array` variant to `ScalarValue`, variant scalars would certainly benefit from that.

AdamGS: or maybe (as unlikely as it might sound), I went through the same rabbit hole when writing the RFC? the parquet spec even references SQL IS NULL at some point, it's very clear that they intended to have two different null values.

vortex-array/src/dtype/dtype_impl.rs
gatesn: If they're not semantically different, then Variant _is_ always nullable?

vortex-array/src/scalar/proto.rs
gatesn: This is all overly complicated.

This is just `ScalarValue::Variant(Scalar)`

AdamGS: what's the dtype of the inner scalar if it's VariantNull?

gatesn: What is VariantNull?

connortsui20: basically we shove an entire nullable scalar into a `ScalarValue::Variant(Scalar)` and then VariantNull is represented by the inner `Scalar` (not inner `ScalarValue`, that's a different thing) being null.

We need this because we need the variant values to have a `DType`

AdamGS: @gatesn that's the key point here. Like @connortsui20 outlined, I believe we need to be able to differentiate between them (as weird as it is) for our variant type to be generally useful.

gatesn: I'm not really following...

I think your `VariantNull == Variant(Scalar { dtype: DType::Null, ... })`

AdamGS: I'll go with that. I'm or tomorrow, but I'll pick it back up on Monday.

gatesn: I was bad at explaining the intuition here.

It's that the column's type is DType::Variant because each row can have its own type.

But scalar _is_ a single row.

So while the top-level scalar must declare its DType as DType::Variant, we must store a row-specific DType inside it.

And if we're storing a row-specific DType, and the data that underpins it, then we're really just storing a scalar.

On the null front, a variant array may store a null buffer. But really I just see that as "shredding the validity". I could choose not to shred the validity I guess

AdamGS: That makes perfect sense. But for variant types from the Parquet family (which includes Spark and Iceberg), that's not how it's defined. 

They have an optional null buffer (like most arrow arrays), and they have a second possible null state, that's is valid in the SQL sense - that's "VariantNull", it's not about shredding or not shredding the validity.

gatesn: I think they're wrong 😀 

But if that is the case, then we should add nullability to DType::Variant, and allow encodings to have a top-level nullability separately from a defined, but internally null, row scalar.

vortex-array/src/scalar/display.rs
```diff
@@ -21,10 +21,8 @@ impl Display for Scalar {
             DType::Struct(..) => write!(f, "{}", self.as_struct()),
             DType::List(..) | DType::FixedSizeList(..) => write!(f, "{}", self.as_list()),
             DType::Extension(_) => write!(f, "{}", self.as_extension()),
-            DType::Variant => match self.value() {
-                None => write!(f, "null"),
-                Some(value) => write!(f, "{value}"),
-            },
+            DType::Variant if self.is_null() => write!(f, "null"),
+            DType::Variant => write!(f, "variant({})", self.as_variant()),
```
AdamGS: Figure worth wrapping the display in with "variant" to make it clearer

vortex-array/src/scalar/scalar_impl.rs
```diff
@@ -190,10 +190,7 @@ impl Scalar {
             DType::FixedSizeList(_, list_size, _) => value.as_list().len() == *list_size as usize,
             DType::Struct(struct_fields, _) => value.as_list().len() == struct_fields.nfields(),
             DType::Extension(_) => self.as_extension().to_storage_scalar().is_zero()?,
-            DType::Variant => matches!(
-                value,
-                ScalarValue::Variant(crate::scalar::VariantValue::Null)
-            ),
+            DType::Variant => value.as_variant().is_null(),
```
AdamGS: Some of these are weird, I think `VariantNull` is the closest variant has for zero?

AdamGS:
Should we put this behind a feature? `unstable_variant` or something like that, just until we're sure about how all pieces fit together.

vortex-array/src/scalar/scalar_impl.rs
connortsui20: Based on the prior art I linked above we should say that the column null is still determined by `is_null` and variant null _must_ be handled by a special `is_variant_null`. So yes I think that for scalars `is_zero` can just delegate to `is_variant_null`! 🤯

vortex-array/src/arrays/constant/vtable/canonical.rs
```diff
@@ -164,6 +164,9 @@ pub(crate) fn constant_canonicalize(array: &ConstantArray) -> VortexResult<Canon
             let storage_self = ConstantArray::new(storage_scalar, array.len()).into_array();
             Canonical::Extension(ExtensionArray::new(ext_dtype.clone(), storage_self))
         }
+        DType::Variant => {
+            unimplemented!("TODO(variant): canonicalization semantics are not defined")
```
gatesn: Do you have any ideas what this might look like? I'm thinking it's very similar to ExtensionArray in that it holds a single child array, and basically does nothing else.

vortex-array/src/scalar/display.rs
```diff
@@ -21,6 +21,8 @@ impl Display for Scalar {
             DType::Struct(..) => write!(f, "{}", self.as_struct()),
             DType::List(..) | DType::FixedSizeList(..) => write!(f, "{}", self.as_list()),
             DType::Extension(_) => write!(f, "{}", self.as_extension()),
+            DType::Variant if self.is_null() => write!(f, "null"),
```
gatesn: Seems odd that this is different to the other variants? Can you handle this inside the as_variant result?

vortex-array/src/scalar/downcast.rs
```diff
@@ -153,6 +153,24 @@ impl Scalar {
         // for this extension type.
         Some(ExtScalar::new_unchecked(self.dtype(), self.value()))
     }
+
+    /// Returns the row-specific scalar wrapped by a variant, panicking if the scalar is not a
+    /// variant.
+    pub fn as_variant(&self) -> &Scalar {
```
gatesn: I think for consistency/symmetry let's wrap this up in a view like all the others.

vortex-array/src/scalar/proto.rs
```diff
@@ -238,6 +241,18 @@ impl ScalarValue {
             .as_ref()
             .ok_or_else(|| vortex_err!(Serde: "Scalar value missing kind"))?;
 
+        if dtype == &DType::Variant {
```
gatesn: Seems weird to handle this here when we have a match below? I think the logic can remain the same

vortex-array/src/scalar/scalar_impl.rs
gatesn: Blergh... I'm still questioning whether we distinguish null in the array vs present variant that holds a null (of possibly any dtype).

gatesn (commented):
Yeah this looks nice.

I want to spent a little bit more time thinking about / looking at the null semantics, but otherwise looking like we're basically there.

vortex-array/src/scalar/scalar_impl.rs
connortsui20: I think we have to since most other systems do that exactly

vortex-array/src/scalar/display.rs
connortsui20: I think we still want to have `DType::Variant(Nullability)` where that is considered the "outer" nullability that we track just like every other logical type. And then we distiguish this from variant null which is much closer to a "zero" value

AdamGS: yeah for sure, forgot about that point 🤦

vortex-array/src/arrays/constant/vtable/canonical.rs
AdamGS: yeah, that's what I have in my WIP branch, the only thing it actually does right now is have a pushdown rule that takes the variant expressions and applies them to the child

gatesn (approved):
Noice

PR #6917, opened 2026-03-12
back compat test

joseph-isaacs:
The part PR for https://github.com/vortex-data/rfcs/pull/23.

We add a generator and a validator, which will naturally be keep up to date with new releases. Future work can add this as a pre-release run.

Next these files must be generated for older versions of vortex.


vortex-test/compat-gen/src/adapter.rs
joseph-isaacs, line 1: This should keep the backwards change to this file and the fixtures

robert3005 (approved):

PR #6919, opened 2026-03-12
wip: fsst dedup

joseph-isaacs:
## Summary

Closes: #000


joseph-isaacs:
@AdamGS did you truncate these measurements look like some floats are. in compress

AdamGS:
I don't think any of this data was every really a float, what do you see that looks wrong in the new formatting?

PR #6920, opened 2026-03-12
Run benchmarks on object storage more times

AdamGS:
Part of the overall attempt to reduce overall noise in benchmarks, just running benchmarks on S3 twice as many times.


connortsui20 (approved):
yes yes

AdamGS:
just running benchmarks to makes sure I didn't break anything

PR #6921, opened 2026-03-12
Minify Duckdb build script

myrrc:
Also, reduce dependency on Itertools in vortex-duckdb


vortex-duckdb/src/convert/expr.rs
```diff
@@ -3,7 +3,7 @@
 
 use std::sync::Arc;
 
-use itertools::Itertools;
+//use itertools::Itertools;
```
0ax1: hello

0ax1 (approved):

PR #6925, opened 2026-03-12
Add FixedSizedList to Arbitrary DType implementation

robert3005:
Even though we implemented all the arbitrary generation logic for FixedSizedList
arrays we didn't enable them in the fuzzer, i.e. added them to DType generation
logic


connortsui20:
oh we disabled it because the arrow conversion is broken, don't actually know if they fixed it yet?

Ah my method is now on the latest version so we should be good if we change the arrow conversion: https://arrow.apache.org/rust/arrow/array/struct.FixedSizeListArray.html#method.try_new_with_length

robert3005:
oh, glad you remembered this one, this was defintely sad bug to find

robert3005:
Ok, updated to use new arrow method

connortsui20 (approved):
finally 🥳

PR #6926, opened 2026-03-12
Remove all_valid/all_invalid from Validity

gatesn:
These functions were performing hidden compute without an ExecutionCtx. 
I have also removed PartialEq, since that was executing arrays!

There now exists a Validity::mask_eq (similar to array_eq) which will evaluate only if necessary

### DIFF
--- a/encodings/zstd/src/compute/cast.rs
+++ b/encodings/zstd/src/compute/cast.rs
@@ -6,6 +6,7 @@ use vortex_array::IntoArray;
 use vortex_array::dtype::DType;
 use vortex_array::dtype::Nullability;
 use vortex_array::scalar_fn::fns::cast::CastReduce;
+use vortex_array::validity::Validity;
 use vortex_error::VortexResult;
 
 use crate::Zstd;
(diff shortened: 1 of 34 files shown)


encodings/zstd/src/compute/cast.rs
```diff
@@ -45,11 +46,10 @@ impl CastReduce for Zstd {
             }
             (Nullability::Nullable, Nullability::NonNullable) => {
                 // null => non-null works if there are no nulls in the sliced range
-                let sliced_len = array.slice_stop() - array.slice_start();
-                let has_nulls = !array
-                    .unsliced_validity
-                    .slice(array.slice_start()..array.slice_stop())?
-                    .all_valid(sliced_len)?;
+                let has_nulls = !matches!(
```
robert3005, line 49: You have to change the comment? This previously was specifically checking the sliced range

robert3005:
Did we not care in the other cases? I think I miss the place where ZSTD and Duckdb exporter enforced a tighter constraint. Maybe the change here is to ask for validity_mask?

robert3005 (approved):

PR #6935, opened 2026-03-13
perf[fsst]: like pushdown using a dfa

joseph-isaacs:
Fsst `like` execution without decompression.

This uses a DFA over the symbol table and the like expression.

Once this is proved out we could think about putting this in fsst-rs?


encodings/fsst/src/dfa/flat_contains.rs
```diff
@@ -0,0 +1,149 @@
...
+//! Flat `u8` transition table DFA for contains matching (`LIKE '%needle%'`).
+//!
+//! Uses an escape-sentinel strategy: the FSST escape code maps to a sentinel
+//! state, and the next literal byte is looked up in a separate byte-level
+//! transition table.
+//! This is to support needles up to u8::MAX long.
+//!
+//! ## Construction (needle = `"aba"`, symbols = `[0:"ab", 1:"ba"]`)
+//!
+//! ### Step 1: KMP byte-level transition table
```
0ax1: Should we spell out once that KMP stands for Knuth-Morris-Pratt](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm)?

encodings/fsst/src/compute/like.rs
```diff
@@ -0,0 +1,474 @@
...
+        array: &FSSTArray,
+        pattern: &ArrayRef,
+        options: LikeOptions,
+        _ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<ArrayRef>> {
+        let Some(pattern_scalar) = pattern.as_constant() else {
+            return Ok(None);
+        };
+
+        if options.case_insensitive {
```
0ax1, line 32: nit: How about `is_case_insensitive`?

encodings/fsst/src/compute/like.rs
```diff
@@ -0,0 +1,474 @@
...
+        let symbols = array.symbols();
+        let symbol_lengths = array.symbol_lengths();
+
+        let Some(matcher) =
+            FsstMatcher::try_new(symbols.as_slice(), symbol_lengths.as_slice(), pattern_str)?
+        else {
+            return Ok(None);
+        };
+
+        let negated = options.negated;
```
0ax1, line 59: nit: `is_negated`?

encodings/fsst/src/dfa/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+                }
+                MatcherInner::Contains(FlatContainsDfa::new(symbols, symbol_lengths, needle)?)
+            }
+        };
+
+        Ok(Some(Self { inner }))
+    }
+
+    /// Run the matcher on a single FSST-compressed code sequence.
+    #[inline]
```
0ax1: Unless you re-export the fn, this shouldn't be `inline`.

encodings/fsst/src/dfa/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        None
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Scan helper
+// ---------------------------------------------------------------------------
+
+// TODO: add N-way ILP overrun scan for higher throughput on short strings.
+#[inline]
```
0ax1: -//-

0ax1 (commented):
Initial first pass. Doing a second in depth round.

encodings/fsst/src/dfa/tests.rs
```diff
@@ -0,0 +1,281 @@
...
+    let arr: ArrayRef = array.into_array();
+    let pattern_arr = ConstantArray::new(pattern, len).into_array();
+    let result = Like
+        .try_new_array(len, LikeOptions::default(), [arr, pattern_arr])?
+        .into_array()
+        .execute::<Canonical>(&mut SESSION.create_execution_ctx())?;
+    Ok(result.into_bool())
+}
+
+#[rstest]
```
0ax1, line 243: Should we test:

- negated `like`
- multi-level fallback:
```
// The string "aabaabaabaab" requires multi-level KMP fallback at the 'a' after "aabaabaab"
#[case(&["aabaabaabaab", "aabaabaax", "xaabaabaab"], "%aabaabaab%", &[true, false, true])]
```
- multi-byte UTF-8 characters:
```
#[case(&["café latte", "naïve approach", "café noir"], "café%", &[true, false, true])]
#[case(&["日本語テスト", "日本語データ", "英語テスト"], "%日本語%", &[true, true, false])]
```

- cross-symbol boundaries - needles longer than 8 bytes:
```
// 10-byte needle, contains: match at start, middle, end, exact, and near-miss
#[case(
    &["abcdefghijxxx", "xxxabcdefghij", "xxabcdefghijxx", "abcdefghij", "abcdefghxx"],
    "%abcdefghij%",
    &[true, true, true, true, false]
)]

// 10-byte prefix: same needle but anchored at the start of the string
#[case(
    &["abcdefghijxxx", "abcdefghij", "xabcdefghij", "abcdefghxx"],
    "abcdefghij%",
    &[true, true, false, false]
)]

// 9-byte needle with KMP-relevant overlap ("abcabcabc"):
// failure table = [0,0,0,1,2,3,4,5,6], so a partial match of "abcabcab"
// followed by a mismatch must fall back to state 5 ("abcab"), not restart.
// This exercises multi-level KMP backtracking across symbol boundaries.
#[case(
    &["xxabcabcabcxx", "abcabcabc", "abcabcabx", "abcabcxx"],
    "%abcabcabc%",
    &[true, true, false, false]
)]
```

0ax1 (approved):

PR #6940, opened 2026-03-13
Fix `bitwise_unary_op`

### LINKED ISSUE #6895: Fuzzing Crash: VortexError in array_ops
### Analysis

**Crash Location**: `fuzz/fuzz_targets/array_ops.rs:33:assert_search_sorted`

**Error Message**:

```
Other error: Expected to find 62965u16 at Found(171) in root: vortex.runend(u16?, len=244) nbytes=87 B (100.00%) [all_valid]
  metadata: RunEndMetadata { ends_ptype: U8, num_runs: 29, offset: 0 }
  ends: vortex.primitive(u8, len=29) nbytes=29 B (33.33%) [nulls=0, min=15u8, max=244u8, strict]
    metadata: EmptyMetadata
    buffer: values host 29 B (align=1) (100.00%)
  values: vortex.primitive(u16?, len=29) nbytes=58 B (66.67%) [nulls=0, min=6939u16, max=65535u16]
    metadata: EmptyMetadata
    buffer: values host 58 B (align=2) (100.00%)
 from left but instead found it at Found(162) in step 2
```

<details>
<summary>Stack Trace</summary>

```
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/db3e99bbab28c6ca778b13222becdea54533d908/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
             at /rustc/db3e99bbab28c6ca778b13222becdea54533d908/library/core/src/panicking.rs:80:14
   2: panic_display<vortex_error::VortexError>
             at /rustc/db3e99bbab28c6ca778b13222becdea54533d908/library/core/src/panicking.rs:259:5
   3: __libfuzzer_sys_run
             at ./fuzz/fuzz_targets/array_ops.rs:33:19
   4: rust_fuzzer_test_input
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/src/lib.rs:363:60
   5: {closure#0}
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/src/lib.rs:62:9
   6: do_call<libfuzzer_sys::test_input_wrap::{closure_env#0}, i32>
             at /rustc/db3e99bbab28c6ca778b13222becdea54533d908/library/std/src/panicking.rs:581:40
   7: __rust_try
   8: catch_unwind<i32, libfuzzer_sys::test_input_wrap::{closure_env#0}>
             at /rustc/db3e99bbab28c6ca778b13222becdea54533d908/library/std/src/panicking.rs:544:19
   9: catch_unwind<libfuzzer_sys::test_input_wrap::{closure_env#0}, i32>
             at /rustc/db3e99bbab28c6ca778b13222becdea54533d908/library/std/src/panic.rs:359:14
  10: test_input_wrap
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/src/lib.rs:60:22
  11: _ZN6fuzzer6Fuzzer15ExecuteCallbackEPKhm
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/libfuzzer/FuzzerLoop.cpp:619:13
  12: _ZN6fuzzer6Fuzzer6RunOneEPKhmbPNS_9InputInfoEbPb
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/libfuzzer/FuzzerLoop.cpp:516:7
  13: _ZN6fuzzer6Fuzzer25ReadAndExecuteSeedCorporaERSt6vectorINS_9SizedFileESaIS2_EE
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/libfuzzer/FuzzerLoop.cpp:834:7
  14: _ZN6fuzzer6Fuzzer4LoopERSt6vectorINS_9SizedFileESaIS2_EE
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/libfuzzer/FuzzerLoop.cpp:872:3
  15: _ZN6fuzzer12FuzzerDriverEPiPPPcPFiPKhmE
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/libfuzzer/FuzzerDriver.cpp:923:6
  16: main
             at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libfuzzer-sys-0.4.12/libfuzzer/FuzzerMain.cpp:20:10
  17: <unknown>
  18: __libc_start_main
  19: _start
```

</details>

### Root Cause Analysis

The crash is a correctness bug in the SearchSorted implementation for RunEnd-encoded arrays. After applying a Mask filter and FillNull on a chunked u16 array (producing a RunEnd-encoded result with 29 runs over 244 elements), searching for value 62965u16 from the left returns Found(171) but a verification step finds the value at index Found(162), indicating the search result is inconsistent. The root cause is likely that the RunEnd encoding's search_sorted logic incorrectly translates between run-level indices and element-level indices, or fails to account for the offset/run-end boundaries when mapping a found position in the values array back to the corresponding position in the expanded array. The fix should be in the RunEnd encoding's SearchSorted implementation, ensuring correct index translation from the compressed runs to the logical element positions.

### Summary

- **Target**: `array_ops`
- **Crash File**: `crash-39852ca4462497e3b10749f2e7ad9699f0831d29`
- **Branch**: ct/fuzzer-report-template
- **Commit**: 09dfb09587a5b9cd7bdbf9d2f91e0c9f54330873
- **Crash Artifact**: https://github.com/vortex-data/vortex/actions/runs/22959766498/artifacts/5873413353

### Reproduction

1. Download the crash artifact:
   - **Direct download**: https://github.com/vortex-data/vortex/actions/runs/22959766498/artifacts/5873413353
   - Extract the zip file (`unzip`)
     - The path should look like `/path/to/array_ops/crash-39852ca4462497e3b10749f2e7ad9699f0831d29`
     - You can create a `./fuzz/artifacts` directory that will be git-ignored in the `vortex` repo
     - Full path would be `./fuzz/artifacts/array_ops/crash-39852ca4462497e3b10749f2e7ad9699f0831d29`

2. Reproduce locally:

```bash
cargo +nightly fuzz run -D --sanitizer=none array_ops /path/to/crash_file -- -rss_limit_mb=0
```

3. Get a backtrace:

```bash
RUST_BACKTRACE=1 cargo +nightly fuzz run -D --sanitizer=none array_ops /path/to/crash_file -- -rss_limit_mb=0
```

```bash
RUST_BACKTRACE=full cargo +nightly fuzz run -D --sanitizer=none array_ops /path/to/crash_file -- -rss_limit_mb=0
```

### Workflow Example

Assuming you download the zipfile to `~/Downloads`, and your working directory is the repository
root, you can follow these steps:

```bash
# Create the artifacts directory if you haven't already.
mkdir -p ./fuzz/artifacts

# Move the zipfile.
mv ~/Downloads/array_ops-crash-artifacts.zip ./fuzz/artifacts/

# Unzip the zipfile.
unzip ./fuzz/artifacts/array_ops-crash-artifacts.zip -d ./fuzz/artifacts/

# You can remove the zipfile now if you want to.
rm ./fuzz/artifacts/array_ops-crash-artifacts.zip
```

You can now reproduce with:

```bash
cargo +nightly fuzz run -D --sanitizer=none array_ops ./fuzz/artifacts/array_ops/crash-39852ca4462497e3b10749f2e7ad9699f0831d29 -- -rss_limit_mb=0
```

If you want a backtrace:

```bash
RUST_BACKTRACE=1 cargo +nightly fuzz run -D --sanitizer=none array_ops ./fuzz/artifacts/array_ops/crash-39852ca4462497e3b10749f2e7ad9699f0831d29 -- -rss_limit_mb=0
```

```bash
RUST_BACKTRACE=full cargo +nightly fuzz run -D --sanitizer=none array_ops ./fuzz/artifacts/array_ops/crash-39852ca4462497e3b10749f2e7ad9699f0831d29 -- -rss_limit_mb=0
```

---

_Auto-created by fuzzing workflow_

connortsui20:
## Summary

Closes: https://github.com/vortex-data/vortex/issues/6895 (which might seem completely unrelated but is relevant because of an incorrect mask)

Fixes a bug in `bitwise_unary_op` where the it used Arrow's `UnalignedBitChunk` iterator, which for buffers larger than 16 bytes (128 bits) uses `align_to::<u64>()` and introduces lead padding zeros when the byte pointer isn't u64-aligned. After applying the operation (e.g. NOT), those padding bits were written into a fresh, aligned output buffer at real data positions, corrupting the first padding bits of the result

This change just delegates to the correct `mut` implementation which is likely also faster.

## Testing

Adds a simple regression test.

### DIFF
--- a/vortex-buffer/src/bit/ops.rs
+++ b/vortex-buffer/src/bit/ops.rs
@@ -8,12 +8,9 @@ use crate::trusted_len::TrustedLenExt;
 
 #[inline]
 pub(super) fn bitwise_unary_op<F: FnMut(u64) -> u64>(buffer: &BitBuffer, op: F) -> BitBuffer {
-    let iter = buffer.unaligned_chunks().iter().map(op);
-    let iter = unsafe { iter.trusted_len() };
-
-    let result = Buffer::<u64>::from_trusted_len_iter(iter).into_byte_buffer();
-
-    BitBuffer::new_with_offset(result, buffer.len(), buffer.offset())
+    let mut buf = buffer.clone().into_mut();
+    bitwise_unary_op_mut(&mut buf, op);
+    buf.freeze()
 }
 
 #[inline]
@@ -89,6 +86,8 @@ pub(super) fn bitwise_binary_op<F: FnMut(u64, u64) -> u64>(
 
 #[cfg(test)]
 mod tests {
+    use std::ops::Not;
+
     use super::*;
     use crate::bitbuffer;
     use crate::buffer;
@@ -125,5 +124,27 @@ mod tests {
         let right = BitBuffer::new(buffer![10u8], 4);
         let result = bitwise_binary_op(&left, &right, |l, r| l ^ r);
         assert_eq!(result, bitbuffer![false, true, true, false]);
+    }
+
+    /// Regression test for a bug where [`bitwise_unary_op`] produced corrupt results when
+    /// the [`BitBuffer`]'s underlying byte pointer was not u64-aligned. Slicing a buffer by
+    /// a non-multiple-of-8 number of bytes can cause this misalignment. The bug only
+    /// manifested for buffers larger than 16 bytes (> 128 bits), because Arrow's
+    /// `UnalignedBitChunk` switches from byte-copying to `align_to` at that threshold.
+    ///
+    /// Issue: <https://github.com/vortex-data/vortex/issues/6895>
+    #[test]
+    fn test_bitwise_unary_not_misaligned_buffer() {
+        // Slice off 1 byte to shift the pointer off u64 alignment. Use 129 bits (17 bytes)
+        // to exceed the 16-byte threshold where `UnalignedBitChunk` uses `align_to`.
+        let padded = BitBuffer::new_set(8 + 129);
+        let buf = padded.slice(8..8 + 129);
+
+        let result = buf.not();
+        assert_eq!(
+            result.true_count(),
+            0,
+            "expected all-false after NOT of all-true"
+        );
     }
 }


robert3005:
Ah, that's why you can't just change this. I made this change in https://github.com/vortex-data/vortex/pull/6880 which means that binary op needs adjustment as well. I don't see how the mut op would be faster though

robert3005:
align_to is not the problem here though? It's that the padding is wrong so offset and length are wrong

connortsui20:
I thought that align to caused the padding to be wrong in this case.

Anyways I don't see much reason to not just delegate to the _mut version, the regression on codspeed is likely because it was just incorrect before.

robert3005:
align_to just splits the buffer (it's a const function). But the padding is clearly not preserved for any buffer in 17-32 range potentially. Anyway I agree with your fix. We should make sure to fix the binary op as well since we're here

connortsui20:
what do you mean by fix the binary op?

robert3005 (approved):
I can follow up with the fix to binary op

robert3005:
Ok, binary op is not a problem since it's only ever used for aligned arrays (i.e. 8 byte aligned) but we really only need 1 byte alignment

vortex-buffer/src/bit/ops.rs
```diff
@@ -8,12 +8,9 @@ use crate::trusted_len::TrustedLenExt;
 
 #[inline]
 pub(super) fn bitwise_unary_op<F: FnMut(u64) -> u64>(buffer: &BitBuffer, op: F) -> BitBuffer {
-    let iter = buffer.unaligned_chunks().iter().map(op);
-    let iter = unsafe { iter.trusted_len() };
-
-    let result = Buffer::<u64>::from_trusted_len_iter(iter).into_byte_buffer();
-
-    BitBuffer::new_with_offset(result, buffer.len(), buffer.offset())
+    let mut buf = buffer.clone().into_mut();
```
gatesn, line 11: @robert3005 we should kill the `into_mut()` functions. If we cannot into_mut, then we do a memcopy for no reason since we're about to overwrite the data _anyway_.

We should just have try_into_mut and the caller can either do in-place, or allocate a new buffer and compute directly into that.

PR #6941, opened 2026-03-13
Ignore justfile

connortsui20:
## Summary

Ignores justfile.

## Testing

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -219,6 +219,9 @@ profile.json.gz
 # Clang compilation database (https://clang.llvm.org/docs/JSONCompilationDatabase.html)
 compile_commands.json
 
+# Commands
+/justfile
+
 # AI Agents
 .claude
 .opencode


robert3005 (dismissed):

.gitignore
```diff
@@ -30,6 +30,9 @@ share/python-wheels/
 *.egg
 MANIFEST
 
+# Commands
```
joseph-isaacs: do this locally?

connortsui20: wdym do this locally? The gitignore is tracked by the repo. Is there a way to have a different gitignore than the repo?

AdamGS (approved):

robert3005 (approved):

.gitignore
```diff
@@ -219,6 +219,9 @@ profile.json.gz
 # Clang compilation database (https://clang.llvm.org/docs/JSONCompilationDatabase.html)
 compile_commands.json
 
+# Commands
+/justfile
```
a10y, line 223: i suspect you wanted to do this without the `/`

connortsui20: justfile only works if it is in the current directory, so this was on purpose (dont want these all over the place)

PR #6957, opened 2026-03-13
Suppress -Wtype-limits warning in DuckDB C++ compilation

joseph-isaacs:


0ax1 (approved):

0ax1 (commented):
Does the warning come from DDB? If so this should be part of the push/push diagnostics macro and move out of the build.rs.

PR #6960, opened 2026-03-13
Type Coercion

gatesn:
This allows Vortex expressions to be typed in contexts that currently don't have their own logical planner, for example expressions passed into PyVortex.


vortex-array/src/expr/transform/coerce.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        // Build new children, inserting casts where needed.
+        let new_children: Vec<Expression> = node
+            .children()
+            .iter()
+            .zip(coerced_dtypes.iter())
+            .map(|(child, target)| {
+                let child_dtype = child.return_dtype(&scope)?;
+                if child_dtype.eq_ignore_nullability(target)
+                    && child_dtype.nullability() == target.nullability()
```
connortsui20, line 53: this is just eq?

vortex-array/src/dtype/coercion.rs
```diff
@@ -0,0 +1,468 @@
...
+impl DType {
+    /// The core primitive — what type can hold both `self` and `other`?
+    /// Returns `None` if no common supertype exists.
+    pub fn least_supertype(&self, other: &DType) -> Option<DType> {
+        // 1. Identity (ignoring nullability): return self with union nullability
+        if self.eq_ignore_nullability(other) {
+            return Some(self.with_nullability(self.nullability() | other.nullability()));
+        }
+
+        let union_null = self.nullability() | other.nullability();
```
connortsui20, line 85: Nit: move this up

connortsui20 (approved):

AdamGS:
We used to talk about expressions as being tightly typed and physical, this seems like a pretty big break from that?

connortsui20:
I think we need to have the ability to upcast due to our type narrowing? I believe the idea is that we want the ability to do query planning within Vortex without relying on other systems, and in that case we need to know what is possible to coerce / cast. I guess it's an open question on if this is the right way to do this.

PR #6963, opened 2026-03-13
 Fix fuzzer issue template

connortsui20:
## Summary

From https://github.com/vortex-data/vortex/pull/6936 I do not understand why we are so against being descriptive in these templates given most of what claude says is complete nonsense. If we want concision, the reproduction steps still need to make sense.

This changes the fuzzer template to have 2 different details sections, one with a full and correct description of how to reproduce, and the other with a single command for people who know what they are doing. The maintainers do not need to look at the first section.

## Testing

N/A

### DIFF
--- a/.github/scripts/fuzz_report/templates/new_issue.md
+++ b/.github/scripts/fuzz_report/templates/new_issue.md
@@ -40,25 +40,33 @@ cargo +nightly fuzz run -D --sanitizer=none {{FUZZ_TARGET}} ./fuzz/artifacts/{{F
 ```
 
 <details>
-<summary>First-time setup: download and extract the crash artifact</summary>
+<summary>Reproduction Steps</summary>
 
-1. Download the crash artifact:
-   - **Direct download**: {{ARTIFACT_URL}}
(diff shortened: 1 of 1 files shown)


joseph-isaacs:
We should optimize for small issue descriptions and someone familiar with the fuzzer.
We could have a hidden getting started section on the bottom, but that should be the less easy thing to access.

joseph-isaacs:
I am very happy to hide more of the issue, I just really don't want to scroll past instructions I already know each time I open a fuzzer issue

connortsui20:
In that case can we hide **all** of the Claude output? It's a huge wall of text that in my experience has rarely been useful.

And if we want to optimize for readability and convenience we should hide the debug output completely as well because that also usually tells us nothing more than the backtrace already gives us.

I can keep going, but to me this conversation makes zero sense when we can just collapse everything into details. The current template is already small enough that you don't need to scroll to anything.

joseph-isaacs (approved):
LG, I am bad at reading diff markdown

Issue #6865, opened 2026-03-10
Tracking Issue: Tensor Types


connortsui20:
# Tensors

We would like to add tensor types as extension types to Vortex.

Tensors are multi-dimensional (n-dimensional) arrays that generalize vectors (1D) and matrices (2D)
to arbitrary dimensions. They are quite common in ML/AI and scientific computing applications. To
name just a few examples:

- Image or video data stored as `height x width x channels`
- Multi-dimensional sensor or time-series data
- Embedding vectors from language models and recommendation systems

## Tensor Types

There are different kinds of Tensors that we would like to add.

- `FixedShapeTensor`: a tensor with a fixed physical shape and dimensions
- `VariableShapeTensor`: a tensor with variable shape but fixed dimensions
- `Matrix`: a 2-dimensional tensor
- `Vector`: a 1-dimensional tensor (flat)
- `UnitVec`: a unit-normalized d-dimensional vector

Implementation of these types and related functionality like expressions (`CosineSimilarity`) will live in the `vortex-tensor` crate.

## Prior Art

- Apache Arrow [Canonical Extensions](https://arrow.apache.org/docs/format/CanonicalExtensions.html)

## Steps / History

- [X] Extensions Types [RFC](https://vortex-data.github.io/rfcs/rfc/0005.html) 
- [ ] Fixed-shape Tensor
  - [X] [RFC](https://vortex-data.github.io/rfcs/rfc/0024.html)
  - [X] https://github.com/vortex-data/vortex/pull/6812
  - [X] https://github.com/vortex-data/vortex/pull/6857
  - [ ] Logical shape cast
  - [ ] Exporters
    - [ ] Arrow
    - [ ] NumPy
    - [ ] PyTorch
- [ ] Variable-size Tensor
- [ ] ~~`UnitVec`: https://github.com/vortex-data/vortex/issues/6855~~
- [ ] Vectors: https://github.com/vortex-data/vortex/pull/6964

## Unresolved Questions

TODO

PR #6964, opened 2026-03-13
Vector Extension Type

connortsui20:
## Summary

Tracking Issue: https://github.com/vortex-data/vortex/issues/6865

Adds a `Vector` extension type and a new `L2Norm` expression.

Additionally adds a `AnyTensor` type that can be matched on for any kind of tensor we want.

Right now the code assumes that everything is built on top of `FixedSizeList`, but in the future that might change.

Additionally make some touchups to the `vortex-tensor` crate in general.

## API Changes

The new `Vector` and `L2Norm` types.

## Testing

Some basic tests.

### DIFF
--- a/vortex-array/src/scalar_fn/vtable.rs
+++ b/vortex-array/src/scalar_fn/vtable.rs
@@ -92,6 +92,8 @@ pub trait ScalarFnVTable: 'static + Sized + Clone + Send + Sync {
         Ok(args.to_vec())
     }
 
+    // TODO(connor): This needs a precondition for the number of args it has, or all implementations
+    // need to return an error if it is wrong.
     /// Compute the return [`DType`] of the expression if evaluated over the given input types.
     ///
     /// # Preconditions
--- a/vortex-array/src/scalar_fn/vtable.rs
+++ b/vortex-array/src/scalar_fn/vtable.rs
@@ -92,8 +92,6 @@ pub trait ScalarFnVTable: 'static + Sized + Clone + Send + Sync {
         Ok(args.to_vec())
     }
 
-    // TODO(connor): This needs a precondition for the number of args it has, or all implementations
-    // need to return an error if it is wrong.
     /// Compute the return [`DType`] of the expression if evaluated over the given input types.
     ///
     /// # Preconditions
--- a/vortex-tensor/src/fixed_shape/vtable.rs
+++ b/vortex-tensor/src/fixed_shape/vtable.rs
@@ -22,7 +22,7 @@ impl ExtVTable for FixedShapeTensor {
     type NativeValue<'a> = &'a ScalarValue;
 
     fn id(&self) -> ExtId {
-        ExtId::new_ref("vortex.fixed_shape_tensor")
+        ExtId::new_ref("vortex.tensor.fixed_shape_tensor")
     }
 
     fn serialize_metadata(&self, metadata: &Self::Metadata) -> VortexResult<Vec<u8>> {
--- a/vortex-tensor/src/scalar_fns/cosine_similarity.rs
+++ b/vortex-tensor/src/scalar_fns/cosine_similarity.rs
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-//! Cosine similarity expression for [`FixedShapeTensor`](crate::fixed_shape::FixedShapeTensor)
-//! arrays.
+//! Cosine similarity expression for tensor-like extension arrays
+//! ([`FixedShapeTensor`](crate::fixed_shape::FixedShapeTensor) and
+//! [`Vector`](crate::vector::Vector)).
 
 use std::fmt::Formatter;
 
@@ -19,6 +20,7 @@ use vortex::array::match_each_float_ptype;
 use vortex::dtype::DType;
 use vortex::dtype::NativePType;
 use vortex::dtype::Nullability;
+use vortex::dtype::extension::Matcher;
 use vortex::error::VortexResult;
 use vortex::error::vortex_bail;
 use vortex::error::vortex_ensure;
@@ -31,18 +33,21 @@ use vortex::scalar_fn::ExecutionArgs;
 use vortex::scalar_fn::ScalarFnId;
 use vortex::scalar_fn::ScalarFnVTable;
 
+use crate::matcher::AnyTensor;
+
 // TODO(connor): We will want to add implementations for unit normalized vectors and also vectors
 // encoded in spherical coordinates.
 /// Cosine similarity between two columns.
 ///
-/// For [`FixedShapeTensor`], computes `dot(a, b) / (||a|| * ||b||)` over the flat backing buffer of
-/// each tensor. The shape and permutation do not affect the result because cosine similarity only
-/// depends on the element values, not their logical arrangement.
+/// Computes `dot(a, b) / (||a|| * ||b||)` over the flat backing buffer of each tensor or vector.
+/// The shape and permutation do not affect the result because cosine similarity only depends on the
+/// element values, not their logical arrangement.
 ///
-/// Right now, both inputs must be [`FixedShapeTensor`] extension arrays with the same dtype and a
-/// float element type. The output is a float column of the same float type.
+/// Both inputs must be tensor-like extension arrays ([`FixedShapeTensor`] or [`Vector`]) with the
+/// same dtype and a float element type. The output is a float column of the same float type.
 ///
 /// [`FixedShapeTensor`]: crate::fixed_shape::FixedShapeTensor
+/// [`Vector`]: crate::vector::Vector
 #[derive(Clone)]
 pub struct CosineSimilarity;
 
@@ -92,10 +97,14 @@ impl ScalarFnVTable for CosineSimilarity {
 
         // We don't need to look at rhs anymore since we know lhs and rhs are equal.
 
-        // Both inputs must be extension types.
+        // Both inputs must be tensor-like extension types.
         let lhs_ext = lhs.as_extension_opt().ok_or_else(|| {
             vortex_err!("cosine_similarity lhs must be an extension type, got {lhs}")
         })?;
+        vortex_ensure!(
+            AnyTensor::matches(lhs_ext),
+            "cosine_similarity inputs must be an `AnyTensor`, got {lhs}"
+        );
 
         // Extract the element dtype from the storage FixedSizeList.
         let element_dtype = lhs_ext
@@ -257,6 +266,7 @@ mod tests {
     use vortex::dtype::Nullability;
     use vortex::dtype::extension::ExtDType;
     use vortex::error::VortexResult;
+    use vortex::extension::EmptyMetadata;
     use vortex::scalar::Scalar;
     use vortex::scalar_fn::EmptyOptions;
     use vortex::scalar_fn::ScalarFn;
@@ -264,6 +274,7 @@ mod tests {
     use crate::fixed_shape::FixedShapeTensor;
     use crate::fixed_shape::FixedShapeTensorMetadata;
     use crate::scalar_fns::cosine_similarity::CosineSimilarity;
+    use crate::vector::Vector;
 
     /// Builds a [`FixedShapeTensor`] extension array from flat f64 elements and a logical shape.
     ///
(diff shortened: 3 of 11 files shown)


vortex-array/src/scalar_fn/vtable.rs
```diff
@@ -92,6 +92,8 @@ pub trait ScalarFnVTable: 'static + Sized + Clone + Send + Sync {
         Ok(args.to_vec())
     }
 
+    // TODO(connor): This needs a precondition for the number of args it has, or all implementations
```
gatesn: There should already be a check using the `arity` of the function

vortex-tensor/src/scalar_fns/cosine_similarity.rs
```diff
@@ -79,46 +82,40 @@ impl ScalarFnVTable for CosineSimilarity {
...
-                    "cosine_similarity storage dtype must be a FixedSizeList, got {}",
-                    lhs_ext.storage_dtype()
-                )
-            })?;
-
-        // Element dtype must be a non-nullable float primitive.
         vortex_ensure!(
-            element_dtype.is_float(),
-            "cosine_similarity element dtype must be a float primitive, got {element_dtype}"
+            AnyTensor::matches(lhs_ext),
```
gatesn: Can you do `lhs_ext.is::<AnyTensor>()`

vortex-tensor/src/scalar_fns/l2_norm.rs
```diff
@@ -0,0 +1,229 @@
...
+/// The input must be a tensor-like extension array with a float element type. The output is a float
+/// column of the same float type.
+#[derive(Clone)]
+pub struct L2Norm;
+
+impl ScalarFnVTable for L2Norm {
+    type Options = EmptyOptions;
+
+    fn id(&self) -> ScalarFnId {
+        ScalarFnId::new_ref("vortex.l2_norm")
```
gatesn: I think the namespace shouldn't be `vortex core` here. 

I think you might want `tensor.l2_norm` and also change the other ext dtypes in this crate? Or `vortex.tensor.l2_norm`?

vortex-array/src/scalar_fn/vtable.rs
connortsui20: I tried backtracing through the code myself but I couldn't find where that is called

vortex-tensor/src/scalar_fns/l2_norm.rs
connortsui20: I'll go with `vortex.tensor.l2_norm` for now (and ill do similar for others), we can change it later if we want

vortex-tensor/src/scalar_fns/l2_norm.rs
```diff
@@ -0,0 +1,228 @@
...
+        Ok(Some(expression.child(0).validity()?))
+    }
+
+    fn is_null_sensitive(&self, _options: &Self::Options) -> bool {
+        false
+    }
+
+    fn is_fallible(&self, _options: &Self::Options) -> bool {
+        // Canonicalization of the storage array can fail.
+        true
```
joseph-isaacs: This is not the meaning of this

joseph-isaacs: Its asks if the operation can fail liked checked_add (overflow). Can this happen here

connortsui20:
```
    /// Returns whether this expression itself is fallible. Conservatively default to *true*.
    ///
    /// An expression is runtime fallible is there is an input set that causes the expression to
    /// panic or return an error, for example checked_add is fallible if there is overflow.
    ///
    /// Note: this is only applicable to expressions that pass type-checking
    /// [`ScalarFnVTable::return_dtype`].
```

In that case this doc comment needs to be more detailed since that was not clear. What is a better description here? How do we distinguish between an execution failure vs a logical failure when we get a `VortexResult` back regardless?

vortex-array/src/scalar_fn/vtable.rs
connortsui20: https://github.com/vortex-data/vortex/pull/7027

vortex-tensor/src/scalar_fns/l2_norm.rs
connortsui20: https://github.com/vortex-data/vortex/pull/7027

vortex-tensor/src/scalar_fns/cosine_similarity.rs
```diff
@@ -244,77 +187,32 @@ fn cosine_similarity_row<T: Float + NativePType>(a: &[T], b: &[T]) -> T {
...
+    use crate::scalar_fns::utils::test_helpers::assert_close;
+    use crate::scalar_fns::utils::test_helpers::constant_tensor_array;
+    use crate::scalar_fns::utils::test_helpers::constant_vector_array;
+    use crate::scalar_fns::utils::test_helpers::tensor_array;
+    use crate::scalar_fns::utils::test_helpers::vector_array;
 
     /// Evaluates cosine similarity between two tensor arrays and returns the result as `Vec<f64>`.
-    fn eval_cosine_similarity(lhs: ArrayRef, rhs: ArrayRef, len: usize) -> VortexResult<Vec<f64>> {
+    fn eval_cosine_similarity(
+        lhs: vortex::array::ArrayRef,
```
AdamGS: nit - remove the fully qualified name

vortex-tensor/src/vector/vtable.rs
```diff
@@ -0,0 +1,142 @@
...
+use vortex::extension::EmptyMetadata;
+use vortex::scalar::ScalarValue;
+
+use crate::vector::Vector;
+
+impl ExtVTable for Vector {
+    type Metadata = EmptyMetadata;
+
+    // TODO(connor): This is just a placeholder for now.
+    type NativeValue<'a> = &'a ScalarValue;
```
AdamGS, line 20: why?

vortex-tensor/src/scalar_fns/cosine_similarity.rs
```diff
@@ -332,20 +226,18 @@ mod tests {
             ],
         )?;
 
-        // Row 0: identical → 1.0, row 1: orthogonal → 0.0.
+        // Row 0: identical -> 1.0, row 1: orthogonal -> 0.0.
```
AdamGS, line 230: 😂

vortex-tensor/src/vector/vtable.rs
connortsui20: It's mainly because we are blocked on https://github.com/vortex-data/vortex/pull/6717 and have to figure some things out on that first

connortsui20: In reality we don't actually care about this now, but once we start using these extension types in the compressor then we do care because we would want a more efficient representation of a vector when (for example) encoded as a `ConstantArray`

vortex-tensor/src/scalar_fns/utils.rs
```diff
@@ -0,0 +1,236 @@
+// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: just a personal taste thing - but these hyper specialized utils just crate fragmentation in the codebase, its more layers and they only exist in this corner

vortex-tensor/src/scalar_fns/cosine_similarity.rs
```diff
@@ -31,26 +27,31 @@ use vortex::scalar_fn::ExecutionArgs;
...
+use crate::scalar_fns::utils::extension_list_size;
+use crate::scalar_fns::utils::extension_storage;
+use crate::scalar_fns::utils::extract_flat_elements;
+
 /// Cosine similarity between two columns.
 ///
-/// For [`FixedShapeTensor`], computes `dot(a, b) / (||a|| * ||b||)` over the flat backing buffer of
-/// each tensor. The shape and permutation do not affect the result because cosine similarity only
-/// depends on the element values, not their logical arrangement.
+/// Computes `dot(a, b) / (||a|| * ||b||)` over the flat backing buffer of each tensor or vector.
```
AdamGS, line 38: Not part of this PR, but I think we also want to do expr(array, scalar) for this sort of thing

AdamGS (approved):
This seems very reasonable to me, some pieces depend on future work, like Variant we probably want a standard way to express these sort of stability/maturity guarantees

vortex-tensor/src/scalar_fns/utils.rs
connortsui20: Not really sure where else to put this though? We need it for both of the scalar fns in here (and will need it for more in the future)

vortex-tensor/src/fixed_shape/vtable.rs
```diff
@@ -22,7 +22,7 @@ impl ExtVTable for FixedShapeTensor {
     type NativeValue<'a> = &'a ScalarValue;
 
     fn id(&self) -> ExtId {
-        ExtId::new_ref("vortex.fixed_shape_tensor")
+        ExtId::new_ref("vortex.tensor.fixed_shape_tensor")
```
danking, line 25: Are we certain no one has written a `vortex.fixed_shape_tensor`?

vortex-tensor/src/scalar_fns/l2_norm.rs
```diff
@@ -0,0 +1,219 @@
...
+
+        let storage = extension_storage(&input)?;
+        let flat = extract_flat_elements(&storage, list_size)?;
+
+        match_each_float_ptype!(flat.ptype(), |T| {
+            let result: PrimitiveArray = (0..row_count)
+                .map(|i| l2_norm_row(flat.row::<T>(i)))
+                .collect();
+
+            Ok(result.into_array())
```
danking, line 123: Do we intend to eventually replace this with a call to BLAS? It seems like extract_flat_elements produces a BLAS compatible matrix.

vortex-tensor/src/scalar_fns/utils.rs
AdamGS: you can just inline them? they seem to be mostly very short and easy to quickly read through, even if it introduces some code duplication

vortex-tensor/src/scalar_fns/utils.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[must_use]
+    pub fn ptype(&self) -> PType {
+        self.elems.ptype()
+    }
+
+    /// Returns the `i`-th row as a typed slice of length `list_size`.
+    #[must_use]
+    pub fn row<T: NativePType>(&self, i: usize) -> &[T] {
+        let slice = self.elems.as_slice::<T>();
+        &slice[i * self.stride..i * self.stride + self.list_size]
```
danking: a personal style thing but I like `slice[i*self.stride..][..self.list_slize]`

vortex-tensor/src/fixed_shape/vtable.rs
connortsui20: We're pretty sure, this is a very recent addition

vortex-tensor/src/scalar_fns/l2_norm.rs
connortsui20: what is BLAS?

vortex-tensor/src/scalar_fns/utils.rs
connortsui20: Hmmm I'm not sure I agree with this, especially since that would mean duplicating the `FlatElements` struct I introduced which feels wrong.

vortex-tensor/src/scalar_fns/cosine_similarity.rs
connortsui20: yeah that is also part of why we want https://github.com/vortex-data/vortex/pull/6717

vortex-tensor/src/scalar_fns/l2_norm.rs
danking: https://www.netlib.org/blas/

Basic Linear Algebra Subprograms. It's _the_ way to do fast linear algebra. There's also LAPACK which mostly deals with matrix transformations e.g. QR factorization.

There's many implementations of the BLAS interface. Perhaps the most interesting is GotoBLAS which this guy hand wrote in assembly. If you're on Intel processors you really want to link against the Intel Matrix Kernel Library (MKL). There's also OpenBLAS which is arch independent and cuBLAS which is for GPUs. Generally, one expects these libraries to be installed on the machine and you dynamically link against them.

connortsui20: is there a standard rust crate that does this? If we can pull that in and it operates over flat elements then that is a trivial thing to add. I looked online and found a few things but am not familiar with this space

PR #6966, opened 2026-03-14
Split up binary ScalarFn

gatesn:
Edit: im not convinced of this PR, but the benchmarks have surfaced some interesting results. So digging in

---
Adds new scalar functions for comparison, arithmetic, and logical operators.

Previously the single binary function typically just switched on the logic for each of these, making it much harder to read and follow.

This also gives us more granular kernel selection, e.g. for arrays that wish to provide a comparison kernel we would previously have to invoke the kernel for _any_ binary function.

### DIFF
--- a/vortex-array/src/scalar_fn/fns/binary/compare.rs
+++ b/vortex-array/src/scalar_fn/fns/binary/compare.rs
@@ -27,6 +27,7 @@ use crate::dtype::Nullability;
 use crate::kernel::ExecuteParentKernel;
 use crate::scalar::Scalar;
 use crate::scalar_fn::fns::binary::Binary;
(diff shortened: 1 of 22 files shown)


vortex-array/src/scalar_fn/fns/binary/compare.rs
```diff
@@ -105,6 +106,56 @@ where
     }
 }
 
+/// Adaptor that bridges [`CompareKernel`] implementations to [`ExecuteParentKernel`] for the
+/// new [`Comparison`] scalar function.
+#[derive(Default, Debug)]
+pub struct ComparisonCompareExecuteAdaptor<V>(pub V);
```
gatesn, line 112: I hate this. we should just rewrite into the new kernel traits

PR #6979, opened 2026-03-16
fix[sparse]: correctly handle cast for all-patched

joseph-isaacs:
Fixes https://github.com/vortex-data/vortex/issues/6932

PR #6980, opened 2026-03-16
Sparse followup

connortsui20:
## Summary

Followup to https://github.com/vortex-data/vortex/pull/6979

We should use the default value here to fast path the null, and also the comment was incorrect since casting can always fail.

## Testing

N/A

### DIFF
--- a/encodings/sparse/src/compute/cast.rs
+++ b/encodings/sparse/src/compute/cast.rs
@@ -14,18 +14,18 @@ use crate::SparseArray;
 
 impl CastReduce for Sparse {
     fn cast(array: &SparseArray, dtype: &DType) -> VortexResult<Option<ArrayRef>> {
-        let casted_fill = if array.patches().num_patches() == array.len() {
-            // When every position is patched the fill is unused — skip casting it
-            // entirely and substitute a zero value for the target dtype.
-            Scalar::zero_value(dtype)
-        } else {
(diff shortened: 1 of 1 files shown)


encodings/sparse/src/compute/cast.rs
```diff
@@ -14,19 +14,19 @@ use crate::SparseArray;
...
             .clone()
             .map_values(|values| values.cast(dtype.clone()))?;
 
+        let casted_fill = if array.patches().num_patches() == array.len() {
+            // When every position is patched the fill scalar is unused and can be undefined.
+            // We skip casting it entirely and substitute a default value for the target dtype.
+            Scalar::default_value(dtype)
+        } else {
+            array.fill_scalar().cast(dtype)?
+        };
```
joseph-isaacs, line 28: why move this

connortsui20: just following the order of the constructor, no other reason

joseph-isaacs: okay, also why default vs zero?

connortsui20: if the type is nested it could be expensive (allocations).

Versus just creating a null scalar which is cheap.

joseph-isaacs (approved):

PR #6984, opened 2026-03-16
Add high-level docs to `vortex-btrblocks`

connortsui20:
## Summary

Decided to add some docs to `vortex-btrblocks` as I went through it to try and understand it better.

## Testing

N/A

### DIFF
--- a/vortex-btrblocks/src/lib.rs
+++ b/vortex-btrblocks/src/lib.rs
@@ -18,6 +18,22 @@
 //! - **Statistical Analysis**: Uses data sampling and statistics to predict compression ratios
 //! - **Recursive Structure Handling**: Compresses nested structures like structs and lists
 //!
+//! # How It Works
+//!
+//! [`BtrBlocksCompressor::compress()`] takes an `&ArrayRef` and returns an `ArrayRef` that may
+//! use a different encoding. It first canonicalizes the input, then dispatches by type.
+//! Primitives go to a type-specific `Compressor` (integer, float, or string). Compound types
+//! like structs and lists recurse into their fields and elements.
+//!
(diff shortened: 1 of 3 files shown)


AdamGS (approved):
love it

vortex-btrblocks/src/lib.rs
```diff
@@ -18,6 +18,22 @@
...
+//!
+//! [`BtrBlocksCompressor::compress()`] takes an `&ArrayRef` and returns an `ArrayRef` that may
+//! use a different encoding. It first canonicalizes the input, then dispatches by type.
+//! Primitives go to a type-specific `Compressor` (integer, float, or string). Compound types
+//! like structs and lists recurse into their fields and elements.
+//!
+//! Each type-specific compressor holds a static list of `Scheme` implementations (e.g.
+//! BitPacking, ALP, Dict). There is no dynamic registry. The compressor evaluates each scheme by
+//! compressing a ~1% sample and measuring the ratio, then picks the best. See `SchemeExt` for
+//! details on how sampling works.
```
joseph-isaacs, line 31: The type-specific one has encodings switched on or off

joseph-isaacs: But we should likely allow fully dynamic

connortsui20: That is kind of the point of why I was trying to understand this better, I am going to see if we can easily make the compressor pluggable (so that we can compress extension types into custom encodings).

joseph-isaacs: You would need to add schemas to a registry?

connortsui20: Yep, or something potentially more complicated, or a reworking of how we register schemes in the first place

PR #6985, opened 2026-03-16
[Experiment] Random IO ideas

AdamGS:
This is a mostly codex-driven experiment to play around with some general ideas I had and never got around to diving into, at least locally I've reached a point where perf looks interesting enough that I want to actually run benchmarks.

While I wat for benchmarks to run, some initial numbers from f3c23a6dec8b52c7df77366346173447172e300a, running clickbench with datafusion against develop (using GNU time):
- peak RSS is down (16.5 GB vs 21.1GB)
- Minor page faults down (~2.2 million vs 2.9 million).
- Voluntary context switches (not sure how to interpret this number, it also a bit noisy): Down to 317 vs 9,422

### DIFF
--- a/scripts/run_clickbench_rss.sh
+++ b/scripts/run_clickbench_rss.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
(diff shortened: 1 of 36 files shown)


danking:
Do you think the vortex-compact improvements are from more parallelism in decompression?

PR #6989, opened 2026-03-16
fix: Zstd canonicalize with >2GB decompressed buffers

SumedhArani:
Fixes a panic in `reconstruct_views` when a `ZstdArray`'s decompressed string data exceeds ~2 GiB. The function builds `BinaryView` structs that use `u32` offsets into the decompressed buffer; when the buffer is larger than `i32::MAX` the offset cast panics.

This is the same class of bug fixed for VarBin/FSST in #5961, but Zstd was not addressed because its decompressed buffer interleaves `u32` length prefixes with string data (unlike VarBin/FSST which have a separate lengths array).

The fix splits the decompressed buffer at value boundaries (zero-copy via `ByteBuffer::slice`) when approaching `i32::MAX`, using `BinaryView`'s native `buffer_index` field to reference the correct segment. The `i32::MAX` limit matches the convention established in #5961 per the Arrow spec that `BinaryView` offsets are logically signed. The existing `pub const MAX_BUFFER_LEN` from `vortex_array::arrays::varbinview::build_views` is reused rather than redefining the limit locally.

Signed-off-by: Sumedh Arani <sumedh@langchain.dev>

### DIFF
--- a/encodings/zstd/public-api.lock
+++ b/encodings/zstd/public-api.lock
@@ -224,4 +224,4 @@ pub fn vortex_zstd::ZstdMetadata::clear(&mut self)
 
 pub fn vortex_zstd::ZstdMetadata::encoded_len(&self) -> usize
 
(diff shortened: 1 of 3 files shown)


SumedhArani:
@a10y, would appreciate if you could take a look at this? Thanks!

a10y (approved):

PR #6990, opened 2026-03-17
Extract pub const NAME for all array encoding structs

jiaqizho:
## Summary

Add a `pub const NAME: &str` to every array encoding struct and derive `pub const ID: ArrayId` from `Self::NAME` instead of an inline string literal. This eliminates duplicated encoding name strings and gives external code a stable constant to reference when matching on encoding IDs.

relative to https://github.com/vortex-data/vortex/pull/6974

## Testing

N/A

### DIFF
--- a/encodings/alp/src/alp/array.rs
+++ b/encodings/alp/src/alp/array.rs
@@ -273,7 +273,8 @@ pub struct ALPArray {
 pub struct ALP;
 
 impl ALP {
-    pub const ID: ArrayId = ArrayId::new_ref("vortex.alp");
+    pub const NAME: &str = "vortex.alp";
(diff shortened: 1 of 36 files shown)


gatesn (changes requested):
I'm not sure I understand the goal here? Why is there duplicate strings in the existing world but not in the new world?

connortsui20:
Was there a reason for this change in relation to #6974?

I think I still personally like this change but the reasoning isn't super clear here, where are the duplicated strings in the current codebase?

jiaqizho:
> Was there a reason for this change in relation to #6974?
> 
> I think I still personally like this change but the reasoning isn't super clear here, where are the duplicated strings in the current codebase?

Yes, after refactoring the implementation from #6974, this change is no longer needed, so closed.

PR #6995, opened 2026-03-17
fix[vortex-array]: fix rebuild_trim_elements overflow

### LINKED ISSUE #6973: vortex-array: rebuild_trim_elements overflows calculating end offset
### What happened?

Seeing this:
```
`offsets + sizes` somehow overflowed:
  Arrow error: Arithmetic overflow: Overflow happened on: 215 + 46
```

It seems to me that instead of/in addition to casting offsets/sizes to the larger width we should consider casting to the the next width? I guess the problem is that the "end" offset is implicit, so it's fine to narrow the start offset of the last range even though the end offset overflows the width.

### Steps to reproduce

I have a vortex file and minimal reproduction test if necessary

### Environment

- Vortex version: 0.59.4 + cherry picks

### Additional context

_No response_

asubiotto:
This commit just casts offsets and sizes to u64/i64 before summing for simplicitly. We could make the logic more complicated but not sure it's worth it.

## Summary

Closes: #6973 

## Testing

Regression test added

### DIFF
--- a/vortex-array/src/arrays/listview/rebuild.rs
+++ b/vortex-array/src/arrays/listview/rebuild.rs
@@ -14,8 +14,10 @@ use crate::arrays::ListViewArray;
 use crate::builders::builder_with_capacity;
 use crate::builtins::ArrayBuiltins;
 use crate::compute;
+use crate::dtype::DType;
 use crate::dtype::IntegerPType;
 use crate::dtype::Nullability;
+use crate::dtype::PType;
 use crate::match_each_integer_ptype;
 use crate::scalar::Scalar;
 use crate::scalar_fn::fns::operators::Operator;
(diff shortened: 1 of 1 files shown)


connortsui20 (approved):
I think this is probably fine given we narrow these in the compressor anyways. If for some reason this is bad (it probably isn't) then we can just do the next size up.

asubiotto:
This shouldn't be relevant for the offsets/sizes width. This code path essentially only computes the "end" offset which is cast to a usize below to slice the elements.

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -294,21 +296,16 @@ impl ListViewArray {
...
             } else {
-                (
-                    self.offsets().cast(self.sizes().dtype().clone())?,
-                    self.sizes().clone(),
-                )
-            };
+                PType::I64
+            });
+            let offsets = self.offsets().cast(wide_dtype.clone())?;
+            let sizes = self.sizes().cast(wide_dtype)?;
```
0ax1, line 308: For types smaller than 64bit this will unconditionally widen => allocate a new buffer + do an element-wise copy.

asubiotto: Yes, this is what I mean by "we could make the logic more complicated". To avoid the unconditional widening we would probably have to attempt the sum and then fall back to widening to the next largest width.

0ax1: Yeah I see, I guess we can act on this in case it shows up in perf profiles.

asubiotto: Definitely! We're always profiling after all 😄

PR #6997, opened 2026-03-17
perf: skip allocation for prim cast if possible

0ax1:
Reinterpret int primitive arrays in-place if byte widths match rather than allocating a new buffer.

### DIFF
--- a/vortex-array/src/arrays/primitive/compute/cast.rs
+++ b/vortex-array/src/arrays/primitive/compute/cast.rs
@@ -13,8 +13,11 @@ use crate::ExecutionCtx;
 use crate::IntoArray;
 use crate::arrays::Primitive;
 use crate::arrays::PrimitiveArray;
+use crate::compute;
 use crate::dtype::DType;
 use crate::dtype::NativePType;
+use crate::dtype::Nullability;
+use crate::dtype::PType;
 use crate::match_each_native_ptype;
 use crate::scalar_fn::fns::cast::CastKernel;
 use crate::vtable::ValidityHelper;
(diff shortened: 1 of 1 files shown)


0ax1:
In my understanding we can't reuse existing buffers for in-place casts rn due to the `CastKernel` signature which takes the array by `array: &PrimitiveArray`. If we'd pass the array by value we could do an in-place cast when there is sufficient capacity.

vortex-array/src/arrays/primitive/compute/cast.rs
```diff
@@ -61,34 +84,35 @@ impl CastKernel for Primitive {
     }
 }
 
+/// Returns `true` if all valid values in `array` are representable as `target_ptype`.
+fn values_fit_in(array: &PrimitiveArray, target_ptype: PType) -> bool {
+    let target_dtype = DType::Primitive(target_ptype, Nullability::NonNullable);
+    compute::min_max(&array.clone().into_array())
```
joseph-isaacs, line 97: This is not free

0ax1: Yeah that's the main q. still worth it given we know?
```
        if array.ptype().is_int()
            && new_ptype.is_int()
            && array.ptype().byte_width() == new_ptype.byte_width()
```

0ax1: If min_max is true, this is a win, otherwise extra cost.

0ax1: As discussed offline: The cast would otherwise bail at a later stage. Therefore, there's no extra cost in doing min_max.

robert3005 (approved):

PR #6999, opened 2026-03-17
Fix scalar partial ordering comparison

connortsui20:
## Summary

When we compare `Scalar`s, we have to be careful that we do not use `<` or `>` if we do not know that the scalars have different types. If they have different types, the result will not panic or raise an error, it will just return `false` for `>`.

This changes the `scalar_cmp` function inside the comparison execution to use the `partial_cmp` method directly, where we map the optional result to an error and raise it.

## Testing

Adds 2 regression tests, the first one is the one I stumbled on and the second is a more targeted one

## Unresolved Questions

- Why is this function public?
- Are there any other places that we do scalar comparison with the built-in `>` operators? If so, we may need to fix them.


vortex-array/src/scalar_fn/fns/binary/compare.rs
```diff
@@ -176,24 +176,43 @@ fn arrow_compare_arrays(
...
+    // Equality and inequality can be determined without `partial_cmp`.
+    match operator {
+        CompareOperator::Eq => return Ok(Scalar::bool(lhs == rhs, nullability)),
+        CompareOperator::NotEq => return Ok(Scalar::bool(lhs != rhs, nullability)),
+        _ => {}
     }
+
+    // We do this instead of `<` and `>` to ensure we do not lose a type mismatch error.
+    let ordering = lhs.partial_cmp(rhs).ok_or_else(|| {
+        vortex_error::vortex_err!(
```
robert3005: import?

vortex-array/src/scalar_fn/fns/binary/compare.rs
```diff
@@ -479,6 +502,49 @@ mod tests {
...
+        let s_scalar = Scalar::extension::<Timestamp>(
+            TimestampOptions {
+                unit: TimeUnit::Seconds,
+                tz: None,
+            },
+            Scalar::from(1704067200i64),
+        );
+
+        // Ordering comparisons must error on incompatible types.
+        assert!(super::scalar_cmp(&ms_scalar, &s_scalar, CompareOperator::Gt).is_err());
```
robert3005: why not import this?

robert3005 (dismissed):
we are missing some imports here

vortex-array/src/scalar_fn/fns/binary/compare.rs
```diff
@@ -176,24 +177,43 @@ fn arrow_compare_arrays(
             CompareOperator::Lte => cmp::lt_eq(&lhs, &rhs)?,
         }
     };
-    from_arrow_array_with_len(&array, left.len(), nullable)
+
+    from_arrow_array_with_len(&arrow_array, left.len(), nullable)
 }
 
-pub fn scalar_cmp(lhs: &Scalar, rhs: &Scalar, operator: CompareOperator) -> Scalar {
+pub fn scalar_cmp(lhs: &Scalar, rhs: &Scalar, operator: CompareOperator) -> VortexResult<Scalar> {
```
joseph-isaacs, line 184: surely we check the dtypes matched above? is that not a better thing to do here?

connortsui20: unfortunately not

connortsui20: actually, I dont even really understand what the pre and post conditions are of execute, it is not well documented on `ScalarFnVTable`

vortex-array/src/scalar_fn/fns/binary/compare.rs
```diff
@@ -176,24 +177,43 @@ fn arrow_compare_arrays(
...
-            CompareOperator::Lt => lhs < rhs,
-            CompareOperator::Lte => lhs <= rhs,
-        };
+        return Ok(Scalar::null(DType::Bool(Nullability::Nullable)));
+    }
+
+    let nullability = lhs.dtype().nullability() | rhs.dtype().nullability();
 
-        Scalar::bool(b, lhs.dtype().nullability() | rhs.dtype().nullability())
+    // Equality and inequality can be determined without `partial_cmp`.
```
joseph-isaacs: why?

connortsui20: nice catch, we should just catch the type error first

joseph-isaacs:
lets run benchmarks first just to check for regressions

connortsui20:
test failure fixed, but that shouldnt stop the benchmarks from running

joseph-isaacs (approved):

PR #7000, opened 2026-03-17
chore: clean up `#[inline]` annotations

0ax1:
Remove `#[inline]` from private, pub(crate), and inner functions where the compiler already has full visibility, and from large (>20 line) functions.

Add `#[inline]` to `BitBuffer` operator trait impls (BitOr, BitAnd, Not, BitXor) - non-generic impls on a concrete type that are otherwise opaque cross-crate with LTO disabled.


connortsui20:
Seems like there are some things that needed to be inlined in here? related to bitpacking and between eval

0ax1:
> Seems like there are some things that needed to be inlined in here? related to bitpacking and between eval

Yeah, I'll double check any regressions for sure.

0ax1:
@connortsui20 I think the current regressions are flaky benchmarks. If you look at the benchmark dashboards you can see how they oscillate back and forth 10% over time.

connortsui20 (approved):
nice

PR #7002, opened 2026-03-17
PatchedArray: basics and wiring

a10y:
## Summary

This is the first PR in a series addressing the PatchedArray RFC: https://github.com/vortex-data/rfcs/pull/27

This PR adds a new `PatchedArray` array variant, which is slated to only be used with FastLanes array types. The design is largely documented in the RFC, but briefly

* Wraps another child array
* Holds a buffer of offsets and patch values. They are both sorted by chunk/lane so there is efficient random as well as bulk access for CUDA

We are able to pushdown the following at reduce time:
* Filters pushes through to the child, and we trim the patch indices to the range of chunks covered by the filter
* Slicing that reduces the chunk range
* Compare always pushes through to the child
* Take pushes through to the inner then constructs a hashmap of the patches

There will be follow ups to add the wiring into CUDA and to update how BitPacked and ALP arrays are written.

## Testing

There are unit tests for all of the reducers and kernels

### DIFF
--- a/vortex-array/src/arrays/patched/array.rs
+++ b/vortex-array/src/arrays/patched/array.rs
@@ -76,6 +76,11 @@ impl PatchedArray {
             "Creating PatchedArray from Patches only supported for primitive arrays"
         );
 
+        vortex_ensure!(
+            patches.num_patches() <= u32::MAX as usize,
+            "PatchedArray does not support > u32::MAX patch values"
+        );
+
         let values_ptype = patches.dtype().as_ptype();
 
         let TransposedPatches {
--- a/vortex-array/src/arrays/patched/array.rs
+++ b/vortex-array/src/arrays/patched/array.rs
@@ -27,8 +27,6 @@ use crate::stats::ArrayStats;
 use crate::validity::Validity;
 
 /// An array that partially "patches" another array with new values.
-///
-/// Patched arrays implement the set of nodes that do this instead here...I think?
 #[derive(Debug, Clone)]
 pub struct PatchedArray {
     /// The inner array that is being patched. This is the zeroth child.
--- a/vortex-array/src/arrays/patched/array.rs
+++ b/vortex-array/src/arrays/patched/array.rs
@@ -216,7 +216,7 @@ fn transpose<I: IntegerPType, V: NativePType>(
         lane_offsets[index] += lane_offsets[index - 1];
     }
 
-    // Loop over patches, writing thme to final positions
+    // Loop over patches, writing them to final positions
     let indices_out = indices_buffer.spare_capacity_mut();
     let values_out = values_buffer.spare_capacity_mut();
     for (index, &value) in std::iter::zip(indices_in, values_in) {
--- a/vortex-array/src/arrays/patched/array.rs
+++ b/vortex-array/src/arrays/patched/array.rs
@@ -14,12 +14,12 @@ use crate::DynArray;
 use crate::ExecutionCtx;
 use crate::IntoArray;
 use crate::arrays::PrimitiveArray;
-use crate::arrays::patched::PatchAccessor;
 use crate::arrays::patched::TransposedPatches;
 use crate::arrays::patched::patch_lanes;
 use crate::buffer::BufferHandle;
 use crate::dtype::IntegerPType;
 use crate::dtype::NativePType;
+use crate::dtype::PType;
 use crate::match_each_native_ptype;
 use crate::match_each_unsigned_integer_ptype;
 use crate::patches::Patches;
@@ -48,7 +48,7 @@ pub struct PatchedArray {
     /// lane offsets. The PType of these MUST be u32
     pub(super) lane_offsets: BufferHandle,
     /// indices within a 1024-element chunk. The PType of these MUST be u16
-    pub(super) indices: BufferHandle,
+    pub(super) indices: ArrayRef,
     /// patch values corresponding to the indices. The ptype is specified by `values_ptype`.
     pub(super) values: ArrayRef,
 
@@ -86,6 +86,12 @@ impl PatchedArray {
             values,
         } = transpose_patches(patches, ctx)?;
 
+        let indices = PrimitiveArray::from_buffer_handle(
+            BufferHandle::new_host(indices),
+            PType::U16,
+            Validity::NonNullable,
+        )
+        .into_array();
         let values = PrimitiveArray::from_buffer_handle(
             BufferHandle::new_host(values),
             values_ptype,
@@ -102,19 +108,24 @@ impl PatchedArray {
             offset: 0,
             len,
             lane_offsets: BufferHandle::new_host(lane_offsets),
-            indices: BufferHandle::new_host(indices),
+            indices,
             values,
             stats_set: ArrayStats::default(),
         })
     }
+}
 
-    /// Get an accessor, which allows ranged access to patches by chunk/lane.
-    pub fn accessor(&self) -> PatchAccessor<'_> {
-        PatchAccessor {
-            n_lanes: self.n_lanes,
-            lane_offsets: self.lane_offsets.as_host().reinterpret::<u32>(),
-            indices: self.indices.as_host().reinterpret::<u16>(),
-        }
+impl PatchedArray {
+    pub(crate) fn seek_to_lane(&self, chunk: usize, lane: usize) -> Range<usize> {
+        assert!(chunk < self.n_chunks);
+        assert!(lane < self.n_lanes);
+
+        let lane_offsets = self.lane_offsets.as_host().reinterpret::<u32>();
+
+        let start = lane_offsets[chunk * self.n_lanes + lane] as usize;
+        let stop = lane_offsets[chunk * self.n_lanes + lane + 1] as usize;
+
+        start..stop
     }
 
     /// Slice the array to just the patches and inner values that are within the chunk range.
--- a/vortex-array/src/arrays/patched/array.rs
+++ b/vortex-array/src/arrays/patched/array.rs
@@ -116,7 +116,13 @@ impl PatchedArray {
 }
 
 impl PatchedArray {
-    pub(crate) fn seek_to_lane(&self, chunk: usize, lane: usize) -> Range<usize> {
+    /// Get a range of indices that can be used to access the `indices` and `values` children
+    /// to retrieve all patches for a specified lane.
+    ///
+    /// # Panics
+    ///
+    /// Note that this function will panic if the caller requests out of bounds chunk/lane ordinals.
+    pub(crate) fn lane_range(&self, chunk: usize, lane: usize) -> Range<usize> {
         assert!(chunk < self.n_chunks);
         assert!(lane < self.n_lanes);
 
--- a/vortex-array/src/arrays/patched/array.rs
+++ b/vortex-array/src/arrays/patched/array.rs
@@ -12,17 +12,19 @@ use crate::ArrayRef;
 use crate::Canonical;
 use crate::DynArray;
 use crate::ExecutionCtx;
+use crate::IntoArray;
+use crate::arrays::PrimitiveArray;
 use crate::arrays::patched::PatchAccessor;
 use crate::arrays::patched::TransposedPatches;
 use crate::arrays::patched::patch_lanes;
 use crate::buffer::BufferHandle;
 use crate::dtype::IntegerPType;
 use crate::dtype::NativePType;
-use crate::dtype::PType;
 use crate::match_each_native_ptype;
 use crate::match_each_unsigned_integer_ptype;
 use crate::patches::Patches;
 use crate::stats::ArrayStats;
+use crate::validity::Validity;
 
 /// An array that partially "patches" another array with new values.
 ///
@@ -50,14 +52,17 @@ pub struct PatchedArray {
     /// indices within a 1024-element chunk. The PType of these MUST be u16
     pub(super) indices: BufferHandle,
     /// patch values corresponding to the indices. The ptype is specified by `values_ptype`.
-    pub(super) values: BufferHandle,
-    /// PType of the scalars in `values`. Can be any native type.
-    pub(super) values_ptype: PType,
+    pub(super) values: ArrayRef,
 
     pub(super) stats_set: ArrayStats,
 }
 
 impl PatchedArray {
+    /// Create a new `PatchedArray` from a child array and a set of [`Patches`].
+    ///
+    /// # Errors
+    ///
+    /// The `inner` array must be primitive type, and it must have the same DType as the patches.
     pub fn from_array_and_patches(
         inner: ArrayRef,
         patches: &Patches,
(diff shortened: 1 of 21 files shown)


vortex-array/src/arrays/patched/array.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub(super) offset: usize,
+    /// Total length.
+    pub(super) len: usize,
+
+    /// lane offsets. The PType of these MUST be u32
+    pub(super) lane_offsets: BufferHandle,
+    /// indices within a 1024-element chunk. The PType of these MUST be u16
+    pub(super) indices: BufferHandle,
+    /// patch values corresponding to the indices. The ptype is specified by `values_ptype`.
+    pub(super) values: BufferHandle,
```
joseph-isaacs: I really think we should reconsider this.

This array (+ constant) should be used to replace the sparse array so I think we really want to store things other than PTypes?

vortex-array/src/arrays/patched/compute/filter.rs
```diff
@@ -0,0 +1,145 @@
...
+                (first / 1024, last.div_ceil(1024))
+            }
+        };
+
+        // If all chunks already covered, there is nothing to do.
+        if chunk_start == 0 && chunk_stop == array.n_chunks {
+            return Ok(None);
+        }
+
+        let sliced = array.slice_chunks(chunk_start..chunk_stop)?;
```
joseph-isaacs, line 46: you cannot do this

a10y: why? this doesn't access buffers just slices

vortex-array/src/arrays/patched/array.rs
a10y: I'm not sure if this array is well-suited for non-primitive types. How many lanes does a `List<Struct{u32,utf8}>` have? I was tempted to actually put this into the vortex-fastlanes crate tbh

a10y: i've updated to use child array for values, which should give us some flexibility to change things in the future

vortex-array/src/arrays/patched/compute/filter.rs
```diff
@@ -0,0 +1,145 @@
...
+            AllOr::All | AllOr::None => {
+                // This is handled as the precondition to this method, see the FilterReduce
+                // documentation.
+                unreachable!("mask must be a MaskValues here")
+            }
+            AllOr::Some(indices) => {
+                let first = indices[0];
+                let last = indices[indices.len() - 1];
+
+                (first / 1024, last.div_ceil(1024))
```
joseph-isaacs: You really want to take(1) and on the slices iterator and backwards slice iterator here

vortex-array/src/arrays/patched/array.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    /// Offset into the first chunk
+    pub(super) offset: usize,
+    /// Total length.
+    pub(super) len: usize,
+
+    /// lane offsets. The PType of these MUST be u32
+    pub(super) lane_offsets: BufferHandle,
+    /// indices within a 1024-element chunk. The PType of these MUST be u16
+    pub(super) indices: BufferHandle,
```
joseph-isaacs: my feeling is that we want this to be an array too in-case we later do want to compress this array.

a10y: the issue is if i make this an array then i'm not sure how to implement scalar_at for PatchedArray

a10y: ok i worked around this

a10y:
@claude review please

a10y:
@claude i have addressed your main comments. please give another review.

a10y:
@claude please review

a10y:
@claude PR is ready for another review

a10y:
Docker launches seem broken right now and seems like possibly an upstream thing

vortex-buffer/src/buffer.rs
```diff
@@ -523,6 +523,34 @@ impl<T> Buffer<T> {
...
+            "ByteBuffer length not a multiple of the value length"
+        );
+
+        let v_len = self.length / size_of::<V>();
+        let v_ptr = self.bytes.as_ptr().cast::<V>();
+
+        // SAFETY: we checked that alignment and length are suitable to treat this as a &[V].
+        unsafe { std::slice::from_raw_parts(v_ptr, v_len) }
+    }
+}
```
joseph-isaacs: did we never do this before?

vortex-test/compat-gen/src/fixtures/arrays/synthetic/encodings/patched.rs
joseph-isaacs, line 1: shall we add a few more of these with different types? e.g. string?

vortex-array/src/arrays/patched/array.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// values that might be patched.
+    pub(super) n_lanes: usize,
+
+    /// Offset into the first chunk
+    pub(super) offset: usize,
+    /// Total length.
+    pub(super) len: usize,
+
+    /// lane offsets. The PType of these MUST be u32
+    pub(super) lane_offsets: BufferHandle,
```
joseph-isaacs: are you sure we don't want to have the option to compress these?

vortex-array/src/arrays/patched/array.rs
```diff
@@ -0,0 +1,290 @@
...
+use crate::validity::Validity;
+
+/// An array that partially "patches" another array with new values.
+#[derive(Debug, Clone)]
+pub struct PatchedArray {
+    /// The inner array that is being patched. This is the zeroth child.
+    pub(super) inner: ArrayRef,
+
+    /// Number of 1024-element chunks. Pre-computed for convenience.
+    pub(super) n_chunks: usize,
```
joseph-isaacs: why store this?

joseph-isaacs: this will increase the stack size and maybe register pressure do avoid / 1024?

vortex-array/src/arrays/patched/array.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    /// Number of lanes the patch indices and values have been split into. Each of the `n_chunks`
+    /// of 1024 values is split into `n_lanes` lanes horizontally, each lane having 1024 / n_lanes
+    /// values that might be patched.
+    pub(super) n_lanes: usize,
```
joseph-isaacs, line 116: this too

vortex-array/src/arrays/patched/array.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    /// Offset into the first chunk
```
joseph-isaacs: the inner or 1k patches?

vortex-array/src/arrays/patched/array.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub(super) offset: usize,
+    /// Total length.
```
joseph-isaacs: len of what?

vortex-array/src/arrays/patched/vtable/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+    fn child_name(_array: &Self::Array, idx: usize) -> String {
+        match idx {
+            0 => "inner".to_string(),
+            1 => "patch_indices".to_string(),
+            2 => "patch_values".to_string(),
+            _ => vortex_panic!("invalid child index for PatchedArray: {idx}"),
+        }
+    }
+
+    #[allow(clippy::cast_possible_truncation)]
```
joseph-isaacs: please can we not use this

vortex-array/src/arrays/patched/vtable/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+    /// This may not match the length of the wrapping PatchedArray, if for example
+    /// in a filter or slice it may be sliced to the nearest chunk boundary.
+    #[prost(uint64, tag = "1")]
+    pub(crate) inner_len: u64,
+
+    /// Offset within the first chunk of `inner` where data begins.
+    ///
+    /// This may become nonzero after slicing.
+    #[prost(uint32, tag = "2")]
+    pub(crate) offset: u32,
```
joseph-isaacs: does inner have chunks?

vortex-array/src/arrays/patched/vtable/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+pub struct Patched;
+
+impl ValidityChild<Patched> for Patched {
+    fn validity_child(array: &PatchedArray) -> &ArrayRef {
+        &array.inner
+    }
+}
+
+#[derive(Clone, prost::Message)]
+pub struct PatchedMetadata {
```
joseph-isaacs, line 67: I wonder if you can pack this into a u128?

vortex-array/src/arrays/patched/vtable/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+    /// Length of the `inner` child.
+    ///
+    /// This may not match the length of the wrapping PatchedArray, if for example
+    /// in a filter or slice it may be sliced to the nearest chunk boundary.
+    #[prost(uint64, tag = "1")]
+    pub(crate) inner_len: u64,
+
```
joseph-isaacs: This is used only to fetch

vortex-array/src/arrays/patched/vtable/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+        vortex_ensure!(
+            children.len() == 3,
+            "PatchedArray must have exactly 3 children"
+        );
+
+        array.inner = children.remove(0);
+        array.indices = children.remove(0);
+        array.values = children.remove(0);
+
+        Ok(())
```
joseph-isaacs: on debug check the dtypes?

vortex-buffer/src/buffer.rs
a10y: I coudln't find anything no

vortex-test/compat-gen/src/fixtures/arrays/synthetic/encodings/patched.rs
a10y: Executing anything other than primitive would fail

vortex-array/src/arrays/patched/array.rs
a10y: This I think we need to store incase we change the default ptype -> lane mapping in the future

vortex-array/src/arrays/patched/vtable/mod.rs
a10y: I think I can eliminate this now that I've fixed slicing

vortex-array/src/arrays/patched/array.rs
robert3005: I think this kind of change would require new arrays. The data layout would change

vortex-array/src/arrays/patched/vtable/mod.rs
a10y: what do you mean by this

vortex-array/src/arrays/patched/vtable/mod.rs
a10y: hm, is there a reason to do that over struct? can protobuf even encode u128?

vortex-array/src/arrays/patched/vtable/mod.rs
a10y: removed in favor of fallible case + documented `as` casts

vortex-array/src/arrays/patched/array.rs
a10y: the array. this is analagous to len in BitPackedArray rn

joseph-isaacs: can you add a more detailed doc str

vortex-array/src/arrays/patched/vtable/mod.rs
joseph-isaacs: I am asking why we store this, and its likely cause the inner may have a different len to the result patch array?

vortex-array/src/arrays/patched/vtable/mod.rs
joseph-isaacs: I mean using less and smaller fields

a10y:
@claude review please

vortex-array/src/arrays/patched/vtable/mod.rs
a10y: i've now removed it so we don't need to store it anymore. inner should now always have the same len as self

vortex-array/src/arrays/patched/vtable/mod.rs
a10y: I removed some fields and packed the rest into a single uint64 packed field

vortex-array/src/arrays/patched/array.rs
a10y: As long as you have the n_lanes saved you can always access the data without issue

vortex-array/src/arrays/patched/array.rs
a10y: removed

vortex-array/src/arrays/patched/array.rs
a10y: updated to store as child

a10y:
I'm pretty sure the CodSpeed results are just noise since the PatchedArray isn't wired into anything here

a10y:
@claude please review

vortex-buffer/src/buffer.rs
a10y: this is no longer used so i removed it

a10y:
@joseph-isaacs / @robert3005 when you get a chance I've addressed or responded to all your latest batch of comments

vortex-array/src/arrays/patched/vtable/mod.rs
joseph-isaacs: Sorry @a10y this was on me. I was thinking that we can have 4 fields that total u128 in size. I don't think we need to do this manual packing

a10y: reverted this the fields are now separated

joseph-isaacs (approved):

joseph-isaacs:
Make sure when we wire this in we make it an unstable encoding

vortex-file/src/strategy.rs
```diff
@@ -104,6 +105,7 @@ pub static ALLOWED_ENCODINGS: LazyLock<ArrayRegistry> = LazyLock::new(|| {
     session.register(Delta);
     session.register(FoR);
     session.register(FSST);
+    session.register(Patched);
```
joseph-isaacs: Remove this

joseph-isaacs (changes requested):

vortex-file/src/strategy.rs
joseph-isaacs: We should keep the old patches for a while and have this as an unstable_encoding

joseph-isaacs:
You will also need to replace with children (with with_slots)

a10y:
done

joseph-isaacs (approved):

PR #7005, opened 2026-03-18
Check for overflow in decimal sum

gatesn:


vortex-array/src/aggregate_fn/fns/sum/decimal.rs
```diff
@@ -14,7 +14,7 @@ use crate::scalar::DecimalValue;
 /// Accumulate a decimal array into the sum state.
 /// Returns Ok(true) if saturated (overflow), Ok(false) if not.
 pub(super) fn accumulate_decimal(inner: &mut SumState, d: &DecimalArray) -> VortexResult<bool> {
-    let SumState::Decimal(acc) = inner else {
+    let SumState::Decimal { value, dtype } = inner else {
```
robert3005, line 17: value.decimal_dtype() is the dtype you want

robert3005: it is not

robert3005 (approved):

AdamGS:
Does this PR fix https://github.com/vortex-data/vortex/issues/7004?

PR #7006, opened 2026-03-18
MinMax AggregateFn

gatesn:
Port min-max to an AggregateFn.

Note that we could push-down over dictionary arrays with the given property (similarly for RunEnd):

```python
def is_support_invariant(agg_func) -> bool:
    """
    Returns True if this aggregate function depends only on the set of
    distinct values in its input, not on how many times each appears.

    Formally, for any multiset S:

        agg(S) == agg(set(S))

    This is the property that allows the aggregate to be evaluated purely
    over dictionary values, skipping the codes column entirely. Since a
    dictionary already contains exactly the distinct values present, no
    codes scan is needed.

    Examples:
        is_support_invariant(min)           # True
        is_support_invariant(max)           # True
        is_support_invariant(count_distinct) # True
        is_support_invariant(bool_or)       # True
        is_support_invariant(bool_and)      # True
        is_support_invariant(any_value)     # True
        is_support_invariant(median)        # False — needs frequencies
        is_support_invariant(mode)          # False — entirely about frequencies
        is_support_invariant(sum)           # False
        is_support_invariant(count)         # False
```


vortex-array/src/aggregate_fn/fns/min_max/varbin.rs
joseph-isaacs, line 1: do we truncate this? Should that be a min/max option?

joseph-isaacs:
`is_support_invariant` just seems like you should use a reduce rule here?

joseph-isaacs:
I think this sounds like a good property `is_support_invariant`.

>     This is the property that allows the aggregate to be evaluated purely
    over dictionary values, skipping the codes column entirely. Since a
    dictionary already contains exactly the distinct values present, no
    codes scan is needed.
    
    
    
This is untrue its only over defined/reachable values, you cannot ignore codes.

vortex-array/src/aggregate_fn/fns/min_max/varbin.rs
gatesn: We could in the future, the old function didn't do it so presumably it lives somewhere else in stats computation?

gatesn:
I guess we could have an "approximate: bool" flag in the options? Or somehow? Like is it exact min/max, or is it lower/upper bound?

joseph-isaacs:
You can but a reachable bit-set from the codes, we have that logic. I would think we do that

gatesn:
We already do that in this PR btw

robert3005 (approved):

joseph-isaacs:
How was this tested?

PR #7008, opened 2026-03-18
Decimal sum uses smallest physical type that supports output dtype

robert3005:
Instead of always casting values to i256 we take the primitive value of
the return dtype and perform operations in that space

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/sum/decimal.rs
+++ b/vortex-array/src/aggregate_fn/fns/sum/decimal.rs
@@ -1,10 +1,18 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use crate::dtype::DecimalDType;
+use crate::dtype::DecimalType;
 use itertools::Itertools;
+use num_traits::AsPrimitive;
+use num_traits::CheckedAdd;
+use num_traits::NumOps;
+use vortex_buffer::BitBuffer;
+use vortex_buffer::Buffer;
+use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_error::vortex_panic;
-use vortex_mask::AllOr;
+use vortex_mask::Mask;
 
 use super::SumState;
 use crate::arrays::DecimalArray;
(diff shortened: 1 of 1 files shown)


vortex-array/src/aggregate_fn/fns/sum/decimal.rs
```diff
@@ -13,46 +20,72 @@ use crate::scalar::DecimalValue;
...
+            let decimal_sum = sum
+                .map(DecimalValue::from)
+                // We have to make sure that the decimal value fits the precision of the decimal dtype.
+                .filter(|v| v.fits_in_precision(*dtype));
+
+            match decimal_sum {
+                Some(v) => *value = v,
+                None => return Ok(true),
             }
             Ok(false)
```
joseph-isaacs: Can you pull this out into a fn, I do think the clippy lint is helpful here

vortex-array/src/aggregate_fn/fns/sum/decimal.rs
```diff
@@ -13,46 +20,72 @@ use crate::scalar::DecimalValue;
 
 /// Accumulate a decimal array into the sum state.
 /// Returns Ok(true) if saturated (overflow), Ok(false) if not.
+#[allow(clippy::cognitive_complexity)]
```
joseph-isaacs:
```suggestion
#[expect(clippy::cognitive_complexity)]
```

vortex-array/src/aggregate_fn/fns/sum/decimal.rs
```diff
@@ -13,46 +20,72 @@ use crate::scalar::DecimalValue;
...
-                    None => return Ok(true),
-                }
-            }
-            Ok(false)
-        }),
-        AllOr::Some(validity) => match_each_decimal_value_type!(d.values_type(), |T| {
-            for (&v, valid) in d.buffer::<T>().iter().zip_eq(validity.iter()) {
-                if valid {
-                    match value.checked_add(&DecimalValue::from(v)) {
-                        Some(r) => {
```
joseph-isaacs, line 43: where did this upcast?

vortex-array/src/aggregate_fn/fns/sum/decimal.rs
```diff
@@ -13,46 +20,72 @@ use crate::scalar::DecimalValue;
...
+    bool: AsPrimitive<I>,
+{
+    let mut sum = initial;
+    for (v, valid) in values.iter().zip_eq(validity) {
+        let v: I = v.as_() * valid.as_();
+
+        sum = CheckedAdd::checked_add(&sum, &v)?;
     }
+    Some(sum)
 }
```
joseph-isaacs, line 101: I wonder if this reduces ILP has to check if the add is allowed before trying the next

vortex-array/src/aggregate_fn/fns/sum/decimal.rs
robert3005: `checked_add` upcasts to i256

joseph-isaacs: Can we remove that function, this will happen again and again!

joseph-isaacs (approved):

PR #7009, opened 2026-03-18
Sum aggregate doesn't include nan values

### LINKED ISSUE #5152: `sum` statistic should not include `NaN`s

robert3005:
NaN will cause sum to be NaN. Instead we have NaNCount already that users can
use to handle columns with NaNs.

fix #5152 

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/sum/primitive.rs
+++ b/vortex-array/src/aggregate_fn/fns/sum/primitive.rs
@@ -54,7 +54,9 @@ fn accumulate_primitive_all(inner: &mut SumState, p: &PrimitiveArray) -> VortexR
             signed: |_T| { vortex_panic!("float sum state with signed input") },
             floating: |T| {
                 for &v in p.as_slice::<T>() {
-                    *acc += ToPrimitive::to_f64(&v).vortex_expect("float to f64");
+                    if !v.is_nan() {
+                        *acc += ToPrimitive::to_f64(&v).vortex_expect("float to f64");
+                    }
                 }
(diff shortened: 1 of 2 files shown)


vortex-array/src/aggregate_fn/fns/sum/primitive.rs
```diff
@@ -98,7 +100,7 @@ fn accumulate_primitive_valid(
             signed: |_T| { vortex_panic!("float sum state with signed input") },
             floating: |T| {
                 for (&v, valid) in p.as_slice::<T>().iter().zip_eq(validity.iter()) {
-                    if valid {
+                    if valid && !v.is_nan() {
```
joseph-isaacs, line 103: what about other special values should we have those?

robert3005: I think Infininty makes sense, while NaN does not, i.e. NaN is an indication something went wrong in computation but Infinity can be a real value

joseph-isaacs: That is what I would think max would tell us, but I am not sure really?

robert3005: We filter out NaN but not infinities. Need to update the saturation behaviour

gatesn:
You should also make the is_saturated check test for nans since `+Inf + -Inf = NaN`, in which case you're saturating anyways

gatesn (approved):

PR #7011, opened 2026-03-18
fix[vortex-array]: handle lazily sliced REE to_arrow

asubiotto:
## Summary

This previous code was short-circuiting with a zero offset, with the assumption that ends were already correctly physically sliced. However, slicing is a lazy operation using metadata, so this commit fixes a couple of edge cases by executing the fast path only if the ends are correcly sliced. Otherwise, ends are trimmed and transformed with the offset in for correct arrow array construction.

## Testing

Enhanced tests to reproduce e2e arrow execution error and added some unit tests to build_run_array

### DIFF
--- a/vortex-array/src/arrow/executor/run_end.rs
+++ b/vortex-array/src/arrow/executor/run_end.rs
@@ -118,22 +118,35 @@ fn build_run_array<R: RunEndIndexType>(
 where
     R::Native: std::ops::Sub<Output = R::Native> + Ord,
 {
-    if offset == 0 {
-        return Ok(
-            Arc::new(RunArray::<R>::try_new(ends.as_primitive::<R>(), values)?) as ArrowArrayRef,
-        );
-    }
-
     let offset_native = R::Native::from_usize(offset)
         .ok_or_else(|| vortex_err!("Offset {offset} exceeds run-end index capacity"))?;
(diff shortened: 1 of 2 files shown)


vortex-array/src/arrow/executor/run_end.rs
```diff
@@ -118,22 +118,35 @@ fn build_run_array<R: RunEndIndexType>(
...
+
+    // Trim to only include runs covering the [offset, offset+length) range.
+    // Runs beyond this would produce duplicate adjusted ends, violating
+    // Arrow's strict-ordering requirement for RunArray.
+    let num_runs = ends_prim
+        .values()
+        .iter()
+        .position(|&e| e - offset_native >= length_native)
+        .map(|i| i + 1)
+        .unwrap_or(ends_prim.len());
```
joseph-isaacs: can we binary search for this?

joseph-isaacs: maybe `partition_point`?

asubiotto: Done, thanks

asubiotto:
Updated to binary search.

joseph-isaacs (approved):
LG, thanks

PR #7012, opened 2026-03-18
Use duckdb allocator for vortex-duckdb

myrrc:
DB extensions are encouraged to use host DB's allocation routines.
Do that for duckdb.

Overriding a global allocator is generally a bad idea, a better idea would
be overriding allocator only for stuff around duckdb intergration around a
memory pool

### DIFF
--- a/vortex-duckdb/include/vortex.h
+++ b/vortex-duckdb/include/vortex.h
@@ -14,6 +14,10 @@ extern "C" {
 #include "duckdb.h"
 
 
+extern void *duckdb_malloc(uintptr_t size);
+
+extern void duckdb_free(void *ptr);
+
 /**
  * Global symbol visibility in the Vortex extension:
  * - Rust functions use C ABI with "_rust" suffix (e.g., vortex_init_rust)
--- a/vortex-duckdb/src/lib.rs
+++ b/vortex-duckdb/src/lib.rs
@@ -39,6 +39,35 @@ mod copy;
 #[cfg(test)]
 mod e2e_test;
 
+use core::alloc::{GlobalAlloc, Layout};
+use core::ffi::c_void;
+use core::ptr;
+
+unsafe extern "C-unwind" {
+    fn duckdb_malloc(size: usize) -> *mut c_void;
+    fn duckdb_free(ptr: *mut c_void);
+}
+
+struct DuckAllocator;
+
+unsafe impl GlobalAlloc for DuckAllocator {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        let ptr = unsafe { duckdb_malloc(layout.size()) };
+        if ptr.is_null() {
+            ptr::null_mut()
+        } else {
+            ptr as *mut u8
+        }
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        unsafe { duckdb_free(ptr as *mut c_void) }
+    }
+}
+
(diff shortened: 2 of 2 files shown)


vortex-duckdb/src/lib.rs
```diff
@@ -39,6 +39,35 @@ mod copy;
...
+        }
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, _layout: Layout) {
+        unsafe { duckdb_free(ptr as *mut c_void) }
+    }
+}
+
+#[global_allocator]
+static GLOBAL: DuckAllocator = DuckAllocator;
```
0ax1, line 69: `DuckAllocator` _allocate me 2 more ducks please 🙂_

AdamGS, line 69: I think this might break the benchmarks, we currently set mimalloc somewhere, we need to make sure its in the leaf crates and not in `vortex-bench`

AdamGS: might also have weird consequences on the sqllogictest crate

vortex-duckdb/include/vortex.h
```diff
@@ -14,6 +14,10 @@ extern "C" {
 #include "duckdb.h"
 
 
+extern void *duckdb_malloc(uintptr_t size);
```
0ax1, line 17: Can't edit this file manually, its bindgen generated. `duckdb_free` and `duckdb_malloc` should come from the duckdb c header and bindgen generated for the rust end..

vortex-duckdb/src/lib.rs
0ax1: Yeah, in general we shouldn't set the allocator in libraries but only in executables right? That said, hooking into the duckdb allocator seems reasonable. Good to keep in mind that duckdb to my recollection uses diff allocators by default between macos and linux. For linux, jemalloc should be the default. We should in any case benchmark the perf difference. My guess here would be though that mimalloc leads to better results.

0ax1: https://doc.rust-lang.org/std/alloc/index.html

> The #[global_allocator] can only be used once in a crate or its recursive dependencies.

AdamGS: that's the thing, we use vortex-duckdb as library for tests and benchmarks, maybe this should be behind some default feature so tests and benchmarks can opt-in/out

AdamGS: I think it makes perfect sense to use this for the duckdb benchmarks and this is also why its now a stand-alone binary, just need to make sure this doesn't break the benchmarks as they are now.

gatesn:
I agree with using their allocator for buffers, but I'm not sure we need to use it for all objects do we?

AdamGS:
should we make buffers generic over allocators like `Vec` is?

joseph-isaacs:
Shall we move it to a discussion?

0ax1:
> I agree with using their allocator for buffers, but I'm not sure we need to use it for all objects do we?

For context, rn `duckdb_malloc` just calls C `malloc`, nothing else so we're not losing out on anything by not calling `duckdb_malloc`: 
```
void *duckdb_malloc(size_t size) {
	return malloc(size);
}
```

https://github.com/duckdb/duckdb/blob/32ab504f62893f1ed3a9e9289204bd4fdd346325/src/main/capi/helper-c.cpp#L490 

cc @gatesn

PR #7014, opened 2026-03-18
Add backwards compatibly handling of sum stats of extension types

AdamGS:
After #6910, Older files with extension types will have zone stats that are just based on the storage type (which might be arbitrary) and which is rarely useful. we currently don't handle its existence at all, so reading these files might panic or otherwise error. This PR falls back to the storage type on the read path, but might not be the desirable solution.


gatesn (approved):
Sadly this is probably correct.

Could you expand on the comment though? Explain that we used to arbitrarily sum the storage type, but this isn't valid for all extension types (e.g. timestamps).

PR #7018, opened 2026-03-18
Pluggable Compressor

connortsui20:
## Summary: Extensible and Pluggable Compressor

Tracking Issue: https://github.com/vortex-data/vortex/issues/7216

You can see a lot of the details in the tracking issue.

This is a major step in supporting extension types as a first-class feature in Vortex. The entire compressor has been rewritten, see the tracking issue for full design details and motivation.

The new `vortex-compressor` crate extracts the encoding-agnostic compression framework from `vortex-btrblocks`, inverting the dependency graph so that encoding crates can implement a single `Scheme` trait and register themselves with the compressor. Additionally, `vortex-btrblocks` remains the "batteries-included" default compressor, and depends on `vortex-compressor`.

---

The compression benchmark comment is [here](https://github.com/vortex-data/vortex/pull/7018#issuecomment-4083359849).

For reviewers: I would just look at the whole `vortex-compressor` and `vortex-btrblocks` crates instead of the git diff since basically everything has changed.

## Changes

- [x] Extract `vortex-compressor` crate with unified `Scheme` trait
- [x] Migrate `vortex-btrblocks` to depend on `vortex-compressor`
- [ ] Verify what APIs we want to preserve from `vortex-btrblocks` (re-exports)
- [ ] Figure out why the `RunEndScheme` exclusion in `rle.rs` is broken and re-enable it

## Testing

Existing tests pass, so that's a good sign. I added a few new tests that check the newer parts of the compressor as well.


joseph-isaacs:
Please can you make a PR that moves the files in one go and other that does real changes

connortsui20:
@joseph-isaacs It's not super clear that that is possible...

Even though things have been pulled out of `vortex-btrblocks`, everything that was pulled out has been changed anyways, so I don't think that really helps review.

Edit: We will keep this PR open but also have parallel PRs that will pull out just the new `vortex-compressor` crate, and then we can slowly move things over from there. That's less about reviewing (that will likely make it harder to review) and more about making sure we don't break things in weird ways.

Edit: Split out into https://github.com/vortex-data/vortex/pull/7104

Edit: That ^ was a terrible idea

gatesn (commented):
I like the direction a lot.

But I think we should follow more closely the conventions we already have for plugins, i.e. a vtable for `Scheme` with all the associated machinery and associated type for options.

I'd also like to think about how we can define compression graphs, or "pipelines" for compressing arrays of certain types.

It might also be worth replacing stats with AggregateFnRef. We're going to do this in the main vortex-array crate soon, and we may as well start doing stats this way within the compressor if we're going to change it a lot.

connortsui20:
If we want to do that, we should do that in a followup PR after we merge this, as the `Scheme` in this PR is much closer to the existing `Scheme` we have on `develop`.

So we will go through the process of merging this and then we can look at extending `Scheme` to be closer to the rest of our `VTable`s.

vortex-array/src/executor.rs
```diff
@@ -190,6 +190,7 @@ impl dyn DynArray + '_ {
 ///
 /// Accumulates a trace of execution steps. Individual steps are logged at TRACE level for
 /// real-time following, and the full trace is dumped at DEBUG level when the context is dropped.
+#[derive(Debug, Clone)]
 pub struct ExecutionCtx {
```
connortsui20, line 194: Is this ok? @joseph-isaacs @gatesn

a10y:
@claude please review

connortsui20:
@a10y Fixed some issues in the latest commit, the things that are actually still wrong are detailed in the (updated) PR description. Everything else that claude said is not an actual concern.

a10y (approved):
let's get this in and iterate

connortsui20:
I am checking the compression size regressions locally. I'm not worrying as much about the throughput regressions because I sort of believe we should have a separate compressor that optimizes for that instead, and also the regressions are quite small.

PR #7025, opened 2026-03-18
skip: clean up ci prebuild image usage

joseph-isaacs:
Can remove the -v2 prefix now


0ax1 (approved):
confused about the pr desc: "can remove v2" but what the pr does it adding it.

PR #7026, opened 2026-03-18
Canonical variant array

AdamGS:
This PR introduces a minimal new canonical array for the new variant DType, its not very useable at the moment but I figured its a piece of mostly boilerplate code that is worth splitting out of the bigger branch.


vortex-array/src/arrays/variant/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+impl VariantArray {
+    /// Creates a new non-nullable VariantArray wrapping the given child.
+    pub fn new(child: ArrayRef) -> Self {
+        Self::new_nullable(child, Nullability::NonNullable)
+    }
+
+    /// Creates a new VariantArray with the given nullability.
+    pub fn new_nullable(child: ArrayRef, nullability: Nullability) -> Self {
```
connortsui20: Do we know that `VariantArray` is going to have a child `ArrayRef` or is that just so you can implement some of the required methods?

Also, we probably just want `new` to take a `Nullablity` directly instead of having a `new_nullable` to be consistent with the other nested arrays.

AdamGS: The general idea is that the variant array has a single child which should be a more concrete encoding of the "variant" family.

AdamGS: also fixed the `new` thing

vortex-array/src/arrays/variant/vtable/mod.rs
```diff
@@ -59,7 +59,7 @@ impl VTable for Variant {
     }
 
     fn stats(array: &Self::Array) -> StatsSetRef<'_> {
-        array.stats_set.to_ref(array.as_ref())
+        array.child.statistics()
```
AdamGS, line 62: I think this is the right thing here, might change as I go deeper into the specific encoding

connortsui20 (approved):

PR #7063, opened 2026-03-19
Compress int dict array values

connortsui20:
## Summary

Compress dictionary-encoded integer array values.

Note that we already do this for dictionary-encoded float array values.

## Testing

N/A

### DIFF
--- a/vortex-btrblocks/src/compressor/integer/mod.rs
+++ b/vortex-btrblocks/src/compressor/integer/mod.rs
@@ -703,21 +703,32 @@ impl Scheme for DictScheme {
 
         let dict = dictionary_encode(stats);
 
-        // Cascade the codes child
-        // Don't allow SequenceArray as the codes child as it merely adds extra indirection without actually compressing data.
-        let mut new_excludes = vec![IntCode::Dict, IntCode::Sequence];
-        new_excludes.extend_from_slice(excludes);
+        // Cascade the codes child.
+        // Don't allow SequenceArray as the codes child as it merely adds extra indirection without
+        // actually compressing data.
+        let mut codes_excludes = vec![IntCode::Dict, IntCode::Sequence];
+        codes_excludes.extend_from_slice(excludes);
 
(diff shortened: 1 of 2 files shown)


vortex-btrblocks/src/compressor/integer/mod.rs
```diff
@@ -703,21 +703,32 @@ impl Scheme for DictScheme {
...
 
-        // Cascade the codes child
-        // Don't allow SequenceArray as the codes child as it merely adds extra indirection without actually compressing data.
-        let mut new_excludes = vec![IntCode::Dict, IntCode::Sequence];
-        new_excludes.extend_from_slice(excludes);
+        // Cascade the codes child.
+        // Don't allow SequenceArray as the codes child as it merely adds extra indirection without
+        // actually compressing data.
+        let mut codes_excludes = vec![IntCode::Dict, IntCode::Sequence];
+        codes_excludes.extend_from_slice(excludes);
```
joseph-isaacs, line 710: what goes wrong with sequence array?

connortsui20: we want to have duplicates in dictionary codes by definition which means the codes can never be a sequence array

joseph-isaacs: that makes sense, is that documented?

connortsui20: the comment right above?

connortsui20: Note that Im working on something that will make this kind of reasoning a lot more obvious, if you are interested: https://github.com/vortex-data/vortex/pull/7018

connortsui20:
whoops just meant to make this a draft

robert3005:
I think you're unlikely to find practical improvements here since the writer already limits dictionaries to 65k unique values, i.e. u16. Not sure what's the formula for best case improvement here but it likely won't be a lot

PR #7065, opened 2026-03-19
Print to stdout in `compress-bench` and `random-access-bench`

connortsui20:
## Summary

For these 2, if you don't specify an output file it doesn't print to stdout. This change fixes that (brings it in line with the other benchmarks).

## Testing

N/A


AdamGS (approved):
🚢

PR #7068, opened 2026-03-19
Interned Identifiers

### LINKED ISSUE #6823: Use string interning for all ID types
All ids should go via a global static string interner in vortex-utils. This will allow constant-time comparisons.

We want global static instead of per-session since in theory arrays can be moved between sessions right now.

gatesn:
Fixes #6823


encodings/alp/src/alp/array.rs
```diff
@@ -273,7 +273,13 @@ pub struct ALPArray {
...
 
 impl ALP {
-    pub const ID: ArrayId = ArrayId::new_ref("vortex.alp");
+    pub const ID: &'static str = "vortex.alp";
+
+    /// Returns the cached [`ArrayId`] for this encoding.
+    pub fn array_id() -> ArrayId {
+        static CACHED: std::sync::OnceLock<ArrayId> = std::sync::OnceLock::new();
+        *CACHED.get_or_init(|| ArrayId::new(Self::ID))
+    }
```
joseph-isaacs, line 282: Does this not have an atomic? On read

gatesn: We could force vtables to hold onto an instance of their ID by having `VTable::id(&Array) -> &ArrayId`?

But it means no more ZSTs for vtables

gatesn:
Lol yeah.... I'm not sure this will work unless we hold the Id on the vtable

gatesn:
Closing until I have time to try this again

PR #7070, opened 2026-03-19
chore: `AggregateFnRef` [de]serialize

blaginin:


vortex-array/src/aggregate_fn/erased.rs
```diff
@@ -34,6 +39,37 @@ use crate::dtype::DType;
...
+            );
+        }
+
+        Ok(agg_fn)
+    }
+
+    /// Serialize this aggregate function's options to bytes.
+    ///
+    /// Returns `Ok(None)` if the function is not serializable.
+    pub fn serialize(&self) -> VortexResult<Option<Vec<u8>>> {
```
gatesn: I think this should instead be `self.options().serialize()` and not have this function at all

vortex-array/src/aggregate_fn/erased.rs
```diff
@@ -34,6 +39,37 @@ use crate::dtype::DType;
 pub struct AggregateFnRef(pub(super) Arc<dyn DynAggregateFn>);
 
 impl AggregateFnRef {
+    /// Deserialize an aggregate function from its ID and serialized metadata bytes.
+    ///
+    /// Looks up the aggregate function plugin by ID in the session's registry
+    /// and delegates deserialization to it.
+    pub fn deserialize(id: &str, metadata: &[u8], session: &VortexSession) -> VortexResult<Self> {
```
gatesn: I don't know whether this should live here, or `AggregateSession::deserialize(&self, id: &AggregateFnId, metadata: &[u8])`

vortex-array/src/aggregate_fn/erased.rs
blaginin: maybe https://github.com/vortex-data/vortex/pull/7070/commits/df97bacb289332a4a87961e204c97db5c07c0ee0 so it'll be similar to expr experience?

vortex-array/src/aggregate_fn/proto.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// Extension trait for serializing an [`AggregateFnRef`] to protobuf.
+///
+/// Note: the serialization format is not stable and may change between versions.
+pub trait AggregateFnSerializeProtoExt {
+    /// Serialize this aggregate function to its protobuf representation.
+    fn serialize_proto(&self) -> VortexResult<pb::AggregateFn>;
+}
+
+impl AggregateFnSerializeProtoExt for AggregateFnRef {
```
gatesn: You should just be able to `impl AggregateFnRef { ... }`, no need for a trait

blaginin: sure, https://github.com/vortex-data/vortex/pull/7070/commits/ab9c381f90943ae88d990fbd88efb7beb7c86d77

gatesn (approved):
Excellent, thank you!

PR #7072, opened 2026-03-19
Show truncation indicator in array display

dimitarvdimitrov:
When displaying array values that exceed the display limit (16), we now show `...` at the end instead of silently truncating. Makes it obvious the output is incomplete.

### DIFF
--- a/vortex-array/src/display/mod.rs
+++ b/vortex-array/src/display/mod.rs
@@ -477,7 +477,7 @@ impl dyn DynArray + '_ {
             DisplayOptions::CommaSeparatedScalars {
                 omit_comma_after_space,
             } => {
-                let opening_brace= if f.alternate() { "[\n" } else { "[" };
+                let opening_brace = if f.alternate() { "[\n" } else { "[" };
                 let closing_brace = if f.alternate() { "\n]" } else { "]" };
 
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

joseph-isaacs (commented):
we need to add custom trunc len. but this can be a follow up

joseph-isaacs (commented):
would be nice to include the last few values too?

robert3005:
I have added the logic to print trailer as well. We now print LIMIT - 3, ..., LAST 3 items when display is truncated

vortex-array/src/display/mod.rs
```diff
@@ -480,17 +480,31 @@ impl dyn DynArray + '_ {
...
+                        "{}",
+                        (self.len() - 3..self.len())
+                            .map(|i| self
+                                .scalar_at(i)
+                                .map_or_else(|e| format!("<error: {e}>"), |s| s.to_string()))
+                            .format(sep)
+                    )?;
+                }
+                let closing_brace = if f.alternate() { "\n]" } else { "]" };
+                write!(f, "{closing_brace}")
```
joseph-isaacs: can we simplify this

PR #7077, opened 2026-03-20
fix fuzzer issue #7074

joseph-isaacs:
maybe?


vortex-array/src/aggregate_fn/fns/min_max/bool.rs
```diff
@@ -13,6 +13,10 @@ use crate::dtype::Nullability::NonNullable;
 use crate::scalar::Scalar;
 
 pub(super) fn accumulate_bool(partial: &mut MinMaxPartial, array: &BoolArray) -> VortexResult<()> {
+    if array.is_empty() {
```
gatesn, line 16: Blergh, yeah I wonder if we change the Accumulator to never invoke the vtable for empty arrays? All the accumulators need to handle it at the moment.

onursatici (approved):

PR #7079, opened 2026-03-20
chore: cache cpu feat check for micro-benchmarks

0ax1:


robert3005 (approved):
I'm sorry, should have realised there's a main

PR #7080, opened 2026-03-20
[Experiment] Self-defined arrow-export

AdamGS:
## Summary

This is an attempt at playing around with making arrow-exporting both more extendable for Vortex extension types, support arrow's extension types and allow encodings to define their own preferred way to export to arrow (given some inputs).

My main motivation here is wanting to be able to export the upcoming `parquet-variant` encoding into a canonical arrow extension type.

## API Changes

todo!()

## Testing

todo!()


robert3005:
I think you really want any conversion of vortex arrays to other arrays be an execution

AdamGS:
WDYM by execution?

robert3005:
Implementation of an Executable trait or something similar so that we can customise it dependeing on the context. For example cuda arrow export is different than host arrow export

AdamGS:
I think there has to be some mix, you must shift some stuff into the extensions/encodings themselves

gatesn:
The plan was to just have the vortex-arrow crate depend on all the code for the builtin arrow types, not sure there's anything else to change really?

AdamGS:
1. This PR has gotten way too big, this is mostly me trying out some ideas, no real intent on merging any of it (definitely not in its current form).
2. I think even what your'e describing with `vortex-arrow` (which solves the issue for "core" encodings and extensions) is probably not enough, you also want to allow extension dtypes to impact this behavior in some way (like allowing them to export into arrow canonical types).
3. This is also missing the other direction (importing arrow into Vortex). Maybe that's on the session, or something else?

PR #7081, opened 2026-03-20
Fuzzer includes unstable_encodings

robert3005:
We should know if there's something broken with them even if they're not enabled
by default so we can enable them by default

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -18,7 +18,7 @@ version = { workspace = true }
 cargo-fuzz = true
 
 [features]
-default = ["native"]
+default = ["native", "vortex/unstable_encodings"]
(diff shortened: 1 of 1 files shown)


0ax1 (approved):

fuzz/Cargo.toml
```diff
@@ -18,7 +18,7 @@ version = { workspace = true }
 cargo-fuzz = true
 
 [features]
-default = ["native"]
+default = ["native", "vortex/unstable_encodings"]
```
joseph-isaacs, line 21: next time can we make this explicit. I think we should run the fuzzer with and without this

robert3005: I think the overall diff is small enough that you’re just adding compute with little practical benefit

PR #7082, opened 2026-03-20
feat[array]: compositional array display logic

joseph-isaacs:
Breaks up the display logic for array tree into multiple composable extractors


0ax1 (approved):
🚢 it

PR #7089, opened 2026-03-20
do not validate already valid arrow varbin arrays

onursatici:
## Summary

varbin::try_new validates every element of the input array, but for arrow arrays this is already validated, so we can skip validation.

This validation can take longer than compressing and writing the entire array into disk.

on varbinview this is already fixed and we skip validation correctly

### DIFF
--- a/vortex-array/src/arrow/convert.rs
+++ b/vortex-array/src/arrow/convert.rs
@@ -275,12 +275,15 @@ where
             DataType::Utf8 | DataType::LargeUtf8 => DType::Utf8(nullable.into()),
             dt => vortex_panic!("Invalid data type for ByteArray: {dt}"),
         };
-        Ok(VarBinArray::try_new(
-            value.offsets().clone().into_array(),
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrow/convert.rs
```diff
@@ -275,12 +275,15 @@ where
             DataType::Utf8 | DataType::LargeUtf8 => DType::Utf8(nullable.into()),
             dt => vortex_panic!("Invalid data type for ByteArray: {dt}"),
         };
```
joseph-isaacs, line 277: at least to a debug assert here

onursatici: `new_unchecked` has this under `new_unchecked_from_handle`:
```rust
        #[cfg(debug_assertions)]
        Self::validate(&offsets, &bytes, &dtype, &validity)
            .vortex_expect("[Debug Assertion]: Invalid `VarBinArray` parameters");

```
so we already have dbg assert validate here

joseph-isaacs (approved):

PR #7103, opened 2026-03-20
Ensure that serialized arrays have 0 offset

### LINKED ISSUE #7094: SparseArray slices can serialize with non-zero patch offsets, but SparseVTable::build asserts offset == 0
## Summary

There appears to be an internal contract mismatch in Vortex sparse encoding:

- slicing a `SparseArray` preserves a non-zero patch offset
- serialized sparse metadata keeps that offset
- deserialization then asserts that sparse patches must start at offset `0`

That makes some sliced sparse arrays unreadable after round-trip.

## Relevant Code

`SparseArray::slice()` delegates to `Patches::slice()`:
- `vortex-sparse/src/ops.rs`
- `vortex-array/src/patches.rs`

`Patches::slice()` preserves the logical offset:

```rust
Some(Self {
    array_len: range.len(),
    offset: range.start + self.offset(),
    indices,
    values,
    chunk_offsets,
    offset_within_chunk,
})
```

But `SparseVTable::build()` asserts offset must be `0`:

```rust
assert_eq!(
    metadata.0.patches.offset(),
    0,
    "Patches must start at offset 0"
);
```

## Observed Behavior

At read time we hit panics like:

```text
assertion `left == right` failed: Patches must start at offset 0
  left: 16384
 right: 0
```

### Steps to reproduce
1. create a sparse array
    - `DictArray`
    - whose `codes` child is a `SparseArray`
    - the dict array is sliced with a non-zero row offset
    - the sliced value is serialized and read back
3. slice it with a non-zero start
4. serialize it
5. deserialize it

### Environment

- Vortex version: `0.58.0`
- `rustc`: `1.94.0 (4a4ef493e 2026-03-02)`
- `cargo`: `1.94.0 (85eff7c80 2026-01-15)`

### Additional context

_No response_

robert3005:
While the writer will always make sure this is the case users of the api can
serialise arrays with arbitrary offsets which then fail to deserialise

fix #7094 

### DIFF
--- a/encodings/fastlanes/src/bitpacking/vtable/mod.rs
+++ b/encodings/fastlanes/src/bitpacking/vtable/mod.rs
@@ -247,6 +247,18 @@ impl VTable for BitPacked {
     }
 
     fn serialize(metadata: Self::Metadata) -> VortexResult<Option<Vec<u8>>> {
+        vortex_ensure!(
+            metadata.offset == 0,
+            "BitPackedArray offset must be 0 for serialization, got {}",
+            metadata.offset
+        );
+        if let Some(ref patches) = metadata.patches {
+            vortex_ensure!(
+                patches.offset()? == 0,
+                "BitPackedArray patches offset must be 0 for serialization, got {}",
+                patches.offset()?
+            );
+        }
         Ok(Some(metadata.serialize()))
     }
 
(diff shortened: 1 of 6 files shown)


robert3005:
We rely on being able to ser/de RunEnd metadata for arrow conversion without depending on the RunEndArray directly.

robert3005:
Ok, this needs a bit more work

PR #7105, opened 2026-03-20
Only pass vtable once via ExtDType<V>

gatesn:


vortex-array/src/extension/tests/divisible_int.rs
```diff
@@ -51,10 +51,7 @@ impl ExtVTable for DivisibleInt {
         Ok(Divisor(n))
     }
 
-    fn validate_dtype(
-        &self,
-        ext_dtype: &crate::dtype::extension::ExtDType<Self>,
-    ) -> VortexResult<()> {
+    fn validate_dtype(ext_dtype: &crate::dtype::extension::ExtDType<Self>) -> VortexResult<()> {
```
robert3005: can we import this?

vortex-array/src/scalar/arrow.rs
```diff
@@ -466,14 +466,12 @@ mod tests {
             }
 
             fn validate_dtype(
-                &self,
                 _ext_dtype: &crate::dtype::extension::ExtDType<Self>,
```
robert3005: can we also import this and the one on line 477?

vortex-array/src/extension/tests/divisible_int.rs
gatesn: Was there before me, will be there after me!

robert3005 (approved):

PR #7106, opened 2026-03-20
Adamg/fix rle delta nullability patch

AdamGS:
An attempt at fixing #7090. First commit is the initial fix, then the second commit is fixing an edge case in the original fix.


robert3005:
I opened #7118 which is a simpler fix

PR #7107, opened 2026-03-20
Norm-extracted vectors

connortsui20:
## Summary

Tracking Issue: https://github.com/vortex-data/vortex/issues/6865

Related PR: https://github.com/vortex-data/vortex/pull/7018

Adds a new encoding specific to the vector extension type.

Note that we cannot actually utilize this by adding it to the compressor until we make it pluggable (see #7018). When that does eventually land, we can simply create a `NormVectorScheme` that uses the `NormVectorArray`.

### Implementation

We currently do not have a good way of broadcasting a multiplication onto a `FixedSizeList`, so this implementation hand rolls the norm multiplication. Additionally, we still do not have https://github.com/vortex-data/vortex/pull/6717 so the `scalar_at` implementation is also slow.

## API Changes

Adds a new encoding type `NormVector`

## Testing

Some simple tests including roundtrips.


vortex-tensor/src/encodings/norm/vtable/operations.rs
connortsui20, line 1: Note that once we get `ScalarValue::Array` (https://github.com/vortex-data/vortex/pull/6717), this will be a lot less ugly.

vortex-tensor/src/encodings/norm/array.rs
```diff
@@ -0,0 +1,221 @@
...
+    ///
+    /// This must have the same dtype as the elements of the vector array.
+    pub(crate) norms: ArrayRef,
+
+    /// Stats set owned by this array.
+    pub(crate) stats_set: ArrayStats,
+}
+
+impl NormVectorArray {
+    /// Creates a new [`NormVectorArray`] from a unit-normalized vector array and its L2 norms.
```
gatesn: I'm a little confused by the docs. It is an array of vectors, and "its L2 norms". What does "its" refer to? 

(I know the answer but only by reading the rest of the file)

vortex-tensor/src/encodings/norm/array.rs
```diff
@@ -0,0 +1,221 @@
...
+        Ok(ext.clone())
+    }
+
+    /// Encodes a [`Vector`] extension array into a [`NormVectorArray`] by computing L2 norms and
+    /// dividing each vector by its norm.
+    ///
+    /// The input must be a [`Vector`] extension array with floating-point elements. Nullable inputs
+    /// are supported; the validity mask is preserved and the normalized data for null rows is
+    /// unspecified.
+    pub fn compress(vector_array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Self> {
```
gatesn, line 177: Worth noting that this operation is lossy per floating point arithmetic.

vortex-tensor/src/encodings/norm/array.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let list_size = extension_list_size(&ext)?;
+        let row_count = vector_array.len();
+        let nullability = Nullability::from(vector_array.dtype().is_nullable());
+
+        // Compute L2 norms using the scalar function. If the input is nullable, the norms will
+        // also be nullable (null vectors produce null norms).
+        let storage = extension_storage(&vector_array)?;
+        let l2_norm_expr =
+            Expression::try_new(ScalarFn::new(L2Norm, EmptyOptions).erased(), [root()])?;
```
gatesn, line 189: Short cut would be `L2Norm.new_expr(...)`

gatesn: Even then, we should have a nicer way of doing this!!

vortex-tensor/src/encodings/norm/array.rs
```diff
@@ -0,0 +1,221 @@
...
+        let norms_prim: PrimitiveArray = vector_array.apply(&l2_norm_expr)?.execute(ctx)?;
+        let norms_array = norms_prim.clone().into_array();
+
+        // Extract flat elements from the (always non-nullable) storage for normalization.
+        let flat = extract_flat_elements(&storage, list_size, ctx)?;
```
gatesn, line 194: What is FlatElements?

vortex-tensor/src/encodings/norm/array.rs
```diff
@@ -0,0 +1,221 @@
...
+        let list_size = extension_list_size(&ext)?;
+        let row_count = vector_array.len();
+        let nullability = Nullability::from(vector_array.dtype().is_nullable());
+
+        // Compute L2 norms using the scalar function. If the input is nullable, the norms will
+        // also be nullable (null vectors produce null norms).
+        let storage = extension_storage(&vector_array)?;
+        let l2_norm_expr =
+            Expression::try_new(ScalarFn::new(L2Norm, EmptyOptions).erased(), [root()])?;
+        let norms_prim: PrimitiveArray = vector_array.apply(&l2_norm_expr)?.execute(ctx)?;
```
gatesn, line 190: You said vector array can be nullable, but you don't check validity when iterating below

vortex-tensor/src/encodings/norm/array.rs
```diff
@@ -0,0 +1,221 @@
...
+        let norms_array = norms_prim.clone().into_array();
+
+        // Extract flat elements from the (always non-nullable) storage for normalization.
+        let flat = extract_flat_elements(&storage, list_size, ctx)?;
+
+        match_each_float_ptype!(flat.ptype(), |T| {
+            let norms_slice = norms_prim.as_slice::<T>();
+
+            let normalized_elems: PrimitiveArray = (0..row_count)
+                .flat_map(|i| {
```
gatesn: `i` might be total garbage if it is null

vortex-tensor/src/encodings/norm/array.rs
```diff
@@ -0,0 +1,221 @@
...
+        let storage = extension_storage(&vector_array)?;
+        let l2_norm_expr =
+            Expression::try_new(ScalarFn::new(L2Norm, EmptyOptions).erased(), [root()])?;
+        let norms_prim: PrimitiveArray = vector_array.apply(&l2_norm_expr)?.execute(ctx)?;
+        let norms_array = norms_prim.clone().into_array();
+
+        // Extract flat elements from the (always non-nullable) storage for normalization.
+        let flat = extract_flat_elements(&storage, list_size, ctx)?;
+
+        match_each_float_ptype!(flat.ptype(), |T| {
```
gatesn, line 196: This all seems a bit convoluted? Can you just run-end encode the norms by the vector lengths, then use a divide function?

gatesn (changes requested):

vortex-tensor/src/encodings/norm/array.rs
connortsui20: It's a helper function I use for all of the other tensor types to make sure we can deal with `ConstantArray` correctly:

```rust
/// Extracts the flat primitive elements from a tensor storage array (FixedSizeList).
///
/// When the input is a [`ConstantArray`] (e.g., a literal query vector), only a single row is
/// materialized to avoid expanding it to the full column length.
```

vortex-tensor/src/encodings/norm/array.rs
connortsui20: whoops

connortsui20:
ok just realized that a bunch of stuff was wrong because I changed from non-nullable to nullable halfway through, will fix everything

vortex-tensor/src/encodings/norm/array.rs
connortsui20: So this is what the decompress path would look like (the compress path would be similar):

```rust
        // We need to multiply each vector element with its respective norm. We do not have any kind
        // of "broadcast" expression to each of the `FixedSizeList` elements, so we can mimic this
        // by multiplying the normalized vector array with a `RunEnd(Sequence)` array.
        let base: PValue = list_size.into();
        let multiplier: PValue = base;
        let ends_ptype = base.ptype();
        let ends_nullability = Nullability::NonNullable;
        let ends =
            SequenceArray::try_new(base, multiplier, ends_ptype, ends_nullability, num_vectors)?;

        let runend = RunEndArray::try_new(ends.into_array(), self.norms.clone())?;

        let storage = extension_storage(&self.vector_array)?;
        let fsl: FixedSizeListArray = storage.execute(ctx)?;
        let elements = fsl.elements();
        debug_assert!(elements.dtype().is_primitive());

        let decompress = elements.binary(runend.into_array(), Operator::Mul)?;

        let denormalized_elements: PrimitiveArray = decompress.execute(ctx)?;

        // SAFETY: TODO
        let fsl = unsafe {
            FixedSizeListArray::new_unchecked(
                denormalized_elements.into_array(),
                list_size,
                self.validity()?,
                num_vectors,
            )
        };

        Ok(ExtensionArray::new(ext.clone(), fsl.into_array()).into_array())
    }
```

I don't think that this is actually less convoluted? But maybe we do this regardless because the optimizer might be able to do something

connortsui20:
Also @gatesn the `.binary(???, Operator::Div)` won't work because we need a safe divide by 0, and `Operator::Div` will return an `ArrowError` that I can't easily intercept.

Do you think it is fine to just have this broadcast logic for decompression and not compression? (see the latest [commit](https://github.com/vortex-data/vortex/pull/7107/commits/e288bb1e6543a6dc126fa24969f977036e7f2280))

gatesn:
I still see this as a ScalarFn though no?

connortsui20:
If you're talking about how some of our encodings are scalar, I still don't think that is right because scalar fns are supposed to logically transform the data in some meaningful way whereas this is a physical transform and the logical data needs to stay identical.

connortsui20:
Im going to close this since https://github.com/vortex-data/vortex/pull/7225 did a bunch of changes that overlap here, want to open a fresh pr regardless

PR #7118, opened 2026-03-23
Cleanup fastlanes encodings crate

robert3005:
* Remove dead code only used in tests
* Simplify signatures that needlessly return Results
* Add test coverage for bit_transpose functions
* Propagate patch offsets when compressing patches

### DIFF
--- a/encodings/fastlanes/src/bit_transpose/validity.rs
+++ b/encodings/fastlanes/src/bit_transpose/validity.rs
@@ -37,7 +37,6 @@ pub fn transpose_validity(validity: &Validity, ctx: &mut ExecutionCtx) -> Vortex
     }
 }
 
-#[inline]
 pub fn transpose_bitbuffer(bits: BitBuffer) -> BitBuffer {
     let (offset, len, bytes) = bits.into_inner();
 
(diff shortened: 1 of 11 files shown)


encodings/fastlanes/src/bit_transpose/validity.rs
```diff
@@ -37,7 +37,6 @@ pub fn transpose_validity(validity: &Validity, ctx: &mut ExecutionCtx) -> Vortex
     }
 }
 
-#[inline]
```
joseph-isaacs, line 40: ?

encodings/fastlanes/src/bit_transpose/validity.rs
```diff
@@ -79,11 +78,10 @@ pub fn untranspose_validity(validity: &Validity, ctx: &mut ExecutionCtx) -> Vort
...
```
joseph-isaacs, line 82: ?

joseph-isaacs:
Mind pulling out the fix and the general refactor into two?

encodings/fastlanes/src/bit_transpose/validity.rs
robert3005: I don't think that inline was helping anything, it was added without any research or justification. It was a leftover from before that function got rewritten

encodings/fastlanes/src/rle/array/rle_compress.rs
```diff
@@ -142,49 +140,50 @@ mod tests {
...
         assert_eq!(encoded.len(), 6);
     }
 
     #[test]
     fn test_empty_length() {
         let values: Buffer<u32> = Buffer::empty();
-        let array = values.into_array();
-        let encoded = RLEArray::encode(&array.to_primitive()).unwrap();
+        let encoded =
+            RLEArray::encode(&PrimitiveArray::new(values, Validity::NonNullable)).unwrap();
```
joseph-isaacs, line 186: could use prim from_iter

joseph-isaacs (approved):

PR #7122, opened 2026-03-23
chore[array]: add filter bool bench

joseph-isaacs:
Add filter bool benchmarks to later optimize

### DIFF
--- a/vortex-array/benches/filter_bool.rs
+++ b/vortex-array/benches/filter_bool.rs
@@ -4,7 +4,7 @@
 //! Benchmarks for filtering BoolArray by a boolean mask.
 //!
 //! Tests multiple mask patterns (mostly-true, mostly-false, random, correlated runs)
-//! with both uniform-random and power-law distributions, across array sizes from 1K to 500K.
+//! with both uniform-random and power-law distributions, across array sizes from 1K to 250K.
 
 #![allow(clippy::unwrap_used)]
 #![allow(clippy::cast_possible_truncation)]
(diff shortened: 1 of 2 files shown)


vortex-array/benches/filter_bool.rs
```diff
@@ -0,0 +1,325 @@
...
+            builder.freeze()
+        }
+    }
+}
+
+#[divan::bench(args = DENSITIES)]
+fn head_to_head_old(bencher: Bencher, density: f64) {
+    let mut rng = StdRng::seed_from_u64(ARRAY_SEED);
+    let src = BitBuffer::from_iter((0..DENSITY_SWEEP_SIZE).map(|_| rng.random_bool(0.5)));
+    let true_count = make_density_mask(DENSITY_SWEEP_SIZE, density, &mut mask_rng()).true_count();
```
0ax1: why not compute both in with inputs rather than calc true count upfront? 
```
let true_count = make_density_mask(DENSITY_SWEEP_SIZE, density, &mut mask_rng()).true_count()
```

vortex-array/benches/filter_bool.rs
```diff
@@ -0,0 +1,325 @@
...
+            let mut bits: Vec<bool> = vec![true; n];
+            bits[n / 2] = false;
+            Mask::from_buffer(BitBuffer::from_iter(bits))
+        })
+        .bench_values(|m| array.filter(m).unwrap());
+}
+
+#[divan::bench(args = LARGE_SIZES)]
+fn filter_ultra_sparse(bencher: Bencher, n: usize) {
+    let array = make_random_bool_array(n, &mut StdRng::seed_from_u64(ARRAY_SEED));
```
0ax1, line 439: shall we black box all bench inputs here, and elsewhere

0ax1 (approved):

PR #7124, opened 2026-03-23
fix[delta]: compress with nulls undef values use last valid value (or 0)

joseph-isaacs:
When we have undef values this can affect decompression with a undef value (use the previous valid value instead).

### DIFF
--- a/encodings/fastlanes/src/delta/array/delta_compress.rs
+++ b/encodings/fastlanes/src/delta/array/delta_compress.rs
@@ -17,13 +17,18 @@ use vortex_buffer::BufferMut;
 use vortex_error::VortexResult;
 
 use crate::bit_transpose::transpose_validity;
+use crate::fill_forward_nulls;
 
 pub fn delta_compress(
     array: &PrimitiveArray,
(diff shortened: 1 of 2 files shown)


encodings/fastlanes/src/delta/array/delta_compress.rs
```diff
@@ -124,4 +129,30 @@ mod tests {
...
+
+    /// Regression test: delta + bitpacked encoding must correctly round-trip nullable arrays
+    /// where null positions contain arbitrary values. Without fill-forward, the delta cumulative
+    /// sum propagates corrupted values from null positions.
+    #[test]
+    fn delta_bitpacked_trailing_nulls() {
+        use vortex_array::IntoArray;
+        use vortex_array::ToCanonical;
+
+        use crate::bitpack_compress::bitpack_encode;
```
0ax1, line 141:
```   
use crate::bitpack_compress::bitpack_encode;
use crate::delta_compress;
```
Move up?

0ax1 (approved):

robert3005:
Thanks @joseph-isaacs

PR #7125, opened 2026-03-23
perf[array]: bool filter kernel optimisation

joseph-isaacs:
Iterate bit buffers instead of indices or slices when filtering boolean arrays

### DIFF
--- a/vortex-array/src/arrays/bool/compute/filter.rs
+++ b/vortex-array/src/arrays/bool/compute/filter.rs
@@ -2,12 +2,10 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use vortex_buffer::BitBuffer;
-use vortex_buffer::BitBufferMut;
 use vortex_buffer::get_bit;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_mask::Mask;
-use vortex_mask::MaskIter;
 
 use crate::ArrayRef;
 use crate::IntoArray;
@@ -17,8 +15,9 @@ use crate::arrays::BoolArray;
 use crate::arrays::bool::BoolArrayExt;
 use crate::arrays::filter::FilterReduce;
 
-/// If the filter density is above 80%, we use slices to filter the array instead of indices.
-const FILTER_SLICES_DENSITY_THRESHOLD: f64 = 0.8;
+/// Below this density threshold, use the sparse path which iterates only set
+/// bits in the mask. Above it, the word-level PEXT approach is faster.
+const SPARSE_DENSITY_THRESHOLD: f64 = 0.05;
 
 impl FilterReduce for Bool {
     fn filter(array: ArrayView<'_, Bool>, mask: &Mask) -> VortexResult<Option<ArrayRef>> {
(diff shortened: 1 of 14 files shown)


vortex-array/src/arrays/filter/execute/primitive.rs
```diff
@@ -17,6 +19,16 @@ pub fn filter_primitive(array: &PrimitiveArray, mask: &Arc<MaskValues>) -> Primi
...
 
+    // u8 uses a specialized byte-compress LUT: 256-entry table avoids
+    // materializing indices/slices and processes 8 elements per mask byte.
+    if array.ptype() == PType::U8 {
+        let filtered_buffer =
+            byte_compress::filter_buffer_u8(array.to_buffer::<u8>(), mask.as_ref());
+        // SAFETY: We filter both the validity and the buffer with the same mask, so they must have
+        // the same length.
+        return unsafe { PrimitiveArray::new_unchecked(filtered_buffer, filtered_validity) };
+    }
```
joseph-isaacs: might be worth adding a todo to look at this for {i/u}{16/u32}?

vortex-array/src/arrays/filter/execute/primitive.rs
```diff
@@ -17,6 +19,16 @@ pub fn filter_primitive(array: &PrimitiveArray, mask: &Arc<MaskValues>) -> Primi
         .vortex_expect("primitive validity should be derivable");
     let filtered_validity = filter_validity(validity, mask);
 
+    // u8 uses a specialized byte-compress LUT: 256-entry table avoids
+    // materializing indices/slices and processes 8 elements per mask byte.
+    if array.ptype() == PType::U8 {
```
joseph-isaacs: i8? or bytewidth == 1?

joseph-isaacs (commented):
I cannot approve but I do approve this

vortex-array/src/arrays/filter/execute/primitive.rs
robert3005: I am investigating it now and will only merge once I have numbers.

robert3005: Fwiw this is always better for byte width 1 and better for other widths if mask density is around 0.5 or above.

vortex-array/src/arrays/bool/compute/filter.rs
```diff
@@ -28,47 +29,279 @@ impl FilterReduce for Bool {
...
+
+/// Byte-level LUT PEXT fallback.
+///
+/// Processes each byte of the u64 independently using a precomputed 256-entry
+/// lookup table per mask byte. Each byte PEXT is a single table lookup with no
+/// data dependencies between bytes, making this faster than the parallel-prefix
+/// approach (~12ns vs ~18ns per word).
+///
+/// This is the best software PEXT for platforms without BMI2 (ARM64) and for
+/// AMD Zen 1/2 where the hardware PEXT instruction is microcoded.
```
onursatici: fallback is only called with `std::arch::is_x86_feature_detected!("bmi2")`, but wouldn't AMD Zen 1/2 would report that they have bmi2 available, it is just slow

vortex-array/src/arrays/filter/execute/byte_compress.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    // Fast path: byte-wide values benefit from avoiding index materialization more often. Wider
+    // values need enough selected values to justify scanning every mask byte directly.
+    if mask_offset == 0
+        && (size_of::<T>() == 1 || mask.density() >= BYTE_COMPRESS_DENSITY_THRESHOLD)
+    {
+        return filter_aligned(src, mask_bytes, true_count);
+    }
+
+    // Slow path: lower-density wide values or unaligned masks are better handled by the generic path.
```
onursatici: we can fallback for lower density values but can we not handle the first few mask bits and get the rest of the mask which is byte aligned in the fast path?

vortex-array/src/arrays/bool/compute/filter.rs
```diff
@@ -28,47 +29,279 @@ impl FilterReduce for Bool {
...
+
+    let out_u64s = true_count.div_ceil(64);
+    let mut output: Vec<u64> = Vec::with_capacity(out_u64s + 1);
+    let out_ptr = output.as_mut_ptr();
+    let mut out_idx: usize = 0;
 
+    // u128 accumulator: overflow naturally held in upper 64 bits, eliminating
+    // the tricky `extracted >> (popcount - accum_bits)` re-derivation. Just
+    // flush low 64 bits and shift down when full.
+    let mut accum: u128 = 0;
```
onursatici, line 147: this is cool

vortex-array/src/arrays/bool/compute/filter.rs
robert3005: yes, this is a bit of leftover. I would make it simple and if there's bmi2 let it run. I don't think we will have significant footprint on Zen 1/2

vortex-array/src/arrays/filter/execute/byte_compress.rs
robert3005: This is not entirely about alignment, it's more about density. I can have it handle non 0 offset better

robert3005:
I have added handling for non 0 offsets for byte_compress method

onursatici (approved):

PR #7126, opened 2026-03-23
Testing FFI from C side

myrrc:
Add Catch2 target and first tests for C API

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -708,7 +708,7 @@ jobs:
           mkdir -p vortex-cxx/build
           cmake -S vortex-cxx -B vortex-cxx/build -DVORTEX_ENABLE_TESTING=ON -DVORTEX_ENABLE_ASAN=ON
           cmake --build vortex-cxx/build --parallel $(nproc)
-          ctest --test-dir vortex-cxx/build -V
+          ctest --test-dir vortex-cxx/build -j $(nproc) -V
       - name: Build and run the example in release mode
         run: |
           cmake -S vortex-cxx/examples -B vortex-cxx/examples/build -DCMAKE_BUILD_TYPE=Release
@@ -843,8 +843,8 @@ jobs:
           cd vortex-ffi
           mkdir build
           cmake -Bbuild
-          cmake --build build --parallel $(nproc)
-          ctest --test-dir build -V
+          cmake --build build -j $(nproc)
+          ctest --test-dir build -j $(nproc)
 
   check-java-publish-build:
     runs-on: ${{ matrix.target.runs-on }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -821,6 +821,31 @@ jobs:
         run: |
           find flatbuffers/ -type f -name "*.fbs" | sed 's/^flatbuffers\///' | xargs -I{} -n1 flatc -I flatbuffers.HEAD --conform-includes flatbuffers --conform flatbuffers/{} flatbuffers.HEAD/{}
 
+  ffi-c-test:
+    name: "C API test build"
+    timeout-minutes: 10
+    runs-on: >-
+      ${{ github.repository == 'vortex-data/vortex'
+          && format('runs-on={0}/runner=amd64-medium/image=ubuntu24-full-x64-pre-v2/tag=cxx-build', github.run_id)
+          || 'ubuntu-latest' }}
+    steps:
+      - uses: runs-on/action@v2
+        if: github.repository == 'vortex-data/vortex'
+        with:
+          sccache: s3
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/setup-prebuild
+      - name: "regenerate FFI header file"
+        run: |
+          cargo +$NIGHTLY_TOOLCHAIN build -p vortex-ffi
+      - name: Build and run C++ unit tests
+        run: |
+          cd vortex-ffi
+          mkdir build
+          cmake -Bbuild
+          cmake --build build --parallel $(nproc)
+          ctest --test-dir build -V
+
   check-java-publish-build:
     runs-on: ${{ matrix.target.runs-on }}
     container:
--- a/vortex-ffi/CMakeLists.txt
+++ b/vortex-ffi/CMakeLists.txt
@@ -1,0 +1,52 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+cmake_minimum_required(VERSION 3.10)
+
+project(VortexFFI
+    VERSION 0.0.1
+    LANGUAGES C CXX)
+set(CMAKE_C_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror -Wextra -Wpedantic")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wextra -Wpedantic")
+
+option(BUILD_TESTING "Build tests" ON)
+
+if (NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Debug)
+endif()
+
+string(TOLOWER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_LOWER)
+set(LIBRARY_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../target/${CMAKE_BUILD_TYPE_LOWER}")
+
+if(WIN32)
+    set(LIBRARY_PATH "${LIBRARY_DIR}/libvortex_ffi.dll")
+elseif(APPLE)
+    set(LIBRARY_PATH "${LIBRARY_DIR}/libvortex_ffi.dylib")
+else()
+    set(LIBRARY_PATH "${LIBRARY_DIR}/libvortex_ffi.so")
+endif()
+
+set(LIBRARY_HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/cinclude")
+
+message("Library dir ${LIBRARY_DIR}
+Library path ${LIBRARY_PATH}
+Headers path ${LIBRARY_HEADERS}")
+
+if (NOT EXISTS "${LIBRARY_PATH}")
+    message(FATAL_ERROR "FFI library not found")
+endif()
+
+add_library(vortex_ffi SHARED IMPORTED)
+set_target_properties(vortex_ffi PROPERTIES
+    IMPORTED_LOCATION "${LIBRARY_PATH}"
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -821,6 +821,31 @@ jobs:
...
+      - uses: ./.github/actions/setup-prebuild
+      - name: "regenerate FFI header file"
+        run: |
+          cargo +$NIGHTLY_TOOLCHAIN build -p vortex-ffi
+      - name: Build and run C++ unit tests
+        run: |
+          cd vortex-ffi
+          mkdir build
+          cmake -Bbuild
+          cmake --build build --parallel $(nproc)
```
0ax1: `cmake --build build --parallel $(nproc)` wild that this is not the default.

.github/workflows/ci.yml
```diff
@@ -821,6 +821,31 @@ jobs:
...
+          ctest --test-dir build -V
```
0ax1: Should the tests also be run in parallel? Not a bottleneck right now but maybe useful in the future.

vortex-ffi/cinclude/vortex.h
```diff
@@ -747,6 +747,13 @@ void vx_session_free(vx_session *ptr);
  */
 vx_session *vx_session_new(void);
 
+/**
+ * Clone a Vortex session, returning an owned copy.
+ *
+ * The caller is responsible for freeing the session with [`vx_session_free`].
+ */
+vx_session *vx_session_clone(vx_session *session);
```
0ax1: Session doesn't have to mutable right. This probably should be `const`.

vortex-ffi/src/dtype.rs
```diff
@@ -206,10 +206,9 @@ pub unsafe extern "C-unwind" fn vx_dtype_decimal_scale(dtype: *const vx_dtype) -
 pub unsafe extern "C-unwind" fn vx_dtype_struct_dtype(
     dtype: *const vx_dtype,
 ) -> *const vx_struct_fields {
-    // TODO(joe): propagate this error up instead of expecting
-    let struct_dtype = vx_dtype::as_ref(dtype)
-        .as_struct_fields_opt()
-        .vortex_expect("not a struct dtype");
+    let Some(struct_dtype) = vx_dtype::as_ref(dtype).as_struct_fields_opt() else {
```
0ax1, line 209: What's the idea around error handling from Rust to Vortex? In other instances we currently use an error out parameter, rather than returning null:
```
 error_out: *mut *mut vx_error
```
We don't have to solve this as part of this PR. But asked differently why does it change as part of it?

vortex-ffi/src/dtype.rs
```diff
@@ -219,10 +218,9 @@ pub unsafe extern "C-unwind" fn vx_dtype_struct_dtype(
 /// Do NOT free the returned dtype pointer - it shares the lifetime of the list dtype.
 #[unsafe(no_mangle)]
 pub unsafe extern "C-unwind" fn vx_dtype_list_element(dtype: *const vx_dtype) -> *const vx_dtype {
-    // TODO(joe): propagate this error up instead of expecting
-    let element_dtype = vx_dtype::as_ref(dtype)
-        .as_list_element_opt()
-        .vortex_expect("not a list dtype");
+    let Some(element_dtype) = vx_dtype::as_ref(dtype).as_list_element_opt() else {
```
0ax1, line 221: Same question here.

vortex-ffi/src/session.rs
```diff
@@ -24,3 +24,12 @@ pub unsafe extern "C-unwind" fn vx_session_new() -> *mut vx_session {
         VortexSession::default().with_handle(RUNTIME.handle()),
     ))
 }
+
+#[unsafe(no_mangle)]
```
0ax1: Idiomatic is having `#[unsafe(no_mangle)]` under the comment.

vortex-ffi/src/session.rs
```diff
@@ -24,3 +24,12 @@ pub unsafe extern "C-unwind" fn vx_session_new() -> *mut vx_session {
...
+/// Clone a Vortex session, returning an owned copy.
+///
+/// The caller is responsible for freeing the session with [`vx_session_free`].
+pub unsafe extern "C-unwind" fn vx_session_clone(session: *mut vx_session) -> *mut vx_session {
```
0ax1: Session does not need to be mut here I think. Something along the lines of:
```
pub unsafe extern "C-unwind" fn vx_session_clone(session: *const vx_session) -> *mut vx_session {
    let session = vx_session::as_ref(session);
    vx_session::new(Box::new(session.clone()))
}
```
should work.

vortex-ffi/test/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+include(CTest)
+include(FetchContent)
+
+# TODO submodule may be cleaner since we want same testing for vortex-cxx
```
0ax1: I think `FetchContent_Declare` is the right choice, might as well drop the comment.

vortex-ffi/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wextra -Wpedantic")
+
+option(BUILD_TESTING "Build tests" ON)
+
+if (NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Debug)
+endif()
+
+string(TOLOWER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_LOWER)
+set(LIBRARY_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../target/${CMAKE_BUILD_TYPE_LOWER}")
```
0ax1: Does this actually still work if the C++ target is built with reldebug or more general with modes that are not `debug` or `release`?

vortex-ffi/README.md
```diff
@@ -40,3 +40,18 @@ Stable builds use the checked-in header file at `cinclude/vortex.h`.
 - **For header changes**: Use nightly toolchain to regenerate headers after modifying FFI code
 - **For regular development**: Stable toolchain builds work with existing checked-in headers
 - **CI validation**: Automated checks verify header freshness using nightly toolchain
+
+### Testing
+
+Build the test library
+
+```
+cmake -Bbuild
```
0ax1, line 49: Should the example include the parallel flag?

vortex-ffi/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+
+if (NOT EXISTS "${LIBRARY_PATH}")
+    message(FATAL_ERROR "FFI library not found")
+endif()
+
+add_library(vortex_ffi SHARED IMPORTED)
+set_target_properties(vortex_ffi PROPERTIES
+    IMPORTED_LOCATION "${LIBRARY_PATH}"
+    INTERFACE_INCLUDE_DIRECTORIES "${LIBRARY_HEADERS}"
+    INTERFACE_LINK_OPTIONS "LINKER:-rpath,${LIBRARY_DIR}"
```
0ax1, line 71: Did you consider linking statically, then we could get rid of `INTERFACE_LINK_OPTIONS "LINKER:-rpath,${LIBRARY_DIR}`? Or do we deliberately want to exercise the dynamic lib flow?

.github/workflows/ci.yml
myrrc: Cmake has different backends which have different defaults. If we use Ninja, the parallel build is the default, but I don't think we should require ninja

.github/workflows/ci.yml
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: No, because vx_session_free (box_wrapper!) requires a non-const vx_session. We can make it an Arc but I doubt it's a better choice.

vortex-ffi/src/session.rs
myrrc: +1

vortex-ffi/README.md
myrrc: +1

vortex-ffi/CMakeLists.txt
myrrc: I think for testing we want the dynamic library as it's faster, but we do want to expose .a/.lib, I've added both libraries now.

vortex-ffi/src/dtype.rs
myrrc: We both return null and set the error. I don't want to change api signatures but I don't also want to panic. We can change this later to use the error out parameter.

vortex-ffi/CMakeLists.txt
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: Ok, the input argument may be const, that's right

vortex-ffi/test/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+
+FetchContent_Declare(
+  Catch
+  GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+  GIT_TAG        v3.8.1
+)
+FetchContent_MakeAvailable(Catch)
+include(Catch)
+
+file(GLOB TEST_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
```
0ax1, line 14: We should enable: `wall wextra wpendatic`. To be consistent with the rest of vortex.

vortex-ffi/src/session.rs
0ax1: _pretty please_

vortex-ffi/test/CMakeLists.txt
0ax1: ignore this is enabled on top level cmake

vortex-ffi/test/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+        vx_struct_fields_builder_add_field(builder, name, dtype);
+    }
+    const vx_struct_fields *fields = vx_struct_fields_builder_finalize(builder);
+    REQUIRE(fields != nullptr);
+
+    const size_t len = vx_struct_fields_nfields(fields);
+    CHECK(len == STRUCT_LEN);
+    for (size_t i = 0; i < len; ++i) {
+        const vx_string *name = vx_struct_fields_field_name(fields, i);
+        const vx_dtype *dtype = vx_struct_fields_field_dtype(fields, i);
```
0ax1, line 81: Does the memory returned from `vx_struct_fields_field_dtype` & `vx_struct_fields_field_name` leak?

myrrc: No, because both are borrowed references

myrrc: Actually, I'm wrong, dtype is owned

0ax1 (approved):
lgtm after the last comments are addressed

PR #7130, opened 2026-03-23
New Parquet-Variant encoding

AdamGS:
## Summary

This PR introduces a new encoding to support Arrow's canonical extension type, but does not integrate it with anything else.

It does include basic pieces like slice/take/filter, and it can (or at least should) roundtrip with the equivalent arrow array.

## Testing

The encoding includes a bunch of basic tests, both for making sure it roundtrips with arrow and for the various nullability cases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ members = [
     "vortex-python",
     "vortex-tui",
     "vortex-web/crate",
+    "vortex-sqllogictest",
     "vortex-test/compat-gen",
     "vortex-test/e2e-cuda",
     "xtask",
@@ -49,13 +50,13 @@ members = [
     "encodings/zigzag",
     "encodings/zstd",
     "encodings/bytebool",
+    "encodings/parquet-variant",
     # Benchmarks
     "benchmarks/lance-bench",
     "benchmarks/compress-bench",
     "benchmarks/datafusion-bench",
     "benchmarks/duckdb-bench",
     "benchmarks/random-access-bench",
-    "vortex-sqllogictest",
 ]
 exclude = ["java/testfiles", "wasm-test"]
 resolver = "2"
@@ -184,6 +185,8 @@ opentelemetry-otlp = "0.31.0"
 opentelemetry_sdk = "0.31.0"
 parking_lot = { version = "0.12.3", features = ["nightly"] }
 parquet = "58"
+parquet-variant = "58"
+parquet-variant-compute = "58"
 paste = "1.0.15"
 pco = "1.0.1"
 pin-project-lite = "0.2.15"
--- a/encodings/parquet-variant/Cargo.toml
+++ b/encodings/parquet-variant/Cargo.toml
@@ -1,0 +1,40 @@
+[package]
+name = "vortex-parquet-variant"
+authors = { workspace = true }
+categories = { workspace = true }
+description = "Vortex Parquet Variant array"
+edition = { workspace = true }
+homepage = { workspace = true }
+include = { workspace = true }
+keywords = { workspace = true }
+license = { workspace = true }
+readme = { workspace = true }
+repository = { workspace = true }
+rust-version = { workspace = true }
+version = { workspace = true }
+publish = false
+
+[lints]
+workspace = true
+
+[dependencies]
+arrow-array = { workspace = true }
+arrow-buffer = { workspace = true }
+arrow-schema = { workspace = true }
+chrono = { workspace = true }
+parquet-variant = { workspace = true }
+parquet-variant-compute = { workspace = true }
+prost = { workspace = true }
+vortex-array = { workspace = true }
+vortex-buffer = { workspace = true }
+vortex-error = { workspace = true }
+vortex-mask = { workspace = true }
+vortex-proto = { workspace = true }
+vortex-session = { workspace = true }
+
+[dev-dependencies]
+rstest = { workspace = true }
+vortex-array = { workspace = true, features = ["_test-harness"] }
+
+[package.metadata.cargo-machete]
+ignored = ["getrandom_v03"]
--- a/encodings/parquet-variant/src/array.rs
+++ b/encodings/parquet-variant/src/array.rs
@@ -20,6 +20,7 @@ use vortex_array::validity::Validity;
 use vortex_buffer::BitBuffer;
 use vortex_error::VortexResult;
 use vortex_error::vortex_ensure;
+use vortex_error::vortex_ensure_eq;
 use vortex_mask::Mask;
 
 /// Array storage for Arrow's canonical `arrow.parquet.variant` extension type.
@@ -59,17 +60,8 @@ pub struct ParquetVariantArray {
 }
 
 impl ParquetVariantArray {
-    /// Creates a Parquet Variant array that follows the canonical extension storage layout.
-    pub fn try_new(
-        metadata: ArrayRef,
-        value: Option<ArrayRef>,
-        typed_value: Option<ArrayRef>,
-    ) -> VortexResult<Self> {
-        Self::try_new_with_validity(Validity::AllValid, metadata, value, typed_value)
-    }
-
     /// Creates a Parquet Variant array with explicit outer validity.
-    pub fn try_new_with_validity(
+    pub fn try_new(
         validity: Validity,
         metadata: ArrayRef,
         value: Option<ArrayRef>,
@@ -81,17 +73,19 @@ impl ParquetVariantArray {
         );
         let len = metadata.len();
         if let Some(validity_len) = validity.maybe_len() {
-            vortex_ensure!(
-                validity_len == len,
+            vortex_ensure_eq!(
+                validity_len,
+                len,
                 "validity length must match metadata length"
             );
         }
         if let Some(ref v) = value {
-            vortex_ensure!(v.len() == len, "value length must match metadata length");
+            vortex_ensure_eq!(v.len(), len, "value length must match metadata length");
         }
         if let Some(ref tv) = typed_value {
-            vortex_ensure!(
-                tv.len() == len,
+            vortex_ensure_eq!(
+                tv.len(),
+                len,
                 "typed_value length must match metadata length"
             );
         }
@@ -161,8 +155,7 @@ impl ParquetVariantArray {
             .map(|tv| ArrayRef::from_arrow(tv.as_ref(), typed_value_nullable))
             .transpose()?;
 
-        let pv =
-            ParquetVariantArray::try_new_with_validity(validity, metadata, value, typed_value)?;
+        let pv = ParquetVariantArray::try_new(validity, metadata, value, typed_value)?;
         Ok(VariantArray::new(pv.into_array()).into_array())
     }
 
@@ -243,6 +236,7 @@ mod tests {
     use vortex_array::arrays::Variant;
     use vortex_array::dtype::DType;
     use vortex_array::dtype::Nullability;
+    use vortex_array::validity::Validity;
     use vortex_buffer::buffer;
     use vortex_error::VortexResult;
 
@@ -356,7 +350,8 @@ mod tests {
     fn test_to_arrow_basic() -> VortexResult<()> {
         let metadata = VarBinViewArray::from_iter_bin([b"\x01\x00", b"\x01\x00"]).into_array();
         let value = VarBinViewArray::from_iter_bin([b"\x10", b"\x11"]).into_array();
-        let pv_array = ParquetVariantArray::try_new(metadata, Some(value), None)?;
+        let pv_array =
+            ParquetVariantArray::try_new(Validity::NonNullable, metadata, Some(value), None)?;
 
         let mut ctx = LEGACY_SESSION.create_execution_ctx();
         let variant_arr = pv_array.to_arrow(&mut ctx)?;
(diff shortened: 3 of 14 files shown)


encodings/parquet-variant/src/lib.rs
```diff
@@ -0,0 +1,1135 @@
...
+            )));
+            arrays.push(tv_arrow);
+        }
+
+        let struct_array = arrow_array::StructArray::try_new(Fields::from(fields), arrays, nulls)?;
+        Ok(ArrowVariantArray::try_new(&struct_array)?)
+    }
+}
+
+impl VTable for ParquetVariant {
```
a10y: it would be nice to break this file out into the folder structure the other encodings use

Cargo.toml
```diff
@@ -182,7 +183,9 @@ opentelemetry = "0.31.0"
 opentelemetry-otlp = "0.31.0"
 opentelemetry_sdk = "0.31.0"
 parking_lot = { version = "0.12.3", features = ["nightly"] }
-parquet = "57.1"
+parquet = "57.2"
+parquet-variant = "57.2"
```
a10y: > "This crate is under active development and is not yet ready for production use."

How real is this?

a10y:
It looks like this supports one `typed_value` per column. How would you represent something like this example, which shreds out multiple nested columns from an object?

image

Physically I think this differs from shredding out a `struct` column with two fields.

AdamGS:
I think in that case it is a struct, because they are named

Cargo.toml
AdamGS: honestly - unclear to me? Its definitely not complete, but it is correct in the things it exposes (as far as I can tell). There's a large issue somewhere with all the missing pieces, I might tackle some of them if we hit any bumps

encodings/parquet-variant/src/lib.rs
AdamGS: done

AdamGS:
`typed_value` will be a struct (I think that's "object" in variant terms), but not directly of the `event_type` and `event_ts` types, but of ParquetVariant arrays that, but roughly:
```
Struct {
    "event_ts": Variant {
        ..,
        typed_value: Timestamp
    },
    "event_type": Variant {
        ..,
        typed_value: String
    },
}
```

encodings/parquet-variant/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+arrow-array = { workspace = true }
+arrow-buffer = { workspace = true }
+arrow-schema = { workspace = true }
+chrono = { workspace = true }
+parquet-variant = { workspace = true }
```
gatesn, line 25: ❤️

encodings/parquet-variant/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Vortex support for Arrow's canonical `arrow.parquet.variant` extension type.
```
gatesn, line 4: Maybe worth screaming that this layout is experimental?

We could even add an `is_experimental()` flag to the VTable and just bail unless some env var is set?

AdamGS: currently not even published, but we should definitely make it hard to use for a while after we merge

robert3005: you can use `unsatble_encodings` feature flag

AdamGS: is there a `really_unstable_encoding`?

Cargo.toml
```diff
@@ -32,6 +32,7 @@ members = [
     "vortex-jni",
     "vortex-python",
     "vortex-tui",
+    "vortex-sqllogictest",
```
AdamGS, line 36: just moving things around

.github/workflows/ci.yml
```diff
@@ -202,7 +202,7 @@ jobs:
             target: wasm32-unknown-unknown
             env:
               rustflags: "RUSTFLAGS='-A warnings --cfg getrandom_backend=\"wasm_js\"'"
-            args: "--target wasm32-unknown-unknown --exclude vortex --exclude vortex-cuda --exclude vortex-cub --exclude vortex-nvcomp --exclude vortex-datafusion --exclude vortex-duckdb --exclude vortex-tui --exclude vortex-zstd --exclude vortex-test-e2e-cuda --exclude vortex-sqllogictest"
+            args: "--target wasm32-unknown-unknown --exclude vortex --exclude vortex-cuda --exclude vortex-cub --exclude vortex-nvcomp --exclude vortex-datafusion --exclude vortex-duckdb --exclude vortex-tui --exclude vortex-zstd --exclude vortex-test-e2e-cuda --exclude vortex-sqllogictest --exclude vortex-parquet-variant"
```
gatesn: Why?

AdamGS: at some point it broke it, let me check again

AdamGS: might've imagined it or maybe it something else, seems good now 🥳

encodings/parquet-variant/src/array.rs
```diff
@@ -0,0 +1,458 @@
...
+    pub fn try_new(
+        metadata: ArrayRef,
+        value: Option<ArrayRef>,
+        typed_value: Option<ArrayRef>,
+    ) -> VortexResult<Self> {
+        Self::try_new_with_validity(Validity::AllValid, metadata, value, typed_value)
+    }
+
+    /// Creates a Parquet Variant array with explicit outer validity.
+    pub fn try_new_with_validity(
```
connortsui20: I feel like this should just be `try_new`? That would be consistent with a lot of our other canonical array constructors. And if you want to be extra consistent you can pull out a `validate` method, but that's not strictly necessary.

encodings/parquet-variant/src/array.rs
```diff
@@ -0,0 +1,458 @@
...
+        }
+        if let Some(ref v) = value {
+            vortex_ensure!(v.len() == len, "value length must match metadata length");
+        }
+        if let Some(ref tv) = typed_value {
+            vortex_ensure!(
+                tv.len() == len,
+                "typed_value length must match metadata length"
+            );
+        }
```
connortsui20, line 90: `vortex_ensure_eq!`

We really should just plumb through the codebase and replace all instances of these...

encodings/parquet-variant/src/array.rs
```diff
@@ -0,0 +1,458 @@
...
+        let validity = arrow_variant
+            .nulls()
+            .map(|nulls| {
+                if nulls.null_count() == nulls.len() {
+                    Validity::AllInvalid
+                } else {
+                    Validity::from(BitBuffer::from(nulls.inner().clone()))
+                }
+            })
+            .unwrap_or(Validity::NonNullable);
```
connortsui20, line 143: Is this guaranteed on the arrow side that `None` implies non-nullable? I feel like I remember running some oddness around this before...

Maybe worth a comment if this is totally fine

Edit: Actually looking down below it seems like both `Validity::NonNullable | Validity::AllValid => None` so I'm not actually sure what to do here?

Edit again: We have this code in `vortex-array/src/arrow/convert.rs`, I think we **must** pass nullability into `from_arrow_variant` to make this correct?

```rust
fn nulls(nulls: Option<&NullBuffer>, nullable: bool) -> Validity {
    if nullable {
        nulls
            .map(|nulls| {
                if nulls.null_count() == nulls.len() {
                    Validity::AllInvalid
                } else {
                    Validity::from(BitBuffer::from(nulls.inner().clone()))
                }
            })
            .unwrap_or_else(|| Validity::AllValid)
    } else {
        assert!(nulls.map(|x| x.null_count() == 0).unwrap_or(true));
        Validity::NonNullable
    }
}
```

encodings/parquet-variant/src/array.rs
```diff
@@ -0,0 +1,458 @@
...
+            .map(|v| ArrayRef::from_arrow(v as &dyn ArrowArray, value_nullable))
+            .transpose()?;
+
+        let typed_value = arrow_variant
+            .typed_value_field()
+            .map(|tv| ArrayRef::from_arrow(tv.as_ref(), typed_value_nullable))
+            .transpose()?;
+
+        let pv =
+            ParquetVariantArray::try_new_with_validity(validity, metadata, value, typed_value)?;
```
connortsui20: Yeah I think it makes sense to just have `try_new` that takes validity, and also we could consider having a `new_unchecked` that does no validation here instead since we trust that the arrow array is well-formed (lengths are correct).

encodings/parquet-variant/src/array.rs
```diff
@@ -0,0 +1,458 @@
...
+                let mask = array.clone().execute::<Mask>(ctx)?;
+                match mask {
+                    Mask::AllTrue(_) => None,
+                    Mask::AllFalse(l) => Some(NullBuffer::new_null(l)),
+                    Mask::Values(values) => Some(NullBuffer::from(
+                        arrow_buffer::BooleanBuffer::from(values.bit_buffer().clone()),
+                    )),
+                }
+            }
+        };
```
connortsui20: We have `use vortex::array::arrow::executor::validity::to_arrow_null_buffer;` for this which is `pub(super)` right now but arguably should be `pub`.

encodings/parquet-variant/src/kernel.rs
```diff
@@ -0,0 +1,268 @@
...
+            .map(|v| v.slice(range.clone()))
+            .transpose()?;
+        let typed_value = array
+            .typed_value
+            .as_ref()
+            .map(|tv| tv.slice(range))
+            .transpose()?;
+        Ok(Some(
+            ParquetVariantArray::try_new_with_validity(validity, metadata, value, typed_value)?
+                .into_array(),
```
connortsui20: Do we care about an unchecked constructor? If we're just validating lengths it's probably not worth it...

encodings/parquet-variant/src/vtable.rs
```diff
@@ -0,0 +1,467 @@
...
+            }
+            (None, None) => {}
+            _ => return false,
+        }
+        match (&array.typed_value, &other.typed_value) {
+            (Some(a), Some(b)) => a.array_eq(b, precision),
+            (None, None) => true,
+            _ => false,
+        }
+    }
```
connortsui20, line 146: Do we want to do array equality via structural equality or value equality? Since we can have different representations for the same data we might want to just do a loop of `scalar_at` in here (after the validity and metadata checks).

encodings/parquet-variant/src/operations.rs
```diff
@@ -0,0 +1,701 @@
...
+        Scalar::try_new(
+            array.dtype.clone(),
+            Some(ScalarValue::Variant(Box::new(inner))),
+        )
+    }
+}
+
+fn scalar_from_variant_storage(
+    metadata: &[u8],
+    value: Option<&vortex_array::ArrayRef>,
```
connortsui20: import?

encodings/parquet-variant/src/array.rs
```diff
@@ -0,0 +1,458 @@
...
+///
+/// 3. **Typed-value child nullability**: the `typed_value` child carries its own `DType`
+///    (which includes nullability).
+#[derive(Clone, Debug)]
+pub struct ParquetVariantArray {
+    pub(crate) dtype: DType,
+    pub(crate) validity: Validity,
+    pub(crate) metadata: ArrayRef,
+    pub(crate) value: Option<ArrayRef>,
+    pub(crate) typed_value: Option<ArrayRef>,
```
connortsui20, line 57: Crazy idea after a quick pass over everything, is it worth having a dedicated enum that enforces at least 1 of these exist? There's a lot of logic that has to check for that everywhere (including in some free standing functions). Maybe that's overkill though and introduces more complexity than it is worth?

connortsui20 (commented):
Looks good! I gave it a quick pass but haven't looked at the scalar operations yet

encodings/parquet-variant/src/array.rs
AdamGS: Definitely a good idea

encodings/parquet-variant/src/array.rs
AdamGS: will take a look

encodings/parquet-variant/src/array.rs
AdamGS: I get that its nicer and I would change it here, but if there's no tooling for that we just have to accept its going to slip through here and there 🤷

encodings/parquet-variant/src/array.rs
AdamGS: done

encodings/parquet-variant/src/array.rs
AdamGS: done

encodings/parquet-variant/src/array.rs
AdamGS: I was re-reading the arrow implementation, and I think all four states are actually valid there. Both will be missing when you declare a shredding schema with a field that doesn't exist it the actual data.

AdamGS: They acknowledge that its not a strict adherence to the spec, not sure what to do about this for now.

gatesn (approved):

encodings/parquet-variant/src/vtable.rs
AdamGS: honestly not sure, is this really useable outside of tests?

PR #7135, opened 2026-03-24
Convert offsets into subtraction in RunEnd and Patches

robert3005:
Offsets were required because we didn't have a good system for optimising and
reducing arrays, now that we do we only need to keep them in fastlanes arrays


robert3005:
This is not ready

PR #7139, opened 2026-03-24
skip[fuzz]: remove undefined binary name

joseph-isaacs:
Use a single binary with a different workflow name and feature flags.


.github/workflows/run-fuzzer.yml
```diff
@@ -4,9 +4,14 @@ on:
...
       fuzz_target:
-        description: "The cargo fuzz target name (e.g., file_io, array_ops, compress_roundtrip)"
+        description: "The cargo fuzz binary target name (e.g., file_io, array_ops)"
         required: true
         type: string
+      fuzz_name:
+        description: "Display/storage name for this fuzz run. Defaults to fuzz_target if not set."
+        required: false
+        type: string
+        default: ""
```
0ax1, line 14: Why do we need default?

0ax1 (approved):

0ax1:
One q, disabled auto-merge.

.github/workflows/run-fuzzer.yml
joseph-isaacs: this should default the target

PR #7141, opened 2026-03-24
fix fuzz_report script

a10y:
Use the fuzz_target instead of the fuzz_name for the templated commands in the the fuzzer issues.

Before it would use the name as the target, e.g. array_ops_unstable_encodings which when you copy-paste would fail due to no such fuzzer target.

## Summary

Closes: #000

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -38,7 +38,6 @@ jobs:
     uses: ./.github/workflows/report-fuzz-crash.yml
     with:
       fuzz_target: file_io
-      fuzz_name: file_io
       crash_file: ${{ needs.io_fuzz.outputs.first_crash_name }}
       artifact_url: ${{ needs.io_fuzz.outputs.artifact_url }}
       artifact_name: file_io-crash-artifacts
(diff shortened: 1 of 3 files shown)


.github/workflows/fuzz.yml
```diff
@@ -38,6 +38,7 @@ jobs:
     uses: ./.github/workflows/report-fuzz-crash.yml
     with:
       fuzz_target: file_io
+      fuzz_name: file_io
```
joseph-isaacs: why?

.github/workflows/report-fuzz-crash.yml
```diff
@@ -6,6 +6,9 @@ on:
       fuzz_target:
         required: true
         type: string
+      fuzz_name:
+        required: true
+        type: string
```
joseph-isaacs, line 11: can you make this match the other workflow. I think think should override not be required

joseph-isaacs (changes requested):

.github/workflows/report-fuzz-crash.yml
joseph-isaacs: I don't mind what we end with just have these match please.

a10y: what do you mean? this already matches the naming of the parameter in run-fuzzer.yml

a10y: ah, you meant required. fixed

joseph-isaacs (approved):

Issue #6535, opened 2026-02-16
Vortex CUDA support


a10y:
Support for executing Vortex file scans in CUDA

- [ ] Filter
- [ ] ZSTD
- [ ] Pipelined AoT kernel
- [ ] Patches

a10y:
Logging some updates from last week:

1. We have some explicitly fused kernels, specifically FOR + BP merged with #6588
2. We have the initial prototype of CUDA-side dynamic dispatch merged with #6636 

Goals for this week in CUDA land:

* Implement a new `PatchesArray` that uses the lane-wise layout instead of the sorted CPU layout
* ?

PR #7145, opened 2026-03-24
[wip] PatchedArray wiring: Part 2

a10y:
## Summary

Part of #6535 epic and successor to #7002 

This wires PatchedArray through to BitPackedArray, replacing the interior Patches entirely. This change is backward compatible.

## API Changes

1. Updates `VTable::build` to return `ArrayRef` instead of `V::Array`. This allows us to lazily deserialize a previously written value as a new type in a backward-compatible way
1. Adds a new `LazyPatchedArray` which is just a dummy node that holds on to an array and some `Patches`, has no reductions and can just be executed into a transposed `PatchedArray`
1. Removes the `patches` child from `BitPackedArray`
1. Removes the `BitPackedArray::encode()` and a mess of other functions, instead moving everything into a new `BitPackedEncoder` fluent API

## Testing

Reuses our existing and extensive test suite for BP


vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -418,928 +418,928 @@ impl MaterializedPlan {
     }
 }
 
-#[cfg(test)]
```
a10y, line 421: todo: add these back

a10y:
With current state of branch and PatchedArray, it looks like externalizing the patches adds ~1-2% to most of the ClickBench files.

image


At lot of that is due to wasted space in the `lane_offsets` where there are many leading zeros.

That would compress easily. The problem with compressing these is that I think we then lose ability to decode in a single kernel launch.

a10y:
We _could_ add a LeadingZeros encoding which has a GPU implementation that does not launch any kernels, just orchestrates a calloc + memcpy on the GPU.

EDIT: this doesn't actually save a kernel, under the hood cudaMemset is just doing a kernel launch apparently.

encodings/fastlanes/src/bitpacking/vtable/mod.rs
```diff
@@ -313,6 +286,14 @@ impl VTable for BitPacked {
     ) -> VortexResult<Option<ArrayRef>> {
         PARENT_KERNELS.execute(array, parent, child_idx, ctx)
     }
+
```
joseph-isaacs: you could add a execute_parent kernel that fuses BitPacked decompress and patch application

joseph-isaacs (commented):
We could do one array at a time?

a10y:
Closing this in favor of multiple PRs

PR #7147, opened 2026-03-24
fix: remove nested `block_on` call in `cuda_execute`

0ax1:
`cuda_execute` itself is async so  we can't run a nested `block_on`. In the grand scheme of things, the mechanisms to allocate CUDA pinned host and device memory, as well as copying host memory to the device need to be reworked as outlined in the added TODO comment.

### DIFF
--- a/vortex-cuda/src/dynamic_dispatch/plan_builder.rs
+++ b/vortex-cuda/src/dynamic_dispatch/plan_builder.rs
@@ -9,7 +9,6 @@
 
 use std::sync::Arc;
 
-use futures::executor::block_on;
 use vortex::array::ArrayRef;
 use vortex::array::DynArray;
 use vortex::array::ExecutionCtx;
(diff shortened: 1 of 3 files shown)


vortex-cuda/src/dynamic_dispatch/plan_builder.rs
joseph-isaacs, line 514: this is a scalar cast?

0ax1: should be cleaned up, and i'll revert, as its unrelated to this pr.

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -490,14 +500,26 @@ impl PlanBuilderState<'_> {
 }
 
 /// Extract a FoR reference scalar as u64 bits.
```
joseph-isaacs, line 502: do we reinterp this in cuda too?

0ax1: all signed types are mapped to unsigned, see the docs

0ax1:
```
    /// Allocate output, upload the plan to the device, and launch the
    /// `dynamic_dispatch` kernel.
    ///
    /// The CUDA kernels are instantiated for unsigned types only.
    /// Encoding transforms (FoR, ZigZag, ALP) are bit-identical
    /// regardless of signedness.
    ///
    /// `CudaSlice::drop` enqueues `free` on the stream after kernel execution.
    pub fn execute(
        self,
        output_ptype: PType,
        len: usize,
        device_buffers: Vec<BufferHandle>,
        ctx: &mut CudaExecutionCtx,
    ) -> VortexResult<Canonical> {
        let unsigned_ptype = match output_ptype {
            PType::U8 | PType::I8 => PType::U8,
            PType::U16 | PType::I16 => PType::U16,
            PType::U32 | PType::I32 | PType::F32 => PType::U32,
            PType::U64 | PType::I64 => PType::U64,
            other => vortex_bail!("dynamic dispatch does not support PType {:?}", other),
        };
        match_each_unsigned_integer_ptype!(unsigned_ptype, |T| {
            self.execute_typed::<T>(output_ptype, len, device_buffers, ctx)
        })
    }
```
Obviously, this is a simplification to get things off the ground.

0ax1: > do we reinterp this in cuda too?

So the reinterpretation happens by calling the respective unsigned kernel.

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
0ax1: dropped the change

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
0ax1: I'll also need to add support for mixed types, e.g. diff types for dict values and codes.

robert3005 (approved):

PR #7148, opened 2026-03-24
ffi: expose array validity, allow creating primitive arrays

myrrc:
- Allow querying array validity from FFI.
- Allow creating null and primitive vx_arrays from FFI.
- Expose utility functions to query dtype and ptype from array
  directly instead of array -> dtype -> variant [-> ptype] chain.

Breaking changes:
- rename `vx_array_is_null` to `vx_array_element_is_invalid`.
- rename `vx_array_null_count` to `vx_array_invalid_count`.
- make functions operating on indices take size_t instead of u32.

### DIFF
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -19,6 +19,52 @@
 #define BinaryView_MAX_INLINED_SIZE 12
 
 /**
+ * The variant tag for a Vortex data type.
+ */
+typedef enum {
+    /**
+     * Null type.
+     */
+    DTYPE_NULL = 0,
+    /**
+     * Boolean type.
+     */
+    DTYPE_BOOL = 1,
+    /**
+     * Primitive types (e.g., u8, i16, f32, etc.).
+     */
+    DTYPE_PRIMITIVE = 2,
+    /**
+     * Variable-length UTF-8 string type.
+     */
+    DTYPE_UTF8 = 3,
+    /**
+     * Variable-length binary data type.
+     */
+    DTYPE_BINARY = 4,
+    /**
+     * Nested struct type.
+     */
+    DTYPE_STRUCT = 5,
+    /**
+     * Nested list type.
+     */
+    DTYPE_LIST = 6,
+    /**
+     * User-defined extension type.
+     */
+    DTYPE_EXTENSION = 7,
+    /**
+     * Decimal type with fixed precision and scale.
+     */
+    DTYPE_DECIMAL = 8,
+    /**
+     * Nested fixed-size list type.
+     */
+    DTYPE_FIXED_SIZE_LIST = 9,
+} vx_dtype_variant;
+
+/**
  * Variant enum for Vortex primitive types.
  */
 typedef enum {
@@ -68,51 +114,25 @@ typedef enum {
     PTYPE_F64 = 10,
 } vx_ptype;
 
-/**
- * The variant tag for a Vortex data type.
- */
 typedef enum {
     /**
-     * Null type.
+     * Items can't be null
      */
-    DTYPE_NULL = 0,
+    VX_VALIDITY_NON_NULLABLE = 0,
     /**
-     * Boolean type.
+     * All items are valid
      */
-    DTYPE_BOOL = 1,
+    VX_VALIDITY_ALL_VALID = 1,
     /**
-     * Primitive types (e.g., u8, i16, f32, etc.).
+     * All items are invalid
      */
-    DTYPE_PRIMITIVE = 2,
+    VX_VALIDITY_ALL_INVALID = 2,
     /**
-     * Variable-length UTF-8 string type.
+     * Items validity is determined by a boolean array. True values in boolean
+     * array are valid, false values are invalid (null)
      */
-    DTYPE_UTF8 = 3,
-    /**
-     * Variable-length binary data type.
-     */
-    DTYPE_BINARY = 4,
-    /**
-     * Nested struct type.
-     */
-    DTYPE_STRUCT = 5,
-    /**
-     * Nested list type.
-     */
-    DTYPE_LIST = 6,
-    /**
-     * User-defined extension type.
-     */
-    DTYPE_EXTENSION = 7,
-    /**
-     * Decimal type with fixed precision and scale.
-     */
-    DTYPE_DECIMAL = 8,
-    /**
-     * Nested fixed-size list type.
-     */
-    DTYPE_FIXED_SIZE_LIST = 9,
-} vx_dtype_variant;
+    VX_VALIDITY_ARRAY = 3,
+} vx_validity_type;
 
 /**
  * Equalities, inequalities, and boolean operations over possibly null values.
@@ -296,15 +316,20 @@ typedef struct Nullability Nullability;
 typedef struct Primitive Primitive;
 
 /**
- * Base type for all Vortex arrays.
+ * Arrays are reference-counted handles to owned memory buffers that hold
+ * scalars. These buffers can be held in a number of physical encodings to
+ * perform lightweight compression that exploits the particular data
+ * distribution of the array's values.
  *
- * All built-in Vortex array types can be safely cast to this type to pass into functions that
- * expect a generic array type. e.g.
+ * Every data type recognized by Vortex also has a canonical physical
+ * encoding format, which arrays can be canonicalized into for ease of
+ * access in compute functions.
  *
- * ```cpp
- * auto primitive_array = vx_array_primitive_new(...);
- * vx_array_len((*vx_array) primitive_array));
- * ```
+ * As an implementation detail, vx_array Arc'ed inside, so cloning an
+ * array is a cheap operation.
+ *
+ * Unless stated explicitly, all operations with vx_array don't take
+ * ownership of it, and thus it must be freed by the caller.
  */
 typedef struct vx_array vx_array;
 
@@ -398,6 +423,16 @@ typedef struct vx_struct_fields vx_struct_fields;
  */
 typedef struct vx_struct_fields_builder vx_struct_fields_builder;
 
+typedef struct {
+    vx_validity_type type;
+    /**
+     * If type is not VX_VALIDITY_ARRAY, this is NULL.
+     * If type is VX_VALIDITY_ARRAY, this is set to an owned boolean validity
+     * array which must be freed by the caller.
+     */
+    const vx_array *array;
+} vx_validity;
+
 /**
  * Options supplied for opening a file.
  */
@@ -478,6 +513,40 @@ const vx_array *vx_array_clone(const vx_array *ptr);
 void vx_array_free(const vx_array *ptr);
 
 /**
+ * Check if array's dtype is nullable.
+ * As a particular example, a Null array is nullable.
+ */
+bool vx_array_is_nullable(const vx_array *array);
+
+/**
+ * Check array's dtype against a variant.
+ * Equivalent to vx_get_dtype_variant(vx_array_dtype(array)).
+ *
+ * Example:
+ *
+ * const vx_array* array = vx_array_new_null(1);
+ * assert(vx_array_has_dtype(array, DTYPE_NULL));
+ * vx_array_free(array);
+ *
+ */
+bool vx_array_has_dtype(const vx_array *array, vx_dtype_variant variant);
+
+/**
+ * Check whether array has a Primitive dtype with a specific ptype.
+ *
+ * const vx_array* array = vx_array_new_null(1);
+ * assert(!vx_array_is_primitive(array, PTYPE_U32));
+ * vx_array_free(array);
+ *
+ */
+bool vx_array_is_primitive(const vx_array *array, vx_ptype ptype);
+
+/**
+ * Return array's validity as a type and a boolean array.
+ */
+void vx_array_get_validity(const vx_array *array, vx_validity *validity, vx_error **error);
+
+/**
  * Get the length of the array.
  */
 size_t vx_array_len(const vx_array *array);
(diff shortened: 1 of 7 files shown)


vortex-ffi/src/array.rs
```diff
@@ -1,37 +1,124 @@
...
-    /// ```cpp
-    /// auto primitive_array = vx_array_primitive_new(...);
-    /// vx_array_len((*vx_array) primitive_array));
-    /// ```
     dyn DynArray,
     vx_array
 );
 
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_array_is_nullable(array: *const vx_array) -> bool {
```
0ax1: This wasn't introduced as a pattern in this PR, but we should consider documenting in the code what `C-unwind` implies / allows for (compared to `extern "C"`), explaining how the unwinding can be leveraged.

vortex-ffi/cinclude/vortex.h
```diff
@@ -401,6 +408,20 @@ const vx_array *vx_array_clone(const vx_array *ptr);
  */
 void vx_array_free(const vx_array *ptr);
 
+bool vx_array_is_nullable(const vx_array *array);
```
0ax1, line 519: All public C API should have proper docs.

vortex-ffi/cinclude/vortex.h
```diff
@@ -401,6 +408,20 @@ const vx_array *vx_array_clone(const vx_array *ptr);
...
+
+bool vx_array_is(const vx_array *array, vx_dtype_variant variant);
```
0ax1: `vx_array_has_dtype?`

vortex-ffi/cinclude/vortex.h
```diff
@@ -418,10 +439,30 @@ const vx_array *vx_array_get_field(const vx_array *array, uint32_t index, vx_err
...
+const vx_array *vx_array_new_null(size_t len);
+
+/**
+ * Create a new primitive array from an existing buffer.
+ * It is caller's responsibility to ensure ptr points to a buffer of correct type.
+ * Buffer contents are copied.
+ * Takes ownership of validity.array if it is set.
+ * validity can't be NULL.
+ */
+const vx_array *vx_array_new_primitive(vx_ptype ptype,
```
0ax1, line 600: The array returned here is const, what if I later have a fn that wants to modify the array?

vortex-ffi/cinclude/vortex.h
```diff
@@ -418,10 +439,30 @@ const vx_array *vx_array_get_field(const vx_array *array, uint32_t index, vx_err
...
+                                       const void *ptr,
```
0ax1, line 601: The `ptr` parameter name should be more descriptive. Shall we have in the fn docs describing what each param is for, here and elsewhere?

vortex-ffi/cinclude/vortex.h
```diff
@@ -764,7 +805,8 @@ vx_array_sink *vx_array_sink_open_file(const vx_session *session,
                                        vx_error **error_out);
 
 /**
- * Pushed a single array chunk into a file sink.
+ * Push an array into a file sink.
+ * Does not take ownership of array
  */
 void vx_array_sink_push(vx_array_sink *sink, const vx_array *array, vx_error **error_out);
```
0ax1, line 1067: Same here, needs more docs.

vortex-ffi/src/array.rs
```diff
@@ -1,37 +1,124 @@
...
+    vx_array::as_ref(array).dtype().is_nullable()
+}
+
+#[unsafe(no_mangle)]
```
0ax1, line 73: docs, docs, docs

vortex-ffi/src/array.rs
```diff
@@ -1,37 +1,124 @@
...
+
+#[repr(C)]
+pub struct vx_validity {
+    r#type: vx_validity_type,
+    /// If type is VX_VALIDITY_ARRAY, this is set, NULL otherwise
+    array: *const vx_array,
+}
+
+impl From<&vx_validity> for Validity {
+    fn from(value: &vx_validity) -> Self {
```
0ax1: `value` or maybe go with `validity`?

0ax1: This would then read `Validity::Array(vx_array::as_ref(validity.array)..)`.

vortex-ffi/src/array.rs
```diff
@@ -1,37 +1,124 @@
...
+        }
+    }
+}
+
+/// Return array's validity.
+/// If validity.type is VX_VALIDITY_ARRAY, returns an owned vx_array in
+/// validity.array which must be freed by the caller.
+/// If validity.type is not VX_VALIDITY_ARRAY, validity.array is NULL.
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_array_get_validity(
```
0ax1, line 166: Is there a reason we're using `get` here in the name but not for `vx_array_len`? Would `vx_array_validity`be sufficient?

vortex-ffi/src/array.rs
```diff
@@ -1,37 +1,124 @@
...
+    array: *const vx_array,
+    validity: *mut vx_validity,
+    error: *mut *mut vx_error,
+) {
+    let array = vx_array::as_ref(array);
+    try_or_default(error, || {
```
0ax1, line 171: Can we return the validity, rather than using an output parameter?

vortex-ffi/src/array.rs
```diff
@@ -81,17 +168,19 @@ pub unsafe extern "C-unwind" fn vx_array_slice(
...
+    error: *mut *mut vx_error,
 ) -> bool {
-    let array = vx_array::as_ref(array);
-    // TODO(joe): propagate this error up instead of expecting
-    array
-        .is_invalid(index as usize)
-        .vortex_expect("is_invalid failed")
+    try_or_default(error, || {
+        vortex_ensure!(!array.is_null());
+        vx_array::as_ref(array).is_invalid(index)
```
0ax1, line 240: Rust fn says `is_invalid` but we're using `is_null`. We're using both terms in the codebase. I care less about which we use but consistency.

vortex-ffi/src/array.rs
```diff
@@ -104,9 +193,59 @@ pub unsafe extern "C-unwind" fn vx_array_null_count(
...
+    validity: &vx_validity,
+) -> *const vx_array {
+    let slice = unsafe { std::slice::from_raw_parts(ptr, len) };
+    let buffer = Buffer::copy_from(slice);
+    let array = PrimitiveArray::new(buffer, validity.into());
+    vx_array::new(array.into_array())
+}
+
+/// Create a new primitive array from an existing buffer.
+/// It is caller's responsibility to ensure ptr points to a buffer of correct type.
```
0ax1: Doc formatting looks a bit scatty, can we block format the docs? 

```
/// Create a new primitive array from an existing buffer.

/// It is caller's responsibility to ensure ptr points to a buffer of correct
/// type. Buffer contents are copied. Takes ownership of validity.array if 
/// it is set. validity can't be NULL.

File: vortex-ffi/src/array.rs [hunk 13]
```diff
@@ -104,9 +193,59 @@ pub unsafe extern "C-unwind" fn vx_array_null_count(
     try_or_default(error_out, || Ok(array.invalid_count()?.try_into()?))
 }
 
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_array_new_null(len: usize) -> *const vx_array {
+    vx_array::new(NullArray::new(len).into_array())
+}
+
+fn primitive<T: vortex::dtype::NativePType>(
```
0ax1: How sth like `new_primitive_from_raw_parts` to make the fn name more descriptive?

vortex-ffi/src/array.rs
```diff
@@ -104,9 +193,59 @@ pub unsafe extern "C-unwind" fn vx_array_null_count(
...
+    vx_array::new(array.into_array())
+}
+
+/// Create a new primitive array from an existing buffer.
+/// It is caller's responsibility to ensure ptr points to a buffer of correct type.
+/// Buffer contents are copied.
+/// Takes ownership of validity.array if it is set.
+/// validity can't be NULL.
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_array_new_primitive(
```
0ax1: In the Rust version, the buffer is moved, taking ownership, in this version we copy and create an additional buffer. Is that what we want to go for?

vortex-ffi/src/error.rs
```diff
@@ -19,6 +19,13 @@ box_wrapper!(
     vx_error
 );
 
+pub(crate) fn write_error(error_out: *mut *mut vx_error, error_string: &str) {
```
0ax1: I guess part of the contract here is that `error_out` is not set elsewhere before, otherwise the previous error is leaked.

vortex-ffi/cinclude/vortex.h
```diff
@@ -418,10 +439,30 @@ const vx_array *vx_array_get_field(const vx_array *array, uint32_t index, vx_err
...
 
-bool vx_array_is_null(const vx_array *array, uint32_t index, vx_error **_error_out);
+/**
+ * Check whether array's element at index is null.
+ * Sets error if index is out of bounds or underlying validity array is
+ * corrupted.
+ */
+bool vx_array_element_is_null(const vx_array *array, size_t index, vx_error **error);
 
 uint32_t vx_array_null_count(const vx_array *array, vx_error **error_out);
```
0ax1: Not introduced in this PR but this should be `size_t`.

vortex-ffi/test/common.h
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+#include <catch2/catch_test_macros.hpp>
```
0ax1, line 5: Missing a pragma once.

vortex-ffi/test/common.h
```diff
@@ -0,0 +1,24 @@
...
+inline std::string to_string(vx_error *err) {
+    const vx_string *msg = vx_error_get_message(err);
+    return {vx_string_ptr(msg), vx_string_len(msg)};
+}
+
+inline std::string_view to_string_view(const vx_string *msg) {
+    return {vx_string_ptr(msg), vx_string_len(msg)};
+}
+
+inline std::string_view to_string_view(vx_error *err) {
```
0ax1, line 18: Should they have different fn name prefixes? They kinda do different things no?

vortex-ffi/cinclude/vortex.h
0ax1: And shall we consider matching the Rust name `vx_array_null_count` => `..invalid_count`.

vortex-ffi/src/array.rs
```diff
@@ -104,9 +193,59 @@ pub unsafe extern "C-unwind" fn vx_array_null_count(
...
+}
+
+fn primitive<T: vortex::dtype::NativePType>(
+    ptr: *const T,
+    len: usize,
+    validity: &vx_validity,
+) -> *const vx_array {
+    let slice = unsafe { std::slice::from_raw_parts(ptr, len) };
+    let buffer = Buffer::copy_from(slice);
+    let array = PrimitiveArray::new(buffer, validity.into());
```
0ax1, line 273: In `impl From<&vx_validity> for Validity`, we clone the arc `validity::Array(vx_array::as_ref(value.array).clone())` (aka atomic ref counted pointer). 

Saying we don't take ownership of validity as mentioned in the `new_primitive` docs.

0ax1 (commented):
We need to double check the consistency of size types for indices:
```
vx_array_get_field` | `uint32_t` |
| `vx_array_slice` | `uint32_t` |
| `vx_array_get_u8` etc. | `uint32_t` |
| `vx_array_null_count` return | `uint32_t
```
should prob all be size_t.

vortex-ffi/cinclude/vortex.h
myrrc: Arrays are a Box<Arc<T>> inside so you can have mutable operations with a const vx_array*. That's not ideal but that's how our arc_dyn_wrapper! works currently, this function just matches the signature

vortex-ffi/src/array.rs
myrrc: Yes, and we also have vx_array_dtype and not vx_array_get_dtype, but
vx_array_get_u8 (for primitives) and vx_dtype_get_variant (because we have a struct vx_dtype_variant)

vortex-ffi/src/array.rs
myrrc: C APIs usually write to a pointer if a returned value is more than a machine word, and sizeof(vx_validity) is two words.

vortex-ffi/cinclude/vortex.h
0ax1: Can we leave a big TODO here to address that? This is confusing for a C programmer looking at the signature not knowing anything about the internals.

vortex-ffi/test/common.h
myrrc: They both extract a view of the error string messsage, so they do the same thing

vortex-ffi/test/common.h
myrrc: +1

vortex-ffi/src/error.rs
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: +1

vortex-ffi/test/common.h
0ax1: Fair enough, should all fns be prefixed with `vx_` in C?

vortex-ffi/src/array.rs
0ax1: I think we can be a bit more liberal returning data larger than 8 byte. In most cases the compiler will prob be able to optimize better if we return by value. I assume C also has RVO. Have a peek at: https://godbolt.org/z/4beja8K11

myrrc: I agree it isn't an issue with modern compilers but that's a common convention.

vortex-ffi/src/array.rs
myrrc: I don't think there is a way for us to move the pointer between FFI boundaries. Also, this version doesn't care about stack vs. heap buffers (these are separate entities in Rust).

vortex-ffi/src/array.rs
myrrc: This is an implementation detail, but from the user's standpoint, if the user sets the array, after calling vx_array_new_primitive, the array still must be freed manually, and not doing so would result in a leak.

vortex-ffi/test/common.h
myrrc: These functions are test-only so I don't think it's necessary.

vortex-ffi/cinclude/vortex.h
```diff
@@ -296,15 +316,20 @@ typedef struct Nullability Nullability;
 typedef struct Primitive Primitive;
 
 /**
- * Base type for all Vortex arrays.
+ * Arrays are typed views of memory buffers that hold scalars. These
```
0ax1: Views might be a bit misleading in the sense that arrays own their in memory buffers. They are ref counted handles to an owned buffer.

vortex-ffi/cinclude/vortex.h
```diff
@@ -296,15 +316,20 @@ typedef struct Nullability Nullability;
...
+ * buffers can be held in a number of physical encodings to perform
+ * lightweight compression that exploits the particular data distribution
+ * of the array's values.
+ *
+ * Every data type recognized by Vortex also has a canonical physical
```
0ax1, line 324: Appreciate you adding the comments! I do wonder though whether this might become difficult to keep in sync with the Rust ref and whether we should consider to link the rust docs? 🤷 Let's leave as is for now.

0ax1: If applicable I mean.

vortex-ffi/cinclude/vortex.h
```diff
@@ -490,57 +559,93 @@ size_t vx_array_len(const vx_array *array);
...
+int64_t vx_array_get_i64(const vx_array *array, size_t index);
 
-int64_t vx_array_get_storage_i64(const vx_array *array, uint32_t index);
+int64_t vx_array_get_storage_i64(const vx_array *array, size_t index);
 
-uint16_t vx_array_get_f16(const vx_array *array, uint32_t index);
+uint16_t vx_array_get_f16(const vx_array *array, size_t index);
 
-uint16_t vx_array_get_storage_f16(const vx_array *array, uint32_t index);
+uint16_t vx_array_get_storage_f16(const vx_array *array, size_t index);
```
0ax1, line 640: 👍

vortex-ffi/src/array.rs
```diff
@@ -1,12 +1,21 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
+#![allow(non_camel_case_types)]
```
0ax1, line 3: Should this be global?

vortex-ffi/src/array.rs
0ax1:
```
fn primitive_from_raw<T: vortex::dtype::NativePType>(
    ptr: *const T,
    len: usize,
    validity: &vx_validity,
) -> *const vx_array {
    let slice = unsafe { std::slice::from_raw_parts(ptr, len) };
    let buffer = Buffer::copy_from(slice);
    let array = PrimitiveArray::new(buffer, validity.into());
```

vortex-ffi/src/array.rs
myrrc: Yes, because there are multiple vx_ structs exposed. I can put an allow() on all of them, but is this much better?

vortex-ffi/src/error.rs
```diff
@@ -19,6 +19,15 @@ box_wrapper!(
     vx_error
 );
 
+/// Write an error message to `error` which has not been populated before.
+pub(crate) fn write_error(error: *mut *mut vx_error, message: &str) {
+    assert!(error.is_null());
```
0ax1: Did you mean !null? The outer pointer needs to be set right?

vortex-ffi/src/array.rs
```diff
@@ -72,59 +215,125 @@ pub unsafe extern "C-unwind" fn vx_array_get_field(
...
+    })
+}
+
+/// Create a new array with DTYPE_NULL dtype.
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_array_new_null(len: usize) -> *const vx_array {
+    vx_array::new(NullArray::new(len).into_array())
+}
+
+fn primitive_from_raw<T: vortex::dtype::NativePType>(
```
0ax1: The fn should be marked unsafe, as it is the callers responsibility to uphold the preconditions:
```
/// # Safety
/// `ptr` must be valid for `len` reads of `T`, properly aligned, and must not be null if `len > 0`.
unsafe fn primitive_from_raw<T: vortex::dtype::NativePType>(
    ptr: *const T,
    len: usize,
    validity: &vx_validity,
) -> *const vx_array {
    let slice = unsafe { std::slice::from_raw_parts(ptr, len) };
    let buffer = Buffer::copy_from(slice);
    let array = PrimitiveArray::new(buffer, validity.into());
    vx_array::new(array.into_array())
}
```

vortex-ffi/src/error.rs
myrrc: +1

0ax1 (approved):

PR #7153, opened 2026-03-25
fix(python): cap substrait below 0.85.0

### LINKED ISSUE #7152: Incompatible with substrait v0.85.0+
### What happened?

```
Traceback (most recent call last):
  File "/home/model/.cache/pypoetry/virtualenvs/daydaynew-pcZtFd17-py3.12/lib64/python3.12/site-packages/vortex/substrait.py", line 23, in <module>
    from substrait.extensions.extensions_pb2 import SimpleExtensionDeclaration, SimpleExtensionURI
ImportError: cannot import name 'SimpleExtensionURI' from 'substrait.extensions.extensions_pb2' (/home/model/.cache/pypoetry/virtualenvs/daydaynew-pcZtFd17-py3.12/lib/python3.12/site-packages/substrait/extensions/extensions_pb2.py). Did you mean: 'SimpleExtensionURN'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  ...
  File "/home/model/.cache/pypoetry/virtualenvs/daydaynew-pcZtFd17-py3.12/lib64/python3.12/site-packages/vortex/__init__.py", line 6, in <module>
    from . import _lib, arrays, dataset, expr, file, io, ray, registry, scan
  File "/home/model/.cache/pypoetry/virtualenvs/daydaynew-pcZtFd17-py3.12/lib64/python3.12/site-packages/vortex/dataset.py", line 18, in <module>
    from .arrow.expression import ensure_vortex_expression
  File "/home/model/.cache/pypoetry/virtualenvs/daydaynew-pcZtFd17-py3.12/lib64/python3.12/site-packages/vortex/arrow/expression.py", line 14, in <module>
    from ..substrait import extended_expression
  File "/home/model/.cache/pypoetry/virtualenvs/daydaynew-pcZtFd17-py3.12/lib64/python3.12/site-packages/vortex/substrait.py", line 27, in <module>
    from substrait.gen.proto.algebra_pb2 import Expression, FunctionArgument
ModuleNotFoundError: No module named 'substrait.gen'
```

In Substrait 0.85.0, `SimpleExtensionURI` was removed. The Vortex Python bindings have not been updated to use the new implementation, and they also do not set an appropriate upper bound on the dependency version.

See:

https://github.com/substrait-io/substrait/issues/970
https://github.com/substrait-io/substrait/releases/tag/v0.85.0
https://github.com/vortex-data/vortex/blob/f2008239f4836b927622b143217165ed1259125a/vortex-python/pyproject.toml#L9

### Steps to reproduce

1. Install vortex-data
2. Import vortex

### Environment

- Vortex 0.64.0
- Python 3.12
- Rocky Linux 9.4 (Blue Onyx)

### Additional context

_No response_

RealXuChe:
## Summary

Fixes [https://github.com/vortex-data/vortex/issues/7152](https://github.com/vortex-data/vortex/issues/7152).

In Substrait 0.85.0, `SimpleExtensionURI` was removed. The Vortex Python bindings have not been updated to use the new implementation, and they also do not set an appropriate upper bound on the dependency version.

This change caps the Python dependency at `substrait<0.85.0` until the code is updated for the URN schema.

This workaround is used because it is currently difficult to make Vortex support the latest Substrait. Vortex currently consumes Substrait expressions produced by PyArrow. However, the current Arrow tree still uses Substrait v0.44.0. Until Arrow is updated, it is difficult for us to support the latest Substrait easily because of our dependency on Arrow.

See:

PyArrow uses `v0.44.0`: https://github.com/apache/arrow/blob/f9315d4e7fb61ac85a77c651dcee84dbfad88472/cpp/thirdparty/versions.txt#L109

We convert PyArrow Expression into Substrait type: https://github.com/vortex-data/vortex/blob/f2008239f4836b927622b143217165ed1259125a/vortex-python/python/vortex/arrow/expression.py#L47-L50

## Testing

Did not run tests. This is a dependency metadata change only.

### DIFF
--- a/vortex-python/pyproject.toml
+++ b/vortex-python/pyproject.toml
@@ -6,7 +6,7 @@ dynamic = ["version", "description", "authors"]
 readme = "README.md"
 dependencies = [
     "pyarrow>=17.0.0",
-    "substrait>=0.23.0",
+    "substrait>=0.23.0,<0.85.0",
     "typing-extensions>=4.5.0",
 ]
 requires-python = ">= 3.11"


AdamGS (approved):

RealXuChe:
I found an important mistake in this PR, and I want to correct it here.

I was misled by the way the Substrait Python packaging is split. I also only did a quick check in an existing project before opening the PR, rather than testing the install in a clean environment from scratch. That rough testing caused me to miss the problem in this PR.

Most of my comments in the issue/PR are still correct:

- The reported failure is real, caused exactly by the removal of `SimpleExtensionURI`
- The removal was introduced in [Substrait specification](https://github.com/substrait-io/substrait/) 0.85.0
- PyArrow still uses an old Substrait schema, so we can't support the latest Substrait easily before PyArrow does.

However, I misunderstood how Substrait is packaged:

- The Python package `substrait` does not come from the `substrait` spec repository. It comes from the separate [Python binding repo](https://github.com/substrait-io/substrait-python)
- The Python package `substrait` has its own versioning and does not match the spec version
- [`substrait-python` `v0.29.0`](https://github.com/substrait-io/substrait-python/releases/tag/v0.29.0) bumped Substrait specification to the 0.85.0 schema

(`substrait-protobuf` and `substrait` share the same `substrait.*` import namespace, really misleading!)

So this PR was wrong. I wrote the bound as if the PyPI `substrait` package version matched the spec version, which it does not. The version upper bound we should actually set is `substrait<0.29.0`.

I have now tested the new fix in a clean virtual environment, and I confirmed that this corrected bound works.

For this issue, I will send a follow-up PR to correct the dependency version bound.

PR #7156, opened 2026-03-25
feat[vortex-datafusion]: use PartitionedFile instead of path in create_reader

asubiotto:
The motivation for this change is that we have a need to create a custom VortexReadAt implementation based on some custom values in a PartitionedFile extension.

I don't think this will break any users because the reader module was private.

### DIFF
--- a/vortex-datafusion/src/persistent/mod.rs
+++ b/vortex-datafusion/src/persistent/mod.rs
@@ -7,7 +7,7 @@ mod cache;
 mod format;
 pub mod metrics;
 mod opener;
-mod reader;
+pub mod reader;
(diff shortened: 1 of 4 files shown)


vortex-datafusion/src/persistent/mod.rs
```diff
@@ -7,7 +7,7 @@ mod cache;
 mod format;
 pub mod metrics;
 mod opener;
-mod reader;
+pub mod reader;
```
AdamGS, line 10: this is a very good catch 😂

AdamGS (approved):

PR #7157, opened 2026-03-25
Add tsan for vortex-ffi, fix asan instrumentation

myrrc:
- Add proper instrumentation for vortex-ffi (didn't check memory leaks before due to Mimalloc).
- Add memory sanitizer and thread sanitizer runtimes for Rust.
- Remove `verbosity=1` from sanitizer output as it's too noisy.
- Remove `VX_DUCKDB_DEBUG` since we don't check duckdb code in these jobs.
- Make address sanitizer fail on error to fail Github action job.
- Add cross Rust-C address and thread sanitizers.
- Fix existing sanitizer errors (memory leak in vx_sink).
- Allow building FFI C part library with Rust's target triple via `TARGET_TRIPLE` CMake option.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -379,25 +379,35 @@ jobs:
           flags: ${{ matrix.suite }}
           use_oidc: true
 
-  rust-test:
-    name: "Rust tests (sanitizer)"
-    timeout-minutes: 40
+  rust-test-sanitizer:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - sanitizer: asan
+            sanitizer_flags: "-Zsanitizer=address -Zsanitize=leak"
+          - sanitizer: msan
+            sanitizer_flags: "-Zsanitizer=memory"
+          - sanitizer: tsan
+            sanitizer_flags: "-Zsanitizer=thread"
+    name: "Rust tests (${{ matrix.sanitizer }})"
     runs-on: >-
       ${{ github.repository == 'vortex-data/vortex'
           && format('runs-on={0}/runner=amd64-medium/image=ubuntu24-full-x64-pre-v2/tag=rust-test-sanitizer', github.run_id)
           || 'ubuntu-latest' }}
+    timeout-minutes: 40
     env:
-      # Add debug symbols and enable ASAN/LSAN with better output
-      ASAN_OPTIONS: "symbolize=1:print_stats=1:check_initialization_order=1:detect_leaks=1:halt_on_error=0:verbosity=1:leak_check_at_exit=1"
-      LSAN_OPTIONS: "verbosity=1:report_objects=1"
+      ASAN_OPTIONS: "symbolize=1:check_initialization_order=1:detect_leaks=1:leak_check_at_exit=1"
+      LSAN_OPTIONS: "report_objects=1"
       ASAN_SYMBOLIZER_PATH: "/usr/bin/llvm-symbolizer"
-      # Link against DuckDB debug build
-      VX_DUCKDB_DEBUG: "1"
-      # Keep frame pointers for better stack traces
-      CARGO_PROFILE_DEV_DEBUG: "true"
-      CARGO_PROFILE_TEST_DEBUG: "true"
-      # Skip slow tests that are too expensive under sanitizer
+      MSAN_OPTIONS: "symbolize=1"
+      MSAN_SYMBOLIZER_PATH: "/usr/bin/llvm-symbolizer"
+      TSAN_OPTIONS: "symbolize=1"
+      TSAN_SYMBOLIZER_PATH: "/usr/bin/llvm-symbolizer"
       VORTEX_SKIP_SLOW_TESTS: "1"
+      # -Cunsafe-allow-abi-mismatch=sanitizer: libraries like compiler_builtins
+      # unset -Zsanitizer flag and we should allow that.
+      RUSTFLAGS: "-A warnings -Cunsafe-allow-abi-mismatch=sanitizer --cfg disable_loom --cfg vortex_nightly -C debuginfo=2 -C opt-level=0 -C strip=none"
     steps:
       - uses: runs-on/action@v2
         if: github.repository == 'vortex-data/vortex'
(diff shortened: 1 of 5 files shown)


myrrc:
@0ax1 notes we should not use Mimalloc by default in vortex-ffi, it should be opt-in.

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
           # Run tests with sanitizers and debug output
           cargo +$NIGHTLY_TOOLCHAIN nextest run \
             --locked \
             --all-features \
             --no-fail-fast \
             --target x86_64-unknown-linux-gnu \
-            -p vortex-buffer -p vortex-ffi -p vortex-fastlanes -p vortex-fsst -p vortex-alp -p vortex-array
+            -p vortex-buffer -p vortex-fastlanes -p vortex-fsst -p vortex-alp -p vortex-array
+
+          # vortex-ffi requires --no-default-features as otherwise we pull in
```
0ax1: As discussed offline, we should not enable mimalloc by default in the ffi lib.

.github/workflows/ci.yml
```diff
@@ -379,20 +379,18 @@ jobs:
           flags: ${{ matrix.suite }}
           use_oidc: true
 
-  rust-test:
-    name: "Rust tests (sanitizer)"
+  rust-test-asan:
+    name: "Rust tests (address sanitizer)"
```
0ax1: The three `name: "Rust tests (X sanitizer)` should parametrize over the type in a GH matrix. Except for the sanitizer type these roughly look the same.

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
+            -p vortex-buffer -p vortex-fastlanes -p vortex-fsst -p vortex-alp -p vortex-array
+
+          # vortex-ffi requires --no-default-features as otherwise we pull in
+          # Mimalloc which interferes with sanitizers
+          # cargo nextest reports less issues than cargo test
+          cargo +$NIGHTLY_TOOLCHAIN test --locked --no-default-features \
+            -Zbuild-std --no-fail-fast --target x86_64-unknown-linux-gnu \
+            --lib vortex-ffi -- --no-capture
+
+  rust-ffi-test-asan:
```
0ax1: Why only run the ffi tests with ASAN?

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
+          echo "Building vortex-ffi"
+          cargo +$NIGHTLY_TOOLCHAIN build --locked --no-default-features \
+            -Zbuild-std --target $triple
+
+          echo "Building C++ tests"
+          cmake -Bbuild -DWITH_ASAN=1 -DTARGET_TRIPLE=$triple
+          cmake --build build -j
+
+          echo "Running C++ tests"
+          set -o pipefail
```
0ax1, line 506: Curious whether `set -o pipefail` isn't the default for inline bash script in GH. If not move it to the top of the script.

0ax1: cargo has these runner wrapper env vars to wrap with an outer command so you don't have to guess the path: `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER`. See `cuda-test-sanitizer`.

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
+    name: "Rust/C++ ffi tests (address sanitizer)"
+    timeout-minutes: 40
+    runs-on: >-
+      ${{ github.repository == 'vortex-data/vortex'
+          && format('runs-on={0}/runner=amd64-medium/image=ubuntu24-full-x64-pre-v2/tag=rust-c++-test-address-sanitizer', github.run_id)
+          || 'ubuntu-latest' }}
+    env:
+      ASAN_OPTIONS: "symbolize=1:print_stats=1:check_initialization_order=1:detect_leaks=1:leak_check_at_exit=1"
+      LSAN_OPTIONS: "report_objects=1"
+      CARGO_PROFILE_DEV_DEBUG: "true"
```
0ax1: These shouldn't be needed or should they:
- CARGO_PROFILE_DEV_DEBUG
- CARGO_PROFILE_TEST_DEBUG

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
+          rustup toolchain install $NIGHTLY_TOOLCHAIN
+          rustup component add --toolchain $NIGHTLY_TOOLCHAIN rust-src rustfmt clippy llvm-tools-preview
+      - name: Install rustfilt
+        run: |
+          cargo install rustfilt
+      - name: Rust Tests
+        env:
+          RUSTFLAGS: "-Zsanitizer=address -Zexternal-clangrt --cfg disable_loom --cfg vortex_nightly -C debuginfo=2 -C opt-level=0 -C strip=none"
+        run: |
+          cd vortex-ffi
```
0ax1: Can we do this without `cd`. Changing dirs == more stateful.

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
+          triple="x86_64-unknown-linux-gnu"
```
0ax1: Can we get the triple from a command, rather than hardcoding it.

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
+          sccache: s3
+      - uses: actions/checkout@v6
+      - uses: ./.github/actions/setup-prebuild
+      - name: Install nightly for sanitizer
+        run: |
+          rustup toolchain install $NIGHTLY_TOOLCHAIN
+          rustup component add --toolchain $NIGHTLY_TOOLCHAIN rust-src rustfmt clippy llvm-tools-preview
+      - name: Rust Tests
+        env:
+          RUSTFLAGS: "-A warnings -Zsanitizer=memory -Cunsafe-allow-abi-mismatch=sanitizer --cfg disable_loom --cfg vortex_nightly -C debuginfo=2 -C opt-level=0 -C strip=none"
```
0ax1: Document what `-Cunsafe-allow-abi-mismatch=sanitizer` does.

0ax1: `-C debuginfo=2 -C opt-level=0 -C strip=none"` should be the default for dev/test, I think.

.github/workflows/ci.yml
```diff
@@ -416,14 +414,159 @@ jobs:
...
+        run: |
+          # Build with full debug info first (helps with caching)
+          cargo +$NIGHTLY_TOOLCHAIN build --locked --all-features  \
+            -Zbuild-std --target x86_64-unknown-linux-gnu \
```
0ax1: I think you can get the the triple from rustc, rather than having to hardcode it.

vortex-ffi/cinclude/vortex.h
```diff
@@ -742,37 +740,45 @@ void vx_session_free(vx_session *ptr);
...
 
 /**
- * Closes an array sink, must be called to ensure all the values pushed to the sink are written
+ * Close an array sink.
+ * Takes ownership of sink and frees it.
+ * Must be called to ensure all the values pushed to the sink are written
  * to the external resource.
  */
-void vx_array_sink_close(vx_array_sink *sink, vx_error **error_out);
+void vx_array_sink_close(vx_array_sink *sink, vx_error **error);
```
0ax1: docs, docs, docs for the parameters, here and elsewhere

0ax1: why remove `_out`?

vortex-ffi/CMakeLists.txt
```diff
@@ -73,6 +81,14 @@ set_target_properties(vortex_ffi_shared PROPERTIES
 
 if (BUILD_TESTING)
     enable_language(CXX)
+    set(CMAKE_CXX_STANDARD 20)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wextra -Wpedantic")
```
0ax1, line 99: Why only set this for the test target?     
`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wextra -Wpedantic")`

0ax1: What flags would cmake pick if you build a C file with clang++

vortex-ffi/build.rs
```diff
@@ -2,21 +2,28 @@
...
     println!("cargo:rerun-if-changed=build.rs");
 
-    // Skip header generation in environments where cbindgen macro expansion fails
     if env::var("MIRI").is_ok() || env::var("MIRIFLAGS").is_ok() {
-        println!("cargo:warning=Skipping header generation under miri (cbindgen incompatible)");
+        println!("cargo:info=Skipping header generation under miri (cbindgen incompatible)");
+        return;
+    }
+
+    if env::var("RUSTFLAGS")
```
0ax1, line 25: We need to check at the top for env var changing:
`println!("cargo:rerun-if-env-changed=RUSTFLAGS/MIRI/MIRIFLAGS");`

vortex-ffi/src/sink.rs
```diff
@@ -22,97 +21,99 @@ use vortex::io::session::RuntimeSessionExt;
...
 /// to the external resource.
 #[unsafe(no_mangle)]
 pub unsafe extern "C-unwind" fn vx_array_sink_close(
     sink: *mut vx_array_sink,
-    error_out: *mut *mut vx_error,
+    error: *mut *mut vx_error,
 ) {
-    try_or_default(error_out, || {
-        let vx_array_sink { sink, writer } = *unsafe { Box::from_raw(sink) };
+    try_or_default(error, || {
```
0ax1: Surprised this isn't implemented in terms of `impl From` / `into`.

.github/workflows/ci.yml
myrrc: We can with something like `gcc -dumpmachine` but it's more ugly since we're running on a specified architecture in CI anyway.

0ax1:
```
rustc -vV
rustc 1.93.1 (01f6ddf75 2026-02-11)
binary: rustc
commit-hash: 01f6ddf7588f42ae2d7eb0a2f21d44e8e96674cf
commit-date: 2026-02-11
host: aarch64-unknown-linux-gnu
release: 1.93.1
LLVM version: 21.1.8
```

.github/workflows/ci.yml
myrrc: We can't parametrize as the flags are different, and we set different environment variables. We can add these to the matrix as well but I doubt it will be less ugly.

.github/workflows/ci.yml
myrrc: It's not the default apparently

.github/workflows/ci.yml
myrrc: +1

vortex-ffi/CMakeLists.txt
myrrc: It would still pick C flags, I assume? I don't want to introduce CXX if a consumer doesn't want to build tests.

.github/workflows/ci.yml
0ax1:
```
  rust-test-sanitizer:
    name: "Rust tests (${{ matrix.sanitizer.name }} sanitizer)"
    timeout-minutes: 40
    runs-on: >-
      ${{ github.repository == 'vortex-data/vortex'
          && format('runs-on={0}/runner=amd64-medium/image=ubuntu24-full-x64-pre-v2/tag=rust-test-{1}-sanitizer', github.run_id, matrix.sanitizer.name)
          || 'ubuntu-latest' }}
    strategy:
      fail-fast: false
      matrix:
        sanitizer:
          - name: address
            rustflags: "-A warnings -Zsanitizer=address -Zsanitizer=leak -Cunsafe-allow-abi-mismatch=sanitizer --cfg disable_loom --cfg vortex_nightly"
            build_std: ""
          - name: memory
            rustflags: "-A warnings -Zsanitizer=memory -Cunsafe-allow-abi-mismatch=sanitizer --cfg disable_loom --cfg vortex_nightly"
            build_std: "-Zbuild-std"
          - name: thread
            rustflags: "-A warnings -Zsanitizer=thread -Cunsafe-allow-abi-mismatch=sanitizer --cfg disable_loom --cfg vortex_nightly"
            build_std: "-Zbuild-std"
    env:
      # Each sanitizer only reads its own *_OPTIONS var, so setting all is safe.
      ASAN_OPTIONS: "symbolize=1:print_stats=1:check_initialization_order=1:detect_leaks=1:leak_check_at_exit=1"
      LSAN_OPTIONS: "report_objects=1"
      MSAN_OPTIONS: "symbolize=1:print_stats=1"
      TSAN_OPTIONS: "symbolize=1"
      ASAN_SYMBOLIZER_PATH: "/usr/bin/llvm-symbolizer"
      MSAN_SYMBOLIZER_PATH: "/usr/bin/llvm-symbolizer"
      TSAN_SYMBOLIZER_PATH: "/usr/bin/llvm-symbolizer"
      CARGO_PROFILE_DEV_DEBUG: "true"
      CARGO_PROFILE_TEST_DEBUG: "true"
      VORTEX_SKIP_SLOW_TESTS: "1"
    steps:
      - uses: runs-on/action@v2
        if: github.repository == 'vortex-data/vortex'
        with:
          sccache: s3
      - uses: actions/checkout@v6
      - name: Install nightly for sanitizer
        run: |
          rustup toolchain install $NIGHTLY_TOOLCHAIN
          rustup component add --toolchain $NIGHTLY_TOOLCHAIN rust-src rustfmt clippy llvm-tools-preview
      - name: Rust Tests
        env:
          RUSTFLAGS: ${{ matrix.sanitizer.rustflags }}
        run: |
          triple=$(rustc -vV | awk '/host:/ { print $2 }')

          cargo +$NIGHTLY_TOOLCHAIN build --locked --all-features \
            ${{ matrix.sanitizer.build_std }} --target $triple \
            -p vortex-buffer -p vortex-fastlanes -p vortex-fsst -p vortex-alp -p vortex-array

          cargo +$NIGHTLY_TOOLCHAIN nextest run \
            --locked --all-features --no-fail-fast \
            ${{ matrix.sanitizer.build_std }} --target $triple \
            -p vortex-buffer -p vortex-fastlanes -p vortex-fsst -p vortex-alp -p vortex-array

          # vortex-ffi requires --no-default-features as otherwise we pull in
          # Mimalloc which interferes with sanitizers.
          # cargo nextest reports fewer leaks than cargo test.
          cargo +$NIGHTLY_TOOLCHAIN test --locked --no-default-features \
            -Zbuild-std --no-fail-fast --target $triple \
            --lib vortex-ffi -- --no-capture
```

.github/workflows/ci.yml
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: +1 for _out, but I suggest leaving the docs for another PR

robert3005 (approved):
I have adjusted required checks. Let ship this and make adjustments if necessary

PR #7160, opened 2026-03-25
spark: pass format options into reader

a10y:
## Summary

In Spark, pass format properties into reader.

This mirrors the existing implementation of ParquetDataSourceV2 which rips the relevant S3/Azure/etc. properties off of the active SparkSession's hadoopConf. This is also similar to what we do in Vortex iceberg already.

## Testing

Added a new test suite for reading/writing to S3 using the Adobe S3Mock testcontainer.

### DIFF
--- a/java/vortex-spark/src/main/java/dev/vortex/spark/config/VortexS3Properties.java
+++ b/java/vortex-spark/src/main/java/dev/vortex/spark/config/VortexS3Properties.java
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+package dev.vortex.spark.config;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import java.util.Map;
+import java.util.Optional;
+
+public final class VortexS3Properties {
+    private static final String ACCESS_KEY = "aws_access_key_id";
+    private static final String SECRET_KEY = "aws_secret_access_key";
+    private static final String SESSION_TOKEN = "aws_session_token";
+    private static final String REGION = "aws_region";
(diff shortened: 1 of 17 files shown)


java/vortex-spark/src/main/java/dev/vortex/spark/read/VortexPartitionReader.java
```diff
@@ -47,7 +47,7 @@ public ColumnarBatch get() {
      * Initialize the Vortex File and ArrayStream resources.
      */
     void initNativeResources() {
-        file = Files.open(partition.getPath());
+        file = Files.open(partition.getPath(), partition.getFormatOptions());
```
a10y, line 50: this is the thing we needed. everything else is just plumbing

java/vortex-spark/src/main/java/dev/vortex/spark/config/VortexS3Properties.java
```diff
@@ -0,0 +1,72 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+package dev.vortex.spark.config;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+import java.util.Map;
+import java.util.Optional;
+
+public final class VortexS3Properties {
```
a10y, line 11: i ripped this straight out of our iceberg fork

a10y:
@claude please review. make no mistakes. be extremely javapilled

a10y:
@claude please re-review. I should have addressed all of the RED comments, and most of the green ones. I don't think the YELLOW ones are relevant.

as always, stay javapilled.

a10y:
@claude ok i addressed the props in cache key thing. happy now?

robert3005 (approved):

PR #7168, opened 2026-03-25
Shade jackson in vortex-jni and vortex-spark and avoid scala version conflicts

robert3005:
Java dependencies work in mysterious ways but mostly it's jackson-scala that's
to blame. Because scala-lang is just a versioned library and 2.13 > 2.12 I had to remove consistent-versions since they specifically forbid doing the thing we need to - have two conflicting version sets across spark versions


a10y (approved):

robert3005:
There's still some sadness I'm trying to figure out here

PR #7174, opened 2026-03-26
Enable sccache for windows builds

AdamGS:
# Summary

Now that windows builds are faster due to the underlying image, maybe this will be another measurable improvement


AdamGS:
Moved this work upstream https://github.com/runs-on/action/pull/32

PR #7175, opened 2026-03-26
chore[gpu]: split plan into unmaterialized / materialized

0ax1:
This allows for checking whether the shared memory usage blows past the max GPU device shared memory before launching the dyn dispatch kernel, as well as having a separate materialize step we can fine tune that moves buffers from the host to the GPU.


0ax1:
I'll rework how we properly account for and allocate shared memory for the GPU in a follow up.

0ax1:
Note the state of this PR is kinda transient and will be reworked as part of: https://github.com/vortex-data/vortex/commit/2cb615ef29487b683d3c2860175067414880f13d

a10y (approved):

0ax1:
@a10y :heart:

PR #7176, opened 2026-03-26
Filter indices without an iterator for Bool

AdamGS:
## Summary

Noticed this difference as part of #6740 that changed some of the inlining around `collect_bool`, making the closure not inlined.
The comment over `filter_indices` used to be on a different function, and the iterator and generics seems wasteful.

This seems to be about 30µs faster than develop pre #6470 for most affected benchmarks, which is about 3% (probably within the noise, but the affect seems very consistent).

### DIFF
--- a/vortex-array/src/arrays/bool/compute/filter.rs
+++ b/vortex-array/src/arrays/bool/compute/filter.rs
@@ -28,11 +28,7 @@ impl FilterReduce for Bool {
             .vortex_expect("AllTrue and AllFalse are handled by filter fn");
 
         let buffer = match mask_values.threshold_iter(FILTER_SLICES_DENSITY_THRESHOLD) {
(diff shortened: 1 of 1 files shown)


robert3005:
I think you have some other random code in here

AdamGS:
@robert3005 I decided to trust github for once and it betrayed me :( I'll fix it once the underlying PR merges.

robert3005:
If you look at bitbuffer there’s a function that’s comparable to collect_bool but also includes the value and the index. Maybe that’s even faster?

AdamGS:
I don't think that works here because you need the offset don't you? looking at `map_cmp`.

robert3005:
You’re right the iteration is over different thing

robert3005 (approved):

PR #7178, opened 2026-03-26
Unify benchmark dispatch into a workflow

AdamGS:
This PR started as something else, but I still think its nicer to split the flows into another workflow to handle triggering the benchmarks themselves.

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -4,7 +4,13 @@ name: PR Benchmarks
 
 concurrency:
   # The group causes runs to queue instead of running in parallel.
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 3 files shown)


0ax1 (approved):

.github/workflows/bench-pr.yml
```diff
@@ -137,17 +121,16 @@ jobs:
           comment-tag: bench-pr-comment-${{ matrix.benchmark.id }}
 
       - name: Comment PR on failure
-        if: failure() && inputs.mode == 'pr' && github.event.pull_request.head.repo.fork == false
+        if: failure() && github.event.pull_request.head.repo.fork == false
         uses: thollander/actions-comment-pull-request@v3
         with:
           message: |
-            # 🚨🚨🚨❌❌❌ BENCHMARK FAILED ❌❌❌🚨🚨🚨
+            # BENCHMARK FAILED
```
0ax1, line 128: ty

PR #7180, opened 2026-03-26
Specialize export of chunked list arrays

AdamGS:
## Summary

If we get Chunked<List> array, it currently goes through the much more expensive export code path that rebuilds it as a ListView before casting it back, which is wasteful. Instead, we can convert each chunk into a more natural list, then concat them.

### DIFF
--- a/vortex-array/src/arrow/executor/list.rs
+++ b/vortex-array/src/arrow/executor/list.rs
@@ -17,6 +17,7 @@ use crate::ArrayRef;
 use crate::Canonical;
 use crate::DynArray;
 use crate::ExecutionCtx;
+use crate::arrays::Chunked;
 use crate::arrays::List;
 use crate::arrays::ListArray;
 use crate::arrays::ListView;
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

vortex-array/src/arrow/executor/list.rs
```diff
@@ -31,7 +32,7 @@ use crate::dtype::NativePType;
 use crate::dtype::Nullability;
 use crate::vtable::ValidityHelper;
 
-/// Convert a Vortex array into an Arrow GenericBinaryArray.
+/// Convert a Vortex VarBinArray into an Arrow [`GenericListArray`](arrow_array:array::GenericListArray).
```
a10y, line 35: This should be `ListArray` no?

vortex-array/src/arrow/executor/list.rs
```diff
@@ -68,7 +80,7 @@ pub(super) fn to_arrow_list<O: OffsetSizeTrait + NativePType>(
     list_view_zctl::<O>(zctl, elements_field, ctx)
 }
 
-/// Convert a Vortex VarBinArray into an Arrow GenericBinaryArray.
+/// Convert a Vortex VarBinArray into an Arrow [`GenericListArray`](arrow_array:array::GenericListArray).
```
a10y, line 83: this comment also is wrong (and was already wrong before)

vortex-array/src/arrow/executor/list.rs
AdamGS: No, `ListArray` is a type alias for `GenericListArray<i32>`, this can also return `LargeListArray` which is `GenericListArray<i64>`

AdamGS: oh I see what you're saying, can't believe I only fixed half the comment 🤦

PR #7186, opened 2026-03-26
feat: make dispatch plan dynamically sized

0ax1:
Replace the fixed-size `DynamicDispatchPlan` struct with a variable-length packed byte buffer, removing the `MAX_STAGES` and `MAX_SCALAR_OPS` limits.

### DIFF
--- a/vortex-cuda/benches/dynamic_dispatch_cuda.rs
+++ b/vortex-cuda/benches/dynamic_dispatch_cuda.rs
@@ -12,6 +12,7 @@ use std::time::Duration;
 use criterion::BenchmarkId;
 use criterion::Criterion;
 use criterion::Throughput;
+use cudarc::driver::CudaSlice;
 use cudarc::driver::DevicePtr;
 use cudarc::driver::LaunchConfig;
 use cudarc::driver::PushKernelArg;
@@ -59,7 +60,7 @@ fn run_timed(
     cuda_ctx: &mut CudaExecutionCtx,
     array_len: usize,
     output_buf: &CudaDeviceBuffer,
-    device_plan: &Arc<cudarc::driver::CudaSlice<u8>>,
+    device_plan: &Arc<CudaSlice<u8>>,
     shared_mem_bytes: u32,
 ) -> VortexResult<Duration> {
     let cuda_function = cuda_ctx.load_function("dynamic_dispatch", &[PType::U32])?;
@@ -115,7 +116,7 @@ struct BenchRunner {
     _plan: CudaDispatchPlan,
     smem_bytes: u32,
     len: usize,
-    device_plan: Arc<cudarc::driver::CudaSlice<u8>>,
+    device_plan: Arc<CudaSlice<u8>>,
     output_buf: CudaDeviceBuffer,
     _plan_buffers: Vec<vortex::array::buffer::BufferHandle>,
 }
--- a/vortex-cuda/benches/dynamic_dispatch_cuda.rs
+++ b/vortex-cuda/benches/dynamic_dispatch_cuda.rs
@@ -59,7 +59,7 @@ fn run_timed(
     cuda_ctx: &mut CudaExecutionCtx,
     array_len: usize,
     output_buf: &CudaDeviceBuffer,
-    device_plan: &Arc<cudarc::driver::CudaSlice<CudaDispatchPlan>>,
+    device_plan: &Arc<cudarc::driver::CudaSlice<u8>>,
     shared_mem_bytes: u32,
 ) -> VortexResult<Duration> {
     let cuda_function = cuda_ctx.load_function("dynamic_dispatch", &[PType::U32])?;
@@ -115,8 +115,7 @@ struct BenchRunner {
     _plan: CudaDispatchPlan,
     smem_bytes: u32,
     len: usize,
-    // Keep alive
-    device_plan: Arc<cudarc::driver::CudaSlice<CudaDispatchPlan>>,
+    device_plan: Arc<cudarc::driver::CudaSlice<u8>>,
     output_buf: CudaDeviceBuffer,
     _plan_buffers: Vec<vortex::array::buffer::BufferHandle>,
 }
@@ -134,7 +133,7 @@ impl BenchRunner {
         let device_plan = Arc::new(
             cuda_ctx
                 .stream()
-                .clone_htod(std::slice::from_ref(&dispatch_plan))
+                .clone_htod(dispatch_plan.as_bytes())
                 .expect("htod plan"),
         );
 
--- a/vortex-cuda/src/dynamic_dispatch/mod.rs
+++ b/vortex-cuda/src/dynamic_dispatch/mod.rs
@@ -20,6 +20,8 @@
 #![allow(non_snake_case)]
 #![allow(clippy::cast_possible_truncation)]
 
+use std::ptr::read_unaligned;
+use std::slice::from_raw_parts;
 use std::sync::Arc;
 
 use cudarc::driver::DevicePtr;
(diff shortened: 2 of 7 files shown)


vortex-cuda/benches/dynamic_dispatch_cuda.rs
```diff
@@ -59,7 +59,7 @@ fn run_timed(
     cuda_ctx: &mut CudaExecutionCtx,
     array_len: usize,
     output_buf: &CudaDeviceBuffer,
-    device_plan: &Arc<cudarc::driver::CudaSlice<CudaDispatchPlan>>,
+    device_plan: &Arc<cudarc::driver::CudaSlice<u8>>,
```
AdamGS: import

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -47,12 +49,191 @@ pub use plan_builder::find_unfusable_nodes;
...
+///
+/// # Safety
+///
+/// The caller must ensure `T` is a `#[repr(C)]` type with no padding that
+/// contains uninitialised bytes.  All the types we serialise (`PlanHeader`,
+/// `PackedStage`, `ScalarOp`) satisfy this because they are bindgen-generated
+/// `#[repr(C)]` structs whose padding bytes are always written before
+/// serialisation.
+fn as_bytes<T: Sized>(val: &T) -> &[u8] {
+    unsafe { from_raw_parts(val as *const T as *const u8, size_of::<T>()) }
```
AdamGS: can you use `cast` here? if we ever want to run miri on this stuff, it helps some of its more complex checks

AdamGS: IDK how common this is, but assuming we use it a bunch, we could have macro that does both `repr(C)`/packed/whatever we care about, and implements a marker trait that this function can use.

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -47,12 +49,191 @@ pub use plan_builder::find_unfusable_nodes;
...
+/// [PackedStage 0][ScalarOp × N0]          — variable
+/// [PackedStage 1][ScalarOp × N1]          — variable
+/// ...
+/// ```
+///
+/// This is uploaded to the device and cooperatively copied into shared memory
+/// by the kernel.
+#[derive(Clone)]
+pub struct CudaDispatchPlan {
+    bytes: Vec<u8>,
```
AdamGS: Is there a reason to use Vec<u8> here and not a Vortex buffer/bytes?

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -47,12 +49,191 @@ pub use plan_builder::find_unfusable_nodes;
...
+
+    /// The raw packed plan bytes, ready for upload to the device.
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.bytes
+    }
+
+    /// Number of stages in the plan.
+    pub fn num_stages(&self) -> u8 {
+        let header: PlanHeader =
+            unsafe { read_unaligned(self.bytes.as_ptr() as *const PlanHeader) };
```
AdamGS: following above, if its a vortex buffer with alignment, we can read_aligned here!

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -47,12 +49,191 @@ pub use plan_builder::find_unfusable_nodes;
...
+        header.num_stages
+    }
+
+    /// Parse and return a read-only view of the stage at `index`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `index >= num_stages()`.
+    pub fn stage(&self, index: usize) -> ParsedStage {
```
AdamGS, line 209: You don't have to allocate this, you can zero-copy parse the underlying data into a `ParsedStage<'a>` thing, which will also have a much cheaper clone. I can't see where its used so I'm not sure if that fits here.

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -81,13 +262,12 @@ impl SourceOp {
...
 
-    /// Decode run-end encoding. Offsets reference shared memory regions
-    /// populated by earlier input stages.
+    /// Decode run-end encoding.
     pub fn runend(
-        ends_smem_offset: u32,
-        values_smem_offset: u32,
-        num_runs: u64,
-        offset: u64,
+        ends_smem_offset: u32,   // smem region holding run-end endpoints
```
AdamGS: put the argument docs in the rustdoc?

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -47,12 +49,191 @@ pub use plan_builder::find_unfusable_nodes;
...
-pub type CudaDispatchPlan = DynamicDispatchPlan;
+impl CudaDispatchPlan {
+    /// Build a packed plan from a sequence of stages.
+    ///
+    /// The last stage is the output pipeline; earlier stages are input stages.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `stages` is empty or the serialized plan exceeds 65535 bytes.
+    pub fn new(stages: &[Stage]) -> Self {
```
AdamGS: given that everything is clone and you end up cloning and copying stuff anyway, why take the stages by reference? I think it might be better served by a `FromIterator` implementation that accepts an owned iterator.

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -47,12 +49,191 @@ pub use plan_builder::find_unfusable_nodes;
...
+
+        let mut total_size = header_size;
+        for stage in stages {
+            total_size += stage_header_size;
+            total_size += stage.scalar_ops.len() * scalar_op_size;
+        }
+        assert!(
+            total_size <= u16::MAX as usize,
+            "packed plan size {total_size} exceeds u16::MAX"
+        );
```
AdamGS, line 161: I don't think it matters a lot, but you can do this in the same pass, if you fail you panic anyway

vortex-cuda/src/dynamic_dispatch/mod.rs
AdamGS: the more I think about, the more I like limiting this more. This only works for "real" (I'm sure there's a name for this) types, it won't work correctly for stuff that has indirection (String, HashMap, Box<T>), making it harder to misuse seems valuable.

AdamGS: (doesn't have to be a proc macro, can also be a regular one or something like an unsafe trait)

vortex-cuda/src/dynamic_dispatch/mod.rs
0ax1: Yeah these slipped in C style

vortex-cuda/src/dynamic_dispatch/mod.rs
0ax1: I'll try going with
```
fn as_bytes<T: Sized>(val: &T) -> &[u8] {
    unsafe { from_raw_parts(std::ptr::addr_of!(*val).cast(), size_of::<T>()) }
}
```

vortex-cuda/src/dynamic_dispatch/mod.rs
0ax1: should be byte buffer yeh

vortex-cuda/src/dynamic_dispatch/mod.rs
0ax1: The cost of copying is fine in favor how of not going with an output parameter. This is not a hot-path.

vortex-cuda/src/dynamic_dispatch/mod.rs
0ax1: done

AdamGS (approved):

PR #7191, opened 2026-03-27
Cache python class imports

AdamGS:
## Summary

Was reading the arrow-rs changelog and I ran into [this](https://github.com/apache/arrow-rs/pull/9439) PR, seems like `PyOnceLock` was built for this and they have some very promising benchmarks [there](https://github.com/apache/arrow-rs/pull/9439/changes#r2955818325).

I've also wrapped all static strings that are passed into `pyo3` with the `intern` macro, which prevents allocating a new `PyString` on every call.


danking (approved):
fantastic

danking:
Ah we don't have any python-side benchmarks :|

PR #7193, opened 2026-03-27
Make sure all tokio runtime propegate observability in the same way

AdamGS:
## Summary

Just fixing custom-labels or any other tracing propagation through our tokio runtime types, which stopped working at some point.


vortex-io/src/runtime/tokio.rs
```diff
@@ -159,13 +151,11 @@ impl BlockingRuntime for TokioRuntime {
     }
 }
 
-#[cfg(feature = "tokio")]
```
a10y, line 162: is this in purpose?

a10y (approved):

vortex-io/src/runtime/tokio.rs
AdamGS: yeah, its already at the module level

PR #7194, opened 2026-03-27
move BitPacked slice from kernel -> parent reduce

a10y:
We can move slicing to run in the optimizer since it does not access data buffers.


a10y:
lol wrong base branch

PR #7197, opened 2026-03-27
feat(cuda): single-pass dispatch plan builder

0ax1:
Refactor the dynamic dispatch plan builder to walk the encoding tree exactly once, discovering unfusable subtrees and computing shared memory requirements in the same pass. The result is a 3-variant enum (`Fused`, `PartiallyFused`, `Unfusable`) that replaces the previous `Result<Option<>>` API and eliminates the separate`find_unfusable_nodes` traversal.

Shared memory is now validated upfront in `DispatchPlan::new` - before any subtree kernels are executed - so we never pay GPU cost for a plan that will not fit.

The plan stages are split into `smem_stages` (fully decoded into persistent shared memory) and `output_stage` (tiled through a scratch region), making the two-phase kernel execution model explicit in the host-side data structures. Shared memory allocation invariants are documented on `FusedPlan`.


0ax1:
Right now we're loading all dict values and all runend ends & codes into GPU shared memory. I didn't change this as part of this PR to scope it. But in a follow up I want to change the logic to tile/window through dict and runed with a fixed shared memory budget. In case of dict, this might be based on some LRU cache indirection or similar.

vortex-cuda/src/dynamic_dispatch/mod.rs
```diff
@@ -71,15 +67,15 @@ fn as_bytes<T: Sized>(val: &T) -> &[u8] {
 /// This is NOT a C ABI struct — it exists purely on the Rust side and is
 /// serialized into the packed plan byte buffer by [`CudaDispatchPlan::new`].
 #[derive(Clone)]
-pub struct Stage {
+pub struct MaterializedStage {
     pub input_ptr: u64,
     pub smem_offset: u32,
```
joseph-isaacs, line 74: can you explain this

vortex-cuda/src/hybrid_dispatch/mod.rs
```diff
@@ -41,105 +41,73 @@
...
+            for subtree in &pending_subtrees {
+                let canonical = subtree.clone().execute_cuda(ctx).await?;
+                subtree_buffers.push(canonical.into_primitive().into_parts().buffer);
+            }
+
+            let num_subtrees = subtree_buffers.len();
+            let materialized = plan.materialize_with_subtrees(subtree_buffers, ctx)?;
+            let num_stages = materialized.dispatch_plan.num_stages();
+            trace!(encoding = %array.encoding_id(), num_stages, num_subtrees, "partially-fused dispatch");
+            materialized.execute(output_ptype, array.len(), ctx)
```
joseph-isaacs, line 98: why do you have a distinction between these (fused is partially fused without children)?

vortex-cuda/src/hybrid_dispatch/mod.rs
```diff
@@ -41,105 +41,73 @@
...
+    trace!(encoding = %array.encoding_id(), dtype = ?array.dtype(), len = array.len(), "try GPU dispatch");
+
+    match DispatchPlan::new(array)? {
+        DispatchPlan::Fused(plan) => {
+            let output_ptype = PType::try_from(array.dtype())?;
+            let materialized = plan.materialize(ctx)?;
+            let num_stages = materialized.dispatch_plan.num_stages();
+            trace!(encoding = %array.encoding_id(), num_stages, "fully-fused dispatch");
+            materialized.execute(output_ptype, array.len(), ctx)
+        }
```
joseph-isaacs, line 80: can you not impl execute for the plan?

vortex-cuda/src/hybrid_dispatch/mod.rs
0ax1: Fully fused means full dynamic dispatch where no buffers need to be prepared by a separate standalone kernel.

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -71,70 +47,39 @@ pub struct MaterializedPlan {
...
-            };
-        }
-        return false;
+        let arr = array.as_::<Dict>();
+        // As of now the dict dyn dispatch kernel requires
+        // codes and values to have the same byte width.
+        return match (
+            PType::try_from(arr.values().dtype()),
+            PType::try_from(arr.codes().dtype()),
+        ) {
```
joseph-isaacs, line 64: is this hard to fix?

vortex-cuda/src/hybrid_dispatch/mod.rs
0ax1: The dispatch plan is the unmaterialized one. You have to materialize first:
```
/// A plan whose source buffers have been copied to the device, ready for kernel launch.
pub struct MaterializedPlan {
    /// Packed plan byte buffer, to upload to the device.
    pub dispatch_plan: CudaDispatchPlan,
    /// Device buffer handles that must be kept alive while the plan is in use.
    pub device_buffers: Vec<BufferHandle>,
    /// Dynamic shared memory bytes needed to launch this plan.
    pub shared_mem_bytes: u32,
}
```
`/// A plan whose source buffers have been copied to the device, ready for kernel launch.`

Part of this PR is being able to create the unmaterialized `DispatchPlan` without having to materialize, so we can check stuff like would the shared memory we'd allocate for this dyn dispatch kernel be within resource limits of the device.

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -364,19 +388,23 @@ impl UnmaterializedPlan {
     ///
     /// When the plan builder encounters a `SliceArray`, it resolves the slice
     /// by invoking the child's `reduce_parent`, `execute_parent`.
```
joseph-isaacs, line 387: you cannot do this?

vortex-cuda/src/dynamic_dispatch/mod.rs
0ax1: https://github.com/vortex-data/vortex/pull/7197#discussion_r3009236675

joseph-isaacs: add a doc str

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -364,19 +388,23 @@ impl UnmaterializedPlan {
...
-            return self.walk(reduced);
+            return self.walk(reduced, pending_subtrees);
         }
 
         // execute_parent: (for types with SliceExecuteAdaptor/SliceKernel, like BitPacked)
         let mut ctx = ExecutionCtx::new(VortexSession::empty().with::<ArraySession>());
         if let Some(executed) = child.vtable().execute_parent(&child, &array, 0, &mut ctx)? {
-            return self.walk(executed);
+            return self.walk(executed, pending_subtrees);
         }
```
joseph-isaacs: dont do this

0ax1: yeh

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
0ax1: i don't actually think so. we need to pass the types for each stage of the dyn dispatch plan to the gpu and then branch there. sth like `match_each_xyz_type` which we do in rust.

vortex-cuda/src/dynamic_dispatch/mod.rs
0ax1: done

joseph-isaacs (approved):

PR #7201, opened 2026-03-29
feat: `Mean` aggregate

blaginin:


vortex-array/src/aggregate_fn/fns/count/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+use crate::aggregate_fn::EmptyOptions;
+use crate::dtype::DType;
+use crate::dtype::Nullability;
+use crate::dtype::PType;
+use crate::scalar::Scalar;
+
+/// Return the count of non-null elements in an array.
+///
+/// See [`Count`] for details.
+pub fn count(array: &ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<u64> {
```
gatesn: If this is counting things in an array, should it return a usize?

vortex-array/src/aggregate_fn/fns/count/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    fn reset(&self, partial: &mut Self::Partial) {
+        *partial = 0;
+    }
+
+    #[inline]
+    fn is_saturated(&self, _partial: &Self::Partial) -> bool {
+        false
+    }
+
+    fn accumulate(
```
gatesn, line 85: We should be able to register a generic aggregate kernel to reduce count-non-null to be `Array.validity().sum()`, then we avoid decompressing all the data.

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+pub struct Mean;
+
+/// Internal accumulation state for [`Mean`].
+pub struct MeanPartial {
+    sum: f64,
+    count: u64,
+}
+
+fn partial_struct_dtype() -> DType {
+    DType::Struct(
```
gatesn: Can you static LazyLock this entire dtype inside this function?

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+/// The partial state is a struct `{sum: f64, count: u64}` so that partials from
+/// different accumulators can be correctly combined via weighted addition.
+#[derive(Clone, Debug)]
+pub struct Mean;
+
+/// Internal accumulation state for [`Mean`].
+pub struct MeanPartial {
+    sum: f64,
+    count: u64,
+}
```
joseph-isaacs: This will compute with a pretty large error?

joseph-isaacs (changes requested):
This should be decimal for decimals?

vortex-array/src/aggregate_fn/fns/mean/mod.rs
joseph-isaacs: $$\mu_n = \sum_{i=0}^n x_i/n \iff \mu_{n+1} = \mu_n + \frac{x_{n+1} - \mu_n}{1+n}$$

joseph-isaacs: you need only hold the partial mean and count

gatesn: You can incrementally push the next value into this partial, but you cannot combine two partials afaik?

joseph-isaacs: For combining you do. $$\mu_{AB} = \frac{n_A \mu_A + n_B \mu_B}{n_A + n_B}$$

blaginin: I thought in that formula there's even more division and hence more rounding errors?

I think there are some smart algorithms, but I haven't seen a [simple implementation](https://github.com/apache/datafusion/blob/1a0af76f780708ff9d9e1940686748bd743352d3/datafusion/functions-aggregate/src/average.rs#L324-L324) ever being a problem in df

gatesn: 100% agree. Let's do what DataFusion odes

joseph-isaacs: DF sum is T and ours is f64

joseph-isaacs: Error comes from numbers of different scales being combined, not from number of OPS.

Since we are storing these on disk we cannot "just change it later".

gatesn: DF only supports AVG for floats, decimals, and durations. And coerces all floats to f64.

vortex-array/src/aggregate_fn/fns/count/mod.rs
blaginin: https://github.com/vortex-data/vortex/pull/7201/commits/575672ad697c71e28991e496b5676c03053672ca

added generic kernels first, but then it's hard to express "I never want `Count::accumulate` to be called" - so switched to `try_accumulate` instead

vortex-array/src/aggregate_fn/fns/mean/mod.rs
joseph-isaacs:
```
| Input Type   | DuckDB | DataFusion | Velox      |
|--------------|--------|------------|------------|
| Integer types| DOUBLE | Float64    | DOUBLE     |
| FLOAT/REAL   | DOUBLE | Float64    | REAL       |
| DOUBLE       | DOUBLE | Float64    | DOUBLE     |
| DECIMAL(p,s) | DOUBLE | Float64    | DECIMAL(p,s)|
```

blaginin: thought about this more - i think it should decimal for decimals? - happy to change

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Compute the arithmetic mean of an array.
+///
+/// See [`Mean`] for details.
+pub fn mean(array: &ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Scalar> {
+    let options = PairOptions(EmptyOptions, EmptyOptions);
+    let vtable = Mean::combined();
+
+    let coerced_dtype = vtable.coerce_args(&options, array.dtype())?;
+    let coerced = array.cast(coerced_dtype.clone())?;
```
blaginin: shouldn't `Accumulator` do that for me?

vortex-array/src/aggregate_fn/combined.rs
joseph-isaacs, line 1: we can factor this out of this PR, we could at a later date add the back to on disk version will be the same?

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+/// Compute the arithmetic mean of an array.
+///
+/// Implemented as `Sum / Count` via [`BinaryCombined`].
+///
+/// Coercion / return type:
+/// - Booleans and primitive numeric types are coerced to `f64` and the result
+///   is a nullable `f64`.
+/// - Decimals are kept as decimals with widened precision and scale
+///   (`+4` each, capped at [`MAX_PRECISION`] / [`MAX_SCALE`]), matching
+///   DataFusion's `coerce_avg_type`.
```
joseph-isaacs, line 51: I think you definitely want higher for the sum, that is only 10k max sized values for an overflow

vortex-array/src/aggregate_fn/combined.rs
blaginin: > disk version will be the same

i blocked serialization for now?

> we can factor this out of this PR

why?

vortex-array/src/aggregate_fn/fns/mean/mod.rs
blaginin: yeah, but `Sum.partial_dtype()` also adds +10

joseph-isaacs: Then this comment need to explain what the +4 is for, 4 extra least sig figures?

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    fn finalize(&self, sum: ArrayRef, count: ArrayRef) -> VortexResult<ArrayRef> {
+        let target = match sum.dtype() {
+            DType::Decimal(..) => sum.dtype().with_nullability(Nullability::Nullable),
+            _ => DType::Primitive(PType::F64, Nullability::Nullable),
+        };
+        let sum_cast = sum.cast(target.clone())?;
+        let count_cast = count.cast(target)?;
+        sum_cast.binary(count_cast, Operator::Div)
```
joseph-isaacs, line 92: handle div by zero?

joseph-isaacs:
This is a mean? not avg?

blaginin:
I think the pattern is `avg` is the function and `mean` is an alias

joseph-isaacs:
I think we want mean with avg as an alias

joseph-isaacs:
or really arithmetic_mean

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        let target = DType::Primitive(PType::F64, Nullability::Nullable);
+        let sum_cast = left_scalar.cast(&target)?;
+        let count_cast = right_scalar.cast(&target)?;
+
+        let sum = sum_cast.as_primitive().typed_value::<f64>();
+        let count = count_cast.as_primitive().typed_value::<f64>();
+        let value = match (sum, count) {
+            (None, _) | (_, None) | (_, Some(0.0)) => return Ok(Scalar::null(target)), // Sum overflowed
```
joseph-isaacs, line 107: empty or overflowed?

vortex-array/src/aggregate_fn/combined.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    /// Field name for the left child in the partial struct dtype.
+    fn left_name(&self) -> &'static str {
+        "left"
+    }
+
+    /// Field name for the right child in the partial struct dtype.
+    fn right_name(&self) -> &'static str {
+        "right"
+    }
```
joseph-isaacs, line 76: why default these?

vortex-array/src/aggregate_fn/combined.rs
```diff
@@ -0,0 +1,264 @@
...
+impl<T: BinaryCombined> Combined<T> {
+    /// Construct a new combined aggregate vtable.
+    pub fn new(inner: T) -> Self {
+        Self(inner)
+    }
+}
+
+impl<T: BinaryCombined> AggregateFnVTable for Combined<T> {
+    type Options = CombinedOptions<T>;
+    type Partial = (LeftPartial<T>, RightPartial<T>);
```
joseph-isaacs, line 143: How will this partial be seralized?

vortex-array/src/aggregate_fn/combined.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+}
+
+fn struct_dtype(left_name: &str, right_name: &str, left: DType, right: DType) -> DType {
+    DType::Struct(
+        StructFields::new(
+            FieldNames::from_iter([FieldName::from(left_name), FieldName::from(right_name)]),
+            vec![left, right],
+        ),
+        Nullability::NonNullable,
```
joseph-isaacs: method on combined?

joseph-isaacs (approved):

vortex-array/src/aggregate_fn/combined.rs
blaginin: those don't matter, so may as well use left/right

vortex-array/src/aggregate_fn/combined.rs
blaginin: as a struct scalar with left/right fields

PR #7205, opened 2026-03-30
feat: Introduce writeBatchFfi in Java JNI to avoid IPC serialization overhead

JingsongLi:
## Summary

Closes: #7204

Currently writeBatch accepts Arrow IPC serialized bytes (byte[]), which requires serialization/deserialization overhead. We wants a writeBatchFfi path that directly accepts raw Arrow C Data Interface pointers (arrow_array_addr and arrow_schema_addr), avoiding the IPC round-trip.

We have conducted some internal tests, and the cost of serialization may reach 20% of the cost of writing.

(From Apache Paimom community)

## Testing

`JNIWriterTest`

### DIFF
--- a/java/vortex-jni/src/main/java/dev/vortex/api/VortexWriter.java
+++ b/java/vortex-jni/src/main/java/dev/vortex/api/VortexWriter.java
@@ -43,6 +43,18 @@ public interface VortexWriter extends AutoCloseable {
     void writeBatch(byte[] arrowData) throws IOException;
 
     /**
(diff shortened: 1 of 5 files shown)


a10y:
We fixed the wasm-integration check earlier today, I think you can pull latest from the develop branch and should fix it

robert3005 (approved):
I think this is in the direction of the apis we need. We will also need to add writing arrow recordbatch stream

JingsongLi:
Thanks @a10y @robert3005 for your feedback. Rebased and force pushed.

PR #7208, opened 2026-03-30
fix: restrict wasm-bindgen to browser wasm, fixing wasip1 builds

0ax1:
wasm-bindgen 0.2.115 changed its cfg guard from
`cfg(all(target_arch = "wasm32", target_os = "unknown"))` to `cfg(target_family = "wasm")`, causing `__wbindgen_*` imports to be emitted for all wasm targets including wasm32-wasip1. Combined with wasm-test/Cargo.lock being untracked (deleted in #6187), CI picked up the new version and the wasmer integration test broke.

Narrow the `wasm-bindgen-futures` dependency and the wasm runtime module in vortex-io from `cfg(target_arch = "wasm32")` to `cfg(all(target_arch = "wasm32", target_os = "unknown"))`, which matches only wasm32-unknown-unknown (browser) and excludes WASI targets. For the fuzz crate's wasip1 target, provide a session without an async runtime handle since the fuzz workloads are synchronous.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -315,6 +315,7 @@ unexpected_cfgs = { level = "deny", check-cfg = [
     "cfg(codspeed)",
     "cfg(disable_loom)",
     "cfg(vortex_nightly)",
+    'cfg(target_os, values("unknown"))',
 ] }
 warnings = "warn"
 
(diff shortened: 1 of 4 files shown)


AdamGS (approved):

AdamGS:
🚢

PR #7212, opened 2026-03-31
High level C Scan API

myrrc:
Expose data source, partition, and scan primitives through C with rust-side and C-side tests.

### DIFF
--- a/vortex-ffi/CMakeLists.txt
+++ b/vortex-ffi/CMakeLists.txt
@@ -100,7 +100,6 @@ if (BUILD_TESTS)
     set(CMAKE_CXX_STANDARD 20)
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wextra -Wpedantic")
-
     enable_testing()
     add_subdirectory(test)
 endif()
--- a/vortex-ffi/cbindgen.toml
+++ b/vortex-ffi/cbindgen.toml
@@ -1,6 +1,6 @@
 language = "C"
 braces = "SameLine"
-pragma_once = true
+pragma_once = false
 cpp_compat = true
 usize_is_size_t = true
 style = "type"
@@ -8,10 +8,52 @@ style = "type"
 header = """
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
+#pragma once
+#include <stdint.h>
 
 //
 // THIS FILE IS AUTO-GENERATED, DO NOT MAKE EDITS DIRECTLY
 //
+
+// https://arrow.apache.org/docs/format/CDataInterface.html#structure-definitions
+// We don't want to bundle nanoarrow or similar just for these two definitions.
+// If you use your own Arrow library, define this macro and
+// typedef FFI_ArrowSchema ArrowSchema;
+// typedef FFI_ArrowArrayStream ArrowArrayStream;
+#ifndef USE_OWN_ARROW
+struct ArrowSchema {
+    const char* format;
+    const char* name;
+    const char* metadata;
+    int64_t flags;
+    int64_t n_children;
+    struct ArrowSchema** children;
+    struct ArrowSchema* dictionary;
+    void (*release)(struct ArrowSchema*);
+    void* private_data;
+};
+struct ArrowArray {
+    int64_t length;
+    int64_t null_count;
+    int64_t offset;
+    int64_t n_buffers;
+    int64_t n_children;
+    const void** buffers;
+    struct ArrowArray** children;
+    struct ArrowArray* dictionary;
+    void (*release)(struct ArrowArray*);
+    void* private_data;
+};
+struct ArrowArrayStream {
+    int (*get_schema)(struct ArrowArrayStream*, struct ArrowSchema* out);
+    int (*get_next)(struct ArrowArrayStream*, struct ArrowArray* out);
+    const char* (*get_last_error)(struct ArrowArrayStream*);
+    void (*release)(struct ArrowArrayStream*);
+    void* private_data;
+};
+typedef struct ArrowSchema FFI_ArrowSchema;
+typedef struct ArrowArrayStream FFI_ArrowArrayStream;
+#endif
 """
 
 [export.rename]
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -1,11 +1,51 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
+#pragma once
+#include <stdint.h>
 
 //
 // THIS FILE IS AUTO-GENERATED, DO NOT MAKE EDITS DIRECTLY
 //
 
-#pragma once
+// https://arrow.apache.org/docs/format/CDataInterface.html#structure-definitions
+// We don't want to bundle nanoarrow or similar just for these two definitions.
+// If you use your own Arrow library, define this macro and
+// typedef FFI_ArrowSchema ArrowSchema;
+// typedef FFI_ArrowArrayStream ArrowArrayStream;
+#ifndef USE_OWN_ARROW
+struct ArrowSchema {
+    const char *format;
+    const char *name;
+    const char *metadata;
+    int64_t flags;
+    int64_t n_children;
+    struct ArrowSchema **children;
+    struct ArrowSchema *dictionary;
+    void (*release)(struct ArrowSchema *);
+    void *private_data;
+};
+struct ArrowArray {
+    int64_t length;
+    int64_t null_count;
+    int64_t offset;
+    int64_t n_buffers;
+    int64_t n_children;
+    const void **buffers;
+    struct ArrowArray **children;
+    struct ArrowArray *dictionary;
+    void (*release)(struct ArrowArray *);
+    void *private_data;
+};
+struct ArrowArrayStream {
+    int (*get_schema)(struct ArrowArrayStream *, struct ArrowSchema *out);
+    int (*get_next)(struct ArrowArrayStream *, struct ArrowArray *out);
+    const char *(*get_last_error)(struct ArrowArrayStream *);
+    void (*release)(struct ArrowArrayStream *);
+    void *private_data;
+};
+typedef struct ArrowSchema FFI_ArrowSchema;
+typedef struct ArrowArrayStream FFI_ArrowArrayStream;
+#endif
 
 #include <stdarg.h>
 #include <stdbool.h>
@@ -134,6 +174,12 @@ typedef enum {
     VX_VALIDITY_ARRAY = 3,
 } vx_validity_type;
 
+typedef enum {
+    VX_CARD_UNKNOWN = 0,
+    VX_CARD_ESTIMATE = 1,
+    VX_CARD_MAXIMUM = 2,
+} vx_cardinality;
+
 /**
  * Equalities, inequalities, and boolean operations over possibly null values.
  * For most operations, if either side is null, the result is null.
@@ -223,6 +269,33 @@ typedef enum {
      */
     LOG_LEVEL_TRACE = 5,
 } vx_log_level;
+
+typedef enum {
+    VX_SELECTION_INCLUDE_ALL = 0,
+    /**
+     * Include rows at the indices in vx_scan_selection.idx.
+     */
+    VX_SELECTION_INCLUDE_RANGE = 1,
+    /**
+     * Exclude rows at the indices in vx_scan_selection.idx.
+     */
+    VX_SELECTION_EXCLUDE_RANGE = 2,
+} vx_scan_selection_include;
+
+typedef enum {
+    /**
+     * No estimate is available.
+     */
+    VX_ESTIMATE_UNKNOWN = 0,
+    /**
+     * The value in vx_estimate.estimate is exact.
+     */
+    VX_ESTIMATE_EXACT = 1,
+    /**
+     * The value in vx_estimate.estimate is an upper bound.
+     */
+    VX_ESTIMATE_INEXACT = 2,
+} vx_estimate_type;
 
 /**
  * Physical type enum, represents the in-memory physical layout but might represent a different logical type.
@@ -370,6 +443,14 @@ typedef struct vx_array_sink vx_array_sink;
 typedef struct vx_binary vx_binary;
 
 /**
+ * A reference to one or more (possibly remote) paths.
+ * Creating vx_data_source opens the first matched path to read the schema.
+ * All other I/O is deferred until a scan is requested. Multiple scans may
+ * be requested from a single data source.
+ */
+typedef struct vx_data_source vx_data_source;
+
+/**
  * A Vortex data type.
  *
  * Data types in Vortex are purely logical, meaning they confer no information about how the data
@@ -404,6 +485,14 @@ typedef struct vx_expression vx_expression;
 typedef struct vx_file vx_file;
 
 /**
+ * A partition is an independent unit of work. Call vx_partition_next repeatedly to
+ * retrieve arrays, then free the partition with vx_partition_free.
+ */
+typedef struct vx_partition vx_partition;
+
+typedef struct vx_scan vx_scan;
+
+/**
  * A handle to a Vortex session.
  */
 typedef struct vx_session vx_session;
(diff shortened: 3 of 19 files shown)


vortex-ffi/cinclude/vortex.h
```diff
@@ -1,10 +1,51 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
+#include <stdint.h>
```
0ax1, line 4: pragma once missing. you can add prefixing a custom line to bindgen.

vortex-ffi/cinclude/vortex.h
```diff
@@ -1,10 +1,51 @@
...
 //
 // THIS FILE IS AUTO-GENERATED, DO NOT MAKE EDITS DIRECTLY
 //
 
+// https://arrow.apache.org/docs/format/CDataInterface.html#structure-definitions
+// We don't want to bundle nanoarrow or similar just for these two definitions.
+// If you use your own Arrow library, define this macro and
+// typedef FFI_ArrowSchema ArrowSchema;
+// typedef FFI_ArrowArrayStream ArrowArrayStream;
+#ifndef USE_OWN_ARROW
```
0ax1, line 15: Do we require ABI compat with their def? And if so how do we make sure that we stay in sync with their def.

vortex-ffi/cinclude/vortex.h
```diff
@@ -224,6 +271,33 @@ typedef enum {
     LOG_LEVEL_TRACE = 5,
 } vx_log_level;
 
+typedef enum {
+    VX_S_INCLUDE_ALL = 0,
```
0ax1: ` VX_S_INCLUDE_ALL = 0` Can we go with a longer more descriptive name, rather than abbrev `S`?

vortex-ffi/cinclude/vortex.h
```diff
@@ -224,6 +271,33 @@ typedef enum {
...
+    VX_S_INCLUDE_RANGE = 1,
+    /**
+     * Exclude rows at the indices in vx_scan_selection.idx.
+     */
+    VX_S_EXCLUDE_RANGE = 2,
+} vx_scan_selection_include;
+
+typedef enum {
+    /**
+     * No estimate is available.
```
0ax1, line 287: Unclear from this ctx alone for what estimation this enum is used for.

vortex-ffi/cinclude/vortex.h
```diff
@@ -369,6 +445,14 @@ typedef struct vx_array_sink vx_array_sink;
  */
 typedef struct vx_binary vx_binary;
 
+/**
+ * A reference to one or more (possibly remote) files.
+ * Creating vx_data_source opens the first matched file to read the schema.
+ * All other I/O is deferred until a scan is requested. Multiple scans may
+ * be requested from a single data source.
+ */
+typedef struct vx_data_source vx_data_source;
```
0ax1, line 451: A data source doens't need to be a file. Ref the Scan API RFC, a data source can be:
```
a Vortex file
a Vortex layout tree
a Vortex table
a Parquet file
an Iceberg table
a remote storage service exposing scan endpoints
```

vortex-ffi/cinclude/vortex.h
```diff
@@ -435,6 +527,68 @@ typedef struct {
...
+
+typedef void (*vx_glob)(const char *glob, vx_glob_callback callback, vx_error **err);
+
+/**
+ * Options for creating a data source.
+ *
+ * "files" is the only required field.
+ * If any "fs_*" callback is provided, all of them must be provided.
+ */
+typedef struct {
```
0ax1, line 530: This mixes what to scan with how to do that.

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+        Ok(0)
+    })
+}
+
+// Scan partition to ArrowArrayStream.
+// Consumes partition fully: subsequent calls to vx_partition_scan_arrow or
+// vx_partition_next are undefined behaviour.
+// This call blocks current thread until underlying stream is fully consumed.
+//
+// Caller doesn't have to free partition after calling this function.
```
0ax1: More like must not free `partition` right? We take ownership of the heap allocation in `vx_partition_scan_arrow`.

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_scan_next(
+    scan: *mut vx_scan,
+    err: *mut *mut vx_error,
+) -> *mut vx_partition {
+    let scan = vx_scan::as_mut(scan);
+    let mut scan = scan.lock().expect("failed to lock mutex");
+    let scan = &mut *scan;
+    unsafe {
+        let ptr = scan as *mut VxScanState;
```
0ax1: Aren't we aliasing the guard here?

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+/// On success returns a partition.
+/// On exhaustion (no more partitions in scan) returns NULL but doesn't set
+/// "err".
+/// On error returns NULL and sets "err".
+///
+/// This function is thread-safe: callers running a multi-threaded pipeline
+/// should call it concurrently and dispatch each partition to a dedicated
+/// worker thread.
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_scan_next(
```
0ax1: should this be suffixed with `_partition`?

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+                    let partition = vx_partition::new(Box::new(partition));
+                    ptr::write(ptr, VxScanState::Started(stream));
+                    Ok(partition)
+                }
+                None => on_finish(),
+            }
+        };
+
+        let owned = ptr::read(ptr);
+        try_or_default(err, || match owned {
```
0ax1, line 299: Aren't we fully serializing here? `on_stream` holds on to the lock guard, even if aliased rn, and the call into `match RUNTIME.block_on(stream.next())`.

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+            )))))
+        })
+    })
+}
+
+/// Scan's dtype.
+/// Must be called before first call to vx_scan_next.
+/// On error returns NULL and sets "err".
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_scan_dtype(
```
0ax1, line 247: The doc should note that the returned dtype needs to be freed by the caller.

vortex-ffi/cinclude/vortex.h
gatesn: It's pretty standard practice to copy/paste the arrow headers. I don't think they've changed in many years

vortex-ffi/src/scan.rs
gatesn: Why is there a lock at all...?

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+            ),
+        };
+        let array_stream = partition.execute()?;
+        let dtype = array_stream.dtype();
+
+        let schema = dtype.to_arrow_schema()?;
+        let schema = Arc::new(schema);
+        let data_type = DataType::Struct(schema.fields().clone());
+
+        let session = vx_session::as_ref(session);
```
0ax1, line 358: In theory, the session ref here can get deallocated before it is used in the closure. So the layz iter we return here might refer to a dangling pointer. We at least need to mention in the docs somewhere that the session should stay alive.

vortex-ffi/cbindgen.toml
```diff
@@ -8,10 +8,51 @@ style = "type"
...
 // THIS FILE IS AUTO-GENERATED, DO NOT MAKE EDITS DIRECTLY
 //
+
+// https://arrow.apache.org/docs/format/CDataInterface.html#structure-definitions
+// We don't want to bundle nanoarrow or similar just for these two definitions.
+// If you use your own Arrow library, define this macro and
+// typedef FFI_ArrowSchema ArrowSchema;
+// typedef FFI_ArrowArrayStream ArrowArrayStream;
+#ifndef USE_OWN_ARROW
+struct ArrowSchema {
```
0ax1, line 24: Refer to arrow docs?

vortex-ffi/CMakeLists.txt
```diff
@@ -94,10 +94,9 @@ set_target_properties(vortex_ffi_shared PROPERTIES
 
 if (BUILD_TESTS)
     enable_language(CXX)
-    set(CMAKE_CXX_STANDARD 20)
+    set(CMAKE_CXX_STANDARD 23)
```
gatesn: Does this mean we cannot be consumed by projects <23?

vortex-ffi/src/scan.rs
0ax1: So the pointer is invalid after the call.

vortex-ffi/src/scan.rs
gatesn: We should just return a borrowed DType, not clone it

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+/// On error returns NULL and sets "err".
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_scan_dtype(
+    scan: *const vx_scan,
+    err: *mut *mut vx_error,
+) -> *const vx_dtype {
+    try_or(err, ptr::null(), || {
+        let scan = vx_scan::as_ref(scan).lock().expect("failed to lock mutex");
+        let VxScanState::Pending(ref scan) = *scan else {
+            vortex_bail!("dtype unavailable: scan already started");
```
gatesn, line 254: Why is this true...?

vortex-ffi/CMakeLists.txt
0ax1: If someone wants to build the tests they need 23 support in the compiler.  @myrrc Why switch on this actually. Can we try to stick everything to 20?

0ax1: More of a consistency q.

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+        None
+    } else {
+        Some(vx_expression::as_ref(opts.filter).clone())
+    };
+
+    let selection = &opts.selection;
+    let selection = match selection.include {
+        vx_scan_selection_include::VX_S_INCLUDE_ALL => Selection::All,
+        vx_scan_selection_include::VX_S_INCLUDE_RANGE => {
+            let buf = unsafe { slice::from_raw_parts(selection.idx, selection.idx_len) };
```
0ax1, line 158: If selection is null `from_raw_parts` is UB. Should we explicitly assert !null?

vortex-ffi/src/data_source.rs
```diff
@@ -0,0 +1,296 @@
...
+    rows: u64,
+}
+
+/// Write data source's row count estimate into "row_count".
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_data_source_get_row_count(
+    ds: *const vx_data_source,
+    row_count: *mut vx_data_source_row_count,
+) {
+    let rc = unsafe { &mut *row_count };
```
0ax1, line 130: Same here should we assert for pointers being not null?

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+///
+/// On success returns a partition.
+/// On exhaustion (no more partitions in scan) returns NULL but doesn't set
+/// "err".
+/// On error returns NULL and sets "err".
+///
+/// This function is thread-safe: callers running a multi-threaded pipeline
+/// should call it concurrently and dispatch each partition to a dedicated
+/// worker thread.
+#[unsafe(no_mangle)]
```
0ax1, line 271: Should we have a test where this is called from multiple threads?

0ax1 (commented):
We should prob consider putting ` #[deny(missing_docs)]` for `ffi` on a module level.

vortex-ffi/test/scan.cpp
```diff
@@ -0,0 +1,602 @@
...
+UniqueArrayStream sample_array_stream() {
+    UniqueSchema schema = sample_schema();
+    UniqueArray arr;
+
+    REQUIRE(ArrowArrayInitFromSchema(arr.get(), schema.get(), nullptr) == NANOARROW_OK);
+    REQUIRE(ArrowArrayStartAppending(arr.get()) == NANOARROW_OK);
+
+    auto ages = sample_age();
+    auto heights = sample_height();
+    for (auto [name, age] : std::views::zip(ages, heights) ) {
```
0ax1: name and age are swapped here

vortex-ffi/src/data_source.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Options for creating a data source.
+///
+/// "files" is the only required field.
+/// If any "fs_*" callback is provided, all of them must be provided.
+#[repr(C)]
+#[cfg_attr(test, derive(Default))]
+pub struct vx_data_source_options {
+    /// Required: files names. May be a glob pattern like "*.vortex".
+    pub files: *const c_char,
```
0ax1: Right now, only `files` is wired in. Should the PR only ship the options that the API currently provides?

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+            let buf = Buffer::copy_from(buf);
+            Selection::ExcludeByIndex(buf)
+        }
+    };
+
+    let ordered = opts.ordered;
+
+    let start = opts.row_range_begin;
+    let end = opts.row_range_end;
+    let row_range = (start > 0 || end > 0).then_some(Range { start, end });
```
0ax1, line 174: `let row_range = (start > 0 || end > 0).then_some(Range { start, end });`  I guess there's no scenario where the caller actually wants 0..0 which would be the empty range but here translates into `None` which leads to all rows being read.

vortex-ffi/src/scan.rs
```diff
@@ -0,0 +1,686 @@
...
+
+    let limit = (opts.limit != 0).then_some(opts.limit);
```
0ax1, line 176: Same question here.

vortex-ffi/test/scan.cpp
```diff
@@ -0,0 +1,602 @@
...
+            REQUIRE(age_left == age_right);
+        }
+    }
+}
+
+TEST_CASE("Scan Arrow schema", "[scan]") {
+    vx_session *session = vx_session_new();
+    defer { vx_session_free(session); };
+
+    TempPath path = write_sample(session, fs::current_path() / "arrow-scan-schema.vortex");
```
0ax1: nit: Isn't it a bit brittle relying on the `x.vortex` suffix to be distinct? If I copy and paste a test and change the name it'll race.

vortex-ffi/cinclude/vortex.h
myrrc: It's not missing, generated after this include, I'll move it.

vortex-ffi/CMakeLists.txt
myrrc: FFI standard itself is C17, C++ is only for the tests, but yeah, let's stay at C++20.

vortex-ffi/src/scan.rs
myrrc: We need a mutex so as worker threads could get partitions without outer synchronization.

gatesn: Feels like that's the job of the caller, we should just document that it is not thread-safe

myrrc: Ok, agreed, let's do as little work as possible.

vortex-ffi/src/scan.rs
myrrc: In Rust API we consume self when scan is turned into partition stream, so I matched that. In theory we can save the dtype for FFI but is there any point in doing that?

vortex-ffi/src/scan.rs
myrrc: The main reason of this is to make vx_scan_options zero-initialization-friendly, i.e. provide sane defaults when user does `opts = {}`. This is documented, and it doesn't make sense to request a scan with 0 rows in return, so I think we're good with this option, but I'd like a second opinion here @joseph-isaacs

vortex-ffi/src/data_source.rs
myrrc: +1

vortex-ffi/src/scan.rs
myrrc: Session deallocation is manual, so it won't be deallocated at least until this function returns.

0ax1 (approved):

PR #7214, opened 2026-03-31
fix[vortex-array]: fix overflow on FSL element take indices

asubiotto:
The take index types used on an FSL might be valid narrow types. This type was used as indices to take the inner elements, which could result in an overflow with long enough lists.

## Summary

Closes: #000

### DIFF
--- a/vortex-array/src/arrays/fixed_size_list/tests/take.rs
+++ b/vortex-array/src/arrays/fixed_size_list/tests/take.rs
@@ -9,6 +9,7 @@ use super::common::create_empty_fsl;
 use super::common::create_large_fsl;
 use super::common::create_nullable_fsl;
 use super::common::create_single_element_fsl;
+use crate::ArrayRef;
 use crate::DynArray;
 use crate::IntoArray;
 use crate::arrays::FixedSizeListArray;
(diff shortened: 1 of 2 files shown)


asubiotto:
We maybe could use usize for the element indices to avoid the generic type parameter

vortex-array/src/arrays/fixed_size_list/tests/take.rs
```diff
@@ -130,6 +131,38 @@ fn test_take_fsl_with_null_indices_preserves_elements() {
...
+        buffer![0u8; 320].into_array(), 16,
+        Validity::from_iter((0..20).map(|i| i != 5)), 20,
+    ),
+    buffer![0u8, 16, 5].into_array(),
+    FixedSizeListArray::new(
+        buffer![0u8; 48].into_array(), 16,
+        Validity::from_iter([true, true, false]), 3,
+    ),
+)]
+fn test_element_index_overflow(
```
connortsui20, line 157: I tried running this on develop and it did not fail, is there something else that happens here?

connortsui20:
seems like there is a nice perf gain from this, so might as well do this?

vortex-array/src/arrays/fixed_size_list/tests/take.rs
```diff
@@ -130,6 +131,38 @@ fn test_take_fsl_with_null_indices_preserves_elements() {
...
+    #[case] fsl: FixedSizeListArray,
+    #[case] indices: ArrayRef,
+    #[case] expected: FixedSizeListArray,
+) {
+    let result = fsl.take(indices.to_array()).unwrap();
```
robert3005, line 162: To actually test the take logic we need an execute here

robert3005: I will make a pr to change the macro. This is a big footgun

robert3005 (approved):

connortsui20:
um... do we want to write a regression test that actually fails?

robert3005:
We have it. I already merged a change to the macro that exposes the bug

asubiotto:
Yeah, sorry it got lost in the rstest parameterization. The take was lazily wrapping the fsl in a dictionary

asubiotto:
Still couldn't get it to fail without a to_fixed_size_list. Here's a PR with the test changes: https://github.com/vortex-data/vortex/pull/7229

robert3005:
I made the macro expand the left hand side as actual vs right as expected. This test was passing them the other way... maybe we need to do both

PR #7181, opened 2026-03-26
Array VTables Migration 1a

gatesn:
This introduces the end-state struct `Array<V>` per the vtables design doc: https://docs.vortex.dev/developer-guide/internals/vtables

It keeps the old ArrayAdapter<V> around for now, to be removed at a later date.

# Array VTable Migration Plan

## Context

The Array vtable system is the last major plugin point that hasn't been migrated to the unified
vtable pattern described in `docs/developer-guide/internals/vtables.md`. ScalarFn, AggregateFn,
and ExtDType already follow the target pattern. Array is the largest migration (~40 encodings
across vortex-array, encodings/*, vortex-python, vortex-cuda).

**Current state:** `VTable` trait + `ArrayAdapter<V>` (#[repr(transparent)]) + `DynArray` public
sealed trait + `vtable!` macro. Each concrete array struct (e.g. `PrimitiveArray`) owns dtype, len,
stats and derefs to `dyn DynArray` via unsafe casts through `ArrayAdapter`.

**Target state:** `ArrayVTable` trait + `Array<V>` generic data struct (owns dtype, len, stats,
derefs to `V::Array`) + sealed `DynArray` thin forwarder + `ArrayRef` newtype. Follows the
`FooRef → DynFoo → Foo<V> → FooVTable` dispatch chain.

---

## Phase 1: Introduce `Array<V>`, rename trait, change signatures

**Goal:** Replace the core VTable machinery in one atomic change. All encodings switch from
`ArrayAdapter<V>` to `Array<V>` at once since VTable method signatures change from `&Self::Array`
to `&Array<Self>`.

This is a large but mechanical change — each encoding's update follows a predictable recipe.

### 1.1 Rename `VTable` → `ArrayVTable`

- `vortex-array/src/vtable/mod.rs`: rename trait
- Codebase-wide find-and-replace: `impl VTable for` → `impl ArrayVTable for`, `V: VTable` →
  `V: ArrayVTable`, etc.

### 1.2 Define `Array<V: ArrayVTable>`

New file `vortex-array/src/vtable/typed.rs`:

```rust
pub struct Array<V: ArrayVTable> {
    vtable: V,
    dtype: DType,
    len: usize,
    array: V::Array,    // encoding-specific data
    stats: ArrayStats,
}

impl<V: ArrayVTable> Deref for Array<V> {
    type Target = V::Array;
    fn deref(&self) -> &V::Array { &self.array }
}
```

[...]

PR #7215, opened 2026-03-31
Array VTables

gatesn:
See https://github.com/vortex-data/vortex/pull/7181

This is a large break that changes how Vortex arrays are held in memory. Previously they were modeled as Arc<dyn Array> where all array impls used a vtable macro to unsafely transmute into a dyn Array.

This was cute, but meant that we were performing a lot of unnecessary clones and Arc allocations for things that should never have required them. We were also limited by how expressive we could make APIs since Arc is considered a foreign type.

This PR makes ArrayRef a struct, makes typed arrays a struct of `Array<V: VTable>`, and adds an `ArrayView<'a, V>`. This means all arrays are Arc allocated up-front in any of these forms, and we pass the Arc allocation around everywhere allowing us to cheaply clone in either the type-erased or typed structs.

The main user-facing change is that array constructors should now live on the vtable struct, e.g. `Primitive::new(...)` instead of `PrimitiveArray::new(...)`


.github/workflows/ci.yml
```diff
@@ -746,7 +746,7 @@ jobs:
           tool: cargo-codspeed
       - name: Build benchmarks
         env:
-          RUSTFLAGS: "-C target-feature=+avx2"
+          RUSTFLAGS: "-C target-feature=+avx2 -C debug-assertions=yes"
```
lwwmanning: might not want to merge this?

vortex-array/src/array/vtable/dyn_.rs
```diff
@@ -202,38 +201,21 @@ impl<V: VTable> DynVTable for V {
...
     ) -> VortexResult<Option<ArrayRef>> {
-        let Some(result) = V::execute_parent(downcast::<V>(array), parent, child_idx, ctx)? else {
+        let Some(result) = V::execute_parent(array.as_::<V>(), parent, child_idx, ctx)? else {
             return Ok(None);
         };
 
         if cfg!(debug_assertions) {
             vortex_ensure!(
-                result.as_ref().len() == parent.len(),
+                result.len() == parent.len(),
```
lwwmanning, line 210: probably worth doing this check even outside the debug_assertions block? dtype comparison can be expensive, but length should be quite fast

vortex-array/src/arrays/scalar_fn/rules.rs
```diff
@@ -105,18 +104,20 @@ impl ArrayParentReduceRule<ScalarFnVTable> for ScalarFnSliceReduceRule {
...
         ))
     }
 }
 
 #[derive(Debug)]
 struct ScalarFnAbstractReduceRule;
 impl ArrayReduceRule<ScalarFnVTable> for ScalarFnAbstractReduceRule {
-    fn reduce(&self, array: &ScalarFnArray) -> VortexResult<Option<ArrayRef>> {
+    fn reduce(&self, array: ArrayView<'_, ScalarFnVTable>) -> VortexResult<Option<ArrayRef>> {
+        // TODO(ngates): blergh!
```
lwwmanning: what does "blergh" mean? :D

lwwmanning (approved):
minor nits, but worth merging asap as long as all the benchmark suites succeed / continue to be flat / better. it's a clear improvement from a code concepts point of view

PR #7218, opened 2026-03-31
Support partitionBy in VortexSparkDataSource

robert3005:
Support partitionBy in spark writer


java/vortex-spark/src/main/java/dev/vortex/spark/write/PartitionedVortexDataWriter.java
```diff
@@ -0,0 +1,428 @@
...
+            }
+            ResolvedTransform rt = resolvedTransforms[i];
+            sb.append(URLEncoder.encode(rt.directoryKey, StandardCharsets.UTF_8));
+            sb.append("=");
+
+            String value = evaluateTransform(rt, row);
+            sb.append(URLEncoder.encode(value, StandardCharsets.UTF_8));
+        }
+        return sb.toString();
+    }
```
a10y, line 204: is this code really not in spark somewhere

a10y (approved):

java/vortex-spark/src/main/java/dev/vortex/spark/write/PartitionedVortexDataWriter.java
robert3005: I spent a long time looking, let me look again. In all fairness all of this logic is not datasource specific. I am very confused why they don't have shared handling. It could also be that we have to be a FileSource and I think while initially that's simpler there's things that it makes harder in the long term

robert3005: All of this logic exists only for file datasources but then you're married to hadoop. I think we are fine to reimplement it.

robert3005:
We don't implement filter pushdown yet so even though we can read and write parititions we don't prune them. Also we don't remove the parititon columns from the data yet

robert3005:
This pr needs more work - we need to remove partition values from data

kesavkolla:
It would be nice to see full support for hive style partitioning with vortex. For both read/write and also filter pushdown

robert3005:
This pr added everything but filter pushdown. We don’t have filter pushdown in spark data source at all right now

PR #7219, opened 2026-03-31
buffered strategy to not use eof for the final chunk

### LINKED ISSUE #7234: Zone maps are not all at the end of the file, instead at the end of a column, even if that's very early in the file.

### LINKED ISSUE #7236: Chunked of Dicts splits up codes and values (vs values living at the end of the codes for just that chunk)

onursatici:
## Summary

fixes buffered layout writer so it doesn't write the final chunk on the eof pointer. Eof should only be used for data that the writer wants to place at the end of the file. Buffered writer was writing regular buffered data to there which did mess up ordering of some segments.

Previously struct writer was using a transposed stream without spawning a task per column, on that world buffering was deadlocky. That is changed for a while to spawn now, so we should be deadlock safe. 

I did try converting all clickbench files repeatedly, as well as the public bi datasets and randomly generated wide tables but I couldn't deadlock this. 

fixes https://github.com/vortex-data/vortex/issues/7234
fixes https://github.com/vortex-data/vortex/issues/7236

## Testing

add a vortex file test that asserts the dict layout segments are in the right order, as well as zone maps across columns

### DIFF
--- a/vortex-file/src/tests.rs
+++ b/vortex-file/src/tests.rs
@@ -62,6 +62,7 @@ use vortex_buffer::ByteBufferMut;
 use vortex_buffer::buffer;
 use vortex_error::VortexResult;
 use vortex_io::session::RuntimeSession;
+use vortex_layout::Layout;
 use vortex_layout::scan::scan_builder::ScanBuilder;
 use vortex_layout::session::LayoutSession;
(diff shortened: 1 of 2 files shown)


gatesn (approved):
Let's try it!

robert3005 (approved):

PR #7220, opened 2026-03-31
use trait upcasting and drop the as_any/as_any_mut boilerplate

a10y:
Since Rust 1.86 (which is before our MSRV) there has been builtin trait upcasting, meaning instead of forcing all our traits to have a dummy `fn as_any(&self) -> &dyn Any` methods with trivial impls, we can instead just take a pointer to the trait object and do `as &dyn Any`.

This reduces the amount of boilerplate in trait impls, which is nice. Another benefit is that in the past blanket impls like `SessionVar` would end up getting pulled in for dispatch of calls to as_any() which has caused me frustration. Now the dispatch is clearer.

This isn't a breaking change b/c it doesn't touch the VTables, which is the implementer surface area

### DIFF
--- a/encodings/fastlanes/src/bitpacking/vtable/mod.rs
+++ b/encodings/fastlanes/src/bitpacking/vtable/mod.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::any::Any;
 use std::hash::Hash;
 use std::sync::Arc;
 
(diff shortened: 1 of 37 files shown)


robert3005 (approved):

AdamGS (approved):

robert3005:
it's interesting that this had changed execution of take benchmarks?

a10y:
So this is the diff that fixes things:

image

The issue is that ArrayRef is an `Arc<dyn DynArray>` which is also a &dyn DynArray, but when we try and match on it we actually want to match on the un-Arc'ed thing. This syntax doesn't really let you do that. I think we're better off keeping the explicit trait methods which handle dispatching through the Arc.

Issue #7210, opened 2026-03-31
Default write strategy is a little unexpected


gatesn:
- [ ]  Why do we store sum in the zone map?
- [x]  min/max_is_truncated does not compress to constant if possible in zone map
- [x]  sum stat in zone map is not compressed (e.g. narrowed, or bitpacked)

robert3005:
The is_truncated is there for fixed sized columns since we thought it would be simpler to have uniform handling and it’s a constant bool column

robert3005:
FWIW the sum in zone map is for pushdown in DF. However, DF had a very bad performance when using sum stat. They have fixed the behaviour on develop https://github.com/apache/datafusion/pull/20768 but it's not released yet

robert3005:
@gatesn do you have example of where SUM is not compressed? I think the value is so small that it doesn't compress well with anything we have beyond constant

PR #7221, opened 2026-03-31
Add compressor for constant nonnullable and all valid bool arrays

robert3005:
Add compressor for constant bool arrays.

This should make part of #7210 less unexpected


vortex-compressor/src/stats/bool.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Clone, Debug)]
+pub struct BoolStats {
+    pub(crate) src: BoolArray,
+    /// Number of null values.
+    pub(crate) null_count: u32,
+    /// Number of non-null values.
+    pub(crate) value_count: u32,
+    /// Number of distinct non-null values (0, 1, or 2).
+    pub(crate) distinct_values_count: u32,
+}
```
joseph-isaacs, line 21: why do we have this? Do you expect other schemes? Surely we can be smarter with what we count here?

robert3005: sorry, this is dumb

a10y (approved):

vortex-compressor/src/stats/bool.rs
connortsui20, line 1: I think a lot of the generate methods are not needed since we are only ever going to have a few compression schemes for bool (so we don't really need options)

vortex-compressor/src/builtins/constant.rs
```diff
@@ -22,6 +23,58 @@ use crate::ctx::CompressorContext;
...
+    }
+
+    fn compress(
+        &self,
+        _compressor: &CascadingCompressor,
+        data: &mut ArrayAndStats,
+        _ctx: CompressorContext,
+    ) -> VortexResult<ArrayRef> {
+        let stats = data.bool_stats();
+        Ok(ConstantArray::new(stats.source().scalar_at(0)?, stats.source().len()).into_array())
```
connortsui20, line 74: So technically we don't need to check if this array is_constant but this feels a bit dangerous.

robert3005: we check it in expected_compression_ratio

vortex-compressor/src/stats/bool.rs
robert3005: I removed the additional methods

connortsui20:
Heads up: https://github.com/vortex-data/vortex/pull/7230

PR #7223, opened 2026-03-31
perf: speed up TryFrom<Scalar> for Primitive

joseph-isaacs:
Simple


robert3005 (approved):

vortex-array/src/scalar/convert/primitive.rs
robert3005, line 88: you probs want to change this as well

PR #7232, opened 2026-04-01
Require two humans for bot authored PRs

gatesn:


AdamGS:
As it seems like we're going to allow cladue to open PRs directly, can we write down our general AI approach/policy and discuss it?

lwwmanning (approved):

PR #7238, opened 2026-04-01
feat[array]: owned matchers

joseph-isaacs:
We want to be able to downcast arrays based on a matcher into an owned type.

open questions:

We could extend array try_into to use a matcher


gatesn (changes requested):
This is subsumed by #7215

joseph-isaacs:
Mind showing which part?

joseph-isaacs:
@gatesn I think we still want this correct?

PR #7240, opened 2026-04-01
fix mut slots

joseph-isaacs:
Removes the temp `with_slots` with a `slot_mut` method this allows for arrays mutate the children inplace.


encodings/fsst/src/array.rs
```diff
@@ -519,9 +503,9 @@ impl FSSTArray {
     }
 }
 
-impl ValidityChild<FSST> for FSST {
-    fn validity_child(array: &FSSTArray) -> &ArrayRef {
-        &array.codes_array
+impl ValidityVTable<FSST> for FSST {
+    fn validity(array: &FSSTArray) -> VortexResult<Validity> {
+        Ok(array.codes().validity())
```
robert3005: you have the validity logic in codes that's a lot simpler than getting codes

vortex-array/src/array/mod.rs
```diff
@@ -345,25 +345,20 @@ impl dyn DynArray + '_ {
...
-            slot_idx < nslots,
-            "slot index {} out of bounds for array with {} slots",
-            slot_idx,
-            nslots
-        );
-        let mut slots = self.slots().to_vec();
-        slots[slot_idx] = Some(replacement);
+    /// Requires unique ownership of the `ArrayRef` (panics if Arc refcount > 1).
+    pub fn take_slot(self: &mut ArrayRef, slot_idx: usize) -> Option<ArrayRef> {
+        let vtable = self.vtable().clone_boxed();
```
robert3005, line 352: part of me wants to do unsafe things here to erase lifetimes and get rid of this clone. Replacing a slot wouldn't touch the vtable

joseph-isaacs: This free for zst (all out current VTables).

onursatici: yea we discussed this offline with Joe, it is a bit weird to have this hidden cost creep up depending on whether you cloned the array before or not. if Arc refcount is one we do a different thing, so that kinda defeats the purpose of ArrayRef.clone being cheap. Clone is cheap but execute just became more expensive because you cloned

onursatici (changes requested):
can you rebase so we know for certain that the wide table decompression is still fast after these changes. Currently I don't think your branch includes today's fix

joseph-isaacs:
I did locally and it was. I can do this but would prefer to fix after the fact

onursatici (approved):

vortex-array/src/arrays/masked/execute.rs
```diff
@@ -204,7 +204,7 @@ fn mask_validity_variant(
     mask: &Mask,
     ctx: &mut ExecutionCtx,
 ) -> VortexResult<VariantArray> {
-    let child = array.child().clone();
```
robert3005, line 207: can you remove the clone here?

joseph-isaacs: is a follow up okay?

robert3005 (approved):

vortex-array/src/arrays/masked/execute.rs
robert3005: yes

gatesn (changes requested):
The regressions seem real? 20%?

robert3005:
How did you conclude they're real? The result analysis shows `No clear signal (environment too noisy confidence)`. We can rerun the benchmarks but I don't think there's actual regression here

PR #7247, opened 2026-04-01
Crosshairs for filemap

gatesn:
This enables hovering and selecting segments from the file map, showing the selected segments as highlighted in the layout tree. It also has a tooltip showing the segment number, path, byte offset, and size.

Screenshot 2026-04-01 at 16 20 12


0ax1 (approved):

0ax1:
preview?

PR #7249, opened 2026-04-01
docs: add back array children docs moved from slots

joseph-isaacs:
Add back docs for each array child removed in https://github.com/vortex-data/vortex/pull/6870


connortsui20 (approved):
nice

PR #7251, opened 2026-04-01
Normalized Vector

connortsui20:
## Summary

Continuation of https://github.com/vortex-data/vortex/pull/7107.

Since we want to add a bunch of new vector encodings that rely on unit normalization (spherical transform, turboquant), we would like to introduce this as an encoding.

Note that once we eventually have https://github.com/vortex-data/vortex/pull/6717, the operations will be a lot less ugly.

Also, we can only do the broadcasting trick with `RunEnd` and `Sequence` array multiplication on the decompress side, we can't do it on the compress side because we may potentially divide by 0 (and we can't intercept that and say return 0 in that case).

There is an open question on if, instead of representing this as a physical encoding, we should have scalar encodings like this one (others include FoR and ZigZag) be scalar functions. Whatever we decide, we should probably just merge this first as we can always change this later.

## Open Questions

- Do we want a helper function for creating a new expression and applying it to an array?

## API Changes

Adds a new vector encoding to the `vortex-tensor` crate.

## Testing

Basic tests.

### DIFF
--- a/vortex-tensor/public-api.lock
+++ b/vortex-tensor/public-api.lock
@@ -2,6 +2,122 @@ pub mod vortex_tensor
 
 pub mod vortex_tensor::encodings
 
+pub mod vortex_tensor::encodings::norm
+
+pub struct vortex_tensor::encodings::norm::NormVector
+
+impl core::clone::Clone for vortex_tensor::encodings::norm::NormVector
+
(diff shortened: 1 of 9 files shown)


connortsui20:
Discussed offline, we don't want this yet (we will hardcode it into the encodings we want)

PR #7254, opened 2026-04-01
Silence unused warning

connortsui20:
## Summary

```
warning: value assigned to `nbytes` is never read
  --> vortex-layout/src/layouts/buffered.rs:75:25
   |
75 |                         nbytes -= chunk_size;
   |                         ^^^^^^^^^^^^^^^^^^^^
   |
   = help: maybe it is overwritten before being read?
   = note: `#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default
```

Because we break immediately after `nbytes` is never used again.

## Testing

N/A


robert3005:
this doesnt seem right?

a10y (approved):
ah yea because we break right after

PR #7255, opened 2026-04-02
RLE handles decompression of indices where invalid positions are clobbered

robert3005:
The bug here is that invalid positions get clobbered and there's no values for
indices to refer to

This was found by a fuzzer but autoclosed as the generation logic changed.


robert3005:
I added handling of cases where the underlying invalid positions are completely clobbered with random values. This should handle all the cases now

encodings/fastlanes/src/rle/array/rle_compress.rs
```diff
@@ -51,46 +52,50 @@ where
...
     }
 
     if !remainder.is_empty() {
         // Repeat the last value for padding to prevent
         // accounting for an additional value change.
         let mut padded_chunk = [values[len - 1]; FL_CHUNK_SIZE];
         padded_chunk[..remainder.len()].copy_from_slice(remainder);
-        process_chunk((len / FL_CHUNK_SIZE) * FL_CHUNK_SIZE, &padded_chunk);
+        let last_idx_chunk = indices_uninit
+            .last_mut()
```
0ax1: Can `.last_mut()` on the chunked spare capacity return the wrong chunk if the allocator over-provisions the buffer?

encodings/fastlanes/src/rle/array/rle_decompress.rs
```diff
@@ -55,42 +54,61 @@ where
...
+            // Single-value chunk: fill directly to avoid out-of-bounds index
+            // access. The indices may contain values other than 0 when they
+            // have been further compressed (e.g., as a masked constant).
+            buffer_values.fill(chunk_values[0]);
+        } else if has_invalid {
+            // When the indices array has invalid (null) positions, those
+            // positions may contain arbitrary garbage values after further
+            // compression. Clamp all indices into [0, num_chunk_values) to
+            // prevent out-of-bounds access in the fastlanes decoder.
+            let mut sanitized = *chunk_indices;
```
0ax1: `let mut sanitized = *chunk_indices;` this copies 1024 elements, right?

encodings/fastlanes/src/rle/array/rle_compress.rs
robert3005: above when we split it into chunks we discard the last uneven chunk that comes from overallocation. This would be the even chunk

encodings/fastlanes/src/rle/array/rle_decompress.rs
robert3005: yes, I realise that since we mutate we should just 0 allocate this

encodings/fastlanes/src/rle/array/rle_compress.rs
0ax1:
```
Expected: allocator returns exactly padded_len (2048 u16s)
═══════════════════════════════════════════════════════════

spare_capacity_mut() → [u16; 2048]
as_chunks_mut::<1024>() → (&mut [[u16; 1024]; 2], &mut [])
                            ▲                 ▲
                          chunk 0           chunk 1
                          (loop)          (last_mut ✓)


Small overallocation: allocator returns 2560 u16s (+512)
════════════════════════════════════════════════════════

spare_capacity_mut() → [u16; 2560]
as_chunks_mut::<1024>() → (&mut [[u16; 1024]; 2], &mut [u16; 512])
                            ▲                 ▲           ▲
                          chunk 0           chunk 1     remainder
                          (loop)          (last_mut ✓)  (discarded)

                          ↑ robert3005's argument: overallocation
                            lands in the discarded remainder


Large overallocation: allocator returns 4096 u16s (+2048)
═════════════════════════════════════════════════════════

spare_capacity_mut() → [u16; 4096]
as_chunks_mut::<1024>() → (&mut [[u16; 1024]; 4], &mut [])
                            ▲          ▲          ▲          ▲
                          chunk 0    chunk 1    chunk 2    chunk 3
                          (loop)    NEEDED     (ghost)   (last_mut ✗)
                                    but
                                    SKIPPED!
                                      ▲
                                      │
                              uninitialized memory
                              exposed by set_len(2048)
```

encodings/fastlanes/src/rle/array/rle_decompress.rs
robert3005: `%` here was excessive, this is a simple min

encodings/fastlanes/src/rle/array/rle_compress.rs
robert3005: ok, haven't thought about it overallocating to that level but you have a point

robert3005: addressed

0ax1 (approved):

PR #7264, opened 2026-04-02
Add comment showing change in benchmark file sizes

a10y:
Adds an additional comment as part of the PR benchmarks reporting that shows the change in size of SQL benchmark files after applying the diff from the PR.

Only applies to vortex and vortex-compact

Similar to our other benchmarks, tracks state over time via a GZipped JSON file stored in S3 bucket.

I've verified the Python scripts locally

### DIFF
--- a/.github/workflows/sql-benchmarks.yml
+++ b/.github/workflows/sql-benchmarks.yml
@@ -223,7 +223,6 @@ jobs:
             ${{ matrix.scale_factor && format('--scale-factor {0}', matrix.scale_factor) || '' }}
 
       - name: Install uv
-        if: inputs.mode == 'pr'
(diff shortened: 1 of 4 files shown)


joseph-isaacs:
Thoughts on having a single file per benchmark (I think we should move other benchmarks in that direction too?)

a10y:
@joseph-isaacs I don't really care, I was just trying to match what we have today

a10y:
I went ahead and updated it to use different file in S3 bucket for each benchmark

AdamGS (approved):
sgtm

connortsui20:
some things failed but Im pretty sure its just because of github falling over, lets just run them again

Issue #6854, opened 2026-03-09
Tracking Issue: UUID Extension Type


connortsui20:
# UUID

This is a tracking issue for adding a UUID extension type to Vortex.

UUID is a 16-byte (128-bit) type that represents a universally unique identifier.

Our implementation will make use of the popular [`uuid`](https://docs.rs/uuid/latest/uuid/) Rust crate.

TODO?

## Prior Art

- Arrow has a canonical extension type [UUID](https://arrow.apache.org/docs/format/CanonicalExtensions.html#uuid)

## Steps / History

Related: [RFC#5](https://github.com/vortex-data/rfcs/blob/develop/proposed/0005-extension.md)

- [X] https://github.com/vortex-data/vortex/pull/6832
- [x] Validate that the array contains valid UUID (according to optional version as well)
- [ ] Add to Arrow export
- [ ] Add to DuckDB export
- [ ] Add expression implementations
  - [ ] `Cast` for `to_string` and `from_string`? How do we do pluggable expressions? 
  - [ ] Figure out what expressions are needed
- [ ] ? 

## Unresolved Questions

- Should this be implemented on top of a `FixedSizeBinary` type instead?

CC @gatesn @asubiotto

PR #7266, opened 2026-04-02
feat[vortex-array]: support executing UUID to arrow

asubiotto:
## Summary

Addresses: #6854

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,7 @@ arrow-cast = "58"
 arrow-data = "58"
 arrow-ipc = "58"
 arrow-ord = "58"
-arrow-schema = "58"
+arrow-schema = { version = "58", features = ["canonical_extension_types"] }
 arrow-select = "58"
 arrow-string = "58"
 async-fs = "2.2.0"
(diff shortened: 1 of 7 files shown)


connortsui20:
I think we need to decouple some of this logic. We can't have the arrow executor for an extension type that the core vortex crates don't know about exist there.

However, I think it does make sense to have an `FSL<u8>` executor into arrow FSB, and then the vector extension type can delegate to that.

CC @AdamGS as I remember you had some thoughts on how interop with arrow and our extension types should look like?

joseph-isaacs:
Surely we need to have a registry of exporter kernels?

AdamGS:
Going to read this more carefully over the weekend/next week, I have a similar issue coming up with ParquetVariant

asubiotto:
Happy to implement whatever we think is the best path. I don't have any opinions here.

asubiotto:
(Very) draft PR to add a registry of exporter kernels: https://github.com/vortex-data/vortex/pull/7796. This is mostly claude and I haven't reviewed deeply but want to confirm this is the direction we want to go in.

asubiotto:
This is now available in 0.71.0

PR #7267, opened 2026-04-02
feat: `Count` aggegate

blaginin:
Count from https://github.com/vortex-data/vortex/pull/7201/files

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/count/mod.rs
+++ b/vortex-array/src/aggregate_fn/fns/count/mod.rs
@@ -112,8 +112,12 @@ impl AggregateFnVTable for Count {
 #[cfg(test)]
 mod tests {
     use vortex_buffer::buffer;
-    use vortex_error::{VortexExpect, VortexResult};
+    use vortex_error::VortexExpect;
+    use vortex_error::VortexResult;
 
+    use crate::ArrayRef;
+    use crate::ExecutionCtx;
+    use crate::IntoArray;
     use crate::LEGACY_SESSION;
     use crate::VortexSessionExecute;
     use crate::aggregate_fn::Accumulator;
@@ -129,7 +133,6 @@ mod tests {
     use crate::dtype::PType;
     use crate::scalar::Scalar;
     use crate::validity::Validity;
-    use crate::{ArrayRef, ExecutionCtx, IntoArray};
 
     pub fn count(array: &ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<usize> {
         let mut acc = Accumulator::try_new(Count, EmptyOptions, array.dtype().clone())?;
--- a/vortex-array/src/aggregate_fn/fns/count/mod.rs
+++ b/vortex-array/src/aggregate_fn/fns/count/mod.rs
@@ -1,0 +1,261 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_error::VortexExpect;
+use vortex_error::VortexResult;
+
+use crate::ArrayRef;
+use crate::Columnar;
+use crate::DynArray;
+use crate::ExecutionCtx;
+use crate::aggregate_fn::AggregateFnId;
+use crate::aggregate_fn::AggregateFnVTable;
+use crate::aggregate_fn::EmptyOptions;
+use crate::dtype::DType;
+use crate::dtype::Nullability;
+use crate::dtype::PType;
+use crate::scalar::Scalar;
+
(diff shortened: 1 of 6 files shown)


vortex-array/src/aggregate_fn/fns/count/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    ) -> VortexResult<Self::Partial> {
+        Ok(0u64)
+    }
+
+    fn combine_partials(&self, partial: &mut Self::Partial, other: Scalar) -> VortexResult<()> {
+        let val = other
+            .as_primitive()
+            .typed_value::<u64>()
+            .vortex_expect("count partial should not be null");
+        *partial += val;
```
joseph-isaacs, line 67: what about overflow?

vortex-array/src/aggregate_fn/fns/count/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+/// The identity value is zero.
+#[derive(Clone, Debug)]
+pub struct Count;
+
+impl AggregateFnVTable for Count {
+    type Options = EmptyOptions;
+    type Partial = u64;
+
+    fn id(&self) -> AggregateFnId {
+        AggregateFnId::new_ref("vortex.count")
```
joseph-isaacs, line 31: are you sure (also @gatesn ) we want a global namespace of scalar-fns and aggregations?

vortex-array/src/aggregate_fn/vtable.rs
```diff
@@ -105,6 +105,22 @@ pub trait AggregateFnVTable: 'static + Sized + Clone + Send + Sync {
...
+    /// This is useful for aggregates that only depend on array metadata (e.g., validity)
+    /// rather than the encoded data, avoiding unnecessary decompression.
+    fn try_accumulate(
+        &self,
+        _state: &mut Self::Partial,
+        _batch: &ArrayRef,
+        _ctx: &mut ExecutionCtx,
+    ) -> VortexResult<bool> {
+        Ok(false)
+    }
```
joseph-isaacs, line 122: can you remove this. We want a different system to short cut decompression, similar to execute_parent

vortex-array/src/aggregate_fn/fns/count/mod.rs
blaginin: from `nan_count` i assume we don't really think this would happen in practice

vortex-array/src/aggregate_fn/vtable.rs
blaginin: i think then let's wait for that system to merge - any pr i should follow?

vortex-array/src/aggregate_fn/fns/count/mod.rs
gatesn: I think that's probably fair for a count to not exceed u64

vortex-array/src/aggregate_fn/vtable.rs
gatesn: Isn't this the equivalent of "reduce"? i.e. something that can operate without looking at data buffers.

So maybe we just need to rename this to `fn reduce(...)` and Joe will be happy!

vortex-array/src/aggregate_fn/fns/count/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+use crate::ExecutionCtx;
+use crate::aggregate_fn::AggregateFnId;
+use crate::aggregate_fn::AggregateFnVTable;
+use crate::aggregate_fn::EmptyOptions;
+use crate::dtype::DType;
+use crate::dtype::Nullability;
+use crate::dtype::PType;
+use crate::scalar::Scalar;
+
+/// Count the number of non-null elements in an array.
```
a10y, line 19: why non-null? Shouldn't this just be array.len() ?

a10y: if not can we rename this to CountNonNull? Or add an option? We'll probably want a vanilla Count(*) aggregate to embed in stats

gatesn: Non-null matches SQL semantics

gatesn: I would maybe call the other one RowCount so it's blindingly obvious what it does?

a10y: yea ok, looking at DF that is actually just COUNT over a projected lit(1)

gatesn (approved):

PR #7279, opened 2026-04-03
Implement compare/between for bitpacked arrays - improve performance by up to 2X

AdamGS:
After some iterations, this seems to improve performance in any case we actually measure. It does require merging and releasing https://github.com/spiraldb/fastlanes/pull/125.

### DIFF
--- a/encodings/fastlanes/src/bitpacking/compute/compare.rs
+++ b/encodings/fastlanes/src/bitpacking/compute/compare.rs
@@ -28,6 +28,7 @@ use vortex_error::VortexResult;
 
 use crate::BitPacked;
 use crate::BitPackedArrayExt;
+use crate::BitPackedData;
 
 impl CompareKernel for BitPacked {
     fn compare(
@@ -43,6 +44,7 @@ impl CompareKernel for BitPacked {
         if !constant.dtype().is_int() {
             return Ok(None);
         }
+
         match_each_integer_ptype!(lhs.dtype().as_ptype(), |T| {
             let value = T::try_from(&constant)?;
             compare_constant::<T>(lhs, value, rhs.dtype().nullability(), operator, ctx).map(Some)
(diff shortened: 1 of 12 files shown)


joseph-isaacs:
can you measure compile time and code size delta? of the fastlanes crate?

encodings/fastlanes/src/bitpacking/compute/compare.rs
```diff
@@ -0,0 +1,647 @@
...
+    T: NativePType + FastLanesComparable<Bitpacked = U>,
+    U: UnsignedPType + BitPacking + BitPackingCompare,
+    C: Fn(T, T) -> bool + Copy,
+{
+    let mut bits = collect_chunk_masks::<U>(
+        array.data(),
+        array.len(),
+        array.offset(),
+        |bit_width, packed_chunk, chunk_matches| unsafe {
+            U::unchecked_unpack_cmp(bit_width, packed_chunk, chunk_matches, compare, value);
```
a10y, line 152: is the crux of this that we fuse unpacking with comparison op and avoid the second allocation/scan?

joseph-isaacs: ye, you save a lot on mem bandwidth, esp. for wider types

a10y: TIL that we had fused cmp in fastlanes crate this whole time

robert3005: We lacked a good way to manage fused exection, I think now we have a bit of a better mechanism

AdamGS:
@joseph-isaacs do we have an accepted way of measuring code size?

AdamGS:
Main difference in build time is that `vortex-fastlanes` jumps from ~15 seconds to almost 50 (from 9th slowest crate to 2nd), command was `cargo build -p vortex --release --timings --all-features`.

While this isn't great for our local experience, I think it'll still parallelize fine with other heavy dependencies, and its pretty similar to what users will experience if they pull something like DF (which has multiple crates that takes 50-60 seconds to build)

joseph-isaacs:
Diff size of and bin using this with and without the compare+bitpack

Issue #7282, opened 2026-04-03
`validity()` no longer needs to return a `VortexResult` anywhere


connortsui20:
I am pretty sure that we have no uses of `validity()` returning an error now.

We have this in `vortex-array/src/array/vtable/validity.rs` which always returns `Ok`:

```rust
impl<V: VTable> ValidityVTable<V> for ValidityVTableFromValidityHelper
where
    V::ArrayData: ValidityHelper,
{
    fn validity(array: ArrayView<'_, V>) -> VortexResult<Validity> {
        Ok(array.data().validity().clone())
    }
}
```

connortsui20:
Ok there are actually cases where we do need it (see dict validity) but to me those should just be `expect`ed?

PR #7285, opened 2026-04-03
Fix Array::validity

gatesn:
The vtables refactor ended up obscuring the validity function in favor of ValidityHelper, which was incorrect in a few places.


connortsui20:
@claude review

connortsui20 (approved):

connortsui20:
lets yolo it

I still think that it is weird that validity returns a result (if something has gone wrong with _looking up my validity_ then I feel like that is on me). See https://github.com/vortex-data/vortex/issues/7282

PR #7287, opened 2026-04-04
Hoist Slots

gatesn:
This PR continues the Array VTables refactor by hoisting the common "slots" field out of ArrayData structs and into the generic Array<V> struct.

Part of #6544
Closes #7240


vortex-array/src/arrays/dict/array.rs
```diff
@@ -192,23 +138,71 @@ impl DictData {
...
+            }
+            AllOr::None => {}
+            AllOr::Some(mask) => {
+                match_each_integer_ptype!(codes_primitive.ptype(), |P| {
+                    #[allow(
+                        clippy::cast_possible_truncation,
+                        clippy::cast_sign_loss,
+                        reason = "codes are non-negative indices; a negative signed code would wrap to a large usize and panic on the bounds-checked array index"
+                    )]
+                    for (valid, &idx) in mask.iter().zip(codes_primitive.as_slice::<P>()) {
```
lwwmanning: this used to be as below, I don't yet understand why that changed?

```
                    buf.set_indices().for_each(|idx| {
                        values_vec[codes[idx] as usize] = referenced_value;
                    })
```

vortex-array/src/arrays/dict/array.rs
```diff
@@ -81,23 +75,15 @@ impl DictData {
     /// that all values are referenced.
     pub unsafe fn set_all_values_referenced(mut self, all_values_referenced: bool) -> Self {
         self.all_values_referenced = all_values_referenced;
-
-        #[cfg(debug_assertions)]
```
lwwmanning, line 85: where did this check go?

vortex-array/src/arrays/chunked/array.rs
```diff
@@ -33,186 +37,127 @@ pub(super) const CHUNK_OFFSETS_SLOT: usize = 0;
...
     }
 
-    #[inline]
-    pub fn chunk_offsets(&self) -> Buffer<u64> {
-        self.chunk_offsets_data().to_buffer()
+    fn chunk_offsets(&self) -> Buffer<u64> {
+        let chunk_offsets = self
+            .chunk_offsets_array()
+            .clone()
+            .execute::<Canonical>(&mut LEGACY_SESSION.create_execution_ctx())
```
lwwmanning: this canonicalizes every time now... wait, did we not allow compressing offsets before? that would be reasonable, and honestly we should be careful about compressing these because of e.g., needing to decompress on every access...

lwwmanning: also, what's the deal with LEGACY_SESSION?

vortex-array/src/arrays/listview/array.rs
```diff
@@ -299,106 +279,67 @@ impl ListViewData {
     ///
     /// Note that leading and trailing unreferenced elements **ARE** allowed.
     pub unsafe fn with_zero_copy_to_list(mut self, is_zctl: bool) -> Self {
-        if cfg!(debug_assertions) && is_zctl {
-            validate_zctl(
```
lwwmanning, line 303: where'd this check go?

encodings/fastlanes/src/bitpacking/array/mod.rs
```diff
@@ -403,114 +266,100 @@ impl BitPackedData {
     }
 
     pub fn into_parts(self, len: usize, nullability: Nullability) -> BitPackedDataParts {
```
lwwmanning: do we still want this function, given that we have `BitPacked::into_parts` as well (which actually has the correct patches/validity)? or if we're keeping it, should strip patches & maybe also validity our from `BitPackedDataParts`?

vortex-array/src/arrays/dict/array.rs
lwwmanning: I suspect this is why [bench_many_nulls[0.01]](https://codspeed.io/vortex-data/vortex/branches/ngates%2Fslots?uri=vortex-array%2Fbenches%2Fdict_unreferenced_mask.rs%3A%3Abench_many_nulls%5B0.01%5D&runnerMode=Simulation&utm_source=github&utm_medium=comment-v2&utm_content=benchmark) is so much slower

vortex-array/src/arrays/chunked/array.rs
gatesn: Discussed offline

lwwmanning (approved):

joseph-isaacs:
This likely broke:
 - The shared array
 - The zstd array
 - The Pco array
 - The FSST array

joseph-isaacs:
Ill work on the fix

PR #7294, opened 2026-04-06
Allow Claude to write

gatesn:
* Split Claude automation into two workflows with separate trust boundaries: claude-write.yml for write-capable issue automation and follow-up edits on Claude-owned PRs, and claude-review.yml for read-only PR review.
* Sets up a GitHub app (vortex-claude) so PRs pushed by Claude can run CI
* Only claude-write can access the GitHub App private key, and only through the claude-automation environment. No other workflow gets that key.
* claude-write uses a short-lived GitHub App installation token, not GITHUB_TOKEN, so PRs opened by Claude trigger normal pull_request CI.
* claude-write runs only for trusted collaborators with write/maintain/admin access.
* claude-write is issue-driven by default. It may also run from a PR conversation comment, but only when that PR is a same-repo PR originally opened by the Vortex Claude GitHub App (CLAUDE_APP_LOGIN).
* claude-write refuses fork PR content entirely. There is no promotion path for forks.
* claude-review has no GitHub App private key and no write-capable repo token. It uses the built-in GITHUB_TOKEN only for reading repo state and posting review comments.
* claude-review can review same-repo PRs, but it must not create commits, push branches, open PRs, or update PR branches.
* PR conversation comments on Claude-owned PRs are routed to claude-write; claude-review explicitly stands down on that path to avoid duplicate runs.
* Both workflows reject bot-triggered invocations and gate access before running Claude logic.
* Both workflows use persist-credentials: false on checkout so git operations do not silently fall back to the built-in GITHUB_TOKEN.
* Fork PRs are always refused for Claude automation because fork content is treated as untrusted prompt input.


connortsui20:
@claude review

connortsui20:
I think that we have tools in place now to do this (2 humans in the loop always)? I'll close for now.

robert3005 (dismissed):

robert3005 (approved):

PR #7296, opened 2026-04-06
Bring back file-size comments

a10y:
The core issue that caused benchmarks to fail last time was that the existing `cat-s3.sh` script didn't handle uploading files that do not exist.

The fix is to add handling that will attempt to perform a first PUT if the head-object returns 404/403


connortsui20 (approved):
yolo

PR #7298, opened 2026-04-06
add LazyPatchedArray

a10y:
First step as part of replacing arrays with interior patches with wrapped PatchedArray in backward-compatible fashion.

We create a new wrapper array, `LazyPatchedArray`. This only lives in-memory as an execution shim to bridge old-style `Patches` to new-style `PatchedArray`.

The follow up to this PR will update the `VTable::build()` method to return ArrayRef, and update BitPacked::build() to return either BitPackedArray with interior Patches, or PatchedArray wrapping BitPacked, based on an experimental flag.


gatesn (changes requested):
Don't we want to just replace e.g. fastlanes.bitpacked with a new deserializer that returns Patched(BitPacked) ?

a10y:
what is a deserializer?

PR #7302, opened 2026-04-06
ArrayPlugin

gatesn:
Rename / clean up to match other VTable designs and make it easier to implement custom Array deserializers.


vortex-array/src/array/plugin.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn deserialize(
+        &self,
+        dtype: &DType,
+        len: usize,
+        metadata: &[u8],
+        buffers: &[BufferHandle],
+        children: &dyn ArrayChildren,
+        session: &VortexSession,
+    ) -> VortexResult<ArrayRef>;
+}
```
a10y, line 43: this is exactly what i need

a10y (approved):

PR #7305, opened 2026-04-07
Fix StructArray / StructData after migration

gatesn:
Fixes a decompression regression on wide struct datasets by removing StructData metadata duplication and deriving struct metadata from the outer cached dtype/len.


vortex-array/src/arrays/struct_/vtable/mod.rs
```diff
@@ -73,14 +74,6 @@ impl VTable for Struct {
             vortex_bail!("Expected struct dtype, found {:?}", dtype)
         };
 
-        if data.names() != struct_dtype.names() {
```
gatesn, line 76: Note that these only existed because StructData duplicated the information, so we had to check the duplicates matched!

robert3005 (approved):

PR #7310, opened 2026-04-07
Do we need to explicitly install ninja?

robert3005:
This is likely preinstalled


joseph-isaacs:
You can also check in the prebuild

joseph-isaacs (approved):

PR #7311, opened 2026-04-07
Migrate to maintained version of setup-zig

robert3005:
Zig has moved to codeberg, there were no updates to the action yet since the
migration


robert3005:
This is a lie, the reference still works

PR #7312, opened 2026-04-07
fix[vortex-array]: fix offset_within_chunk underflow on patches array

asubiotto:
slice_start_idx was missing normalization to absolute coordinates.

### DIFF
--- a/vortex-array/src/patches.rs
+++ b/vortex-array/src/patches.rs
@@ -656,7 +656,7 @@ impl Patches {
         let values = self.values().slice(slice_start_idx..slice_end_idx)?;
         let indices = self.indices().slice(slice_start_idx..slice_end_idx)?;
 
-        let chunk_offsets = self
+        let new_chunk_offsets = self
             .chunk_offsets
             .as_ref()
             .map(|chunk_offsets| -> VortexResult<ArrayRef> {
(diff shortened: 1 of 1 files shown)


vortex-array/src/patches.rs
```diff
@@ -670,12 +670,15 @@ impl Patches {
...
+                let absolute_chunk_base = chunk_offsets
                     .scalar_at(0)?
                     .as_primitive()
                     .as_::<usize>()
                     .ok_or_else(|| vortex_err!("chunk offset does not fit in usize"))?;
-                Ok(slice_start_idx - base_offset)
+                let absolute_slice_start = self.chunk_offset_at(0)?
+                    + self.offset_within_chunk.unwrap_or(0)
+                    + slice_start_idx;
+                Ok(absolute_slice_start - absolute_chunk_base)
```
robert3005: I might be naive but doesn't this reduce to `self.offset_within_chunk.unwrap_or(0) + slice_start_idx` since  `self.chunk_offset_at(0)` is the same as the logic for `absolute_chunk_base`

asubiotto: I don't think so, because `absolute_chunk_base` is the chunk offset of the sliced chunk (not parent chunk). I have updated the PR with better names which hopefully makes this clearer.

robert3005 (approved):
Thanks for explaining it to me

PR #7314, opened 2026-04-07
add env flag for Patched array

a10y:
## Summary

Part of roll out of Patched array. This PR adds a new environment variable, `VORTEX_EXPERIMENTAL_PATCHED_ARRAY`.

When the env var is set

1. On read, any flat layout containing a `BitPacked` array with interior patches will be decoded as a `Patched` array wrapping a `BitPacked` with no interior patches
2. The default BtrBlocks compressor will write a Patched array of BitPacked instead of BitPacked arrary with interior patches.

This allows us to do some more soak testing of the encoding before we switch it to be the default.

## Testing

On a previous commit in this PR, I've setup the benchmarks to run with the experimental Patched encoding enabled, the benchmark results all seem within the margin of error. The file sizes are generally the same but in the worst case, 2-3% larger.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -109,6 +109,7 @@ jobs:
       fuzz_target: array_ops
       fuzz_name: array_ops_unstable_encodings
       extra_features: "vortex/unstable_encodings"
+      extra_env: "VORTEX_EXPERIMENTAL_PATCHED_ARRAY=1"
       jobs: 16
     secrets:
       R2_FUZZ_ACCESS_KEY_ID: ${{ secrets.R2_FUZZ_ACCESS_KEY_ID }}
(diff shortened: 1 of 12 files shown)


.github/workflows/fuzz.yml
```diff
@@ -109,6 +109,7 @@ jobs:
       fuzz_target: array_ops
       fuzz_name: array_ops_unstable_encodings
       extra_features: "vortex/unstable_encodings"
+      extra_env: "VORTEX_EXPERIMENTAL_PATCHED_ARRAY=1"
```
joseph-isaacs, line 112: seems odd we need both?

robert3005 (approved):

.github/workflows/fuzz.yml
robert3005: I think we should convert cargo feature flags to environment variables/config. It's not useful if you're using a non rust package and have to recompile the world to try out a feature

a10y: Yea, I think this is something we want as a runtime/env flag not a compile flag

PR #7325, opened 2026-04-07
Unify id creation for ScalarFns and rename is_null to vortex.is_null

robert3005:
Use ScalarFnId::from instead of new_ref and rename is_null fn to vortex.is_null


vortex-array/src/scalar_fn/fns/dynamic.rs
joseph-isaacs, line 1: should be namespace these into vortex.scalar_fn... or something else?

Also while we are here what dynamic expr is this?

robert3005: haha, `DynamicComparisonExpr`

gatesn (approved):
If folks have serialized expressions, they may need to register a custom ScalarFnPlugin to deserialize the old is_null ID

PR #7327, opened 2026-04-07
Field pruning in nested struct scan

brancz:
## Summary

I'm trying to have scans read only a subset of a nested struct. It's of the shape: `List<Struct<a: Utf8, b: Int64>`, so I can't use regular projections in DataFusion, since those only support indices. This is my best attempt, combined with an optimizer rule that modifies the table schema to the subset under certain circumstances.

## Testing

Added a unit test that fails without the second commit.

## Notes

Happy to think through other approaches to this, this was just the first thing that I came up with.

### DIFF
--- a/vortex-array/src/arrow/executor/struct_.rs
+++ b/vortex-array/src/arrow/executor/struct_.rs
@@ -320,6 +320,53 @@ mod tests {
         Ok(())
     }
 
+    /// Test that converting a struct to Arrow with fewer target fields (field
+    /// pruning) works by falling through to the cast path.
+    #[test]
+    fn struct_to_arrow_with_field_pruning() -> VortexResult<()> {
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrow/executor/struct_.rs
```diff
@@ -47,23 +47,41 @@ pub(super) fn to_arrow_struct(
...
+                _ => 0,
+            };
+            // Skip the fast path only when target has strictly fewer fields
+            // (nested field pruning). When target has same or more fields,
+            // use the fast path which will validate the count.
+            let can_fast_path = match target_fields {
+                None => true,
+                Some(fields) => fields.len() >= n_struct_fields,
+            };
+            if can_fast_path {
```
connortsui20, line 67: you can use `is_none_or` here

PR #7337, opened 2026-04-08
Buffer Allocator Abstraction

gatesn:
Initial PR to move towards a session-defined buffer allocation.

1. Sets up a MemorySession (we could move this and all buffer handle types to vortex-io and treat the i/o crate as engine integration?)
2. Defines an allocator trait, returning a WritableHostBuffer
3. Freeze the WriteableHostBuffer into a host ByteBuffer

For now, we keep the existing behavior for device buffers (i.e. vortex-cuda constructs its own device BufferHandle inside its segment source).

This change configures host VortexReadAt sources (ObjectStoreReadAt and FileReadAt) to use the allocator from the session. It also demonstrates using the allocator from within an array (chunked canonical), although it doesn't do the plumbing work all the way into the array builders.

A follow up PR will implement a pooling allocator.

Further PRs wlll port over the execution logic to use the allocator from the ExecutionCtx, eventually using clippy to forbid direct ByteBufferMut access from within our (non allocator) crates.


vortex-array/src/memory.rs
```diff
@@ -0,0 +1,375 @@
...
+            .finish()
+    }
+}
+
+/// Allocator for exact-size writable host buffers.
+pub trait BufferAllocator: Debug + Send + Sync + 'static {
+    /// Allocate a writable host buffer with the requested byte length and alignment.
+    fn allocate(&self, len: usize, alignment: Alignment) -> VortexResult<WritableHostBuffer>;
+}
+
```
joseph-isaacs: can we call this a host allocator?

0ax1:
Can this get a little RFC upfront?

joseph-isaacs:
Why does an allocator return `WritableHostBuffer` vs a buffer handle?

gatesn:
> Why does an allocator return `WritableHostBuffer` vs a buffer handle?

BufferHandle isn't writeable

gatesn:
@0ax1 https://github.com/vortex-data/rfcs/pull/44/changes

joseph-isaacs (approved):

PR #7339, opened 2026-04-08
fix: avoid ListView take_reduce rebuild for dense selections

dimitarvdimitrov:
The `take_reduce` path rebuilds the ListView, which is expensive. 

For dense selections (density above the rebuild threshold) there's little garbage to reclaim, so skip the rebuild and let the regular take path handle it. The rebuild is still worthwhile for sparse takes where we'd otherwise drag around a lot of unused elements when exporting (e.g. to DuckDB).

This PR does a very crude estimation of the number of elements that stay in the ListViewArray.

Potentially we should do this estimation at export time in the `ListViewExporter`, but that's less trivial.


vortex-array/src/arrays/listview/compute/take.rs
joseph-isaacs: remove this too?

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -26,8 +25,7 @@ use crate::scalar::Scalar;
 /// However, we also do not want to carry around a large amount of garbage data. Below this
 /// threshold of the density of the selection mask, we will rebuild the [`ListViewArray`], removing
 /// any garbage data.
-#[allow(unused)]
-const REBUILD_DENSITY_THRESHOLD: f64 = 0.1;
+const REBUILD_DENSITY_THRESHOLD: f32 = 0.1;
```
joseph-isaacs: can we have only one of these

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -41,48 +41,83 @@ const REBUILD_DENSITY_THRESHOLD: f64 = 0.1;
...
+        // compute functions have run, at the "top" of the operator tree. However, we cannot do
+        // this right now, so we will just rebuild every time (similar to `ListArray`).
+        let taken = take_metadata(array, indices)?;
         Ok(Some(
-            new_array
+            taken
                 .rebuild(ListViewRebuildMode::MakeZeroCopyToList)?
                 .into_array(),
         ))
     }
```
joseph-isaacs: why do you always rebuild? This will have the same problem as before

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -41,48 +41,83 @@ const REBUILD_DENSITY_THRESHOLD: f64 = 0.1;
...
 }
+
+/// Shared metadata-only take: take `offsets`, `sizes` and `validity` at `indices` while reusing
+/// the original `elements` buffer as-is.
+fn take_metadata(
```
joseph-isaacs:
```suggestion
fn apply_take(
```

vortex-array/src/arrays/listview/vtable/mod.rs
```diff
@@ -214,4 +215,13 @@ impl VTable for ListView {
...
+
+    fn execute_parent(
+        array: ArrayView<'_, Self>,
+        parent: &ArrayRef,
+        child_idx: usize,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<ArrayRef>> {
+        PARENT_KERNELS.execute(array, parent, child_idx, ctx)
+    }
 }
```
joseph-isaacs: you don't need this

gatesn:
Am I missing something? Why would we _ever_ rebuild? The whole point of ListView is to have cheap take/filter.

It should be explicit from the user if they want to garbage collect

dimitarvdimitrov:
> Am I missing something? Why would we _ever_ rebuild? The whole point of ListView is to have cheap take/filter.
> 
> It should be explicit from the user if they want to garbage collect

The rebuild is still worthwhile for sparse takes where we'd otherwise drag around a lot of unused elements when exporting e.g. duckdb's ListViewExporter https://github.com/vortex-data/vortex/blob/11d607e123d25186f8b16f1bead1a4835930aeff/vortex-duckdb/src/exporter/list_view.rs#L80-L85

ideally we do this as heuristics in the exporter as opposed to in each operation like take. I was thinking of doing that next, but it might take a bit longer

gatesn:
> ideally we do this as heuristics in the exporter as opposed to in each operation like take

Yeah, 100%.  Agree priority might not be to do it now, but eager GC impacts all Vortex users

dimitarvdimitrov:
looking into why the benchmarks regressed...

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -17,72 +19,99 @@ use crate::dtype::Nullability;
...
-
+        // compute functions have run, at the "top" of the operator tree. However, we cannot do
+        // this right now, so we will just rebuild every time (similar to `ListArray`).
+        let taken = apply_take(array, indices)?;
         Ok(Some(
-            new_array
+            taken
                 .rebuild(ListViewRebuildMode::MakeZeroCopyToList)?
                 .into_array(),
         ))
```
joseph-isaacs: NOOO, don't always rebuild!

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -17,72 +19,91 @@ use crate::dtype::Nullability;
...
 ///
-/// However, we also do not want to carry around a large amount of garbage data. Below this
-/// threshold of the density of the selection mask, we will rebuild the [`ListViewArray`], removing
-/// any garbage data.
-#[allow(unused)]
-const REBUILD_DENSITY_THRESHOLD: f64 = 0.1;
+// TODO(connor)[ListView]: Ideally, we would only rebuild after all `take`s and `filter`
+//  compute functions have run, at the "top" of the operator tree. However, we cannot do this
+//  right now, so we will just rebuild every time (similar to [`ListArray`]).
+pub const REBUILD_DENSITY_THRESHOLD: f32 = 0.1;
```
joseph-isaacs: i think this should move? and also be pub crate?

dimitarvdimitrov: any suggestion for where to move it? it's now used in arrays/listview/compute and arrays/filter/execute

vortex-array/src/arrays/listview/compute/take.rs
```diff
@@ -17,72 +19,91 @@ use crate::dtype::Nullability;
...
 ///
-/// However, we also do not want to carry around a large amount of garbage data. Below this
-/// threshold of the density of the selection mask, we will rebuild the [`ListViewArray`], removing
-/// any garbage data.
-#[expect(unused)]
-const REBUILD_DENSITY_THRESHOLD: f64 = 0.1;
+// TODO(connor)[ListView]: Ideally, we would only rebuild after all `take`s and `filter`
+//  compute functions have run, at the "top" of the operator tree. However, we cannot do this
+//  right now, so we will just rebuild every time (similar to [`ListArray`]).
+pub(crate) const REBUILD_DENSITY_THRESHOLD: f32 = 0.1;
```
joseph-isaacs: why does this live here and not in the compute mod still as pub(crate)?

joseph-isaacs (approved):
Fix this and this is ready to go

PR #7347, opened 2026-04-08
Allow loading "foreign" plugins for UI/TUI/Serde

gatesn:
This allows us to enable deserialization and re-serialization of unknown plugins by preserving their serialized metadata.

We cannot necessarily execute them in-memory, but it should be sufficient to traverse the expression trees. For arrays/layouts, we cannot construct the child trees without the vtable logic that derives the child dtypes, therefore a foreign encoding becomes a leaf node.


robert3005 (approved):
my only hesitation are foreign layouts but you can still traverse what you can from the layout

PR #7348, opened 2026-04-08
Move Array::serialize -> ArrayPlugin::Serialize

gatesn:
The cheeky thing here is that now we require a VortexSession to serialize metadata:
1. Makes it hard to serialize arrays you didn't intend to. All arrays must be registered at least.
2. Allows ScalarFnArray to serialize metadata using a dedicated ArrayPlugin for that ScalarFn. So not all scalar function vtables need to define array serde semantics, they just register an optional plugin. This will allow us to port existing scalar arrays to scalar functions without a wire break


connortsui20:
@claude review

connortsui20 (approved):
ship it (after lint fix and claude review)

connortsui20:
@claude review

Issue #7216, opened 2026-03-31
Tracking Issue: Pluggable Compressor


connortsui20:
# Pluggable Compressor

This is a tracking issue for extracting a pluggable compression framework (`vortex-compressor`) from `vortex-btrblocks`, as well as making general improvements.

Related: https://github.com/vortex-data/vortex/issues/6872

## Motivation

The `vortex-btrblocks` compressor currently depends on every encoding crate in the workspace, and extension types (Vector, UUID, Tensor, JSON) have no mechanism for type-specific compression. We need to invert the dependency graph so that encoding crates can implement compression schemes without circular dependencies.

_Theoretically, there was a way to hack pluggability into the existing compressor without a complete rewrite, but it would not provide the level of expressiveness needed to fully support extension types and encodings as a first-class citizen._

## Design

<details>

- Unified `Scheme` trait replaces the old type-specific `IntegerScheme`/`FloatScheme`/`StringScheme` traits and `IntCode`/`FloatCode`/`StringCode` enums. Schemes are identified by opaque `SchemeId` (obtained only via `SchemeExt::id()`). The old `Compressor`/`CompressorExt`/`CanonicalCompressor` traits and `IntCompressor`/`FloatCompressor`/`StringCompressor` structs are replaced by a `CascadingCompressor` that selects from a vec of `&'static dyn Scheme`.
- The new `vortex-compressor` crate contains the framework (trait definitions, cascading compressor, stats, sampling) with zero encoding dependencies (other than built-in ones from `vortex-array`).
- `ArrayAndStats` bundle replaces the old pattern of passing arrays and stats caches separately. Stats are generated lazily on first access via typed methods (`integer_stats()`, `float_stats()`, `string_stats()`). Each scheme declares any expensive required stats via `stats_options()` (specifically, distinct values and their frequencies via a hash map), and the compressor merges all eligible schemes' options before generating stats so that expensive computations only run when needed.
- `vortex-btrblocks` remains the batteries-included writer, except now it depends on `vortex-compressor` for the compression logic and registers all encoding-specific schemes (BitPacking, FoR, ALP, FSST, etc.) that we host in `encodings/`.
[...]

PR #7350, opened 2026-04-08
Remove generic rle

connortsui20:
## Summary

We have this generic RLE code from the old compressor that doesn't make as much sense anymore.
This change hardcodes the 2 schemes (int and float) instead of using generics that make things harder to trace.

This is a purely cosmetic change.

## API Changes

Removes olds constants and just has `FloatRLEScheme` and `IntRLEScheme`.


connortsui20:
I simply dont believe this given that there is no change in the files, and this could ONLY affect the compressor: https://github.com/vortex-data/vortex/pull/7350#issuecomment-4208985914

a10y (approved):

a10y:
Let's watch the develop benchmark runs for regressions to suss out if this is real

robert3005 (approved):

PR #7354, opened 2026-04-09
Fix patch_chunk index OOB when slicing ALP arrays mid-chunk

abnobdoss:
## Summary

`patch_chunk` panics with index-out-of-bounds when decompressing sliced ALP arrays with patches. The slice boundary must fall mid-chunk (non-1024-aligned) to trigger it.

**Root cause:** `Patches::slice()` slices `chunk_offsets` at chunk granularity but `indices`/`values` at element granularity. When a slice ends mid-chunk, `patches_end_idx` can exceed `patches_indices.len()`. The start index already used `saturating_sub` — the end index didn't.

**Fix:** `.saturating_sub(offset_within_chunk).min(patches_indices.len())` on the end index computation in `patch_chunk`, `search_index_chunked`, and `search_index_chunked_batch`.

**AI disclosure:** Root cause analysis, fix implementation, and test writing were done with Claude Code under my direction and review. I identified the bug during development benchmarking at 100M rows and verified the fix against my workload.

## Testing

- Unit test calling `patch_chunk` with inputs that OOB without the fix.
- Integration test: ALP encode 3-chunk array with patches, slice mid-chunk, decode. Tests both start-aligned and mid-chunk-start variants.
- All existing tests pass (2538 vortex-array, 167 vortex-alp).

### DIFF
--- a/encodings/alp/src/alp/compress.rs
+++ b/encodings/alp/src/alp/compress.rs
@@ -546,4 +546,44 @@ mod tests {
 
         assert_arrays_eq!(decoded, array);
     }
+
+    /// Regression test for patch_chunk index-out-of-bounds when slicing a multi-chunk
+    /// ALP array mid-chunk with patches in the trailing chunk.
+    ///
+    /// The bug: chunk_offsets are sliced at chunk granularity (1024-row boundaries)
+    /// but patches indices/values are sliced at element granularity. When a slice ends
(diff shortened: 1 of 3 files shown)


robert3005 (approved):
offset_within_chunk has to always be used with saturating sub, we don't know where exactly the indicies fall in the chunk. I thought that this wasn't necessary for subsequent chunks after the current chunk but we don't know if chunks have any patches at all

Issue #7355, opened 2026-04-09
AVX2 primitive take treats valid u8 index 255 as invalid when values.len() == 256


st1page:
# Title

AVX2 primitive take treats valid `u8` index `255` as invalid when `values.len() == 256`

# Body

## Summary

I found a correctness bug in the AVX2 primitive take path that affects dictionary decoding during canonical execution.

When:

- the dictionary code type is `u8`
- the dictionary has exactly `256` values
- a valid code is `255`
- the AVX2 take path is used

the SIMD path treats index `255` as invalid and substitutes `0` instead of the real value.

This surfaced for us as corruption during canonical execution of a compact-encoded nullable float column:

- source logical type: nullable `f64`
- encoded as `ALP(f64?)`
- inner encoded child: `Dict(i64?)`
- `values_len = 256`
- `codes_ptype = U8`
- bad segment: valid value `5078555.0` became `0.0`

## Why this seems to happen

In `vortex-array/src/arrays/primitive/compute/take/avx2.rs`, the AVX2 path computes:

```rust
let max_index = Idx::from(values.len()).unwrap_or_else(|| Idx::max_value());
```

For `Idx = u8` and `values.len() == 256`, `Idx::from(256)` overflows and falls back to `255`.

Then the gather path uses a strict comparison:

```rust
// Create a mask for valid indices (where the max_idx > provided index).
let invalid_mask = unsafe { $mask_indices(max_idx_vec, indices_vec) };
```

So index `255` is treated as invalid because the effective predicate becomes `255 > 255`, which is false.

The masked gather then substitutes zero.

## Minimal reproduction

The important part is that short inputs may fall through to the scalar remainder path and look correct. You need enough indices to exercise the AVX2 gather loop.

Pseudo-repro:

1. Build a primitive values array with length `256`
2. Use `u8` indices
3. Fill the indices array with repeated `255`
4. Make the indices array long enough to enter the SIMD loop

Observed behavior on x86_64 with AVX2:

- short take like `[255]` returns the correct value
- wider take like `[255; 20]` returns zeros in the SIMD-produced slots and correct values in the scalar remainder slots

In our case, this reproduces through dictionary execution:

- `codes.execute()` still returns `255` at the bad positions
- `Dict::execute()` returns `0` at those same positions
- `ALP::execute()` then decodes those zeros into `0.0`

## Concrete evidence from local repro

- Original compressed array `scalar_at()` is correct
- serialize/decode roundtrip is correct
[...]

PR #7356, opened 2026-04-09
fix: avoid AVX2 take misclassifying u8 max index

st1page:
## Summary
- fall back to the scalar take path when values.len() cannot be represented by the unsigned index type
- add a regression test for u8 indices taking the last valid code from a 256-value array

## Why
This fixes the u8 / values.len() == 256 boundary bug reported in #7355, where the AVX2 path can treat valid index 255 as out-of-bounds and substitute zero.

## Testing
- cargo test -p vortex-array test_avx2_take_u8_last_valid_index_for_256_values
- cargo test -p vortex-array avx2_take


robert3005:
made an alternative in #7359

PR #7359, opened 2026-04-09
AVX2 take handles indices that are equal to the index type max value

robert3005:
The in the avx2 take code correctly states that we need to fitler out values
greater than but then it also filters out values that are equal, this fixes the
edge case for values that are equal to the max index value

### DIFF
--- a/vortex-array/src/arrays/primitive/compute/take/avx2.rs
+++ b/vortex-array/src/arrays/primitive/compute/take/avx2.rs
@@ -14,6 +14,7 @@ use std::arch::x86_64::_mm_setzero_si128;
 use std::arch::x86_64::_mm_shuffle_epi32;
 use std::arch::x86_64::_mm_storeu_si128;
 use std::arch::x86_64::_mm_unpacklo_epi64;
+use std::arch::x86_64::_mm256_andnot_si256;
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/primitive/compute/take/avx2.rs
```diff
@@ -223,7 +222,7 @@ macro_rules! impl_gather {
                     // Create a vec of the max idx.
                     let max_idx_vec = unsafe { $splat(max_idx as _) };
                     // Create a mask for valid indices (where the max_idx > provided index).
-                    let invalid_mask = unsafe { $mask_indices(max_idx_vec, indices_vec) };
+                    let invalid_mask = unsafe { _mm256_or_si256($mask_indices(max_idx_vec, indices_vec), $mask_equal(max_idx_vec, indices_vec)) };
```
joseph-isaacs: do we really need two more instrs?

robert3005: alternatively we can make a nand

robert3005: Changed it so we have 1 extra instruction. Could also generify this further and split the edge case to separate kernel

joseph-isaacs (approved):

PR #7360, opened 2026-04-09
feat: add an arrays slots macro

joseph-isaacs:
Adds a array_slot(VTable) attr macro which can be applied to a slots struct. This gives types access to the slots for an array.

e.g.

```rust
#[array_slots(ALP)]
pub struct ALPSlots {
    /// The ALP-encoded values array.
    pub encoded: ArrayRef,
    /// The indices of exception values that could not be ALP-encoded.
    pub patch_indices: Option<ArrayRef>,
    /// The exception values that could not be ALP-encoded.
    pub patch_values: Option<ArrayRef>,
    /// Chunk offsets for the patch indices/values.
    pub patch_chunk_offsets: Option<ArrayRef>,
}
```


robert3005:
@connortsui20

connortsui20:
just at a quick glance, I think it is super important that we have documentation that explains what gets generated, instead of having to read to proc macro code

robert3005:
Fair, I had to review with an ide to see macro expansions

joseph-isaacs:
I am happy to add just wanted to see what is was and if this was the right approach?

connortsui20 (approved):

PR #7361, opened 2026-04-09
ScalarFns as Arrays

gatesn:
Provide a utility for serializing scalar functions as arrays


joseph-isaacs:
Nice

vortex-array/src/arrays/scalar_fn/vtable/mod.rs
```diff
@@ -51,7 +50,7 @@ pub type ScalarFnArray = Array<ScalarFnVTable>;
 
 #[derive(Clone, Debug)]
 pub struct ScalarFnVTable {
-    pub(super) scalar_fn: ScalarFnRef,
+    pub(super) id: ScalarFnId,
```
joseph-isaacs, line 53: why did you change this?

vortex-array/src/arrays/scalar_fn/plugin.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    fn serialize(
+        &self,
+        array: &ArrayRef,
+        session: &VortexSession,
+    ) -> VortexResult<Option<Vec<u8>>> {
+        // We serialize the scalar function options, along with any scalar function array data.
+        let scalar_fn = array.as_::<ExactScalarFn<V>>();
+        <V as ScalarFnArrayVTable>::serialize(&self.0, &scalar_fn, session)
+    }
```
joseph-isaacs, line 71: can we put this behind an env var?

gatesn: No.... why? This is still opt-in, you have to register the plugin

vortex-array/src/arrays/scalar_fn/vtable/mod.rs
gatesn: We were storing more than we needed, and had to check for equality later anyway. So reduced it to minimal information

joseph-isaacs (approved):

connortsui20 (approved):
send it

PR #7369, opened 2026-04-09
AllInvalid arrays have non empty distinct info

robert3005:
This is required to handle cases where we are attempting to dict compress and
a sample ends up being all null

### DIFF
--- a/vortex-compressor/src/stats/float.rs
+++ b/vortex-compressor/src/stats/float.rs
@@ -179,7 +179,13 @@ where
             null_count: u32::try_from(array.len())?,
             value_count: 0,
             average_run_length: 0,
-            erased: TypedStats { distinct: None }.into(),
+            erased: TypedStats {
+                distinct: Some(DistinctInfo {
+                    distinct_values: HashSet::with_capacity_and_hasher(0, FxBuildHasher),
+                    distinct_count: 0,
+                }),
+            }
+            .into(),
         });
     }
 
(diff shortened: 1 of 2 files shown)


vortex-compressor/src/stats/float.rs
```diff
@@ -179,7 +179,13 @@ where
             null_count: u32::try_from(array.len())?,
             value_count: 0,
             average_run_length: 0,
-            erased: TypedStats { distinct: None }.into(),
+            erased: TypedStats {
+                distinct: Some(DistinctInfo {
+                    distinct_values: HashSet::with_capacity_and_hasher(0, FxBuildHasher),
+                    distinct_count: 0,
```
connortsui20, line 185: distinct count has to be 0, this is an invariant that other code relies on.

Let me check where

connortsui20 (dismissed):

vortex-compressor/src/stats/float.rs
connortsui20: we use it in the constant schemes where if the distinct count is not greater than 1, it must be exactly 1.

I think it is actually fine though since that exact code path wont happen, but I would prefer if we just set distinct count to 1 and then add a big TODO and red flag that this is a hack. Once I finish the vector search stuff I can revisit this and come up with a more robust solution while adding all the things that would be good to have in the compressor (https://github.com/vortex-data/vortex/issues/7216)

robert3005: not sure why setting any value is meaningful here, the idea is that they will not look at it and we just need to have a present value. This is also internally consistent where all null arrays aren't constant

connortsui20: its just that I rely on that fact at other callsites, so it might panic there instead. I guess it doesn't really matter either way since both of these fixes are technically incorrect

connortsui20 (approved):
i will fix this next week

PR #7370, opened 2026-04-09
ALP decoding plugin

a10y:
Follow up to #7314 this time for ALP


encodings/alp/src/lib.rs
```diff
@@ -22,14 +22,21 @@ use vortex_array::aggregate_fn::AggregateFnVTable;
...
 
 mod alp;
 mod alp_rd;
 
 /// Initialize ALP encoding in the given session.
 pub fn initialize(session: &VortexSession) {
-    session.arrays().register(ALP);
+    // If we're using the experimental Patched encoding, register a shim
+    // for ALP with interior patches to decode as Patched array.
+    if *USE_EXPERIMENTAL_PATCHES {
```
connortsui20, line 35: This isn't part of this PR but is there a reason we put this constant in vortex-fastlanes and not in vortex-array?

a10y: At the time it was really only relevant within vortex-fastlanes crate but now I think we can probably move it

connortsui20 (approved):

joseph-isaacs (approved):

PR #7371, opened 2026-04-09
perf: zero copy validity export to duckdb

joseph-isaacs:
Zero copy export validity similarly to how we export data for Primitive or Decimal.


vortex-duckdb/cpp/include/duckdb_vx/vector.h
```diff
@@ -50,6 +50,11 @@ void duckdb_vx_vector_set_vector_data_buffer(duckdb_vector ffi_vector, duckdb_vx
 // Set the data pointer for the vector. This is the start of the values array in the vector.
 void duckdb_vx_vector_set_data_ptr(duckdb_vector ffi_vector, void *ptr);
 
+// Set the validity pointer for the vector to external data, and store the buffer in auxiliary
+// to keep it alive. This enables zero-copy export of validity masks.
+void duckdb_vx_vector_set_validity_data(duckdb_vector ffi_vector, void *validity_ptr, idx_t capacity,
```
myrrc, line 55: If validity_ptr points to buffer, just pass the buffer

vortex-duckdb/cpp/vector.cpp
```diff
@@ -56,6 +56,21 @@ class DataVector : public Vector {
...
+
+    inline ValidityMask &GetValidity() {
+        return validity;
+    };
+};
+
+// Same hack for ValidityMask: access protected fields via inheritance.
+class ExternalValidityMask : public ValidityMask {
+public:
+    inline void SetExternal(validity_t *ptr, idx_t cap,
```
myrrc, line 68: Same here, pass just the buffer and derive ptr from it

vortex-duckdb/cpp/vector.cpp
```diff
@@ -81,6 +96,26 @@ extern "C" void duckdb_vx_vector_set_data_ptr(duckdb_vector ffi_vector, void *pt
...
+
+    // Use the shared_ptr aliasing constructor: the control block ref-counts the
+    // ExternalVectorBuffer (preventing the Rust buffer from being freed),
+    // while the stored pointer satisfies ValidityMask's buffer_ptr<ValidityBuffer> type.
+    auto ext_buf = reinterpret_cast<shared_ptr<vortex::ExternalVectorBuffer> *>(buffer);
+    auto keeper = shared_ptr<TemplatedValidityData<validity_t>>(
+        *ext_buf, reinterpret_cast<TemplatedValidityData<validity_t> *>(ext_buf->get()));
+
+    // Set validity_mask, capacity, and validity_data (which keeps the buffer alive).
+    ext_validity->SetExternal(reinterpret_cast<validity_t *>(validity_ptr), capacity,
```
myrrc, line 115: Technically this will slice the class to base's validity, but as derived class doesn't have any members, it's fine. Worth adding a comment

vortex-duckdb/cpp/vector.cpp
```diff
@@ -56,6 +56,21 @@ class DataVector : public Vector {
...
+};
+
+// Same hack for ValidityMask: access protected fields via inheritance.
+class ExternalValidityMask : public ValidityMask {
+public:
+    inline void SetExternal(validity_t *ptr, idx_t cap,
+                            buffer_ptr<ValidityBuffer> keeper) {
+        validity_mask = ptr;
+        capacity = cap;
+        validity_data = std::move(keeper);
```
myrrc, line 72: I don't think this is right as TemplatedValidityMask(V* ptr, idx_t capacity) constructor sets the pointer but doesn't change validity_data, so these two aren't derived.

joseph-isaacs: This like better?

vortex-duckdb/cpp/vector.cpp
```diff
@@ -82,6 +97,29 @@ extern "C" void duckdb_vx_vector_set_data_ptr(duckdb_vector ffi_vector, void *pt
...
 
+extern "C" void duckdb_vx_vector_set_validity_data(duckdb_vector ffi_vector,
+                                                   idx_t u64_offset,
+                                                   idx_t capacity,
+                                                   duckdb_vx_vector_buffer buffer) {
+    auto dvector = reinterpret_cast<vortex::DataVector *>(ffi_vector);
+    auto &validity = dvector->GetValidity();
+    // ExternalValidityMask adds no members, so this reinterpret_cast is a safe
+    // "downcast" that only exposes access to ValidityMask's protected fields.
+    auto ext_validity = reinterpret_cast<vortex::ExternalValidityMask *>(&validity);
```
myrrc: should be a static_cast since we're downcasting to derived type

vortex-duckdb/src/exporter/constant.rs
```diff
@@ -65,14 +65,14 @@ impl ColumnExporter for ConstantExporter {
...
         _offset: usize,
-        _len: usize,
+        len: usize,
         vector: &mut VectorRef,
         _ctx: &mut ExecutionCtx,
     ) -> VortexResult<()> {
         match self.value.as_ref() {
             None => {
                 // TODO(ngates): would be good if DuckDB supported constant null vectors.
-                vector.set_all_false_validity();
```
myrrc, line 75: set_all_false_validity is more optimal

myrrc (commented):
C++ side LGTM

vortex-duckdb/cpp/include/duckdb_vx/vector.h
joseph-isaacs: validity_ptr is not the buffer is something a few levels of ptr deep. We could fix, but would also want do change Primitive/Decimal Export at once

myrrc (approved):
LGTM

PR #7372, opened 2026-04-09
fix(cuda): handle validity in GPU kernels

0ax1:
All GPU kernels should now  handle validity or fall back to the CPU otherwise in cases they don't.

Dynamic dispatch:
- Propagate root array validity through FusedPlan to output.
- Reject Dict with nullable codes and RunEnd with nullable ends (out-of-bounds risk).
- Short-circuit AllInvalid (skip kernel) and empty arrays.

Standalone kernels:
- RunEnd/Zstd: replace unreachable!()/unimplemented!() with vortex_bail!()
  for unsupported nullable cases, enabling graceful CPU fallback.
- ALP: remove stale TODO; patch validity scatter is unnecessary since the
  encoder already strips null positions from the exception list.

### DIFF
--- a/vortex-cuda/benches/dynamic_dispatch_cuda.rs
+++ b/vortex-cuda/benches/dynamic_dispatch_cuda.rs
@@ -135,6 +135,7 @@ impl BenchRunner {
             dispatch_plan,
             device_buffers,
             shared_mem_bytes,
+            ..
         } = plan.materialize(cuda_ctx).vortex_expect("materialize plan");
 
         let device_plan = Arc::new(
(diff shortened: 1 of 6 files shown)


joseph-isaacs (approved):
do you actually do any work in the plan. is that in a different pr

0ax1:
> do you actually do any work in the plan. is that in a different pr

the plan doesn't really have to do sth in addition here other than attaching the validity to the output array.  the kernels are fine operating on garbage data. the ones that aren't require non-nullable `dict with nullable codes and RunEnd with nullable ends`.

Issue #7297, opened 2026-04-06
Tracking Issue: Vector Similarity Search


connortsui20:
# Vector Similarity Search

We would like to add vector search support to Vortex.

## Plan

At a minimum, we likely need to support (some of this already exists):

- a stable vector representation
- reusable normalization (either as an encoding or potentially a serialized scalar fn - issue potetially coming soon)  
- exact and approximate execution paths (since quantization is lossy)
- vector-specific encodings (like TurboQuant or Spherical transforms)
- compressor/runtime optimizations for common query shapes
- persisted indexes (probably supported by a new layout)
- a clear user-facing model for lossy storage and approximate compute

Note that we still have not decided how to communicate or expose approximation/lossiness in a principled way.

## Design

<details>

- `Vector` should be a first-class data model for fixed-length float embeddings, with clean integration into the dtype system, expression system, serialization, and compression pipeline.
- Similarity metrics should be modeled as first-class operations, not just ad hoc helper functions. At minimum this includes `inner_product`, `cosine_similarity`, and `l2_norm`.
- Unit-norm factoring should be a reusable encoding or representation rather than something hardcoded inside TurboQuant. This gives us a common substrate for cosine similarity, normalized-dot-product rewrites, and future vector encodings.
- Vector encodings should be pluggable and composable in the same spirit as the rest of Vortex. TurboQuant is the first serious candidate, but it should not be the only path and should not have to own normalization, approximation semantics, and query execution by itself (especially since it is lossy).
- Similarity search needs dedicated execution paths for the common “constant query vector against a column of vectors” case. This shows up naturally as `ConstantArray` on one side, and we should not pay the same cost as the fully general pairwise case.

There are also more high-level things that likely deserve their own issues:

- We should distinguish exact scan, encoded scan, and indexed search as separate but interoperable execution strategies. These are different tradeoffs and should be visible to the planner.
- Vector search likely wants a first-class top-k / nearest-neighbor operator rather than forcing users to manually compose “compute similarity for every row, then sort, then limit”.
- Persisted vector indexes should be part of the story. Even if we start with exact flat scan and encoded scan, the architecture should leave room for ANN indexes and row-id based retrieval.
- We need a principled API for approximation and lossiness. Right now `ApproxOptions` is a useful start, but it is not yet a full user-facing contract for what is allowed to be approximate, when approximation is chosen automatically, or how accuracy tradeoffs are surfaced.

</details>

## Steps / History

[...]

PR #7374, opened 2026-04-09
Decompose TurboQuant into 2 `ScalarFnArray`s and `DictArray`

connortsui20:
## Summary

Decomposes `TurboQuant` into:

```text
ScalarFnArray(L2Denorm, [
    ScalarFnArray(SorfTransform, [
        Extension<Vector>(
            FixedSizeListArray(
                DictArray(codes=Primitive<u8>, values=Primitive<f32>),
                padded_dim
            )
        )
    ]),
    norms
])
```

This makes the implementation more modular and turns the TurboQuant-specific pieces into reusable building blocks.

Also defines SORF sign generation with a frozen local `SplitMix64` implementation and cleans up related vector compute code.

## API Changes

- add `SorfTransform`
- remove the `TurboQuant` encoding/array type
- keep `TurboQuantScheme` as the compressor entry point
- simplify a all scalar fn APIs as we no longer need `ApproxOptions`

Note that `ApproxOptions` doesn't actually make sense here because we are doing exact compute here, it is the encoding itself that is lossy. Until we figure out what the exact semantics are of a lossy encoding I will remove this.

## Testing

More tests:
- TurboQuant roundtrip / structural coverage
- SORF roundtrip + norm preservation
- deterministic SplitMix64 / sign generation coverage
- readthrough behavior (when children are normalized) for vector similarity ops


vortex-tensor/src/scalar_fns/l2_norm.rs
```diff
@@ -122,9 +122,9 @@ impl ScalarFnVTable for L2Norm {
         let tensor_flat_size = tensor_match.list_size();
         let element_ptype = tensor_match.element_ptype();
 
-        // L2Norm(L2Denorm(normalized, norms)) == norms, since normalized vectors have unit norm
-        // and L2 norms are non-negative. This avoids decompressing the TQ child just to recompute
-        // norms that are already stored.
+        // L2Norm(L2Denorm(normalized, norms)) is defined to read back the authoritative stored
+        // norms. Exact callers of lossy encodings like TurboQuant opt into that storage semantics
+        // instead of forcing a decode-and-recompute path here.
```
gatesn, line 127: This seems wrong to me... 

As in, I don't think ScalarFn::execute should be matching on child array type should it? CC @joseph-isaacs

gatesn (approved):

vortex-tensor/src/scalar_fns/l2_norm.rs
connortsui20: we resolved this, this is fine

PR #7375, opened 2026-04-09
wip: Optimize BitBuffer methods across the board

joseph-isaacs:
Key optimizations:
- `Not for &BitBuffer`: allocate directly instead of clone+try_into_mut
  which always failed (clone shares Arc, so 2 refs = failure)
  → 27-58% faster bitwise NOT on references
- `iter_bits()`: process u64 chunks instead of byte-by-byte
  → 13% faster iteration
- `PartialEq`: fast path using memcmp for byte-aligned buffers
- `append_buffer()`: memcpy fast path when both sides are byte-aligned
  → 20-52% faster buffer appends
- `append_false()`: remove unnecessary branch (new bytes are zero-init)
  → 65% faster single-bit appends
- `from_indices()`: use set_bit_unchecked directly on the byte slice
- `FromIterator` tail: batch remaining items in u64 words
  → 13% faster from_iter
- `sliced()`: use bitwise_unary_op_copy to avoid clone+fail path,
  fix byte range bug in aligned path
- `filter_bitbuffer_by_indices`: detect contiguous runs for bulk copy
- `filter_bitbuffer_by_slices`: use slice()+append_buffer() instead
  of per-bit get+append
- Add #[inline] to hot methods: set, set_to, append, true_count, etc.

https://claude.ai/code/session_0163XH8LLYAkU2qNQGbmYjhB

### DIFF
--- a/vortex-buffer/src/bit/buf_mut.rs
+++ b/vortex-buffer/src/bit/buf_mut.rs
@@ -181,9 +181,17 @@ impl BitBufferMut {
 
     /// Create a bit buffer of `len` with `indices` set as true.
     pub fn from_indices(len: usize, indices: &[usize]) -> BitBufferMut {
+        let num_bytes = len.div_ceil(8);
         let mut buf = BitBufferMut::new_unset(len);
-        // TODO(ngates): for dense indices, we can do better by collecting into u64s.
-        indices.iter().for_each(|&idx| buf.set(idx));
+        let slice = &mut buf.buffer.as_mut_slice()[..num_bytes];
+
+        // Batch set bits by writing directly to the byte slice.
+        let ptr = slice.as_mut_ptr();
+        for &idx in indices {
+            debug_assert!(idx < len);
+            // SAFETY: idx < len ensures idx/8 < num_bytes.
+            unsafe { set_bit_unchecked(ptr, idx) };
+        }
         buf
     }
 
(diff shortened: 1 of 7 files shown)


vortex-buffer/src/bit/ops.rs
```diff
@@ -87,6 +98,57 @@ pub(super) fn bitwise_unary_op_mut<F: FnMut(u64) -> u64>(buffer: &mut BitBufferM
     }
 }
 
+/// Apply a binary operation with an owned left operand, mutating in-place when possible.
+///
+/// Tries `try_into_mut` on the left operand. If the backing storage has exclusive access,
+/// the operation is performed in-place (zero allocation). Otherwise, falls back to
+/// [`bitwise_binary_op`] which allocates a new buffer.
+pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
```
robert3005, line 106: maybe this should just be a method on the bitbuffer? Not sure why I made these separate functions

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -514,24 +525,39 @@ impl BitBufferMut {
...
-            .as_slice()
-            .view_bits::<bitvec::prelude::Lsb0>();
-
-        // Copy from source buffer (accounting for its offset) to destination (accounting for our offset + len)
-        let source_range = buffer.offset()..buffer.offset() + bit_len;
-        self_slice[start_bit_pos..end_bit_pos].copy_from_bitslice(&other_slice[source_range]);
+        let dst_bit_offset = start_bit_pos % 8;
+        let src_bit_offset = buffer.offset();
+
+        if dst_bit_offset == 0 && src_bit_offset == 0 {
```
robert3005, line 536: I don't think src offset matters here, you care that src end is byte aligned

PR #7376, opened 2026-04-09
CUDA execution for external patches

a10y:
## Summary

This fixes up CUDA execution to adaptively support interior or exterior patches for bit unpacking.

This does not implement Dynamic Dispatch


vortex-cuda/src/kernel/encodings/bitpacked.rs
```diff
@@ -86,10 +95,81 @@ pub fn bitpacked_cuda_launch_config(output_width: usize, len: usize) -> VortexRe
 
 unsafe impl DeviceRepr for GPUPatches {}
 
+/// The patching style, depends on what is supported by the array.
+pub(crate) enum PatchKind {
```
0ax1: We could do this without the extra `PatchKind` enum I think. We could use an `Option<DevicePatches>` and instead of using the `Patched` variant,  construct `DevicePatches` directly, like:
```
  let lane_offsets = array
            .lane_offsets()
            .clone()
            .execute_cuda(ctx)
            .await?
            .into_primitive()
            .into_data_parts()
            .buffer;

        let indices = array
            .patch_indices()
            .clone()
            .execute_cuda(ctx)
            .await?
            .into_primitive()
            .into_data_parts()
            .buffer;

        let values = array
            .patch_values()
            .clone()
            .execute_cuda(ctx)
            .await?
            .into_primitive()
            .into_data_parts()
            .buffer;

        let device_patches = DevicePatches {
            lane_offsets,
            indices,
            values,
        };
```

vortex-cuda/src/kernel/encodings/bitpacked.rs
```diff
@@ -86,10 +95,81 @@ pub fn bitpacked_cuda_launch_config(output_width: usize, len: usize) -> VortexRe
...
+    /// Execute into a set of `DevicePatches` which can be scheduled for fused
+    /// application on the GPU.
+    pub(crate) async fn execute(
+        self,
+        ctx: &mut CudaExecutionCtx,
+    ) -> VortexResult<Option<DevicePatches>> {
+        match self {
+            PatchKind::None => Ok(None),
+            PatchKind::Interior(patches) => {
+                let transposed = transpose_patches(&patches, ctx).await?;
```
0ax1: Should we consider to transpose them on the GPU?

vortex-cuda/src/kernel/encodings/bitpacked.rs
```diff
@@ -86,10 +95,81 @@ pub fn bitpacked_cuda_launch_config(output_width: usize, len: usize) -> VortexRe
...
+                Ok(Some(transposed))
+            }
+            PatchKind::Patched {
+                lane_offsets,
+                patch_indices,
+                patch_values,
+            } => {
+                // Get back some
+                let lane_offsets = lane_offsets
```
0ax1: I guess this is just the reality of separate kernel calls instead of dyn dispatch, but 3 CUDA dispatches seems intense.

vortex-cuda/src/kernel/encodings/bitpacked.rs
```diff
@@ -86,10 +95,81 @@ pub fn bitpacked_cuda_launch_config(output_width: usize, len: usize) -> VortexRe
...
+
+                let values = patch_values
+                    .execute_cuda(ctx)
+                    .await?
+                    .into_primitive()
+                    .into_data_parts()
+                    .buffer;
+
+                Ok(Some(DevicePatches {
+                    lane_offsets: ctx.ensure_on_device(lane_offsets).await?,
```
0ax1: This reads a bit weird right? Do sth on the GPU but then copy the buffer to the CPU. The buffer might be on the CPU if it already was canonical. I do wonder though how we could change the API to make this better. Shouldn't buffers always already be on the GPU when calling execute cuda?

vortex-cuda/src/kernel/encodings/patched.rs
```diff
@@ -0,0 +1,279 @@
...
+                output.freeze(),
+                validity,
+            )))
+        })
+    }
+}
+
+/// Apply patches on top of the existing values (CPU fallback).
+#[allow(clippy::too_many_arguments)]
+fn apply_patches_primitive<V: NativePType>(
```
0ax1: There's an identical impl of this in `vortex-array/src/arrays/patched/vtable/mod.rs`.

vortex-cuda/src/kernel/encodings/patched.rs
```diff
@@ -0,0 +1,279 @@
...
+        let len = array.as_ref().len();
+
+        // Execute inner on GPU and copy back to host
+        let inner_canonical = array.inner().clone().execute_cuda(ctx).await?;
+        let inner_primitive = inner_canonical.into_host().await?.into_primitive();
+        let validity = inner_primitive.validity()?.clone();
+        let ptype = inner_primitive.ptype();
+
+        // Execute patch components on GPU and copy back to host
+        let lane_offsets = array
```
0ax1: I wonder whether we should ever actually take this path but rather panic here. Some form of plan building could prevent taking this path in the first place. In case this is interleaved into other GPU decompression, any movement between host and device and back would tank performance.

vortex-cuda/src/kernel/encodings/patched.rs
```diff
@@ -0,0 +1,279 @@
...
+    use vortex_array::arrays::Patched;
+    use vortex_array::patches::Patches;
+
+    use super::*;
+    use crate::CanonicalCudaExt;
+    use crate::session::CudaSession;
+
+    #[crate::test]
+    fn test_patched_bitpacked() -> VortexResult<()> {
+        let mut cuda_ctx = CudaSession::create_execution_ctx(&VortexSession::empty())
```
0ax1, line 252: Could add tests for sliced arrays, and arrays that are larger than one FL chunk.

vortex-cuda/src/kernel/encodings/patched.rs
0ax1: How about falling back to the standalone patches kernel?

vortex-cuda/src/kernel/encodings/bitpacked.rs
a10y: I actually did try a single-threaded GPU kernel approach previously, which avoids the copy overhead but was ~10x slower than the CPU implementation so it negated the benefits.


image

vortex-cuda/src/kernel/encodings/bitpacked.rs
a10y: The reason we need this is if we call the cuda execute on a PrimitiveArray that lives on the cpu, then execute_cuda() is a no-op (maybe it shouldn't be?)

vortex-cuda/src/kernel/encodings/patched.rs
a10y: fixed

vortex-cuda/src/kernel/encodings/patched.rs
a10y: added some more tests

vortex-cuda/src/kernel/encodings/bitpacked.rs
0ax1: Yeah, that's seems like an impl detail. I'd assume for someone new to Vortex this behavior would be unexpected. I kinda want the buffers having to live on device when execute cuda is called. But nothing we need to address in this PR obviously.

vortex-cuda/src/kernel/encodings/bitpacked.rs
a10y: yea :/

vortex-cuda/src/kernel/encodings/bitpacked.rs
a10y: removed

vortex-cuda/kernels/src/patched.cu
```diff
@@ -0,0 +1,65 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#include "types.cuh"
+
+/// Apply patches to an output array using the transposed Patched array format.
+///
+/// This kernel uses a thread-per-lane model where each thread is assigned to
+/// one (chunk, lane) slot and applies all patches in that slot.
+template <typename ValueT>
+__device__ void patched(ValueT *const output,
```
0ax1, line 11: Nice! Ideally we could move to a world where we drop the other `patches.cu` eventually and only operate on transposed patches on the GPU, maybe.

vortex-cuda/src/kernel/encodings/patched.rs
```diff
@@ -0,0 +1,498 @@
...
+                .execute_cuda(ctx)
+                .await?
+                .into_primitive()
+                .into_data_parts()
+                .buffer;
+
+            match_each_integer_ptype!(bitpacked.ptype(bitpacked.dtype()), |P| {
+                return decode_bitpacked::<P>(
+                    bitpacked.into_owned(),
+                    P::default(),
```
0ax1, line 99: We could also handle the fused case here right: FoR + BP?

vortex-cuda/src/kernel/encodings/patched.rs
```diff
@@ -0,0 +1,498 @@
...
+            // The inner BitPacked should not have its own interior patches since they've
+            // been externalized into the Patched wrapper
+            if bitpacked.patches().is_some() {
+                return Err(vortex_err!(
+                    "Patched(BitPacked) should not have interior patches in BitPacked child"
+                ));
+            }
+
+            // Execute the components
+            let lane_offsets = array
```
0ax1, line 69: This could go in a separate fn:
```
            let lane_offsets = array
                .lane_offsets()
                .clone()
                .execute_cuda(ctx)
                .await?
                .into_primitive()
                .into_data_parts()
                .buffer;

            let patch_indices = array
                .patch_indices()
                .clone()
                .execute_cuda(ctx)
                .await?
                .into_primitive()
                .into_data_parts()
                .buffer;

            let patch_values = array
                .patch_values()
                .clone()
                .execute_cuda(ctx)
                .await?
                .into_primitive()
                .into_data_parts()
                .buffer;
```

0ax1 (approved):
LGTM - Added a couple more thoughts.

a10y:
Benchmark results for

```
cargo bench -p vortex-cuda --bench bitpacked_cuda -- patch
```

image

a10y:
I think we're going to change layout of Patched array to be sorted w/chunk offsets instead of lane-split. See https://github.com/vortex-data/vortex/pull/7440

PR #7380, opened 2026-04-10
fix: preserve operand width in DecimalValue checked arithmetic

### LINKED ISSUE #7022: DecimalValue checked_add should not upcast to i256 unconditionally
https://github.com/vortex-data/vortex/blob/bf61c076049fc50455f8d9286642d310e2476137/vortex-array/src/scalar/typed_view/decimal/dvalue.rs#L133

Find the max size of left and right and use that.

This just hides an upcast which can lead to other large scalar for no reason.

abnobdoss:
## Summary

Closes: #7022

`DecimalValue::checked_add/sub/mul/div` unconditionally upcast both operands to `i256` and returned `DecimalValue::I256`, producing unnecessarily wide scalars even when both inputs were narrow (e.g. `I32 + I32 → I256`).

Operate at `max(self, other)` width instead, matching the pattern in `aggregate_fn/fns/sum/decimal.rs`. The old `checked_binary_op` helper was replaced with a local macro so each op dispatches with its own trait.

**AI disclosure:** Analysis, implementation, and tests were done with Claude Code under my direction and review.

## API Changes

No public API signature changes (verified via `./scripts/public-api.sh`).

Overflow is now caught at the target width rather than silently widening (e.g. `I8(i8::MAX) + I8(1)` now returns `None` instead of `I256(128)`). **This felt like the most faithful reading of the issue, but I'd appreciate a sanity check that returning `None` on target-width overflow is the desired semantics rather than, say, promoting to the next wider variant.** No in-tree caller depends on the old behavior: `sum/mod.rs` pre-widens the accumulator by +10 precision, `decimal/scalar.rs::checked_binary_numeric` requires both operands to share the same width, and `sum/constant.rs` uses `I128` as the multiplier.

## Testing

- Updated `test_decimal_value_checked_*` to assert the correct variant.
- Added tests for variant preservation, mixed-width promotion, and overflow at the target width (including `i8::MIN / -1`).
- All existing tests pass (212 decimal, 14 sum).


joseph-isaacs:
I256 is not the maximum of the inputs. I would accept a PR doing this

abnobdoss:
Hi @joseph-isaacs, sorry not sure I understood - are you saying this change is fine? If so, what would the process be to get this approved / merged?

vortex-array/src/scalar/typed_view/decimal/dvalue.rs
```diff
@@ -19,6 +19,19 @@ use crate::dtype::NativeDecimalType;
...
 use crate::match_each_decimal_value;
+use crate::match_each_decimal_value_type;
+
+/// Performs a checked binary operation at the wider of the two operand types.
+macro_rules! checked_binary_op {
+    ($self:expr, $other:expr, $op:path) => {{
+        let target = $self.decimal_type().max($other.decimal_type());
+        match_each_decimal_value_type!(target, |T| {
+            let a: T = $self.cast()?;
+            let b: T = $other.cast()?;
```
joseph-isaacs: How can this fail?

vortex-array/src/scalar/typed_view/decimal/dvalue.rs
```diff
@@ -19,6 +19,19 @@ use crate::dtype::NativeDecimalType;
 use crate::dtype::ToI256;
 use crate::dtype::i256;
 use crate::match_each_decimal_value;
+use crate::match_each_decimal_value_type;
+
+/// Performs a checked binary operation at the wider of the two operand types.
+macro_rules! checked_binary_op {
```
joseph-isaacs: Can we signify the upcast
```suggestion
macro_rules! checked_widening_binary_op {
```

joseph-isaacs (commented):
Thanks for this sorry, for delay just a few small ones and we are good to go

abnobdoss:
Thank you @joseph-isaacs! I've updated the PR per the comments.

abnobdoss:
Hi @joseph-isaacs, could you please take a look when you get a chance?

abnobdoss:
Hi @joseph-isaacs, could you please take a look when you get a chance?

robert3005 (approved):

PR #7381, opened 2026-04-10
fix[vortex-array]: don't destroy encodings in filter grandchild

asubiotto:
I believe this is a regression. A Filter(Slice(Ree)) is pretty common and would eagerly canonicalize its child preventing execute_parent kernels like RunEnd's FilterKernel from firing. This is an issue with dict encoding too.

This commit executes its child one step so that execute_parent kernels may match.

## Summary
See above

## Testing

Regression test to explicitly verify execute_until result encoding.

### DIFF
--- a/encodings/runend/src/compute/filter.rs
+++ b/encodings/runend/src/compute/filter.rs
@@ -115,6 +115,8 @@ fn filter_run_end_primitive<R: NativePType + AddAssign + From<bool> + AsPrimitiv
 #[cfg(test)]
 mod tests {
     use vortex_array::IntoArray;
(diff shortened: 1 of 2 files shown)


asubiotto:
@joseph-isaacs I think this is what I mentioned in our sync and the cause behind seeing filters destroy ree/dict encodings.

joseph-isaacs (changes requested):
I will look into this, but this is not the way to do this

joseph-isaacs:
https://github.com/vortex-data/vortex/pull/7386 i think this is the fix. i'll get it merged on monday

asubiotto:
Closing in favor of #7386

PR #7388, opened 2026-04-10
Remove portable_simd take implementation and remove unused cargo flags

robert3005:
First of all we never use nightly so portable simd never gets enabled, second of
all it looks like avx512 doesn't bring any benefits beyond avx2 for simd
gather per gcc mailing list and Intel GDS advisory. AVX2 on modern intel
machines is comparable to Zen 5 machines and on Zen 5 machines AVX512, AVX2 and
scalar versions behave the same.


0ax1 (approved):

Cargo.toml
```diff
@@ -315,9 +315,6 @@ unsafe_op_in_unsafe_fn = "deny"
 unused_lifetimes = "deny"
 unused_qualifications = "deny"
 unexpected_cfgs = { level = "deny", check-cfg = [
-    "cfg(codspeed)",
```
a10y, line 318: looks like you still need this one?

0ax1:
_"at least we tried, portable simd"_

.github/workflows/ci.yml
```diff
@@ -409,7 +409,7 @@ jobs:
       VORTEX_SKIP_SLOW_TESTS: "1"
       # -Cunsafe-allow-abi-mismatch=sanitizer: libraries like compiler_builtins
       # unset -Zsanitizer flag and we should allow that.
-      RUSTFLAGS: "-A warnings -Cunsafe-allow-abi-mismatch=sanitizer --cfg disable_loom --cfg vortex_nightly -C debuginfo=2 -C opt-level=0 -C strip=none"
```
0ax1, line 412: why `--cfg disable_loom`

Cargo.toml
0ax1: ➕

a10y:
I guess this never really did anything on Neon since it doesn't have a Gather right?

Cargo.toml
robert3005: my search fooooo is not good enough

.github/workflows/ci.yml
robert3005: there's no more any loom tests that I could find

robert3005:
yes, Neon doesn't have any specialised gather. SVE does but only hardware with support for that are AWS Graviton instances. I wonder how this looks like on on Nvidia Grace cpus. Anyway, only benefit I could find with benchmarking was AVX2 on intel machines.

PR #7389, opened 2026-04-10
Allow "," in paths without globs

myrrc:
Allow paths like "1.vortex,2.vortex" if each sub-part doesn't contain glob expressions.


joseph-isaacs:
is this standard duckdb syntax?

a10y:
Possible to get some tests?

a10y:
Actually yea, this might not work for DuckDB? I think instead we likely want to support multiple glob patterns in the multifile reader, and then in duckdb extension we can dispatch based on if input is string or list.

myrrc:
+1, https://github.com/vortex-data/vortex/pull/7390 is a better implementation.

PR #7390, opened 2026-04-10
support multiple globs in DuckDB and MultiFileDataSource

a10y:
## Summary

Adds support for multiple globs in DuckDB, similar to Parquet,

E.g. instead of `read_vortex('*.vortex')` now you can also do `read_vortex(['s3://bucket1/path/*.vortex', '/a/b/*.vortex'])`

Note that multiple file systems are supported in a single scan.

This is implemented and wired through into the underlying Rust MultiFileDataSource.

## Testing

We add new DuckDB e2e tests for multi-glob

Couldn't figure out a great way to test multiple file systems, just multiple globs


myrrc (approved):
LGTM

PR #7391, opened 2026-04-10
add vector similarity benchmark

connortsui20:
## Summary

Adds basic benchmarking setup for vector similarity. Right now it is just a bunch of random vectors.

Note that the numbers dont really mean anything right now as we have yet to optimize anything (namely I have not yet added the inner product / cosine similarity optimizations pushed through both the SORF transform and the dictionary for constant array).

In the future we will add proper benchmarking on real datasets (likely in `vortex-bench`, and also maybe we will integrate https://github.com/zilliztech/vectordbbench

## Testing

N/A

### DIFF
--- a/vortex-tensor/Cargo.toml
+++ b/vortex-tensor/Cargo.toml
@@ -31,6 +31,14 @@ num-traits = { workspace = true }
 prost = { workspace = true }
 
 [dev-dependencies]
+divan = { workspace = true }
+mimalloc = { workspace = true }
 rand = { workspace = true }
 rand_distr = { workspace = true }
 rstest = { workspace = true }
+vortex-btrblocks = { path = "../vortex-btrblocks" }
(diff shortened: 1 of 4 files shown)


gatesn (approved):

vortex-tensor/Cargo.toml
```diff
@@ -31,6 +31,14 @@ num-traits = { workspace = true }
 prost = { workspace = true }
 
 [dev-dependencies]
+divan = { workspace = true }
+mimalloc = { workspace = true }
 rand = { workspace = true }
 rand_distr = { workspace = true }
 rstest = { workspace = true }
+vortex-btrblocks = { workspace = true }
```
a10y: switch this to `{ path = "../vortex-btrblocks" }` and you should be fine

PR #7393, opened 2026-04-10
Add TakeExecute implementation for FilterArray

robert3005:
This seems to be an improvement if the take array is small. Falls back to regular execution path if we have too many indices


robert3005:
Random access benchmark changes seem real. I think take on a bitbuffer would always be faster than take on the filtered array that needs to copy values. I will do some refactoring here before this is ready for review

vortex-buffer/src/bit/select.rs
```diff
@@ -0,0 +1,428 @@
...
+///
+/// The returned position is relative to the logical start (i.e., 0-indexed from `offset`).
+/// Returns `None` if `nth` is out of bounds.
+///
+/// Uses architecture-specific optimizations:
+/// - **aarch64**: NEON `vcnt`-based popcount for the word-level scan.
+/// - **x86_64 + BMI2**: `pdep` + `tzcnt` for the final in-word select.
+/// - **Scalar fallback**: 4× unrolled word scan with `count_ones`, byte-level narrowing.
+#[inline]
+pub fn bit_select(bytes: &[u8], offset: usize, len: usize, nth: usize) -> Option<usize> {
```
joseph-isaacs, line 19: I think this we want a fast path that tried to skip whole chunks 512 bits at a time with a ==0?

vortex-array/src/arrays/filter/take.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::arrays::filter::FilterArrayExt;
+use crate::builtins::ArrayBuiltins;
+use crate::dtype::DType;
+use crate::executor::ExecutionCtx;
+use crate::kernel::ParentKernelSet;
+use crate::scalar::Scalar;
+
+pub(super) const PARENT_KERNELS: ParentKernelSet<Filter> =
+    ParentKernelSet::new(&[ParentKernelSet::lift(&TakeExecuteAdaptor(Filter))]);
+
```
joseph-isaacs: can you move this into either the vtable or a kernels files

vortex-buffer/src/bit/select.rs
robert3005: we can add vbmi2 here like in count_ones and bit transpose

robert3005:
~I need to have a look into statpopgen, clearly it exercises codepath I didn't think about~ all noise

joseph-isaacs (commented):
Looks good to me, lets run the benchmarks one more time and check for regressions

joseph-isaacs (approved):

PR #7397, opened 2026-04-11
Add Vector benchmarks to suite

connortsui20:
## Summary

Adds a `vector-search-bench` crate similar to [VectorDBBench](https://github.com/zilliztech/vectordbbench).

The benchmark brute-forces cosine similarity search over public VectorDBBench datasets (Cohere, OpenAI, etc). 

Since Vortex is not a database, we do not measure things like vector inserts and deletes. instead we just measure storage size and throughput for four targets:

- Hand-rolled Rust baseline over `&[f32]`
- Uncompressed (canonical) Vortex
- Default compressed (which ends up being ALPrd)
- TurboQuant (plus Recall@10 for the lossy TurboQuant path).

Every variant goes through a correctness check against the uncompressed scan before timing.

Not sure if this makes sense to have on a PR benchmark yet since this can only happen on a very specific array tree that will have a very specific optimized implementation.

## Testing

N/A


connortsui20:
claude code web fell over so this doesnt even compile

connortsui20:
Superseded by https://github.com/vortex-data/vortex/pull/7399

PR #7401, opened 2026-04-12
perf[vortex-array]: use from_trusted_len_iter in primitive casts

asubiotto:
Some of our scan profiles show 10% of scan cpu time is spent in integer widening casts (nullable dictionary codes). This commit simplifies and optimizes primitive casts by hoisting a lot of hot loop branching logic.

Specifically, this commit relies on values_fit_in to verify representability so that we can avoid a potential validity and error check in the hot loop. Additionally from_trusted_len_iter lets the destination BufferMut optimize the actual cast instead of using push_unchecked for each element.

Results locally running the benchmark from #7400. Before:
```
$ cargo bench -p vortex-array --bench cast_primitive
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.16s
     Running benches/cast_primitive.rs (target/release/deps/cast_primitive-598823f32b8f3db0)
Timer precision: 41 ns
cast_primitive      fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ cast_u16_to_u32  384.2 µs      │ 491.6 µs      │ 395.1 µs      │ 397.3 µs      │ 100     │ 100
```
After:
```
cargo bench -p vortex-array --bench cast_primitive
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.17s
     Running benches/cast_primitive.rs (target/release/deps/cast_primitive-598823f32b8f3db0)
Timer precision: 41 ns
cast_primitive      fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ cast_u16_to_u32  6.874 µs      │ 543.7 µs      │ 6.999 µs      │ 12.7 µs       │ 100     │ 100
```

## Summary

Closes: #000

### DIFF
--- a/vortex-array/benches/cast_primitive.rs
+++ b/vortex-array/benches/cast_primitive.rs
@@ -9,6 +9,7 @@ use vortex_array::builtins::ArrayBuiltins;
 use vortex_array::dtype::DType;
 use vortex_array::dtype::Nullability;
 use vortex_array::dtype::PType;
+use vortex_array::expr::stats::Stat;
 
 fn main() {
     divan::main();
@@ -28,6 +29,8 @@ fn cast_u16_to_u32(bencher: Bencher) {
         }
     }))
     .into_array();
+    // Pre-compute min/max so values_fit_in is a cache hit during the benchmark.
+    arr.statistics().compute_all(&[Stat::Min, Stat::Max]).ok();
     bencher.with_inputs(|| arr.clone()).bench_refs(|a| {
         #[expect(clippy::unwrap_used)]
         a.cast(DType::Primitive(PType::U32, Nullability::Nullable))
--- a/vortex-array/src/arrays/primitive/compute/cast.rs
+++ b/vortex-array/src/arrays/primitive/compute/cast.rs
@@ -1,13 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use num_traits::AsPrimitive;
 use vortex_buffer::Buffer;
 use vortex_buffer::BufferMut;
 use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
-use vortex_error::vortex_err;
-use vortex_mask::AllOr;
-use vortex_mask::Mask;
 
 use crate::ArrayRef;
 use crate::ExecutionCtx;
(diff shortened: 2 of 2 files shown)


asubiotto:
I *think* it's correct to use `values_fit_in` for floats too. I initially had this as an int to int only optimization.

vortex-array/src/arrays/primitive/compute/cast.rs
```diff
@@ -104,30 +99,12 @@ fn values_fit_in(
...
-            }
-            Ok(buffer.freeze())
-        }
-    }
+/// Caller must ensure all valid values are representable via `values_fit_in`;
+/// `unwrap_or_default` only fires at invalid positions where the physical
+/// value is out of range.
+fn cast<F: NativePType, T: NativePType>(array: &[F]) -> Buffer<T> {
+    BufferMut::from_trusted_len_iter(array.iter().map(|&src| T::from(src).unwrap_or_default()))
+        .freeze()
```
joseph-isaacs: You need to filter the cast by the nullability. 

We don't have a test that add null values over very large primitive values.

We can do this for all valid or nonnullable values

asubiotto: We wrap the result in the new_validity and this is what makes the nullable case fast (important for dictionary codes). We cast invalid positions here, but the result is ::Default (if not representable) and wrapped with validity so I think this is fine. Will add a test with large primitive values.

asubiotto: Done. Chose to make it a separate test so that it's explicit.

vortex-array/benches/cast_primitive.rs
```diff
@@ -28,6 +29,8 @@ fn cast_u16_to_u32(bencher: Bencher) {
         }
     }))
     .into_array();
+    // Pre-compute min/max so values_fit_in is a cache hit during the benchmark.
+    arr.statistics().compute_all(&[Stat::Min, Stat::Max]).ok();
```
asubiotto, line 33: This is reasonable to do to avoid a noisy benchmark on the first run. I think it's reasonable to assume these will be computed.

vortex-array/src/arrays/primitive/compute/cast.rs
joseph-isaacs: Maybe we can use an `as` here and avoid the unwrap? making this branchless

asubiotto: Yep! Glad you think so, I kept T::from since I didn't want to deviate from what was previously done. Will update.

asubiotto: Done

robert3005 (approved):

PR #7408, opened 2026-04-13
Ignore dead_code warnings on scalar bit transpose code

robert3005:
Likely we shouldn't use expect here but we also don't test on architectures
where this wouldn't be dead code


encodings/fastlanes/src/bit_transpose/scalar.rs
```diff
@@ -12,6 +12,7 @@ use crate::bit_transpose::TRANSPOSE_8X8;
 /// This version uses 64-bit gather + parallel bit operations instead of
 /// extracting bits one by one. Typically 5-10x faster than the basic scalar version.
 #[inline(never)]
+#[allow(dead_code)]
```
0ax1, line 15: In practice looks like a `if x86`

0ax1 (approved):

PR #7409, opened 2026-04-13
Remove needless fmt_sql

gatesn:
Remove bespoke `fmt_sql` implementations for scalar functions that do not need special syntax. These overrides only duplicated the generic call formatting and could drift from the registered `ScalarFnId`; relying on the default keeps expression display consistent, preserves qualified function names/options, and reduces boilerplate for new scalar functions.


connortsui20 (approved):
nice

PR #7412, opened 2026-04-13
perf: intern identifiers (array, layout, etc).

joseph-isaacs:
We use id for scalar_fns, array and layouts. This we used to match rules among other things.

This PR replace Arc<str> ID with interned str (u32) id.


encodings/alp/src/alp/array.rs
```diff
@@ -238,7 +239,13 @@ impl Display for ALPData {
 pub struct ALP;
 
 impl ALP {
-    pub const ID: ArrayId = ArrayId::new_ref("vortex.alp");
+    pub const ID: &'static str = "vortex.alp";
+
+    /// Returns the cached [`ArrayId`] for this encoding.
+    pub fn array_id() -> ArrayId {
```
gatesn: Any reason we need the wrapping function?

encodings/alp/src/alp/array.rs
```diff
@@ -238,7 +239,13 @@ impl Display for ALPData {
 pub struct ALP;
 
 impl ALP {
-    pub const ID: ArrayId = ArrayId::new_ref("vortex.alp");
+    pub const ID: &'static str = "vortex.alp";
```
gatesn: Should we avoid exposing this as &'static str? Makes it too easy to go down the non-interned path?

vortex-session/src/registry.rs
```diff
@@ -4,17 +4,171 @@
...
+        let state = self.state.load_full();
+        if let Some(symbol) = state.interner.get(s) {
+            return symbol;
+        }
+
+        let s: &'static str = Box::leak(s.to_owned().into_boxed_str());
+        let mut next = (*state).clone();
+        let symbol = next.interner.get_or_intern_static(s);
+        debug_assert_eq!(symbol.to_usize(), next.strings.len());
+        next.strings.push(s);
```
gatesn: Symbol is `get_or_intern_static`, what if someone else interned the string. You're now pushing a second string into next.strings?

encodings/alp/src/alp/array.rs
joseph-isaacs: where else do we put the static either here or it global, (you cannot have in an impl block)

vortex-array/src/vtable/dyn_.rs
```diff
@@ -0,0 +1,239 @@
...
+/// Dynamically typed vtable trait.
+///
+/// This trait is sealed, therefore users should implement the strongly typed [`VTable`] trait
+/// instead. The [`ArrayVTableExt::vtable`] function can be used to lift the implementation into
+/// this object-safe form.
+///
+/// This trait contains the implementation API for Vortex arrays, allowing us to keep the public
+/// [`DynArray`] trait API to a minimum.
+pub trait DynVTable: 'static + private::Sealed + Send + Sync + Debug {
+    #[allow(clippy::too_many_arguments)]
```
connortsui20: expect

vortex-array/src/vtable/dyn_.rs
connortsui20, line 1: dead code

connortsui20 (approved):

vortex-array/src/vtable/dyn_.rs
connortsui20: actually this is dead code

PR #7416, opened 2026-04-13
Propagate min/max/string length statistics to duckdb

myrrc:
Propagate min/max/max_length/nullable stats for single-file reads in duckdb.

Fix a bug in file statistics merge where it assumed sum to be applied only on primitive types.

### DIFF
--- a/vortex-cuda/src/layout.rs
+++ b/vortex-cuda/src/layout.rs
@@ -3,6 +3,7 @@
 
 //! A CUDA-optimized flat layout that inlines small constant array buffers into layout metadata.
 
+use std::any::Any;
 use std::collections::BTreeSet;
 use std::ops::BitAnd;
 use std::ops::Range;
@@ -382,7 +383,7 @@ impl LayoutReader for CudaFlatReader {
         .boxed())
     }
 
-    fn as_any(&self) -> &dyn std::any::Any {
+    fn as_any(&self) -> &dyn Any {
         self
     }
 }
--- a/vortex-cuda/src/layout.rs
+++ b/vortex-cuda/src/layout.rs
@@ -381,6 +381,10 @@ impl LayoutReader for CudaFlatReader {
         }
         .boxed())
     }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
 }
 
 /// A [`LayoutStrategy`] that writes a [`CudaFlatLayout`] with constant array buffers inlined
--- a/vortex-duckdb/cpp/table_function.cpp
+++ b/vortex-duckdb/cpp/table_function.cpp
@@ -179,10 +179,10 @@ c_statistics(ClientContext &context, const FunctionData *bind_data, column_t col
         return string_stats(statistics, type);
     }
     case LogicalTypeId::STRUCT: {
-        // Duckdb's has_null has a different meaning for structs.
-        // our DType non-nullability is recursive i.e. a non-null struct can't
-        // contain null children.
-        // Duckdb applies non-nullability only for the outer layer.
+        // TODO(myrrc)
+        // Duckdb's has_null has a different semantics for structs.
+        // If we propagate our has_null, this breaks Duckdb optimizer.
+        // You can reproduce it in struct.slt test in vortex-sqllogictests:
         return {};
     }
     default:
--- a/vortex-duckdb/cpp/table_function.cpp
+++ b/vortex-duckdb/cpp/table_function.cpp
@@ -107,6 +107,9 @@ unique_ptr<BaseStatistics> numeric_stats(duckdb_column_statistics &stats, Logica
         NumericStats::SetMax(out, UnwrapValue(stats.max));
         duckdb_destroy_value(&stats.max);
     }
+    if (!stats.has_null) {
+        out.Set(StatsInfo::CANNOT_HAVE_NULL_VALUES);
+    }
     return out.ToUnique();
 }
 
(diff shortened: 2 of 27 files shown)


vortex-duckdb/src/datasource.rs
```diff
@@ -412,6 +445,17 @@ impl<T: DataSourceTableFunction> TableFunction for T {
         Ok(false)
     }
 
+    fn statistics<'a>(
+        _client_context: &ClientContextRef,
+        bind_data: &'a Self::BindData,
+        column_index: usize,
+    ) -> &'a ColumnStatistics {
+        match &bind_data.stats {
+            DataSourceStatistics::All(items) => &items[column_index],
```
gatesn: We should lazily construct stats for each column. Could be way fewer columns that DuckDB asks for

vortex-file/src/footer/mod.rs
```diff
@@ -146,6 +146,10 @@ impl Footer {
         self.statistics.as_ref()
     }
 
+    pub fn take_statistics(&mut self) -> Option<FileStatistics> {
```
gatesn: Why would we take stats?

vortex-file/src/file.rs
```diff
@@ -73,6 +73,10 @@ impl VortexFile {
         self.footer.statistics()
     }
 
+    pub fn take_file_stats(&mut self) -> Option<FileStatistics> {
```
gatesn: Again, kill it

joseph-isaacs:
Looks like some real regressions here.

Opening all the files AoT is not a good idea, its very expensive.

gatesn (commented):
I think the regression might be eagerly merging all stats for all columns, when we only need min/max/str_len for columns that DuckDB asks us about

myrrc:
statistics isn't the right function here as parquet only reports it for a single file (I guess they also tried eager readers first). I'll keep it, but also we need get_partition_stats with cached metadata

vortex-layout/src/reader.rs
```diff
@@ -27,10 +28,12 @@ pub type LayoutReaderRef = Arc<dyn LayoutReader>;
 
 /// A [`LayoutReader`] is used to read a [`crate::Layout`] in a way that can cache state across multiple
 /// evaluation operations.
-pub trait LayoutReader: 'static + Send + Sync {
+pub trait LayoutReader: 'static + Send + Sync + Any {
```
robert3005: I don't think you need this

vortex-duckdb/src/datasource.rs
```diff
@@ -97,15 +106,21 @@ pub(crate) trait DataSourceTableFunction: Sized + Debug {
     }
 
     /// Bind the table function and return a [`DataSourceRef`].
-    fn bind(ctx: &ClientContextRef, input: &BindInputRef) -> VortexResult<DataSourceRef>;
+    fn bind(ctx: &ClientContextRef, input: &BindInputRef) -> VortexResult<MultiLayoutDataSource>;
+}
+
+#[derive(Debug, Clone)]
+struct ColumnTriple {
```
robert3005: Call it a DuckdbField? Triple doesn't tell me what it's used for besied that it has 3 things

vortex-cuda/src/layout.rs
```diff
@@ -381,6 +381,10 @@ impl LayoutReader for CudaFlatReader {
         }
         .boxed())
     }
+
+    fn as_any(&self) -> &dyn std::any::Any {
```
robert3005: you want to import this

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -88,6 +93,103 @@ double c_table_scan_progress(ClientContext &context,
...
+    case LogicalTypeId::VARCHAR:
+    case LogicalTypeId::BLOB: {
+        return string_stats(statistics, type);
+    }
+    case LogicalTypeId::STRUCT: {
+        // Duckdb's has_null has a different meaning for structs.
+        // our DType non-nullability is recursive i.e. a non-null struct can't
+        // contain null children.
+        // Duckdb applies non-nullability only for the outer layer.
+        return {};
```
robert3005, line 186: This comment doesn't make sense to me. The two interpretations for nulls on structs are parent structs are laid out on top of children nulls or child nulls have to contain all the parent nulls as well. Since stats in duckdb also look recursive, there's a child_stats field I think our definitions would be similar?

vortex-duckdb/src/datasource.rs
```diff
@@ -163,6 +176,71 @@ fn progress(bytes_read: &AtomicU64, bytes_total: &AtomicU64) -> f64 {
...
+            Scalar::try_new(dtype.clone(), Some(value.clone()))
+                .vortex_expect("scalar dtype and value are incompatible")
+                .try_to_duckdb_scalar()
+                .vortex_expect("can't convert Scalar to duckdb Value")
+        });
+
+        let max_string_length = stats
+            .max_string_length
+            .map_or(0, |len| (1u64 << 63) | (len as u64));
+        let has_null = dtype.is_nullable();
```
robert3005: This is not quite right, you have to look at the null_count. Nullable types might not have nulls

0ax1:
@claude review in depth for correctness and performance

vortex-duckdb/src/datasource.rs
```diff
@@ -412,6 +486,33 @@ impl<T: DataSourceTableFunction> TableFunction for T {
...
+        // regression. Duckdb's Parquet reader only gets metadata for multiple
+        // files with a UNION BY NAME and we don't support it (yet)
+        // https://github.com/duckdb/duckdb/blob/471de9f0e0e157ae672e56710e8c43b132a5ddc4/src/include/duckdb/common/multi_file/multi_file_function.hpp#L691
+        if children.len() != 1 {
+            return None;
+        }
+        let MultiLayoutChild::Opened(ref reader) = children[0] else {
+            return None;
+        };
+        let stats_sets = match reader.as_any().downcast_ref::<FileStatsLayoutReader>() {
```
robert3005, line 519: This is a little bit meh, I think we need to work on the api

robert3005 (commented):
There's couple of things that can be improved. The general problem we have to solve (not here, right now) is how we get stats out of layouts

vortex-duckdb/src/datasource.rs
myrrc: +1, I have a branch which adds layout statistics but I'd prefer to merge it separately

vortex-duckdb/cpp/table_function.cpp
myrrc: +1, better to add a comment

robert3005 (approved):

PR #7417, opened 2026-04-13
perf: remove `ExecutionCtx` from `OperationsVTable::scalar_at`

0ax1:
## Summary

- Removes all usages of `ExecutionCtx` in `scalar_at`

## Profiling

Profiled with [apmc](https://github.com/nickhrsims/apmc). ClickBench, Vortex format, Apple Silicon M4 Max, 3 iterations:

| Metric | Before | After | Delta |
|---|---|---|---|
| Cycles | 1,973.5B | 1,942.6B | **-1.6%** |
| Instructions | 4,730.2B | 4,722.5B | -0.16% |
| IPC | 2.40 | 2.43 | **+1.25%** |
| Wall clock | 43.09s | 41.92s | **-2.7%** |
| Dispatch stalls | 593.7B | 573.9B | **-3.3%** |
| L1D cache misses | 80.7B | 80.6B | ~0% |
| Branch mispredicts | 11.2B | 11.2B | ~0% |

### DIFF
--- a/encodings/pco/src/array.rs
+++ b/encodings/pco/src/array.rs
@@ -626,11 +626,7 @@ impl ValidityVTable<Pco> for Pco {
 }
 
 impl OperationsVTable<Pco> for Pco {
-    fn scalar_at(
-        array: ArrayView<'_, Pco>,
-        index: usize,
-        _ctx: &mut ExecutionCtx,
-    ) -> VortexResult<Scalar> {
+    fn scalar_at(array: ArrayView<'_, Pco>, index: usize) -> VortexResult<Scalar> {
         let mut ctx = LEGACY_SESSION.create_execution_ctx();
         let unsliced_validity = child_to_validity(&array.slots()[0], array.dtype().nullability());
         array
--- a/encodings/pco/src/array.rs
+++ b/encodings/pco/src/array.rs
@@ -28,10 +28,8 @@ use vortex_array::ArrayView;
 use vortex_array::ExecutionCtx;
 use vortex_array::ExecutionResult;
 use vortex_array::IntoArray;
-use vortex_array::LEGACY_SESSION;
 use vortex_array::Precision;
 use vortex_array::ToCanonical;
-use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::Primitive;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::buffer::BufferHandle;
@@ -54,6 +52,7 @@ use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
 use vortex_error::vortex_ensure;
 use vortex_error::vortex_err;
+use vortex_mask::Mask;
 use vortex_session::VortexSession;
 
 use crate::PcoChunkInfo;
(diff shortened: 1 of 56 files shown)


joseph-isaacs (changes requested):
Where do we create this execution ctx.

This idea is that this should never happen.

0ax1:
> Where do we create this execution ctx.
> 
> This idea is that this should never happen.

We did in `Patched`. Removed as part of https://github.com/vortex-data/vortex/commit/f70e24120f0a581f77e92f8794433293d50d44d0.

encodings/zstd/src/array.rs
```diff
@@ -1020,11 +1020,7 @@ impl ValidityVTable<Zstd> for Zstd {
 }
 
 impl OperationsVTable<Zstd> for Zstd {
-    fn scalar_at(
-        array: ArrayView<'_, Zstd>,
-        index: usize,
-        _ctx: &mut ExecutionCtx,
-    ) -> VortexResult<Scalar> {
+    fn scalar_at(array: ArrayView<'_, Zstd>, index: usize) -> VortexResult<Scalar> {
         let mut ctx = LEGACY_SESSION.create_execution_ctx();
```
0ax1: This should also go away.

encodings/pco/src/array.rs
```diff
@@ -626,11 +626,7 @@ impl ValidityVTable<Pco> for Pco {
 }
 
 impl OperationsVTable<Pco> for Pco {
-    fn scalar_at(
-        array: ArrayView<'_, Pco>,
-        index: usize,
-        _ctx: &mut ExecutionCtx,
-    ) -> VortexResult<Scalar> {
+    fn scalar_at(array: ArrayView<'_, Pco>, index: usize) -> VortexResult<Scalar> {
         let mut ctx = LEGACY_SESSION.create_execution_ctx();
```
0ax1: and here..

encodings/zstd/src/array.rs
joseph-isaacs, line 1028: We can plumb the _ctx -> ctx?

0ax1: Currently having a peek whether ctx should be here at all. Or more in general, scalar_ at shouldn't need a ctx in the first place?

a10y: I added this for Patched, so I could execute a range of one of the children for fast random access. but we can probably implement the same thing using search_sorted, similar to what we previously do with BitPacked.

that doesn't change pco/zstd though

vortex-array/src/arrays/patched/vtable/operations.rs
```diff
@@ -31,13 +25,18 @@ impl OperationsVTable<Patched> for Patched {
...
         let range = array.lane_range(chunk, lane)?;
 
-        // Get the range of indices corresponding to the lane, potentially decoding them to avoid
-        // the overhead of repeated scalar_at calls.
-        let patch_indices = array
+        // patch_indices is always Primitive after execution (enforced by require_child! in
+        // Patched::execute), and PrimitiveArray::slice returns a PrimitiveArray directly.
+        let patch_indices: PrimitiveArray = array
             .patch_indices()
             .slice(range.clone())?
```
a10y, line 32: i don't think this downcast is generally safe to do? there's nothing that enforces the patch_indices child is primitive right?

I think the thing we do is change this to do a search_sorted over the `array.patch_indices().slice(..)` for `chunk_index` ?

0ax1:
As discussed offline with @joseph-isaacs, we for now want to keep the API as is.

joseph-isaacs:
Sorry to close this. We in the future highly likely want to use the execution context thorough out execution. I think for now we should accept this perf loss. 

We can re-examine this in the future.

PR #7418, opened 2026-04-13
fix[ci]: bench no fail-fast random access and compress bench

joseph-isaacs:
## Summary

Closes: #000


connortsui20 (approved):
do we want to do this for all of our benchmarks?

connortsui20:
or what about pr benchmarks?

PR #7431, opened 2026-04-14
feat[gpu]: dyn dispatch patches infrastructure

0ax1:
Adds patches infra for dyn dispatch, without applying them to bitpacking and alp yet. As part of that, the types used in dyn dispatch are trimmed down in size. This is to not reduce warp occupancy on the SMs due to allocating more registers on the GPU.

### DIFF
--- a/vortex-cuda/kernels/src/dynamic_dispatch.cu
+++ b/vortex-cuda/kernels/src/dynamic_dispatch.cu
@@ -195,9 +195,7 @@ template <typename T>
 __device__ inline GPUPatches unpack_source_patches(uint64_t patches_ptr,
                                                     uint32_t stage_len,
                                                     uint32_t element_offset) {
-    if (patches_ptr == 0) {
-        return { nullptr, nullptr, nullptr };
-    }
+    static_assert((sizeof(T) & (sizeof(T) - 1)) == 0, "sizeof(T) must be a power of two");
     uint8_t *base = reinterpret_cast<uint8_t *>(patches_ptr);
     constexpr uint32_t FL_CHUNK = 1024;
     constexpr uint32_t N_LANES = (sizeof(T) < 8) ? 32 : 16;
--- a/vortex-cuda/kernels/src/dynamic_dispatch.cu
+++ b/vortex-cuda/kernels/src/dynamic_dispatch.cu
@@ -192,9 +192,8 @@ __device__ inline void bitunpack(const T *__restrict packed,
 /// The packed layout is: [lane_offsets (uint32 × lo_count)] [indices (uint16 × num_patches)]
 /// [values (T × num_patches, aligned)].  Returns null pointers when patches_ptr == 0.
 template <typename T>
-__device__ inline GPUPatches unpack_source_patches(uint64_t patches_ptr,
-                                                    uint32_t stage_len,
-                                                    uint32_t element_offset) {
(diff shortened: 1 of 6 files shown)


vortex-cuda/kernels/src/dynamic_dispatch.h
```diff
@@ -166,17 +166,17 @@ union ScalarParams {
     /// `output_ptype` (on the enclosing ScalarOp) to determine the values'
     /// element type.
     struct DictParams {
-        uint32_t values_smem_byte_offset; // byte offset to decoded dict values in smem
+        uint16_t values_smem_byte_offset; // byte offset to decoded dict values in smem
```
0ax1, line 169: Current max shared memory with guard is 48KB.

vortex-cuda/kernels/src/dynamic_dispatch.h
```diff
@@ -188,11 +188,11 @@ struct ScalarOp {
...
 /// pool so that stages with different element widths can coexist.
 struct PackedStage {
-    uint64_t input_ptr;        // global memory pointer to this stage's encoded input
-    uint32_t smem_byte_offset; // byte offset within dynamic shared memory for output
-    uint32_t len;              // number of elements this stage produces
-
+    uint64_t input_ptr;   // global memory pointer to this stage's encoded input
+    uint64_t patches_ptr; // device ptr to packed source patches (0 = none)
     struct SourceOp source;
+    uint32_t len;              // number of elements this stage produces
```
0ax1, line 194: Happy to expand this again to u64, but for now let's assume and guard we stay within u32. If we go past u32, we should also try micro/marco-benchmarking in that range.

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -620,28 +641,15 @@ impl FusedPlan {
...
-        // Same for codes.
-        let mut pipeline = if codes_elem_bytes != self.output_elem_bytes {
-            self.walk_mixed_width_child(codes, pending_subtrees)?
-        } else {
-            self.walk(codes, pending_subtrees)?
-        };
+        let values_len: u32 = values
+            .len()
+            .try_into()
+            .map_err(|_| vortex_err!("Dict values length {} exceeds u32::MAX", values.len()))?;
```
0ax1, line 648: This doesn't fail decompression but leads to falling back to the CPU.

vortex-cuda/kernels/src/dynamic_dispatch.cu
```diff
@@ -183,6 +184,79 @@ __device__ inline void bitunpack(const T *__restrict packed,
...
+// ═══════════════════════════════════════════════════════════════════════════
+// Source-patch helpers
+// ═══════════════════════════════════════════════════════════════════════════
+
+/// Reconstruct a GPUPatches struct from a packed device pointer.
+/// The packed layout is: [lane_offsets (uint32 × lo_count)] [indices (uint16 × num_patches)]
+/// [values (T × num_patches, aligned)].  Returns null pointers when patches_ptr == 0.
+template <typename T>
+__device__ inline GPUPatches
+unpack_source_patches(uint64_t patches_ptr, uint32_t stage_len, uint32_t element_offset) {
```
a10y, line 196: i think this changes once we change patched layout

PR #7435, opened 2026-04-14
Remove metadata from `DynArray`

connortsui20:
## Summary

Apparently we just don't need this now? And we want to use `array_serialize` to get metadata instead.

This also fixes a few incorrect tests that we had.

## API Changes

Removes the `metadata` method on `DynArray`.

## Testing

Fixes some tests.

### DIFF
--- a/encodings/parquet-variant/src/vtable.rs
+++ b/encodings/parquet-variant/src/vtable.rs
@@ -244,13 +244,13 @@ mod tests {
     use vortex_array::IntoArray;
     use vortex_array::Precision;
     use vortex_array::arrays::VarBinViewArray;
-    use vortex_array::arrays::Variant;
     use vortex_array::arrays::VariantArray;
     use vortex_array::dtype::DType;
     use vortex_array::dtype::Nullability;
     use vortex_array::dtype::PType;
     use vortex_array::serde::SerializeOptions;
     use vortex_array::serde::SerializedArray;
+    use vortex_array::session::ArraySession;
     use vortex_array::session::ArraySessionExt;
     use vortex_array::validity::Validity;
     use vortex_buffer::BitBuffer;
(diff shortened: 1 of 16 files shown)


encodings/parquet-variant/src/vtable.rs
```diff
@@ -278,8 +281,6 @@ mod tests {
             concat.extend_from_slice(buf.as_ref());
         }
         let concat = concat.freeze();
-        session.arrays().register(ParquetVariant);
-        session.arrays().register(Variant);
```
joseph-isaacs, line 280: ?

gatesn (approved):

encodings/parquet-variant/src/vtable.rs
gatesn: ParquetVariant is registered above. 
Variant is the canonical array, it's now registrered in default ArraySession

vortex-python/src/io.rs
```diff
@@ -292,7 +292,7 @@ impl PyVortexWriteOptions {
     /// ```python
     /// >>> vx.io.VortexWriteOptions.compact().write(sprl, "tiny.vortex")
     /// >>> os.path.getsize('tiny.vortex')
-    /// 55120
+    /// 55152
```
connortsui20, line 295: @gatesn is this expected with only the changes in this PR?

vortex-array/src/serde.rs
```diff
@@ -701,101 +691,3 @@ impl TryFrom<BufferHandle> for SerializedArray {
...
-    use crate::array::ArrayId;
-    use crate::dtype::DType;
-    use crate::dtype::Nullability;
-    use crate::flatbuffers as fba;
-    use crate::session::ArraySession;
-
-    static SESSION: LazyLock<VortexSession> = LazyLock::new(VortexSession::empty);
-
-    #[test]
-    fn unknown_array_encoding_allow_unknown() {
```
connortsui20, line 725: this test didnt actually test anything, and was broken

vortex-python/src/io.rs
gatesn: Lol no

gatesn (changes requested):

gatesn:
Screenshot 2026-04-14 at 15 49 50

Size change is attributed to us always serializing the IDs of all encodings that exist in the registry. This PR adds dict + variant (which should have been there all along!!)

gatesn (approved):

vortex-python/src/io.rs
connortsui20: see https://github.com/vortex-data/vortex/pull/7435#issuecomment-4246701485

PR #7439, opened 2026-04-15
Remove more scalar_at from sparse arrays

gatesn:
| Benchmark case | Before (mean) | After (mean) | Speedup |
|---|---:|---:|---:|
| `extend_from_array_non_zctl_overlapping (1000, 8)` | `1.413 ms` | `80.71 us` | `17.5x` |
| `extend_from_array_non_zctl_overlapping (1000, 32)` | `2.586 ms` | `83.67 us` | `30.9x` |
| `extend_from_array_non_zctl_overlapping (10000, 8)` | `9.708 ms` | `560.3 us` | `17.3x` |
| `extend_from_array_zctl (1000, 8)` | `1.116 ms` | `14.31 us` | `78.0x` |
| `extend_from_array_zctl (1000, 64)` | `5.511 ms` | `38.68 us` | `142.5x` |
| `extend_from_array_zctl (10000, 8)` | `11.01 ms` | `108.4 us` | `101.6x` |

Before = commit `d72bf9b93` (pre-fix), After = commit `c4ad4e2ad` (the ListViewBuilder scalar-at removal fix).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10881,6 +10881,7 @@ dependencies = [
 name = "vortex-sparse"
 version = "0.1.0"
 dependencies = [
+ "codspeed-divan-compat",
  "itertools 0.14.0",
  "num-traits",
  "prost 0.14.3",
--- a/encodings/sparse/Cargo.toml
+++ b/encodings/sparse/Cargo.toml
@@ -27,6 +27,11 @@ vortex-mask = { workspace = true }
 vortex-session = { workspace = true }
 
 [dev-dependencies]
+divan = { workspace = true }
 itertools = { workspace = true }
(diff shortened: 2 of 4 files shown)


robert3005 (dismissed):

robert3005:
I pushed a commit to fix the build. Benchmarks need to be trimmed, even on my machine they take ~10ms which is way too slow for microbenchmark imho. The improvement here is more modest 35-50%

robert3005 (approved):

robert3005:
I have cut down the benchmarks by 20x. Let's see

joseph-isaacs (dismissed):
Benchmarks too long to run

PR #7441, opened 2026-04-15
chore[gpu]: don't inline bitunpack lane impls

0ax1:
Fully inlining the bitunpack kernels adds memory pressure such that the dynamic dispatch kernel previously operated at the 32 registers per thread max. Preventing to inline the bit unpack line impls drops the register count per thread to 24. This is relevant for future changes which will requires more registers such as patches support on the GPU.

The performance hit we take for bitunpack kernels here is 10%. It's probably worthwhile to investigate whether there might a tradeoff here to get similar perf with less aggressive inlining in the future. One thing we could also look at is trading in register spills via launch bounds for more occupancy.

Splitting out the lane implementations into headers is needed such that the dynamic dispatch kernel ptx can be compiled without the standalone bitunpack kernels which are not relevant for the dynamic dispatch kernel. This reduces the amount of assembly for the dynamic dispatch to 48k lines from 128k lines. Besides nvcc compile times, this is relevant for the dynamic dispatch kernel in terms of ptx to device compilation which should be as fast as possible.


0ax1:
command used to list hardware counters:
```
CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="sudo ncu \
  --kernel-name dynamic_dispatch_u32 \
  --launch-skip 2 \
  --launch-count 1 \
  --metrics \
launch__registers_per_thread,\
launch__occupancy_limit_registers,\
launch__shared_mem_per_block_dynamic,\
launch__occupancy_limit_shared_mem,\
sm__warps_active.avg.pct_of_peak_sustained_active,\
sm__warps_active.avg.per_cycle_active,\
sm__pipe_alu_cycles_active.avg.pct_of_peak_sustained_active,\
sm__inst_executed_pipe_lsu.avg.pct_of_peak_sustained_active,\
smsp__warps_issue_stalled_barrier.avg.pct_of_peak_sustained_active,\
smsp__warps_issue_stalled_long_scoreboard.avg.pct_of_peak_sustained_active" \
cargo bench -p vortex-cuda --bench dynamic_dispatch_cuda \
  -- dict_64vals_bp6bw_codes_for_bp6bw_values/dynamic_dispatch_u32/100M
```

robert3005 (approved):

PR #7446, opened 2026-04-15
Vector datasets catalog and downloader

connortsui20:
## Summary

We will want to add vector benchmarking soon (see https://github.com/vortex-data/vortex/pull/7399 for a draft).

This adds a simple catalog for the vector datasets hosted by `https://assets.zilliz.com/benchmark` for [VectorDBBench](https://github.com/zilliztech/vectordbbench), which both describes the shape of the datasets (are things partitioned, randomly shuffled, are there neighbors lists for top k, etc).

Also handles downloading everything.

I had to verify that all of this stuff was correct by looking at the S3 buckets themselves:

```sh
aws s3 ls s3://assets.zilliz.com/benchmark/ --region us-west-2 --no-sign-request
```

<details>

```sh
for d in bioasq_large_10m bioasq_medium_1m cohere_large_10m cohere_medium_1m \
         cohere_small_100k gist_medium_1m gist_small_100k glove_medium_1m \
         glove_small_100k laion_large_100m  \
         openai_large_5m openai_medium_500k openai_small_50k \
         sift_large_50m sift_medium_5m sift_small_500k; do
  echo "=== $d ==="
  aws s3 ls s3://assets.zilliz.com/benchmark/$d/ --region us-west-2 --no-sign-request
done
```

</details>

And this script from the main repo helped too: https://github.com/zilliztech/VectorDBBench/blob/main/vectordb_bench/backend/dataset.py

---

Things that are not implemented that I would like to add:

- Is the dataset pre-normalized for cosine similarity? This is not so obvious to me without actually working with the datasets, so I will do this later.
- Some datasets have scalar labels for all vectors that help mimic similarity + filter by some other column. Some of them also have neighbor lists for these specific filtered queries. So that is something we'll probably want to add in the future.

## Testing

N/A

### DIFF
--- a/vortex-bench/src/vector_dataset/download.rs
+++ b/vortex-bench/src/vector_dataset/download.rs
@@ -1,0 +1,225 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! URL builders and idempotent download driver for vector benchmark datasets.
+//!
+//! The upstream bucket is `https://assets.zilliz.com/benchmark/<prefix>/`. Within each
+//! prefix the train split is named according to a four-way convention:
+//!
+//! - `Single`: `train.parquet`
+//! - `SingleShuffled`: `shuffle_train.parquet`
+//! - `Partitioned`: `train-NN-of-MM.parquet`
+//! - `PartitionedShuffled`: `shuffle_train-NN-of-MM.parquet`
+//!
+//! `test.parquet` and (when present) `neighbors.parquet` live alongside the train files.
+
+use std::path::PathBuf;
+use std::time::Duration;
+
+use anyhow::Context;
+use anyhow::Result;
+use bytes::Bytes;
+use futures::StreamExt;
+use indicatif::ProgressBar;
+use indicatif::ProgressStyle;
+use reqwest::Client;
+use reqwest::IntoUrl;
+use tokio::fs::File;
+use tokio::io::AsyncWriteExt;
+use tokio::task::JoinSet;
+use tracing::info;
+use tracing::warn;
+
+use crate::datasets::data_downloads::download_data;
+use crate::utils::file::idempotent_async;
+use crate::vector_dataset::catalog::VectorDataset;
+use crate::vector_dataset::layout::LayoutSpec;
+use crate::vector_dataset::layout::TrainLayout;
+use crate::vector_dataset::paths;
+
+/// Bucket root for all VectorDBBench datasets we mirror against.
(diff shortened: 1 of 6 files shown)


connortsui20:
We probably want to mirror this somewhere. @AdamGS @robert3005 is there an easy way to do this?

AdamGS:
R2 is probably the easiest? Whatever we use for the clickbench data

vortex-bench/src/vector_dataset/download.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    Ok(DatasetPaths {
+        train_files: train_targets,
+        test,
+        neighbors,
+    })
+}
+
+/// Stream a large file to disk with a byte-progress bar.
+async fn download_with_progress(client: &Client, url: &str, output: &PathBuf) -> Result<()> {
```
AdamGS, line 164: Why do we need another one of these? Can't this be part of the general download utils we have here?

AdamGS: (this function and a bunch of the following ones)

vortex-bench/src/vector_dataset/download.rs
```diff
@@ -0,0 +1,289 @@
...
+        );
+    }
+
+    #[test]
+    fn shuffle_train_url() {
+        let urls = train_urls(VectorDataset::BioasqMedium1m, LayoutSpec::single_shuffled());
+        assert_eq!(
+            urls,
+            vec![
+                "https://assets.zilliz.com/benchmark/bioasq_medium_1m/shuffle_train.parquet"
```
AdamGS: we really don't want test that hit actual external data, if benchmarks run its good enough

connortsui20: note that these tests dont actually hit external endpoints, it just checks the string. But yeah Ill just remove them

vortex-bench/src/vector_dataset/download.rs
connortsui20: maybe? But afaict we dont use a `reqwest` client anywhere else

AdamGS: we definitely do
Screenshot 2026-04-15 at 18 30 53

AdamGS (approved):
overall, nothing objectionable here, lets ship it

connortsui20:
ok im yoloing this since this doesnt affect anyone else. At some point it would be good to unify the downloading but if we are going to do that then we might as well implement the catalog idea that @joseph-isaacs had.

AdamGS:
@joseph-isaacs what's the catalog idea? worth writing down somewhere?

PR #7456, opened 2026-04-15
small metal benchmark runners

joseph-isaacs:
## Summary

Closes: #000


scripts/bench-taskset.sh
```diff
@@ -18,12 +18,17 @@ fi
...
+    else
+        cpu_count="$(nproc)"
+        BENCH_CPUS="2-$((cpu_count - 1))"
+    fi
 fi
 
 if command -v numactl >/dev/null 2>&1; then
-    exec numactl --physcpubind="$BENCH_CPUS" --localalloc "$@"
+    exec numactl --physcpubind="$BENCH_CPUS" --membind=0 "$@"
 fi
```
AdamGS, line 32: reviewing the numactl [docs](https://man7.org/linux/man-pages/man8/numactl.8.html) again, I think we can do `--cpunodebind=0 --membind=0`

joseph-isaacs: mind adding this as a suggestion?

AdamGS: just realized its actually not the same, lets keep it as is

.github/workflows/nightly-bench.yml
```diff
@@ -48,4 +48,4 @@ jobs:
       matrix:
         machine_type:
           - id: x86
-            instance_name: c6id.8xlarge
```
0ax1, line 51: consider turning off freq scaling always running at max: 

```
# Pin CPU frequency governor to performance to prevent dynamic frequency scaling.
for gov in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do
  [[ -w "$gov" ]] && echo performance > "$gov" || true
done
```

scripts/bench-taskset.sh
```diff
@@ -18,12 +18,17 @@ fi
...
-    cpu_count="$(nproc)"
-    BENCH_CPUS="2-$((cpu_count - 1))"
+    if command -v numactl >/dev/null 2>&1; then
+        # All CPUs on NUMA node 0, skipping CPUs 0-1 to avoid OS interference
+        BENCH_CPUS=$(numactl --hardware | awk '/^node 0 cpus:/{sep=""; for(i=4;i<=NF;i++){if($i+0>1){printf "%s%s",sep,$i; sep=","}}}')
+    else
+        cpu_count="$(nproc)"
+        BENCH_CPUS="2-$((cpu_count - 1))"
+    fi
 fi
```
joseph-isaacs, line 29:
```suggestion

# Pin CPU frequency governor to performance to prevent dynamic frequency scaling.
for gov in /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor; do
  [[ -w "$gov" ]] && echo performance > "$gov" || true
done

```

Like this?

scripts/bench-taskset.sh
```diff
@@ -18,12 +18,17 @@ fi
 
 
 if [[ -z "${BENCH_CPUS:-}" ]]; then
-    cpu_count="$(nproc)"
-    BENCH_CPUS="2-$((cpu_count - 1))"
+    if command -v numactl >/dev/null 2>&1; then
```
AdamGS, line 21: worth setting [`TOKIO_WORKER_THREADS`](https://docs.rs/tokio/latest/tokio/#cpu-bound-tasks-and-blocking-code) here to make sure tokio only spawns one thread per CPU core we use

AdamGS (approved):
ship it, we can always follow up once we have a new baseline

PR #7458, opened 2026-04-15
Add new `vector-search-bench` benchmarking crate

connortsui20:
## Summary

Adds a new `vector-search-bench` crate.

Right now this is just utilities is preparation for adding proper vector search benchmarks (that are pulled from disk, not just in-memory). This just includes data downloading and file preparation, conversion (from parquet lists to vortex vector arrays), and some different compression and scan utilities.

## Testing

Some basic unit tests but the real stress test will come later when we actually benchmark stuff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10043,6 +10043,30 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ba73ea9cf16a25df0c8caa16c51acb937d5712a8429db78a3ee29d5dcacd3a65"
 
 [[package]]
+name = "vector-search-bench"
+version = "0.1.0"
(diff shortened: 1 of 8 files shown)


AdamGS (approved):

benchmarks/vector-search-bench/src/tests.rs
```diff
@@ -0,0 +1,156 @@
+// SPDX-License-Identifier: Apache-2.0
```
AdamGS: nit - move to `vector-search-bench/tests`?

connortsui20: decided to just delete the tests

PR #7462, opened 2026-04-16
Unify benchmark file download logic

robert3005:
Stop repeating same download logic all over codebase

There's still slight variation of this logic in statpopgen vcf download


0ax1 (approved):

0ax1:
Ideally, we could have a download api in vortex-utils. We also download stuff in vortex-duckdb build.rs etc. Can do in a follow up. As part of that make all downloads atomic etc.

PR #7463, opened 2026-04-16
Correct rust version channel

robert3005:
I had noticed when setting up machines for benchmarking that 1.90 isn't actually
a valid rust version and changing it to 1.90.0 was required to make it work


0ax1 (commented):
Should we change all 1.90 to 1.90.0? Do a global grep.

0ax1 (approved):

PR #7466, opened 2026-04-16
Use `vx-bench` to run benchmarks in CI

AdamGS:
## Summary

Use the same `vx-bench` tool we want to use locally to run the benchmarks in CI, removing bespoke bash argument parsing logic, generating data with it.


.github/workflows/sql-benchmarks.yml
```diff
@@ -87,21 +221,46 @@ on:
...
+              ],
               "scale_factor": "1"
-            },
+            }
           ]
 
 jobs:
   bench:
     timeout-minutes: 120
+    env:
```
AdamGS, line 261: @joseph-isaacs you previously enabled those only at query time but not when generating the data, what behavior is desired here?

AdamGS: if you look at the file sizes here, it does make a difference

bench-orchestrator/bench_orchestrator/runner/builder.py
```diff
@@ -29,21 +29,24 @@ def __init__(
...
-        results = {}
+        results: dict[ExecutionBackend, Path] = {}
 
-        # Build each binary
         env = os.environ.copy()
         env["RUSTFLAGS"] = self.config.rustflags
 
-        for engine in engines:
-            binary_name = engine.binary_name
+        for backend in backends:
```
joseph-isaacs, line 48: backend?

.github/workflows/sql-benchmarks.yml
joseph-isaacs: This should have been enabled when generating the file!

AdamGS: I figured, so keeping this change.

bench-orchestrator/bench_orchestrator/runner/builder.py
AdamGS: not a thing anymore

bench-orchestrator/README.md
```diff
@@ -253,10 +269,30 @@ vx-bench clean --older-than "30 days" --no-keep-labeled
...
+The SQL benchmark workflow uses explicit JSON target selection instead of parsing `engine:format`
+strings in shell:
+
+```bash
+uv run --project bench-orchestrator vx-bench prepare-data tpch \
+  --formats-json '["parquet","vortex","vortex-compact"]' \
+  --opt scale-factor=1.0
+
+uv run --project bench-orchestrator vx-bench run tpch \
+  --targets-json '[{"engine":"datafusion","format":"parquet"},{"engine":"duckdb","format":"vortex"}]' \
```
joseph-isaacs, line 290: should we put the opt in this json?

AdamGS: I was thinking about making this some sort of manifest format that we can just have different version of, WDYT?

joseph-isaacs: ye I think we should have this. happy to move out of this one

joseph-isaacs (approved):

AdamGS:
I feel like this is already pretty big, lets merge it and I'll do a follow-up today/tomorrow

bench-orchestrator/README.md
AdamGS: I feel like this is already pretty big, lets merge it and I'll do a follow-up today/tomorrow

PR #7468, opened 2026-04-16
Unify all runtime env var access to functions wrapping LazyLock

AdamGS:
## Summary

Env variable access isn't free, this is a fairly common pattern we already used in a couple of places, this PR just makes it consistent across the codebase.

## API Changes

Replaced the experimental patches from a public `LazyLock` to a function. I think this should've been a feature an not a runtime thing, but that's a different change.

I've also removed the `EnvVarGuard` from `vortex-utils`, it was only used in one place and I've just replaced it with `temp-env` that we already use elsewhere. 

## Testing

For the few tests that toggled env variables, I've changed them to only run on nextest, relying on it process isolation.


robert3005 (approved):

joseph-isaacs:
Shall we unify all the env vars into a single unstable one?

PR #7469, opened 2026-04-16
vortex-array: Attempt casting inner array to target for extensions

brancz:
## Summary

Getting
```
No CastKernel to cast canonical array vortex.ext from vortex.timestamp[ns, tz=UTC](i64) to i64
```
When running a datafusion query like:
```
SELECT cast(timestamp as bigint) as timestamp from table;
```

Where the timestamp array is a utc timestamp type column.

Since the inner array of a timestamp array is int64 already, all we really need to do is try to attempt to cast the inner array of the extension type to the target type. That's exactly what this PR does.

## Testing

Added a unit test.

### DIFF
--- a/vortex-array/src/arrays/extension/compute/cast.rs
+++ b/vortex-array/src/arrays/extension/compute/cast.rs
@@ -17,7 +17,9 @@ impl CastReduce for Extension {
         dtype: &DType,
     ) -> vortex_error::VortexResult<Option<ArrayRef>> {
         if !array.dtype().eq_ignore_nullability(dtype) {
-            return Ok(None);
+            // Target is not the same extension type.
+            // Delegate to the storage array's cast.
+            return Ok(Some(array.storage_array().cast(dtype.clone())?));
         }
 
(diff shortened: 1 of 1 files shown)


connortsui20 (approved):
whoops, I must have forgot this!

connortsui20:
So just a heads up that we do not know if this behavior will be supported in this exact way in the future as we have not fully figured out the exact semantics of extension type casting. Note that there will definitely be a way to do this cast, but we might enforce that you have to "define" that this cast is allowed somewhere (rather than say that we always allow casting directly from the storage array).

cc @joseph-isaacs

connortsui20:
So we talked offline and we actually do not want this behavior for extension types (we might want to enforce that you cannot cast to or from extension types). Instead, we want to enforce that conversions on extension types have to go through scalar functions like `timestamp_to_i64` where the semantics are well defined specifically for that type.

So we are going to revert this, but in order not to break you we will make sure we add functionality so that you can do this operation.

cc @joseph-isaacs @gatesn

PR #7480, opened 2026-04-16
Use cargo-zigbuild to link against old libc

robert3005:
We can use zig and link directly without having to spin up old oses


AdamGS (approved):
❤️

PR #7485, opened 2026-04-16
encodings/fsst: add OptimizedDecompressor for throughput

joseph-isaacs:
## Summary

- Adds `OptimizedDecompressor` in `encodings/fsst/src/decompressor.rs`, replacing the default `fsst-rs` decompressor in the canonicalization path
- Processes compressed codes in 32-code blocks (4×8-byte chunks) using SWAR escape detection — checks all 8 bytes of a `u64` for `0xFF` in a single branchless operation
- Pre-converts the symbol table to `u64` at construction time (~2.3KB total, fits in L1 cache), eliminating `Symbol::to_u64()` per lookup
- Re-enters the 32-code fast path after each escape rather than permanently dropping to scalar processing
- Runtime dispatch to a BMI1/BMI2/POPCNT-optimized codepath on x86-64

## Test plan

- [ ] Existing unit tests in `decompressor.rs` cover: basic decompression, escape codes, empty input, matches baseline on random strings, matches baseline on full-byte-range inputs with many escapes, large corpus correctness
- [ ] `cargo test -p vortex-fsst`
- [ ] `cargo bench -p vortex-fsst --bench fsst_compress -- decompress_fsst` to observe throughput improvement

### DIFF
--- a/encodings/fsst/src/canonical.rs
+++ b/encodings/fsst/src/canonical.rs
@@ -21,6 +21,7 @@ use vortex_error::VortexResult;
 
 use crate::FSST;
 use crate::FSSTArrayExt;
+use crate::decompressor::OptimizedDecompressor;
 
 pub(super) fn canonicalize_fsst(
     array: ArrayView<'_, FSST>,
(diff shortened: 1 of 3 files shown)


robert3005:
if this ends up being a perf win we should migrate this to fsst-rs imho

PR #7489, opened 2026-04-16
perf[gpu]: reduce register pressure in dyn dispatch

0ax1:
We decrease the number of values per tile in the output stage each GPU thread uses, as well as limit the register count to 32 in the launch bounds. This brings the dynamic dispatch kernel into a reasonably close range compared to the standalone kernel for now.

Type | Dynamic dispatch | Standalone | Ratio |
|---|---|---|---|
| u8 bw6 | 172 µs | 79 µs | 2.17× |
| u16 bw6 | 140 µs | 88 µs | 1.59× |
| u32 bw6 | 184 µs | 148 µs | 1.24× |
| u64 bw8 | 303 µs | 276 µs | 1.10×|


0ax1:
It's good to keep in mind, that dyn dispatch with only running bp is an odd scenario. We can spend more time on fine tuning bp in context of dyn dispatch but should prioritize optimizing end-to-end perf at this point.

robert3005 (approved):

a10y (approved):

PR #7490, opened 2026-04-16
Unify download management

connortsui20:
## Summary

Unifies the download management for benchmarks. Also makes the downloads smarter with AIMD and nicer progress bars.

## Testing

I just ran it in my terminal and it works well enough.

Let me know if we want a video for this and I can figure that out.

https://github.com/user-attachments/assets/85a7a0a6-3b88-4f35-bcbc-41dbdba37aff

### DIFF
--- a/vortex-bench/src/clickbench/benchmark.rs
+++ b/vortex-bench/src/clickbench/benchmark.rs
@@ -7,13 +7,13 @@ use std::path::Path;
 
 use anyhow::Result;
 use url::Url;
-use vortex::error::VortexExpect;
 
 use crate::Benchmark;
 use crate::BenchmarkDataset;
 use crate::IdempotentPath;
 use crate::TableSpec;
 use crate::clickbench::*;
+use crate::utils::file::resolve_data_url;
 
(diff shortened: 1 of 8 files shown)


robert3005 (approved):

robert3005:
I think you need to rebase for benchmarks to run since some of the logic changed

connortsui20:
did something go wrong here? This is saying that file sizes are changing too which is somewhat unexpected but maybe it is true if this is somehow wrong?

robert3005:
I think that while the compressor is stable if you change the order you feed files into the compressor the compression might change due to different numbers being drawn from the generator.

connortsui20:
This seems strange, why would the order in which I download things affect the order that the compressor compresses files? Downloading files should be an atomic operation, in that we download all of the files and then continue, so I don't know why this changes things.

I'll dig into this a bit more.

connortsui20:
ok I think that was a phantom from other changes on develop, everything seems fine now except https://github.com/vortex-data/vortex/pull/7490#issuecomment-4262289666?

PR #7491, opened 2026-04-16
DuckDB backend for benchmarks website

AdamGS:
## Summary

This PR started as a joke with @joseph-isaacs, but I actually think the result is quite nice (even considering the diff size).

This PR changes the benchmarks website backend from having a specialized ETL in JS, to processing the raw JSON into DuckDB tables, and using SQL to serve all endpoints.

### Tooling

Moving to SQL also makes it easier to explore data locally, and this change includes tools and documentation on how to get the data into a local DuckDB instance, including generating all required SQL statements and downloading the data.

### Docs

This PR includes new docs about the benchmarks website:
- README.md includes general setup, development and how to explore data locally.
- SCHEMA.md describes the various tables and views, and how the API uses them
- ETL.md attempts at describing the full lifecycle of the data, from being generated by the post-merge CI run, how various fields actually look and the processing steps it goes through before reaching the API.


joseph-isaacs:
Does this help with filter over graph points?

AdamGS:
This PR is mostly just switches the logic from JS to SQL with DuckDB, the thing I realized though is that I would like to make the server more generic in general and maybe reshape the data so creating new graphs/groups becomes easier (hopefully).

benchmarks-website/src/components/Header.jsx
```diff
@@ -25,8 +24,8 @@ export default function Header({
           </button>
           <a href="/" className="logo-link">
             <picture>
-                <source srcset="/vortex_black_nobg.svg" media="(prefers-color-scheme: light)"/>
-                <source srcset="/vortex_white_nobg.svg" media="(prefers-color-scheme: dark)"/>
+                <source srcSet="/vortex_black_nobg.svg" media="(prefers-color-scheme: light)"/>
+                <source srcSet="/vortex_white_nobg.svg" media="(prefers-color-scheme: dark)"/>
```
AdamGS, line 28: my browser was just angry at this

AdamGS:
I actually think this is a really good step, the follow ups are:
1. Normalize the raw data into more columns/fields.
2. Benchmarks report actual raw data, instead of rations or any processing.
3. Figure out a way to make the layouts and charts just config

robert3005:
You’re going to make yet another charting tool by converting charts to config. You’d think there’s something for it already?

AdamGS:
yeah I sort of hope to find something minimal here eventually, I think that's currently a stretch goal, getting nicer data and easier exploration seems much more important

benchmarks-website/src/components/Header.jsx
palaska: the white logo should have transparent background, I think claude might be able to fix the svg

AdamGS: I think it does? It was just angry that it was `srcset` and not `srcSet`

joseph-isaacs (approved):

PR #7493, opened 2026-04-16
perf: idempotent_dir for dataset generation

joseph-isaacs:
Have a `.success` file for dataset generate to speed up repeated benchmark runs

### DIFF
--- a/vortex-bench/src/bin/data-gen.rs
+++ b/vortex-bench/src/bin/data-gen.rs
@@ -11,7 +11,6 @@ use std::process::Command;
 
 use clap::Parser;
 use clap::value_parser;
-use tracing::info;
 use vortex::error::VortexExpect;
 use vortex_bench::Benchmark;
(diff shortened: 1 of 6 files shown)


connortsui20 (approved):

vortex-bench/src/bin/data-gen.rs
```diff
@@ -53,39 +54,50 @@ async fn main() -> anyhow::Result<()> {
...
-
-        if args
-            .formats
-            .iter()
-            .any(|f| matches!(f, Format::OnDiskVortex))
-        {
-            convert_parquet_directory_to_vortex(&base_path, CompactionStrategy::Default).await?;
-        }
+    // All conversions are only meaningful for local file URLs.
+    if benchmark.data_url().scheme() != "file" {
```
AdamGS, line 58: a bunch of this stuff is handled in the underlying benchmarks, if we want to have unified handling here, lets pull all of them out and document that on the trait,

PR #7498, opened 2026-04-16
feat: update the benchmark dataset and generator to add runner to the data id

joseph-isaacs:
e.g. `tpch/sf_1/q01/ec2_c6id.8xlarge/datafusion:vortex-file-compressed`

There is a local migration script.


AdamGS:
this requires rebasing, the script doesn't exist after #7466

AdamGS:
If benchmarks pass, lets merge it

AdamGS:
@joseph-isaacs this breaks the baseline

AdamGS (changes requested):
This shouldn't invalidate all the history, either migrate it somehow or handle it during processing

joseph-isaacs:
I should have mentioned that I would migrate the data manually

AdamGS (approved):
we can move the runner info into `QueryMeasurement`, just makes it easier to process later

AdamGS:
superseded by #7622

PR #7500, opened 2026-04-16
allow make view inline

joseph-isaacs:
..


robert3005 (changes requested):
I will need to see godbolt for this

PR #7502, opened 2026-04-16
chore: remove deprecated compute functions from encodings

joseph-isaacs:


AdamGS:
not sure what happened here, is this just #7497?

PR #7503, opened 2026-04-16
Separate positional from by name cast

### LINKED ISSUE #6900: `StructArray` field names should be unique
### Originally Discussed in https://github.com/vortex-data/vortex/discussions/6010

<sup>Originally posted by **connortsui20** January 16, 2026</sup>

# `StructArray` field names should be unique

We have this in our doc comment for `StructArray`:

https://github.com/vortex-data/vortex/blob/a30298b4aa8c227716a00f87a27fa3bcd0a0b127/vortex-array/src/arrays/struct_/array.rs#L71-L97

To me, this is insane behavior, and not at all expected.

I am guessing that we have this behavior because the Arrow specification does not enforce this at all.

## Similar issues

- [arrow-rs implementation]( https://docs.rs/arrow/latest/arrow/array/struct.StructArray.html#method.column_by_name)
- Apache (old?) issue: https://issues.apache.org/jira/browse/ARROW-11178
- Other Apache issue (failed to migrate?): https://github.com/apache/arrow/issues/27083

Also, looking at other Arrow issues like https://github.com/apache/arrow-rs/issues/4908, it doesn't seem like they have considered things like casting and schema evolution in the context of having duplicate field names.

Edit: Arrow-rs issue https://github.com/apache/arrow-rs/issues/9205

## Expected Behavior

I would expect field name uniqueness to be enforced.

I cannot think of any good reason to have duplicate names, and can think of many reasons **not** to have this behavior...

## Actual Behavior

Field names can be duplicated (and all duplicates can have **DIFFERENT TYPES**). When looking up fields with a given name, we (arbitrarily) get the first one.

## Reproduction Steps

It is quite easy to reproduce this, simply create a struct that has 2 or more fields with the same name and different (incompatible) types.

### Fuzzer Crashes

We have seen the fuzzer crash because of this a few times, here are some examples (closed in favor of this issue):
- https://github.com/vortex-data/vortex/issues/5865
- https://github.com/vortex-data/vortex/issues/6887
- https://github.com/vortex-data/vortex/issues/6929
- https://github.com/vortex-data/vortex/issues/6961

Technically this is not a bug but a feature, but this is so unexpected that I would consider this similar...

robert3005:
In order to handle full generality of casts we need to distinguish between
casting fields by name and by position. The positional casts are useful for non
sql systems where names are just metadata and the order of values matters.
However, in order to integrate with sql systesm we need by names casts as well.
It's not possible to handle them without any indication from the user which
behaviour they care about

fix #6900 

### DIFF
--- a/vortex-array/src/scalar_fn/fns/cast/mod.rs
+++ b/vortex-array/src/scalar_fn/fns/cast/mod.rs
@@ -3,6 +3,7 @@
 
 mod kernel;
 
+use std::fmt::Display;
 use std::fmt::Formatter;
 
 pub use kernel::*;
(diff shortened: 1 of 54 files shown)


vortex-proto/src/generated/vortex.expr.rs
```diff
@@ -163,3 +165,29 @@ pub struct CaseWhenOpts {
...
+    /// The values are not transformed in any way and thus are considered stable
+    /// (if the ProtoBuf definition does not change) and safe for programmatic use.
+    pub fn as_str_name(&self) -> &'static str {
+        match self {
+            Self::ByPosition => "BY_POSITION",
+            Self::ByName => "BY_NAME",
+        }
+    }
+    /// Creates an enum from field names used in the ProtoBuf definition.
+    pub fn from_str_name(value: &str) -> ::core::option::Option<Self> {
```
robert3005, line 186: remove this

vortex-array/src/scalar_fn/fns/cast/mod.rs
```diff
@@ -70,89 +193,103 @@ impl ScalarFnVTable for Cast {
         _metadata: &[u8],
         session: &VortexSession,
     ) -> VortexResult<Self::Options> {
-        let proto = pb::CastOpts::decode(_metadata)?.target;
-        DType::from_proto(
+        let proto = pb::CastOpts::decode(_metadata)?;
```
robert3005, line 196: remove leading underscore

robert3005:
Looks like we want to invert defaults here where the cast by default is name based from talking to @gatesn

robert3005:
Ok, I will add a cast registry

robert3005:
@AdamGS I have dug through datafusion issues. The top level one is https://github.com/apache/datafusion/issues/17330 where what Datafusion really wants is to remove `cast_column` and have `cast` with `cast_column` behaviour. So long term datafusion will have one cast. However, Arrow cast is not the datafusion cast given they're migrating away from it.

I will make a pluggable cast

robert3005:
I will have to redo this with deeper changes. Will close this and open a new pr #7588 is the first step we need for a real solution

PR #7505, opened 2026-04-16
Refresh PyVortex to be more Arrow focussed

gatesn:


vortex-array/src/expr/transform/coerce.rs
```diff
@@ -35,7 +36,17 @@ pub fn coerce_expression(expr: Expression, scope: &DType) -> VortexResult<Expres
...
+            .as_opt::<Binary>()
+            .is_some_and(|operator| operator.is_comparison())
+        {
+            match comparison_literal_coercions(&node, &child_dtypes) {
+                Some(coerced_dtypes) => coerced_dtypes,
+                None => node.scalar_fn().coerce_args(&child_dtypes)?,
+            }
+        } else {
+            node.scalar_fn().coerce_args(&child_dtypes)?
+        };
```
AdamGS, line 49: unrelated but can we use this in DF to deal with Decimal casting?

vortex-array/src/expr/plan.rs
```diff
@@ -0,0 +1,172 @@
+// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: these seem really simple, should they actually be part of the public API?

vortex-python/src/expr/mod.rs
```diff
@@ -406,6 +408,24 @@ pub fn cast(child: PyExpr, dtype: PyDType) -> PyResult<PyExpr> {
     })
 }
 
+/// Plan an expression against a Vortex dtype.
```
AdamGS, line 411: Worth explaining what "planning" is.

vortex-layout/src/scan/scan_builder.rs
```diff
@@ -205,7 +207,8 @@ impl<A: 'static + Send> ScanBuilder<A> {
 
     /// The [`DType`] returned by the scan, after applying the projection.
     pub fn dtype(&self) -> VortexResult<DType> {
-        self.projection.return_dtype(self.layout_reader.dtype())
+        plan_expression(self.projection.clone(), self.layout_reader.dtype())?
```
AdamGS, line 210: Ok after reading through all this PR, isn't this a significant behavior change? We used to talk about Vortex expressions as very strict and "physical", but now we implicitly cast and adapt them

robert3005: Hmmm, we do need some way of planning the query and executing expressions that don't necessarily match. Vortex expressions shouldn't do anything implicitly and instead we have this planning stage that inserts all the additional operations. However, there's still a limit on the autocoercions you'd want to do and we should limit them to only the nonfallible ones

docs/user-guide/pyarrow.md
```diff
@@ -29,25 +29,50 @@ Use {func}`~vortex.open` to lazily open a Vortex file:
...
+
+```{doctest} pycon
+>>> import pyarrow.compute as pc
+>>> table = f.to_table(columns=['tip_amount'], filter=pc.field('tip_amount') > 10)
+>>> table.num_rows > 0
+True
+```
+
+Use `filter_policy="pushdown"` to raise when a PyArrow expression cannot be pushed into Vortex. Use
+`filter_policy="fallback"` to read the rows and apply the PyArrow filter after the scan.
```
a10y, line 74: what is default?

docs/user-guide/pyarrow.md
```diff
@@ -29,25 +29,50 @@ Use {func}`~vortex.open` to lazily open a Vortex file:
...
+Vortex expressions are the stable pushdown API. PyVortex plans them against the file schema before
+the scan runs:
+
+```{doctest} pycon
+>>> table = f.to_table(columns=['tip_amount'], filter=vx.col('tip_amount') > 10)
+>>> table.num_rows > 0
+True
+```
+
+PyArrow compute expressions are accepted as compatibility input. PyVortex converts them through
```
a10y, line 63:
```suggestion
PyArrow compute functions are also accepted. PyVortex converts them through
```

vortex-array/src/expr/transform/coerce.rs
```diff
@@ -65,6 +76,31 @@ pub fn coerce_expression(expr: Expression, scope: &DType) -> VortexResult<Expres
...
+    let lhs_literal = node.child(0).is::<Literal>();
+    let rhs_literal = node.child(1).is::<Literal>();
+    if lhs_literal == rhs_literal {
+        return None;
+    }
+
+    let literal_idx = usize::from(rhs_literal);
+    let context_idx = usize::from(lhs_literal);
+    let literal_dtype = &child_dtypes[literal_idx];
+    let context_dtype = &child_dtypes[context_idx];
```
a10y, line 89: nit: cute, match would have been more readable

vortex-array/src/expr/plan.rs
gatesn: I think if we expose to Python + Java + C++ etc, then it would be nice to have them all in one place

vortex-layout/src/scan/scan_builder.rs
gatesn: Using Vortex directly from Python feels much more like the front-end of a query engine.

Honestly, I'm not sure where to draw the line.

If you use Vortex from inside a query engine, you have almost certainly already done physical planning. If you use Vortex directly, you probably haven't. It's a shame that Arrow Datasets also don't do physical planning, else we could use that as our "engine interface".

AdamGS: I just wanted to point out that this is a significant conceptual change which is bigger than what the title hints at.

One way to draw the line is to have a logical expression layer to handle this, which as far as I can tell is what `pyarrow` does, and is "bound" to a specific schema during execution (Like during `dataset.to_table(..)`)

PR #7506, opened 2026-04-17
L2 denorm scheme

connortsui20:
## Summary

We want to merge https://github.com/vortex-data/vortex/pull/7499 soon, but it is not a fair comparison because we do not normalize the uncompressed vectors flavor, but we do normalize the TurboQuant-encoded vectors. This just makes the comparison fair.

I've marked this as semi-unstable (there is no `with_turboquant` equivalent on the compressor builder, but the ID is clearly unstable) just so we can get the benchmark in #7499 over the line in a reasonable state.

## Testing

It already works for the TurboQuant scheme, this is just a reduction of that so we don't need to test this.


AdamGS (approved):
sure

PR #7507, opened 2026-04-17
Add extension constant pushdown rule and fix `InnerProduct` rule

connortsui20:
## Summary

Most of our optimizer rules interact with `ConstantArray`, as often there is an optimized implementation when a child is known to be constant.

We can have both of these arrays that mean the same thing, but the second is not easily matched on:

```
ConstantArray(
    scalar: Scalar::Extension(...)
)

ExtensionArray(
    storage_array: ConstantArray(
        scalar: Scalar::*
    )
)
```

This change adds a rule to convert the second representation into the first. It additionally fixes a bug in the `InnerProduct` reduction (right now it is in `execute` but it should really be a reduction rule, that will come in the future) where it checks for the second case instead of the first.

## Testing

Basic tests.

### DIFF
--- a/vortex-array/src/arrays/extension/compute/rules.rs
+++ b/vortex-array/src/arrays/extension/compute/rules.rs
@@ -6,24 +6,59 @@ use vortex_error::VortexResult;
 use crate::ArrayRef;
 use crate::IntoArray;
 use crate::array::ArrayView;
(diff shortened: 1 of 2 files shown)


AdamGS (approved):
This makes sense to me

PR #7509, opened 2026-04-17
chore: speed up CUDA kernel compilation

0ax1:
Previously, changing a .cu file triggered cargo to rerun vortex-cuda's build script, which recompiled all CUDA kernels and regenerated bindgen output, causing a full Rust recompilation. Besides skipping the Rust recompilation, we now do timestamp-based PTX compilation: Only recompile .cu files whose PTX is older than the source or any header.

### DIFF
--- a/vortex-cuda/build.rs
+++ b/vortex-cuda/build.rs
@@ -3,10 +3,11 @@
 
 #![expect(clippy::unwrap_used)]
 #![expect(clippy::expect_used)]
+#![expect(clippy::panic)]
 #![expect(clippy::use_debug)]
 
 use std::env;
-use std::fs::File;
+use std::fs;
 use std::io;
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


vortex-cuda/kernel-build/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+// SPDX-License-Identifier: Apache-2.0
```
AdamGS: why is this file needed?

0ax1: don't we need at least the lib.rs for a build.rs?

AdamGS: oh I get it, worth having a comment here?

vortex-cuda/kernel-build/build.rs
```diff
@@ -0,0 +1,226 @@
...
+        .expect("kernel-build must be inside vortex-cuda");
+
+    // Source directory for kernels (hand-written and generated .cu/.cuh files)
+    let kernels_src = vortex_cuda_dir.join("kernels/src");
+    // Output directory for compiled .ptx files - separate by profile.
+    let kernels_gen = vortex_cuda_dir.join("kernels/gen").join(&profile);
+
+    fs::create_dir_all(&kernels_gen).expect("Failed to create kernels/gen directory");
+
+    println!("cargo:rerun-if-env-changed=PROFILE");
```
AdamGS: is this actually necessery?

vortex-cuda/kernel-build/build.rs
```diff
@@ -0,0 +1,226 @@
...
+        Path::new(&manifest_dir)
+            .join("src/bit_unpack_gen.rs")
+            .display()
+    );
+    generate_unpack::<u8>(&kernels_src, 32).expect("Failed to generate unpack for u8");
+    generate_unpack::<u16>(&kernels_src, 32).expect("Failed to generate unpack for u16");
+    generate_unpack::<u32>(&kernels_src, 32).expect("Failed to generate unpack for u32");
+    generate_unpack::<u64>(&kernels_src, 16).expect("Failed to generate unpack for u64");
+
+    if !is_cuda_available() {
```
AdamGS: can this be moved higher up?

vortex-cuda/kernel-build/build.rs
0ax1: yes bc the profile translates into diff ptx outputs. debug translates into debug ptx.

vortex-cuda/kernel-build/build.rs
0ax1: hmm i guess? though then you can't have a peek at the generated files on macos which we might want still.

0ax1: generate unpack is cuda source code gen.

AdamGS (approved):

PR #7510, opened 2026-04-17
Use dev profile for editable python builds

AdamGS:
## Summary

Has this ever happened to you? You just want to make some minor change, or run some small `uv` command, and suddenly you're re-building all of Vortex in release mode and your laptop sounds like a jet taking off.

Now you don't have to do that! Any editable build (which is the default thing `uv` does) build will be in `dev` profile, while things like `maturin develop -r` or our publishing will stay in release mode.


robert3005 (approved):

vortex-python/test/conftest.py
```diff
@@ -2,19 +2,5 @@
...
 
 import logging
-import os
-import pathlib
-import subprocess
 
 logging.basicConfig(level=logging.DEBUG)
-
-
-def pytest_sessionstart():
```
AdamGS, line 12: I've also removed this test plugin that I think tried to avoid expensie rebuilds when running tests, now even a clean build will just build everything with a `dev` profile so this is unnecessery

PR #7511, opened 2026-04-17
Revert "chore: speed up CUDA kernel compilation (#7509)"

0ax1:
Needs to be reverted as CUDA changes are not properly picked up by build.rs Turns out there's no way to skip Rust build time and linkage, even if only the CUDA files changed that need to be re-generated.


AdamGS (approved):

0ax1:
I'm really just trying to get my commit count up.

PR #7516, opened 2026-04-17
fix issue with FileSystem::glob handling of non-existent files

### LINKED ISSUE #6599: FileSystem::glob with non-glob should verify file exists
### What happened?

Currently if there are no glob characters in the path we don't bother running the list files endpoint. We probably need some way to verify that the file actually exists, either with a HEAD or allow listing to "list" a file.

### Steps to reproduce

FileSystem::glob an empty file

### Environment

Develop

### Additional context

_No response_

a10y:
## Summary

Closes: #6599 

When `FileSystem::glob()` is called without any glob patterns, it returns a file stream with either a single element or no elements, depending on if the file exists in the FS. 

## API Changes

None

## Testing

3 unit tests:

1. globbing single file
2. Globbing non-existent file
3. Globbing file where multiple have same prefix only returns stream for exact-match

### DIFF
--- a/vortex-io/src/filesystem/glob.rs
+++ b/vortex-io/src/filesystem/glob.rs
@@ -23,13 +23,21 @@ impl dyn FileSystem + '_ {
         validate_glob(pattern)?;
 
         // If there are no glob characters, the pattern is an exact file path.
-        // Return it directly without listing the filesystem.
+        // Use [`list`](FileSystem::list) with the full path as the prefix so that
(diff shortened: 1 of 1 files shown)


gatesn (approved):

a10y:
There's something weird here, since `ObjectStore::list("x")` doesn't return `"x"` if `"x"` is a file path. This probably needs to use head_object instead, and plumb that through FileSystem

PR #7518, opened 2026-04-17
fix potential UB in ByteBool encoding

a10y:
## Summary

Our existing ByteBool constructor accepts arbitrary bytes as input, and will attempt to transmute it to a `&[bool]`. If those bytes are not `0x00` or `0x01` that can trigger UB on release builds.

## API Changes

We eliminate the `ByteBool::as_slice() -> &[bool]` method, replacing it with an accessor to the truthy byte values.

We don't really need to access this as a bool-slice anyway since we can do all operations on the bytes and wrap them back up as a ByteBool.

We also eliminate some old From constructors for ByteBool that I think we probably don't want lying around

## Tests

No additional tests needed

### DIFF
--- a/encodings/bytebool/public-api.lock
+++ b/encodings/bytebool/public-api.lock
@@ -10,10 +10,6 @@ pub fn vortex_bytebool::ByteBool::from_vec<V: core::convert::Into<vortex_array::
 
 pub fn vortex_bytebool::ByteBool::new(buffer: vortex_array::buffer::BufferHandle, validity: vortex_array::validity::Validity) -> vortex_bytebool::ByteBoolArray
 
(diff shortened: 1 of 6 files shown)


AdamGS (approved):

a10y:
Hmm, actually this begs the question of, do we want to match Arrow 8-bit Bool representation, or do we want to force it to be a Rust `bool`?

https://arrow.apache.org/docs/format/CanonicalExtensions.html#bit-boolean

a10y:
I think this should probably use 8-bit bool instead (i.e. truthy bytes = true, 0x00 = false).

Let me fix this up.

PR #7519, opened 2026-04-17
Implement validity for Between scalar function

robert3005:
Avoid using IsNotNull(expr) fallback which can lead to infinite recursion. As a followup we should make validity required thing for scalar functions


robert3005:
apparently this was not enough :(

joseph-isaacs (approved):

PR #7520, opened 2026-04-17
Revert "Add extension constant pushdown rule and fix `InnerProduct` rule"

joseph-isaacs:
Reverts vortex-data/vortex#7507


connortsui20:
we only want to revert the rule, not the other changes

PR #7522, opened 2026-04-17
Fix reduction/execute cycle

connortsui20:
## Summary

Fixes the regression caused by https://github.com/vortex-data/vortex/pull/7507

The reduction rule would go from `Extension(Constant(..))` into `Constant(Extension(..))`, but the `execute` would just revert that, creating a cycle!

## Testing

N/A

### DIFF
--- a/vortex-array/src/arrays/constant/vtable/canonical.rs
+++ b/vortex-array/src/arrays/constant/vtable/canonical.rs
@@ -10,6 +10,7 @@ use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 
 use crate::Canonical;
+use crate::ExecutionCtx;
 use crate::IntoArray;
 use crate::array::ArrayView;
 use crate::arrays::BoolArray;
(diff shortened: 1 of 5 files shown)


vortex-array/src/arrays/constant/vtable/canonical.rs
```diff
@@ -163,7 +167,16 @@ pub(crate) fn constant_canonicalize(array: ArrayView<'_, Constant>) -> VortexRes
             let s = scalar.as_extension();
 
             let storage_scalar = s.to_storage_scalar();
-            let storage_self = ConstantArray::new(storage_scalar, array.len()).into_array();
+
+            // NB: We need to execute the constant array to be canonical because there is a
+            // reduction rule that turns `Extension(Constant(..))` into `Constant(Extension(..))`,
+            // and if we don't do this we create an infinite cycle.
+            // See `ExtensionConstantRule` for more details.
```
a10y, line 174: should we just remove the reduction rule?

connortsui20: It's kind of hard to say since if we don't remove that reduction rule, all of our optimizations that look for a `ConstantArray` child now don't work.

connortsui20:
i dont really believe the TPC-H SF1 regression is real given SF10 is fine

vortex-array/src/arrays/constant/vtable/canonical.rs
joseph-isaacs: We should keep the rule 100%. We must have a standard way to check is constant.

joseph-isaacs (approved):
We should add a cycle detection mode to executor for debug

vortex-array/src/arrays/constant/vtable/canonical.rs
a10y: yea makes sense

PR #7563, opened 2026-04-20
feat: GPU dyn dispatch patches support

0ax1:
Integrates the structural plumbing as well as applying the patches in source and scalar ops within the GPU dynamic dispatch kernel. None of the CUDA dynamic dispatch benchmarks regressed by adding patches support.

As part of this change, the fastlanes lane count for a given type is now determined at compile time via `FL_LANES<type>`.


vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -69,10 +72,10 @@ fn is_dyn_dispatch_compatible(array: &ArrayRef) -> bool {
     let id = array.encoding_id();
     if id == ALP.id() {
         let arr = array.as_::<ALP>();
-        return arr.patches().is_none() && arr.dtype().as_ptype() == PType::F32;
+        return arr.dtype().as_ptype() == PType::F32;
```
a10y, line 75: oh no f64?

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -119,9 +122,17 @@ fn is_dyn_dispatch_compatible(array: &ArrayRef) -> bool {
...
+///
+/// Patches are tied to their owning ops, mirroring the CUDA side where
+/// `patches_ptr` lives on `BitunpackParams` / `AlpParams`:
+/// - `source_patches` for the source op (BitPacked exceptions)
+/// - Each scalar op carries its own `Option<Patches>` (ALP exceptions)
 struct Stage {
     source: SourceOp,
-    scalar_ops: Vec<ScalarOp>,
+    /// Patches from the source op (e.g. BitPacked overflow exceptions).
+    source_patches: Option<Patches>,
```
a10y, line 133: do you think this would change a lot if we changed PatchedArray to use the Patches layout and then moved the interior patches to wrapping with PatchedArray?

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
0ax1: Oh this was me being deliberately over-restrictive when initially mapping out the plan builder logic. Good catch. I'll wire in f64 in a follow up.

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
0ax1: Hmm, I don't think the physical plan, consisting of one or more `Stage`s, would really change. That'd be more a question of how we walk the array tree when we construct/extract the different ops to send to the GPU.

a10y (approved):

PR #7567, opened 2026-04-20
Report sum stat for numerical columns to DF

AdamGS:
## Summary

Report the `sum` stat we have up to DF, which improved a lot of its stats overhead now AND includes the clickbench specialized planning that all other engines have.


robert3005:
Last I checked the sum fix wasn’t released yet

PR #7569, opened 2026-04-20
avoid including experimental encoding in footer

a10y:
## Summary

We have been publishing a footer reference to `vortex.patched` experimental encoding in the last few releases.

The issue is that we should not be including the Patched encoding in the default array context

This removes Patched from the default array context, instead registering it conditionally where we register the other unstable encodings.

I've verified that after this change writing a file with default settings no longer includes Patched encoding

image


robert3005 (approved):

a10y:
I deleted the `patched.vortex` compat files for now and will re-enable them once we stabilize

PR #7580, opened 2026-04-20
fix: RepeatedScan::execute is CPU-work not I/O

danking:
I also renamed `Handle::spawn_blocking` to `Handle::spawn_blocking_io` which better reflects what it does.

### DIFF
--- a/vortex-cuda/src/pooled_read_at.rs
+++ b/vortex-cuda/src/pooled_read_at.rs
@@ -106,7 +106,7 @@ impl VortexReadAt for PooledFileReadAt {
         async move {
             let mut target = pool.get(length)?;
             let target = handle
-                .spawn_blocking(move || {
+                .spawn_blocking_io(move || {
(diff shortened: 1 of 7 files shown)


robert3005 (approved):
This is effectively the same, you're just going through a different interface

danking:
In Spiral, it is not, we use a Rayon thread pool for CPU-heavy work.

danking:
but honestly the overhead of actually spawning this (very tiny, very quick) work seems not worth it. It makes my example lose about 2/5 of its throughput.

a10y:
It can be pretty expensive depending on how complex your projection expression and layout tree are

danking:
@robert3005 the benchmarks wouldn't show a change unless y'all used my CPUSegregatedExecutor which puts `spawn_cpu` on a Rayon thread pool.

robert3005:
I was curious whether spawn_blocking vs spawn made a difference

PR #7583, opened 2026-04-21
Make `StatsCache` and `ArrayAndStats` have interior mutability

connortsui20:
## Summary

It was unfortunate that `ArrayAndStats` had to be taken by `&mut`, so this change makes updates use interior mutability.

## API Changes

`Scheme` takes `ArrayAndStats` by shared reference instead of by exclusive reference.

## Testing

N/A since there isn't any semantic change here, just API changes.

### DIFF
--- a/vortex-btrblocks/public-api.lock
+++ b/vortex-btrblocks/public-api.lock
@@ -56,9 +56,9 @@ impl core::marker::StructuralPartialEq for vortex_btrblocks::schemes::decimal::D
 
 impl vortex_compressor::scheme::Scheme for vortex_btrblocks::schemes::decimal::DecimalScheme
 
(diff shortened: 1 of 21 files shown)


robert3005 (approved):

a10y:
Wait, why is this better? isn't it preferable to not have extra unnecessary locking?

connortsui20:
@a10y The fastpath on the mutex will basically never lock because the compressor only runs single-threaded, the only reason is to keep this `Sync`.

a10y:
I don't think this affects the Sendness of the compressor schemes?

connortsui20:
Sorry, to keep the array and stats sync

PR #7585, opened 2026-04-21
Implement smarter sampler

connortsui20:
## Summary

Right now, we have a very naive sampling strategy in the compressor. We will run the estimation function in the order of the `Scheme` declaration, regardless of how expensive it is to estimate things.

This changes the sampling algorithm to first do all of the cheap estimations, and only when all of them fail do we do the more expensive sampling / callbacks.

Edit: so benchmarks show no real change, but theoretically if someone comes along and has a really slow compress scheme, this should lessen the impact of that.

## Testing

Will run benchmarks.

### DIFF
--- a/vortex-btrblocks/src/schemes/integer.rs
+++ b/vortex-btrblocks/src/schemes/integer.rs
@@ -17,6 +17,7 @@ use vortex_compressor::builtins::FloatDictScheme;
 use vortex_compressor::builtins::StringDictScheme;
 use vortex_compressor::estimate::CompressionEstimate;
 use vortex_compressor::estimate::DeferredEstimate;
+use vortex_compressor::estimate::EstimateScore;
 use vortex_compressor::estimate::EstimateVerdict;
 use vortex_compressor::scheme::AncestorExclusion;
 use vortex_compressor::scheme::ChildSelection;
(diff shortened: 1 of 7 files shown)


vortex-compressor/src/ctx.rs
```diff
@@ -30,6 +30,7 @@ pub struct CompressorContext {
     /// Merged stats options from all eligible schemes at this compression site.
     merged_stats_options: GenerateStatsOptions,
 
+    // TODO(connor): Replace this with an `im::Vector`
```
robert3005, line 33: huh?

vortex-btrblocks/src/schemes/integer.rs
```diff
@@ -737,20 +738,29 @@ impl Scheme for SequenceScheme {
...
-                // TODO(connor): This doesn't really make sense?
-                // Since two values are required to store base and multiplier the compression ratio is
-                // divided by 2.
-                Ok(EstimateVerdict::Ratio(encoded.len() as f64 / 2.0))
+                // TODO(connor): We should pass this array back to the compressor in the case that
+                // we do want to sequence encode this so that we do not need to recompress.
+                if sequence_encode(data.array_as_primitive(), exec_ctx)?.is_none() {
+                    return Ok(EstimateVerdict::Skip);
+                }
+                // TODO(connor): Should we get the actual ratio here?
```
robert3005, line 762: we don't if you do the todo above that avoids creating the array

vortex-compressor/src/ctx.rs
connortsui20: `im::Vector` is a copy on write vector in a 3rd party library

robert3005 (approved):

PR #7588, opened 2026-04-22
Add ability to override function behaviour via registry in VortexSession

robert3005:
This logic isn't used yet but will be used to allow us to customise behaviour of
functions depending on an integration point, i.e. Datafusion can have it's
casting logic that is different from arrow casting logic while everyone using
vortex can still continue calling `cast` and not specialize for the engine
they're using

Thing to consider is whether we want require passing session to optimise or whether we should remove the implicit optimise calls and defer them to execute loop

The next pr will replace struct casting logic with Arrow and DF specific
behaviour.


vortex-array/src/optimizer/mod.rs
```diff
@@ -6,34 +6,89 @@
...
     fn optimize_recursive(&self) -> VortexResult<ArrayRef>;
 }
 
 impl ArrayOptimizer for ArrayRef {
     fn optimize(&self) -> VortexResult<ArrayRef> {
-        Ok(try_optimize(self)?.unwrap_or_else(|| self.clone()))
+        Ok(try_optimize(self, None)?.unwrap_or_else(|| self.clone()))
+    }
+
+    fn optimize_ctx(&self, ctx: &ExecutionCtx) -> VortexResult<ArrayRef> {
```
robert3005: this should be OptimiserSession?

vortex-session/src/registry.rs
```diff
@@ -287,3 +288,86 @@ impl<T: Clone> Context<T> {
...
+/// Each entry stores an `Arc<dyn Any + Send + Sync>` wrapping a caller-supplied concrete type `F`
+/// (typically a function pointer). Callers recover the original type by passing the same `F` to
+/// [`FnRegistry::find`]; `find` returns `Some(Arc<F>)` on a type match and `None` otherwise.
+///
+/// Used for pluggable dispatch keyed by an `(outer, inner)` identifier pair — for example the
+/// optimizer's parent-reduce registry keys by `(parent_encoding_id, child_encoding_id)` so that
+/// downstream crates can override the rule that would normally run from the child encoding's
+/// static `PARENT_RULES` set.
+#[derive(Clone, Debug, Default)]
+pub struct FnRegistry(Arc<DashMap<(Id, Id), Arc<dyn Any + Send + Sync>>>);
```
robert3005: we want a Fn here

joseph-isaacs (commented):
let's run benchmarks.

vortex-session/src/registry.rs
```diff
@@ -287,3 +288,86 @@ impl<T: Clone> Context<T> {
...
+    /// Register a function under `(outer, inner)`, replacing any existing entry.
+    pub fn register<F: Any + Send + Sync>(&self, outer: Id, inner: Id, f: F) {
+        self.0.insert((outer, inner), Arc::new(f));
+    }
+
+    /// Look up a function registered under `(outer, inner)`, downcasting to `F`.
+    ///
+    /// Returns `None` if no function is registered, or if the registered value is not of type `F`.
+    pub fn find<F: Any + Send + Sync>(&self, outer: Id, inner: Id) -> Option<Arc<F>> {
+        let entry = self.0.get(&(outer, inner))?;
```
joseph-isaacs: This might be expensive in a hot loop

vortex-array/src/executor.rs
```diff
@@ -106,6 +106,9 @@ impl ArrayRef {
     /// For safety, we will error when the number of execution iterations reaches a configurable
     /// maximum (default 128, override with `VORTEX_MAX_ITERATIONS`).
     pub fn execute_until<M: Matcher>(self, ctx: &mut ExecutionCtx) -> VortexResult<ArrayRef> {
+        // Clone the session up-front so each `optimize_ctx` call below borrows from this owned
+        // value rather than re-borrowing through `&mut ctx`.
+        let session = ctx.session().clone();
```
joseph-isaacs: can we not clone?

vortex-array/src/optimizer/session.rs
```diff
@@ -0,0 +1,43 @@
...
+
+//! Session state for pluggable parent-reduce rules.
+//!
+//! [`OptimizerSession`] wraps an [`FnRegistry`] keyed by `(parent_encoding_id, child_encoding_id)`
+//! and is consulted by the optimizer during execution, before the child encoding's static
+//! `PARENT_RULES` are tried. Entries are typed as [`ReduceParentFn`](super::ReduceParentFn).
+//!
+//! The registry is empty by default. Downstream crates register `ReduceParentFn` values to add
+//! new parent-reduce rules or override ones that the child encoding would otherwise run from its
+//! static `PARENT_RULES`.
```
joseph-isaacs: in the end we will have reduce/execute, reduce/execute_parnet, how will we organize all of these?

joseph-isaacs: I think we will add all of these here?

vortex-session/src/registry.rs
robert3005: Maybe not, since we can have other functions here

joseph-isaacs: We cannot use a DashMap or (Id, Id). We need a ArcSwap or something better!

vortex-array/src/optimizer/session.rs
robert3005: Yes, I think we just have a registry of function pointers here. I don't know if we have conflicts

vortex-array/src/optimizer/session.rs
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Session state for pluggable parent-reduce rules.
```
joseph-isaacs: Can we call this ArrayKernels of something desc reduce/execute & reduce/execute_parnet

vortex-session/src/registry.rs
joseph-isaacs: We might even want to avoid the Arc<T> and leek it?

vortex-array/src/optimizer/session.rs
joseph-isaacs: what do you mean you think they might?

robert3005: oh they all have ~same~ different signature

vortex-array/src/optimizer/kernels.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /// Look up the [`ReduceParentFn`] registered for `(outer, child)`.
+    ///
+    /// Returns an owned [`Arc`] so the session-variable borrow can be dropped before invoking the
+    /// function.
+    pub fn find_reduce_parent(&self, outer: Id, child: Id) -> Option<Arc<ReduceParentFn>> {
+        self.registry
+            .find(self.hash_fn_ids(outer, child, FnKind::ReduceParent))
+    }
```
joseph-isaacs: There can be more than one?

vortex-array/src/optimizer/kernels.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Return `true` if a [`ReduceParentFn`] is registered for `(outer, child)`.
+    pub fn contains_reduce_parent(&self, outer: Id, child: Id) -> bool {
+        self.registry
+            .contains(self.hash_fn_ids(outer, child, FnKind::ReduceParent))
+    }
```
joseph-isaacs: I think we want one registry per find?

vortex-session/src/registry.rs
```diff
@@ -287,3 +291,108 @@ impl<T: Clone> Context<T> {
...
+/// committing this registry to a specific schema. For example, `ArrayKernels` hashes
+/// `(parent_encoding_id, child_encoding_id, FnKind)` so one registry can hold several kinds of
+/// pluggable kernel keyed by encoding pairs without collisions.
+///
+/// This type is intended for read-heavy registries that are populated during session setup and
+/// then shared. Reads load a snapshot through [`ArcSwap`] and do not block on writers. Writes
+/// clone the current map and swap in a new snapshot, so externally serialize concurrent
+/// [`FnRegistry::register`] calls if every write must be retained.
+#[derive(Debug, Default)]
+pub struct FnRegistry(ArcSwap<HashMap<u64, Arc<dyn Any + Send + Sync>>>);
```
joseph-isaacs: Shall we store one per fn type to avoid the downcast to the known type up extraction?

joseph-isaacs:
I really think we want one lookup for either of the kernels types execute/reduce?

vortex-session/src/registry.rs
robert3005: yes

robert3005:
@joseph-isaacs I think I applied the changes you have been thinking about

vortex-array/src/optimizer/kernels.rs
```diff
@@ -0,0 +1,113 @@
...
+use vortex_error::VortexResult;
+use vortex_session::Ref;
+use vortex_session::SessionExt;
+use vortex_session::registry::Id;
+use vortex_utils::aliases::DefaultHashBuilder;
+use vortex_utils::aliases::hash_map::HashMap;
+
+use crate::ArrayRef;
+
+/// Shared hasher used to combine `(outer, child, FnKind)` tuples into [`FnRegistry`] keys.
```
joseph-isaacs, line 34: no kind?

vortex-array/src/optimizer/kernels.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// Implementations must preserve the parent's logical length and dtype, matching the invariant
+/// required of static parent-reduce rules.
+pub type ReduceParentFn =
+    fn(child: &ArrayRef, parent: &ArrayRef, child_idx: usize) -> VortexResult<Option<ArrayRef>>;
+
+/// Session-scoped registry of optimizer kernel functions.
+#[derive(Debug, Default)]
+pub struct ArrayKernels {
+    reduce_parent: ArcSwap<HashMap<u64, Arc<[ReduceParentFn]>>>,
```
joseph-isaacs, line 51: I would new type this, but not a blocker

vortex-array/src/optimizer/mod.rs
```diff
@@ -6,34 +6,83 @@
...
 pub mod rules;
 
 /// Extension trait for optimizing array trees using reduce/reduce_parent rules.
 pub trait ArrayOptimizer {
-    /// Optimize the root array node only by running reduce and reduce_parent rules to fixpoint.
+    /// Optimize the root array node by running reduce and reduce_parent rules to fixpoint.
+    ///
+    /// This uses only static rules registered on encoding vtables. Use [`Self::optimize_ctx`]
+    /// when session-registered [`ArrayKernels`] should participate.
     fn optimize(&self) -> VortexResult<ArrayRef>;
```
joseph-isaacs, line 38: This should be deprecated?

joseph-isaacs (approved):

PR #7589, opened 2026-04-22
Prunning expressions can reference rowcount 

### LINKED ISSUE #7187: Add RowCount stat to zone maps for more precise pruning
Zone maps currently lack a RowCount stat. This limits certain pruning strategies
that need to compare a count against the total number of rows in a zone.

For example, to falsify `is_not_null(x)` the ideal condition is:

    null_count == row_count => is_not_null is all false

Without RowCount, we have to fall back to `is_constant && null_count > 0`, which
only catches the subset of cases where the zone is constant.

Adding RowCount would enable more precise pruning for is_not_null and potentially
other expressions that need to reason about total row counts.

robert3005:
This lets us effectively prune expressions like IsNotNull

fix #7187


vortex-layout/src/layouts/zoned/zone_map.rs
```diff
@@ -160,18 +170,77 @@ impl ZoneMap {
...
+        if num_zones == 1 || last_zone_len == zone_len {
+            let ends = PrimitiveArray::from_iter([num_zones as u64]).into_array();
+            let values = PrimitiveArray::from_iter([last_zone_len]).into_array();
+            return Ok(RunEnd::try_new(ends, values, ctx)?.into_array());
+        }
+
+        let ends = PrimitiveArray::from_iter([num_zones as u64 - 1, num_zones as u64]).into_array();
+        let values = PrimitiveArray::from_iter([zone_len, last_zone_len]).into_array();
+        Ok(RunEnd::try_new(ends, values, ctx)?.into_array())
+    }
```
robert3005: this can be simplified

vortex-layout/src/layouts/zoned/zone_map.rs
```diff
@@ -160,18 +170,77 @@ impl ZoneMap {
...
+        names.push(FieldName::from(ZONE_ROW_COUNT_FIELD));
+        fields.push(row_count);
+
+        StructArray::try_new(
+            FieldNames::from_iter(names),
+            fields,
+            num_zones,
+            self.array.struct_validity(),
+        )
+    }
```
robert3005: This should just be `Merge`

vortex-array/src/expr/pruning/pruning_expr.rs
```diff
@@ -14,11 +14,17 @@ use crate::dtype::FieldPath;
 use crate::dtype::FieldPathSet;
 use crate::expr::Expression;
 use crate::expr::StatsCatalog;
+use crate::expr::col;
 use crate::expr::get_item;
 use crate::expr::root;
 use crate::expr::stats::Stat;
 
 pub type RequiredStats = Relation<FieldPath, Stat>;
+pub const ROW_COUNT_FIELD: &str = "__vortex_row_count";
```
joseph-isaacs: we cannot do this

robert3005: What do you suggest?

robert3005:
I have tried making it an expression. The problem you run into is that this expression isn't constant so you can't rewrite it into literal but also that expression cannot be inferred from the existing arrays properties

robert3005:
Ok, this is good to go now, a lot more code than I'd wish but at least it's pure array rewrite

vortex-array/src/expr/exprs.rs
```diff
@@ -701,3 +705,21 @@ pub fn dynamic(
...
+pub fn stats_expression(agg: AggregateFnRef) -> Expression {
+    StatsExpression.new_expr(agg, [])
+}
+
+/// Creates a [`StatsExpression`] wrapping the [`RowCount`] aggregate.
+///
+/// This is the canonical way to refer to the row count of the current evaluation scope inside
+/// a pruning predicate.
+pub fn row_count() -> Expression {
+    stats_expression(RowCount.bind(crate::aggregate_fn::EmptyOptions))
```
robert3005:
```suggestion
    stats_expression(RowCount.bind(EmptyOptions))
```

vortex-layout/src/layouts/zoned/zone_map.rs
joseph-isaacs, line 42: its odd a zonemap doesn't know its row len?

robert3005: I think the layout knows. We can plumb that through that's a good point

robert3005:
After talking to @joseph-isaacs I have trimmed it down to just a RowCount ScalarFn. We can add generic stats aggregate expressions later

vortex-array/src/scalar_fn/internal/row_count.rs
```diff
@@ -0,0 +1,161 @@
...
+use crate::scalar_fn::EmptyOptions;
+use crate::scalar_fn::ExecutionArgs;
+use crate::scalar_fn::ScalarFnId;
+use crate::scalar_fn::ScalarFnVTable;
+
+/// Zero-argument placeholder for the row count of the current evaluation scope.
+///
+/// `RowCount` is emitted while building pruning predicates that need a
+/// scope-level value which is not stored as a regular stats column, such as the
+/// row count of the current file or zone. The layer that owns that scope must
```
joseph-isaacs, line 32: Explain how unstable/internal this is

joseph-isaacs (approved):

PR #7591, opened 2026-04-22
Use layout file splits when DF re-partitions individual files

AdamGS:
## Summary

Instead of just splitting files arbitrarily, align it with split layouts to make better use of Vortex's internal pruning and other behaviors.

### DIFF
--- a/vortex-datafusion/src/persistent/opener.rs
+++ b/vortex-datafusion/src/persistent/opener.rs
@@ -10,7 +10,6 @@ use datafusion_common::DataFusionError;
 use datafusion_common::Result as DFResult;
 use datafusion_common::ScalarValue;
 use datafusion_common::exec_datafusion_err;
-use datafusion_datasource::FileRange;
 use datafusion_datasource::PartitionedFile;
 use datafusion_datasource::TableSchema;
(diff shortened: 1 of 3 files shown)


AdamGS:
@joseph-isaacs do you expect the full benchmarks to behave differently than the SQL ones?

vortex-datafusion/src/persistent/opener.rs
```diff
@@ -421,33 +441,88 @@ impl FileOpener for VortexOpener {
...
+    let row_range = 0..row_count;
+    let mut split_points: Vec<_> = SplitBy::Layout
+        .splits(layout_reader, &row_range, &[FieldMask::All])
+        .map_err(|e| exec_datafusion_err!("Failed to compute Vortex natural splits: {e}"))?
+        .into_iter()
+        .collect();
+
+    if split_points.first().copied() != Some(0) {
+        split_points.insert(0, 0);
+    }
```
a10y: i thought this was always the case

a10y (approved):

vortex-datafusion/src/persistent/opener.rs
AdamGS: good catch

a10y (approved):

PR #7592, opened 2026-04-22
Route Arrow - DType extension conversion through the registry

palaska:
- Arrow <-> DType extension conversion now consults the DTypeSession extension registry.
  - Any registered extension round-trips through Arrow by default, not just temporal types and Parquet Variant.
  - Extensions can register an Arrow canonical alias; FixedShapeTensor uses this to emit `arrow.fixed_shape_tensor` with JSON metadata, so pyarrow sees it as a first-class FixedShapeTensorArray.


vortex-array/src/dtype/arrow.rs
```diff
@@ -23,26 +23,58 @@ use arrow_schema::Schema;
...
 use crate::extension::datetime::TimeUnit;
 use crate::extension::datetime::Timestamp;
 
+const ARROW_EXT_NAME_VARIANT: &str = "arrow.parquet.variant";
+
+/// `(vortex_id, arrow_canonical_name)` pairs — single source of truth for bijection between
+/// Vortex-internal extension ids and Arrow canonical extension names. Canonical extensions
+/// serialize metadata as raw UTF-8 (typically JSON) rather than base64-wrapped bytes.
+const CANONICAL_ALIASES: &[(&str, &str)] =
+    &[("vortex.fixed_shape_tensor", "arrow.fixed_shape_tensor")];
```
connortsui20: Whoops I forgot to change `vortex.fixed_shape_tensor` to `vortex.tensor.fixed_shape_tensor`... Can you add that to this PR? I could do it myself but then these won't be synced. Maybe the str ID should just be a constant and both call sites use that?

AdamGS:
ok I have thoughts but I'm going to head out soon and this required more thinking, I'll try and post a review first thing tomorrow @palaska

vortex-array/src/dtype/arrow.rs
AdamGS: Just putting what we talked about over slack here - I think we should avoid hard-coding extensions here, this whole part of the code shouldn't know about them and while we can have some best-practices/defaults or assumed semantics for some of these, it should be done externally and allow for customization by the user.

vortex-array/src/dtype/arrow.rs
```diff
@@ -55,6 +87,15 @@ pub trait TryFromArrowType<T>: Sized {
     fn try_from_arrow(value: T) -> VortexResult<Self>;
 }
 
+/// Conversion from Arrow types to Vortex types using a session's extension dtype registry to
+/// resolve `ARROW:extension:name` metadata into [`DType::Extension`] values.
+///
+/// Unregistered or malformed extension metadata falls back to the storage dtype.
+pub trait FromArrowWithSession<T>: Sized {
```
AdamGS: Should this be a function with a default impl on `TryFromArrowType`?

vortex-array/src/dtype/session.rs
```diff
@@ -22,15 +24,21 @@ pub type ExtDTypeRegistry = Registry<ExtDTypePluginRef>;
 #[derive(Debug)]
 pub struct DTypeSession {
     registry: ExtDTypeRegistry,
+    arrow_canonical: RwLock<ArrowCanonicalAliases>,
```
joseph-isaacs: We cannot use a RwLock, it has very high contention. It rarely written to?

vortex-array/src/dtype/session.rs
```diff
@@ -22,15 +24,21 @@ pub type ExtDTypeRegistry = Registry<ExtDTypePluginRef>;
...
+}
+
+#[derive(Debug, Default)]
+struct ArrowCanonicalAliases {
+    entries: Vec<(ExtId, &'static str)>,
```
joseph-isaacs: What is &'static str shall we give it a type alias?

vortex-array/src/extension/tests/mod.rs
```diff
@@ -3,4 +3,4 @@
 
 //! Test extension types for exercising the [`ExtVTable`] contract.
 
-mod divisible_int;
+pub(crate) mod divisible_int;
```
joseph-isaacs, line 6: ?

vortex-tensor/src/types/fixed_shape/mod.rs
```diff
@@ -3,6 +3,12 @@
 
 //! Fixed-shape Tensor extension type.
 
+/// Vortex extension id for [`FixedShapeTensor`].
+pub(crate) const ID: &str = "vortex.tensor.fixed_shape_tensor";
```
joseph-isaacs: Why did you expose this?

vortex-tensor/src/types/fixed_shape/proto.rs
joseph-isaacs, line 1: Why remove this in this PR

vortex-tensor/src/lib.rs
```diff
@@ -47,8 +43,11 @@ pub const SCALAR_FN_ARRAY_TENSOR_PLUGIN_ENV: &str = "VX_SCALAR_FN_ARRAY_TENSOR_P
 
 /// Initialize the Vortex tensor library with a Vortex session.
 pub fn initialize(session: &VortexSession) {
-    session.dtypes().register(Vector);
-    session.dtypes().register(FixedShapeTensor);
+    let dtypes = session.dtypes();
+    dtypes.register(Vector);
+    dtypes.register(FixedShapeTensor);
+    dtypes.register_arrow_canonical(ExtId::new(fixed_shape::ID), fixed_shape::ARROW_EXT_NAME);
```
joseph-isaacs: don't do this ExtId::new(fixed_shape::ID)

vortex-tensor/src/types/vector/vtable.rs
```diff
@@ -19,7 +21,7 @@ impl ExtVTable for Vector {
     type NativeValue<'a> = &'a ScalarValue;
 
     fn id(&self) -> ExtId {
-        ExtId::new("vortex.tensor.vector")
+        ExtId::new(ID)
```
joseph-isaacs:
```suggestion
        static ID: CachedId = CachedId::new(...);
        *ID
```

vortex-tensor/src/types/fixed_shape/mod.rs
palaska: re-exposed via `pub(crate)` so initialize can register the canonical alias from one source

vortex-array/src/dtype/session.rs
palaska: Claude recommended `ArcSwap<HashMap>`, lock-free reads and atomic write swaps

vortex-tensor/src/types/fixed_shape/proto.rs
palaska: arrow requires metadata to be UTF-8 json so i swapped, do we want to keep the proto around?

vortex-array/src/extension/tests/mod.rs
palaska: for the newly added extension_roundtrip tests, i needed a test extension type with metadata. What is your comment here? 😄

vortex-tensor/src/types/fixed_shape/proto.rs
joseph-isaacs: @connortsui20 is this okay?

joseph-isaacs: But in general JSON is less space efficient than proto

vortex-array/src/dtype/session.rs
joseph-isaacs: Yes, I think wrap it in a newtype just like https://github.com/vortex-data/vortex/pull/7588

vortex-tensor/src/types/fixed_shape/mod.rs
joseph-isaacs: use `fn id(&self) -> ExtId` you can extract the str from there

vortex-array/src/extension/tests/mod.rs
joseph-isaacs: It was do we need to do this

vortex-array/src/dtype/arrow.rs
```diff
@@ -210,21 +243,123 @@ impl FromArrowType<(&DataType, Nullability)> for DType {
...
+    }
+}
+
+/// Canonical extensions store UTF-8 bytes directly; non-canonical extensions base64-encode so
+/// arbitrary binary plugin output survives the String-typed metadata channel.
+fn decode_extension_metadata(field: &Field, is_canonical: bool) -> VortexResult<Cow<'_, [u8]>> {
+    match field.extension_type_metadata() {
+        None | Some("") => Ok(Cow::Borrowed(&[])),
+        Some(s) if is_canonical => Ok(Cow::Borrowed(s.as_bytes())),
+        Some(s) => BASE64_STANDARD
```
palaska: @joseph-isaacs should I get rid of this and assert valid utf 8 on extension metadata serde? Then this codepath will be simpler (String::from_utf8) and we can make `serialize_metadata` return `VortexResult<String>` instead of bytes.

joseph-isaacs: Do we store BASE_64 json in the vortex file for this type? Surely that is a waste of space

palaska: this is only the arrow serde path. Metadata for the fixed shape tensor is utf8 encoded json bytes, no base64. I guess its still larger than proto but it's metadata at the end of the day, your call :)

vortex-tensor/src/types/fixed_shape/proto.rs
connortsui20: I think it's fine so that we are compat with arrow

palaska:
Closing this one and filed https://github.com/vortex-data/vortex/pull/7726 instead which has a smaller scope

PR #7595, opened 2026-04-22
Bump msrv and rust-toolchain to 1.91 (and lance for benchmarks)

AdamGS:
We're two major version back on our lance comparison (which is about 2 months of changes), but it also seems like more and more dependencies break because we're on 1.90.

libs.rs says 68% of crates.io requests ask for this stable version or newer, not sure what was our bar in the past but I figured its worth at least checking it out.

I've reviewed the changes and none of them seem substantial, they are all tests and some lints that were improved and don't warn anymore.

### DIFF
--- a/encodings/fastlanes/src/bitpacking/array/bitpack_compress.rs
+++ b/encodings/fastlanes/src/bitpacking/array/bitpack_compress.rs
@@ -540,13 +540,15 @@ mod test {
             .for_each(|&idx| values[idx] = patch_value);
 
         let array = PrimitiveArray::from_iter(values);
-        let bitpacked = bitpack_encode(&array, 4, None, &mut ctx).unwrap();
+        let bitpacked = bitpack_encode(&array, 4, None, &mut ctx)?;
 
-        let patches = bitpacked.patches().unwrap();
+        let patches = bitpacked
+            .patches()
+            .ok_or_else(|| vortex_error::vortex_err!("expected patches"))?;
         let chunk_offsets = patches
             .chunk_offsets()
             .as_ref()
-            .unwrap()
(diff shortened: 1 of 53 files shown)


robert3005:
We have agreed to last 3 major versions just like datafusion. We can upgrade if there’s a reason

AdamGS:
@robert3005 given that 1.95 was released last week, should we bump to 1.92 while we're doing this?

robert3005:
While we can bump to 1.92 I wouldn't bump beyond what we have a need for

a10y:
I think we've roughly been following DF to date and trying to keep a window of stable versions, but you're right we haven't bumped it in awhile. This is 4 releases behind latest stable so seems safe to me

EDIT: I see rob's comment to the same effect now buried in the middle of a bunch of benchmark result comments

AdamGS:
So lets keep it at this? @a10y @robert3005

encodings/fastlanes/src/bitpacking/array/bitpack_compress.rs
```diff
@@ -540,13 +540,15 @@ mod test {
...
 
-        let patches = bitpacked.patches().unwrap();
+        let patches = bitpacked
+            .patches()
+            .ok_or_else(|| vortex_error::vortex_err!("expected patches"))?;
         let chunk_offsets = patches
             .chunk_offsets()
             .as_ref()
-            .unwrap()
+            .ok_or_else(|| vortex_error::vortex_err!("expected chunk offsets"))?
```
robert3005: can you import all the vortex_err invocations

AdamGS: done

robert3005 (approved):

a10y (approved):

PR #7596, opened 2026-04-22
Extend DType::least_supertype to cover FixedSizeList, List, and tensor extensions

palaska:
Widening Vector<f32, 768> and Vector<f64, 768> now returns Vector<f64, 768> instead of None.

### DIFF
--- a/vortex-array/src/dtype/coercion.rs
+++ b/vortex-array/src/dtype/coercion.rs
@@ -79,13 +79,27 @@ impl DType {
     /// The core primitive — what type can hold both `self` and `other`?
     /// Returns `None` if no common supertype exists.
     pub fn least_supertype(&self, other: &DType) -> Option<DType> {
-        // 1. Same type modulo outer nullability: union outer and return. Nested types
-        //    whose only difference is inner nullability fall through to their own arms.
-        if self.with_nullability(other.nullability()) == *other {
-            return Some(self.with_nullability(self.nullability() | other.nullability()));
+        let union_null = self.nullability() | other.nullability();
+
+        if let (
+            DType::FixedSizeList(lhs_elem, lhs_size, _),
+            DType::FixedSizeList(rhs_elem, rhs_size, _),
+        ) = (self, other)
+            && lhs_size == rhs_size
+        {
+            let elem = lhs_elem.least_supertype(rhs_elem)?;
+            return Some(DType::FixedSizeList(Arc::new(elem), *lhs_size, union_null));
         }
 
-        let union_null = self.nullability() | other.nullability();
+        if let (DType::List(lhs_elem, _), DType::List(rhs_elem, _)) = (self, other) {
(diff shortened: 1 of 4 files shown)


vortex-array/src/dtype/coercion.rs
connortsui20: So now that we do the nested type widening, this is wrong because the "ignore nullability" part doesn't recurse.

So these tests will fail (allows the coercion when it shouldn't):

```rust
    #[test]
    fn list_can_coerce_from_rejects_nullable_source_elements() {
        let target = DType::List(
            Arc::new(DType::Primitive(PType::F64, NonNullable)),
            NonNullable,
        );
        let source = DType::List(
            Arc::new(DType::Primitive(PType::F32, Nullable)),
            NonNullable,
        );

        assert!(!target.can_coerce_from(&source));
    }

    #[test]
    fn fsl_can_coerce_from_rejects_nullable_source_elements() {
        let target = DType::FixedSizeList(
            Arc::new(DType::Primitive(PType::F64, NonNullable)),
            4,
            NonNullable,
        );
        let source = DType::FixedSizeList(
            Arc::new(DType::Primitive(PType::F32, Nullable)),
            4,
            NonNullable,
        );

        assert!(!target.can_coerce_from(&source));
    }

```

vortex-array/src/dtype/coercion.rs
```diff
@@ -194,12 +214,24 @@ impl DType {
...
-        // 7. Extension: delegate to vtable
+        // 7. FixedSizeList / List widening: defer to least_supertype + identity check
+        if matches!(
+            (self, other),
+            (DType::FixedSizeList(..), DType::FixedSizeList(..))
+                | (DType::List(..), DType::List(..))
+        ) {
+            return other
+                .least_supertype(self)
+                .is_some_and(|st| st.eq_ignore_nullability(self))
```
connortsui20: This erases inner nullability

vortex-array/src/dtype/coercion.rs
connortsui20: The fix should be to add this _before_ the `eq_ignore_nullability` check:

```rust
        // 1. FixedSizeList / List widening: recurse on element dtype and preserve both outer and
        //    inner nullability constraints.
        if let (
            DType::FixedSizeList(target_elem, target_size, _),
            DType::FixedSizeList(source_elem, source_size, _),
        ) = (self, other)
        {
            return target_size == source_size
                && (self.is_nullable() || !other.is_nullable())
                && target_elem.can_coerce_from(source_elem);
        }
        if let (DType::List(target_elem, _), DType::List(source_elem, _)) = (self, other) {
            return (self.is_nullable() || !other.is_nullable())
                && target_elem.can_coerce_from(source_elem);
        }
```

(note the recurse into `can_coerce_from`)

connortsui20 (changes requested):

connortsui20 (approved):
Nit: Ordering of numbers in the comments is off, but otherwise LGTM!

PR #7599, opened 2026-04-22
Faster BoolArray::min_max via true_count instead of set_slices

### LINKED ISSUE #7598: Bool Min/Max is slow
We can probably just use `popcount` to implement this

robert3005:
There's no need to compute valid slices, we can just use true_count

fix #7598 


AdamGS (approved):
love it

PR #7603, opened 2026-04-23
feat(cuda): fuse narrower-than-output Dict codes and RunEnd ends

0ax1:
Dict codes and RunEnd ends that are narrower than the output type (e.g. u8 BitPacked codes in a u32 Dict) previously required a separate kernel launch. They are now fused by decoding at the source's native width and widening to T in shared memory.

### DIFF
--- a/vortex-cuda/kernels/src/dynamic_dispatch.cu
+++ b/vortex-cuda/kernels/src/dynamic_dispatch.cu
@@ -39,9 +39,16 @@
 //
 // ## Mixed-width support
 //
-// LOAD sources from pending subtrees may have a narrower type than the
-// output (e.g. u8 dict codes in a u32 plan). load_element() widens
-// to T via static_cast — no separate widen kernel or smem intermediate.
+// Dict codes, RunEnd ends, and other child arrays may have a narrower
+// element type than the output T. Two mechanisms handle this:
+//
+// LOAD       load_element() dispatches on the per-stage PTypeTag to
+//            read at the source's native width and static_cast to T.
+// BITUNPACK  bitunpack_typed() unpacks at the source's native width,
+//            then widens to T in-place via a backward scan
+//            (widen_inplace). The smem region is pre-allocated at
+//            max(source_width, T) bytes per element by the Rust plan
+//            builder, so the widen never overflows.
 
 #include <assert.h>
 #include <cuda.h>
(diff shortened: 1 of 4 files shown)


vortex-cuda/kernels/src/dynamic_dispatch.cu
```diff
@@ -203,6 +210,22 @@ scatter_patches_chunk(const GPUPatches &patches, T *__restrict out, uint32_t chu
 // Source ops
 // ═══════════════════════════════════════════════════════════════════════════
 
+/// Widen U-sized elements in shared memory to T-sized, in-place.
+/// Backward scan ensures no unread element is overwritten since
```
a10y, line 214: clever

vortex-cuda/kernels/src/dynamic_dispatch.cu
```diff
@@ -203,6 +210,22 @@ scatter_patches_chunk(const GPUPatches &patches, T *__restrict out, uint32_t chu
...
+/// sizeof(T) >= sizeof(U) guarantees the write at index i touches
+/// only bytes beyond those of src[i].
+template <typename T, typename U>
```
a10y, line 217: nit: src/dst or narrow/wide would be a bit clearer but makes sense

vortex-cuda/kernels/src/dynamic_dispatch.cu
```diff
@@ -236,6 +259,68 @@ __device__ inline void bitunpack(const T *__restrict packed,
...
+    const uint32_t n_chunks = (chunk_len + dst_off + FL_CHUNK - 1) / FL_CHUNK;
+    const uint32_t total_elems = n_chunks * FL_CHUNK;
+
+    // Narrow source: unpack at native width, then widen to T.
+    switch (source_ptype) {
+    case PTYPE_U8:
+    case PTYPE_I8: {
+        auto *narrow = reinterpret_cast<uint8_t *>(dst);
+        bitunpack<uint8_t>(reinterpret_cast<const uint8_t *>(packed), narrow, chunk_start, chunk_len, src);
+        __syncthreads();
```
a10y, line 298: does this need to be syncthreads or can it be syncwarp?

a10y (approved):

a10y:
@0ax1 I actually think the widen in place might not be sound? You're going to have two warps operating concurrently per-block (since block size is 64), both sweeping from the end of their range. I think if the warps diverge by a lot can't they clobber each other? Seems like it'd be hard for them to diverge that much in practice but technically possible right?

0ax1:
> @0ax1 I actually think the widen in place might not be sound? You're going to have two warps operating concurrently per-block (since block size is 64), both sweeping from the end of their range. I think if the warps diverge by a lot can't they clobber each other? Seems like it'd be hard for them to diverge that much in practice but technically possible right?

Oh damn, good catch. Yeah, I'll need to think of sth here.

connortsui20:
going to revert this because tests fail on CI

0ax1:
thx @connortsui20 !

vortex-cuda/kernels/src/dynamic_dispatch.cu
0ax1: I haven't seen perf regression for using `__syncthreads`, so I can take a conservative choice here.

vortex-cuda/kernels/src/dynamic_dispatch.cu
0ax1: Will add that in the follow up.

PR #7604, opened 2026-04-23
Allow configuring session-wide vortex options in DF

AdamGS:
## Summary

This PR shits the DF config to use the `Extension` API which seems like the right thing? It keeps all the other nice behaviors will allowing users to configure session wide things (either through SQL or directly).

### DIFF
--- a/vortex-datafusion/src/persistent/format.rs
+++ b/vortex-datafusion/src/persistent/format.rs
@@ -15,8 +15,9 @@ use datafusion_common::DataFusionError;
 use datafusion_common::GetExt;
 use datafusion_common::Result as DFResult;
 use datafusion_common::Statistics;
-use datafusion_common::config::ConfigField;
-use datafusion_common::config_namespace;
+use datafusion_common::config::ConfigExtension;
+use datafusion_common::config::ExtensionOptions;
+use datafusion_common::extensions_options;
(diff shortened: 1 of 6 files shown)


vortex-datafusion/src/persistent/format.rs
```diff
@@ -259,26 +271,35 @@ impl VortexFormatFactory {
...
 impl FileFormatFactory for VortexFormatFactory {
     #[expect(clippy::disallowed_types, reason = "required by trait signature")]
     fn create(
         &self,
-        _state: &dyn Session,
+        state: &dyn Session,
         format_options: &std::collections::HashMap<String, String>,
     ) -> DFResult<Arc<dyn FileFormat>> {
-        let mut opts = self.options.clone().unwrap_or_default();
+        let mut opts = state
```
a10y, line 296: shouldn't we delete the options field from the format factory if we always pull it out of the state now?

AdamGS: I don't think you do, you have two code paths here.
1. The one enabled by this PR - User sets a session-level config, which propagates down to new tables.
2. User executes a DDL statement like `CREATE EXTERNAL TABLE t ... OPTIONS (my_vortex_conf=7)`; which goes through the factory.

PR #7606, opened 2026-04-23
fix deadlock on unknown policy request

myrrc:
Querying allows_unknown + allow_unknown leads to a deadlock in FFI tests.


AdamGS (approved):
LGTM, we should probably consider doing this for more values? VortexSession is probably very easy to deadlock this way

PR #7608, opened 2026-04-23
add another oneshot tsan suppression

myrrc:
Another false positive causing errors in
https://github.com/vortex-data/vortex/actions/runs/24559595293/job/71804664847?pr=7509#step:10:215

Also add tsan flags to FFI test job.

### DIFF
--- a/vortex-ffi/test/scan.cpp
+++ b/vortex-ffi/test/scan.cpp
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
-#include <iostream>
 #include <mutex>
 #include <nanoarrow/common/inline_types.h>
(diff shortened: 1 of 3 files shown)


vortex-ffi/test/scan.cpp
```diff
@@ -520,10 +520,15 @@ TEST_CASE("Multithreaded scan", "[datasource]") {
             arrays[i] = array;
         });
     }
+
     for (auto &thread : threads) {
         thread.join();
     }
 
+    vx_partition* partition = vx_scan_next_partition(scan, &error);
```
0ax1: What is the exact error and why can it be ignored. Add a comment inline.

myrrc: There is no error and this is checked with require_no_error. partition is null as it's exhausted.

0ax1 (approved):

PR #7611, opened 2026-04-23
Potentially better FSST estimation

connortsui20:
## Summary

## API Changes

TODO maybe?

## Testing

Benchmarks run.


connortsui20:
so compress time definitely improved, but there are regressions in file size. I think I know why (chooses FSST instead of Dict), and I can fix that

connortsui20:
so it is quite hard to do this (or at least codex has not been able to find holistically good improvements, so I'm going to close this.

PR #7622, opened 2026-04-24
feat: add a runner field to all runners

joseph-isaacs:
Adds a runner field to all benchmark runner which could be later used to diff runs from different runners


vortex-bench/src/runner.rs
```diff
@@ -411,6 +420,18 @@ impl SqlBenchmarkRunner {
...
+    matches!(std::env::var("CI").as_deref(), Ok("true"))
+}
+
+fn validate_benchmark_runner_id(benchmark_runner: &str, is_ci: bool) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        !is_ci || benchmark_runner != DEFAULT_BENCHMARK_RUNNER_ID,
+        "benchmark runner must not be unknown in CI; pass --runner"
+    );
+    Ok(())
+}
```
AdamGS, line 433: does this condition and check work locally? it seems weird

AdamGS (approved):

PR #7625, opened 2026-04-24
Refinement Types

connortsui20:
## Summary

Tracking issue: https://github.com/vortex-data/vortex/issues/6872

Adds refinement types as extension types by adding an `is_refinement` method to the `ExtVTable`.

The main benefit of declaring an extension type as a refinement type is that many scalar functions can push down operations directly onto the storage array since the domain of values of the refinement type is a $\subseteq$ of the domain of the storage type.

This additionally adds a "reduction rule" (it's not actually a rule here, this is checked on construction and in expressions, I don't really know if this fits into our current reduction rule system?) that will "peel back" the refinement layers of extension types if the scalar function cannot handle / has no knowledge of the refinement type, but does know about the storage type.

I think we can also do this for non-scalar array parents? Not sure.

I've also left a TODO where it would be nice if there was a way to express that a refinement type is closed under some operation. For example, if we had a `PositiveInteger` refinement type over `Primitive`, we know that it is closed under addition (ignoring overflow), so `add: PositiveInteger x PositiveInteger -> PositiveInteger`, but with this change alone we can only get `add: PositiveInteger x PositiveInteger -> Primitive`. It is unclear if this is even possible to express.

## API Changes

Adds `is_refinement` to the `ExtVTable`.

## Testing

Some basic refinement type unit tests.

### DIFF
--- a/vortex-array/src/dtype/extension/vtable.rs
+++ b/vortex-array/src/dtype/extension/vtable.rs
@@ -28,6 +28,26 @@ pub trait ExtVTable: 'static + Sized + Send + Sync + Clone + Debug + Eq + Hash {
     /// Returns the ID for this extension type.
     fn id(&self) -> ExtId;
 
+    /// Returns `true` when this extension type logically refines its storage dtype.
+    ///
+    /// A refinement asserts that every value in this extension's logical domain is also a
+    /// value of the storage dtype's domain, with additional invariants enforced by this
+    /// vtable. The canonical examples are `NormalizedVector` (refines `Vector` by
+    /// requiring unit L2 norm) and a hypothetical `PositiveInt` (refines `Primitive(U64)`
+    /// by excluding zero).
+    ///
+    /// The default is `false`: plain extension types (e.g. `Uuid` over
+    /// `FixedSizeList<u8, 16>`) encode a distinct logical type in their storage that is
+    /// not substitutable with it.
+    ///
+    /// This flag is consumed by `ScalarFnRefinementPeelRule` (in
+    /// `arrays::scalar_fn::rules`) to decide whether to transparently peel a refinement
+    /// input and retry a scalar fn on the source type. See that rule for the full
+    /// semantics; the `TODO(connor)` there also tracks refinement-closure preservation.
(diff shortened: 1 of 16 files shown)


vortex-array/src/dtype/extension/vtable.rs
```diff
@@ -28,6 +28,26 @@ pub trait ExtVTable: 'static + Sized + Send + Sync + Clone + Debug + Eq + Hash {
...
+    /// `FixedSizeList<u8, 16>`) encode a distinct logical type in their storage that is
+    /// not substitutable with it.
+    ///
+    /// This flag is consumed by `ScalarFnRefinementPeelRule` (in
+    /// `arrays::scalar_fn::rules`) to decide whether to transparently peel a refinement
+    /// input and retry a scalar fn on the source type. See that rule for the full
+    /// semantics; the `TODO(connor)` there also tracks refinement-closure preservation.
+    fn is_refinement(&self) -> bool {
+        false
+    }
```
joseph-isaacs, line 49: lets make this an enum

General | Refinement

connortsui20: Not sure what that gives us? Also not even sure what this method would be called.

joseph-isaacs: Extensibility. Properties or Specialisation, ..?

connortsui20: Those things would need to compose, right? In which case we probably either need more than a single method or return a struct of things. For now, I think a boolean is sufficient until we actually figure out the semantics of pushing things down to refinement/extension types (see the comment in `vortex-array/src/arrays/scalar_fn/array.rs`).

connortsui20:
this is proving to be harder than expected to get right

vortex-array/src/dtype/extension/vtable.rs
joseph-isaacs: I guess you might expect them to be compositional, I was not

connortsui20:
This is more complicated than I initially thought, so I'm going to punt this to later

PR #7627, opened 2026-04-24
Validate zoned build

connortsui20:
## Summary

Adds some validation to the `ZonedLayout` building.

## Testing

Some error catching tests.

### DIFF
--- a/vortex-layout/src/layouts/zoned/mod.rs
+++ b/vortex-layout/src/layouts/zoned/mod.rs
@@ -20,6 +20,8 @@ use vortex_array::stats::stats_from_bitset_bytes;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
+use vortex_error::vortex_ensure;
+use vortex_error::vortex_ensure_eq;
 use vortex_error::vortex_panic;
 use vortex_session::VortexSession;
 use vortex_session::registry::ReadContext;
@@ -118,6 +120,11 @@ impl VTable for Zoned {
         children: &dyn LayoutChildren,
         _ctx: &ReadContext,
     ) -> VortexResult<Self::Layout> {
+        vortex_ensure_eq!(
+            children.nchildren(),
+            2,
+            "ZonedLayout expects exactly 2 children (data, zones)"
+        );
         Ok(ZonedLayout {
             dtype: dtype.clone(),
             children: children.to_arc(),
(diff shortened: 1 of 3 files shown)


vortex-layout/src/layouts/zoned/mod.rs
```diff
@@ -194,7 +211,15 @@ impl DeserializeMetadata for ZonedMetadata {
     type Output = Self;
 
     fn deserialize(metadata: &[u8]) -> VortexResult<Self::Output> {
+        vortex_ensure!(
+            metadata.len() >= 4,
+            "Zoned metadata must contain at least 4 bytes for zone length, got {}",
+            metadata.len()
+        );
```
robert3005, line 208: line 220 does this

vortex-layout/src/layouts/zoned/mod.rs
```diff
@@ -118,6 +120,21 @@ impl VTable for Zoned {
         children: &dyn LayoutChildren,
         _ctx: &ReadContext,
     ) -> VortexResult<Self::Layout> {
+        vortex_ensure!(metadata.zone_len != 0, "Zone length must be greater than 0");
```
robert3005: do we forbid writing empty zone layouts anywhere?

robert3005 (changes requested):
I would like to hold the invariant that we will not look at the zone layout unless we have a filter and right now we are going to read it just so we can check row count

vortex-layout/src/layouts/zoned/mod.rs
connortsui20: it was mostly just to have a better error message

vortex-layout/src/layouts/zoned/writer.rs
```diff
@@ -80,6 +81,11 @@ impl LayoutStrategy for ZonedStrategy {
         mut eof: SequencePointer,
         session: &VortexSession,
     ) -> VortexResult<LayoutRef> {
+        vortex_ensure!(
+            self.options.block_size > 0,
+            "ZonedStrategy requires block_size > 0 when writing"
+        );
```
connortsui20, line 87: Here is the write-side check

robert3005 (approved):

connortsui20:
actually I am going to run the benchmarks first just to be sure

connortsui20:
i cant tell if the regressions are real (they are from a different PR)

PR #7628, opened 2026-04-24
Improve stats reporting to DF

AdamGS:
## Summary

By comparing the physical plans between Parquet and Vortex, I've found a few things we're doing that are either wrong or suboptimal:
1. Reporting inexact column and file byte sizes if they are missing (which they are), which at the very least leads DF down a suboptimal path which introduces an extra RepartitionExec which isn't necessary.
2. He have a cute trick to report cardinality estimation for constant columns, problem is if we said we had exactly one record even if `IsConstant` was false.

There are still a few stats things that could probably be improved here, especially which stats are calculated and stored where and how that's configured, but I think for now this is a very nice improvement.

### DIFF
--- a/vortex-datafusion/src/convert/stats.rs
+++ b/vortex-datafusion/src/convert/stats.rs
@@ -9,6 +9,7 @@ use vortex::dtype::Nullability;
 use vortex::dtype::PType;
 use vortex::error::VortexExpect;
 use vortex::error::VortexResult;
+use vortex::expr::stats::Precision as VortexPrecision;
 use vortex::expr::stats::Stat;
 use vortex::scalar::Scalar;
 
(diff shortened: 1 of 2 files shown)


joseph-isaacs:
this definitely did something. 

got a 2x regression too.

AdamGS:
the regression is because we now that we have an absent size stat, DF uses the row count which makes it mis-judge the size of a colume. I do think its better to report semantically correct things even if they end up being worse in some cases (fwiw - I think we're still better than parquet for TPC-DS q22).

joseph-isaacs (approved):

robert3005:
We should decide what to do with UncompressedSizeInBytes. The idea is cute but not sure what we should be calculating and computation of it is expensive

AdamGS:
thing is - we already do! `CompressionStrategy` computes all of them. What if we introduce it into the PRUNING_STATS and just opt out in all nested types (struct and list) to avoid what @asubiotto ran into? 
Seems like there's some good perf just beyond getting these right, rn just for DF but soon for DuckDB thanks to @myrrc and any future query engine.

PR #7630, opened 2026-04-24
fix(sequence): fall back to canonical cast when params can't fit target

### LINKED ISSUE #5102: SequenceArray cannot cast to a narrower type
### Describe the bug

The sequence array `[30, 20, 10, 0]` should be castable to u8, but it is not because its multiplier is -10.

Sequence Array should use one PType for calculation and a different PType for the return value.

See also: https://github.com/vortex-data/vortex/pull/5101

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

changjoon-park:
## Summary

Closes #5102.

`SequenceArray` represents arithmetic sequences as `(base, multiplier, len)`. When cast to a narrower or differently-signed target type, the current implementation tries to cast the stored `base` and `multiplier` to the target type directly. This fails when either parameter cannot be represented in the target type even if every produced value is representable.

Concrete failing case (matches the test file's `TODO(DK)` comment):

```
Sequence(base=100i32, multiplier=-10i32, len=5)
  → values [100, 90, 80, 70, 60]   (all fit in u8)
  → current: fails with "Cannot cast -10i32 to u8"
  → expected: succeed with [100u8, 90, 80, 70, 60]
```

## Fix

Return `Ok(None)` from `CastReduce::cast` when casting `base` or `multiplier` fails. This delegates to the canonical cast path, which decodes the sequence into individual values and casts them one by one — correctly succeeding whenever all produced values fit the target type.

This matches the existing behavior for float targets, which already returns `Ok(None)` in the same file. `CastReduce`'s `Ok(None)` contract is for "rule doesn't apply, use another approach" situations.

## What this PR does not do

The issue suggests a more invasive fix: separate `PType`s for calculation vs. return value. That would preserve Sequence's compact representation through the cast and avoid the canonical decode. This PR takes the narrower approach of using the existing canonical fallback to make the operation correct; a parameter-type redesign would be a separate PR with its own design discussion.

## Testing

- Reproduced the bug first: uncommented `#[case::negative_step(...)]` and confirmed `test_cast_sequence_conformance::case_3_negative_step` failed with `Cannot cast -10i32 to u8`.
- After fix: same case passes, and all 5 conformance cases pass.
- `cargo test -p vortex-sequence` — 57/57 pass.
- `cargo test -p vortex --lib` — 5/5 pass.
- `cargo +nightly fmt --all -- --check` — clean.
- `cargo clippy -p vortex-sequence --all-targets --all-features` — no warnings.

No public API changes other than extending the set of casts that succeed.

### DIFF
--- a/encodings/sequence/src/compute/cast.rs
+++ b/encodings/sequence/src/compute/cast.rs
@@ -54,10 +54,20 @@ impl CastReduce for Sequence {
                 Some(ScalarValue::Primitive(array.multiplier())),
             )?;
 
-            let new_base_scalar =
-                base_scalar.cast(&DType::Primitive(*target_ptype, Nullability::NonNullable))?;
-            let new_multiplier_scalar = multiplier_scalar
-                .cast(&DType::Primitive(*target_ptype, Nullability::NonNullable))?;
+            let target_scalar_dtype = DType::Primitive(*target_ptype, Nullability::NonNullable);
+
+            // If either the base or multiplier cannot be represented in the target
+            // type (e.g., a negative multiplier with an unsigned target), fall back
+            // to the canonical cast path by returning `Ok(None)`. That path decodes
+            // individual values and casts them one by one, which correctly succeeds
+            // whenever all produced values fit the target type even if the stored
+            // parameters do not.
+            let Ok(new_base_scalar) = base_scalar.cast(&target_scalar_dtype) else {
+                return Ok(None);
+            };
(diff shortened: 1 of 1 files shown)


robert3005:
The fix here should be to have two ptypes not fallback to canonical as the issue describes.

changjoon-park:
> The fix here should be to have two ptypes not fallback to canonical as the issue describes.

make sense. the two-ptype approach is a larger redesign than I' d scoped

thanks for the review !

PR #7631, opened 2026-04-24
Late materialization support for duckdb

myrrc:
Add file_index, file_row_number virtual columns.
Add file-based filtering (range, selection) to ScanRequest.
Add partition index method.
Add late materialization support and row id columns support in duckdb.

### DIFF
--- a/vortex-duckdb/cpp/table_function.cpp
+++ b/vortex-duckdb/cpp/table_function.cpp
@@ -10,6 +10,7 @@ DUCKDB_INCLUDES_BEGIN
 #include "duckdb.h"
 #include "duckdb/catalog/catalog.hpp"
 #include "duckdb/common/insertion_order_preserving_map.hpp"
+#include "duckdb/common/multi_file/multi_file_reader.hpp"
 #include "duckdb/function/table_function.hpp"
 #include "duckdb/main/capi/capi_internal.hpp"
 #include "duckdb/main/connection.hpp"
@@ -19,6 +20,8 @@ DUCKDB_INCLUDES_END
 using namespace duckdb;
 using vortex::CData;
 using vortex::IntoErrString;
+constexpr column_t COLUMN_IDENTIFIER_FILE_INDEX = MultiFileReader::COLUMN_IDENTIFIER_FILE_INDEX;
+constexpr column_t COLUMN_IDENTIFIER_FILE_ROW_NUMBER = MultiFileReader::COLUMN_IDENTIFIER_FILE_ROW_NUMBER;
 
 struct CTableFunctionInfo final : TableFunctionInfo {
     explicit CTableFunctionInfo(const duckdb_vx_tfunc_vtab_t &vtab) : vtab(vtab) {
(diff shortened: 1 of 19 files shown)


vortex-duckdb/cpp/table_function.cpp
```diff
@@ -19,6 +20,8 @@ DUCKDB_INCLUDES_END
 using namespace duckdb;
 using vortex::CData;
 using vortex::IntoErrString;
+constexpr column_t COLUMN_IDENTIFIER_FILE_INDEX = MultiFileReader::COLUMN_IDENTIFIER_FILE_INDEX;
+// constexpr column_t COLUMN_IDENTIFIER_FILE_ROW_NUMBER = MultiFileReader::COLUMN_IDENTIFIER_FILE_ROW_NUMBER;
```
joseph-isaacs: ?

vortex-duckdb/src/duckdb/value.rs
```diff
@@ -203,6 +203,10 @@ impl Value {
         }
     }
 
+    pub fn new_u64(value: u64) -> Self {
+        unsafe { Self::own(cpp::duckdb_create_uint64(value as u64)) }
+    }
```
joseph-isaacs: Value::from(55)

vortex-duckdb/src/datasource.rs
```diff
@@ -67,25 +67,21 @@ use crate::duckdb::DataChunkRef;
...
-/// it uninitialized in this case, we define COLUMN_IDENTIFIER_EMPTY as a
-/// virtual column.
-/// See virtual_columns in vortex-duckdb/cpp/table_function.cpp
-static EMPTY_COLUMN_IDX: u64 = 18446744073709551614;
+/// File index virtual column, may be requested either by user or optimizer.
+static FILE_INDEX_COLUMN_IDX: u64 = 9223372036854775810;
+
+/// See duckdb/src/common/constants.cpp
+fn is_virtual_column(id: u64) -> bool {
+    return id >= 9223372036854775808u64;
```
joseph-isaacs: can you scope this to a struct?

vortex-duckdb/src/datasource.rs
```diff
@@ -423,30 +428,36 @@ impl<T: DataSourceTableFunction> TableFunction for T {
...
                 break;
             }
         }
 
         assert!(!chunk.is_empty());
 
+        if let Some(pos) = global_state.file_idx_pos {
+            chunk
+                .get_vector_mut(pos)
+                .reference_value(&Value::new_u64(local_state.file_idx as u64));
```
joseph-isaacs: hold a Value once?

joseph-isaacs (approved):

vortex-duckdb/src/datasource.rs
myrrc: But we need different Values for different file indices.

myrrc:
@joseph-isaacs The particularly hacky stuff here is adding partition indices, however, I can't imagine another way to filter files in a stable manner.

joseph-isaacs:
nice. let me review this tomorrow and get this over the line. what is the perf results?

vortex-duckdb/src/convert/table_filter.rs
```diff
@@ -125,3 +131,96 @@ pub fn try_from_table_filter(
...
+            std::cmp::Ordering::Greater => j += 1,
+        }
+    }
+    result
+}
+
+/// For constant comparison on IN filters over file_index or file_row_number
+/// virtual column, create a selection and a range covering the same range as
+/// expressions do.
+pub fn try_from_virtual_column_filter(
```
joseph-isaacs, line 169: Can you add a todo saying we would prefer to use an expression and have those handles selections efficiently.

vortex-duckdb/src/convert/table_filter.rs
```diff
@@ -125,3 +131,96 @@ pub fn try_from_table_filter(
...
+        match left[i].cmp(&right[j]) {
+            std::cmp::Ordering::Equal => {
+                result.push(left[i]);
+                i += 1;
+                j += 1;
+            }
+            std::cmp::Ordering::Less => i += 1,
+            std::cmp::Ordering::Greater => j += 1,
+        }
+    }
```
joseph-isaacs, line 162: this should be in a method Selection::merge

vortex-duckdb/src/convert/table_filter.rs
```diff
@@ -125,3 +131,96 @@ pub fn try_from_table_filter(
...
+                if let Some(r) = range {
+                    start = start.max(r.start);
+                    end = end.min(r.end);
+                }
+            }
+            let range = (start < end).then_some(start..end);
+            let sel = indices
+                .map(|v| Selection::IncludeByIndex(Buffer::from_iter(v)))
+                .unwrap_or(Selection::All);
+            Ok((sel, range))
```
joseph-isaacs, line 219: this is just selection merge?

vortex-scan/src/lib.rs
```diff
@@ -125,6 +125,10 @@ pub struct ScanRequest {
     /// A row selection to apply to the scan. The selection identifies rows within the specified
     /// row range.
     pub selection: Selection,
+    /// If we're operating on files, what files to read
+    pub file_selection: Selection,
```
joseph-isaacs, line 129: This only makes sense on a specific type of scan

vortex-scan/src/lib.rs
```diff
@@ -125,6 +125,10 @@ pub struct ScanRequest {
...
+    /// If we're operating on files, what files to read
+    pub file_range: Option<Range<u64>>,
```
joseph-isaacs, line 131: we don't need both

vortex-duckdb/src/convert/table_filter.rs
myrrc: Selection::merge is a more generic (and hard) method to implement. Here we're sure we're handling either Selection::All or Selection::IncludeByIndex.

vortex-scan/src/lib.rs
myrrc: For duckdb we do, because it emits file ranges. Yes, you can turn a range into a selection, and it would probably be fast, but I think an extra parameter isn't that bad.

myrrc:
> What's with the perf

Main impact should be clickbench Q23 stabilisation.
Current benchmarks show 
```
clickbench_q23/duckdb:vortex-file-compressed 🚀 	295280316 	1597941501 	0.18
```

That's likely noise, but local runs showed its speedup as well.

vortex-scan/src/lib.rs
myrrc: I agree, but we need to filter out lazy non-materialized files. Is there another way?

PR #7632, opened 2026-04-24
feat: teach Array<Struct> try_concat(&[Array<Struct>])

danking:
Look ma, no execution context!

### DIFF
--- a/vortex-array/src/arrays/chunked/vtable/canonical.rs
+++ b/vortex-array/src/arrays/chunked/vtable/canonical.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use itertools::Itertools as _;
 use vortex_buffer::Buffer;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
@@ -62,12 +63,10 @@ pub(super) fn _canonicalize(
 ///
 /// The caller guarantees there are at least 2 chunks.
 fn pack_struct_chunks(chunks: Vec<ArrayRef>, ctx: &mut ExecutionCtx) -> VortexResult<StructArray> {
-    let executed_chunks = chunks
+    chunks
         .into_iter()
         .map(|c| c.execute::<StructArray>(ctx))
-        .collect::<Result<Vec<_>, _>>()?;
-
-    StructArray::try_concat(executed_chunks.as_slice())
+        .process_results(|iter| StructArray::try_concat(iter))?
 }
 
 /// Packs [`ListViewArray`]s together into a chunked `ListViewArray`.
--- a/vortex-array/src/arrays/chunked/vtable/canonical.rs
+++ b/vortex-array/src/arrays/chunked/vtable/canonical.rs
@@ -18,13 +18,11 @@ use crate::arrays::StructArray;
 use crate::arrays::chunked::ChunkedArrayExt;
 use crate::arrays::listview::ListViewArrayExt;
 use crate::arrays::listview::ListViewRebuildMode;
-use crate::arrays::struct_::StructArrayExt;
 use crate::builders::builder_with_capacity_in;
 use crate::builtins::ArrayBuiltins;
 use crate::dtype::DType;
 use crate::dtype::Nullability;
 use crate::dtype::PType;
-use crate::dtype::StructFields;
 use crate::memory::HostAllocatorExt;
 use crate::validity::Validity;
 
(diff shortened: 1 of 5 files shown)


vortex-array/src/arrays/struct_/array.rs
```diff
@@ -454,3 +456,68 @@ impl Array<Struct> {
         self.remove_column(name)
     }
 }
+
+impl TryFrom<&[Array<Struct>]> for Array<Struct> {
```
a10y: i think it's a bit misleading to use this as the try_from impl. I personally would've expected StructArray::try_from([array1, array2]) to merge horizontally.

I think something like `StructArray::concat()` or `StructArray::from_chunks` is a better place for this to live

vortex-array/src/validity.rs
```diff
@@ -447,6 +448,38 @@ impl From<&Nullability> for Validity {
...
+    ///
+    /// Returns None if the vector is empty.
+    pub fn concat(validities: Vec<(Validity, usize)>) -> Option<Self> {
+        let (first, _) = validities.first()?;
+
+        if !matches!(first, Validity::Array(_)) {
+            let target = std::mem::discriminant(first);
+            if validities
+                .iter()
+                .all(|(v, _)| std::mem::discriminant(v) == target)
```
connortsui20: I feel like a comment would be nice here to explain how none of the validity variants are compatible (it took some time to think about why `NonNullable` is incompatible with `AllValid`, I think it's actually not possible if the validities come from a chunked array but that might not be obvious if the validities are not coming from a chunked array?)

vortex-array/src/arrays/struct_/array.rs
robert3005: fwiw that's what `merge` expression is

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -454,3 +456,68 @@ impl Array<Struct> {
...
+    type Error = VortexError;
+
+    fn try_from(value: &[Array<Struct>]) -> Result<Self, Self::Error> {
```
connortsui20: just renaming `value` to `struct_arrays` would make this much easier to grok imo

vortex-array/src/arrays/struct_/array.rs
joseph-isaacs: I like from chunks. 

I think keep away from operation like names here.

a10y: Yea I'm just saying I think this is non-obvious enough that it should leave in an appropriately named method rather than the generic tryfrom impl

danking: `try_concat`. Done.

vortex-array/src/arrays/struct_/array.rs
danking: `chunks`. Done.

vortex-array/src/validity.rs
danking: I changed the code a bit. Does it seem more obviously correct now?

connortsui20: yep this is good

vortex-array/src/arrays/chunked/vtable/canonical.rs
```diff
@@ -64,36 +61,13 @@ pub(super) fn _canonicalize(
...
+fn pack_struct_chunks(chunks: Vec<ArrayRef>, ctx: &mut ExecutionCtx) -> VortexResult<StructArray> {
+    let executed_chunks = chunks
+        .into_iter()
+        .map(|c| c.execute::<StructArray>(ctx))
+        .collect::<Result<Vec<_>, _>>()?;
 
-    // SAFETY: field_arrays are built from corresponding chunks of same length, dtypes match by
-    // construction.
-    Ok(unsafe { StructArray::new_unchecked(field_arrays, struct_dtype.clone(), len, validity) })
+    StructArray::try_concat(executed_chunks.as_slice())
```
gatesn: Should this just take impl Iterator?

danking: Done.

vortex-array/src/arrays/struct_/array.rs
```diff
@@ -453,4 +455,71 @@ impl Array<Struct> {
     pub fn remove_column_owned(&self, name: impl Into<FieldName>) -> Option<(Self, ArrayRef)> {
         self.remove_column(name)
     }
+
+    pub fn try_concat<T>(chunks: impl IntoIterator<Item = T>) -> VortexResult<Self>
+    where
+        T: Borrow<Array<Struct>>,
+    {
+        let mut it = chunks.into_iter();
+        // .map(|chunk| (chunk.dtype(), chunk, chunk.len(), chunk.validity()));
```
robert3005:
```suggestion
```

robert3005 (commented):
If you remove that one comment we can get this merged

robert3005 (approved):

PR #7634, opened 2026-04-24
`zoned` cleanup

connortsui20:
## Summary

Cleans up `ZonedLayout`. This is just a cosmetic change as I was trying to understand it. Also adds some tests that I thought would be helpful.

## Testing

N/A


robert3005 (approved):
I will allow it this time but not sure if pure cleanup prs like this are overall what we want

PR #7646, opened 2026-04-26
Benchmarks v3 migration to duckdb

connortsui20:
This is a one-shot migration binary to take all of the data from `data.json.gz` and bring it into a duckdb database.

Simply gathers and aggregates everything into memory and writes data in chunks with arrow arrays. Insert row-by-row took way too long, and the appender API in duckdb does not support `BIGINT[]` for some reason...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10353,6 +10353,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "vortex-bench-migrate"
+version = "0.1.0-alpha.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "duckdb",
+ "flate2",
+ "reqwest 0.13.2",
+ "rstest",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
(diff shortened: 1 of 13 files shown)


connortsui20:
There are some group regressions that I will eventually figure out if we need them at all:

```
Groups in both v2 and v3:
  + Clickbench
  + Compression
  + Compression Size
  + PolarSignals Profiling
  + Random Access
  + Statistical and Population Genetics
  + TPC-DS (NVMe) (SF=1)
  + TPC-H (NVMe) (SF=1)
  + TPC-H (NVMe) (SF=10)
  + TPC-H (NVMe) (SF=100)
  + TPC-H (S3) (SF=1)
  + TPC-H (S3) (SF=10)
  + TPC-H (S3) (SF=100)
Groups only in v2 (regression candidates):
  - TPC-DS (NVMe) (SF=10)
  - TPC-H (NVMe) (SF=1000)
  - TPC-H (S3) (SF=1000)
Groups only in v3:
  + Fineweb
Chart count diffs:
  Compression : v2=10 v3=6 (delta=-4)
  Compression Size : v2=6 v3=4 (delta=-2)
```

PR #7667, opened 2026-04-27
feat(cuda): prefer standalone kernels in `auto` mode

0ax1:
In `auto` dispatch mode, `try_gpu_dispatch` now checks `has_standalone_kernel` before the dynamic-dispatch plan builder.

Standalone kernels are preferred when they can decode the entire tree in a single launch without recursing into `execute_cuda` for child encodings.

### DIFF
--- a/vortex-cuda/src/dynamic_dispatch/plan_builder.rs
+++ b/vortex-cuda/src/dynamic_dispatch/plan_builder.rs
@@ -47,6 +47,7 @@ use super::ptype_to_tag;
 use super::tag_to_ptype;
 use crate::CudaBufferExt;
 use crate::CudaExecutionCtx;
+use crate::executor::CudaDispatchMode;
 use crate::kernel::load_patches_to_gpu;
 
 /// A plan whose source buffers have been copied to the device, ready for kernel launch.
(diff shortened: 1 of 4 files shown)


vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -121,6 +122,33 @@ fn is_dyn_dispatch_compatible(array: &ArrayRef) -> bool {
         || id == Sequence.id()
 }
 
+/// Returns `true` if a registered standalone kernel can decode the entire
+/// `array` tree in a single launch without recursing into `execute_cuda`
+/// for child encodings.
+pub fn has_standalone_kernel(array: &ArrayRef) -> bool {
```
a10y, line 128: in theory this should probably examine the session right?

joseph-isaacs (approved):

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
0ax1: I guess we could but for some kernels, like dict, we def want to prefer dynamic dispatch. Part of this is that we only want to call the standalone kernel if it can decompress the whole tree.

0ax1: Probably worth considering whether we should remove dict, and runend maybe as standalone kernels.

PR #7675, opened 2026-04-27
feat(ffi): add literal expression support

m7kss1:
## Summary

Add support for literal expressions in `vortex-ffi`

Motivation example: 

Сallers need constants in expression trees to create scan predicates that can be pushed down, for example:

```sql
age >= 67
price >= decimal(1512.0)
```

Introduce `vx_scalar` as an opaque ffi handle that exposes a typed Vortex scalar (`DType` + `optional ScalarValue`) and `vx_expression_literal` that allows to create literal expression nodes from scalar handle

### Scalar:

`vx_scalar_new_*` creates a Rust `Scalar` and returns it as an owned opaque boxed handle (`vx_scalar *`). APIs that take `const vx_scalar *` borrow the handle; `vx_expression_literal` clones the underlying scalar into the expression, so the caller can free the original handle after construction

```c
vx_scalar *vx_scalar_new_u8(uint8_t value, bool is_nullable);
vx_scalar *vx_scalar_new_utf8(const char *ptr, size_t len, bool is_nullable, vx_error **err);
vx_scalar *vx_scalar_new_null(const vx_dtype *dtype, vx_error **err);

const vx_dtype *vx_scalar_dtype(const vx_scalar *scalar);
bool vx_scalar_is_null(const vx_scalar *scalar);
vx_scalar *vx_scalar_clone(const vx_scalar *scalar);
void vx_scalar_free(vx_scalar *scalar);
```

### Literal expression:

`vx_expression_literal` clones the scalar into the expression, so the original scalar can be freed immediately after expression creation:

```c
vx_expression *root = vx_expression_root();
vx_expression *age = vx_expression_get_item("age", root);

vx_scalar *threshold_scalar = vx_scalar_new_u8(67, false);
vx_expression *threshold = vx_expression_literal(threshold_scalar, &error);
vx_scalar_free(threshold_scalar);

vx_expression *filter = vx_expression_binary(VX_OPERATOR_GTE, age, threshold);

vx_scan_options options = {};
options.filter = filter;
```

## Testing

Verifying new behavior and functionality works correctly

AI disclosure: Claude was used to add tests

### DIFF
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -1106,6 +1106,44 @@ void vx_expression_free(vx_expression *ptr);
 vx_expression *vx_expression_root(void);
 
 /**
+ * Create a literal boolean expression.
+ */
+vx_expression *vx_expression_literal_bool(bool value, bool is_nullable);
+
+/**
+ * Create a literal primitive expression.
+ *
+ * `value` must point to one value of the C type corresponding to `ptype`.
+ * For `PTYPE_F16`, the value is represented as `uint16_t`.
+ * If `value` is NULL, returns NULL.
+ */
+vx_expression *vx_expression_literal_primitive(vx_ptype ptype, const void *value, bool is_nullable);
+
+/**
+ * Create a literal UTF-8 expression.
+ *
+ * The input string is borrowed and copied into the returned expression.
+ * If `value` is NULL, returns NULL.
+ */
+vx_expression *vx_expression_literal_utf8(const vx_string *value, bool is_nullable);
+
+/**
+ * Create a literal binary expression.
+ *
+ * The input bytes are borrowed and copied into the returned expression.
+ * If `value` is NULL, returns NULL.
+ */
+vx_expression *vx_expression_literal_binary(const vx_binary *value, bool is_nullable);
+
+/**
+ * Create a typed null literal expression.
+ *
+ * The null literal uses a nullable version of `dtype`.
+ * If `dtype` is NULL, returns NULL.
+ */
+vx_expression *vx_expression_literal_null(const vx_dtype *dtype);
+
+/**
  * Create an expression that selects (includes) specific fields from a child
  * expression. Child expression must have a DTYPE_STRUCT dtype. Errors in
  * vx_array_apply if the child expression doesn't have a specified field.
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -490,13 +490,12 @@ typedef struct vx_file vx_file;
 typedef struct vx_partition vx_partition;
 
 /**
- * A `vx_scalar` stores a logical `DType` with an optional `ScalarValue`.
- * The `ScalarValue` is absent for a null scalar. Otherwise, the `DType`
- * gives the `ScalarValue` its meaning; for example, a tuple value can
- * represent a list, fixed-size list, or struct scalar depending on the
- * `DType`.
+ * A typed scalar value.
  *
- * The C API exposes only the complete pair as `vx_scalar`.
+ * A `vx_scalar` represents a single value with an associated `DType`.
+ * Its value is either null or a `ScalarValue`. Null values are allowed only
+ * when the associated `DType` allows nulls. Non-null values are represented
+ * by `ScalarValue` and interpreted using the `DType`.
  */
 typedef struct vx_scalar vx_scalar;
 
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -490,6 +490,16 @@ typedef struct vx_file vx_file;
 typedef struct vx_partition vx_partition;
 
 /**
+ * Opaque handle to a `Scalar` struct: a pair of a logical `DType` and an optional
+ * `ScalarValue` payload.
+ *
+ * The C API exposes only the complete pair as `vx_scalar`, so
+ * callers do not need to know enum layouts or maintain the `DType`/`ScalarValue`
+ * invariants themselves.
+ */
+typedef struct vx_scalar vx_scalar;
+
+/**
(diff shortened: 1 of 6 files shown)


m7kss1:
note: the current design is not ideal and will be changed. the ffi usage is expected to resemble something like this:

```cpp
vx_scalar *scalar = ...; // opaque class
vx_expression *expr = vx_expression_literal(scalar);
```

vortex-ffi/cinclude/vortex.h
```diff
@@ -1105,6 +1105,44 @@ void vx_expression_free(vx_expression *ptr);
...
+ */
+vx_expression *vx_expression_literal_primitive(vx_ptype ptype, const void *value, bool is_nullable);
+
+/**
+ * Create a literal UTF-8 expression.
+ *
+ * The input string is borrowed and copied into the returned expression.
+ * If `value` is NULL, returns NULL.
+ */
+vx_expression *vx_expression_literal_utf8(const vx_string *value, bool is_nullable);
```
myrrc: Thanks for the contribution. I believe we want a dedicated vx_scalar type and not just expressions, as scalar are useful for other purposes.

vortex-ffi/src/expression.rs
```diff
@@ -62,6 +73,104 @@ pub unsafe extern "C" fn vx_expression_root() -> *mut vx_expression {
...
+    is_nullable: bool,
+) -> *mut vx_expression {
+    vx_expression::new(lit(Scalar::bool(value, Nullability::from(is_nullable))))
+}
+
+/// Create a literal primitive expression.
+///
+/// `value` must point to one value of the C type corresponding to `ptype`.
+/// For `PTYPE_F16`, the value is represented as `uint16_t`.
+/// If `value` is NULL, returns NULL.
```
myrrc: Please add an API usage example. You may reference primitive array creation in src/array.rs

m7kss1:
There is 1 possible way to change proposed API:

Now we follow the simple rule:

If C ABI can carry the type, the constructor carries it; if the type is Vortex-specific, the caller might pass a `vx_dtype` 

```c
vx_scalar *vx_scalar_new_u8(uint8_t value, bool is_nullable); -- uint8_t and u8
vx_scalar *vx_scalar_new_i8(int8_t value, bool is_nullable); -- int8_t and i8

vx_scalar *vx_scalar_new_null(const vx_dtype *dtype, vx_error **err); -- null type problem
vx_scalar_new_struct(const vx_dtype *struct_dtype, ...) -- field names, field order, field dtypes or nullability
```

But for C-ABI types it could be limited to one base constructor with an explicit type specification:

```c
vx_scalar * vx_scalar_new_primitive(void * value, type, size, alignment, ...)
```

IMHO, the second approach is worse - for ffi using `void *` forces the caller to handle size/alignment correctly, which is error-prone

vortex-ffi/cinclude/vortex.h
```diff
@@ -489,6 +489,16 @@ typedef struct vx_file vx_file;
...
 typedef struct vx_partition vx_partition;
 
+/**
+ * Opaque handle to a `Scalar` struct: a pair of a logical `DType` and an optional
+ * `ScalarValue` payload.
+ *
+ * The C API exposes only the complete pair as `vx_scalar`, so
+ * callers do not need to know enum layouts or maintain the `DType`/`ScalarValue`
+ * invariants themselves.
+ */
```
myrrc: I think the API should have comments about how the things work as opposed to why they work in a specific way. So for this specific example, I'd add a comment like
"A vx_scalar is a pair of a DType and an optional value of this DType. Value may not be present if DType is nullable.".

vortex-ffi/cinclude/vortex.h
```diff
@@ -1252,6 +1296,236 @@ vx_array_iterator *vx_file_scan(const vx_session *session,
...
+vx_scalar *vx_scalar_new_f32(float value, bool is_nullable);
+
+/**
+ * Create a 64-bit floating point scalar.
+ */
+vx_scalar *vx_scalar_new_f64(double value, bool is_nullable);
+
+/**
+ * Create a 16-bit floating point scalar.
```
myrrc, line 1384: There is an option to use _Float16 but I'd like a second opinion here.
@0ax1

vortex-ffi/src/scalar.rs
```diff
@@ -0,0 +1,844 @@
...
+pub unsafe extern "C-unwind" fn vx_scalar_clone(scalar: *const vx_scalar) -> *mut vx_scalar {
+    if scalar.is_null() {
+        return ptr::null_mut();
+    }
+    vx_scalar::new(vx_scalar::as_ref(scalar).clone())
+}
+
+/// Return the data type of a scalar.
+///
+/// The returned data type handle borrows storage from the scalar handle and must
```
myrrc: Let's add comment about dtype that its lifetime is bound to lifetime of the scalar

myrrc (approved):
LGTM sans small comments. Let's wait for Alex about _Float16 exposure and we're good to go.

vortex-ffi/cinclude/vortex.h
0ax1: We can postpone the decision on this. We could probably be more explicit about the bitcasting into u16 required.

vortex-ffi/cinclude/vortex.h
```diff
@@ -489,6 +489,17 @@ typedef struct vx_file vx_file;
  */
 typedef struct vx_partition vx_partition;
 
+/**
+ * A `vx_scalar` stores a logical `DType` with an optional `ScalarValue`.
+ * The `ScalarValue` is absent for a null scalar. Otherwise, the `DType`
```
myrrc: > The `ScalarValue` is absent for a null scalar

I think this reads in two ways:
- ScalarValue is null only for Dtype::Null (this is incorrect)
- ScalarValue may be null if Dtype is Null or Dtype is Nullable (this is correct)

PR #7684, opened 2026-04-28
Make struct cast implementation pluggable

robert3005:
This pr moves Struct Cast implementation to be overridable by VortexSession.

The default implementation stays the same, i.e. it resolves the fields by name.

In follow up we will add back the positional cast and add both implementations
to the fuzzer

### DIFF
--- a/vortex-array/src/arrays/struct_/compute/cast.rs
+++ b/vortex-array/src/arrays/struct_/compute/cast.rs
@@ -16,6 +16,7 @@ use crate::arrays::scalar_fn::ExactScalarFn;
 use crate::arrays::struct_::StructArrayExt;
 use crate::builtins::ArrayBuiltins;
 use crate::dtype::DType;
+use crate::dtype::StructFields;
 use crate::matcher::Matcher;
 use crate::scalar::Scalar;
 use crate::scalar_fn::fns::cast::Cast;
@@ -50,30 +51,7 @@ pub(crate) fn struct_cast(
         return Ok(None);
     };
 
-    let source_sdtype = array.struct_fields();
-
-    let mut cast_fields = Vec::with_capacity(target_sdtype.nfields());
-    // Re-order, handle fields by value instead.
-    for (target_name, target_type) in target_sdtype.names().iter().zip_eq(target_sdtype.fields()) {
-        match source_sdtype.find(target_name) {
-            None => {
-                // No source field with this name => evolve the schema compatibly.
-                // If the field is nullable, we add a new ConstantArray field with the type.
-                vortex_ensure!(
-                    target_type.is_nullable(),
-                    "CAST for struct only supports added nullable fields"
-                );
-
-                cast_fields
-                    .push(ConstantArray::new(Scalar::null(target_type), array.len()).into_array());
-            }
-            Some(src_field_idx) => {
-                // Field exists in source field. Cast it to the target type.
-                let cast_field = array.unmasked_field(src_field_idx).cast(target_type)?;
-                cast_fields.push(cast_field);
-            }
-        }
-    }
+    let cast_fields = struct_cast_fields(array, target_sdtype)?;
 
     let validity = array
         .validity()?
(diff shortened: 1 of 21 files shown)


vortex-array/src/executor.rs
```diff
@@ -402,13 +406,44 @@ fn execute_step(array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Executi
...
+///
+/// The returned [`Arc`] is owned so the caller can drop the [`ArrayKernels`] borrow before
+/// invoking the functions.
+fn plugin_execute_parent(
+    ctx: &ExecutionCtx,
+    parent: &ArrayRef,
+    child: &ArrayRef,
+) -> Option<Arc<[ExecuteParentFn]>> {
+    ctx.session()
+        .get_opt::<ArrayKernels>()
```
joseph-isaacs: Hold this and pass it into the loop, so you avoid the repeated fetch

vortex-array/src/optimizer/kernels.rs
```diff
@@ -72,32 +132,73 @@ impl ArrayKernels {
...
+        )));
+    }
+
+    /// Look up the [`ExecuteParentFn`]s registered for `(parent, child)`.
+    ///
+    /// Returns an owned [`Arc`] so the session-variable borrow can be dropped before invoking the
+    /// functions.
+    pub fn find_execute_parent(&self, parent: Id, child: Id) -> Option<Arc<[ExecuteParentFn]>> {
+        let id = hash_fn_id(parent, child);
+        self.execute_parent.load().get(&id).cloned()
```
joseph-isaacs, line 201: would be nice to used `ArcSwap::cached` here. It might change the API slightly

vortex-array/src/optimizer/kernels.rs
```diff
@@ -45,23 +57,71 @@ static FN_HASHER: LazyLock<DefaultHashBuilder> = LazyLock::new(DefaultHashBuilde
...
 /// Session-scoped registry of optimizer kernel functions.
-#[derive(Debug, Default)]
+///
+/// Each kernel kind has its own storage map, keyed by `(outer_id, child_id)`. Registering a
+/// function for a key replaces any previous function registered for the same key and kind, so
+/// downstream sessions can override built-in plugins by re-registering them.
+#[derive(Debug)]
 pub struct ArrayKernels {
     reduce_parent: ArcSwap<HashMap<u64, Arc<[ReduceParentFn]>>>,
+    execute_parent: ArcSwap<HashMap<u64, Arc<[ExecuteParentFn]>>>,
```
joseph-isaacs: can we have the u64 as a newtype or at least type alias

vortex-array/src/arrays/struct_/compute/rules.rs
joseph-isaacs, line 1: Shall we clean up the deprecated here?

vortex-array/src/arrays/struct_/compute/cast.rs
```diff
@@ -24,88 +27,131 @@ impl CastKernel for Struct {
...
     ) -> VortexResult<Option<ArrayRef>> {
-        let Some(target_sdtype) = dtype.as_struct_fields_opt() else {
-            return Ok(None);
-        };
+        cast_struct(array, dtype)
+    }
+}
 
-        let source_sdtype = array.struct_fields();
+pub(crate) fn struct_cast_execute_parent(
```
joseph-isaacs, line 24: why is this an execute if `_ctx`?

vortex-array/src/arrays/struct_/compute/rules.rs
robert3005: I noticed there's more refactoring to do here

vortex-array/src/arrays/struct_/compute/cast.rs
robert3005: This should be just a reduce rule apart from the case where you want to convert validity from nullable into non nullable which MIGHT require compute. We need to refactor validity casting

connortsui20:
Is this related to https://github.com/vortex-data/vortex/issues/6900?

robert3005:
It is related. We need to acknowledge that different systems have different requirements and we need to be able to override the behaviour

robert3005:
as I was refactoring this logic I made https://github.com/vortex-data/vortex/pull/7710 to cleanup validity casting

vortex-array/src/arrays/struct_/compute/cast.rs
```diff
@@ -6,106 +6,149 @@ use vortex_error::VortexResult;
...
+    let dtype = parent.options;
+    if array.dtype() == parent.options {
+        return Ok(Some(array.array().clone()));
+    }
 
-        let fields_match_order = target_sdtype.nfields() == source_sdtype.nfields()
-            && target_sdtype
-                .names()
-                .iter()
-                .zip(source_sdtype.names().iter())
```
joseph-isaacs, line 37: this is a perf optimisation to avoid O(n^2) name lookups

joseph-isaacs (approved):
Later I want to make cases composable (not just pluggable) so we could document that this is non-composable due to the `vortex_ensure!`

vortex-array/src/arrays/struct_/compute/cast.rs
connortsui20: Did that actually come up as expensive in profiling? Otherwise I think the more naive version is harder to read and likely the same perf?

robert3005: I brought it back, and refactored the code

PR #7693, opened 2026-04-28
bug: type_id

joseph-isaacs:
I don't think we should use type_id any more?

What I believe happens is that the `type_id` for struct is crate local and this breaks equality of these objects.


a10y:
you're calling ArrayRef::as_any() on line 336 of typed.rs, and for some reason the SessionVar impl takes precedence instead of the DynArray impl of as_any.

SessionVar::as_any returns self while DynArray unwraps the arc into its inner ref so you get the real V instead of the Arc'ed version, that's what's happening

I think this is b/c SessionVar defines a blanket impl, whereas the ArrayRef.inner() returns &Arc<DynArray> and the compiler needs to insert a deref coercion to get the &DynArray out, and I guess the compiler picks blanket impl over deref coercion chain.

To fix this particular issue, you can be more explicit in the callsite. This is a footgun though and I'd be very much in favor of renaming the SessionVar trait method

PR #7695, opened 2026-04-28
remove SessionVar blanket impl, replace with type impls

a10y:
The blanket impl is a bit of a footgun, replace with impls for each of the session extension types

See discussion in https://github.com/vortex-data/vortex/pull/7693#issuecomment-4335695797

Playground link that illustrates what the blanket impl is doing: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=86edbba2a1f4d7e4fb46f281fb21a8c3


joseph-isaacs (approved):

joseph-isaacs:
Can you add a doc str explaining that you need to do this

gatesn:
Could we just rename the as_any/as_any_mut functions instead to not conflict?

PR #7696, opened 2026-04-28
ci[gpu]: run cuda micro-benchmarks with codspeed

0ax1:


0ax1:
~~Giving this a couple of runs first to check whether the benches are stable across runs.~~ seems stable

robert3005 (approved):

PR #7698, opened 2026-04-28
feat: iterative execution for DateTimePartsArray

joseph-isaacs:


encodings/datetime-parts/src/array.rs
```diff
@@ -208,46 +209,58 @@ impl VTable for DateTimeParts {
...
+}
 
 #[derive(Clone, Debug)]
 pub struct DateTimePartsData {}
 
-impl Display for DateTimePartsData {
-    fn fmt(&self, _f: &mut Formatter<'_>) -> std::fmt::Result {
-        Ok(())
-    }
+pub struct DateTimePartsParts {
```
0ax1, line 226: This has the same fields as `DateTimePartsSlots` do we need both?

joseph-isaacs: I think consistency beats duplication?

0ax1 (approved):

PR #7701, opened 2026-04-28
Tell agents to use Cargo Nextest for testing

robert3005:
Cargo test is very slow compared to nextest


connortsui20 (approved):
note that `nextest` does not run doc tests so maybe we should add a prompt for that (like "if you modified any docs please run the doc tests" or something)

PR #7702, opened 2026-04-28
Java CI is self contained and writes files through java instead of rust

robert3005:
This should speed up builds and avoid build up of rust artifacts


a10y (commented):
deleting code >>>

a10y (approved):

PR #7708, opened 2026-04-28
example raster compression using Vortex

a10y:
Just an experiment for now. Idea being, why not store raster data in Vortex.

I've run some local benchmarks extracting crops out of a local GeoTIFF. 

Vortex wins by a wide margin, but it's kind of cheating because

- The raster data doesn't get compressed at all (even with Compact compressor)
- TIFF access is dominated by the DEFLATE overhead

Screenshot 2026-04-28 at 15 10 01

## Next steps

- [ ] Find compressor scheme that works for raster tiles, which are `List<u8>`
- [ ] Benchmark E2E inference pipeline VS rasterio + torch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3893,6 +3893,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
+name = "fax"
+version = "0.2.7"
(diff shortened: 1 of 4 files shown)


a10y:
I suspect we could actually squeeze out a lot of compression if we retiled into 32x32 (1024 element) vectors, and FFOR'ed each of them. Then every tile becomes a _column_, not a row, and each one should bitpack appropriately. This also lets us combine projection pruning for tile selection + filtering to extract a subset of a tile. Makes it very easy to turn random rectangular ranges of the image into something we can trivially push down into a scan. Plus, each 1024-element vector would probably pack relatively well.

PR #7709, opened 2026-04-28
Add `NormalizedVector` extension type

connortsui20:
## Summary

Adds a `NormalizedVector` extension type to `vortex-tensor`, which wraps a `Vector` extension array. Since we may want to add refinement types in the future, we chose to wrap a `Vector` storage array instead of a `FixedSizeList` storage array.

Additionally does a bunch of general cleanup that was obviously bad when I was actually trying to add this (and I wasn't really able to pull it out into a separate PR because it is only obvious when working with this new type).

There are still a bunch of TODOs that I'd like to fix but I'm probably going to do those all at once in a later PR.

## Testing

Adds basic tests for the new `NormalizedVector` type and also adds some more complex tests for logic that was missing before for the older code.

### DIFF
--- a/vortex-tensor/src/types/vector/matcher.rs
+++ b/vortex-tensor/src/types/vector/matcher.rs
@@ -13,8 +13,8 @@ use vortex_error::vortex_panic;
 use crate::types::normalized_vector::NormalizedVector;
 use crate::types::vector::Vector;
 
-/// Matcher that accepts any vector-shaped extension type — both plain
-/// [`Vector`] and [`NormalizedVector`](crate::normalized_vector::NormalizedVector).
+/// Matcher that accepts any vector-shaped extension type (both plain [`Vector`] and
+/// [`NormalizedVector`]).
 ///
 /// To match a plain [`Vector`] only (excluding [`NormalizedVector`]), pair this matcher with a
 /// negated `is::<AnyNormalizedVector>()` check; to match a `NormalizedVector` only, use
--- a/vortex-tensor/src/types/vector/matcher.rs
+++ b/vortex-tensor/src/types/vector/matcher.rs
@@ -10,13 +10,16 @@ use vortex_error::VortexResult;
 use vortex_error::vortex_ensure;
 use vortex_error::vortex_panic;
 
+use crate::types::normalized_vector::NormalizedVector;
 use crate::types::vector::Vector;
 
-/// Matcher that accepts only the [`Vector`] extension type.
+/// Matcher that accepts any vector-shaped extension type — both plain
+/// [`Vector`] and [`NormalizedVector`](crate::normalized_vector::NormalizedVector).
 ///
-/// Use [`AnyTensor`](crate::matcher::AnyTensor) instead when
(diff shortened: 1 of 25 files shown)


gatesn:
> P1
SorfTransform cannot generally preserve NormalizedVector
This mirrors the child wrapper kind, so a NormalizedVector child produces a NormalizedVector output after inverse SORF plus truncation. That is not valid for an arbitrary unit vector in the padded dimension: the inverse orthogonal transform preserves norm before truncation, but truncating can drop energy. The docs also say this produces a plain Vector. Return Vector here unless the type system carries stronger provenance that proves the child came from the zero-padded SORF subspace.

> P2
f16 normalization tolerance is too loose
The tolerance scales as 10 * sqrt(dim) * eps. For f16 at common embedding sizes like 768 dimensions this is around 0.27, so NormalizedVector::try_new can accept rows whose norm is far from 1.0. Downstream L2Norm then treats the dtype as authoritative and returns 1.0, so exact validation should use a tighter or capped tolerance, especially for f16.

> P2
normalize_as_l2_denorm drops FixedShapeTensor shape metadata
The helper still accepts AnyTensor and documents FixedShapeTensor input, but the slow path now always builds a NormalizedVector child, so a FixedShapeTensor normalizes into the vector family and L2Denorm later executes as a plain Vector. This regresses the previous behavior where the normalized child kept the original tensor extension dtype. Either restrict this helper to Vector inputs or preserve the original extension family for FixedShapeTensor.

vortex-tensor/src/types/vector/matcher.rs
```diff
@@ -33,17 +41,41 @@ pub struct VectorMatcherMetadata {
...
+        } else if ext_dtype.is::<Vector>() {
+            (ext_dtype.storage_dtype(), false)
+        } else {
             return None;
-        }
+        };
 
-        let DType::FixedSizeList(element_dtype, list_size, _) = ext_dtype.storage_dtype() else {
+        let DType::FixedSizeList(element_dtype, list_size, _) = fsl_dtype else {
             vortex_panic!("`Vector` type somehow did not have a `FixedSizeList` storage type")
```
joseph-isaacs, line 79: we cannot do this its SO expensive.

This could be called in a tight execute loop.

joseph-isaacs: You must at least impl a match

connortsui20: I am confused by this, how else are we supposed to implement a matcher if we can't inspect the ext vtable?

Are you just saying that doing `is::<AnyNormalizedVector>` is faster? That is incredibly surprising to me and certainly not clear or documented anywhere clearly in our codebase.

connortsui20:
@joseph-isaacs this branch doesn't affect any of our current benchmarks (it doesn't go through this codepath)

connortsui20:
This is getting quite complicated, but it is unclear if this is inherent complexity or because the design decisions I chose were bad. I'm going to try a pull out some changes out of this, but it is hard to say if this is worth it.

connortsui20:
it would be nice to have this in the future...

PR #7710, opened 2026-04-29
Validity casting doesn't hide usage of execution context

robert3005:
Casting validity correctly advertises that it might do compute by requiring
context.

We split the logic of casting between CastReduce rules which can handle all
NonNullalbe -> Nullable and Nullable -> NonNullable casts if stats have been
calculated and CastKernels which need to handle full generality of casts, in
practice Nullable to NonNullable casts


encodings/fsst/src/compute/cast.rs
```diff
@@ -1,46 +1,106 @@
...
+
+        // `FSST::try_new` only validates metadata; the local SESSION's ctx is sufficient.
+        Ok(Some(build_with_codes_validity(
+            array,
+            dtype,
+            new_codes_validity,
+            &mut SESSION.create_execution_ctx(),
+        )?))
+    }
+}
```
joseph-isaacs, line 64: hmmm

vortex-array/src/arrays/listview/compute/cast.rs
```diff
@@ -21,23 +42,33 @@ impl CastReduce for ListView {
         };
 
         // Cast the elements to the target element type.
+        let new_elements = array.elements().cast((**target_element_type).clone())?;
+        let Some(validity) = array
+            .validity()?
+            .try_cast_nullability(dtype.nullability(), array.len())?
+        else {
+            return Ok(None);
```
joseph-isaacs, line 47: failing here is expensive

encodings/fsst/src/compute/cast.rs
robert3005: This is wrong

joseph-isaacs:
We need to ensure that reduce rule exit early if they cannot match and do little more unless we know they match, since we try to apply them so often.

encodings/fsst/src/compute/cast.rs
```diff
@@ -1,46 +1,106 @@
...
 use vortex_error::VortexResult;
+use vortex_session::VortexSession;
 
 use crate::FSST;
 use crate::FSSTArrayExt;
+
+/// Local session used as a default `ExecutionCtx` source for the trait-bound `CastReduce`
+/// path, which doesn't receive a context. `FSST::try_new` only validates metadata so this
+/// is harmless.
+static SESSION: LazyLock<VortexSession> =
```
robert3005: This is interesting

robert3005:
The thing that's extremely repetitive here is `dtype.eq_ignore_nullability(target)` which I am not quite sure we should abstract fully

vortex-array/src/validity.rs
```diff
@@ -360,28 +361,101 @@ impl Validity {
...
+            Self::Array(is_valid) => {
+                Ok(is_valid
+                    .statistics()
+                    .get_as::<bool>(Stat::Min)
+                    .and_then(|min| min.as_exact())
+                    // min true => all true
+                    .filter(|min| *min)
+                    .map(|_| Self::NonNullable))
+            }
+        }
```
joseph-isaacs: please can we not do this test and change the method name to reflect this

joseph-isaacs (approved):
Lets run benchmarks first

PR #7715, opened 2026-04-29
Add a new AggregateFn for UncompressedSize

AdamGS:
## Summary

This stat turns out to be [very useful](https://github.com/vortex-data/vortex/pull/7628) for helping execution engines plan joins better, but its currently requires a full-decompression. This PR includes the basic structure for a new AggregateFn, including implementation for all canonical types.

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list.rs
+++ b/vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list.rs
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_error::VortexResult;
+use vortex_error::vortex_err;
+
+use crate::ExecutionCtx;
+use crate::arrays::ListViewArray;
+use crate::arrays::listview::ListViewArrayExt;
+use crate::arrays::listview::ListViewRebuildMode;
+
+pub(super) fn list_uncompressed_size_in_bytes(
+    array: &ListViewArray,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<u64> {
+    let mut size = if array.is_empty() {
+        0
+    } else {
+        let rebuilt = array.rebuild(ListViewRebuildMode::MakeExact)?;
+        super::uncompressed_size_in_bytes_u64(rebuilt.elements(), ctx)?
+    };
+
+    let view_buffer_size = u64::try_from(array.len())
+        .map_err(|e| vortex_err!("Failed to convert list array length to u64: {e}"))?
+        .checked_mul(8)
+        .ok_or_else(|| vortex_err!("uncompressed size in bytes overflowed u64"))?;
+
+    size = size
+        .checked_add(view_buffer_size)
+        .and_then(|size| size.checked_add(view_buffer_size))
+        .ok_or_else(|| vortex_err!("uncompressed size in bytes overflowed u64"))?;
+    size = size
+        .checked_add(super::validity_uncompressed_size_in_bytes(
+            array
+                .as_ref()
+                .validity()?
+                .execute_mask(array.as_ref().len(), ctx)?,
+        )?)
+        .ok_or_else(|| vortex_err!("uncompressed size in bytes overflowed u64"))?;
+
+    Ok(size)
+}
--- a/vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list_view.rs
+++ b/vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list_view.rs
@@ -9,6 +9,7 @@ use super::validity_uncompressed_size_in_bytes;
 use crate::ExecutionCtx;
 use crate::arrays::ListViewArray;
 use crate::arrays::listview::ListViewArrayExt;
+use crate::arrays::listview::ListViewRebuildMode;
 
 pub(super) fn list_view_uncompressed_size_in_bytes(
     array: &ListViewArray,
@@ -17,7 +18,8 @@ pub(super) fn list_view_uncompressed_size_in_bytes(
     let mut size = if array.is_empty() {
         0
     } else {
-        uncompressed_size_in_bytes_u64(array.elements(), ctx)?
+        let rebuilt = array.rebuild(ListViewRebuildMode::MakeExact)?;
+        uncompressed_size_in_bytes_u64(rebuilt.elements(), ctx)?
     };
 
     let view_buffer_size = u64::try_from(array.len())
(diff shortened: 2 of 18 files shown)


vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub(super) fn list_uncompressed_size_in_bytes(
+    array: &ListViewArray,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<u64> {
+    let mut size = if array.is_empty() {
+        0
+    } else {
+        let rebuilt = array.rebuild(ListViewRebuildMode::MakeExact)?;
+        super::uncompressed_size_in_bytes_u64(rebuilt.elements(), ctx)?
```
robert3005: Can we import all those `super::` methods

AdamGS: done

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    Ok(size)
+}
+
+/// Sum the canonical, recursively uncompressed buffer sizes for an array.
+///
+/// Applies to all types and returns a non-null `u64`. Encoding kernels can return this aggregate
+/// directly from metadata to avoid decoding arrays whose uncompressed size is known.
+#[derive(Clone, Debug)]
+pub struct UncompressedSizeInBytes;
```
joseph-isaacs, line 98: So this is only buffers of the array and all children. 

Are we sure we should not include metadata?

Maybe just justify why?

AdamGS: I don't think we do now, and it should really be a rounding error right?
It also seems to be more of an implementation details and not the number execution engines care about.

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list_view.rs
```diff
@@ -0,0 +1,45 @@
...
+pub(super) fn list_view_uncompressed_size_in_bytes(
+    array: &ListViewArray,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<u64> {
+    let mut size = if array.is_empty() {
+        0
+    } else {
+        let rebuilt = array.rebuild(ListViewRebuildMode::MakeExact)?;
+        uncompressed_size_in_bytes_u64(rebuilt.elements(), ctx)?
+    };
```
robert3005, line 23: I know this is the existing behaviour, however, we likely want to change it to just take elements size

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
```diff
@@ -0,0 +1,570 @@
...
+        &self,
+        partial: &mut Self::Partial,
+        batch: &Columnar,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<()> {
+        let size = match batch {
+            Columnar::Canonical(canonical) => canonical_uncompressed_size_in_bytes(canonical, ctx)?,
+            Columnar::Constant(constant) => {
+                let array = constant.clone().into_array();
+                materialized_uncompressed_size_in_bytes(&array)
```
robert3005: this should delegate to the constant array implementation that extrapolates for canonical types

robert3005 (changes requested):
Constant impl needs changing, this reads a bit too much like LLM generated code

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
joseph-isaacs: You need a more precise definition of this value. I guess its sum of buffers in the fully canonical variant?

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list_view.rs
AdamGS: as in - just `uncompressed_size_in_bytes_u64(array.elements(), ctx)?`?

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
AdamGS: meant to do this as a followup, but glad to do it here.

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/list_view.rs
AdamGS: well keep it for now, I'll try to do it better in a followup

robert3005: I actually don't think you can do fundamentally better. You can only really do different by overestimating and including the dead bytes

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
robert3005: Let's define it as byte size of buffers and children in canonical representation. Can you add a note?

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
```diff
@@ -0,0 +1,645 @@
...
+use crate::aggregate_fn::AggregateFnId;
+use crate::aggregate_fn::AggregateFnVTable;
+use crate::aggregate_fn::DynAccumulator;
+use crate::aggregate_fn::EmptyOptions;
+use crate::array::ArrayView;
+use crate::arrays::Constant;
+use crate::arrays::ConstantArray;
+use crate::arrays::varbinview::BinaryView;
+#[cfg(test)]
+use crate::builders::builder_with_capacity;
```
robert3005: can you move this to the tests module?

vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
```diff
@@ -0,0 +1,645 @@
...
+}
+
+#[cfg(test)]
+fn materialized_uncompressed_size_in_bytes(array: &ArrayRef) -> u64 {
+    let mut builder = builder_with_capacity(array.dtype(), array.len());
+    unsafe {
+        builder.extend_from_array_unchecked(array);
+    }
+    builder.finish().nbytes()
+}
```
robert3005: as well as this

robert3005 (approved):

PR #7721, opened 2026-04-29
Late materialization support for duckdb

myrrc:
Add file_index, file_row_number virtual columns.
Add file-based filtering (range, selection) to ScanRequest.
Add partition index method.
Add late materialization support and row id columns support in duckdb.

Attempt 1 was accidentally merged at https://github.com/vortex-data/vortex/pull/7631 and reverted

### DIFF
--- a/vortex-duckdb/src/convert/table_filter.rs
+++ b/vortex-duckdb/src/convert/table_filter.rs
@@ -1,14 +1,17 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::ops::Range;
 use std::sync::Arc;
 
 use itertools::Itertools;
+use vortex::buffer::Buffer;
 use vortex::dtype::DType;
 use vortex::dtype::Nullability;
 use vortex::error::VortexExpect;
 use vortex::error::VortexResult;
 use vortex::error::vortex_bail;
+use vortex::error::vortex_err;
 use vortex::expr::Expression;
 use vortex::expr::and_collect;
 use vortex::expr::get_item;
@@ -21,10 +24,13 @@ use vortex::scalar::Scalar;
 use vortex::scalar_fn::ScalarFnVTableExt;
 use vortex::scalar_fn::fns::binary::Binary;
 use vortex::scalar_fn::fns::operators::CompareOperator;
+use vortex::scan::selection::Selection;
 
 use crate::cpp::DUCKDB_VX_EXPR_TYPE;
+use crate::duckdb::ExtractedValue;
 use crate::duckdb::TableFilterClass;
 use crate::duckdb::TableFilterRef;
+use crate::duckdb::ValueRef;
 
 pub fn try_from_table_filter(
     value: &TableFilterRef,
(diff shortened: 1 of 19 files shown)


myrrc:
The disagreement point with @joseph-isaacs was about propagating partition range/selection to scan reader vs. not doing this.

Unfortunately, you can't do it other way right now because partition stream is non-deterministic and you can't filter based on the stream itself due to file-level pruning.

Iterating the stream twice is not an option either because there is some significant work (opening files) for each iteration, and since we don't use cached footers (we push them to cache but never read them) it's costly as well.

We've agreed on merging this change, and then I'd work on adding stable partition IDs so partition filtering and, consequently, row range could happen out of the Scan request.

vortex-duckdb/src/convert/table_filter.rs
```diff
@@ -125,3 +131,96 @@ pub fn try_from_table_filter(
...
+            };
+            Ok((Selection::All, range))
+        }
+        TableFilterClass::ConjunctionAnd(conj) => {
+            let mut start = 0u64;
+            let mut end = u64::MAX;
+            let mut indices: Option<Vec<u64>> = None;
+            for child in conj.children() {
+                let (sel, range) = try_from_virtual_column_filter(child)?;
+                if let Selection::IncludeByIndex(buf) = sel {
```
joseph-isaacs, line 204: shall we add a select intersection method?

vortex-duckdb/src/datasource.rs
```diff
@@ -67,25 +75,25 @@ use crate::duckdb::DataChunkRef;
...
-/// If DuckDB requests a zero-column projection from read_vortex like count(*),
-/// its planner tries to get any column:
-/// https://github.com/duckdb/duckdb/blob/dc11eadd8f0a7c600f0034810706605ebe10d5b9/src/planner/operator/logical_get.cpp#L149
-///
-/// If you define COLUMN_IDENTIFIER_EMPTY, planner takes it, otherwise the
-/// first column. As we don't want to fill the output chunk and we can leave
-/// it uninitialized in this case, we define COLUMN_IDENTIFIER_EMPTY as a
-/// virtual column.
-/// See virtual_columns in vortex-duckdb/cpp/table_function.cpp
-static EMPTY_COLUMN_IDX: u64 = 18446744073709551614;
```
joseph-isaacs, line 88: why did you remove this?

vortex-scan/src/lib.rs
```diff
@@ -169,6 +176,11 @@ pub trait Partition: 'static + Send {
     /// Downcast the partition to a concrete type.
     fn as_any(&self) -> &dyn Any;
 
+    /// Some unique identifier for partition if it's present.
+    /// Used mainly to filter out unused partitions with Duckdb's
+    /// late materialization support
+    fn index(&self) -> usize;
```
joseph-isaacs, line 190: I wonder if usize work well with serialisation. 

Let's say i want to reference globs by index, what is stopping the order changing or another file being added?

vortex-layout/src/scan/multi.rs
joseph-isaacs, line 270: what happens here when the if ready stream changes. this will change the order of the partitions. 

Let's say cause we cache a footer so it goes from deferred to open. 

I think we want a different index

vortex-duckdb/src/datasource.rs
myrrc: I've moved the comment into chunk export itself. We don't specifically handle empty column anymore.

vortex-scan/src/lib.rs
myrrc: We don't reference globs by index. Once the files are expanded, and the data source is created, the order stays the same, so the index is stable.

vortex-duckdb/src/convert/table_filter.rs
myrrc: A generic intersection is a better but harder solution. Currently we deal only with Selection::All and Selection::IncludeByIndex (and we know duckdb will only provide these to us) so I chose to implement only this intersection type.

vortex-layout/src/scan/multi.rs
myrrc: This is fine. If the ready stream changes, we will have to provide a different DataSource as partitions() consumes self.
If we decide to to any deferred-ready optimizations like re-using a footer cache, we will do it before call to partitions() in the DataSource.

joseph-isaacs (approved):

vortex-duckdb/src/datasource.rs
```diff
@@ -557,53 +610,96 @@ fn extract_schema_from_dtype(dtype: &DType) -> VortexResult<Vec<DuckdbField>> {
...
         .collect::<FieldNames>();
 
-    select(names, root())
+    // file_index column will be filled later when exporting the chunk.
+    let select = select(names, root());
+    let projection = if file_row_number_column_pos.is_some() {
+        // row_idx will be rearranged to correct position in scan(), prepend
+        // here
+        let row_idx = cast(row_idx(), DType::Primitive(PType::I64, false.into()));
+        let row_idx_struct = pack([("file_row_number", row_idx)], false.into());
```
joseph-isaacs, line 666: what happens if a file has this column name in it `file_row_number`.

PR #7724, opened 2026-04-29
Fix cosine similarity optimization bug

connortsui20:
## Summary

`CosineSimilarity` fast paths for `L2Denorm` used only the decoded normalized children, so lossy encodings could return a non-zero cosine for rows whose actual stored norm was 0.

This fix makes those fast paths also execute the stored norm children and return 0 when a denorm stored norm, or the plain-side norm in the one-denorm case, is zero while preserving the existing validity mask.

## Testing

Some regression tests.

### DIFF
--- a/vortex-tensor/src/scalar_fns/cosine_similarity.rs
+++ b/vortex-tensor/src/scalar_fns/cosine_similarity.rs
@@ -14,7 +14,6 @@ use vortex_array::arrays::ScalarFnArray;
 use vortex_array::arrays::scalar_fn::ScalarFnArrayView;
 use vortex_array::arrays::scalar_fn::plugin::ScalarFnArrayParts;
 use vortex_array::arrays::scalar_fn::plugin::ScalarFnArrayVTable;
-use vortex_array::builtins::ArrayBuiltins;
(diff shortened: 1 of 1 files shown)


vortex-tensor/src/scalar_fns/cosine_similarity.rs
```diff
@@ -244,22 +242,39 @@ impl CosineSimilarity {
...
+            let dots = dot.as_slice::<T>();
+            let norms_l = norms_l.as_slice::<T>();
+            let norms_r = norms_r.as_slice::<T>();
+            let buffer: Buffer<T> = (0..len)
+                .map(|i| {
+                    if norms_l[i] == T::zero() || norms_r[i] == T::zero() {
+                        T::zero()
+                    } else {
+                        dots[i]
+                    }
```
danking, line 271: definitely premature optimization, but I wonder if rustc is able to see that this can be written in a branchless way.

```rust
let either_is_zero = norms_l[i] == T::zero() || norms_r[i] == T::zero();
T::from(!either_is_zero) * dots[i]
```

danking (approved):

PR #7726, opened 2026-04-30
Round-trip extension dtypes through Arrow

palaska:
Vortex extension dtypes registered on a session now survive the Arrow boundary in both directions: schema-level (Schema / RecordBatch / nested fields) and leaf-array-level (`pa.ExtensionArray` over the C ABI).


vortex-array/src/arrow/convert.rs
```diff
@@ -380,23 +387,34 @@ fn remove_nulls(data: arrow_data::ArrayData) -> arrow_data::ArrayData {
 
 impl FromArrowArray<&ArrowStructArray> for ArrayRef {
     fn from_arrow(value: &ArrowStructArray, nullable: bool) -> VortexResult<Self> {
+        Self::from_arrow_with_session(value, nullable, &LEGACY_SESSION)
```
joseph-isaacs, line 390: can we deprecate this method

palaska: Do you mean the ones without session arg? Can do but the surface area is huge, it should probably be a separate mechanical PR

joseph-isaacs: Yep. could we do a PR which only does that and another (stacked/following) that does this semantic change?

vortex-array/src/arrow/executor/mod.rs
```diff
@@ -96,6 +99,15 @@ impl ArrowArrayExecutor for ArrayRef {
             None => preferred_arrow_type(&self)?,
         };
 
+        // Temporal extensions stay wrapped — `to_arrow_temporal` reads their metadata.
+        // Other extensions unwrap to storage; their identity lives on the Field.
+        if let DType::Extension(ext) = self.dtype()
+            && ext.metadata_opt::<AnyTemporal>().is_none()
+        {
+            let ext = self.execute::<ExtensionArray>(ctx)?;
+            return ext.storage_array().clone().execute_arrow(data_type, ctx);
```
joseph-isaacs, line 108: Does this impl all arrow ext type have the same layout as vortex ones

palaska: Extension is just a wrapper around storage types so yes, they will have the same layout. One difference is there is no extension DataType in arrow, extension types are just `Field` metadata. We just convert the storage type and attach the right metadata on the field when converting a vortex ext type to arrow.

joseph-isaacs: That is not what an extension type is

palaska:
Closing, see https://github.com/vortex-data/vortex/pull/7880

PR #7727, opened 2026-04-30
Expression pushdown for duckdb

### LINKED ISSUE #3964: Support table_function_pushdown_expression in DuckDB extension
This allows duckdb to ask us which expressions we support for push-down into a table filter.

Not sure how this interacts with supports_complex_pushdown

myrrc:
- Support expression pushdown similarly to how it's done for
filter pushdown.
- Support pushing down constants, comparisons, IN/NOT in, struct_extract, contains, prefix, suffix, and LIKE/NOT LIKE.

### DIFF
--- a/vortex-duckdb/cpp/table_function.cpp
+++ b/vortex-duckdb/cpp/table_function.cpp
@@ -15,6 +15,11 @@ DUCKDB_INCLUDES_BEGIN
 #include "duckdb/main/capi/capi_internal.hpp"
 #include "duckdb/main/connection.hpp"
 #include "duckdb/parser/parsed_data/create_table_function_info.hpp"
+#include "duckdb/planner/expression/bound_operator_expression.hpp"
+#include "duckdb/planner/expression/bound_comparison_expression.hpp"
+#include "duckdb/planner/expression/bound_between_expression.hpp"
+#include "duckdb/planner/expression/bound_conjunction_expression.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
 DUCKDB_INCLUDES_END
 
 using namespace duckdb;
@@ -263,11 +268,22 @@ void function(ClientContext &, TableFunctionInput &input, DataChunk &output) {
     }
 }
 
-void c_pushdown_complex_filter(ClientContext &,
-                               LogicalGet &,
-                               FunctionData *bind_data,
-                               vector<unique_ptr<Expression>> &filters) {
-    auto &bind = bind_data->Cast<CTableBindData>();
+using FilterVec = vector<unique_ptr<Expression>>;
+
+/*
+ * Table filter pushdown is used for two tasks in duckdb:
+ *
+ * 1. Prune files based on filename or hive partitioning, see Parquet
+ * filter pushdown. We don't use this because we do own file-level pruning in
+ * FileStatsLayoutReader, and we don't support hive partitioning yet.
+ *
+ * 2. Avoid reading unused file data. Filter expressions are pushed to Vortex,
+ * converted to Vortex expressions and used during the scan.
+ * Duckdb pushes a subset of expressions i.e. equality operators, and also
+ * expressions which return true in pushdown_expression.
+ */
+void pushdown_complex_filter(const FunctionData &bind_data, FilterVec &filters) {
+    const auto &bind = bind_data.Cast<CTableBindData>();
     void *const ffi_bind = bind.ffi_data->DataPtr();
     duckdb_vx_error error_out = nullptr;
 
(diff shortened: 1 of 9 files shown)


myrrc:
No visible impact

myrrc:
I've found the case where it has useful impact.
Consider this query:

```sql

COPY (SELECT '2020-01-01 02:20:30' ts) TO 'file.parquet';
EXPLAIN SELECT ts::TIMESTAMP FROM 'file.parquet' WHERE ts::DATE = '2020-01-01';

COPY (SELECT '2020-01-01 02:20:30' ts) TO 'file.vortex';
EXPLAIN SELECT ts::TIMESTAMP FROM 'file.vortex' WHERE ts::DATE = '2020-01-01';
```

Vortex plan:

```
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│   CAST(ts AS TIMESTAMP)   │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│           FILTER          │
│    ────────────────────   │
│ (CAST(ts AS DATE) = '2020 │
│       -01-01'::DATE)      │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│        READ_VORTEX        │
│    ────────────────────   │
│   Function: Vortex Scan   │
│      Projections: ts      │
│           ~1 row          │
└───────────────────────────┘
```

Timestamp casting is not propagated, duckdb applies one filter and one cast. You can't push down casted type into Vortex because there will be a conflict.

Parquet plan:

```
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│   CAST(ts AS TIMESTAMP)   │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│        PARQUET_SCAN       │
│    ────────────────────   │
│         Function:         │
│        PARQUET_SCAN       │
│                           │
│      Projections: ts      │
│                           │
│          Filters:         │
│ (CAST(ts AS DATE) = '2020 │
│       -01-01'::DATE)      │
│           ~1 row          │
└───────────────────────────┘
```

Parquet doesn't do type pushdown, but if it did, it could read casted ts directly, because second cast (as date) is pushed into filter expression, and projection is free to read whichever it wants.

This would allow us to decouple filter and projection casts.

joseph-isaacs:
That example is a little contrived but we should still of course support this

myrrc:
@joseph-isaacs as casting is a tricky topic, I'll split BOUND_CAST pushdown to a separate PR

vortex-duckdb/src/convert/table_filter.rs
```diff
@@ -123,8 +123,10 @@ pub fn try_from_table_filter(
             )
         }
         TableFilterClass::ExpressionRef(expr) => {
-            // TODO(ngates): figure out which column ID DuckDB is using for the expression.
-            vortex_bail!("expression table filter is not supported: {}", expr);
+            match super::expr::try_from_bound_expression_with_col_sub(expr, col)? {
```
joseph-isaacs: import please

vortex-duckdb/src/convert/expr.rs
```diff
@@ -33,20 +34,89 @@ use vortex::scalar_fn::fns::operators::Operator;
...
+                case_insensitive: false,
+            };
+            Like.new_expr(opts, [string, target])
+        }
+        matchers @ ("contains" | "prefix" | "suffix") => {
+            let children: Vec<_> = func.children().collect();
+            vortex_ensure!(children.len() == 2);
+            let Some(value) = try_from_expression_inner(children[0], col_sub)? else {
+                return Ok(None);
+            };
```
joseph-isaacs, line 88: where these not being pushed down before?

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -263,11 +268,92 @@ void function(ClientContext &, TableFunctionInput &input, DataChunk &output) {
...
+
+        for (const auto &child : expr.Cast<BoundOperatorExpression>().children) {
+            if (!pushdown_expression(*child)) {
+                return false;
+            }
+        }
+        return true;
+    }
+    default:
+        return false;
```
joseph-isaacs: put this into rust

joseph-isaacs (commented):
Lets change the C++ and this looks good

vortex-duckdb/src/convert/table_filter.rs
myrrc: +1

vortex-duckdb/src/convert/expr.rs
myrrc: These were pushed as part of complex filter but not table filters.

vortex-duckdb/cpp/table_function.cpp
myrrc: We can't do this in Rust unfortunately, because pushdown_expression callback doesn't give us FunctionData.

myrrc: Fixed with an ugly extern exposure.

joseph-isaacs (approved):

PR #7728, opened 2026-04-30
skip: remove parameter names from public api lock files

joseph-isaacs:
The names should append the public api signature.


AdamGS:
can we just remove them? I don't think we actually look at them, and we don't use them to actually communicate API changes in the changelog/release notes.

joseph-isaacs:
I agree with this, do we have a plan to make these useful? @robert3005 @gatesn

AdamGS (approved):

PR #7729, opened 2026-04-30
skip[fuzz]: allow generation of longer arrays 

joseph-isaacs:
Allow the fuzzer to create longer arrays to exercise more file_io logic (e.g. zone_map partitions).

Adds a new trait `ArbitraryWith` and `ArbitraryArrayConfig` to express this.

```rust
/// Trait for generating arbitrary values with a caller-provided configuration.
pub trait ArbitraryWith<'a, C>: Sized {
    /// Generate an arbitrary value using the provided configuration.
    fn arbitrary_with_config(u: &mut Unstructured<'a>, config: &C) -> Result<Self>;
}

/// Configuration for arbitrary array generation.
#[derive(Clone, Debug)]
pub struct ArbitraryArrayConfig {
    /// Fixed dtype, or `None` to generate one from [`Unstructured`].
    pub dtype: Option<DType>,
    /// Inclusive range for the total array length.
    pub len: RangeInclusive<usize>,
}
```

### DIFF
--- a/encodings/runend/src/arbitrary.rs
+++ b/encodings/runend/src/arbitrary.rs
@@ -9,6 +9,7 @@ use vortex_array::LEGACY_SESSION;
 use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::arrays::arbitrary::ArbitraryArray;
+use vortex_array::arrays::arbitrary::ArbitraryArrayConfig;
 use vortex_array::arrays::arbitrary::ArbitraryWith;
 use vortex_array::dtype::DType;
 use vortex_array::dtype::Nullability;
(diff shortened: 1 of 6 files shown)


encodings/runend/src/arbitrary.rs
```diff
@@ -46,14 +47,28 @@ impl ArbitraryRunEndArray {
...
             let runend_array = RunEnd::try_new(ends, values, &mut ctx)
                 .vortex_expect("Empty RunEndArray creation should succeed");
             return Ok(ArbitraryRunEndArray(runend_array));
         }
 
         // Generate arbitrary values for each run
-        let values = ArbitraryArray::arbitrary_with(u, Some(num_runs), dtype)?.0;
+        let values = ArbitraryArray::arbitrary_with_config(
+            u,
+            &vortex_array::arrays::arbitrary::ArbitraryArrayConfig {
```
robert3005: can you import this?

robert3005 (approved):

PR #7730, opened 2026-04-30
Actually wire the pluggable expression convertor

### LINKED ISSUE #7731: Custom ExpressionConvertor is not propagated from VortexSource to VortexOpener
### What happened?

  `VortexSource` exposes `with_expression_convertor(...)`, and `try_pushdown_filters` uses `self.expression_convertor` to decide whether a
  DataFusion physical expression can be pushed down.

  However, when `VortexSource::create_file_opener` constructs `VortexOpener`, it appears to pass a fresh `DefaultExpressionConvertor` instead of
  the converter stored on `VortexSource`:

```rust
expression_convertor: Arc::new(DefaultExpressionConvertor::default()),
```

  I expected the opener to receive the same converter configured on the source, e.g.:

```rust
expression_convertor: Arc::clone(&self.expression_convertor),
```

Otherwise a custom converter can mark a custom scalar UDF/predicate as pushdown-supported at the VortexSource level, but the actual opener/scan path falls back to the default converter when converting the pushed predicate/projection. 

That means custom scalar UDF pushdown may not survive into the file-opening / scan-building phase.

This seems especially relevant after #5853, which made expression conversion pluggable.

### Steps to reproduce

  1. Implement a custom ExpressionConvertor that supports a custom DataFusion scalar UDF or any custom PhysicalExpr.

  2. Use it with:
       let source = VortexSource::new(table_schema, session)
          .with_expression_convertor(Arc::new(MyExpressionConvertor));

  3. Push down a filter/projection that only MyExpressionConvertor supports.
  4. Observe that VortexSource::try_pushdown_filters consults the custom converter, but VortexSource::create_file_opener constructs VortexOpener
     with DefaultExpressionConvertor.
  5. The opener then uses the default converter in its projection/predicate conversion path, so the custom expression is not handled.

### Environment

  - Vortex branch: develop
  - Area: vortex-datafusion
  - Related PR: #5853

### Additional context

Potential fix may be as small as changing VortexSource::create_file_opener to pass through the configured converter:

  expression_convertor: Arc::clone(&self.expression_convertor)

AdamGS:
## Summary

- Closes https://github.com/vortex-data/vortex/issues/7731

Turns out I didn't wire the expression convertor extension point correctly. This PR both fixes that AND adds a test to make sure this behavior is maintained.

### DIFF
--- a/vortex-datafusion/src/persistent/source.rs
+++ b/vortex-datafusion/src/persistent/source.rs
@@ -301,15 +301,13 @@ impl VortexSource {
         self.options = opts;
         self
     }
-}
 
-impl FileSource for VortexSource {
-    fn create_file_opener(
+    fn create_vortex_opener(
         &self,
         object_store: Arc<dyn ObjectStore>,
         base_config: &FileScanConfig,
         partition: usize,
-    ) -> DFResult<Arc<dyn FileOpener>> {
+    ) -> DFResult<VortexOpener> {
         let batch_size = self
             .batch_size
             .vortex_expect("batch_size must be supplied to VortexSource");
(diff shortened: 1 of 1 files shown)


vortex-datafusion/src/persistent/source.rs
```diff
@@ -301,15 +301,13 @@ impl VortexSource {
         self.options = opts;
         self
     }
-}
 
-impl FileSource for VortexSource {
-    fn create_file_opener(
+    fn create_vortex_opener(
```
AdamGS, line 305: introduce new private function because currently `FileSource` can't be downcasted, so we want to return a real type in tests.

robert3005 (approved):

PR #7740, opened 2026-04-30
reenable LTO for release builds and benchmarks

myrrc:


0ax1 (changes requested):
The rationale from prev discussions was that our perf should not rely on LTO. We can't make assumption about other crates that import Vortex whether LTO is enabled. Story would be diff if Vortex would be the executable top-level binary. Should we leave perf on the table due to missing cross crate opt, we should scan our APIs for missing inline annotations which also benefit clients that don't use LTO.

AdamGS:
last time I checked LTO caused some regression for datafusion, glad to see that doesn't seem to be the case (maybe it was a different level of LTO though?), glad to see that's not the case anymore.

0ax1:
We agreed on enabling LTO for the DuckDB extension build but not turn this on by default in Vortex. @myrrc Can we close this PR then?

PR #7741, opened 2026-04-30
feat: teach ArrayStream instrument(span)

danking:


vortex-array/src/stream/instrumented.rs
```diff
@@ -0,0 +1,38 @@
...
+    fn dtype(&self) -> &DType {
+        self.stream.inner().dtype()
+    }
+}
+
+impl<S: Stream> Stream for InstrumentedArrayStream<S> {
+    type Item = S::Item;
+
+    fn poll_next(
+        self: std::pin::Pin<&mut Self>,
```
a10y: Can probably import this

a10y (approved):

vortex-array/src/stream/instrumented.rs
danking: Done.

a10y:
you might need the ol' `./scripts/public-api.sh`

robert3005 (approved):

PR #7743, opened 2026-04-30
Array operations fuzzer operates on smaller arrays

robert3005:
I believe some of the reference implementations are very slow so we can't
operate on bigger arrays


joseph-isaacs (approved):
We can do this. But we do really want the 16k size going forwards

robert3005:
I kept the file fuzzer working on 16k arrays since that looked fine. I think we need to spend some time on reference impls in the fuzzer

PR #7744, opened 2026-05-01
Improve intersect_by_rank performance

robert3005:
Intersect by rank is an operation for performing filter on top of another filter

This pr improves
Dense on Dense masks: 10-18x
Sparse on Dense masks: 8x
Sparse on Sparse masks: 2x

### DIFF
--- a/vortex-buffer/src/bit/buf_mut.rs
+++ b/vortex-buffer/src/bit/buf_mut.rs
@@ -165,11 +165,21 @@ impl BitBufferMut {
     }
 
     /// Create a bit buffer of `len` with `indices` set as true.
-    pub fn from_indices(len: usize, indices: &[usize]) -> BitBufferMut {
-        let mut buf = BitBufferMut::new_unset(len);
-        // TODO(ngates): for dense indices, we can do better by collecting into u64s.
-        indices.iter().for_each(|&idx| buf.set(idx));
-        buf
+    pub fn from_indices(len: usize, indices: impl IntoIterator<Item = usize>) -> BitBufferMut {
+        let mut buffer = BufferMut::<u64>::zeroed(len.div_ceil(64));
+        for idx in indices {
+            assert!(idx < len, "index {idx} exceeds len {len}");
+            buffer.as_mut_slice()[idx / 64] |= 1 << (idx % 64);
+        }
+
+        let mut buffer = buffer.into_byte_buffer();
+        buffer.truncate(len.div_ceil(8));
+
+        Self {
+            buffer,
+            offset: 0,
+            len,
+        }
     }
 
     /// Invokes `f` with indexes `0..len` collecting the boolean results into a new `BitBufferMut`
--- a/vortex-mask/src/intersect_by_rank.rs
+++ b/vortex-mask/src/intersect_by_rank.rs
@@ -429,6 +429,21 @@ where
     intersect_mask_driven::<Portable, _>(self_buffer, mask_indices, true_count)
 }
 
+/// Check if a mask is sparse.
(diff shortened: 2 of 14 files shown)


robert3005:
I will go over this tomorrow @joseph-isaacs I looked at #7098 and #7393 which both optimised slightly different cases of this function. I tried to combine the two.

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -165,11 +165,21 @@ impl BitBufferMut {
     }
 
     /// Create a bit buffer of `len` with `indices` set as true.
-    pub fn from_indices(len: usize, indices: &[usize]) -> BitBufferMut {
-        let mut buf = BitBufferMut::new_unset(len);
-        // TODO(ngates): for dense indices, we can do better by collecting into u64s.
-        indices.iter().for_each(|&idx| buf.set(idx));
-        buf
+    pub fn from_indices(len: usize, indices: impl IntoIterator<Item = usize>) -> BitBufferMut {
+        let mut buffer = BufferMut::<u64>::zeroed(len.div_ceil(64));
```
joseph-isaacs, line 169: add a todo to add a sized iterator variant, then you can switch over the ratio of len to indices

vortex-mask/src/intersect_by_rank.rs
```diff
@@ -29,22 +465,70 @@ impl Mask {
...
+            (Self::AllFalse(_), _) | (_, Self::AllFalse(_)) => Self::new_false(self.len()),
+            (Self::Values(self_values), Self::Values(mask_values)) => {
+                let self_is_very_sparse = self_values.true_count() < self.len().div_ceil(64);
+                // The mask-driven path becomes worthwhile around ~3% mask density: each set
+                // bit costs a select + push, but we save a per-self-chunk popcount + deposit.
+                let mask_is_very_sparse = mask_values.true_count().saturating_mul(32) < mask.len();
+
+                if let Some(mask_indices) = mask_values.indices.get() {
+                    if let Some(self_indices) = self_values.indices.get()
+                        && mask_indices.len() < self.len().div_ceil(64)
```
joseph-isaacs, line 482: do this first

joseph-isaacs (approved):

joseph-isaacs:
Add benchmark numbers to the desc

vortex-mask/src/intersect_by_rank.rs
```diff
@@ -29,22 +465,70 @@ impl Mask {
...
-                        .collect(),
+        match (self, mask) {
+            (Self::AllTrue(_), _) => mask.clone(),
+            (_, Self::AllTrue(_)) => self.clone(),
+            (Self::AllFalse(_), _) | (_, Self::AllFalse(_)) => Self::new_false(self.len()),
+            (Self::Values(self_values), Self::Values(mask_values)) => {
+                let self_is_very_sparse = self_values.true_count() < self.len().div_ceil(64);
+                // The mask-driven path becomes worthwhile around ~3% mask density: each set
+                // bit costs a select + push, but we save a per-self-chunk popcount + deposit.
+                let mask_is_very_sparse = mask_values.true_count().saturating_mul(32) < mask.len();
```
joseph-isaacs: can you make this function, and make it clear this is a approx threshold that might be configurable in the future

vortex-mask/src/intersect_by_rank.rs
robert3005: This is one special case where the indices are cached on the mask, we want to avoid asking for mask_indices if they aren't cached

PR #7747, opened 2026-05-01
`-isystem` for Duckdb headers, build with `-Werror`

myrrc:
We want all warnings as errors for our code, not for duckdb one.


vortex-duckdb/cpp/CMakeLists.txt
```diff
@@ -23,19 +22,19 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 # Enable compiler warnings (matching build.rs flags).
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -Werror")
+set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Wall -Wextra -Wpedantic -Werror -O3")
```
0ax1: Why change the release flags in this PR? Also should release just expand on default flags? -O0 will be overridden by -O3 etc.

vortex-duckdb/build.rs
```diff
@@ -327,10 +328,14 @@ fn c2rust(crate_dir: &Path, duckdb_include_dir: &Path) {
     }
 }
 
-fn cpp(duckdb_include_dir: &Path) {
+fn cpp(duckdb_include_dir: &Path, build_type: &str) {
+    let mut flags = vec!["-Wall", "-Wextra", "-Wpedantic", "-Werror"];
+    if build_type == "release" {
+        flags.extend_from_slice(&["-flto=auto"]);
```
0ax1: Can we split out the LTO changes from this PR? Does enabling LTO for the C++ code in vortex-duckdb even have an effect on perf? Other than that, my idea was that we enable LTO in the ext repo but not in the vortex repo itself.

0ax1: auto means full lto for clang which is very slow. we should try to use thin for both gcc and clang.

vortex-duckdb/build.rs
```diff
@@ -391,8 +391,12 @@ fn main() {
 
     let crate_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
     let duckdb_dir = crate_dir.join("duckdb");
-    let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
```
0ax1, line 417: `library_dir` no longer depends on the build mode. If we build in debug with `VX_DUCKDB_DEBUG=1`, it ends up in the same path as the pre-built downloaded dylib. So if we build with `VX_DUCKDB_DEBUG=1` in debug, and without `VX_DUCKDB_DEBUG=1` in release after, the build.rs will treat the existing `libduckdb.*` in `library_dir` as a cache hit. We then refer to a stale artifact.

vortex-duckdb/build.rs
```diff
@@ -330,7 +330,7 @@ fn c2rust(crate_dir: &Path, duckdb_include_dir: &Path) {
 fn cpp(duckdb_include_dir: &Path) {
     cc::Build::new()
         .std("c++20")
-        .flags(["-Wall", "-Wextra", "-Wpedantic"])
+        .flags(["-Wall", "-Wextra", "-Wpedantic", "-Werror"])
```
0ax1, line 356: Are we sure this is safe to do at this point? I don't think we can rely on the following covering all cases.
```
    _Pragma("GCC diagnostic push")                                             \
    _Pragma("GCC diagnostic ignored \"-Wall\"")                                \
    _Pragma("GCC diagnostic ignored \"-Wextra\"")                              \
    _Pragma("GCC diagnostic ignored \"-Wpedantic\"")                           \
    _Pragma("GCC diagnostic ignored \"-Wunused-parameter\"")                   \
    _Pragma("GCC diagnostic ignored \"-Wtype-limits\"")
```
The issue is that DDB has def in header files that we don't control. I would have used `Weverything` to ignore all warnings but it doesn't exist for GCC.

vortex-duckdb/build.rs
myrrc: +1

vortex-duckdb/build.rs
myrrc: Yeah, I've manually re-built duckdb and it doesn't trigger any warnings.

0ax1: If the DuckDB source changes or someone uses a compiler with a different set of warnings, the compilation will error out. My take is we should not add `Werror` here to let builds complete regardless, as we don't have control over the DuckDB header def.

myrrc: We do have control over it because we pin duckdb to a specific version. The only way how the source code may change is us manually upgrading duckdb.

vortex-duckdb/build.rs
0ax1: Moving the extracted stuff out of the out dir is nice for caching but also means `cargo clean` won't be able wipe a broken checkout, e.g. if a user hits ctrl-c halfway through when the duckdb files get extracted.

vortex-duckdb/build.rs
0ax1: We don't control the compiler chain of all clients.

vortex-duckdb/build.rs
```diff
@@ -355,7 +355,9 @@ fn cpp(duckdb_include_dir: &Path) {
         .std("c++20")
         .flags(["-Wall", "-Wextra", "-Wpedantic"])
         .cpp(true)
-        .include(duckdb_include_dir)
+        // We don't want compiler warnings inside duckdb headers, pass as flags
+        .flag("-isystem")
+        .flag(duckdb_include_dir)
         .include("include")
         .include("cpp/include")
```
0ax1, line 362: Build with `Werror` here too. The CMake were we added it to is only for IDE integration with JetBrains.

0ax1: I guess the super clean solution would be to generate the CMake IDE file from build.rs and make it transient.

myrrc: +1

0ax1 (approved):

PR #7751, opened 2026-05-01
Allow running reduce_parent operations on stack allocated parents

robert3005:
Add indirection logic to support stack allocated array parents in reduce rules.

This lets us avoid heap allocation for commonly used operations that are often
immediately optimised away

### DIFF
--- a/encodings/fastlanes/src/bitpacking/vtable/operations.rs
+++ b/encodings/fastlanes/src/bitpacking/vtable/operations.rs
@@ -34,6 +34,7 @@ mod test {
 
     use vortex_array::ArrayRef;
     use vortex_array::IntoArray;
+    use vortex_array::ParentRef;
     use vortex_array::VortexSessionExecute;
     use vortex_array::arrays::PrimitiveArray;
     use vortex_array::arrays::SliceArray;
@@ -63,9 +64,9 @@ mod test {
 
     fn slice_via_reduce(array: &BitPackedArray, range: Range<usize>) -> BitPackedArray {
         let array_ref = array.clone().into_array();
-        let slice_array = SliceArray::new(array_ref.clone(), range);
+        let slice_array = SliceArray::new(array_ref.clone(), range).into_array();
         let sliced = array_ref
-            .reduce_parent(&slice_array.into_array(), 0)
+            .reduce_parent(&ParentRef::from_array_ref(&slice_array), 0)
             .expect("execute_parent failed")
             .expect("expected slice kernel to execute");
         sliced.as_::<BitPacked>().into_owned()
--- a/encodings/fastlanes/src/bitpacking/vtable/operations.rs
+++ b/encodings/fastlanes/src/bitpacking/vtable/operations.rs
@@ -64,9 +64,9 @@ mod test {
 
     fn slice_via_reduce(array: &BitPackedArray, range: Range<usize>) -> BitPackedArray {
         let array_ref = array.clone().into_array();
-        let slice_array = SliceArray::new(array_ref.clone(), range).into_array();
+        let slice_array = SliceArray::new(array_ref.clone(), range);
         let sliced = array_ref
-            .reduce_parent(&ParentRef::from_array_ref(&slice_array), 0)
+            .reduce_parent(&ParentRef::from_array(&slice_array), 0)
             .expect("execute_parent failed")
             .expect("expected slice kernel to execute");
         sliced.as_::<BitPacked>().into_owned()
--- a/vortex-array/src/array/erased.rs
+++ b/vortex-array/src/array/erased.rs
@@ -92,6 +92,7 @@ impl ArrayRef {
         &self.0.data
     }
 
+    #[allow(clippy::inline_always)]
     #[inline(always)]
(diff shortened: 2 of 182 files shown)


vortex-array/src/array/erased.rs
```diff
@@ -200,16 +207,26 @@ impl ArrayRef {
 
     /// Wraps the array in a [`FilterArray`] such that it is logically filtered by the given mask.
     pub fn filter(&self, mask: Mask) -> VortexResult<ArrayRef> {
-        FilterArray::try_new(self.clone(), mask)?
+        let dtype = self.dtype().clone();
+        let len = mask.true_count();
+        let data = FilterData::try_new(self.len(), mask)?;
+        ArrayParts::new(Filter, dtype, len, data)
+            .with_slots(vec![Some(self.clone())])
+            .optimize()?
```
joseph-isaacs: can you revert this?

vortex-array/src/matcher.rs
```diff
@@ -14,6 +15,21 @@ pub trait Matcher {
 
     /// Try to match the given array, returning the matched view type if successful.
     fn try_match<'a>(array: &'a ArrayRef) -> Option<Self::Match<'a>>;
```
joseph-isaacs: deprecated doc str

vortex-array/src/array/parent.rs
```diff
@@ -0,0 +1,225 @@
...
+/// available so matchers and rule bodies can dispatch without distinguishing the two
+/// cases.
+#[derive(Clone, Copy)]
+pub struct ParentRef<'a> {
+    array_ref: Option<&'a ArrayRef>,
+    encoding_id: ArrayId,
+    dtype: &'a DType,
+    len: usize,
+    data: &'a dyn Any,
+    slots: &'a [Option<ArrayRef>],
```
joseph-isaacs, line 49: This should be a struct View, not sure on the name

vortex-array/src/array/typed.rs
```diff
@@ -52,6 +57,118 @@ impl<V: VTable> ArrayParts<V> {
...
+    /// Try to apply parent-reduction rules using both static rules and any
+    /// session-registered [`ArrayKernels`](crate::optimizer::kernels::ArrayKernels).
+    ///
+    /// See [`ArrayParts::optimize`] for the static-only variant.
+    pub fn optimize_ctx(self, session: &VortexSession) -> VortexResult<Optimized<V>> {
+        self.optimize_inner(Some(session))
+    }
+
+    fn optimize_inner(self, session: Option<&VortexSession>) -> VortexResult<Optimized<V>> {
+        // Validate without materializing — `try_from_parts` would allocate, so we
```
joseph-isaacs: take &self and return Option<V>

encodings/fastlanes/src/bitpacking/vtable/operations.rs
```diff
@@ -63,9 +64,9 @@ mod test {
 
     fn slice_via_reduce(array: &BitPackedArray, range: Range<usize>) -> BitPackedArray {
         let array_ref = array.clone().into_array();
-        let slice_array = SliceArray::new(array_ref.clone(), range);
+        let slice_array = SliceArray::new(array_ref.clone(), range).into_array();
```
joseph-isaacs: remove the into_arrow

vortex-array/src/array/erased.rs
robert3005: we cannot, FilterArray has an Arc of ArrayParts

robert3005: we need to generate new methods

vortex-array/src/array/typed.rs
```diff
@@ -52,6 +57,82 @@ impl<V: VTable> ArrayParts<V> {
...
+            let Some(child) = slot else { continue };
+
+            // Session kernels take precedence over static `PARENT_RULES`, matching
+            // the existing optimizer's ordering.
+            if let Some(kernels) = &kernels
+                && let Some(plugins) = kernels.find_reduce_parent(parent_id, child.encoding_id())
+            {
+                for plugin in plugins.as_ref() {
+                    if let Some(reduced) = plugin(child, &parent_ref, slot_idx)? {
+                        return Ok(Some(cascade(reduced, session)?));
```
robert3005: we should check here if slot_idx equals slots length and not call cascade if it is

robert3005: We cannot do that, the regular optimise loop resets if optimisations are applied therefore we MUST cascade since there might be more optimistaions applied to the new array after reduction

joseph-isaacs:
Some of those benchmarks are doing extra real work

robert3005:
This is not ready yet, there’s still something wrong with matchers

AdamGS:
What's the state of this PR?

robert3005:
I think this pr is in the right shape. I need to go over it again to see if anything is missing. But realistically it's ready for review

robert3005:
@AdamGS @joseph-isaacs this is ready for review. Since this touches every reduce rule we should make sure we are happy with the new api

PR #7753, opened 2026-05-01
Cache min/max indices in patches

palaska:
Cache `min`/`max` patch indices in `Patches` and short-circuit `search_index` when the query falls outside that range. Speeds up ALP/BitPacked `scalar_at`. This should help with slicing any patched array into a region that doesn't overlap the patch range

Screenshot 2026-05-01 at 16 26 04


joseph-isaacs (changes requested):
Indices arrays have min/max stats these should be used instead.

We might not currently save or use this, but by these benchmarks we should.

If these benchmarks are cleaned up I am happy to merge those first.

joseph-isaacs:
See: https://github.com/vortex-data/vortex/blob/e03bd1dcc63a34f260a16c05395146e26d1c9df3/docs/developer-guide/benchmarking.md

robert3005 (commented):
Meta point, this is why you want to have special logic that's shared across all arrays, if we have Patched array then that could be one place where all of this logic could live without hacks. Alas we need Patched array to get over the line

palaska:
Stats lookups (rwlock + hashmap + scalar conversion) were too expensive to be in the hot path (search_index is called in tight loops) so I still went with a OnceLock approach, combined with stats.

palaska:
I think RwLock -> ArcSwap change made setting stats more expensive..

joseph-isaacs:
It does look that way. I wonder if it made reading them faster?

robert3005:
Github is failing me - ~remove extra Arc around arcswap~ tihs is not going to work, arcswap isn't clone

PR #7757, opened 2026-05-02
[WIP] some io stuff

AdamGS:
## Summary

Closes: #000

### DIFF
--- a/vortex-layout/src/scan/tasks.rs
+++ b/vortex-layout/src/scan/tasks.rs
@@ -136,9 +136,11 @@ pub fn split_exec<A: 'static + Send>(
         }
     };
 
-    // Step 4: execute the projection only if the filter kept rows.
-    let reader = Arc::clone(&ctx.reader);
-    let projection = ctx.projection.clone();
+    // Step 4: execute the projection, only at the mask for rows which match the filter.
+    let projection_future =
+        ctx.reader
+            .projection_evaluation(&row_range, &ctx.projection, filter_mask.clone())?;
+
     let mapper = Arc::clone(&ctx.mapper);
     let array_fut = async move {
         let mask = filter_mask.await?;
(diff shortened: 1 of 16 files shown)


vortex-layout/src/scan/tasks.rs
```diff
@@ -136,18 +136,18 @@ pub fn split_exec<A: 'static + Send>(
...
     };
 
-    // Step 4: execute the projection, only at the mask for rows which match the filter
-    let projection_future =
-        ctx.reader
-            .projection_evaluation(&row_range, &ctx.projection, filter_mask.clone())?;
-
+    // Step 4: execute the projection only if the filter kept rows.
+    let reader = Arc::clone(&ctx.reader);
+    let projection = ctx.projection.clone();
```
a10y: 👍

a10y (approved):
looks good, definitely the change to task execution is good, i'm a bit confused where we actually cancel requests now?

vortex-layout/src/scan/tasks.rs
```diff
@@ -136,18 +136,18 @@ pub fn split_exec<A: 'static + Send>(
...
     let mapper = Arc::clone(&ctx.mapper);
     let array_fut = async move {
         let mask = filter_mask.await?;
         if mask.all_false() {
             return Ok(None);
         }
 
+        let projection_future =
+            reader.projection_evaluation(&row_range, &projection, MaskFuture::ready(mask))?;
```
a10y: actually, do we want to do this? different layouts should handle projection with empty Mask efficiently on their own I would hope? plus now we defer creating the projection_evaluation which kills some of the pre-fetching right?

a10y (changes requested):

a10y:
I think this directionally makes sense, just want to make sure we wire in cancelation. This will probably fix a lot of issues we've had reported about Vortex doing a surprising amount of IO against S3.

vortex-layout/src/scan/tasks.rs
AdamGS: I don't think they actually do? For example, in flat layout we do:
```rust
let mut array = array.clone().await?;
let mask = mask.await?;
```

PR #7759, opened 2026-05-02
Refactor DictEncoder to not use ArrayAccessors

robert3005:
ArrayAccessor is not the fastest api and we can do better by accessing the value iterators directly


robert3005:
Need to figure out what to do about WASM here, cloudflare crate requires 64 bit pointers

robert3005:
What do you think about this one @gatesn. We can change the estimator to be datasketches-rust. The two libraries I used optimise the small case a bit better.

robert3005:
I gave up on trying to use cardinality estimator for dict compression. I have kept whatever was useful from that pr and maybe will attempt again in the future

vortex-compressor/benches/dict_encode.rs
```diff
@@ -39,16 +33,16 @@ fn make_array() -> PrimitiveArray {
 #[divan::bench]
 fn encode_generic(bencher: Bencher) {
     let array = make_array().into_array();
-    bencher
-        .with_inputs(|| (&array, SESSION.create_execution_ctx()))
-        .bench_refs(|(array, ctx)| dict_encode(array, ctx).unwrap());
+    bencher.with_inputs(|| &array).bench_refs(|array| {
+        dict_encode(array, &mut LEGACY_SESSION.create_execution_ctx()).unwrap()
```
myrrc: Do we want to use LEGACY_SESSION?

vortex-array/src/builders/dict/primitive.rs
```diff
@@ -123,21 +139,43 @@ where
...
+            Mask::AllFalse(_) => {
+                if let Some(code) = self.encode_null() {
+                    unsafe { codes.push_n_unchecked(code, array.len()) }
+                }
+            }
+            Mask::Values(v) => {
+                let bit_buff = v.bit_buffer();
+                for (&value, valid) in prim.as_slice::<T>().iter().zip(bit_buff) {
+                    if !valid {
+                        let Some(code) = self.encode_null() else {
```
myrrc, line 168: This should be a vortex_ensure!()? If we find a non-valid item without a NULL code, something is broken

vortex-array/src/builders/dict/primitive.rs
```diff
@@ -123,21 +139,43 @@ where
...
+                            break;
+                        };
+                        unsafe { codes.push_unchecked(code) }
+                    } else {
+                        let Some(code) = self.encode_value(value) else {
```
myrrc, line 173: Same as above?

vortex-array/src/builders/dict/bytes.rs
```diff
@@ -71,18 +80,16 @@ impl<Code: UnsignedPType> BytesDictBuilder<Code> {
...
-    fn lookup_bytes(&self, idx: usize) -> Option<&[u8]> {
-        self.values_nulls.value(idx).then(|| {
-            let bin_view = &self.views[idx];
-            if bin_view.is_inlined() {
-                bin_view.as_inlined().value()
-            } else {
-                &self.values[bin_view.as_view().as_range()]
-            }
-        })
+    fn lookup_bytes(&self, idx: usize) -> &[u8] {
```
myrrc, line 83: Do we guarantee now to invoke this only on valid items? Worth adding a comment and/or a `debug_assert!(self.values_nulls.value(idx).is_none())`

myrrc (approved):
LGTM

vortex-compressor/benches/dict_encode.rs
robert3005: Noo, I must have messed up

vortex-array/src/builders/dict/primitive.rs
robert3005: the thing here is we overflew the encoder limit so we want to terminate encoding and have caller start a new one

vortex-array/src/builders/dict/bytes.rs
robert3005: This is only looking up in the existing view array so by definition this only ever holds valid items

vortex-array/src/builders/dict/primitive.rs
myrrc: +1. Can you add a comment explaining that? Not obvious from the code?

encodings/sequence/src/compress.rs
```diff
@@ -150,8 +150,6 @@ fn encode_primitive_array<P: NativePType + Into<PValue> + CheckedAdd + CheckedSu
 mod tests {
     use std::sync::LazyLock;
 
-    #[expect(unused_imports)]
```
AdamGS, line 153: 😢

AdamGS (approved):

Issue #6538, opened 2026-02-16
CUDA backlog


a10y:
Potential items

- [ ] JIT pipeline kernel
- [ ] GSST

??

myrrc:
Duplicate of #6535

PR #7776, opened 2026-05-04
feat[vortex-cuda]: GPU FSST decompression kernel

asubiotto:
## Summary

This commit implements on-GPU decompression of the existing FSST encoding. This kernel achieves ~42% max throughput utilization as compared to the `throughput_cuda` benchmark on a DGX spark. CPU work is required to compute the output offsets.

The core performance win is buffering up to 24 bytes of decompressed data in three u64 registers and emitting the widest aligned stores possible up to u128 (st.global.v2.u64).

The 256-entry symbol table (≤ 2 KB) is read directly from global memory. Staging it into shared memory measured ~3% slower at 10M rows and ~15% slower at 1M rows. The hypothesis is that L1 already holds the table after a few iterations and the explicit shared copy adds bank-conflict latency on the warp-divergent symbols[code] reads; the gap is wider at 1M because the kernel is less bandwidth-bound there.

Further optimizations would require an encoding change. Splits-style intra-string parallelism (one GPU thread per ~32-compressed-byte chunk instead of per-string) was prototyped on top of this kernel and measured an additional +30% kernel throughput at 1M clickbench URLs, +26% at 5M, +12% at 10M.

Four kernel variants are generated for the unsigned widths of codes_offsets (u8/u16/u32/u64); signed integer ptypes are reinterpreted as their unsigned equivalent on the Rust side, so the bit pattern is preserved without copying.

Addresses: #6538 

## Testing

Unit tests against the CPU implementation on small and larger dataset.


vortex-cuda/src/kernel/encodings/fsst.rs
```diff
@@ -0,0 +1,267 @@
...
+                    dtype,
+                    validity,
+                )
+            };
+            return Ok(Canonical::VarBinView(empty));
+        }
+
+        // Partial-null arrays fall back to CPU.
+        if !matches!(validity, Validity::NonNullable | Validity::AllValid) {
+            vortex_bail!("CUDA FSST decode does not yet support arrays with nulls");
```
asubiotto: I don't know about this. We can probably still use the GPU kernel for mixed nullability as long as there are no escape codes in the null positions. In general, I think this holds true, but I preferred to cautiously reject any nulls. Maybe we can do a CPU verification pass?

a10y: the on-device validity handling should be easy right? `if (!validity[sid]) return;` ?

vortex-cuda/kernels/src/fsst.cu
```diff
@@ -0,0 +1,190 @@
...
+            scratch_mid = (scratch_mid >> 8) | (scratch_high << 56);
+            scratch_high >>= 8;
+            out_pos += 1;
+            scratch_bytes -= 1;
+        }
+    };
+
+    while (in_pos < in_end) {
+        // Drain to scratch_bytes < 16 so the next ≤8-byte symbol fits.
+        while (scratch_bytes >= 16) {
```
a10y: this can be `>` right?

vortex-cuda/kernels/src/fsst.cu
```diff
@@ -0,0 +1,190 @@
...
+                                          const uint8_t *__restrict symbol_lengths,
+                                          const uint32_t *__restrict output_offsets,
+                                          uint8_t *__restrict output_bytes,
+                                          uint64_t sid) {
+    OffT in_pos = codes_offsets[sid];
+    const OffT in_end = codes_offsets[sid + 1];
+    uint32_t out_pos = output_offsets[sid];
+    const uint32_t out_end = output_offsets[sid + 1];
+
+    uint64_t scratch_low = 0, scratch_mid = 0, scratch_high = 0;
```
a10y: it might make this more readable if this were wrapped up into a struct with some methods for draining and pushing, would internalize the bit shifting stuff. but not necessary

vortex-cuda/kernels/src/fsst.cu
```diff
@@ -0,0 +1,190 @@
...
+//
+// `drain_step` picks the largest aligned store the gates allow (alignment of
+// out_pos, scratch occupancy, remaining out_end room). Bytes leave from the
+// low end (scratch_low byte 0); the kept bytes slide N positions toward that
+// low end across all three registers i.e. each u64 right-shifts by N*8 and
+// pulls the next register's low bits up to fill the vacated high bits.
+//
+//   width   gate                                          ptx
+//   ------  -------------------------------------------   ----------------
+//   16 B    out_pos % 16 == 0, scratch ≥ 16, room ≥ 16    st.global.v2.u64
```
a10y: ah yea, so i guess this is why you stop at 3 registers, b/c this is the largest write instruction

a10y:
Very cool!

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: Done

vortex-cuda/kernels/src/fsst.cu
asubiotto: Good catch. Fixed

vortex-cuda/kernels/src/fsst.cu
asubiotto: Done. It is indeed much more readable, thanks for the suggestion!

vortex-cuda/kernels/src/fsst.cu
asubiotto: Yeah, exactly. We basically want to buffer as much as needed to find alignment + a full u128 write with the restriction that we push up to 8 bytes at a time.

asubiotto:
Thanks for the review! Addressed the comments.

vortex-cuda/src/kernel/encodings/fsst.rs
```diff
@@ -0,0 +1,282 @@
...
+            .clone()
+            .execute::<PrimitiveArray>(ctx.execution_ctx())?;
+
+        // Prefix-sum lens to per-string u32 output offsets so the kernel
+        // knows where to write each decoded string.
+        #[expect(clippy::cast_possible_truncation)]
+        let output_offsets: Vec<u32> = match_each_integer_ptype!(lens.ptype(), |P| {
+            let mut out = Vec::with_capacity(lens.len() + 1);
+            let mut acc: usize = 0;
+            out.push(0u32);
```
0ax1: Why limit the offsets to u32? I think we'd want to support larger inputs. I guess we could consider templating the CUDA kernel on the offset type to save memory for the offset buffer. Then we could also support smaller offset types than u32.

0ax1 (commented):
Fantastic contribution, one question inline.

vortex-cuda/src/kernel/encodings/fsst.rs
0ax1: I see we're already doing that for code offsets but not for output offsets.

vortex-cuda/kernels/src/fsst.cu
```diff
@@ -0,0 +1,211 @@
...
+            out_pos += 1;
+            cursor -= 1;
+        }
+    }
+};
+
+template <typename OffT>
+__device__ inline void fsst_decode_string(const uint8_t *__restrict codes_bytes,
+                                          const OffT *__restrict codes_offsets,
+                                          const uint64_t *__restrict symbols,
```
joseph-isaacs: You likely want this in shared memory?

vortex-cuda/kernels/src/fsst.cu
```diff
@@ -0,0 +1,211 @@
...
+                                          const uint8_t *__restrict symbol_lengths,
```
joseph-isaacs: Could we put this in a struct with its own docs?

vortex-cuda/kernels/src/fsst.cu
asubiotto: I tried that (see commit message) but it turned out to be slower.

vortex-cuda/kernels/src/fsst.cu
joseph-isaacs, line 1: Worth noting the limitation that this can only decompress u32::MAX long str arrays?

vortex-cuda/src/kernel/encodings/fsst.rs
0ax1:
```
// `codes_offsets` is templated over the four unsigned integer widths
// (u8/u16/u32/u64). `output_offsets` is always uint32_t because of the
// MAX_BUFFER_LEN output limit.
```
Just saw this.

vortex-cuda/kernels/src/fsst.cu
joseph-isaacs: How did you implement this?

vortex-cuda/src/kernel/encodings/fsst.rs
```diff
@@ -0,0 +1,282 @@
...
+            .execute::<PrimitiveArray>(ctx.execution_ctx())?;
+        let codes_offsets = fsst
+            .codes()
+            .offsets()
+            .clone()
+            .execute::<PrimitiveArray>(ctx.execution_ctx())?;
+
+        // Prefix-sum lens to per-string u32 output offsets so the kernel
+        // knows where to write each decoded string.
+        #[expect(clippy::cast_possible_truncation)]
```
joseph-isaacs: This looks unsound to me

vortex-cuda/src/kernel/encodings/fsst.rs
```diff
@@ -0,0 +1,282 @@
...
+        let output_offsets: Vec<u32> = match_each_integer_ptype!(lens.ptype(), |P| {
+            let mut out = Vec::with_capacity(lens.len() + 1);
+            let mut acc: usize = 0;
+            out.push(0u32);
+            for &l in lens.as_slice::<P>() {
+                acc += l as usize;
+                out.push(acc as u32);
+            }
+            out
+        });
```
joseph-isaacs: would be nice if this was computed in the kernel too?

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: Yes, this is leftover from the hackathon. Ignore. I will template on output offsets.

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: No, I think you're right. I was confused with the varbinview buffer limit + this was leftover from my initial attempt, but offsets > i32 are representable just as separate buffers. Will template on the output offset type.

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: Or just use u64, but that'd be a little wasteful.

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: How would you suggest to do this? The kernel needs to know the output offsets in order to decode and I think the only thing that would work would be to do an O(n^2) prefix sum of all lengths for every string. I think it's probably cheaper to execute this linear computation with dependencies on the CPU. A GSST encoding would obviate the need for this so I would just punt on this.

vortex-cuda/kernels/src/fsst.cu
asubiotto: Will fix.

vortex-cuda/kernels/src/fsst.cu
asubiotto: This is what I tried:
```
diff --git a/vortex-cuda/kernels/src/fsst.cu b/vortex-cuda/kernels/src/fsst.cu
index 28ca66d2e..5ce282737 100644
--- a/vortex-cuda/kernels/src/fsst.cu
+++ b/vortex-cuda/kernels/src/fsst.cu
@@ -187,6 +187,17 @@ __device__ inline void fsst_decode_string(const uint8_t *__restrict codes_bytes,
                                              const uint8_t *__restrict validity_bits,                        \
                                              uint8_t *__restrict output_bytes,                               \
                                              uint64_t num_strings) {                                         \
+        __shared__ uint64_t sm_symbols[256];                                                                 \
+        __shared__ uint8_t sm_symbol_lengths[256];                                                           \
+        for (uint32_t i = threadIdx.x; i < 256; i += blockDim.x) {                                           \
+            sm_symbols[i] = symbols[i];                                                                      \
+            sm_symbol_lengths[i] = symbol_lengths[i];                                                        \
+        }                                                                                                    \
+        __syncthreads();                                                                                     \
+                                                                                                             \
         const uint64_t elements_per_block = (uint64_t)blockDim.x * ELEMENTS_PER_THREAD;                      \
         const uint64_t block_start = (uint64_t)blockIdx.x * elements_per_block;                              \
         const uint64_t block_end = (block_start + elements_per_block < num_strings)                          \
@@ -196,8 +207,8 @@ __device__ inline void fsst_decode_string(const uint8_t *__restrict codes_bytes,
         for (uint64_t sid = block_start + threadIdx.x; sid < block_end; sid += blockDim.x) {                 \
             fsst_decode_string<OffT>(codes_bytes,                                                            \
                                      codes_offsets,                                                          \
-                                     symbols,                                                                \
-                                     symbol_lengths,                                                         \
+                                     sm_symbols,                                                             \
+                                     sm_symbol_lengths,                                                      \
                                      output_offsets,                                                         \
                                      validity_bits,                                                          \
                                      output_bytes,                                                           \
```

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: I chose to just use u64, not sure templating is worth it. LMK and I can change the approach.

vortex-cuda/kernels/src/fsst.cu
asubiotto: Done

vortex-cuda/kernels/src/fsst.cu
asubiotto: Done. The host->device still needs to use independent args I think, but this cleaned up the inner kernel function.

asubiotto:
Updated to use u64 output offsets and cleaned up the kernel parameters into an args struct.

0ax1:
Gave the benchmark a spin also on my end on a GH200 (which has 4 TB/s memory bandwidth):

```
  CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="sudo ncu \
    --kernel-name fsst_u32 \
    --launch-skip 2 \
    --launch-count 1 \
    --metrics \
  smsp__thread_inst_executed_per_inst_executed,\
  smsp__thread_inst_executed_pred_on_per_inst_executed,\
  smsp__sass_average_branch_targets_threads_uniform,\
  smsp__sass_branch_targets_threads_uniform,\
  smsp__sass_branch_targets_threads_divergent,\
  smsp__sass_inst_executed_op_branch,\
  smsp__warp_issue_stalled_branch_resolving_per_warp_active,\
  smsp__average_warps_issue_stalled_branch_resolving_per_issue_active,\
  smsp__warps_eligible,\
  smsp__issue_active,\
  smsp__inst_issued_per_issue_active" \
  cargo bench -p vortex-cuda --bench fsst_cuda --target aarch64-unknown-linux-gnu -- cuda/fsst/decompress/10M


---------------------------------------------------------------------------- ----------- -------------
Metric Name                                                                  Metric Unit  Metric Value
---------------------------------------------------------------------------- ----------- -------------
smsp__average_warps_issue_stalled_branch_resolving_per_issue_active.max_rate        inst             1
smsp__average_warps_issue_stalled_branch_resolving_per_issue_active.pct                %         84.03
smsp__average_warps_issue_stalled_branch_resolving_per_issue_active.ratio           inst          0.84
smsp__inst_issued_per_issue_active.max_rate                                   inst/cycle             1
smsp__inst_issued_per_issue_active.pct                                                 %           100
smsp__inst_issued_per_issue_active.ratio                                      inst/cycle             1
smsp__issue_active.avg                                                             cycle  4,200,828.18
smsp__issue_active.max                                                             cycle     4,341,189
smsp__issue_active.min                                                             cycle     4,062,625
smsp__issue_active.sum                                                             cycle 2,218,037,277
smsp__sass_average_branch_targets_threads_uniform.max_rate                                           1
smsp__sass_average_branch_targets_threads_uniform.pct                                  %         55.52
smsp__sass_average_branch_targets_threads_uniform.ratio                                           0.56
smsp__sass_branch_targets_threads_divergent.avg                                             124,624.90
smsp__sass_branch_targets_threads_divergent.max                                                162,591
smsp__sass_branch_targets_threads_divergent.min                                                 80,209
smsp__sass_branch_targets_threads_divergent.sum                                             65,801,947
smsp__sass_branch_targets_threads_uniform.avg                                               155,567.00
smsp__sass_branch_targets_threads_uniform.max                                                  203,499
smsp__sass_branch_targets_threads_uniform.min                                                   99,768
smsp__sass_branch_targets_threads_uniform.sum                                               82,139,378
smsp__sass_inst_executed_op_branch.avg                                              inst    280,191.90
smsp__sass_inst_executed_op_branch.max                                              inst       365,698
smsp__sass_inst_executed_op_branch.min                                              inst       180,368
smsp__sass_inst_executed_op_branch.sum                                              inst   147,941,325
smsp__thread_inst_executed_per_inst_executed.max_rate                                               32
smsp__thread_inst_executed_per_inst_executed.pct                                       %         32.40
smsp__thread_inst_executed_per_inst_executed.ratio                                               10.37
smsp__thread_inst_executed_pred_on_per_inst_executed.max_rate                                       32
smsp__thread_inst_executed_pred_on_per_inst_executed.pct                               %         25.77
smsp__thread_inst_executed_pred_on_per_inst_executed.ratio                                        8.25
smsp__warp_issue_stalled_branch_resolving_per_warp_active.max_rate                                   1
smsp__warp_issue_stalled_branch_resolving_per_warp_active.pct                          %          7.53
smsp__warp_issue_stalled_branch_resolving_per_warp_active.ratio                                   0.08
smsp__warps_eligible.avg                                                            warp 18,937,555.73
smsp__warps_eligible.max                                                            warp    19,878,429
smsp__warps_eligible.min                                                            warp    18,067,927
smsp__warps_eligible.sum                                                            warp 9,999,029,425
---------------------------------------------------------------------------- ----------- -------------

cuda/cuda/fsst/decompress/10M
                        time:   [2.9651 ms 2.9839 ms 3.0055 ms]
                        thrpt:  [286.19 GiB/s 288.26 GiB/s 290.08 GiB/s]
                 change:
                        time:   [-12.955% -12.300% -11.621%] (p = 0.00 < 0.05)
                        thrpt:  [+13.149% +14.025% +14.883%]
                        Performance has improved.
```

So the main bottleneck for this kernel is warp divergence: 
```
smsp__sass_average_branch_targets_threads_uniform.pct 55.52%
```

For comparison we e.g. get
```
  cuda/cuda/runend_100runs/dispatch_u32/10M
  time:   [26.832 µs 29.795 µs 35.405 µs]
  thrpt:  [1052.2 GiB/s 1250.3 GiB/s 1388.4 GiB/s]

  cuda/cuda/runend_100runs/dispatch_u32/100M
  time:   [159.44 µs 160.21 µs 161.11 µs]
  thrpt:  [2312.2 GiB/s 2325.2 GiB/s 2336.4 GiB/s]
```
for runend.

vortex-cuda/benches/fsst_cuda.rs
```diff
@@ -0,0 +1,92 @@
...
+use vortex::session::VortexSession;
+use vortex_cuda::CudaSession;
+use vortex_cuda::executor::CudaArrayExt;
+use vortex_cuda_macros::cuda_available;
+use vortex_cuda_macros::cuda_not_available;
+use vortex_fsst::test_utils::make_fsst_clickbench_urls;
+
+use crate::timed_launch_strategy::TimedLaunchStrategy;
+
+const BENCH_SIZES: &[(usize, &str)] = &[(1_000_000, "1M"), (5_000_000, "5M"), (10_000_000, "10M")];
```
0ax1: At this point we run all benchmarks for vortex-cuda with 100M and re-use the constant from `vortex-cuda/benches/bench_config/mod.rs`. 10M inputs proved to be too noisy when running the benchmarks in codspeed and can't saturate larger GPUs like a GH200.

asubiotto:
> Gave the benchmark a spin also on my end on a GH200 (which has TB/s memory bandwidth):
> 
> ```
>   CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="sudo ncu \
>     --kernel-name fsst_u32 \
>     --launch-skip 2 \
>     --launch-count 1 \
>     --metrics \
>   smsp__thread_inst_executed_per_inst_executed,\
>   smsp__thread_inst_executed_pred_on_per_inst_executed,\
>   smsp__sass_average_branch_targets_threads_uniform,\
>   smsp__sass_branch_targets_threads_uniform,\
>   smsp__sass_branch_targets_threads_divergent,\
>   smsp__sass_inst_executed_op_branch,\
>   smsp__warp_issue_stalled_branch_resolving_per_warp_active,\
>   smsp__average_warps_issue_stalled_branch_resolving_per_issue_active,\
>   smsp__warps_eligible,\
>   smsp__issue_active,\
>   smsp__inst_issued_per_issue_active" \
>   cargo bench -p vortex-cuda --bench fsst_cuda --target aarch64-unknown-linux-gnu -- cuda/fsst/decompress/10M
> 
> 
> ---------------------------------------------------------------------------- ----------- -------------
> Metric Name                                                                  Metric Unit  Metric Value
> ---------------------------------------------------------------------------- ----------- -------------
> smsp__average_warps_issue_stalled_branch_resolving_per_issue_active.max_rate        inst             1
> smsp__average_warps_issue_stalled_branch_resolving_per_issue_active.pct                %         84.03
> smsp__average_warps_issue_stalled_branch_resolving_per_issue_active.ratio           inst          0.84
> smsp__inst_issued_per_issue_active.max_rate                                   inst/cycle             1
> smsp__inst_issued_per_issue_active.pct                                                 %           100
> smsp__inst_issued_per_issue_active.ratio                                      inst/cycle             1
> smsp__issue_active.avg                                                             cycle  4,200,828.18
> smsp__issue_active.max                                                             cycle     4,341,189
> smsp__issue_active.min                                                             cycle     4,062,625
> smsp__issue_active.sum                                                             cycle 2,218,037,277
> smsp__sass_average_branch_targets_threads_uniform.max_rate                                           1
> smsp__sass_average_branch_targets_threads_uniform.pct                                  %         55.52
> smsp__sass_average_branch_targets_threads_uniform.ratio                                           0.56
> smsp__sass_branch_targets_threads_divergent.avg                                             124,624.90
> smsp__sass_branch_targets_threads_divergent.max                                                162,591
> smsp__sass_branch_targets_threads_divergent.min                                                 80,209
> smsp__sass_branch_targets_threads_divergent.sum                                             65,801,947
> smsp__sass_branch_targets_threads_uniform.avg                                               155,567.00
> smsp__sass_branch_targets_threads_uniform.max                                                  203,499
> smsp__sass_branch_targets_threads_uniform.min                                                   99,768
> smsp__sass_branch_targets_threads_uniform.sum                                               82,139,378
> smsp__sass_inst_executed_op_branch.avg                                              inst    280,191.90
> smsp__sass_inst_executed_op_branch.max                                              inst       365,698
> smsp__sass_inst_executed_op_branch.min                                              inst       180,368
> smsp__sass_inst_executed_op_branch.sum                                              inst   147,941,325
> smsp__thread_inst_executed_per_inst_executed.max_rate                                               32
> smsp__thread_inst_executed_per_inst_executed.pct                                       %         32.40
> smsp__thread_inst_executed_per_inst_executed.ratio                                               10.37
> smsp__thread_inst_executed_pred_on_per_inst_executed.max_rate                                       32
> smsp__thread_inst_executed_pred_on_per_inst_executed.pct                               %         25.77
> smsp__thread_inst_executed_pred_on_per_inst_executed.ratio                                        8.25
> smsp__warp_issue_stalled_branch_resolving_per_warp_active.max_rate                                   1
> smsp__warp_issue_stalled_branch_resolving_per_warp_active.pct                          %          7.53
> smsp__warp_issue_stalled_branch_resolving_per_warp_active.ratio                                   0.08
> smsp__warps_eligible.avg                                                            warp 18,937,555.73
> smsp__warps_eligible.max                                                            warp    19,878,429
> smsp__warps_eligible.min                                                            warp    18,067,927
> smsp__warps_eligible.sum                                                            warp 9,999,029,425
> ---------------------------------------------------------------------------- ----------- -------------
> 
> cuda/cuda/fsst/decompress/10M
>                         time:   [2.9651 ms 2.9839 ms 3.0055 ms]
>                         thrpt:  [286.19 GiB/s 288.26 GiB/s 290.08 GiB/s]
>                  change:
>                         time:   [-12.955% -12.300% -11.621%] (p = 0.00 < 0.05)
>                         thrpt:  [+13.149% +14.025% +14.883%]
>                         Performance has improved.
> ```
> 
> So the main bottleneck for this kernel is warp divergence:
> 
> ```
> smsp__sass_average_branch_targets_threads_uniform.pct 55.52%
> ```

Nice, wish I had one of those lying around 😂 Yes, that's the main idea behind the GSST splits. Precomputing splits on the CPU showed +30% kernel throughput at 1M clickbench URLs, +26% at 5M, +12% at 10M on the DGX spark. While we can precompute splits I think maybe it's a little too much CPU prep and this is rather a motivation for the GSST encoding? Happy to introduce splits in a follow up PR if we want to do that.

vortex-cuda/benches/fsst_cuda.rs
asubiotto: Yeah the difference is that with FSST each input is a clickbench URL. Happy to use 100M if you want, but I think that's around 10GB.

0ax1:
> Nice, wish I had one of those lying around 😂 Yes, that's the main idea behind the GSST splits. Precomputing splits on the CPU showed +30% kernel throughput at 1M clickbench URLs, +26% at 5M, +12% at 10M on the DGX spark. While we can precompute splits I think maybe it's a little too much CPU prep and this is rather a motivation for the GSST encoding? Happy to introduce splits in a follow up PR if we want to do that.

Heh, I mean this is clearly not a blocker but I wanted to pin down the exact numbers. Can we def land as is from a perf end.

vortex-cuda/benches/fsst_cuda.rs
0ax1: Yeah fair, we can keep 10M but let's add comment on the constant why it's diff here or so.

vortex-cuda/src/kernel/encodings/fsst.rs
```diff
@@ -0,0 +1,285 @@
...
+    let host_bytes = host_bytes.slice(0..total_size);
+
+    let (buffers, views) = match_each_integer_ptype!(lens.ptype(), |P| {
+        build_views(
+            0,
+            MAX_BUFFER_LEN,
+            host_bytes.into_mut(),
+            lens.as_slice::<P>(),
+        )
+    });
```
joseph-isaacs, line 195: Can we build the views on the GPU?

vortex-cuda/kernels/src/fsst.cu
0ax1: As the ncu perf overview points out, we're not memory latency bound.

vortex-cuda/benches/fsst_cuda.rs
0ax1: Oh we also need to wire in the bench to codspeed in: `.github/workflows/codspeed.yml`. We can put it next to zstd which is nice for comparison. ` - { shard: 7, name: "Encodings 4", packages: "vortex-sparse vortex-zigzag vortex-zstd" }`.

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: Probably since we already compute the output offsets. It's just a question of modding by i32::MAX to split up the output into a couple of buffers.

vortex-cuda/src/kernel/encodings/fsst.rs
asubiotto: From our conversation earlier, it seems like we should at least try this out. `cub` gives us prefix sums

0ax1 (approved):

vortex-cuda/benches/fsst_cuda.rs
asubiotto: Added it to `Standalone kernels`

vortex-cuda/src/kernel/encodings/fsst.rs
0ax1: As discussed, we can address this in a follow up.

PR #7779, opened 2026-05-04
Fix logo on benchmarks website

connortsui20:
## Summary

They were inconsistent with our current branding material. Only the black version changes on the current [website](https://bench.vortex.dev/).

## Testing

N/A


connortsui20:
actually it might be the case that both of these are wrong

connortsui20:
ok it is right now

mprammer (approved):

PR #7782, opened 2026-05-04
fix: make ExtDType metadata deserialization total over byte input

### LINKED ISSUE #7773: ExtDType deserialize can panic
Note: I am testing out the Claude write workflow with this issue

### What happened?

date.rs (line 89) does metadata[0]; empty metadata panics.
time.rs (line 85) does data[0]; same issue.
timestamp.rs (line 122) checks data.len() >= 3, but later slices timezone bytes using the declared length. A malformed tz_len can still panic.

Metadata serde should be total over byte input: return VortexError, never panic.

### Steps to reproduce

N/A

### Environment

 N/A

### Additional context

_No response_

gatesn:
Fixes #7773

Generated with Claude Code

### DIFF
--- a/vortex-array/src/extension/datetime/date.rs
+++ b/vortex-array/src/extension/datetime/date.rs
@@ -87,6 +87,7 @@ impl ExtVTable for Date {
     }
 
     fn deserialize_metadata(&self, metadata: &[u8]) -> VortexResult<Self::Metadata> {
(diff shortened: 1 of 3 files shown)


a10y (approved):
nice

a10y:
Should consider adding some limited fuzzing passes to all of our serde to make sure it never panics

PR #7783, opened 2026-05-04
feat: check VortexReadAt::read_at results in the I/O driver

danking:
This ensures misbehaving VortexReadAt implementations cannot trigger later bad behavior by the I/O driver. Instead, the driver raises an error as close as possible to the badly behaving VortexReadAt implementation.

### DIFF
--- a/vortex-file/src/read/request.rs
+++ b/vortex-file/src/read/request.rs
@@ -160,7 +160,7 @@ impl CoalescedRequest {
                     req,
                 )
             }
-            if req.offset.saturating_add(req.length as u64) <= range.end {
+            if req.offset.saturating_add(req.length as u64) > range.end {
                 vortex_bail!(
                     "CoalescedRequest: sub-request for length {} at file offset {} exceeds the coalesced range: {}..{}. {:?}",
                     req.length,
--- a/vortex-file/src/read/request.rs
+++ b/vortex-file/src/read/request.rs
@@ -175,6 +175,16 @@ impl CoalescedRequest {
         })
     }
 
+    #[allow_unused)]
+    pub fn range(&self) -> &Range<u64> {
+        &self.range
+    }
+
+    #[allow_unused)]
+    pub fn alignment(&self) -> Alignment {
+        self.alignment
+    }
+
     pub fn requests(&self) -> &[ReadRequest] {
         &self.requests
     }
--- a/vortex-file/src/read/request.rs
+++ b/vortex-file/src/read/request.rs
@@ -160,6 +160,16 @@ impl CoalescedRequest {
                     req,
                 )
             }
+            if req.offset.saturating_add(req.length as u64) <= range.end {
+                vortex_bail!(
+                    "CoalescedRequest: sub-request for length {} at file offset {} exceeds the coalesced range: {}..{}. {:?}",
+                    req.length,
+                    req.offset,
+                    range.start,
+                    range.end,
+                    req,
+                )
+            }
         }
         Ok(Self {
             range,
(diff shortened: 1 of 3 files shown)


vortex-file/src/read/request.rs
```diff
@@ -133,6 +146,50 @@ impl Debug for CoalescedRequest {
...
+            for req in self.requests.iter() {
+                let request_offset = req.offset.checked_sub(self.range.start).ok_or_else(|| {
+                    vortex_err!(
+                        "CoalescedRequest: sub-request for length {} at file offset {} preceeds coalesced range: {}..{}.",
+                        req.length,
+                        req.offset,
+                        self.range.start,
+                        self.range.end,
+                    )
+                })?;
```
a10y: if we're going to check this it should be on construction, not on resolution

vortex-file/src/read/request.rs
```diff
@@ -107,6 +109,17 @@ impl Debug for ReadRequest {
...
     pub(crate) fn resolve(self, result: VortexResult<BufferHandle>) {
+        let result = result.and_then(|buffer| {
+            if self.length != buffer.len() {
+                vortex_bail!(
+                    "ReadRequest: expected buffer of length {} but received {}.",
+                    self.length,
+                    buffer.len()
+                )
+            }
+            Ok(buffer)
```
a10y: can we assert this inside of the driver loop instead? in FileSegmentSource

```result
        let drive_fut = async move {
            stream
                .map(move |req| {
                    let reader = reader.clone();
                    async move {
                        let result = reader
                            .read_at(req.offset(), req.len(), req.alignment())
                            .await;
                        // check here that buffer length == req.len()
                        req.resolve(result);
                    }
                })
                .buffer_unordered(concurrency)
                .collect::<()>()
                .await
        };
```

a10y (commented):
what do you think?

vortex-file/src/read/request.rs
danking: Done.

vortex-file/src/read/request.rs
danking: done.

vortex-file/src/read/request.rs
```diff
@@ -132,6 +137,51 @@ impl Debug for CoalescedRequest {
 }
 
 impl CoalescedRequest {
+    pub fn try_new(
```
a10y, line 140: i would also be ok with this panicking since passing a request that is outside of the coalescing window is an implementer error. but this is fine

vortex-file/src/read/request.rs
```diff
@@ -132,6 +137,51 @@ impl Debug for CoalescedRequest {
...
+            if req.offset < range.start {
+                vortex_bail!(
+                    "CoalescedRequest: sub-request for length {} at file offset {} precedes coalesced range: {}..{}. {:?}",
+                    req.length,
+                    req.offset,
+                    range.start,
+                    range.end,
+                    req,
+                )
+            }
```
a10y: nit: there's the `vortex_ensure!` macro for this pattern

a10y (approved):
optional nit, i think you probably need to regen lockfiles as well

vortex-file/src/read/request.rs
```diff
@@ -132,6 +137,51 @@ impl Debug for CoalescedRequest {
...
+        Ok(Self {
+            range,
+            alignment,
+            requests,
+        })
+    }
+
+    pub fn requests(&self) -> &[ReadRequest] {
+        &self.requests
+    }
```
joseph-isaacs, line 190: Java?? ☕ ⚰️

vortex-file/src/read/request.rs
danking: my how times have changed https://spiraldb.slack.com/archives/C066263CX0Q/p1747837979403279

vortex-file/src/read/request.rs
danking: it's the only way to have read-only public fields so I can enforce my invariants.

PR #7785, opened 2026-05-04
Add FilterPushdown support to spark data source

robert3005:
Add spark filter pushdown to spark datasource. This pr is big because it exhaustively covers all possible pushdown expressions


robert3005:
contains #8038 since spark tests exercise this case

java/vortex-spark/src/main/java/dev/vortex/spark/read/SparkPredicateToVortexExpression.java
```diff
@@ -0,0 +1,522 @@
...
+        return Optional.of(Expression.like(
+                column.get(), Expression.literal(pattern), /* negated= */ false, /* caseInsensitive= */ false));
+    }
+
+    /**
+     * Build a LIKE pattern from a literal substring, escaping the {@code %}, {@code _}, and {@code \} meta-characters
+     * so the Spark {@code STARTS_WITH}/{@code ENDS_WITH}/{@code CONTAINS} semantics (exact substring match) are
+     * preserved.
+     */
+    private static String buildLikePattern(String literal, boolean leadingWildcard, boolean trailingWildcard) {
```
onursatici, line 266: we are pushing escaped like expressions unchanged to vortex, I think that is the right thing but I will look at how we handle them in vortex, I am not sure if we do it correctly for all of them

vortex-jni/src/expression.rs
```diff
@@ -259,3 +344,209 @@ pub extern "system" fn Java_dev_vortex_jni_NativeExpression_literalString(
...
+        Ok(DecimalValue::I128(v))
+    } else {
+        Ok(DecimalValue::I256(value))
+    }
+}
+
+/// Sign-extend a two's-complement big-endian byte slice into an `i256`.
+fn i256_from_twos_complement_be(bytes: &[u8]) -> vortex::dtype::i256 {
+    let mut le = [0u8; 32];
+    let len = bytes.len().min(32);
```
onursatici, line 444: can this ever be larger than 32 and we are truncating it?

robert3005: I think we need to error here since we don't handle decimals bigger than 32 bytes

java/vortex-spark/src/main/java/dev/vortex/spark/read/SparkPredicateToVortexExpression.java
robert3005: I have merged a fix to handle it correctly in vortex as I was testing this https://github.com/vortex-data/vortex/pull/8038. We implicitly assume escape characters are the same which is the case right now

onursatici: seems like we were not handling them on stat falsification:
https://github.com/vortex-data/vortex/pull/8103

robert3005: Thanks for fixing it

java/vortex-spark/src/main/java/dev/vortex/spark/read/SparkPredicateToVortexExpression.java
```diff
@@ -0,0 +1,522 @@
...
+            }
+            sb.append(c);
+        }
+        if (trailingWildcard) {
+            sb.append('%');
+        }
+        return sb.toString();
+    }
+
+    private static BinaryOp toBinaryOp(String name) {
```
AdamGS, line 284: 😢

robert3005: all I get is `GenericScalarExpression` that has a name...

onursatici (approved):

PR #7791, opened 2026-05-05
from_arrow gets session arg

palaska:
## Summary

This PR is a mechanical refactor that deprecates `from_arrow` in favor of `from_arrow_in`, that accepts a Vortex session argument. This will be needed for arrow roundtrip with customized sessions with extension types.

This is a prerequisite for https://github.com/vortex-data/vortex/pull/7726

## API Changes

`from_arrow` is deprecated. `from_arrow_in` should be used and it requires VortexSession to be passed.


joseph-isaacs:
Should we deprecate the old API and add this new one or just break it?

vortex-array/src/arrow/mod.rs
```diff
@@ -24,10 +25,8 @@ use crate::ArrayRef;
 use crate::LEGACY_SESSION;
 use crate::VortexSessionExecute;
 
-pub trait FromArrowArray<A> {
-    fn from_arrow(array: A, nullable: bool) -> VortexResult<Self>
-    where
-        Self: Sized;
+pub trait FromArrowArray<A>: Sized {
+    fn from_arrow(array: A, nullable: bool, session: &VortexSession) -> VortexResult<Self>;
 }
```
joseph-isaacs, line 37: Should this be a session or a ExecutionCtx?

joseph-isaacs: This is only for kernel lookup so its a session, right?

palaska:
> Should we deprecate the old API and add this new one or just break it?

Breaking doesn't sound too bad to me, wdyt?

vortex-array/src/arrow/mod.rs
robert3005: This should be a session

palaska: Many of the call sites don't need execution context so I thought forcing them to create one wouldn't be ideal. The follow up PR just calls `session.dtypes()` and DType <-> ArrowDataType is a logical conversion so session makes sense to me

vortex-array/src/arrow/mod.rs
```diff
@@ -24,10 +25,19 @@ use crate::ArrayRef;
...
+        array: A,
+        nullable: bool,
+        session: &VortexSession,
+    ) -> VortexResult<Self>;
+
+    /// Convert an Arrow array to a Vortex array using the legacy global session.
+    #[deprecated(note = "Use `from_arrow_with_session` instead")]
+    fn from_arrow(array: A, nullable: bool) -> VortexResult<Self> {
+        Self::from_arrow_with_session(array, nullable, &LEGACY_SESSION)
+    }
```
joseph-isaacs, line 36: So the plan is the release this and then later rename `from_arrow_with_session` -> `from_arrow` (with session) once after removing the deprecated `from_arrow` (no session)?

palaska: I think so, or we can find a better name (maybe `from_arrow_in`?) and just delete the old function?

palaska:
Closing, see https://github.com/vortex-data/vortex/pull/7880

PR #7793, opened 2026-05-05
pluggable arrow exec

a10y:
## Summary

Closes: #000

### DIFF
--- a/vortex-array/src/arrow/canonical.rs
+++ b/vortex-array/src/arrow/canonical.rs
@@ -1,0 +1,132 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! [`CanonicalArrowEncoder`] — fallback Vortex → Arrow encoder for canonical encodings.
+//!
+//! The canonical encoder handles every `DataType` that maps directly to a canonical Vortex
+//! encoding (`Bool`, `Primitive`, `VarBinView`, `ListView`, `Struct`, `FixedSizeList`,
+//! `Decimal`, `Extension`). For now it also handles a few non-canonical optimizations
+//! (offset-based byte/list arrays); those are slated to move into encoding-keyed
+//! [`ArrowEncoder`](super::ArrowEncoder) plugins in subsequent work.
+
+use arrow_array::ArrayRef as ArrowArrayRef;
+use arrow_array::types::*;
+use arrow_schema::DataType;
+use vortex_error::VortexResult;
+use vortex_error::vortex_bail;
+use vortex_error::vortex_ensure;
+
+use crate::ArrayRef;
+use crate::ExecutionCtx;
+use crate::arrow::ArrowEncoder;
+use crate::arrow::ArrowSession;
+use crate::arrow::executor::bool::to_arrow_bool;
+use crate::arrow::executor::byte::to_arrow_byte_array;
+use crate::arrow::executor::byte_view::to_arrow_byte_view;
+use crate::arrow::executor::decimal::to_arrow_decimal;
+use crate::arrow::executor::dictionary::to_arrow_dictionary;
+use crate::arrow::executor::fixed_size_list::to_arrow_fixed_list;
+use crate::arrow::executor::list::to_arrow_list;
+use crate::arrow::executor::list_view::to_arrow_list_view;
+use crate::arrow::executor::null::to_arrow_null;
+use crate::arrow::executor::primitive::to_arrow_primitive;
+use crate::arrow::executor::run_end::to_arrow_run_end;
+use crate::arrow::executor::struct_::to_arrow_struct;
+use crate::arrow::executor::temporal::to_arrow_temporal;
+use crate::dtype::DType;
+
+/// Returns the canonical Arrow [`DataType`] for a Vortex [`DType`].
+///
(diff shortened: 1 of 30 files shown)


vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,533 @@
...
+    dtype_converters: ArrowDTypeConverterRegistry,
+    /// User-registered decoder chain. Walked before [`ArrowSession::default_decoder`].
+    decoders: RwLock<Vec<ArrowDecoderRef>>,
+    /// User-registered dtype-reader chain. Walked before [`ArrowSession::default_dtype_reader`].
+    dtype_readers: RwLock<Vec<ArrowDTypeReaderRef>>,
+    canonical_encoder: RwLock<Option<ArrowEncoderRef>>,
+    /// Fallback decoder used after the user chain has declined.
+    default_decoder: RwLock<Option<ArrowDecoderRef>>,
+    /// Fallback dtype reader used after the user chain has declined.
+    default_dtype_reader: RwLock<Option<ArrowDTypeReaderRef>>,
```
joseph-isaacs: A RwLock still has contention if all accessors and readers, maybe an `ArcSwap`. I think we might want a wrapper for this pattern, not needed here however!

vortex-array/src/arrow/encoder.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Convert `array` into an Arrow array of type `target`.
+    ///
+    /// Returning [`Ok(None)`] tells the dispatcher to canonicalize the array and re-dispatch
+    /// through the canonical encoder. Encoders may decline requests they don't recognize but
+    /// must not silently mis-convert.
+    fn to_arrow_array(
+        &self,
+        array: ArrayRef,
+        target: &DataType,
```
joseph-isaacs, line 52: I think we want this to be an optional so if we don't mind we can return encoding specific values
```suggestion
        target: Option<&DataType>,
```

Sorry if I missed a comment regarding this.

vortex-array/src/arrow/dtype_converter.rs
```diff
@@ -0,0 +1,62 @@
...
+/// external crates can override built-in behavior.
+///
+/// Returning [`Ok(None)`] passes the request to the next reader in the chain.
+pub trait ArrowDTypeReader: 'static + Send + Sync + Debug {
+    /// Try to read a Vortex [`DType`] from an Arrow [`Field`].
+    ///
+    /// Implementations typically inspect [`Field::metadata`] for the `ARROW:extension:name`
+    /// key and dispatch on it.
+    fn try_read_dtype(&self, field: &Field) -> VortexResult<Option<DType>>;
+}
```
joseph-isaacs, line 62: why have both fields and data_types? I guess nullability?

vortex-array/src/arrow/encoder.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Returning [`None`] defers to the canonical encoder's preference (e.g.
+    /// [`DataType::Utf8View`] for [`crate::dtype::DType::Utf8`]). Implementations should only
+    /// override the canonical preference when they can produce a cheaper Arrow representation
+    /// (for example, [`crate::arrays::VarBin`] preferring offset-based [`DataType::Utf8`] over
+    /// [`DataType::Utf8View`]).
+    fn preferred_arrow_type(
+        &self,
+        array: &ArrayRef,
+        session: &ArrowSession,
+    ) -> VortexResult<Option<DataType>>;
```
joseph-isaacs, line 42: two passes might be expensive?

vortex-array/src/arrow/canonical.rs
joseph-isaacs, line 1: we need to make sure this doens't regress any benchmarks

vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,533 @@
...
+        plugin: impl Into<ArrowEncoderRef>,
+    ) {
+        self.encoder_by_encoding.register(key, plugin);
+    }
+
+    /// Register a forward array encoder keyed by extension [`ExtId`].
+    pub fn register_encoder_for_extension(
+        &self,
+        key: impl Into<Id>,
+        plugin: impl Into<ArrowEncoderRef>,
```
joseph-isaacs, line 152: I think we should make these `ArrowEncoderRef` I would think we end up using the same for for many ids so we don't want a different instance for each? 

Using might may lead to that?

vortex-array/src/arrow/encoder.rs
a10y: the idea was that you must pass in an explicit target physical type, whether that's derived from user or by calling `preferred_arrow_type()` first

vortex-array/src/arrow/dtype_converter.rs
a10y: yea both is kinda funny. i think if you want to cover extension dtypes then you need the Field b/c that has metadata. and also nullability yea.

palaska:
looks like 3 people started doing the same thing in parallel 😄 https://github.com/vortex-data/vortex/pull/7726

a10y:
It's a hot topic

a10y:
I felt like this one was overly complex, I think I have a slightly simpler version over in https://github.com/vortex-data/vortex/pull/7824

PR #7794, opened 2026-05-05
Try to fix benchmakrs-website tests on windows

robert3005:
If we drop the db before we open we should avoid the double open problem


joseph-isaacs (approved):

benchmarks-website/migrate/tests/end_to_end.rs
```diff
@@ -277,7 +277,9 @@ fn open_target_db_removes_orphan_wal() {
     assert!(wal.exists(), "precondition: orphan wal staged");
     assert!(!target.exists(), "precondition: no main db file");
 
-    let _conn = migrate::open_target_db(&target).unwrap();
+    {
+        let _conn = migrate::open_target_db(&target).unwrap();
+    }
```
joseph-isaacs, line 282: should we explicitly drop this?

PR #7795, opened 2026-05-05
Add Patches lookup benchmarks

palaska:
Isolating benchmarks before this optimization: https://github.com/vortex-data/vortex/pull/7753


vortex-array/benches/patches_lookup.rs
```diff
@@ -0,0 +1,122 @@
...
+    let patches = narrow_band_patches();
+    let queries: Vec<usize> = (PATCH_HIGH..(PATCH_HIGH + NUM_QUERIES)).collect();
+
+    bencher
+        .with_inputs(|| (&patches, &queries))
+        .bench_refs(|(patches, queries)| {
+            for &q in queries.iter() {
+                divan::black_box(patches.get_patched(q).unwrap());
+            }
+        });
```
joseph-isaacs: it would be better if you pass in a Vec::with_cap(...) and push all results there since you wouldn't need a black box and you will not measure alloc time.

joseph-isaacs: This will also just measure `create_execution_ctx`?

palaska: I deleted this bench, measuring `search_index` should be sufficient. I think black_box is correct though, Vec push costs a memory write per call and seems slower.

vortex-array/benches/patches_lookup.rs
```diff
@@ -0,0 +1,100 @@
...
+    let mut indices: Vec<u64> = index_iter.collect();
+    indices.sort();
+    indices.dedup();
+    let values: Buffer<i32> = (0..indices.len() as i32).collect();
+    Patches::new(
+        ARRAY_LEN,
+        0,
+        Buffer::from(indices).into_array(),
+        values.into_array(),
+        None,
```
robert3005: We should build the offsets, they're there to make things closer to O(1) instead O(log(n)). Need to make sure we actually compute them though

palaska: I guess we should bench both

joseph-isaacs: Why we should just ensure we always have these offsets?

robert3005 (approved):

vortex-array/benches/patches_lookup.rs
robert3005: We haven't done the work to propagate it everywhere but you're correct, we should make sure it's everywhere

PR #7806, opened 2026-05-06
Fix table output for random access benchmark

robert3005:

### DIFF
--- a/benchmarks/random-access-bench/src/main.rs
+++ b/benchmarks/random-access-bench/src/main.rs
@@ -313,6 +313,13 @@ fn format_to_engine(format: Format) -> Engine {
     }
 }
 
+fn table_targets(formats: &[Format]) -> Vec<Target> {
+    formats
+        .iter()
+        .map(|format| Target::new(format_to_engine(*format), *format))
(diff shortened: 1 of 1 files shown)


benchmarks/random-access-bench/src/main.rs
danking, line 476: The trouble is here. Even though targets now only has `Vortex, Vortex`, `Arrow, Parquet`, and `Arrow, Lance`, you're sending in way more measurements. Which value is displayed for Dataset A and `Vortex, Vortex`? The NVME & footer? non-NVME & footer? NVME & non-footer? non-NVME & non-footer?

The Target is not rich enough to capture all the possible ways in which a benchmark is indexed.

danking (changes requested):
x

benchmarks/random-access-bench/src/main.rs
robert3005: Interesting that it continues to display fine for json output. I have made them produce display from the source of data originally and looks like this got broken

robert3005:
THIS NEEDS A LOT MORE WORK

PR #7809, opened 2026-05-06
feat: vortex-datafusion segment cache

ch-sc:
## Summary

`vortex-layout` already provides a `SegmentCache` trait and a `MokaSegmentCache` implementation, and `vortex-file`'s `OpenOptions` already has `with_segment_cache()` that wires it into every file open. `vortex-datafusion` just did not thread this through and so every new query re-reads data segments.

This PR:

  - Adds `segment_cache: Option<Arc<dyn SegmentCache>>` to `VortexOpener` and `VortexSource`.
  - Passes the cache into `OpenOptions::with_segment_cache()` when opening each file in `VortexOpener::open()`.
  - Exposes `VortexSource::with_segment_cache()` so callers can inject a shared cache instance.

Closes: #000

## API Changes

```rust
 impl VortexSource {
      /// Sets a [`SegmentCache`] to reuse decoded segment bytes across scans of the same file.
      pub fn with_segment_cache(self, segment_cache: Arc<dyn SegmentCache>) -> Self;
  }
```

## Testing

The `segment_cache` field defaults to `None`, so all existing behavior is unchanged when the cache is not configured.


AdamGS:
I think there's an issue here where that the segment caches uses `SegmentId` for its keys, so you can't be re-used across files as they are not unique IIRC.
In some cases this will probably work correctly, but often a single `VortexSource` will be shared between multiple files.

ch-sc:
Ah, I see. `SegmentId` is file dependent, but only consists of a `u32`. 

I don't have a clear solution in mind, but if `SegmentId` would contain file information (maybe path and an e-tag for updates to the file) we could resolve the file dependency.

AdamGS:
I wonder if there's a mix of something like `SegmentCacheFactory` and tracking per-file state in `VortexSource`/`VortexFormat`/`ListingTable`? I need to think about it sore more. 

Another way to solve this is just cache at the `VortexReadAt` level, which we already do allow customizing, but that will override the built-in `ObjectStore` IO, so you do give up some functionality.

ch-sc:
I played around with a SegmentCacheBuilder that namespaces a single shared cache across files. Introducing a FileIdentity (path + version) for the namespacing. 

I mainly wanted to see how much benefits we would get from this. Might not be the best path for the project. You can take a look at the code here: https://github.com/massive-com/vortex/pull/5

PR #7814, opened 2026-05-06
Exectuion and optimization tracing harness for testing

AdamGS:
## Summary

Introduces an internal tracing API that enables testing execution and optimization flows with snapshot testing! Just wrap your code into a callaback and call `trace_op`, and you get back both the result and a snapshot you can assert.

For example:
```rust
    #[test]
    fn trace_optimize_reduce_fixpoint() -> VortexResult<()> {
        let values = PrimitiveArray::from_iter([0i32, 1, 2, 3]).into_array();
        let filter =
            FilterArray::try_new(values.clone(), Mask::new_true(values.len()))?.into_array();

        let traced = trace_op(|| filter.optimize())?;

        assert!(traced.output.is::<Primitive>());
        assert_arrays_eq!(traced.output, values);
        insta::assert_snapshot!(traced.trace.to_string(), @r"
            optimize root=vortex.filter(i32, len=4) session=false
            reduce TrivialFilterRule: vortex.filter(i32, len=4) -> vortex.primitive(i32, len=4)
            done output=vortex.primitive(i32, len=4)
        ");

        Ok(())
    }
```

The API for tests is pretty nice, internally there are still improvements that can be made here, but this is a good step towards building more confidence and making the execution model clearer.


vortex-array/src/executor.rs
```diff
@@ -154,29 +154,63 @@ impl ArrayRef {
...
     pub fn execute_until<M: Matcher>(self, ctx: &mut ExecutionCtx) -> VortexResult<ArrayRef> {
         let mut current_array = self;
         let mut current_builder: Option<Box<dyn ArrayBuilder>> = None;
         let mut stack: Vec<StackFrame> = Vec::new();
         let max_iterations = max_iterations();
 
-        for _ in 0..max_iterations {
+        crate::trace_op!(record_execute_until_start::<M>(&current_array));
+
+        for iteration in 1..=max_iterations {
```
joseph-isaacs: 1... ?

AdamGS: cleaned this up.

myrrc (approved):
LGTM, a very valuable PR.

vortex-array/src/optimizer/mod.rs
```diff
@@ -71,22 +72,24 @@ fn try_optimize(
...
+    trace_op!(record_optimize_start(array, session.is_some()));
+
     // Apply reduction rules to the current array until no more rules apply.
-    let mut loop_counter = 0;
-    'outer: loop {
-        if loop_counter > 100 {
-            vortex_bail!("Exceeded maximum optimization iterations (possible infinite loop)");
-        }
-        loop_counter += 1;
+    for _ in 0..=100 {
```
connortsui20, line 80: 🤣

robert3005 (approved):

connortsui20 (dismissed):
just continuing the conversation on slack: I think that we should first justify why we cannot use `tracing` instead of handrolling these traces before we merge this. My gut feeling is that tracing would remove a bunch of boilerplate here and would make things simpler, but it could be the case that `tracing` is not expressive enough. Regardless, that should be justified.

robert3005:
I don't think you want tracing for this. You can get structure out of the execution context that we control instead of munging strings. In particular I would consider tracing informational and not a thing you should use for testing behaviour.

robert3005 (dismissed):
I had misread some of the code. I think this should be stored on the execution context instead of a thread local

robert3005 (approved):

PR #7822, opened 2026-05-06
Slots are a Box<[...]> instead of Vec<...>

robert3005:
Trying to see if the one pointer less is beneficial


robert3005:
This is not worth it

PR #7823, opened 2026-05-06
Use smallvec for internal stats storage

robert3005:
Instead of always allocating fixed sized vec we can use a smallvec instead


a10y (approved):
Do we think these allocations are significant?

robert3005:
I am not sure, trying things out to see benchmarks

robert3005:
looks like microbenchmark wise this is a win since we no longer have indirect vec allocation and we batch it in the owner arc<lock>

vortex-array/src/stats/stats_set.rs
```diff
@@ -33,7 +35,7 @@ use crate::scalar::ScalarValue;
 
 #[derive(Default, Debug, Clone)]
 pub struct StatsSet {
-    values: Vec<(Stat, Precision<ScalarValue>)>,
+    values: SmallVec<[(Stat, Precision<ScalarValue>); Stat::CARDINALITY]>,
```
joseph-isaacs: This we should actually choose a smaller number since we don't usually see all stats in the Vec.

vortex-array/src/stats/stats_set.rs
```diff
@@ -42,20 +44,16 @@ impl StatsSet {
     /// # Safety
     ///
     /// This method will not panic or trigger UB, but may lead to duplicate stats being stored.
-    pub unsafe fn new_unchecked(values: Vec<(Stat, Precision<ScalarValue>)>) -> Self {
+    pub unsafe fn new_unchecked(
+        values: SmallVec<[(Stat, Precision<ScalarValue>); Stat::CARDINALITY]>,
```
joseph-isaacs: Type alias please

joseph-isaacs (changes requested):

vortex-array/src/stats/stats_set.rs
joseph-isaacs: Otherwise you might as well have an array?

robert3005:
I made the smallvec be 4 elements in size. empirically we only ever populate more if we write or if there's multiple operations

joseph-isaacs (approved):

Issue #7686, opened 2026-04-28
Tracking Issue: Geospatial Vector types


a10y:
## Description

Support for geospatial vector data is growing among the analytics community. For example

- [GeoParquet](https://geoparquet.org/releases/v1.1.0/) standard, a standard way for Parquet files to store vector data (`MULTIPOLYGON`, `LINESTRING`, etc.) as well as other feature components and store them as-such instead
- [GeoArrow](https://geoarrow.org/format.html) standard, which formalizes a set of extension types on top of Apache Arrow
- OGC Simple Feature Access provides a vast catalog of geospatial function types
- Iceberg V3 adds first-party support for the `GEOMETRY` type

## Goals

- Any data stored in a GeoParquet file should be convertible to a Vortex file
- We should support exposing geospatial schemas to our query engines that support it. Currently that is DuckDB and DataFusion. Iceberg v3 also supports it but that is in a separate repo.
- Not an explicit goal for this epic, but we should perform at least on par with other formats in common geospatial benchmarks. Currently the one that I've found is Apache Sedona's [SpatialBench](https://sedona.apache.org/spatialbench/)

## Components

- [x] Make Arrow execution pluggable -> https://github.com/vortex-data/vortex/pull/7824
- [x] Implement basic support for WKB type -> https://github.com/vortex-data/vortex/pull/7722
- [x] Expose geometry columns to DuckDB -> also https://github.com/vortex-data/vortex/pull/7722
- [ ] Expose geometry to DataFusion
- [ ] Implement stats for Geometry types (Dependent on https://github.com/vortex-data/vortex/issues/7707)
- [ ] Add [SpatialBench](https://sedona.apache.org/spatialbench/) to benchmarks: DuckDB (#8543, ...); DataFusion lane still open

## Out of scope for now

- Support for union types and singular `geometry` unified extension type
- Raster data support
- Interop with SedonaDB (?)

## Questions

- Most systems seem to use WKB, should we just use that, but then have compressed codecs that shred it?
  - _For now, we should focus on WKB as lowest common denominator and then can move on to other encodings_

## Implementation history

- Working on PR to enable session-pluggable Arrow encoding to/from Vortex - https://github.com/vortex-data/vortex/pull/7824
- PR opened for basic WKB support in Vortex, wiring into DuckDB

jiayuasu:
[...]

PR #7824, opened 2026-05-07
pluggable registry for input/export arrow kernels

a10y:
## Summary

  Adds a pluggable `ArrowSession` registry on `VortexSession` for round-tripping Vortex extension types in and out of Arrow extension types. Unblocks Arrow round-trip for `arrow.uuid` today, with `arrow.parquet.variant`, GeoArrow, and tensor types as the next consumers.

  Part of #7686.

  ## API changes

  The session exposes two trait-driven plugin slots:

  - `ArrowExportVTable` — dispatched by **target Arrow extension name** (`ARROW:extension:name`). Implementations turn a Vortex `ArrayRef` into an Arrow `ArrayRef` shaped to the requested `Field`. Also provides `to_arrow_field` for schema inference when only a Vortex `DType` is in
  hand.
  - `ArrowImportVTable` — dispatched by **source Arrow extension name** carried on the incoming `Field`. Implementations turn an Arrow `ArrayRef` back into a Vortex `ArrayRef`, including any storage re-encoding (e.g. `FixedSizeBinary[16]` → `FixedSizeList<u8; 16>` for UUID).

  Both traits return `Unsupported(input)` to defer to the next plugin or to the canonical fallback, so multiple plugins can register against the same key and probe in order.

  New session entry points (`vortex-array/src/arrow/session.rs`):

  - `ArrowSession::to_arrow_field` / `to_arrow_schema` — Vortex `DType` → Arrow `Field`/`Schema`, recursing into containers so nested extension fields go through the registered plugin.
  - `ArrowSession::from_arrow_field` / `from_arrow_schema` — inverse direction, plugin-aware.
  - `ArrowSession::from_arrow_record_batch` / `execute_record_batch` — `RecordBatch` round-trip.
  - `ArrowSessionExt` extension trait so any `SessionExt` can call `session.arrow().…`.

  The default session pre-registers the builtin UUID plugin (`vortex-array/src/extension/uuid/arrow.rs`).

  ## What's *not* in the plugin layer

  `Date`, `Time`, and `Timestamp` are Vortex builtin extensions that map directly to native Arrow temporal types, so they continue to go through the canonical executor (`vortex-array/src/arrow/executor/temporal.rs`) rather than the plugin registry. The plugin layer is reserved for **Arrow extension types** that the canonical path can't express.

  ## DataFusion wiring

  `vortex-datafusion` now goes through the session for schema/array conversion:

  - `convert/schema.rs::calculate_physical_schema` uses `ArrowSession::to_arrow_field` so extension metadata survives projection.
  - `persistent/format.rs` and `persistent/opener.rs` route schema inference through the session.
  - `persistent/sink.rs` uses `from_arrow_record_batch`, passing the original schema separately from `RecordBatch::schema()` to preserve `ARROW:extension:name` metadata that DataFusion strips at runtime.

  ## Tests

  Two new end-to-end tests in `vortex-datafusion/src/persistent/tests.rs`:

  - `arrow_uuid_extension_roundtrip` — write Arrow UUID column to a Vortex file via the session, `SELECT *` it back, assert the field still carries the `Uuid` extension type and the values match.
  - `arrow_uuid_extension_roundtrip_nested_struct` — same flow with the UUID nested in a top-level `Struct`, exercising recursive session-aware schema inference.


vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,239 @@
...
+/// A plugin that lets users plugin conversion between Vortex extension types and Arrow arrays
+/// and data types.
+pub trait ArrowVTable: 'static + Send + Sync + Debug {
+    /// Vortex extension type ID that this plugin handles.
+    fn vortex_ext_id(&self) -> ExtId;
+
+    /// The name of the Vortex extension type handled by this plugin (e.g. `"arrow.uuid"`), if any.
+    fn arrow_ext_name(&self) -> Option<&'static str> {
+        None
+    }
```
joseph-isaacs: when would I not put a name here?

vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    /// Look up the plugin registered for the given Vortex extension ID.
+    pub fn for_vortex_ext(&self, id: &ExtId) -> Option<ArrowVTableRef> {
+        self.by_vortex_ext.find(id)
+    }
+
+    /// Look up the plugin registered for the given Arrow extension name.
+    pub fn for_arrow_ext(&self, name: &str) -> Option<ArrowVTableRef> {
+        self.by_arrow_ext.find(&Id::new(name))
+    }
```
joseph-isaacs: Do we need this? Its much slower than by id? I would like to encourage that users hold a Id long term

vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,239 @@
...
+            by_vortex_ext: Registry::default(),
+            by_arrow_ext: Registry::default(),
+        };
+
+        // Builtin extension-type plugins. User registrations with the same ID will replace them.
+        this.register(Uuid);
+        this.register(Date);
+        this.register(Time);
+        this.register(Timestamp);
+
```
joseph-isaacs, line 188: Are we sure there is only one exporter for each vortex type? I am not sure that is true?

I would think there is one per (vortex-type, arrow-type) We might have multiple arrow-types from a single vx one?

vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,239 @@
...
+            let len = array.len();
+            let arrow = plugin.execute_arrow(array, target_field, ctx)?;
+            vortex_error::vortex_ensure!(
+                arrow.len() == len,
+                "Arrow array length does not match Vortex array length after conversion to {:?}",
+                arrow
+            );
+            return Ok(arrow);
+        }
+        canonical_execute_arrow(array, target.map(Field::data_type), ctx)
```
joseph-isaacs: Its not clear if this current framework will help with these exporters. I think it should

vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,239 @@
...
+use crate::extension::uuid::Uuid;
+
+/// A plugin that lets users plugin conversion between Vortex extension types and Arrow arrays
+/// and data types.
+pub trait ArrowVTable: 'static + Send + Sync + Debug {
+    /// Vortex extension type ID that this plugin handles.
+    fn vortex_ext_id(&self) -> ExtId;
+
+    /// The name of the Vortex extension type handled by this plugin (e.g. `"arrow.uuid"`), if any.
+    fn arrow_ext_name(&self) -> Option<&'static str> {
```
joseph-isaacs: Is this a general exporter for vortex <-> arrow or specific for extensions?

vortex-array/src/arrow/session.rs
a10y: When you have a one-way mapping from Vortex extension -> Arrow. For example Vortex's Date/Time/Timestamp types don't have an arrow extension. I think most implementations will fill this in FWIW

a10y:
Perf seems within margin of error? I still need to fix up variant handling before we can merge.

vortex-array/src/arrow/executor/list.rs
```diff
@@ -96,10 +96,11 @@ fn list_to_list<O: OffsetSizeTrait + NativePType>(
         .to_buffer::<O>()
         .into_arrow_offset_buffer();
 
-    let elements = array
-        .elements()
-        .clone()
-        .execute_arrow(Some(elements_field.data_type()), ctx)?;
+    let elements = ctx.session().clone().arrow().execute_arrow(
```
joseph-isaacs, line 99: `ctx.session().clone().arrow()` show expensive is this?

vortex-array/src/arrow/executor/mod.rs
```diff
@@ -87,92 +90,22 @@ impl ArrowArrayExecutor for ArrayRef {
...
+        // `ctx` from being passed back through to the session method.
+        let session = ctx.session().clone();
+        let target = match data_type {
+            Some(dt) => Some(Field::new("", dt.clone(), self.dtype().is_nullable())),
+            // No target supplied: if the source dtype tree contains any Vortex extension,
+            // synthesize a Field via session-aware inference so registered plugins run and
+            // ARROW:extension:name metadata is preserved end-to-end. For non-extension
+            // trees we leave target as None so canonical preferred-type logic (e.g.
+            // VarBin → Utf8 instead of Utf8View) keeps running.
+            None if dtype_has_extension(self.dtype()) => {
```
joseph-isaacs: this is odd

vortex-array/src/arrow/session.rs
```diff
@@ -0,0 +1,777 @@
...
+///
+/// Plugins that don't handle the supplied array return [`Unsupported`][Self::Unsupported]
+/// with ownership of the input so the session can probe the next plugin or fall back to the
+/// canonical path. Errors are propagated through [`VortexResult`].
+pub enum ArrowImport {
+    /// The plugin does not handle this input; the session may try another plugin.
+    Unsupported(ArrowArrayRef),
+    /// A successful import.
+    Imported(ArrayRef),
+}
```
joseph-isaacs, line 92: do we need two?

vortex-array/src/dtype/arrow.rs
```diff
@@ -258,6 +263,12 @@ impl DType {
     }
 
     /// Returns the Arrow [`DataType`] that best corresponds to this Vortex [`DType`].
+    ///
+    /// **Prefer `ArrowSession::to_arrow_data_type` (or `to_arrow_field`).** This method
+    /// has no awareness of registered Arrow extension plugins, so any [`DType::Extension`]
+    /// outside the builtin temporal set will fail or silently lose its
+    /// `ARROW:extension:name` metadata. The session methods recurse through containers
+    /// and dispatch plugins at every extension node.
     pub fn to_arrow_dtype(&self) -> VortexResult<DataType> {
```
joseph-isaacs, line 273: deprecate?

vortex-array/src/arrow/executor/list.rs
a10y: i think this is unnecessary

vortex-array/src/arrow/session.rs
a10y: i guess probably not, i just like giving them meaningful names

a10y: they return different things in the success enum

vortex-array/src/dtype/arrow.rs
a10y: yea, i can clean this shim up a bit

vortex-array/src/arrow/executor/list.rs
a10y: actually it is necessary b/c ExecutionCtx isheld mutably and we need an immutable ref.

this is just an Arc clone so should be cheap compared to execute_arrow

a10y:
@palaska I just noticed, for vortex-tensor `Vector` type you made a new Arrow extension type up. Is there a reason not to use the canonical [Arrow tensor type](https://arrow.apache.org/docs/format/CanonicalExtensions.html#fixed-shape-tensor) instead?

a10y:
There's some fundamental weirdness with Parquet Variant.

Variant itself is the only canonical DType that does not have an obvious Arrow encoding. The closest is the `arrow.parquet.variant` Arrow extension type, but we have decided to make Vortex's Parquet Variant be an encoding. I think we _also_ need to make Vortex parquet variant be an extension type wrapping `DType::Variant` storage type.

So, this leads us to an interesting situation where we need to implement the `ArrowExportVTable` method to output the field type:

```
    /// Build the Arrow [`Field`] this plugin produces for the given Vortex extension
    /// `dtype`. Used during schema inference.
    fn to_arrow_field(
        &self,
        name: &str,
        dtype: &ExtDTypeRef,
        session: &ArrowSession,
    ) -> VortexResult<Option<Field>>;
```

According to the [Arrow spec](https://arrow.apache.org/docs/format/CanonicalExtensions.html#parquet-variant), this field type must be a STRUCT with a required `metadata` BINARY field, and then either/both of a BINARY `value` field, or a `typed_value` field of the appropriate type.

We don't capture information about shredding in the system, so we don't know here what fields our returned struct field should have...

vortex-array/src/arrow/executor/list.rs
a10y: not sure i have an idea of cleaner way to do this

vortex-array/src/arrow/executor/mod.rs
a10y: cleaned this up a bit

vortex-array/src/dtype/arrow.rs
a10y: deprecated and moved the impl out to a crate-local function

palaska:
> @palaska I just noticed, for vortex-tensor `Vector` type you made a new Arrow extension type up. Is there a reason not to use the canonical [Arrow tensor type](https://arrow.apache.org/docs/format/CanonicalExtensions.html#fixed-shape-tensor) instead?

No very good reason tbh, I guess Vector is a special case of FixedShapeTensor and we could have used it, but since Vortex already has the distinction between those two, I thought we could mirror it. I was planning to have something like [this](https://github.com/spiraldb/spiraldb/pull/4826/changes#diff-fcc73e32373d1419f31320eb9705169b79b0f926b23345e89e57a0a4b1c7d042) in Spiral. Lmk if you prefer otherwise, I can put up a PR!

robert3005:
for variant I think it's fine if we force you to specify the output type and otherwise refuse. IMHO converting things unconditionally to parquet variant could be surprising but I also think there's not enough usage

joseph-isaacs (approved):
Approving to unblock I will do a stronger review post merge

PR #7825, opened 2026-05-07
perf[array]: small vec for slots

joseph-isaacs:
try out small vec for arrays (either 2 or 4), here we try 4


vortex-array/src/arrays/listview/array.rs
```diff
@@ -470,7 +472,7 @@ impl Array<ListView> {
         }
         let dtype = self.dtype().clone();
         let len = self.len();
-        let slots = self.slots().to_vec();
+        let slots: ArraySlots = self.slots().iter().cloned().collect();
```
robert3005, line 475: I think maybe there's nicer way of doing this?

encodings/alp/src/alp_rd/array.rs
```diff
@@ -457,8 +458,9 @@ impl ALPRDData {
         left_parts: &ArrayRef,
         right_parts: &ArrayRef,
         patches: Option<&Patches>,
-    ) -> Vec<Option<ArrayRef>> {
-        let mut slots = vec![Some(left_parts.clone()), Some(right_parts.clone())];
+    ) -> ArraySlots {
+        let mut slots: ArraySlots =
+            vortex_array::smallvec::smallvec![Some(left_parts.clone()), Some(right_parts.clone())];
```
robert3005: these could be imported

robert3005 (approved):

PR #7826, opened 2026-05-07
fix[py]: remove mimalloc in library

### LINKED ISSUE #7760: Importing `vortex` after `pyarrow.dataset` impacts the `pyarrow` runtime, causing SIGSEGV
## What happened?

Importing `vortex` _after_ importing `pyarrow.dataset` corrupts pyarrow's runtime state. 
Calling in to pyarrow can then segfault.

### Steps to reproduce

```python                                                                                                                                                                                           
import pyarrow.dataset
import vortex
import pyarrow as pa

pa.array([1])

# Process finished with exit code 139 (interrupted by signal 11:SIGSEGV)
```
### Environment

- Vortex version 0.70.0
- Python: 3.11
- macOS: 26.4.1

## Suspected root cause                                                                                                                                                                             

Maybe `vortex`'s Python extension statically links its own copy of the Arrow runtime? If imported after `pyarrow.dataset`, this means parts of Arrow's process-global state could be overwritten, with subsequent pyarrow calls dereferencing function pointers or type handles that point into a runtime it's not coordinated with. 

## Possible fix                                                                                                                                                                  

Reuse the existing pyarrow runtime when it's already loaded?
Ref: [pyarrow C++ integration docs](https://arrow.apache.org/docs/python/integration/extending.html).

### Additional context

Reversing the import order lets pyarrow's registrations win, leaving the runtime self-consistent:

* `pyarrow.dataset` → `vortex`
  
   

* `vortex` → `pyarrow.dataset`

joseph-isaacs:
Remove `mimalloc` feature all together.

Fixes #7760

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11070,7 +11070,6 @@ dependencies = [
  "bytes",
  "itertools 0.14.0",
  "log",
- "mimalloc",
  "object_store 0.13.2",
  "parking_lot",
  "pyo3",
--- a/vortex-python/Cargo.toml
+++ b/vortex-python/Cargo.toml
@@ -28,9 +28,10 @@ default = ["extension-module", "tui"]
 # duplicate PyInit__lib symbols.
 extension-module = []
 tui = ["dep:vortex-tui"]
-# Set the default allocator to mimalloc, this is enabled by default
-# for the maturin wheels, but it is not a default feature to not conflict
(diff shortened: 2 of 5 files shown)


AdamGS (approved):

AdamGS:
Lets just remove the feature.

PR #7827, opened 2026-05-07
feat: all_non_distinct aggregation make fuzzer faster

joseph-isaacs:
Add a fused all(non_distinct) aggregation (which should be a streaming `non_distinct` and all). 

Use this for array equality.


robert3005 (approved):
This would be better if we had streaming execution, right now we have to fuse this in the aggregator

gatesn:
Can I request that we name this better?

This is the same thing as Spark's equal_null right? https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.equal_null.html

"distinct" implies to me that it comparing values across rows

gatesn:
Also why do we need this fused? Isn't this the point of saturating aggregates? Can this be a separate "all" aggregate function, and "equal_null" scalar function?

joseph-isaacs:
this is the duckdb name

-- DuckDB
a IS NOT DISTINCT FROM b

-- DataFusion SQL
a <=> b

-- Velox / Presto-style function
NOT distinct_from(a, b)

-- Daft Python
daft.functions.eq_null_safe(a, b)

-- Spark SQL / PySpark
a <=> b
col("a").eqNullSafe(col("b"))

joseph-isaacs:
you cannot actually short circuit without applying the no distinct the underlying array

PR #7829, opened 2026-05-07
TurboQuant again!

connortsui20:
## Summary

Moves TurboQuant out of `vortex-tensor` into a new `vortex-turboquant` crate.

The first commit was mostly copying and pasting a bunch of code, as well as adding the `unpack` method to replace canonicalization. The second commit was cleaning up everything holistically.

A lot of the code in `vortex-tensor` was reviewed pretty lightly because we knew that it was unstable, but now that we are more certain about the implementation (not necessarily about the exact design, but the actual implementation of the TQ algorithms), I think it is worth reviewing everything as a whole.

## Testing

These tests were mostly there before, but now there are more!


vortex-turboquant/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+pub use vector::unpack::turboquant_unpack;
+pub use vtable::TurboQuant;
+pub use vtable::TurboQuantMetadata;
+
+/// Register the TurboQuant extension type with a Vortex session.
+pub fn initialize(session: &vortex_session::VortexSession) {
+    use vortex_array::dtype::session::DTypeSessionExt;
+    use vortex_tensor::vector::Vector;
+
+    session.dtypes().register(Vector);
```
gatesn: I think you want to call `vortex_tensor::initialize(...)`?

We should probably document that initialize functions should be idempotent!

vortex-turboquant/src/vector/normalize.rs
```diff
@@ -0,0 +1,143 @@
...
+    // - Every valid nonzero row is divided by its exact stored norm, so valid rows are unit norm.
+    // - Invalid rows carry invalid validity on both children and are not part of the L2Denorm
+    //   semantic contract.
+    unsafe { L2Denorm::new_array_unchecked(normalized, norms, row_count) }
+}
+
+fn normalize_elements<T>(
+    elements: &PrimitiveArray,
+    norms: &PrimitiveArray,
+    is_valid: &[bool],
```
gatesn: We could pass in the validity mask and zip the iter instead of allocating a vec above

vortex-turboquant/src/scalar_fns/pack.rs
```diff
@@ -0,0 +1,177 @@
...
+        &self,
+        _dtype: &DType,
+        _len: usize,
+        _metadata: &[u8],
+        _children: &dyn ArrayChildren,
+        _session: &VortexSession,
+    ) -> VortexResult<ScalarFnArrayParts<Self>> {
+        vortex_bail!("TQPack scalar-fn arrays are not serializable")
+    }
+}
```
connortsui20: Do we want to serialize a `TQPack` array? It doesn't really make a lot of sense (though it does for `TQUnpack` as it means we can store the TQ compressed array and get back a `Vector` array)

connortsui20:
codex has some findings that I will have to investigate tomorrow:

<details>
 
[P1] Lazy TQPack still cannot be used as the write-time path. initialize() only registers the TQUnpack array plugin, and TQPack’s array serialization returns None / deserialization bails, so writing TQPack::try_new_array(...).into_array() will fail before pack_vector runs. The file test currently writes an already-executed TurboQuant array, so it does not cover the actual lazy write path. See lib.rs (line 81), pack.rs (line 158), and file.rs (line 30).



[P2] vortex_turboquant::initialize() is not self-contained for lazy TQUnpack file reads. TQUnpack deserialization requires the parent dtype to downcast as AnyVector, but initialize() no longer registers vortex_tensor::vector::Vector. The current file tests mask this by separately calling vortex_tensor::initialize(&session). A session that only calls the TurboQuant initializer can register TQUnpack but still fail to deserialize its Vector parent dtype. See lib.rs (line 76) and unpack.rs (line 167).



[P2] The public SORF dimension padding path uses unchecked next_power_of_two(). SorfMatrix::try_new and SorfTransform::return_dtype can panic in debug builds on oversized dimensions instead of returning VortexResult, and validate_sorf_options does not reject the bad dimension before serialized metadata reaches this path. Use checked_next_power_of_two() and reject zero/overflow in validation. See rotation.rs (line 79) and vtable.rs (line 96).

</details>

joseph-isaacs:
Please can we think a little about the reviewer this is a 4k+ PR with a large move and a new feature

vortex-turboquant/src/scalar_fns/encode.rs
joseph-isaacs, line 1: are there any tests?

joseph-isaacs (dismissed):
please can this be split up

vortex-turboquant/src/scalar_fns/metadata.rs
```diff
@@ -0,0 +1,47 @@
...
+        let bit_width = u8::try_from(self.bit_width)
+            .map_err(|_| vortex_err!("TurboQuant bit_width does not fit u8"))?;
+        let num_rounds = u8::try_from(self.num_rounds)
+            .map_err(|_| vortex_err!("TurboQuant num_rounds does not fit u8"))?;
+
+        TurboQuantConfig::try_new(bit_width, self.seed, num_rounds)
+    }
+}
+
+pub(super) fn serialize_config(config: &TurboQuantConfig) -> Vec<u8> {
```
gatesn, line 39: Seems a bit weird these aren't on the TQScalarFnMetadata impl? Don't mind that much 🤷 - you could also just inline them to the call site they're so small

vortex-turboquant/src/tests/malformed.rs
```diff
@@ -0,0 +1,198 @@
...
+use vortex_buffer::Buffer;
+use vortex_error::VortexResult;
+
+use super::execute_tq_unpack_from_metadata;
+use super::test_session;
+use super::vector_validity;
+use crate::TurboQuant;
+use crate::TurboQuantMetadata;
+
+fn validity_for_nullability(nullability: Nullability) -> Validity {
```
gatesn: Does this not exist as Validity::from(Nullability)?

vortex-turboquant/src/scalar_fns/decode.rs
```diff
@@ -0,0 +1,213 @@
...
+        child: ArrayRef,
+        config: &TurboQuantConfig,
+        len: usize,
+    ) -> VortexResult<ScalarFnArray> {
+        ScalarFnArray::try_new(TQUnpack::new(config).erased(), vec![child], len)
+    }
+}
+
+impl ScalarFnVTable for TQUnpack {
+    type Options = TurboQuantConfig;
```
gatesn: Do you need this? Isn't this entirely encapsulated by the child's ext dtype metadata?

You should model this as e.g. options the user must provide in their SQL query `SELECT tq.unpack(..., <options>)`

vortex-turboquant/src/vector/normalize.rs
```diff
@@ -0,0 +1,146 @@
...
+    let output_len = num_vectors
+        .checked_mul(dimensions)
+        .ok_or_else(|| vortex_err!("TurboQuant normalized vector length overflow"))?;
+    let mut output = BufferMut::<T>::with_capacity(output_len);
+
+    for i in 0..num_vectors {
+        let row_values = &values[i * dimensions..][..dimensions];
+        let norm = norm_values[i];
+
+        if !mask.value(i) || norm == T::zero() {
```
gatesn: I think you probably want to iterate the mask, rather than call `.value(i)` on every iteration

gatesn: In fact may even be worth a criterion benchmark and a quick skim of samply for any red falgs

vortex-turboquant/src/vector/quantize.rs
```diff
@@ -0,0 +1,150 @@
...
+        .checked_mul(padded_dim)
+        .ok_or_else(|| vortex_err!("TurboQuant codes length overflow"))?;
+    let mut all_indices = BufferMut::<u8>::with_capacity(codes_len);
+    let mut padded = vec![0.0f32; padded_dim];
+    let mut transformed = vec![0.0f32; padded_dim];
+
+    let f32_slice = f32_elements.as_slice();
+    let dimension = dimension as usize;
+    for row in 0..num_vectors {
+        if !mask.value(row) {
```
gatesn: Same here, avoid the function call for extracting bit indices

vortex-turboquant/src/vector/storage.rs
```diff
@@ -0,0 +1,167 @@
...
+    codes_validity: &Validity,
+    len: usize,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<()> {
+    let struct_mask = struct_validity.execute_mask(len, ctx)?;
+    let norms_mask = norms_validity.execute_mask(len, ctx)?;
+    let codes_mask = codes_validity.execute_mask(len, ctx)?;
+
+    for row in 0..len {
+        if !struct_mask.value(row) {
```
gatesn: And also here.

vortex-turboquant/src/vector/unpack.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    let mut decoded = vec![0.0f32; padded_dim];
+    let mut inverse = vec![0.0f32; padded_dim];
+    let output_len = num_vectors
+        .checked_mul(dimensions)
+        .ok_or_else(|| vortex_err!("TurboQuant decoded vector length overflow"))?;
+    let mut output = BufferMut::<T>::with_capacity(output_len);
+
+    for i in 0..num_vectors {
+        if !mask.value(i) {
```
gatesn: In fact... maybe worth auditing for this everywhere!

vortex-turboquant/src/scalar_fns/encode.rs
connortsui20: There is an entire tests module

robert3005 (dismissed):
Can we use prs as atomic units instead of commits, please. Github isn't made for reviewing individual commits

connortsui20:
@joseph-isaacs how do you suggest I split this PR up?

The best alternative here that I can think of is that I split this into a) copy and paste a bunch of code from vortex-tensor, b) clean up, and c) add the scalar functions, which is frankly a waste of time if we know most of the code is going to change.

This is all due to the fact that we decided to change the design quite significantly but the implementation is still mostly the same.

vortex-turboquant/src/scalar_fns/metadata.rs
connortsui20: I mean we only use the struct as an intermediate type, these are not really methods

vortex-turboquant/src/scalar_fns/decode.rs
connortsui20: We do actually need this in order to (lossily) reconstruct the original vectors:

```rust
/// Configuration for lossy TurboQuant packing.
#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub struct TurboQuantConfig {
    bit_width: u8,
    seed: u64,
    num_rounds: u8,
}
```

None of this information can be taken from the dtype

vortex-turboquant/src/vector/normalize.rs
connortsui20: Might as well optimize this now while we are here since I can imagine a sparse column of vectors will not be encoded super well

connortsui20:
I will clean up my git history and we can figure out if it makes sense to split changes out of this or if everything together is required for context for reviewing

gatesn (approved):

PR #7831, opened 2026-05-07
bench[gpu]: CUDA device memory pool benchmarks

0ax1:
Adds benchmarks that compared pooled vs unpooled memory allocation. Main insight being that the memory release threshold for a pool can be adjusted by the `CU_MEMPOOL_ATTR_RELEASE_THRESHOLD` attribute.

```
cuda/core_primitives/device_alloc_reuse/default_pool/1GiB
                        time:   [31.125 ms 31.208 ms 31.292 ms]
                        change: [-5.1888% -2.7137% -0.3753%] (p = 0.04 < 0.05)
 
cuda/core_primitives/device_alloc_reuse/default_pool_75pct_threshold/1GiB
                        time:   [1.2864 µs 1.3728 µs 1.4528 µs]
                        change: [-7.2072% +0.4684% +8.9330%] (p = 0.91 > 0.05)
```

As part of this, `LOAD_SIZES` got renamed to `BUFFER_SIZE` for which we drop the loop as there's only a single buffer size value.

### DIFF
--- a/vortex-cuda/benches/core_primitives.rs
+++ b/vortex-cuda/benches/core_primitives.rs
@@ -20,11 +20,15 @@ use cudarc::driver::CudaStream;
 use cudarc::driver::HostSlice;
 use cudarc::driver::SyncOnDrop;
 use cudarc::driver::result;
+use cudarc::driver::sys;
 use cudarc::driver::sys::CU_MEMHOSTALLOC_WRITECOMBINED;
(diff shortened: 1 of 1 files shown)


0ax1:
Like the other allocation or copy only benchmarks, they do not run as part of Codspeed but are used to run locally to build an intuition for the performance of core primitives.

robert3005 (approved):

PR #7832, opened 2026-05-07
fsst: regression test for i32 offset overflow in fsst_compress

mprammer:
Regression test for #7833 — FSST `i32` offset overflow. Wrapped in `#[should_panic(expected = "to offset of type i32")]` so it merges green; drop the attribute alongside the fix and the trailing `assert_eq!(compressed.len(), len)` becomes the live assertion.

Gated `#[test_with::env(CI)]` + `#[test_with::no_env(VORTEX_SKIP_SLOW_TESTS)]` to match `vortex-btrblocks/src/schemes/integer.rs:1113`.

### DIFF
--- a/encodings/fsst/src/tests_large.rs
+++ b/encodings/fsst/src/tests_large.rs
@@ -18,8 +18,8 @@ use vortex_array::dtype::Nullability;
 use crate::fsst_compress;
 use crate::fsst_train_compressor;
 
-/// Regression: `fsst_compress` must accept inputs whose cumulative compressed
-/// bytes exceed `i32::MAX`. Today this panics in
+/// Regression for #7833: `fsst_compress` must accept inputs whose cumulative
+/// compressed bytes exceed `i32::MAX`. Today this panics in
 /// `vortex-array/src/arrays/varbin/builder.rs:62` because `fsst_compress_iter`
 /// (`encodings/fsst/src/compress.rs:72`) hardcodes `VarBinBuilder::<i32>` for
 /// the FSST output buffer regardless of input size.
--- a/encodings/fsst/src/tests_large.rs
+++ b/encodings/fsst/src/tests_large.rs
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Stress regression tests for FSST compression at the i32-offset boundary.
+//!
+//! Gated to CI runs (collected but skipped when `CI` is unset; opt-out with
+//! `VORTEX_SKIP_SLOW_TESTS=1`) because of the multi-GiB memory footprint.
+
+use rand::SeedableRng;
+use rand::rngs::StdRng;
+use rand::seq::IndexedRandom;
+use vortex_array::LEGACY_SESSION;
+use vortex_array::VortexSessionExecute;
+use vortex_array::arrays::varbin::builder::VarBinBuilder;
+use vortex_array::dtype::DType;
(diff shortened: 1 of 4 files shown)


encodings/fsst/src/tests_large.rs
```diff
@@ -0,0 +1,74 @@
...
+    const POOL_LEN: usize = 64 * 1024 * 1024;
+
+    // Printable ASCII alphabet so the result is valid UTF-8.
+    const ALPHABET: &[u8; 95] =
+        b" !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
+
+    let mut rng = StdRng::seed_from_u64(0xC0DE_C011_B711);
+    let pool: Vec<u8> = (0..POOL_LEN)
+        .map(|_| *ALPHABET.choose(&mut rng).unwrap())
+        .collect();
```
joseph-isaacs, line 55: do you need this? vs using a single char (so the test runs faster)

encodings/fsst/src/tests_large.rs
```diff
@@ -0,0 +1,74 @@
...
+    for i in 0..N {
+        let off = (i.wrapping_mul(31337)) % (POOL_LEN - STRING_LEN);
+        builder.append_value(&pool[off..off + STRING_LEN]);
+    }
+    let array = builder.finish(DType::Utf8(Nullability::NonNullable));
+
+    let compressor = fsst_train_compressor(&array);
+    let len = array.len();
+    let dtype = array.dtype().clone();
+    let mut ctx = LEGACY_SESSION.create_execution_ctx();
```
joseph-isaacs, line 67: could we directly build the fsst array to save some time?

joseph-isaacs (approved):

encodings/fsst/src/tests_large.rs
mprammer: Tried first: FSST collapses repetitive bytes, so the output never crosses 2 GiB.

encodings/fsst/src/tests_large.rs
mprammer: Bug is in fsst_compress_iter (compress.rs:72); constructing the array directly skips the panicking path.

gatesn:
Can we just.... also fix it??

connortsui20 (changes requested):
we should just fix this

connortsui20:
We should just write the offsets as i64 (as an aside, why do we use signed integers here and not unsigned?), since the compressor is going to narrow the codes regardless:

```rust
let utf8 = data.array_as_utf8().into_owned();
let compressor_fsst = fsst_train_compressor(&utf8);
let fsst = fsst_compress(&utf8, utf8.len(), utf8.dtype(), &compressor_fsst, exec_ctx);

let uncompressed_lengths_primitive = fsst
    .uncompressed_lengths()
    .clone()
    .execute::<PrimitiveArray>(exec_ctx)?
    .narrow()?;
let compressed_original_lengths = compressor.compress_child(
    &uncompressed_lengths_primitive.into_array(),
    &compress_ctx,
    self.id(),
    0,
    exec_ctx,
)?;

let codes_offsets_primitive = fsst
    .codes()
    .offsets()
    .clone()
    .execute::<PrimitiveArray>(exec_ctx)?
    .narrow()?;
let compressed_codes_offsets = compressor.compress_child(
    &codes_offsets_primitive.into_array(),
    &compress_ctx,
    self.id(),
    1,
    exec_ctx,
)?;
```

PR #7836, opened 2026-05-07
fix(fsst): pick i32 vs i64 codes offsets per call

### LINKED ISSUE #7833: FSST: `fsst_compress` panics on cumulative output >2 GiB (i32 offset overflow)
### What happened

`fsst_compress` panics when cumulative output bytes cross `i32::MAX`:

```
Other error: Failed to convert sum of N and M to offset of type i32
   at vortex-array/src/arrays/varbin/builder.rs:62
```

The output buffer is built with `VarBinBuilder::<i32>` regardless of input size — `encodings/fsst/src/compress.rs:72`. A related per-string i32 ceiling sits on `uncompressed_lengths` at `compress.rs:73,84`.

### Steps to reproduce

Regression test in #7832:

```
CI=1 cargo test --release -p vortex-fsst fsst_compress_offsets
```

connortsui20:
Fixes #7833.

`fsst_compress_iter` previously hardcoded `VarBinBuilder::<i32>` for the
FSST output, panicking once cumulative compressed bytes crossed
`i32::MAX`.

## Approach

`fsst_compress` now does an upfront pass over the input to compute the
total uncompressed byte count, then picks `VarBinBuilder<i32>` when the
FSST upper-bound compressed size (`2 * total + 7 * n`) fits in
`i32::MAX`, falling back to `VarBinBuilder<i64>` only when it would
actually overflow. Common case keeps the compact i32 layout; pathological
inputs are correctly handled.

`fsst_compress_iter` (single-pass iterator API) keeps its public
signature and now always uses i64. It can't size-estimate without
consuming the iterator, and direct callers are test-only.

`vortex-array/src/arrays/varbin/compute/compare.rs`: latent bug surfaced
by the i64 path. With i64 offsets, `Datum::try_new` produces an Arrow
`LargeBinary`/`LargeUtf8`, but the RHS scalar was hardcoded to
`Binary`/`Utf8`. Arrow refuses `LargeBinary == Binary`. The RHS now
picks the matching Arrow scalar from `lhs.data_type()`.

## Tests

The previous ~5 GiB `#[ignore]`d regression test is replaced with:
- Three boundary unit tests of the `upper_bound_fits_i32` helper.
- A small round-trip that asserts a fresh FSST array keeps `i32`
  `codes_offsets` for typical inputs.

The i64 path is exercised only for inputs whose worst-case compressed
size exceeds `i32::MAX`, which is too expensive to test directly; the
boundary unit tests cover the dispatch.

## Checks

- `cargo test -p vortex-fsst --lib` — 83 passed (was 78 before this PR).
- `cargo test -p vortex-array --lib arrays::varbin` — 93 passed.
- `cargo test -p vortex-btrblocks --lib` — 35 passed.
- `cargo +nightly fmt -- --check`, `cargo clippy -p vortex-fsst -p vortex-array --all-targets --all-features` — clean.
- `./scripts/public-api.sh` — no public-api.lock changes.

### DIFF
--- a/encodings/fsst/src/compress.rs
+++ b/encodings/fsst/src/compress.rs
@@ -21,6 +21,7 @@ use crate::FSSTArray;
 pub fn fsst_compress<A: ArrayAccessor<[u8]>>(
     strings: A,
     len: usize,
+    total_uncompressed: usize,
     dtype: &DType,
     compressor: &Compressor,
     ctx: &mut ExecutionCtx,
@@ -29,11 +30,7 @@ pub fn fsst_compress<A: ArrayAccessor<[u8]>>(
     // bounds the compressed size at `2 * uncompressed + 7` per string, so
     // if the upper bound fits in `i32::MAX` the actual offsets are
     // guaranteed to fit; otherwise we widen to `i64` to avoid the overflow
-    // tracked in #7833. The upfront pass over the iterator is cheap
-    // relative to the compression pass.
-    let total_uncompressed =
-        strings.with_iterator(|iter| iter.map(|opt| opt.map_or(0, <[u8]>::len)).sum::<usize>());
-
+    // tracked in #7833.
     if upper_bound_fits_i32(total_uncompressed, len) {
         strings.with_iterator(|iter| {
             fsst_compress_iter_with::<i32, _>(iter, len, dtype.clone(), compressor, ctx)
@@ -87,13 +84,13 @@ fn upper_bound_fits_i32(total_uncompressed: usize, len: usize) -> bool {
 
 /// Compress from an iterator of bytestrings using FSST.
 ///
-/// Single-pass callers that don't have access to the array's total
-/// uncompressed length always produce `i64` codes offsets for safety.
-/// Callers with an [`ArrayAccessor`] should prefer [`fsst_compress`],
-/// which picks `i32` offsets in the common case.
+/// `total_uncompressed` is the total byte length of all strings in the input;
+/// callers typically have it cheaply available (e.g. `VarBinArray::bytes().len()`).
+/// It selects the narrowest codes-offsets type per [`upper_bound_fits_i32`].
 pub fn fsst_compress_iter<'a, I>(
     iter: I,
     len: usize,
+    total_uncompressed: usize,
     dtype: DType,
     compressor: &Compressor,
     ctx: &mut ExecutionCtx,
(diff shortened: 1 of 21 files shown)


connortsui20:
@mprammer seems to work well

joseph-isaacs:
Can we please edit these descriptions

encodings/fsst/src/compress.rs
```diff
@@ -69,7 +69,10 @@ where
     I: Iterator<Item = Option<&'a [u8]>>,
 {
     let mut buffer = Vec::with_capacity(DEFAULT_BUFFER_LEN);
-    let mut builder = VarBinBuilder::<i32>::with_capacity(len);
+    // Offsets are widened to i64 because the cumulative compressed bytes can
+    // exceed i32::MAX for large inputs (see issue #7833). Per-string sizes
+    // still fit in i32.
+    let mut builder = VarBinBuilder::<i64>::with_capacity(len);
```
joseph-isaacs: really? we can do better than this

joseph-isaacs (changes requested):

encodings/fsst/src/tests.rs
```diff
@@ -107,3 +110,57 @@ fn test_fsst_array_ops() {
...
+///
+/// Marked `#[ignore]` because the test allocates ~2.5 GiB for the input and
+/// ~2.5 GiB for the FSST output (~5 GiB total), which is too much to run by
+/// default even in CI. To run it explicitly:
+///
+/// ```text
+/// cargo test --release -p vortex-fsst -- --ignored fsst_compress_offsets
+/// ```
+#[test]
+#[ignore = "allocates ~5 GiB; run with --ignored"]
```
joseph-isaacs: Why even have this test?

encodings/fsst/src/compress.rs
robert3005: For reader - we have uncrompressed lengths and we know the total uncompressed length, it's easy from that to figure out which one to use

connortsui20:
So this new approach just collects the sum of the lengths of the strings to figure out which builder to use. I am happy to clean this up later, but @mprammer since this is your issue feel free to pick this branch up. Seems to work well, but I don't have time to dig into this.

encodings/fsst/src/compress.rs
```diff
@@ -24,7 +25,24 @@ pub fn fsst_compress<A: ArrayAccessor<[u8]>>(
...
     ctx: &mut ExecutionCtx,
 ) -> FSSTArray {
-    strings.with_iterator(|iter| fsst_compress_iter(iter, len, dtype.clone(), compressor, ctx))
+    // Pick the narrowest sufficient codes-offsets type. The FSST contract
+    // bounds the compressed size at `2 * uncompressed + 7` per string, so
+    // if the upper bound fits in `i32::MAX` the actual offsets are
+    // guaranteed to fit; otherwise we widen to `i64` to avoid the overflow
+    // tracked in #7833. The upfront pass over the iterator is cheap
+    // relative to the compression pass.
+    let total_uncompressed =
```
joseph-isaacs: Is the signature wrong?

joseph-isaacs: Would changing the method sig help here?

connortsui20:
I dont think this is right, I think claude doesn't really know what it is doing. I think that just having i64 always is really the best solution here, and as the benchmarks showed before it had 0 affect on perf and no file changes

PR #7839, opened 2026-05-08
fix[bitpacked]: slice patches in execute method (not reduce).

joseph-isaacs:
Problem: BitPacked's SliceReduce implementation called patches.slice(), which reads buffers (search_index does binary search on buffer data, execute_scalar executes on chunk offsets). This violates the SliceReduce contract that requires metadata-only operations.

### DIFF
--- a/encodings/fastlanes/src/bitpacking/compute/slice.rs
+++ b/encodings/fastlanes/src/bitpacking/compute/slice.rs
@@ -10,6 +10,7 @@ use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
 use vortex_array::arrays::slice::SliceKernel;
 use vortex_array::arrays::slice::SliceReduce;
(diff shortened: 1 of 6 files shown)


encodings/fastlanes/src/bitpacking/compute/slice.rs
```diff
@@ -15,34 +17,57 @@ use crate::bitpacking::array::BitPackedArrayExt;
...
+            .flatten();
+
+        Ok(Some(slice_bitpacked(array, range, patches)?))
     }
 }
 
+fn slice_bitpacked(
+    array: ArrayView<'_, BitPacked>,
+    range: Range<usize>,
+    patches: Option<vortex_array::patches::Patches>,
```
robert3005: you want to import this

robert3005 (approved):

PR #7841, opened 2026-05-08
Sparse Dictionary Canonicalize

gatesn:
When the number of unique codes is much much smaller than the number of values, we should collect unique codes, filter the values, and remap the codes into the smaller dictionary's value space.

I don't expect this PR to immediately have any impact. But I have some other changes that propagate filter masks deeper into the tree so they actually hit the DictLayout


joseph-isaacs:
I think you want this idea: https://github.com/vortex-data/vortex/blob/74d803be99cec66c248d3a6855e518951fc98a13/vortex-array/src/arrays/dict/array.rs#L144 but could be sampled?

gatesn:
The problem is, we now run a rather expensive check on every canonicalize... Need to think about how we can constrain this some more. 

I think it may involve storing the approximate unique count of codes as an array statistic

vortex-array/src/arrays/dict/compute/slice.rs
```diff
@@ -12,25 +12,36 @@ use crate::arrays::Constant;
...
-                    .into_array(),
-                ))
+                };
+                if sliced_code.is_empty() {
+                    Ok(Some(sliced.into_array()))
+                } else {
+                    Ok(Some(unsafe {
+                        sliced.set_all_values_referenced(true).into_array()
+                    }))
+                }
```
robert3005, line 44: can't you return `values` in this code path? Am I missing something very obvious?

vortex-array/src/arrays/dict/vtable/mod.rs
```diff
@@ -186,6 +188,16 @@ impl VTable for Dict {
             )));
         }
 
+        if !array.has_all_values_referenced()
+            // `take(FilterArray(...))` is also represented as a dictionary. Compacting that shape
```
robert3005, line 192: note to self to check this when #7393 is ready

vortex-duckdb/src/exporter/cache.rs
```diff
@@ -20,6 +19,5 @@ use crate::duckdb::Vector;
 pub struct ConversionCache {
     pub dict_cache: DashMap<usize, (ArrayRef, ReusableDict)>,
     pub values_cache: DashMap<usize, (ArrayRef, Arc<Mutex<Vector>>)>,
-    pub canonical_cache: DashMap<usize, (ArrayRef, Canonical)>,
```
robert3005, line 23: little bit confused why this would be removed here

vortex-array/src/arrays/dict/vtable/sparse.rs
```diff
@@ -0,0 +1,321 @@
...
+use crate::arrays::Primitive;
+use crate::arrays::PrimitiveArray;
+use crate::arrays::dict::DictArraySlotsExt;
+use crate::builtins::ArrayBuiltins;
+use crate::dtype::DType;
+use crate::dtype::PType;
+use crate::executor::ExecutionCtx;
+use crate::validity::Validity;
+
+// TODO: Replace this fixed sparse-dictionary threshold with a cost model that accounts for values
```
robert3005, line 25: this should be an issue number

vortex-array/src/arrays/dict/vtable/sparse.rs
```diff
@@ -0,0 +1,321 @@
...
+    remapped_codes: PrimitiveArray,
+}
+
+#[cold]
+#[inline(never)]
+pub(super) fn sparse_canonicalize_dict(
+    array: &DictArray,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<Option<Canonical>> {
+    let codes = array.codes().as_::<Primitive>().into_owned();
```
robert3005, line 44: this looks wrong, and will panic in practice

robert3005: This will be fine but it's unclear from here that we have previously canonicalzed codes

vortex-array/src/arrays/dict/vtable/sparse.rs
```diff
@@ -0,0 +1,321 @@
...
+    let Some(sparse_codes) = collect_sparse_codes(&codes, array.values().len(), ctx)? else {
```
robert3005, line 45: It's weird that you're passing things by reference here if you made the value owned line before

vortex-array/src/arrays/dict/vtable/sparse.rs
```diff
@@ -0,0 +1,321 @@
...
+    codes_len.saturating_mul(SPARSE_CANONICALIZE_CODES_PER_VALUE_THRESHOLD) < values_len
+        || should_sample_sparse_canonicalize(codes_len, values_len)
+}
+
+#[inline]
+fn should_sample_sparse_canonicalize(codes_len: usize, values_len: usize) -> bool {
+    // Sampling is only a preflight for cases that are not sparse by row count alone. Keep it away
+    // from tiny dictionary domains and near-dense slices where the estimator overhead dominates.
+    values_len >= SPARSE_CANONICALIZE_MIN_SAMPLED_VALUES_LEN
+        && codes_len.saturating_mul(SPARSE_CANONICALIZE_SAMPLED_CODES_PER_VALUE_THRESHOLD)
```
robert3005, line 176: part of me would want to remove `SPARSE_CANONICALIZE` prefix so they're different

vortex-array/src/arrays/dict/vtable/sparse.rs
```diff
@@ -0,0 +1,321 @@
...
+            .unwrap_or_else(|_| vortex_panic!("dictionary code {code} does not fit usize"));
+        let new_code = value_remap[old_code];
+        debug_assert_ne!(new_code, usize::MAX);
+
+        remapped_codes.push(new_code as u64);
+    }
+
+    Ok(SparseDictCodes {
+        unique_codes: PrimitiveArray::new(Buffer::from_iter(unique_codes), Validity::NonNullable),
+        remapped_codes: PrimitiveArray::new(Buffer::from_iter(remapped_codes), validity),
```
robert3005, line 218: why are you copying here

robert3005 (changes requested):
I think beyond the exporter changes that are unclear to me why they're in this pr the rest of the code is fine, I mostly left some questions since it looks like too much llm code

Issue #6162, opened 2026-01-27
[reader-v2, layout reader] Nullable struct layouts are unsound


joseph-isaacs:
We currently we partition expressions over a struct layouts ignoring the struct validity. This causes undefined values to be examined by the inner expression.

The other problem is we cannot round trip a file with type `{a: {b: i32}?}` we get back `{a: {b: i32?}?}`

PR #7844, opened 2026-05-08
Stop ignoring tests

### LINKED ISSUE #7835: Run ignored tests on nightly CI
We should have an action that runs nightly and tests all of the ignored tests.

robert3005:
These were ignored for reasons unknown. I removed ignores, fixed one case of
actually failing test and converted one of them to an expected panic

fix #7835

### DIFF
--- a/vortex-array/src/scalar_fn/fns/get_item.rs
+++ b/vortex-array/src/scalar_fn/fns/get_item.rs
@@ -258,7 +258,7 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "apply() has a bug with null propagation from struct validity to non-nullable child fields"]
+    #[should_panic = "Cannot create null scalar with non-nullable dtype i32"]
     fn get_nullable_field() {
         let st = StructArray::try_new(
             FieldNames::from(["a"]),
--- a/vortex-array/src/arrays/filter/execute/listview.rs
+++ b/vortex-array/src/arrays/filter/execute/listview.rs
@@ -71,13 +71,14 @@ pub fn filter_listview(array: &ListViewArray, selection_mask: &Arc<MaskValues>) 
 
 #[cfg(test)]
 mod test {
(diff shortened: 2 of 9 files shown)


vortex-array/src/scalar_fn/fns/get_item.rs
```diff
@@ -258,7 +258,7 @@ mod tests {
     }
 
     #[test]
-    #[ignore = "apply() has a bug with null propagation from struct validity to non-nullable child fields"]
+    #[should_panic = "Cannot create null scalar with non-nullable dtype i32"]
```
connortsui20, line 261: This doesn't seem right? Is this actually the correct behavior?

connortsui20 (approved):

vortex-array/src/scalar_fn/fns/get_item.rs
robert3005: it's not the correct behaviour but instead of ignoring the test we let it run and assert it panics. This should be fixed and if someone fixes it this test will start failing

connortsui20: Sure, but now nobody is tracking this issue? And now we are overloading `should_panic` with both "this is a test that should panic" and "this is a bug that someone needs to fix, without an actual reason". That was why I had those ignores, because they are easy to find.

robert3005: Nobody was tracking this issue anyway, adding an ignore or should_panic is not tracking an issue. You need something else, for example this issue is https://github.com/vortex-data/vortex/issues/6162

connortsui20: My point was that we could easily grep for this, and now we can't.

I will create an issue: https://github.com/vortex-data/vortex/issues/7878

PR #7850, opened 2026-05-08
benchmarks-website: duckdb conn pool

a10y:
A single visit to the benchmarks website makes ~360 concurrent API requests to this axum server.

The previous `Arc<Mutex<Connection>>` was causing all of these requests to serialize (multiply by however many concurrent tabs had the benchmarks open).

This creates a connection pool of DuckDB conns, and adds a mutex on the ingestion endpoint to avoid attempting to open two concurrent DB connections for writing (which DuckDB does not allow and would cause an error).

### DIFF
--- a/benchmarks-website/server/src/app.rs
+++ b/benchmarks-website/server/src/app.rs
@@ -22,6 +22,7 @@ use anyhow::Result;
 use axum::Router;
 use axum::routing::get;
 use axum::routing::post;
+use tokio::sync::Mutex;
 use tower_http::compression::CompressionLayer;
 
 use crate::api;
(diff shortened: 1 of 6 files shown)


benchmarks-website/server/src/app.rs
```diff
@@ -35,8 +36,11 @@ use crate::ingest;
 /// or a small `String`).
 #[derive(Clone)]
 pub struct AppState {
-    /// Mutex-guarded DuckDB connection. See [`crate::db`].
+    /// r2d2 pool of DuckDB connections. See [`crate::db`].
     pub db: DbHandle,
+    /// Serializes `/api/ingest` so concurrent ingests can't race on the same
+    /// rows and trigger a DuckDB write-write conflict. Reads are unaffected.
+    pub ingest_lock: Arc<Mutex<()>>,
```
AdamGS, line 43: really calls out to a `RwLock` around the handle or something? put the valid states into the type system

a10y: i think we actually want a separate Connection for the ingest endpoint. it's ok for there to be multiple readonly conns in the process but there can only be one writable. so actually i should make this an `Arc<Mutex<Connection>>` and leave the pool for the other one in place.

a10y:
Closing in favor of just doing this in the other benchmarks PR

PR #7852, opened 2026-05-08
Use many connections on the benchmarks server

connortsui20:
## Summary

Test website: http://ec2-18-219-54-101.us-east-2.compute.amazonaws.com:3000/

Turns out we can just clone the connection and have each of them write (with retry). Now all threads can read and write concurrently instead of being serialized through a lock.

You can look at this [file](https://github.com/vortex-data/vortex/blob/8afd24b705cae0bd6ebe8fd09b1ab7cc93e5961f/benchmarks-website/server/ARCHITECTURE.md) to see the architecture of the new server.

## Testing

What's that?!


benchmarks-website/server/src/db.rs
```diff
@@ -31,8 +32,25 @@ use crate::records::RandomAccessTime;
...
+
+impl DbHandle {
+    fn new(root: Connection) -> Self {
+        Self {
+            root: Arc::new(Mutex::new(root)),
+        }
+    }
+
+    fn connection(&self) -> Result<Connection> {
+        let root = self.root.lock();
```
a10y, line 56: i don't love that this still requires locking, even though the criticial section is pretty short lived...guess it's fine 🤷

a10y (approved):

benchmarks-website/server/src/db.rs
connortsui20: I think that we cant really get around this? Also it is unlikely that this would have contention anyways

a10y: this method is called on every request right?

AdamGS: at this point, why do we even need the lock? can't we just clone here?

a10y: the type isn't send so it needs to be wrapped in something with interior mut

connortsui20: Oh I guess it might have some contention, but duckdb is going to have a lot of overhead in creating txns for all of these requests (which is further dominated by the actual queries) so I don't think it's a problem

connortsui20: these issues should all be resolved now

PR #7853, opened 2026-05-08
Widen `VarBinBuilder` offets during FSST compress

### LINKED ISSUE #7833: FSST: `fsst_compress` panics on cumulative output >2 GiB (i32 offset overflow)

connortsui20:
## Summary

Fixes: #7833

I don't think that a more complicated solution here makes a lot of sense. We narrow these offsets is the compressor anyways, so there is no reason to do something more complicated than this.

Additionally adds the logic for comparing with arrow large binary arrays that was missing before.

To be honest, I feel like it doesn't really make any sense to allow the user to specify what the offsets and sizes types should be at all, I think we can have the builder _always_ use the 64-bit types because we know that they will be compressed by the compressor anyways.

## Testing

Adds some regression tests. I've ignored the big test because it allocates too much memory.

You can run the ignored test with:

```
cargo t --release -p vortex-fsst varbin_compress_offsets_overflow_i32 -- --ignored --no-capture
```

### DIFF
--- a/encodings/fsst/src/compress.rs
+++ b/encodings/fsst/src/compress.rs
@@ -69,8 +69,12 @@ where
     I: Iterator<Item = Option<&'a [u8]>>,
 {
     let mut buffer = Vec::with_capacity(DEFAULT_BUFFER_LEN);
-    let mut builder = VarBinBuilder::<i32>::with_capacity(len);
+
+    // Offsets are widened to i64 because the cumulative compressed bytes can exceed i32::MAX for
+    // large inputs (see issue #7833). Per-string sizes still fit in i32.
+    let mut builder = VarBinBuilder::<i64>::with_capacity(len);
     let mut uncompressed_lengths: BufferMut<i32> = BufferMut::with_capacity(len);
+
     for string in iter {
         match string {
             None => {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10781,6 +10781,7 @@ dependencies = [
  "prost 0.14.3",
  "rand 0.10.1",
  "rstest",
+ "test-with",
  "vortex-array",
  "vortex-buffer",
  "vortex-error",
--- a/encodings/fsst/Cargo.toml
+++ b/encodings/fsst/Cargo.toml
@@ -33,6 +33,7 @@ _test-harness = ["dep:rand", "vortex-array/_test-harness"]
 divan = { workspace = true }
 rand = { workspace = true }
 rstest = { workspace = true }
+test-with = { workspace = true }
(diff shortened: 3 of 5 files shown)


joseph-isaacs (changes requested):
please have the test run in ci. what is the problem with doing it?

connortsui20:
Oh I forgot about test with

encodings/fsst/src/compress.rs
```diff
@@ -69,8 +69,12 @@ where
     I: Iterator<Item = Option<&'a [u8]>>,
 {
     let mut buffer = Vec::with_capacity(DEFAULT_BUFFER_LEN);
-    let mut builder = VarBinBuilder::<i32>::with_capacity(len);
+
+    // Offsets are widened to i64 because the cumulative compressed bytes can exceed i32::MAX for
+    // large inputs (see issue #7833). Per-string sizes still fit in i32.
+    let mut builder = VarBinBuilder::<i64>::with_capacity(len);
     let mut uncompressed_lengths: BufferMut<i32> = BufferMut::with_capacity(len);
```
robert3005, line 76: shouldn't this match?

connortsui20: To me the offsets aren't part of the "type" of the strings so its fine to just change this, but the lengths are definitely part of the "type" so if we want to allow 2 billion + length strings then that is a meaningful enough difference that we should discuss that in a separate PR

robert3005: I confused what this was representing, these aren't cumulative that's why they can stay u32

robert3005:
I have been looking at similar function in our dictionary encoding and I think we have made wrong assumptions. We should kill ArrayAccessor usage here and use the array directly. Then figuring out which width to choose for the downstream builder is easy

encodings/fsst/src/compress.rs
joseph-isaacs, line 76: we should at least justify this being i32?

joseph-isaacs (approved):

PR #7857, opened 2026-05-09
Faster Mask::from_slices

gatesn:
Do not iterate individual bits to construct a mask from slices

### DIFF
--- a/vortex-mask/src/lib.rs
+++ b/vortex-mask/src/lib.rs
@@ -281,10 +281,14 @@ impl Mask {
             return Self::AllTrue(len);
         }
 
-        let mut buf = BitBufferMut::new_unset(len);
+        let mut buf = BitBufferMut::with_capacity(len);
+        let mut cursor = 0;
         for (start, end) in slices.iter().copied() {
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

vortex-mask/src/lib.rs
```diff
@@ -281,10 +281,14 @@ impl Mask {
...
 
-        let mut buf = BitBufferMut::new_unset(len);
+        let mut buf = BitBufferMut::with_capacity(len);
+        let mut cursor = 0;
         for (start, end) in slices.iter().copied() {
-            (start..end).for_each(|idx| buf.set(idx));
+            buf.append_n(false, start - cursor);
+            buf.append_n(true, end - start);
+            cursor = end;
         }
```
joseph-isaacs, line 290: If there is a high bias, it would be better to set the one with a high true count and then only write the unset on

gatesn: I guess this is approximately knowable from the number of slices giving us an average run length?

PR #7876, opened 2026-05-11
Add more benchmarks for take on filter array

robert3005:
As part of optimising this execution flow we need to have benchmarks merged
before to reliably measure changes


robert3005:
These might need to be cut down in size, let's see what codspeed reports

joseph-isaacs (dismissed):
some of these benchmarks are too long, also do we need them all?

robert3005:
I think you're right, we can cover things with tests. I have gotten carried away optimising this and didn't consider whether things need to be benchmarks

joseph-isaacs:
we can add a few just not this many

joseph-isaacs:
there are still `2ms` runs

robert3005:
the slowest benchmark is now 400us

joseph-isaacs:
says there are 60 new benchmarks

robert3005:
but 24 were skipped because they have been renamed. Let me trim some more

robert3005:
I have cut it down to the essentials, it should be only couple extra benchmarks for variable length types like list

myrrc (approved):

PR #7877, opened 2026-05-11
Updated Variant array and the new VariantGet expression

AdamGS:
## Summary

This PR includes two big changes as Variant moves closer to readiness.

1. Potentially holding the `shredded` child of a variant array in the canonical VariantArray
2. A `VariantGet` expression that can pull extract data out of variant arrays, either in a typed way or as a more opaque `Variant`.

For reviewers, some relevant context might be:
1. The [VariantGet](https://github.com/vortex-data/rfcs/pull/58) RFC: this RFC takes some lessons I've learned working on this into account and reflects my updated view of this problem.
2. The original [Variant type](https://vortex-data.github.io/rfcs/rfc/0015.html) RFC

I think the Parquet spec is also a pretty good read and a very heavy influence of this work - [`Shredding`](https://parquet.apache.org/docs/file-format/types/variantshredding/) and the [`Variant type`](https://parquet.apache.org/docs/file-format/types/variantencoding/).


encodings/parquet-variant/src/array.rs
```diff
@@ -196,45 +126,206 @@ impl ParquetVariantData {
...
+    array: &Array<ParquetVariant>,
+) -> VortexResult<ArrayRef> {
+    // The spec requires at least one of `value`/`typed_value` to be present
+    // (matching the Arrow canonical extension contract). When we lift `typed_value` out into
+    // the outer `VariantArray::shredded` slot and the original had no `value`, synthesize an
+    // all-null `value` so the remaining `ParquetVariant` still satisfies that invariant and
+    // can round-trip back through `to_arrow`.
+    let value = array.value_array().cloned().or_else(|| {
+        array.typed_value_array().map(|_| {
+            VarBinViewArray::from_iter_nullable_bin(std::iter::repeat_n(None::<&[u8]>, array.len()))
```
robert3005: There must be a better way of doing this, this is just Constant null array?

vortex-array/src/canonical.rs
joseph-isaacs, line 758: want to add a todo for using `recursively_canonicalize_slots` for all arrays here.

vortex-array/src/scalar_fn/fns/variant_get/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+        for idx in 0..input.len() {
+            let scalar = input.execute_scalar(idx, ctx)?;
+            let output = variant_get_scalar(&scalar, options, &dtype)?;
+            chunks.push(ConstantArray::new(output, 1).into_array());
+        }
+
+        let array = ChunkedArray::try_new(chunks, dtype.clone())?.into_array();
+        if dtype.is_variant() {
+            VariantArray::try_new(array, None).map(|array| array.into_array())
+        } else {
```
joseph-isaacs: this looks very slow, can you add a comment explaining this?

vortex-array/src/scalar_fn/fns/variant_get/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+    fn return_dtype(&self, options: &Self::Options, arg_dtypes: &[DType]) -> VortexResult<DType> {
+        let input_dtype = &arg_dtypes[0];
+        vortex_ensure!(
+            matches!(input_dtype, DType::Variant(_)),
+            "VariantGet input must be Variant, found {input_dtype}"
+        );
+
+        Ok(options
+            .dtype()
+            .map_or(DType::Variant(Nullability::Nullable), DType::as_nullable))
```
joseph-isaacs, line 125: Why do you store as option dtype as possibly non-nullable? There is a hidden cast

encodings/parquet-variant/src/array.rs
```diff
@@ -196,45 +126,206 @@ impl ParquetVariantData {
...
             .map(|tv| ArrayRef::from_arrow(tv.as_ref(), typed_value_nullable))
             .transpose()?;
+        ParquetVariant::try_new(validity, metadata, value, typed_value).map(IntoArray::into_array)
+    }
+}
 
-        let pv = ParquetVariant::try_new(validity, metadata, value, typed_value)?;
-        Ok(VariantArray::new(pv.into_array()).into_array())
+pub(crate) fn core_storage_without_typed_value(
+    array: &Array<ParquetVariant>,
```
robert3005: This is a weird type, &ParquetVariantArray?

vortex-array/src/arrays/variant/mod.rs
```diff
@@ -15,30 +18,388 @@ use crate::array::EmptyArrayData;
...
+/// Accessors for canonical variant storage.
+///
+/// A canonical variant array keeps raw variant semantics in `core_storage` and may carry a
+/// row-aligned, storage-agnostic `shredded` typed tree for selected paths. The shredded child may
+/// have any dtype; its dtype is recorded during serialization and validated by normal child
+/// deserialization.
 pub trait VariantArrayExt: TypedArrayRef<Variant> {
-    fn child(&self) -> &ArrayRef {
-        self.as_ref().slots()[0]
+    /// Returns the raw storage that preserves the full variant value for every row.
```
joseph-isaacs: So this does contain shredded data again

vortex-array/src/arrays/variant/mod.rs
```diff
@@ -15,30 +18,388 @@ use crate::array::EmptyArrayData;
...
-        let len = child.len();
-        let stats = child.statistics().to_owned();
-        unsafe {
-            Array::from_parts_unchecked(
-                ArrayParts::new(Variant, dtype, len, EmptyArrayData)
-                    .with_slots(smallvec![Some(child)]),
-            )
+    pub fn new(core_storage: ArrayRef) -> Self {
+        Self::try_new(core_storage, None).vortex_expect("invalid VariantArray core_storage")
+    }
```
joseph-isaacs: These methods have odd names and parameters

vortex-array/src/arrays/variant/mod.rs
```diff
@@ -15,30 +18,388 @@ use crate::array::EmptyArrayData;
...
+    /// Returns the optional row-aligned typed shredded tree for selected variant paths.
+    fn shredded(&self) -> Option<&ArrayRef> {
+        self.as_ref().slots()[SHREDDED_SLOT].as_ref()
     }
 }
 impl<T: TypedArrayRef<Variant>> VariantArrayExt for T {}
 
 impl Array<Variant> {
+    /// Creates a new `VariantArray` with raw core storage and optional shredded storage.
+    pub fn try_new(core_storage: ArrayRef, shredded: Option<ArrayRef>) -> VortexResult<Self> {
```
joseph-isaacs, line 60: here explain or link to what you pass here

encodings/parquet-variant/src/operations.rs
```diff
@@ -198,7 +223,12 @@ fn scalar_from_shredded_object_scalar(
         if !unshredded.is_null() {
             let unshredded = unshredded.as_struct();
```
onursatici: `as_struct` here should not panic here as long as the underlying parquet array is not corrupt right? Any present unshredded data when there is a shredded value has to be a struct

vortex-array/src/arrays/variant/vtable/kernel.rs
```diff
@@ -0,0 +1,211 @@
...
+use crate::scalar_fn::fns::variant_get::VariantGetOptions;
+use crate::scalar_fn::fns::variant_get::VariantPath;
+use crate::scalar_fn::fns::variant_get::VariantPathElement;
+use crate::validity::Validity;
+
+pub(super) const PARENT_KERNELS: ParentKernelSet<Variant> =
+    ParentKernelSet::new(&[ParentKernelSet::lift(&VariantGetExecute)]);
+
+#[derive(Default, Debug)]
+struct VariantGetExecute;
```
robert3005: I would remove `Execute` suffix

robert3005:
@claude can you review this?

vortex-array/src/arrays/variant/vtable/kernel.rs
```diff
@@ -0,0 +1,211 @@
...
+        let Some(field) = current_struct.unmasked_field_by_name_opt(name.as_ref()) else {
+            return Ok(None);
+        };
+        current = mask_with_validity(field.clone(), current_struct.validity()?)?;
+    }
+
+    Ok(Some(current))
+}
+
+fn mask_with_validity(array: ArrayRef, validity: Validity) -> VortexResult<ArrayRef> {
```
robert3005: I would just call mask directly. it should deal with it

vortex-array/src/arrays/variant/vtable/kernel.rs
```diff
@@ -0,0 +1,211 @@
...
+    if validity.no_nulls() {
+        return Ok(array);
+    }
+
+    let len = array.len();
+    array.mask(validity.to_array(len))
+}
+
+fn all_valid(array: &ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<bool> {
```
robert3005: This is a weird optimisation, let's remove it for now

encodings/parquet-variant/src/vtable.rs
```diff
@@ -193,27 +258,26 @@ impl VTable for ParquetVariant {
...
+    fn execute(array: Array<Self>, ctx: &mut ExecutionCtx) -> VortexResult<ExecutionResult> {
+        let shredded = array
+            .typed_value_array()
+            .cloned()
+            .map(|typed_value| {
+                logical_shredded_from_parquet_typed_value(array.metadata_array(), typed_value, ctx)
+            })
+            .transpose()?;
+        let core_storage = core_storage_without_typed_value(&array)?;
         Ok(ExecutionResult::done(
```
onursatici, line 279: I know there must be lots of discussion about this and I think having variant as a dtype is fine, but it is still weird for me to have a fully shredded int array and we still have to canonicalise it into a vortex variant array with empty core_storage, instead of returning a primitive array

vortex-array/src/arrays/variant/vtable/operations.rs
```diff
@@ -16,6 +17,19 @@ impl OperationsVTable<Variant> for Variant {
...
+        if !core_storage.is_valid(index, ctx)? {
+            return Ok(Scalar::null(array.dtype().clone()));
+        }
+
+        let Some(shredded) = array.shredded() else {
+            return core_storage.execute_scalar(index, ctx);
+        };
+
+        let typed = shredded.execute_scalar(index, ctx)?;
+        let fallback = (typed.is_null() || typed.dtype().is_struct())
```
robert3005, line 31: Why struct as well?

vortex-array/src/scalar_fn/fns/variant_get/mod.rs
```diff
@@ -0,0 +1,701 @@
...
+            .map_or(DType::Variant(Nullability::Nullable), DType::as_nullable);
+        let mut chunks = Vec::with_capacity(input.len());
+
+        for idx in 0..input.len() {
+            let scalar = input.execute_scalar(idx, ctx)?;
+            let output = variant_get_scalar(&scalar, options, &dtype)?;
+            chunks.push(ConstantArray::new(output, 1).into_array());
+        }
+
+        let array = ChunkedArray::try_new(chunks, dtype.clone())?.into_array();
```
robert3005: this is very weird. How does this become canonical?

vortex-array/src/arrays/chunked/vtable/mod.rs
```diff
@@ -245,6 +245,7 @@ impl VTable for Chunked {
                 // TODO(joe)[#7674]: iterative execution here too
                 Ok(ExecutionResult::done(_canonicalize(array.as_view(), ctx)?))
             }
+            DType::Variant(..) => Ok(ExecutionResult::done(array)),
```
robert3005: This seems incorrect. Can you add a test for canonicalizing variant_get expression?

onursatici: aren't we potentially returning a non canonical array here? the executor would try to call execute repeatedly getting back the same array then bail failing to match the array with canonical?

robert3005 (commented):
I think everything apart from ChunkedArray(Variant) execute are just nits

vortex-array/src/arrays/masked/execute.rs
```diff
@@ -158,27 +158,38 @@ fn mask_validity_extension(
...
         }
         Validity::Array(_) => {
-            // Child has an array-backed validity stored as its first child.
+            // Core storage has an array-backed validity stored as its first child.
             // Combine with the mask and replace that child via with_children.
-            let combined = Validity::and(child_validity, validity)?;
-            let new_child = child.with_slot(0, combined.to_array(len))?;
-            Ok(VariantArray::new(new_child))
+            let combined = Validity::and(core_validity, validity)?;
+            core_storage.with_slot(0, combined.to_array(len))?
```
onursatici: should we assert slot 0 is validity?
core_storage is parquet variant I think and that should work, but if we have Chunked or VariantArray as the core_storage somehow then this will eithre panic or corrupt the array I think

vortex-array/src/arrays/variant/vtable/kernel.rs
```diff
@@ -0,0 +1,211 @@
...
+        if all_valid(&typed, ctx)? {
+            return Ok(Some(typed));
+        }
+
+        // Null typed rows are not necessarily missing from the logical variant value;
+        // fill those rows from core storage and keep valid typed rows unchanged.
+        let fallback = make_fallback(ctx)?;
+        let typed_mask = typed.is_not_null()?;
+        typed_mask
+            .zip(typed, fallback)?
```
onursatici, line 143: as a follow up, we are computing the entire `fallback` array here but only take indices where typed_mask is false, so we can narrow the core_storage to only the false indices and compute from there then zip

vortex-array/src/arrays/masked/execute.rs
AdamGS: yeah this was bad, fixed.

vortex-array/src/arrays/chunked/vtable/mod.rs
AdamGS: addressed

vortex-array/src/canonical.rs
```diff
@@ -1017,7 +1069,6 @@ impl Matcher for AnyCanonical {
             || array.is::<VarBinView>()
             || array.is::<Variant>()
             || array.is::<Extension>()
-            || array.is::<Variant>()
```
AdamGS, line 1020: this is just a type, it already exists two lines above

vortex-array/src/scalar_fn/fns/variant_get/mod.rs
AdamGS: This is the most general behavior we have, ideally we push it down into children that can do smarter stuff.

vortex-array/src/arrays/variant/mod.rs
AdamGS: I'll clarify, the idea is that the shredded data is in either one and works in both cases.

vortex-array/src/scalar_fn/fns/variant_get/mod.rs
AdamGS: `VariantGet` is inherently nullable because the path might not exist.

vortex-array/src/arrays/variant/mod.rs
AdamGS: can you be more specific?

vortex-array/src/arrays/variant/vtable/kernel.rs
AdamGS: will add a todo here

vortex-array/src/arrays/variant/vtable/operations.rs
AdamGS: Its basically
1. If the typed array is null
2. The shredded object is a struct which we want to merge together.

AdamGS: will add a comment

encodings/parquet-variant/src/vtable.rs
AdamGS: For `ParquetVariant` there's no real "empty" variant, and I think canonicalization should always maintain the DType. You can always get it out using `VariantGet` or maybe `cast` (which I think is what DuckDB does), we can figure out the semantics of that.

robert3005 (approved):

AdamGS:
I'm not merging this yet because I'm some stuff on top of it, and letting more people weigh in.

onursatici (approved):

PR #7880, opened 2026-05-11
Wire Vector into Arrow conversion

palaska:
Register `Vector` as an Arrow import/export plugin on the new `ArrowSession`, so a Vortex Vector extension array can go to Arrow and back without losing its identity.

Note: If the plugin is not registered, imports fall back to the plain FSL (no error, but the Vector identity is dropped).

### DIFF
--- a/vortex-tensor/src/types/vector/arrow.rs
+++ b/vortex-tensor/src/types/vector/arrow.rs
@@ -1,0 +1,398 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Arrow plugin impls for the [`Vector`] extension type.
+//!
+//! [`Vector`] stores `FixedSizeList<float, N>` and is round-tripped through Arrow as a
+//! `FixedSizeList<float, N>` carrying the `vortex.tensor.vector` extension name on the field
+//! metadata. The element layout is identical on both sides, so the conversion is structural.
+
+use std::sync::Arc;
+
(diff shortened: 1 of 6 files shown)


vortex-tensor/src/types/vector/arrow.rs
```diff
@@ -0,0 +1,398 @@
...
+use crate::types::vector::Vector;
+
+/// Arrow extension name used to identify [`Vector`] fields on the wire.
+pub const ARROW_VECTOR_EXTENSION_NAME: &str = "vortex.tensor.vector";
+
+static ARROW_VECTOR: CachedId = CachedId::new(ARROW_VECTOR_EXTENSION_NAME);
+
+// Arrow's `Field::set_metadata` takes a `std::collections::HashMap<String, String>`, so we
+// emit that type here instead of the workspace `vortex_utils::aliases::hash_map::HashMap`.
+#[allow(clippy::disallowed_types)]
```
connortsui20: Nit: can you change this to `expect(` with a `reason = `?

a10y (approved):
nice

connortsui20 (commented):
This seems great to me, but I did not work on the arrow exporter stuff so I feel like I'm not the best person to review this

vortex-tensor/src/types/vector/arrow.rs
joseph-isaacs: this

joseph-isaacs (changes requested):

PR #7884, opened 2026-05-11
perf[layout]: stop using dict layout if it offers poor compression

joseph-isaacs:


onursatici (commented):
I wonder why clickbench files increased in size, do you know if the fallback layout still compresses them as dict arrays?

PR #7887, opened 2026-05-11
fix[array]: get_item reduce + struct wrong nullability

### LINKED ISSUE #7878: `apply()` nullability bug
### What happened?

`apply()` has a bug with null propagation from struct validity to non-nullable child fields.

This test should pass, but it panics now.

###### `vortex-array/src/scalar_fn/fns/get_item.rs`

```rust
    fn get_nullable_field() {
        let st = StructArray::try_new(
            FieldNames::from(["a"]),
            vec![buffer![1i32].into_array()],
            1,
            Validity::AllInvalid,
        )
        .unwrap()
        .into_array();

        let get_item_expr = get_item("a", root());
        let item = st.apply(&get_item_expr).unwrap();
        // The dtype should be nullable since it inherits struct validity
        assert_eq!(
            item.dtype(),
            &DType::Primitive(PType::I32, Nullability::Nullable)
        );
    }
```

### Steps to reproduce

See tests above

### Environment

All environments

### Additional context

https://github.com/vortex-data/vortex/pull/7844/

joseph-isaacs:
fixes: https://github.com/vortex-data/vortex/issues/7878

### DIFF
--- a/vortex-array/src/arrays/struct_/compute/rules.rs
+++ b/vortex-array/src/arrays/struct_/compute/rules.rs
@@ -109,7 +109,7 @@ impl ArrayParentReduceRule<Struct> for StructGetItemRule {
             Validity::AllInvalid => {
                 // If everything is invalid, the field is also all invalid
                 Ok(Some(
-                    ConstantArray::new(Scalar::null(field.dtype().clone()), field.len())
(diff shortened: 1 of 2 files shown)


robert3005 (approved):
and here I thought this is more complicated

PR #7889, opened 2026-05-11
Reorder agg kernel dispatch, and have Combined use inner accumulators

gatesn:
Aggregate kernel dispatch previously tried the vtable before the plugin registry, meaning plugins could not intercept the aggregate function as they can with array execution.

This change also fixed Combined to use inner accumulators, instead of just inner partials. This fixes the kernel dispatch for the delegated aggregate functions. e.g. a custom Sum<->Dict kernel previously would not have been used for Combiner<Avg><->Dict, even though it delegates to Sum + Count.

### DIFF
--- a/vortex-array/src/aggregate_fn/accumulator.rs
+++ b/vortex-array/src/aggregate_fn/accumulator.rs
@@ -151,12 +151,12 @@ impl<V: AggregateFnVTable> DynAccumulator for Accumulator<V> {
             return Ok(());
         }
 
-        // 3. Iteratively execute one step at a time, re-checking the kernel registry against
-        //    each intermediate encoding. (The initial batch's encoding was already checked in
-        //    step 1, so execute first.)
+        // 3. Iteratively check the registry against each intermediate encoding, executing one
+        //    step between checks. Mirrors the loop in `GroupedAccumulator::accumulate_list_view`.
+        //    Iteration 0 re-checks the initial encoding — a redundant HashMap miss, the price of
+        //    keeping the loop body uniform.
         let mut batch = batch.clone();
         for _ in 0..max_iterations() {
-            batch = batch.execute(ctx)?;
             if batch.is::<AnyCanonical>() {
                 break;
             }
@@ -180,6 +180,8 @@ impl<V: AggregateFnVTable> DynAccumulator for Accumulator<V> {
                 self.vtable.combine_partials(&mut self.partial, result)?;
                 return Ok(());
             }
+
+            batch = batch.execute(ctx)?;
         }
 
         // 4. Otherwise, execute the batch until it is columnar and accumulate it into the state.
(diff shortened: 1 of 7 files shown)


vortex-array/src/aggregate_fn/accumulator.rs
```diff
@@ -99,14 +116,45 @@ impl<V: AggregateFnVTable> DynAccumulator for Accumulator<V> {
...
+                );
+                self.vtable.combine_partials(&mut self.partial, result)?;
+                return Ok(());
+            }
+        }
+
+        // 2. Allow the vtable to short-circuit on the raw array before decompression.
         if self.vtable.try_accumulate(&mut self.partial, batch, ctx)? {
             return Ok(());
         }
```
joseph-isaacs, line 166: we later want to move this also to the session?

vortex-array/src/arrays/constant/compute/uncompressed_size.rs
```diff
@@ -31,6 +35,17 @@ impl DynAggregateKernel for ConstantUncompressedSizeKernel {
...
+        // Honor a cached exact stat if present. This preserves the optimization that
+        // `UncompressedSizeInBytes::try_accumulate` provides for unregistered encodings;
+        // without it, the kernel would always recompute the size and ignore the cache.
+        if let Some(Precision::Exact(size_scalar)) =
+            batch.statistics().get(Stat::UncompressedSizeInBytes)
+        {
+            let size = u64::try_from(&size_scalar)
+                .map_err(|e| vortex_err!("Failed to convert uncompressed size stat to u64: {e}"))?;
+            return Ok(Some(Scalar::primitive(size, Nullability::NonNullable)));
+        }
```
joseph-isaacs: Can we move this out of the kernel and into a accessor layer above this?

Maybe we cannot but this makes the compute hard to reason about. I guess we don't set this which is better!

vortex-array/src/aggregate_fn/accumulator.rs
gatesn: As in, we want even canonical dispatch to be via the registry?

vortex-array/src/arrays/constant/compute/uncompressed_size.rs
joseph-isaacs, line 1: where did this go do we instead execute with columnar?

gatesn: We were always supposed to. This only existed I think because of the stats lookup

gatesn: https://github.com/vortex-data/vortex/pull/7889/changes#diff-1ef8d0153ab1d80b1932558ccb589325eede81d32a712f919e9d74745ebe2244R172-R174

joseph-isaacs (approved):

vortex-array/src/aggregate_fn/accumulator.rs
```diff
@@ -99,31 +116,75 @@ impl<V: AggregateFnVTable> DynAccumulator for Accumulator<V> {
...
+                );
+                self.vtable.combine_partials(&mut self.partial, result)?;
+                return Ok(());
+            }
+        }
+
+        // 2. Allow the vtable to short-circuit on the raw array before decompression.
+        if self.vtable.try_accumulate(&mut self.partial, batch, ctx)? {
+            return Ok(());
+        }
```
robert3005, line 166: why is this whole block different from the block below?

vortex-array/src/aggregate_fn/accumulator.rs
joseph-isaacs: Well it depends how you want pluggable semantics

vortex-array/src/aggregate_fn/accumulator.rs
joseph-isaacs: Looks like there are two parts, one `try_accumulate` is only tried once and two there a duplicated code..

Issue #7882, opened 2026-05-11
Tracking Issue: Add `DType::Union`


connortsui20:
This is a tracking issue for adding the `DType::Union` variant and canonical encoding for `Union`.

## Motivation

See the parent epic issue: https://github.com/vortex-data/vortex/issues/7705

## Design

We will be following the [Arrow specification of Union](https://arrow.apache.org/docs/format/Columnar.html#union-layout), though we have an important decision to make here. Arrow specifies both a dense and a sparse union layout. Both layouts (encodings in Vortex) would have child arrays for each type that a value could be, as well as a types child array that specifies which type in the `Union` array a given row is (and thus, which child array to look at to fetch the correct value).

Note that we cannot reuse the `StructFields` type for `Union` as we need to also track the `i8` tags for each of the possible variants of the union.

### Dense vs Sparse

[Dense](https://arrow.apache.org/docs/format/Columnar.html#dense-union) encoding: each child array contains only the values that actually belong to that type, packed contiguously. An additional `offsets` buffer tells us which slot of the chosen child holds the value for that row.

Slicing and filtering requires walking `type_ids` to recompute per-child offsets.

```
type_ids: [0, 1, 0, 1, 0]
offsets:  [0, 0, 1, 1, 2]
child 0 (int):    [10, 20, 30]
child 1 (string): ["a", "b"]
```

[Sparse](https://arrow.apache.org/docs/format/Columnar.html#sparse-union) encoding: every child array has the same length as the union itself, so the value at row `i` is simply found at position `i` of child `type_ids[i]`.

The "inactive" slots of each child are unused but must still be addressable. There is no `offsets` buffer.

```
type_ids: [0, 1, 0, 1, 0]
child 0 (int):    [10, _, 20, _, 30]   // _ slots are inactive
child 1 (string): [_, "a", _, "b", _]
```

### Why Sparse should be Canonical (and not Dense)

We want to support quick execution of things like slice and filter on our canonical arrays, and sparse is clearly superior to dense in this case (since dense would have to perform compute just to figure out how to slice child arrays).

Additionally, the space loss of sparse in Arrow does not necessarily translate to Vortex because our canonical types do not require the children to also be canonical. This means that actually sparse children under a sparse `Union` (for example, if I have 99000 integers and 1000 strings all separated) can be encoded with a `SparseArray`, so we would not lose much space compared to the dense union encoding.

[...]

PR #7890, opened 2026-05-11
Add `DType::Union` and `UnionVariants`

connortsui20:
## Summary

Tracking issue: https://github.com/vortex-data/vortex/issues/7882

Adds the scaffolding and boilerplate for the new logical `DType::Union` type.

This is the first PR of probably a bunch, so it simply adds the `DType::Union` variant and the `UnionVariants` type that `DType::Union` holds. Other than that, it is mostly just proto and flatbuffers code and other boilerplate that is easy to implement.

A lot of the code is essentially duplicated from implementations of `Struct`. It could have been nice if we could reuse `StructFields`, but we need to additionally store a type IDs field that maps the different variant types to tag indexes (i.e. 0, 1, 2, 5), so we might as well have a new type for this that is named more appropriately.

Also note that the tags do not have to be consecutive (to support potential schema evolution, and also it does make things like casts easier).

I've also enforced that variant names are distinct. The Arrow spec does not make any guarantee about this, but I've chosen the sane path that DuckDB also [chooses](https://duckdb.org/docs/current/sql/data_types/union).

Finally, I reordered some of the dtype match statements so that struct and union are always next to each other after list and fsl. This did cause a bit of extra churn... but with so many types now I think it is important that similar things are next to each other.

Edit: There's 1000+ lines of flatbuffers diff because I think my local `flatc` install is different from the one on CI?

## API Changes

Adds `Union` variant to `DType`

## Testing

Basic testing (mostly for the serialization roundtrips).


vortex-array/src/aggregate_fn/fns/uncompressed_size_in_bytes/mod.rs
```diff
@@ -287,22 +288,25 @@ fn checked_len_mul(len: usize, width: usize, name: &str) -> VortexResult<u64> {
...
+        | DType::Binary(_) => true,
         DType::List(element_dtype, _) | DType::FixedSizeList(element_dtype, ..) => {
             supports_uncompressed_size_in_bytes(element_dtype)
         }
         DType::Struct(fields, _) => fields
             .fields()
             .all(|field| supports_uncompressed_size_in_bytes(&field)),
+        DType::Union(variants, _) => variants
+            .variants()
+            .all(|variant| supports_uncompressed_size_in_bytes(&variant)),
```
connortsui20, line 299: Happy to remove this until the actual uncompressed_size_in_bytes is implemented later?

fuzz/src/array/mod.rs
```diff
@@ -510,12 +495,28 @@ fn actions_for_dtype(dtype: &DType) -> HashSet<ActionType> {
...
+            // Struct supports: Compress, Slice, Take, Filter, MinMax, Mask, ScalarAt
+            // Does NOT support: SearchSorted (requires scalar comparison), Compare, Cast, Sum, FillNull
+            let struct_actions = [Compress, Slice, Take, Filter, MinMax, Mask, ScalarAt];
+            sdt.fields()
+                .map(|child| actions_for_dtype(&child))
+                .fold(struct_actions.into(), |acc, actions| {
+                    acc.intersection(&actions).copied().collect()
+                })
+        }
+        DType::Union(..) => todo!("TODO(connor)[Union]: unimplemented"),
```
connortsui20: This is the only real change in this file

vortex-array/src/scalar/scalar_impl.rs
```diff
@@ -188,8 +188,11 @@ impl Scalar {
             DType::Utf8(_) => value.as_utf8().is_empty(),
             DType::Binary(_) => value.as_binary().is_empty(),
             DType::List(..) => value.as_list().is_empty(),
+            // TODO(connor): This seems wrong...
             DType::FixedSizeList(_, list_size, _) => value.as_list().len() == *list_size as usize,
+            // TODO(connor): This also seems wrong...
             DType::Struct(struct_fields, _) => value.as_list().len() == struct_fields.nfields(),
```
connortsui20, line 194: https://github.com/vortex-data/vortex/issues/7891

connortsui20:
I think my local version of `flatc` is wrong? Or rather its just different from CI

```
❯ flatc --version
flatc version 25.12.19
```

fuzz/src/array/mod.rs
joseph-isaacs: then please don't do this in the same pr

connortsui20: https://github.com/vortex-data/vortex/pull/7897

connortsui20:
https://github.com/vortex-data/vortex/pull/7899

fuzz/src/array/mod.rs
connortsui20: done

vortex-array/src/dtype/union.rs
```diff
@@ -0,0 +1,623 @@
...
+///
+/// A `UnionVariants` describes the possible alternative types for each row of a union, along with a
+/// per-variant `i8` type tag. We use the term **variants** (rather than "fields") because a union
+/// is a sum type: each row chooses exactly one alternative.
+///
+/// Per Arrow's spec, the per-row type tag is an `int8`. By default, tag `i` selects the child at
+/// offset `i` (`type_ids = [0, 1, ..., N-1]`). Schemas may also use non-consecutive tags (e.g.
+/// `[0, 5, 7]`), in which case the value of `type_ids[i]` is the tag used in the data to select the
+/// child at offset `i`. Supporting non-consecutive tags from v1 lets the schema remove children
+/// without renumbering the remaining tags.
```
joseph-isaacs, line 32: why i8 not u8? 

Its worth explaining why we have type_ids and non-consecutive tags too?

vortex-array/src/dtype/union.rs
```diff
@@ -0,0 +1,623 @@
...
+///
+/// Variant names must be distinct. Unlike [`StructFields`](crate::dtype::StructFields), which
+/// permits duplicate field names for Arrow/Parquet round-trip fidelity, duplicates have no
+/// meaningful semantics in a sum type and are rejected at construction.
+///
```
joseph-isaacs, line 37: so we cannot round trip arrow?

If you use index access not name then it does make sense to have duplicate fields. Fields are just metadata in arrow.

vortex-array/src/dtype/struct_.rs
```diff
@@ -138,8 +138,11 @@ impl FieldDTypeInner {
 /// // Accessing a field by name will yield the first
 /// assert_eq!(fields.field("int_col").unwrap(), DType::Primitive(PType::I32, Nullability::Nullable));
 /// ```
+#[allow(
```
joseph-isaacs, line 141: expect please

connortsui20: expect doesnt work here (for some reason, I have no idea)

vortex-array/src/dtype/union.rs
connortsui20: Arrow uses i8 and the maximum number of variants is 128 so this is fine imo

connortsui20:
@joseph-isaacs wants to split this out, so I will remove `UnionVariants` from this PR

PR #7892, opened 2026-05-12
tracing logger for duckdb

### LINKED ISSUE #6820: DuckDB: set up tracing:: subscriber
We have calls to tracing:: logger inside Vortex but we can't use it in DuckDB as we don't have any subscriber.
tracing_subscriber is quite a big dependency so I'd like to avoid using it.

We'd need to write our own subscriber and initialize it when opening a database.
It should check DuckDB's own logging level and filter out our messages accordingly.

myrrc:
- Add stdout logger for Vortex in duckdb (we use it only for debugging anyway).
- Move some log messages from debug to trace
- Don't cast file_row_number to i64, use u64 as target column.
- Fix dynamic filter printing

Resolves #6820

### DIFF
--- a/vortex-duckdb/src/lib.rs
+++ b/vortex-duckdb/src/lib.rs
@@ -6,6 +6,7 @@
 use std::ffi::CStr;
 use std::ffi::c_char;
 use std::sync::LazyLock;
+use std::sync::OnceLock;
 
 use vortex::VortexSessionDefault;
 use vortex::error::VortexExpect;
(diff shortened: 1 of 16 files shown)


joseph-isaacs (dismissed):
Please put this at the top level

AdamGS (approved):

vortex-duckdb/src/lib.rs
```diff
@@ -45,6 +46,20 @@ static RUNTIME: LazyLock<CurrentThreadRuntime> = LazyLock::new(CurrentThreadRunt
 static SESSION: LazyLock<VortexSession> =
     LazyLock::new(|| VortexSession::default().with_handle(RUNTIME.handle()));
 
+// Duckdb's logger requires a *Context as first argument which
+// would be hard to integrate with tracing::. We use logging for
+// debugging only anyway, so that's good enough.
+fn init_tracing() {
```
AdamGS, line 52: you don't need a lock here, `try_init` already has a lock under the hood, you can just do:
```
_ = tracing_subscriber::fmt()
                .with_writer(std::io::stdout)
                .try_init();
```

PR #7893, opened 2026-05-12
faster cast

joseph-isaacs:
## Summary

Closes: #000


robert3005:
in codspeed those two look the same. I wonder if this is because of callgrind

PR #7894, opened 2026-05-12
Add Executor::spawn_io

AdamGS:
## Summary

Allow `Executor` implementation to split their async IO and CPU work on different runtimes.

### DIFF
--- a/vortex-io/src/object_store/read_at.rs
+++ b/vortex-io/src/object_store/read_at.rs
@@ -118,63 +118,145 @@ impl VortexReadAt for ObjectStoreReadAt {
         let range = offset..(offset + length as u64);
 
         async move {
-            let mut buffer = allocator.allocate(length, alignment)?;
+            let blocking_handle = handle.clone();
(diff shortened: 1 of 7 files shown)


gatesn (approved):

vortex-io/src/object_store/read_at.rs
```diff
@@ -118,63 +118,145 @@ impl VortexReadAt for ObjectStoreReadAt {
...
-
-                    buffer
-                }
-            };
-
-            Ok(BufferHandle::new_host(buffer.freeze()))
+            let blocking_handle = handle.clone();
+
+            handle
+                .spawn_io(async move {
```
robert3005, line 124: if you're doing this can you then remove the outer async move?

AdamGS: sounds about right, let me take a look

AdamGS: done

PR #7896, opened 2026-05-12
Switch python runtime to CurrentThreadRuntime

robert3005:
We want to unify the language bindings to have the same behaviour when interacting with vortex. This pr brings python bindings in line with C and Java in using CurrentThreadRuntime by default

Vortex uses shared runtime underneath python api. When no background threads are configured the python thread drives the work on the scan. This means multiple Python threads can make progress independently as long as each thread owns the reader it is consuming

```python
from concurrent.futures import ThreadPoolExecutor

import pyarrow.compute as pc
import vortex as vx


def sum_column(path: str, column: str) -> int | float:
    reader = vx.open(path).to_arrow([column], batch_size=64_000)
    total = 0

    for batch in reader:
        value = pc.sum(batch.column(column)).as_py()
        if value is not None:
            total += value

    return total


columns = ["tip_amount", "fare_amount", "total_amount"]
with ThreadPoolExecutor(max_workers=len(columns)) as threads:
    totals = list(threads.map(lambda column: sum_column("example.vortex", column), columns))
```

Alternatively users who want vortex to work in the background, independently of user level python threads, can configure worker count to desired value.

```python
import vortex as vx
import vortex.runtime as vxrt

previous_workers = vxrt.worker_count()
vxrt.set_worker_threads_to_available_parallelism()

try:
    reader = vx.open("example.vortex").to_arrow(batch_size=64_000)
    table = reader.read_all()
finally:
    vxrt.set_worker_threads(previous_workers)
```

These examples are added to the docs


gatesn:
I'm not sure we necessarily want to expose session? I was making the decision that in Python land we can just assume a single global instance? 

Also would you mind adding to the PR description how multi-threading works in the new design? Do users have to have multiple python threads? Or is there a default background pool? Can I start one?

robert3005:
I think one global session is fine as long as we don't support free threaded python. In free threaded model python is more like any other language but also I don't know anyone yet using python 3.14.

docs/user-guide/vortex-python.md
```diff
@@ -186,6 +188,54 @@ tip_amount: double
...
+
+Applications that want Vortex work to continue on background threads can configure the shared
+runtime through `vortex.runtime`:
+
+```python
+import vortex as vx
+import vortex.runtime as vxrt
+
+previous_workers = vxrt.worker_count()
+vxrt.set_worker_threads_to_available_parallelism()
```
gatesn: I don't particularly like this API...

I think the default should be to have a background pool, and to respect VORTEX_MAX_THREADS=n env var

I think you want `vx.set_worker_threads(n=None)`, where None resets to (core count - 1). 

So no separate runtime module. Too much namespacing for canonical Python

vortex-python/python/vortex/_lib/runtime.pyi
```diff
@@ -0,0 +1,6 @@
+#  SPDX-License-Identifier: Apache-2.0
+#  SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+def set_worker_threads(n: int) -> None: ...
+def set_worker_threads_to_available_parallelism() -> None: ...
```
gatesn: Remove this func

vortex-python/python/vortex/_lib/runtime.pyi
```diff
@@ -0,0 +1,6 @@
...
+def worker_count() -> int: ...
```
gatesn: set_worker_threads => worker_threads() , not worker_count

vortex-python/python/vortex/arrays.py
```diff
@@ -78,6 +78,7 @@ def _Array_to_arrow_table(self: _arrays.Array) -> pyarrow.Table:
     Examples
     --------
 
+    >>> import vortex
```
gatesn: We need all this? And is it not import vortex as vx?

gatesn (changes requested):
I think we should default to `core count - 1`, and allow the user to adjust from there

docs/user-guide/vortex-python.md
robert3005: Fair, it's a bit unnatural for python to have to spin up threads just to call some native method

gatesn (approved):

PR #7898, opened 2026-05-12
substr expression

myrrc:


myrrc:
No speedup, closing

vortex-array/src/scalar_fn/fns/substring/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+        } else {
+            None
+        };
+
+        let nullable = string_arr.dtype().is_nullable();
+        // Execute string array to Arrow as Utf8: arrow_string::substring does
+        // not support Utf8View.
+        let arrow_array = string_arr.execute_arrow(Some(&DataType::Utf8), ctx)?;
+        let result = arrow_string::substring::substring(arrow_array.as_ref(), start - 1, length)?;
+        from_arrow_array_with_len(result.as_ref(), len, nullable)
```
joseph-isaacs: this is likely very slow, we can do this via varbinview offset rewrite and we should apply push down to each encoding type

vortex-array/src/arrays/varbinview/compute/substring.rs
```diff
@@ -0,0 +1,111 @@
...
+                        .unwrap_or(total_len);
+                    let new_len = end - start;
+                    let new_offset = r.offset as usize + start;
+
+                    if new_len <= BinaryView::MAX_INLINED_SIZE {
+                        let buf = array.buffer(r.buffer_index as usize);
+                        BinaryView::new_inlined(&buf[new_offset..new_offset + new_len])
+                    } else {
+                        let buf = array.buffer(r.buffer_index as usize);
+                        let prefix: [u8; 4] = buf[new_offset..new_offset + 4]
```
joseph-isaacs, line 82: these will fit

joseph-isaacs:
worth looking at arrow-rs here. You can delegate to arrow-re or at least compare as a baseline

substring(array, start: u64, length: Option<u64>).

PR #7899, opened 2026-05-12
Update flatbuffers generated

connortsui20:
## Summary

I am not sure why but my local flatc produces different results than CI. Do we need to update something?

This happens on both my linux machine and mac.

```
❯ flatc --version
flatc version 25.12.19
```

## Testing

N/A


connortsui20:
This is because of https://github.com/google/flatbuffers/pull/8564 and https://github.com/google/flatbuffers/pull/8828 which is somewhat recent

robert3005:
You have to update versions used by CI as well

.github/actions/setup-flatc/action.yml
```diff
@@ -3,7 +3,7 @@ description: "Download and install flatc binary"
 inputs:
   flatc_version:
     description: "Version of the flatc binary"
-    default: "25.9.23"
+    default: "25.12.19"
```
robert3005, line 6: this doesn't get run on our ami images, you will have to update them and then make a pr. Annoyingly there will be an amount of time when all prs will fail because of it :/

connortsui20: building now: https://github.com/vortex-data/.github-private/actions/runs/25746194678/job/75609981370

connortsui20:
Do I need to rerun this: https://github.com/vortex-data/.github-private/actions/runs/25746194678/job/75609981444

robert3005 (approved):

PR #7900, opened 2026-05-12
refactor[fsst]: take ArrayRef in fsst_compress, decide offset width upfront

mprammer:
## Summary
Refactors `fsst_compress` and `fsst_train_compressor` to take an `ArrayRef` and dispatch internally on the input encoding (`VarBinView` or `VarBin`), so callers don't need to know which string encoding they hold. The codes-offsets width (i32 vs i64) is decided upfront from the FSST worst-case bound (2 × input bytes + 7 per non-null row), which always covers the actual output regardless of how well the data compresses.

Follow-up to #7853 addressing review feedback on #7833.

### DIFF
--- a/encodings/fsst/src/compress.rs
+++ b/encodings/fsst/src/compress.rs
@@ -1,160 +1,359 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-// Compress a set of values into an Array.
+//! FSST compression entry points.
+//!
+//! [`fsst_compress`] and [`fsst_train_compressor`] take an [`ArrayRef`] and dispatch
+//! on the input encoding ([`VarBinView`] or [`VarBin`]). Callers don't need to know
+//! which string encoding they hold.
 
 use fsst::Compressor;
-use fsst::Symbol;
+use num_traits::AsPrimitive;
+use vortex_array::ArrayRef;
 use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
-use vortex_array::accessor::ArrayAccessor;
+use vortex_array::arrays::PrimitiveArray;
+use vortex_array::arrays::VarBin;
+use vortex_array::arrays::VarBinArray;
+use vortex_array::arrays::VarBinView;
+use vortex_array::arrays::VarBinViewArray;
+use vortex_array::arrays::varbin::VarBinArrayExt;
 use vortex_array::arrays::varbin::builder::VarBinBuilder;
+use vortex_array::arrays::varbinview::BinaryView;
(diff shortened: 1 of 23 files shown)


joseph-isaacs:
What happens if the compress actually increase the array code size this estimate might still be incorrect

encodings/fsst/src/compress.rs
```diff
@@ -1,160 +1,342 @@
...
-{
-    let mut lines = Vec::with_capacity(8_192);
+/// Train an FSST [`Compressor`] from a [`VarBinViewArray`]'s non-null rows.
+pub fn fsst_train_compressor(
+    strings: &VarBinViewArray,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<Compressor> {
+    let mask = strings.validity()?.execute_mask(strings.len(), ctx)?;
+    let views = strings.views();
+    let mut lines: Vec<&[u8]> = Vec::with_capacity(views.len());
```
connortsui20: can you do `mask.true_count()` here

mprammer: Done.

robert3005:
the upper bound of fsst is len*2+7, we can choose the type that fits this value and that WILL always work, so anything in 2**31-4 length will be u32 and anything above will be u64

robert3005:
I think you still want a compress method that can take ArrayRef so callers don't have to know what they're compressing and we don't have to expose internal of how fsst does compression, then internally you can switch on the encoding type.

connortsui20:
was there a merge that went wrong? There seems to be a bunch of changes in here that are not relevant to this PR

mprammer:
re-merged develop

robert3005:
Rebased onto develop as a single clean commit (the earlier merge noise is gone — the diff is now 23 fsst-related files).

Status of review feedback:
- `mask.true_count()` — done (thread resolved).
- Offset-width estimate vs actual output size (@joseph-isaacs, @robert3005): the width decision now uses the FSST worst-case bound `2 × input_bytes + 7 × non_null_rows`; if that fits in `i32::MAX` the output always does, regardless of how the data compresses. Covered by the `offset_width_boundary` unit test and the `fsst_compress_offsets_overflow_i32` regression test.
- `fsst_compress`/`fsst_train_compressor` now take `ArrayRef` and dispatch internally on `VarBinView`/`VarBin`, so callers don't depend on the input encoding.

The rebase also folds in develop-side changes: the #8324 test optimization (empty symbol table, ~half the memory) now runs through the new API, `schemes/string/fsst.rs` in vortex-btrblocks (split out by #8207) was ported to the new signatures, and the `public-api.lock` changes were dropped since lockfiles were removed in #8099.

robert3005 (approved):

PR #7902, opened 2026-05-12
Add `UnionVariants` type and embed it in `DType::Union`

connortsui20:
## Summary

Tracking issue: https://github.com/vortex-data/vortex/issues/7882

Replaces `DType::Union(Nullability)` with `DType::Union(UnionVariants, Nullability)`. 

`UnionVariants` carries a `FieldNames` list, per-variant `FieldDType`s, and an `i8` type-id vector (supporting non-consecutive tags like `[0, 5, 7]` so children can be removed without renumbering). We use `i8` because that is what arrow uses, and because there is already a max of 128 different union variants for any given array.

A lot of the code is essentially duplicated from implementations of Struct. It could have been nice if we could reuse StructFields, but we need to additionally store a type IDs field that maps the different variant types to tag indexes (i.e. 0, 1, 2, 5), so we might as well have a new type for this that is named more appropriately.

Also note that the tags do not have to be consecutive (to support potential schema evolution, and also it does make things like casts easier).

I've also enforced that variant names are distinct. The Arrow spec does not make any guarantee about this, but I've chosen the sane path that DuckDB also [chooses](https://duckdb.org/docs/current/sql/data_types/union).

I'm going to guess that some of us will want to just allow duplicate variant names, so we should just discuss below.

## API Changes

Changes `DType::Union` to carry `UnionVariants`.

## Testing

Some basic roundtrip testing for the serialization formats.

### DIFF
--- a/vortex-array/src/dtype/union.rs
+++ b/vortex-array/src/dtype/union.rs
@@ -1,0 +1,623 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use std::fmt;
+use std::hash::Hash;
+use std::sync::Arc;
+use std::sync::OnceLock;
+
+use itertools::Itertools;
+use vortex_error::VortexExpect;
+use vortex_error::VortexResult;
+use vortex_error::vortex_ensure;
+use vortex_error::vortex_ensure_eq;
+use vortex_utils::aliases::hash_map::HashMap;
+
+use crate::dtype::DType;
+use crate::dtype::FieldDType;
+use crate::dtype::FieldName;
+use crate::dtype::FieldNames;
+use crate::dtype::Nullability;
+
+/// Type information for a union array.
+///
+/// A `UnionVariants` describes the possible alternative types for each row of a union, along with a
+/// per-variant `i8` type tag. We use the term **variants** (rather than "fields") because a union
+/// is a sum type: each row chooses exactly one alternative.
+///
+/// Per Arrow's spec, the per-row type tag is an `int8`. By default, tag `i` selects the child at
+/// offset `i` (`type_ids = [0, 1, ..., N-1]`). Schemas may also use non-consecutive tags (e.g.
+/// `[0, 5, 7]`), in which case the value of `type_ids[i]` is the tag used in the data to select the
+/// child at offset `i`. Supporting non-consecutive tags from v1 lets the schema remove children
+/// without renumbering the remaining tags.
+///
+/// Variant names must be distinct. Unlike [`StructFields`](crate::dtype::StructFields), which
+/// permits duplicate field names for Arrow/Parquet round-trip fidelity, duplicates have no
+/// meaningful semantics in a sum type and are rejected at construction.
+///
+/// ```
+/// use vortex_array::dtype::{DType, Nullability, PType, UnionVariants};
+///
+/// let variants = UnionVariants::new_consecutive(
+///     ["a", "b"].into(),
+///     vec![
+///         DType::Primitive(PType::I32, Nullability::NonNullable),
+///         DType::Utf8(Nullability::NonNullable),
+///     ],
+/// )
+/// .unwrap();
+///
+/// assert_eq!(variants.len(), 2);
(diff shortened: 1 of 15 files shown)


vortex-array/src/dtype/union.rs
```diff
@@ -0,0 +1,623 @@
...
+/// per-variant `i8` type tag. We use the term **variants** (rather than "fields") because a union
+/// is a sum type: each row chooses exactly one alternative.
+///
+/// Per Arrow's spec, the per-row type tag is an `int8`. By default, tag `i` selects the child at
+/// offset `i` (`type_ids = [0, 1, ..., N-1]`). Schemas may also use non-consecutive tags (e.g.
+/// `[0, 5, 7]`), in which case the value of `type_ids[i]` is the tag used in the data to select the
+/// child at offset `i`. Supporting non-consecutive tags from v1 lets the schema remove children
+/// without renumbering the remaining tags.
+///
+/// Variant names must be distinct. Unlike [`StructFields`](crate::dtype::StructFields), which
```
joseph-isaacs, line 34: I am not sure I agree with this. 

This will stop arrow round trip.

vortex-array/src/dtype/union.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    /// Create a new [`UnionVariants`] with explicit `type_ids`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if names, dtypes, or type IDs do not all have the same length, or if there
+    /// are any duplicate names or type ids.
+    pub fn try_new(names: FieldNames, dtypes: Vec<DType>, type_ids: Vec<i8>) -> VortexResult<Self> {
+        Self::validate_shape(&names, dtypes.len(), &type_ids)?;
```
joseph-isaacs, line 211: this look too expensive for release.

vortex-array/src/dtype/union.rs
```diff
@@ -0,0 +1,623 @@
...
+        ))))
+    }
+
+    /// Create a new [`UnionVariants`] with consecutive `type_ids = [0, 1, ..., N-1]`.
+    ///
+    /// # Errors
+    ///
+    /// `names` and `dtypes` must have the same length, and `names.len()` cannot be more than
+    /// `i8::MAX as usize + 1` (128).
+    pub fn new_consecutive(names: FieldNames, dtypes: Vec<DType>) -> VortexResult<Self> {
```
joseph-isaacs, line 227: shall we just call this new?

vortex-array/src/dtype/union.rs
```diff
@@ -0,0 +1,623 @@
...
+        self.0
+            .type_ids
+            .iter()
+            .enumerate()
+            .all(|(i, &tag)| i8::try_from(i).is_ok_and(|i| i == tag))
+    }
+
+    /// Find the offset of a variant by name. Returns `None` if no variant has the name.
+    pub fn find(&self, name: impl AsRef<str>) -> Option<usize> {
+        self.0.indices().get(name.as_ref()).copied()
```
joseph-isaacs, line 300: i would think this is slower for small variants?

vortex-array/src/dtype/union.rs
connortsui20: Allowing this would prevent duckdb round trip which explicitly disallows duplicates.

So allowing duplicates just because arrow forgot to specify it in the spec does not seem worth being completely incompatible with duckdb.

vortex-array/src/dtype/union.rs
connortsui20: What do you mean release? Are you just saying that we need a new_unchecked?

Even if there were 128 relatively long names for each of the union variants, I dont think this is that expensive, and if somehow this ends up in a hot loop we can easily optimize this.

connortsui20:
I need to override the flatc backcompat check here because we split this into 2 PRs

vortex-array/src/dtype/union.rs
robert3005: Looks like arrow doesn’t care about names since the children are resolved via types buffer. Why does duckdb care? Duplicate names will never conflict since children are always mutually exclusive.

PR #7904, opened 2026-05-12
Use async_fs instead of LocalFileSystem in java bindings for local files

robert3005:
Found this crate and I think it's a bit more ergonomic than object_store for
local files


gatesn:
Why just for Java bindings if it's better?

robert3005:
I have this change in my python pr already.

robert3005:
this doesn't seem worth it but I have a different change

PR #7905, opened 2026-05-13
Add minimal go bindings

robert3005:
This adds go bindings matching our java and c bindings. As go can interoprate
directly with C we leverage vortex-ffi directly instead of dedicated crate like
vortex-jni


robert3005:
I think the C changes to expose Arrow to Vortex conversions would be useful regardless. I will make a separate pr

robert3005:
We should change the write method to take vortex arrays and use arrow to vortex conversion

robert3005:
@gatesn you would depend on this like `vortex "github.com/vortex-data/vortex/lang/go"` but we would need to make sure to make tags like `lang/go/v0.71.0` so that go package manager picks it up

Issue #7838, opened 2026-05-08
gpu-scan-cli scan --gpu-file panics with "as_slice must be called on host buffer"


manish-sethi:
### What happened?

## Summary                                                

`gpu-scan-cli scan --gpu-file` panics on every Vortex file I produce via `gpu-scan-cli convert`. 

  Immediate panic on every tokio worker:                                                                                                                                          
   
  2026-05-07T01:11:35Z INFO vortex_cuda: Registering CUDA kernels                                                                                                                 
                                                                                                                                                                                  
  thread 'tokio-rt-worker' panicked at vortex-error/src/lib.rs:361:13:                                                                                                            
  Assertion failed error: as_slice must be called on host buffer                                                                                                                  
     0: PrimitiveData::as_slice                                                                                                                                                   
     1: Primitive::scalar_at                                                                                                                                                      
     2: ArrayRef::execute_scalar                                                                                                                                                  
     3: Patches::chunk_offset_at                                                                                                                                                  
     4: Patches::search_index                                                                                                                                                     
     5: Patches::slice                                      
     6: BitPacked::slice                                                                                                                                                          
     7: BitPacked::reduce_parent                            
     8: ArrayRef::slice                                                                                                                                                           
     9: FoR::slice
    10: FoR::reduce_parent                                                                                                                                                        
    11: ArrayRef::slice                                                                                                                                                           
    12: CudaFlatReader::projection_evaluation
    13: ChunkedReader::projection_evaluation                                                                                                                                      
    14: StructReader::projection_evaluation                                                                                                                                       
    15: scan::tasks::split_exec
                                                                                                                                                                                  
  (Many threads panic in parallel; backtrace is the same on each.)                   
                                                                                                                          
                                                        

### Steps to reproduce

  ```bash                                                                                                                                                                         
  # 1. Build the tools                                      
  cargo build --release -p vortex-tui -p gpu-scan-cli
                                                                                                                                                                                  
  # 2. Ingest a Parquet file (TPC-H SF100 `lineitem`)
  ./target/release/vx convert lineitem.parquet                                                                                                                                    
  # → lineitem.vortex                                                                                                                                                   
                                                                                                                                                                                  
  # 3. Re-encode for CUDA                                                                                                                                                         
  ./target/release/gpu-scan-cli convert \                                                                                                                                         
    --input  lineitem.vortex \                                                                                                                                                    
    --output lineitem.cuda.vortex                                                                                                                                                 
[...]

PR #7911, opened 2026-05-13
fix[gpu]: CUDA sliced patch views with device patches

0ax1:
### Summary

For `Slice(ALP)` the GPU plan builder was calling `Patches::slice` for device resident patch metadata. Within slice scalar access panicked when the patch buffers are device resident.

This change adds a CUDA materialization path for sliced patch views. The planner keeps the original Patches plus an optional slice range; materialization slices only chunk_offsets, leaves indices/values on device, and records indices_base when the sliced chunk_offsets still refer into unsliced patch arrays.

### Fixed Issue

https://github.com/vortex-data/vortex/issues/7838

### DIFF
--- a/vortex-cuda/kernels/src/dynamic_dispatch.cu
+++ b/vortex-cuda/kernels/src/dynamic_dispatch.cu
@@ -171,9 +171,6 @@ scalar_op(T *values, const struct ScalarOp &op, char *__restrict smem, uint64_t 
         // straddle two FL chunks, so each value needs its own lookup.
         if (op.params.alp.patches_ptr != 0) {
             const auto &patches = *reinterpret_cast<const GPUPatches *>(op.params.alp.patches_ptr);
-            // The sliced chunk_offsets array starts at original chunk
-            // (offset / FL_CHUNK). PatchesCursor indexes from 0, so
-            // subtract that base to get the index into chunk_offsets.
             const uint32_t chunk_start = patches.offset / FL_CHUNK;
 #pragma unroll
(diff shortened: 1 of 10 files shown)


robert3005:
I think where we tried to fix this for bitpacking we disabled some cuda tests that were panicing. Can you reenable them here?

0ax1:
> I think where we tried to fix this for bitpacking we disabled some cuda tests that were panicing. Can you reenable them here?

That's a diff issue. When a `Slice(BitPacked)` is planned through reduce rules, it stays a `SliceArray` instead of becoming a sliced `BitPacked`. The standalone `BitPackedExecutor` expects a `BitPackedArray`, and dynamic dispatch also does not fuse that `SliceArray` yet. We can prob special case BP like we do for ALP in walk_slice.

robert3005:
I fail to see the difference apart from one array being alp and the other one being bitpacked. However, we should figure out a way for dynamic dispatch to self generate itself from code blocks of each encoding so we don't end up fixing every encoding combination. In this case we would still need patched array to have the common part though

0ax1:
> I fail to see the difference apart from one array being alp and the other one being bitpacked. However, we should figure out a way for dynamic dispatch to self generate itself from code blocks of each encoding so we don't end up fixing every encoding combination. In this case we would still need patched array to have the common part though

We went from reduce to execute for bp. So we get a SliceArray for which we need special case handling, like for ALP in walk slice. That special casing I want to add in a separate PR.

robert3005 (approved):

PR #7912, opened 2026-05-13
fix[gpu]: handle sliced BP arrays in CUDA

0ax1:


vortex-cuda/src/dynamic_dispatch/plan_builder.rs
```diff
@@ -518,6 +524,30 @@ impl FusedPlan {
             return self.walk(reduced, pending_subtrees);
         }
 
+        // BitPacked with patches does not reduce through Slice. Slice the
+        // packed buffer here, and defer patch slicing to CUDA materialization.
+        if child.encoding_id() == BitPacked.id() {
```
joseph-isaacs, line 529: do we need to do ALP patches too?

joseph-isaacs (approved):

vortex-cuda/src/dynamic_dispatch/plan_builder.rs
0ax1: yeah we do the same thing for alp too

PR #7913, opened 2026-05-13
register_splits to get both offset and relative row_range

onursatici:
## Summary

register_splits had a row_range that was local, so nested chunked layouts failed to get the right global split registered.  Add the global offset and send it with the local row_range so layout readers have all information to register the global correct split offset.

I could have done this without changing the API but then the outer chunk layout has to create a new local btreemap and send that then merge with the global btreemap with the chunk offset applied, and that felt expensive

### DIFF
--- a/vortex-cuda/src/layout.rs
+++ b/vortex-cuda/src/layout.rs
@@ -49,6 +49,7 @@ use vortex::layout::LayoutReader;
 use vortex::layout::LayoutReaderRef;
 use vortex::layout::LayoutRef;
 use vortex::layout::LayoutStrategy;
+use vortex::layout::SplitRange;
 use vortex::layout::VTable;
 use vortex::layout::layouts::SharedArrayFuture;
 use vortex::layout::segments::SegmentId;
@@ -284,10 +285,11 @@ impl LayoutReader for CudaFlatReader {
     fn register_splits(
         &self,
         _field_mask: &[FieldMask],
-        row_range: &Range<u64>,
+        split_range: &SplitRange,
         splits: &mut BTreeSet<u64>,
     ) -> VortexResult<()> {
-        splits.insert(row_range.start + self.layout.row_count);
+        split_range.check_bounds(self.layout.row_count)?;
+        splits.insert(split_range.root_row_range().end);
         Ok(())
     }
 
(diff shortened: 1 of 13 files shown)


vortex-io/public-api.lock
```diff
@@ -866,13 +866,13 @@ pub async fn vortex_io::object_store::ObjectStoreWrite::shutdown(&mut self) -> s
...
+impl<T> vortex_io::VortexWrite for core::io::cursor::Cursor<T> where core::io::cursor::Cursor<T>: std::io::Write
 
-pub fn std::io::cursor::Cursor<T>::flush(&mut self) -> impl core::future::future::Future<Output = std::io::error::Result<()>>
+pub fn core::io::cursor::Cursor<T>::flush(&mut self) -> impl core::future::future::Future<Output = std::io::error::Result<()>>
 
-pub fn std::io::cursor::Cursor<T>::shutdown(&mut self) -> impl core::future::future::Future<Output = std::io::error::Result<()>>
+pub fn core::io::cursor::Cursor<T>::shutdown(&mut self) -> impl core::future::future::Future<Output = std::io::error::Result<()>>
 
-pub fn std::io::cursor::Cursor<T>::write_all<B: vortex_io::IoBuf>(&mut self, B) -> impl core::future::future::Future<Output = std::io::error::Result<B>>
+pub fn core::io::cursor::Cursor<T>::write_all<B: vortex_io::IoBuf>(&mut self, B) -> impl core::future::future::Future<Output = std::io::error::Result<B>>
```
joseph-isaacs: ??

joseph-isaacs (commented):
You could pass down an offset btree?

onursatici:
yea I thought about that but that would introduce unnecessary mutable state. We would keep changing the btree offset before passing down to each child

vortex-io/public-api.lock
onursatici: I think this happened because I have bumped my toolchain to 1.91, where Cursor supposedly moved to core but std still has the reexport

joseph-isaacs:
It would be a stack alloced wrapper

vortex-io/public-api.lock
robert3005: you need to rebase

onursatici:
yea true, it is &mut now anyway. Still I think I prefer this because it is more explicit and if projection eval or pruning eval methods in the reader decide to use the global offset in the future, we can use this SplitRange struct to give them this information

joseph-isaacs (approved):

PR #7914, opened 2026-05-13
Fix `IsSorted` return dtype

connortsui20:
## Summary

This is why catch all arms are dangerous!

Also: I made extension type not sortable, I think this is correct? Though it might also break timestamp is sorted, I have no idea.

## Testing

How should we test this?

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/is_sorted/mod.rs
+++ b/vortex-array/src/aggregate_fn/fns/is_sorted/mod.rs
@@ -242,22 +242,34 @@ impl AggregateFnVTable for IsSorted {
     fn return_dtype(&self, _options: &Self::Options, input_dtype: &DType) -> Option<DType> {
         match input_dtype {
             DType::Null
-            | DType::Struct(..)
             | DType::List(..)
             | DType::FixedSizeList(..)
-            | DType::Variant(..) => None,
-            _ => Some(DType::Bool(Nullability::NonNullable)),
+            | DType::Struct(..)
+            | DType::Union(_)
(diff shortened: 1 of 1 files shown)


vortex-array/src/aggregate_fn/fns/is_sorted/mod.rs
```diff
@@ -242,22 +242,34 @@ impl AggregateFnVTable for IsSorted {
     fn return_dtype(&self, _options: &Self::Options, input_dtype: &DType) -> Option<DType> {
         match input_dtype {
             DType::Null
-            | DType::Struct(..)
             | DType::List(..)
             | DType::FixedSizeList(..)
-            | DType::Variant(..) => None,
-            _ => Some(DType::Bool(Nullability::NonNullable)),
```
joseph-isaacs, line 249: This should have been written in the negative variant.

joseph-isaacs (approved):

vortex-array/src/aggregate_fn/fns/is_sorted/mod.rs
connortsui20: I dont understand?

robert3005:
Is sorted on extension dtype needs to call the vtable

vortex-array/src/aggregate_fn/fns/is_sorted/mod.rs
robert3005: You enumerate the ones that can instead of the ones that can’t

PR #7916, opened 2026-05-13
Consistent `DType` match

connortsui20:
## Summary

This is a purely cosmetic change.

Since we now have a whole 12 dtype enum variants (with the addition of `Union`), we might as well be more consistent with the ordering of match statements.

Also moves `Extension` to the end of the enum.

If you are strongly against this PR, then just close it for me.

## Testing

N/A


joseph-isaacs:
Ill leave this one to you and @robert3005

robert3005 (approved):

PR #7917, opened 2026-05-13
bump parquet to v58 for lance

myrrc:
Also replace base16ct with manual printing.


benchmarks/lance-bench/Cargo.toml
```diff
@@ -19,11 +19,11 @@ lance = { version = "4", default-features = false }
 lance-encoding = { version = "4" }
 
 anyhow = { workspace = true }
-arrow-cast = { version = "57" }
```
robert3005, line 22: That version was separate because lance is slow to upgrade. Just change it to 58

AdamGS: this has to match whatever lance is using, I don't think you can just bump it. What's the motivation here?

robert3005:
This is merged now with lance v6 upgrade

PR #7919, opened 2026-05-14
DataFusion: Push down decimals with some explicit coercion

AdamGS:
## Summary

Allow some version of pushing down work and casts on decimals to overcome the difference in semantics between DataFusion and Vortex.

The underlying issue here is that DataFusion is happy handling different decimals in one expression, while Vortex isn't. To overcome that, we make sure to inject some cast expressions when they are required.


AdamGS:
this approached was just nullified by #8032

PR #7920, opened 2026-05-14
Use async_fs file for java and python writes to avoid object_store buffering on local disks

robert3005:
Object store writes are tuned for multipart uploads, it's simpler to use std
file for local writes


robert3005:
This is recreation of #7904 with small addition of having python also go through local file for writes. This already happens in C/C++ and is not necessary in rust since we don't expose api to write to a path so the user has to resolve the path to file handle

gatesn (approved):

PR #7927, opened 2026-05-14
Add OnPair string compression encoding with predicate pushdown

joseph-isaacs:
## Summary

This PR introduces a new `vortex-onpair` encoding that integrates the [OnPair](https://github.com/gargiulofrancesco/onpair_cpp) short-string compression library into Vortex. OnPair is a dictionary-based compressor optimized for string data with two key features:

1. **Fast random access**: Individual rows can be decompressed without scanning the entire column
2. **Compressed-domain predicate evaluation**: Equality, prefix matching (`LIKE 'prefix%'`), and substring matching (`LIKE '%substr%'`) can be evaluated directly on compressed data without decompression

### Changes

- **`vortex-onpair-sys`**: Low-level FFI bindings to the OnPair C++ library
  - C ABI shim (`onpair_shim.h/cpp`) wrapping the OnPair C++ API
  - CMake build configuration that fetches and builds `onpair_cpp` via `FetchContent`
  - Boost dependency stripping (replaces `boost::unordered_flat_map` with `std::unordered_map`)
  - Safe Rust wrapper (`Column` type) around the C++ `OnPairColumn` handle

- **`vortex-onpair`**: Vortex array encoding and compute kernels
  - `OnPairArray` type implementing the `VTable` trait for Vortex integration
  - Compression entry point (`onpair_compress`) with configurable bit-width (9-16 bits, default 12)
  - Canonicalization to `VarBinViewArray` via bulk decompression
  - Compute kernel implementations:
    - `CompareKernel`: Pushdown of `Eq`/`NotEq` to compressed-domain `equals()`
    - `LikeKernel`: Pushdown of `LIKE` patterns to `starts_with()` and `contains()`
    - `CastKernel`: Nullability-only casts between Utf8/Binary
    - `FilterKernel` and `SliceReduce`: Fall back to canonicalization
  - Serialization/deserialization with metadata persistence
  - Lazy column materialization for cheap clones

### Design Notes

- The C++ `OnPairColumn` is lazily reconstructed on first use (e.g., canonicalization or predicate pushdown), keeping clone-only paths cheap
- Null entries are indexed by the column (mapped to empty payloads) with nullness preserved on the outer array's validity slot
- The default preset is "dict-12": 12-bit codes with a dictionary capped at 4,096 entries
- Predicate pushdown is wired through standard Vortex compute kernels, enabling automatic filter optimization

## Testing

- Added roundtrip tests verifying compression and decompression correctness
- Added nullable array handling tests
- Added scalar access tests
- Metadata serialization tests with golden file validation
- All tests pass with the new encoding integrated into the Vortex array system


a10y:
curious how it would do if wired into the compressor

a10y:
partsupp boost is huge

image

Issue #7707, opened 2026-04-28
Epic: Stats and AggregateFns


gatesn:
## Goal

Make Vortex statistics pluggable by modeling pruning stats as aggregate-function state exposed through physical bound expressions. The concrete success case is demonstrating a Bloom-filter zone-map stat for UTF-8 equality pruning added through plugins: a custom aggregate function, scalar function, and rewrite rule, without changing built-in pruning logic.

## Direction

Stats are aggregate-function partials/results when the aggregate has pruning semantics. Not every aggregate is a useful pruning stat; the rewrite path should depend only on aggregates that can prove bounds.

Keep expressions physical for this epic. Falsification turns concrete predicates into normal Vortex expressions over physical scalar functions and `stat(expr, AggregateFnRef)`. We are not adding a logical expression layer yet.

Use `stat(expr, AggregateFnRef)` as the bound-expression primitive. It returns the stat value for the current stats scope, or null when unavailable. Falsification produces expressions containing `stat(...)`; simplification/execution decides whether anything is proven.

Aggregate functions advertise whether a stored aggregate can satisfy a requested aggregate through `AggregateFnRef::can_satisfy(...)`. Exact descriptor matches are preferred; compatible approximate aggregates, such as bounded max satisfying max, may be used when the stored aggregate is a sound bound.

Zone maps store aggregate-function descriptors, using `Display` for `AggregateFnRef`, and use those descriptors as stats-table column names. At read time, the zone map lowers bound expressions by matching available descriptors against requested aggregates.

Expression expansion is acceptable for now. Rewrites may produce multiple physical proof expressions, and each zone map can lower unavailable aggregates to null.

All new stats-facing APIs should live under `vortex-array/src/stats/`. Scalar function implementations may live with scalar functions, but should be re-exported through `vortex_array::stats`.

## Phase 1: Bound Expressions and Pruning Aggregates

- [x] Add `StatFn` / `stat(expr, AggregateFnRef)` under `vortex-array/src/stats/`. #7930 #8024
  - Re-export through `vortex_array::stats`.
  - Use `StatFn::new_expr(...)` directly in tests and rewrites.
- [x] Add pruning aggregate functions needed by the rewrite path. #7933 #8025
  - Include `Min`, `Max`, `AllNull`, `AllNonNull`, `AllNan`, `AllNonNan`, and bounded min/max variants for binary/string stats.
  - Keep `NullCount` as a legacy bridge for existing stats, not as a pruning proof aggregate.
[...]

PR #7928, opened 2026-05-14
Remove chunked special case from stat execution

gatesn:
Part of #7707.

Stack: 1/3

Base: `develop`
Next: #7929

## Summary

- Remove the explicit chunked-array special case from `StatFn::execute`.
- Rely on the existing chunked scalar-function pushdown path to preserve per-chunk stat results.
- Keep the existing chunked stats regression test covering the behavior.

## Checks

- `cargo test -p vortex-array stat_expr_reads_cached_sum_per_chunk`

### DIFF
--- a/vortex-array/src/scalar_fn/fns/stat.rs
+++ b/vortex-array/src/scalar_fn/fns/stat.rs
@@ -13,10 +13,7 @@ use crate::ArrayRef;
 use crate::ExecutionCtx;
 use crate::IntoArray;
 use crate::aggregate_fn::AggregateFnRef;
-use crate::arrays::Chunked;
-use crate::arrays::ChunkedArray;
 use crate::arrays::ConstantArray;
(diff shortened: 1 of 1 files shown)


vortex-array/src/scalar_fn/fns/stat.rs
```diff
@@ -115,26 +112,6 @@ impl ScalarFnVTable for StatFn {
     ) -> VortexResult<ArrayRef> {
         let input = args.get(0)?;
         let dtype = stat_dtype(options.aggregate_fn(), input.dtype())?;
-
-        // Recurse into each chunk so the output keeps per-chunk granularity (one constant
```
gatesn, line 119: I realized we already have a rule that pushes scalar functions through chunked arrays

connortsui20 (approved):
yolo

PR #7934, opened 2026-05-14
Add pruning aggregate functions

gatesn:
## Summary

- add separate `Min` and `Max` aggregate function modules that delegate to the existing `min_max` implementation for now
- return nullable min/max aggregate partials and results so empty or all-null groups can be represented
- add `AllNull` and `AllNonNull` aggregate functions for null pruning proofs
- register the new aggregate functions and expose `stats::all_null(...)` / `stats::all_non_null(...)`
- bridge legacy stats into aggregate accumulation, including nullability-only dtype differences
- bridge `AllNull` / `AllNonNull` through legacy `NullCount` stats when the proof is sound

## Notes

`Min` and `Max` currently delegate through `min_max`, so each computes the unused bound. A future dedicated implementation can avoid that extra compute.

`AllNull` and `AllNonNull` are proof-oriented null aggregates. They let rewrites express null pruning without introducing new `row_count()` placeholders, while existing legacy `NullCount` stats can still feed them when exact enough.

## Tests

- `cargo +nightly fmt --all`
- `cargo test -p vortex-array min_`
- `cargo test -p vortex-array max_`
- `cargo test -p vortex-array stat_expr`
- `./scripts/public-api.sh`


joseph-isaacs:
@claude cna you rteview this pr, waht are system changes here

vortex-array/src/aggregate_fn/fns/all_nan/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn id(&self) -> AggregateFnId {
+        AggregateFnId::new("vortex.all_nan")
+    }
+
+    fn serialize(&self, _options: &Self::Options) -> VortexResult<Option<Vec<u8>>> {
+        Ok(None)
+    }
+
+    fn return_dtype(&self, _options: &Self::Options, _input_dtype: &DType) -> Option<DType> {
+        Some(DType::Bool(Nullability::NonNullable))
```
joseph-isaacs: Nullable, only all-valid for floats

gatesn:
@claude give this another pass

vortex-array/src/scalar_fn/fns/stat.rs
```diff
@@ -133,7 +140,55 @@ fn stat_array(
...
+            .statistics()
+            .with_typed_stats_set(|stats| stats.get_as::<u64>(Stat::NullCount))
+            .and_then(|null_count| match null_count {
+                Precision::Exact(count) => Some(count == 0),
+                Precision::Inexact(0) => Some(true),
+                Precision::Inexact(_) => None,
+            })
+            .map(ScalarValue::Bool)
+    } else if aggregate_fn.is::<AllNan>() {
+        let len = u64::try_from(len)?;
```
robert3005, line 164: why do this if you only check the ptype afterwards?

vortex-array/src/scalar_fn/fns/stat.rs
```diff
@@ -133,7 +140,55 @@ fn stat_array(
...
+            .with_typed_stats_set(|stats| stats.get_as::<u64>(Stat::NullCount))
+            .and_then(|null_count| match null_count {
+                Precision::Exact(count) => Some(count == len),
+                Precision::Inexact(count) => (count < len).then_some(false),
+            })
+            .map(ScalarValue::Bool)
+    } else if aggregate_fn.is::<AllNonNull>() {
+        array
+            .statistics()
+            .with_typed_stats_set(|stats| stats.get_as::<u64>(Stat::NullCount))
```
robert3005, line 156: there's a `array.statistics().get_as` that's shorter and does the same thing?

vortex-array/src/stats/expr.rs
```diff
@@ -172,4 +199,280 @@ mod tests {
...
+        array
+            .statistics()
+            .set(Stat::NullCount, Precision::inexact(ScalarValue::from(2u64)));
+
+        let result = array
+            .apply(&stat(
+                root(),
+                AggregateFn::new(AllNonNull, EmptyOptions).erased(),
+            ))?
+            .execute::<Canonical>(&mut LEGACY_SESSION.create_execution_ctx())?
```
robert3005, line 314: no new usages of legacy session allowed, add a static to the test

vortex-array/src/stats/expr.rs
```diff
@@ -172,4 +199,280 @@ mod tests {
...
+
+    #[test]
+    fn stat_expr_reads_cached_all_null_false_from_inexact_low_null_count() -> VortexResult<()> {
+        let array = PrimitiveArray::from_option_iter::<i32, _>([None, Some(2), None]).into_array();
+        array
+            .statistics()
+            .set(Stat::NullCount, Precision::inexact(ScalarValue::from(2u64)));
+
+        let result = array
+            .apply(&super::all_null(root()))?
```
robert3005, line 230: those super look ugly but I guess they're not explicitly forbidden by lint rules

vortex-array/src/stats/expr.rs
```diff
@@ -172,4 +199,280 @@ mod tests {
...
+        let array = PrimitiveArray::from_option_iter::<i32, _>([None, Some(2), None]).into_array();
+        array
+            .statistics()
+            .set(Stat::NullCount, Precision::inexact(ScalarValue::from(3u64)));
+
+        let result = array
+            .apply(&stat(
+                root(),
+                AggregateFn::new(AllNull, EmptyOptions).erased(),
+            ))?
```
robert3005, line 252: this is inconsistent, sometimes you use the shorthand and sometimes you don't?

vortex-array/src/expr/expression.rs
robert3005: I think you now have semantic conflict with develop

vortex-array/src/aggregate_fn/fns/bounded_max/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub max_bytes: NonZeroUsize,
+}
+
+impl Display for BoundedMaxOptions {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.max_bytes.get())
+    }
+}
+
+/// Compute a byte-bounded upper bound for the maximum non-null value of a UTF8/Binary array.
```
robert3005, line 46: I think I am missing something, this is Max in the context of pruning. Why do we need another expression? Is this because we don't have aggregate -> stats expression conversion logic we talked about?

vortex-array/src/aggregate_fn/fns/bounded_min/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+/// Compute a byte-bounded lower bound for the minimum non-null value of a UTF8/Binary array.
+#[derive(Clone, Debug)]
+pub struct BoundedMin;
+
+/// Partial accumulator state for the bounded minimum aggregate.
+pub struct BoundedMinPartial {
+    min: Option<Scalar>,
```
robert3005, line 52: I think you should change this to be an enum of `enum Value { min(Scalar), null(dtype) }`

vortex-array/src/aggregate_fn/fns/all_non_null/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[test]
+    fn all_non_null_false_with_nulls() -> VortexResult<()> {
+        let mut ctx = LEGACY_SESSION.create_execution_ctx();
+        let dtype = DType::Primitive(PType::I32, Nullability::Nullable);
+        let mut acc = Accumulator::try_new(AllNonNull, EmptyOptions, dtype)?;
+
+        let batch = PrimitiveArray::from_option_iter([Some(1i32), None, Some(3)]).into_array();
+        acc.accumulate(&batch, &mut ctx)?;
+
+        assert!(!bool::try_from(&acc.finish()?)?);
```
robert3005, line 138: are `as bool` casts forbidden? This looks weird

vortex-array/src/aggregate_fn/fns/all_non_nan/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+        acc.accumulate(&batch, &mut ctx)?;
+
+        assert!(!bool::try_from(&acc.finish()?)?);
+        Ok(())
+    }
+
+    #[test]
+    fn all_non_nan_unsupported_for_non_float() -> VortexResult<()> {
+        let dtype = DType::Primitive(PType::I32, Nullability::Nullable);
+        assert!(Accumulator::try_new(AllNonNan, EmptyOptions, dtype).is_err());
```
robert3005, line 155: note to self: we need some shorthand for this but that's the current way to do this

robert3005 (changes requested):
I think it's mostly nits but I also feel like this could use self review

vortex-array/src/aggregate_fn/fns/bounded_max/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+#[derive(Clone, Debug)]
+pub struct BoundedMax;
```
joseph-isaacs, line 48: I would call this truncated max, but this does feel like a stat/pruning concept not an aggregation one?

gatesn: But the whole epic is saying Stat == AggregateFn...

PR #7937, opened 2026-05-14
Teach zoned pruning to lower StatFn

gatesn:
## Summary

- lower StatFn-based pruning predicates over zoned stats instead of the legacy StatsCatalog path
- map null/NaN pruning aggregates to existing NullCount/NaNCount zone stats plus row_count
- remove the old zoned pruning fallback to checked_pruning_expr

## Tests

- cargo test -p vortex-array stats::rewrite::builtins
- cargo test -p vortex-array all_nan
- cargo test -p vortex-array all_non_nan
- cargo test -p vortex-layout zone_map
- cargo test -p vortex-layout -- --test-threads=1
- cargo clippy -p vortex-array --all-targets --all-features
- cargo clippy -p vortex-layout --all-targets --all-features
- ./scripts/public-api.sh
- git diff --check


vortex-array/src/aggregate_fn/fns/all_nan/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::IntoArray;
+use crate::aggregate_fn::AggregateFnId;
+use crate::aggregate_fn::AggregateFnVTable;
+use crate::aggregate_fn::EmptyOptions;
+use crate::aggregate_fn::fns::nan_count::nan_count;
+use crate::dtype::DType;
+use crate::dtype::Nullability;
+use crate::scalar::Scalar;
+
+/// Compute whether every value in an array is NaN.
```
gatesn, line 18: We should explain why we need these aggregate functions in addition to NullCount. Specifically, only certain aggregate functions are able to be used for pruning. They must represent row-wise bounds on the value of a row. Null-count contains cross-row information, therefore we cannot slice / concat the stats columns and for them to remain valid pruning expressions.

vortex-array/src/aggregate_fn/fns/all_non_nan/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::IntoArray;
+use crate::aggregate_fn::AggregateFnId;
+use crate::aggregate_fn::AggregateFnVTable;
+use crate::aggregate_fn::EmptyOptions;
+use crate::aggregate_fn::fns::nan_count::nan_count;
+use crate::dtype::DType;
+use crate::dtype::Nullability;
+use crate::scalar::Scalar;
+
+/// Compute whether every value in an array is not NaN.
```
gatesn, line 18: Same here

vortex-array/src/scalar_fn/fns/stat.rs
```diff
@@ -158,6 +160,34 @@ fn stat_array(
                 Precision::Inexact(_) => None,
             })
             .map(ScalarValue::Bool)
+    } else if aggregate_fn.is::<AllNan>() {
+        let len = u64::try_from(len)?;
+        if !has_nans(array.dtype()) {
+            Some(len == 0)
```
gatesn: Are we sure we want this to be the case for non-float empty arrays? I would have thought non-float arrays (1) would simply not support this aggregate function, i.e. the return_dtype call returns None. Or (2), can never have "True" for AllNan

vortex-array/src/stats/rewrite/builtins.rs
```diff
@@ -392,17 +428,44 @@ fn stat_expr(expr: &Expression, stat: Stat) -> Option<Expression> {
...
 
+fn with_nan_predicate(
+    lhs: &Expression,
+    rhs: &Expression,
+    value_predicate: Expression,
+) -> Expression {
+    with_all_non_nan_predicate([lhs, rhs], value_predicate)
+}
+
+fn with_all_non_nan_predicate<'a>(
```
gatesn, line 506: I think we need a few comments on these helper functions

vortex-array/src/stats/rewrite/builtins.rs
```diff
@@ -379,11 +390,36 @@ fn all_non_null(expr: &Expression) -> Expression {
...
+fn has_nans(dtype: &DType) -> bool {
+    matches!(dtype, DType::Primitive(ptype, _) if ptype.is_float())
+}
+
 fn stat_expr(expr: &Expression, stat: Stat) -> Option<Expression> {
     if let Some(literal) = literal_stat(expr, stat) {
         return Some(literal);
     }
 
+    if expr.is::<Literal>() {
```
gatesn, line 485: When would this be hit, but not above?

vortex-layout/src/layouts/zoned/zone_map.rs
```diff
@@ -106,6 +129,109 @@ impl ZoneMap {
         let substituted = substitute_row_count(applied, &row_count_array)?;
         substituted.execute::<Mask>(&mut ctx)
     }
+
+    fn lower_stats(&self, predicate: Expression) -> VortexResult<Expression> {
```
gatesn, line 135: Add a comment describing why we need this

vortex-array/src/stats/rewrite/builtins.rs
robert3005: I think with the satisfies change these rewrite rules shouldn't be necessary. You can just emit min/max and bounded min/max stats will satisfy them

gatesn: Great catch

robert3005 (approved):
Would be nice to have a bit more comments on how expression are rewritten.

PR #7938, opened 2026-05-15
Use aggregate descriptors for zoned stats

gatesn:
This PR adds a new layout type: vortex.zoned

During deserialization, we read the old zone map layout `vortex.stats` into a `vortex.zoned` in memory.

The new zoned layout now stores aggregate descriptors instead of legacy `Stat` bits, enabling arbitrary aggregate functions to be used as zone statistics.

FLUPs:
* Move FooterStats over to using AggregateFns. Or even just a root-level ZonedLayout with a single zone?


vortex-file/src/footer/mod.rs
```diff
@@ -107,6 +107,7 @@ impl Footer {
             &layout_read_ctx,
             &array_read_ctx,
             session.layouts().registry(),
+            session,
             session.allows_unknown(),
```
robert3005: should we maybe just pass session here, not sure

vortex-layout/src/layouts/zoned/builder.rs
```diff
@@ -176,6 +167,102 @@ fn supports_file_stats(dtype: &DType) -> bool {
...
+        .iter()
+        .map(|aggregate_fn| {
+            let mut accumulator = aggregate_fn.accumulator(array.dtype())?;
+            accumulator.accumulate(array, ctx)?;
+            accumulator.partial_scalar()
+        })
+        .collect()
+}
+
 fn stats_builder_with_capacity(
```
robert3005: do we still need these? I thought now this would get changed to the aggregators?

robert3005 (commented):
I think we can delete some of the old write path? We no longer seem to be using old stats accumulators

robert3005 (approved):

vortex-file/src/tests.rs
```diff
@@ -1886,7 +1886,10 @@ async fn test_segment_ordering_zonemaps_after_data() -> VortexResult<()> {
 
     // Find all zoned layouts and verify data segments come before zone map segments.
     fn check_zoned_ordering(layout: &dyn Layout, segment_specs: &[SegmentSpec]) {
-        if layout.encoding_id().as_ref() == "vortex.stats" {
+        if matches!(
+            layout.encoding_id().as_ref(),
+            "vortex.zoned" | "vortex.stats"
```
AdamGS: nit - use consts here that come from our types?

vortex-layout/src/layouts/zoned/builder.rs
```diff
@@ -126,369 +101,63 @@ impl StatsAccumulator {
...
-    fn append_null(&mut self);
-
-    fn finish(&mut self) -> NamedArrays;
-}
-
-struct StatNameArrayBuilder {
-    stat: Stat,
+struct AggregateStatsArrayBuilder {
+    aggregate_fn: AggregateFnRef,
+    descriptor: String,
```
AdamGS: what's this?

vortex-layout/src/layouts/zoned/schema.rs
```diff
@@ -3,16 +3,109 @@
...
+use vortex_error::vortex_bail;
+use vortex_error::vortex_err;
+use vortex_session::VortexSession;
 
 pub const MAX_IS_TRUNCATED: &str = "max_is_truncated";
 pub const MIN_IS_TRUNCATED: &str = "min_is_truncated";
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct AggregateSpec {
+    id: String,
```
AdamGS, line 26: should this be CachedId?

vortex-layout/src/layouts/zoned/schema.rs
```diff
@@ -47,8 +140,55 @@ pub(crate) fn stats_table_dtype(column_dtype: &DType, present_stats: &[Stat]) ->
...
+    aggregate_fn.state_dtype(column_dtype).or_else(|| {
+        if let DType::Extension(ext) = column_dtype {
+            aggregate_fn.state_dtype(ext.storage_dtype())
+        } else {
+            None
+        }
+    })
+}
+
+pub(crate) fn default_bounded_stat_max_bytes() -> std::num::NonZeroUsize {
```
AdamGS, line 148: if we really need this as a `NonZeroUsize`, than `NonZeroUsize::new_unchecked` is a `const fn`, which is much nicer IMO.

vortex-layout/src/layouts/zoned/builder.rs
robert3005: I think maybe it would be better if we inlined `aggregate_descriptor` function, it's way too ominous for what it does

vortex-layout/src/layouts/zoned/builder.rs
```diff
@@ -3,66 +3,43 @@
...
-use crate::layouts::zoned::schema::MAX_IS_TRUNCATED;
-use crate::layouts::zoned::schema::MIN_IS_TRUNCATED;
+use crate::layouts::zoned::schema::aggregate_descriptor;
+use crate::layouts::zoned::schema::aggregate_state_dtype;
 
-/// Accumulates write-time statistics for each logical zone.
-pub struct StatsAccumulator {
-    builders: Vec<Box<dyn StatsArrayBuilder>>,
+/// Accumulates aggregate-function partials for each logical zone.
+pub(crate) struct AggregateStatsAccumulator {
```
AdamGS, line 27: Why the change in visibility here?

vortex-layout/src/layouts/zoned/builder.rs
gatesn: It was memoized string representation of AggregateFnRef, used to name the stats columns. I don't think it needs to be cached anymore though, good point

vortex-layout/src/layouts/zoned/schema.rs
AdamGS: I meant to comment this on a non proto type :(

gatesn: I've removed this struct. It was a weird intermediate between the prost struct, and the in-memory AggregateFnRef. But in theory, yes it could have been

vortex-layout/src/layouts/zoned/builder.rs
gatesn: In general I lean towards less public API surface. Happy to open it if someone asks

vortex-file/src/tests.rs
gatesn: Switched to using `layout.is::<Zoned>()`

PR #7943, opened 2026-05-15
skip[ci]: fmt fix in ci

joseph-isaacs:
Try to auto-fix fmt errors using a github-bot


joseph-isaacs:
@claude review this can we do this in a better way security and code org wise?

robert3005 (approved):

PR #7944, opened 2026-05-15
Use aggregate functions for stats dtype

AdamGS:
## Summary

We shouldn't implement the logic twice, IDK how this will deal with existing files but I guess that's why we have those tests?


AdamGS:
😢 need to figure out back compact here

joseph-isaacs:
Lets be careful about moving these over.

gatesn:
@claude figure out the right fix here

gatesn:
Claude you are stupid

PR #7945, opened 2026-05-15
Allow writing Variant to files and test parquet-variant IO

AdamGS:
## Summary

Integrate Variant better into the compressor so that it can be used with the default write strategy.


vortex-file/src/strategy.rs
```diff
@@ -111,6 +114,7 @@ pub static ALLOWED_ENCODINGS: LazyLock<HashSet<ArrayId>> = LazyLock::new(|| {
 
     if use_experimental_patches() {
         allowed.insert(Patched.id());
+        allowed.insert(ParquetVariant.id());
```
robert3005: this shouldn't be here

AdamGS: removed

robert3005 (dismissed):

vortex-compressor/src/compressor.rs
```diff
@@ -250,8 +252,15 @@ impl CascadingCompressor {
...
             }
-            Canonical::Variant(_) => {
-                vortex_bail!("Variant arrays can not be compressed")
+            Canonical::Variant(variant_array) => {
+                let core_storage =
+                    self.compress_physical_slots(variant_array.core_storage(), exec_ctx)?;
+                let shredded = variant_array
+                    .shredded()
+                    .map(|arr| self.compress_physical_slots(arr, exec_ctx))
+                    .transpose()?;
```
robert3005, line 269: Why do you explicitly enumare the slots here instead of allowing the array to decide? Shouldn't this be 
```
self.compress(variant_array.core_storage(), exec_ctx)
variant_array
                    .shredded()
                    .map(|arr| self.compress(arr, exec_ctx))
                    .transpose()?
```

AdamGS: IIRC because `core_storage` is potentially recursive, you end up with a stack overflow here.

AdamGS: canonicalization for variant does something similar? this structure introduces some complexity for this sort of stuff.

robert3005: core_storage must be variant but I think shredded does not.

AdamGS: pushed something, WDYT?

robert3005 (approved):

robert3005:
I think you want similar fix in the canonical code path

AdamGS:
I agree, I have some other fixes there in the works as part of a separate PR.

AdamGS:
@robert3005 I've fixed the canonicalization issue if you want to take another look.

robert3005 (approved):

PR #7947, opened 2026-05-15
Fix segfault when calling vx CLI from Python bindings

### LINKED ISSUE #7910: `vx --help` segfaults
### What happened?

Installed vx via `uv add vortex-data`. Output:
```
[1]    74544 segmentation fault  vx --help
```
Happy to provide any useful information on request.

### Steps to reproduce

1. `uv add vortex-data`
2. `vx --help`

### Environment

```
uv 0.11.13 (Homebrew 2026-05-11 aarch64-apple-darwin)`

python 3.12.7

ProductName:            macOS
ProductVersion:         15.7.1
BuildVersion:           24G231
```

### Additional context

_No response_

joseph-isaacs:
## Summary

Closes: #7910

The `launch_from` function in the native CLI module was calling `Cli::parse_from()`, which internally calls `std::process::exit()` on parse errors (including `--help` and `--version`). When `launch_from` is invoked from Python bindings via a dynamically loaded dylib, calling `std::process::exit()` from library code causes a segfault.

This PR changes the implementation to:

1. Use `Cli::try_parse_from()` instead of `parse_from()` to surface parse errors as `clap::Error` values rather than calling `exit()`
2. Convert file-not-found errors to return `Err()` instead of calling `.exit()`
3. Update `main.rs` to handle `clap::Error` specially by calling `exit()` only from the binary entry point, not from library code
4. Add a regression test verifying that `--help` surfaces as a `clap::Error` rather than exiting

The changes preserve the exact behavior of the standalone binary (same exit codes and formatting) while allowing the library function to be safely called from Python without segfaulting.

## Testing

Added a regression test `help_flag_surfaces_clap_error()` that verifies `vx --help` returns a `clap::Error` with `DisplayHelp` kind rather than calling `std::process::exit()`. The test uses the async test harness and confirms the error can be downcast to `clap::Error`.

Existing behavior is preserved: the standalone binary still exits with the same codes and formatting as before, since `main.rs` now explicitly calls `clap_err.exit()` for clap errors.

https://claude.ai/code/session_0177K9Q4KzUyxKUPu5xokEDZ

### DIFF
--- a/vortex-tui/src/lib.rs
+++ b/vortex-tui/src/lib.rs
@@ -123,11 +123,15 @@ mod native_cli {
     ) -> anyhow::Result<()> {
         let _ = env_logger::try_init();
 
-        let cli = Cli::parse_from(args);
+        // Use `try_parse_from` rather than `parse_from` so we never call
+        // `std::process::exit` from library code — that segfaults when
+        // `launch_from` runs inside a Python-loaded dylib.
+        // See https://github.com/vortex-data/vortex/issues/7910.
+        let cli = Cli::try_parse_from(args)?;
 
         let path = cli.command.file_path();
         if !std::fs::exists(path)? {
-            Cli::command()
+            return Err(Cli::command()
                 .error(
                     clap::error::ErrorKind::Io,
                     format!(
@@ -135,7 +139,7 @@ mod native_cli {
                         path.to_str().vortex_expect("file path")
                     ),
                 )
-                .exit()
+                .into());
         }
 
         match cli.command {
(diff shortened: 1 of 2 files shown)


vortex-tui/src/lib.rs
```diff
@@ -123,19 +123,23 @@ mod native_cli {
     ) -> anyhow::Result<()> {
         let _ = env_logger::try_init();
 
-        let cli = Cli::parse_from(args);
+        // Use `try_parse_from` rather than `parse_from` so we never call
+        // `std::process::exit` from library code — that segfaults when
+        // `launch_from` runs inside a Python-loaded dylib.
+        // See https://github.com/vortex-data/vortex/issues/7910.
+        let cli = Cli::try_parse_from(args)?;
```
robert3005, line 130: I guess I understand how it works out for a cli

robert3005 (approved):

PR #7951, opened 2026-05-15
duckdb: flatten runend arrays on export if requested

myrrc:
There was a bug in runend exporter that didn't flatten the array.
This produced dictionary arrays instead of flat arrays which required duckdb to
re-flatten them which was a regression in statpopgen.

### DIFF
--- a/vortex-duckdb/src/exporter/mod.rs
+++ b/vortex-duckdb/src/exporter/mod.rs
@@ -4,6 +4,7 @@
 mod all_invalid;
 mod bool;
 mod cache;
+mod canonical;
 mod constant;
 mod decimal;
 mod dict;
@@ -22,13 +23,11 @@ mod vector;
 pub use cache::ConversionCache;
 pub use decimal::precision_to_duckdb_storage_size;
 use vortex::array::ArrayRef;
-use vortex::array::Canonical;
 use vortex::array::ExecutionCtx;
 use vortex::array::arrays::Constant;
 use vortex::array::arrays::Dict;
 use vortex::array::arrays::List;
 use vortex::array::arrays::StructArray;
-use vortex::array::arrays::TemporalArray;
 use vortex::array::arrays::struct_::StructArrayExt;
 use vortex::buffer::BitChunks;
 use vortex::encodings::runend::RunEnd;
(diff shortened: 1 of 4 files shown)


vortex-duckdb/src/exporter/mod.rs
```diff
@@ -191,7 +192,8 @@ fn new_array_exporter_with_flatten(
     };
 
     let array = match array.try_downcast::<RunEnd>() {
-        Ok(array) => return run_end::new_exporter(array, cache, ctx),
+        Ok(array) if !flatten => return run_end::new_exporter(array, cache, ctx),
+        Ok(array) => array.into_array(), // .into_array() does flattening
```
joseph-isaacs: that is not true?

joseph-isaacs: add a `new_exporter_with_flatten` here?

myrrc: The comment is misleading, the idea was that we fall back to canonicalization which does flattening.

myrrc: +1

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -165,6 +165,32 @@ pub trait ColumnExporter: 'static {
     ) -> VortexResult<()>;
 }
 
+fn canonicalize_array(
```
joseph-isaacs: I would name this `canonical::new_exporter`

vortex-duckdb/src/exporter/run_end.rs
```diff
@@ -32,11 +34,18 @@ struct RunEndExporter<E: IntegerPType> {
...
 
-pub(crate) fn new_exporter(
+pub(crate) fn new_exporter_with_flatten(
     array: RunEndArray,
     cache: &ConversionCache,
     ctx: &mut ExecutionCtx,
+    flatten: bool,
 ) -> VortexResult<Box<dyn ColumnExporter>> {
+    // Our canonicalization is 3x faster than creating a dictionary vector and
+    // letting duckdb flatten it for us.
```
joseph-isaacs: This is not really helpful. It worth saving that our canonicalize is faster than ddb, but the 3x is misleading

joseph-isaacs (approved):

vortex-duckdb/src/exporter/run_end.rs
myrrc: +1

vortex-duckdb/src/exporter/mod.rs
myrrc: +1

PR #7952, opened 2026-05-15
array tree layout

onursatici:
## Summary

Introduce array tree layout that gathers the array tree flatbuffer definitions as a vortex array in a segment that resides just before the zone map segments. This supersedes the env var to embed the array trees into the footer.
env var is still supported but deprecated.

  ## API Changes

  - **Breaking:** `vortex_layout::VTable::new_reader` gains a `ctx: &LayoutReaderContext`
    parameter. Out-of-tree implementors must add this one argument; intermediate layouts
    forward `ctx` to recursive `Layout::new_reader_in_ctx` calls on children, leaves can ignore
    it. Migration is adding that new ctx parameter per impl.
  - **Deprecated:** `FLAT_LAYOUT_INLINE_ARRAY_NODE=1` env var continues to populate
    `FlatLayout::array_tree` inline metadata as before, but now emits a one-shot
    `tracing::warn` on first read directing users at `with_array_tree(true)`. Will be removed
    once downstream tooling has migrated.


onursatici:
closing in favour of https://github.com/vortex-data/vortex/pull/8066

PR #7953, opened 2026-05-15
Add UncompressedSizeInBytes pushdown for many encodings

AdamGS:
## Summary

This PR introduces pushdown of `UncompressedSizeInBytes` for a bunch of encodings, much of it is using a shared kernel for fixed-sized types that is applicable in a ton of places.


AdamGS:
@claude review this PR

onursatici (commented):
I think it looks good overall. One thing I am curious about is if we should remove the fixed width kernel registrations from encodings and lets them fall to `try_accumulate`, which does call the fixed width kernel anyway? we can instead check if fixed width in explicit variable length encoding kernels and return None?

AdamGS:
I don't have a good heuristic of when its better to have things in shared locations vs pushdown kernels, either makes sense here I guess so I'll do that.

encodings/sparse/src/compute/uncompressed_size.rs
```diff
@@ -0,0 +1,62 @@
...
+            sparse
+                .as_ref()
+                .validity()?
+                .execute_mask(sparse.len(), ctx)?,
+        )?;
+
+        let total = base
+            .checked_add(patches)
+            .and_then(|v| v.checked_add(validity))
+            .ok_or_else(|| vortex_err!("total size overflow"))?;
```
onursatici, line 58: Sorry I think I missed this initially. We might be undercounting here for sparse, because the decompressed varbinview would have the byte views per all elements and the buffer for the long strings. 
So for short strings we should add 16 bytes - length, and for long strings  we should add 16 bytes

AdamGS: that's a good point, not sure I have a good way of handling this here, I'll think about it some more.

AdamGS: this is also true for Dict right?

robert3005: if you had average value size (or better some distribution of values) you could use this as a value for size of patches

PR #7979, opened 2026-05-18
BufferMut::zeroed_aligned stores actually allocated length instead of passed length

robert3005:
If `length * size_of::<T>` overflows the length of the buffer is wrong

### DIFF
--- a/vortex-buffer/src/buffer_mut.rs
+++ b/vortex-buffer/src/buffer_mut.rs
@@ -68,7 +68,11 @@ impl<T> BufferMut<T> {
         let mut bytes = BytesMut::zeroed((len * size_of::<T>()) + *alignment);
         bytes.advance(bytes.as_ptr().align_offset(*alignment));
         unsafe { bytes.set_len(len * size_of::<T>()) };
-        let actual_len = if size_of::<T>() == 0 { 0 } else { bytes.len() / size_of::<T>() };
+        let actual_len = if size_of::<T>() == 0 {
+            0
+        } else {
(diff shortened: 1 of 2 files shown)


vortex-buffer/src/buffer_mut.rs
```diff
@@ -68,9 +68,14 @@ impl<T> BufferMut<T> {
         let mut bytes = BytesMut::zeroed((len * size_of::<T>()) + *alignment);
         bytes.advance(bytes.as_ptr().align_offset(*alignment));
         unsafe { bytes.set_len(len * size_of::<T>()) };
+        let actual_len = if size_of::<T>() == 0 {
```
joseph-isaacs: why would I ever have a buffer of zst?

vortex-buffer/src/buffer_mut.rs
```diff
@@ -68,9 +68,14 @@ impl<T> BufferMut<T> {
...
+            0
+        } else {
+            bytes.len() / size_of::<T>()
+        };
```
joseph-isaacs: could do
`bytes.len().checked_div(size_of::<T>()).unwrap_or(0);`

vortex-buffer/src/buffer_mut.rs
robert3005: I wanted to avoid this panicing for silly, there's no reason to allocate a buffer for ZST

joseph-isaacs (approved):
I still think our buffer should assert non zst at compile time

PR #8010, opened 2026-05-18
CI: build benchmark binaries once for all targets

myrrc:


myrrc:
Not worth it, download time dominates the workload

PR #8011, opened 2026-05-18
Fast-path comparison and constant encoding for bit-packed arrays

joseph-isaacs:
## Summary

This PR adds two complementary optimizations for bit-packed arrays:

1. **Out-of-range constant comparison** (`compare.rs`): When comparing a `BitPackedArray` against a constant that falls outside the packable range `[0, 2^bit_width - 1]`, every packed lane has the same ordering relation to the constant. This collapses the comparison to a constant boolean result (modulo patches and validity), reducing the result to a `ConstantArray<bool>` in the hot path or a `BitBuffer` with per-position overlays at patched indices. The range check is `O(1)` on the constant alone and strictly cheaper than encoding it into the bit-packed representation.

2. **Constant-only bit-packing** (`bitpack_constant` and `bitpack_encode_constant`): When encoding a uniform-constant input, the standard FastLanes packing kernel is unnecessary. Instead, we analytically compute the bit pattern that each lane produces (a periodic stream of the constant's low bits), replicate it across chunks via `memset`/`memcpy`, and fall back to a single standard pack call only for the trailing partial chunk if needed. This avoids calling `BitPacking::pack` for the common case of constant-filled arrays.

Both optimizations are layout-aware and integrate cleanly with the existing encode/compare pipelines. In-range constants in comparisons fall through to the canonical decompress-then-compare path; the plan for accelerating those cases is documented in `docs/inrange_compare_plan.md`.

### Changes

- **`encodings/fastlanes/src/bitpacking/compute/compare.rs`** (new): Implements `CompareKernel` for `BitPacked` with fast-path handling for out-of-range constants. Includes comprehensive unit tests covering patches, nullability, and edge cases.
- **`encodings/fastlanes/src/bitpacking/array/bitpack_compress.rs`**: Adds `bitpack_constant()` (low-level buffer synthesis), `constant_lane_words()` (analytical bit-pattern computation), and `bitpack_encode_constant()` (public API). Includes roundtrip tests verifying correctness against the standard encode path.
- **`encodings/fastlanes/src/bitpacking/array/mod.rs`**: Adds `value_fits_bit_width()` helper for `O(1)` range checking.
- **`encodings/fastlanes/docs/inrange_compare_plan.md`** (new): Detailed plan for accelerating in-range constant comparisons via SWAR and bit-sliced techniques, with benchmarking guidance.
- **`encodings/fastlanes/benches/bitpack_constant.rs`** (new): Benchmarks constant encoding against the standard pipeline.
- **`encodings/fastlanes/benches/bitpack_compare.rs`** (new): Benchmarks out-of-range comparisons against the decompress-then-compare baseline.
- **`encodings/fastlanes/src/bitpacking/vtable/kernels.rs`**: Registers the new `CompareKernel`.
- **`encodings/fastlanes/src/bitpacking/compute/mod.rs`**: Declares the new `compare` module.
- **`encodings/fastlanes/Cargo.toml`**: Adds benchmark targets.
- **`encodings/fastlanes/public-api.lock`**: Updated for new public APIs.

## Testing

- **Unit tests in `compare.rs`**: Cover out-of-range comparisons (above and below range), patches, nullability, and fallthrough for in-range constants across all six comparison operators.
- **Unit tests in `bitpack_compress.rs`**: Verify that `bitpack_constant()` produces identical packed buffers to the standard encode path and that roundtrip unpacking recovers the original constant values.
- **Benchmarks**: Two new benchmark suites measure the fast paths against baselines to validate performance gains.

All existing tests pass; the new kernels return `Ok(None)` for cases they don't accelerate, so the canonical paths remain the correctness fallback.

https://claude.ai/code/session_0156Z1mXHNghcuT1yX3pscE8

### DIFF
--- a/encodings/fastlanes/Cargo.toml
+++ b/encodings/fastlanes/Cargo.toml
@@ -59,3 +59,11 @@ required-features = ["_test-harness"]
 name = "bit_transpose"
 harness = false
 required-features = ["_test-harness"]
+
+[[bench]]
+name = "bitpack_constant"
+harness = false
+
+[[bench]]
+name = "bitpack_compare"
+harness = false
--- a/encodings/fastlanes/benches/bitpack_compare.rs
+++ b/encodings/fastlanes/benches/bitpack_compare.rs
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Compare an already-packed `BitPackedArray` against a constant value. Compares the
+//! out-of-range fast path (constant outside `[0, 2^bit_width - 1]`) against an explicit
+//! "decompress, then compare" baseline.
+//!
+//! Sized to finish quickly. Run with `cargo bench -p vortex-fastlanes --bench bitpack_compare`.
+
+#![expect(clippy::unwrap_used)]
+#![expect(clippy::cast_possible_truncation)]
+
+use divan::Bencher;
+use divan::counter::ItemsCount;
+use vortex_array::ArrayRef;
+use vortex_array::ExecutionCtx;
+use vortex_array::IntoArray;
+use vortex_array::LEGACY_SESSION;
+use vortex_array::VortexSessionExecute;
+use vortex_array::arrays::BoolArray;
+use vortex_array::arrays::ConstantArray;
+use vortex_array::arrays::PrimitiveArray;
+use vortex_array::builtins::ArrayBuiltins;
+use vortex_array::scalar_fn::fns::operators::Operator;
+use vortex_array::validity::Validity;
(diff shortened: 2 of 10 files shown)


joseph-isaacs:
Superseded by a stacked split:

- #8012 — adds the `bitpack_compare` divan bench as a baseline (decompress + Arrow compare).
- #8013 — adds the out-of-range compare fast path on `BitPacked`, the `bitpack_constant` analytical encoder + its bench, and the in-range ordering plan doc. Stacked on #8012 so the bench numbers in that PR are the baseline the speedup is measured against.

Closing this one to consolidate review on the split. The branch `claude/optimize-bitpack-comparison-KGPS3` will be left untouched on the remote in case any of the inline comments need to be re-anchored.

PR #8015, opened 2026-05-19
fastlanes: streaming compare + between kernels for BitPacked

joseph-isaacs:
Adds `CompareKernel` and `BetweenKernel` for `BitPacked` that walk the
encoded array one 1024-element FastLanes block at a time through a single
reused scratch buffer, splice any `Patches` into the block in place via a
sorted-index cursor, then fold a `Fn(T) -> bool` predicate over the block
and write the bits directly into the output bit buffer. The materialised
primitive never appears.

The inner predicate-fold matches the canonical `BitBuffer::collect_bool`
shape — pack 64 bools into a `u64` in a tight loop — which rustc
auto-vectorises into the same `pcmpeq` + `psllq` (vector shift to bit
position) + `por` (OR into accumulator) pattern that `arrow-ord::apply_op`
lowers to. Verified via `objdump` on the bench binary (344 monomorphised
`stream_predicate` variants emit those SIMD instructions in the inner loop).

Smallest possible diff: only adds the two kernels and a private helper
shared between them, no benches, no public-API expansion beyond the two
trait impls.

    encodings/fastlanes/public-api.lock                |   8 +
    encodings/fastlanes/src/bitpacking/compute/between.rs            | 248 +
    encodings/fastlanes/src/bitpacking/compute/compare.rs            | 187 +
    encodings/fastlanes/src/bitpacking/compute/mod.rs                |   3 +
    encodings/fastlanes/src/bitpacking/compute/stream_predicate.rs   | 211 +
    encodings/fastlanes/src/bitpacking/vtable/kernels.rs             |   4 +
    6 files changed, 661 insertions(+)

Checks:
- cargo nextest run -p vortex-fastlanes  (278 passed)
- cargo clippy -p vortex-fastlanes --all-targets --all-features
- cargo +nightly fmt -p vortex-fastlanes --check
- ./scripts/public-api.sh  (only adds the two new trait impls)


encodings/fastlanes/src/bitpacking/compute/stream_predicate.rs
```diff
@@ -0,0 +1,145 @@
...
+    let bits = BitBufferMut::from_buffer(words.into_byte_buffer(), 0, len);
+    let validity = array.validity()?.union_nullability(nullability);
+    Ok(BoolArray::new(bits.freeze(), validity).into_array())
+}
+
+/// Overwrite the unpacked block in place with any patches falling in
+/// `[chunk_start, chunk_start + block.len())`, then advance `cursor` past them. Sorted
+/// indices mean the cursor only moves forward across the whole walk.
+#[inline]
+fn splice_patches<T, I>(
```
AdamGS, line 81: ~I think cursor is always 0 here and then iterates up to the `usize::min(indices.len(), chunk_start + block.len())`~

AdamGS: nvm, missed the loop there.

AdamGS, line 81: this is a bit of a weird structure, we can try and improve it in a followup

encodings/fastlanes/src/bitpacking/compute/stream_predicate.rs
```diff
@@ -0,0 +1,145 @@
...
+    P: Fn(T) -> bool,
+{
+    debug_assert_eq!(range.len(), block.len());
+    let start_bit = range.start;
+    let active_len = range.len();
+    if active_len == 0 {
+        return;
+    }
+
+    if start_bit.is_multiple_of(64) {
```
AdamGS: why 64?

joseph-isaacs: A word, to allow aligned access

AdamGS: can you make it a const?

AdamGS (approved):

PR #8016, opened 2026-05-19
chore[gpu]: error out in case of mixed cpu/gpu execution

0ax1:
Mixed GPU/CPU execution is not implemented yet. We therefore throw an explicit error in case we try decompressing a device buffer on the host.

### DIFF
--- a/vortex-cuda/src/executor.rs
+++ b/vortex-cuda/src/executor.rs
@@ -26,6 +26,7 @@ use vortex::array::arrays::struct_::StructDataParts;
 use vortex::array::buffer::BufferHandle;
 use vortex::dtype::PType;
 use vortex::error::VortexResult;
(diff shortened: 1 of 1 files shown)


joseph-isaacs (approved):

robert3005 (approved):

vortex-cuda/src/executor.rs
```diff
@@ -417,19 +417,27 @@ impl CudaArrayExt for ArrayRef {
...
                     encoding = %self.encoding_id(),
                     error = %e,
                     "No GPU execution path available, falling back to CPU"
                 );
+                e
             }
+        };
+
+        if !self.is_host() {
+            return Err(vortex_err!(
```
robert3005: you could do vortex_bail! here

PR #8021, opened 2026-05-19
Add Samply and benchmark skills

gatesn:
Adds samply and benchmarking iteration skills.

I have worked on these locally for a while and reached a point where, at least in Codex, I can just ask it to "make clickbench Q10 faster" and it will get there.


.agents/skills/bench-performance/scripts/summarize_flow_tracing.py
robert3005, line 1: what's v2 about this file?

robert3005: oh, we have V2_TRACE...

onursatici (approved):

PR #8025, opened 2026-05-19
Add pruning aggregate functions

gatesn:
Adds the pruning aggregate functions used by stats rewrites, bounded string/binary min-max variants, and the aggregate compatibility relation for matching stored bounds to requested bounds.


gatesn:
## Review comments carried over from auto-closed #7934

#7934 was closed/merged into the stack branch while the stack was being repaired. Copying the review notes here so they remain visible on the replacement PR.

### Review summaries

- [robert3005 requested changes](https://github.com/vortex-data/vortex/pull/7934#pullrequestreview-4306678378): "I think it's mostly nits but I also feel like this could use self review"
- [joseph-isaacs commented](https://github.com/vortex-data/vortex/pull/7934#pullrequestreview-4299196727)
- [joseph-isaacs commented](https://github.com/vortex-data/vortex/pull/7934#pullrequestreview-4309013364)
- [gatesn commented](https://github.com/vortex-data/vortex/pull/7934#pullrequestreview-4309034239)

### Inline review comments

- [joseph-isaacs on `vortex-array/src/aggregate_fn/fns/all_nan/mod.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3249119463): Nullable, only all-valid for floats
- [robert3005 on `vortex-array/src/scalar_fn/fns/stat.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255846347): why do this if you only check the ptype afterwards?
- [robert3005 on `vortex-array/src/scalar_fn/fns/stat.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255849036): there's a `array.statistics().get_as` that's shorter and does the same thing?
- [robert3005 on `vortex-array/src/stats/expr.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255849753): no new usages of legacy session allowed, add a static to the test
- [robert3005 on `vortex-array/src/stats/expr.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255850647): those super look ugly but I guess they're not explicitly forbidden by lint rules
- [robert3005 on `vortex-array/src/stats/expr.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255851791): this is inconsistent, sometimes you use the shorthand and sometimes you don't?
- [robert3005 on `vortex-array/src/expr/expression.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255854335): I think you now have semantic conflict with develop
- [robert3005 on `vortex-array/src/aggregate_fn/fns/bounded_max/mod.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255860628): I think I am missing something, this is Max in the context of pruning. Why do we need another expression? Is this because we don't have aggregate -> stats expression conversion logic we talked about?
- [robert3005 on `vortex-array/src/aggregate_fn/fns/bounded_min/mod.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255863730): I think you should change this to be an enum of `enum Value { min(Scalar), null(dtype) }`
- [robert3005 on `vortex-array/src/aggregate_fn/fns/all_non_null/mod.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255867272): are `as bool` casts forbidden? This looks weird
- [robert3005 on `vortex-array/src/aggregate_fn/fns/all_non_nan/mod.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3255870366): note to self: we need some shorthand for this but that's the current way to do this
- [joseph-isaacs on `vortex-array/src/aggregate_fn/fns/bounded_max/mod.rs`](https://github.com/vortex-data/vortex/pull/7934#discussion_r3257863302): I would call this truncated max, but this does feel like a stat/pruning concept not an aggregation one?
- [gatesn reply on that thread](https://github.com/vortex-data/vortex/pull/7934#discussion_r3257881691): But the whole epic is saying Stat == AggregateFn...

### Bot review findings worth preserving

From [Claude first pass](https://github.com/vortex-data/vortex/pull/7934#issuecomment-4460788173):

- Add/confirm tests and docs for empty float `AllNan` / `AllNonNan` vacuous semantics.
- Investigate whether the Min/Max `min_max()` double-compute is related to CodSpeed regressions.
- The `StatsRewriteRule` / session `#[allow(dead_code)]` plumbing is probably foundation work, but should be documented or exercised when active.

From [Claude second pass](https://github.com/vortex-data/vortex/pull/7934#issuecomment-4472139251):

- `BoundedMax` may conflate `Empty` and `Unknown` in cross-shard merging; keep or add tests covering `combine_partials` behavior.
- `try_accumulate` always short-circuits for the boolean pruning aggregates; document the relationship with `accumulate` to avoid drift.
- Non-float `AllNan` / `AllNonNan` guards in `stat_array` were called out as potentially dead or inconsistent depending on intended support.
- Add comments to bounded min/max noting they delegate to `min_max()` and compute the unused bound for now.

vortex-array/src/aggregate_fn/fns/bounded_min/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+        Ok(BoundedMinOptions {
+            max_bytes: NonZeroUsize::new(max_bytes).vortex_expect("checked non-zero max_bytes"),
+        })
+    }
+
+    fn return_dtype(&self, options: &Self::Options, input_dtype: &DType) -> Option<DType> {
+        supported_dtype(options, input_dtype).map(DType::as_nullable)
+    }
+
+    fn can_satisfy(
```
robert3005, line 114: I think you also can satisfy if options.max_bytes >= requested.max_bytes?

robert3005 (commented):
I think the only comment is boundedmin/max and we probably also want the other way around that boundedmin can satisfy min. Otherwise looks good

robert3005 (approved):

PR #8030, opened 2026-05-20
duckdb-bench --baseline option

myrrc:
This isn't a replacement for scripts/compare-benchmark-jsons.py but a way to
quickly compare benchmark query results with baseline.

Add a --baseline parameter for duckdb-bench to use previously exported
--gh-json-v3 file as a baseline

### DIFF
--- a/benchmarks/compress-bench/src/main.rs
+++ b/benchmarks/compress-bench/src/main.rs
@@ -192,7 +192,7 @@ async fn run_compress(
 
     match display_format {
         DisplayFormat::Table => {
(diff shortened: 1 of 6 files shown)


AdamGS:
this functionality exists in the python tool (vx-bench), I would really love to avoid adding more features to the underlying binaries.

0ax1:
`--gh-json-v3 file as a baseline` An interesting addition would be to be able to pass a URL from a PR which the Python runner compares against.

PR #8032, opened 2026-05-20
Remove type coercion

gatesn:
This belongs in a type analysis phase inside an engine, rather than inside the Vortex physical function implementations.

It is currently unused.


robert3005:
Just running benchmarks to make sure nothing is broken, we can merge after that

robert3005 (approved):

PR #8037, opened 2026-05-20
layout reader ctx

onursatici:
## Summary

Introduce `LayoutReaderContext`, a typed-data registry threaded through reader construction so an ancestor layout can publish values that descendant layouts look up by type at construction time.

Breaking: Layout::new_reader and VTable::new_reader gain a &LayoutReaderContext parameter. Out-of-tree impls will get a compile error and must add it


vortex-layout/src/reader.rs
```diff
@@ -217,7 +222,62 @@ impl LazyReaderChildren {
...
+/// [`Self::with`] returns a derived context that copies the existing map and adds (or
+/// replaces) one entry. The original is unchanged, allowing concurrent reader-tree
+/// constructions to each derive their own context without races. When the same `T` is
+/// published at multiple ancestors, the innermost (most-recently-published) wins.
+///
+/// Contexts are cheap to clone (shared via `Arc`), so they can be captured by lazy
+/// children helpers and survive any single stack frame.
+#[derive(Clone, Default)]
+pub struct LayoutReaderContext {
+    values: Arc<HashMap<TypeId, Arc<dyn Any + Send + Sync>>>,
```
joseph-isaacs: No TypeId please

joseph-isaacs (changes requested):
Can we use a Array like Id instead of a TypeId

vortex-layout/src/reader.rs
onursatici: would that import bug that hit our use of type id be relevant for this type of use as well? I can use the layout id here, I think it is a bit of a weirder api to have but should work

gatesn:
What's wrong with TypeId? We use this structure for all ephemeral context objects, i.e. VortexSession.

It works correctly, is very fast, and helps keep it clear that this stuff is never serialized.

The general `Id` is more used for things that are serialized and then later looked up in a registry on the other end.

vortex-layout/src/layouts/chunked/mod.rs
```diff
@@ -74,12 +74,14 @@ impl VTable for Chunked {
         name: Arc<str>,
         segment_source: Arc<dyn SegmentSource>,
         session: &VortexSession,
+        ctx: &crate::LayoutReaderContext,
```
joseph-isaacs: can you remove the crate::?

vortex-layout/src/reader.rs
```diff
@@ -217,7 +222,61 @@ impl LazyReaderChildren {
...
+    /// The original context is unchanged. If a value was already published under the
+    /// same `id`, the new one replaces it in the returned context — so two ancestors
+    /// using the same well-known static id give descendants "nearest ancestor wins".
+    pub fn with<T: Any + Send + Sync>(&self, id: Id, value: Arc<T>) -> Self {
+        let mut values = HashMap::clone(&self.values);
+        values.insert(id, value);
+        Self {
+            values: Arc::new(values),
+        }
+    }
```
joseph-isaacs: can this be its own file

vortex-layout/src/layout.rs
```diff
@@ -63,11 +64,19 @@ pub trait Layout: 'static + Send + Sync + Debug + private::Sealed {
...
+    ///
+    /// `ctx` is the typed-data registry threaded through reader construction (see
+    /// [`LayoutReaderContext`]). Top-level callers (file open, tests) typically pass
+    /// `&LayoutReaderContext::new()`; recursive callers inside layout implementations
+    /// must propagate the `ctx` they were handed so ancestor-published values reach
+    /// descendants.
     fn new_reader(
         &self,
         name: Arc<str>,
         segment_source: Arc<dyn SegmentSource>,
```
joseph-isaacs, line 84: describe other params

joseph-isaacs (approved):

joseph-isaacs:
A type id is not a stable id outside on a single compiler patch version.

PR #8039, opened 2026-05-20
Execution rule for Chunked<ParquetVariant>

AdamGS:
## Summary

When we have Chunked<ParquetVariant> array, we can (sometimes) keep it as a ParquetVariant array instead of just shifting the ChunkedArray into a canonical Variant array's core storage child.

### DIFF
--- a/encodings/parquet-variant/src/kernel.rs
+++ b/encodings/parquet-variant/src/kernel.rs
@@ -17,6 +17,9 @@ use vortex_array::ArrayRef;
 use vortex_array::ArrayView;
 use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
+use vortex_array::arrays::Chunked;
+use vortex_array::arrays::ChunkedArray;
+use vortex_array::arrays::chunked::ChunkedArrayExt;
 use vortex_array::arrays::dict::TakeExecute;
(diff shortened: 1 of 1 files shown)


robert3005:
@joseph-isaacs has done quite a bit of work on Chunked execution. Maybe we should do the swizzling in the builder execution mode?

robert3005 (approved):

encodings/parquet-variant/src/kernel.rs
```diff
@@ -45,8 +49,128 @@ pub(crate) static PARENT_KERNELS: ParentKernelSet<ParquetVariant> = ParentKernel
...
+                    .ok_or_else(|| vortex_err!("validated ParquetVariant value child is missing"))
+            })
+            .collect::<VortexResult<Vec<_>>>()?;
+        Some(ChunkedArray::try_new(value_chunks, value_dtype)?.into_array())
+    } else {
+        None
+    };
+
+    let typed_value = if let Some(typed_value_dtype) = layout.typed_value_dtype {
+        let typed_value_chunks = chunks
```
joseph-isaacs, line 158: does this not get pulled out into the canonical array variant on construction

joseph-isaacs (commented):
did you verify this actually gets hit? 

it looks like the right idea

PR #8042, opened 2026-05-20
Explicit Precision::Absent variant instead of Option<Precision>

AdamGS:
This PR adds a new `Precision::Absent` variant **only after serialization** because its a more idiomatic API. The serialization format is unchanged.


vortex-jni/src/scan.rs
```diff
@@ -216,7 +216,7 @@ pub extern "system" fn Java_dev_vortex_jni_NativeScan_partitionCount(
         let (rows, cardinality) = match scan.partition_count() {
```
robert3005, line 216: you should change this to not return Option

AdamGS: fixed in [6949da9](https://github.com/vortex-data/vortex/pull/8042/commits/6949da9da938e9a1e8ff814c5ac7de1f1009e4ef)

robert3005 (approved):

PR #8048, opened 2026-05-21
perf: remove implicit `ListViewArray` rebuild during `take` and `filter` operations

### LINKED ISSUE #8052: Defer `ListViewArray` compaction and improve rebuild heuristic
Right now we always compact `ListViewArray` based on a density heuristic after ops like `take` and `filter `to clean up sparse arrays. But when to compact wrt execution is dependent whole execution tree. If we have nested takes, for example, we should only compact after the last op. 

Whether to compact should be an explicit decision. Tradeoffs are different for different use cases and GC is expensive.

mhk197:
## Summary

Removes implicit rebuilds from `ListViewArray` `take` and `filter` compute kernels, adds density-introspection methods to support deciding when to rebuild explicitly at materialization boundaries, and defers rebuilding until array export to duckdb and arrow.

**Motivation.** The previous code rebuilt the elements buffer eagerly on every `take` or `filter` whose row-fraction dropped below `REBUILD_DENSITY_THRESHOLD`. In an execution tree like `take → take → ...`, an eager mid-pipeline rebuild costs an allocation and a full copy of referenced ranges that the next operator may immediately sparsify away.

The row-fraction heuristic was also inaccurate. It doesn't account for per-row size variance, unreferenced elements, and duplicate references. Instead, `ListViewArray::estimate_density` uses sum of `sizes` instead of row-fraction. This will overestimate density when there are overlapping references, but it is typically preferable to not compact.

**Changes:**

- **Drop implicit rebuild from `TakeReduce::take` and `TakeExecute::take`**
- **Drop implicit rebuild from `filter_listview`**
- **Add methods to calculate reference density**: `compute_referenced_elements_mask`, `compute_density`, and `estimate_density`
- **Estimate density and conditionally rebuild on export boundaries for duckdb and arrow**

## API Changes

Adds `ListViewArray::estimate_density`, `ListViewArray::compute_density`, and `ListViewArray::compute_referenced_elements_mask`.

## Testing

- New `vortex-array/src/arrays/listview/tests/density.rs`

### DIFF
--- a/vortex-array/src/arrays/listview/array.rs
+++ b/vortex-array/src/arrays/listview/array.rs
@@ -17,11 +17,11 @@ use vortex_mask::Mask;
 
 use crate::ArrayRef;
 use crate::ArraySlots;
+use crate::ExecutionCtx;
 use crate::LEGACY_SESSION;
 #[expect(deprecated)]
 use crate::ToCanonical as _;
 use crate::VortexSessionExecute;
-use crate::aggregate_fn::fns::sum::sum;
 use crate::array::Array;
 use crate::array::ArrayParts;
 use crate::array::TypedArrayRef;
@@ -33,9 +33,7 @@ use crate::arrays::PrimitiveArray;
 use crate::arrays::bool;
 use crate::dtype::DType;
 use crate::dtype::IntegerPType;
-use crate::expr::stats::Precision;
 use crate::expr::stats::Stat;
-use crate::expr::stats::StatsProvider;
 use crate::match_each_integer_ptype;
 use crate::validity::Validity;
 
@@ -409,52 +407,51 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
     /// Walks every `(offset, size)` pair, canonicalizes both `offsets` and `sizes`,
     /// and allocates a `BitBuffer` of length `elements.len()`, so it is extremely costly.
     ///
-    /// Returns `None` when `elements` is empty.
-    fn compute_referenced_elements_mask(&self) -> Option<Mask> {
+    /// Returns `Ok(None)` when `elements` is empty.
+    fn compute_referenced_elements_mask(
+        &self,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<Mask>> {
         let len = self.elements().len();
         if len == 0 {
-            return None;
+            return Ok(None);
         }
 
-        let offsets_dtype = self.offsets().dtype();
-        let sizes_dtype = self.sizes().dtype();
-
-        #[expect(deprecated)]
-        let offsets_primitive = self.offsets().to_primitive();
-        #[expect(deprecated)]
-        let sizes_primitive = self.sizes().to_primitive();
+        let offsets_primitive = self.offsets().clone().execute::<PrimitiveArray>(ctx)?;
+        let sizes_primitive = self.sizes().clone().execute::<PrimitiveArray>(ctx)?;
 
         let mut buf = BitBufferMut::new_unset(len);
         let offset_len = self.as_ref().len();
 
-        match_each_integer_ptype!(offsets_dtype.as_ptype(), |O| {
-            match_each_integer_ptype!(sizes_dtype.as_ptype(), |S| {
+        match_each_integer_ptype!(offsets_primitive.ptype(), |O| {
+            match_each_integer_ptype!(sizes_primitive.ptype(), |S| {
                 let offsets_slice = offsets_primitive.as_slice::<O>();
                 let sizes_slice = sizes_primitive.as_slice::<S>();
 
-                (0..offset_len).for_each(|i| {
-                    #[allow(clippy::cast_possible_truncation)]
-                    let start = offsets_slice[i] as usize;
-                    #[allow(clippy::cast_possible_truncation)]
-                    let size = sizes_slice[i] as usize;
+                for i in 0..offset_len {
+                    let start =
+                        usize::try_from(offsets_slice[i]).vortex_expect("offset must fit in usize");
+                    let size =
+                        usize::try_from(sizes_slice[i]).vortex_expect("size must fit in usize");
                     buf.fill_range(start, start + size, true);
-                });
+                }
             })
         });
 
-        Some(Mask::from_buffer(buf.freeze()))
+        Ok(Some(Mask::from_buffer(buf.freeze())))
     }
 
     /// Exact fraction of `elements` referenced by some view, in `[0.0, 1.0]`. Extremely costly.
     ///
-    /// Returns `None` when `elements` is empty.
-    fn compute_density(&self) -> Option<f32> {
-        self.compute_referenced_elements_mask()
+    /// Returns `Ok(None)` when `elements` is empty.
+    fn compute_density(&self, ctx: &mut ExecutionCtx) -> VortexResult<Option<f32>> {
+        Ok(self
+            .compute_referenced_elements_mask(ctx)?
             .map(|mask| match mask {
                 Mask::AllTrue(_) => 1.0,
                 Mask::AllFalse(_) => 0.0,
                 Mask::Values(values) => values.true_count() as f32 / self.elements().len() as f32,
-            })
+            }))
     }
 
     /// Upper-bound estimate of [`compute_density`](Self::compute_density) via
@@ -462,8 +459,8 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
     ///
     /// Exact for non-overlapping views, but overcounts when multiple views share the same elements.
     ///
-    /// Returns `Ok(None)` when `elements` is empty
-    fn estimate_density(&self) -> VortexResult<Option<f32>> {
+    /// Returns `Ok(None)` when `elements` is empty.
+    fn estimate_density(&self, ctx: &mut ExecutionCtx) -> VortexResult<Option<f32>> {
         let n_elts = self.elements().len();
         if n_elts == 0 {
             return Ok(None);
@@ -474,17 +471,14 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
             return Ok(Some(0.0));
         }
 
-        // Try to fetch the cached sum stat, otherwise fall back to calculating it on the spot
-        let sizes_sum = if let Some(Precision::Exact(scalar)) = sizes.statistics().get(Stat::Sum)
-            && let Some(sum) = scalar.as_primitive().as_::<u64>()
-        {
-            sum
-        } else {
-            sum(sizes, &mut LEGACY_SESSION.create_execution_ctx())?
-                .as_primitive()
-                .as_::<u64>()
-                .ok_or_else(|| vortex_err!("could not cast sum of sizes to u64"))?
-        };
+        // compute_stat short-circuits on a cached exact Sum and otherwise computes-and-caches.
+        let sizes_sum = sizes
+            .statistics()
+            .compute_stat(Stat::Sum, ctx)?
+            .ok_or_else(|| vortex_err!("Sum stat unavailable for sizes"))?
+            .as_primitive()
+            .as_::<u64>()
+            .ok_or_else(|| vortex_err!("could not cast sum of sizes to u64"))?;
 
         let estimate = (sizes_sum as f32 / n_elts as f32).clamp(0.0, 1.0);
 
--- a/vortex-array/src/arrays/listview/array.rs
+++ b/vortex-array/src/arrays/listview/array.rs
@@ -408,6 +408,7 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
     /// and allocates a `BitBuffer` of length `elements.len()`, so it is extremely costly.
     ///
     /// Returns `Ok(None)` when `elements` is empty.
+    #[allow(clippy::cognitive_complexity, clippy::unnecessary_fallible_conversions)]
     fn compute_referenced_elements_mask(
         &self,
         ctx: &mut ExecutionCtx,
--- a/vortex-array/src/arrays/listview/array.rs
+++ b/vortex-array/src/arrays/listview/array.rs
@@ -407,16 +407,13 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
     /// Walks every `(offset, size)` pair, canonicalizes both `offsets` and `sizes`,
     /// and allocates a `BitBuffer` of length `elements.len()`, so it is extremely costly.
     ///
-    /// Returns `Ok(None)` when `elements` is empty.
-    #[allow(clippy::cognitive_complexity, clippy::unnecessary_fallible_conversions)]
-    fn compute_referenced_elements_mask(
-        &self,
-        ctx: &mut ExecutionCtx,
-    ) -> VortexResult<Option<Mask>> {
+    /// **Preconditions**
+    ///
+    /// Assumes that `self.elements()` is non-empty.
+    #[allow(clippy::cognitive_complexity)]
+    fn compute_referenced_elements_mask(&self, ctx: &mut ExecutionCtx) -> VortexResult<Mask> {
+        debug_assert!(!self.elements().is_empty());
         let len = self.elements().len();
-        if len == 0 {
-            return Ok(None);
-        }
 
         let offsets_primitive = self.offsets().clone().execute::<PrimitiveArray>(ctx)?;
         let sizes_primitive = self.sizes().clone().execute::<PrimitiveArray>(ctx)?;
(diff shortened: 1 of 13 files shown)


vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -25,6 +25,18 @@ use crate::match_each_integer_ptype;
...
+///
+/// We don't touch `elements` on the metadata-only path since reorganizing it can be expensive.
+/// However, we also don't want to drag around a large amount of garbage data when the selection
+/// is sparse. Below this fraction of list rows retained, the rebuild is worth it.
+/// Rebuilding is needed when exporting the ListView's elements.
+///
+// TODO(connor)[ListView]: Ideally, we would only rebuild after all `take`s and `filter`
+//  compute functions have run, at the "top" of the operator tree. However, we cannot do this
+//  right now, so we will just rebuild every time (similar to [`ListArray`]).
+const REBUILD_DENSITY_THRESHOLD: f32 = 0.1;
```
connortsui20: This needs to be updated (the comment)?

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +402,94 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+            return None;
+        }
+
+        let offsets_dtype = self.offsets().dtype();
+        let sizes_dtype = self.sizes().dtype();
+
+        #[expect(deprecated)]
+        let offsets_primitive = self.offsets().to_primitive();
+        #[expect(deprecated)]
+        let sizes_primitive = self.sizes().to_primitive();
```
connortsui20: Since this is a new function can you require passing `ExecutionCtx` and use `execute` here?

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +402,94 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+        match_each_integer_ptype!(offsets_dtype.as_ptype(), |O| {
+            match_each_integer_ptype!(sizes_dtype.as_ptype(), |S| {
+                let offsets_slice = offsets_primitive.as_slice::<O>();
+                let sizes_slice = sizes_primitive.as_slice::<S>();
+
+                (0..offset_len).for_each(|i| {
+                    #[allow(clippy::cast_possible_truncation)]
+                    let start = offsets_slice[i] as usize;
+                    #[allow(clippy::cast_possible_truncation)]
+                    let size = sizes_slice[i] as usize;
```
connortsui20: This needs to handle truncation because usize might be 32 bit on different platforms and the offsets or sizes might be 64 bit (just do `usize::try_from` with a `vortex_expect`)

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +402,94 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+            return Ok(Some(0.0));
+        }
+
+        // Try to fetch the cached sum stat, otherwise fall back to calculating it on the spot
+        let sizes_sum = if let Some(Precision::Exact(scalar)) = sizes.statistics().get(Stat::Sum)
+            && let Some(sum) = scalar.as_primitive().as_::<u64>()
+        {
+            sum
+        } else {
+            sum(sizes, &mut LEGACY_SESSION.create_execution_ctx())?
```
connortsui20: Yeah can you just pass the execution ctx through the function

connortsui20 (changes requested):

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +402,94 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+            return Ok(None);
+        }
+
+        let sizes = self.sizes();
+        if sizes.is_empty() {
+            return Ok(Some(0.0));
+        }
+
+        // Try to fetch the cached sum stat, otherwise fall back to calculating it on the spot
+        let sizes_sum = if let Some(Precision::Exact(scalar)) = sizes.statistics().get(Stat::Sum)
```
connortsui20: Use `compute_stat` here since it will compute and also store it (and then you dont need the other branch)

mhk197: I don't think it stores it https://github.com/vortex-data/vortex/blob/fc03517e8d322a2f5a31118a05a3ae091346e508/vortex-array/src/stats/array.rs#L167-L176

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,91 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+        let mut buf = BitBufferMut::new_unset(len);
+        let offset_len = self.as_ref().len();
+
+        match_each_integer_ptype!(offsets_primitive.ptype(), |O| {
+            match_each_integer_ptype!(sizes_primitive.ptype(), |S| {
+                let offsets_slice = offsets_primitive.as_slice::<O>();
+                let sizes_slice = sizes_primitive.as_slice::<S>();
+
+                for i in 0..offset_len {
+                    let start =
```
joseph-isaacs: don't use a fallible convert use the numtrait infallible one. 

less branching

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,91 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+        // compute_stat short-circuits on a cached exact Sum and otherwise computes-and-caches.
+        let sizes_sum = sizes
+            .statistics()
+            .compute_stat(Stat::Sum, ctx)?
+            .ok_or_else(|| vortex_err!("Sum stat unavailable for sizes"))?
+            .as_primitive()
+            .as_::<u64>()
+            .ok_or_else(|| vortex_err!("could not cast sum of sizes to u64"))?;
+
+        let estimate = (sizes_sum as f32 / n_elts as f32).clamp(0.0, 1.0);
```
joseph-isaacs: how can this go above or below 0,1 maybe a panic here. 

or debug assert

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -25,6 +26,13 @@ use crate::match_each_integer_ptype;
 use crate::scalar::Scalar;
 use crate::scalar_fn::fns::operators::Operator;
 
+/// Density threshold to decide whether to rebuild a sparse `ListViewArray`.
+///
+/// A `ListViewArray` can accumulate unreferenced bytes in its `elements` buffer after
+/// metadata-only operations like `take` and `filter`. When density (referenced fraction of `elements`)
+/// falls below this threshold, the benefits of a rebuild may outweigh its cost.
+const REBUILD_DENSITY_THRESHOLD: f32 = 0.1;
```
joseph-isaacs: explain it's a guess and likely different for different element dtypes and different users of the list

vortex-array/src/arrays/listview/rebuild.rs
```diff
@@ -376,6 +384,16 @@ impl ListViewArray {
             self.rebuild_zero_copy_to_list()
         }
     }
+
+    pub fn should_rebuild(&self, exact: bool, ctx: &mut ExecutionCtx) -> VortexResult<bool> {
```
joseph-isaacs: comment. explain this. also this looks like a footgun. 

add the docstr with a detail message here or point to the threshold!

vortex-array/src/arrow/executor/list_view.rs
```diff
@@ -27,15 +27,16 @@ pub(super) fn to_arrow_list_view<O: OffsetSizeTrait + IntegerPType>(
...
     ctx: &mut ExecutionCtx,
 ) -> VortexResult<arrow_array::ArrayRef> {
-    // Check for Vortex ListViewArray and convert directly.
-    let array = match array.try_downcast::<ListView>() {
-        Ok(array) => return list_view_to_list_view::<O>(array, elements_field, ctx),
-        Err(array) => array,
+    let array = array.execute::<ListViewArray>(ctx)?;
+
+    // If array is sparse, rebuild before handing it to Arrow
+    let array = if array.should_rebuild(false, ctx)? {
```
joseph-isaacs: justify this

joseph-isaacs (commented):
looks. good overall

did you do this for duckdb too? that is important. since that might be passed sparse element lists. 

also a single definitive comment explaining what you did.

vortex-array/src/arrays/listview/array.rs
connortsui20: I would prefer a debug assert here

vortex-array/src/arrays/listview/rebuild.rs
connortsui20: Like Joe said, this seems like a footgun. I think that the caller should just decide, and the `REBUILD_DENSITY_THRESHOLD` can just be `DEFAULT_REBUILD_DENSITY_THRESHOLD` or similar

vortex-array/src/arrays/listview/array.rs
mhk197:
```
Elements: [0,1]
ListView: [[0,1], [1]]
----
estimate: 3/2
```

Can't go below 0 though ofc, so I can use `min(1.0)` and debug assert `> 0`

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,91 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+            })
+        });
+
+        Ok(Some(Mask::from_buffer(buf.freeze())))
+    }
+
+    /// Exact fraction of `elements` referenced by some view, in `[0.0, 1.0]`. Extremely costly.
+    ///
+    /// Returns `Ok(None)` when `elements` is empty.
+    fn compute_density(&self, ctx: &mut ExecutionCtx) -> VortexResult<Option<f32>> {
```
connortsui20: I think this should not return an option, the only case that it returns None is if the elements is empty, which means that there can only be offsets and sizes all equal to 0

mhk197: Setting to `1.0` in that case

vortex-array/src/arrays/listview/rebuild.rs
mhk197: I'm just going to remove this and leave it up to the caller

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,96 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+                let offsets_slice = offsets_primitive.as_slice::<O>();
+                let sizes_slice = sizes_primitive.as_slice::<S>();
+
+                for i in 0..offset_len {
+                    let start: usize = offsets_slice[i].as_();
+                    let size: usize = sizes_slice[i].as_();
+                    buf.fill_range(start, start + size, true);
+                }
+            })
+        });
```
connortsui20, line 456: Instead of allowing cognitive complexity, can you split this out into a generic function?

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,96 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+    /// Returns a [`Mask`] of length `elements.len()` where each bit is set iff that
+    /// position in `elements` is referenced by at least one view.
+    ///
+    /// Walks every `(offset, size)` pair, canonicalizes both `offsets` and `sizes`,
+    /// and allocates a `BitBuffer` of length `elements.len()`, so it is extremely costly.
+    ///
+    /// **Preconditions**
+    ///
+    /// Assumes that `self.elements()` is non-empty.
+    #[allow(clippy::cognitive_complexity)]
```
connortsui20: See above, but for the future we try and prefer `#[expect(clippy::cognitive_complexity, reason = "...")]`

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,98 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+
+    /// Returns a [`Mask`] of length `elements.len()` where each bit is set iff that
+    /// position in `elements` is referenced by at least one view.
+    ///
+    /// Walks every `(offset, size)` pair, canonicalizes both `offsets` and `sizes`,
+    /// and allocates a `BitBuffer` of length `elements.len()`, so it is extremely costly.
+    ///
+    /// **Preconditions**
+    ///
+    /// Assumes that `self.elements()` is non-empty.
```
connortsui20: If it is just a debug assert it is not a precondition, you should just make it an assert and say that this will panic if it is non-empty (and explain that the caller should be checking that it is not empty first)

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,98 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+                }
+            })
+        });
+
+        Ok(Mask::from_buffer(buf.freeze()))
+    }
+
+    /// Exact fraction of `elements` referenced by some view, in `[0.0, 1.0]`. Extremely costly.
+    ///
+    /// Returns `Ok(1.0)` when `elements` is empty.
```
connortsui20:
```suggestion
    /// Returns `Ok(1.0)` when `elements` is empty instead of dividing by 0.
```

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,98 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+        Ok(density)
+    }
+
+    /// Upper-bound estimate of [`compute_density`](Self::compute_density) via
+    /// `sum(sizes) / elements.len()`, clamped to `[0.0, 1.0]`.
+    ///
+    /// Exact for non-overlapping views, but overcounts when multiple views share the same elements.
+    ///
+    /// Returns `Ok(1.0)` when `elements` is empty.
+    fn estimate_density(&self, ctx: &mut ExecutionCtx) -> VortexResult<f32> {
```
connortsui20: maybe we should just call this `upper_bound_density`?

vortex-array/src/arrays/listview/array.rs
```diff
@@ -396,6 +400,98 @@ pub trait ListViewArrayExt: TypedArrayRef<ListView> {
...
+        }
+
+        // compute_stat short-circuits on a cached exact Sum and otherwise computes
+        let sizes_sum = sizes
+            .statistics()
+            .compute_stat(Stat::Sum, ctx)?
+            .ok_or_else(|| vortex_err!("Sum stat unavailable for sizes"))?
+            .as_primitive()
+            .as_::<u64>()
+            .ok_or_else(|| vortex_err!("could not cast sum of sizes to u64"))?;
```
connortsui20: ~~Does this do an implicit upcast?~~

Yes it does, so these should be vortex_expect

vortex-array/src/arrays/listview/array.rs
mhk197: noted

vortex-array/src/arrays/listview/array.rs
```diff
@@ -311,6 +315,31 @@ impl Default for ListViewData {
     }
 }
 
+/// Walks parallel `(offset, size)` slices and sets each range `[offset, offset + size]` in `buf`.
+///
+/// **Preconditions**
```
connortsui20, line 320: Nit: I think the Rust idiomatic practice is to label this `# Panics`

https://rust-lang.github.io/rust-clippy/stable/index.html?search=clippy%3A%3Amissing_panics_doc#missing_panics_doc

(also below)

vortex-array/src/arrow/executor/list_view.rs
```diff
@@ -27,15 +29,19 @@ pub(super) fn to_arrow_list_view<O: OffsetSizeTrait + IntegerPType>(
...
-        Ok(array) => return list_view_to_list_view::<O>(array, elements_field, ctx),
-        Err(array) => array,
+    let array = array.execute::<ListViewArray>(ctx)?;
+
+    // If the array is sufficiently sparse, rebuild before handing it to Arrow. Otherwise downstream
+    // consumers hold an elements buffer containing unreferenced data in memory indefinitely,
+    // and any compute pass over that buffer wastes work on data nothing references.
+    let density = array.upper_bound_density(ctx)?;
+    let array = if density < DEFAULT_REBUILD_DENSITY_THRESHOLD {
+        array.rebuild(ListViewRebuildMode::MakeZeroCopyToList)?
```
connortsui20, line 39: Might as well do `::MakeExact` here?

Now that I think about this, do all Arrow implementations allow offsets that don't start at 0? @joseph-isaacs do you know?

https://arrow.apache.org/docs/format/Columnar.html#variable-size-list-layout

vortex-array/src/stats/array.rs
```diff
@@ -153,6 +153,8 @@ impl StatsSetRef<'_> {
         f(&mut lock.iter())
     }
 
+    /// Returns the value of `stat` by either fetching it from cache if it exists and is [`Precision::Exact`], or falling back to
+    /// computation. The underlying compute kernels will cache the computed stat in the latter case.
```
connortsui20, line 157: nice

connortsui20 (approved):
LGTM

PR #8050, opened 2026-05-21
Use Uint8 for GT field in statpopgen instead of UInt64

myrrc:
With current U64 we're mostly measuring Duckdb upcast performance.
list_sum(GT) upcasts GT to u128 which takes most CPU time.
Since GT can fit in U8, use this to measure our performance.


danking (approved):

joseph-isaacs (changes requested):
lets audit all the fields

PR #8057, opened 2026-05-22
Support reporting statistics in spark datasource

robert3005:
Spark mostly focuses on sizeInBytes which we populate from file sizes with
scaling. We also report numRows since that exists in our datasource.

### DIFF
--- a/java/vortex-spark/src/main/java/dev/vortex/spark/read/VortexScan.java
+++ b/java/vortex-spark/src/main/java/dev/vortex/spark/read/VortexScan.java
@@ -3,38 +3,62 @@
 
 package dev.vortex.spark.read;
 
+import dev.vortex.api.DataSource;
+import dev.vortex.api.Session;
+import dev.vortex.spark.VortexSparkSession;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalLong;
 import org.apache.spark.sql.connector.catalog.CatalogV2Util;
 import org.apache.spark.sql.connector.catalog.Column;
+import org.apache.spark.sql.connector.expressions.NamedReference;
 import org.apache.spark.sql.connector.expressions.filter.Predicate;
 import org.apache.spark.sql.connector.read.Batch;
 import org.apache.spark.sql.connector.read.Scan;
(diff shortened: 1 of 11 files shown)


java/vortex-spark/src/main/java/dev/vortex/spark/read/VortexScan.java
```diff
@@ -83,4 +107,70 @@ public Batch toBatch() {
...
+        int tableDefaultSize = tableSchema.defaultSize();
+        if (tableDefaultSize <= 0) {
+            return fileBytes;
+        }
+
+        double scaled = SQLConf.get().fileCompressionFactor()
+                * fileBytes.getAsLong()
+                / tableDefaultSize
+                * readSchema.defaultSize();
+        return OptionalLong.of((long) scaled);
```
a10y, line 166: what is this doing?

java/vortex-spark/src/main/java/dev/vortex/spark/read/VortexScan.java
```diff
@@ -83,4 +107,70 @@ public Batch toBatch() {
...
+    }
+
+    private record VortexStatistics(OptionalLong numRows, OptionalLong sizeInBytes) implements Statistics {
+
+        @Override
+        public Map<NamedReference, ColumnStatistics> columnStats() {
+            return Map.of();
+        }
```
a10y, line 174: in the future this is where we'd report the file stats right?

robert3005: Yes but this doesn't happen right now anywhere in spark and you'd have to read all the footers here to produce this value which might be bad

java/vortex-spark/src/main/java/dev/vortex/spark/read/VortexScan.java
robert3005: This is taking the logic that spark has for the size of the relation - it's trying to scale the file size by the ratio of read schema and file schema. Basically try to take column pruning into consideration

a10y (approved):

joseph-isaacs:
This is a break

robert3005:
indeed, we have added a new argument to the datasource constructors

PR #8059, opened 2026-05-22
Add bit-packed widening cast pushdown

joseph-isaacs:
## Summary

- Add a bit-packed `CastKernel` path for widening integer casts.
- Decode FastLanes chunks directly into the wider primitive output buffer instead of materializing a full narrow intermediate.
- Keep patch handling and FoR fused-decode patch mapping compatible with the existing public patch API.

Previously the problem with this kernels include:
 - walking the elements twice (min/max, then cast)
 - Not mutating the array in-place if the bit-width match (always)

### DIFF
--- a/encodings/fastlanes/src/bitpacking/array/unpack_iter.rs
+++ b/encodings/fastlanes/src/bitpacking/array/unpack_iter.rs
@@ -181,12 +181,11 @@ impl<T: PhysicalPType, S: UnpackStrategy<T>> UnpackedChunks<T, S> {
         })
     }
 
-    /// Decode all chunks (initial, full, and trailer) into the output range.
-    /// This consolidates the logic for handling all three chunk types in one place.
+    /// Decode all chunks (initial, full, and trailer) directly into the output range.
     pub fn decode_into(&mut self, output: &mut [MaybeUninit<T>]) {
+        debug_assert_eq!(output.len(), self.len);
         let mut local_idx = 0;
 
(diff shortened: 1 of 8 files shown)


joseph-isaacs:
@claude can you review this PR for design and ensure the usage of the btrblock iterator is consistent

encodings/fastlanes/src/bitpacking/array/unpack_iter.rs
```diff
@@ -202,29 +201,78 @@ impl<T: PhysicalPType, S: UnpackStrategy<T>> UnpackedChunks<T, S> {
...
+            );
+            local_idx += chunk_len;
+        }
+
+        debug_assert_eq!(local_idx, self.len);
+    }
+
     /// Unpack full chunks into output range starting at the given index.
     /// Returns the next local index to write to.
     fn decode_full_chunks_into_at(
```
robert3005, line 264: we could replace this function and its use with the one above passing identity function to it. This feels now overly specific to decompression

robert3005 (approved):

encodings/fastlanes/src/bitpacking/array/unpack_iter.rs
robert3005, line 185: I guess this function could go as well or just be an alias to the new code with identity

robert3005:
Where do the regressions come from?

joseph-isaacs:
updated the desc.

PR #8061, opened 2026-05-22
perf: aggregate min/max

joseph-isaacs:
Adds a divan benchmark exercising the min/max aggregation over primitive
arrays (i32/i64/f64, with and without nulls) so we can measure and inspect
the codegen of the max reduction path.

### DIFF
--- a/vortex-array/benches/aggregate_max.rs
+++ b/vortex-array/benches/aggregate_max.rs
@@ -14,50 +14,12 @@ fn main() {
 
 const N: usize = 1_000_000;
 
-fn gen_i32(null_density: f64, seed: u64) -> Vec<Option<i32>> {
-    let mut rng = StdRng::seed_from_u64(seed);
-    (0..N)
-        .map(|_| {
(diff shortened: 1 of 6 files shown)


robert3005 (approved):

vortex-array/benches/aggregate_max.rs
```diff
@@ -0,0 +1,51 @@
...
+
+#[divan::bench]
+fn max_i32(bencher: Bencher) {
+    let mut rng = StdRng::seed_from_u64(1);
+    let data: Vec<i32> = (0..N).map(|_| rng.random::<i32>()).collect();
+    bencher
+        .with_inputs(|| PrimitiveArray::from_iter(data.iter().copied()).into_array())
+        .bench_refs(|a| {
+            a.statistics()
+                .compute_max::<i32>(&mut LEGACY_SESSION.create_execution_ctx())
```
robert3005: can you create a local session here?

PR #8062, opened 2026-05-22
perf[array]: cast to prim faster

joseph-isaacs:
Remove a double pass of the array, once for min/max another to cast.

Can try to reinterpret the same buffer.

### DIFF
--- a/vortex-array/src/arrays/primitive/compute/cast.rs
+++ b/vortex-array/src/arrays/primitive/compute/cast.rs
@@ -103,6 +103,62 @@ impl CastKernel for Primitive {
     }
 }
 
+/// Cast values from `F` to `T`. For infallible casts this is a pure pass; for fallible casts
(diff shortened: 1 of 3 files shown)


vortex-array/src/arrays/primitive/compute/cast.rs
```diff
@@ -131,9 +164,119 @@ fn values_fit_in(
...
+            Stat::Max,
+            Precision::Exact(ScalarValue::Primitive(PValue::from(max))),
+        );
+
+        if PValue::from(min).cast::<T>().is_err() || PValue::from(max).cast::<T>().is_err() {
+            vortex_bail!(
+                Compute: "Cannot cast {} to {} — values exceed target range",
+                F::PTYPE, T::PTYPE,
+            );
+        }
```
robert3005: what does this protect against?

robert3005 (approved):

PR #8063, opened 2026-05-22
deser array tree from a vortex array

onursatici:
## Summary

Add support for serialising a `ArrayNode` tree into a vortex array, with array stats. Before we always serialised these into flatbuffers that were appended to the tail of the serialised array buffers. This PR adds the scaffolding to get a vortex array that can later at read time used to deserialise the buffers into ArrayRef's.

Main user of this would be the array tree layout, which will land in a followup PR

### DIFF
--- a/vortex-array/src/serde/columnar.rs
+++ b/vortex-array/src/serde/columnar.rs
@@ -1,0 +1,1244 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Columnar serialization of array trees — a parallel to [`SerializedArray`] that sources
+//! its encoding-tree metadata from a columnar representation rather than a flatbuffer.
+//!
+//! `SerializedArray` parses a per-array-tree flatbuffer (`fba::Array`) and navigates it via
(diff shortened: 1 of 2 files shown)


vortex-array/src/serde/columnar.rs
```diff
@@ -0,0 +1,1113 @@
...
+    "is_constant",
+    "is_sorted",
+    "is_strict_sorted",
+];
+
+/// The canonical dtype of a [`StatsColumns`] struct: 11 nullable typed columns mirroring
+/// `fba::ArrayStats`. Binary columns hold `ScalarValue::to_proto_bytes` blobs (decoded
+/// using the array's dtype at read time); the `*_exact` bools tag whether `min`/`max`
+/// were exact. `sum` is exact-only by construction so there is no `sum_exact` column.
+pub static STATS_COLUMNS_DTYPE: LazyLock<DType> = LazyLock::new(|| {
```
joseph-isaacs: this is soon doing to be a extensible scalar (that is from the agg partial).

vortex-layout/src/layouts/dict/mod.rs
joseph-isaacs, line 95: can you import

joseph-isaacs (commented):
My main question is about the hard coded stat serde

PR #8068, opened 2026-05-22
Tell duckdb if we're blocked on IO

myrrc:
If we're running Duckdb over slow storage (S3), getting ArrayRef's may block. Before this PR, scan function would block until getting an array. Now we can tell Duckdb we're waiting for IO, and it will suspend the worker thread and reschedule it when we get the array.

This doesn't have any impact if we're running over fast storage like NVME.

### DIFF
--- a/vortex-duckdb/src/table_function.rs
+++ b/vortex-duckdb/src/table_function.rs
@@ -41,6 +41,7 @@ use vortex::file::v2::FileStatsLayoutReader;
 use vortex::io::kanal_ext::KanalExt as _;
 use vortex::io::runtime::BlockingRuntime as _;
 use vortex::io::runtime::current::ThreadSafeIterator;
+use vortex::io::runtime::current::TryRecv;
 use vortex::layout::scan::multi::MultiLayoutChild;
 use vortex::layout::scan::multi::MultiLayoutDataSource;
 use vortex::metrics::tracing::get_global_labels;
@@ -131,6 +132,7 @@ impl<'a> TableInitInput<'a> {
 
 type ScanItem = VortexResult<(ArrayRef, Arc<ConversionCache>)>;
 type DataSourceIterator = ThreadSafeIterator<ScanItem>;
+type ScanItemFuture = BoxFuture<'static, Option<ScanItem>>;
 
 pub struct TableFunctionGlobal {
     iterator: DataSourceIterator,
@@ -158,8 +160,10 @@ pub struct TableFunctionLocal {
     exporter: Option<ArrayExporter>,
     partition_index: u64,
     file_index: usize,
-    // Aggregate scan accumulated partials. Empty for non-aggregate scan
+    /// Aggregate scan accumulated partials. Empty for non-aggregate scan
     partials: Vec<Box<dyn DynAccumulator>>,
+    /// In progress next item
+    pending: Option<ScanItemFuture>,
 }
 
 pub struct PartitionData {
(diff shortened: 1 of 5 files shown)


vortex-duckdb/src/table_function.rs
```diff
@@ -443,11 +448,43 @@ fn convert_result(array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Struc
...
+        let value = match aggregate {
+            ColumnAggregate::Real { .. } => {
+                let accum = accum_iter.next().vortex_expect("partial for real agg");
+                Value::try_from(accum.finish()?)?
+            }
+            ColumnAggregate::CountStar => Value::from(row_count),
+        };
+        chunk.get_vector_mut(idx).reference_value(&value);
+    }
+    chunk.set_len(1);
```
joseph-isaacs, line 479: ?

vortex-duckdb/src/table_function.rs
```diff
@@ -517,22 +535,77 @@ fn scan_aggregate(
...
+        // Duckdb calls this on a background task thread pool which is separate
+        // from table function processing pool. This call therefore doesn't
+        // block the query worker
+        let item = RUNTIME.block_on(&mut future);
+        // When this assignment is happening, thread related to "local_state"
+        // is suspended. It will be resumed only when wait_and_fetch returns
+        // so there's no race condition
+        local_state.pending = Some(Box::pin(std::future::ready(item)));
+    };
+}
```
joseph-isaacs, line 584: this needs very careful explanation.

It also shouldn't be in this mod I would this in the iterator def?

vortex-duckdb/src/table_function.rs
myrrc: Moved out of scan_aggregate because it had too big indentation, see below.

vortex-duckdb/src/table_function.rs
```diff
@@ -443,11 +448,43 @@ fn convert_result(array: ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Struc
...
+        return Ok(());
+    }
+
+    let mut accumulators = global_state.partials.lock();
+    let row_count = global_state.row_count.load(Ordering::Acquire) as i64;
+    let mut accum_iter = accumulators.iter_mut();
+    for (idx, aggregate) in global_state.aggregates.iter().enumerate() {
+        let value = match aggregate {
+            ColumnAggregate::Real { .. } => {
+                let accum = accum_iter.next().vortex_expect("partial for real agg");
```
joseph-isaacs, line 472: should this panic?

myrrc: Yeah, by definition we build partials for non-count(*) aggregates above, so they're present.

vortex-duckdb/src/table_function.rs
```diff
@@ -131,6 +132,7 @@ impl<'a> TableInitInput<'a> {
 
 type ScanItem = VortexResult<(ArrayRef, Arc<ConversionCache>)>;
 type DataSourceIterator = ThreadSafeIterator<ScanItem>;
+type ScanItemFuture = BoxFuture<'static, Option<ScanItem>>;
 
 pub struct TableFunctionGlobal {
     iterator: DataSourceIterator,
```
joseph-isaacs, line 138: is this used in the scan?

myrrc:
In addition to the comments ^ benchmarks likely show no impact because we can't really reproduce slow IO even on our S3. I'll think whether we can get a good reproduction where we will see a speedup, otherwise there's still no sense to merge this

myrrc:
This PR is superceded by #9355

Issue #7830, opened 2026-05-07
Tracking Issue: TurboQuant


connortsui20:
This is a tracking issue for adding TurboQuant quantization as a feature in Vortex, but notably it will _not_ be a new physical encoding for `Vector`, but rather it will be a new logical type.

## Motivation

We would like to support the lossy compression (quantization) of data in Vortex. However, it was not entirely clear how we would achieve this in a file format, where many of our invariants rely on the assumption that our physical encodings are all lossless.

After some internal discussions, we have realized that "lossy" data and compression MUST live at the logical layer, not at the physical layer. It is logical because losing data is a full modification of the data, not just a different way of storing it.

For TurboQuant is particular, this means that it **cannot** be an encoding of the `Vector` type. There are too many assumptions that break if were an encoding, such as unit-normalization breaks for quantized vectors, and scalar functions and canonicalization do not return the same result as if we had not quantized the vectors with TQ.

However, this does not mean we want to remove TQ entirely. We see that there is value in building tooling for users who want to _purposefully_ write TQ-quantized vectors into Vortex files, and reading it back knowing that some of the information in the original data has been lost.

Users should be able to write whatever data they want into Vortex, however they want to structure it. But if they want to use a lossy compression / quantization scheme, they need to first transform the data themselves before writing it to Vortex, and they additionally need to make sure that the default Vortex compressor does not recompress the data that they have specially modified.

## Design

It is unclear if this is the ideal way to generally design lossy schemes, but it is certainly the easiest in terms of moving forward.

TurboQuant will live in a new `vortex-turboquant` crate that lives outside of the top-level `vortex` dependency tree, mimicking a third-party crate. This will carry a new `TurboQuant` extension type over a struct array that carries all of the components needed to quantize vector data, which is just the norms of the vectors and the codes (indices) into the centroid book (values), and the centroids can be constructed at read time (cached on the # dimensions and bit width).

Creating a new extension type has the benefit that we do not need to customize the behavior of the compressor, as the default compressor will just compress the inner storage `StructArray` and not canonicalize into a `Vector` array.

But on the flip side, this means that we can no longer canonicalize into a `Vector` array, and we have to reimplement all of the scalar functions on vectors (inner product, cosine similarity, etc) specifically for TQ.

[...]

PR #8070, opened 2026-05-22
TurboQuant: Add distortion benchmark

connortsui20:
## Summary

Adds distortion / error benchmarking option for TurboQuant. Note that this benchmark is still over the old implementation of TurboQuant, but because the algorithm is identical it shouldn't affect any numbers.

We will cut over to the new version atomically once everything is implemented in `vortex-turboquant`.

## Testing

distortion

### DIFF
--- a/benchmarks/vector-search-bench/scripts/plot-turboquant-distortion.py
+++ b/benchmarks/vector-search-bench/scripts/plot-turboquant-distortion.py
@@ -11,7 +11,7 @@
 """Sweep bits-vs-distortion for TurboQuant and plot the curves.
 
 Calls `vector-search-bench distortion` for each (dataset, bits) combination, parses the
-table from stdout, and plots reconstruction NMSE and pairwise cosine-error curves with
+table from stdout, and plots reconstruction NMSE and squared cosine-error curves with
 mean/median/max shown on a log-scaled y-axis.
 
 Each `--dataset` value may optionally pin a train layout with a colon, e.g.
@@ -37,7 +37,7 @@ from pathlib import Path
 
 import matplotlib.pyplot as plt
 from matplotlib.lines import Line2D
-from matplotlib.ticker import MaxNLocator
+from matplotlib.ticker import MaxNLocator, NullLocator
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
 DEFAULT_BINARY = REPO_ROOT / "target" / "release" / "vector-search-bench"
@@ -46,9 +46,9 @@ METRIC_NAMES = [
     "reconstruction NMSE mean",
     "reconstruction NMSE median",
     "reconstruction NMSE max",
-    "decoded cosine err mean",
-    "decoded cosine err median",
-    "decoded cosine err max",
+    "decoded cosine sqerr mean",
+    "decoded cosine sqerr median",
+    "decoded cosine sqerr max",
 ]
 
 
@@ -143,15 +143,33 @@ def run_one(
     )
 
 
+# Refined small-b values from `main.tex` line 273-274 ("for b = 1, 2, 3, 4 we have
+# D_mse approx 0.36, 0.117, 0.03, 0.009"). Tighter than the general sqrt(3)*pi/2 * 4^(-b)
+# upper bound, which is what we fall back to for b >= 5.
+_NMSE_UPPER_REFINED = {1: 0.36, 2: 0.117, 3: 0.03, 4: 0.009}
+
+
 def nmse_bound_stage1(bits: int) -> float:
-    """Paper's NMSE upper bound for TurboQuant_mse (Stage 1).
+    """Paper's Stage-1 unit-norm reconstruction upper bound for TurboQuant_mse.
 
     From the Stage 1 theorem (`main.tex`, line 272): for a unit-norm vector `x` quantized
-    to `b` bits per coordinate, `E[||x - x'||^2] <= (sqrt(3)*pi/2) / 4^b`. Because `x` is
-    unit-norm, `||x - x'||^2` equals the normalized squared error `||x - x'||^2 / ||x||^2`,
-    so the bound applies to the `reconstruction NMSE mean` curve directly.
+    to `b` bits per coordinate, `E[||x - x'||^2] <= (sqrt(3)*pi/2) * 4^(-b)`. TurboQuant
+    internally normalizes each input before quantizing, so the bound applies to per-row
+    NMSE = `||x - x'||^2 / ||x||^2 = ||unit(x) - unit(x')||^2` directly. For small `b`
+    (1..=4) the paper gives tighter refined values; we splice those in.
     """
+    if bits in _NMSE_UPPER_REFINED:
+        return _NMSE_UPPER_REFINED[bits]
     return (math.sqrt(3.0) * math.pi / 2.0) / (4.0**bits)
+
+
+def nmse_lower_bound(bits: int) -> float:
+    """Paper's Shannon lower bound on Stage-1 unit-norm reconstruction.
+
+    From `main.tex` line 297: `D_mse(Q) >= 1/4^b` for any randomized `b`-bit quantizer.
+    Independent of dimension; applies to NMSE for the same reason as the upper bound.
+    """
+    return 1.0 / (4.0**bits)
 
 
 def compression_ratio(bits: int, dim: int) -> float:
@@ -168,18 +186,13 @@ def compression_ratio(bits: int, dim: int) -> float:
     return original_bytes / per_vector_bytes
 
 
-def cosine_bound(bits: int, dim: int) -> float:
-    """Paper's Stage-2 inner-product bound, rendered as an absolute-error envelope.
+def cosine_sqerr_lower_bound(bits: int, dim: int) -> float:
+    """Paper's Shannon lower bound on Stage-2 squared inner-product distortion.
 
-    From the Stage 2 theorem (`main.tex`, line 288): for unit y and an `x` quantized via
-    TurboQuant_prod (Stage 2, MSE + QJL residual), `E[|<y, x> - <y, x'>|^2] <=
-    sqrt(3)*pi^2/d * 4^(-b)`. Taking sqrt gives an upper envelope on the RMS error per
-    bit width, and by Jensen also on the mean abs error.
-
-    Caveat: Vortex currently implements only Stage 1 (no QJL residual correction). The
-    Stage 1 inner-product error is biased and can sit *above* this Stage-2 envelope.
+    From `main.tex` line 298: `D_prod(Q) >= ||y||^2 / d * 1/4^b` for any randomized
+    `b`-bit quantizer. With unit probes (`||y||^2 = 1`) this is `1 / (d * 4^b)`.
     """
-    return math.pi * (3.0**0.25) / math.sqrt(dim) / (2.0**bits)
+    return 1.0 / (dim * (4.0**bits))
 
 
 DATASET_PALETTE = [
@@ -224,7 +237,17 @@ def plot(runs: list[Run], args: argparse.Namespace) -> None:
         }
     )
 
-    fig, axes = plt.subplots(1, 3, figsize=(20, 6.5), constrained_layout=True)
+    # GridSpec with a dedicated bottom strip for the caption so the long text gets a real
+    # subplot rect: no clipping by `bbox_inches`, no overlap with axis labels, no reliance
+    # on matplotlib's `wrap=True` heuristic. Plot row gets the lion's share so the bottom
+    # caption strip doesn't dominate visually; legends are anchored above the axes via
+    # `bbox_to_anchor` (see `add_legends`), and constrained_layout reserves space for them
+    # inside the plot row.
+    fig = plt.figure(figsize=(22, 9.5), constrained_layout=True)
+    gs = fig.add_gridspec(2, 3, height_ratios=[12, 1])
+    axes = [fig.add_subplot(gs[0, i]) for i in range(3)]
+    caption_ax = fig.add_subplot(gs[1, :])
+    caption_ax.axis("off")
     fig.suptitle(
         f"TurboQuant distortion vs bits per coordinate"
         f"     (samples={args.samples:,}, seed={args.seed}, rounds={args.rounds})",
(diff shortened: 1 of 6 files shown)


connortsui20:
whoops forgot to make sure the binary is always up to date (edit: fixed)

danking:
@claude review with careful attention paid to the validity of the statistical arguments.

benchmarks/vector-search-bench/src/distortion.rs
```diff
@@ -0,0 +1,310 @@
...
+        bail!(
+            "distortion: flat element count {} not divisible by row count {num_rows}",
+            flat.len(),
+        );
+    }
+    Ok(flat.len() / num_rows)
+}
+
+/// Per-vector normalized reconstruction MSE. Rows whose original squared norm is below `1e-10`
+/// are dropped because their normalized error is numerically undefined.
```
danking: I think we need to switch to the correct definition of mean square error, $\frac{1}{N} \sum_i (x_i - \hat{x_i})^2$. This also means we don't need to elide low magnitude rows, because MSE is well defined regardless of norm.

benchmarks/vector-search-bench/src/distortion.rs
```diff
@@ -0,0 +1,310 @@
...
+    let mut sorted = samples.to_vec();
+    sorted.sort_by(|a, b| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal));
+    let mid = sorted.len() / 2;
+    let median = if sorted.len() % 2 == 1 {
+        sorted[mid]
+    } else {
+        0.5 * (sorted[mid - 1] + sorted[mid])
+    };
+
+    let max = samples.iter().copied().fold(f32::NEG_INFINITY, f32::max);
```
danking: I prefer `reduce(...).expect()` when we know it is non-empty. The choice of `f32::NEG_INFINITY` is correct in the sense that it is the zero element of max but I don't like having to reason about that! haha

benchmarks/vector-search-bench/src/distortion.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    let reconstruction = stats(&reconstruction_errors(&original, &decoded_flat, dim, n));
+
+    let half = n / 2;
+    let true_cosines = compute_cosines(emb.slice(0..half)?, emb.slice(half..2 * half)?, &mut ctx)?;
+    let decoded_cosines = compute_cosines(
+        decoded.slice(0..half)?,
+        decoded.slice(half..2 * half)?,
+        &mut ctx,
+    )?;
```
danking: Claude mentions this as well but I think this section has a few errors.

Equation 2 from the TurboQuant paper defines the inner product error as: for _all_ pair of vectors, x and y, in $R^d$, the following quantity
Image
is upper bounded by (from an unnumbered equation under "Inner Product TurboQuant"):
Image

We need to respect the quantifier and expectation ordering. We need to fix a bit width, an x vector, and a y vector, and then compute the squared cosine distance:
```rust
d_prods = []
for seed in random_seeds:
    diff = compute_cosines(y, x) - compute_cosines(y, decode(encode(x, seed)))
    d_prods.append(diff * diff)

d_prod_mean = sum(d_prods) / len(d_prods)
```
That value, `d_prod_mean`, in the limit of looping over all seeds, is bound by the error.

benchmarks/vector-search-bench/src/distortion.rs
```diff
@@ -0,0 +1,310 @@
...
+    let decoded_cosine = stats(&abs_diff(&true_cosines, &decoded_cosines));
```
danking: I don't think this is right. The paper uses the square of the difference.

Image

benchmarks/vector-search-bench/src/distortion.rs
danking: I'm not sure how to choose y. I think choosing a set of random vectors on startup is reasonable. The property should hold for any y you choose.

Again, another unnumbered equation, this one under "1.1 Problem Definition", but this argues that if we choose `y = x`, then the expectation of the difference is just zero. That seems like a separately useful statistic to measure and display.
Image

benchmarks/vector-search-bench/scripts/plot-turboquant-distortion.py
```diff
@@ -0,0 +1,549 @@
...
+    """Paper's normalized-MSE upper bound for TurboQuant_mse (Stage 1).
+
+    From the Stage 1 theorem (`main.tex`, line 272): for a unit-norm vector `x` quantized
+    to `b` bits per coordinate, `E[||x - x'||^2] <= (sqrt(3)*pi/2) / 4^b`. Vortex ships
+    this stage, so the bound applies to the `reconstruction MSE mean` curve directly.
+    """
+    return (math.sqrt(3.0) * math.pi / 2.0) / (4.0**bits)
+
+
+def mse_bound_stage2(bits: int) -> float:
```
danking: Dead code; why do we have this?

danking (changes requested):
I hid ChatGPT's comment because it is quite long and I think not quite correct about some statistical things, but you may find it helpful to read over it if anything I've written below is confusing.

connortsui20:
issue 1 is not a real issue because I say "normalized" in the chart but I adjusted it anyways, issue 2 is fake, and I just fixed issue 3 and 4

benchmarks/vector-search-bench/src/distortion.rs
connortsui20: ok so I decided to just remove the "normalization" factor of the dimensions since higher dimensions should theoretically result in lower distortion anyways, so not its just sum of squared errors.

connortsui20: oh wait I need to make these all unit-norm first

benchmarks/vector-search-bench/src/distortion.rs
connortsui20: I think it is fine to just have a bunch of random pairs of x and y, I think a specific random y and then a random set of x is fine in the expectation calculation because everything (should be) independent.

robert3005 (approved):

benchmarks/vector-search-bench/src/distortion.rs
```diff
@@ -0,0 +1,365 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! TurboQuant distortion measurement on real vector datasets.
+//!
+//! Reports per-vector NMSE (`||x - x'||^2 / ||x||^2 = ||unit(x) - unit(x')||^2`) and per-
```
danking: This equality is only true if the norm of x is equal to the norm of its reconstruction. I think we should just elide the equality.

danking: A "per-vector NMSE" doesn't make sense to me. NSME is "normalized mean square error", right? I agree that $||x - x'||^2 / ||x||^2$ is a normalized squared error, but there's no mean/expectation. I think the correct verbiage is:


```suggestion
//! Reports the normalized mean square error of this TurboQuant implementation over vectors from a configurable dataset [...]
```

No "per-vector" stuff.

benchmarks/vector-search-bench/src/distortion.rs
```diff
@@ -0,0 +1,365 @@
...
+    (0..num_rows)
+        .map(|row| {
+            let start = row * dim;
+            let end = start + dim;
+            let orig = &original[start..end];
+            let recon = &reconstructed[start..end];
+            let norm_sq: f32 = orig.iter().map(|&v| v * v).sum();
+            if norm_sq == 0.0 {
+                return 0.0;
+            }
```
danking, line 220: This gives me heebie jeebies. Any zero vector in the test data _erroneously reduces_ our estimate of the reconstruction error.

Suppose, for example, if all my test vectors are the zero vector and all my reconstructions are the all ones vector. The reconstruction error, as defined by the paper, would be $\sqrt{d}$! That's huge! But this function reports the reconstruction error as zero!

Are there any zero vectors in the test data?

benchmarks/vector-search-bench/src/distortion.rs
```diff
@@ -0,0 +1,365 @@
...
+        bit_width: config.bits,
+        seed: config.seed,
+        num_rounds: config.rounds,
+    };
+    let encoded = turboquant_encode(emb, &tq_config, &mut ctx)?;
+    let decoded_ext: ExtensionArray = encoded.execute(&mut ctx)?;
+    let decoded = decoded_ext.into_array();
+    let decoded_flat = extract_flat_f32(&decoded, &mut ctx)?;
+
+    let reconstruction = stats(&reconstruction_nmse(&original, &decoded_flat, dim, n));
```
danking, line 154: This is still not equivalent to $D_{mse}$ from the paper. This is, given some specific seed,

$\mathbb{E}_{x \in \mathbb{R}^d}[||x - Q^{-1}(Q(x, \textrm{seed}))||^2]$

The paper defines $D_{mse}$ as, for some specific vector x:

$\mathbb{E}_{\mathrm{seed} \in R}[||x - Q^{-1}(Q(x, \textrm{seed}))||^2]$

I realize you're trying to replicate Figure 3. I think we can do that, but it's a less useful figure than Figure 1. Figure 1 shows the _distribution_ of $D_{prod}$ over multiple bit widths and over all pairings of the 100,000 data vectors and the 1,000 query vectors. I'm not sure why the paper lacks a similar figure for $D_{mse}$. That seems to me a very reasonable figure to generate. Regardless, if you want to construct Figure 3, it should be constructed according to this:

$E_{x \in \mathbb{R}^d} [ \mathbb{E}_{\mathrm{seed} \in R}[||x - Q^{-1}(Q(x, \textrm{seed}))||^2]]$

Expectations are linear operators so your for loops can be `for seed: for x:` or `for x: for seed:`; however, there you _must_ average over multiple seeds for your plots to fairly characterize Vortex's TurboQuant.

benchmarks/vector-search-bench/src/distortion.rs
danking: Actually, why not just normalize the vectors after loading them? Then everything downstream can follow the paper's formalism exactly (e.g. we can just use mean square error here instead of normalizing). If any vectors are zero vectors, we just throw them out while loading.

danking (dismissed):
There are still some issues with the statistics. I'm happy to jump on a zoom tomorrow morning to talk through it.

benchmarks/vector-search-bench/src/distortion.rs
connortsui20: I just changed it to:

```
//! Reports the normalized mean square error (`||x - x'||^2 / ||x||^2`) and the squared
//! cosine-similarity error (`(cos(y_i, x_i) - cos(y_i, x'_i))^2`) against a set of independently
//! sampled unit-norm probe vectors `y_i`, after a full encode and decode roundtrip through the
//! [`vortex_tensor::encodings::turboquant`] scheme.
```

benchmarks/vector-search-bench/src/distortion.rs
connortsui20: I wanted to let this test the TurboQuant internal normalization which is why I don't normalize on read, but yes it a bit strange to just add 0 in these cases.

I'm pretty sure none of these datasets actually have 0 vectors but i'll change it regardless

benchmarks/vector-search-bench/src/distortion.rs
connortsui20: So I had the idea that this is still the same even for a fixed seed, or rather a feeling that expectation for a fixed seed and random vectors should be the same as random seed and fixed (or random) vectors.

I had claude look into this:

  The single-seed NMSE estimator is unbiased for a quantity that respects the bound. Theorem 1 gives

  $$D_{mse}(u) := \mathbb{E}_Q\!\left[\|u - Q^{-1}(Q(u))\|^2\right] \le B$$

  for any $`u \in \mathbb{S}^{d-1}`$ with $`B = \frac{\sqrt{3}\pi}{2} \cdot 4^{-b}`$. TurboQuant's decoder restores the original
  norm, so the per-row NMSE $`\|x_i - \hat{x}_i\|^2 / \|x_i\|^2`$ equals $`\|u_i - \hat{u}_i\|^2`$ for $`u_i = x_i/\|x_i\|`$, which
   is a single realization of $`D_{mse}(u_i)`$. The estimator

  $$\hat{T} = \frac{1}{N}\sum_i \frac{\|x_i - \hat{x}_i\|^2}{\|x_i\|^2}$$

  satisfies

  $$\mathbb{E}_Q[\hat{T}] = \frac{1}{N}\sum_i D_{mse}(u_i) \le B$$

  by linearity, since each $`D_{mse}(u_i) \le B`$ pointwise. So $`\hat{T}`$ isn't per-vector $`D_{mse}`$ (agreed), but it's an
  unbiased estimator of the dataset average of $`D_{mse}`$, bounded by the same constant. §5.1 / Figure 3 in the paper plots the
  same single-realization dataset-mean error against the bounds without any seed averaging, so I'd argue this matches the paper's
  own empirical standard.

  Two things still benefit from a seed loop if you want them: tighter variance (between-seed variance in $`\hat{T}`$ doesn't shrink
   with $`N`$, so rerunning with a different `config.seed` will move the number), and the `max` column being biased high by Jensen,

  $$\mathbb{E}_Q\!\left[\max_i \frac{\|x_i - \hat{x}_i\|^2}{\|x_i\|^2}\right] \ge \max_i D_{mse}(u_i).$$

  Can add a `num_seeds` knob in a follow-up if either reads as load-bearing here.

danking: The paper does not claim that, for an arbitrary choice of seed, the following is true

$D_{mse}(u_i) \le B$ 

The paper only proves that the expectation of that expression over many seeds,

$\mathbb{E} [ D_{mse}(u_i) ]$

is bounded above by $B$. So, in particular, that third equation does not follow by linearity. Moreover, it is not true that $E_Q[\hat{T}]=\frac{1}{N}\sum_i D_{mse}(u_i)$. The subscript Q on the expectation requires you to average over seeds, not vectors.

---

Concretely, any particular seed could return a value greater than `B` on every vector and still not violate the theorem. Now, it seems unlikely both that there exists such a pathological seed and that you picked that seed for this assessment.

Based on the figures you've shared, you seem to have selected a seed which exhibits decent behavior on these vectors. Averaging over seeds makes your argument that we correctly implemented TurboQuant stronger, but it is up to you to decide if that matters.

---

> Figure 3 in the paper plots the same single-realization dataset-mean error against the bounds without any seed averaging,

This is an assertion or inference. The paper, sadly, does not clearly describe how to reproduce Figure 3.

PR #8071, opened 2026-05-22
`ListLayout` leaf version

mhk197:
Adds `ListLayout`, a structured layout for `DType::List` columns that stores `elements`, `offsets`, and (when nullable) `validity` as independent child layouts, each with its own configurable strategy. 

`ListLayoutStrategy` canonicalizes input to `ListViewArray`, rebuilds it via `list_from_list_view`, and writes each child concurrently.

`ListReader` fetches `offsets` and `validity` concurrently. Then uses `offsets[0]..offsets[-1]` to derive a bounded elements range, rebase the offsets to start at zero, fetch only that slice of `elements`, and assemble the `ListArray`. This means partial-range reads only fetch the elements they actually need rather than the whole elements buffer.

### DIFF
--- a/vortex-layout/src/layouts/list/mod.rs
+++ b/vortex-layout/src/layouts/list/mod.rs
@@ -19,8 +19,8 @@ use vortex_error::vortex_ensure;
 use vortex_error::vortex_err;
 use vortex_error::vortex_panic;
 use vortex_session::VortexSession;
-use vortex_session::registry::ReadContext;
 
+use crate::LayoutBuildContext;
 use crate::LayoutChildType;
 use crate::LayoutEncodingRef;
 use crate::LayoutId;
@@ -124,7 +124,7 @@ impl VTable for List {
         metadata: &<Self::Metadata as DeserializeMetadata>::Output,
         _segment_ids: Vec<SegmentId>,
         children: &dyn LayoutChildren,
-        _ctx: &ReadContext,
+        _ctx: &LayoutBuildContext<'_>,
     ) -> VortexResult<Self::Layout> {
         validate_children(dtype, children.nchildren())?;
 
--- a/vortex-layout/src/layouts/list/mod.rs
+++ b/vortex-layout/src/layouts/list/mod.rs
@@ -68,7 +68,7 @@ impl VTable for List {
     }
 
     fn metadata(layout: &Self::Layout) -> Self::Metadata {
-        ProstMetadata(ListLayoutMetadata::new(layout.offsets_ptype))
+        ProstMetadata(ListLayoutMetadata::new(layout.offsets_ptype()))
     }
 
     fn segment_ids(_layout: &Self::Layout) -> Vec<SegmentId> {
@@ -124,10 +124,6 @@ impl VTable for List {
         children: &dyn LayoutChildren,
         _ctx: &ReadContext,
     ) -> VortexResult<Self::Layout> {
-        let elements_dtype = dtype
-            .as_list_element_opt()
-            .ok_or_else(|| vortex_err!("ListLayout requires a List dtype, got {dtype}"))?;
-
         let expected_children = if dtype.is_nullable() {
             NUM_CHILDREN_NULLABLE
         } else {
@@ -139,9 +135,14 @@ impl VTable for List {
             children.nchildren(),
         );
 
+        let elements_dtype = dtype
+            .as_list_element_opt()
+            .ok_or_else(|| vortex_err!("ListLayout requires a List dtype, got {dtype}"))?;
         let elements = children.child(ELEMENTS_CHILD_INDEX, elements_dtype.as_ref())?;
+
         let offsets_dtype = DType::Primitive(metadata.offsets_ptype(), Nullability::NonNullable);
         let offsets = children.child(OFFSETS_CHILD_INDEX, &offsets_dtype)?;
+
         let validity = if dtype.is_nullable() {
             Some(children.child(VALIDITY_CHILD_INDEX, &DType::Bool(Nullability::NonNullable))?)
         } else {
@@ -153,7 +154,6 @@ impl VTable for List {
             elements,
             offsets,
             validity,
-            offsets_ptype: metadata.offsets_ptype(),
         })
     }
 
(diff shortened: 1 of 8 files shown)


vortex-layout/src/layouts/list/reader.rs
```diff
@@ -0,0 +1,414 @@
...
+        &self,
+        field_mask: &[FieldMask],
+        split_range: &SplitRange,
+        splits: &mut BTreeSet<u64>,
+    ) -> VortexResult<()> {
+        self.offsets
+            .register_splits(field_mask, split_range, splits)?;
+        if let Some(validity) = &self.validity {
+            validity.register_splits(field_mask, split_range, splits)?;
+        }
```
joseph-isaacs, line 295: This looks odd why would they be different?

vortex-layout/src/layouts/list/reader.rs
```diff
@@ -0,0 +1,414 @@
...
+use crate::segments::SegmentSource;
+
+/// Reader for [`ListLayout`].
+pub struct ListReader {
+    layout: ListLayout,
+    name: Arc<str>,
+    session: VortexSession,
+    elements: LayoutReaderRef,
+    offsets: LayoutReaderRef,
+    validity: Option<LayoutReaderRef>,
```
joseph-isaacs, line 55: Are we sure this needs a layout?

vortex-layout/src/layouts/list/writer.rs
```diff
@@ -0,0 +1,394 @@
...
+        let (sequence_id, array) = chunk?;
+
+        // Canonicalize to ListView, then rebuild into zctl
+        let mut exec_ctx = session.create_execution_ctx();
+        let ListDataParts {
+            elements,
+            offsets,
+            validity,
+            ..
+        } = list_from_list_view(array.execute::<ListViewArray>(&mut exec_ctx)?)?.into_data_parts();
```
joseph-isaacs: Why to execute to a ListArray?

vortex-layout/src/layouts/list/writer.rs
```diff
@@ -0,0 +1,394 @@
...
+
+/// Strategy for writing list-typed arrays.
+///
+/// Single-chunk only. The strategy:
+///  1. Canonicalizes the input chunk into a [`ListViewArray`].
+///  2. Calls [`list_from_list_view`] to rebuild it into zero-copy-to-list form
+///     (sorted, gapless, non-overlapping offsets) and produce a [`ListArray`].
+///  3. Writes the `elements`, `offsets`, and (when nullable) `validity` columns into
+///     separately configurable downstream strategies, producing a single [`ListLayout`].
+///
```
joseph-isaacs: How does the sizing of the splits works, what is it based on?

gatesn: We just talked through this. This is the layout that really highlights why row-batches suck... and really why engines should be pushing down unnest operations into the source. But... they don't.

For now, the best we can do is register the validity and offsets splits, so we just delegate to the child.

vortex-layout/src/layouts/list/reader.rs
mhk197: they may be chunked differently and offsets will have one extra entry

vortex-layout/src/layouts/list/reader.rs
robert3005: You can probably can by with this being a segment

vortex-layout/src/layouts/list/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+                iter.next()
+                    .ok_or_else(|| vortex_err!("missing validity child"))
+            })
+            .transpose()?;
+        Ok(())
+    }
+}
+
+/// Validates expected number of children based on `dtype`
+#[inline]
```
connortsui20: We generally dont want to inline functions with branches, and also because this is a private function the compiler will likely inline this anyways.

inline is only useful for _public_ functions that are small

vortex-layout/src/layouts/list/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+fn validate_children(dtype: &DType, n_children: usize) -> VortexResult<()> {
+    let mut expected = NUM_CHILDREN_NON_NULLABLE;
+
+    if dtype.is_nullable() {
+        expected += 1;
+    };
+
+    vortex_ensure!(
+        n_children == expected,
```
connortsui20: vortex_ensure_eq!

vortex-layout/src/layouts/list/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+    #[prost(enumeration = "PType", tag = "1")]
+    offsets_ptype: i32,
+}
+
+impl ListLayoutMetadata {
+    pub fn new(offsets_ptype: PType) -> Self {
+        let mut metadata = Self::default();
+        metadata.set_offsets_ptype(offsets_ptype);
+        metadata
+    }
```
connortsui20, line 281: This seems like a strange constructor? Why not just do `Self { offsets_ptype }`?

vortex-layout/src/layouts/list/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+        vec![]
+    }
+
+    fn nchildren(layout: &Self::Layout) -> usize {
+        let mut n = NUM_CHILDREN_NON_NULLABLE;
+        if layout.dtype.is_nullable() {
+            n += 1;
+        }
+
+        n
```
connortsui20: I feel like I would prefer reading just an if branch and then return either `NUM_CHILDREN_NON_NULLABLE` or `NUM_CHILDREN_NON_NULLABLE + 1`?

connortsui20:
Is there a tracking issue that explains why we do not want to store sizes as a (potentially optional) layout child?

It would be nice to see some sort of design doc (or if that already exists, please attach it in the PR description).

vortex-layout/src/layouts/list/reader.rs
```diff
@@ -0,0 +1,414 @@
...
+        Ok(MaskFuture::ready(mask))
+    }
+
+    fn filter_evaluation(
+        &self,
+        _row_range: &Range<u64>,
+        _expr: &Expression,
+        _mask: MaskFuture,
+    ) -> VortexResult<MaskFuture> {
+        todo!()
```
connortsui20: todo for when?

vortex-layout/src/layouts/list/reader.rs
```diff
@@ -0,0 +1,414 @@
...
+                match validity_fut {
+                    Some(fut) => fut.await.map(Some),
+                    None => Ok(None),
+                }
+            },)?;
+
+            // Bound the elements read using offsets[0] and offsets[-1]
+            let elements_range = calculate_elements_range(&offsets, &session)?;
+
+            // Rebase the offsets so they start at zero
```
connortsui20: Nit: prose comments should end in periods (ask an llm to do this for you!).

vortex-layout/src/layouts/list/reader.rs
```diff
@@ -0,0 +1,414 @@
...
+            let validity = create_validity(validity_array, nullability);
+            let array = ListArray::try_new(elements, rebased_offsets, validity)?.into_array();
+
+            // Apply mask and expression
+            let mask = mask.await?;
+            let array = if !mask.all_true() {
+                array.filter(mask)?
+            } else {
+                array
+            };
```
connortsui20: @joseph-isaacs I feel like we should have this optimization in the `.filter` method for arrays instead of constructing the filter array and then immediately optimizing it away?

Basically for the "built-in" array methods we might as well do the optimizations immediately.

connortsui20 (commented):
Mostly nits, didn't get a chance to read through writer.rs though

vortex-layout/src/layouts/list/mod.rs
mhk197: conversion from enum to i32 in prost needs this apparently, see `DictLayout`

PR #8072, opened 2026-05-22
Add ArcSwapMap and use it throughout our session registries

robert3005:
ArcSwap is faster than a lock for read. These session are mutable but mutations
are rare and retrievals are common

### DIFF
--- a/vortex-array/src/aggregate_fn/accumulator.rs
+++ b/vortex-array/src/aggregate_fn/accumulator.rs
@@ -4,7 +4,6 @@
 use vortex_error::VortexResult;
 use vortex_error::vortex_ensure;
 use vortex_error::vortex_err;
-use vortex_session::SessionExt;
 
 use crate::ArrayRef;
 use crate::Columnar;
@@ -12,7 +11,6 @@ use crate::ExecutionCtx;
 use crate::aggregate_fn::AggregateFn;
 use crate::aggregate_fn::AggregateFnRef;
 use crate::aggregate_fn::AggregateFnVTable;
-use crate::aggregate_fn::session::AggregateFnSession;
 use crate::aggregate_fn::session::AggregateFnSessionExt;
 use crate::columnar::AnyColumnar;
 use crate::dtype::DType;
--- a/vortex-array/src/aggregate_fn/accumulator.rs
+++ b/vortex-array/src/aggregate_fn/accumulator.rs
@@ -142,7 +142,7 @@ impl<V: AggregateFnVTable> DynAccumulator for Accumulator<V> {
         }
 
         let session = ctx.session().clone();
-        let kernels = &session.aggregate_fns().kernels.load();
+        let aggregate_fns = session.aggregate_fns();
 
         // 1. Kernel registry first: a registered `(encoding, aggregate_fn)` kernel is strictly
         //    more specific than the vtable's `try_accumulate` short-circuit. Checking the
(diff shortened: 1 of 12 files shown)


vortex-array/src/aggregate_fn/session.rs
```diff
@@ -49,10 +49,10 @@ pub type AggregateFnRegistry = Registry<AggregateFnPluginRef>;
 /// Session state for aggregate function vtables.
 #[derive(Debug)]
 pub struct AggregateFnSession {
-    registry: AggregateFnRegistry,
```
onursatici, line 52: shall we delete this type alias now that it is unused?

vortex-array/src/aggregate_fn/session.rs
```diff
@@ -107,15 +107,20 @@ impl Default for AggregateFnSession {
 
 impl AggregateFnSession {
     /// Returns the aggregate function registry.
```
onursatici: this doc is now stale

vortex-array/src/aggregate_fn/accumulator.rs
```diff
@@ -142,21 +142,19 @@ impl<V: AggregateFnVTable> DynAccumulator for Accumulator<V> {
         }
 
         let session = ctx.session().clone();
-        let kernels = &session.aggregate_fns().kernels;
+        let kernels = &session.aggregate_fns().kernels.load();
```
onursatici: I think holding this for the entire body is mostly fine, but if we have some recursion here it might fallback to be as slow as the rwlock. I mean we load once then hold that guard and call the kernel, if it is a chunked array then that calls aggregate and calls load once more etc. I believe arcswap has a limited number of fast permits per thread and if we exhaust them then it falls back to refcount increments.

If we narrow the load scope to just one kernel execution in the loop below then that problem goes away. But it is unlikely that we will hit that level of recursion and the perf degradation is not that bad, it falls back to what rwlock does so up to you

robert3005: I think you're right. Let me narrow it down

robert3005: I have pushed the retrieval into the function so now it's only held during lookup

vortex-array/src/aggregate_fn/accumulator.rs
```diff
@@ -187,13 +182,8 @@ impl<V: AggregateFnVTable> DynAccumulator for Accumulator<V> {
...
             }
 
-            let kernels_r = kernels.read();
-            let batch_id = batch.encoding_id();
-            let kernel = kernels_r
-                .get(&(batch_id, Some(self.aggregate_fn.id())))
-                .or_else(|| kernels_r.get(&(batch_id, None)))
-                .copied();
-            drop(kernels_r);
+            let kernel =
```
joseph-isaacs: ❤️

vortex-array/src/aggregate_fn/session.rs
```diff
@@ -43,16 +42,13 @@ use crate::arrays::dict::compute::is_constant::DictIsConstantKernel;
 use crate::arrays::dict::compute::is_sorted::DictIsSortedKernel;
 use crate::arrays::dict::compute::min_max::DictMinMaxKernel;
 
-/// Registry of aggregate function vtables.
-pub type AggregateFnRegistry = Registry<AggregateFnPluginRef>;
```
joseph-isaacs, line 47: Shall we have a pattern like this for ArcSwap?

joseph-isaacs (approved):
Can we break this out into a pattern with a struct or type wrapper?

onursatici (approved):

robert3005:
I will make a follow up.

robert3005:
Looks like this broke the backcompat check https://github.com/vortex-data/vortex/actions/runs/26544463401/job/78201950621?pr=8072. Need to investigate this tomorrow as it's passing locally

PR #8076, opened 2026-05-24
Fix TurboQuant centroid initialization wasting most codes at high dimensions

mprammer:
TurboQuant's Lloyd–Max scalar quantizer seeded its centroids uniformly across the full support `[-1, 1]`, but after rotation each coordinate's marginal concentrates around 0 with standard deviation `1/sqrt(dimension)`. For `dimension >= 256` most seeds start in the near-zero-mass tails, where the conditional-mean update's zero-denominator guard freezes them on the first iteration — so a large fraction of codes end up quantizing coordinate values that never occur. At 8 bits and `dimension = 1024` that left reconstruction error at ~40x the paper's Theorem 1 bound and stopped it falling as bits increased. The fix seeds centroids on `±sqrt(bit_width) · sigma`, where the mass actually is, so no code freezes.

The seed scales with `sqrt(bit_width)` rather than a fixed multiple of sigma because the optimal spread grows with codebook size: a constant tight enough to be optimal at low bit widths under-covers the tail at high ones, and a constant wide enough for high bit widths wastes resolution lower down. So any single constant is dominated — 2.5σ is best in the mid range, 3.0σ at the top, neither everywhere — while `sqrt(bit_width)·sigma` ties the best constant at every bit width and is strictly best at 8 bits. Reconstruction error as a multiple of the Theorem 1 bound at `dimension = 1024`, lower is better (the pattern holds across dimensions):

| bits | `[-1,1]` (before) | 2.5σ | 3.0σ | 3.5σ | `√bits·σ` (this PR) |
|---|---|---|---|---|---|
| 4 | 1.15 | 0.89 | 0.89 | 0.89 | 0.89 |
| 6 | 3.66 | 0.97 | 1.04 | 1.18 | 0.97 |
| 8 | 40.6 | 1.25 | 1.21 | 1.49 | 1.17 |

The same seed change lands in both Lloyd–Max copies (`vortex-tensor` and `vortex-turboquant`), kept in lockstep by a cross-crate parity test, with a regression test asserting no code freezes across the dimension/bit grid and an `#[ignore]`d `sweep_centroid_init` that reproduces the table. Note the codebook itself changes: `vortex-tensor` arrays store their centroids inline so existing files are unaffected, but the `vortex-turboquant` extension path derives centroids at decode, so data written through it will dequantize against the new codebook.

### DIFF
--- a/vortex-tensor/src/encodings/turboquant/centroids.rs
+++ b/vortex-tensor/src/encodings/turboquant/centroids.rs
@@ -84,7 +84,41 @@ impl HalfIntExponent {
     }
 }
 
-/// Compute optimal centroids via the Max-Lloyd (Lloyd-Max) algorithm.
+/// How far to spread the initial centroids, as a multiple of the coordinate standard deviation
+/// `sigma = 1 / sqrt(dimension)`.
+///
+/// Seeding centroids across the full support `[-1, 1]` strands most of them in the near-zero-mass
+/// tails, where the zero-denominator guard in [`mean_between_centroids`] freezes them for every
+/// iteration; scaling the seed by `sigma` keeps every cell on live probability mass.
+#[derive(Clone, Copy, Debug)]
+enum InitSpread {
+    /// A constant multiple of `sigma`, independent of bit width. Only the sweep test constructs
+    /// this; production uses [`InitSpread::SqrtRate`].
+    #[cfg_attr(not(test), allow(dead_code))]
+    Fixed(f64),
+    /// `coeff * sqrt(bit_width)` multiples of `sigma`. A codebook with more levels needs a wider
+    /// seed to keep its outermost cells on live probability mass, so the spread grows with the bit
+    /// width — mirroring how a quantizer's optimal loading factor grows with rate.
+    SqrtRate(f64),
+}
+
+impl InitSpread {
+    /// The seed half-width, in multiples of `sigma`, for the given bit width.
+    fn sigmas(self, bit_width: u8) -> f64 {
+        match self {
+            InitSpread::Fixed(sigmas) => sigmas,
+            InitSpread::SqrtRate(coeff) => coeff * f64::from(bit_width).sqrt(),
+        }
+    }
+}
+
+/// Default centroid initialization. The seed half-width grows as `sqrt(bit_width)` standard
+/// deviations, tracking the bit-width-dependent optimum and beating every fixed multiple in
+/// `sweep_centroid_init` (including vLLM's `3.5 sigma`).
+const DEFAULT_INIT_SPREAD: InitSpread = InitSpread::SqrtRate(1.0);
+
+/// Compute optimal centroids via the Max-Lloyd (Lloyd-Max) algorithm with the
+/// [default initialization](DEFAULT_INIT_SPREAD).
 ///
 /// Operates on the marginal distribution of a single coordinate of a randomly rotated unit vector
 /// in d dimensions.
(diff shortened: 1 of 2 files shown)


connortsui20:
If we are going to do this, can this also figure out better boundaries? These are initialization changes, but we should also change the boundaries from -1, 1 to something like -10.5sigma, 10.5sigma (following the vLLM impl)

mprammer:
We checked this — the outer boundary is inert. Holding the init fixed (3σ here, to isolate the boundary) and sweeping the edge over ±1 (support), ±9σ, and ±10.5σ gives identical distortion (d=1024, error as a multiple of the Theorem 1 bound, lower is better):

| bits | ±1 (support) | ±9σ | ±10.5σ |
|------|--------------|-----|--------|
| 4    | 0.89         | 0.89| 0.89   |
| 6    | 1.04         | 1.04| 1.04   |
| 8    | 1.21         | 1.21| 1.21   |

The marginal sits within a few σ of 0, so the outermost cell's conditional mean is set by the near-zero mass no matter where the far boundary lands, independent of the init. (At d=128, ±1 is already ≈11σ; at higher d the extra span carries no mass.) We left the boundaries at [-1, 1] to keep the diff minimal, but it doesn't matter either way — the init spread is the only lever that moves distortion, which is what this PR changes. The full sweep, including the vLLM 3.5σ/10.5σ config, is reproducible through the ignored `sweep_centroid_init` test.

connortsui20:
in that case, I want a proof of why we sqrt(bit width) * sigma is optimal (either as a formal mathematical proof or numerically with a chart). It is still super unclear to me why changing the initial parameters has such a massive affect on the distortion, and it is further unclear why this is supposed to be "ideal"

danking:
@claude give this a detailed review paying special attention to the accuracy and precision of this method.

danking (approved):
I believe this is a correct improvement to TurboQuant but it *changes the interpretation of already written vectors*. I think this is OK as long as no releases of Vortex have been made since the original TQ PR merged.

danking:
Here's my reasoning for approving:

1. #8076 only changes to where the centroids are initialized. The algorithm still searches for an optimal quantization within [-1, 1]. The initial bins are kinda weird: the first and last bin can be much larger than the other bins: because they cover the gaps between `[-1, 1]` and the interval from which we sample. I think this is OK, if there happens to be a ton of mass near -1, the first centroid will be pulled towards that mass. The second centroid will then shift to fill that vacuum.

2. This Stanford course (EE269) argues that you should derive an analytical approximation of the CDF, sample some points in Unif[-1, 1] and send them through CDF^-1 to get decent starting centroids. What #8076 does is use a Uniform distribution on a truncated interval. That's also an analytical approximation (albeit not a great one!). https://web.stanford.edu/class/ee269/Lecture_nonuniform_quantization.pdf

I also think Claude is wrong to bring up sqrt(2 ln N). AFAICT, what it is saying is true for the Normal distribution. In particular, you need to truncate the Normal distribution's support from (-inf, inf) to something reasonable. Claude may be correct that setting the boundaries to sqrt(2 ln N) is "optimal" under some settings for the Normal; however, that's a distraction. We've already chosen the interval: [-1, 1]. As I describe above, we're just trying to choose a decent initialization. And as Martin points out, the optimizer gets stuck if the initial points land in a "too flat" section of the distribution.

connortsui20:
Id like to have a different PR that only changes the centroid initialization, rather than also have this init spread thing that I still don't understand.

connortsui20:
closing in favor of https://github.com/vortex-data/vortex/pull/8116. I don't understand the `IntiSpread` construct, but maybe it is something we want to have (I think it would be good to have a diagram explanation)?

Regardless, it is in the old vortex-tensor implementation only so merging this wouldn't do anything anyways

PR #8102, opened 2026-05-26
Use plain functions instead of vtab in duckdb

myrrc:
Remove vtab in favour of plain functions. We already exported them as C functions, so the change was purely removing mangling.
With vtab we had Rust-side functions duplicated 2 times (1 for ordinary scan and 1 for list scan). Remove duplication as well.
Split projection, ffi, and wrappers into their own respective files.


vortex-duckdb/src/ffi.rs
```diff
@@ -0,0 +1,221 @@
...
+    error_out: *mut cpp::duckdb_vx_error,
+) -> bool {
+    let bind_data = unsafe { bind_data.cast::<TableFunctionBind>().as_mut() }
+        .vortex_expect("bind_data null pointer");
+    let expr = unsafe { Expression::borrow(expr) };
+    try_or(error_out, || pushdown_complex_filter(bind_data, expr))
+}
+
+#[unsafe(no_mangle)]
+unsafe extern "C-unwind" fn duckdb_table_function_scan(
```
0ax1, line 104: `duckdb_table_function_scan` We have an ABI mismatch. This returns `bool` on the C side and `void` in Rust.

vortex-duckdb/src/ffi.rs
```diff
@@ -0,0 +1,221 @@
...
+    let node_stats =
+        unsafe { node_stats_out.as_mut() }.vortex_expect("node_stats_out null pointer");
+
+    match cardinality(bind_data) {
+        Cardinality::Unknown => {}
+        Cardinality::Estimate(c) => {
+            node_stats.has_estimated_cardinality = true;
+            node_stats.estimated_cardinality = c as _;
+        }
+        Cardinality::Maximum(c) => {
```
0ax1: We changed cardinality from `Inexact` to `Maximum` which sets a hard upper bound. Why is this sound?

vortex-duckdb/src/ffi.rs
```diff
@@ -0,0 +1,221 @@
...
+            node_stats.has_max_cardinality = true;
+            node_stats.max_cardinality = c as _;
+            node_stats.has_estimated_cardinality = true;
+            node_stats.estimated_cardinality = c as _;
+        }
+    }
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn duckdb_table_function_init_global(
```
0ax1, line 157: Given we now have more freestanding functions, can we throw more unit tests at it?

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -15,58 +15,69 @@ DUCKDB_INCLUDES_BEGIN
...
-    }
-
-    const duckdb_vx_tfunc_vtab_t vtab;
-};
+extern "C" {
+extern void duckdb_table_function_to_string(void *bind_data, duckdb_vx_string_map map);
+extern bool
+duckdb_table_function_statistics(void *bind_data, size_t column_index, duckdb_column_statistics *stats_out);
+extern double duckdb_table_function_scan_progress(void *global_state);
+extern void duckdb_table_function_get_partition_data(void *global_data,
```
0ax1: We re-declare the Rust exports here. This should be cbindgen'd.

myrrc: cbindgen doesn't give you the "extern" modifer in vortex.h, so the builds fail (I've tried that).

0ax1: Supposedly you can do sth like:
```
  [fn]
  prefix = "extern"
```
in you cbindgen toml which you can combine with
```
language = "C"
cpp_compat = true
```

vortex-duckdb/src/ffi.rs
myrrc: +1

vortex-duckdb/src/ffi.rs
myrrc: I assume the question is for cardinality() callback, not the ffi function. Yes, there is a bug there, fixed.

vortex-duckdb/src/ffi.rs
myrrc: I suggest for this, but in a separate PR given this one is large already.

vortex-duckdb/cpp/table_function.cpp
myrrc: +1

vortex-duckdb/src/duckdb/table_init_input.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use std::fmt::Debug;
+use std::fmt::Formatter;
+use std::fmt::Result;
+
+use crate::cpp;
+use crate::duckdb::TableFilterSet;
+use crate::duckdb::TableFilterSetRef;
+
+pub struct TableInitInput<'a> {
```
0ax1, line 12: You can use `#[repr(transparent)]` if you want a transparent wrapper on which you can impl your own fns but where you don't need field access indirection by the way.

0ax1 (approved):

PR #8103, opened 2026-05-26
like stat falsification respects escape characters

onursatici:
## Summary

In pruning of like expressions. We didn't remove the escape characters in the pattern, so we were getting wrong prefixes and exact patterns forwarded to further stat falsification

## Testing

unit tests

### DIFF
--- a/encodings/fsst/src/compute/like.rs
+++ b/encodings/fsst/src/compute/like.rs
@@ -291,6 +291,65 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


robert3005:
I have added a bail logic to the fsst dfa transform. It might be useful to apply the same fix there so we get support

robert3005 (approved):

onursatici:
fsst dfa fix here: https://github.com/vortex-data/vortex/pull/8107

PR #8104, opened 2026-05-26
feat: support nullable Arrow Device array export

0ax1:


vortex-cuda/src/arrow/canonical.rs
```diff
@@ -202,6 +206,58 @@ fn export_canonical(
...
+
+    let len = logical_validity.len();
+    let arrow_bitmap = if arrow_offset == 0 {
+        logical_validity.sliced()
+    } else {
+        let physical_bits = BitBuffer::collect_bool(len + arrow_offset, |physical_index| {
+            physical_index >= arrow_offset && logical_validity.value(physical_index - arrow_offset)
+        });
+        BitBuffer::new_with_offset(physical_bits.inner().clone(), len, arrow_offset)
+    };
```
robert3005: imho you logical_validity.slice(arrow_offset..).sliced().into_inner()

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -202,6 +202,59 @@ fn export_canonical(
...
+    Ok((Some(validity), null_count))
+}
+
+/// Convert logical validity bits into an Arrow validity byte buffer.
+///
+/// Arrow consumers read validity at `ArrowArray.offset + row`, so
+/// non-zero offsets need leading padding bits before logical row 0.
+fn to_arrow_validity_byte_buffer(
+    logical_validity: BitBuffer,
+    arrow_offset: usize,
```
onursatici: if the underlying offset of logical_validity matches arrow_offset, should we bypass padding?

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -202,6 +202,59 @@ fn export_canonical(
...
+    ctx: &mut CudaExecutionCtx,
+) -> VortexResult<(Option<BufferHandle>, i64)> {
+    let mask = validity.execute_mask(len, ctx.execution_ctx())?;
+    let null_count = i64::try_from(mask.false_count())?;
+    if null_count == 0 {
+        return Ok((None, 0));
+    }
+
+    let validity_buffer = to_arrow_validity_byte_buffer(mask.into_bit_buffer(), arrow_offset)?;
```
onursatici: if mask is all false, we are currently allocating a bit buffer here then go through it bit by bit and allocate another bit buffer from it below. I don't think it is a huge waste, but if there is a clean way to allocate a 0 byte buffer on the gpu directly maybe we can use that here?

vortex-cuda/src/arrow/canonical.rs
0ax1: yeah

vortex-cuda/src/arrow/canonical.rs
0ax1: good point, i added proper handling for all true/false.

vortex-cuda/src/arrow/canonical.rs
0ax1: It's not quite the same, I changed the `BitBufferMut` to `with_capacity` and added a comment why slice wouldn't work.

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -202,6 +208,65 @@ fn export_canonical(
...
+    validity: Validity,
+    len: usize,
+    arrow_offset: usize,
+    ctx: &mut CudaExecutionCtx,
+) -> VortexResult<(Option<BufferHandle>, i64)> {
+    let mask = validity.execute_mask(len, ctx.execution_ctx())?;
+    let null_count = i64::try_from(mask.false_count())?;
+    if null_count == 0 {
+        return Ok((None, 0));
+    }
```
robert3005: we don't need this, it's AllTrue below

vortex-cuda/src/arrow/canonical.rs
```diff
@@ -202,6 +208,65 @@ fn export_canonical(
...
+        .await?;
+
+    Ok((Some(validity), null_count))
+}
+
+/// Convert logical validity bits into an Arrow validity byte buffer.
+///
+/// Arrow consumers read validity at `ArrowArray.offset + row`, so
+/// non-zero offsets need leading padding bits before logical row 0.
+fn to_arrow_validity_byte_buffer(
```
robert3005: I am confused by this method. The underlying bytes are already aligned to full bytes, the offset is taken from the underlying bitbuffer which makes this whole operation unnecessary. All you want to is to get the underlying buffer and discard the offset

robert3005 (changes requested):
I think I originally missed something here. I don't understand why the offseting logic is necessary, that's how the buffers are laid out already

vortex-cuda/src/arrow/canonical.rs
0ax1: wiped

vortex-cuda/src/arrow/canonical.rs
0ax1: Yeh duh this was backwards actually, fixing.

0ax1: done

robert3005 (approved):

PR #8106, opened 2026-05-26
wipe duckdb source in build.rs if not fully extracted before

onursatici:
## Summary

Fix `Result::unwrap()` on an `Err` value: InvalidArchive("Symlink is not inherently safe")` when building vortex-duckdb and if a partially extracted archive exists on target

### DIFF
--- a/vortex-duckdb/build.rs
+++ b/vortex-duckdb/build.rs
@@ -7,6 +7,7 @@
 
 use std::env;
 use std::fs;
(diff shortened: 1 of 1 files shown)


AdamGS (approved):
nice 🥳

0ax1:
Do we prefer this over atomic swap-in of extracted dirs from tmp?

PR #8108, opened 2026-05-26
Fix native library publish for non amd64 platforms

robert3005:
When publishing native packages make sure different architectures don't override
each other

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -10,25 +10,40 @@ permissions:
 jobs:
   build:
     name: Build ${{ matrix.target }}
-    runs-on: ${{ matrix.runs-on }}
+    runs-on: >-
+      ${{ github.repository == 'vortex-data/vortex'
+          && matrix.runner
+          || matrix.fallback_runner }}
     timeout-minutes: 120
     strategy:
       fail-fast: false
       matrix:
         include:
           - target: aarch64-apple-darwin
-            runs-on: macos-latest
+            runner: macos-latest
+            fallback_runner: macos-latest
             archive: tgz
           - target: x86_64-apple-darwin
-            runs-on: macos-15-intel
+            runner: macos-15-intel
+            fallback_runner: macos-15-intel
             archive: tgz
+          # Linux targets cross-compile from an amd64 runs-on runner via
+          # cargo-zigbuild with glibc 2.31 pinning, matching the Python wheel
+          # and JNI builds.
           - target: aarch64-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -10,25 +10,40 @@ permissions:
...
     strategy:
       fail-fast: false
       matrix:
         include:
           - target: aarch64-apple-darwin
-            runs-on: macos-latest
+            runner: macos-latest
+            fallback_runner: macos-latest
             archive: tgz
           - target: x86_64-apple-darwin
```
AdamGS, line 26: maybe we should take the opportunity and just remove it?

.github/workflows/release-binaries.yml
```diff
@@ -10,25 +10,40 @@ permissions:
...
           - target: x86_64-unknown-linux-gnu
-            runs-on: ubuntu-24.04
+            runner: runs-on=${{ github.run_id }}/runner=amd64-medium/image=ubuntu24-full-x64-pre-v2/tag=release-vx-amd64-linux
+            fallback_runner: ubuntu-24.04
             archive: tgz
     steps:
+      - uses: runs-on/action@v2
+        if: github.repository == 'vortex-data/vortex' && contains(matrix.target, 'linux')
+        with:
+          sccache: s3
```
AdamGS, line 45: I think in general we try not to use caches for binary releases to make sure they are not poisoned? I might be wrong here.

.github/workflows/release-binaries.yml
AdamGS: also - can't we cross-compile to mac from linux?

java/vortex-jni/build.gradle.kts
```diff
@@ -112,31 +112,43 @@ tasks.register("makeTestFiles") {
...
         execOps.exec {
             workingDir = rootProject.projectDir.absoluteFile.parentFile
             executable = "cargo"
             args("build", "--package", "vortex-jni")
         }
 
-        copy {
-            from("${rootProject.projectDir.absoluteFile.parentFile}/target/debug/libvortex_jni.so")
-            into("$projectDir/src/main/resources/native/linux-amd64")
+        val osName = System.getProperty("os.name").lowercase()
```
AdamGS, line 132: I guess there's no `--target` 😢

AdamGS (approved):
LGTM

.github/workflows/release-binaries.yml
robert3005: I haven't looked into it. For removal we can wait until macos 27 is out imho and see what people do then

PR #8109, opened 2026-05-26
Use plain functions for copy function in duckdb

myrrc:
This also fixes the segfault of empty "name" passed to copy function, see

https://dataengineeringcentral.substack.com/p/benchmarking-vortex-file-format-vs

Run bindgen before cpp so that changes to ffi for duckdb would propagate to vortex.h before C++ code compilation.


vortex-duckdb/src/copy.rs
```diff
@@ -22,133 +21,111 @@ use vortex::file::WriteOptionsSessionExt;
...
 /// flushed to disk.
-pub struct GlobalState {
+pub struct CopyFunctionGlobal {
     write_task: Mutex<Option<Task<VortexResult<WriteSummary>>>>,
     sink: Option<Sender<VortexResult<ArrayRef>>>,
-    // Pool of background workers helping to drive the write task.
-    // Note that this is optional and without it, we would only drive the task when DuckDB calls
-    // into us, and we call `RUNTIME.block_on`.
-    #[expect(dead_code)]
-    worker_pool: CurrentThreadWorkerPool,
```
0ax1, line 58: Why did we remove this? Like the comment says we now only make progress on `block_on` rather than driving execution in the background.

myrrc: As an application plugin, we should never create extra threads to do work, and should rely on host threads as we do for table function. But I agree this should be done in a separate PR since we don't have benchmarks to measure write speed.

vortex-duckdb/build.rs
```diff
@@ -485,6 +485,6 @@ fn main() {
     let duckdb_include_dir = inner_dir.join("src").join("include");
     println!("cargo:rerun-if-changed=cpp/include");
     c2rust(&crate_dir, &duckdb_include_dir);
-    cpp(&duckdb_include_dir);
     rust2c(&crate_dir);
+    cpp(&duckdb_include_dir);
```
0ax1: We could clarify the names of these fns while we're at it. 
```
rust2c(&crate_dir);
cpp(&duckdb_include_dir);
c2rust
```
In isolation, without seeing the full context, it's not clear by name what they do.

vortex-duckdb/src/copy.rs
```diff
@@ -22,133 +21,111 @@ use vortex::file::WriteOptionsSessionExt;
...
-    }
+    })
+}
+
+pub fn copy_to_initialize_global(
+    client_context: &ClientContextRef,
+    bind_data: &CopyFunctionBind,
+    file_path: String,
+) -> VortexResult<CopyFunctionGlobal> {
+    // The channel size 32 was chosen arbitrarily.
```
0ax1, line 118: Would be nice to be more intentional about the queue size or document if sizing has an effect on perf or behavior.

vortex-duckdb/src/copy.rs
0ax1: +1

vortex-duckdb/src/copy.rs
myrrc: +1, but this should be a separate PR.

vortex-duckdb/build.rs
myrrc: +1

0ax1 (approved):

PR #8110, opened 2026-05-26
Reduce monomorphization by dispatching over unsigned integer widths

joseph-isaacs:
## Summary

This PR reduces code bloat from generic monomorphization by dispatching over unsigned integer widths (4 types) instead of all signed and unsigned integer types (8 types) in hot paths where values are known to be non-negative.

The key insight is that offsets, indices, lengths, and similar values are always non-negative. By reinterpreting signed buffers as their unsigned counterparts (same width, identical bit patterns), we can dispatch over 4 unsigned widths instead of 8 integer types. For nested matches (e.g., offset type × size type), this reduces the cross-product from 8×8=64 to 4×4=16 monomorphizations.


myrrc (approved):
LGTM

PR #8116, opened 2026-05-27
TurboQuant: better centroid initialization 

connortsui20:
## Summary

Followup of https://github.com/vortex-data/vortex/pull/8076

Changes the centroid initialization for `vortex-turboquant` to use `+- sqrt(bit_width) * sigma` instead of just [-1, 1]

## Testing

```sh
uv run benchmarks/vector-search-bench/scripts/plot-turboquant-distortion.py \
            --dataset cohere-small-100k:single \
            --dataset openai-medium-500k:single \
            --dataset bioasq-medium-1m \
            --dataset glove-small-100k \
            --dataset gist-small-100k \
            --dataset sift-small-500k \
            --output ~/Downloads/distortion-sqrt-sigma.png
```

distortion-3-sigma


danking:
@claude review with careful attention paid to the validity of the statistical arguments.

vortex-tensor/src/encodings/turboquant/centroids.rs
```diff
@@ -19,6 +31,8 @@ use vortex_utils::aliases::dash_map::DashMap;
 use crate::encodings::turboquant::MAX_BIT_WIDTH;
 use crate::encodings::turboquant::MIN_DIMENSION;
 
+// NB: Some of these numbers were arbitrarily chosen...
```
danking: Prefer calling out which specific constants are chosen without a clear principle with a comment on that constant.

vortex-tensor/src/encodings/turboquant/centroids.rs
danking: This PR does not change this, but this is the wrong name. This is the number of trapezoids we draw. We integrate at 1000 + 1 points. Consider if INTEGRATION_POINTS is two we'll split the interval into two trapezoids like this:

```
+
|\
|  \
|    \
|      +------+
|      |      |
|      |      |
+------+------+

0     dx     2dx
```

And we'll evaluate the function at all three points: 0, dx, and 2dx. It's a bit of a roundabout way of computing the area of the trapezoid but it works. We compute:

$0.5 f(x) + f(dx) + 0.5 f(2 dx)$

which is a bit easier to read as:

$\frac{f(x) + f(dx)}{2} + \frac{f(dx) + f(2dx)}{2}$.

danking (dismissed):
Two nits that can be addressed separately. Let's get this merged and shipped into Spiral.

connortsui20:
let me fix the nits and rebase on develop

vortex-tensor/src/encodings/turboquant/centroids.rs
connortsui20: I changed it to:

```rust
// NB: All of these constants were chosen arbitrarily.
```

vortex-tensor/src/encodings/turboquant/centroids.rs
connortsui20: changed to `INTEGRATION_TRAPEZOIDS`

danking (approved):
LGTM

PR #8120, opened 2026-05-27
perf[buffer]: iteration for fallible operations with validity

joseph-isaacs:
Currently use (and arrow) handle fallible operations with scalar (non-SIMD) code.

This PR add a trait and methods to have fast SIMD checked operations (includes cast) but verified else where that `checked_add` benefits


joseph-isaacs:
Open question is where to put this code?

robert3005:
Sounds like we want a crate in between the array and vortex-buffer or this could be a feature flag in vortex-buffer

joseph-isaacs:
I was thinking `vortex-compute` what only works with dtype, buffers and rust native types?

I cannot remember if this was a problem before?

robert3005:
there was no problem with vortex-compute iirc. There were some constructs that made it hard in the past but I think we're fine now.

vortex-array/benches/cast_primitive.rs
```diff
@@ -51,3 +49,38 @@ fn cast_u16_to_u32(bencher: Bencher) {
                 .execute::<Canonical>(ctx)
         });
 }
+
+/// Narrowing fallible cast that goes through `try_map_with_mask`. Inputs are bounded
+/// so every value fits, isolating the kernel's per-lane checked-cast overhead.
+#[divan::bench(args = SIZES)]
+fn cast_u32_to_u8(bencher: Bencher, n: usize) {
+    let mut rng = StdRng::seed_from_u64(42);
```
AdamGS, line 58: nit - make seed const

vortex-compute/src/lane_kernels/source.rs
```diff
@@ -0,0 +1,90 @@
...
+
+//! Read-only lane source — the [`IndexedSource`] trait and the [`LaneZip`] adapter.
+
+/// A length-known source supporting unchecked indexed reads.
+///
+/// Implemented for `&[T]` (with `T: Copy`) and for [`LaneZip`] over two `IndexedSource`s.
+/// The kernels in this crate require this trait instead of `Iterator` so that lane
+/// reads carry no inter-iteration data dependency — the autovectorizer treats each
+/// lane independently.
+pub trait IndexedSource {
```
AdamGS, line 12: These traits really look like `Buf/BufMut` from `bytes` which we already implement for buffers and are implemented for similar things

vortex-compute/src/lane_kernels/source.rs
```diff
@@ -0,0 +1,90 @@
...
+        // SAFETY: caller guarantees i < self.len().
+        unsafe { *<[T]>::get_unchecked(self, i) }
+    }
+}
+
+/// Pair of two [`IndexedSource`]s of equal length. Yields `(A::Item, B::Item)` per lane.
+///
+/// Use this to drive a binary kernel from two columns. Length equality is enforced
+/// at construction.
+pub struct LaneZip<A, B>(pub A, pub B);
```
AdamGS, line 60: Seems like a good candidate for implementing `Iterator` and `ExactSizeIterator`

vortex-compute/src/lane_kernels/map_into.rs
```diff
@@ -0,0 +1,498 @@
...
+
+use crate::lane_kernels::source::IndexedSource;
+
+/// Extension trait providing out-of-place lane-kernel methods on any [`IndexedSource`].
+///
+/// All methods have default implementations and are inherited via the blanket
+/// `impl<S: IndexedSource> IndexedSourceExt for S` below. Bring the trait into
+/// scope (`use vortex_compute::lane_kernels::IndexedSourceExt;`) to call
+/// them with method syntax: `values.try_map_masked_into(&mask, &mut out, f)`.
+pub trait IndexedSourceExt: IndexedSource + Sized {
```
AdamGS, line 20: Can we skip the ext traits? just increases the mental load of tracking where things happen IMO

robert3005: What's another way of doing this?

AdamGS: why can't it be default functions on the core trait?

vortex-compute/src/lane_kernels/map_in_place.rs
```diff
@@ -0,0 +1,367 @@
...
+                let val = unsafe { values.get_unchecked(idx) };
+                let result = f(val);
+                // SAFETY: caller guarantees base + count <= len.
+                unsafe { values.set_unchecked(idx, result) };
+            }
+        }
+
+        let mut values = self;
+        let len = values.len();
+        let chunks_count = len / 64;
```
AdamGS: I think making 64 a const here will make it both much clearer (because you can document it once), and the ability to do some more complex `cfg` based on CPU features (similar code in [Arrow](https://github.com/apache/arrow-rs/blob/c4a831a1c81be3c20dff510d73954fa6712d90d5/arrow-arith/src/aggregate.rs#L271))

vortex-array/src/arrays/primitive/compute/cast.rs
```diff
@@ -114,53 +115,101 @@ where
...
+
+    // Returns `true` if every value of `from` is representable in `to` without loss.
+    fn casts_losslessly_to(from: PType, to: PType) -> bool {
+        from.least_supertype(to) == Some(to)
+    }
+
+    // Skip the fallible kernel when type widening or (cached) min/max prove every value fits.
+    let target_dtype = DType::Primitive(T::PTYPE, Nullability::NonNullable);
+    let infallible = casts_losslessly_to(F::PTYPE, T::PTYPE)
+        || cached_values_fit_in(array, &target_dtype) == Some(true);
```
AdamGS: nit - `unwrap_or_default`

vortex-array/src/arrays/primitive/compute/cast.rs
```diff
@@ -114,53 +115,101 @@ where
...
         vortex_err!(
             Compute: "Cannot cast {} to {} — value exceeds target range",
             F::PTYPE, T::PTYPE,
         )
     };
-    let buffer: Buffer<T> = match &mask {
-        Mask::AllTrue(_) => BufferMut::try_from_trusted_len_iter(
+
+    // Returns `true` if every value of `from` is representable in `to` without loss.
+    fn casts_losslessly_to(from: PType, to: PType) -> bool {
```
AdamGS, line 126: doesn't need to be a function

vortex-array/src/arrays/primitive/compute/cast.rs
```diff
@@ -114,53 +115,101 @@ where
...
+    // Skip the fallible kernel when type widening or (cached) min/max prove every value fits.
+    let target_dtype = DType::Primitive(T::PTYPE, Nullability::NonNullable);
+    let infallible = casts_losslessly_to(F::PTYPE, T::PTYPE)
+        || cached_values_fit_in(array, &target_dtype) == Some(true);
+
+    let len = array.len();
+
+    // If F and T have the same byte width, try to take unique ownership of the buffer.
+    let same_bit_width = F::PTYPE.byte_width() == T::PTYPE.byte_width();
+    let owned: Option<BufferMut<F>> = if same_bit_width {
```
AdamGS: nit `same_bit_width.then(...)`

vortex-compute/src/lane_kernels/source.rs
joseph-isaacs: I think the random access read/write

vortex-compute/src/lane_kernels/map_into.rs
joseph-isaacs: I really don't think its worth having default functions here we likely want more of these

vortex-compute/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: Should this be in `vortex-buffer`?

vortex-compute/src/lane_kernels/map_into.rs
AdamGS: not sure I understand

vortex-compute/src/lib.rs
joseph-isaacs: I did think about putting this there, but it just seemed like the wrong place

vortex-array/src/arrays/primitive/compute/cast.rs
joseph-isaacs: I prefer this only that the body does easily read like that?

vortex-compute/src/lane_kernels/map_into.rs
joseph-isaacs: Do this really matter?

AdamGS (approved):

robert3005:
we need to add the crate to the benchmarks, since it's a new one it has to be enumerate in the ci matrix

PR #8121, opened 2026-05-27
Make `Mask` execution strict about nullable booleans

joseph-isaacs:
## Summary

Previously `Mask::execute` accepted **nullable** boolean arrays and silently
coerced null elements to `false` (folding validity into the value bits). That
behaviour was flagged with a `TODO` questioning whether the implicit coercion
was correct.

This PR makes `Mask` execution **strict**: `Mask::execute` now requires a
non-nullable boolean array (`Bool(NonNullable)`) and errors on nullable input.
The null-as-`false` coercion is no longer implicit — callers that want
SQL-style `NULL`-as-not-matching semantics opt in explicitly via
`array.fill_null(false)` before executing into a `Mask`.


gatesn:
Why is this better than before?

joseph-isaacs:
First the semantics of the `execute::<Mask>` is very surprising, main reason. Two removes an extra alloc and `&`

gatesn:
I'm not sure the semantics are that surprising...? We can remove the alloc/& by checking the nullability of the dtype.

joseph-isaacs:
Mapping null to false is a hidden default (useful only in layouts) as seen in this PR.

robert3005:
What are cases where you would want to discard validity?

gatesn:
I really don't understand the alternative. Don't execute to a mask? Or force the user to call array.null_as_false().execute::<Mask>().

execute::<MaskNullAsFalse>() just feels clunky to me.

robert3005:
I think the only possible interpretation of execute to mask is that we fold nulls as false. The problem you run into is that it's not always correct to do and you should execute to BoolArray. If you know you have non nullable bool array like validity you just get a shortcut.

joseph-isaacs:
`array.null_as_false().execute::()` this heap allocates. Mask should never touch validity.

robert3005:
I propose that we 1. have `Mask` be the `MaskNullAsFalse` and 2. current `Mask` renamed to `NonNullableMask`

gatesn:
> `array.null_as_false().execute::()` this heap allocates. Mask should never touch validity.

Heap allocates what exactly?

Null-as-false is a perfectly valid scalar function. We should just treat it like one. You're over-optimizing if you're trying to avoid wrapping this up in a ScalarFnArray.

joseph-isaacs:
@robert3005 thoughts

robert3005:
I recently optimised in place owned bitbuffer operations such that things like null as false will attempt to reuse the existing buffer where possible. Is there still a difference then?

joseph-isaacs:
Yes the semantics are the *ONLY* important change here

robert3005:
Ok by what are we saving by this not being a scalar function? I think we should change the semantics of mask to require non nullable bool but then... I think I agree with @gatesn, why is this another executable target?

onursatici (approved):

robert3005 (approved):

PR #8123, opened 2026-05-27
Benchmark reporting improvements (better engine-specific comparison, better file size comparison) 

AdamGS:
## Summary

This PR includes two changes:
1. Merge the file size tracking into the timing stuff we already have, so its included in the same comment and has the same logic to figure out the baseline which is currently buggy.
2. Try and do inter-engine perf comparison, which is relevant when we do engine-specific work.


AdamGS:
GH is having a bit of a hard time, but it should work IMO.

joseph-isaacs:
Verdict and Within-engine, could we add a helper somewhere in the comment in a hidden section, explaining them

AdamGS:
btw changed "within-engine" to "engines", just happened after I triggered this run

AdamGS:
most up to date format: https://github.com/vortex-data/vortex/pull/8123#issuecomment-4554174006

myrrc (approved):
LGTM

PR #8125, opened 2026-05-27
Support for importing Arrow extension arrays into core vortex arrays and ParquetVariant support.

AdamGS:
## Summary

Allows importing/exporting vortex and arrow arrays in the case where the arrow side is an extension but Vortex isn't.

This PR also includes the motivating change - support for ParquetVariant import/export.

## API Changes

Changes the signature of `ArrowImportVTable::from_arrow_array` to include both the `Field` (which has the full arrow-side type info) and a full  vortex `DType`.

### DIFF
--- a/encodings/parquet-variant/src/arrow.rs
+++ b/encodings/parquet-variant/src/arrow.rs
@@ -48,6 +48,8 @@ impl ArrowExportVTable for ParquetVariant {
         dtype: &DType,
         _session: &ArrowSession,
     ) -> VortexResult<Option<Field>> {
+        // TODO(#8135): This is wrong and won't work for shredded arrays.
+        // We need to be able to access array metadata to accurately provide Arrow schemas.
         if !dtype.is_variant() {
             return Ok(None);
         }
--- a/encodings/parquet-variant/src/arrow.rs
+++ b/encodings/parquet-variant/src/arrow.rs
@@ -1,0 +1,123 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use std::sync::Arc;
+
+use arrow_array::Array as _;
+use arrow_array::ArrayRef as ArrowArrayRef;
+use arrow_array::cast::AsArray;
+use arrow_schema::DataType;
+use arrow_schema::Field;
+use arrow_schema::extension::EXTENSION_TYPE_NAME_KEY;
+use parquet_variant_compute::VariantArray as ArrowVariantArray;
+use vortex_array::ArrayRef;
+use vortex_array::ExecutionCtx;
+use vortex_array::IntoArray;
+use vortex_array::VTable;
+use vortex_array::arrow::ArrowExport;
+use vortex_array::arrow::ArrowExportVTable;
+use vortex_array::arrow::ArrowImport;
+use vortex_array::arrow::ArrowImportVTable;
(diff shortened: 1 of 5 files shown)


connortsui20:
do we know all of the possible arrow extension types that are considered not extension by Vortex? Is it just Variant?

vortex-tensor/src/types/vector/arrow.rs
```diff
@@ -362,13 +366,11 @@ mod tests {
...
-        let ext = dtype
-            .as_extension_opt()
-            .expect("vector dtype should be an extension dtype")
-            .clone();
+        let field = Field::new("embedding", DataType::Int32, false);
 
         let int_array: ArrowArrayRef = Arc::new(Int32Array::from(vec![1, 2, 3]));
-        let result = <Vector as ArrowImportVTable>::from_arrow_array(&Vector, int_array, &ext)?;
+        let result =
+            <Vector as ArrowImportVTable>::from_arrow_array(&Vector, int_array, &field, &dtype)?;
```
connortsui20, line 375: It does feel a bit strange that we pass in both the `Vector` vtable as well as the extension dtype here? Maybe we don't care

AdamGS:
@connortsui20 8-bit boolean too.

encodings/parquet-variant/src/arrow.rs
```diff
@@ -0,0 +1,123 @@
...
+impl ArrowExportVTable for ParquetVariant {
+    fn arrow_ext_id(&self) -> Id {
+        *ARROW_PARQUET_VARIANT
+    }
+
+    fn vortex_ext_id(&self) -> Id {
+        ParquetVariant.id()
+    }
+
+    fn to_arrow_field(
```
robert3005, line 125: Don’t you need to fix this as well to handle non ext dtypes?

AdamGS: yeah, there's also a bunch of other things here that might be wrong.

AdamGS: addressed, also realized some other small things around it needed to be changed.

robert3005: But don't you actually need to implement it? Shouldn't this produce an arrow field with variant type metadata?

robert3005:
The methods here still look weird, in particular to_arrow_field will never be called unless the source is an extension dtype which is not the case for things like 8-bit bool. However, the export should work just fine as long as the exporter doesn't look at the target dtype. The more confounding factor though is that for variant there's no real implementation you can have because of #8135, i.e you need to know physical information in order to export it. @a10y @palaska have you thought about these cases? Basically the code as it is now (and even after this pr is still subtly broken in a way that I find hard to unwind)

robert3005 (approved):
Let's merge this. I think we need to do better on schema inference but this will require some more thought

a10y:
My thought with these VTables originally was that the only thing that should be supported as pluggable is Arrow extension <-> Vortex extension. I think it starts to get confusing to allow mapping between arbitrary `DType`s and `Field`s.

Re 8-bit bool: I wonder if the 8-bit bool encoding actually should be removed in favor of an extension type around u8? If you really care about getting back bytebools in Arrow that's a decision you need to make at write time, it's not just scanning any other bool column IMO.

robert3005:
Variant is a real problem though but maybe that's not pluggable but then again there's at least 3 variant impls that we know about which behave differently

PR #8133, opened 2026-05-27
bool ree

blaginin:


encodings/runend/src/compress.rs
```diff
@@ -94,6 +95,67 @@ pub fn runend_encode(
     (ends, values)
 }
 
+/// Run-end encode a `BoolArray`, returning a tuple of `(ends, values)`.
+///
+/// The `values` array is a [`BoolArray`] for ordinary inputs. For all-invalid inputs it is a
+/// single-row null [`ConstantArray`].
+pub fn runend_encode_bool(
```
gatesn, line 102: We used to have a custom run-end-bool encoding at some point.

The point is you don't need to store the "values" array because you know it just flip-flops. So you could say, values starts at true, and maybe offsets are `0, 0, 4, 10, ...`

gatesn: But also... we already support run-end bools because we sometimes push-down a predicate expression over non-bool run-end types. So we can use this as a compression scheme now, and update later to preferred run-end bool if we want one.

robert3005: whenever we push down expression that end up with run end bools we immediately canonicalise them. The format we have is really wasteful for bools.

joseph-isaacs:
Can you show us a dataset where this matters?

vortex-compressor/src/stats/bool.rs
```diff
@@ -84,12 +98,70 @@ impl BoolStats {
...
+    let mut runs = 1;
+    for value in iter {
+        if value != previous {
+            previous = value;
+            runs += 1;
+        }
+    }
+
+    runs
+}
```
joseph-isaacs, line 146: these looks *very slow*

PR #8136, opened 2026-05-28
fix: speed up coverage tests and LLVM tools path

0ax1:


.github/workflows/rust-instrumented.yml
```diff
@@ -47,13 +47,19 @@ jobs:
           sccache: s3
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
       - uses: ./.github/actions/setup-prebuild
```
joseph-isaacs, line 49: add llvm to the prebuild image?

0ax1: Yeh but not in this PR.

encodings/fsst/src/tests.rs
```diff
@@ -111,60 +111,28 @@ fn test_fsst_array_ops() {
...
-    println!("training FSST compressor");
     let compressor = fsst_train_compressor(&array);
     let len = array.len();
     let dtype = array.dtype().clone();
     let mut ctx = LEGACY_SESSION.create_execution_ctx();
 
-    println!("compressing to FSST");
     let compressed = fsst_compress(array, len, &dtype, &compressor, &mut ctx);
     assert_eq!(compressed.len(), len);
+    assert_eq!(compressed.codes().offsets().dtype().as_ptype(), PType::I64);
```
joseph-isaacs: so we unconditionally upcast to u64 now?

if you think this is equiv then we should add a comment here also explaining this (or link to above)

connortsui20:
fyi we are not cutting over this version of the benchmarks website with server so those changes won't affect anything

encodings/fsst/src/tests.rs
0ax1: reverting to the prev test. should be moved to a diff test target in a follow up probably.

.github/workflows/rust-instrumented.yml
joseph-isaacs: You want to land this first?

connortsui20 (approved):

.github/workflows/rust-instrumented.yml
0ax1: aight

0ax1: done

robert3005 (approved):

PR #8137, opened 2026-05-28
Use selection in repeated scans

### LINKED ISSUE #7915: Tracking Issue: random access Lance parity

Big improvements:

Moving the needle (won't do):

TODOs:

Additional context:

myrrc:
Only self.row_range was used before, meaning for selection we always re-read the original files. A symptom of this was 97% of CPU time in spawn() calls in random access benchmarks.

This PR adds self.selection usage to ranges/splits calculation.
As a result, Selection is no longer needed in TaskContext as it's evaluated before spawning the split task.

This PR also moves lance projection calculation out of take() to make the comparison more fair (take() re-reads schema all the time).

### DIFF
--- a/vortex-layout/src/scan/repeated_scan.rs
+++ b/vortex-layout/src/scan/repeated_scan.rs
@@ -17,6 +17,7 @@ use vortex_array::iter::ArrayIterator;
 use vortex_array::iter::ArrayIteratorAdapter;
 use vortex_array::stream::ArrayStream;
 use vortex_array::stream::ArrayStreamAdapter;
+use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_io::runtime::BlockingRuntime;
 use vortex_io::session::RuntimeSessionExt;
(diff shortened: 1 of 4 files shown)


vortex-scan/src/selection.rs
```diff
@@ -152,6 +152,19 @@ impl Selection {
     }
 }
 
+/// true if "roaring" contains an element that's in "range"
+pub fn selection_roaring_intersects(roaring: &roaring::RoaringTreemap, range: &Range<u64>) -> bool {
+    // vortex-web's wasm has roaring 0.11.3 which doesn't have
```
robert3005: we can update this

vortex-layout/src/scan/repeated_scan.rs
robert3005: there's a method called `selection.row_mask` that does the intersection with the row range and produces the mask. I think you want to change the flow here such that you do this check before you create a task

vortex-scan/src/selection.rs
myrrc: +1, but let's make it a separate PR to keep this one small.

vortex-layout/src/scan/repeated_scan.rs
myrrc: It's already called inside split_exec, so we can return early and avoid task planning

myrrc:
Lance slowdown is an outlier, it was 0.76x previous run, so likely noise.

vortex-scan/src/selection.rs
```diff
@@ -43,6 +43,10 @@ impl Selection {
 
     /// Extract the [`RowMask`] for the given range from this selection.
     pub fn row_mask(&self, range: &Range<u64>) -> RowMask {
+        if range.start >= range.end {
+            return RowMask::new(0, Mask::AllFalse(0));
```
onursatici, line 47: nit: we lose the row offset here, we can return range.start instead of 0. Being all false it doesn't really matter

onursatici (approved):

PR #8138, opened 2026-05-28
ci: explicitly list used AMIs

0ax1:


robert3005:
I think you want to always take the latest

robert3005:
We should make sure these images expire so we publish new versions with patches

0ax1:
> We should make sure these images expire so we publish new versions with patches

There's something unsettling about un-pinned builds, but I tend to agree. Keeping the names explicitly but switched to `-*` such that the latest should be picked.

robert3005 (approved):

PR #8140, opened 2026-05-28
Add dtype helper macro

connortsui20:
## Summary

I often have found myself getting annoyed at the amount of boilerplate needed to define different dtypes.

This is a simple macro that autogenerates dtypes and nested dtypes. I specifically exclude struct fields from this macro since people might have different opinions on what the syntax should be.

Also adds a `const` version of `DecimalDType` construction since I would like to have the dtype macro fail if we try to construct something bad.

## Testing

Basic unit tests, including some deeper nested dtypes.

### DIFF
--- a/vortex-array/src/dtype/decimal/mod.rs
+++ b/vortex-array/src/dtype/decimal/mod.rs
@@ -87,6 +87,38 @@ impl DecimalDType {
             .unwrap_or_else(|e| vortex_panic!(e, "Failed to create DecimalDType"))
     }
 
+    /// `const fn` constructor for a `DecimalDType`.
+    ///
+    /// Enforces the same constraints as [`Self::try_new`] but is usable in `const` contexts
+    /// (e.g. inside `const { ... }` blocks), which means invalid `(precision, scale)` pairs
+    /// are rejected at compile time.
+    ///
+    /// # Panics
+    ///
+    /// Panics (at compile time when called in a `const` context) if `precision` is zero,
+    /// `precision` exceeds [`MAX_PRECISION`], `scale` exceeds [`MAX_SCALE`], or `scale` is
+    /// positive and greater than `precision`.
+    #[allow(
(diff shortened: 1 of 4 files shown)


vortex-array/src/dtype/decimal/mod.rs
```diff
@@ -87,6 +87,38 @@ impl DecimalDType {
...
+            "decimal precision exceeds MAX_PRECISION"
+        );
+        assert!(scale <= MAX_SCALE, "decimal scale exceeds MAX_SCALE");
+        assert!(
+            scale <= 0 || (scale as u8) <= precision,
+            "decimal scale is greater than precision"
+        );
+
+        // SAFETY: `precision > 0` was asserted above.
+        let precision = unsafe { NonZero::new_unchecked(precision) };
```
joseph-isaacs, line 118: use the safe ctr?

connortsui20: The safe one returns an option

joseph-isaacs: so this that instead of an assert above.

vortex-array/src/dtype/macros.rs
```diff
@@ -0,0 +1,387 @@
...
+        let inner = StructFields::from_iter([("mean", dtype!(f64?)), ("count", dtype!(u32))]);
+        let outer = StructFields::from_iter([
+            ("id", dtype!(i64)),
+            ("label", dtype!(utf8?)),
+            ("values", dtype!(list(f32)?)),
+            ("codes", dtype!(fixed_size_list(u8?)[4])),
+            ("metrics", dtype!(struct(inner)?)),
+        ]);
+        let dt = dtype!(struct(outer.clone())?);
+        assert_eq!(dt, DType::Struct(outer, Nullable));
```
joseph-isaacs, line 385: lets add structs!

PR #8142, opened 2026-05-28
fix: code coverage ci task

0ax1:
The installed LLVM tools reside under 
```
RUST_SYSROOT="$(rustc --print sysroot)"
RUST_HOST="$(rustc --print host-tuple)"
LLVM_TOOLS_BIN="${RUST_SYSROOT}/lib/rustlib/${RUST_HOST}/bin"
```
which is different from what we previously expected.

Further, splitting up the commands like
```
RUST_SYSROOT="$(rustc +$NIGHTLY_TOOLCHAIN --print sysroot)"
RUST_HOST="$(rustc +$NIGHTLY_TOOLCHAIN --print host-tuple)"
LLVM_TOOLS_BIN="${RUST_SYSROOT}/lib/rustlib/${RUST_HOST}/bin"
```
leads to non-zero exit code in case any of them fail. This is unlike the previous setup where only the outermost command `dirname` was checked for its exit code which is 0, even if `""` is passed as an argument.

### DIFF
--- a/.github/workflows/fuzz-coverage.yml
+++ b/.github/workflows/fuzz-coverage.yml
@@ -40,11 +40,14 @@ jobs:
             rustup component add llvm-tools-preview --toolchain $NIGHTLY_TOOLCHAIN
 
           # Verify llvm-profdata is accessible (cargo-fuzz needs it for coverage merging)
-          LLVM_PROFDATA="$(rustc +$NIGHTLY_TOOLCHAIN --print sysroot)/lib/rustlib/$(rustc +$NIGHTLY_TOOLCHAIN -vV | sed -n 's|host: ||p')/bin/llvm-profdata"
+          RUST_SYSROOT="$(rustc +$NIGHTLY_TOOLCHAIN --print sysroot)"
+          RUST_HOST="$(rustc +$NIGHTLY_TOOLCHAIN --print host-tuple)"
+          LLVM_TOOLS_BIN="${RUST_SYSROOT}/lib/rustlib/${RUST_HOST}/bin"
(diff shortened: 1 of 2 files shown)


.github/workflows/fuzz-coverage.yml
```diff
@@ -40,11 +40,14 @@ jobs:
             rustup component add llvm-tools-preview --toolchain $NIGHTLY_TOOLCHAIN
 
           # Verify llvm-profdata is accessible (cargo-fuzz needs it for coverage merging)
-          LLVM_PROFDATA="$(rustc +$NIGHTLY_TOOLCHAIN --print sysroot)/lib/rustlib/$(rustc +$NIGHTLY_TOOLCHAIN -vV | sed -n 's|host: ||p')/bin/llvm-profdata"
+          RUST_SYSROOT="$(rustc +$NIGHTLY_TOOLCHAIN --print sysroot)"
```
0ax1, line 43: Also changed to the more robust way to find the path here.

.github/workflows/rust-instrumented.yml
```diff
@@ -47,14 +47,33 @@ jobs:
...
+
       - name: Rust Tests
         if: ${{ matrix.suite == 'tests' }}
         run: |
           cargo nextest run --locked --workspace --all-features --no-fail-fast
       - name: Generate coverage report
         run: |
-          set -euo pipefail
-          LLVM_TOOLS_BIN="$(dirname "$(rustup which llvm-profdata)")"
+          RUST_SYSROOT="$(rustc --print sysroot)"
```
0ax1, line 74: `set -euo pipefail` is not needed. GH does set `set -e`.

robert3005 (approved):

PR #8144, opened 2026-05-29
feat: add experimental `onpair` string encoding

joseph-isaacs:
Adds the `vortex-onpair` array encoding under `encodings/experimental/onpair`, sourcing the compression algorithm from the standalone `onpair` crate (local path dependency) rather than vendored code.

- vortex-onpair: Vortex array wrapping, serialisation, and cast/filter pushdown only; train/encode/decode live in the onpair crate.
- btrblocks: register OnPairScheme alongside FSSTScheme so the sample-based selector keeps the smaller per column; delta-encode the monotonic dict_offsets/codes_offsets children (>= 2048 rows) when it wins.
- vortex-file: register the OnPair encoding and allow it in the write strategy.


Cargo.toml
```diff
@@ -188,6 +190,7 @@ num_enum = { version = "0.7.3", default-features = false }
 object_store = { version = "0.13.1", default-features = false }
 once_cell = "1.21"
 oneshot = { version = "0.2.0", features = ["async"] }
+onpair = { path = "../onpair" }
```
joseph-isaacs: This is noted and will be fixed shortly

joseph-isaacs:
fineweb regression are quite small due to like contains and prefix.

Cargo.toml
```diff
@@ -188,6 +190,7 @@ num_enum = { version = "0.7.3", default-features = false }
 object_store = { version = "0.13.1", default-features = false }
 once_cell = "1.21"
 oneshot = { version = "0.2.0", features = ["async"] }
+onpair = { version = "0.0.3", git = "https://github.com/spiraldb/onpair.git", rev = "53e8ca6081d377e9933d999cef286e26bf52e2c7" }
```
AdamGS, line 193: We can't have git dependencies in the trees of published crates.

encodings/experimental/onpair/public-api.lock
```diff
@@ -0,0 +1,235 @@
+pub mod vortex_onpair
```
AdamGS: 🔥

encodings/fastlanes/src/delta/compute/cast.rs
```diff
@@ -20,18 +20,31 @@ impl CastReduce for Delta {
         };
 
         let source_ptype = array.dtype().as_ptype();
-        // TODO(DK): narrows can be safe but we must decompress to compute the maximum value.
-        if target_ptype.is_signed_int() || source_ptype.bit_width() > target_ptype.bit_width() {
+        // Only a same-width cast (e.g. a nullability change) can be served by
```
AdamGS: some of the cast stuff here should probably be a separate PR

robert3005 (approved):
I think the structure makes sense. We will need fuzzer to run over it and audit compute operations

PR #8147, opened 2026-05-29
Always use `FileStatsLayoutReader` if available for `VortexFile` reader

connortsui20:
## Summary

We use the file statistics when constructing a data source, but we do always do that for the layout reader.

Question: do we want to cache the layout reader in the vortex file instead of constructing it in every other method? (this was why I added a constructor)

## Testing

We'll see what happens in benchmarks.


myrrc (approved):
LGTM. Let's see the benchmarks

connortsui20:
what in the world happened with https://github.com/vortex-data/vortex/pull/8147#issuecomment-4573934805!!!

PR #8148, opened 2026-05-29
try fix file size comment 

joseph-isaacs:


connortsui20:
seems to not work always?

PR #8153, opened 2026-05-29
Introduce basic binary compressor schemes and refactor some other bits

AdamGS:
## Summary

The important piece of this PR is the new `BinaryDictScheme` and `BinaryConstantScheme`, it also includes a bunch of simplifications in the compressor, including moving types into the file that actually includes their functionality, and removes some unnecessary utility functions.


vortex-compressor/src/stats/cache.rs
```diff
@@ -135,13 +135,22 @@ impl ArrayAndStats {
...
     }
 
+    /// Returns the array as an [`ArrayView<VarBinView>`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the array is not a varbinview array.
+    pub fn array_as_utf8(&self) -> ArrayView<'_, VarBinView> {
+        self.array_as_varbinview()
+    }
```
connortsui20: Do we still need this?

vortex-compressor/src/stats/varbinview.rs
connortsui20, line 1: might as well change the module name to varbinview.rs?

connortsui20 (approved):
nice

vortex-compressor/src/stats/cache.rs
AdamGS: good catch

PR #8155, opened 2026-05-29
feat[gpu]:  arrow device array decimal export

0ax1:


vortex-cuda/kernels/src/decimal_cast.cu
```diff
@@ -0,0 +1,76 @@
...
+// Vortex may use narrower decimal storage, so Arrow Device export widens values
+// to match the schema-implied physical layout consumed by cuDF and other Arrow
+// readers.
+// Converts a decimal storage value to Arrow's 128-bit decimal physical representation.
+template <typename Input>
+__device__ __forceinline__ int128_t decimal_to_i128(Input value) {
+    if constexpr (std::is_same_v<Input, int128_t>) {
+        return value;
+    } else if constexpr (std::is_same_v<Input, int256_t>) {
+        return int128_t {value.parts[0], value.parts[1]};
```
onursatici, line 37: can this end up truncating? I think it is currently possible in vortex to do this:

```rust
  let array = DecimalArray::from_iter(
      [i256::from_parts(0, 1)], // 2^128, so does not fit into i128
      DecimalDType::new(38, 0), // this is normally i128
  );
```

when exporting this array we would pick the i256 -> i128 kernel and truncate the values without checking for overflow.

probably the right fix is for vortex to reject constructing such arrays

robert3005 (approved):

vortex-cuda/kernels/src/decimal_cast.cu
0ax1: done

Issue #8119, opened 2026-05-27
Variant write-path features


AdamGS:
This is a tracking issue for more sophisticated write-path features for Variant.

## Motivation

Shredding variant arrays can be done in many ways and is an active research topic.

## Steps

- Convert JSON or similar data into Variant arrays - including shredding https://github.com/vortex-data/vortex/pull/8391

## Unresolved questions

1. How do we define JSON? To we want JSON-specific functions?

## Relevant resources

https://db.in.tum.de/~durner/papers/json-tiles-sigmod21.pdf

## Implementation history

...

PR #8156, opened 2026-05-29
Compress JSON as variant

AdamGS:
## Summary

- Part of https://github.com/vortex-data/vortex/issues/8119

This PR introduces a compression scheme to compress JSON arrays into ParquetVariant arrays.

It includes the following changes:
1. BinaryFSST compression scheme, useful for somewhat structured binary data (like the parquet-variant encoding).
2. `VariantToJson` array, that wraps a variant array, and when executed it turns a Variant array into JSON. This array is feature-flagged behind `parquet-variant/json`.
3. A new compression scheme to turn JSON data into variat - `JsonToVariantScheme`, which is behind a new `parquet-variant` feature.


AdamGS:
I think the publish dry run fails because I use both `vortex-parquet-variant` and `vortex-json`, both are not released yet.

AdamGS:
I think I learned some useful stuff here, but the better API is #8391

PR #8158, opened 2026-05-29
Simplify fsst_compress_offsets_overflow_i32 test with empty compressor

joseph-isaacs:
## Summary

Refactor the `fsst_compress_offsets_overflow_i32` test to use an empty FSST compressor instead of training one on random data. This simplification:

1. **Reduces memory allocation** from ~5 GiB to ~3.4 GiB by eliminating the random data pool and using deterministic escape-based expansion (2x input size) instead of incompressible random data (~1x input size).

2. **Removes randomness** by replacing `StdRng` and random pool generation with a single reused buffer of repeated bytes, making the test deterministic and faster.

3. **Maintains regression coverage** by exercising the same output-offset overflow path. The empty compressor's escape factor (2x + 7 bytes) is the worst-case FSST expansion, so this is the cheapest way to reach the i32::MAX boundary while still crossing it.

4. **Adds explicit assertion** that compressed output actually exceeds i32::MAX, preventing silent test degradation if FSST's escape behavior changes.

The test still allocates ~1.1 GiB of input and ~2.25 GiB of output, so it remains gated to CI and respects `VORTEX_SKIP_SLOW_TESTS`.

## Testing

Existing test infrastructure covers this change. The test is gated to CI runs and validates that the regression (i32 offset overflow in FSST output) is properly exercised by asserting the compressed byte count exceeds i32::MAX.

https://claude.ai/code/session_01212r9Sii7DxuVZ1UJ9oqx1

### DIFF
--- a/encodings/fsst/src/tests.rs
+++ b/encodings/fsst/src/tests.rs
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use rand::SeedableRng;
-use rand::rngs::StdRng;
-use rand::seq::IndexedRandom;
+use fsst::CompressorBuilder;
 use vortex_array::ArrayRef;
 use vortex_array::IntoArray;
 use vortex_array::LEGACY_SESSION;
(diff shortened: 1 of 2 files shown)


encodings/fsst/src/tests.rs
```diff
@@ -118,11 +116,20 @@ fn test_fsst_array_ops() {
 /// [`VarBinBuilder<i32>`] for the FSST output and panicked in
 /// [`VarBinBuilder::append_value`] once cumulative compressed bytes crossed the boundary.
 ///
-/// The input is built with [`VarBinBuilder<i64>`] so the input itself does not panic, which
-/// confirms the overflow is on the FSST output side. After the fix the test must succeed
-/// with the row count preserved.
+/// We force the output past [`i32::MAX`] with an empty FSST compressor: it has no symbols, so
```
AdamGS, line 119: is this test the best documented part of the codebase?

PR #8160, opened 2026-05-29
don't account array deallocation in random access bench

myrrc:


benchmarks/random-access-bench/src/main.rs
```diff
@@ -247,7 +247,7 @@ async fn benchmark_random_access(
 
     loop {
         let start = Instant::now();
-        let _row_count = accessor.take(indices).await?;
+        let _arr = accessor.take(indices).await?;
         runs.push(start.elapsed());
```
joseph-isaacs: shall we explicitly drop the array after `start.elapsed()`

myrrc: +1

joseph-isaacs (approved):

PR #8162, opened 2026-05-29
BitBuffer binary operations support in place updates to lhs

robert3005:
Revive #7375

This speeds up And/Or/Xor - 50% to 4x time
Append - 2x (by shifting the work elsewhere in the code)
append_buffer - 10-50%

### DIFF
--- a/vortex-buffer/src/bit/ops.rs
+++ b/vortex-buffer/src/bit/ops.rs
@@ -7,6 +7,52 @@ use crate::Buffer;
 use crate::ByteBufferMut;
 use crate::trusted_len::TrustedLenExt;
 
+/// Apply a unary operation to a borrowed [`BitBuffer`], always allocating a new buffer.
+///
+/// Use this instead of [`bitwise_unary_op`] when you only have a reference, since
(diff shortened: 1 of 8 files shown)


vortex-buffer/src/bit/ops.rs
```diff
@@ -4,86 +4,106 @@
...
+    map_u64_words_in_place(buf.as_mut_slice(), op);
+    buf.freeze()
+}
+
+/// Apply a unary operation to an owned [`BitBuffer`], mutating in-place when possible.
+///
+/// Tries to get exclusive access via `try_into_mut`. If the backing storage is shared
+/// (Arc refcount > 1), falls back to [`bitwise_unary_op_copy`].
+#[inline]
+pub(super) fn bitwise_unary_op<F: FnMut(u64) -> u64>(buffer: BitBuffer, op: F) -> BitBuffer {
```
robert3005, line 142: This went too far, the bitwise_unary_op_copy should be allocating empty buffer and writing to it

AdamGS (approved):
I think this all makes sense, this sort of code is naturally dense but as far as I can tell its correct and reasonable.

PR #8169, opened 2026-05-31
Aggregating bounded min/max values has to not ignore null values

robert3005:
Propagating bounded min/max is not the same as propagating nulls. In the case of
bounded min/max we have to conclude that null value could have come from trying
to produce the bounded value which means that there's no valid bounded value while there's still valid values in the array

handle #8166

### DIFF
--- a/vortex-file/src/tests.rs
+++ b/vortex-file/src/tests.rs
@@ -1986,3 +1986,100 @@ async fn test_can_prune_composite_predicates() -> VortexResult<()> {
 
     Ok(())
 }
+
+// Regression test for https://github.com/vortex-data/vortex/issues/8166.
+//
+// A binary column whose per-zone max value has an all-0xff prefix (longer than the
+// statistics-truncation length) cannot be upper-bounded, so the max stat is written as NULL.
+// Filtering with `col > literal` must still return the row, but a pruning bug drops it.
+#[tokio::test]
+#[cfg_attr(miri, ignore)]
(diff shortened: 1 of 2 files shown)


gatesn (changes requested):
This seems like a very odd behaviour for truncated min/max to me, just leaving a request changes so I can read thoroughly on Monday.

robert3005:
This comes from the fact that bounded min/max is not transitive if you exclude nulls. The problem we run into is when producing a bounded max returns null because the bound exceeds the limits we have defined which then is not included in the min/max of the file which then can lead to skipped values.

robert3005:
talked to @joseph-isaacs. ~There's a simpler fix, when we overflow on max we store truncated_len + 1 field which would be bigger than any of the existing values~ this doesn't work, we need to increment

joseph-isaacs:
what is the counterexample?

robert3005:
take `upper_bound([FF, FF, FF, 0], 2)` then `[FF, FF, FF]` is not >= `[FF, FF, FF, 0]`

joseph-isaacs:
But `0xFF` > `0xF`?

robert3005:
yes but I don't follow

joseph-isaacs:
Okay we need to store a flag to specify `MAX_VALUE`, such that for `x. x <= MAX_VALUE`

robert3005:
done in #8515

PR #8197, opened 2026-06-01
Document why the Arrow exporter keeps Decimal128 as the default decimal width

joseph-isaacs:
## Summary

Originally this PR enabled emitting narrower Arrow decimals (`Decimal32`/`Decimal64`) by default from `to_data_type_naive`. E2E verification against `datafusion-bench` with **real generated data** showed this is **not safe on the versions Vortex ships**, so the behavior change has been reverted; the net diff is now an explanatory comment documenting *why* the narrower widths stay gated.

## What the verification found

- **arrow-rs 58 / DataFusion 53** both have the `Decimal32`/`Decimal64` types (added in arrow-rs ~v56 and DataFusion 51.0.0), and Vortex's Arrow executor already produces all four widths on request. So the *types* exist.
- But **DataFusion 53 keeps decimal `SUM`/`AVG` within the same physical width family** — `Decimal32 → i32`, `Decimal64 → i64` — instead of widening the accumulator (`datafusion-functions-aggregate` `sum.rs` / `average.rs`). A `Decimal32` `SUM`/`AVG` is capped at precision 9 / `i32`, so it overflows once the running total exceeds ~1e9, regardless of row count.
  - **TPC-DS Q1** (`avg(sum(sr_return_amt))`, `sr_return_amt` = `decimal(7,2)`, inferred schema → `Decimal32`) fails with `Arithmetic Overflow in AvgAccumulator`. Tracked upstream in [apache/datafusion#17489](https://github.com/apache/datafusion/issues/17489).
  - **TPC-H Q1/Q6** pass — but only because TPC-H registers explicit `Decimal128(15,2)` schemas, so its plan never sees the narrow type.
- The **Spark/JNI bindings assume 128-bit decimals** (`SparkToArrowSchema` hard-codes `bitWidth = 128`; the JNI `TestMinimal` writes `Decimal128(9,2)` and asserts an exact round-trip), so any narrowing breaks the round-trip — confirmed by the `TestMinimal.testFullScan` failure.

There is no narrowing tier that's safe across all current consumers: `Decimal32` breaks DataFusion aggregation + JNI; `Decimal64` would break the Spark `decimal(10,2)` round-trip too. The original guard comment was correct.

## Change

Revert to the `Decimal128` default and replace the terse `// commented out until DataFusion improves...` note with a comment recording the concrete reasons (accumulator overflow + 128-bit binding assumption, with the upstream issue link). The Arrow executor still emits `Decimal32`/`Decimal64` when a consumer explicitly requests that target type — only the inferred default stays `Decimal128`.

## Follow-up option

If we want the storage/bandwidth win for narrow decimals without breaking these consumers, the path is to make narrowing **opt-in** (default stays `Decimal128`, enabled via a session/config flag). Happy to do that as a separate change.

## Checks

- `cargo test -p vortex-array --lib dtype::arrow` ✅
- `datafusion-bench tpch --formats parquet,vortex --queries 1,6` (SF=1, real data) ✅ — row counts validated
- Decimal mapping now byte-for-byte equivalent to `develop` (comment-only diff), restoring the green baseline for the TPC-DS bench and JNI jobs.

https://claude.ai/code/session_01RctLpues7aLnsxJ86XH9pC

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,35 +146,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
-name = "apache-avro"
-version = "0.21.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "36fa98bc79671c7981272d91a8753a928ff6a1cd8e4f20a44c45bd5d313840bf"
-dependencies = [
- "bigdecimal",
- "bon",
- "bzip2",
- "crc32fast",
- "digest 0.10.7",
- "liblzma",
- "log",
- "miniz_oxide",
- "num-bigint",
- "quad-rand",
- "rand 0.9.4",
- "regex-lite",
- "serde",
- "serde_bytes",
- "serde_json",
- "snap",
- "strum 0.27.2",
- "strum_macros 0.27.2",
- "thiserror 2.0.18",
- "uuid",
- "zstd",
-]
-
-[[package]]
 name = "approx"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


AdamGS:
this used to break some benchmarks in some annoying ways

robert3005:
@AdamGS do you remember where this broke previously?

robert3005:
TPC-DS
```
thread 'main' (75198) panicked at /home/runner/_work/vortex/vortex/vortex-bench/src/runner.rs:402:37:
Other error: query 1 failed: Execution error: Arithmetic Overflow in AvgAccumulator
```

AdamGS:
I remembered Sum and and Avg, but I'll open a DF issue.

AdamGS:
another fix for now will be to have schemas for tpc-ds/tpc-h, if we declare them wide enough we'll avoid the overflow.

robert3005:
I remember now that arrow-java doesn't support them yet so for jni we would need a way to force only i128/i256

joseph-isaacs:
needs to gate these too if < DF 0.54.0

AdamGS:
DF issue: https://github.com/apache/datafusion/issues/22713
DF PR: https://github.com/apache/datafusion/pull/22714

PR #8209, opened 2026-06-02
Chunk reader children cache

### LINKED ISSUE #7915: Tracking Issue: random access Lance parity

myrrc:
Don't reinitialize ChunkedReader children with their own caches each time.

### DIFF
--- a/vortex-layout/src/reader.rs
+++ b/vortex-layout/src/reader.rs
@@ -2,12 +2,17 @@
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
 use std::any::Any;
+use std::fmt::Debug;
+use std::fmt::Formatter;
+use std::fmt::Result;
 use std::ops::Range;
 use std::sync::Arc;
+use std::sync::Weak;
 
 use futures::future::BoxFuture;
 use futures::try_join;
 use once_cell::sync::OnceCell;
+use parking_lot::Mutex;
 use vortex_array::ArrayRef;
 use vortex_array::IntoArray;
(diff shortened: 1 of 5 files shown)


vortex-layout/src/reader.rs
```diff
@@ -201,26 +205,39 @@ impl ArrayFutureExt for ArrayFuture {
     }
 }
 
+/// Per child layout reader cache
+#[derive(Clone)]
+pub struct LazyReaderChildrenCache(Arc<Vec<OnceCell<LayoutReaderRef>>>);
```
AdamGS: does this mean we keep the readers alive for as long as the `ChunkedReader` is?

myrrc: Yeah, all children are. This makes sense only for repeatable reads.

AdamGS: How common are they outside of this benchmark? The use of locks like `OnceCell` causes the code to serialize over these code paths, and we keep adding all kinds of workarounds

myrrc: I agree with this approach in general, but this cache doesn't show up in the benchmarks and makes requests faster.

joseph-isaacs:
What is the scope of the cache?

myrrc:
Chunk reader lifetime

joseph-isaacs:
Why is the original cache not enough

myrrc:
Obsolete over https://github.com/vortex-data/vortex/pull/8244

PR #8210, opened 2026-06-02
use mimalloc only on MacOS

myrrc:
Vortex being an application plugin shouldn't bring our own allocator.
MacOS has a bad default allocator so on MacOS we still use mimalloc for
local benchmarks


joseph-isaacs (commented):
What about other non-linux os?

AdamGS:
but we don't bring an allocator, its just an external allocator to get faster and more consistent experience. DF also benchmarks with mimalloc.

myrrc:
Regressions, won't do

PR #8211, opened 2026-06-02
Expose row_idx in vortex-jni bindings

robert3005:
These are necessary for supporting iceberg row positions

### DIFF
--- a/java/vortex-jni/src/main/java/dev/vortex/api/Expression.java
+++ b/java/vortex-jni/src/main/java/dev/vortex/api/Expression.java
@@ -41,6 +41,14 @@ public final class Expression {
         return new Expression(NativeExpression.root());
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


java/vortex-jni/src/main/java/dev/vortex/api/Expression.java
```diff
@@ -41,6 +41,14 @@ public static Expression root() {
         return new Expression(NativeExpression.root());
     }
 
+    /**
+     * The row-index expression. When evaluated as part of a Vortex scan it yields the zero-based index of each row
+     * within the file as a non-nullable {@code u64}. It cannot be evaluated outside of a scan.
```
joseph-isaacs: before filtering

joseph-isaacs (approved):

PR #8213, opened 2026-06-02
split registration to respect nested projection and filters

onursatici:
## Summary

split registration only cared about the immediate accessed fields of the projection and filter expression. We do handle these nested expressions correctly at layout readers, but we did still register splits as if we are referencing all nested fields.

so if we had a `select a.x, a.y`, before, we were registering splits for `Prefix(a)`, meaning all fields under a would have their splits registered, ending up with many more tasks than needed. Now we do pass `Prefix(a.y) | Prefix(a.x)` correctly to splits

### DIFF
--- a/vortex-array/src/dtype/field.rs
+++ b/vortex-array/src/dtype/field.rs
@@ -238,47 +238,56 @@ impl Display for FieldPath {
     }
 }
 
-#[derive(Default, Clone, Debug, PartialEq, Eq)]
-/// Contains a set of field paths, and can answer an efficient field path contains queries.
+#[derive(Default, Clone, Debug)]
+/// A set of field paths supporting efficient `contains` queries.
+///
+/// Paths are stored as inserted. Prefix-minimization—collapsing a path into an ancestor that
+/// already covers it—is deferred until the set is iterated via [`IntoIterator`], so insertion stays
+/// cheap.
 pub struct FieldPathSet {
-    /// While this is currently a set wrapper it can be replaced with a trie.
+    /// While this is currently a set wrapper it can be replaced with a trie, at which point the
+    /// deferred minimization in [`IntoIterator`] becomes cheap.
     // TODO(joe): this can be replaced with a `FieldPath` trie
     set: HashSet<FieldPath>,
 }
 
 impl FieldPathSet {
-    /// Checks if a set contains a field path
+    /// Checks if the set contains exactly this field path.
     pub fn contains(&self, path: &FieldPath) -> bool {
         self.set.contains(path)
     }
 
-    /// Iterates over the field paths in the set.
+    /// Iterates over the field paths in the set, as inserted (not prefix-minimized).
     pub fn iter(&self) -> impl Iterator<Item = &FieldPath> {
         self.set.iter()
     }
 
-    /// Inserts a field path prefix unless an existing prefix already covers it.
-    ///
-    /// Any existing field paths covered by the new prefix are removed.
-    pub fn insert_prefix(&mut self, path: FieldPath) {
-        if self
-            .set
+    /// Inserts a field path. Prefix-minimization is deferred until the set is iterated.
+    pub fn insert(&mut self, path: FieldPath) {
+        self.set.insert(path);
+    }
+}
+
+/// Reduces field paths to their minimal covering set: any path that has another path in the set as
+/// a prefix is redundant and dropped.
+fn minimal_covering_set(paths: impl IntoIterator<Item = FieldPath>) -> Vec<FieldPath> {
+    let mut covering: Vec<FieldPath> = Vec::new();
+    for path in paths {
+        if covering
             .iter()
             .any(|existing| path.parts().starts_with(existing.parts()))
         {
-            return;
+            continue;
         }
-
-        self.set
-            .retain(|existing| !existing.parts().starts_with(path.parts()));
-        self.set.insert(path);
+        covering.retain(|existing| !existing.parts().starts_with(path.parts()));
+        covering.push(path);
     }
+    covering
+}
 
-    /// Inserts field path prefixes, retaining only the minimal covering set.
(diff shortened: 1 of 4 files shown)


vortex-array/src/expr/analysis/referenced_field_paths.rs
joseph-isaacs, line 1: can we use the `NodeVisitor` pattern here

vortex-array/src/expr/analysis/referenced_field_paths.rs
```diff
@@ -0,0 +1,172 @@
...
+pub fn referenced_field_paths(expr: &Expression, scope: &DType) -> VortexResult<FieldPathSet> {
+    // Validate the whole expression so plain GetItem paths and Select paths behave consistently.
+    expr.return_dtype(scope)?;
+
+    let mut field_paths = FieldPathSet::default();
+    collect_referenced_field_paths(expr, scope, &FieldPath::root(), &mut field_paths)?;
+    Ok(field_paths)
+}
+
+fn collect_referenced_field_paths(
```
joseph-isaacs: can you add some comments here

vortex-array/src/expr/analysis/referenced_field_paths.rs
```diff
@@ -0,0 +1,172 @@
...
+/// A standalone root expression is represented by [`FieldPath::root`], which conservatively
+/// selects all fields. When one referenced path is a prefix of another, only the prefix is returned.
+/// Scalar functions other than `GetItem` and `Select` conservatively reference each complete child
+/// output.
+pub fn referenced_field_paths(expr: &Expression, scope: &DType) -> VortexResult<FieldPathSet> {
+    // Validate the whole expression so plain GetItem paths and Select paths behave consistently.
+    expr.return_dtype(scope)?;
+
+    let mut field_paths = FieldPathSet::default();
+    collect_referenced_field_paths(expr, scope, &FieldPath::root(), &mut field_paths)?;
```
joseph-isaacs: debug assert that imm access matches the prefix set of this?

vortex-array/src/expr/analysis/referenced_field_paths.rs
```diff
@@ -0,0 +1,172 @@
...
+    Ok(field_paths)
+}
+
+fn collect_referenced_field_paths(
+    expr: &Expression,
+    scope: &DType,
+    requested_path: &FieldPath,
+    field_paths: &mut FieldPathSet,
```
joseph-isaacs: Should we use a Vec and dedup at the end?

onursatici: I want to use the field path set because once we make that a trie this deduping would stop being O(n^2)

vortex-array/src/dtype/field.rs
```diff
@@ -251,6 +251,35 @@ impl FieldPathSet {
...
+
+    /// Iterates over the field paths in the set.
+    pub fn iter(&self) -> impl Iterator<Item = &FieldPath> {
+        self.set.iter()
+    }
+
+    /// Inserts a field path prefix unless an existing prefix already covers it.
+    ///
+    /// Any existing field paths covered by the new prefix are removed.
+    pub fn insert_prefix(&mut self, path: FieldPath) {
```
joseph-isaacs: this is insert then?

vortex-array/src/expr/analysis/referenced_field_paths.rs
joseph-isaacs: Keep the name just use a Vec into and put the dedup in the into iter

vortex-array/src/expr/analysis/referenced_field_paths.rs
```diff
@@ -0,0 +1,257 @@
...
+        // `Select` keeps requested paths whose head is included, expanding a whole-scope request
+        // into one path per included field.
+        if let Some(selection) = node.as_opt::<Select>() {
+            let child_dtype = node.child(0).return_dtype(self.scope)?;
+            let child_fields = child_dtype
+                .as_struct_fields_opt()
+                .ok_or_else(|| vortex_err!("Select child is not a struct"))?;
+            let included_fields = selection.normalize_to_included_fields(child_fields.names())?;
+
+            let mut narrowed = Vec::new();
```
joseph-isaacs: capacity?

vortex-array/src/expr/analysis/referenced_field_paths.rs
```diff
@@ -0,0 +1,257 @@
...
+        // `GetItem` prepends its field to every requested path before descending into its child.
+        if let Some(field_name) = node.as_opt::<GetItem>() {
+            let prefixed = requested
+                .iter()
+                .map(|path| {
+                    FieldPath::from_iter(
+                        once(Field::Name(field_name.clone())).chain(path.parts().iter().cloned()),
+                    )
+                })
+                .collect();
```
joseph-isaacs, line 109: can you append and rev at the end?

vortex-array/src/expr/analysis/referenced_field_paths.rs
```diff
@@ -0,0 +1,264 @@
...
+                    Some(Field::Name(name)) => Some(name.clone()),
+                    _ => None,
+                })
+                .collect()
+        };
+        debug_assert_eq!(
+            referenced_heads,
+            immediate_scope_access(expr, scope_fields),
+            "referenced field path heads must match the immediately accessed scope fields"
+        );
```
joseph-isaacs, line 64: does this need to be in the macro? for dead code checks?

joseph-isaacs (approved):
Good find

vortex-array/src/expr/analysis/referenced_field_paths.rs
onursatici: I wanted to leave the entire block out of release builds as the debug_assert won't do anything, and it marks it clear that this is only for the debug assertion

PR #8214, opened 2026-06-02
Rename ArrayEq/Hash Precision to `EqMode`

AdamGS:
## Summary

This PR renames the `Precision` enum that we use to do array hashing and equality checks to `Accuracy`, as we already have a `Precision` enum we use for stats, that is also part of the binary format.

## API Changes

Just the renamed enum, which should be relatively minor.

## Testing

N/A

### DIFF
--- a/encodings/alp/src/alp/array.rs
+++ b/encodings/alp/src/alp/array.rs
@@ -8,7 +8,6 @@ use std::hash::Hash;
 use std::hash::Hasher;
 
 use prost::Message;
-use vortex_array::Accuracy;
 use vortex_array::Array;
 use vortex_array::ArrayEq;
 use vortex_array::ArrayHash;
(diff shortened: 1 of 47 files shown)


robert3005 (approved):

joseph-isaacs (changes requested):
accuracy is the wrong word

joseph-isaacs:
exactness?

AdamGS:
how about splitting it into two functions on the same trait? the enum seems confusing

a10y:
should we just call this `EqMode` or something?

encodings/alp/src/alp/array.rs
```diff
@@ -55,14 +55,14 @@ use crate::alp::rules::RULES;
 pub type ALPArray = Array<ALP>;
 
 impl ArrayHash for ALPData {
-    fn array_hash<H: Hasher>(&self, state: &mut H, _precision: Precision) {
+    fn array_hash<H: Hasher>(&self, state: &mut H, _accuracy: EqMode) {
```
joseph-isaacs, line 58: should call these mode?

joseph-isaacs (approved):

encodings/alp/src/alp/array.rs
AdamGS: there are like 10 million of them, I changed the trait so future impls will get filled in by IDEs

joseph-isaacs: It is a simple find and replace

PR #8217, opened 2026-06-02
fix session deadlock

onursatici:
## Summary

`MemorySession` is registered lazily. Because it implements default we don't need to register it to the session and the first call to `ctx.allocator()` would hold the session dashmap shard lock and insert the allocator. The problem is when we are in a aggregate function we are already potentially holding the same dashmap shard's lock, because the aggregate functino registry is also read from the session.

because session vars are type id keyed and type id's change on every build, if the aggregate function registry and memory session end up being on the same dashmap shard, we deadlock running an aggregate function.

This is fixing with a band aid, I am eagerly registering memory session because that is the only one that we forgot to register, hence we won't ever write lock the session after initialising now. I think the right solution is to not hold the read lock of the dashmap shard while executing kernels, which I will do in a follow up

### DIFF
--- a/vortex/src/lib.rs
+++ b/vortex/src/lib.rs
@@ -12,6 +12,7 @@ use vortex_array::dtype::session::DTypeSession;
 // vortex::expr is in the process of having its dependencies inverted, and will eventually be
 // pulled back out into a vortex_expr crate.
 pub use vortex_array::expr;
+use vortex_array::memory::MemorySession;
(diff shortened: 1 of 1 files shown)


AdamGS:
I think we should have a baseline structure that isn't dashmap-dependent, instead just a struct with normal children, but it can follow up this change.

robert3005 (approved):

robert3005:
fwiw I merged a change couple mins ago that makes the aggregate session registry not hold the lock

PR #8218, opened 2026-06-02
Implement ZipKernel for ListViewArray

robert3005:
This can be very cheap with ListViews and ends up being very sad because we use
a builder by default

### DIFF
--- a/vortex-array/src/arrays/listview/compute/zip.rs
+++ b/vortex-array/src/arrays/listview/compute/zip.rs
@@ -57,15 +57,15 @@ impl ZipKernel for ListView {
 
         let len = if_true.len();
 
+        let result_elements_dtype = if_true
+            .elements()
+            .dtype()
+            .union_nullability(if_false.elements().dtype().nullability());
+
         // `if_false`'s elements share the element dtype up to nullability; normalize so both chunks
         // of the concatenated elements array have an identical dtype.
-        let true_elements = if_true.elements().clone();
-        let element_dtype = true_elements.dtype().clone();
-        let false_elements = if if_false.elements().dtype() == &element_dtype {
-            if_false.elements().clone()
-        } else {
-            if_false.elements().cast(element_dtype.clone())?
-        };
+        let true_elements = if_true.elements().cast(result_elements_dtype.clone())?;
+        let false_elements = if_false.elements().cast(result_elements_dtype.clone())?;
 
         // `if_false` views index into the second half of the concatenated elements.
         let false_shift = true_elements.len() as u64;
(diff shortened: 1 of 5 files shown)


vortex-array/src/arrays/listview/compute/zip.rs
```diff
@@ -0,0 +1,367 @@
...
+        let len = if_true.len();
+
+        // `if_false`'s elements share the element dtype up to nullability; normalize so both chunks
+        // of the concatenated elements array have an identical dtype.
+        let true_elements = if_true.elements().clone();
+        let element_dtype = true_elements.dtype().clone();
+        let false_elements = if if_false.elements().dtype() == &element_dtype {
+            if_false.elements().clone()
+        } else {
+            if_false.elements().cast(element_dtype.clone())?
```
onursatici: shouldn't we union the nullability here? If `if_true` is non nullable and `if_false` is nullable, we can't coerce `if_false` to non nullable

vortex-array/src/arrays/listview/compute/zip.rs
```diff
@@ -0,0 +1,367 @@
...
+        };
+
+        // `if_false` views index into the second half of the concatenated elements.
+        let false_shift = true_elements.len() as u64;
+
+        // Concatenate the two `elements` arrays without copying. If either side is already a
+        // `ChunkedArray` (e.g. the result of a previous list-view zip), splice its chunks in
+        // directly rather than nesting chunked arrays.
+        let mut chunks = Vec::new();
```
joseph-isaacs: capacity?

vortex-array/src/arrays/listview/compute/zip.rs
joseph-isaacs: cast shortcuts itself

vortex-array/src/arrays/listview/compute/zip.rs
```diff
@@ -0,0 +1,367 @@
...
+        let false_offsets = to_u64(if_false.offsets(), ctx)?;
+        let false_sizes = to_u64(if_false.sizes(), ctx)?;
+
+        let mut offsets = BufferMut::<u64>::with_capacity(len);
+        let mut sizes = BufferMut::<u64>::with_capacity(len);
+        for i in 0..len {
+            // SAFETY: the loop runs exactly `len` times and both buffers reserved `len`.
+            unsafe {
+                if mask.value(i) {
+                    offsets.push_unchecked(true_offsets[i]);
```
joseph-isaacs: i think a mut iterator for the output might be faster?

vortex-array/src/arrays/listview/compute/zip.rs
```diff
@@ -0,0 +1,367 @@
...
+        .execute::<Buffer<u64>>(ctx)
+}
+
+/// Combine the two list-level validities, taking `if_true`'s validity where `mask` is set and
+/// `if_false`'s where it is not.
+fn zip_validity(
+    if_true: Validity,
+    if_false: Validity,
+    mask: &Mask,
+    nullability: Nullability,
```
joseph-isaacs: validity gives you this?

danking (approved):
I am convinced that Rob's [latest commit](https://github.com/vortex-data/vortex/pull/8218/changes/0ef75690a1add53823c51bfc390b631bc52fd6f5) addresses all four review comments.

PR #8222, opened 2026-06-02
Share FSST symbol table state between executed variants of arrays.

AdamGS:
## Summary

This PR has two changes:
1. Bumps the FSST dependency to include 0.5.11 which has faster `Compressor::rebuild_from`.
2. When executing an FSST array through slice/filter/take, we keep the compressor instead of creating new lazy instances, as part of a `FSSTSymbolTable` type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3272,9 +3272,9 @@ dependencies = [
 
 [[package]]
 name = "fsst-rs"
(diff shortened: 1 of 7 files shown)


joseph-isaacs (approved):
shall we make a symbols struct?

PR #8225, opened 2026-06-02
Add `select_unpredictable` hint to search_sorted

AdamGS:
## Summary

Seems to be the main difference between the impl we have and std, docs - https://doc.rust-lang.org/std/hint/fn.select_unpredictable.html

Doesn't seem to make a difference in our benchmarks, probably worth further investigation.


robert3005:
those hints used to be not available outside of std, glad this changed?

robert3005 (approved):

PR #8228, opened 2026-06-02
Faster take for runend array

robert3005:
Maybe this is faster?

### DIFF
--- a/encodings/runend/src/compute/take.rs
+++ b/encodings/runend/src/compute/take.rs
@@ -7,85 +7,479 @@ use vortex_array::ArrayRef;
 use vortex_array::ArrayView;
 use vortex_array::ExecutionCtx;
 use vortex_array::IntoArray;
+use vortex_array::arrays::ConstantArray;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::arrays::dict::TakeExecute;
+use vortex_array::dtype::UnsignedPType;
 use vortex_array::match_each_integer_ptype;
-use vortex_array::search_sorted::SearchResult;
-use vortex_array::search_sorted::SearchSorted;
-use vortex_array::search_sorted::SearchSortedSide;
+use vortex_array::match_each_unsigned_integer_ptype;
+use vortex_array::scalar::Scalar;
 use vortex_array::validity::Validity;
 use vortex_buffer::Buffer;
+use vortex_buffer::BufferMut;
 use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
+use vortex_mask::AllOr;
+use vortex_mask::Mask;
 
 use crate::RunEnd;
 use crate::array::RunEndArrayExt;
+use crate::iter::trimmed_ends_iter;
+
+const SORTED_LINEAR_RUNS_PER_INDEX_THRESHOLD: usize = 16;
+const UNSORTED_LINEAR_RUNS_PER_INDEX_THRESHOLD: usize = 4;
+/// Sorting the indices and merging only beats per-index binary search once the run ends are too
+/// large to stay cache-resident; below this run count binary search wins.
+const UNSORTED_LINEAR_MIN_RUNS: usize = 1 << 19;
+/// Use a dense logical-position-to-run-index table when the array length is at most this many
+/// times the number of valid indices: building the table is O(array_len) and each index then
+/// resolves with a single unconditional gather.
+const TABLE_LEN_PER_INDEX_THRESHOLD: usize = 8;
 
 impl TakeExecute for RunEnd {
-    #[expect(
-        clippy::cast_possible_truncation,
-        reason = "index cast to usize inside macro"
-    )]
     fn take(
         array: ArrayView<'_, Self>,
         indices: &ArrayRef,
         ctx: &mut ExecutionCtx,
     ) -> VortexResult<Option<ArrayRef>> {
         let primitive_indices = indices.clone().execute::<PrimitiveArray>(ctx)?;
+        let indices_validity = primitive_indices.validity()?;
+        let indices_mask = indices_validity.execute_mask(primitive_indices.len(), ctx)?;
 
-        let checked_indices = match_each_integer_ptype!(primitive_indices.ptype(), |P| {
-            primitive_indices
-                .as_slice::<P>()
-                .iter()
-                .copied()
-                .map(|idx| {
-                    let usize_idx = idx as usize;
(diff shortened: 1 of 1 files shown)


encodings/runend/src/compute/take.rs
```diff
@@ -32,13 +36,20 @@ impl TakeExecute for RunEnd {
...
 
         let checked_indices = match_each_integer_ptype!(primitive_indices.ptype(), |P| {
             primitive_indices
                 .as_slice::<P>()
                 .iter()
                 .copied()
-                .map(|idx| {
+                .enumerate()
+                .map(|(idx_pos, idx)| {
+                    if !indices_mask.value(idx_pos) {
```
joseph-isaacs: that looks very slow?

encodings/runend/src/compute/take.rs
```diff
@@ -9,83 +9,451 @@ use vortex_array::ExecutionCtx;
...
 
-    // TODO(joe): use the validity mask to skip search sorted.
-    let physical_indices = match_each_integer_ptype!(ends.ptype(), |I| {
-        let end_slices = ends.as_slice::<I>();
-        let physical_indices_vec: Vec<u64> = indices
+    let physical_indices = match_each_unsigned_integer_ptype!(ends.ptype(), |I| {
+        let ends = ends.as_slice::<I>();
+        // Run indices fit in u32 for any realistic array; the narrower physical indices halve
+        // the memory traffic of the downstream take on the values.
+        if u32::try_from(ends.len()).is_ok() {
```
robert3005: this should compare to u32::MAX

encodings/runend/src/compute/take.rs
```diff
@@ -9,83 +9,451 @@ use vortex_array::ExecutionCtx;
...
+            ends.len(),
+            stats.count,
+            SORTED_LINEAR_RUNS_PER_INDEX_THRESHOLD,
+        )
+    {
+        return physical_indices_linear_sorted(ends, offset, indices, validity_mask);
+    }
+
+    // Sorted indices never reach this point when the condition holds: runs <= array_len <=
+    // 8 * count implies the sorted linear scan above was taken.
```
robert3005: this comment is wrong

encodings/runend/src/compute/take.rs
```diff
@@ -9,83 +9,451 @@ use vortex_array::ExecutionCtx;
...
+) -> VortexResult<bool> {
+    let mut previous_idx = None;
+    let mut sorted = true;
+
+    for idx in indices {
+        let idx = idx.as_();
+        if check_bounds {
+            check_index(idx, array_len)?;
+        }
+        if previous_idx.is_some_and(|previous_idx| previous_idx > idx) {
```
robert3005: can we use >= here?

robert3005: no, this is inverted comparison

robert3005: we should avoid the option here and start this count from 1 and explicitly handle 0 and 1 length indices

encodings/runend/src/compute/take.rs
```diff
@@ -9,83 +9,451 @@ use vortex_array::ExecutionCtx;
...
+    Ok(sorted)
+}
+
+fn valid_indices_sorted_masked<T: AsPrimitive<usize>>(
+    indices: &[T],
+    is_valid: impl Iterator<Item = bool>,
+    array_len: usize,
+    check_bounds: bool,
+) -> VortexResult<bool> {
+    let mut previous_idx = None;
```
robert3005: same as the all valid version

encodings/runend/src/compute/take.rs
```diff
@@ -9,83 +9,451 @@ use vortex_array::ExecutionCtx;
...
+    index.min(ends.len() - 1)
+}
+
+fn advance_run<I: UnsignedPType>(ends: &[I], run_idx: &mut usize, logical_idx: usize) {
+    while *run_idx + 1 < ends.len() && run_end_le_logical_idx(ends[*run_idx], logical_idx) {
+        *run_idx += 1;
+    }
+}
+
+fn run_end_le_logical_idx<I: UnsignedPType>(run_end: I, logical_idx: usize) -> bool {
```
robert3005: this should be inlined

encodings/runend/src/compute/take.rs
```diff
@@ -9,83 +9,469 @@ use vortex_array::ExecutionCtx;
...
+    indices: &[T],
+    validity_mask: &Mask,
+) -> Buffer<O>
+where
+    I: UnsignedPType,
+    T: AsPrimitive<usize>,
+    O: UnsignedPType,
+    usize: AsPrimitive<O>,
+{
+    let table = run_index_table::<I, O>(ends, offset, array_len);
```
robert3005, line 356: we should avoid the pathological case here

encodings/runend/src/compute/take.rs
```diff
@@ -7,85 +7,480 @@ use vortex_array::ArrayRef;
...
+
+const SORTED_LINEAR_RUNS_PER_INDEX_THRESHOLD: usize = 16;
+const UNSORTED_LINEAR_RUNS_PER_INDEX_THRESHOLD: usize = 4;
+/// Sorting the indices and merging only beats per-index binary search once the run ends are too
+/// large to stay cache-resident; below this run count binary search wins.
+const UNSORTED_LINEAR_MIN_RUNS: usize = 1 << 19;
+/// Use a dense logical-position-to-run-index table when the array length is at most this many
+/// times the number of valid indices: building the table is O(array_len) and each index then
+/// resolves with a single unconditional gather.
+const TABLE_LEN_PER_INDEX_THRESHOLD: usize = 8;
```
onursatici, line 37: how did we get these? I don't think we have micro-benches for take on runend, I wonder if we should have some in case we need to experiment different values here

onursatici (commented):
this looks good to me, did you ever get your answer for if this is any faster?

encodings/runend/src/compute/take.rs
robert3005: I realised I have added benchmarks separately to make this pr but never pr'ed them. This was result of those benchmarks. I will add benchmarks first. We can also drop the linear path optimisation as that only triggers at 500k elements

onursatici: yea the unsorted linear path is rare and complex but I think it is neat and looks correct so we can keep it

gatesn (approved):

PR #8230, opened 2026-06-03
statpopgen: store GT genotypes as u8 (schema + builder + regression test)

joseph-isaacs:
## What

Stores the `statpopgen` `GT` (genotype) field as `list(u8)` instead of `list(u64)`. GT dosage is only ever `NULL`, `0`, `1`, or `2`, so `u8` is the natural width and 8× narrower than `u64`.


0ax1:
ping me again once the benchmarks pass

joseph-isaacs:
They did all pass

0ax1 (approved):

PR #8231, opened 2026-06-03
Don't recalculate chunk offsets for ChunkedReader

myrrc:
Reuse them from parent layout.

### DIFF
--- a/vortex-layout/src/layouts/chunked/mod.rs
+++ b/vortex-layout/src/layouts/chunked/mod.rs
@@ -103,10 +103,9 @@ impl VTable for Chunked {
         let new_children = OwnedLayoutChildren::layout_children(children);
 
         // Recalculate chunk offsets based on new children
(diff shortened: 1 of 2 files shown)


vortex-layout/src/layouts/chunked/mod.rs
```diff
@@ -135,12 +131,11 @@ pub struct ChunkedLayout {
...
-
-        chunk_offsets.push(0);
+        let mut chunk_offsets = vec![0; children.nchildren() + 1];
         for i in 0..children.nchildren() {
-            chunk_offsets.push(chunk_offsets[i] + children.child_row_count(i));
+            chunk_offsets[i + 1] = chunk_offsets[i] + children.child_row_count(i);
         }
+
         assert_eq!(
             chunk_offsets[children.nchildren()],
```
joseph-isaacs, line 140: revert or use mut iter if you want perf

joseph-isaacs (approved):

PR #8234, opened 2026-06-03
ViewedLayoutChildren child layout cache

### LINKED ISSUE #7915: Tracking Issue: random access Lance parity

myrrc:
Avoid re-parsing flatbuffers when you re-access the same children

### DIFF
--- a/vortex-layout/src/children.rs
+++ b/vortex-layout/src/children.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 
 use flatbuffers::Follow;
 use itertools::Itertools;
+use once_cell::sync::OnceCell;
 use vortex_array::dtype::DType;
 use vortex_error::VortexResult;
 use vortex_error::vortex_bail;
@@ -105,6 +106,7 @@ pub(crate) struct ViewedLayoutChildren {
     layout_read_ctx: ReadContext,
     layouts: LayoutRegistry,
     allow_unknown: bool,
+    cache: Arc<[OnceCell<LayoutRef>]>,
 }
 
 impl ViewedLayoutChildren {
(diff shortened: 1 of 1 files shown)


vortex-layout/src/children.rs
joseph-isaacs: update this!

vortex-layout/src/children.rs
```diff
@@ -181,50 +192,58 @@ impl LayoutChildren for ViewedLayoutChildren {
     }
 
     fn child(&self, idx: usize, dtype: &DType) -> VortexResult<LayoutRef> {
-        if idx >= self.nchildren() {
-            vortex_bail!("Child index out of bounds: {} of {}", idx, self.nchildren());
+        if idx >= self.cache.len() {
+            vortex_bail!("Child index out of bounds: {} of {}", idx, self.cache.len());
```
joseph-isaacs: revert this

vortex-layout/src/children.rs
```diff
@@ -121,13 +123,22 @@ impl ViewedLayoutChildren {
         layouts: LayoutRegistry,
         allow_unknown: bool,
     ) -> Self {
+        // SAFETY: guaranteed by caller
+        let nchildren = unsafe {
+            fbl::Layout::follow(flatbuffer.as_ref(), flatbuffer_loc)
+                .children()
+                .unwrap_or_default()
+                .len()
+        };
```
joseph-isaacs:
```suggestion
        let nchildren = unsafe {
            fbl::Layout::follow(flatbuffer.as_ref(), flatbuffer_loc)
              
        }  .children()
                .unwrap_or_default()
                .len();
```

joseph-isaacs (approved):

PR #8235, opened 2026-06-03
Initial impl for sort pushdown in DataFusion FileSource implementation

AdamGS:
## Summary

Naive implementation of `try_pushdown_sort`, its mostly just parts of the parquet impl + making sure we propagate the information into the Vortex scan.

### DIFF
--- a/vortex-datafusion/src/persistent/source.rs
+++ b/vortex-datafusion/src/persistent/source.rs
@@ -13,13 +13,17 @@ use datafusion_datasource::file::FileSource;
 use datafusion_datasource::file_scan_config::FileScanConfig;
 use datafusion_datasource::file_stream::FileOpener;
 use datafusion_execution::cache::cache_manager::FileMetadataCache;
+use datafusion_physical_expr::EquivalenceProperties;
+use datafusion_physical_expr::LexOrdering;
 use datafusion_physical_expr::PhysicalExprRef;
+use datafusion_physical_expr::PhysicalSortExpr;
 use datafusion_physical_expr::conjunction;
 use datafusion_physical_expr::projection::ProjectionExprs;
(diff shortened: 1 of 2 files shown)


AdamGS:
going to add tests later today

joseph-isaacs:
Do we have any benchmarks that care about sortness

vortex-datafusion/src/persistent/source.rs
```diff
@@ -383,6 +389,64 @@ impl FileSource for VortexSource {
...
+        }
+
+        let is_descending = sort_order
+            .as_ref()
+            .is_some_and(|s| s.first().options.descending);
+
+        if !is_descending {
+            let mut this = self.clone();
+            this.ordered = true;
+            return Ok(SortOrderPushdownResult::Inexact {
```
onursatici: is this a bit optimistic? I don't know how datafusion sort operators treat this but would we want to fallback to a near sorted optimised sort strategy always if we have an ascending sort of a column that exists in the file?

AdamGS: ordered here is just "the order of the file", instead of returning batches in whichever order we get them.

AdamGS:
@joseph-isaacs we do (some of the top-k style queries), but I'm not sure we have a benchmarks with a sorted table in the way DF can understand.

vortex-datafusion/src/persistent/source.rs
onursatici: ah I mean returning `Inexact` instead of `Unknown` here because we only know that the column exists at this point, and not much about how it is sorted

AdamGS: you're right, I've simplified it for now.

onursatici (approved):

PR #8239, opened 2026-06-03
perf(fastlanes): fuse bit-packed compare into a transposed mask + untranspose

joseph-isaacs:
## Summary

Replaces the unpack-then-compare streaming kernel for compare-against-constant with the FastLanes fused `unpack_cmp`:

- compare each value **as it is unpacked**, accumulating results straight into a transposed 1024-bit mask (`[u64; 16]`, one register-resident word per lane — no `[bool; 1024]`/`[T; 1024]` scratch),
- a single SIMD `untranspose_bits` per block rotates the mask into logical row order, copied directly into the output bit buffer,
- inline patches are spliced in afterwards; sliced (`offset != 0`) arrays fall back to the scalar streaming predicate.


Cargo.toml
```diff
@@ -412,3 +412,8 @@ debug = false
 debug-assertions = false
 strip = "debuginfo"
 incremental = false
+
+# Pin to the in-development FastLanes branch (PR #141 fused [u64;16] cmp mask +
+# PR #145 width-generic BMI/VBMI untranspose) until a release is cut.
+[patch.crates-io]
+fastlanes = { git = "https://github.com/spiraldb/fastlanes", rev = "6c10ea72cf693a17e994aa6401604ebedbeda453" }
```
joseph-isaacs: We will remove this before we merge this PR

encodings/fastlanes/src/bitpacking/compute/compare_fused.rs
```diff
@@ -0,0 +1,146 @@
...
+        return stream_predicate::<T, _>(array, nullability, move |v| cmp(v, rhs), ctx);
+    }
+
+    // Over-allocate to whole 1024-bit blocks in padded coordinates so every block - including the
+    // trailing partial - has room for a full untranspose at a 64-bit-word-aligned offset.
+    let num_chunks = (offset + len).div_ceil(CHUNK_SIZE);
+    let mut words: BufferMut<u64> = BufferMut::zeroed(num_chunks * WORDS_PER_CHUNK);
+
+    let chunks = array.unpacked_chunks::<T>()?;
+    {
```
robert3005, line 91: why is this a block?

encodings/fastlanes/src/bitpacking/compute/compare_fused.rs
```diff
@@ -0,0 +1,146 @@
...
+    let bits = BitBufferMut::from_buffer(words.into_byte_buffer(), offset, len).freeze();
+    let validity = array.validity()?.union_nullability(nullability);
+    Ok(BoolArray::new(bits, validity).into_array())
+}
+
+/// Branchlessly write a single bit in a packed `u64` word buffer: clear the bit, then OR in the
+/// new value. Avoids a data-dependent branch per patch in the patch-fixup loop, and touches the
+/// target word through a single bounds-checked `&mut`.
+#[inline]
+fn set_bit(words: &mut [u64], idx: usize, value: bool) {
```
robert3005: we should have set bit already?

robert3005 (approved):

PR #8242, opened 2026-06-03
Make `from_arrow` not panic

connortsui20:
## Summary

Removes the assertion in the `nulls` function so that `from_arrow` errors instead of panicking if the `nullable` does not match the actually nullability of the arrow array.

Also documents the `FromArrowArray` trait and method.

## Testing

2 basic unit tests.

### DIFF
--- a/vortex-array/src/arrow/convert.rs
+++ b/vortex-array/src/arrow/convert.rs
@@ -57,14 +57,15 @@ use arrow_buffer::buffer::NullBuffer;
 use arrow_buffer::buffer::OffsetBuffer;
 use arrow_schema::DataType;
 use arrow_schema::TimeUnit as ArrowTimeUnit;
-use itertools::Itertools;
 use vortex_buffer::Alignment;
 use vortex_buffer::BitBuffer;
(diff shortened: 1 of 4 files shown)


vortex-array/src/arrow/convert.rs
```diff
@@ -476,20 +483,25 @@ impl<K: ArrowDictionaryKeyType> FromArrowArray<&DictionaryArray<K>> for DictArra
...
     } else {
-        assert!(nulls.map(|x| x.null_count() == 0).unwrap_or(true));
-        Validity::NonNullable
+        let null_count = nulls.map(NullBuffer::null_count).unwrap_or(0);
+        vortex_ensure_eq!(
+            null_count,
+            0,
+            "Cannot convert an Arrow array containing {null_count} nulls into a non-nullable Vortex array"
+        );
+        Ok(Validity::NonNullable)
```
connortsui20, line 504: This was the main reason for this change, it doesn't make much sense to assert when basically all callers return results

robert3005 (approved):

PR #8244, opened 2026-06-04
Allow caching LayoutReader in VortexFile

### LINKED ISSUE #7915: Tracking Issue: random access Lance parity

myrrc:
Not-reinitializing LayoutReader saves us some time.
As a related change, make root() static throughout the program. Before ScanBuilder::new() created a new root()
which didn't compare with the old one pointer-eq-wise which was an issue

### DIFF
--- a/vortex-file/src/file.rs
+++ b/vortex-file/src/file.rs
@@ -8,6 +8,7 @@
 
 use std::ops::Range;
 use std::sync::Arc;
+use std::sync::OnceLock;
 
 use itertools::Itertools;
 use vortex_array::ArrayRef;
@@ -51,6 +52,29 @@ pub struct VortexFile {
     segment_source: Arc<dyn SegmentSource>,
     /// The Vortex session used to open this file.
     session: VortexSession,
+    /// None id LayoutReader caching is turned off
+    layout_reader_cache: Option<OnceLock<Arc<dyn LayoutReader>>>,
+}
+
+fn layout_reader(
+    segment_source: Arc<dyn SegmentSource>,
+    footer: &Footer,
+    session: &VortexSession,
+) -> VortexResult<Arc<dyn LayoutReader>> {
+    let root_reader = footer
+        .layout()
+        // TODO(ngates): we may want to allow the user pass in a name here?
+        .new_reader("".into(), segment_source, session, &Default::default())?;
+
+    Ok(if let Some(stats) = footer.statistics().cloned() {
+        Arc::new(FileStatsLayoutReader::new(
+            root_reader,
+            stats,
+            session.clone(),
+        ))
+    } else {
+        root_reader
+    })
 }
 
 impl VortexFile {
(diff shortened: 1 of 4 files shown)


vortex-file/src/file.rs
```diff
@@ -59,12 +82,25 @@ impl VortexFile {
         footer: Footer,
         segment_source: Arc<dyn SegmentSource>,
         session: VortexSession,
-    ) -> Self {
-        Self {
+        cache_layout_reader: bool,
+    ) -> VortexResult<Self> {
+        // false diagnostic which leads to compilation error
+        #[expect(clippy::if_then_some_else_none)]
+        let layout_reader_cache = if cache_layout_reader {
```
robert3005: I think it wants you to do `cache_layout_reader.then(|| ...)`

vortex-file/src/file.rs
```diff
@@ -59,12 +82,25 @@ impl VortexFile {
         footer: Footer,
         segment_source: Arc<dyn SegmentSource>,
         session: VortexSession,
-    ) -> Self {
-        Self {
+        cache_layout_reader: bool,
```
robert3005: I don't think you want a bool like this. What I would do instead is have a OnceLock with the layoutreader and take it from there. We slightly increase the size but we don't have to pay the cost of creating layout reader on open.

robert3005 (dismissed):
I think you want to make it a oncelock on VortexFile

vortex-file/src/file.rs
myrrc: Yep, which fails because closure returns a result and .then expects an option

vortex-file/src/file.rs
myrrc: But we want it optional

vortex-file/src/file.rs
```diff
@@ -59,12 +82,25 @@ impl VortexFile {
...
+    ) -> VortexResult<Self> {
+        // false diagnostic which leads to compilation error
+        #[expect(clippy::if_then_some_else_none)]
```
joseph-isaacs: accept clippy

vortex-file/src/file.rs
joseph-isaacs: this is a break we don't need to do this!, add a new method

joseph-isaacs: Don't create the layout reader here

vortex-file/src/file.rs
```diff
@@ -59,12 +83,27 @@ impl VortexFile {
...
+        footer: Footer,
+        segment_source: Arc<dyn SegmentSource>,
+        session: VortexSession,
+    ) -> VortexResult<Self> {
+        Ok(Self {
+            footer,
+            segment_source,
+            session,
+            layout_reader_cache: Some(OnceLock::new()),
+        })
```
joseph-isaacs: new(..).with_caching()?

myrrc: +1

vortex-file/src/file.rs
```diff
@@ -145,10 +159,27 @@ impl VortexFile {
 
     /// Initiate a scan of the file, returning a builder for configuring the scan.
     pub fn scan(&self) -> VortexResult<ScanBuilder<ArrayRef>> {
-        Ok(ScanBuilder::new(
-            self.session.clone(),
-            self.layout_reader()?,
-        ))
+        let reader = match &self.layout_reader_cache {
```
joseph-isaacs: this should be in self.layout_reader!

myrrc: +

joseph-isaacs (approved):

PR #8245, opened 2026-06-04
case_when with is_null -> fill_null

onursatici:
## Summary

zip kernels are generally slower than fill_null kernels, and `CASE WHEN is_null(x) THEN c ELSE x END` needs to resolve x twice, whereas `fill_null(x, c)` does resolve x once


vortex-array/src/scalar_fn/fns/case_when.rs
```diff
@@ -251,6 +255,49 @@ impl ScalarFnVTable for CaseWhen {
...
+            (then, els)
+        } else {
+            return Ok(None);
+        };
+
+        match fill.as_opt::<Literal>() {
+            Some(scalar) if !scalar.is_null() => {
+                Ok(Some(crate::expr::fill_null(x.clone(), fill.clone())))
+            }
+            _ => Ok(None),
```
joseph-isaacs: return x

robert3005 (approved):

PR #8248, opened 2026-06-04
fix issue with FileSystem path handling

### LINKED ISSUE #6599: FileSystem::glob with non-glob should verify file exists

robert3005:
fix #6599

this is a revival of #7516 that needed some extra love

In this pr we make sure to handle all the edge cases of paths that work differently locally vs over http. This is in addition to #7516 which fixed glob handling of non existent files

### DIFF
--- a/vortex-duckdb/src/filesystem.rs
+++ b/vortex-duckdb/src/filesystem.rs
@@ -147,6 +147,22 @@ impl FileSystem for DuckDbFileSystem {
         .boxed()
     }
 
+    async fn head(&self, path: &str) -> VortexResult<Option<FileListing>> {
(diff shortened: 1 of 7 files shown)


joseph-isaacs:
should be also handle esc and `%` codes

robert3005:
I have made the coverage more expansive. It boils down to using `Path::parse` instead of `Path::from`

a10y (approved):

PR #8253, opened 2026-06-04
feat(vortex-row): row-oriented byte encoder (size + encode passes)

joseph-isaacs:
## Summary

Adds `vortex-row`, a new crate that encodes one or more columnar Vortex arrays into a single
`ListView<u8>` whose per-row byte slices are lexicographically comparable. The byte order
matches tuple ordering of the input values under per-column sort options, so the output works
directly as a sort key / row key — the Vortex analogue of `arrow-row`.

This PR includes the base row-encoding API, the two scalar-function passes, the byte codec,
focused tests, a `row_encode` benchmark, and the row byte-layout documentation. The crate is
marked `publish = false`, so no `public-api.lock` is tracked while the API is still settling.

FSST-specific row-encoding experiments are intentionally not included in this PR.

## Design

Encoding runs as two scalar functions wired behind the `RowEncoder` API:

1. **Size pass — `RowSize`.** Walks the N input columns once, classifies each column as
   fixed- or variable-width, accumulates the fixed-width prefix per row, and lazily collects
   per-row variable lengths. Returns `Struct { fixed: u32, var: u32 }` so callers read per-row
   widths without materializing the constant `fixed` slot as a per-row buffer.
2. **Encode pass — `RowEncode`.** Uses those sizes to compute totals, allocate one contiguous
   elements buffer, build per-row absolute offsets, then writes each column left-to-right into
   its per-row slot via a write cursor that doubles as the ListView `sizes` array, so no
   separate finalize step is needed.

The converter is effectively 2 passes for the pure-fixed-width case and 3 when
variable-length columns require the prefix-sum offsets pass.

Per-column ordering is controlled by `RowSortField { descending, nulls_first }`: descending
reverses the encoded value bytes, and leading sentinel bytes place nulls before or after
non-nulls independently of sort direction.

## API Layout

`convert_columns(cols: &[ArrayRef], fields: &[RowSortField], ctx) -> VortexResult<ListViewArray>`
is the one-shot entry point; `RowEncoder` is the reusable form.

| Item | File |
|---|---|
| `RowEncoder`, `convert_columns(_with_options)`, `compute_row_sizes(_with_options)` | `src/encoder.rs` |
| `RowEncode` scalar fn + encode driver | `src/encode.rs` |
| `RowSize` scalar fn + size/classify pass (`compute_sizes`) | `src/size.rs` |
| `RowEncodingOptions`, `RowSortField` | `src/options.rs` |
| per-dtype byte codec (`field_size` / `field_encode`) | `src/codec.rs` |
| `initialize(session)` + re-exports | `src/lib.rs` |

## Type Coverage

Supported: nulls, booleans, integer/float primitives, decimals up to 128 bits, UTF-8 and
binary, structs, and fixed-size lists.

Rejected: extension types, variant, union, and variable-size list arrays. `Decimal256` is also
not implemented. Temporal extensions could be added later by normalizing them to storage arrays
at the `RowEncoder` boundary once the temporal ordering contract is made explicit.

## Docs

Adds `docs/specs/row-encoding.md`, a FoundationDB-tuple-style byte-sort specification with:

- sentinel summary table
- `BE(value)` definition and examples
- per-type encoding rules
- unsupported-type table
- worked row example at the end

The spec and `vortex-row/README.md` both mark the byte layout as experimental.

## Testing

- `cargo test -p vortex-row` — 23 tests passed.
- `uv run --all-packages make -C docs doctest` — 268 doctests passed.
- `cargo +nightly fmt --all` — clean.
- `cargo clippy -p vortex-row --all-targets --all-features -- -D warnings` — clean.
- `git diff --check` — clean.


dimitarvdimitrov (commented):
SGTM

dimitarvdimitrov (approved):

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+
+fn add_size_primitive(arr: &PrimitiveArray, sizes: &mut [u32]) {
+    let width = byte_width_u32(arr.ptype().byte_width());
+    add_size_const(sizes, encoded_size_for_fixed(width));
+}
+
+fn add_size_decimal(arr: &DecimalArray, sizes: &mut [u32]) {
+    let width = byte_width_u32(arr.values_type().byte_width());
+    add_size_const(sizes, encoded_size_for_fixed(width));
+}
```
AdamGS, line 389: all these are just one-liners - `sizes.iter_mut().for_each(...)`

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+    let bits = arr.clone().into_bit_buffer();
+    let non_null = field.non_null_sentinel();
+    let xor = if field.descending { 0xFF } else { 0x00 };
+    match resolve_validity(arr.as_ref().validity()?, arr.len(), ctx)? {
+        ValidityKind::AllValid => {
+            for i in 0..bits.len() {
+                let pos = (row_offsets[i] + col_offset[i]) as usize;
+                out[pos] = non_null;
+                // false=0x01, true=0x02 so false < true; XOR for descending
+                let raw = if bits.value(i) { 0x02u8 } else { 0x01u8 };
```
AdamGS, line 547: `as u8 + 1`?

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+                col_offset[i] += BOOL_ENCODED_SIZE;
+            }
+        }
+        ValidityKind::Mask(mask) => {
+            let null = field.null_sentinel();
+            for i in 0..bits.len() {
+                let pos = (row_offsets[i] + col_offset[i]) as usize;
+                if mask.value(i) {
+                    out[pos] = non_null;
+                    let raw = if bits.value(i) { 0x02u8 } else { 0x01u8 };
```
AdamGS, line 558: same as above

vortex-row/src/options.rs
```diff
@@ -0,0 +1,193 @@
...
+    #[inline]
+    pub(crate) fn null_sentinel(&self) -> u8 {
+        if self.nulls_first {
+            // Nulls before non-nulls (smaller byte sorts first).
+            0x00
+        } else {
+            // Nulls after non-nulls (larger byte sorts later).
+            0x02
+        }
+    }
```
AdamGS, line 86: These (`non_null_sentinel`, `null_sentinel`) shouldn't be part of `RowSortField`, they don't really fit the rest of the pattern

vortex-row/src/options.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+const FIELDS_INLINE: usize = 4;
```
AdamGS, line 89: explain? not sure I get what's going on here

vortex-row/src/options.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use smallvec::SmallVec;
+
+/// Per-column ordering options for row-oriented encoding.
+///
+/// A `RowSortField` describes how one input column contributes to a row key. Descending order
+/// reverses the encoded value bytes for that column. Null placement is controlled separately,
+/// so nulls keep the requested position relative to non-null values in either direction.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct RowSortField {
```
AdamGS, line 15: nit - `RowSortFieldOptions`?

vortex-row/src/tests.rs
```diff
@@ -0,0 +1,615 @@
+// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: I think there are two types of tests that are missing here:
1. Lock in some reference values (Like in the docs page).
2. some property testing - generate a dtype, generate values for that dtype and assert their order doesn't change.

docs/specs/row-encoding.md
```diff
@@ -0,0 +1,539 @@
+# Row Encoding Byte Sort Specification
```
AdamGS, line 1: can more of this content be in the `lib.rs`? I think that's a much more accessible location, and more helpful for reading the code.

AdamGS, line 1: How much trust do you have in this scheme?

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
+// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: This file is almost half this PR, probably worth splitting it into a module with multiple files

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+fn byte_width_u32(width: usize) -> u32 {
+    u32::try_from(width).vortex_expect("native byte width must fit in u32")
+}
+
+/// Pre-resolved per-row validity for the row encoders.
+///
+/// Encoders pattern-match on this once before their inner loop so the no-nulls fast path
+/// avoids per-row `mask.value(i)` branches entirely, and the nullable path materializes the
+/// mask exactly once.
+pub(crate) enum ValidityKind {
```
AdamGS, line 93: Isn't this just `AllOr` from `vortex-mask`?

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+    1 + blocks * VARLEN_BLOCK_TOTAL_U32
+}
+
+/// Constant per-row size in bytes for fixed-width encodings (including 1-byte sentinel).
+#[inline]
+const fn encoded_size_for_fixed(value_bytes: u32) -> u32 {
+    1 + value_bytes
+}
+
+fn byte_width_u32(width: usize) -> u32 {
```
AdamGS, line 84: byte width always fit in a usize, why can't we just cast?

vortex-row/src/size.rs
```diff
@@ -0,0 +1,252 @@
...
+
+/// Classification of a single input column for the size pass.
+///
+/// Tracks each column's within-row byte offset (the constant prefix from all preceding
+/// fixed-width columns) and, for fixed columns, whether any variable-length column has
+/// appeared yet — the encode pass uses this to choose between the arithmetic-write fast
+/// path (no varlen before this column, so the within-row position is constant per row) and
+/// the cursor-write path.
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum ColKind {
```
AdamGS, line 47: ColumnKind

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+    // Outer sentinel: 1 byte per row.
+    for s in sizes.iter_mut() {
+        *s = s.checked_add(1).vortex_expect("per-row size overflow");
+    }
+    match row_width_for_dtype(elem_dtype)? {
+        RowWidth::Fixed(w) => {
+            // Each row has `list_size` fixed-width elements regardless of null parent mask.
+            let body = w
+                .checked_mul(u32::try_from(list_size).vortex_expect("list_size fits u32"))
+                .vortex_expect("FSL body width overflow");
```
AdamGS, line 483: why panic?

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+/// would overflow `u32` is also reported as an error rather than silently saturating.
+pub(crate) fn row_width_for_dtype(dtype: &DType) -> VortexResult<RowWidth> {
+    match dtype {
+        DType::Null => Ok(RowWidth::Fixed(1)),
+        DType::Bool(_) => Ok(RowWidth::Fixed(BOOL_ENCODED_SIZE)),
+        DType::Primitive(ptype, _) => Ok(RowWidth::Fixed(encoded_size_for_fixed(byte_width_u32(
+            ptype.byte_width(),
+        )))),
+        DType::Decimal(dt, _) => {
+            let vt = DecimalType::smallest_decimal_value_type(dt);
```
AdamGS, line 185: is shrinking the decimal here sound? I think it makes the behavior much less predictable.

vortex-row/src/codec.rs
```diff
@@ -0,0 +1,1257 @@
...
+            RowWidth::Variable => Ok(RowWidth::Variable),
+        },
+        DType::Struct(fields, _) => {
+            // Struct is fixed iff all its fields are fixed; sum their widths plus a sentinel.
+            let mut total: u32 = 1; // outer sentinel
+            for field_dtype in fields.fields() {
+                match row_width_for_dtype(&field_dtype)? {
+                    RowWidth::Fixed(w) => {
+                        total = total.checked_add(w).ok_or_else(|| {
+                            vortex_error::vortex_err!("Struct row width overflows u32")
```
AdamGS, line 215: import

AdamGS:
This PR is pretty big, I've made an effort but given that its already merged I have other priorities.

PR #8255, opened 2026-06-04
fix grouped sum index for listview array

onursatici:
## Summary

grouped sum fallback should index the groups correctly when summing

### DIFF
--- a/vortex-array/src/aggregate_fn/accumulator_grouped.rs
+++ b/vortex-array/src/aggregate_fn/accumulator_grouped.rs
@@ -229,11 +229,12 @@ impl<V: AggregateFnVTable> GroupedAccumulator<V> {
         )?;
         let mut states = builder_with_capacity(&self.partial_dtype, offsets.len());
 
-        for (offset, size) in offsets.iter().zip(sizes.iter()) {
+        for (i, (offset, size)) in offsets.iter().zip(sizes.iter()).enumerate() {
             let offset = offset.to_usize().vortex_expect("Offset value is not usize");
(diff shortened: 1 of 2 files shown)


vortex-array/src/aggregate_fn/accumulator_grouped.rs
```diff
@@ -229,11 +229,12 @@ impl<V: AggregateFnVTable> GroupedAccumulator<V> {
...
         let mut states = builder_with_capacity(&self.partial_dtype, offsets.len());
 
-        for (offset, size) in offsets.iter().zip(sizes.iter()) {
+        for (i, (offset, size)) in offsets.iter().zip(sizes.iter()).enumerate() {
             let offset = offset.to_usize().vortex_expect("Offset value is not usize");
             let size = size.to_usize().vortex_expect("Size value is not usize");
 
-            if validity.value(offset) {
+            // validity is for the outer list view, so it must be indexed with `i`
+            if validity.value(i) {
```
gatesn, line 237: I think a Mask iterator is far faster than value for every function

robert3005 (approved):

vortex-array/src/aggregate_fn/accumulator_grouped.rs
onursatici: Yea but I don't think we have that now. I can add one though separately

onursatici: https://github.com/vortex-data/vortex/pull/8258

PR #8257, opened 2026-06-04
fix: `[profile.bench]` `codegen-units = 16`

joseph-isaacs:
#7824 changed this field?

We use 16 since this is cargo release default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -389,7 +389,7 @@ debug = "full"
 inherits = "release"
 
 [profile.bench]
-codegen-units = 1
+codegen-units = 16
 debug = "full"
 lto = false
 


a10y (approved):

AdamGS:
don't we want to bench in the most optimized profile?

AdamGS:
having a description/explaination in the commit message would be useful here, or maybe a comment in the file itself.

joseph-isaacs:
We might want that but we should change this in a stand alone PR

AdamGS:
why? if we're going to revert this we can just directly get to the end state

joseph-isaacs:
Going to merge to speed up the CI. 

I think it might be worth looking at the diff and add some inlines?

PR #8258, opened 2026-06-04
Mask bool iterator

onursatici:
## Summary

Bool iterator to mask to be used by code that gets the bool values from a mask in a loop


joseph-isaacs:
This looks like a very slow way to iterate. We should use a u64 packed validity iterator. 

I am not sure if we should merge then and then that. But each call site will change

robert3005:
BitIterator works on u64 words, not sure what do you mean @joseph-isaacs ?

onursatici:
to get the benefit of word at a time iteration we would need to hand out u64's not bool's to the loop consumers

joseph-isaacs (approved):

onursatici:
we can switch to something like https://github.com/vortex-data/vortex/pull/8262 to get the perf improvement but we need to yield u64 values, not bool

PR #8259, opened 2026-06-04
ref(vortex-io): unify cloud object storage api

m7kss1:
## Summary

Cloud object store construction was duplicated across `vortex-jni` and `vortex-duckdb`,
each with hand-rolled per-scheme builders (`AmazonS3Builder`, `MicrosoftAzureBuilder`,
`GoogleCloudStorageBuilder`), hardcoded scheme lists, and inconsistent credential handling

Introduce `FileLocation` in `vortex-io` as the single resolution point for any
URL or path string - no manual builder code required

```rust
// Handles local paths, file://, s3://, gs://, az://, https:// - anything object_store supports
// Use credentials from env (AWS_ACCESS_KEY_ID etc.)
match FileLocation::resolve(url)? {
    FileLocation::Local(path) => { /* open with async_fs */ }
    FileLocation::Remote { store, path } => { /* open with ObjectStore */ }
}

// Explicit credentials override env (e.g. per-connection props passed)
// props example: {"access_key_id": "XXX", "secret_access_key": XXX", "region": "eu-west-X"}
FileLocation::resolve_with_props(url, java_properties.iter()
    .map(|(k, v)| (k.as_str(), v.as_str())))?
```

By the way, now object store works out of the box in CLI :

```bash
# Pass config arguments from env
$ AWS_REGION=us-east-1 vx query s3://bucket/hits.vortex --sql "select count(*) from hits" 

# Or directly from argv
$ vx query s3://bucket/hits.vortex --sql "select count(*) from hits" \
        --store-option region=us-east-1,allow_http=true 

$ vx tree gs://bucket/hits.vortex --store-option ...
```

Closes: #000

## Testing

AI disclosure: XXX


m7kss1:
note: verified only on `s3`-compatible storages, `azure` and `gcs` remain untested for now

AdamGS:
Hi @m7kss1 thanks for your PR! I'll review it next week.

robert3005:
You have one failing test case on windows `object_store::cloud::tests::test_resolve`. Otherwise I think this looks good.

vortex-io/src/object_store/cloud.rs
```diff
@@ -0,0 +1,299 @@
...
+use vortex_error::VortexResult;
+use vortex_error::vortex_bail;
+use vortex_error::vortex_err;
+use vortex_utils::aliases::hash_map::HashMap;
+
+use crate::object_store::registry::Registry;
+
+/// The process-global registry used by [`FileLocation::resolve`] to cache lazily-constructed
+/// stores.
+static DEFAULT_REGISTRY: LazyLock<Registry> = LazyLock::new(Registry::default);
```
AdamGS: I think its better to have the static registries in the user crates (duckdb, python, JNI). Other places probably want to have the flexibility to control the lifetime in a way that makes sense to them.

vortex-io/src/object_store/cloud.rs
```diff
@@ -0,0 +1,299 @@
...
+            tracing::trace!("using LocalFileSystem object store");
+            Arc::new(LocalFileSystem::default())
+        }
+        ObjectStoreScheme::AmazonS3 => {
+            tracing::trace!("using AmazonS3 object store");
+            let mut builder = AmazonS3Builder::new()
+                .with_url(url.to_string())
+                // Use a generic S3 endpoint to avoid DNS resolution issues with
+                // region-specific endpoints.
+                .with_endpoint("https://s3.amazonaws.com")
```
AdamGS: This should be configurable

vortex-io/src/object_store/registry.rs
```diff
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: Apache Software Foundation (ASF)
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
```
AdamGS: The copyright here is intentional, we pulled this code from an Apache project.

vortex-io/src/object_store/cloud.rs
```diff
@@ -0,0 +1,299 @@
...
+/// per-property configuration passthrough.
+// The cognitive-complexity lint only triggers under some feature unifications (it depends on how
+// the `tracing` macros expand), so we use `allow` rather than `expect` to avoid an unfulfilled
+// expectation in minimal-feature builds.
+#[allow(clippy::cognitive_complexity)]
+pub fn make_object_store(
+    url: &Url,
+    properties: &HashMap<String, String>,
+) -> VortexResult<Arc<dyn ObjectStore>> {
+    static OBJECT_STORES: LazyLock<Mutex<HashMap<String, Arc<dyn ObjectStore>>>> =
```
AdamGS: Like above - I don't think we should have static state, especially having two different static states that are not shared and have very different behavior.

AdamGS (changes requested):
I've added a few more focused comments, but I think there are two bigger issues here:
1. I don't think this quite solves the issue you describe in the PR.
While it does centralize the implementations (which I agree is desirable), it keeps some of inconsistencies between them. The registry is a lot more configurable than `make_object_store`, which hard-codes a few assumptions.
2. It introduces a bunch of new public APIs like the `Registry` but more importantly `ObjectStore`, which is currently a crate we use in non-rust integrations and CAN work with for rust codebases when it makes sense for the user. I think our current IO traits will likely change, but still we should probably use them externally.

m7kss1:
@AdamGS thanks, that's a good point. the idea behind this approach was to avoid making any major changes to the current workflow and reuse as much as possible from the original implementation, to unlock the possibility for future refactoring in other crates. though, i think we can make them directly inside this MR, so I've temporarily marked MR as a draft and will ping you once i've finished the changes

vortex-io/src/object_store/cloud.rs
m7kss1: @AdamGS indeed, but if we make them configurable, should we keep the current ones as defaults + change api to allow use custom properties? in trunk, these settings are hard-coded, so this should be handled extra carefully (ref: https://github.com/vortex-data/vortex/blob/develop/vortex-jni/src/object_store.rs#L71-L78)

vortex-tui/src/segments.rs
```diff
@@ -17,6 +18,10 @@ use crate::segment_tree::collect_segment_tree;
 pub struct SegmentsArgs {
     /// Path to the Vortex file
     pub file: PathBuf,
+
+    /// Object store options for remote files (see `--store-option`).
+    #[command(flatten)]
+    pub store: crate::store_options::StoreOptions,
```
AdamGS, line 24: nit - import here

vortex-python/src/object_store/registry.rs
```diff
@@ -1,192 +0,0 @@
-// SPDX-License-Identifier: Apache-2.0
```
AdamGS, line 1: This is still important for things like JNI, DuckDB and python, so they won't keep creating new connection if one already exists. `ObjectStore` implementations are backed by a connection pool so sharing is very beneficial.

PR #8264, opened 2026-06-05
perf: branchless mask-select for listview zip

joseph-isaacs:
## Summary

Replaces the per-element, data-dependent branch in the listview zip kernel's offset/size selection with a branchless, chunk-at-a-time mask select that the compiler can auto-vectorize.

For each 64-bit mask chunk, each bit is expanded to a full-width lane mask and both sides are blended with `(t & m) | (f & !m)` via a shared `select_column` helper, so the inner loop is branch-free regardless of mask shape. `if_false` offsets are shifted into the second half of the concatenated `elements` as before; sizes are taken verbatim from the chosen side.

> The primitive-array zip kernel that was previously bundled here has been split into its own PR.

### Changes

- `vortex-array/src/arrays/listview/compute/zip.rs`: branchless chunked offset/size select via `select_column`; added a test that spans the 64-bit chunk boundary + remainder.
- `vortex-array/benches/listview_zip.rs`: a small divan bench — one Fragmented (alternating) mask with non-nullable and nullable inputs.

### Benchmark

Because the branchless select is mask-shape-independent, the bench uses a single Fragmented mask (the worst case for the replaced per-element branch) with both nullability variants. `LEN` is 8,192: listview zip cost is dominated by element concatenation and per-list canonicalization, so a few thousand lists already exercise the select while keeping each case well under a few hundred microseconds (~37 µs non-null, ~41 µs nullable locally).

The kernel change itself is a single-to-low-double-digit-percent win on the offset/size select; end-to-end listview zip is dominated by element concatenation + canonicalization, so the select is a small slice of total cost.

### Testing

- `vortex-array` listview zip tests pass (incl. the new multi-chunk-span test); `cargo +nightly fmt` and `clippy -D warnings` (default + all-features) clean.

https://claude.ai/code/session_01N5ivPiCJy7dGQjMEP7ips9

### DIFF
--- a/vortex-array/src/arrays/listview/compute/zip.rs
+++ b/vortex-array/src/arrays/listview/compute/zip.rs
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use std::mem::MaybeUninit;
 use std::ops::BitAnd;
 use std::ops::BitOr;
 use std::ops::Not;
 
 use vortex_buffer::Buffer;
 use vortex_buffer::BufferMut;
+use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_mask::Mask;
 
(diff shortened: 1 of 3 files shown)


vortex-array/src/arrays/listview/compute/zip.rs
```diff
@@ -85,24 +87,56 @@ impl ZipKernel for ListView {
...
+            let sizes_out = sizes.spare_capacity_mut();
+
+            // We matched `Mask::Values` above, so the bit buffer is materialized. Walk it as 64-bit
+            // chunks and branchlessly blend both sides per row, letting the compiler vectorize the
+            // inner select instead of mispredicting a data-dependent branch per element.
+            let mask_bits = mask
+                .values()
+                .vortex_expect("mask is Mask::Values")
+                .bit_buffer();
+            let chunks = mask_bits.chunks();
```
robert3005: The conventional wisdom tells you that you want to use unaligned_chunks for faster iteration speed.

robert3005 (approved):

PR #8267, opened 2026-06-05
use octet_length() instead of len() in clickbench

myrrc:
Clickhouse's length() is non-standard SQL which returns number of bytes and not
Unicode code points. Replace this for ClickBench to Duckdb's octet_length() to match
original intent. As Duckdb doesn't support octet_length for strings and uses non-standard strlen(), add a macro to convert octet_length to strlen().
We can't use strlen() directly because DataFusion's strlen() counts code points, not bytes.


joseph-isaacs (approved):

benchmarks/duckdb-bench/src/main.rs
```diff
@@ -178,6 +178,11 @@ fn main() -> anyhow::Result<()> {
                 args.threads,
             )?;
             ctx.register_tables(&*benchmark, format)?;
+
+            // Duckdb doesn't support octet_length for strings but we need this
+            // in ClickBench.
+            ctx.execute_query_result("create macro octet_length(a) as strlen(a)")?;
```
joseph-isaacs: we really should move this into a benchmark specific `setup.sql` function. I think that other benchmarks might be affected

myrrc: No, because duckdb doesn't support octet length so there were no queries with it.

joseph-isaacs: this is a bug waiting to happen

PR #8269, opened 2026-06-05
Deprecate vortex-array public APIs that use the hidden LEGACY_SESSION

joseph-isaacs:
## Summary

Several public `vortex-array` methods construct an `ExecutionCtx` from the hidden global `LEGACY_SESSION` static instead of accepting an explicit `ExecutionCtx`. This PR marks them `#[deprecated]` so downstream callers migrate to the context-threading APIs (matching the existing `IntoArrowArray` deprecation pattern).

## Deprecated public APIs

| API | File | Suggested replacement |
|---|---|---|
| `BoolTyped::true_count` | `vortex-array/src/variants.rs` | `sum(array, ctx)` with an explicit `ExecutionCtx` |
| `PrimitiveTyped::value` | `vortex-array/src/variants.rs` | `is_valid` / `execute_scalar` with an explicit `ExecutionCtx` |
| `PrimitiveTyped::value_unchecked` | `vortex-array/src/variants.rs` | `execute_scalar` with an explicit `ExecutionCtx` |
| `from_arrow_array_with_len` | `vortex-array/src/arrow/datum.rs` | thread an explicit `ExecutionCtx` through `execute_scalar` |

Each carries a `#[deprecated(note = …)]` explaining why and what to use instead.

### DIFF
--- a/encodings/datetime-parts/src/compute/compare.rs
+++ b/encodings/datetime-parts/src/compute/compare.rs
@@ -200,11 +200,12 @@ fn compare_dtp(
 }
 
 #[cfg(test)]
-#[allow(deprecated)]
 mod test {
     use rstest::rstest;
+    use vortex_array::ArrayRef;
     use vortex_array::LEGACY_SESSION;
     use vortex_array::VortexSessionExecute;
+    use vortex_array::aggregate_fn::fns::sum::sum;
     use vortex_array::arrays::PrimitiveArray;
     use vortex_array::arrays::TemporalArray;
     use vortex_array::dtype::IntegerPType;
@@ -231,6 +232,16 @@ mod test {
         .expect("Failed to construct DateTimePartsArray from TemporalArray")
     }
 
+    /// Count the true values in a boolean array using an explicit execution context.
+    fn true_count(array: &ArrayRef) -> usize {
+        let mut ctx = LEGACY_SESSION.create_execution_ctx();
+        sum(array, &mut ctx)
+            .unwrap()
+            .as_primitive()
+            .as_::<usize>()
+            .unwrap()
+    }
+
     #[rstest]
     #[case(Validity::NonNullable, Validity::NonNullable)]
     #[case(Validity::NonNullable, Validity::AllValid)]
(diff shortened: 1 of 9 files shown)


connortsui20 (commented):
use expect deprecated and maybe add a reason for each?

encodings/datetime-parts/src/compute/rules.rs
```diff
@@ -200,6 +202,16 @@ mod tests {
 
     const SECONDS_PER_DAY: i64 = 86400;
 
+    /// Count the true values in a boolean array using an explicit execution context.
+    fn true_count(array: &ArrayRef) -> usize {
```
AdamGS: this is a test-only function, but it doesn't do what the comment says it does

encodings/datetime-parts/src/compute/compare.rs
```diff
@@ -230,6 +232,16 @@ mod test {
         .expect("Failed to construct DateTimePartsArray from TemporalArray")
     }
 
+    /// Count the true values in a boolean array using an explicit execution context.
+    fn true_count(array: &ArrayRef) -> usize {
```
AdamGS: same as below

vortex-array/src/arrow/datum.rs
```diff
@@ -105,6 +105,9 @@ impl ArrowDatum for Datum {
 /// # Error
 ///
 /// The provided array must have length
+#[deprecated(
+    note = "Relies on the hidden global `LEGACY_SESSION`; thread an explicit `ExecutionCtx` through `execute_scalar` instead"
+)]
```
AdamGS, line 110: I think the note is wrong here?

encodings/datetime-parts/src/compute/rules.rs
joseph-isaacs: I don't see the problem with this test the ctx is explicit to this test module not hidden in the true count

joseph-isaacs: Agreed — done in f8e0ae8. `true_count` now takes `ctx: &mut ExecutionCtx`, and each test constructs the context explicitly (`LEGACY_SESSION.create_execution_ctx()`) before calling it, so the ctx is visible in the test rather than hidden in the helper. Also updated the doc comment to say "the provided execution context".

---

vortex-array/src/arrow/datum.rs
joseph-isaacs: Good catch — updated in 94eae41 to point at the actual replacement: `note = "Relies on the hidden global LEGACY_SESSION; use from_arrow_columnar with an explicit ExecutionCtx instead"`.

---

AdamGS (approved):

PR #8270, opened 2026-06-05
perf: branchless primitive zip kernel

joseph-isaacs:
## Summary

Adds a dedicated primitive zip kernel that selects values branchlessly per row.

The generic zip path copies runs of `if_true`/`if_false` between mask boundaries — fast for clustered masks but degrading to per-element work on fragmented masks. This kernel walks the mask as 64-bit chunks and blends both sides per row with no data-dependent branch, so the inner loop stays branch-free and auto-vectorizable regardless of mask shape. Result validity reuses the shared `zip_validity` helper, which expresses validity selection as a (lazy) zip over the two validity bitmaps.

> The branchless boolean zip kernel (#8275) this builds on has now merged into `develop`; this branch has been rebased on top of it, so the diff here is primitive-only.

### DIFF
--- a/vortex-array/Cargo.toml
+++ b/vortex-array/Cargo.toml
@@ -197,6 +197,10 @@ name = "bool_zip"
 harness = false
 
 [[bench]]
+name = "primitive_zip"
+harness = false
+
+[[bench]]
 name = "take_primitive"
 harness = false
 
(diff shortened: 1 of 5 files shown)


robert3005 (approved):
the only comment I have is that unaligned_chunks is a faster iterator than `chunks` but we can leave that for later

PR #8274, opened 2026-06-05
Pass execution context through list view rebuild

connortsui20:
## Summary

TODO is there a tracking issue for deprecating the old to_canonical path? This PR makes a solid amount of progress in removing deprecated stuff.

Threads `ExecutionCtx` through `ListView::rebuild` and the paths it reaches (`reset_offsets`, `compact`, `list_from_list_view`, `recursive_list_from_list_view`) so they reuse the caller's session instead of each minting a fresh `LEGACY_SESSION` context internally.

`reset_offsets` in `list_view_from_list` now propagates with `?` instead of `vortex_expect` since execution can fail once it runs through it (https://github.com/vortex-data/vortex/pull/8274#discussion_r3371872039) 

## API Changes

A few places that now get an `ExecutionCtx` passed through them.

## Testing

Tests and benchmarks also now use a local empty session. This is just a lot of mechanical / not semantic changes.

### DIFF
--- a/encodings/zstd/benches/listview_rebuild.rs
+++ b/encodings/zstd/benches/listview_rebuild.rs
@@ -3,17 +3,24 @@
 
 #![expect(clippy::unwrap_used)]
 
+use std::sync::LazyLock;
+
 use divan::Bencher;
 use vortex_array::IntoArray;
-use vortex_array::LEGACY_SESSION;
 use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::ListViewArray;
 use vortex_array::arrays::VarBinViewArray;
 use vortex_array::arrays::listview::ListViewRebuildMode;
+use vortex_array::session::ArraySession;
 use vortex_array::validity::Validity;
 use vortex_buffer::Buffer;
+use vortex_session::VortexSession;
 use vortex_zstd::Zstd;
 use vortex_zstd::ZstdData;
+
+/// A shared session for the `ListView` rebuild benchmark, used to create execution contexts.
+static SESSION: LazyLock<VortexSession> =
+    LazyLock::new(|| VortexSession::empty().with::<ArraySession>());
 
 #[divan::bench(sample_size = 1000)]
 fn rebuild_naive(bencher: Bencher) {
@@ -23,7 +30,7 @@ fn rebuild_naive(bencher: Bencher) {
     let validity = dudes.validity().unwrap();
     let dudes = Zstd::try_new(
         dtype,
-        ZstdData::from_array(dudes, 9, 1024, &mut LEGACY_SESSION.create_execution_ctx()).unwrap(),
+        ZstdData::from_array(dudes, 9, 1024, &mut SESSION.create_execution_ctx()).unwrap(),
         validity,
     )
     .unwrap()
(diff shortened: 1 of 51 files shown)


connortsui20:
I dont trust codspeed here

encodings/zstd/benches/listview_rebuild.rs
```diff
@@ -41,9 +48,12 @@ fn rebuild_naive(bencher: Bencher) {
 
     let list_view = ListViewArray::new(dudes, offsets, sizes, Validity::NonNullable);
 
-    bencher
-        .with_inputs(|| &list_view)
-        .bench_refs(|list_view| list_view.rebuild(ListViewRebuildMode::MakeZeroCopyToList))
+    bencher.with_inputs(|| &list_view).bench_refs(|list_view| {
+        list_view.rebuild(
+            ListViewRebuildMode::MakeZeroCopyToList,
+            &mut SESSION.create_execution_ctx(),
```
joseph-isaacs: you will bench creating this

vortex-array/src/arrays/listview/conversion.rs
```diff
@@ -41,7 +39,9 @@ pub fn list_view_from_list(list: ListArray, ctx: &mut ExecutionCtx) -> VortexRes
     // We reset the offsets here because mostly for convenience, and also because callers of this
     // function might not expect the output `ListViewArray` to have a bunch of leading and trailing
     // garbage data when they turn it back into a `ListArray`.
-    let list = list.reset_offsets(false).vortex_expect("This can't fail");
+    let list = list
+        .reset_offsets(false, ctx)
+        .vortex_expect("This can't fail");
```
joseph-isaacs: is this still true?

connortsui20: it used to be the case that it couldnt fail, but now that there is execution in it I guess it can fail

vortex-array/src/arrays/listview/conversion.rs
```diff
@@ -29,8 +27,8 @@ use crate::match_each_integer_ptype;
 
 /// Creates a `ListViewArray` from a `ListArray` by computing `sizes` from `offsets`.
 ///
-/// The output `ListViewArray` will be zero-copyable back to a `ListArray`, and additionally it
-/// will not have any leading or trailing garbage data.
+/// The output `ListViewArray` will be zero-copyable back to a `ListArray`, and additionally it will
+/// not have any leading or trailing garbage data.
 pub fn list_view_from_list(list: ListArray, ctx: &mut ExecutionCtx) -> VortexResult<ListViewArray> {
```
joseph-isaacs, line 32: Also should this be a method on the ListArray?

vortex-array/src/builders/listview.rs
```diff
@@ -343,8 +349,11 @@ impl<O: IntegerPType, S: IntegerPType> ArrayBuilder for ListViewBuilder<O, S> {
         let uninit_range = self.offsets_builder.uninit_range(extend_length);
 
         // This should be cheap because we didn't compress after rebuilding.
-        #[expect(deprecated)]
-        let new_offsets = listview.offsets().to_primitive();
+        let new_offsets = listview
+            .offsets()
+            .clone()
+            .execute::<PrimitiveArray>(&mut ctx)
+            .vortex_expect("failed to execute list view offsets into a PrimitiveArray");
```
joseph-isaacs, line 356: can this fail?

vortex-array/src/arrays/listview/conversion.rs
connortsui20: I wanted it standalone since it is unclear if this should live on list view or list

vortex-array/src/builders/listview.rs
connortsui20: probably, but not much we can do about it because of the builders API

joseph-isaacs: Right do we have a issue for addressing this?

joseph-isaacs (approved):

PR #8277, opened 2026-06-05
feat(vortex-array): add Interleave array encoding

joseph-isaacs:
Adds an `Interleave` array: a lazy, **random-access** gather of `N` value arrays into one array, taking output row `i` from `values[array_indices[i]][row_indices[i]]`.


robert3005:
annoyingly arrow calls your `Merge` -> `interleave` and we already have `merge` that zips two structs. I guess you could consider the existing merge the special case of no selector child where all children are kept

joseph-isaacs:
Can you link what you mean, I based this off arrow-rs naming

robert3005:
I thought arrow interleave just takes a single value per row but they take tuple per row. This is the same then

joseph-isaacs:
@robert3005 are we good to merge this?

robert3005 (commented):
Will approve once we have working ci. I think this is the right approach.

robert3005 (approved):

PR #8299, opened 2026-06-08
Don't use SharedArray for dict values

myrrc:
SharedArray prevents us from using FSST data in queries like
SELECT byte_length(str), from SharedArray we already get decompressed
VarBinView data.

Check what of performance impact is not using the array


joseph-isaacs:
We really don't want to do this?

PR #8303, opened 2026-06-08
Scalar functions pushdown to Duckdb

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

Subtasks:

Upstreaming:

Additional work:
  BoundColumnRef.

myrrc:
- Rewrite duckdb query plan to push down expresions of form `SELECT fn(col) FROM '1.vortex'` to just `SELECT col FROM '1.vortex'` if "col" isn't used without fn in the query. 
- Add printing of pushed down functions as "SELECT projections".
- Push down `strlen` in SELECT and in WHERE.

Depends on #8371 and #8341

### DIFF
--- a/vortex-duckdb/cpp/optimizer.cpp
+++ b/vortex-duckdb/cpp/optimizer.cpp
@@ -93,13 +93,12 @@ void FindGetsAndProjections(LogicalOperator &op, Analyses &analyses, Projections
         LogicalProjection &projection = op.Cast<LogicalProjection>();
         D_ASSERT(projection.children.size() == 1);
         auto &child = *projection.children[0];
-        if (child.type != LOGICAL_GET) {
+        if (!is_passthrough(projection) || child.type != LOGICAL_GET) {
             break;
         }
         // The GET itself is recorded when recursion reaches it below. Only
         // passthrough projections wrapping a vortex GET act as aliases.
-        if (auto &get = child.Cast<LogicalGet>();
-            get.function.bind == duckdb_vx_table_function_bind && is_passthrough(projection)) {
+        if (auto &get = child.Cast<LogicalGet>(); get.function.bind == duckdb_vx_table_function_bind) {
             projections.emplace(projection.table_index, projection);
         }
         break;
--- a/vortex-duckdb/cpp/optimizer.cpp
+++ b/vortex-duckdb/cpp/optimizer.cpp
@@ -1,0 +1,250 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/planner/operator/logical_projection.hpp"
+#include "duckdb_vx/optimizer.h"
+#include "duckdb_vx/table_function.h"
+#include "vortex.h"
+#include <optional>
+
+/**
+ * Our optimizer runs after all duckdb optimizers. Functions that can be pushed
+ * down as complex filters (e.g. WHERE str != '') are already pushed down.
+ * If there are any functions left, this means they were not pushed down and
+ * may produce conflicts (e.g. WHERE prefix("str", 'h')).
+ */
+
+extern "C" duckdb_state duckdb_vx_optimizer_extension_register(duckdb_database ffi_db) {
+    D_ASSERT(ffi_db);
+    const DatabaseWrapper &wrapper = *reinterpret_cast<DatabaseWrapper *>(ffi_db);
+    DatabaseInstance &db = *wrapper.database->instance;
+    try {
+        DBConfig::GetConfig(db).GetCallbackManager().Register(VortexOptimizerExtension());
+    } catch (const std::exception &e) {
+        ErrorData data(e);
+        DUCKDB_LOG_ERROR(db, "Failed to create Vortex optimizer extension:\t" + data.Message());
+        return DuckDBError;
+    }
+    return DuckDBSuccess;
+}
+
+void VortexOptimizeFunction(OptimizerExtensionInput &input, LogicalOperatorPtr &plan) {
+    plan = TryPushdownScalarFunctions(input.context, std::move(plan));
+}
+
+LogicalOperatorPtr TryPushdownScalarFunctions(ClientContext &context, LogicalOperatorPtr plan) {
+    Analyses analyses;
+    Projections projections;
+    FindGetsAndProjections(*plan, analyses, projections);
+    if (analyses.empty()) {
+        return plan;
+    }
+    ScalarFnCollect(analyses, projections).VisitOperator(*plan);
+
+    bool any_pushed = false;
+    for (auto &[_, analysis] : analyses) {
+        for (auto &[column_index, expr] : analysis.col_to_fn) {
+            if (expr == nullptr) { // Conflict for column
+                continue;
+            }
+            const TableColumnStorageIndex storage_index =
+                analysis.get.GetColumnIds()[column_index].GetPrimaryIndex();
+            TableFunctionProjectionExpressionInput input {analysis.get, *expr, storage_index};
+            if (projection_expression_pushdown(context, input)) {
+                analysis.get.types[column_index] = expr->return_type;
+                analysis.get.returned_types[storage_index] = expr->return_type;
+                any_pushed = true;
+            } else { // failed to push down expression, can't replace it
+                expr = nullptr;
+            }
+        }
+    }
+
+    if (any_pushed) {
+        ScalarFnReplace(analyses, projections).VisitOperator(*plan);
+    }
+    return plan;
+}
+
+// A passthrough projection only forwards its child columns, e.g. a VIEW's
+// "SELECT col".
+static bool is_passthrough(const LogicalProjection &projection) {
+    if (projection.expressions.empty()) {
+        return false; // don't register empty projections in Projections
+    }
+    for (const auto &e : projection.expressions) {
+        if (e->GetExpressionType() != ExpressionType::BOUND_COLUMN_REF) {
+            return false;
+        }
+    }
+    return true;
+}
+
+void FindGetsAndProjections(LogicalOperator &op, Analyses &analyses, Projections &projections) {
+    using enum LogicalOperatorType;
+    switch (op.type) {
+    case LOGICAL_GET: {
+        if (auto &get = op.Cast<LogicalGet>(); get.function.bind == duckdb_vx_table_function_bind) {
+            analyses.emplace(get.table_index, GetAnalysis {get, {}});
+        }
+        break;
+    }
+    case LOGICAL_PROJECTION: {
+        LogicalProjection &projection = op.Cast<LogicalProjection>();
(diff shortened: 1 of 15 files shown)


myrrc:
Local run shows 2x speedup for clickbench 27 (arm, 250->140ms), CI run shows a slowdown, so something wrong is going on.

On a local x86 box the speedup is 680ms -> 345 ms .

vortex-sqllogictest/slt/duckdb/projection_expression_pushdown.slt
```diff
@@ -0,0 +1,427 @@
...
+----
+Hello
+Hey
+Hi
+
+# explain: strlen is pushed down with direct reads from vortex
+query TT
+EXPLAIN (FORMAT json) SELECT strlen(str) FROM '$__TEST_DIR__/pe-pushdown.vortex';
+----
+<REGEX>:SELECT projections
```
joseph-isaacs: explain this test

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -170,15 +172,35 @@ struct CTableBindResult {
...
+bool projection_expression_pushdown(ClientContext &, const TableFunctionProjectionExpressionInput &input) {
+    const auto &bind = input.get.bind_data->Cast<CTableBindData>();
+
+    // This is a flaw of Duckdb API which doesn't allow passing non-const
+    // expressions. We never modify the value on Rust side.
+    auto ffi_expr = reinterpret_cast<duckdb_vx_expr>(const_cast<Expression *>(&input.expression));
+    void *const ffi_bind = bind.ffi_data->DataPtr();
+    duckdb_vx_error error_out = nullptr;
+
+    const bool ret = duckdb_table_function_pushdown_projection_expression( //
```
0ax1, line 184: Stray comment? `//`

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -170,15 +172,35 @@ struct CTableBindResult {
     vector<string> &names;
 };
 
+bool projection_expression_pushdown(ClientContext &, const TableFunctionProjectionExpressionInput &input) {
+    const auto &bind = input.get.bind_data->Cast<CTableBindData>();
+
+    // This is a flaw of Duckdb API which doesn't allow passing non-const
```
0ax1, line 178: nit: `DuckDB's API`

vortex-duckdb/cpp/optimizer.cpp
```diff
@@ -0,0 +1,243 @@
...
+    ExpressionPtr &bound_col_base = expr.children[0];
+    const auto &bound_col = bound_col_base->Cast<BoundColumnRefExpression>();
+    const auto binding = Resolve(bound_col.binding, analyses, projections);
+    if (!binding) {
+        return nullptr;
+    }
+
+    const auto &[analysis, column_index] = *binding;
+    if (conflict(analysis, column_index)) {
+        return std::move(*ptr);
```
0ax1, line 233: Why move?

vortex-duckdb/cpp/optimizer.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    bound_col_base->return_type = analysis.get.types[column_index];
+    return std::move(bound_col_base);
```
0ax1, line 240: Same here, why move. Unless we have to use it, or there's a measurable perf benefit the complexity of using `move` is probably not worth it.

vortex-duckdb/cpp/optimizer.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+    ExpressionPtr &bound_col_base = expr.children[0];
+    const auto &bound_col = bound_col_base->Cast<BoundColumnRefExpression>();
+    const auto binding = Resolve(bound_col.binding, analyses, projections);
+    if (!binding) {
+        return nullptr;
+    }
+
+    const auto &[analysis, column_index] = *binding;
+    if (conflict(analysis, column_index)) {
```
0ax1: `conflict` is a bit brief to be guessable what it does.

vortex-duckdb/cpp/optimizer.cpp
```diff
@@ -0,0 +1,243 @@
...
+        return nullptr;
+    }
+    const auto &bound_col = expr.children[0]->Cast<BoundColumnRefExpression>();
+    const auto binding = Resolve(bound_col.binding, analyses, projections);
+    if (!binding) {
+        return nullptr;
+    }
+    auto &col_to_fn = binding->analysis.col_to_fn;
+
+    if (auto it = col_to_fn.find(binding->column_index); it == col_to_fn.end()) {
```
0ax1, line 185: I've probably asked this before but isn't there a `contains` fn. And if we'd need a higher C++ std, can we have one in a utility file on our end?

vortex-duckdb/cpp/table_function.cpp
myrrc: This is a formatter hack.

vortex-duckdb/cpp/optimizer.cpp
myrrc: unique_ptr doesn't have a copy operator.

vortex-duckdb/cpp/optimizer.cpp
myrrc: We can't use it here because contains() returns bool and we want to use the
result iterator in "else" clause. Even if we had "contains", we'd have to recompute hash which we don't do here

0ax1:
@claude review the c++ code, anything unsafe/unsound?

0ax1 (approved):
Shape of the code looks good, but we should run C++ code executed as part of the SLT tests with sanitizers.

vortex-sqllogictest/slt/duckdb/projection_expression_pushdown.slt
myrrc: +1

vortex-sqllogictest/slt/duckdb/projection_expression_pushdown.slt
```diff
@@ -0,0 +1,427 @@
...
+
+include ../setup.slt.no
+
+# str has storage_index=1. This tests cases where column_index (i.e. 0 in
+# SELECT strlen(str)) is different from storage_index.
+query I
+CREATE TABLE pep_test (before INTEGER, str VARCHAR, after INTEGER);
+----
+
+query I
```
AdamGS: btw - you I think you can also do `statement ok` for stuff like this

vortex-duckdb/src/convert/expr.rs
```diff
@@ -56,6 +62,20 @@ fn try_from_bound_function(
     col_sub: Option<&Expression>,
 ) -> VortexResult<Option<Expression>> {
     let expr = match func.scalar_function.name() {
```
AdamGS, line 64: ~We should be matching on ID here, names are allowed to conflict IMO.~ NVM I realize this is a duckdb type 😢

vortex-duckdb/src/projection.rs
```diff
@@ -43,6 +42,9 @@ pub struct DuckdbField {
     pub name: String,
     pub logical_type: LogicalType,
     pub dtype: DType,
+    /// Function to use instead of get_item(col, root()), e.g. len(col).
+    /// It does not include column name so it's just "len" and not "len(col)"
```
AdamGS: This is not a function, the comments seems to be out of sync with the code

vortex-duckdb/src/projection.rs
```diff
@@ -86,47 +89,92 @@ impl Projection {
...
+            };
+        }
 
-            select(names, root())
-        };
+        let has_fn_columns = fn_col_count > 0;
+        let mut all_exprs = Vec::with_capacity(
+            (ids.len() + has_file_row_number as usize) * has_fn_columns as usize,
+        );
+        let mut named_fields = Vec::with_capacity(ids.len() * !has_fn_columns as usize);
```
AdamGS:
```rust
let (mut all_expres, mut named_fields) = has_fn_columns {
    ...
} else {
    ...
}
```

vortex-duckdb/src/projection.rs
```diff
@@ -86,47 +89,92 @@ impl Projection {
...
+            return Projection {
+                projection,
+                file_index_column_pos,
+                file_row_number_column_pos,
+            };
+        }
 
-            select(names, root())
-        };
+        let has_fn_columns = fn_col_count > 0;
```
AdamGS: Like above - this should probably be like `has_pushdown_columns` or something like it?

vortex-duckdb/src/projection.rs
```diff
@@ -86,47 +89,92 @@ impl Projection {
...
+            let column_id = if has_projection_ids {
+                let column_id: usize = column_id.as_();
+                column_ids[column_id]
+            } else {
+                column_id
+            };
+            if is_virtual_column(column_id) {
+                continue;
+            }
+            let column_id: usize = column_id.as_();
```
AdamGS, line 156: Do you get `column_id` twice? How are they different?

vortex-duckdb/src/projection.rs
```diff
@@ -86,47 +89,92 @@ impl Projection {
...
+            let name = column_fields[column_id].name.as_str();
+            if !has_fn_columns {
+                named_fields.push(name);
+                continue;
+            }
+
+            let column_field = &column_fields[column_id];
+            let expr = match &column_field.projection_fn {
```
AdamGS:
```suggestion
            let expr = column_field.projection_fn.cloned().unwrap_or_else(|| get_item(name, root()));
```

vortex-duckdb/src/projection.rs
```diff
@@ -86,47 +89,92 @@ impl Projection {
...
+                None => get_item(name, root()),
+                Some(func) => func.clone(),
+            };
+            all_exprs.push((name, expr));
+        }
+
+        let projection = if has_fn_columns {
+            // If file_row_number was requested, it's in all_exprs as first
+            // element
+            pack(all_exprs, false.into())
```
AdamGS: not sure I understand the comment here

vortex-duckdb/src/projection.rs
myrrc: +1

vortex-sqllogictest/slt/duckdb/projection_expression_pushdown.slt
myrrc: +1

vortex-duckdb/src/projection.rs
myrrc: If we have projections, then column id is an index into projections, if we don't, then it's just an index. I agree the logic is ugly but this is what Duckdb supplies us.

vortex-duckdb/src/projection.rs
myrrc: cloned() isn't implemented for Option<Expression>

vortex-duckdb/src/projection.rs
myrrc: +1

PR #8320, opened 2026-06-09
feat[gpu]: export as Arrow device array binary

0ax1:
Turns out whilst Arrow device array supports binary, cuDF does not. Therefore, we don't currently exercise binary as part of the cuDF harness.

### DIFF
--- a/vortex-cuda/kernels/src/arrow_binary.cu
+++ b/vortex-cuda/kernels/src/arrow_binary.cu
@@ -1,0 +1,221 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#include "config.cuh"
+
+#include <limits.h>
+#include <stdint.h>
+
+namespace {
+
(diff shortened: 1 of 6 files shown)


vortex-cuda/kernels/src/arrow_binary.cu
```diff
@@ -0,0 +1,212 @@
...
+        uint8_t byte = 0;
+        const uint64_t first_bit = byte_idx * 8;
+        for (uint64_t bit_idx = 0; bit_idx < 8 && first_bit + bit_idx < len; bit_idx++) {
+            if (get_bit(input, input_offset + first_bit + bit_idx)) {
+                byte |= static_cast<uint8_t>(1u << bit_idx);
+            }
+        }
+        output[byte_idx] = byte;
+    }
+}
```
joseph-isaacs: 🤖 ⏰ 
Use a word level left shift

0ax1: done

robert3005 (approved):

PR #8321, opened 2026-06-09
Optimize slice for dict and minor changes in other arrays

AdamGS:
## Summary

Includes a few optimizations around array slicing:
1. In the general case, no need to allocate a full new Scalar to check what stats can be inherited.
2. BoolArray, Primitive and VarBinView can use existing unchecked constructors. 
3. For dict - specialize primitive codes and call them directly instead of going through optimization.

### DIFF
--- a/vortex-array/src/array/erased.rs
+++ b/vortex-array/src/array/erased.rs
@@ -47,13 +47,13 @@ use crate::arrays::VarBinView;
 use crate::buffer::BufferHandle;
 use crate::builders::ArrayBuilder;
 use crate::dtype::DType;
-use crate::dtype::Nullability;
 use crate::expr::stats::Precision;
 use crate::expr::stats::Stat;
 use crate::expr::stats::StatsProviderExt;
 use crate::matcher::Matcher;
 use crate::optimizer::ArrayOptimizer;
 use crate::scalar::Scalar;
+use crate::scalar::ScalarValue;
 use crate::stats::StatsSetRef;
 use crate::validity::Validity;
 
@@ -239,13 +239,10 @@ impl ArrayRef {
                     matches!(
                         stat,
                         Stat::IsConstant | Stat::IsSorted | Stat::IsStrictSorted
-                    ) && value.as_ref().as_exact().is_some_and(|v| {
-                        Scalar::try_new(DType::Bool(Nullability::NonNullable), Some(v.clone()))
-                            .vortex_expect("A stat that was expected to be a boolean stat was not")
-                            .as_bool()
-                            .value()
-                            .unwrap_or_default()
-                    })
+                    ) && value
+                        .as_ref()
+                        .as_exact()
+                        .is_some_and(|v| matches!(v, ScalarValue::Bool(true)))
                 }));
             });
         }
--- a/vortex-array/src/arrays/bool/compute/rules.rs
+++ b/vortex-array/src/arrays/bool/compute/rules.rs
@@ -47,12 +47,16 @@ impl ArrayParentReduceRule<Bool> for BoolMaskedValidityRule {
 
         // Merge the parent's validity mask into the child's validity
         // TODO(joe): make this lazy
-        Ok(Some(
-            BoolArray::new(
-                array.to_bit_buffer(),
-                array.validity()?.and(parent.validity()?)?,
-            )
-            .into_array(),
-        ))
+        // Safety:
+        // we know all elements are valid, the AND operation will fail if mismatched.
+        unsafe {
+            Ok(Some(
+                BoolArray::new_unchecked(
+                    array.to_bit_buffer(),
+                    array.validity()?.and(parent.validity()?)?,
+                )
+                .into_array(),
+            ))
+        }
     }
 }
--- a/vortex-array/src/arrays/bool/compute/rules.rs
+++ b/vortex-array/src/arrays/bool/compute/rules.rs
@@ -45,18 +45,15 @@ impl ArrayParentReduceRule<Bool> for BoolMaskedValidityRule {
             return Ok(None);
         }
 
+        let bit_buffer = array.to_bit_buffer();
         // Merge the parent's validity mask into the child's validity
         // TODO(joe): make this lazy
+        let validity = array.validity()?.and(parent.validity()?)?;
+
         // Safety:
         // we know all elements are valid, the AND operation will fail if mismatched.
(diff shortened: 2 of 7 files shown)


vortex-array/src/array/erased.rs
```diff
@@ -239,13 +239,10 @@ impl ArrayRef {
...
-                        Scalar::try_new(DType::Bool(Nullability::NonNullable), Some(v.clone()))
-                            .vortex_expect("A stat that was expected to be a boolean stat was not")
-                            .as_bool()
-                            .value()
-                            .unwrap_or_default()
-                    })
+                    ) && value
+                        .as_ref()
+                        .as_exact()
+                        .is_some_and(|v| matches!(v, ScalarValue::Bool(true)))
```
connortsui20, line 245: nice

vortex-array/src/arrays/bool/compute/rules.rs
```diff
@@ -47,12 +47,16 @@ impl ArrayParentReduceRule<Bool> for BoolMaskedValidityRule {
...
-                array.to_bit_buffer(),
-                array.validity()?.and(parent.validity()?)?,
-            )
-            .into_array(),
-        ))
+        // Safety:
+        // we know all elements are valid, the AND operation will fail if mismatched.
+        unsafe {
+            Ok(Some(
+                BoolArray::new_unchecked(
```
connortsui20: Nit: I usually like to push the unsafe as far down as possible (and make SAFETY all caps) but I wont block this

AdamGS: its just so ugly to have the unsafe inside the `Ok(Some(..))`

vortex-array/src/arrays/dict/compute/slice.rs
```diff
@@ -12,31 +13,32 @@ use crate::arrays::Constant;
...
-                    )
-                    .into_array(),
-                ))
-            } else {
-                Ok(Some(
-                    ConstantArray::new(Scalar::null(array.dtype().clone()), sliced_code.len())
-                        .into_array(),
-                ))
-            };
+            return slice_constant_code(array, code.scalar(), sliced_code.len());
```
connortsui20, line 42: question (not review): do we have the optimization where if we slice something it can BECOME a constant array? The only cases I can think of this happening would be in runend where we happen to slice into a run

vortex-array/src/arrays/dict/compute/slice.rs
```diff
@@ -45,6 +47,42 @@ impl SliceReduce for Dict {
...
+    code: &Scalar,
+    len: usize,
+) -> VortexResult<Option<ArrayRef>> {
+    let code = code.as_primitive().as_::<usize>();
+    if let Some(code) = code {
+        let values = array.values().slice(code..code + 1)?;
+        Ok(Some(
+            DictArray::new(ConstantArray::new(0u8, len).into_array(), values).into_array(),
+        ))
+    } else {
```
connortsui20, line 81: cant we just create a constant array immediately here? In fact, if the codes are constant we probably want a reduce rule to collapse into a Masked Constant array

vortex-array/src/arrays/bool/compute/rules.rs
connortsui20: fair enough

vortex-array/src/arrays/dict/compute/slice.rs
AdamGS: I don't think so, but maybe I was just looking in the wrong place.

connortsui20 (approved):
lgtm, though I am still wondering about those constant codes optimization, it feels like we should have optimized those away before we ever call slice? And then if slicing causes codes to be constant then thats an additional optimization. That can be in a separate PR

vortex-array/src/arrays/dict/compute/slice.rs
AdamGS: That sounds right, I'll look into it tomorrow.

vortex-array/src/arrays/bool/compute/rules.rs
joseph-isaacs: I personally agree with connor

vortex-array/src/arrays/dict/compute/slice.rs
```diff
@@ -45,6 +47,42 @@ impl SliceReduce for Dict {
...
+
+fn is_inheritable_true_slice_stat(stat: Stat, value: &Precision<ScalarValue>) -> bool {
+    matches!(
+        stat,
+        Stat::IsConstant | Stat::IsSorted | Stat::IsStrictSorted
+    ) && value
+        .as_ref()
+        .as_exact()
+        .is_some_and(|value| matches!(value, ScalarValue::Bool(true)))
+}
```
joseph-isaacs, line 68: Seems like this should be on the slice generic method/SliceArray ?

vortex-array/src/arrays/varbinview/compute/slice.rs
```diff
@@ -16,16 +16,20 @@ use crate::arrays::varbinview::BinaryView;
...
+                    array
+                        .views_handle()
+                        .slice_typed::<BinaryView>(range.clone()),
+                    Arc::clone(array.data_buffers()),
+                    array.dtype().clone(),
+                    array.validity()?.slice(range)?,
+                )
+                .into_array(),
+            ))
+        }
```
joseph-isaacs: can we move the unsafe scope down, maybe move out other calls?

vortex-array/src/arrays/dict/compute/slice.rs
AdamGS: It is, we specialize primitive here so we need to re-implement it.

vortex-array/src/arrays/dict/compute/slice.rs
AdamGS: can't do it in SliceReduce because it doesn't look at data and doesn't execute.

vortex-array/src/arrays/dict/compute/slice.rs
AdamGS: added a comment justifying it.

vortex-array/src/arrays/bool/compute/rules.rs
AdamGS: addressed

PR #8322, opened 2026-06-09
Optimize buffer ops

AdamGS:
## Summary

This PR includes a few optimization for buffer-level ops:
1. Empty buffer allocations are don't allocate anymore, using a shared static ptr.
2. Moved some alignment checks to direct comparison or mask-based checks instead of division.
3. More of a refactoring - but more of the alignment related checks are now functions on `Alignment`, instead of having less specific checks in different callsites.

After this PR is merged, I'll follow up and try to remove `bitvec` as a dependency, its currently used in a couple of pretty random places and I suspect there's nothing special about them compared to our own `BitBuffer`.

### DIFF
--- a/vortex-buffer/src/bit/buf_mut.rs
+++ b/vortex-buffer/src/bit/buf_mut.rs
@@ -650,6 +650,8 @@ impl FromIterator<bool> for BitBufferMut {
 
 #[cfg(test)]
 mod tests {
+    use rstest::rstest;
+
     use crate::BufferMut;
     use crate::bit::buf_mut::BitBufferMut;
     use crate::bitbuffer;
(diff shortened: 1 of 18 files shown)


robert3005:
I made https://github.com/vortex-data/vortex/pull/8162 that optimizes some of these code paths

AdamGS:
@robert3005 do you want to merge that first?

robert3005:
that would be ideal, the pr I made is a revival of an older pr already

AdamGS:
I'll review it

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -518,17 +520,16 @@ impl BitBufferMut {
                 self.buffer.as_mut_slice()[dst_byte + full_bytes] |= src_bytes[full_bytes] & mask;
             }
         } else {
-            // Use bitvec for unaligned bit copying.
-            let self_slice = self
```
robert3005, line 522: We have benchmarked this and bitvec is faster for anything that's bigger than 128 bits. I think we want to keep it

AdamGS: brought it back

robert3005 (approved):

PR #8324, opened 2026-06-09
Optimize `fsst_compress_offsets_overflow_i32` by 2X locally

AdamGS:
## Summary

Optimize the infamous `fsst_compress_offsets_overflow_i32` test by dropping the symbol table - instead compressing with an empty symbol table to get constant escapes.

Locally this change drops the test's runtime from 136s to 57s (in normal test profile), which should have a real impact on our CI runtimes.

This change is provided by the new fable-5 model.

### DIFF
--- a/encodings/fsst/src/tests.rs
+++ b/encodings/fsst/src/tests.rs
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-use rand::SeedableRng;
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

joseph-isaacs:
We could run this in release with debug assert too?

0ax1:
Thaaaank you

AdamGS:
@joseph-isaacs we can, last time I looked running tests in release was much slower and specializing one test is brittle, I think this already makes it much better.

PR #7722, opened 2026-04-29
add vortex-geo crate and WKB extension type

a10y:
## Summary

Part of the implementation of #7686 

This PR adds a new crate, `vortex-geo`, which will hold all extension types, custom layouts, and encodings necessary to store geospatial vector datasets in Vortex files.

The goal of this crate is to enable integration with DuckDB Spatial, GeoDataFusion, SedonaDB, and Iceberg v3 geo types.

This initial PR implements an extension type for the Well-Known Binary encoding (WKB). This encoding is the most common format for geospatial data for analytics, it's what both GeoParquet and DuckDB use to represent geometry types.

We also wire this into vortex DuckDB extension to support converting Geometry columns between Vortex and DuckDB formats.

## API Changes

Adds new crate `vortex-geo` with extension type `WellKnownBinary`, (`geo.wkb`)

Adds support for geometry columns to DuckDB as well, e.g. you can now do something like

```sql
SELECT building_id, building_name, ST_Area(geometry) AS area 
FROM read_vortex("buildings.vortex")
ORDER BY area DESC
LIMIT 10;
```

It won't be very performant yet, not until we support better layouts + stats for geometry.

## Testing

We have unit tests for metadata serde, round trip conversion between Vortex <> DuckDB geometry format, and an additional E2E test that demonstrates reading a Vortex file with geometry data and providing it to the DuckDB Spatial extension.

PR #8325, opened 2026-06-09
WKB import/export VTables to Arrow

a10y:
## Summary

Part of #7686 

Following up on #7722 , which added the WellKnownBinary extension type and handling for exporting to DuckDB vectors.

This PR adds support for import/export to Arrow for the extension type.

## Testing

Unit tests are added to exercise both code paths


AdamGS (approved):
Looks reasonable

PR #8327, opened 2026-06-10
Fix alias + cte reading from Vortex files

myrrc:
TODO


myrrc:
Depends on #8632

myrrc:
The upstream patch has been merged, but we need to migrate to duckdb 2.0 to take care of this. Will re-open again

PR #8333, opened 2026-06-10
Rename Validity::no_nulls to definitely_no_nulls and add execute_no_nulls

joseph-isaacs:
## Summary

**PR 1 of a 4-PR stack** preparing `Validity` for lazy validity arrays, where a `Validity::Array` may resolve to all-true/all-false once executed.

- Rename `Validity::no_nulls()` → `definitely_no_nulls()`. The boolean only proves the absence of nulls for `NonNullable`/`AllValid`; for a (lazy) `Array` variant, `false` means "unknown without compute", not "has nulls". The new name and doc comment make the conservative semantics explicit.
- Add `Validity::execute_no_nulls(length, ctx)`, the exact counterpart built on `execute_mask()`, for call sites that need a definitive answer.
- All 9 existing call sites are mechanically renamed; they are all conservative fast paths with fallbacks, except the CUDA ListView export `vortex_ensure!`, where a `false` answer is a hard error rather than a slow path — that one is switched to `execute_no_nulls` so a lazy all-valid child validity no longer spuriously fails the export.

## Stack

1. **→ this PR**: `no_nulls()` → `definitely_no_nulls()` + `execute_no_nulls()`
2. `mask_eq()` semantics fix for mixed variants
3. `definitely_all_invalid()` helper
4. Lazy `ScalarFn` array validity

## Checks

- `cargo build -p vortex-array -p vortex-zstd -p vortex-parquet-variant`
- `cargo nextest run -p vortex-array` (2962 passed)
- `cargo clippy -p vortex-array -p vortex-zstd -p vortex-parquet-variant --all-targets`
- `cargo +nightly fmt --all`
- Could not build/test `vortex-cuda` locally: its `vortex-nvcomp` build script requires the CUDA toolkit, which is unavailable in this environment.

https://claude.ai/code/session_01VPQ7dfZtijfrsjAipwXvEj

---
_Generated by [Claude Code](https://claude.ai/code/session_01VPQ7dfZtijfrsjAipwXvEj)_

### DIFF
--- a/vortex-array/src/validity.rs
+++ b/vortex-array/src/validity.rs
@@ -112,11 +112,28 @@ impl Validity {
         }
     }
 
-    /// Returns `true` if this validity guarantees no null values, i.e. it is either
+    /// Returns `true` if this validity *definitely* contains no null values, i.e. it is either
     /// [`Validity::NonNullable`] or [`Validity::AllValid`].
+    ///
+    /// Returning `false` does not prove the presence of nulls: a [`Validity::Array`] may still
+    /// resolve to all-valid once executed. Callers must treat `false` as "unknown without
+    /// compute" and either fall back to a null-handling path or execute the validity.
     #[inline]
-    pub fn no_nulls(&self) -> bool {
+    pub fn definitely_no_nulls(&self) -> bool {
         matches!(self, Self::NonNullable | Self::AllValid)
+    }
+
+    /// Returns whether this validity contains no null values, executing the validity array if
+    /// necessary.
+    ///
+    /// This is the exact counterpart to [`Self::definitely_no_nulls`]: use it when the caller
+    /// needs a definitive answer rather than a cheap, conservative one.
+    pub fn execute_no_nulls(&self, length: usize, ctx: &mut ExecutionCtx) -> VortexResult<bool> {
+        match self {
+            Self::NonNullable | Self::AllValid => Ok(true),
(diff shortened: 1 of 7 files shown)


vortex-array/src/validity.rs
```diff
@@ -112,13 +112,30 @@ impl Validity {
...
 
-    /// Returns `true` if this validity guarantees no null values, i.e. it is either
+    /// Returns `true` if this validity *definitely* contains no null values, i.e. it is either
     /// [`Validity::NonNullable`] or [`Validity::AllValid`].
+    ///
+    /// Returning `false` does not prove the presence of nulls: a [`Validity::Array`] may still
+    /// resolve to all-valid once executed. Callers must treat `false` as "unknown without
+    /// compute" and either fall back to a null-handling path or execute the validity.
     #[inline]
-    pub fn no_nulls(&self) -> bool {
```
AdamGS, line 118: `has_no_nulls`?

joseph-isaacs: This is an approximate function. Do you think `has_no_nulls` says that?

AdamGS: They all sound vague to me, does it really need to be a function vs allowing callers to just have it explicitly there?

joseph-isaacs: Do you have a suggestion.

the {maybe/definitely} convention for ternary logic is pretty common

AdamGS (approved):

PR #8334, opened 2026-06-10
Fix Validity::mask_eq semantics for mixed variants

joseph-isaacs:
## Summary

**PR 2 of a 4-PR stack** (stacked on #8333) preparing `Validity` for lazy validity arrays.

`mask_eq` previously returned `false` for any mixed-variant pairing without executing — e.g. a `Validity::Array` that resolves to all-true compared against `Validity::AllValid`. With lazy validity arrays, unresolved `Array` variants frequently hold constant masks, making this silently wrong rather than merely conservative.

### DIFF
--- a/vortex-array/src/validity.rs
+++ b/vortex-array/src/validity.rs
@@ -112,11 +112,28 @@ impl Validity {
         }
     }
 
-    /// Returns `true` if this validity guarantees no null values, i.e. it is either
+    /// Returns `true` if this validity *definitely* contains no null values, i.e. it is either
     /// [`Validity::NonNullable`] or [`Validity::AllValid`].
+    ///
+    /// Returning `false` does not prove the presence of nulls: a [`Validity::Array`] may still
+    /// resolve to all-valid once executed. Callers must treat `false` as "unknown without
+    /// compute" and either fall back to a null-handling path or execute the validity.
     #[inline]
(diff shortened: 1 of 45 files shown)


joseph-isaacs:
shall we just remove mask_eq?

vortex-array/src/validity.rs
```diff
@@ -263,18 +263,22 @@ impl Validity {
...
-                Ok(a == b)
-            }
-            _ => Ok(false),
+            // Fast paths that avoid executing: constant variants with known-equal masks.
+            (
+                Validity::NonNullable | Validity::AllValid,
+                Validity::NonNullable | Validity::AllValid,
+            )
+            | (Validity::AllInvalid, Validity::AllInvalid) => Ok(true),
+            _ => Ok(self.execute_mask(length, ctx)? == other.execute_mask(length, ctx)?),
```
AdamGS: What's the difference between `execute_mask` and `execute::<Mask>`?

AdamGS: if the former is preferable, seems like removing the ability to do `execute::<Mask>` is also desirable here

joseph-isaacs: `Mask::execute_mask` tried to avoid using the `ArrayRef::execute` loop. I think we want both.

AdamGS (approved):

PR #8336, opened 2026-06-10
Make ScalarFn array validity lazy when the function defines a validity expression

joseph-isaacs:
## Summary

Lazy validity for `ScalarFn` arrays. (Originally part of a 4-PR stack; the prerequisites — `definitely_no_nulls`/`execute_no_nulls` in #8333, the `mask_eq` semantics fix in #8334, and `definitely_all_null` — have all since landed on `develop`, so this PR is now a single standalone commit rebased onto current develop.)

Previously `ValidityVTable<ScalarFn>` always **eagerly executed** the validity expression via the legacy session. Now, when the scalar function provides a validity expression over its inputs, the expression is converted into a **lazy ScalarFn array DAG** instead:

- `Literal` nodes → `ConstantArray`
- `ArrayExpr` leaves → unwrap to the child array they hold
- interior nodes → lazy `ScalarFn` arrays via `Array::<ScalarFn>::try_new_with_len`
- constant results are folded back into `AllValid`/`AllInvalid` via `child_to_validity`

### Why the eager path remains for some functions

Functions that don't define a validity expression (Kleene `and`/`or`, where validity depends on the computed *values*) keep the eager path. The erased fallback for these is `is_not_null(expr)` — self-referential: lazily materializing it means resolving the validity of the inner node, which spawns another `is_not_null` DAG over a fresh copy of the same node, recursing without ever shrinking. This manifested as a stack overflow in element-wise execution paths (`execute_scalar` → `is_invalid` → `validity()`), caught by `test_bool_consistency`. The new `ScalarFnRef::validity_opt` exposes whether a function defines its own validity expression so the vtable can pick the right path.

## Checks

- `cargo nextest run -p vortex-array` (3197 passed)
- `cargo nextest run -p vortex-file -p vortex-layout -p vortex-scan -p vortex-btrblocks -p vortex-compressor -p vortex-datafusion -p vortex-fastlanes -p vortex-runend -p vortex-sparse -p vortex-fsst -p vortex-alp -p vortex-zstd -p vortex-pco -p vortex-datetime-parts -p vortex-ipc` (1429 passed)
- `cargo clippy -p vortex-array --all-targets`, `cargo +nightly fmt --check`
- Could not run `vortex-cuda`, `vortex-duckdb`, or lance-dependent benches locally (CUDA toolkit / DuckDB source download / `protoc` unavailable in the sandbox) — CI covers them.

https://claude.ai/code/session_01VPQ7dfZtijfrsjAipwXvEj

### DIFF
--- a/vortex-array/src/scalar_fn/erased.rs
+++ b/vortex-array/src/scalar_fn/erased.rs
@@ -132,10 +132,20 @@ impl ScalarFnRef {
 
     /// Transforms the expression into one representing the validity of this expression.
     pub fn validity(&self, expr: &Expression) -> VortexResult<Expression> {
-        Ok(self.0.validity(expr)?.unwrap_or_else(|| {
+        Ok(self.validity_opt(expr)?.unwrap_or_else(|| {
             // TODO(ngates): make validity a mandatory method on VTable to avoid this fallback.
             IsNotNull.new_expr(EmptyOptions, [expr.clone()])
         }))
+    }
+
+    /// Transforms the expression into one representing the validity of this expression,
+    /// returning `None` if the function does not define a validity expression.
+    ///
+    /// When `None` is returned, the validity can only be determined by executing the
+    /// expression itself (e.g. Kleene logic `and`/`or`), and [`Self::validity`] falls back to
+    /// `is_not_null` over the expression.
+    pub fn validity_opt(&self, expr: &Expression) -> VortexResult<Option<Expression>> {
+        self.0.validity(expr)
     }
 
     /// Execute the expression given the input arguments.
--- a/vortex-array/src/arrays/scalar_fn/vtable/validity.rs
+++ b/vortex-array/src/arrays/scalar_fn/vtable/validity.rs
@@ -8,13 +8,16 @@ use crate::ArrayRef;
 use crate::ExecutionCtx;
 use crate::IntoArray;
 use crate::VortexSessionExecute;
+use crate::array::Array;
 use crate::array::ArrayView;
 use crate::array::ValidityVTable;
+use crate::array::child_to_validity;
 use crate::arrays::ConstantArray;
 use crate::arrays::scalar_fn::ScalarFnArrayExt;
 use crate::arrays::scalar_fn::vtable::ArrayExpr;
 use crate::arrays::scalar_fn::vtable::FakeEq;
 use crate::arrays::scalar_fn::vtable::ScalarFn;
+use crate::dtype::Nullability;
 use crate::expr::Expression;
 use crate::expr::lit;
 use crate::legacy_session;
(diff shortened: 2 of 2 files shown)


joseph-isaacs:
Investigated the CodSpeed `varbinview_zip_*` regressions — they are a codegen-layout artifact, not a real cost of this change:

1. **The changed code never runs in these benchmarks.** Instrumenting `ScalarFn::validity` with a counter shows **0 calls** during `varbinview_zip_block_mask`/`varbinview_zip_fragmented_mask` (vs ~47k calls in the bool-consistency tests, confirming the instrumentation works). The benches go straight through the dedicated `ZipKernel for VarBinView` and never query a `ScalarFn` array's validity.

2. **Bisecting the two files in this PR** reproduces the full ~19% delta locally only when `arrays/scalar_fn/vtable/validity.rs` changes — the file whose code is never executed here. The `scalar_fn/erased.rs` change alone shows no delta.

3. **`codegen-units=1` erases the regression entirely** (branch 3: 505µs median vs branch 3 + this PR's files: 494µs). With default codegen units, adding code to that module shifts LLVM's CGU partitioning and changes inlining in the unrelated hot zip loop, which CodSpeed's instruction-counting simulation faithfully reports.

The same flavor of noise shows in the rest of the stack: the `bitwise_not_vortex_buffer_mut` and `chunked_*_canonical_into` benchmarks flip between ±20–47% on adjacent PRs that don't touch vortex-buffer or the chunked builders at all.

I can't acknowledge regressions on the CodSpeed dashboard from this session — that needs someone with dashboard access.

https://claude.ai/code/session_01VPQ7dfZtijfrsjAipwXvEj

---

vortex-array/src/scalar_fn/erased.rs
```diff
@@ -134,12 +134,22 @@ impl ScalarFnRef {
 
     /// Transforms the expression into one representing the validity of this expression.
     pub fn validity(&self, expr: &Expression) -> VortexResult<Expression> {
-        Ok(self.0.validity(expr)?.unwrap_or_else(|| {
+        Ok(self.validity_opt(expr)?.unwrap_or_else(|| {
```
robert3005, line 135: do you want to remove the TODO?

joseph-isaacs:
**CodSpeed failure triage for head `70cdf0c`** — not caused by this PR:

- All 6 flagged regressions are **WallTime-mode** measurements of raw SIMD lane kernels (`arrow_checked_add_u32_neon`, `mul_u64/i64_nonnull_neon`, `multiply_shapes_neon`, `dbp_split_kernel_mixed_null_neon`, `mul_i32_nonnull_avx512`). Those benches exercise the vectorized arithmetic kernels directly, below the array layer — they never query a `ScalarFn` array's validity, which is the only behavior this PR changes (2 files, `scalar_fn` only).
- CodSpeed's own report carries two warnings that explain the numbers: *"Unknown Walltime execution environment"* (walltime on shared hosted runners is inconsistent) and *"Different runtime environments detected"* (BASE and HEAD were measured on different machines — the mix of NEON and AVX512 rows in one comparison points the same way).
- The deterministic **Simulation** mode reports **0 regressions across 2,244 benchmarks** on this head.

There is no code fix to port; the check needs a re-run or a dashboard acknowledgment, neither of which I can trigger from this session — flagging for someone with CodSpeed access. Everything else on this head is green (3,648 vortex-array tests pass locally after adapting the commit to the new `Expression` enum API).

https://claude.ai/code/session_01VPQ7dfZtijfrsjAipwXvEj

---

PR #8339, opened 2026-06-10
Add arrow import/export for vortex-json

AdamGS:
## Summary

Adds import/export from `vortex-json` to Arrow's JSON canonical extension type.


vortex-json/src/arrow.rs
```diff
@@ -0,0 +1,140 @@
...
+        _session: &ArrowSession,
+    ) -> VortexResult<Option<Field>> {
+        let DType::Extension(ext_dtype) = dtype else {
+            return Ok(None);
+        };
+        if !ext_dtype.is::<Json>() {
+            return Ok(None);
+        }
+
+        let mut field = Field::new(name, DataType::Utf8, dtype.is_nullable());
```
a10y: do we always want Utf8? why not stringview?

in my WKB branch I just delegated this to the session

vortex-json/src/arrow.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn arrow_ext_id(&self) -> Id {
+        *ARROW_JSON
+    }
+
+    fn from_arrow_field(&self, field: &Field) -> VortexResult<Option<DType>> {
+        if !has_valid_json_extension(field) {
+            return Ok(None);
+        }
+
+        let storage_dtype = DType::from_arrow(field);
```
a10y: we should not use DType::from_arrow anymore. should probably be deprecated actually. I think you can just use Utf8 directly in this case?

vortex-json/src/arrow.rs
```diff
@@ -0,0 +1,140 @@
...
+        dtype: &DType,
+    ) -> VortexResult<ArrowImport> {
+        let DType::Extension(ext_dtype) = dtype else {
+            return Ok(ArrowImport::Unsupported(array));
+        };
+        if !ext_dtype.is::<Json>() || !has_valid_json_extension(field) {
+            return Ok(ArrowImport::Unsupported(array));
+        }
+
+        let storage = ArrayRef::from_arrow(array.as_ref(), field.is_nullable())?;
```
a10y, line 134: we should deprecate this as well and use the session

AdamGS: mind if I do it in a followup? I'll need to wire the session here

robert3005 (approved):

PR #8341, opened 2026-06-10
Push down some expressions to Dict layout reader's cached values

myrrc:
When we access values of Dict layout reader, it canonicalizes them and stores
them in a SharedArray. This means we always pay the cost of canonicalization
which in turn means we can't do https://github.com/vortex-data/vortex/issues/8310 .

In order to solve this issue, we need to apply some expressions to the values
array before canonicalizing it. However, we can't push down arbitrary
expressions as it may be beneficial to apply them over canonicalized array.

One example of such expressions is LIKE over a Dict array with few codes used.
Applying LIKE to whole values array is not beneficial.

This PR adds a hardcoded internal is_negative_cost estimation for expressions
that we want to push before canonicalization. A hint for these are expressions
which don't depend on individual input size. As an example, for every string,
len(string) doesn't read the string itself but reads the metadata and thus is
O(1) on individual input.

We don't push down fallible (like cast) or null sensitive (like IS NULL)
expressions as well because we want to propagate the errors at call site
rather than upfront.

### DIFF
--- a/vortex-array/src/expr/analysis/annotation.rs
+++ b/vortex-array/src/expr/analysis/annotation.rs
@@ -42,6 +42,25 @@ pub fn descendent_annotations<A: AnnotationFn>(
     let mut visitor = AnnotationVisitor {
         annotations: Default::default(),
         annotate,
+        propagate_up: true,
+    };
+    expr.accept(&mut visitor).vortex_expect("Infallible");
+    visitor.annotations
+}
+
+/// Walk the expression tree and annotate each expression with zero or more
+/// annotations.
+///
+/// Returns a map of each expression to all annotations. Annotations of
+/// children are not propagated to parents.
+pub fn direct_annotations<A: AnnotationFn>(
+    expr: &Expression,
+    annotate: A,
+) -> Annotations<'_, A::Annotation> {
+    let mut visitor = AnnotationVisitor {
+        annotations: Default::default(),
+        annotate,
+        propagate_up: false,
     };
     expr.accept(&mut visitor).vortex_expect("Infallible");
     visitor.annotations
@@ -50,6 +69,7 @@ pub fn descendent_annotations<A: AnnotationFn>(
 struct AnnotationVisitor<'a, A: AnnotationFn> {
     annotations: Annotations<'a, A::Annotation>,
     annotate: A,
+    propagate_up: bool,
 }
 
 impl<'a, A: AnnotationFn> NodeVisitor<'a> for AnnotationVisitor<'a, A> {
@@ -70,6 +90,9 @@ impl<'a, A: AnnotationFn> NodeVisitor<'a> for AnnotationVisitor<'a, A> {
     }
 
     fn visit_up(&mut self, node: &'a Expression) -> VortexResult<TraversalOrder> {
+        if !self.propagate_up {
+            return Ok(TraversalOrder::Continue);
+        }
         let child_annotations = node
             .children()
             .iter()
--- a/vortex-array/src/expr/transform/partition.rs
+++ b/vortex-array/src/expr/transform/partition.rs
@@ -3,6 +3,7 @@
 
 use std::fmt::Display;
 use std::fmt::Formatter;
+use std::hash::Hash;
 
 use itertools::Itertools;
 use vortex_error::VortexExpect;
@@ -49,11 +50,22 @@ where
 {
     // Annotate each expression with the annotations that any of its descendent expressions have.
     let annotations = descendent_annotations(&expr, annotate_fn);
+    partition_annotations(expr.clone(), scope, annotations)
+}
 
+pub fn partition_annotations<A>(
+    expr: Expression,
+    scope: &DType,
+    annotations: Annotations<A>,
+) -> VortexResult<PartitionedExpr<A>>
+where
+    A: Display + Clone + Eq + Hash,
+    FieldName: From<A>,
+{
     // Now we split the original expression into sub-expressions based on the annotations, and
     // generate a root expression to re-assemble the results.
-    let mut splitter = StructFieldExpressionSplitter::<A::Annotation>::new(&annotations);
-    let root = expr.clone().rewrite(&mut splitter)?.value;
+    let mut splitter = StructFieldExpressionSplitter::<A>::new(&annotations);
+    let root = expr.rewrite(&mut splitter)?.value;
 
     let mut partitions = Vec::with_capacity(splitter.sub_expressions.len());
     let mut partition_annotations = Vec::with_capacity(splitter.sub_expressions.len());
--- a/vortex-array/src/scalar_fn/mod.rs
+++ b/vortex-array/src/scalar_fn/mod.rs
@@ -9,6 +9,10 @@
 
 use vortex_session::registry::Id;
 
+use crate::scalar_fn::fns::byte_length::ByteLength;
+use crate::scalar_fn::fns::get_item::GetItem;
+use crate::scalar_fn::fns::literal::Literal;
+
 mod vtable;
 pub use vtable::*;
 
@@ -47,4 +51,18 @@ mod sealed {
 
     /// This can be the **only** implementor for [`super::typed::DynScalarFn`].
     impl<V: ScalarFnVTable> Sealed for TypedScalarFnInstance<V> {}
+}
+
+/// A scalar function has a negative cost if applying it to an array and
+/// canonicalizing is cheaper than canonicalizing an array and applying it.
+///
+/// Example of negative cost expressions are byte_length() and get_item() since
+/// they don't depend on input size.
+///
+/// Example of non-negative cost expression is like() as it's linear over
+/// individual input.
+pub fn is_negative_cost(id: ScalarFnId) -> bool {
+    id == ScalarFnVTable::id(&ByteLength)
+        || id == ScalarFnVTable::id(&GetItem)
+        || id == ScalarFnVTable::id(&Literal)
 }
--- a/vortex-layout/src/layouts/dict/reader.rs
+++ b/vortex-layout/src/layouts/dict/reader.rs
@@ -19,9 +19,16 @@ use vortex_array::arrays::SharedArray;
 use vortex_array::builtins::ArrayBuiltins;
 use vortex_array::dtype::DType;
 use vortex_array::dtype::FieldMask;
+use vortex_array::dtype::Nullability;
 use vortex_array::expr::Expression;
+use vortex_array::expr::direct_annotations;
+use vortex_array::expr::is_root;
+use vortex_array::expr::label_tree;
+use vortex_array::expr::pack;
 use vortex_array::expr::root;
+use vortex_array::expr::transform::partition_annotations;
 use vortex_array::optimizer::ArrayOptimizer;
+use vortex_array::scalar_fn::is_negative_cost;
 use vortex_error::VortexError;
 use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
@@ -101,10 +108,7 @@ impl DictReader {
                     )
                     .vortex_expect("must construct dict values array evaluation")
                     .map_err(Arc::new)
-                    .map(move |array| {
-                        let array = array?;
-                        Ok(SharedArray::new(array).into_array())
-                    })
+                    .map(move |array| Ok(SharedArray::new(array?).into_array()))
                     .boxed()
                     .shared()
             })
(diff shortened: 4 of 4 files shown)


vortex-array/src/scalar_fn/mod.rs
```diff
@@ -48,3 +48,18 @@ mod sealed {
...
+ *
+ * Example of negative cost expressions are byte_length() and get_item() since
+ * they don't depend on input size.
+ *
+ * Example of non-negative cost expression is like()
+ */
+pub fn is_negative_cost(id: ScalarFnId) -> bool {
+    id == Id::new_static("vortex.byte_length")
+        || id == Id::new_static("vortex.get_item")
+        || id == Id::new_static("vortex.literal")
```
joseph-isaacs: lets use the typed version of these ids?

vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -155,6 +159,49 @@ impl DictReader {
...
+            if is_negative_cost(node.id())
+                && references_root(&node)
+                && !fallible.get(&node).copied().unwrap_or(true)
+                && !null_sensitive.get(&node).copied().unwrap_or(true)
+            {
+                inner = Some(node);
+                Ok(Transformed {
+                    value: root(),
+                    changed: true,
+                    order: TraversalOrder::Skip,
```
joseph-isaacs: we have a partition function for exactly this!

myrrc: You can't implement this using partition<> because it pushes children annotations up to the parents. This breaks on

```
let expr = cast(byte_length(root()), dtype);
// expect inner = byte_length(root), outer = cast(root, dtype)
```

But instead we get

```
inner = cast(...)
```

because cast inherits the annotation from a child. This is also an implementation that's harder to reason about. Let's leave this one.

joseph-isaacs: Do you know which part incorrect with this method?

myrrc: partition<> splits an expression into non-overlapping parts, each element gets into one group, union of groups is the original one. This task requires a single split of expressions where a split point is replaced with inner. Inner and outer expressions are not non-overlapping parts.
This is a different task to solve.
partition<> is implemented correctly but isn't suitable for our task.

joseph-isaacs: I think we can modify `partition` to take a annotated expr tree instead of taking an annotation function then the logic for annotation can live here

myrrc: I've added the comment about this.

vortex-array/src/expr/transform/partition.rs
```diff
@@ -20,12 +20,62 @@ use crate::expr::analysis::AnnotationFn;
...
+/// canonicalizing.
+/// right is the optional inner part that we want to apply to array before
+/// canonicalizing.
+///
+/// We want to push to array only if expression has a negative cost, is
+/// infallible and null-insensitive.
+///
+/// This is a specialized version of partition() which we probably want to merge
+/// into partition<> at some point
+pub fn split_expression_for_pushdown(expr: Expression) -> (Option<Expression>, Option<Expression>) {
```
joseph-isaacs:
```suggestion
TODO(..): unify this with the expression partitioning logic, by extending the expr annotator
pub fn split_expression_for_pushdown(expr: Expression) -> (Option<Expression>, Option<Expression>) {
```

vortex-array/src/scalar_fn/mod.rs
```diff
@@ -48,3 +52,18 @@ mod sealed {
...
+
+/*
+ * A scalar function has a negative cost if applying it to an array and
+ * canonicalizing is cheaper than canonicalizing an array and applying it.
+ *
+ * Example of negative cost expressions are byte_length() and get_item() since
+ * they don't depend on input size.
+ *
+ * Example of non-negative cost expression is like()
+ */
```
joseph-isaacs: use ///

vortex-array/src/expr/transform/partition.rs
```diff
@@ -20,12 +20,62 @@ use crate::expr::analysis::AnnotationFn;
...
 use crate::expr::root;
 use crate::expr::traversal::NodeExt;
 use crate::expr::traversal::NodeRewriter;
 use crate::expr::traversal::Transformed;
 use crate::expr::traversal::TraversalOrder;
+use crate::scalar_fn::is_negative_cost;
+
+fn references_root(expr: &Expression) -> bool {
+    is_root(expr) || expr.children().iter().any(references_root)
+}
```
joseph-isaacs: This is slow!

vortex-array/src/expr/transform/partition.rs
```diff
@@ -348,4 +407,61 @@ mod tests {
...
+        assert_eq!(outer, None);
+        assert_eq!(inner, Some(byte_length(root())));
+        join_split_expr(&expr, outer, inner);
+    }
+
+    #[test]
+    fn split_expr_no_pushdown() {
+        // We can push down lit(), but it we replace
+        // lit() with root(), the semantics change.
+        let expr = like(root(), lit(1u64));
```
joseph-isaacs: the types are wrong

vortex-array/src/expr/transform/partition.rs
```diff
@@ -20,12 +20,62 @@ use crate::expr::analysis::AnnotationFn;
...
+pub fn split_expression_for_pushdown(expr: Expression) -> (Option<Expression>, Option<Expression>) {
+    let labelled_expr = expr.clone();
+    let fallible = label_is_fallible(&labelled_expr);
+    let null_sensitive = label_null_sensitive(&labelled_expr);
+    let mut inner: Option<Expression> = None;
+
+    let outer = expr
+        .transform_down(|node| {
+            if is_negative_cost(node.id())
+                && references_root(&node)
```
joseph-isaacs: you don't need this

vortex-array/src/expr/transform/partition.rs
joseph-isaacs, line 1: Please can you use partition

vortex-array/src/scalar_fn/mod.rs
```diff
@@ -48,3 +52,16 @@ mod sealed {
...
     impl<V: ScalarFnVTable> Sealed for TypedScalarFnInstance<V> {}
 }
+
+/// A scalar function has a negative cost if applying it to an array and
+/// canonicalizing is cheaper than canonicalizing an array and applying it.
+///
+/// Example of negative cost expressions are byte_length() and get_item() since
+/// they don't depend on input size.
+///
+/// Example of non-negative cost expression is like()
```
joseph-isaacs: why is like not negative cost?

vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -229,13 +227,18 @@ impl LayoutReader for DictReader {
...
         let codes_eval = self
             .codes
             .projection_evaluation(row_range, &root(), mask)
             .map_err(|err| err.with_context("While evaluating projection on codes"))?;
-        let expr = expr.clone();
 
+        let (expr_outer, expr_inner) = split_expression_for_pushdown(expr.clone(), self.dtype())?;
+
+        let values_eval = if let Some(inner) = expr_inner {
+            self.values_eval(inner)
```
joseph-isaacs: i think we might want to avoid the shared array value cache if the expr is negative cost?

vortex-array/src/expr/transform/partition.rs
```diff
@@ -19,13 +20,50 @@ use crate::expr::analysis::Annotation;
...
+
+/// Split expression into two parts:
+///
+/// left is the outer part that we want to apply to array after canonicalizing.
+/// right is the optional inner part that we want to apply to array before
+/// canonicalizing.
+///
+/// We want to push to array only if expression has a negative cost, is
+/// infallible and null-insensitive.
+pub fn split_expression_for_pushdown(
```
joseph-isaacs: This function should not live here

vortex-array/src/expr/transform/partition.rs
```diff
@@ -19,13 +20,50 @@ use crate::expr::analysis::Annotation;
...
+) -> VortexResult<(Expression, Option<Expression>)> {
+    let references_root = label_tree(&expr, is_root, |acc, &child| acc | child);
+    let annotations = annotations(&expr, |expr| {
+        let signature = expr.signature();
+        if !signature.is_fallible()
+            && !signature.is_null_sensitive()
+            && is_negative_cost(expr.id())
+            && references_root.get(&expr).copied().unwrap_or(true)
+        {
+            vec![""] // we want a single annotation so an empty literal is fine
```
joseph-isaacs: Can you either use an enum or a name const/static variable

vortex-array/src/expr/analysis/annotation.rs
```diff
@@ -84,3 +101,19 @@ impl<'a, A: AnnotationFn> NodeVisitor<'a> for AnnotationVisitor<'a, A> {
...
+    fn visit_down(&mut self, node: &'a Self::NodeTy) -> VortexResult<TraversalOrder> {
+        let annotations = (self.annotate)(node);
+        if annotations.is_empty() {
+            Ok(TraversalOrder::Continue)
+        } else {
+            self.annotations
+                .entry(node)
+                .or_default()
+                .extend(annotations);
+            Ok(TraversalOrder::Skip)
```
joseph-isaacs: Why skip here?

vortex-array/src/expr/analysis/annotation.rs
```diff
@@ -47,6 +47,23 @@ pub fn descendent_annotations<A: AnnotationFn>(
...
+) -> Annotations<'_, A::Annotation> {
+    let mut visitor = DirectAnnotationVisitor {
+        annotations: Default::default(),
+        annotate,
+    };
+    expr.accept(&mut visitor).vortex_expect("Infallible");
+    visitor.annotations
+}
+
+struct DirectAnnotationVisitor<'a, A: AnnotationFn> {
```
joseph-isaacs: doc str please

vortex-array/src/expr/analysis/annotation.rs
```diff
@@ -47,6 +47,23 @@ pub fn descendent_annotations<A: AnnotationFn>(
     visitor.annotations
 }
 
+pub fn annotations<A: AnnotationFn>(
```
joseph-isaacs: The name is wrong, it uses a specific direction annotation

vortex-array/src/expr/analysis/annotation.rs
```diff
@@ -84,3 +101,19 @@ impl<'a, A: AnnotationFn> NodeVisitor<'a> for AnnotationVisitor<'a, A> {
         Ok(TraversalOrder::Continue)
     }
 }
+
+impl<'a, A: AnnotationFn> NodeVisitor<'a> for DirectAnnotationVisitor<'a, A> {
```
joseph-isaacs: explain the semantics. 

i think this is odd since it will stop when you first annotate a tree it will not annotate the rest of the subtree, that might be what you want.

vortex-array/src/expr/transform/partition.rs
```diff
@@ -19,13 +20,50 @@ use crate::expr::analysis::Annotation;
...
+        } else {
+            vec![]
+        }
+    });
+    let partition = partition_annotations(expr.clone(), dtype, annotations)?;
+    if partition.partitions.is_empty() {
+        Ok((partition.root, None))
+    } else {
+        Ok((partition.root, Some(partition.partitions[0].clone())))
```
joseph-isaacs: This is fishy. 

What does the zeroth partition contain, what if there are others?

vortex-array/src/scalar_fn/mod.rs
myrrc: It's linear in terms of input since it needs to evaluate it.

vortex-array/src/expr/transform/partition.rs
myrrc: There won't be others since we have only one label.

vortex-layout/src/layouts/dict/reader.rs
myrrc: We can't because the semantics of the shared future break then (there are errors in CI about "unsupported type for normalify")

vortex-array/src/scalar_fn/mod.rs
joseph-isaacs: Update the comment with why

joseph-isaacs: What else shall we put here?

vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -155,6 +157,40 @@ impl DictReader {
...
+            vec![ANNOTATION]
+        } else {
+            vec![]
+        }
+    });
+    let partition = partition_annotations(expr.clone(), dtype, annotations)?;
+    if partition.partitions.is_empty() {
+        Ok((partition.root, None))
+    } else {
+        Ok((partition.root, Some(partition.partitions[0].clone())))
```
joseph-isaacs, line 197: debug assert

vortex-layout/src/layouts/dict/reader.rs
joseph-isaacs: I am saying you need to get the uncompressed uncached values if the expr is negative cost

vortex-layout/src/layouts/dict/reader.rs
```diff
@@ -229,13 +265,18 @@ impl LayoutReader for DictReader {
...
     ) -> VortexResult<BoxFuture<'static, VortexResult<ArrayRef>>> {
         // TODO: fix up expr partitioning with fallible & null sensitive annotations
-        let values_eval = self.values_array();
         let codes_eval = self
             .codes
             .projection_evaluation(row_range, &root(), mask)
             .map_err(|err| err.with_context("While evaluating projection on codes"))?;
-        let expr = expr.clone();
 
+        let (expr_outer, expr_inner) = split_expression_for_pushdown(expr.clone(), self.dtype())?;
```
joseph-isaacs, line 283:
```suggestion
        let (expr_outer, expr_inner) = split_expression_for_negative_cost_pushdown(expr.clone(), self.dtype())?;
```

vortex-layout/src/layouts/dict/reader.rs
myrrc: No, we want to have cached values if we have already computed this expression.

values_eval() is using a different cache (self.values_evals) with values_array() (self.values_array) so there is no regression if we first hit projection_evaluation() without an expression that can be pushed down and then with one.

myrrc: Ok, I see, the regression's is in values_array_uncanonical.

myrrc:
The implementation with annotations and packing is slower than the original I've proposed. On #8303 the difference is 150ms (my solution) vs 250-300ms (the solution you proposed with annotations).

myrrc:
Replacing values_eval with uncompressed values doesn't change the performance.

myrrc:
For easier searching, the optimal solution is https://github.com/vortex-data/vortex/commit/ed16fc6ed7b8abbf91eaf4cc6267656294967802. I agree we can memoize references_root to avoid N^2, and use uncompressed values instead of values_eval().

joseph-isaacs (approved):

PR #8342, opened 2026-06-10
feat(vortex-geo): native Point extension type and GeoDistance scalar function

HarukiMoriarty:
## Summary

This PR adds a native point type to `vortex-geo`. Points are by far the most common geometry in analytical datasets, and a columnar representation makes their coordinates directly accessible without parsing WKB.

It also adds the scalar function: point-to-point distance with PostGIS `ST_Distance` semantics (planar/Euclidean, results in CRS units).

## API Changes

Adds to `vortex-geo`, all registered through `vortex_geo::initialize`:

- Extension type `Point` (`vortex.geo.point`): a location stored as `Struct<x, y, z?, m?>` of non-nullable `f64`, where `z?` is an optional elevation and `m?` an optional measure. 
- `Coordinate`: the internal value a point scalar unpacks to.
- Scalar function `GeoDistance` (`vortex.geo.distance`): per-row distance between two equal-length point columns; either or both operands may be constant, in which case the query point is decoded once and broadcast.

## Testing

Unit tests cover dtype validation for every GeoArrow dimension (and rejection of invalid storage), round-tripping a point column through scalar execution back to the original coordinates, WKT display for all four dimensions, and distance over all operand shapes: column-to-constant (either side), column-to-column, and constant-to-constant.

### DIFF
--- a/vortex-geo/src/extension/coordinate.rs
+++ b/vortex-geo/src/extension/coordinate.rs
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-//! Coordinate building blocks for geometry extension types: the `Struct<x, y, z?, m?>` storage,
-//! its [`Dimension`], and the decoded [`Coordinate`] value.
+//! Coordinate building blocks for geometry extension types: the
+//! `Struct<x: f64, y: f64, z?: f64, m?: f64>` storage, its [`Dimension`], and the decoded
+//! [`Coordinate`] value.
 //!
 //! The coordinate fields, where `?` marks an optional field, are:
 //! - `x` — longitude or easting
@@ -160,6 +161,11 @@ pub(crate) fn xy_columns(
         .storage_array()
         .clone()
         .execute::<StructArray>(ctx)?;
+    vortex_ensure!(
+        !storage.dtype().is_nullable(),
+        "point storage must be non-nullable to read unmasked x/y, was {}",
+        storage.dtype()
+    );
     let xs = storage
         .unmasked_field_by_name("x")?
         .clone()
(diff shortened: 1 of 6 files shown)


vortex-geo/src/extension/coordinate.rs
```diff
@@ -0,0 +1,195 @@
...
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<(PrimitiveArray, PrimitiveArray)> {
+    let storage = points
+        .clone()
+        .execute::<ExtensionArray>(ctx)?
+        .storage_array()
+        .clone()
+        .execute::<StructArray>(ctx)?;
+    let xs = storage
+        .unmasked_field_by_name("x")?
```
joseph-isaacs, line 180: you are sure, should check the struct is non-nullable.

joseph-isaacs: at least a debug assert

vortex-geo/src/extension/coordinate.rs
```diff
@@ -0,0 +1,195 @@
...
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Coordinate building blocks for geometry extension types: the `Struct<x, y, z?, m?>` storage,
+//! its [`Dimension`], and the decoded [`Coordinate`] value.
+//!
+//! The coordinate fields, where `?` marks an optional field, are:
+//! - `x` — longitude or easting
+//! - `y` — latitude or northing
+//! - `z?` — elevation
+//! - `m?` — measure: an arbitrary per-point value such as distance along a route or a timestamp
```
joseph-isaacs, line 12: why do we always prefer f64 vs f32?

vortex-geo/src/extension/coordinate.rs
```diff
@@ -0,0 +1,195 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Coordinate building blocks for geometry extension types: the `Struct<x, y, z?, m?>` storage,
```
joseph-isaacs: can you fill in the types here?

vortex-geo/src/extension/coordinate.rs
HarukiMoriarty: cause `GeoArrow` and `WKB` both fix coordinates as float64.

vortex-geo/src/extension/coordinate.rs
HarukiMoriarty: done.

vortex-geo/src/extension/coordinate.rs
HarukiMoriarty: I add a `vortex_ensure`.

connortsui20: maybe you want to add a `parse_storage` helper that produces a typed struct where the construct validates the storage array? So instead of calling `xy_columns` on the storage array you do `parse_storage(storage_array, &mut ctx)` and then that given you a `ParsedCoordinate` struct that holds the primitive arrays you want.

Let me know if that makes sense or not! You can look at the turboquant code for inspiration (though not that we are going to delete that / I need to delete that soon)

HarukiMoriarty: Got a try, can you look at it? @connortsui20

HarukiMoriarty:
Superseded by #8372 — same change, re-opened from an in-repo branch (`vortex-data:nemo/geo-point`) instead of my fork. All review comments from this PR have been addressed there (latest commit also adopts the suggested `parse_storage` pattern).

PR #8344, opened 2026-06-10
Stats predicate satisfiers

gatesn:
Add satisfier rewrites for binary comparisons, boolean composition, and between expressions. The rewrites prove predicates true from min/max stats and add null/NaN guards where comparisons require every row to evaluate true.


robert3005:
This has shipped already

PR #8356, opened 2026-06-11
Minor touchups for vortex-datafusion

AdamGS:
## Summary

Some minor touchups to the `FileSource` based DataFusion integration.

The two substantial changes here are:
1. When reading files, if we didn't have the footer in the cache, make sure to insert it. That can happen when using `ListingTable` without stats inference, or when using `FileScanConfig` directly in another table provider.
2. On write - move the schema-to-dtype logic outside of the loop. It only needs to happen once and the `dtype` is cloned per write task.

### DIFF
--- a/vortex-datafusion/src/persistent/opener.rs
+++ b/vortex-datafusion/src/persistent/opener.rs
@@ -17,6 +17,7 @@ use datafusion_datasource::PartitionedFile;
 use datafusion_datasource::TableSchema;
 use datafusion_datasource::file_stream::FileOpenFuture;
 use datafusion_datasource::file_stream::FileOpener;
+use datafusion_execution::cache::cache_manager::CachedFileMetadataEntry;
 use datafusion_execution::cache::cache_manager::FileMetadataCache;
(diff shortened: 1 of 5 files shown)


vortex-datafusion/src/persistent/opener.rs
```diff
@@ -192,22 +191,41 @@ impl FileOpener for VortexOpener {
...
+                    entry
+                        .file_metadata
+                        .as_any()
+                        .downcast_ref::<CachedVortexMetadata>()
+                        .map(|vortex_metadata| vortex_metadata.footer().clone())
+                });
+            let footer_cache_hit = cached_footer.is_some();
+
+            if let Some(footer) = cached_footer {
+                open_opts = open_opts.with_footer(footer);
```
joseph-isaacs, line 208: This now has a cache!

AdamGS: where?

robert3005 (approved):
what does Joe mean?

a10y (approved):

PR #8359, opened 2026-06-11
chore: clean up Arrow device export

0ax1:
Cleanups from a review of the Arrow device export path, no behavior change:


0ax1:
@claude review

0ax1:
@claude keep all comments concise e.g.
```
/// cuDF's Arrow Device import currently rejects both Arrow `Binary` and Arrow `BinaryView`
/// (unlike `Utf8View`, which it accepts as strings), so standard `Binary` is exported as the
/// layout other Arrow Device consumers accept most widely. This path keeps conversion on the
/// CUDA stream by building `i32` offsets from view sizes and gathering inline/out-of-line view
/// bytes into one contiguous values buffer.
async fn export_binary(
```
commit and push

0ax1:
License error is unrelated to my PR.

onursatici (approved):

PR #8360, opened 2026-06-11
fix: keep Arrow device schemas aligned with exports

0ax1:
Route export_device_array_with_schema through the exporter so schema derivation sees the same rebuilt host layout that gets exported. This fixes host ListView fallback cases where the Arrow C schema could describe a different child layout than the emitted array.


0ax1:
@claude review

0ax1:
@claude fix the build / lint issues

robert3005 (approved):

PR #8364, opened 2026-06-11
Exclude benchmarks and fuzzing from default builds

myrrc:
This PR adds default-members in Cargo.toml which exclude benchmarks and fuzzing
from the default build. This should have an impact on CI as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,52 @@ members = [
     "vortex-geo",
 ]
 exclude = ["java/testfiles", "wasm-test"]
+
+# Benchmark harnesses and fuzzing account for 40% of a cold workspace build
+# and they aren't needed for ordinary builds.
+default-members = [
+    "vortex-array-macros",
+    "vortex-error",
+    "vortex-buffer",
+    "vortex-mask",
+    "vortex-utils",
+    "vortex-session",
+    "vortex-flatbuffers",
+    "vortex-metrics",
+    "vortex-io",
+    "vortex-proto",
+    "vortex-array",
+    "vortex-tensor",
+    "vortex-json",
+    "vortex-compressor",
+    "vortex-btrblocks",
+    "vortex-layout",
+    "vortex-scan",
+    "vortex-file",
+    "vortex-ipc",
+    "vortex",
+    "vortex-datafusion",
+    "vortex-duckdb",
+    "vortex-cuda",
+    "vortex-cuda/cub",
+    "vortex-cuda/ffi",
+    "vortex-cuda/gpu-scan-cli",
+    "vortex-cuda/macros",
+    "vortex-cuda/nvcomp",
+    "vortex-cxx",
+    "vortex-ffi",
+    "vortex-jni",
+    "vortex-python",
+    "vortex-tui",
+    "vortex-web/crate",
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -388,6 +448,9 @@ unwrap_in_result = "deny"
 unwrap_used = "deny"
 use_debug = "deny"
 
+[profile.dev.package."*"]
+debug = false
```
robert3005: don't quite get this? Strip debug info from dev profile?

myrrc: Debug info for non-workspace crates. I'll double check, but it should work this way.

AdamGS: Its for non-workspace dependencies. I think the debug info is a feature of dev builds.

AdamGS:
I'm not sure I agree with the first point, but the latter seems very reasonable, no reason to treat auxiliary code the same as the code crates.

robert3005 (approved):

AdamGS (changes requested):
debug info is a feature of `dev` builds. If you don't want it, you can build the specific crates you care about or use `ci` profile.

robert3005 (dismissed):

AdamGS:
you have a conflict in CI

.github/workflows/sql-benchmarks.yml
```diff
@@ -349,7 +349,7 @@ jobs:
           if [ "${{ inputs.mode }}" != "pr" ]; then
             packages="$packages --bin lance-bench"
           fi
-          cargo build $packages --profile release_debug --features unstable_encodings
+          cargo build --workspace $packages --profile release_debug --features unstable_encodings
```
AdamGS: does it make a difference here were we pass the binaries explicitly?

Cargo.toml
```diff
@@ -70,6 +70,65 @@ members = [
...
+    "encodings/alp",
+    "encodings/datetime-parts",
+    "encodings/fsst",
+    "encodings/pco",
+    "encodings/sparse",
+    "encodings/zigzag",
+    "encodings/zstd",
+    "encodings/bytebool",
+    "encodings/parquet-variant",
+    "encodings/experimental/onpair",
```
AdamGS: Can we have a wildcard here like we have for members? `"encodings/*"`

.github/workflows/bench.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         env:
           RUSTFLAGS: "-C target-cpu=native"
         run: |
-          cargo build --bin ${{ matrix.benchmark.id }} --profile release_debug ${{ matrix.benchmark.build_args }} --features unstable_encodings
+          cargo build --workspace --bin ${{ matrix.benchmark.id }} --profile release_debug ${{ matrix.benchmark.build_args }} --features unstable_encodings
```
AdamGS: same as below - we build a specific binary here, does `--workspace` makes a difference?

.github/workflows/sql-benchmarks.yml
myrrc: +1, I've run a regex replace which wasn't a good idea.

Cargo.toml
myrrc: +1

PR #8365, opened 2026-06-11
aggregate_fn: Return NaN instead of null for mean of all-null input

dimitarvdimitrov:
`Mean::finalize_scalar` returned null when the count was zero, while the array `finalize` path computes `sum/count = 0/0 = NaN` for the same input. A mean over an all-null group therefore gave different results depending on which accumulator we're using.

https://github.com/vortex-data/vortex/blob/90d743356722a8d5ca7e39053229654778bacf46/vortex-array/src/aggregate_fn/fns/mean/mod.rs#L85-L93

Since nulls are skipped during accumulation (as in standard SQL aggregation), an all-null input is an empty mean. Both paths now let the division produce NaN.

Note this only matters for the count = 0 case: sum overflow still returns null.

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/mean/mod.rs
+++ b/vortex-array/src/aggregate_fn/fns/mean/mod.rs
@@ -104,7 +104,9 @@ impl BinaryCombined for Mean {
         let sum = sum_cast.as_primitive().typed_value::<f64>();
         let count = count_cast.as_primitive().typed_value::<f64>();
         let value = match (sum, count) {
-            (None, _) | (_, None) | (_, Some(0.0)) => return Ok(Scalar::null(target)), // Sum overflowed
+            (None, _) | (_, None) => return Ok(Scalar::null(target)), // Sum overflowed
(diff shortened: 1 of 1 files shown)


robert3005 (approved):

joseph-isaacs:
This looks incorrect to me. I think this should return Null

robert3005:
I thought that NaN made more semantic sense but you can argue either way. Just have to document what is mean of an empty array

robert3005:
Basically I think NaN as a result composes from partials more naturally without any additional checks

dimitarvdimitrov:
if we combine NaNs, then the associativity of combining partials breaks. `mean([a, b, c]) != combine_partial(mean_partial([a, b]), mean_partial([c]))` because NaN always turns anything else into a NaN, but we can ignore nulls.

both duckdb and datafusion return nulls instead of nans for all-null input. 

opened https://github.com/vortex-data/vortex/pull/8389 to revert this commit and fix `finalize`

PR #8369, opened 2026-06-11
fix[file]: read the one-row pruning result in can_prune instead of re…

### LINKED ISSUE #8366: Regression: can_prune stops pruning and/or/eq predicates
### What happened?

We'd been on Vortex 0.69 for a while and, upgrading to 0.74, hit a regression caused by the removal of `ScalarFnConstantRule` in #7575.

`can_prune` evaluates a predicate's stat-falsification expression over the one-row file-stats array, but it only accepted a constant-folded result (`Columnar::Constant`) and treated a materialized one-row boolean (`Columnar::Canonical`) as "cannot prove".

After #7575 removed bottom-up constant folding, composite falsifications no longer fold to a constant: boolean trees (`and`/`or`) and `eq` (whose falsification is internally `or(min > lit, lit > max)`) now execute to a one-row `Canonical`, so `can_prune` discards the result and stops pruning. Bare `gt`/`lt` comparisons still fold through the compare kernel's one-level constant fast path, so only composite and `eq` predicates regressed.

### Steps to reproduce

 1. Write a Vortex file containing a struct column whose statistics bound the values — e.g. an age column with values [15, 18, 22, 25] (min 15, max 25) and a price column [120, 130, 140, 150].
 2. Open the file (SESSION.open_options().open_buffer(buf)?).
 3. Call file.can_prune(&eq(col("age"), lit(5)))? ... 5 is outside the [15, 25] min/max, so the file provably contains no match. Expected Ok(true) (prunable); actual Ok(false).
 4. Call file.can_prune(&and(gt(col("age"), lit(30)), lt(col("price"), lit(100))))? ...both branches are falsified by the stats. Expected Ok(true); actual Ok(false).
 5. Call file.can_prune(&or(gt(col("age"), lit(30)), lt(col("age"), lit(10))))? ... both branches falsified. Expected Ok(true); actual Ok(false).
 6. As a control, call file.can_prune(&gt(col("age"), lit(30)))? ... a bare comparison, which still folds through the compare kernel's fast path. Returns Ok(true) correctly, showing only composite/eq predicates regressed.

### Environment

- Observed on 0.74.0, upgrading from 0.69.0 (last release before #7575).
- Present in every release since #7575 (2026-04-20): 0.70.0 through 0.74.0, and develop HEAD.
- Platform-independent

### Additional context

_No response_

tomsanbear:
## Summary

Closes: #8366

`VortexFile::can_prune` stopped pruning for every `and`/`or` and `eq` predicate after #7575 removed bottom-up constant-folding. It evaluates a predicate's stat-falsification expression over the one-row file-stats array, but only read a folded `Columnar::Constant` result and mapped `Columnar::Canonical(_)` to `false`, discarding the computed boolean. Post-#7575, composite falsifications (boolean `and`/`or` trees, and the `or(min > lit, lit > max)` that `eq` expands to) execute to a one-row `Canonical` instead of folding, so pruning silently stopped for them. Bare `gt`/`lt` comparisons still fold through the compare kernel's one-level constant fast path, so only composite and `eq` predicates regressed.

It went unnoticed upstream because `can_prune` has no in-repo callers, the scan path prunes through `FileStatsLayoutReader::evaluate_file_stats`, which already reads the one-row result out correctly. This change makes `can_prune` mirror that read-out (execute to `Canonical`, take the row-0 scalar) and adds an end-to-end regression test covering bare, `and`/`or`, and `eq` predicates with non-falsifiable controls.

## API Changes

No signature changes. This restores the runtime behavior of the public `VortexFile::can_prune`: it now returns `true` (prunable) for `and`/`or`/`eq` predicates that file statistics prove cannot match, where it previously returned `false`. No engine integration is affected. The scan path already pruned correctly via `evaluate_file_stats`; only direct callers of `can_prune` saw the regression. No docs changes needed.

## Testing

- New `test_can_prune_composite_predicates` (`vortex-file/src/tests.rs`): writes a file with stat-bounded `age`/`price` columns and asserts pruning for bare, `and`, `or`, and `eq` falsifiable predicates, plus non-falsifiable controls that must not prune.
- Verified non-vacuous: the test fails against the pre-fix `Columnar::Canonical(_) => false` arm and passes after.
- `cargo nextest run -p vortex-file` — 62/62 pass (existing suite + new test).
- `cargo +nightly fmt -p vortex-file` and `cargo clippy -p vortex-file --all-targets` clean.

## AI assistance

Developed with assistance from Claude Code (agentic AI) for root-cause analysis, the fix, and the regression test. All changes were reviewed and verified by me against the Vortex source and a downstream test suite.

### DIFF
--- a/vortex-file/src/file.rs
+++ b/vortex-file/src/file.rs
@@ -12,7 +12,7 @@ use std::sync::OnceLock;
 
 use itertools::Itertools;
 use vortex_array::ArrayRef;
-use vortex_array::Canonical;
+use vortex_array::Columnar;
 use vortex_array::IntoArray;
 use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::ConstantArray;
@@ -245,12 +245,16 @@ impl VortexFile {
         let applied = substitute_row_count(applied, &row_count_replacement)?;
 
         let mut ctx = self.session.create_execution_ctx();
-        let result = applied
-            .execute::<Canonical>(&mut ctx)?
-            .into_bool()
-            .into_array()
-            .execute_scalar(0, &mut ctx)?;
-        Ok(result.as_bool().value() == Some(true))
+        Ok(match applied.execute::<Columnar>(&mut ctx)? {
+            Columnar::Constant(s) => s.scalar().as_bool().value() == Some(true),
+            Columnar::Canonical(c) => {
+                c.into_array()
+                    .execute_scalar(0, &mut ctx)?
+                    .as_bool()
+                    .value()
+                    == Some(true)
+            }
+        })
     }
 
     pub fn splits(&self) -> VortexResult<Vec<Range<u64>>> {
--- a/vortex-file/src/file.rs
+++ b/vortex-file/src/file.rs
@@ -12,7 +12,7 @@ use std::sync::OnceLock;
 
 use itertools::Itertools;
 use vortex_array::ArrayRef;
-use vortex_array::Columnar;
+use vortex_array::Canonical;
 use vortex_array::IntoArray;
 use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::ConstantArray;
@@ -245,10 +245,12 @@ impl VortexFile {
         let applied = substitute_row_count(applied, &row_count_replacement)?;
 
         let mut ctx = self.session.create_execution_ctx();
-        Ok(match applied.execute::<Columnar>(&mut ctx)? {
-            Columnar::Constant(s) => s.scalar().as_bool().value() == Some(true),
-            Columnar::Canonical(_) => false,
-        })
+        let result = applied
+            .execute::<Canonical>(&mut ctx)?
+            .into_bool()
+            .into_array()
+            .execute_scalar(0, &mut ctx)?;
+        Ok(result.as_bool().value() == Some(true))
     }
 
     pub fn splits(&self) -> VortexResult<Vec<Range<u64>>> {
--- a/vortex-file/src/tests.rs
+++ b/vortex-file/src/tests.rs
@@ -35,6 +35,7 @@ use vortex_array::dtype::PType::I32;
 use vortex_array::dtype::StructFields;
 use vortex_array::expr::and;
 use vortex_array::expr::cast;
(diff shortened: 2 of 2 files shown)


vortex-file/src/file.rs
```diff
@@ -245,10 +245,12 @@ impl VortexFile {
         let applied = substitute_row_count(applied, &row_count_replacement)?;
 
         let mut ctx = self.session.create_execution_ctx();
-        Ok(match applied.execute::<Columnar>(&mut ctx)? {
-            Columnar::Constant(s) => s.scalar().as_bool().value() == Some(true),
-            Columnar::Canonical(_) => false,
-        })
+        let result = applied
+            .execute::<Canonical>(&mut ctx)?
+            .into_bool()
```
robert3005: you can skip this into_bool

vortex-file/src/file.rs
```diff
@@ -245,10 +245,12 @@ impl VortexFile {
         let applied = substitute_row_count(applied, &row_count_replacement)?;
 
         let mut ctx = self.session.create_execution_ctx();
-        Ok(match applied.execute::<Columnar>(&mut ctx)? {
-            Columnar::Constant(s) => s.scalar().as_bool().value() == Some(true),
-            Columnar::Canonical(_) => false,
```
joseph-isaacs, line 250: add the `.execute_scalar(0, &mut ctx)?;` == 0 here

joseph-isaacs:
Great find thanks!

Can we keep the `execute::<Columnar>` and just check if the canonical first value is a boolean

tomsanbear:
> Great find thanks!
> 
> Can we keep the `execute::<Columnar>` and just check if the canonical first value is a boolean

for sure, let me make a quick patch to address this and robert's comment

vortex-file/src/file.rs
tomsanbear: addressed now in 46623180e

tomsanbear:
@joseph-isaacs @robert3005 updated with the change to maintain the previous match arm

joseph-isaacs (approved):

PR #8370, opened 2026-06-11
File based pruning handles single row canonical arrays instead of just constant arrays

### LINKED ISSUE #8366: Regression: can_prune stops pruning and/or/eq predicates

robert3005:
Fix #8366

### DIFF
--- a/vortex-file/src/file.rs
+++ b/vortex-file/src/file.rs
@@ -12,7 +12,6 @@ use std::sync::OnceLock;
 
 use itertools::Itertools;
 use vortex_array::ArrayRef;
-use vortex_array::Columnar;
 use vortex_array::IntoArray;
 use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::ConstantArray;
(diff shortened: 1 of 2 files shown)


robert3005:
let's go with #8369

PR #8371, opened 2026-06-11
Constant comparison and byte_length OnPair kernels

myrrc:
This PR adds 2 kernels for OnPair:
- byte_length which uses uncompressed_lengths only
- constant comparison to avoid decompressing for queries like URL != ''

This PR also adds validity() for FSST's kernel for byte_length so that
evaluating validity wouldn't mean decompressing whole FSST array.

### DIFF
--- a/encodings/experimental/onpair/src/compute/byte_length.rs
+++ b/encodings/experimental/onpair/src/compute/byte_length.rs
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+use vortex_array::IntoArray;
+use vortex_array::ValidityVTable;
+use vortex_array::arrays::ConstantArray;
+use vortex_array::builtins::ArrayBuiltins;
+use vortex_array::dtype::DType;
+use vortex_array::dtype::PType;
+use vortex_array::scalar::Scalar;
+use vortex_array::scalar_fn::fns::byte_length::ByteLengthKernel;
+use vortex_array::validity::Validity;
+
+use crate::OnPair;
+use crate::OnPairArraySlotsExt;
+
+// TODO(myrrc): this and FSST comparison should be in the same trait.
+// https://github.com/vortex-data/vortex/tree/myrrc/onpair-compare-trait
+impl ByteLengthKernel for OnPair {
+    fn byte_length(
+        array: vortex_array::ArrayView<'_, Self>,
+        _ctx: &mut vortex_array::ExecutionCtx,
+    ) -> vortex_error::VortexResult<Option<vortex_array::ArrayRef>> {
+        let nullable = array.dtype().nullability();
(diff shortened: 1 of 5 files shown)


encodings/experimental/onpair/src/compute/compare.rs
```diff
@@ -0,0 +1,143 @@
...
+            CompareOperator::Lt => BitBuffer::new_unset(lhs.len()),
+            _ => lengths
+                .binary(
+                    ConstantArray::new(Scalar::zero_value(lengths.dtype()), lengths.len())
+                        .into_array(),
+                    operator.into(),
+                )?
+                .execute(ctx)?,
+        };
+        Ok(Some(
```
joseph-isaacs, line 54: lets make a trait over the two string arrays we can use to impl this?

myrrc: I started with a trait but it's a worse approach, because the only thing we can re-implement is comparison with an empty constant and then the diff is twice as large
I don't think over-generalisation is a good thing here.

encodings/experimental/onpair/src/compute/compare.rs
```diff
@@ -0,0 +1,143 @@
...
+            _ => return Ok(None),
+        };
+        if is_empty != Some(true) {
+            return Ok(None);
+        }
+
+        let lengths = lhs.uncompressed_lengths();
+        let buffer = match operator {
+            // every value is greater than an empty string
+            CompareOperator::Gte => BitBuffer::new_set(lhs.len()),
```
joseph-isaacs, line 43: todo once we can have a constant with validity

robert3005: Isn’t Mask scalar function that?

encodings/experimental/onpair/src/compute/compare.rs
myrrc: FSST uses code->row mapping, OnPair uses code range->row mapping, so non-empty constant comparison requires re-tokenizing the constant over the dictionary. We can't generalize both operations to a single trait.
And if we have a trait, we would be duplicating the same operations, but not in the function for CompareKernel but in this new trait. I'm against this extra complexity.

joseph-isaacs: Can I see a skeleton trait and we can go from there

myrrc: https://github.com/vortex-data/vortex/tree/myrrc/onpair-compare-trait

joseph-isaacs: We need a change the onpair to get a lpm exposed.

encodings/experimental/onpair/src/compute/byte_length.rs
joseph-isaacs, line 1: Add a todo with unify with FSST

joseph-isaacs (approved):

PR #8372, opened 2026-06-11
feat(vortex-geo): native Point extension type and GeoDistance scalar function

HarukiMoriarty:
## Summary

This PR adds a native point type to `vortex-geo`. Points are by far the most common geometry in analytical datasets, and a columnar representation makes their coordinates directly accessible without parsing WKB.

It also adds the scalar function: point-to-point distance with PostGIS `ST_Distance` semantics (planar/Euclidean, results in CRS units).

## API Changes

Adds to `vortex-geo`, all registered through `vortex_geo::initialize`:

- Extension type `Point` (`vortex.geo.point`): a location stored as `Struct<x, y, z?, m?>` of non-nullable `f64`, where `z?` is an optional elevation and `m?` an optional measure. 
- `Coordinate`: the internal value a point scalar unpacks to.
- Scalar function `GeoDistance` (`vortex.geo.distance`): per-row distance between two equal-length point columns; either or both operands may be constant, in which case the query point is decoded once and broadcast.

## Testing

Unit tests cover dtype validation for every GeoArrow dimension (and rejection of invalid storage), round-tripping a point column through scalar execution back to the original coordinates, WKT display for all four dimensions, and distance over all operand shapes: column-to-constant (either side), column-to-column, and constant-to-constant.

---
Supersedes #8342 (same change, moved from my fork to an in-repo branch).


vortex-geo/src/extension/point.rs
```diff
@@ -0,0 +1,165 @@
...
+use vortex_array::scalar::Scalar;
+use vortex_array::scalar::ScalarValue;
+use vortex_error::VortexResult;
+
+use super::GeoMetadata;
+use super::coordinate::Coordinate;
+use super::coordinate::coordinate_dimension;
+use super::coordinate::coordinate_from_struct;
+
+/// A single location: `geoarrow.point`, stored as `Struct<x, y, z?, m?>` of `f64`.
```
joseph-isaacs:
```suggestion
/// A single location: `geoarrow.point`, stored as `Struct<x: f64, y: f64, z: f64?, m: f64?>`.
```

vortex-geo/src/extension/coordinate.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+}
+
+/// Validate that `dtype` is a coordinate struct of non-nullable `f64` fields, returning its
+/// [`Dimension`]. Any of the four GeoArrow dimensions validates.
+pub(crate) fn coordinate_dimension(dtype: &DType) -> VortexResult<Dimension> {
+    let DType::Struct(fields, _) = dtype else {
+        vortex_bail!("coordinate storage must be a Struct, was {dtype}");
+    };
+    let names: Vec<&str> = fields.names().iter().map(|n| n.as_ref()).collect();
```
joseph-isaacs: why alloc?

vortex-geo/src/extension/coordinate.rs
```diff
@@ -0,0 +1,250 @@
...
+    for (i, field) in fields.fields().enumerate() {
+        vortex_ensure!(
+            matches!(
+                field,
+                DType::Primitive(PType::F64, Nullability::NonNullable)
```
joseph-isaacs, line 117: I thought that two fields are Nullable?

vortex-geo/src/extension/coordinate.rs
HarukiMoriarty: Removed. The names are now staged in a stack buffer inside `from_field_names` so the slice-pattern match still works, and coordinate_dimension zips names with fields directly instead of collecting.

vortex-geo/src/extension/coordinate.rs
HarukiMoriarty: z/m are optional fields (per dimension), not nullable ones — the GeoArrow spec requires coordinate fields to be non-nullable, with "only the outer level allowed to have nulls". So a point can be missing entirely, but a present point can't have a null ordinate.

Ref: https://geoarrow.org/format.html

connortsui20: oh maybe you should make it `Struct<x, y, ?z, ?m>` instead, I forgot that ? also means possibly null, not just optional. Or you can do `Struct<x, y, {z}, {m}>`

joseph-isaacs (approved):

PR #8373, opened 2026-06-11
[WIP] Use Kahan (Neumaier) summation for float sums

dimitarvdimitrov:
Stacked on #8367  so the CodSpeed run shows the cost of compensated summation against the freshly-added benchmarks.

In draft because we probably don't want to eat the perf cost

#### What this pr does

- `SumState::Float` now holds a [`KahanSum`](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) (running sum + compensation term) implementing Neumaier's variant of Kahan compensated summation. When an addition produces a non-finite value the compensation update is skipped, preserving the existing overflow-to-infinity and `inf + -inf => NaN` (saturated) semantics.
- Float sum **partials** are a struct `{sum: f64, compensation: f64}` instead of a plain `f64`, so precision survives partial flush/combine boundaries (chunked arrays, kernel partials, accumulator merges):
  - `Sum::combine_partials` accepts both the struct partial and a plain `f64`.
  - The legacy stats bridge in `Accumulator::accumulate` falls through on a partial dtype mismatch instead of erroring, and `Sum::try_accumulate` consumes the cached f64 `Stat::Sum` for floats so the stats short-circuit is preserved.

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/sum/mod.rs
+++ b/vortex-array/src/aggregate_fn/fns/sum/mod.rs
@@ -161,7 +161,7 @@ impl AggregateFnVTable for Sum {
                     .as_primitive()
                     .typed_value::<f64>()
                     .vortex_expect("checked non-null");
-                *acc += val;
+                acc.add(val);
                 false
             }
             SumState::Decimal { value, dtype } => {
@@ -189,7 +189,7 @@ impl AggregateFnVTable for Sum {
             None => Scalar::null(partial.return_dtype.as_nullable()),
             Some(SumState::Unsigned(v)) => Scalar::primitive(*v, Nullability::Nullable),
             Some(SumState::Signed(v)) => Scalar::primitive(*v, Nullability::Nullable),
-            Some(SumState::Float(v)) => Scalar::primitive(*v, Nullability::Nullable),
+            Some(SumState::Float(v)) => Scalar::primitive(v.value(), Nullability::Nullable),
             Some(SumState::Decimal { value, .. }) => {
                 let decimal_dtype = *partial
                     .return_dtype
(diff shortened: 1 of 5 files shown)


vortex-array/src/aggregate_fn/fns/sum/mod.rs
```diff
@@ -113,7 +118,13 @@ impl AggregateFnVTable for Sum {
     }
 
     fn partial_dtype(&self, options: &Self::Options, input_dtype: &DType) -> Option<DType> {
-        self.return_dtype(options, input_dtype)
+        let return_dtype = self.return_dtype(options, input_dtype)?;
+        Some(match &return_dtype {
+            // Float partials carry the Kahan compensation term so that precision is preserved
+            // when partial sums are flushed and re-combined.
+            DType::Primitive(ptype, _) if ptype.is_float() => float_partial_dtype(),
```
gatesn, line 125: CC @joseph-isaacs - but this would be a breaking change for serialized SUM statistics if we had started to serialize them in this way.

Do other engines use this?

gatesn: In general, we should treat Vortex as a collection of physical kernels that some query engine may resolve to based on the engine's own semantics. 

Therefore, this should be its own aggregate function. And DuckDB / DataFusion aggregate push-down can pick this one if it needs to have matching execution semantics.

robert3005: Datafusion supports reporting Sum statistic, that was blocked though until 54 that released this week since sum stats were very slow

dimitarvdimitrov: we spoke with Joe offline too. Keeping backwards compat would make this change more complicated.

the only one I could find was duckdb which has a dedicated `fsum` (aka `sum_kahan`) function. Because of that I think it's ok to keep a less accurate sum aggregation

PR #8374, opened 2026-06-11
feat(vortex-geo): Arrow import/export for the native Point type

HarukiMoriarty:
## Summary

This PR adds support for import/export to Arrow for the `Point` extension type, as the
`geoarrow.point` Arrow extension with separated (struct) coordinates.

Stacked on #8372.

## Testing

Unit tests are added to exercise both code paths, plus a Vortex → Arrow → Vortex round-trip.


vortex-geo/src/extension/coordinate.rs
```diff
@@ -202,17 +251,38 @@ mod tests {
...
 
+    /// Each dimension round-trips through its field names and canonical storage dtype.
+    #[test]
+    fn storage_dtype_roundtrips_dimension() -> VortexResult<()> {
+        let cases = [
+            (Dimension::Xy, ["x", "y"].as_slice()),
+            (Dimension::Xyz, ["x", "y", "z"].as_slice()),
+            (Dimension::Xym, ["x", "y", "m"].as_slice()),
+            (Dimension::Xyzm, ["x", "y", "z", "m"].as_slice()),
+        ];
```
a10y: nit: we have `` `rstest::cases` `` for this. just search for that in the repo and you'll see several examples. you might need to import rstest into this crate to use it

vortex-geo/src/extension/point.rs
```diff
@@ -55,18 +84,143 @@ impl ExtVTable for Point {
...
+    fn from_arrow_field(&self, field: &Field) -> VortexResult<Option<DType>> {
+        let Ok(point_meta) = field.try_extension_type::<PointType>() else {
+            return Ok(None);
+        };
+        if point_meta.coord_type() != CoordType::Separated {
+            vortex_bail!(
+                "geoarrow.point with interleaved coordinates is not supported; \
+                 re-encode with separated (struct) coordinates"
+            );
+        }
```
a10y: there is a `vortex_ensure!()` macro which captures the if-not-vortex_bail flow

vortex-geo/src/extension/coordinate.rs
HarukiMoriarty: Done, also applied to other places.

vortex-geo/src/extension/point.rs
HarukiMoriarty: Done.

a10y (approved):

PR #8378, opened 2026-06-11
Remove `len` parameter from `ScalarFnArray::try_new`

connortsui20:
## Summary

This didn't need to take length because it already gets it from the array children.

## API Changes

Removes `len` parameter.

## Testing

N/A


robert3005 (approved):

connortsui20:
@claude fix the merge conflicts

joseph-isaacs:
What is the scalar fn doesn't have children?

`pub fn try_new(scalar_fn: ScalarFnRef, children: Vec<ArrayRef>)` if the vec.len is empty?

robert3005:
There’s still a method that lets you provide len. I think it’s nicer since the other cases are very few compared to cases with non empty children

joseph-isaacs:
No cause now if the children is empty this becomes a bail or panic! For no reason

robert3005:
This is the same thing as constructing chunkedarray from children and dtype or just children, having empty children panics.

PR #8314, opened 2026-06-09
aggregate fns to have grouped aggregate kernels for sum and count

onursatici:
## Summary

aggregate functions to be able to do grouped aggregations before the fallback that slices group by group.

Implements count for all arrays and sum for primitives.

## API Changes

Grouped aggregate kernels now receive a single `GroupedArray` enum, covering `ListViewArray` and `FixedSizeListArray`, instead of exposing separate methods for each list representation

PR #8379, opened 2026-06-11
Support out of order grouped aggregate accumulation

gatesn:
Our original API for grouped aggregate functions only supported groups that we pre-sorted into complete lists. That meant large groups had to be globally sorted before we could start accumulating their internal state, unlike the grouped-aggregate APIs used by query engines.

This PR updates the grouped accumulator API to accept caller-assigned dense `GroupIds` directly.

## Summary

- Accept `GroupIds { ids: ArrayRef, num_groups }`, including repeated, out-of-order, and absent IDs.
- Dispatch grouped kernels on both the value encoding and group-ID encoding.
- Cache materialized and validated group IDs across clones and aggregate functions.
- Keep a list-group compatibility adapter for existing `list_sum` callers.
- Let each aggregate own its dense grouped-state representation, with the existing generic partial-state implementation as the fallback.
- Store grouped count state as one `u64` per group.
- Store grouped sum state in monomorphic `u64`, `i64`, `f64`, or decimal vectors, with overflow and empty state kept separately.
- Preserve SQL null-on-empty sum semantics, including all-null groups and groups absent from a batch.
- Select exact decimal state width (`i8` through `i256`) once when the accumulator is created.
- Merge count, primitive-sum, float-sum, and decimal-sum partial arrays without `execute_scalar` per row.
- Adapt between direct typed scatter for unclustered IDs and per-run reduction for clustered IDs.
- Use one stable counting-sort gather per morsel in the universal fallback instead of one `take` per group.

## Spiral impact

Spiral already interns group keys into dense `u32` state slots, so it can pass those IDs directly without rearranging values into ordered list groups. This covers the common count/sum machinery while allowing Spiral to retain its stronger integer/decimal mean states and other specialized paths.

## Local performance

Medians below use 65,536 non-null `i32` rows and include canonical result materialization. Shuffled inputs contain the same multiset of IDs as their clustered counterpart.

| Groups | Count clustered | Count shuffled | Sum clustered | Sum shuffled |
|---:|---:|---:|---:|---:|
| 128 | 16.6 μs | 16.5 μs | 22.0 μs | 36.3 μs |
| 4,096 | 18.5 μs | 16.1 μs | 30.2 μs | 26.9 μs |
| 65,536 | 18.3 μs | 37.2 μs | 60.4 μs | 90.2 μs |

The original 65,536-row / 4,096-group shuffled sum benchmark was approximately 215 μs with per-group `SumPartial` state. Typed storage plus adaptive scatter reduces it to approximately 26.9 μs while also materializing the result and tracking SQL empty-state semantics, an approximately 8× improvement.

On the existing roughly 1K-row cases, representative medians are 1.9 μs for clustered-null count, 3.7 μs for nullable all-valid `i32` sum, 3.9 μs for all-valid `f64` sum, and 6.2 μs for clustered-null `i32` sum.

## Validation

- `cargo +nightly fmt --all`
- `cargo test -p vortex-array --lib` (3,376 passed, 1 ignored)
- `cargo bench -p vortex-array --bench aggregate_grouped`
- `cargo clippy -p vortex-array --all-targets --all-features`
- `RUSTDOCFLAGS="-D warnings" cargo doc --profile ci --no-deps -p vortex-array`


joseph-isaacs:
We wait that vortex was a linear scan engine. This breaks it correct?

onursatici (commented):
I prefer this API, not only that it supports out of order groups but also it can keep the memory footprint lower by not having us to materialise each group before calling aggregate.

I think you are conflicting with https://github.com/vortex-data/vortex/pull/8314
Sum and count kernels you added make sense to me but we went with having a encoding agnostic kernel support on the registry on the merged PR instead of having the kernel in the aggregate function's vtable

gatesn:
> We wait that vortex was a linear scan engine. This breaks it correct?

No, the idea is that I can take an array, compute the group indices (i.e. using a scalar function), then update the aggregate state per group. This still doesn't require a shuffle / sort. But it does mean aggregate state may grow large. So in the future this is where we would want the ability to spill partials / reconstruct later.

vortex-array/benches/aggregate_grouped.rs
```diff
@@ -45,44 +43,42 @@ fn total_element_count(group_sizes: &[usize]) -> usize {
...
-fn contiguous_list_view(elements: ArrayRef, group_sizes: &[usize]) -> ArrayRef {
-    let mut offset = 0usize;
-    let offsets: Buffer<u32> = group_sizes
+struct DenseGroupedInput {
+    values: ArrayRef,
+    group_ids: Vec<u32>,
+    num_groups: usize,
+}
+
+fn dense_grouped_input(values: ArrayRef, group_sizes: &[usize]) -> DenseGroupedInput {
```
AdamGS: nit - this is basically the constructor for `DenseGroupedInput`

vortex-array/src/aggregate_fn/fns/count/grouped.rs
```diff
@@ -1,216 +1,22 @@
...
-
-    Ok(Some(grouped_count(groups.elements(), &group_ranges, ctx)?))
-}
-
-/// Count the valid elements of each group described by `group_ranges` (element `(offset, size)`
-/// pairs) into a non-nullable `U64` array, one entry per group.
-fn grouped_count(
-    elements: &ArrayRef,
-    group_ranges: &GroupRanges,
+pub(super) fn try_accumulate_grouped(
```
AdamGS: nit - this seems short enough to inline into the vtable

onursatici (commented):
I think we should choose either to have encoding agnostic grouped kernels that we register in the registry, like `develop` does with `AggregateFnSession::register_grouped_kernel`, or have the fast path as methods on the  aggregate function vtable, like this PR does with `AggregateFnVTable::try_accumulate_grouped`

  If we merge this as is, we will have two ways that do very similar things

gatesn:
Agreed, in general I want to move everything towards session kernels, including both aggregate functions and scalar functions.

vortex-array/src/aggregate_fn/fns/count/grouped.rs
```diff
@@ -1,216 +1,22 @@
...
-        let actual = grouped_count_actual(&groups, &elem_dtype)?;
-        let direct = PrimitiveArray::new(buffer![1u64, 2], Validity::NonNullable).into_array();
-        assert_arrays_eq!(&actual, &direct);
-        Ok(())
+) -> VortexResult<bool> {
+    let validity = batch.validity()?.execute_mask(batch.len(), ctx)?;
+    for (&group_id, valid) in group_ids.iter().zip(validity.iter()) {
+        if valid {
+            states[group_id as usize] += 1;
+        }
```
joseph-isaacs: This also looks kinda slow

vortex-array/src/aggregate_fn/fns/sum/grouped.rs
```diff
@@ -1,367 +1,297 @@
...
+        AllOr::All => {
+            for idx in 0..len {
+                f(idx);
+            }
+        }
+        AllOr::None => {}
+        AllOr::Some(indices) => {
+            for &idx in indices {
+                f(idx);
+            }
```
joseph-isaacs: Is this not a method on Mask?

vortex-array/src/aggregate_fn/vtable.rs
```diff
@@ -146,6 +157,37 @@ pub trait AggregateFnVTable: 'static + Sized + Clone + Send + Sync {
...
+    /// Try to accumulate a raw values batch into dense per-group states before decompression.
+    ///
+    /// `group_ids` is parallel to `batch` and contains caller-assigned dense ordinals in
+    /// `0..states.len()`. Ids may repeat, appear out of order, or be absent from the batch.
+    /// Returns `true` when the batch was fully handled.
+    fn try_accumulate_grouped(
+        &self,
+        _states: &mut [Self::Partial],
+        _batch: &ArrayRef,
+        _group_ids: &[u32],
```
joseph-isaacs: I think we should use an ArrayRef here so we can dispatch on different encoding ids? RLE, Constant

vortex-array/src/aggregate_fn/accumulator_grouped.rs
```diff
@@ -221,199 +74,351 @@ impl<V: AggregateFnVTable> GroupedAccumulator<V> {
...
+            self.vtable
+                .combine_partials(&mut self.partials[group_id as usize], partial)?;
+        }
+        Ok(())
     }
 
-    fn push_result(&mut self, state: ArrayRef) -> VortexResult<()> {
+    fn merge_group(
+        &mut self,
+        into: u32,
```
joseph-isaacs:
```suggestion
        into_num_groups: u32,
```

vortex-array/src/aggregate_fn/accumulator_grouped.rs
```diff
@@ -221,199 +74,351 @@ impl<V: AggregateFnVTable> GroupedAccumulator<V> {
...
-    fn accumulate_list(&mut self, groups: &ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<()>;
+    /// Accumulate a values batch into dense group state.
+    ///
+    /// `group_ids` is parallel to `batch`. Each id must be a caller-assigned group ordinal in
+    /// `0..num_groups`; ids may repeat, appear out of order, or be absent from a given batch.
+    fn accumulate(
+        &mut self,
+        batch: &ArrayRef,
+        group_ids: &[u32],
+        num_groups: usize,
```
joseph-isaacs: Should this be a Group struct?

vortex-array/src/aggregate_fn/accumulator_grouped.rs
```diff
@@ -221,199 +74,351 @@ impl<V: AggregateFnVTable> GroupedAccumulator<V> {
...
+        }
+
+        Ok(false)
+    }
+
+    fn accumulate_fallback(
+        &mut self,
+        batch: &ArrayRef,
+        group_ids: &[u32],
+        ctx: &mut ExecutionCtx,
```
joseph-isaacs, line 700: Why don't we know the num_groups here?

PR #8381, opened 2026-06-11
Ignore PyO3 rustsec

connortsui20:
## Summary

Ignores the pyo3 rustsec because there are other dependencies that haven't upgraded yet.

## Testing

N/A


gatesn (dismissed):
Let's make this not a required check for branches, but still required before release

robert3005:
These aren't required checks already

AdamGS (approved):
I think this is fine, as far as I can tell we don't use those code paths, and this is a relatively minor dependency.

PR #8382, opened 2026-06-11
Run RustSec before publish

gatesn:


robert3005 (changes requested):
How do you retrigger this on the same commit? I don't like that you will have permanently broken release if you don't check this. Would prefer a release that we might need to yank

.github/workflows/ci.yml
```diff
@@ -391,11 +391,11 @@ jobs:
         checks:
           - advisories
           - bans licenses sources
-    # Prevent sudden announcement of a new advisory from failing ci:
-    continue-on-error: ${{ matrix.checks == 'advisories' }}
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
       - uses: EmbarkStudios/cargo-deny-action@bb137d7af7e4fb67e5f82a49c4fce4fad40782fe  # v2
+        # Prevent sudden announcement of a new advisory from failing CI or blocking PR merges.
+        continue-on-error: ${{ matrix.checks == 'advisories' }}
```
connortsui20, line 398: do we want to just delete this and only check on release?

robert3005:
I think this check should run on every pr and before cutting release you have to check the status of it on develop. The ideal solution is to have a release bot that would check this and only tag if the check passes no tag and then check

PR #8383, opened 2026-06-12
Remove debug symbols from vortex-fastlanes on `dev` profile

AdamGS:
## Summary

Inspired by @myrrc's #8634, this change makes a few subtle changes to our `dev` and `ci` build profiles. The main change here is removing debug symbols from `vortex-fastlanes`, which for a full build of the main package (`cargo build -p vortex --all-features --all-targets`) reduces build time by about 33% (from 90 seconds to under 60).

The others are:
1. For the `ci` profile, enables `debug_assertions` and remove the explicit `strip` which seems redundant, and doesn't seem to make a measurable difference in build times.
2. Remove `vortex-bench` as a dev dependency for `vortex`, which is just weird and unused.

### DIFF
--- a/vortex/Cargo.toml
+++ b/vortex/Cargo.toml
@@ -65,7 +65,6 @@ tokio = { workspace = true, features = ["full"] }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 vortex = { path = ".", features = ["tokio"] }
-vortex-bench = { workspace = true, features = ["unstable_encodings"] }
 vortex-tensor = { workspace = true }
 
 [features]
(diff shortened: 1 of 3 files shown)


robert3005 (approved):

myrrc (approved):

vortex/Cargo.toml
```diff
@@ -65,7 +65,6 @@ tokio = { workspace = true, features = ["full"] }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 vortex = { path = ".", features = ["tokio"] }
-vortex-bench = { workspace = true, features = ["unstable_encodings"] }
```
AdamGS, line 68: This is also potentially a pretty big speedup in many cases

joseph-isaacs (approved):

PR #8384, opened 2026-06-12
Install libc debug symbols for codspeed benchmarks

AdamGS:
## Summary

Install libc's debug symbols for codspeed runs, to get better symbolication for some of the code we end up calling.

For example, for the noisy `chunked_bool_canonical_into` we currently get (note the unknown spans on the left):
Screenshot 2026-06-12 at 09 03 39

With this change we get (note all the new blue spans, other colors are just noise):
Screenshot 2026-06-12 at 09 04 22

They include both memory allocation function but often more useful to us - SIMD instructions! On this PR's runs with a quick look I've found:
- `_int_malloc`
- `malloc_consolidate`
- `__memset_avx2_unaligned_erms`
- `__memcpy_avx_unaligned_erms`
- `__memcmp_avx2_movbe`
- `tcache_get_n`

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -51,6 +51,9 @@ jobs:
         run: sudo bash scripts/setup-benchmark.sh
       - uses: ./.github/actions/setup-prebuild
       - uses: ./.github/actions/system-info
+      - run: |
+          sudo apt-get update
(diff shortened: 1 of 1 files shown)


myrrc (approved):

0ax1 (approved):
❤️

PR #8388, opened 2026-06-12
Timeout slow tests

AdamGS:
## Summary

It seems like the rate of tests hanging has gone up significantly recently which is very wasteful. This config marks tests as slow at 30s, and times them out after 3 periods (AKA 90 seconds).

As far as I'm aware there's only 1 test that approaches 1 minute (also raised its priority), so this change should timeout any normal run.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,3 +1,6 @@
+[profile.default]
+slow-timeout = { period = "30s", terminate-after = 3 }
+
 [[profile.default.overrides]]
-filter = 'test(compress_large_int)'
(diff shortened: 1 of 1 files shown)


.config/nextest.toml
```diff
@@ -1,3 +1,6 @@
+[profile.default]
+slow-timeout = { period = "30s", terminate-after = 3 }
```
joseph-isaacs:
```suggestion
slow-timeout = { period = "30s", terminate-after = 5 }
```

joseph-isaacs (approved):

PR #8389, opened 2026-06-12
Return null for mean of all-null input

dimitarvdimitrov:
Reverts #8365: a mean over zero non-null values should be null (as in SQL, where nulls are skipped and an empty mean is null), not NaN.

Both reduce paths previously produced `0/0 = NaN` for an all-null group: `Mean::finalize` divided the sum and count arrays directly, and `Mean::finalize_scalar` computed `sum / count`. Now `finalize` masks out zero-count entries and `finalize_scalar` returns null for a zero count, so both paths return null.

NaN handling is unchanged; it stays controlled by `NumericalAggregateOpts::skip_nans` from #8457.

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/mean/mod.rs
+++ b/vortex-array/src/aggregate_fn/fns/mean/mod.rs
@@ -6,6 +6,7 @@ use vortex_error::vortex_bail;
 
 use crate::ArrayRef;
 use crate::ExecutionCtx;
+use crate::IntoArray;
 use crate::aggregate_fn::Accumulator;
 use crate::aggregate_fn::AggregateFnId;
 use crate::aggregate_fn::AggregateFnVTable;
@@ -17,6 +18,7 @@ use crate::aggregate_fn::combined::CombinedOptions;
 use crate::aggregate_fn::combined::PairOptions;
 use crate::aggregate_fn::fns::count::Count;
 use crate::aggregate_fn::fns::sum::Sum;
+use crate::arrays::ConstantArray;
 use crate::builtins::ArrayBuiltins;
 use crate::dtype::DType;
 use crate::dtype::Nullability;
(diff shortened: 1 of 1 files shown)


vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -89,7 +91,15 @@ impl BinaryCombined for Mean {
         };
         let sum_cast = sum.cast(target.clone())?;
         let count_cast = count.cast(target)?;
-        sum_cast.binary(count_cast, Operator::Div)
+        let mean = sum_cast.binary(count_cast, Operator::Div)?;
+        // Nulls are skipped during accumulation, so an all-null group has a count of zero and
+        // the division produces 0/0 = NaN. The mean of an empty group is null (as in SQL), so
+        // mask out zero-count entries. This matches `finalize_scalar`.
+        let non_empty = count.binary(
+            ConstantArray::new(0u64, count.len()).into_array(),
```
robert3005: doesn't this need to be `target` dtype?

dimitarvdimitrov:
sorry this was prematurely opened by claude. i'm not sure about the `mean(nan, 1) == 0.5`

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -89,7 +91,15 @@ impl BinaryCombined for Mean {
...
+            Operator::NotEq,
+        )?;
+        mean.mask(non_empty)
```
joseph-isaacs, line 108: ?

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -232,11 +241,84 @@ mod tests {
...
+    /// (`finalize_scalar`) and the grouped array path (`finalize`).
+    ///
+    /// Note that `Sum` skips NaN values while `Count` counts them as valid, so NaN elements
+    /// reduce the mean rather than poisoning it, e.g. `mean(NaN, 1, null) = 1/2`.
+    fn mean_cases() -> Vec<(Vec<Option<f64>>, Option<f64>)> {
+        vec![
+            (vec![Some(f64::NAN), Some(1.0), None], Some(0.5)),
+            (vec![Some(f64::NAN), Some(1.0), Some(1.0)], Some(2.0 / 3.0)),
+            (vec![None, None, Some(f64::NAN)], Some(0.0)),
+            (vec![Some(f64::NAN), Some(1.0), Some(1.0)], Some(2.0 / 3.0)),
```
joseph-isaacs: This looks wrong

robert3005:
I am going to drop this here https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.mean.html. I think long term we will need `skipna` flag (or similar) to make this correct we need count to skip nans as well

dimitarvdimitrov:
> I am going to drop this here [pandas.pydata.org/docs/reference/api/pandas.DataFrame.mean.html](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.mean.html). I think long term we will need `skipna` flag (or similar) to make this correct we need count to skip nans as well

i just found out about https://github.com/vortex-data/vortex/pull/7009, i think we'll need the option now so we have the default stats adhere to the IEEE standard and our stats be more useful. WIP

dimitarvdimitrov:
we talked about changing only stats to contain the sum as "sum of non-NaNs." But in order to use the sum stat, we have to now always look for the NanCount stat and compute it if it's not present.

are we ok with that perf change for old arrays? I suppose we could do the sum and the nan check in one pass and not change it too much, but that'd be a new accumulator implementation; do you guys have better ideas?

robert3005:
We need to compute it anyway to make this correct. I am working through adding skipNaN option to our aggregates

robert3005:
We have skip nan

PR #8391, opened 2026-06-12
Add JSON to Parquet Variant conversion with shredding

AdamGS:
## Summary

This PR adds a `vortex.json_to_variant` expression for converting UTF-8 or `vortex.json` extension arrays into Variant arrays.

The expression lives in `vortex-json` so JSON-facing expression metadata and serialization are independent of a concrete Variant encoding. `vortex-parquet-variant` provides the concrete execution kernels, parsing JSON through `parquet_variant_compute` and producing Parquet Variant storage.


mprammer (approved):

gatesn (dismissed):
One sec!

encodings/parquet-variant/src/fns/json_to_variant.rs
```diff
@@ -0,0 +1,570 @@
...
+/// # Not an inverse of `variant_to_json`
+///
+/// `json_to_variant` and [`VariantToJson`](crate::VariantToJson) are lossy, normalizing
+/// conversions and are NOT inverses of each other:
+/// - JSON whitespace is not preserved.
+/// - Object keys may be reordered: Variant metadata stores keys in sorted order, so
+///   `variant_to_json` emits fields in a canonical order, not source order.
+/// - Number formatting and precision change: e.g. `1.0` may render as `1`, exponent forms and
+///   very large numbers are re-rendered, and floating-point values are re-encoded.
+/// - Duplicate object keys are collapsed to a single entry.
```
connortsui20: Is the key value pair that gets removed well defined? Like is it whichever one comes first wins?

gatesn:
Do these functions mix DType (JSON) with encoding (ParquetVariant)? At least the docs seem to suggest that they do.

We probably want vortex-json to expose to_json and from_json functions (or maybe this just a cast kernel) and then implement them using execute_parent for ParquetVariant.

But it seems like you are actually defining a more general DType::JSON -> DType::Variant (any variant encoding) function? In which case, it cannot be a cast because you want the shredding options, but the definition should probably still live in the vortex-json crate, and implement general-purpose shredding (per the canonical Variant array), without touching the non-shredded data?

encodings/parquet-variant/src/fns/variant_to_json.rs
```diff
@@ -0,0 +1,503 @@
...
+    ) -> fmt::Result {
+        write!(f, "variant_to_json(")?;
+        expr.child(0).fmt_sql(f)?;
+        write!(f, ")")
+    }
+
+    fn return_dtype(&self, _options: &Self::Options, arg_dtypes: &[DType]) -> VortexResult<DType> {
+        let input_dtype = &arg_dtypes[0];
+        vortex_ensure!(
+            matches!(input_dtype, DType::Variant(_)),
```
connortsui20: we have an `is_variant` method now

encodings/parquet-variant/src/fns/variant_to_json.rs
```diff
@@ -0,0 +1,503 @@
...
+
+        let arrow_json = parquet_variant_compute::variant_to_json(&arrow_storage)?;
+        let storage = ArrayRef::from_arrow(&arrow_json, input_nullable)?;
+
+        ExtensionArray::try_new_from_vtable(Json, EmptyMetadata, storage).map(IntoArray::into_array)
+    }
+
+    fn is_null_sensitive(&self, _options: &Self::Options) -> bool {
+        false
+    }
```
connortsui20: are the null semantics of internal nulls and nested nulls (afaik these are not the same things) documented anywhere for the json and variant conversions?

encodings/parquet-variant/src/fns/json_to_variant.rs
```diff
@@ -0,0 +1,570 @@
...
+        match child_idx {
+            0 => ChildName::from("input"),
+            _ => unreachable!("Invalid child index {child_idx} for JsonToVariant expression"),
+        }
+    }
+
+    fn return_dtype(&self, _options: &Self::Options, arg_dtypes: &[DType]) -> VortexResult<DType> {
+        let input_dtype = &arg_dtypes[0];
+        vortex_ensure!(
+            input_dtype.is_utf8()
```
connortsui20: Do we really want Vortex to essentially "autocast" from UTF8 to JSON and then to Variant? IMO it makes more sense to wrap a UTF8 array in JSON (so the validation logic can live there) and then we are only allowed to shred the JSON after it has been validated (which fits nicely because the JSON storage already might already be shredded variant).

I guess it is kind of okay for now because json storage array _is_ just a utf8 array, but in that case we should be more intentional with our conversions.

connortsui20 (dismissed):

encodings/parquet-variant/src/fns/json_to_variant.rs
AdamGS: Json is an extension type, I don't see a reason to force users to go through it if they already have strings.

AdamGS:
@gatesn I've really tried to keep expressions that can't be evaluated directly from the core. I thought execute_parent rules were less flexible but I guess I can try and do that.

encodings/parquet-variant/src/fns/variant_to_json.rs
AdamGS: The difference between null and "VariantNull" are documented in a few places around variant and ParquetVariant.

connortsui20: sorry I meant that it might be worth justifying why this is not null sensitive (Im guessing because variant null semantics are going to be different from the top-level null semantics, but that might not be super obvious to everyone), because I am guessing this function is definitely _variant_ null sensitive

AdamGS:
I think I addressed the comments. I've also made the PR a bit smaller and more focused, just on `json_to_variant` which I think is the more interesting code path.

encodings/parquet-variant/src/kernel.rs
```diff
@@ -64,6 +72,12 @@ pub(crate) fn initialize(session: &VortexSession) {
         TakeExecuteAdaptor(ParquetVariant),
     );
     kernels.register_execute_parent_kernel(VariantGet.id(), ParquetVariant, VariantGetKernel);
+    kernels.register_execute_parent_kernel(JsonToVariant.id(), VarBinView, JsonToVariantKernel);
```
robert3005: imho this implicit flow is weird here, let's require cast to json first

robert3005 (changes requested):
I think the only thing is the implicit JsonToVariant for VarBinView. We should require an extension json type at this level.

robert3005 (approved):

PR #8395, opened 2026-06-12
Fix weird malformed license notice in test file

AdamGS:
## Summary

I was trying [hawkeye](https://github.com/korandoru/hawkeye) which has some nice features `reuse` doesn't have, and it found this weird issue.


vortex-array/src/arrays/datetime/test.rs
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// SPDX-FileCopyrightText: Copyright the Vortex contributorsuse crate::dtype::Nullability;
```
a10y, line 2: sorry, but think we need to contact `Vortex contributorsuse crate::dtype::Nullability;` and ask for their permission before we change the license

a10y (approved):

PR #8400, opened 2026-06-12
perf(scan): intra-file decode parallelism — sub-split large chunk spans

lukekim:
## Summary

`SplitBy::Layout` now sub-divides any span between adjacent chunk boundaries wider than `IDEAL_SPLIT_SIZE` (100k rows) into evenly sized row-range splits, so a file with few large chunks (e.g. a single flat layout, or byte-targeted int columns that coalesce to ~262k rows/chunk) decodes across multiple cores instead of one.

## Correctness

Subdivision only inserts points strictly between existing adjacent boundaries — it never moves or removes one — so the half-open ranges consumers derive (`tuple_windows`) remain a contiguous, non-overlapping, exact partition of the same rows. Spans at or below the cap pass through untouched (fast-path no-op). All boundary consumers (`RepeatedScan`, `VortexFile::splits()`, the DataFusion repartitioner) operate on arbitrary ranges; sub-chunk ranges were already exercised by `SplitBy::RowCount`. The arithmetic saturates at `u64::MAX`.

## API/Observable behavior

For files whose merged (projected-column) chunk boundaries leave spans > 100k rows, `VortexFile::splits()` (incl. Python bindings) returns more, smaller ranges; scans emit smaller batches; DataFusion gets real repartitioning where a single-chunk file previously collapsed to one partition. Fine-grained files (e.g. ~8k-row string chunks from the default 1 MiB block target) are untouched.

## Testing

- Unit/property/overflow tests for `subdivide_large_spans` (no-op, large single chunk, mixed gaps, exact-coverage property, `u64::MAX` boundary).
- E2E: 250k-row single flat chunk → splits all ≤ the cap, contiguous, exact endpoints; full + filtered scans match the unsplit data.
- E2E (rstest): fixed-size `SplitBy::RowCount` scans (unaligned 33,333 and exceeds-file 300,000 cases).
- E2E: ~120-byte string column via the default write strategy keeps its natural fine-grained chunk splits (bounded relative to the cap, not to writer defaults).
- `cargo nextest run -p vortex-layout -p vortex-file` — 174 passed on this branch.


robert3005 (approved):
Sorry for the delay reviewing it. I think this is good to go. You will need to rebase/merge develop

PR #8401, opened 2026-06-12
perf(executor): per-ExecutionCtx ArrayKernels snapshot

lukekim:
## Summary

Resolves `ArrayKernels` once at `ExecutionCtx` construction into a crate-private `KernelSnapshot` (a single `ArcSwapMap::load_full` of the execute-parent kernel map), instead of cloning the session and probing the sharded session-variable `DashMap` on every array node in the `execute_until` / single-step execution paths.

### Why

- Removes a per-array-node session clone + `DashMap` shard `RwLock` probe from the hot execution loop.
- Stops holding the session-variable read guard across kernel invocation (previously a plugin kernel touching the session registry could contend/deadlock on the same shard).

## Semantics

The registry is session-scoped and mutable via its public `register_*` methods. An `ExecutionCtx` sees a point-in-time snapshot taken at construction; later registrations are picked up by the next context (contexts are created per evaluation). Kernel lookup order is unchanged: registered plugin kernels are tried before static `execute_parent` kernels, with the same `(parent, child)` hashing.

Adds a `pub(crate)` `ArcSwapMap::load_full` accessor (snapshot-that-outlives-the-call, complementing `read`). `KernelSnapshot` and `ArrayKernels::snapshot()` are `pub(crate)` — no new public API.

## Testing

- `cargo nextest run -p vortex-array` — 2893 passed (includes `struct_cast_execute_parent_uses_session_plugin`, covering the register → snapshot → execute path).
- `cargo clippy --all-targets -p vortex-array` — clean on the spiceai-53 variant of this change; develop port re-verified via the full test build.


robert3005:
I think we can merge this. @joseph-isaacs you mentioned something about the execute loop optimisation, is this in the right direction?

robert3005 (approved):
I think this is the right direction. Apologies for the conflicts.

gatesn:
So I was actually refactoring execute_parent to be _only_ kernels, and ran into the exact same performance problem. I hope you don't mind that I inlined your fix into this PR https://github.com/vortex-data/vortex/pull/8482

PR #8403, opened 2026-06-13
feat: sequence cast compute

### LINKED ISSUE #5102: SequenceArray cannot cast to a narrower type

siddarth2810:
## Summary
This PR fixes `SequenceArray` casts where internal arithmetic needs to use a different type than the declared output dtype.

Previously, `SequenceArray` effectively used the same primitive type for both internal arithmetic and for exposing values. That broke cases like a negative-step signed sequence cast to an unsigned dtype, where every generated value still fits the unsigned output type.

Sequence operations now compute using that internal arithmetic type while materialized arrays and scalars are exposed using the array output dtype.

Added regression coverage for casts where `scalar_at` must return a scalar matching the array output dtype rather than the internal arithmetic type.

Closes: #5102 

## API Changes
None

## Review Feedback

Based on review feedback, this PR avoids storing an extra `calculation_ptype` in `SequenceMetadata`.

Instead, it determines a normalized arithmetic type from the sequence values:
`i64` when either `base` or `multiplier` is signed, otherwise `u64`.

## Testing
For testing code changes

```sh
cargo nextest run -p vortex-sequence
cargo +nightly fmt --all
cargo clippy --all-targets --all-features
```

For testing the build
```sh
cargo build --workspace
```

## AI tools Disclosure
Used ChatGPT for understanding code and OpenCode for updating callers and generating tests


connortsui20:
Don't worry about the rustsec issue, that is a known problem

gatesn (commented):
It's a good catch that this was broken before.

But why is it important to store this information? Vs just always computing in i64/u64 space?

siddarth2810:
> It's a good catch that this was broken before.
> 
> But why is it important to store this information? Vs just always computing in i64/u64 space?

Thanks for the quick review :)

The maintainer in the [previous PR](https://github.com/vortex-data/vortex/pull/7630) mentioned to use two ptypes, so I went ahead with that design in mind. 

But after experimenting a bit on this, I think we could get `calculation_ptype` from `base.ptype()`
instead of being passed around or stored separately. 

For deserialization, I think we can use scalar kind before decoding the scalar, so we may not need to store `calculation_ptype` in SequenceMetadata either.

encodings/sequence/src/array.rs
```diff
@@ -70,11 +71,16 @@ pub(super) const SLOT_NAMES: [&str; 0] = [];
 pub struct SequenceData {
     base: PValue,
     multiplier: PValue,
+    calculation_ptype: PType,
```
connortsui20: This is a bit outside of what this PR is trying to do, but could you document this now that we are adding a third type here that is different from the base and multiplier? With just base and multiplier it is obvious what this is doing, but with the addition of `calculation_ptype` this is now harder to understand on a first read. And if possible you could document other places that use this now, that would be great, thanks!

Edit: I am interested to see if your idea about not storing this at all works. That would probably be better for us since that is not a breaking change.

connortsui20 (dismissed):

encodings/sequence/src/array.rs
```diff
@@ -61,6 +60,8 @@ pub struct SequenceMetadata {
     base: Option<vortex_proto::scalar::ScalarValue>,
     #[prost(message, tag = "2")]
     multiplier: Option<vortex_proto::scalar::ScalarValue>,
+    #[prost(enumeration = "PType", optional, tag = "3")]
+    calculation_ptype: Option<i32>,
```
joseph-isaacs: can you explain why in a doc str why we need this, if we need this

encodings/sequence/src/array.rs
siddarth2810: My understanding of this issue was to use two types. `caclculation_ptype` for computing and `output_ptype` for the result type. But yeah, adding another type makes it less obvious in the first read, makes sense. 

I'l work on the idea of not storing this at all. 

Thanks a lot !

encodings/sequence/src/array.rs
siddarth2810: I was trying to decode the base and multiplier as `calculation_ptype` during deserialization, so I added this. But after Gates comments, I found that I could use `scalar_value::Kind` to get the type instead. I'll remove this in the next change

Thanks for the review :)

encodings/sequence/src/array.rs
gatesn, line 232: I think this entire module is a massive foot gun.... (not your fault!)

We should just be passing around Scalar values. Not PValue.

This PType used to return the output ptype (i.e. matches array.dtype().ptype()), but now it returns the ptype used to calculate the result values I think?

Can we remove this function maybe? Or rename it to calculation_ptype()? Make it pub(crate) and not pub? 

I think that's all I would change here!

siddarth2810: I see. Yep you're right, previously the PType used to return the output ptype and now it returns the ptype used to calculate the result values. 

I changed it to calculation_ptype and made it pub(crate).

Rebased and pushed the changes. Thanks so much for the review :)

encodings/sequence/src/array.rs
```diff
@@ -291,14 +336,22 @@ impl VTable for Sequence {
...
+            .as_ref()
+            .ok_or_else(|| vortex_err!("base required"))?;
+
+        let multiplier_metadata = metadata
+            .multiplier
+            .as_ref()
+            .ok_or_else(|| vortex_err!("multiplier required"))?;
+
+        let ptype =
+            SequenceData::infer_calculation_ptype_from_proto(base_metadata, multiplier_metadata)?;
```
joseph-isaacs: why do you find this before the base and mult below?

connortsui20:
Hi @siddarth2810, are you able to address the review comments? Please let us know if you have questions!

robert3005:
@connortsui20 I rebased this pr and addressed all of your comments

robert3005 (dismissed):

connortsui20:
This PR is pretty tricky. This is what claude had to say:

<details>

I reviewed PR 8403 (`feat: sequence cast compute`) by checking out the head commit into a worktree alongside its base (`4c40890`) and differentially testing both.

**What holds up:** the core idea works. I built a 1649-case matrix over (base PValue, multiplier PValue, output ptype, length) and checked canonical decompress, `scalar_at`, `take`, `filter`, `slice`, `min_max`, and serde round-trip against a naive `i128` model — zero mismatches, and zero cases that the base built but the PR does not (149 newly-representable ones, including the genuinely broken `u8` constant sequence longer than 256, which the base could not build at all). `compare`/`list_contains` also stay correct for every non-zero multiplier despite `compare.rs` not being touched. The 63 existing crate tests pass. `vortex-cuda` could not be compiled here (its `nvcomp` build script needs a CUDA download), so those changes are review-only.

Findings:

- `encodings/sequence/src/array.rs:154` — `infer_calculation_ptype` picks i64/u64 from PValue signedness alone, so any signed input forces i64 and the upper half of the u64 range becomes unrepresentable. `Sequence::try_new(PValue::I64(0), PValue::I64(1<<62), PType::U64, NonNullable, 4)` built `[0, 2^62, 2^63, 3*2^62]` on the base and now fails with "final value not expressible"; same for a `U64` base above `i64::MAX` with a signed multiplier. Confirmed regression.
- `encodings/sequence/src/compress.rs:79` — the per-element `NumCast::from(...).vortex_expect(...)` in the decompress loop costs 2.4x: 1M-element u32/i32 decompress goes 0.203 ms to 0.480 ms (release, best-of-30). Values are already proven in range, so an infallible `as_` conversion restores the vectorizable loop. Same pattern at `filter.rs:52` and `take.rs:44`/`:65`.
- `encodings/sequence/src/array.rs:327` — a `u8` sequence with multiplier -10 now writes `Int64Value(-10)` metadata; older readers decode it against the unsigned dtype and hard-error in `int64_from_proto` (`(-10).to_u8() == None`). Forward-incompatible file output with no version gate.
- `encodings/sequence/src/array.rs:210` — `SequenceData::ptype()` was a `pub` export (`vortex::encodings::sequence`) and is now `pub(crate) calculation_ptype()`; `SequenceDataParts::ptype` renamed with changed meaning and now has no reader at all. The PR body says "API Changes: None".
- `encodings/sequence/src/compress.rs:145` — the `try_last` "cannot encode" guard is now unreachable for any ptype narrower than 64 bits; harmless today because the `windows(2).checked_add` scan still catches wrapping runs, but the check and its comment are misleading.
- `encodings/sequence/src/array.rs:205` — the four new `vortex_expect("validated sequence values must fit output ptype")` sites make value-range containment a panic-enforced invariant, but `new_unchecked`'s safety contract never states it; a future unsafe-constructor caller panics mid-decode instead of getting an error.
- `encodings/sequence/src/compute/compare.rs:60` — **pre-existing, not this PR**: Eq against a zero-multiplier (constant) sequence sets only bit 0, so `Sequence(100, 0, len 5) == 100` returns `[true, false, false, false, false]`. Reproduced identically on base and head (810 mismatches, all multiplier 0). `list_contains` shares the helper and is wrong the same way. Worth a separate issue.
```

</details>

The semantics here are a bit tough to grok (and just at a quick glance it's hard for me to understand too since I haven't thought about this in a while). I can take a look at this again tomorrow in more depth to see if the claude review is correct.

robert3005:
No worries, I think I know how to fix it. This is a bit too naive. There’s real cost to doing computation in 64 bit space but first step is making sure we respect the output type.

robert3005:
~So... the practical things that changed here is that checks have to be made in i128 space to avoid values that are bigger than i64 but less than u64 failing.~ The base ptype stays as the array ptype, multiplier gets normalised to i64/u64 and the compute continues in the array dtype space so we shouldn't have any regressions. I need to figure out more cleanup, the arithmetic checking is a bit confusing still

we can keep it all in i64/u64

siddarth2810:
Thanks so much for picking this up. Apologies, I recently joined a fast paced startup and could not continue working on this. Apologies!

robert3005 (approved):

robert3005:
I have fixed up the followups, Thanks for the contribution!

PR #8406, opened 2026-06-14
fix(layout/dict): probe with the configured compressor instead of a hardcoded default

### LINKED ISSUE #8405: DictStrategy dict-fit probe ignores the configured compressor
### What happened?

`DictStrategy` decides whether to apply a dictionary *layout* to a column by probe-compressing the column's first chunk and checking whether the cascade chose a dictionary encoding (`compressed.is::<Dict>()`). That probe is hardcoded to a stock `BtrBlocksCompressor::default()`:

```rust
// vortex-layout/src/layouts/dict/writer.rs
let compressed = BtrBlocksCompressor::default().compress(&chunk, &mut exec_ctx)?;
!compressed.is::<Dict>()
```

So the dict-fit decision **ignores the compressor the caller configured** via `WriteStrategyBuilder::with_btrblocks_builder` / `with_compressor`. A caller who customizes the cascade — e.g. excluding a dictionary scheme — still gets a layout decision made by a compressor they didn't ask for.

This is observable in-tree: `vortex-cuda/gpu-scan-cli` writes with `BtrBlocksCompressorBuilder::default().only_cuda_compatible()`, which deliberately excludes `StringDictScheme` / `BinaryDictScheme` / `FSSTScheme`. Because the probe ignores that, a `vortex.dict` layout still fires for low-cardinality string columns; the GPU scan then skips Dict fields (`gpu-scan-cli/src/main.rs`, `if field.is::<Dict>() { continue; }`), so those columns silently drop off the pure-GPU path the config was built to keep them on.

### Steps to reproduce

1. Build a low-cardinality string column (e.g. 32,768 rows cycling `["alpha","beta","gamma"]`).
2. Write it twice through `SESSION.write_options().with_strategy(...)`:
   - **A:** `WriteStrategyBuilder::default().build()`
   - **B:** `WriteStrategyBuilder::default().with_btrblocks_builder(BtrBlocksCompressorBuilder::default().exclude_schemes([StringDictScheme.id()])).build()`
3. Walk each file's layout tree (`footer().layout()`) for a node whose `encoding_id() == "vortex.dict"`.
4. **Expected:** A has a dict layout, B falls back (no dict layout — StringDict was excluded). **Actual:** both A and B contain a `vortex.dict` layout, because the probe uses the hardcoded default regardless of B's configuration.

### Environment

- Vortex version: `develop` @ `9444d20ae` (source-level logic bug; not release-specific)
- Python/Java version: n/a
- OS: n/a (platform-independent)

### Additional context

I have a pull request/branch with this fix on it, will post shortly.

Discovery was encountered myself while experimenting. Reproduced, tracked and found the fix using coding agents.

tomsanbear:
## Summary

Closes: #8405

`DictStrategy`'s dict-fit probe was hardcoded to `BtrBlocksCompressor::default()`, ignoring the caller's configured compressor. The probe now takes `stats_compressor` as a parameter to `DictStrategy::new`, so caller scheme exclusions are honoured.

`stats_compressor` rather than `data_compressor` because the probe needs all dictionary schemes to detect eligibility. `data_compressor` excludes `IntDictScheme`.

## Testing

New test `dict_probe_honours_configured_compressor`: asserts dict layout with the default builder, asserts no dict layout with `StringDictScheme` excluded.

AI use disclosure: fix authored with assistance from Claude Code.


tomsanbear:
Open design question:

With this change, the probe runs the caller's opaque compressor instead of the stock default. This is empirically a no-op for every current opaque caller (all pass btrblocks-based compressors that emit Dict, or non-dict-favourable vector data). The only behaviour change would be for a hypothetical opaque compressor that never emits a top-level `Dict` while being fed low-cardinality data... it would stop triggering the dict layout. I chose this uniform behaviour because it matches the approach of "the probe reflects the caller's compressor" open to alternatives as i'm still new to the codebase.

vortex-file/src/strategy.rs
```diff
@@ -268,15 +268,20 @@ impl WriteStrategyBuilder {
...
         };
-        let compress_then_flat = CompressingStrategy::new(flat, stats_compressor);
+        let compress_then_flat = CompressingStrategy::new(flat, Arc::clone(&stats_compressor));
 
-        // 3. apply dict encoding or fallback
+        // 3. apply dict encoding or fallback.
+        // The dict-fit probe shares `stats_compressor` (the full configured cascade), not
+        // `data_compressor`: `data_compressor` drops `IntDictScheme` to avoid re-encoding the
+        // codes produced in step 5, but the probe needs every dict scheme to detect eligibility.
+        // The full cascade also honours caller scheme exclusions, unlike a hardcoded default.
```
onursatici: nit: remove this comment, I think it is a bit noisy given the declarations of stats and data compressors are self explanatory

vortex-file/src/strategy.rs
```diff
@@ -268,15 +268,20 @@ impl WriteStrategyBuilder {
...
         let dict = DictStrategy::new(
             coalescing.clone(),
             compress_then_flat.clone(),
             coalescing,
             Default::default(),
-        );
+        )
+        .with_probe_compressor(stats_compressor);
```
joseph-isaacs: Why use this compressor?

onursatici: shall we get this as an argument to `new` with the default being btrblocks?

onursatici (commented):
this makes sense to me

vortex-file/src/strategy.rs
tomsanbear: The probe compresses the first chunk and only checks is dict, the result is discarded. stats_compressor seems like the right fit since it has all dictionary schemes; the data compressor excludes intdictscheme to prevent double dict-encoding the codes that DictStrategy produces downstream

tomsanbear: makes sense, updated the branch with this change and removed the with_... method

onursatici: Maybe we can have a new compressor argument thar defaults to btrblocks instead of reusing the stats ine. I agree using data compressor doesn't make sense but maybe callers want a separate compressor for dict selection than the one used for stats

tomsanbear: I guess it's the tradeoff that callers have to explicitly choose which compressor probes dict eligibility, but that seems better than a default that silently ignores configured scheme exclusions, what do you think?

onursatici: Yea if users want to customise the probe compressor they should be able to change only that while constructing their write strategies. For the default case having the stats compressor works but there would probably be a need to change the probe compressor separately from the stats compressor

tomsanbear: Got it, that makes sense, I can make this change. I'll add a new `with_probe_compressor` and wire it through

tomsanbear: @onursatici addressed in https://github.com/vortex-data/vortex/pull/8406/changes/f15df5d01cecfd2b6a99d8661e055339fee42463

robert3005 (approved):

robert3005:
@tomsanbear you will need to fix the compilation. We can merge it after that

tomsanbear:
@robert3005 branch is now up to date with develop and fixed locally.

PR #8425, opened 2026-06-15
perf(mask): fuse true_count into boolean operators

### LINKED ISSUE #1508: Compute true_count while constructing/evaluating boolean operators
We almost always want to know true_count from the result of mask operations, so we may as well compute them while iterating over the u64s used within BooleanBuffer.

We probably need our own version of `BooleanBuffer::collect_bool` that does a popcnt for each u64.

miniex:
## Summary

Closes: #1508

Stacked on #8436, the bit collect half. the kernel change is reviewed there, so the diff here includes it until #8436 merges.

`Mask`'s boolean operators (`&`, `|`, `bitand_not`, `!`) built the result buffer
then re-scanned it with `true_count()`, walking the bits twice. They now fold
`count_ones` into the combine pass through a shared `bitwise_binary_op_with`, so
it is single-pass for every offset; `!mask` reuses the cached count, no popcount.

## API Changes

Adds three non-breaking `BitBuffer` methods: `bitand_with_true_count`,
`bitor_with_true_count`, `bitand_not_with_true_count`.

## Benchmark

`vortex-buffer/benches/bitbuffer_count.rs` compares two-pass vs fused. The gain
is ~neutral for small borrowed inputs and grows with size, largest on owned-LHS.

## Testing

`cargo test -p vortex-buffer -p vortex-mask` passes, plus a per-bit reference
test across offsets/lengths and a `Mask`-level regression test for #1508. It
also passes under `miri` (strict provenance). `fmt` + `clippy --all-targets
--all-features` clean.

---

### DIFF
--- a/vortex-buffer/benches/bitbuffer_count.rs
+++ b/vortex-buffer/benches/bitbuffer_count.rs
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Benchmarks comparing strategies for computing a bitwise-op result *and* its `true_count`:
+//!
+//! * `*_two_pass`: current behaviour, materialise the result buffer then run the (vectorised)
+//!   `count_ones` kernel as a separate pass (`&a & &b` followed by `.true_count()`).
+//! * `*_fused`: single pass that combines the words and accumulates `count_ones` together
+//!   (`BitBuffer::bitand_with_true_count`).
+
+#![allow(clippy::unwrap_used, clippy::cast_possible_truncation)]
+
+use divan::Bencher;
+use divan::black_box;
+use vortex_buffer::BitBuffer;
+
+fn main() {
+    divan::main();
+}
+
+// Lengths in bits, from L1-resident up to L2/L3.
+const SIZES: &[usize] = &[
+    1_024,   // 128 B
+    8_192,   // 1 KB
+    65_536,  // 8 KB
+    524_288, // 64 KB
+];
+
+fn inputs(len: usize) -> (BitBuffer, BitBuffer) {
+    let lhs =
+        BitBuffer::from_iter((0..len).map(|i| (i.wrapping_mul(2_654_435_761) >> 13) & 1 == 0));
+    let rhs = BitBuffer::from_iter((0..len).map(|i| (i.wrapping_mul(40_503) >> 7) & 1 == 0));
+    (lhs, rhs)
+}
+
+#[divan::bench(args = SIZES)]
+fn and_two_pass(bencher: Bencher, len: usize) {
+    let (lhs, rhs) = inputs(len);
+    bencher.bench_local(|| {
+        let result = black_box(&lhs) & black_box(&rhs);
+        let count = result.true_count();
+        (result, count)
+    });
+}
+
+#[divan::bench(args = SIZES)]
+fn and_fused(bencher: Bencher, len: usize) {
+    let (lhs, rhs) = inputs(len);
+    bencher.bench_local(|| black_box(&lhs).bitand_with_true_count(black_box(&rhs)));
+}
+
+// Owned-LHS scenario: the current code can combine in-place (no allocation) then count
+// separately; the fused path always allocates but does a single pass. A fresh owned, uniquely
+// owned LHS is cloned per iteration so the in-place fast path is actually taken.
+#[divan::bench(args = SIZES)]
+fn and_owned_two_pass(bencher: Bencher, len: usize) {
+    let (lhs, rhs) = inputs(len);
+    bencher
+        .with_inputs(|| lhs.clone())
+        .bench_local_values(|lhs| {
+            let result = lhs & black_box(&rhs);
+            let count = result.true_count();
+            (result, count)
+        });
+}
+
+#[divan::bench(args = SIZES)]
+fn and_owned_fused(bencher: Bencher, len: usize) {
+    let (lhs, rhs) = inputs(len);
+    bencher
+        .with_inputs(|| lhs.clone())
+        .bench_local_values(|lhs| lhs.bitand_with_true_count(black_box(&rhs)));
+}
--- a/vortex-buffer/src/bit/ops.rs
+++ b/vortex-buffer/src/bit/ops.rs
@@ -235,6 +235,113 @@ pub(super) fn bitwise_binary_op<F: FnMut(u64, u64) -> u64>(
     BitBuffer::new(bytes.freeze(), len)
 }
 
+/// Combine `left` and `right` word-by-word with `op`, calling `comb` on each result word so a
+/// caller can fold a reduction (e.g. `count_ones`) into the same pass. The word given to `comb`
+/// has bits past `len` cleared; the buffer keeps `op`'s raw output in those positions.
+pub(super) fn bitwise_binary_op_with<F, C>(
+    left: &BitBuffer,
+    right: &BitBuffer,
+    op: F,
+    comb: C,
+) -> BitBuffer
+where
+    F: FnMut(u64, u64) -> u64,
+    C: FnMut(u64),
+{
+    assert_eq!(left.len(), right.len());
+    let len = left.len();
+    if len == 0 {
+        return BitBuffer::empty();
+    }
+
+    // Byte-aligned operands: logical bits map onto physical `u64` words, so we can read the backing
+    // bytes straight as words instead of shifting through `iter_padded`.
+    if left.offset().is_multiple_of(8) && right.offset().is_multiple_of(8) {
+        let n_bytes = len.div_ceil(8);
+        let l_start = left.offset() / 8;
+        let r_start = right.offset() / 8;
+        let lhs = &left.inner().as_slice()[l_start..l_start + n_bytes];
+        let rhs = &right.inner().as_slice()[r_start..r_start + n_bytes];
+
+        let (lhs_words, lhs_tail) = lhs.as_chunks::<8>();
+        let (rhs_words, rhs_tail) = rhs.as_chunks::<8>();
+
+        let words = lhs_words
+            .iter()
+            .zip(rhs_words)
+            .map(|(l, r)| (u64::from_le_bytes(*l), u64::from_le_bytes(*r)))
+            .chain((!lhs_tail.is_empty()).then(|| (read_u64_le(lhs_tail), read_u64_le(rhs_tail))));
+        return combine_words(words, len, op, comb);
+    }
+
+    // Sub-byte offset: `iter_padded` realigns the bits and zero-pads the final word.
+    let words = left
+        .chunks()
+        .iter_padded()
+        .zip(right.chunks().iter_padded());
+    combine_words(words, len, op, comb)
+}
+
+/// Combine an iterator of `(left, right)` words via `op` into a `len`-bit [`BitBuffer`], folding
+/// `comb` over each in-range result word. Consumes exactly `ceil(len / 64)` words (any extra the
+/// producer yields, e.g. `iter_padded`'s trailing pad word, is dropped). Only the final word can
+/// hold bits past `len`, so it alone is masked for `comb`; the buffer keeps `op`'s raw output.
+#[inline]
+fn combine_words<I, F, C>(words: I, len: usize, mut op: F, mut comb: C) -> BitBuffer
+where
+    I: Iterator<Item = (u64, u64)>,
+    F: FnMut(u64, u64) -> u64,
+    C: FnMut(u64),
+{
+    let n_words = len.div_ceil(64);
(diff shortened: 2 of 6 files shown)


vortex-buffer/benches/bitbuffer_count.rs
joseph-isaacs, line 1: can you reduce the number of benchmarks we add, also all must be under 1ms

vortex-buffer/src/bit/ops.rs
```diff
@@ -220,6 +221,68 @@ pub(super) fn bitwise_binary_op<F: FnMut(u64, u64) -> u64>(
...
+    let len = left.len();
+    if len == 0 {
+        return (BitBuffer::empty(), 0);
+    }
+
+    if left.offset() != 0 || right.offset() != 0 {
+        let buffer = bitwise_binary_op(left, right, op);
+        let true_count = buffer.true_count();
+        return (buffer, true_count);
+    }
```
joseph-isaacs: why do you fail here?

vortex-buffer/benches/bitbuffer_count.rs
joseph-isaacs, line 1: can you also read the benchmarking guidelines

vortex-buffer/src/bit/ops.rs
joseph-isaacs, line 1: this is a great, insight.

```
pub(super) fn bitwise_binary_op_counted<F: FnMut(u64, u64) -> u64, C: u64 -> ()>(
    left: &BitBuffer,
    right: &BitBuffer,
    mut op: F,
    mut comb: C
) -> (BitBuffer) {
```

Then use |_| {} for the original and |c| {count += c} for the counted one.

The compile will drop the original unit comb function, then we can share the copy of the op.

joseph-isaacs, line 1: I have a feeling part of your speed up is a better binary_op impl for BitBuffer would be nice to see that.

miniex: Ah, I learned a lot here, thank you. I will merge to one `binary_op` with a `comb: FnMut(u64)` - `|_| {}` for plain op (compiler drop it) and `|w| count += w.count_ones()` for counted, so the offset case is also gone. And you are right, some of the speedup is just a better kernel not the fused count, so I move it into the shared `binary_op` and bench again.

vortex-buffer/src/bit/ops.rs
miniex: yeah you are right, fallback is still two pass. will fold into the combiner loop and drop it.

vortex-buffer/benches/bitbuffer_count.rs
miniex: yep will do, trim it under 1ms.

miniex:
pushed the combiner version and dropped the fallback. can you check if i got your point right? also i ran the bench on a different machine than before, so the numbers are not 1:1 with the old table.

vortex-buffer/src/bit/ops.rs
```diff
@@ -185,39 +184,114 @@ pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
...
+
+        let full = lhs_words.len();
+        for (i, (l, r)) in lhs_words.iter().zip(rhs_words).enumerate() {
+            let word = op(u64::from_le_bytes(*l), u64::from_le_bytes(*r));
+            // Only the last word can hold bits past `len`; clear them for `comb`.
+            let observed = if lhs_tail.is_empty() && i + 1 == full {
+                word & last_word_mask(len)
+            } else {
+                word
+            };
```
joseph-isaacs: move this out of the hot loop

vortex-buffer/src/bit/ops.rs
```diff
@@ -185,39 +184,114 @@ pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
...
+            comb(observed);
+            // SAFETY: capacity reserved for `n_words` words, each written exactly once.
+            unsafe { out.push_unchecked(word) };
```
joseph-isaacs: use either a mut iterator

vortex-buffer/src/bit/ops.rs
```diff
@@ -185,39 +184,114 @@ pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
...
+
+        let (lhs_words, lhs_tail) = lhs.as_chunks::<8>();
+        let (rhs_words, rhs_tail) = rhs.as_chunks::<8>();
+
+        let full = lhs_words.len();
+        for (i, (l, r)) in lhs_words.iter().zip(rhs_words).enumerate() {
+            let word = op(u64::from_le_bytes(*l), u64::from_le_bytes(*r));
+            // Only the last word can hold bits past `len`; clear them for `comb`.
+            let observed = if lhs_tail.is_empty() && i + 1 == full {
+                word & last_word_mask(len)
```
joseph-isaacs: does this actually get hit, seeing as you have a tail below to handle this too?

joseph-isaacs (commented):
Why are there two loops?

vortex-buffer/src/bit/ops.rs
```diff
@@ -185,39 +184,114 @@ pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
...
+        unsafe { out.push_unchecked(word) };
+    }
+    BitBuffer::new(out.into_byte_buffer().freeze(), len)
+}
 
-    let result = Buffer::<u64>::from_trusted_len_iter(iter).into_byte_buffer();
+/// Mask of the in-range bits in the final word: the low `len % 64` bits, or all 64 when `len` is
+/// a multiple of 64.
+#[inline]
+fn last_word_mask(len: usize) -> u64 {
```
joseph-isaacs, line 322: This seems like it should only be used once for the tail value?

vortex-buffer/src/bit/ops.rs
miniex: yeah it's hit when n_bytes is a multiple of 8 but len isn't a multiple of 64, so no tail runs but the last word still has bits past len

miniex:
> Why are there two loops?

byte-aligned reads the backing bytes directly as u64 for the fast kernel, sub-byte offsets straddle byte boundaries so they need iter_padded to realign.

vortex-buffer/src/bit/ops.rs
miniex: it's the final word not just the tail, same as [the other thread](https://github.com/vortex-data/vortex/pull/8425#discussion_r3413704471). it ends up a single call once i merge the last word and tail into one step.

vortex-buffer/src/bit/ops.rs
miniex: yup

vortex-buffer/src/bit/ops.rs
miniex: yup, will do

miniex:
pushed the updates for your comments. mask moved out of the hot loop, only the final word is masked now, and the unsafe push is gone. can you check if i got it right? also miri caught an out-of-bounds push in the sub-byte path where iter_padded appends one extra word, so i capped it to ceil(len/64).

vortex-buffer/src/bit/ops.rs
```diff
@@ -185,39 +184,117 @@ pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
...
+
+        let (lhs_words, lhs_tail) = lhs.as_chunks::<8>();
+        let (rhs_words, rhs_tail) = rhs.as_chunks::<8>();
+
+        let words = lhs_words
+            .iter()
+            .zip(rhs_words)
+            .map(|(l, r)| (u64::from_le_bytes(*l), u64::from_le_bytes(*r)))
+            .chain((!lhs_tail.is_empty()).then(|| (read_u64_le(lhs_tail), read_u64_le(rhs_tail))));
+        return combine_words(words, len, op, comb);
```
joseph-isaacs, line 274: Do you find `lhs.as_chunks::<8>()` is more performant to `BitChunkIterator`

miniex: for the sizes that actually hit this, yeah. the buffer kernel only runs when both sides are mixed masks, all-true and all-false short-circuit. those are chunk-sized, and as_chunks is the fastest of the three there. it also matches the other bit kernels here, count_ones and select both read via as_chunks::<8> not BitChunkIterator.

`&` at offset 0, as_chunks vs bare BitChunkIterator vs iter_padded:

bench

as_chunks vs bare .iter() is close to a wash and bare edges ahead once it goes memory-bound. chunk size is not capped so a mask can get large, but as_chunks is never much worse and stays consistent with count_ones and select.

<details>
<summary>bench code</summary>

```rust
use divan::Bencher;
use divan::black_box;
use vortex_buffer::BitBuffer;

fn main() {
    divan::main();
}

const SIZES: &[usize] = &[1_024, 8_192, 65_536, 524_288, 4_194_304, 33_554_432];

fn inputs(len: usize) -> (BitBuffer, BitBuffer) {
    let a = BitBuffer::from_iter((0..len).map(|i| (i.wrapping_mul(2_654_435_761) >> 13) & 1 == 0));
    let b = BitBuffer::from_iter((0..len).map(|i| (i.wrapping_mul(40_503) >> 7) & 1 == 0));
    (a, b)
}

fn read_u64_le(bytes: &[u8]) -> u64 {
    let mut buf = [0u8; 8];
    buf[..bytes.len()].copy_from_slice(bytes);
    u64::from_le_bytes(buf)
}

fn combine_as_chunks(a: &BitBuffer, b: &BitBuffer, len: usize) -> Vec<u64> {
    let nbytes = len.div_ceil(8);
    let la = &a.inner().as_slice()[..nbytes];
    let lb = &b.inner().as_slice()[..nbytes];
    let (aw, at) = la.as_chunks::<8>();
    let (bw, bt) = lb.as_chunks::<8>();
    let mut out = Vec::with_capacity(len.div_ceil(64));
    for (x, y) in aw.iter().zip(bw) {
        out.push(u64::from_le_bytes(*x) & u64::from_le_bytes(*y));
    }
    if !at.is_empty() {
        out.push(read_u64_le(at) & read_u64_le(bt));
    }
    out
}

fn combine_bare_iter(a: &BitBuffer, b: &BitBuffer, len: usize) -> Vec<u64> {
    let mut out = Vec::with_capacity(len.div_ceil(64));
    for (x, y) in a.chunks().iter().zip(b.chunks().iter()) {
        out.push(x & y);
    }
    if a.chunks().remainder_len() != 0 {
        out.push(a.chunks().remainder_bits() & b.chunks().remainder_bits());
    }
    out
}

fn combine_bitchunks(a: &BitBuffer, b: &BitBuffer, len: usize) -> Vec<u64> {
    let n_words = len.div_ceil(64);
    let mut out = Vec::with_capacity(n_words);
    for (x, y) in a
        .chunks()
        .iter_padded()
        .zip(b.chunks().iter_padded())
        .take(n_words)
    {
        out.push(x & y);
    }
    out
}

#[divan::bench(args = SIZES)]
fn aligned_as_chunks(bencher: Bencher, len: usize) {
    let (a, b) = inputs(len);
    bencher.bench_local(|| combine_as_chunks(black_box(&a), black_box(&b), len));
}

#[divan::bench(args = SIZES)]
fn aligned_bare_iter(bencher: Bencher, len: usize) {
    let (a, b) = inputs(len);
    bencher.bench_local(|| combine_bare_iter(black_box(&a), black_box(&b), len));
}

#[divan::bench(args = SIZES)]
fn aligned_bitchunks(bencher: Bencher, len: usize) {
    let (a, b) = inputs(len);
    bencher.bench_local(|| combine_bitchunks(black_box(&a), black_box(&b), len));
}
```

</details>

joseph-isaacs (commented):
This is a really great result. 

Do you think we could break out the changes to the bit collect into its own pr and keep the collect+count in this pr. 

then we can examine each perf change in isolation

miniex:
> This is a really great result.
> 
> Do you think we could break out the changes to the bit collect into its own pr and keep the collect+count in this pr.
> 
> then we can examine each perf change in isolation

thanks! makes sense. will split the bit collect into its own pr and keep collect+count here

miniex:
done, split into #8436

miniex:
done, rebased onto #8436.

joseph-isaacs:
Do we see a perf with when fusing count and pack?

miniex:
> Do we see a perf with when fusing count and pack?

no, right now the fused seems slower than two-pass.. after making the kernel fast with the iterator, the vectorized two-pass beats the fused loop.. so fusion doesn't really seem to give a win.. i think it'd be better to close this PR and keep it as the two-pass approach and flesh it out more..! my head was all over the place so i didn't think of it.. thanks!

joseph-isaacs:
I would expect a small win for large bit buffers. I am happy with the other commit

PR #8426, opened 2026-06-15
ffi: create data source from memory buffer

myrrc:
Match rust API: buffer is borrowed under the condition it won't be modified
until data source is freed.

We need this for ClickHouse integration as it hands us memory buffers.

### DIFF
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -885,6 +885,21 @@ const vx_data_source *
 vx_data_source_new(const vx_session *session, const vx_data_source_options *options, vx_error **err);
 
 /**
+ * Create a data source from a single in-memory Vortex file.
+ *
+ * "buffer_len" is the length of "buffer" in bytes.
+ * The bytes are borrowed, not copied: the caller must keep "buffer" alive and
+ * unmodified until the data source is freed.
+ *
+ * The returned pointer is owned by the caller and must be freed with
+ * vx_data_source_free.
+ *
+ * On error, returns NULL and sets "err".
+ */
+const vx_data_source *
+vx_data_source_new_buffer(const vx_session *session, const void *buffer, size_t buffer_len, vx_error **err);
+
+/**
  * Return the schema of the data source as a non-owned dtype.
  * The returned pointer is valid as long as "ds" is alive. Do not free it.
  */
--- a/vortex-ffi/src/data_source.rs
+++ b/vortex-ffi/src/data_source.rs
@@ -5,16 +5,22 @@
 
 use std::ffi::CStr;
 use std::ffi::c_char;
+use std::ffi::c_void;
 use std::ptr;
+use std::slice;
 use std::sync::Arc;
 
+use bytes::Bytes;
+use vortex::buffer::ByteBuffer;
 use vortex::error::VortexResult;
 use vortex::error::vortex_ensure;
 use vortex::expr::stats::Precision::Absent;
 use vortex::expr::stats::Precision::Exact;
 use vortex::expr::stats::Precision::Inexact;
+use vortex::file::OpenOptionsSessionExt;
 use vortex::file::multi::MultiFileDataSource;
 use vortex::io::runtime::BlockingRuntime;
+use vortex::layout::scan::multi::MultiLayoutDataSource;
 use vortex::scan::DataSource;
 use vortex::scan::DataSourceRef;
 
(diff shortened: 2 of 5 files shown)


AdamGS (approved):
Seems reasonable to me, but maybe needs someone with more C/FFI opinions.

vortex-ffi/cinclude/vortex.h
```diff
@@ -884,6 +884,19 @@ void vx_data_source_free(const vx_data_source *ptr);
 const vx_data_source *
 vx_data_source_new(const vx_session *session, const vx_data_source_options *options, vx_error **err);
 
+/**
+ * Create a data source from a single in-memory Vortex file.
```
0ax1, line 888: nit: maybe nice to add a new line after the opening sentence if that is the short summary
```
 * Create a data source from a single in-memory Vortex file.
 * 
 * <more text>
 ```

File: vortex-ffi/cinclude/vortex.h [hunk 2]
```diff
@@ -884,6 +884,19 @@ void vx_data_source_free(const vx_data_source *ptr);
...
+ * "buffer_len" is the length of "buffer" in bytes. The bytes are borrowed, not
+ * copied: the caller must keep "buffer" alive and unmodified until the data
+ * source is freed.
+ * The returned pointer is owned by the caller and must be freed with
```
0ax1, line 894: For comments that are not separated by an empty line, format with (vim speak) `visual select gq`? The comments look a bit shot.

vortex-ffi/cinclude/vortex.h
```diff
@@ -884,6 +884,19 @@ void vx_data_source_free(const vx_data_source *ptr);
...
+ * vx_data_source_free.
```
0ax1, line 895: In C++ code, backticks for symbol names?

vortex-ffi/src/data_source.rs
```diff
@@ -87,6 +93,25 @@ unsafe fn data_source_new(
...
+    vortex_ensure!(!buffer.is_null());
+    let session = vx_session::as_ref(session);
+
+    let bytes: &'static [u8] = unsafe { slice::from_raw_parts(buffer.cast::<u8>(), buffer_len) };
+    let file = session
+        .open_options()
+        .open_buffer(ByteBuffer::from(Bytes::from_static(bytes)))?;
+
+    let data_source =
+        MultiLayoutDataSource::new_with_first(file.layout_reader()?, Vec::new(), session);
```
0ax1: Not sure on `new_with_first` fn name, but not introduced as part of your PR.

vortex-ffi/src/data_source.rs
```diff
@@ -104,6 +129,26 @@ pub unsafe extern "C-unwind" fn vx_data_source_new(
...
+/// Create a data source from a single in-memory Vortex file.
+/// "buffer_len" is the length of "buffer" in bytes. The bytes are borrowed, not
+/// copied: the caller must keep "buffer" alive and unmodified until the data
+/// source is freed.
+/// The returned pointer is owned by the caller and must be freed with
+/// vx_data_source_free.
+///
+/// On error, returns NULL and sets "err".
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_data_source_new_buffer(
```
0ax1, line 124: Did we decide to keep `C-unwind` after all? I vaguely recall you mentioning we didn't want to use that.

vortex-ffi/src/data_source.rs
```diff
@@ -104,6 +129,26 @@ pub unsafe extern "C-unwind" fn vx_data_source_new(
...
+    session: *const vx_session,
+    buffer: *const c_void,
+    buffer_len: usize,
+    err: *mut *mut vx_error,
+) -> *const vx_data_source {
+    try_or(err, ptr::null(), || unsafe {
+        data_source_new_buffer(session, buffer, buffer_len)
```
0ax1: Is it a smell that we always have 2 fn? Here `vx_data_source_new_buffer` & `data_source_new_buffer`? But I'm seeing you're using the other one for tests and it's not public.

vortex-ffi/test/scan.cpp
```diff
@@ -218,6 +219,11 @@ TEST_CASE("Creating datasources", "[datasource]") {
     REQUIRE(error != nullptr);
     vx_error_free(error);
 
+    ds = vx_data_source_new_buffer(session, nullptr, 0, &error);
+    REQUIRE(ds == nullptr);
+    REQUIRE(error != nullptr);
+    vx_error_free(error);
```
0ax1, line 225: Should we comment here why we expect the error to be `nullptr`?

0ax1 (commented):
Couple of nits and questions.

vortex-ffi/test/scan.cpp
```diff
@@ -439,6 +429,50 @@ TEST_CASE("Basic scan", "[datasource]") {
     verify_sample_array(array);
 }
 
+TEST_CASE("Basic scan", "[datasource]") {
+    vx_session *session = vx_session_new();
+    defer {
```
0ax1, line 434: _wild_ :)

vortex-ffi/src/data_source.rs
```diff
@@ -87,6 +93,25 @@ unsafe fn data_source_new(
...
+    buffer_len: usize,
+) -> VortexResult<*const vx_data_source> {
+    vortex_ensure!(!session.is_null());
+    vortex_ensure!(!buffer.is_null());
+    let session = vx_session::as_ref(session);
+
+    let bytes: &'static [u8] = unsafe { slice::from_raw_parts(buffer.cast::<u8>(), buffer_len) };
+    let file = session
+        .open_options()
+        .open_buffer(ByteBuffer::from(Bytes::from_static(bytes)))?;
```
0ax1: `from_static` does not increase any ref count right? So the pointer returned from the ffi fn needs to be fried but the bytes are borrowed?

vortex-ffi/src/data_source.rs
myrrc: +1

vortex-ffi/test/scan.cpp
myrrc: No, it's documented in the function comment already that it sets the pointer to non-nullptr if there is an error.

vortex-ffi/src/data_source.rs
myrrc: Yes, and this is documented in the function comment. We don't want to copy the buffer if we can borrow it, but as we can't do C-Rust memory interop the only way how we can do it is this way.

0ax1 (approved):

vortex-ffi/src/data_source.rs
```diff
@@ -104,6 +110,38 @@ pub unsafe extern "C-unwind" fn vx_data_source_new(
...
+    buffer: *const c_void,
+    buffer_len: usize,
+    err: *mut *mut vx_error,
+) -> *const vx_data_source {
+    try_or(err, ptr::null(), || {
+        vortex_ensure!(!session.is_null());
+        vortex_ensure!(!buffer.is_null());
+
+        let session = vx_session::as_ref(session);
+        let bytes: &'static [u8] =
```
m7kss1, line 135: this approach implies that the data is already in ram, but in ch read buffer is a streaming abstraction, including positional read api over s3/ssd etc. wouldn't it be more reasonable to expose`VortexReadAt` to ffi?

myrrc: I agree, but this is a separate work which is non-trivial as opposed to this function. I have it on my TODO list. The issue with this is that we already have an inverse abstraction (duckdbfs) which we probably want to pull out.

myrrc: https://github.com/vortex-data/vortex/issues/8433

PR #8427, opened 2026-06-15
Yet another renovate configuration

AdamGS:
## Summary

Yet another iteration of our renovate config, including the following changes:
1. Removes some already-default base configs or moved things there.
2. Lock files updated weekly, everything else every second Monday
3. JS/NPM - only update the lockfile and security updates
4. The lance benchmarks have their own group, so they shouldn't impact other dependencies

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,34 +2,30 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "config:recommended",
-    ":automergeStableNonMajor",
-    ":automergePr",
     ":automergeRequireAllStatusChecks",
-    ":combinePatchMinorReleases",
-    ":dependencyDashboard",
     ":separateMultipleMajorReleases",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -2,34 +2,30 @@
...
     "helpers:pinGitHubActionDigests",
-    "schedule:earlyMondays",
-    "docker:disable"
+    "docker:disable",
+    ":maintainLockFilesWeekly"
   ],
-  "lockFileMaintenance": {
-    "enabled": true
-  },
+  "schedule": "* 0-3 * * 1#2,1#4",
```
AdamGS, line 13: Non standard, but is supported by the underlying cron package (and renovate even tests that it works).

robert3005 (approved):
renovate has inscrutable docs to me given the number of options in existence. I assume you have thought about these

AdamGS:
I've tried and I think I got it right

PR #8436, opened 2026-06-15
perf(buffer): read byte-aligned bit operands directly as words

miniex:
## Summary

Split out of #8425 so each perf change can be looked at on its own, as suggested by @joseph-isaacs. This is the "bit collect" half, no count fusion.

`bitwise_binary_op` only took the direct word path when both operands were 64-bit aligned, otherwise falling back to `iter_padded`. It now reads the backing bytes as `u64` via `as_chunks::<8>` for any byte-aligned offset, leaving `iter_padded` for sub-byte offsets only. This also drops the `from_trusted_len_iter` unsafe for a safe `BufferMut` push.

## Testing

`cargo test -p vortex-buffer` passes, plus a per-bit reference test (`binary_op_matches_naive`) across operand offsets and lengths. It also passes under `miri` (strict provenance). `fmt` + `clippy --all-targets --all-features` clean.

---

### DIFF
--- a/vortex-buffer/src/bit/ops.rs
+++ b/vortex-buffer/src/bit/ops.rs
@@ -5,9 +5,8 @@ use std::mem::MaybeUninit;
 
 use crate::BitBuffer;
 use crate::BitBufferMut;
-use crate::Buffer;
+use crate::BufferMut;
 use crate::ByteBufferMut;
-use crate::trusted_len::TrustedLenExt;
 
 /// Read up to 8 bytes as a little-endian `u64`, zero-padding the high bytes when fewer than 8 are
 /// supplied. Using [`u64::from_le_bytes`] keeps the bit-numbering identical on little- and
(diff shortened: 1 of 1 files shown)


vortex-buffer/src/bit/ops.rs
```diff
@@ -185,39 +184,59 @@ pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
...
     assert_eq!(left.len(), right.len());
+    let len = left.len();
+    if len == 0 {
+        return BitBuffer::empty();
+    }
 
-    // If the buffers are aligned, we can use the fast path.
+    // Byte-aligned operands: logical bits map onto physical `u64` words, so we can read the backing
+    // bytes straight as words instead of shifting through `iter_padded`.
     if left.offset().is_multiple_of(8) && right.offset().is_multiple_of(8) {
```
joseph-isaacs:
```suggestion
    // bytes straight as words instead of shifting through `iter_padded`.
    // TODO(joe): make this into an iterator.
if left.offset().is_multiple_of(8) && right.offset().is_multiple_of(8) {
```
Can you add a todo to make this into an iterator

joseph-isaacs: Or maybe do this?

vortex-buffer/src/bit/ops.rs
```diff
@@ -185,39 +184,59 @@ pub(super) fn bitwise_binary_op_lhs_owned<F: FnMut(u64, u64) -> u64>(
...
+
+        let (lhs_words, lhs_tail) = lhs.as_chunks::<8>();
+        let (rhs_words, rhs_tail) = rhs.as_chunks::<8>();
+
+        let words = lhs_words
+            .iter()
+            .zip(rhs_words)
+            .map(|(l, r)| (u64::from_le_bytes(*l), u64::from_le_bytes(*r)))
+            .chain((!lhs_tail.is_empty()).then(|| (read_u64_le(lhs_tail), read_u64_le(rhs_tail))));
+        return combine_words(words, len, op);
```
joseph-isaacs: I think this will make LLVM generate better code

miniex: ok, i check it!

joseph-isaacs:
I just merged in a fix to another crate from develop

miniex:
done, made it a TrustedLen iterator

joseph-isaacs:
I was talking about the input too, You really have found the difference between an aligned and unaligned byte iterator. Lets land this and maybe we can have a follow up

joseph-isaacs (approved):
Thanks for this. I wonder if we can work this into an iterator left and right without a perf loss?

miniex:
> Thanks for this. I wonder if we can work this into an iterator left and right without a perf loss?

thanks for merging! lets dig into that in a follow up.

also #8425 is the collect+count half stacked on this. i'll rebase it onto develop now that this landed so it's just the fuse, then it's ready for review.

PR #8439, opened 2026-06-16
ci: use ctest

joseph-isaacs:
This seems better

### DIFF
--- a/.github/workflows/rust-instrumented.yml
+++ b/.github/workflows/rust-instrumented.yml
@@ -221,7 +221,8 @@ jobs:
       - name: Run tests
         run: |
           set -o pipefail
-          ./vortex-ffi/build/test/vortex_ffi_test 2>&1 | rustfilt
+          ctest --test-dir vortex-ffi/build -N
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-instrumented.yml
```diff
@@ -221,7 +221,8 @@ jobs:
       - name: Run tests
         run: |
           set -o pipefail
-          ./vortex-ffi/build/test/vortex_ffi_test 2>&1 | rustfilt
+          ctest --test-dir vortex-ffi/build -N
```
myrrc, line 224: We've already built the binary

.github/workflows/rust-instrumented.yml
```diff
@@ -221,7 +221,8 @@ jobs:
...
+          ctest --test-dir vortex-ffi/build --output-on-failure --timeout 120 -V 2>&1 | rustfilt
```
myrrc, line 225: ctest convolutes the output on errors, so I'd not do this

myrrc (changes requested):
I don't think we should merge this

Issue #8445, opened 2026-06-16
Migrate all `VTable::execute_parent` rewrites onto the ArrayKernels registry


joseph-isaacs:
Parent rewrites in the executor currently come from two sources: kernels registered on `ArrayKernels` (looked up `via find_execute_parent(parent_id, child_id)`) and the per-encoding `VTable::execute_parent`. 

We should only have `ArrayKernels`, so we can customize the execution of any array, see https://github.com/vortex-data/vortex/pull/7588
  
Concretely this is two pieces of work. 

 ## Move each execute_parent kernels to a registered `ArrayKernels`.

  ### Core arrays (`vortex-array/src/arrays/`)
  - [ ] bool — `bool/vtable/mod.rs:177`
  - [ ] chunked — `chunked/vtable/mod.rs:266`
  - [ ] decimal — `decimal/vtable/mod.rs:196`
  - [ ] dict — `dict/vtable/mod.rs:208`
  - [ ] extension — `extension/vtable/mod.rs:179`
  - [ ] filter — `filter/vtable.rs:174`
  - [ ] fixed_size_list — `fixed_size_list/vtable/mod.rs:87`
  - [ ] list — `list/vtable/mod.rs:195`
  - [ ] listview — `listview/vtable/mod.rs:223`
  - [ ] patched — `patched/vtable/mod.rs:304`
  - [ ] primitive — `primitive/vtable/mod.rs:195`
  - [ ] struct_ — `struct_/vtable/mod.rs:196`
  - [ ] varbin — `varbin/vtable/mod.rs:185`
  - [ ] varbinview — `varbinview/vtable/mod.rs:220`
  - [ ] list — `list/vtable/mod.rs:195`
  - [ ] listview — `listview/vtable/mod.rs:223`
  - [ ] patched — `patched/vtable/mod.rs:304`
  - [ ] primitive — `primitive/vtable/mod.rs:195`
  - [ ] struct_ — `struct_/vtable/mod.rs:196`
  - [ ] varbin — `varbin/vtable/mod.rs:185`
  - [ ] varbinview — `varbinview/vtable/mod.rs:220`
  - [ ] variant — `variant/vtable/mod.rs:196`

  ### Encodings (`encodings/`)
  - [ ] alp — `alp/src/alp/array.rs:192`
  - [ ] alp_rd — `alp/src/alp_rd/array.rs:311`
  - [ ] bytebool — `bytebool/src/array.rs:161`
[...]

PR #8448, opened 2026-06-16
perf: executor use execute_parent kernel registry in `execute_until`

joseph-isaacs:
As part of https://github.com/vortex-data/vortex/issues/8445 we need to use execute_parent kernels for in `execute_until`. 

This PR does that change


vortex-array/src/executor.rs
```diff
@@ -164,6 +163,9 @@ impl ArrayRef {
         let mut stack: Vec<StackFrame> = Vec::new();
         let max_iterations = max_iterations();
 
+        let session = ctx.session().clone();
+        let kernels = session.get::<ArrayKernels>();
```
robert3005: I would really like to get rid of this `get` method. The fact that it's a read that can cause a write can be very surprising

robert3005 (approved):

vortex-array/src/executor.rs
joseph-isaacs: Only once, but ye

gatesn:
Closed by #8482

PR #8449, opened 2026-06-16
fix: cuda san always run

joseph-isaacs:
We forgot to ensure that the CUDA san checks should pass if vortex-cuda is no touched..


0ax1 (approved):

robert3005:
What was wrong here?

PR #8453, opened 2026-06-16
feat(vortex-datafusion): struct scalar conversion + extension-over-struct scan

HarukiMoriarty:
## Summary

1. DataFusion and Vortex can now exchange struct-shaped scalars.
2. Scan can resolve columns whose type is an extension over a struct.

## Testing

add null/non-null struct scalar round-trips.


vortex-datafusion/src/convert/schema.rs
```diff
@@ -204,6 +205,16 @@ fn calculate_physical_field_type(
     })
 }
 
+/// The struct fields of `dtype` if it is a struct, or of an extension type whose storage is
+/// (eventually) a struct -- e.g. the native geo Point over `Struct<x, y>`.
+fn struct_fields(dtype: &DType) -> Option<&StructFields> {
```
AdamGS: nit - this is only used once, can we just inline it?

vortex-datafusion/src/convert/scalars.rs
```diff
@@ -113,7 +120,42 @@ impl TryToDataFusion<ScalarValue> for Scalar {
             ),
             DType::List(..) => todo!("list scalar conversion"),
             DType::FixedSizeList(..) => todo!("fixed-size list scalar conversion"),
-            DType::Struct(..) => todo!("struct scalar conversion"),
+            DType::Struct(struct_fields, _) => {
+                let scalar = self.as_struct();
+                let (fields, arrays): (Vec<Field>, Vec<_>) = struct_fields
```
connortsui20: nit: might as well split this whole branch out into a separate function?

vortex-datafusion/src/convert/scalars.rs
```diff
@@ -288,6 +330,28 @@ impl FromDataFusion<ScalarValue> for Scalar {
                 }
             }
             ScalarValue::Dictionary(_, v) => Scalar::from_df(v.as_ref()),
+            ScalarValue::Struct(array) => {
+                let nullable = array.is_null(0);
```
AdamGS: the struct array here has a `DataType` with the nullability info, I think using it is much clearer.

AdamGS: ok reading down this is not `nullable`, this is an actual check if its null.

vortex-datafusion/src/convert/scalars.rs
```diff
@@ -288,6 +330,28 @@ impl FromDataFusion<ScalarValue> for Scalar {
...
+                if nullable {
+                    Scalar::null(dtype)
+                } else {
+                    let children = array
+                        .columns()
+                        .iter()
+                        .map(|column| {
+                            Scalar::from_df(
+                                &ScalarValue::try_from_array(&**column, 0).unwrap_or_else(|e| {
+                                    unimplemented!(
```
AdamGS: Should return an error, `unimplemented` is semantically code that hasn't been implemented yet (and it panics).

vortex-datafusion/src/convert/scalars.rs
```diff
@@ -288,6 +330,28 @@ impl FromDataFusion<ScalarValue> for Scalar {
...
+                let dtype = DType::from_arrow((array.data_type(), nullable.into()));
+                if nullable {
+                    Scalar::null(dtype)
+                } else {
+                    let children = array
+                        .columns()
+                        .iter()
+                        .map(|column| {
+                            Scalar::from_df(
+                                &ScalarValue::try_from_array(&**column, 0).unwrap_or_else(|e| {
```
AdamGS: `&**column` - might be correct, but there's almost certainly a more idiomatic way to express that.

vortex-datafusion/src/convert/scalars.rs
AdamGS: I think you're using it for both, it should be two different things, one derived from the `DataType` and one from the data itself.

vortex-datafusion/src/convert/scalars.rs
HarukiMoriarty: `from_df` is infallible by design — `fn from_df(value: &ScalarValue) -> Scalar` returns a
`Scalar`, not a `Result` — so we can't return an `Err` here. For the (effectively unreachable)
"struct child can't convert" case, a panic is the only option.

a10y (approved):

AdamGS (approved):

PR #8456, opened 2026-06-16
bench(spatialbench): SpatialBench Q1 on DataFusion-over-Vortex

HarukiMoriarty:
## Status: 

temporary fork dependency, so the target branch is not develop

## Summary

A **SpatialBench** suite (Apache Sedona's geospatial ride-sharing benchmark) built on top of DataFusion. It runs Q1 four ways, `{parquet, vortex} × {wkb, native}`, to show where the speedup comes from.

- **Query:** Q1 — trips within a radius of a point, ranked by distance.
- **Wired:** the `trip` table + Q1; dimension tables and other queries are catalog-ready, not yet wired.
- **Datasets**: `wkb.rs` generates the canonical WKB base; `native.rs` derives the native encodings by decoding WKB → GeoArrow points **in Arrow** (`geoarrow_cast`, so Vortex carries no WKB code), writing a Vortex `Point` file and a GeoParquet file.
- Some glues.

## Results (Q1, scale factor 1, release)

| | Geometry storage | Compute path | Time | vs fastest |
|---|---|---|---|---|
| **vortex-native** | native `Point` extension (columnar x/y) | `GeoDistance` fused into the Vortex scan | **5.7 ms** | **1.0×** |
| vortex-wkb | WKB bytes, Vortex format | geodatafusion above the scan (no pushdown) | 46 ms | 8.1× |
| parquet-native | GeoParquet `geoarrow.point` | geodatafusion per-row UDF, filter in parquet scan | 52 ms | 9.1× |
| parquet-wkb | WKB bytes, Parquet | WKB parse + geodatafusion per-row above the scan UDF | 103 ms | 18× |

## Why the others are slow
**parquet-wkb** (103 ms): the geometry is stored as raw WKB bytes, and the query wraps it in `ST_GeomFromWKB(pickup)`. DataFusion pushes the radius filter down into the parquet reader as a row-filter, but to actually evaluate it the reader has to decode every WKB blob into a geometry and then call geodatafusion's `ST_Distance` UDF once per row. Both of those costs grow with the number of rows, which is why it's the slowest.

**parquet-native** (52 ms): GeoParquet stores the points natively as `geoarrow.point`, and `skip_metadata=false` keeps that extension so geodatafusion recognizes the column as a geometry, so the WKB-parsing cost disappears. But the distance is still computed by geodatafusion's per-row UDF: Parquet has no geo kernel inside the scan, so even with filter pushdown it still materializes the points into Arrow and dispatches the UDF row by row. Dropping the parse is why it beats parquet-wkb, but the per-row compute keeps it well behind Vortex.

**vortex-wkb** (46 ms): this is Vortex without the pushdown. The predicate can't lower into the scan because its operand is `ST_GeomFromWKB(t_pickuploc)`, which the converter doesn't recognize as something it can translate, so the whole filter stays above the scan and geodatafusion computes it. The columnar format still reads faster than Parquet, but the geo work happens above the scan.

**vortex-native** (5.7 ms): both blockers are gone. With `points=native` the query drops the `ST_GeomFromWKB` wrapper, so the operand is just a plain Point column the converter can translate, and the column itself is the native Point extension that GeoDistance knows how to compute over. The distance is fused right into the scan and evaluated over the columnar x and y as they're read, with the filter applied in-scan, so there's no WKB parsing and no UDF round-trip, and only the handful of matching rows ever materialize.

### DIFF
--- a/benchmarks/datafusion-bench/src/lib.rs
+++ b/benchmarks/datafusion-bench/src/lib.rs
@@ -45,13 +45,23 @@ pub fn get_session_context() -> SessionContext {
         .build_arc()
         .expect("could not build runtime environment");
 
-    let factory = VortexFormatFactory::new().with_options(VortexTableOptions {
-        projection_pushdown: true,
-        ..Default::default()
-    });
+    let factory = VortexFormatFactory::new_with_options(
+        SESSION.clone(),
+        VortexTableOptions {
+            projection_pushdown: true,
+            ..Default::default()
+        },
+    );
+
+    let mut config = SessionConfig::from_env().expect("shouldn't fail");
+    // Keep Parquet field metadata so the geoarrow.point extension survives the read.
+    config.options_mut().execution.parquet.skip_metadata = false;
+    // Evaluate (and reorder) the filter inside the parquet scan -- fairest parquet baseline.
+    config.options_mut().execution.parquet.pushdown_filters = true;
+    config.options_mut().execution.parquet.reorder_filters = true;
 
     let mut session_state_builder = SessionStateBuilder::new()
-        .with_config(SessionConfig::from_env().expect("shouldn't fail"))
+        .with_config(config)
         .with_runtime_env(rt)
         .with_default_features();
 
@@ -66,7 +76,10 @@ pub fn get_session_context() -> SessionContext {
         file_formats.push(Arc::new(factory));
     }
 
-    SessionContext::new_with_state(session_state_builder.build())
+    let ctx = SessionContext::new_with_state(session_state_builder.build());
+    // Register geodatafusion's PostGIS-style ST_* UDFs so SpatialBench SQL plans.
+    geodatafusion::register(&ctx);
+    ctx
 }
 
 pub fn make_object_store(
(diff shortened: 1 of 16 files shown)


vortex-bench/spatialbench.sql
joseph-isaacs, line 1: can we comment out the first and uncomment 1-by-1

benchmarks/datafusion-bench/src/lib.rs
```diff
@@ -45,13 +45,23 @@ pub fn get_session_context() -> SessionContext {
...
+            ..Default::default()
+        },
+    );
+
+    let mut config = SessionConfig::from_env().expect("shouldn't fail");
+    // Keep Parquet field metadata so the geoarrow.point extension survives the read.
+    config.options_mut().execution.parquet.skip_metadata = false;
+    // Evaluate (and reorder) the filter inside the parquet scan -- fairest parquet baseline.
+    config.options_mut().execution.parquet.pushdown_filters = true;
+    config.options_mut().execution.parquet.reorder_filters = true;
```
joseph-isaacs: will this affect other benchmarks?

vortex-bench/spatialbench.sql
HarukiMoriarty: Sure!

benchmarks/datafusion-bench/src/lib.rs
HarukiMoriarty: I scope it to only apply for spatial bench.

PR #8457, opened 2026-06-17
Numerical aggregate functions have an option to skip or include nans in calculation, skip by default

robert3005:
Almost all of the time you want to skip nans but for rare cases when you don't
we need to be able to configure it

### DIFF
--- a/encodings/runend/src/compute/min_max.rs
+++ b/encodings/runend/src/compute/min_max.rs
@@ -6,7 +6,7 @@ use vortex_array::ExecutionCtx;
 use vortex_array::aggregate_fn::AggregateFnRef;
 use vortex_array::aggregate_fn::fns::min_max::MinMax;
 use vortex_array::aggregate_fn::fns::min_max::make_minmax_dtype;
-use vortex_array::aggregate_fn::fns::min_max::min_max;
+use vortex_array::aggregate_fn::fns::min_max::min_max_with;
 use vortex_array::aggregate_fn::kernels::DynAggregateKernel;
 use vortex_array::scalar::Scalar;
 use vortex_error::VortexResult;
@@ -28,16 +28,16 @@ impl DynAggregateKernel for RunEndMinMaxKernel {
         batch: &ArrayRef,
         ctx: &mut ExecutionCtx,
     ) -> VortexResult<Option<Scalar>> {
-        if !aggregate_fn.is::<MinMax>() {
+        let Some(options) = aggregate_fn.as_opt::<MinMax>() else {
             return Ok(None);
-        }
+        };
 
         let Some(run_end) = batch.as_opt::<RunEnd>() else {
             return Ok(None);
         };
 
         let struct_dtype = make_minmax_dtype(batch.dtype());
-        match min_max(run_end.values(), ctx)? {
+        match min_max_with(run_end.values(), ctx, *options)? {
             Some(result) => Ok(Some(Scalar::struct_(
                 struct_dtype,
                 vec![result.min, result.max],
--- a/encodings/runend/src/compute/min_max.rs
+++ b/encodings/runend/src/compute/min_max.rs
@@ -6,7 +6,7 @@ use vortex_array::ExecutionCtx;
 use vortex_array::aggregate_fn::AggregateFnRef;
 use vortex_array::aggregate_fn::fns::min_max::MinMax;
 use vortex_array::aggregate_fn::fns::min_max::make_minmax_dtype;
-use vortex_array::aggregate_fn::fns::min_max::min_max_with;
+use vortex_array::aggregate_fn::fns::min_max::min_max;
 use vortex_array::aggregate_fn::kernels::DynAggregateKernel;
 use vortex_array::scalar::Scalar;
 use vortex_error::VortexResult;
@@ -37,7 +37,7 @@ impl DynAggregateKernel for RunEndMinMaxKernel {
         };
 
         let struct_dtype = make_minmax_dtype(batch.dtype());
-        match min_max_with(run_end.values(), ctx, *options)? {
+        match min_max(run_end.values(), ctx, *options)? {
             Some(result) => Ok(Some(Scalar::struct_(
                 struct_dtype,
                 vec![result.min, result.max],
--- a/encodings/sequence/src/compute/min_max.rs
+++ b/encodings/sequence/src/compute/min_max.rs
@@ -32,14 +32,11 @@ impl DynAggregateKernel for SequenceMinMaxKernel {
         batch: &ArrayRef,
         _ctx: &mut ExecutionCtx,
     ) -> VortexResult<Option<Scalar>> {
-        let Some(options) = aggregate_fn.as_opt::<MinMax>() else {
-            return Ok(None);
-        };
-        // The algebraic extrema assume NaN-skipping semantics; decline NaN-including requests
-        // over float sequences.
-        if !options.skip_nans && batch.dtype().is_float() {
+        if !aggregate_fn.is::<MinMax>() {
             return Ok(None);
         }
+        // Sequence arrays are always integer-typed (float multipliers are unsupported), so NaN
+        // handling never applies and `SkipNansOptions` can be ignored.
 
         let Some(seq) = batch.as_opt::<Sequence>() else {
             return Ok(None);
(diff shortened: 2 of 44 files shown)


encodings/runend/src/compute/min_max.rs
```diff
@@ -28,16 +28,16 @@ impl DynAggregateKernel for RunEndMinMaxKernel {
...
-        }
+        };
 
         let Some(run_end) = batch.as_opt::<RunEnd>() else {
             return Ok(None);
         };
 
         let struct_dtype = make_minmax_dtype(batch.dtype());
-        match min_max(run_end.values(), ctx)? {
+        match min_max_with(run_end.values(), ctx, *options)? {
```
robert3005: should keep the old name

robert3005:
Need to make sure that stats are only ever written with skipnan: true

encodings/sequence/src/compute/min_max.rs
```diff
@@ -32,7 +32,12 @@ impl DynAggregateKernel for SequenceMinMaxKernel {
         batch: &ArrayRef,
         _ctx: &mut ExecutionCtx,
     ) -> VortexResult<Option<Scalar>> {
-        if !aggregate_fn.is::<MinMax>() {
+        let Some(options) = aggregate_fn.as_opt::<MinMax>() else {
+            return Ok(None);
+        };
+        // The algebraic extrema assume NaN-skipping semantics; decline NaN-including requests
+        // over float sequences.
+        if !options.skip_nans && batch.dtype().is_float() {
```
robert3005: remove this, it's always int

vortex-array/src/aggregate_fn/fns/max/mod.rs
```diff
@@ -36,15 +46,32 @@ impl MaxPartial {
...
+        }
+
         self.max = Some(match self.max.take() {
             Some(current) => partial_max(max, current).vortex_expect("incomparable max scalars"),
             None => max,
         });
     }
+
+    fn poison(&mut self) {
+        self.max = Some(nan_scalar(&self.element_dtype.as_nonnullable()));
```
robert3005: this should already be nonnullable

vortex-array/src/aggregate_fn/fns/max/mod.rs
```diff
@@ -107,8 +139,37 @@ impl AggregateFnVTable for Max {
...
+        // NaN-aware shortcircuits only apply to the NaN-including float maximum; everything else
+        // takes the default dispatch path.
+        if partial.skip_nans || !partial.element_dtype.is_float() {
+            return Ok(false);
+        }
+        match batch.statistics().get_as::<u64>(Stat::NaNCount) {
+            Precision::Exact(0) => {
+                // NaN-free batch: the cached NaN-skipping maximum (if any) is valid.
+                if let Some(max) = batch.statistics().get(Stat::Max).as_exact() {
+                    partial.merge(max.cast(&partial.element_dtype.as_nonnullable())?);
```
robert3005: we don't need this cast

vortex-array/src/aggregate_fn/fns/min/mod.rs
```diff
@@ -36,15 +46,32 @@ impl MinPartial {
...
+        }
+
         self.min = Some(match self.min.take() {
             Some(current) => partial_min(min, current).vortex_expect("incomparable min scalars"),
             None => min,
         });
     }
+
+    fn poison(&mut self) {
+        self.min = Some(nan_scalar(&self.element_dtype.as_nonnullable()));
```
robert3005: what's up with these casts?

vortex-array/src/aggregate_fn/fns/min/mod.rs
```diff
@@ -107,8 +139,37 @@ impl AggregateFnVTable for Min {
...
+        // NaN-aware shortcircuits only apply to the NaN-including float minimum; everything else
+        // takes the default dispatch path.
+        if partial.skip_nans || !partial.element_dtype.is_float() {
+            return Ok(false);
+        }
+        match batch.statistics().get_as::<u64>(Stat::NaNCount) {
+            Precision::Exact(0) => {
+                // NaN-free batch: the cached NaN-skipping minimum (if any) is valid.
+                if let Some(min) = batch.statistics().get(Stat::Min).as_exact() {
+                    partial.merge(min.cast(&partial.element_dtype.as_nonnullable())?);
```
robert3005: shouldn't need to cast

vortex-array/src/aggregate_fn/fns/min_max/mod.rs
```diff
@@ -146,6 +225,18 @@ impl MinMaxPartial {
             None => max,
         });
     }
+
+    /// Poison the partial state to `{min: NaN, max: NaN}`.
+    fn poison(&mut self) {
+        let nan = nan_scalar(&self.element_dtype.as_nonnullable());
```
robert3005: no dtype change

vortex-array/src/stats/expr.rs
```diff
@@ -29,12 +30,12 @@ pub fn stat(expr: Expression, aggregate_fn: AggregateFnRef) -> Expression {
 
 /// Creates `stat(expr, min_max)`, returning a nullable `{ min, max }` struct statistic.
 pub fn min_max(expr: Expression) -> Expression {
-    stat(expr, MinMax.bind(EmptyOptions))
+    stat(expr, MinMax.bind(SkipNansOptions::default()))
```
robert3005: aggregates from stats should explicitly specify the options to skip nans

vortex-array/src/expr/stats/mod.rs
```diff
@@ -186,17 +187,18 @@ impl Stat {
                     .return_dtype(&EmptyOptions, data_type);
             }
             Self::Sum => {
-                return aggregate_fn::fns::sum::Sum.return_dtype(&EmptyOptions, data_type);
+                return aggregate_fn::fns::sum::Sum
+                    .return_dtype(&SkipNansOptions::default(), data_type);
```
robert3005: This should explicitly specify skinans

encodings/sequence/src/compute/min_max.rs
```diff
@@ -35,6 +35,8 @@ impl DynAggregateKernel for SequenceMinMaxKernel {
         if !aggregate_fn.is::<MinMax>() {
             return Ok(None);
         }
+        // Sequence arrays are always integer-typed (float multipliers are unsupported), so NaN
+        // handling never applies and `SkipNansOptions` can be ignored.
```
joseph-isaacs, line 38: ?

robert3005: this comment is a leftover

fuzz/src/array/min_max.rs
```diff
@@ -13,5 +14,5 @@ pub fn min_max_canonical_array(
     canonical: Canonical,
     ctx: &mut ExecutionCtx,
 ) -> VortexResult<Option<MinMaxResult>> {
-    min_max(&canonical.into_array(), ctx)
+    min_max(&canonical.into_array(), ctx, SkipNansOptions::default())
```
joseph-isaacs: We should not do it in this PR, but the common behaviour for lots of system is to include nan

datafusion:
```sql
select max(t) from (values (2),(3), (0.0/0.0) as t(t) );
+----------+
| max(t.t) |
+----------+
| NaN      |
+----------+
```

joseph-isaacs (commented):
Its good to have options, stats should ignore NaN but the default should be to propergate NaN

Can do this in the future

fuzz/src/array/min_max.rs
robert3005: Yes, I think skipping nans is the ultimately more useful behaviour though

vortex-array/src/aggregate_fn/fns/count/mod.rs
```diff
@@ -80,7 +95,12 @@ impl AggregateFnVTable for Count {
         batch: &ArrayRef,
         ctx: &mut ExecutionCtx,
     ) -> VortexResult<bool> {
-        *state += batch.valid_count(ctx)? as u64;
+        let mut count = batch.valid_count(ctx)? as u64;
+        if state.exclude_nans {
+            // `nan_count` shortcircuits on an exact `Stat::NaNCount` before scanning the batch.
+            count -= nan_count(batch, ctx)? as u64;
```
onursatici: could this underflow on a deliberately wrong vortex file that reports nan_count > size?

onursatici: I think it would, but we trust stats on other places as well. But on the release build this would wrap to report a large count so maybe we should to saturating_sub here?

robert3005: I think at worst you will get a wrong count

robert3005: made it a saturating sub but release rust would just wrap around anyway

onursatici (approved):

PR #8462, opened 2026-06-17
Upgrade pyO3 to 0.29.0

robert3005:
This lets us get over the security advisory


AdamGS (approved):
If CI is happy, I'm happy

PR #8465, opened 2026-06-17
Minimize default sql benchmarks suite

AdamGS:
## Summary

This PR changes the default benchmark suite that runs when adding `action/benchmark-sql` by removing the following benchmarks from the baseline:
1. Appian and TPC-H SF=10 on S3 (two slowest benchmarks)
2. Any `duckdb:duckdb` and `vortex-compact` runs

The full run is unchanged, and will still run post-merge or when using the `action/benchmark-sql-full` label.

Happy to remove more things, but I think these are the least contentious targets we can start with.

### DIFF
--- a/.github/workflows/sql-benchmarks.yml
+++ b/.github/workflows/sql-benchmarks.yml
@@ -277,9 +277,54 @@ on:
               "iterations": "10"
             }
           ]
+      benchmark_profile:
+        required: false
+        type: string
+        description: "Benchmark profile to run: full or base"
+        default: "full"
 
 jobs:
+  resolve-benchmark-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      benchmark_matrix: ${{ steps.matrix.outputs.benchmark_matrix }}
+    steps:
+      - name: Resolve benchmark matrix
+        id: matrix
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/sql-benchmarks.yml
```diff
@@ -277,9 +277,54 @@ on:
...
+              echo "::error::Unknown benchmark_profile '${{ inputs.benchmark_profile }}'. Expected 'full' or 'base'."
+              exit 1
+              ;;
+          esac
+
+          {
+            echo "benchmark_matrix<<__BENCHMARK_MATRIX__"
+            cat resolved-matrix.json
+            echo "__BENCHMARK_MATRIX__"
+          } >> "$GITHUB_OUTPUT"
```
joseph-isaacs: is this overly complex? should be have a way to write a list of strings and have those be converted into bench run commands?

joseph-isaacs (commented):
Like the idea

.github/workflows/sql-benchmarks.yml
AdamGS: we have a JSON format, this is a transformation on top of it, I can also use that.

AdamGS:
This PR is cursed CI just keeps getting queued

robert3005 (approved):

PR #7486, opened 2026-04-16
perf: fsst decompress

joseph-isaacs:
## Summary

Closes: #000

PR #8467, opened 2026-06-17
Run-end encode boolean columns via the generic RunEnd encoding

joseph-isaacs:
## Summary

Boolean columns are now run-end compressed through the **generic `RunEnd`** encoding rather than a bespoke bool encoding. The generic crate already *decodes* bool-valued run-end arrays (`decompress_bool.rs`, and the `DType::Bool` branch of `run_end_canonicalize`); the only missing piece was the *encode* side, which this PR adds.

This supersedes the earlier approach in this PR (a standalone `vortex-runend-bool` encoding). As discussed on the design, `RunEndBool` is really just `RunEnd` whose values are the implicit alternating pattern `[start, !start, …]`, so it is better expressed by composing the generic encoding than by duplicating its compute surface.

### Changes

1. **`vortex-runend` — bool encode path** (`compress.rs`):
   - `runend_encode_bool(ArrayView<Bool>, ctx) -> (ends, values)` mirrors the existing primitive encoders' validity handling — a new run starts whenever the `(value, validity)` pair changes, and per-run validity lives in the `BoolArray` values child.
   - `RunEnd::encode` now accepts `Bool` in addition to `Primitive`.

2. **`vortex-btrblocks` — auto-selection** (`schemes/bool/runend.rs`):
   - New `BoolRunEndScheme` emits a generic `RunEnd{ends, BoolArray values}` and cascades both children. Registered in `ALL_SCHEMES`; gated on average run length (word-at-a-time run counting via `set_slices`).

### Design note (future work)

The generic form materializes an `num_runs`-length alternating values child that is almost entirely redundant. A doc note on `runend_encode_bool` records that a future cycling/repeat-pattern array encoding — one that logically yields `[start, !start, …]` in O(1) space — could stand in as that values child, recovering the single-`start`-bit compactness of a dedicated bool run-end encoding through composition rather than a separate encoding.

### Tests / checks

- `cargo test -p vortex-runend` — 78 passed (new bool encode round-trips incl. interior-null cases + existing decode).
- `cargo test -p vortex-btrblocks` — 39 passed, incl. `test_run_heavy_bool_uses_runend` (run-heavy bool selects generic `RunEnd`) and existing bool/constant tests.
- `cargo clippy -p vortex-runend -p vortex-btrblocks` clean; `cargo +nightly fmt` clean.
- Not run locally: `vortex-duckdb` C++ layer (proxy blocks the DuckDB source download); its Rust uses only the unchanged generic `RunEndArray`.

### DIFF
--- a/encodings/runend/src/array.rs
+++ b/encodings/runend/src/array.rs
@@ -21,6 +21,7 @@ use vortex_array::ExecutionResult;
 use vortex_array::IntoArray;
 use vortex_array::TypedArrayRef;
 use vortex_array::VortexSessionExecute;
+use vortex_array::arrays::Bool;
 use vortex_array::arrays::Primitive;
 use vortex_array::arrays::VarBinViewArray;
 use vortex_array::buffer::BufferHandle;
@@ -44,6 +45,7 @@ use vortex_session::registry::CachedId;
 use crate::compress::runend_decode_primitive;
 use crate::compress::runend_decode_varbinview;
 use crate::compress::runend_encode;
+use crate::compress::runend_encode_bool;
 use crate::decompress_bool::runend_decode_bools;
 use crate::ops::find_physical_index;
 use crate::ops::find_slice_end_index;
@@ -308,8 +310,16 @@ impl RunEnd {
             let slots = smallvec![Some(ends), Some(values)];
             let data = unsafe { RunEndData::new_unchecked(0) };
             Array::try_from_parts(ArrayParts::new(RunEnd, dtype, len, data).with_slots(slots))
+        } else if let Some(barray) = array.as_opt::<Bool>() {
+            let (ends, values) = runend_encode_bool(barray, ctx);
+            let ends = ends.into_array();
+            let len = array.len();
+            let dtype = values.dtype().clone();
+            let slots = smallvec![Some(ends), Some(values)];
+            let data = unsafe { RunEndData::new_unchecked(0) };
+            Array::try_from_parts(ArrayParts::new(RunEnd, dtype, len, data).with_slots(slots))
         } else {
-            vortex_bail!("REE can only encode primitive arrays")
+            vortex_bail!("REE can only encode primitive or bool arrays")
         }
     }
 }
--- a/encodings/runend/src/array.rs
+++ b/encodings/runend/src/array.rs
@@ -22,6 +22,7 @@ use vortex_array::IntoArray;
 use vortex_array::TypedArrayRef;
 use vortex_array::VortexSessionExecute;
 use vortex_array::array_slots;
+use vortex_array::arrays::Bool;
 use vortex_array::arrays::DecimalArray;
 use vortex_array::arrays::ListViewArray;
 use vortex_array::arrays::Primitive;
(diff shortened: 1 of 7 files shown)


joseph-isaacs:
CI status on `3e685c5` — both current failures look external to this PR. Details so a reviewer doesn't have to re-derive them:

**CodSpeed "Performance Regression: -3.26%"** — not attributable to this PR. The two regressed benchmarks are `vortex-array/benches/filter_fixed_width.rs::random_i16[0.95]` and `vortex-compute/benches/lane_kernels.rs::arrow_checked_add_u32_avx2[16384]`. This PR's diff is 7 files confined to `encodings/runend/` and `vortex-btrblocks/` — it touches neither crate, nor either benchmark. It adds a new encode function plus a new BtrBlocks scheme; there is no shared code path by which that could affect primitive filtering or an AVX2 lane kernel.

Two things mark it as measurement noise rather than a real regression:
- The *same* bench file reports `random_i8[0.5]` **improved by +34.76%** in the same run.
- CodSpeed itself warns on this run that the WallTime instrument on standard hosted runners "will lead to inconsistent data" — and the AVX2 result is a WallTime measurement. Earlier runs on this branch have swung between a reported +19.09% overall improvement and this -3.26%, each time implicating different unrelated benchmarks.

There is no defect here to fix, so there is no fix to port.

**Rust tests (linux-musl)** — infrastructure, not a test failure. The job was marked failed while its main step (`Run workspace tests for x86_64-unknown-linux-musl`) was still `in_progress`, with every later step `pending`, after only ~3.5 minutes, and its logs return HTTP 404. No test ever reported a result; this is consistent with the runner being reclaimed mid-run. I have re-run the failed job once — attempt 2 is in progress. If it fails again *with actual test output*, I'll treat it as genuine and investigate.

The `deny.toml` advisories check remains the separate, repo-wide issue described in my earlier comment.

---

joseph-isaacs:
Correcting my previous comment on the musl failure — I said it looked like an aborted runner and would investigate if it recurred. It recurred, I investigated, and the mechanism I guessed was wrong. The conclusion about attribution is stronger, though: **this is a pre-existing failure on `develop`, not something this PR introduces.**

Evidence — the `Linux musl` workflow run history on `develop`:

| Run | Commit | Result |
|---|---|---|
| #7486 | `fa01d35` | success |
| #7488 | `3db6d3c` — "Lock file maintenance (#9139)" | **failure** |
| #7490 | `2dddd80` | **failure** |
| #7492 | `f25a36a` — "Lock file maintenance (#9563)" | **failure** |
| #7498 | `4afdbd4` | **failure** |

`4afdbd4` is precisely the base commit this PR is rebased onto, and its `Rust tests (linux-musl)` job fails at the same step (`Run workspace tests for x86_64-unknown-linux-musl`) after the same ~3 minutes, and trips the same `Alert incident.io` step — with none of this PR's code present. The failure reproduces identically on this branch and on its parent, so it is not attributable to these changes.

The regression window is between `fa01d35` (last green) and `3db6d3c` (first red); the latter is a Renovate lockfile refresh, so a dependency update is the most likely cause. That is a `develop` fix, not something this PR can carry — there is no fix to port here yet.

For the record on what I could verify locally: `cargo test -p vortex-runend` (91) and `-p vortex-btrblocks` (49) pass, along with clippy, rustfmt, and `cargo doc` under `-D warnings`. I could not reproduce the musl target locally. I re-ran the failed job once (attempt 2, same result) and will keep watching this PR until it is green.

---

PR #8468, opened 2026-06-17
Remove duckdbfs

### LINKED ISSUE #6706: Performance regressions in duckdbfs
### What happened?

When #5767 was merged, we saw TPC-h (S3) severe regressions in benchmarks. Then we decided to turn it off in https://github.com/vortex-data/vortex/pull/6564. I've been investigating this regression recently, but there are a couple of issues with this approach:

- Our CI can't run benchmarks for a duckdb hash https://github.com/vortex-data/vortex/pull/6658 https://github.com/vortex-data/vortex/pull/6677
- Moreover, Duckdb 1.5 introduced API changes so we depend on that (tracking changes in https://github.com/vortex-data/vortex/pull/6627)
- My guess was that regression may be negligible in using duckdbfs + httpfs's curl backend (default from 1.5) as opposed to httplib (1.4-). However, our benchmarks in CI fail with curl backend due to 1.4 httpfs curl S3 globbing (list) bug https://github.com/duckdb/duckdb-httpfs/pull/265 and we also have a bug in our code for httplib backend https://github.com/vortex-data/vortex/pull/6685 https://github.com/vortex-data/vortex/pull/6709, which further complicates testing.

myrrc:
DuckdbFS implementation for Vortex was introduced in https://github.com/vortex-data/vortex/pull/6198 as opt-out, but changed to opt-in in https://github.com/vortex-data/vortex/pull/6564 due to performance regressions.
There were multiple issues (https://github.com/vortex-data/vortex/pull/6709, https://github.com/vortex-data/vortex/pull/6565 https://github.com/vortex-data/vortex/pull/6685) associated with it which differ from vortex's file system behaviour.

It also requires additional dependencies CI which are a blocker for 
https://github.com/vortex-data/vortex/pull/8486 since MacOS runner doesn't bundle openssl for x86_64 on arm, and builds fail.

As a long term goal, calling duckdb's blocking IO inside our event loop isn't the right abstraction. We want to allow duckdb to use its own IO outside vortex.

Duckdb fs is  also not maintaned actively so we're removing it


0ax1 (approved):
I'm actually in favor of dropping this. The perf doesn't match and as we never exercise or bench the code the level of maintenance is low.

Issue #8348, opened 2026-06-10
Struct field nested directly inside another struct loses nullability on file round-trip


joseph-isaacs:
Found by the Python API Hypothesis fuzzer (`vortex-python/test/test_fuzz_file_roundtrip.py`).

A nullable struct field nested directly inside another struct comes back non-nullable when read from a file, tripping the dtype `debug_assert_eq!` in `vortex-array/src/stream/adapter.rs` on debug builds:

```python
import os
import tempfile

import pyarrow as pa
import vortex as vx

typ = pa.struct([("a", pa.struct([("b", pa.int32())]))])
t = pa.table({"c0": pa.array([{"a": None}, None], type=typ)})
with tempfile.TemporaryDirectory() as d:
    p = os.path.join(d, "x.vortex")
    vx.io.write(t, p)
    vx.open(p).scan().read_all()
```

```text
pyo3_runtime.PanicException: assertion `left == right` failed:
ArrayStreamAdapter expected array with type {c0={a={b=i32}?}?}, actual {c0={a={b=i32}}?}
```

The expected dtype has the inner field `a` as nullable (`{b=i32}?`) but the dtype read back from the file has dropped the `?`. On release builds the `debug_assert` doesn't fire, so this presumably round-trips with silently altered nullability. The fuzzer currently `assume()`s away struct-of-struct schemas; that guard can be removed once this is fixed.

miniex:
~@joseph-isaacs Shouldn't we close this issue?~ mistake

robert3005:
This still reproduces

PR #8472, opened 2026-06-17
fix(layout): don't panic collecting an empty stream

### LINKED ISSUE #8347: Writing an empty table with a top-level struct column panics: "must have visited at least one chunk"

Writing a zero-row table that contains a top-level struct column panics in `vortex-layout/src/layouts/collect.rs` (`latest_sequence_id.vortex_expect("must have visited at least one chunk")`), which surfaces in Python as `pyo3_runtime.PanicException: Runtime dropped task without completing it, likely it panicked`.

```python
import os
import tempfile

import pyarrow as pa
import vortex as vx

t = pa.table({"c0": pa.array([], type=pa.struct([("a", pa.int32())]))})
with tempfile.TemporaryDirectory() as d:
    vx.io.write(t, os.path.join(d, "x.vortex"))  # panics
```

Zero-row tables without struct columns write fine. The fuzzer currently `assume()`s away empty tables with struct columns; that guard can be removed once this is fixed.

miniex:
## Summary

`CollectStrategy` collects its whole input into a single chunk for a child that requires exactly one chunk. When the input stream is empty, e.g. writing a zero-row table with a nullable struct column whose validity substream is empty, no chunk supplied a sequence id and it panicked with `must have visited at least one chunk`. `CollectStrategy` now yields nothing on empty input, and `FlatLayoutStrategy` returns an empty `ChunkedLayout` instead of requiring a single chunk, so no segment is written for an empty array.

The issue mentions the fuzzer's `assume()` guard could be dropped once this is fixed. I left it in place here: reading a nullable struct nested in a struct back is a separate bug (#8348), so the round-trip only works once both are fixed.

Closes: #8347

## Testing

Added a regression test in `vortex-file/tests/test_write_table.rs` that writes a zero-row nullable struct column; it panicked before this change and now writes and reads back as zero rows. The issue's Python repro (`vx.io.write` of an empty `struct` column) no longer panics. `cargo nextest run -p vortex-layout -p vortex-file` passes (177 tests, including the segment-ordering tests). `fmt --all` + `clippy --all-targets --all-features` clean.

---

### DIFF
--- a/vortex-layout/src/layouts/collect.rs
+++ b/vortex-layout/src/layouts/collect.rs
@@ -48,11 +48,6 @@ impl LayoutStrategy for CollectStrategy {
         // Read the whole stream, then write one Chunked stream to the inner thing
         let dtype = stream.dtype().clone();
 
-        // the child needs exactly one chunk, so an empty stream still yields a single empty
-        // chunk, with its sequence id taken from `eof`.
-        let mut eof = eof;
-        let empty_sequence_id = eof.advance();
-
         let _dtype = dtype.clone();
         let collected_stream = try_stream! {
             pin_mut!(stream);
@@ -65,8 +60,11 @@ impl LayoutStrategy for CollectStrategy {
                 chunks.push(chunk);
             }
 
-            let collected = ChunkedArray::try_new(chunks, _dtype)?.into_array();
-            yield (latest_sequence_id.unwrap_or(empty_sequence_id), collected);
+            // an empty input yields no chunk; the child layout handles it.
+            if let Some(sequence_id) = latest_sequence_id {
+                let collected = ChunkedArray::try_new(chunks, _dtype)?.into_array();
+                yield (sequence_id, collected);
+            }
         };
 
         let adapted = Box::pin(SequentialStreamAdapter::new(dtype, collected_stream));
--- a/vortex-layout/src/layouts/collect.rs
+++ b/vortex-layout/src/layouts/collect.rs
@@ -10,7 +10,6 @@ use futures::pin_mut;
 use vortex_array::ArrayContext;
 use vortex_array::IntoArray;
 use vortex_array::arrays::ChunkedArray;
-use vortex_error::VortexExpect;
 use vortex_error::VortexResult;
 use vortex_session::VortexSession;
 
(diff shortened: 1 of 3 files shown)


onursatici (changes requested):
Thanks for the contribution!
This seems correct, but I think we should fix this on a lower level. 
We generally shouldn't use `eof` if we do not intend the chunk to go towards the end of file, so although it makes it possible to write an empty chunk, I think the right behaviour of collect strategy is to not yield anything on empty input.
If you just make that change the empty stream would propagate to the children and the flat layout would then fail. I think in the flat layout if there is no chunk we should return an empty `ChunkedLayout` which allows for no segments to exist at all, unlike flat layout which enforces one segment. If we do this we can save from writing a segment just to store an empty array

So the shape would be:

- `CollectStrategy`: empty input yields no chunks
- `FlatLayoutStrategy`: empty input returns a zero-row, zero-child `ChunkedLayout`
- non-empty flat input keeps the existing exactly-one-chunk behavior

miniex:
Thanks for the steer - done.

`CollectStrategy` now yields nothing on empty input, and `FlatLayoutStrategy` returns an empty `ChunkedLayout` instead of requiring a single chunk, so no segment is written for an empty array. Dropped the `eof` minting as you suggested. Non-empty flat input keeps the exactly-one-chunk path.

Verified the empty file still opens and scans back to zero rows; full `vortex-layout` + `vortex-file` suites green.

vortex-layout/src/layouts/collect.rs
```diff
@@ -61,8 +60,11 @@ impl LayoutStrategy for CollectStrategy {
                 chunks.push(chunk);
             }
 
-            let collected = ChunkedArray::try_new(chunks, _dtype)?.into_array();
-            yield (latest_sequence_id.vortex_expect("must have visited at least one chunk"), collected);
+            // an empty input yields no chunk; the child layout handles it.
+            if let Some(sequence_id) = latest_sequence_id {
```
onursatici: shall we invert this if and return early, something like:

```rust
  let Some(sequence_id) = latest_sequence_id else {
      return;
  };

  let collected = ChunkedArray::try_new(chunks, _dtype)?.into_array();
  yield (sequence_id, collected);
```

onursatici (commented):
thanks, one final comment and then we can merge

vortex-layout/src/layouts/collect.rs
miniex: done, switched to the let-else guard.

onursatici (approved):

PR #8474, opened 2026-06-17
Fix integer overflow in BitBufferMut::truncate and append_n

ksj1230:
## Summary

Several methods on `BitBufferMut` use unchecked arithmetic on offset/length values, which can overflow in release mode and lead to undefined behavior reachable entirely from safe code (`#![forbid(unsafe_code)]`).

Affected methods:
- `truncate`: `self.offset + len` overflows, causing the buffer to be truncated to an incorrect size.
- `append_n`: `self.offset + self.len + n` overflows, causing incorrect capacity calculation.

Fix: add overflow assertions before the unchecked addition.

Reproduction (all use `#![forbid(unsafe_code)]`):

```rust
// 1. truncate
let buffer = ByteBufferMut::zeroed(10);
let mut bb = BitBufferMut::from_buffer(buffer, usize::MAX - 2, 80);
bb.truncate(3);
bb.set(2);

// 2. append_n
let buffer = ByteBufferMut::zeroed(2);
let mut bb = BitBufferMut::from_buffer(buffer, 1000, 0);
bb.append_n(true, usize::MAX);
bb.set(10000);
```

Related: #7979 (the `BufferMut::zeroed_aligned` overflow I reported to maintainers in April 2025; this PR addresses additional overflow sites found during the same audit)

## Testing

Existing `cargo test -p vortex-buffer` passes (789 tests + 6 doc-tests).
Both PoCs now panic with overflow message instead of UB.

Happy to adjust the approach if maintainers prefer a different fix strategy.

### DIFF
--- a/vortex-buffer/src/bit/buf_mut.rs
+++ b/vortex-buffer/src/bit/buf_mut.rs
@@ -375,7 +375,8 @@ impl BitBufferMut {
         if len > self.len {
             return;
         }
-
+        
+        assert!(self.offset <= usize::MAX - len, "Truncate on BitBufferMut overflowed");
         let end_bit = self.offset + len;
         let new_len_bytes = end_bit.div_ceil(8);
         self.buffer.truncate(new_len_bytes);
@@ -442,6 +443,8 @@ impl BitBufferMut {
             return;
         }
 
+        assert!(self.offset.checked_add(self.len).and_then(|v| v.checked_add(n)).is_some(),
+            "Append on BitBufferMut overflowed");
         let end_bit_pos = self.offset + self.len + n;
         let required_bytes = end_bit_pos.div_ceil(8);
 
--- a/vortex-buffer/src/bit/buf_mut.rs
+++ b/vortex-buffer/src/bit/buf_mut.rs
@@ -375,8 +375,11 @@ impl BitBufferMut {
         if len > self.len {
             return;
         }
-        
-        assert!(self.offset <= usize::MAX - len, "Truncate on BitBufferMut overflowed");
+
+        assert!(
+            self.offset <= usize::MAX - len,
+            "Truncate on BitBufferMut overflowed"
+        );
         let end_bit = self.offset + len;
         let new_len_bytes = end_bit.div_ceil(8);
         self.buffer.truncate(new_len_bytes);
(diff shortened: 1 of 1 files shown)


miniex:
sign off the commit to pass DCO: `git commit --amend -s --no-edit && git push -f`

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -186,7 +186,8 @@ impl BitBufferMut {
     /// Invokes `f` with indexes `0..len` collecting the boolean results into a new `BitBufferMut`
     #[inline]
     pub fn collect_bool<F: FnMut(usize) -> bool>(len: usize, f: F) -> Self {
-        let num_words = len.div_ceil(64);
+        let num_words = len.checked_add(63)
+    		.expect("collect_bool: len overflow") / 64;
         let mut buffer: BufferMut<u64> = BufferMut::with_capacity(num_words);
```
joseph-isaacs: how does this overflow?

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -376,7 +377,8 @@ impl BitBufferMut {
             return;
         }
 
-        let end_bit = self.offset + len;
+        let end_bit = self.offset.checked_add(len)
+    		.expect("BitBufferMut::truncate: offset + len overflow");
```
joseph-isaacs: This is gated 3 lines above

vortex-buffer/src/bit/buf_mut.rs
```diff
@@ -442,7 +444,9 @@ impl BitBufferMut {
             return;
         }
 
-        let end_bit_pos = self.offset + self.len + n;
+        let end_bit_pos = self.offset.checked_add(self.len)
```
AdamGS: instead of doing checked arithmatic, how about asserting these? The code is much clearer that way IMO.

AdamGS:
stylistically - I think we use error messagess that read more like "Operation X on Y overflowed"

vortex-buffer/src/bit/buf_mut.rs
ksj1230: Thanks for the quick review.

You're right. standard `div_ceil` uses division first, so it doesn't overflow. It only causes out-of-memory, which is not a soundness issue. Removed from this PR.

vortex-buffer/src/bit/buf_mut.rs
ksj1230: The guard bounds `len` by `self.len`, but `self.offset` is independent and can be set to a large value via `from_buffer` without validation. So `self.offset + len` can still overflow even when `len <= self.len`.

Miri confirms UB with the PoC in the description:
  error: Undefined Behavior: attempting to offset pointer by 
  2305843009213693951 bytes, but got alloc192 which is only 11 bytes

vortex-buffer/src/bit/buf_mut.rs
ksj1230: Switched to assertions and updated the message style. Thanks!

robert3005:
@AdamGS anything missing here?

AdamGS (approved):
No, just lost track of it

AdamGS:
@ksj1230 thank you for you contribution! I apologize for taking so long.

ksj1230:
Thanks for merging! Happy to help.

PR #8476, opened 2026-06-17
Add Kleene boolean benchmarks

gatesn:


AdamGS (approved):
LGTM.

Should we black_box inputs? I never know.

PR #8478, opened 2026-06-17
Inline Kleene boolean kernels

gatesn:
Next in our list of Arrow dependencies to remove


vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +171,632 @@ fn arrow_execute_boolean(
...
+fn word_source_from_bit_buffer(buffer: &BitBuffer, len: usize) -> Option<WordSource<'_>> {
+    if !buffer.offset().is_multiple_of(8) {
+        return None;
+    }
+
+    let n_bytes = len.div_ceil(8);
+    let start = buffer.offset() / 8;
+    let end = start + n_bytes;
+    Some(WordSource::Bytes(&buffer.inner().as_slice()[start..end]))
+}
```
joseph-isaacs: Mask can do this

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +171,632 @@ fn arrow_execute_boolean(
...
+        AllOr::Some(buffer) => word_source_from_bit_buffer(buffer, len),
+    }
+}
+
+fn read_u64_le(bytes: &[u8]) -> u64 {
+    debug_assert!(bytes.len() <= 8);
+    let mut buf = [0; 8];
+    buf[..bytes.len()].copy_from_slice(bytes);
+    u64::from_le_bytes(buf)
+}
```
joseph-isaacs: this exists

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +171,632 @@ fn arrow_execute_boolean(
...
+
+fn all_validity(nullability: Nullability) -> Validity {
+    match nullability {
+        Nullability::NonNullable => Validity::NonNullable,
+        Nullability::Nullable => Validity::AllValid,
+    }
+}
+
+fn mask_to_validity(mask: Mask, nullability: Nullability) -> Validity {
```
robert3005: this surely exists already?

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +171,632 @@ fn arrow_execute_boolean(
...
+    } else {
+        array.cast(dtype)
+    }
+}
+
+fn boolean_nullability(lhs: &ArrayRef, rhs: &ArrayRef) -> Nullability {
+    lhs.dtype().nullability() | rhs.dtype().nullability()
+}
+
+fn all_validity(nullability: Nullability) -> Validity {
```
robert3005: this should also exist

joseph-isaacs:
Could the perf regression be due to alignment from this new boolean buffer allocated here?

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -1,21 +1,98 @@
...
-use crate::executor::ExecutionCtx;
+use crate::dtype::Nullability;
+use crate::kernel::ExecuteParentKernel;
+use crate::scalar::BoolScalar;
 use crate::scalar::Scalar;
+use crate::scalar_fn::fns::binary::Binary;
 use crate::scalar_fn::fns::operators::Operator;
+use crate::validity::Validity;
+
+/// Trait for encoding-specific Kleene boolean kernels that operate in encoded space.
```
AdamGS: Why only Kleene? Should we do both?

joseph-isaacs:
Is there list a?

Do we have a grasp of the performance of this change (ignoring constant), which would likely a mistake on our part previously?

I what would we expect here?

vortex-array/src/scalar_fn/fns/binary/boolean.rs
gatesn: We should have both. We have an open item to rename our current and function to and_kleene, then the engines should pick the variant that matches their semantics

encodings/bytebool/src/compute.rs
```diff
@@ -106,9 +112,59 @@ impl TakeExecute for ByteBool {
...
+            operator,
+            nullability,
+            ctx,
+        )
+        .map(Some)
+    }
+}
+
+fn truthy_bit_buffer(array: ArrayView<'_, ByteBool>) -> BitBuffer {
+    BitBuffer::from_iter(array.truthy_bytes().iter().map(|byte| *byte != 0))
```
robert3005: from_trusted_len_iter

encodings/bytebool/src/compute.rs
```diff
@@ -106,9 +112,59 @@ impl TakeExecute for ByteBool {
...
+    fn boolean(
+        lhs: ArrayView<'_, Self>,
+        rhs: &ArrayRef,
+        operator: Operator,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<ArrayRef>> {
+        let nullability = lhs.dtype().nullability() | rhs.dtype().nullability();
+        let lhs_values = truthy_bit_buffer(lhs);
+
+        if let Some(rhs) = rhs.as_opt::<vortex_array::arrays::Constant>() {
```
robert3005: Would be nice to import this

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +175,601 @@ fn arrow_execute_boolean(
...
+
+fn bool_scalar_value(scalar: &Scalar) -> VortexResult<Option<bool>> {
+    Ok(scalar
+        .as_bool_opt()
+        .ok_or_else(|| vortex_err!("expected boolean scalar"))?
+        .value())
+}
+
+/// Execute a Kleene boolean operation from boolean value bitmaps and validity values.
+pub fn boolean_buffers(
```
robert3005: How about we name it something better? I was a bit puzzled why these aren't methods on bitbuffer but these need validity as well. How about having kleene_or/kleene_and? kleene_boolean?

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +175,601 @@ fn arrow_execute_boolean(
...
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<ArrayRef> {
+    let len = lhs_values.len();
+    debug_assert_eq!(rhs_values.len(), len);
+
+    if lhs_validity.definitely_no_nulls() && rhs_validity.definitely_no_nulls() {
+        let values = match operator {
+            Operator::And => lhs_values & &rhs_values,
+            Operator::Or => lhs_values | &rhs_values,
+            other => return Err(vortex_err!("Not a boolean operator: {other}")),
```
robert3005: vortex_bail!

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +175,601 @@ fn arrow_execute_boolean(
...
+        &lhs_valid,
+        &rhs_values,
+        &rhs_valid,
+        operator,
+        nullability,
+    )
+}
+
+/// Execute a Kleene boolean operation between boolean value bits and a scalar.
+pub fn boolean_buffer_scalar(
```
robert3005: this is no longer used?

robert3005: No, it's used. Should we replace this with constant_array_boolean? Am I missing something?

vortex-array/src/scalar_fn/fns/binary/boolean.rs
```diff
@@ -73,63 +175,601 @@ fn arrow_execute_boolean(
...
-    Ok(Some(ConstantArray::new(scalar, length).into_array()))
+fn word_source_from_mask(mask: &Mask) -> Option<WordSource<'_>> {
+    match mask.bit_buffer() {
+        AllOr::All => Some(WordSource::Fill(u64::MAX)),
+        AllOr::None => Some(WordSource::Fill(0)),
+        AllOr::Some(buffer) => word_source_from_bit_buffer(buffer),
+    }
+}
+
+fn fused_boolean_word_sources(
```
robert3005, line 474: should this be two methods? There's little logic shared between the two branches

robert3005 (changes requested):
I think apart form some weird structure that looks like refactor done half way this is good

robert3005 (approved):

PR #8481, opened 2026-06-17
fix Fuzz constant OOM for CUDA

a10y:
## Summary

The Fuzz check has been failing for CUDA `compress_gpu` job for a very long time now (at least a month).

After some investigation, it appears this is because ASAN will mprotect a very large memory region that conflicts with some address space that the CUDA driver wants to own, resulting in cudaCtx initialization failing with OOMs like we've been seeing.

I've run a sample Fuzz workflow with this change and confirmed we are able to get past context init now.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -200,6 +200,10 @@ jobs:
       fuzz_target: compress_gpu
       runner: gpu
       extra_features: "cuda"
+      # ASan mprotects the shadow gap PROT_NONE by default, which collides with the large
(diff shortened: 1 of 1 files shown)


a10y:
~First run: https://github.com/vortex-data/vortex/actions/runs/27706306865~

~Actual first run: https://github.com/vortex-data/vortex/actions/runs/27706466820/job/81956199153~

Working run: https://github.com/vortex-data/vortex/actions/runs/27710006987/job/81992356488#step:5:186

0ax1 (approved):
Happy when CI is happy 🙂

lwwmanning (approved):

PR #8482, opened 2026-06-17
Move execute-parent kernels into session registry

gatesn:
Removes the ScalarFnVTable::execute_parent function in favor of purely session-registered kernels.

We already had some kernels on the session, and some on the scalar function. This unifies everything on the session.

I may well follow up by removing ScalarFnVTable::execute also in favor of these kernels.

A definite follow up is to make the structure of our plugin crates uniform. Sometimes kernels live in `compute`, `kernel` or `vtable` modules.

Closes https://github.com/vortex-data/vortex/pull/8401


gatesn:
@claude review it

lwwmanning (approved):

PR #8483, opened 2026-06-17
feat[gpu]: arrow device array stream support

0ax1:
Adds Arrow device stream support which is exercised and tested through the cuDF harness.


vortex-ffi/src/lib.rs
```diff
@@ -53,6 +55,16 @@ static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
 // TODO(ngates): also create a CurrentThreadPool to manage background worker threads.
 static RUNTIME: LazyLock<CurrentThreadRuntime> = LazyLock::new(CurrentThreadRuntime::new);
 
+/// Return the shared FFI runtime for layered FFI crates that drive Vortex streams produced through
+/// `vortex-ffi`.
+///
+/// Streams from `vortex-ffi` partitions spawn their scan work onto this runtime's executor, so a
+/// consumer crate (for example `vortex-cuda`'s Arrow device stream export) must drive them on this
+/// same runtime rather than a private one.
+pub fn runtime() -> &'static CurrentThreadRuntime {
```
myrrc: ffi_runtime()?

vortex-cuda/ffi/src/lib.rs
```diff
@@ -100,6 +109,46 @@ pub unsafe extern "C-unwind" fn vx_cuda_array_export_arrow_device(
...
+pub unsafe extern "C-unwind" fn vx_cuda_partition_scan_arrow_device_stream(
+    session: *const vx_session,
+    partition: *mut vx_partition,
+    out_stream: *mut ArrowDeviceArrayStream,
+    error_out: *mut *mut vx_error,
+) -> c_int {
+    try_or(error_out, VX_CUDA_ERR, || {
+        vortex_ensure!(!partition.is_null(), "null vx_partition");
+
+        let array_stream = unsafe { vx_partition_into_array_stream(partition) }?;
```
myrrc, line 140: Maybe we should move vx_partition_into_array_stream to vortex-cuda? It's not used in main FFI crate.

vortex-cuda/src/arrow/mod.rs
```diff
@@ -242,6 +269,408 @@ impl DeviceArrayExt for ArrayRef {
     }
 }
 
+/// POSIX EIO for Arrow stream producer/export failures.
+const ARROW_STREAM_EIO: c_int = 5;
```
myrrc: libc::EINVAL/EIO?

myrrc (approved):
Changes look good to me, but this PR would benefit from adding C-side tests.

vortex-ffi/src/lib.rs
0ax1: Yeah that's a little more explicit.

vortex-cuda/ffi/src/lib.rs
0ax1: Hmm, I thought about that but landed on that the fn isn't really specific to CUDA.

vortex-cuda/src/arrow/mod.rs
0ax1: I'd need to add another crate dep for that in our toml. But I should prob rename them to LIB_C_ .. at least.

vortex-cuda/src/arrow/mod.rs
```diff
@@ -242,6 +269,408 @@ impl DeviceArrayExt for ArrayRef {
     }
 }
 
+/// POSIX EIO for Arrow stream producer/export failures.
+const LIBC_EIO: c_int = 5;
```
robert3005, line 273: These are the apis that I live for

0ax1: 🙃 Can also return 1 for all error cases.

robert3005 (approved):

PR #8484, opened 2026-06-17
More realistic RunEnd take benchmark params

robert3005:
Average run length of 1 is not a real run end array


AdamGS (approved):
never say never

PR #8486, opened 2026-06-18
Use R2 bucket for duckdb libraries

### LINKED ISSUE #8034: Tracking Issue: own infrastructure for duckdb builds
Currently we use duckdb servers to download libduckdb.so and headers.
This has some drawbacks:
1. Servers sometimes are unavailable, and our builds fail.
2. We don't have libduckdb.so with debug symbols, so we always need to rebuild locally which is a big waste of time.
3. For pre-releases, we can't use duckdb servers because there are no pre-release builds, and we want to test changes before release.

So we need an S3 bucket with release/pre-release .so files for architectures we use as well as headers, and CI infrastructure to update it automatically.

- [x] Use ci-builds.vortex.dev instead of Github Releases #8486
- [ ] Build and upload reldebug builds of Duckdb. Use them in CI and local builds.

myrrc:
- Use ci-builds.vortex.dev R2 bucket as source for duckdb release and commit builds.
- Mirror release builds from duckdb github releases page. Build commits from source.
- For commit builds also try to download from R2 (useful for testing pre-release in CI)
- Gate test_geometry for duckdb under release builds only. Running it for commits means we need to bundle "spatial" extension which is hard on macos due to openssl-dev symbols

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,25 @@ env:
   NIGHTLY_TOOLCHAIN: nightly-2026-02-05
 
 jobs:
+  duckdb-mirror:
+    name: "Mirror DuckDB to R2"
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/duckdb-r2.yml
+    secrets: inherit
+
+  duckdb-ready:
+    name: "DuckDB libraries available in R2"
+    needs: duckdb-mirror
+    if: ${{ !cancelled() }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Verify DuckDB mirror
+        if: ${{ needs.duckdb-mirror.result == 'failure' }}
+        run: |
+          echo "DuckDB mirror failed; downstream builds would 404"
+          exit 1
+
   lint-toml:
     runs-on: ubuntu-latest
     timeout-minutes: 10
@@ -115,6 +134,7 @@ jobs:
 
   rust-docs:
     name: "Rust (docs)"
+    needs: duckdb-ready
     timeout-minutes: 30
     runs-on: >-
       ${{ github.repository == 'vortex-data/vortex'
@@ -204,6 +224,7 @@ jobs:
 
   rust-lint:
     name: "Rust (lint)"
+    needs: duckdb-ready
     timeout-minutes: 30
     runs-on: >-
       ${{ github.repository == 'vortex-data/vortex'
@@ -301,6 +322,7 @@ jobs:
 
   rust-test-other:
     name: "Rust tests (${{ matrix.os }})"
+    needs: duckdb-ready
     timeout-minutes: 30
     strategy:
       fail-fast: false
@@ -422,6 +444,7 @@ jobs:
 
   sqllogic-test:
     name: "SQL logic tests"
+    needs: duckdb-ready
     runs-on: >-
       ${{ github.repository == 'vortex-data/vortex'
           && format('runs-on={0}/runner=amd64-medium/image=ubuntu24-full-x64-pre-v2/tag=sql-logic-test', github.run_id)
--- a/.github/workflows/duckdb-r2.yml
+++ b/.github/workflows/duckdb-r2.yml
@@ -1,0 +1,83 @@
+name: DuckDB R2 mirror
+
+# If we have an archive for a commit or a tag in R2, return it.
+# Otherwise, if it's a tag, download release archive from Duckdb Github
+# page and upload it to R2.
+# If it's a commit, build Duckdb from source and upload it to R2.
+on:
+  workflow_call: { }
+
+concurrency:
+  group: duckdb-r2-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+env:
+  PUBLIC_BASE_URL: "https://ci-builds.vortex.dev"
+  R2_BUCKET: "duckdb-builds"
+  R2_ENDPOINT_URL: "https://52bdeab5651e1584747feefd051fd566.r2.cloudflarestorage.com"
+
+jobs:
+  check:
+    name: "Resolve DuckDB version and check R2"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      version: ${{ steps.resolve.outputs.version }}
+      ref_dir: ${{ steps.resolve.outputs.ref_dir }}
+      release: ${{ steps.resolve.outputs.release }}
+      matrix: ${{ steps.resolve.outputs.matrix }}
+      any_missing: ${{ steps.resolve.outputs.any_missing }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+      - name: Resolve version and check R2
+        id: resolve
+        run: sh scripts/duckdb-r2-resolve.sh
+
+  mirror:
+    name: "Mirror DuckDB ${{ matrix.archive }} to R2"
+    needs: check
+    if: >-
+      needs.check.outputs.any_missing == 'true' &&
+      github.repository == 'vortex-data/vortex' &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    environment: duckdb-build
+    timeout-minutes: 120
+    strategy:
+      fail-fast: false
+      matrix: ${{ fromJSON(needs.check.outputs.matrix) }}
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+
+      - name: Install build dependencies (Linux)
+        if: needs.check.outputs.release != 'true' && runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ninja-build libcurl4-openssl-dev zip unzip
+
+      # MacOS already has ninja and p7zip
+
+      - name: Prepare ${{ matrix.archive }}
+        env:
+          ARCHIVE: ${{ matrix.archive }}
+          REF_DIR: ${{ needs.check.outputs.ref_dir }}
+          RELEASE: ${{ needs.check.outputs.release }}
+          PLATFORM_OS: ${{ matrix.os }}
+        run: sh scripts/duckdb-r2-prepare.sh
+
+      - name: Upload to R2
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.DUCKDB_R2_ACCESS_KEY_ID }}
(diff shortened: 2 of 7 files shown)


myrrc:
Macos deployment error:

```
error:  cannot create /Users/runner/work/vortex/vortex/duckdb-08e34c447bae34eaee3723cac61f2878b6bdf787/data/csv/issue2628_������.csv
        Illegal byte sequence
```

myrrc:
I've verified manually this works for commits, they are uploaded and used from R2 as well.

.github/workflows/ci.yml
```diff
@@ -25,6 +25,25 @@ env:
...
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/duckdb-r2.yml
+    secrets: inherit
+
+  duckdb-ready:
+    name: "DuckDB libraries available in R2"
+    needs: duckdb-mirror
+    if: ${{ !cancelled() }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
```
0ax1, line 39: What's the rationale for the timeout duration here?

.github/workflows/ci.yml
```diff
@@ -25,6 +25,25 @@ env:
   NIGHTLY_TOOLCHAIN: nightly-2026-02-05
 
 jobs:
+  duckdb-mirror:
+    name: "Mirror DuckDB to R2"
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/duckdb-r2.yml
+    secrets: inherit
+
+  duckdb-ready:
```
0ax1, line 34: Do we really need this, or asked diff can we encode the dependency?

.github/workflows/duckdb-r2.yml
```diff
@@ -0,0 +1,197 @@
+name: DuckDB R2 mirror
+
+# Mirror DuckDB libraries referenced by vortex-duckdb/build.rs to R2 when they
```
0ax1: What does mirror mean exactly? Should we extend the text here a bit on how the whole setup works with R2 and caching?

.github/workflows/duckdb-r2.yml
```diff
@@ -0,0 +1,197 @@
...
+  R2_BUCKET: "duckdb-builds"
+  R2_ENDPOINT_URL: "https://52bdeab5651e1584747feefd051fd566.r2.cloudflarestorage.com"
+
+jobs:
+  check:
+    name: "Resolve DuckDB version and check R2"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      version: ${{ steps.resolve.outputs.version }}
```
0ax1, line 28: Do we support individual commits? (by default DDB sets sth like version 0.0.0 or so right ?)

.github/workflows/duckdb-r2.yml
```diff
@@ -0,0 +1,197 @@
...
+      ref_dir: ${{ steps.resolve.outputs.ref_dir }}
+      release: ${{ steps.resolve.outputs.release }}
+      matrix: ${{ steps.resolve.outputs.matrix }}
+      any_missing: ${{ steps.resolve.outputs.any_missing }}
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+      - name: Resolve version and check R2
+        id: resolve
+        run: |
+          set -Eeuo pipefail
```
0ax1: Bit complex and long to inline a shell script into the GH action, wdyt?

.github/workflows/duckdb-r2.yml
```diff
@@ -0,0 +1,197 @@
...
+              --arg os "$os" \
+              --arg arch "$arch" \
+              '{archive: $archive, runner: $runner, os: $os, arch: $arch}')")
+          done
+
+          if [ "${#entries[@]}" -eq 0 ]; then
+            matrix='{"include":[]}'
+            any_missing=false
+          else
+            include=$(printf '%s\n' "${entries[@]}" | jq -sc '.')
```
0ax1: If the runners have python preinstalled, we could maybe consider that.

.github/workflows/duckdb-r2.yml
```diff
@@ -0,0 +1,197 @@
...
+              THREADSAN=0 \
+              BUILD_SHELL=false \
+              BUILD_UNITTESTS=false \
+              ENABLE_UNITTEST_CPP_TESTS=false \
+              BUILD_EXTENSIONS="parquet;tpch;tpcds" \
+              $extra
+
+            lib_dir="${src_dir}/build/release/src"
+            stage="stage"
+            rm -rf "$stage"
```
0ax1: How come we need to clear the dir here? Isn't this empty for each new runner?

0ax1 (commented):
Great to see this coming into shape! Couple of questions inline. In general, will this be able to handle diff build configs: debug, release, asan etc.? The build config should prob be part of the key with which we store builds.

.github/workflows/duckdb-r2.yml
myrrc: Yes, we support commits.

.github/workflows/duckdb-r2.yml
myrrc: +1

myrrc:
> In general, will this be able to handle diff build configs: debug, release, asan etc.?

Yes, but for the first version it won't to keep the diff small

.github/workflows/ci.yml
myrrc: Chosen randomly

.github/workflows/duckdb-r2.yml
myrrc: +1

vortex-duckdb/build.rs
```diff
@@ -358,12 +374,12 @@ fn build_duckdb(version: &DuckDBVersion, duckdb_repo_dir: &Path) {
             ("1", "0")
         };
 
-    // If we're building from a commit  we need to build benchmark
+    // If we're building from a commit we need to build benchmark
     // extensions statically, otherwise DuckDB tries to load them from an http
     // endpoint with version 0.0.1 (all non-tagged builds) which doesn't exist.
     let static_extensions = match version {
-        DuckDBVersion::Release(_) => "parquet;jemalloc",
```
0ax1, line 365: Why remove jemalloc?

myrrc: It's not an extension anymore, the build prints this as a warning in the log.

.github/workflows/ci.yml
```diff
@@ -25,6 +25,25 @@ env:
...
+    name: "DuckDB libraries available in R2"
+    needs: duckdb-mirror
+    if: ${{ !cancelled() }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Verify DuckDB mirror
+        if: ${{ needs.duckdb-mirror.result == 'failure' }}
```
0ax1, line 42: There's more result types than success and failure. We could consider `if: ${{ needs.duckdb-mirror.result != 'success' }}` to signal a failure here.

0ax1: Apparently there's also: `needs.check.outputs.any_missing == 'true'`.

.github/workflows/duckdb-r2.yml
myrrc: +1

.github/workflows/ci.yml
myrrc: Unlikely we can. We need a separate workflow as there is no needs: between files, and the users of this workflow need to check for its status

.github/workflows/ci.yml
myrrc: No, if the job was skipped, it's a success as well, so we need to check only failures.

0ax1: There's also "canceled" which is a failure.

myrrc: I don't think we should track it then? The job itself will fail, and it's blocking for a merge (and other jobs will fail as well as they won't have the binaries).

scripts/duckdb-r2-prepare.sh
```diff
@@ -0,0 +1,61 @@
...
+        "https://github.com/duckdb/duckdb/releases/download/${REF_DIR}/${ARCHIVE}"
+else
+    echo "Building DuckDB commit ${REF_DIR} from source"
+
+    curl -fSL --retry 3 -o duckdb-src.zip \
+        "https://github.com/duckdb/duckdb/archive/${REF_DIR}.zip"
+
+    # macos zip extract error: cannot create
+    # <...>/issue2628_������.csv Illegal byte sequence
+    if [ "$PLATFORM_OS" = "osx" ]; then
```
0ax1, line 26: Bit annoying we need to use diff tools here. Did you consider just using zstd?

scripts/duckdb-r2-prepare.sh
```diff
@@ -0,0 +1,61 @@
...
+        BUILD_UNITTESTS=false \
+        ENABLE_UNITTEST_CPP_TESTS=false \
+        BUILD_EXTENSIONS="parquet;tpch;tpcds" \
+        $extra
+
+    lib_dir="${src_dir}/build/release/src"
+    stage="stage"
+    mkdir -p "$stage"
+
+    cp -a "${lib_dir}/libduckdb.so" "$stage/" 2>/dev/null || true
```
0ax1, line 52: How about putting this in a loop?

scripts/duckdb-r2-resolve.sh
```diff
@@ -0,0 +1,76 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+#
+# Resolve DuckDB version in vortex-duckdb/build.rs and check which archives
+# are present in R2. Output "version", "ref_dir", "release", "matrix", and
+# "any_missing".
+#
+# Required env vars: PUBLIC_BASE_URL, GITHUB_OUTPUT
+
+set -eu
```
0ax1, line 12: How come this is not `set -euo pipefail`?

scripts/duckdb-r2-resolve.sh
```diff
@@ -0,0 +1,76 @@
...
+    esac
+    jq -nc \
+        --arg archive "$archive" \
+        --arg runner "$runner" \
+        --arg os "$os" \
+        --arg arch "$arch" \
+        '{archive: $archive, runner: $runner, os: $os, arch: $arch}' >> "$entries"
+done
+
+if [ -s "$entries" ]; then
```
0ax1, line 59: I still think the shell scripts here are on the edge of being too complex for bash, and python would give us better long term maintainability.

.github/workflows/ci.yml
0ax1: Right, the canceled case is handled here: `if: ${{ !cancelled() }}`.

0ax1 (approved):
I don't love the long shell scripts and we should prob consider switching to python here eventually but let's give this a spin in prod.

PR #8487, opened 2026-06-18
Minimize alignment noise in bitbuffer benchmarks

AdamGS:
## Summary

I'm trying to hunt down some of the noise in our benchmarks. This PR makes the bitbuffer benchmarks use a fixed alignment of 128 bytes.

### DIFF
--- a/vortex-buffer/benches/vortex_bitbuffer.rs
+++ b/vortex-buffer/benches/vortex_bitbuffer.rs
@@ -6,8 +6,11 @@ use std::iter::Iterator;
 use arrow_buffer::BooleanBuffer;
 use arrow_buffer::BooleanBufferBuilder;
 use divan::Bencher;
+use vortex_buffer::Alignment;
 use vortex_buffer::BitBuffer;
 use vortex_buffer::BitBufferMut;
+use vortex_buffer::BufferMut;
+use vortex_buffer::collect_bool_words;
 
 fn main() {
     // Pre-warm CPUID feature detection so the one-time probe cost is never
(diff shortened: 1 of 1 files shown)


0ax1 (approved):

vortex-buffer/benches/vortex_bitbuffer.rs
```diff
@@ -32,12 +35,45 @@ impl FromIterator<bool> for Arrow<BooleanBuffer> {
...
+    let num_words = len.div_ceil(64);
+    let mut buffer = BufferMut::<u64>::with_capacity_aligned(num_words, BENCH_ALIGNMENT);
+    // SAFETY: `collect_bool_words` writes every word before any read.
+    unsafe { buffer.set_len(num_words) };
+    collect_bool_words(buffer.as_mut_slice(), len, f);
+
+    let mut bytes = buffer.into_byte_buffer();
+    bytes.truncate(len.div_ceil(8));
+    BitBufferMut::from_buffer(bytes, 0, len)
+}
```
robert3005, line 68: This is a very funky function

robert3005: I see, this is BitBufferMut::collect_bool with custom alignment

AdamGS: basically, do you think its worth making it a public function on `BitBufferMut`?

robert3005: we can do that later

joseph-isaacs:
Why do you try the default alignment of all buffers

PR #8489, opened 2026-06-18
Build CUDA benchmarks once, but run in parallel

AdamGS:
## Summary

Reducing the amount of GPU machines we need to get from AWS, and save some money.
An easy follow up here is to build on a machine without a GPU, which should also save some time on startup.


0ax1 (approved):
Looks good! Might be worthwhile to comment in the yml why we do this.

PR #8490, opened 2026-06-18
perf: align buffer to `Alignment::DEFAULT_ALIGNMENT = Alignment::new(256)`

joseph-isaacs:
Default alignment of new buffer to 256

### DIFF
--- a/vortex-buffer/src/alignment.rs
+++ b/vortex-buffer/src/alignment.rs
@@ -20,6 +20,12 @@ impl Alignment {
     pub const HOST_COPY: Self = Alignment::new(256);
 
     /// Default alignment for all buffers.
+    ///
+    /// Chosen to be larger than any SIMD register (e.g. AVX-512's 64-byte
+    /// registers) so that buffers can be processed with vectorized loads/stores
+    /// without alignment fixups, and to match the alignment guarantees of the
+    /// CUDA allocator (256 bytes) so host buffers can be copied to/from device
+    /// memory without re-alignment.
     pub const DEFAULT_ALIGNMENT: Self = Alignment::new(256);
 
     /// Create a new alignment.
--- a/vortex-buffer/src/alignment.rs
+++ b/vortex-buffer/src/alignment.rs
@@ -19,6 +19,9 @@ impl Alignment {
     /// Default alignment for device-to-host buffer copies.
     pub const HOST_COPY: Self = Alignment::new(256);
 
+    /// Default alignment for all buffers.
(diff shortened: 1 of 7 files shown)


vortex-buffer/src/buffer.rs
```diff
@@ -47,7 +47,7 @@ impl<T> Default for Buffer<T> {
         Self {
             bytes: Bytes::from_static(EMPTY_BACKING),
             length: 0,
-            alignment: Alignment::of::<T>(),
+            alignment: Alignment::DEFAULT_ALIGNMENT,
```
AdamGS: I think this is too aggressive

robert3005:
```suggestion
            alignment: Alignment::of::<T>(),
```

vortex-buffer/src/buffer_mut.rs
```diff
@@ -38,7 +39,33 @@ impl<T> BufferMut<T> {
...
+    /// Create a new `BufferMut` with the requested alignment and capacity.
+    ///
+    /// The buffer reports `alignment`, but the underlying allocation is over-aligned to the larger
+    /// of `alignment` and `preferred_alignment`.
+    pub fn with_capacity_preferred_aligned(
+        capacity: usize,
+        alignment: Alignment,
+        preferred_alignment: Option<Alignment>,
+    ) -> Self {
+        let actual = max(
```
0ax1, line 64: Is it safe to assume that alignment is always 2^n? Otherwise, you can have an alignment that is larger than 256 but not a multiple of it, e.g. 300.

robert3005: Alignment struct requires the value to be a power of 2

0ax1:
@claude in depth review

robert3005:
I'll change the empty function. Since it never allocated anything over aligning there is a mistake

vortex-buffer/src/alignment.rs
```diff
@@ -19,6 +19,9 @@ impl Alignment {
     /// Default alignment for device-to-host buffer copies.
     pub const HOST_COPY: Self = Alignment::new(256);
 
+    /// Default alignment for all buffers.
+    pub const DEFAULT_ALIGNMENT: Self = Alignment::new(256);
```
0ax1, line 29: We should comment here why we picked this alignment, as this is larger than any SIMD registers.

0ax1 (approved):
Given that there are no perf regresssions, I think we're good to land this once the nits are resolved.

PR #8491, opened 2026-06-18
codegen only used functions in duckdb C API

myrrc:
Make duckdb-vortex compilation slightly faster

### DIFF
--- a/vortex-duckdb/build.rs
+++ b/vortex-duckdb/build.rs
@@ -44,6 +44,145 @@ const SOURCE_FILES: [&str; 17] = [
     "cpp/vector_buffer.cpp",
 ];
 
+// Duckdb C API function we use.
+// This lowers codegen'd src/cpp.rs by four times.
+const DUCKDB_C_API_FUNCTIONS: [&str; 134] = [
+    "duckdb_array_type_array_size",
+    "duckdb_array_type_child_type",
+    "duckdb_array_vector_get_child",
+    "duckdb_client_context_try_get_current_setting",
+    "duckdb_close",
+    "duckdb_column_count",
+    "duckdb_column_logical_type",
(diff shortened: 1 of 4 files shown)


vortex-duckdb/include/vortex.h
```diff
@@ -1,45 +1,24 @@
...
 #include "duckdb.h"
 
 
+#pragma once
+
 #ifdef __cplusplus
 extern "C" {
 #endif // __cplusplus
 
-/**
```
0ax1, line 21: How come these are dropped and why?

vortex-duckdb/build.rs
```diff
@@ -45,6 +44,145 @@ const SOURCE_FILES: [&str; 18] = [
     "cpp/vector_buffer.cpp",
 ];
 
+// Duckdb C API function we use.
+// This lowers codegen'd src/cpp.rs by four times.
+const DUCKDB_C_API_FUNCTIONS: [&str; 134] = [
```
0ax1, line 49: The binary gets smaller but also all API is now hidden behind having to change the build.rs when not exposed explicitly.

0ax1 (approved):
Seems a bit tedious to maintain a custom list, but alright let's try this in practice.

vortex-duckdb/include/vortex.h
myrrc: We had had a double check: one in cbingden option, other in header/trailer. I've removed the hand-written one.

vortex-duckdb/build.rs
myrrc: If you want to extend Rust API, you still likely need to look into duckdb.h/.hpp, and you need to write a Rust wrapper over Rust unsafe {} functions, so we're not losing anything here.

PR #8494, opened 2026-06-18
Split core `vortex-duckdb` from the extension to reduce build overhead

AdamGS:
## Summary

Turns out, if you mark a crate as "staticlib", it will always build the static library, even if you try and pull it as a rust dependency (Like we do for `vortex-duckdb` in `vortex-sqllogictest` and benchmarks).
This PR splits the `vortex-duckdb` crate into two:
1. The core stays in `vortex-duckdb`, which is now a plain old "rlib".
2. A new `vortex-extension` crate that is a "staticlib", which is in the `duckdb-vortex` repo (https://github.com/vortex-data/duckdb-vortex/pull/91).
3. 
This saves about **1GB* per build of `vortex-duckdb`.


0ax1 (approved):
Great stuff! Annoying request, please do a full extension build with this setup to make sure it'll work e2e.

AdamGS:
@myrrc figured out we don't even need `vortex-duckdb-extension`, I'm going to move it to `vortex-data/duckdb-vortex` to replace the submodule!

AdamGS:
Linux build just [passed](https://github.com/vortex-data/duckdb-vortex/actions/runs/27768837471/job/82162991043?pr=91) on duckdb-vortex! Once I'm happy benchmarks also pass here, I'll merge this and change the other repo to point at the latest commit.

PR #8495, opened 2026-06-18
Add `list_length` scalar function

mhk197:
Adds a `list_length` scalar function returning the number of elements in each list of a `List`-like array.

- Computed purely from the list's offsets/sizes — it never reads elements. Different paths for `List`, `ListView`, and `FixedSizeList` arrays. 
- Returns a `U64` array; a null list yields a null length.
- Registered as a built-in (`vortex.list.length`) alongside `list_contains`, and exposed via the `list_length(expr)` expression constructor.

### DIFF
--- a/vortex-array/benches/list_length.rs
+++ b/vortex-array/benches/list_length.rs
@@ -1,0 +1,146 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Benchmarks for the `list_length` scalar function over `List` and `ListView` inputs.
+//!
+//! `list_length` reads only the offsets/sizes (never the elements), so its cost scales with the
+//! number of lists.
+
+#![expect(clippy::unwrap_used)]
+#![expect(clippy::cast_possible_truncation)]
+
+use std::sync::LazyLock;
+
+use divan::Bencher;
+use rand::RngExt;
+use rand::SeedableRng;
+use rand::distr::Uniform;
+use rand::rngs::StdRng;
+use vortex_array::ArrayRef;
+use vortex_array::Canonical;
+use vortex_array::IntoArray;
+use vortex_array::VortexSessionExecute;
+use vortex_array::arrays::BoolArray;
+use vortex_array::arrays::ListArray;
+use vortex_array::arrays::ListViewArray;
+use vortex_array::arrays::PrimitiveArray;
+use vortex_array::expr::list_length;
+use vortex_array::expr::root;
+use vortex_array::validity::Validity;
+use vortex_buffer::Buffer;
+use vortex_session::VortexSession;
+
+fn main() {
+    divan::main();
+}
+
+static SESSION: LazyLock<VortexSession> = LazyLock::new(vortex_array::array_session);
+
+const BASE_LIST_SIZE: usize = 8;
+
+const SMALL: usize = 100;
+const MEDIUM: usize = 10_000;
+const LARGE: usize = 1_000_000;
+
+/// A uniformly-random partition of `num_lists * LIST_SIZE` elements into `num_lists` lists,
+/// plus a validity mask with ~1/8 of lists null at random positions.
+fn random_lists(num_lists: usize) -> (Vec<i32>, Validity) {
+    let mut rng = StdRng::seed_from_u64(num_lists as u64);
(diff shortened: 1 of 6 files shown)


vortex-array/benches/list_length.rs
```diff
@@ -0,0 +1,146 @@
...
+use vortex_array::arrays::ListViewArray;
+use vortex_array::arrays::PrimitiveArray;
+use vortex_array::expr::list_length;
+use vortex_array::expr::root;
+use vortex_array::validity::Validity;
+use vortex_buffer::Buffer;
+use vortex_session::VortexSession;
+
+fn main() {
+    divan::main();
```
gatesn, line 34: I think we use criterion these days right?

vortex-array/src/scalar_fn/fns/list_length.rs
```diff
@@ -0,0 +1,307 @@
...
+        match child_idx {
+            0 => ChildName::from("input"),
+            _ => unreachable!("Invalid child index {child_idx} for list_length()"),
+        }
+    }
+
+    fn return_dtype(&self, _options: &Self::Options, arg_dtypes: &[DType]) -> VortexResult<DType> {
+        match &arg_dtypes[0] {
+            DType::List(_, nullable) => Ok(DType::Primitive(PType::U64, *nullable)),
+            other => vortex_bail!("list_length() requires List, got {other}"),
```
gatesn: May as well support FixedList as well, then implement `reduce` to collapse it into the constant

vortex-array/src/scalar_fn/fns/list_length.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+pub(crate) fn list_length(
+    array: &ArrayRef,
+    nullability: Nullability,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<ArrayRef> {
+    // TODO(mk): short-circuit when array is all null
+
+    let (lengths, validity) = if let Some(list) = array.as_opt::<ListView>() {
```
gatesn: I think the "proper" way to do this would be to define an `AnyList` matcher, then `execute_until::<AnyList>`?

vortex-array/benches/list_length.rs
mhk197: I think its [divan in `vortex-array` crate](https://github.com/vortex-data/vortex/blob/02e085f2582bab02b15e497a3def6d4b1f9bb580/vortex-array/Cargo.toml#L91)? Can switch to criterion though if wanted

robert3005: We only use criterion for cuda benchmarks, divan for the rest. Divan seems largely unmaintained, might need to build our own :/

AdamGS: or just move back to criterion? Development picked up there

vortex-array/src/scalar_fn/fns/list_length.rs
mhk197: Implemented `reduce` for nonnullable fsl, delegated nullable to `execute` since we can't easily get validity (talked offline)

vortex-array/src/scalar_fn/fns/list_length.rs
```diff
@@ -0,0 +1,401 @@
...
+    offsets
+        .slice(1..offsets.len())?
+        .binary(offsets.slice(0..n)?, Operator::Sub)
+}
+
+/// Matches an `Array<List>` or `Array<ListView>`.
+struct AnyList;
+
+impl Matcher for AnyList {
+    type Match<'a> = ();
```
gatesn, line 179: You should define a `enum AnyListView { List(...), FixedList(...) }` , then you can just match on it above in the execute_until

mhk197: do we want to execute `FixedList`? We can just get the size from the dtype

gatesn: You're not executing the FixedList itself, you're basically saying, run execution one step at a time _until_ it matches one of these encodings.

So there may be some scalar function that happens to return a FixedList, then you will terminate and have access to it

mhk197: Discussed offline -- `execute_until` may need refactor to return `Match` instead of discarding it

gatesn (approved):

PR #8496, opened 2026-06-18
Generalize SIMD `take` operations to `Copy`

### LINKED ISSUE #5677: Generalize SIMD `take` operations to `Copy`
What we ultimately want is to match on `size_of::<T>`, and transmute into unsigned types so we don't bloat code-gen. Then we will only SIMD dispatch on types <= word size, fall back to scalar for anything else.

_Originally posted by @gatesn in https://github.com/vortex-data/vortex/pull/5666#discussion_r2605958982_

connortsui20:
## Summary

Closes: https://github.com/vortex-data/vortex/issues/5677

Changes the `NativePType` bound to `Copy`, which allows us to deduplicate a bunch of code. Imo this is also easier to read.

Also a small `is_constant` change.

## Testing

Existing tests should suffice

### DIFF
--- a/vortex-array/src/arrays/primitive/compute/take/avx2.rs
+++ b/vortex-array/src/arrays/primitive/compute/take/avx2.rs
@@ -46,7 +46,6 @@ use crate::arrays::primitive::compute::take::TakeImpl;
 use crate::arrays::primitive::compute::take::take_primitive_scalar;
 use crate::arrays::primitive::vtable::Primitive;
 use crate::dtype::NativePType;
-use crate::dtype::PType;
 use crate::dtype::UnsignedPType;
 use crate::match_each_native_ptype;
 use crate::match_each_unsigned_integer_ptype;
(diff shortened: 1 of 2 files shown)


robert3005 (changes requested):
You need to audit the callers to make sure that T passed into this function is `NativeValue(T)` and not `T`

vortex-array/src/aggregate_fn/fns/is_constant/primitive.rs
```diff
@@ -14,9 +13,10 @@ cfg_if::cfg_if! {
     }
 }
 
-/// Assumes any floating point has been cast into its bit representation for which != and !is_eq are the same
+/// Assumes any floating point has been cast into its bit representation for which `==` and `is_eq`
+/// agree (callers transmute floats to unsigned ints in [`check_primitive_constant`]).
 /// Assumes there's at least 1 value in the slice, which is an invariant of the entry level function.
-pub fn compute_is_constant<T: NativePType, const WIDTH: usize>(values: &[T]) -> bool {
+pub fn compute_is_constant<T: Copy + PartialEq, const WIDTH: usize>(values: &[T]) -> bool {
```
robert3005: mostly here

a10y:
The AVX2 changes seem fine to me

vortex-array/src/aggregate_fn/fns/is_constant/primitive.rs
connortsui20: I don't think I understand, why would this affect callers since all of our `NativePType` satisfy both of these?

connortsui20: also I just changed it to `Eq` instead of `PartialEq`

connortsui20:
not sure which regressions/speedups are real, there are some take speedups maybe because of less hit to the instruction cache?

vortex-array/src/aggregate_fn/fns/is_constant/primitive.rs
robert3005: you changed `is_eq` to `!=`. NativeValue is the struct that implements `!=` using `is_eq`. `is_eq` is not the standard rust behaviour for floats

connortsui20: I see, I'll just revert this part and keep the simd stuff

robert3005 (approved):

PR #8498, opened 2026-06-18
Fix pushed cast nullability conversion

gatesn:
Use DataFusion's physical expression nullability when converting pushed casts into Vortex expressions. This keeps target-field metadata from turning nullable physical casts into non-null Vortex runtime assertions.

### DIFF
--- a/vortex-datafusion/src/convert/exprs.rs
+++ b/vortex-datafusion/src/convert/exprs.rs
@@ -53,10 +53,11 @@ pub struct ProcessedProjection {
 pub(crate) fn make_vortex_predicate(
     expr_convertor: &dyn ExpressionConvertor,
     predicate: &[Arc<dyn PhysicalExpr>],
+    input_schema: &Schema,
 ) -> DFResult<Option<Expression>> {
     let exprs = predicate
         .iter()
-        .map(|e| expr_convertor.convert(e.as_ref()))
+        .map(|e| expr_convertor.convert_with_schema(e.as_ref(), input_schema))
         .collect::<DFResult<Vec<_>>>()?;
 
     Ok(and_collect(exprs))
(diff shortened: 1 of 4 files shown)


vortex-datafusion/src/convert/exprs.rs
```diff
@@ -53,10 +53,11 @@ pub struct ProcessedProjection {
 pub(crate) fn make_vortex_predicate(
     expr_convertor: &dyn ExpressionConvertor,
     predicate: &[Arc<dyn PhysicalExpr>],
+    input_schema: &Schema,
```
AdamGS, line 56: should we replace the current convert?

vortex-datafusion/src/convert/exprs.rs
```diff
@@ -224,23 +248,26 @@ impl ExpressionConvertor for DefaultExpressionConvertor {
         }
 
         if let Some(cast_expr) = df.downcast_ref::<df_expr::CastExpr>() {
-            let cast_dtype = DType::from_arrow(cast_expr.target_field().as_ref());
-            let child = self.convert(cast_expr.expr().as_ref())?;
+            let mut cast_dtype = DType::from_arrow(cast_expr.target_field().as_ref());
+            if let Some(input_schema) = input_schema {
+                cast_dtype = cast_dtype.with_nullability(cast_expr.nullable(input_schema)?.into());
```
AdamGS, line 253: Inside `CastExpr::nullable` there's this comment:
> A cast is nullable if **either** the child is nullable or the
> target field allows nulls.  This conservative rule prevents
> optimizers from assuming a non-null result when a null input could
> still propagate.  `return_field()` continues to expose the exact
> target metadata separately.

So I think this is correct? Did you run into it with some real world example or just a random AI thing?

PR #8500, opened 2026-06-19
Fix is_zero() for FixedSizeList and Struct scalar types

### LINKED ISSUE #7891: `Scalar::is_zero` is probably not correct for `FixedSizeList` and `Struct`
```rust
    /// Returns `true` if the [`Scalar`] has a non-null zero value.
    ///
    /// Returns `None` if the scalar is null, otherwise returns `Some(true)` if the value is zero
    /// and `Some(false)` otherwise.
    pub fn is_zero(&self) -> Option<bool> {
        let value = self.value()?;

        let is_zero = match self.dtype() {
            ...
            ...
        };

        Some(is_zero)
    }

```

I'm pretty sure this is wrong. We need to check these are true but also that the elements themselves are zero.

connortsui20:
## Summary

Closes: https://github.com/vortex-data/vortex/issues/7891

Fixes incorrect `is_zero()` implementations for `FixedSizeList` and `Struct` scalar types that were only checking the number of elements/fields rather than recursively validating that all contained values are themselves zero.

### DIFF
--- a/vortex-array/src/scalar/scalar_impl.rs
+++ b/vortex-array/src/scalar/scalar_impl.rs
@@ -188,10 +188,19 @@ impl Scalar {
             DType::Utf8(_) => value.as_utf8().is_empty(),
             DType::Binary(_) => value.as_binary().is_empty(),
             DType::List(..) => value.as_list().is_empty(),
-            // TODO(connor): This seems wrong...
-            DType::FixedSizeList(_, list_size, _) => value.as_list().len() == *list_size as usize,
-            // TODO(connor): This also seems wrong...
-            DType::Struct(struct_fields, _) => value.as_list().len() == struct_fields.nfields(),
+            // A fixed-size list is zero only if it has the expected number of elements and every
+            // element is itself a non-null zero value.
+            DType::FixedSizeList(_, list_size, _) => {
+                let list = self.as_list();
+                list.len() == *list_size as usize
+                    && (0..list.len())
+                        .all(|i| list.element(i).is_some_and(|e| e.is_zero() == Some(true)))
+            }
+            // A struct is zero only if every one of its fields is itself a non-null zero value.
+            DType::Struct(..) => self
+                .as_struct()
+                .fields_iter()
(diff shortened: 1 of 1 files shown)


vortex-array/src/scalar/scalar_impl.rs
```diff
@@ -188,10 +188,19 @@ impl Scalar {
...
+                let list = self.as_list();
+                list.len() == *list_size as usize
+                    && (0..list.len())
+                        .all(|i| list.element(i).is_some_and(|e| e.is_zero() == Some(true)))
+            }
+            // A struct is zero only if every one of its fields is itself a non-null zero value.
+            DType::Struct(..) => self
+                .as_struct()
+                .fields_iter()
+                .is_some_and(|mut fields| fields.all(|f| f.is_zero() == Some(true))),
```
AdamGS, line 203: The definitions here seem really weird to me, I think its worth documenting the "why" here.
How do other systems handle this?

connortsui20: I'm not sure why they seem weird? I feel like this is the only possible correct definition for these scalars.

AdamGS: In my mind a "zero" struct is an empty struct (as in - 0 fields) of length 0

connortsui20: But this is a scalar, which means it's already length 1. And I would think of this as `is_default` almost where we do not have control over the struct fields, we just call `is_default` on all fields (via auto-derived `Default` impl of the struct).

connortsui20: More importantly, we have to implement the functions in [`vortex-array/src/scalar/scalar_value.rs`](https://github.com/vortex-data/vortex/blob/develop/vortex-array/src/scalar/scalar_value.rs) somehow based on the dtype, and the `is_` functions have to mirror that.

robert3005 (approved):

PR #8503, opened 2026-06-19
VortexSession API cleanup pass

AdamGS:
## Summary

This PR makes the final pass to make the `VortexSession` immutable, removing the `to_builder` function and trying to clear up all call sites.

I've also tried to find all docs referencing it, but the change touches so many places I might've missed something.

One big change here is changing `array_session` (which I think I introduced?) to `default_session_builder`


robert3005 (approved):

robert3005:
We want shorthand methods for the builders as well so we don't have to use `get_mut::<T>()`

robert3005 (changes requested):
I missed the part where this makes it difficult to add things at runtime and will force to teach consumers about unloading. Unloading can be difficult since you need to track all the state that you have borrowed from the session

PR #8504, opened 2026-06-19
Reduce vortex-fastlanes debug symbols by halving `unchecked_unpack_cmp` monomorphization

AdamGS:
## Summary

`vortex-fastlanes` is currently a pretty big crate, due to its underlying structure/generics.

When running `cargo clean; cargo build -p vortex-fastlanes`

Currently on `develop:
- vortex-fastlanes dev build with debug: 229MB rlib 105s
- vortex-fastlanes dev build no debug: 119MB rlib 57s

We currently disable debug symbols for it to improve our development experience, but underlying crates can't enjoy that without explicitly adding the same or similar configuration.

With this branch, the build takes 50s and generates a 133MB rlib, with no performance impact for release builds.
For a release build, this change saves about 2.5MB (out of 7.4MB).

### DIFF
--- a/encodings/fastlanes/src/bitpacking/compute/compare_fused.rs
+++ b/encodings/fastlanes/src/bitpacking/compute/compare_fused.rs
@@ -41,6 +41,7 @@ use vortex_array::dtype::NativePType;
 use vortex_array::dtype::Nullability;
 use vortex_array::dtype::PhysicalPType;
 use vortex_array::match_each_unsigned_integer_ptype;
+use vortex_array::scalar_fn::fns::operators::CompareOperator;
 use vortex_buffer::BitBufferMut;
 use vortex_buffer::BufferMut;
 use vortex_error::VortexExpect;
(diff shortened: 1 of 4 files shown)


encodings/fastlanes/src/bitpacking/compute/compare_fused.rs
```diff
@@ -133,11 +126,94 @@ where
...
+                    bit_width,
+                    packed_chunk,
+                    transposed,
+                    T::is_gt,
+                    rhs,
+                );
+            }
+            CompareOperator::NotEq | CompareOperator::Lte | CompareOperator::Gte => {
+                unreachable!("canonical_operator only returns Eq, Lt, or Gt")
+            }
```
robert3005, line 186: this doesn't seem to be true?

joseph-isaacs:
We should offer this as an option or the default. 

It halves code size. For the cost of a 16 word negation

PR #8508, opened 2026-06-19
Try and improve some noise from cuda builds on macos

AdamGS:
## Summary

Closes: #000


.github/workflows/codspeed.yml
```diff
@@ -100,10 +100,23 @@ jobs:
             --bench fsst_cuda \
             --bench runend_cuda \
             --profile bench
-      - name: Verify CUDA kernels were built
-        run: test -f vortex-cuda/kernels/gen/release/dynamic_dispatch.ptx
+      - name: Locate built CUDA kernels
+        id: cuda_kernels
+        run: |
+          ptx_file="$(find target -type f -path '*/out/kernels/*/dynamic_dispatch.ptx' -print -quit)"
```
0ax1, line 106: Does this work with debug & release configs? The gen ptx differs based on the build mode. I guess only one mode is relevant here only anyway.

PR #8509, opened 2026-06-19
assert_arrays_eq and assert_nth_scalar require ExecutionCtx

robert3005:
Part of migration to populate execution context everywhere


encodings/parquet-variant/src/compute/allnondistinct.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Lets `AllNonDistinct` compare two `ParquetVariant` arrays without canonicalizing them.
+///
+/// `AllNonDistinct` accumulates over a `Struct{lhs, rhs}` batch, so this kernel is registered for
+/// the struct encoding and inspects the two children. When both are `ParquetVariant`, we compare
+/// the typed (`typed_value`) arrays if both sides are shredded, and fall back to the raw `value`
+/// arrays otherwise. Comparing these child arrays directly avoids re-canonicalizing the variant
+/// (which would recurse through the `Variant` canonical form).
+#[derive(Debug)]
+pub struct AllNonDistinctParquetVariant;
```
onursatici: should this be a separate PR?

robert3005: Ok, I have made a change in this pr to the macro as well that used this aggregate function but I had to add it for variant. I will separate these two

robert3005: I have removed it from this pr

onursatici (approved):

PR #8510, opened 2026-06-19
feat(python): package CUDA as an optional extension

0ax1:
Keeps the base vortex-data wheel CPU-only and model CUDA support as a separate `vortex-data-cuda` extension package imported as `vortex_cuda`. The base package exposes `vortex.cuda_extension_installed()` to check whether the extension is importable, while `vortex_cuda.cuda_available()` performs the runtime CUDA probe.

Wires `vortex-data[cuda]` to depend on the exact matching `vortex-data-cuda` version, and have the CUDA package depend on the exact matching `vortex-data` version. Release automation updates those pins with the workspace version, but this change intentionally does not add CUDA wheel publishing jobs yet.

Keeps `vortex-python-cuda` out of the uv workspace so `uv --all-packages` on non-GPU CI does not build the CUDA crate chain. Add explicit CUDA Python checks to the GPU workflow instead, including the uv setup needed to run them.

Embeds generated PTX into `vortex-cuda` when available so installed wheels do not depend on build-machine kernel file paths at runtime.


0ax1:
@robert3005 & @gatesn Lemme know if this directionally makes sense.

robert3005:
@claude have a look. Is there a better way of setting version in pyproject.toml?

.github/workflows/package.yml
```diff
@@ -50,6 +50,30 @@ jobs:
...
+          from pathlib import Path
+
+          version = os.environ["VERSION"]
+          replacements = {
+              Path("vortex-python/pyproject.toml"): "vortex-data-cuda",
+              Path("vortex-python-cuda/pyproject.toml"): "vortex-data",
+          }
+          for path, package in replacements.items():
+              text = path.read_text()
+              text = re.sub(f"{package}==[^\"\\]]+", f"{package}=={version}", text)
```
robert3005: This might silently fail, not sure if we care though, might be a good sanity check to make sure that there was a hit?

robert3005 (approved):
Only the version rewrite is a nit

.github/workflows/package.yml
0ax1: changed this to fail loudly

PR #8511, opened 2026-06-19
Move optimizer kernels into a dedicated KernelSession

gatesn:
I half moved kernels from KernelSession to ArraySession. I didn't fully undo that move.


AdamGS (approved):

AdamGS:
TPC-H doesn't seem improved? I'll follow up next week, maybe there's an encoding mismatch somewhere

AdamGS:
so what's the final form? Kernel under Array or stand alone?

gatesn:
Final form is standalone. I realised we already have ScalarFn and AggregateFnSession anyway in this crate.

This doesn't fix the TPC-H query. That's hopefully this one https://github.com/vortex-data/vortex/pull/8514 !

PR #8512, opened 2026-06-19
Add a dormant, best-effort v4 RDS Postgres dual-write to the CI benchmark emitters

connortsui20:
The CI benchmark emitters currently need a manual `vortex-bench-migrate` run to populate the new v4 RDS Postgres database. This adds a best-effort, env-gated dual-write so the emitters populate it directly, ported from the unmerged `ct/bench-v4` branch onto current `develop`. It ships dormant: every v4 step is gated on `vars.GH_BENCH_INGEST_ROLE_ARN` being set and carries `continue-on-error: true`, and the existing v2 (static S3 JSON) and v3 (DuckDB + `/api/ingest`) paths are left fully intact and run first. With the variable unset the new code is dead-but-safe — a v4 write failure can never fail a workflow. Turning it on (setting the role ARN, repointing the site URL) is a separate ops change after this merges.

What's in the diff:

- A Python port of the `measurement_id` xxhash64 key (`scripts/_measurement_id.py`), the cross-language golden vectors, and a parity test asserting byte-for-byte agreement with the Rust reference. The golden vectors are the contract.
- A `--postgres` mode in `scripts/post-ingest.py`: RDS IAM auth, `verify-full` TLS, a five-table-plus-commit upsert in a single transaction, a NaN/Inf guard, and a best-effort site revalidate. The existing `--server` (v3) path stays stdlib-only and untouched.
- Testcontainer writer tests against Postgres 16, a revalidate test, and an operator cross-check utility.
- The CI wiring (parity test in the required job, testcontainer suite in a docker-gated job) and the dormant v4 ingest step added to the three emitter workflows.

The parity test and the Postgres-16 testcontainer suite run in CI. This branch also carries big-plans orchestration scaffolding under `.big-plans/` (the spine, per-step plan files, and the `plan:` commits); that's transient and will be removed in a follow-up cleanup PR rather than being part of the shipped feature. A few hardening and doc nits surfaced during review are tracked as deferred, non-blocking items in the spine.


connortsui20:
Superseded by #8513, a minimal code-only PR: just the `--postgres` writer, the `measurement_id` port, and the dormant workflow steps. The big-plans orchestration scaffolding and the test suite (testcontainer, revalidate, golden parity, cross-check) are intentionally left out of the repo.

PR #8514, opened 2026-06-19
ScalarFn validity

gatesn:
Makes validity behavior mandatory for ScalarFnVTable so we do not have the odd slightly recursive is_not_null(self) execution using a LEGACY_SESSION.


vortex-array/src/arrays/scalar_fn/vtable/validity.rs
```diff
@@ -3,72 +3,57 @@
...
-                    TypedScalarFnInstance::new(ArrayExpr, FakeEq(child.clone())).erased(),
-                    [],
-                )
-            })
-            .collect::<VortexResult<_>>()?;
-
-        let expr = Expression::try_new(array.scalar_fn().clone(), inputs)?;
-        let validity_expr = array.scalar_fn().validity(&expr)?;
+        let expr = scalar_fn_array_expr(array)?.validity()?;
+        let input = ConstantArray::new(true, array.len()).into_array();
```
robert3005, line 17: this is sketchy? But is the argument that you never touch this?

gatesn: Yeah, there's never a "root". We could make it a null literal maybe to be more obvious?

PR #8515, opened 2026-06-19
Fix bounded max partial state

gatesn:
This is a breaking change to the serialized form of BoundedMax. The ZonedLayout change to persist these types hasn't been released yet though, so it is still safe to change for now.

In case of producing a bounded max value on top of another bounded max value we need to know if the original value was null or whether it became null after producing bound. It's safe to skip chunks in the first case but not in the second thus the zoned stats have to differentiate between the two

handle https://github.com/vortex-data/vortex/issues/8166


AdamGS:
Can you add an explanation of the issue to the PR description?

robert3005 (approved):

PR #8520, opened 2026-06-20
DuckDB chunk exporter

gatesn:
DuckDB vectors are allowed to be shorter than the standard vector size, so the Vortex exporter now lets column exporters cap the next output batch if they wish.

Chunked arrays use this hook to emit one DuckDB vector per inner chunk segment. I ran into a bad case this allowed preserving dictionary vectors for chunked dictionary arrays instead of the catastrophic case of flattening or remapping dictionaries across chunk boundaries.

### DIFF
--- a/vortex-duckdb/src/exporter/mod.rs
+++ b/vortex-duckdb/src/exporter/mod.rs
@@ -5,6 +5,7 @@ mod all_invalid;
 mod bool;
 mod cache;
 mod canonical;
+mod chunked;
 mod constant;
 mod decimal;
 mod dict;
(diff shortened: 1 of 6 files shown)


vortex-duckdb/src/exporter/mod.rs
```diff
@@ -96,8 +99,18 @@ impl ArrayExporter {
             vortex_bail!("Expected {expected_cols} columns in output chunk, got {chunk_cols}");
         }
 
-        let chunk_len = duckdb_vector_size().min(self.remaining);
         let position = self.array_len - self.remaining;
+        let mut chunk_len = duckdb_vector_size().min(self.remaining);
+        if !zero_projection {
+            for field in &self.fields {
+                chunk_len = field.preferred_batch_len(position, chunk_len);
+            }
```
robert3005, line 107: there's a bit of inconsistent style where struct exporter does a fold and this does a loop

robert3005 (approved):

PR #8538, opened 2026-06-22
ReusableDict for RunEnd duckdb exporter

myrrc:
Move shared logic of creating a ReusableDict or using a created one from dict.rs to a shared file. Use this logic for RunEnd duckdb exporter as well.

Make Constant and Sequence exportes obey flatten=true flag, otherwise ReusableDict values exporting failed for them, reading garbage data.

### DIFF
--- a/vortex-duckdb/src/exporter/mod.rs
+++ b/vortex-duckdb/src/exporter/mod.rs
@@ -42,6 +42,7 @@ use vortex::error::vortex_bail;
 use vortex::error::vortex_ensure;
 
 use crate::duckdb::DataChunkRef;
+use crate::duckdb::ReusableDict;
 use crate::duckdb::VectorRef;
 use crate::duckdb::duckdb_vector_size;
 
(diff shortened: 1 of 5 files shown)


myrrc:
Looks like there's some use-after-free happening.
Minimal request that fails with this patch is TPC-H's
```
~/duckdb-vortex/build/reldebug/duckdb -c " CREATE VIEW t AS (SELECT * FROM '/home/myrrc/vortex/vortex-bench/data/tpch/1.0/vortex-file-compressed/partsupp_0.vortex'); SELECT COUNT(*) FROM t a, t b WHERE a.ps_partkey = b.ps_partkey; "
```

myrrc:
We have two options here:
1. Fully canonicalize only for reusable dict.
2. Make Constant and Sequence obey flatten=true and use new_exporter_with_flatten.

The best choice is likely the one that's most performant.

myrrc:
There is a fineweb q1 2x regression with flattening constant/sequence, not sure whether it's a real regression

0ax1:
@claude review in depth

0ax1 (approved):

myrrc:
I'll recheck the performance and merge if it doesn't reproduce

vortex-duckdb/src/exporter/mod.rs
```diff
@@ -195,6 +196,32 @@ fn new_array_exporter(
...
+    values: ArrayRef,
+    cache: &ConversionCache,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<ReusableDict> {
+    let key = values.addr();
+    if let Some(entry) = cache.dict_cache.get(&key) {
+        return Ok(entry.value().1.clone());
+    }
+    let mut dict = ReusableDict::new(values.dtype().try_into()?, values.len());
+    // ReusableDict's values must be flattened. When we call
```
0ax1, line 210: Mentioned this in a prev PR. Can you expand a bit on your block comment size in vim for visual select + gq? This looks a bit scatty in terms of linebreaks.

vortex-duckdb/src/exporter/sequence.rs
```diff
@@ -3,19 +3,30 @@
...
+use crate::exporter::ConversionCache;
+use crate::exporter::canonical;
 
 struct SequenceExporter {
     start: i64,
     step: i64,
 }
 
-pub(crate) fn new_exporter(array: &SequenceArray) -> VortexResult<Box<dyn ColumnExporter>> {
+pub(crate) fn new_exporter_with_flatten(
```
0ax1, line 21: If we only have a single exporter new fn, do we really need the `_with_flatten` suffix? I guess that's more consistent with constant array?

vortex-duckdb/src/exporter/mod.rs
myrrc: I use 80 lines comments for the codebase, is this an issue?

vortex-duckdb/src/exporter/sequence.rs
myrrc: Just to keep things coherent. We could rename the function

vortex-duckdb/src/exporter/mod.rs
0ax1: Looks inconsistent with the rest of the formatting in Vortex. Worthwhile setting it to whatever rustfmt uses.

myrrc:
Turns out this was a flaky run, merging

PR #8542, opened 2026-06-22
Write vortex-compact with Zstdbuffers instead of Zstd with unstable_encodings

### LINKED ISSUE #8541: Zstd-compact regression for strlen() pushdown
After merging https://github.com/vortex-data/vortex/pull/8303 we got a regression on zstd-encoded arrays: https://github.com/vortex-data/vortex/issues/8310#issuecomment-4743898050.

There are two issues with that:

- You can't push zstd-encoded byte_length as it decompresses.
- You need to reuse the cache, as branching over ZSTD doesn't solve the issue you recompute values over and over again for each Dict request.

The solution is to use ZstdBuffers and implement a CompareKernel and ByteLength Kernel for it.

myrrc:
If unstable_encodings feature is set (CI as an example), register ZstdBuffers and not Zstd as default write strategy. This allows using byte_length without decompressing data in Zstd.
This brings down local Clickbench Q27 run from 450 to 190ms.

Resolves: #8541


vortex-btrblocks/src/builder.rs
```diff
@@ -161,6 +161,21 @@ impl BtrBlocksCompressorBuilder {
...
     }
 
+    /// Same as with_compact() but ZstdBuffers is used instead of Zstd
+    #[cfg(all(feature = "zstd", feature = "unstable_encodings"))]
+    pub fn with_compact_buffers(self) -> Self {
+        let builder = self
+            .with_new_scheme(&string::ZstdBuffersScheme)
+            .with_new_scheme(&binary::ZstdBuffersScheme);
+
+        #[cfg(feature = "pco")]
```
myrrc, line 171: Need to register Zstd as well for unstable encodings

myrrc: On second thought, for write strategy we want Zstdbuffers to win, so no, we won't register both.

myrrc:
I've experimented with implementing ByteLength and CompareKernels, but this doesn't add any performance gains.

myrrc:
Wins:

Fineweb: performance improvements on all queries up to 4x (q2),
median improvement 2x.

Losses:

- Size increase in general bencmarks is 5-6%.
- Fineweb size +14.3%.
- Clickbench size on some files: +67% on hits_54,
  +40% on hits_55/hits_27 https://github.com/vortex-data/vortex/issues/8548
- Compression failure on TPC-DS https://github.com/vortex-data/vortex/issues/8549

cc @onursatici

robert3005:
We really shouldn't add one more config here, we should settle on some stable set of encodings, i.e. maybe we can make zstdbuffers the preferred zstd encoding

PR #8544, opened 2026-06-22
perf(duckdb): push down list length expressions

mhk197:
Pushes DuckDB's list-length scalar function into the Vortex scan as the `list_length` expression, so lengths are computed from list offsets/sizes without materializing element values.

Pushdowns supported:
- **Projection** (`SELECT len(list)` / `length(list)` / `array_length(list)`)
- **Filter** (`WHERE array_length(list) >= k`, also `len`/`length`)

Each maps to `cast(list_length(col), i64)` — DuckDB's `len`/`array_length` return `BIGINT` while `list_length` returns `u64`.

`len`/`length` are overloaded with strings/bits, so the filter path needs the argument type to disambiguate. Added a small FFI accessor `duckdb_vx_expr_get_return_type` plus `ExpressionRef::return_type()`, and gate `len`/`length`/`array_length` on the bound child being `LIST`/`ARRAY`.

Does not currently support `array_length(expr, dim)`.

Stacked on #8495.

### DIFF
--- a/vortex-duckdb/cpp/expr.cpp
+++ b/vortex-duckdb/cpp/expr.cpp
@@ -47,6 +47,14 @@ extern "C" duckdb_vx_expr_class duckdb_vx_expr_get_class(duckdb_vx_expr ffi_expr
     return static_cast<duckdb_vx_expr_class>(expr->GetExpressionClass());
 }
 
+extern "C" duckdb_logical_type duckdb_vx_expr_get_return_type(duckdb_vx_expr ffi_expr) {
+    if (!ffi_expr) {
+        return nullptr;
+    }
+    auto expr = reinterpret_cast<Expression *>(ffi_expr);
+    return reinterpret_cast<duckdb_logical_type>(&expr->return_type);
+}
+
 extern "C" const char *duckdb_vx_expr_get_bound_column_ref_get_name(duckdb_vx_expr ffi_expr) {
     if (!ffi_expr) {
         return nullptr;
--- a/vortex-duckdb/cpp/expr.cpp
+++ b/vortex-duckdb/cpp/expr.cpp
@@ -48,9 +48,7 @@ extern "C" duckdb_vx_expr_class duckdb_vx_expr_get_class(duckdb_vx_expr ffi_expr
 }
 
 extern "C" duckdb_logical_type duckdb_vx_expr_get_return_type(duckdb_vx_expr ffi_expr) {
-    if (!ffi_expr) {
-        return nullptr;
-    }
+    D_ASSERT(ffi_expr);
     auto expr = reinterpret_cast<Expression *>(ffi_expr);
     return reinterpret_cast<duckdb_logical_type>(&expr->return_type);
 }
--- a/vortex-duckdb/src/convert/expr.rs
+++ b/vortex-duckdb/src/convert/expr.rs
@@ -22,6 +22,7 @@ use vortex::expr::get_item;
 use vortex::expr::is_not_null;
 use vortex::expr::is_null;
 use vortex::expr::list_contains;
+use vortex::expr::list_length;
 use vortex::expr::lit;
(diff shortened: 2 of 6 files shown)


vortex-duckdb/cpp/expr.cpp
```diff
@@ -38,6 +38,14 @@ extern "C" duckdb_vx_expr_class duckdb_vx_expr_get_class(duckdb_vx_expr ffi_expr
     return static_cast<duckdb_vx_expr_class>(expr->GetExpressionClass());
 }
 
+extern "C" duckdb_logical_type duckdb_vx_expr_get_return_type(duckdb_vx_expr ffi_expr) {
+    if (!ffi_expr) {
```
myrrc: Please use `D_ASSERT(ffi_expr);` instead of if here.
I'll migrate other code to this macro as well.

vortex-duckdb/src/convert/expr.rs
```diff
@@ -57,6 +59,15 @@ fn from_bound_str(value: &duckdb::ExpressionRef) -> VortexResult<String> {
     }
 }
 
+/// Returns true if the expression's bound return type is a `LIST` or fixed-size `ARRAY`. Used to
+/// disambiguate the overloaded `len`/`length` functions, which also apply to strings and bits.
+fn is_list_typed(expr: &duckdb::ExpressionRef) -> bool {
```
myrrc: returns_a_list? Let's move the comment about overloaded functions to the invocation point.

vortex-duckdb/src/convert/expr.rs
```diff
@@ -173,6 +203,10 @@ pub fn can_push_expression(value: &duckdb::ExpressionRef) -> bool {
                 || name == "~~"
                 || name == "!~~"
                 || name == "strlen"
+                || (matches!(name, "len" | "length" | "array_length") && {
```
myrrc: Let's move this to a separate function?

vortex-duckdb/src/convert/expr.rs
```diff
@@ -208,6 +242,18 @@ pub fn try_from_projection_expression(
             let col = cast(col, dtype);
             Some(col)
         }
+        // `len`/`length`/`array_length` over a list column. The column dtype is known here, so
```
myrrc: I think we can reference the comment in can_pushdown_* without duplicating it

vortex-duckdb/src/duckdb/expr.rs
```diff
@@ -33,6 +35,11 @@ impl ExpressionRef {
         unsafe { cpp::duckdb_vx_expr_get_class(self.as_ptr()) }
     }
 
+    /// The (bound) return type of this expression. Borrowed from the expression.
```
myrrc: Do we care whether the return type is bound?
Filter pushdown takes place after bind() is called, so everything you can get is bound, not sure it's worth mentioning specifically.

myrrc (commented):
Overall, PR LGTM sans some small comments.
Please add sqllogictests in slt/duckdb before merging.
In the sqllogic tests, please also add tests where
array length (for same and different columns) is queried both in SELECT and WHERE.

vortex-sqllogictest/slt/duckdb/list_length_pushdown.slt
```diff
@@ -0,0 +1,123 @@
...
+WHERE array_length(b) >= 2
+ORDER BY id;
+----
+4 2
+1 3
+5 4
+
+# Mixing the len/length/array_length aliases across SELECT and WHERE.
+query I
+SELECT len(a) FROM '$__TEST_DIR__/list-length.vortex'
```
myrrc, line 163: Let's add EXPLAIN for all queries in the test to test pushdown happened

myrrc (approved):

robert3005 (approved):

PR #8546, opened 2026-06-22
Use AllNonDistinct in assert_arrays_eq and implement it for variant types

robert3005:
AllNonDistinct is exactly the semantics we want, it should be faster and in case
of failure we run the slow method to produce precise error message

### DIFF
--- a/encodings/parquet-variant/src/compute/allnondistinct.rs
+++ b/encodings/parquet-variant/src/compute/allnondistinct.rs
@@ -53,7 +53,7 @@ impl DynAggregateKernel for AllNonDistinctParquetVariant {
 
         let typed_identical = match (lhs.typed_value_array(), rhs.typed_value_array()) {
             (Some(lhs_typed), Some(rhs_typed)) => {
-                if lhs_typed.dtype().eq_ignore_nullability(rhs.dtype()) {
+                if lhs_typed.dtype().eq_ignore_nullability(rhs_typed.dtype()) {
                     all_non_distinct(lhs_typed, rhs_typed, ctx)?
                 } else {
                     return Ok(None);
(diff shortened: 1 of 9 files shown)


encodings/parquet-variant/src/compute/allnondistinct.rs
```diff
@@ -0,0 +1,65 @@
...
+        let rhs = batch.unmasked_field(1);
+        let (Some(lhs), Some(rhs)) = (
+            lhs.as_opt::<ParquetVariant>(),
+            rhs.as_opt::<ParquetVariant>(),
+        ) else {
+            return Ok(None);
+        };
+
+        let identical = match (lhs.typed_value_array(), rhs.typed_value_array()) {
+            (Some(lhs_typed), Some(rhs_typed)) => all_non_distinct(lhs_typed, rhs_typed, ctx)?,
```
onursatici: does falling here mean the values array is empty for both?

Shouldn't we compare typed_value of both first, then the fallback values for both, then return identical only if both identical?

robert3005: Ok I went through the spec, we can compare the shredded parts if they have equal schemas, if they don't we need to shred the values. We still need to compare the unshredded values since values might be partially shredded

encodings/parquet-variant/src/compute/allnondistinct.rs
```diff
@@ -0,0 +1,195 @@
...
+                } else {
+                    return Ok(None);
+                }
+            }
+            _ => true,
+        };
+
+        if typed_identical {
+            let values_identical = match (lhs.value_array(), rhs.value_array()) {
+                (Some(lhs_value), Some(rhs_value)) => all_non_distinct(lhs_value, rhs_value, ctx)?,
```
onursatici, line 67: shall we do a `(None, None) => return true` here because we know typed_identical is true, so there is no need to fallback if they both are fully shredded

robert3005: done

onursatici (approved):

PR #8550, opened 2026-06-23
group ids as array ref, multi encoding kernel lookup

onursatici:
Signed-off-by: Onur Satici <onur@spiraldb.com>

## Summary

Closes: #000

### DIFF
--- a/vortex-array/src/aggregate_fn/accumulator_grouped.rs
+++ b/vortex-array/src/aggregate_fn/accumulator_grouped.rs
@@ -14,7 +14,6 @@ use crate::aggregate_fn::AggregateFn;
 use crate::aggregate_fn::AggregateFnRef;
 use crate::aggregate_fn::AggregateFnVTable;
 use crate::aggregate_fn::DynAccumulator;
-use crate::aggregate_fn::kernels::GroupedAggregateKernelResult;
 use crate::aggregate_fn::session::AggregateFnSessionExt;
 use crate::array::ArrayId;
 use crate::arrays::PrimitiveArray;
@@ -65,22 +64,6 @@ impl GroupIds {
     /// Create group ids from materialized values.
     pub fn from_iter(ids: impl IntoIterator<Item = u32>, num_groups: usize) -> VortexResult<Self> {
         Self::from_buffer(Buffer::from_iter(ids), num_groups)
-    }
-
-    /// Create group ids containing `0..num_groups`.
-    pub fn range(num_groups: usize) -> VortexResult<Self> {
-        validate_num_groups(num_groups)?;
-        if num_groups == 0 {
-            return Self::from_buffer(Buffer::<u32>::empty(), num_groups);
-        }
-
-        let last = u32::try_from(num_groups - 1).map_err(|_| {
-            vortex_err!(
-                "num_groups {} exceeds dense u32 group id capacity",
-                num_groups
-            )
-        })?;
-        Self::from_buffer((0..=last).collect(), num_groups)
     }
 
     /// Return the encoded ids array.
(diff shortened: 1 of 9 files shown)


vortex-array/src/aggregate_fn/fns/count/grouped.rs
```diff
@@ -3,20 +3,51 @@
...
+use crate::aggregate_fn::kernels::DynGroupedAggregateKernel;
+use crate::aggregate_fn::kernels::GroupedAggregateKernelResult;
+use crate::arrays::PrimitiveArray;
 
-pub(super) fn try_accumulate_grouped(
-    states: &mut [u64],
+#[derive(Debug)]
+pub(crate) struct CountGroupedKernel;
+
+impl DynGroupedAggregateKernel for CountGroupedKernel {
```
gatesn: Don't we typically have non-object-safe typed kernels with associated types, and then lift them into the DynTrait when we register?

vortex-array/src/aggregate_fn/fns/sum/grouped.rs
```diff
@@ -25,6 +35,61 @@ use crate::match_each_native_ptype;
...
+        aggregate_fn: &AggregateFnRef,
+        batch: &ArrayRef,
+        group_ids: &GroupIds,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Option<GroupedAggregateKernelResult>> {
+        let Some(options) = aggregate_fn.as_opt::<Sum>() else {
+            return Ok(None);
+        };
+
+        let columnar = batch.clone().execute::<Columnar>(ctx)?;
```
gatesn: Why execute to Columnar if we never match on Constant?

vortex-array/src/aggregate_fn/accumulator_grouped.rs
```diff
@@ -17,16 +16,107 @@ use crate::aggregate_fn::AggregateFnVTable;
...
+            return Self::from_buffer(Buffer::<u32>::empty(), num_groups);
+        }
+
+        let last = u32::try_from(num_groups - 1).map_err(|_| {
+            vortex_err!(
+                "num_groups {} exceeds dense u32 group id capacity",
+                num_groups
+            )
+        })?;
+        Self::from_buffer((0..=last).collect(), num_groups)
```
gatesn: Wouldn't we want a sequence array?

vortex-array/src/aggregate_fn/kernels.rs
```diff
@@ -35,20 +35,27 @@ pub trait DynAggregateKernel: 'static + Send + Sync + Debug {
...
+    pub fn new(group_ids: GroupIds, partials: ArrayRef) -> Self {
         Self {
             group_ids,
             partials,
         }
     }
 
-    pub fn group_ids(&self) -> &[u32] {
-        self.group_ids.as_ref()
+    pub fn dense(partials: ArrayRef, num_groups: usize) -> VortexResult<Self> {
```
gatesn: Is num_groups == partials.len()?

vortex-array/src/aggregate_fn/session.rs
```diff
@@ -67,15 +68,14 @@ impl SessionVar for AggregateFnSession {
 }
 
 type AggregateKernelKey = (ArrayId, Option<AggregateFnId>);
-type GroupedEncodingKernelKey = (ArrayId, AggregateFnId);
+type GroupedAggregateKernelKey = (AggregateFnId, Option<ArrayId>, Option<ArrayId>);
```
gatesn: Might be worth a struct at this point

gatesn (changes requested):

PR #8556, opened 2026-06-23
GDB pretty-printers

myrrc:
GDB pretty-printers for ArrayRef, `Array<T>`, Id,
and `Arc<dyn T>` like `Arc<dyn ArrayPlugin>`


AdamGS:
can it be in `scripts/`? I think the root of the repo is already very wide.

robert3005 (approved):

PR #8564, opened 2026-06-24
fix(array): don't panic on unsupported arrow types

### LINKED ISSUE #8346: DType::from_arrow panics with `unimplemented!` for Duration, Interval and FixedSizeBinary instead of returning an error

`DType::from_arrow` hits the `unimplemented!()` fallback arm in `vortex-array/src/dtype/arrow.rs` for unsupported Arrow types, which surfaces in Python as a `pyo3_runtime.PanicException` instead of a clean `ValueError`.

```python
import pyarrow as pa
import vortex as vx

vx.array(pa.table({"c0": pa.array([1, 2], type=pa.duration("us"))}))
# pyo3_runtime.PanicException: not implemented: Arrow data type not yet supported: Duration(Microsecond)

vx.array(pa.table({"c0": pa.array([b"abc"], type=pa.binary(3))}))
# pyo3_runtime.PanicException: not implemented: Arrow data type not yet supported: FixedSizeBinary(3)
```

The fallback arm should return a `VortexError` (e.g. via a `TryFrom`-style conversion) so unsupported types fail cleanly. The fuzzer currently excludes `Duration` and `FixedSizeBinary` from generation; once this is fixed, those exclusions can be lifted and the clean error will be routed to the existing unsupported-type skip path.

miniex:
## Summary

`DType::from_arrow` hit the `unimplemented!` fallback arm for unsupported Arrow types (`Duration`, `Interval`, `FixedSizeBinary`), which surfaced in Python as a `PanicException` instead of a clean error. I added a fallible `TryFromArrowType` conversion that returns a `VortexError` on that arm, kept the infallible `FromArrowType` as a thin delegate for existing trusted callers, and routed the `ArrowSession` leaf path and the Python entry points through it so unsupported types raise a clean `ValueError`.

The error keeps the `not supported` marker so it lands in the fuzzer's existing unsupported-type skip path; the `Duration`/`FixedSizeBinary` generation exclusions can be lifted once the fuzzer is on `develop`.

Closes: #8346

## Testing

Added an rstest in `vortex-array/src/dtype/arrow.rs` checking `Duration`/`Interval`/`FixedSizeBinary` error cleanly as a leaf and nested in a field/schema, plus a parametrized regression test in `vortex-python/test/test_array.py` for the issue's repro; both of the issue's Python snippets now raise `ValueError` instead of panicking. `cargo nextest run -p vortex-array` (3060 tests) and `uv run --all-packages pytest vortex-python/test/test_array.py` pass; `fmt --all`, `clippy --all-targets --all-features`, `cargo doc`, `basedpyright` and `ruff` are clean.

---

### DIFF
--- a/vortex-array/src/arrow/session.rs
+++ b/vortex-array/src/arrow/session.rs
@@ -57,7 +57,7 @@ use crate::dtype::FieldName;
 use crate::dtype::FieldNames;
 use crate::dtype::Nullability;
 use crate::dtype::StructFields;
-use crate::dtype::arrow::FromArrowType;
+use crate::dtype::arrow::TryFromArrowType;
 use crate::dtype::arrow::to_data_type_naive;
 use crate::dtype::extension::ExtId;
 use crate::extension::datetime::AnyTemporal;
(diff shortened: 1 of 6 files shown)


AdamGS (approved):
This seems good, just running CI to see if everything else is good.

robert3005:
~Shouldn't this be migrated to the session arrow conversion?~ it's a bigger change

Issue #8434, opened 2026-06-15
Tracking Issue: Stateful `Scheme`s


connortsui20:
This is a tracking issue for making stateful `Scheme`s in the compressor easier to utilize and work with.

## Motivation

There are cases where we can define better `Scheme`s at runtime with more information. For example, we might want to pre-train an `FSST` compressor and then use that table for all of the arrays we want to compress (instead of training them every single time). 

## Design

TODO not super clear yet.

## Steps

- [ ] Initial implementation
- [ ] Documentation
- [ ] Public API stabilization

## Unresolved questions

- [ ] What is the best way to do this???
- [ ] For Stateful `Scheme`s, do we need to accept `Arc<Scheme>`? And should we update `Scheme` to look more like our `VTable`s? Is that even better?

## Implementation history

- Nothing yet.

connortsui20:
@tomsanbear I'm moving this to a dedicated tracking issue:

> Discussed on [Slack](https://vortex-data.slack.com/archives/C07BV3GKAJ2/p1781532902617609)
> 
> I have a working branch exploring the `Arc<dyn Scheme>` approach for stateful schemes, with a pretrained FSST variant as the first use case: [develop...tomsanbear:vortex:feat/fsst-pretrained-compressor](https://github.com/vortex-data/vortex/compare/develop...tomsanbear:vortex:feat/fsst-pretrained-compressor)
> 
> Saw ~10x on the isolated FSST compression path and ~21% end-to-end on a streaming insert workload. Using the `Box::leak` workaround for now per Connor's suggestion. Happy to help when the design for stateful schemes is ready.

Issue #919, opened 2024-09-24
ALP exponent sampling improvements


lwwmanning:
The original ALP paper, as well as the [duckdb implementation](https://github.com/duckdb/duckdb/blob/v1.1.1/src/include/duckdb/storage/compression/alp/algorithm/alp.hpp#L200-L248), does two-level sampling to find exponents, which should materially improve throughput for larger datasets.

Notably, in duckdb, they have row groups of 60 vectors, of 2048 elements each (i.e., 122880 elements). They do a stratified sample (8 samples of 32 elements) within those 122K elements to find the top ~5 combinations of exponents, and then they pick from those top 5 exponent pairs to compress each 2048 element vector individually.

I like the duckdb criteria for determining top N combinations: https://github.com/duckdb/duckdb/blob/v1.1.1/src/include/duckdb/storage/compression/alp/algorithm/alp.hpp#L136-L154

We could either:

1. (Simplest) take the top N candidate exponent pairs from the first chunk of 65K elements, and only consider those as candidates for subsequent chunks of that column
2. (More complex, potentially more robust?) do a stratified sample across all rows of the entire column (i.e., across all chunks), fit per sample, find top N exponents, then pass that set to compress per chunk.

x-at-01:
Regarding sampling overhead and exponent discovery in ALP:

In [fastalp](https://github.com/webc-site/fastalp) (version 0.1.27 on [crates.io](https://crates.io/crates/fastalp)), I evaluated sampling performance and introduced three search pruning techniques:

- Dynamic cost lower-bound pruning: The search cost ceiling is bounded by uncompressed raw storage. If running exception penalties exceed best_cost, parameter evaluation breaks immediately.
- Fast-path for common decimal precision: On 32-sample blocks, if an exponent e produces zero exceptions, the search skips factor exploration, avoiding candidate matrix traversal.
- Monotonic delta heuristic: On continuous series, fastalp compares adjacent diffs against range spread during sampling to decide whether to activate Delta-ALP.
- Performance: Sampling latency is approximately 200 ns per 1024-element block. Single-core register SIMD decompression sustains 23.55 GB/s geometric mean (peaking at 55 to 77 GB/s), and end-to-end compression operates at 3.55 GB/s geometric mean (4.4 to 6.8 GB/s in regular blocks, up to 20+ GB/s with parameter caching).

Sharing these empirical pruning results from fastalp for reference.

[...]

PR #8565, opened 2026-06-24
perf(alp): avoid per-combination allocations when searching exponents

miniex:
## Summary

`find_best_exponents` ran a full `encode` for every `(e, f)` candidate just to size the result, allocating the encoded, patch and chunk-offset buffers across the whole `e`/`f` search space. It now estimates the size in a single allocation-free pass via `estimate_encoded_size_for_exponents`, which reproduces the existing `estimate_encoded_size` exactly, so the chosen exponents (and therefore the compressed output) are bit-for-bit unchanged.

This is the throughput half of #919; the two-level stratified sampling / top-N reuse the issue describes would change which exponents are picked and depends on per-column scheme state (#8434), so I left it out here.

## Benchmark

`encodings/alp/benches/alp_compress.rs`, compress medians (type, length, patch-fraction, valid-fraction):

| case | before | after | speedup |
| --- | --- | --- | --- |
| f64 (1000, 0.0, 1.0) | 39.1 us | 14.0 us | 2.8x |
| f64 (1000, 0.1, 0.25) | 45.0 us | 16.5 us | 2.7x |
| f32 (1000, 0.0, 1.0) | 14.6 us | 5.8 us | 2.5x |
| f64 (10000, 0.0, 1.0) | 53.1 us | 28.6 us | 1.9x |
| f64 (10000, 0.1, 0.25) | 73.7 us | 49.4 us | 1.5x |

## Testing

`cargo nextest run -p vortex-alp -p vortex-btrblocks` passes, including a new equivalence test (`estimate_for_exponents_matches_full_encode`) that asserts the allocation-free estimate equals a full `encode` + `estimate_encoded_size` for every candidate exponent across kept, patched and all-patched inputs. `fmt` + `clippy --all-targets --all-features` + `cargo doc` clean.

---

### DIFF
--- a/encodings/alp/src/alp/mod.rs
+++ b/encodings/alp/src/alp/mod.rs
@@ -30,7 +30,48 @@ mod tests {
     use vortex_array::patches::PatchesMetadata;
     use vortex_array::test_harness::check_metadata;
 
+    use crate::alp::ALPFloat;
+    use crate::alp::Exponents;
     use crate::alp::array::ALPMetadata;
+
+    // The allocation-free estimate must match a full encode + estimate for every candidate
+    // exponent pair, so `find_best_exponents` picks the same exponents and compression is unchanged.
+    fn check_estimate_matches<T: ALPFloat>(values: &[T]) {
+        for e in 0..T::MAX_EXPONENT {
+            for f in 0..e {
+                let exp = Exponents { e, f };
+                let lightweight = T::estimate_encoded_size_for_exponents(values, exp);
+                let (_, encoded, _, patches, _) = T::encode(values, Some(exp));
+                let full = T::estimate_encoded_size(&encoded, &patches);
+                assert_eq!(
+                    lightweight,
+                    full,
+                    "mismatch at e={e}, f={f}, len={}",
+                    values.len()
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn estimate_for_exponents_matches_full_encode() {
+        // Clean 2-decimal values (mostly kept), repeating decimals (many patches), large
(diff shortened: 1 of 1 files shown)


encodings/alp/src/alp/mod.rs
```diff
@@ -120,27 +161,72 @@ pub trait ALPFloat: private::Sealed + Float + Display + NativePType {
...
+        // matches its output; only when every value is patched (no fill value) is the full range used.
+        let mut kept_min: Option<Self::ALPInt> = None;
+        let mut kept_max: Option<Self::ALPInt> = None;
+        let mut all_min: Option<Self::ALPInt> = None;
+        let mut all_max: Option<Self::ALPInt> = None;
+        let mut patch_count = 0usize;
+
+        for &value in values {
+            let encoded = Self::encode_single_unchecked(value, exponents);
+            all_min = Some(all_min.map_or(encoded, |m| m.min(encoded)));
```
AdamGS: can you extract this bounds logic into a function, and always track an `Option<(Self::ALPInt, Self::ALPInt)>`? You always update both, the fallback value is shared and its the same logic for the "kept" and "all" bounds (also worth explaining what each one means).

AdamGS:
very cool optimization! just one ask

encodings/alp/src/alp/mod.rs
miniex: done, two Option<(min, max)> + a shared helper now. thanks!

AdamGS: thanks!

AdamGS (approved):
Thanks for this contribution! great improvement.

PR #8566, opened 2026-06-24
Separate and merge .h/.hpp files in duckdb

myrrc:
- Merge small .h files into vortex_duckdb.h
- Separate C++ code not used in bindgen to .hpp files
- Remove dead code from duckdb: duckdb_vector_to_string, Config,  
ClientContext, duckdb_vx_add_extension_option, DDBString creation.
- Remove rest duckdbfs code. Duckdbfs was partially removed in #8468 but the Rust-side code survived.


vortex-duckdb/cpp/vector.cpp
```diff
@@ -46,10 +37,8 @@ duckdb_vx_sequence_vector(duckdb_vector c_vector, int64_t start, int64_t step, i
     vector->Sequence(start, step, capacity);
 }
 
-namespace vortex {
-
 // This is a complete hack to access the data buffer and pointer of a vector.
-class DataVector : public Vector {
+class DataVector final : public Vector {
```
0ax1: Why are adding final to the classes? Given this is not a public API we expose, is there any perf benefit? otherwise I'd suggest dropping it.

vortex-duckdb/cpp/vector.cpp
```diff
@@ -103,16 +90,16 @@ extern "C" void duckdb_vx_vector_set_validity_data(duckdb_vector ffi_vector,
                                                    idx_t capacity,
                                                    duckdb_vx_vector_buffer buffer,
                                                    void *data_ptr) {
-    auto dvector = reinterpret_cast<vortex::DataVector *>(ffi_vector);
```
0ax1, line 106: this was more readable: `vortex::DataVector`. from looking at `DataVector` alone, I can't  see whether the type comes from ddb or vortex.

vortex-duckdb/src/lib.rs
```diff
@@ -67,12 +65,6 @@ fn init_tracing() {
 /// Note: This also registers extension options. If you want to register options
 /// separately (e.g., before creating connections), call `register_extension_options` first.
 pub fn initialize(db: &DatabaseRef) -> VortexResult<()> {
-    db.config().add_extension_options(
```
0ax1, line 70: removing the filesystem should go in a separate pr. we need to check all docs around that to reflect the change. i think we mention in the ext repo that we support duckdb fs.

myrrc: We've already dropped duckdbfs, the PR just didn't remove the C++-side code.

vortex-duckdb/cpp/vector.cpp
myrrc: There is no perf benefit, it's a correctness guarantee. Underlying types are not DataVectors but Vectors, the only reason this works is because DataVector doesn't add any class members. final is a protection from anyone accidentally inheriting from it and adding class members, which would be UB.

vortex-duckdb/cpp/vector.cpp
myrrc: We can call it a VortexDataVector if you wish

vortex-duckdb/src/lib.rs
myrrc: I'll open a separate PR to the extension repo once we merge this.

vortex-duckdb/cpp/vector.cpp
myrrc: We should also add a static_assert (sizeof(DataVector) == sizeof(Vector);

0ax1 (approved):

PR #8568, opened 2026-06-24
Decimal cast supports inplace lossless casts

robert3005:
This is functionality we have in primitive arrays. We can avoid copying if the
cast can be done in place


robert3005:
added benchmarks separately in https://github.com/vortex-data/vortex/pull/8569

robert3005:
ok, our execute_parent dispatch always passes &ArrayRef :/

robert3005:
This was a long way to figure out we still have borrowed reference

PR #8574, opened 2026-06-24
Validate row-index selections

gatesn:
## Rational for this change

We binary search the row selections and therefore panic if they are out of order. Therefore we should validate them on-construction to be sorted

Include/ExcludeByIndex has try_new constructors.

An alternative would be to have a SortedBuffer<T> wrapper type perhaps?

Include/ExcludeByIndex

### DIFF
--- a/vortex-jni/src/scan.rs
+++ b/vortex-jni/src/scan.rs
@@ -43,7 +43,7 @@ use vortex::scan::PartitionRef;
 use vortex::scan::PartitionStream;
 use vortex::scan::ScanRequest;
 use vortex::scan::selection::Selection;
-use vortex::scan::selection::StrictSortedBuffer;
+use vortex::scan::strict_sorted_buffer::StrictSortedBuffer;
 use vortex_arrow::ArrowSessionExt;
 
 use crate::POOL;
--- a/vortex-jni/src/scan.rs
+++ b/vortex-jni/src/scan.rs
@@ -94,8 +94,8 @@ fn build_scan_request(
 
     let selection = match selection_include {
         0 => Selection::All,
-        1 => Selection::IncludeByIndex(Buffer::copy_from(selection_idx)),
-        2 => Selection::ExcludeByIndex(Buffer::copy_from(selection_idx)),
+        1 => Selection::include_by_index(Buffer::copy_from(selection_idx))?,
+        2 => Selection::exclude_by_index(Buffer::copy_from(selection_idx))?,
         3 => Selection::IncludeRoaring(deserialize_roaring_selection(selection_roaring_bitmap)?),
         4 => Selection::ExcludeRoaring(deserialize_roaring_selection(selection_roaring_bitmap)?),
         other => vortex_bail!("unknown selection include code: {other}"),
--- a/vortex-jni/src/scan.rs
+++ b/vortex-jni/src/scan.rs
@@ -43,6 +43,7 @@ use vortex::scan::PartitionRef;
 use vortex::scan::PartitionStream;
 use vortex::scan::ScanRequest;
 use vortex::scan::selection::Selection;
+use vortex::scan::selection::StrictSortedBuffer;
 use vortex_arrow::ArrowSessionExt;
 
 use crate::POOL;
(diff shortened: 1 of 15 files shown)


vortex-jni/src/scan.rs
```diff
@@ -96,8 +97,12 @@ fn build_scan_request(
...
     let selection = match selection_include {
         0 => Selection::All,
-        1 => Selection::IncludeByIndex(Buffer::copy_from(selection_idx)),
-        2 => Selection::ExcludeByIndex(Buffer::copy_from(selection_idx)),
+        1 => Selection::IncludeByIndex(StrictSortedBuffer::try_new(Buffer::copy_from(
+            selection_idx,
+        ))?),
+        2 => Selection::ExcludeByIndex(StrictSortedBuffer::try_new(Buffer::copy_from(
+            selection_idx,
+        ))?),
```
robert3005, line 103: can you remove the validation from java? JNI actually gets validated selections

gatesn: I have no idea where you want to assert that? Seems odd to accept it unchecked over FFI?

robert3005: If you look at the code you will notice that with this change this validation happens twice. I am asking you to have it happen once again as before this change. The validation happens on the java side of the jni bindings

robert3005: in particular the logic that I am talking about is here https://github.com/vortex-data/vortex/blob/develop/java/vortex-jni/src/main/java/dev/vortex/api/ScanOptions.java#L94

gatesn: I am saying it's weird for this check to happen on the Java side, because then the FFI side just accepts unsafe input? I guess it's accepting raw pointers though so like...!! :D

robert3005: the jni c api is effectively private so wherever it does validation seems of little consequence but agree it lets you use unvalidated indices. It should have been in the rust api. I haven't followed up after I added the java side validation

vortex-scan/src/selection.rs
```diff
@@ -7,22 +7,97 @@ use std::ops::Not;
...
+    /// Return the sorted buffer.
+    pub fn into_inner(self) -> Buffer<T> {
+        self.buffer
+    }
+
+    /// Return true if the buffer contains no values.
+    pub fn is_empty(&self) -> bool {
+        self.buffer.is_empty()
+    }
+
```
joseph-isaacs: its own file please

vortex-scan/src/selection.rs
```diff
@@ -7,22 +7,97 @@ use std::ops::Not;
...
+    fn try_from(value: Buffer<T>) -> Result<Self, Self::Error> {
+        Self::try_new(value)
+    }
+}
+
+impl<T> From<StrictSortedBuffer<T>> for Buffer<T> {
+    fn from(value: StrictSortedBuffer<T>) -> Self {
+        value.into_inner()
+    }
```
joseph-isaacs: please remove this they are too expensive for a from

vortex-scan/src/selection.rs
```diff
@@ -7,22 +7,97 @@ use std::ops::Not;
...
+}
+
+impl<T: Ord> StrictSortedBuffer<T> {
+    /// Create a new buffer, failing if the values are not strictly increasing.
+    pub fn try_new(buffer: Buffer<T>) -> VortexResult<Self> {
+        validate_strictly_sorted(buffer.as_slice())?;
+        Ok(Self { buffer })
+    }
+}
```
joseph-isaacs: shall we have ann unnsafe variant?

vortex-scan/src/selection.rs
```diff
@@ -7,22 +7,97 @@ use std::ops::Not;
...
+#[derive(Clone, Debug)]
+pub struct StrictSortedBuffer<T> {
+    buffer: Buffer<T>,
+}
+
+impl<T> StrictSortedBuffer<T> {
+    /// Return the sorted values.
+    pub fn as_slice(&self) -> &[T] {
+        self.buffer.as_slice()
+    }
```
joseph-isaacs: both this and as ref?

vortex-scan/src/selection.rs
```diff
@@ -7,22 +7,97 @@ use std::ops::Not;
...
+}
+
+impl<T> std::ops::Deref for StrictSortedBuffer<T> {
+    type Target = [T];
+
+    fn deref(&self) -> &Self::Target {
+        self.as_slice()
+    }
+}
```
joseph-isaacs: you can a slice or a buffer here?

robert3005 (approved):

PR #8575, opened 2026-06-24
Support session-time source configuration in DataFusion, and clear up precedence between config sources

AdamGS:
## Rational for this change

Makes it possible to configure Vortex FileSource-based tables using `SET` statements on an appropriately configured DataFusion session. This is mostly useful for testing or interactive workflows.

1. Use DataFusion's `ExtensionOptions` and `ConfigExtension` APIs instead of `config_namespace` to get configuration handling for free.
2. Adds SLT based tests for table and session configuration.
3. Better document the order of precedence between the various sources of config, trying to mirror the built-in file formats.

For users of `VortexSource` and `VortexFormat`, some APIs might change slightly, and if they expose Vortex through an interactive interface, the config might change slightly.


a10y (approved):
looks good!

PR #8576, opened 2026-06-24
Make Shared transparent to parent kernels

gatesn:
## Rational for this change

`SharedArray` is a cache wrapper, but parent-kernel execution currently treats it as an ordinary child encoding. That can force canonicalization before the wrapped array gets a chance to handle the parent operation itself. This makes `SharedArray` transparent for parent reductions and parent execution, so existing encoding-specific kernels still fire through shared wrappers.

No tracked issue.

`SharedArray` now delegates parent reduction to its current wrapped array. Parent execution also checks the wrapped array before falling back to ordinary execution, avoiding unnecessary cache population and canonicalization.

No public API changes. Query results are unchanged; this only changes internal execution dispatch.

### DIFF
--- a/vortex-array/src/executor.rs
+++ b/vortex-array/src/executor.rs
@@ -34,6 +34,8 @@ use crate::ArrayRef;
 use crate::Canonical;
 use crate::IntoArray;
 use crate::array::ArrayId;
+use crate::arrays::Shared;
+use crate::arrays::shared::SharedArrayExt;
 use crate::builders::ArrayBuilder;
 use crate::builders::builder_with_capacity_in;
 use crate::dtype::DType;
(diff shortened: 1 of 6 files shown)


AdamGS:
I think we caused some weird race in github, but benchmarks seem to be running https://github.com/vortex-data/vortex/actions/runs/28109422384

vortex-array/src/executor.rs
```diff
@@ -568,6 +570,35 @@ fn execute_parent_for_child(
     slot_idx: usize,
     kernels: &ParentExecutionKernels,
     ctx: &mut ExecutionCtx,
+) -> VortexResult<Option<ArrayRef>> {
+    if let Some(result) = execute_parent_for_exact_child(parent, child, slot_idx, kernels, ctx)? {
+        return Ok(Some(result));
+    }
```
lwwmanning, line 581: add some sort of test?

vortex-array/src/executor.rs
```diff
@@ -568,6 +570,35 @@ fn execute_parent_for_child(
...
+    {
+        if let Some(result) =
+            execute_parent_for_exact_child(parent, &source, slot_idx, kernels, ctx)?
+        {
+            return Ok(Some(result));
+        }
+        current = source;
+    }
+
+    Ok(None)
```
robert3005: Is there a way to register a kernel for shared for all parents? Then we can avoid this

joseph-isaacs:
I don't think we want this change. This will mean that sometimes we will reach into a canonical cache and fetch the compressed array. 

I think we would want to not construct this array if we wanted that behavoiur

PR #8577, opened 2026-06-24
Chunked dict values pullup

gatesn:
## Rational for this change

Repeated dictionary chunks can share the exact same values array while keeping separate codes arrays. In that shape, later predicate and scalar-function pushdown works better if the common values array is above the chunked codes rather than repeated under every chunk. This adds that optimizer rewrite with a strict pointer-identity precondition.

No tracked issue.

Adds an optimizer rule that rewrites `Chunked<Dict<codes_i, values>>` into `Dict<Chunked<codes_i>, values>` when all chunks are dictionaries with the same values allocation and compatible code dtypes. It also registers the parent kernel needed for dictionary-over-chunked execution and adds tests for both the shared-values rewrite and the distinct-values no-op case.

No public API changes. Optimized array shape may change internally, but logical array values are unchanged.

Note: I don't expect this to impact current develop since we do not return ChunkedArrays from scans that use SplitBy::Layout

### DIFF
--- a/vortex-array/src/arrays/dict/vtable/kernel.rs
+++ b/vortex-array/src/arrays/dict/vtable/kernel.rs
@@ -4,6 +4,7 @@
 use vortex_session::VortexSession;
 
 use crate::ArrayVTable;
+use crate::arrays::Chunked;
 use crate::arrays::Dict;
 use crate::arrays::dict::TakeExecuteAdaptor;
 use crate::optimizer::kernels::ArrayKernelsExt;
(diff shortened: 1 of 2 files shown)


lwwmanning (approved):

vortex-array/src/arrays/dict/vtable/kernel.rs
```diff
@@ -16,6 +17,7 @@ use crate::scalar_fn::fns::fill_null::FillNullExecuteAdaptor;
 pub(crate) fn initialize(session: &VortexSession) {
     let kernels = session.kernels();
     kernels.register_execute_parent_kernel(Binary.id(), Dict, CompareExecuteAdaptor(Dict));
+    kernels.register_execute_parent_kernel(Dict.id(), Chunked, TakeExecuteAdaptor(Chunked));
```
robert3005, line 20: Did this exist before and was not registered?

robert3005 (approved):

PR #8584, opened 2026-06-24
Improve clickbench-sorted to better reflect sorted data

AdamGS:
## Rationale for this change

The current setup is too sorted, which makes any sort pushdown inconsequential.

While keeping the files globally sorted (now just by EventTime), their names are shuffled in a consistent way, so the their name's lexicographical order doesn't match the data's order.

Just benchmarks.


myrrc (approved):

0ax1 (approved):
lgtm. one note: the chage will not invalidate an existing cache such that all all clients automatically migrate to sorted shards.

PR #8586, opened 2026-06-24
Remove data generation code from engine-specific bench binaries

AdamGS:
## Rationale for this change

Part of https://github.com/vortex-data/vortex/issues/9329.

This is a step towards making the benchmark binaries smaller, shifting as much of the functionality into the python tool

Removes data generation from the engine specific binaries, replacing it with an informative error.

None


vortex-bench/src/lib.rs
```diff
@@ -245,6 +245,66 @@ impl CompactionStrategy {
...
+    let requested_formats = formats
+        .iter()
+        .copied()
+        .unique()
+        .map(|format| format!("\"{format}\""))
+        .join(",");
+
+    anyhow::bail!(
+        "prepared data is missing for {}: {missing_data}. Generate it first with \
+         `vx-bench prepare-data {} --formats-json '[{requested_formats}]'` or \
```
0ax1, line 306: In `bail` we print, `--formats-json '["vortex-file-compressed"]'`. But we don't accept `--formats-json '["vortex-file-compressed"]'` on the CLI?

AdamGS: We do! this is the python tool, it has its own CLI args.

myrrc (approved):

0ax1 (approved):
lgtm, just gotta make ci happy here

myrrc (approved):

Issue #1907, opened 2025-01-11
Expression partitioning to respect scope validity


gatesn:
Currently the expression partitioner will not correctly respect the validity of the scope array.

This is blocked on a mask expression existing: #1900

gatesn:
Is this true? Or would any expression that altered the result array validity already be part of the root expr, and therefore respected?

Maybe the problem is that we need to ensure that when getting the child field of a layout, that the parent's validity is pushed down. We could enforce this at write time.

joseph-isaacs:
I think that get_item needs to respect validity and then it should work.

robert3005:
We assert that structs when being written don't have top level nullability

PR #1900, opened 2025-01-10
feat: mask

danking:
Mask sets entries of an array to null. I like the analogy to light: the array is a sequence of lights (each value might be a different wavelength). Null is represented by the absence oflight. Placing a mask (i.e. a piece of plastic with slits) over the array causes those values where the mask is present (i.e. "on", "true") to be dark.

An example in pseudo-code:

```rust
a = [1, 2, 3, 4, 5]
a_mask = [t, f, f, t, f]
mask(a, a_mask) == [null, 2, 3, null, 5]
```

Specializations
---------------

I only fallback to Arrow for two of the core arrays:

- Sparse. I was skeptical that I could do better than decompressing and applying it.
- Constant. If the mask is sparse, SparseArray might be a good choice. I didn't investigate.
- Extension. (Now a follow up due to comments)

For the non-core arrays, I'm missing the following. I'm not clear that I can beat decompression forrun end. The others are easy enough but some amount of typing and testing.

- fastlanes
- fsst
- roaring
- runend
- runend-bool
- zigzag
- ALP (Now a follow up due to comments)

Naming
------

Pandas also calls this operation [`mask`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.mask.html) but accepts an optional second argument which is an array of values to use instead of null (which makes Pandas' mask more like an `if_else`).

Arrow-rs calls this [`nullif`](https://arrow.apache.org/rust/arrow/compute/fn.nullif.html).

Arrow-cpp has [`if_else(condition, consequent, alternate)`](https://arrow.apache.org/docs/cpp/compute.html#cpp-compute-scalar-selections) and [`replace_with_mask(array, mask, replacements)`](https://arrow.apache.org/docs/cpp/compute.html#replace-functions) both of which can implement our `mask` by passing a `NullArray` as the third argument.

## Follow ups:
1. Extension array masking.
2. ALP array masking.
3. Add an expression for masking.
4. Assess the situation with nested structs in files and make sure we're preserving validity correctly.

PR #8590, opened 2026-06-25
fix(layout): preserve nullability of structs nested in structs

### LINKED ISSUE #8348: Struct field nested directly inside another struct loses nullability on file round-trip

miniex:
## Summary

A nullable struct field nested directly inside another single-field struct came back non-nullable on a file round-trip (`{c0={a={b=i32}?}?}` was read back as `{c0={a={b=i32}}?}`), silently dropping the inner `?`. The struct reader's single-partition fast path delegated this scope's reconstruction to the child, so chained single-field structs accumulated `pack` wrappers and each scope's validity landed at the wrong nesting level. I stopped the `partition` splitter from collapsing `pack`/`merge` reconstruction nodes, and made the struct reader take the fast path only when nothing is reconstructed at this level, so reconstruction now flows through the multi-partition path, one level per scope. Multi-field structs were already unaffected.

The fuzzer's struct-of-struct `assume()` guard can be dropped once the fuzzer is on `develop`.

Closes: #8348

## Testing

Added a round-trip regression test in `vortex-file/tests/test_write_table.rs` for the nested nullable-struct schema; it dropped the inner `?` before this change and now round-trips unchanged. `cargo nextest run -p vortex-array -p vortex-layout -p vortex-file` passes (3270 tests, including the `partition` and struct-reader regressions); `cargo +nightly fmt --all --check` and `cargo clippy --all-targets --all-features` are clean.

---

### DIFF
--- a/vortex-array/src/expr/transform/partition.rs
+++ b/vortex-array/src/expr/transform/partition.rs
@@ -27,6 +27,8 @@ use crate::expr::traversal::NodeExt;
 use crate::expr::traversal::NodeRewriter;
 use crate::expr::traversal::Transformed;
 use crate::expr::traversal::TraversalOrder;
+use crate::scalar_fn::fns::merge::Merge;
+use crate::scalar_fn::fns::pack::Pack;
 
 /// Partition an expression into sub-expressions that are uniquely associated with an annotation.
 /// A root expression is also returned that can be used to recombine the results of the partitions
@@ -183,8 +185,11 @@ where
 
     fn visit_down(&mut self, node: Self::NodeTy) -> VortexResult<Transformed<Self::NodeTy>> {
         match self.annotations.get(&node) {
-            // If this expression only accesses a single field, then we can skip the children
-            Some(annotations) if annotations.len() == 1 => {
+            // Single-field expressions can skip their children, except reconstruction nodes
+            // (`pack`/`merge`): collapsing those applies nested-struct validity at the wrong level.
+            Some(annotations)
+                if annotations.len() == 1 && !node.is::<Pack>() && !node.is::<Merge>() =>
+            {
                 let annotation = annotations
                     .iter()
                     .next()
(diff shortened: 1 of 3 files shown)


robert3005:
This might be required but I wonder if there's a more elegant solution to this? As in can we correctly handle pack/merge/get_item validity such that we account for validity when doing expression rewrite so the splitting ends up being correct.

miniex:
You are right, this is workaround not the root fix. I think real cause is already in partition() doc note (#1907): it respects field validity but not the scope validity, so reader patches it later with is_pack_merge masking. When single-field nullable structs are chained, this masking lands at wrong depth.

The clean fix is possible because pieces are already there (Mask merged in #1900, validity() vtable, separate validity child): we can make partition() emit scope validity as own partition, and let root mask the reconstruction. But I think it is bigger than it looks. It touches partitioner, both eval paths, pack/merge validity semantics (both just return lit(true) now), and many tests that assert exact partition shape.

So maybe we land this first as safe fix for #8348, and reopen #1907 for proper refactor as follow up? I am happy to take it.

PR #8596, opened 2026-06-25
Fix flat reader subrange decode reuse

### LINKED ISSUE #8587: perf: Subsplitting large chunks causes some regression for vortex-compact for some benchmarks
Seems like #8400 causes some queries over vortex-compact significant regressions, while most others seem to be better or about the same.

The affected queries include:
- Clickbench Q2, Q3, Q7, Q9, Q10, Q11, Q19, Q25, Q29, 
- TPC-DS Q19, Q33, Q34, Q46, Q50, Q56, Q68, Q79, Q88, Q96

This regression affects both DuckDB and DataFusion, but not always both. It also improves a bunch of queries over vortex-compact, for example TPC-DS Q92.

lukekim:
Fixes #8587.

## Summary
- Memoize `FlatReader`'s decoded array future so synthetic subrange scans share the decoded flat segment instead of issuing repeated decode work.
- Add regression coverage for the query patterns that regressed after #8400, including projection-only, filter-only, filtered projection, computed projection, string filtered projection, and string filtered computed projection cases.

## Validation
- `cargo nextest run -p vortex-layout -E 'test(layouts::flat::reader)'`
- `cargo nextest run -p vortex-layout`
- `cargo clippy -p vortex-layout --all-targets --all-features`
- `git diff --check`
- `cargo bench --workspace`
- Re-ran SQL benches with `/opt/homebrew/bin/uv 0.11.24`; core suites passed: Appian, TPCH, TPCDS, ClickBench, ClickBench sorted, FineWeb, and GH Archive via direct binary rerun. PolarSignals/StatPopGen exposed pre-existing benchmark-definition/runtime backend failures, and bare Public BI requires `--opt dataset=<name>`.


robert3005:
We had this at one point and removed it https://github.com/vortex-data/vortex/pull/4459 because this effectively doubles memory usage of the reader. Any kind of caching like this has to have invalidation logic. We need memory benchmarks to flag changes like these but we have investigated this in the past

lukekim:
@robert3005 thanks for the review. Is it worth adding caching invalidation or is this not a path we should pursue?

robert3005:
Invalidation here is hard since you might come back to the segment between filter and project. Generally we couldn't figure out a good invalidation logic in the past. Overall we need to redo how tasks generation works in readers which is happening in #8518. In the mean time we can add the splitting behaviour behind a config option.

PR #8599, opened 2026-06-25
fix: support 64K alignment

a10y:
## Rationale for this change

Our `Alignment` type currently has a restriction that it fits in a `u16`. This is easy enough to satisfy most of the time.

However, while doing some testing with GH200 machines, which are ARM and which [NVIDIA recommends you setup with 64K page alignment](https://docs.nvidia.com/dccpu/grace-perf-tuning-guide/os-settings.html#page-size), I have seen some production code fail because it uses a custom allocator that passes back page-aligned memory, and attempting to construct 64K alignment is one past the maximum value so it panics.

- Remove panic if `Alignment > u16::MAX`, now you can represent 64K alignment on buffers

`Alignment` now has no restriction that it must fit into a `u16`, so some code which may have previously panicked will no longer panic.

The infallable `From<Alignment> for u16` impl has been removed entirely, there were zero callers within Vortex. This is technically a breaking change and anyone who was relying on this behavior should switch to using the exponent instead which will fit in `u8`.

### DIFF
--- a/vortex-buffer/src/alignment.rs
+++ b/vortex-buffer/src/alignment.rs
@@ -10,8 +10,7 @@ use vortex_error::vortex_err;
 
 /// The alignment of a buffer.
 ///
-/// This type is a wrapper around `usize` that ensures the alignment is a power of 2 and fits into
-/// a `u16`.
+/// This type is a wrapper around `usize` that ensures the alignment is a non-zero power of 2.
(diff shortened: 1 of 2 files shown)


robert3005 (approved):
This is one chonky alignment

a10y:
take it up with Jensen

PR #8600, opened 2026-06-25
perf(datafusion): push down `list_length` expression

mhk197:
Pushes DataFusion's `array_length(expr)`, `array_length(expr, 1)`, and `list_length(expr)` into the Vortex scan as `vortex.list.length`, computed from list offsets without materializing the element values. This is the DataFusion analogue of the DuckDB `array_length` pushdown.

## Semantics

The rewrite is exact — both functions:
- return `UInt64`,
- return `0` for a non-null empty list,
- return `null` for a null list.

The two-argument `array_length(expr, dim)` form where `dim > 1` has multidimensional semantics that `list_length` does not model, so it is explicitly rejected.

## Changes

- Add `datafusion-functions-nested` dependency (home of the `ArrayLength` UDF).

Stacked on #8495

### DIFF
--- a/vortex-datafusion/src/convert/exprs.rs
+++ b/vortex-datafusion/src/convert/exprs.rs
@@ -668,29 +668,12 @@ fn array_length_input(scalar_fn: &ScalarFunctionExpr) -> Option<&Arc<dyn Physica
     }
 }
 
-/// Returns true if `expr` is an integer literal equal to 1. The dimension argument of
-/// `array_length` is coerced to `Int64`, but we accept any integer width defensively.
+/// Returns true if `expr` is an `Int64` literal equal to 1. DataFusion coerces the `array_length`
+/// dimension argument to `Int64`, so that is the only form we need to recognize; any other literal
+/// simply isn't pushed down.
 fn is_dimension_one(expr: &Arc<dyn PhysicalExpr>) -> bool {
-    let Some(literal) = expr.downcast_ref::<df_expr::Literal>() else {
-        return false;
-    };
-
-    let dimension = match literal.value() {
-        ScalarValue::Int8(Some(v)) => i64::from(*v),
-        ScalarValue::Int16(Some(v)) => i64::from(*v),
-        ScalarValue::Int32(Some(v)) => i64::from(*v),
-        ScalarValue::Int64(Some(v)) => *v,
-        ScalarValue::UInt8(Some(v)) => i64::from(*v),
-        ScalarValue::UInt16(Some(v)) => i64::from(*v),
-        ScalarValue::UInt32(Some(v)) => i64::from(*v),
-        ScalarValue::UInt64(Some(v)) => match i64::try_from(*v) {
-            Ok(v) => v,
-            Err(_) => return false,
-        },
-        _ => return false,
-    };
-
-    dimension == 1
+    expr.downcast_ref::<df_expr::Literal>()
+        .is_some_and(|literal| matches!(literal.value(), ScalarValue::Int64(Some(1))))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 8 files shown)


vortex-datafusion/src/convert/exprs.rs
```diff
@@ -598,6 +640,59 @@ fn can_octet_length_be_pushed_down(scalar_fn: &ScalarFunctionExpr, schema: &Sche
...
+        ScalarValue::Int64(Some(v)) => *v,
+        ScalarValue::UInt8(Some(v)) => i64::from(*v),
+        ScalarValue::UInt16(Some(v)) => i64::from(*v),
+        ScalarValue::UInt32(Some(v)) => i64::from(*v),
+        ScalarValue::UInt64(Some(v)) => match i64::try_from(*v) {
+            Ok(v) => v,
+            Err(_) => return false,
+        },
+        _ => return false,
+    };
```
AdamGS: very cursed

vortex-datafusion/src/persistent/tests.rs
```diff
@@ -233,6 +236,114 @@ async fn test_octet_length_pushdown() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+async fn test_array_length_pushdown() -> anyhow::Result<()> {
```
AdamGS: I think this might nicer as an SLT test (see vortex-sqllogictest). You can also assert the plan structure to make sure the DF UDF call disappears.

vortex-datafusion/src/convert/exprs.rs
mhk197: yeah lol -- is there a better way to extract this?

AdamGS: as far as I can tell - no.

AdamGS: the alternatives are allocating another ScalarValue and casting this literal, but that seems unnecessarily wasteful.

mhk197: huh I actually think DF planner coerces this into `I64` so we just have to match on that?

https://github.com/apache/datafusion/blob/45d943dfb8699dc9cb9ef2320e955b73e3e6c03b/datafusion/functions-nested/src/length.rs#L91

https://github.com/apache/datafusion/blob/45d943dfb8699dc9cb9ef2320e955b73e3e6c03b/datafusion/expr/src/type_coercion/functions.rs#L652

vortex-sqllogictest/slt/datafusion/list_length_pushdown.slt
```diff
@@ -0,0 +1,153 @@
...
+
+# Projection-expression pushdown is off by default; enable it so the length computation folds
+# into the Vortex scan rather than being evaluated by a ProjectionExec above it.
+statement ok
+SET vortex.projection_pushdown = true;
+
+# Two list columns so we can exercise list-length pushdown over the same and different columns,
+# in both SELECT and WHERE. Row 5's `b` is an empty list (length 0).
+statement ok
+CREATE TABLE list_test (id INT, a INT[], b INT[]);
```
AdamGS: tests here have a weird flow, I think it should either be querying a file, or ideally creating an external table that's backed by Vortex.

mhk197: corrected and made structure clearer

vortex-datafusion/src/persistent/tests.rs
mhk197: done

vortex-datafusion/src/convert/exprs.rs
mhk197: Did this ^

AdamGS (approved):

PR #8603, opened 2026-06-26
Remove ArrayAccessor

robert3005:
ArrayAccessor was a poorly thought out api that leaves performance on the table.

We never use it in any hot path and only serves as a shortcut for converting
string arrays to a vec of strings. There are simpler methods for doing that in
tests


0ax1:
@claude review in depth

AdamGS (approved):

0ax1 (approved):

robert3005:
Need to fix min_max/is_sorted just noticed it while I was going through the changes

PR #8604, opened 2026-06-26
feat: add metadata bridge for Python CUDA export

0ax1:
Add a private metadata-only bridge on PyVortex arrays so the optional CUDA extension can reconstruct arrays through its own local Vortex session instead of passing Rust ArrayRef values across Python extension modules.

The CUDA extension now parses the metadata tree, calls local array plugins for deserialization, and exports bufferless arrays through the Arrow C Device capsule path. Physical buffer handoff remains a follow-up.


0ax1:
@claude review in depth

AdamGS (approved):
this makes sense to me

robert3005 (approved):

PR #8606, opened 2026-06-26
Expose array buffer replacement

gatesn:
## Rationale for this change

I want to move array buffers between devices, but keep the same array tree.

### DIFF
--- a/encodings/alp/src/alp/array.rs
+++ b/encodings/alp/src/alp/array.rs
@@ -102,6 +102,14 @@ impl VTable for ALP {
         None
     }
 
+    fn with_buffers(
+        &self,
+        array: ArrayView<'_, Self>,
+        buffers: &[BufferHandle],
+    ) -> VortexResult<ArrayParts<Self>> {
+        vortex_array::vtable::with_empty_buffers(self, array, buffers)
+    }
+
     fn serialize(
         array: ArrayView<'_, Self>,
         _session: &VortexSession,
--- a/encodings/bytebool/src/array.rs
+++ b/encodings/bytebool/src/array.rs
@@ -94,6 +94,23 @@ impl VTable for ByteBool {
         }
     }
 
+    fn with_buffers(
+        &self,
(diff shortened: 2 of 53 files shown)


vortex-array/src/array/mod.rs
```diff
@@ -363,6 +366,29 @@ impl<V: VTable> DynArrayData for ArrayData<V> {
         .into_array())
     }
 
+    fn with_buffers(&self, this: &ArrayRef, buffers: Vec<BufferHandle>) -> VortexResult<ArrayRef> {
+        let view = unsafe { ArrayView::new_unchecked(this, &self.data) };
+        let metadata = V::serialize(view, &crate::LEGACY_SESSION)?.ok_or_else(|| {
```
gatesn: This is weird.. let me fix

encodings/alp/src/alp/array.rs
```diff
@@ -102,6 +102,14 @@ impl VTable for ALP {
         None
     }
 
+    fn with_buffers(
+        &self,
+        array: ArrayView<'_, Self>,
+        buffers: &[BufferHandle],
+    ) -> VortexResult<ArrayParts<Self>> {
+        vortex_array::vtable::with_empty_buffers(self, array, buffers)
```
robert3005, line 110: don't quite get this? It's just so you can add the method and assert that it panics?

robert3005: ok no there are some implementations, I thought everything is this

encodings/bytebool/src/array.rs
```diff
@@ -94,6 +94,23 @@ impl VTable for ByteBool {
         }
     }
 
+    fn with_buffers(
```
robert3005, line 97: Shouldn't this be truly with_buffers just like with slots? Try to mutate in place and only make a copy if it fails? Then we cannot take view but typed Array<V>

robert3005 (approved):
we can add unchecked variants later

PR #8609, opened 2026-06-27
Replace ToCanonical trait usages with execute

### LINKED ISSUE #3235: Move logic into append_to_builder
Currently `append_to_builder` for canonical arrays looks something like:

```rust
    fn _append_to_builder(&self, builder: &mut dyn ArrayBuilder) -> VortexResult<()> {
        builder.extend_from_array(self)
    }
```

Where the builder presumably performs some array downcast.

Instead, the encodings should unpack themselves into the builders.

robert3005:
This also removes ArrayBuilder::extend_from_array and moves that logic
in to ArrayRef::append_to_builder

Plumb ExecutionCtx through more methods instead of creating fresh one.

Fixes: #3235 

### DIFF
--- a/encodings/runend/src/arbitrary.rs
+++ b/encodings/runend/src/arbitrary.rs
@@ -4,8 +4,9 @@
 use arbitrary::Arbitrary;
 use arbitrary::Result;
 use arbitrary::Unstructured;
+use arbitrary::unstructured::Int;
+use vortex_array::ArrayRef;
 use vortex_array::IntoArray;
-use vortex_array::VortexSessionExecute;
 use vortex_array::arrays::PrimitiveArray;
 use vortex_array::arrays::arbitrary::ArbitraryArray;
 use vortex_array::arrays::arbitrary::ArbitraryArrayConfig;
@@ -13,7 +14,7 @@ use vortex_array::arrays::arbitrary::ArbitraryWith;
 use vortex_array::dtype::DType;
 use vortex_array::dtype::Nullability;
 use vortex_array::dtype::PType;
-use vortex_array::legacy_session;
+use vortex_array::dtype::UnsignedPType;
 use vortex_array::validity::Validity;
 use vortex_buffer::Buffer;
 use vortex_error::VortexExpect;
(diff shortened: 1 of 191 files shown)


vortex-array/src/arrays/list/vtable/mod.rs
```diff
@@ -190,6 +192,17 @@ impl VTable for List {
             list_view_from_list(array, ctx)?.into_array(),
         ))
     }
+
+    fn append_to_builder(
+        array: ArrayView<'_, Self>,
+        builder: &mut dyn ArrayBuilder,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<()> {
+        let Some(builder) = builder.as_any_mut().downcast_mut::<ListBuilder<u64>>() else {
```
lwwmanning: `builder_with_capacity(DType::List, ...)` returns a `ListViewBuilder<u64, u64>`, so generic callers that allocate the canonical builder and then append a `ListArray` now hit this branch and error out instead of appending. For example, `zip`/`case_when` build the default list builder and append sliced branches through `append_to_builder`. This should either append into the canonical `ListViewBuilder` or fall back through `execute::<ListViewArray>` before appending.

encodings/runend/src/arbitrary.rs
```diff
@@ -71,12 +70,12 @@ impl ArbitraryRunEndArray {
         )?
         .0;
 
+        let len = u.int_in_range(0..=2048)?;
```
lwwmanning, line 75: `len` can now be as high as 2048, but `random_strictly_sorted_ends` still randomly chooses `PType::U8` and later does `u8::try_from(e).vortex_expect(...)` after forcing the last end toward this target. Ordinary fuzzer input can therefore panic during `Arbitrary` instead of returning `arbitrary::Error`. Please choose an end type that can represent the generated max end, cap the target by the chosen type, or return an error instead of `vortex_expect`.

vortex-array/src/arrays/list/vtable/mod.rs
robert3005: This part needs reworking. Needs to do a different downcast here

lwwmanning:
@codex please review this change

lwwmanning:
@codex review again pls

0ax1 (approved):
🚢 it

connortsui20:
Seems like this improves polarsignals benchmark by a huge amount? Is that real?

robert3005:
I think it is real. We have sometimes not delegated through VTable::append_to_builder which this pr fixes. I didn't know about this bug until we hit it elsewhere

joseph-isaacs:
please can we run benchmarks before large changes like this going forwards!

joseph-isaacs:
Do we have a fix for this yet?

robert3005:
Benchmarks aren't here for coverage

PR #8612, opened 2026-06-27
Fix wasm32 build by gating MultiFileSession on non-wasm targets

robert3005:
## Rationale for this change

The `MultiFileSession` holds a `moka` cache that reads `std::time::Instant::now()` during construction. The `Instant` type is unsupported on `wasm32` targets and panics with "time not implemented on this platform". Since multi-file scanning is not available on wasm anyway, this change gates the `MultiFileSession` registration behind a `#[cfg(not(target_arch = "wasm32"))]` guard to allow the wasm32 build to succeed.

Added a `#[cfg(not(target_arch = "wasm32"))]` attribute to the `MultiFileSession` registration in `vortex/src/lib.rs` to prevent instantiation of the session on wasm32 targets where `std::time::Instant` is not available.

No public APIs are changed. This is a build fix that prevents a panic on wasm32 targets. Multi-file scanning functionality remains unavailable on wasm, as it was before.

https://claude.ai/code/session_01P5VUBo8DhakxhhF1Ux2Kte

### DIFF
--- a/vortex/src/lib.rs
+++ b/vortex/src/lib.rs
@@ -303,6 +303,11 @@ impl VortexSessionDefault for VortexSession {
 
         #[cfg(feature = "files")]
         let session = {
+            // `MultiFileSession` holds a `moka` cache whose clock reads `std::time::Instant::now()`
+            // when constructed. `Instant` is unsupported on `wasm32` and panics with "time not
+            // implemented on this platform". Multi-file scanning is not available on wasm anyway, so
+            // only register this session variable on non-wasm targets.
+            #[cfg(not(target_arch = "wasm32"))]
             let session = session.with::<file::multi::MultiFileSession>();
             file::register_default_encodings(&session);
             session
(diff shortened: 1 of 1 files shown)


vortex/src/lib.rs
```diff
@@ -303,6 +303,11 @@ impl VortexSessionDefault for VortexSession {
 
         #[cfg(feature = "files")]
         let session = {
+            // `MultiFileSession` holds a `moka` cache whose clock reads `std::time::Instant::now()`
+            // when constructed. `Instant` is unsupported on `wasm32` and panics with "time not
+            // implemented on this platform". Multi-file scanning is not available on wasm anyway, so
+            // only register this session variable on non-wasm targets.
+            #[cfg(not(target_arch = "wasm32"))]
```
lwwmanning: nit: you could just apply this to the whole block (since it's now a no-op), same spot as `#[cfg(feature = "files")]`

lwwmanning (approved):
approving but I disabled auto-merge, up to you whether you want to address my nit

PR #8613, opened 2026-06-27
Polish treemap view in vortex explorer

robert3005:
Polish treemap view in vortex explorer

treemap-field
treemap-heavy-chunks
treemap-root


0ax1 (commented):
No preview deploy?

0ax1 (commented):
codex:
```
• Findings

  - Medium: vortex-web/src/components/explorer/BlockTreemap.tsx:158 leaves selfSelected.current stale after a treemap-originated selection. After clicking tile A, selecting another node externally, then selecting A externally again, the effect treats
    A as self-originated and refuses to re-root the treemap. The marker set at vortex-web/src/components/explorer/BlockTreemap.tsx:268 should be one-shot: clear it when it matches, and also clear it before handling a non-matching external selection.
```

0ax1:
@claude in depth review

0ax1 (approved):

robert3005:
Addressed the review in `5e5d5f9`:

- **🔴 stale `selfSelected` latch** — it's now a one-shot marker, cleared the moment it's consumed (both on the self-match branch and before handling an external selection). Re-selecting a previously map-selected node from the tree panel now re-roots the zoom instead of being suppressed.
- **Duplicated `arraySubtreeBytes`** — exported from `physicalStats.ts` and imported in `BlockTreemap`; the copy is gone.
- **Per-mousemove stats recompute** — `nodePhysicalStats` now takes the prebuilt `segmentMap`, and `TileTooltip` memoizes its stats by `node`, so they aren't recomputed on every pointer move (matters for the `HeavyChunks` story).
- **Area vs `% of file`** — added a comment on `subtreeBytes` noting tile area is data-only bytes while the tooltip's `% of file` includes metadata, so the two can differ slightly.

Ran `eslint` + `tsc` (clean) and a Storybook build of the three stories.

---

Issue #8380, opened 2026-06-11
Vortex introduced a breaking semantic change to ScalarFnId::new which did not surface as compilation errors in consumers


danking:
### What happened?

Previously we wrote `ScalarFnId::new` and this was reasonable quick, even if invoked across tens of threads.

After a recent Vortex change, this started grabbing locks which substantially reduced the rate at which this method could be called in a multi-threaded environment.
```rust
     type Options = BlobBytesOptions;
 
     fn id(&self) -> ScalarFnId {
-        ScalarFnId::new("spiral.blob.bytes")
+        static ID: CachedId = CachedId::new("spiral.blob.bytes");
+        *ID
     }
 ```

I would have preferred a compilation error. Even better, could `ScalarFnId::new` be implemented as above?

### Steps to reproduce

see above

### Environment

macOS

vortex: 1b19ac90c01f32da7bbfc895dd4f7d6a7bbc4118

myrrc:
I suspect this was during my work on random access improvements. ScalarFnId::new is slow enough for random access to be a blocker since it calls a Registry which in turn inserts into a Dashmap. Maybe we should find a faster way to intern/compare.

gatesn:
This should not be called `new`, that would have made the break obvious.

robert3005:
FWIW if we want to introduce semantic breaks we should make them come with syntactic break so users know they have to make a change

PR #8617, opened 2026-06-29
chore: forbid the locking Id constructors with a clippy lint

miniex:
## Summary

Follow-up to #8614. A `clippy` disallowed-methods lint now forbids `Id::new` and `Id::new_static` workspace-wide, since both intern into a globally locked table (`ThreadedRodeo`) on every call. Static identifiers must use a `CachedId` static instead; the sites that intern a genuinely dynamic string (proto/flatbuffer deserialization, arrow extension lookups, Python read contexts, tests) keep `Id::new` behind an `#[expect(clippy::disallowed_methods, reason = "...")]`. `CachedId` itself now interns through `Id::new_static`, so it is the single sanctioned path for static ids.

This is the enforcement half of #8380; #8614 did the migration. It is stacked on #8614, so it stays a draft until that merges; the shared migration commit drops out once this is rebased onto a merged #8614.

Refs: #8380

## Testing

`cargo clippy --all-targets` is clean across the edited crates (`vortex-array`, `vortex-layout`, `vortex-file`, `vortex-ipc`, `vortex-zstd`, `vortex-python`, `vortex-duckdb`, `vortex-session`), with `disallowed_methods` and `unfulfilled_lint_expectations` both at zero; `cargo +nightly fmt --all --check` clean. `vortex-cuda` could not be built locally (no CUDA toolchain).

---

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -15,4 +15,6 @@ disallowed-methods = [
     { path = "itertools::Itertools::counts", reason = "It uses the default hasher which is slow for primitives. Just inline the loop for better performance.", allow-invalid = true },
     { path = "std::result::Result::and", reason = "This method is a footgun, especially when working with `Result<Validity>`." },
     { path = "std::thread::available_parallelism", reason = "This function might do an unbounded amount of work, use `vortex_utils::parallelism::get_available_parallelism instead" },
+    { path = "vortex_session::registry::Id::new", reason = "Interning a static id on every call grabs the interner lock (#8380). Use a `CachedId` static for static ids; for a dynamic string, annotate the call with `#[expect(clippy::disallowed_methods)]`.", allow-invalid = true },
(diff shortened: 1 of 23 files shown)


robert3005 (approved):

robert3005:
I think you need to rebase and fix

```
error: use of a disallowed method `vortex_session::registry::Id::new`
   --> vortex-python-cuda/src/lib.rs:191:23
    |
191 |     let encoding_id = ArrayId::new(&metadata.encoding_id);
    |                       ^^^^^^^^^^^^
    |
    = note: Interning a static id on every call grabs the interner lock (#8380). Use a `CachedId` static for static ids; for a dynamic string, annotate the call with `#[expect(clippy::disallowed_methods)]`.
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#disallowed_methods
    = note: `-D clippy::disallowed-methods` implied by `-D clippy::all`
    = help: to override `-D clippy::all` add `#[allow(clippy::disallowed_methods)]`
```

Issue #8570, opened 2026-06-24
Casting


gatesn:
# Add Bound DType Cast Functions and Extension Cast Hooks

## Motivation

`Cast` currently handles canonical casts with hard-coded dispatch, and `ExtensionArray` only supports same-extension/nullability casts or unwrapping to storage. This makes metadata-sensitive casts like `timestamp(ms) -> timestamp(ns)` hard to express, even though the extension dtype metadata contains the information needed to decide whether the cast is valid.

## Proposal

Keep casts represented as `ScalarFnArray(Cast, target_dtype, child)` so existing pushdown and `ExecuteParent` behavior continues to work.

Refactor cast execution around a type-erased cast function:

```rust
pub type CastFn =
    Arc<dyn Fn(ArrayRef, &mut ExecutionCtx) -> VortexResult<ArrayRef> + Send + Sync>;
```

The returned function is already bound to the concrete source and target dtypes, so it can capture the target dtype and any extension metadata it needs. `CastFn` is only the callable execution body; binding remains owned by the `Cast` scalar function.

## Design

Add a dtype-level binder as an inherent method on `Cast`, used by `Cast::execute`:

```rust
impl Cast {
    pub(crate) fn bind_dtype_cast(
        source: &DType,
        target: &DType,
    ) -> VortexResult<Option<CastFn>>;
}
```

Binding order:

1. Exact dtype: no-op cast.
2. Same dtype ignoring nullability: nullability cast.
3. Source extension hook: `source_ext.cast_to(target)`.
4. Target extension hook: `target_ext.cast_from(source)`.
5. Built-in non-extension casts: primitive, decimal, bool, varbin, list, struct, etc.

Add optional extension hooks:

```rust
fn cast_to(ext_dtype: &ExtDType<Self>, target: &DType) -> VortexResult<Option<CastFn>> {
    Ok(None)
}

fn cast_from(ext_dtype: &ExtDType<Self>, source: &DType) -> VortexResult<Option<CastFn>> {
    Ok(None)
}
```

These hooks receive the concrete `ExtDType`, so implementations can inspect metadata.

## Extension Behavior

- `ext -> storage/non-ext`: source-side hook may unwrap storage and recursively bind storage-to-target.
- `storage/non-ext -> ext`: no universal default; extension must opt in because storage dtype alone may not prove semantic validity.
- `ext -> ext`: source-side hook handles metadata-sensitive casts, for example timestamp unit scaling when timezone metadata matches.

[...]

PR #8620, opened 2026-06-29
Cast pushdown for duckdb

### LINKED ISSUE #3963: Support table_function_type_pushdown in DuckDB extension
This allows for projection casting to be pushed into the Vortex scan

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

myrrc:
- Backport of https://github.com/duckdb/duckdb/pull/22788 to push down casts, optimizer pass. Extract both passes common parts to optimizer.cpp
- Map 8- and 16-bit types in sqllogictest runner to INTEGER.
- Add tracing subscriber logging to sqllogictest runner to support RUST_LOG=debug.
- Support of CAST (not TRY_CAST) over primitives pushdown on Rust side. U128/I128 are not pushed.

We are overly restrictive here because we don't want clients to face issues like `No CastReduce to cast constant array from vortex.date[days](i32) to vortex.timestamp[µs](i64?)`. We can lift the restriction later.


0ax1:
What about constant+bool arrays?

vortex-array/src/arrays/bool/compute/cast.rs
```diff
@@ -102,4 +124,22 @@ mod tests {
...
+    #[case(crate::dtype::PType::F32)]
+    #[case(crate::dtype::PType::F64)]
+    fn cast_bool_to_primitive(#[case] target: crate::dtype::PType) {
+        let mut ctx = SESSION.create_execution_ctx();
+        let arr = BoolArray::from_iter(vec![true, false, true]).into_array();
+        let out = arr
+            .cast(DType::Primitive(target, Nullability::NonNullable))
+            .unwrap();
+        let out = out.execute::<Canonical>(&mut ctx).unwrap().into_array();
+        assert_eq!(out.len(), 3);
```
0ax1, line 144: We only assert the length here not the type.

0ax1:
@claude in depth review

vortex-duckdb/cpp/include/cast_pushdown.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include "duckdb/planner/logical_operator.hpp"
+
+using namespace duckdb;
+
+/**
+ * Collect CAST(col) expressions. If "col" is used without CAST in "plan",
+ * record in "analyses.conflicts"
+ */
+struct CastCollect final : LogicalOperatorVisitor {
+    Analyses &analyses;
```
0ax1, line 18: What is the ownership model here? Can we access a stale ref? Same for all other refs and pointers.

vortex-duckdb/cpp/include/cast_pushdown.hpp
```diff
@@ -0,0 +1,43 @@
...
+    ExpressionPtr VisitReplace(BoundCastExpression &expr, ExpressionPtr *ptr) override;
+};
+
+/*
+ * For "col" in columns collected by CastCollect, replace CAST(col) to "col"
+ * if "col" doesn't have conflicting usage. Update return types for bound
+ * columns and logical projections referencing this column.
+ */
+struct CastReplace final : LogicalOperatorVisitor {
+    Analyses &analyses;
```
0ax1, line 37: -//-

vortex-duckdb/cpp/include/optimizer.hpp
```diff
@@ -0,0 +1,137 @@
...
+// A passthrough projection only forwards its child columns, e.g. a VIEW's
+// "SELECT col".
+bool IsPassthrough(const LogicalProjection &projection);
+
+// There are no conflicting column usages in the plan
+bool CanPushdownColumn(const GetAnalysis &analysis, TableColumnScanIndex idx);
+
+template <class Collect, class Replace>
+LogicalOperatorPtr TryPushdown(ClientContext &context, LogicalOperatorPtr plan) {
+    Analyses analyses;
```
0ax1, line 107: No init for `analyses` and `projections`? We should add a comment that they are output params on `FindGetsAndProjections` in any case.

vortex-duckdb/cpp/include/optimizer.hpp
```diff
@@ -0,0 +1,137 @@
...
+    Projections projections;
+    FindGetsAndProjections(*plan, analyses, projections);
```
0ax1, line 109: Probably hard to do in C++ but it would be much nicer to return from `FindGetsAndProjections` rather than using output params.

vortex-duckdb/cpp/include/optimizer.hpp
```diff
@@ -0,0 +1,137 @@
...
+    if (analyses.empty()) {
+        return plan;
+    }
+    Collect(analyses, projections).VisitOperator(*plan);
+
+    bool any_pushed = false;
+    for (auto &[_, analysis] : analyses) {
```
0ax1, line 116: Would Sean Parent approve this loop? :)

vortex-duckdb/cpp/cast_pushdown.cpp
```diff
@@ -0,0 +1,171 @@
...
+    LogicalOperatorVisitor::VisitOperator(op);
+}
+
+ExpressionPtr CastCollect::VisitReplace(BoundColumnRefExpression &expr, ExpressionPtr *ptr) {
+    if (const auto binding = Resolve(expr.binding, analyses, projections)) {
+        // Column is used without cast applied to it, register a conflict.
+        // Not emplace() as we need to update the value if it was present
+        binding->analysis.col_to_expr[binding->column_index] = nullptr;
+    }
+    return std::move(*ptr);
```
0ax1, line 92: why move?

0ax1 (commented):
first pass

vortex-duckdb/cpp/include/cast_pushdown.hpp
myrrc: No, the analysis lives on the stack of pushdown function, and the visitor's lifetime is less than lifetime of the function.

For pointers, we reference the query plan which will outlive the optimizer, and we either read (Collect) or modify (Replace) the pointers. Optimizers run sequentially so there's no risk the pointers will be invalidated.

vortex-duckdb/cpp/include/optimizer.hpp
myrrc: These are default-initialized. I can add = {}; but that would be same semantically. On the out parameters note, I think this is trivially visible when looking at the code, since FindGets* doesn't output anything else but these two structs are used later.

vortex-duckdb/cpp/include/cast_pushdown.hpp
0ax1: Can we drop a doc line concisely what the ownership model is?

vortex-duckdb/cpp/include/optimizer.hpp
myrrc: I'm not sure this is the case for C++ but rather recursive/iterative algorithms. If we return from a function, we need to combine results if this function was recursively called, and this triggers a fresh object reallocation every time.

vortex-duckdb/cpp/include/optimizer.hpp
0ax1: Rumors say there's that one lang that guarantees that memory is initialized 😛 .

vortex-duckdb/cpp/cast_pushdown.cpp
myrrc: Because unique_ptr doesn't have a copy constructor

0ax1:
Whats your take on: `float→int casts diverge (truncate vs. round)`

myrrc:
That's a good find, I'll open a PR to Vortex to match the semantics duckdb does since it's what other DBMS do as well. Same for tests, worth adding. I'll rebase once I'm done

vortex-duckdb/cpp/include/optimizer.hpp
0ax1: Yeh

vortex-duckdb/cpp/include/cast_pushdown.hpp
myrrc: +1

myrrc:
I've thought about this for a while, and no, we don't want float cast pushdown because this would be a breaking change for people using Vortex. Let's set the behaviour as it is today.

vortex-duckdb/src/convert/expr.rs
```diff
@@ -189,6 +190,17 @@ fn is_supported_length_alias(func: &BoundFunction) -> bool {
     children.len() == 1 && returns_a_list(children[0])
 }
 
+// We limit casting to Primitive types, because some conversions yield an error
```
0ax1, line 331: Is this something we should fix in the future and leave a TODO? What does Parquet do?

0ax1:
Fable review: 

# Code Review — Cast pushdown for DuckDB

## Verdict

The commit carries **three confirmed correctness defects** in the new C++ optimizer pass, each with a concrete failing query. The new SLT suite is blind in exactly the plan shapes where these bugs live (views/passthrough projections, scan-index ≠ storage-index columns, deleted plan-level negative checks). The remainder are test-coverage gaps, copy-paste duplication between the cast and scalar-function passes, and doc/API-shape cleanups.

---

## Correctness

### 1. `CastCollect` conflict detection is visit-order dependent — CONFIRMED

`vortex-duckdb/cpp/cast_pushdown.cpp:103` (same root cause at `:101`, `:151`)

A cast over a column that is *not* a top-level projection cast records neither a candidate nor a conflict when no `col_to_expr` entry exists yet. Because `LogicalOperatorVisitor::VisitOperator` visits children (residual WHERE filters) **before** the operator's own expressions (duckdb-1.5.3 `src/planner/logical_operator_visitor.cpp:9-12`), a conflicting cast/`TRY_CAST` in a filter below the projection is visited first and silently ignored. `CastReplace::VisitReplace(BoundCastExpression&)` (line 139) then rewrites that filter cast into a bare column ref **without checking its target type or `try_cast` flag**.

**Failure scenario.** Data file with `column01 INTEGER` values `{0,…,0,257}`:

```sql
SELECT column01::UTINYINT FROM 'f.vortex'
WHERE TRY_CAST(column01 AS UTINYINT) IS NOT NULL;
```

The `TRY_CAST` filter stays a `LOGICAL_FILTER` (Rust `can_push_cast` rejects try-casts) and is visited before the projection cast registers its candidate, so the try-cast mismatch check (lines 106–109) never runs. The non-try `CAST` is pushed into the scan and the filter's `TRY_CAST` is replaced by a bare column ref, turning the filter into `column01 IS NOT NULL`. The scan-level cast then raises `Conversion Error: 257 out of range for UTINYINT` — failing a query that previously returned 9 rows. With differing target types in SELECT vs WHERE, the rewritten filter instead **silently returns wrong rows**. The new `cast_pushdown.slt` only tests the same-operator ordering (SELECT cast and TRY_CAST side by side) where expression order masks the hole.

### 2. `projection->types` written at the wrong index — CONFIRMED

`vortex-duckdb/cpp/cast_pushdown.cpp:132` (same pattern at `:160`; carried into the refactored `scalar_fn_pushdown.cpp:85`, `:113`)

`CastReplace` writes `projection->types[column_index]` using the **GET's scan index** (`GetBinding::column_index` from `Resolve`) instead of the passthrough projection's **output position** — an index-space mixup. `Resolve()` in `optimizer.cpp` returns `get_binding.column_index` (index into the GET's column ids), discarding the projection output slot it came from.

**Failure scenario.** A passthrough projection whose colrefs are not the identity mapping, e.g.

```sql
CREATE VIEW v AS SELECT b AS x, a AS y FROM t WHERE a > 0;  -- over a vortex file
SELECT x::BIGINT, y::SMALLINT FROM v;
```

The filter becomes a table filter, leaving `PROJECTION[colref(1), colref(0)] -> GET`, which registers as a passthrough. Both casts push down and the pass writes `projection->types[1] = BIGINT` for output 0 and `projection->types[0] = SMALLINT` for output 1 — the two slots swapped relative to the retyped expressions. With a single-output passthrough over a wider scan (`types.size() == 1`, cast column at scan index 1), the write is **out of bounds**, and DuckDB's bounds-checked `vector::operator[]` aborts an otherwise valid query with `INTERNAL Error: Attempted to access index 1 within vector of size 1`.

### 3. Error-raising casts pushed below residual filters — CONFIRMED

`vortex-duckdb/cpp/cast_pushdown.cpp:21`

`ReachesPushdownGet` allows `LOGICAL_FILTER` in the projection-to-GET chain, so a non-try cast that can raise conversion errors is pushed into the scan and evaluated on **every scanned row**, including rows that an unpushable residual filter would have excluded before the projection cast executed in the original plan.

**Failure scenario.** Data file with rows `(n=1, s='a')`, `(n=257, s='b')`:

```sql
SELECT n::UTINYINT FROM 'f.vortex' WHERE s || 'x' = 'ax';
```

The `||` filter is not pushable and remains a `LOGICAL_FILTER` (only `s` conflicts; `n` is used solely in the top-level cast), so `CAST(n AS UTINYINT)` is pushed into the vortex scan below the filter. The scan casts `n = 257` and raises a conversion error, while the original plan filters that row out first and returns `1`. A previously-succeeding query now errors.

0ax1 (approved):

vortex-duckdb/src/convert/expr.rs
myrrc: #8570

myrrc:
I've fixed the bugs by not allowing the FILTER in single-get chain search. This would likely need changes in upstream pass though, we now want to run after filter_pushdown optimizer and inspect table filters

PR #8621, opened 2026-06-29
bool -> Primitive casts

myrrc:
Support of bool -> Primitive casts.
This is a subtask for pushing expressions of form
"WHERE prefix(...) > 0" to Duckdb.

### DIFF
--- a/vortex-array/src/arrays/bool/compute/cast.rs
+++ b/vortex-array/src/arrays/bool/compute/cast.rs
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
+use num_traits::One;
+use num_traits::Zero;
+use vortex_buffer::BufferMut;
 use vortex_error::VortexResult;
 
 use crate::ArrayRef;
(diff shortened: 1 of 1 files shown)


vortex-array/src/arrays/bool/compute/cast.rs
```diff
@@ -102,4 +124,22 @@ mod tests {
     fn test_cast_bool_conformance(#[case] array: BoolArray) {
         test_cast_conformance(&array.into_array());
     }
+
+    #[rstest]
+    #[case(crate::dtype::PType::I8)]
```
0ax1: rule of thumb: never use `crate::` always import mods such that `crate::` isn't needed.

0ax1 (approved):
Looks good, one nit.

vortex-array/src/arrays/bool/compute/cast.rs
myrrc: +1

vortex-array/src/arrays/bool/compute/cast.rs
```diff
@@ -38,17 +43,34 @@ impl CastKernel for Bool {
...
-        ))
+                .cast_nullability(*new_nullability, array.len(), ctx)?;
+
+        let bits = array.to_bit_buffer();
+        let len = bits.len();
+
+        Ok(Some(match_each_native_ptype!(*new_ptype, |T| {
+            let (one, zero) = (<T as One>::one(), <T as Zero>::zero());
+            let mut buffer = BufferMut::<T>::with_capacity(len);
+            buffer.extend(bits.iter().map(|v| if v { one } else { zero }));
```
robert3005, line 71: Buffer::from_trusted_len_iter

robert3005:
I think these need to be separate

PR #8622, opened 2026-06-29
fix: CUDA bitpacked sliced output allocation

0ax1:
Decode sliced bit-packed arrays in padded coordinates by sizing and launching for offset + len. This keeps the returned offset..offset+len device slice in bounds and ensures the final touched 1024-value chunk is decoded.

### DIFF
--- a/vortex-cuda/src/kernel/encodings/bitpacked.rs
+++ b/vortex-cuda/src/kernel/encodings/bitpacked.rs
@@ -177,14 +177,16 @@ where
     // Get CUDA view of input
     let input_view = device_input.cuda_view::<A::Physical>()?;
 
+    let output_len = offset + len;
+
     // Allocate output buffer
-    let output_slice = ctx.device_alloc::<A>(len.next_multiple_of(1024))?;
+    let output_slice = ctx.device_alloc::<A>(output_len.next_multiple_of(1024))?;
     let output_buf = CudaDeviceBuffer::new(output_slice);
     let output_view = output_buf.as_view::<A>();
 
     let output_width = size_of::<A>() * 8;
     let cuda_function = bitpacked_cuda_kernel(bit_width, output_width, ctx)?;
-    let config = bitpacked_cuda_launch_config(output_width, len)?;
+    let config = bitpacked_cuda_launch_config(output_width, output_len)?;
 
     // We hold this here to keep the device buffers alive.
     let device_patches = if let Some(patches) = patches {
@@ -199,7 +201,7 @@ where
 
     let patches_arg = build_gpu_patches(device_patches.as_ref())?;
 
-    ctx.launch_kernel_config(&cuda_function, config, len, |args| {
+    ctx.launch_kernel_config(&cuda_function, config, output_len, |args| {
         args.arg(&input_view)
             .arg(&output_view)
             .arg(&reference)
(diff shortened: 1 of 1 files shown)


0ax1:
report: 
```
# GPU bit-packed decode overruns its output buffer for sliced arrays

**Component:** `vortex-cuda` — `decode_bitpacked` (`vortex-cuda/src/kernel/encodings/bitpacked.rs`)
**Affects:** `develop` · **Severity:** panic during GPU scan

## Problem

GPU-decoding a **sliced** bit-packed array (non-zero `offset`) overruns the output device buffer:

```rust
// allocation ignores `offset`
let output_slice = ctx.device_alloc::<A>(len.next_multiple_of(1024))?;
...
// slice uses `offset`
output_buf.slice_typed::<A>(offset..(offset + len))
```

A sliced bit-packed array carries a non-zero `offset` (FastLanes records the slice start within the
first 1024-block). The allocation is `len.next_multiple_of(1024)` but the slice end is `offset + len`,
so when `offset > len.next_multiple_of(1024) - len` the slice runs past the allocation and panics at
`device_buffer.rs:320`: `Slice range end {offset+len} exceeds allocation size {…}`.

**Likely fix:** allocate `(offset + len).next_multiple_of(1024)` (and confirm the kernel writes into
the same offset-based window).

## Reproduction

Add to the `tests` module of `bitpacked.rs`, then run on a GPU box:
`cargo nextest run -p vortex-cuda --all-features -E 'test(test_sliced_offset_overruns_output)'`

```rust
#[crate::test]
fn test_sliced_offset_overruns_output() -> VortexResult<()> {

    let mut ctx = vortex_array::array_session().create_execution_ctx();
    let mut cuda_ctx = CudaSession::create_execution_ctx(&crate::cuda_session())
        .vortex_expect("failed to create execution context");

    // 2048 values (two 1024-blocks); all < 64 so they fit in 6 bits (no patches).
    let array = PrimitiveArray::new((0u32..64).cycle().take(2048).collect::<Buffer<_>>(), NonNullable);
    let bp = BitPacked::encode(&array.into_array(), 6, &mut ctx)?;

    // Slice to a 1024-long window at offset 1 -> offset = 1, len = 1024.
    // Decoder allocates 1024 but slices 1..1025.
    let sliced = bp.into_array().slice(1..1025)?;

    let gpu_result = block_on(async {
        sliced.clone().execute_cuda(&mut cuda_ctx).await
            .vortex_expect("GPU decompression failed")
            .into_host().await.map(|a| a.into_array())
    })?;

    assert_arrays_eq!(sliced, gpu_result, &mut ctx);
    Ok(())
}
```

**Output:**
```
panicked at vortex-cuda/src/device_buffer.rs:320:9:
Slice range end 4100 exceeds allocation size 4096
```
(`4100 = (1 + 1024) × 4 bytes`; the overrun equals the slice offset.)

## Impact

Dictionary columns slice their bit-packed `codes` child, producing these non-zero-offset arrays, so
any GPU scan decoding such a column hits this. Found scanning the ClickBench `Title` column on a
GH200 (`offset = 382`, `len = 106496`).
```

0ax1:
Thanks for the heads up on this one: @gargiulofrancesco !

robert3005 (approved):

PR #8625, opened 2026-06-29
feat: support `ST_DWithin` pushdown in vortex

HarukiMoriarty:
## Summary

Insert non-throwing geo predicate `vortex_dwithin` in DuckDB, which later pushdown into vortex, call `distance` scalar function on scanning, significantly improve Q1/Q3 performance in SpatialBench.

1. adding new geo predicate `vortex_dwithin` in DuckDB, which is non-throwing and can be pushdown.
2. adding SQL rewrite so for geo native type, `ST_dwithin` is text rewritten into `vortex_dwithin`.

## Performance

```
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┓
┃ Query ┃ duckdb:parquet (base) ┃   duckdb:vortex ┃ duckdb:vortex-native ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩
│ 1     │               179.6ms │ 132.1ms (0.74x) │       29.0ms (0.16x) │
│ 2     │               268.1ms │ 255.3ms (0.95x) │      312.5ms (1.17x) │
│ 3     │               247.8ms │ 216.4ms (0.87x) │      140.2ms (0.57x) │
│ 4     │               199.6ms │ 146.9ms (0.74x) │      144.5ms (0.72x) │
│ 5     │                 3.40s │   3.43s (1.01x) │        3.03s (0.89x) │
│ 6     │               528.1ms │ 359.2ms (0.68x) │      455.4ms (0.86x) │
│ 7     │               984.0ms │ 983.2ms (1.00x) │      887.7ms (0.90x) │
│ 8     │                 1.08s │ 945.4ms (0.87x) │      997.6ms (0.92x) │
│ 9     │                34.2ms │  33.5ms (0.98x) │       43.6ms (1.27x) │
└───────┴───────────────────────┴─────────────────┴──────────────────────┘
```

Takeaways: Q1 and Q3 is significantly improved due to the single table geo predicate is pushdown into vortex scan. Q1 is more benefit from the manually fast path without going through `geo` crate for calculation. Q3 is also potentially can be benefit from manually calculation.


vortex-duckdb/cpp/expr.cpp
```diff
@@ -21,6 +30,40 @@ extern "C" const char *duckdb_vx_sfunc_name(duckdb_vx_sfunc ffi_func) {
...
+                // (ST_DWithin's bind folds it into bind_data). vortex_dwithin is only pushed, never run.
+                copy.bind = nullptr;
+                set.AddFunction(copy);
+            }
+            CreateScalarFunctionInfo info(std::move(set));
+            info.on_conflict = OnCreateConflict::IGNORE_ON_CONFLICT;
+            catalog.CreateFunction(context, info);
+        });
+    } catch (const std::exception &) {
+        // No `spatial` loaded → no `ST_DWithin` to alias; nothing to register.
```
myrrc: I don't think we should include non-ascii characters like arrows in our comments

vortex-bench/src/spatialbench/benchmark.rs
```diff
@@ -78,12 +78,12 @@ impl Benchmark for SpatialBenchBenchmark {
...
-    /// On the `vortex-native` lane, geometry columns surface as `GEOMETRY`, so drop the
-    /// `ST_GeomFromWKB(..)` wrappers and let DuckDB's `spatial` extension evaluate the `ST_*`
-    /// predicates directly on the native geometry.
+    /// Adapt a query to the storage format. The `vortex-native` lane surfaces geometry as `GEOMETRY`,
+    /// so it drops the `ST_GeomFromWKB(..)` wrappers and routes pushable `ST_DWithin` filters.
     fn query_for_format(&self, query: &str, format: Format) -> String {
         match format {
-            Format::VortexNative => strip_wkb_wrappers(query),
+            // Native geometry is `GEOMETRY`: drop `ST_GeomFromWKB(..)`, route pushable `ST_DWithin`.
+            Format::VortexNative => route_pushable_dwithin(&strip_wkb_wrappers(query)),
```
myrrc: I think CREATE MACRO in init.sql is an easier approach than rewriting the query manually.

vortex-duckdb/cpp/expr.cpp
```diff
@@ -21,6 +30,40 @@ extern "C" const char *duckdb_vx_sfunc_name(duckdb_vx_sfunc ffi_func) {
     return func->name.c_str();
 }
 
+extern "C" duckdb_state duckdb_vx_register_geo_aliases(duckdb_database ffi_db) {
+    if (!ffi_db) {
+        return DuckDBError;
+    }
+    const DatabaseWrapper &wrapper = *reinterpret_cast<DatabaseWrapper *>(ffi_db);
+    try {
+        Connection conn(*wrapper.database->instance);
```
myrrc: Why do we need a transaction here? Can we register the function in the global catalog without it?

vortex-duckdb/src/duckdb/database.rs
```diff
@@ -90,4 +90,14 @@ impl DatabaseRef {
         );
         Ok(())
     }
+
+    /// Register `vortex_dwithin`, a non-throwing alias of the `spatial` extension's `ST_DWithin`, so
+    /// the radius predicate pushes into the Vortex scan.
+    pub fn register_geo_aliases(&self) -> VortexResult<()> {
```
myrrc: Nit: function's name doesn't link to function's comment.

vortex-duckdb/cpp/expr.cpp
HarukiMoriarty: It looks like we cannot not. DuckDB's catalog is MVCC, so `CreateFunction` needs an active transaction (it throws `ActiveTransaction called without active transaction` otherwise). A fresh `Connection` only begins a transaction per query, and we're calling the catalog API directly, so `RunFunctionInTransaction` is just the "begin → run → commit" wrapper to supply one. 
I confirmed by removing it: registration aborts and queries then fail with `vortex_dwithin does not exist`.

vortex-bench/src/spatialbench/benchmark.rs
HarukiMoriarty: Here the queries call `ST_GeomFromWKB`/`ST_DWithin`, which the spatial extension already defines, and you can't shadow those: `CREATE MACRO ST_DWithin(...)` -> `Catalog Error: Macro Function with name "ST_DWithin" already exists` (same for `ST_GeomFromWKB`, both tested).

myrrc: But this means we won't be able to use this function on real queries since you don't have access to the query string in Duckdb. Let's think of another way. Maybe we can delete the spatial function and then create ours? This won't survive extension re-LOAD, but is a temporary good solution.
Otherwise we can write a small pass replacing ST_DWithin with vortex_DWIthin in the graph, but doing nothing else, this should be a single Visitor override.

HarukiMoriarty: It is actually messy than what I thought. Tried replacing `ST_DWithin` in the catalog with our own bind-cleared copy. Single-table filters push (Q1 ~8.6×, Q3 ~4.9×), but it regresses the join queries.
DuckDB's `ST_DWithin`'s bind erases the constant radius into `bind_data`, leaving a 2-arg predicate, which is exactly what spatial's spatial-join optimizer requires (first gate: `func.children.size() != 2`). Clearing the bind to expose the radius gives 3 args, trips that gate, and Q8 drops from `SPATIAL_JOIN` to `BLOCKWISE_NL_JOIN`. Confirmed via EXPLAIN (override off → `SPATIAL_JOIN`, on → `BLOCKWISE_NL_JOIN`).
So the radius-erase is what we need gone to push but spatial needs present to optimize joins, a global replacement can't have both.

HarukiMoriarty: Working on SpatialBench infra itself so we never need `GeomFromWKB` anymore, though selective of rewritten or overwritten or Macro of `ST_DWithin` is still needed. Or maybe we just rename it as `ST_Distance <= xxx`?

vortex-duckdb/cpp/include/expr.h
```diff
@@ -13,6 +13,13 @@ typedef struct duckdb_vx_sfunc_ *duckdb_vx_sfunc;
 
 const char *duckdb_vx_sfunc_name(duckdb_vx_sfunc ffi_func);
 
+/// Shadow the spatial extension's `ST_DWithin` in the user catalog with a copy whose bind keeps
+/// the radius as the third argument (spatial's bind folds it into private bind data), so radius
+/// predicates can push into Vortex scans. Unpushed occurrences still execute via spatial's own
+/// 3-column code path, and the Vortex optimizer pass rebinds them (e.g. join conditions) back to
+/// spatial's folded form, which its spatial-join optimization requires. No-op without `spatial`.
+duckdb_state duckdb_vx_register_st_dwithin_override(duckdb_database ffi_db);
```
myrrc, line 19: Can we make this comment more terse? I understand it's clauded, but let's keep it readable for humans

vortex-duckdb/cpp/vortex_duckdb.cpp
```diff
@@ -269,6 +269,9 @@ extern "C" duckdb_blob duckdb_vx_value_get_geometry(duckdb_value value) {
 
 static void VortexOptimizeFunction(OptimizerExtensionInput &input, unique_ptr<LogicalOperator> &plan) {
     plan = TryPushdownScalarFunctions(input.context, std::move(plan));
+    // Runs before spatial's own optimizer pass: this extension registers at database open,
```
myrrc: Do we guarantee it runs before extension's optimizer pass? Do we have a total order on that? If not, I'd suggest using a pre_optimize_function in vortex extension (second parameter to OptimizerExtension) and put this pass there.

vortex-duckdb/cpp/scalar_fn_pushdown.cpp
```diff
@@ -231,3 +234,47 @@ ScalarFnReplace::ScalarFnReplace(Analyses &analyses, const Projections &projecti
...
+
+// See RestoreStDWithin: rebinding through spatial's own entry lets spatial build its own bind
+// data, so nothing here depends on its internals.
+class StDWithinRestore final : public LogicalOperatorVisitor {
+public:
+    explicit StDWithinRestore(ClientContext &context) : context(context) {
+    }
+
+    ExpressionPtr VisitReplace(BoundFunctionExpression &expr, ExpressionPtr *) override {
+        if (expr.children.size() != 3 || !StringUtil::CIEquals(expr.function.name, "st_dwithin")) {
```
myrrc: Duckdb already compares identifiers case-insensitively, we can use == here.

vortex-duckdb/src/convert/expr.rs
```diff
@@ -73,15 +80,134 @@ fn build_list_length(expr: Expression, nullability: Nullability) -> Expression {
...
+
+/// Lower geo UDFs to native Vortex geo ops so the work runs in the scan. `None` otherwise.
+fn try_from_geo_function(
+    name: &str,
+    func: &BoundFunction,
+    ctx: ConvertCtx<'_>,
+) -> VortexResult<Option<Expression>> {
+    // Catch-all for every bound function: reject non-geo names before touching the children.
+    if !is_geo_function(name) {
+        debug!("bound function {name}");
```
myrrc: Let's remove this debug print

vortex-duckdb/cpp/include/expr.h
HarukiMoriarty: Yeah, of course!

vortex-duckdb/cpp/vortex_duckdb.cpp
HarukiMoriarty: Done as suggested.

vortex-duckdb/src/convert/expr.rs
HarukiMoriarty: Removed.

benchmarks/duckdb-bench/src/lib.rs
```diff
@@ -78,6 +78,11 @@ impl DuckClient {
         for stmt in &statements {
             self.connection().query(stmt)?;
         }
+        // After `LOAD spatial`, shadow `ST_DWithin` so radius filters push. No-op without it.
+        self.db
```
myrrc, line 82: Do we want to register geo-overrides only for the bench or for users in general?
If it's the latter, we should move the initialization out of benchmark

vortex-duckdb/cpp/include/scalar_fn_pushdown.hpp
```diff
@@ -77,6 +77,14 @@ using Projections = unordered_map<TableIndex, LogicalProjection &>;
 
 LogicalOperatorPtr TryPushdownScalarFunctions(ClientContext &context, LogicalOperatorPtr plan);
 
+/*
+ * Rebind 3-argument `st_dwithin` join conditions (bound via the Vortex override; see expr.h)
+ * back through spatial's original entry, whose folded 2-argument form its spatial-join
+ * optimization requires. Runs in the pre-optimize hook, which is guaranteed to precede every
+ * extension's post-optimize pass; filters are left untouched so they can push into Vortex scans.
```
myrrc: To be honest, I don't understand this comment. Maybe we should add an example showing what's done somewhere (i.e. what function gets overriden and restored), and reference it in other places?

vortex-duckdb/cpp/expr.cpp
```diff
@@ -21,6 +32,43 @@ extern "C" const char *duckdb_vx_sfunc_name(duckdb_vx_sfunc ffi_func) {
...
+            // `RestoreStDWithin` rebinds unpushed calls through the original.
+            auto &catalog = Catalog::GetCatalog(context, DatabaseManager::GetDefaultDatabase(context));
+            // Durable catalogs require the modified mark; scalar function entries are never
+            // persisted, so this is metadata-only.
+            MetaTransaction::Get(context).ModifyDatabase(catalog.GetAttached(), DatabaseModificationType());
+            catalog.CreateFunction(context, info);
+        });
+    } catch (const std::exception &) {
+        // No `spatial` loaded, so there is no `ST_DWithin` to override; nothing to register.
+        return DuckDBSuccess;
```
myrrc: What if ModifyDatabase or other calls failed? We want to propagate an error to the user

vortex-duckdb/src/convert/expr.rs
```diff
@@ -73,15 +80,133 @@ fn build_list_length(expr: Expression, nullability: Nullability) -> Expression {
...
+        _ => Ok(None),
+    }
+}
+
+/// Context threaded through expression conversion.
+#[derive(Clone, Copy)]
+struct ConvertCtx<'a> {
+    /// Substituted for `BoundRef` references when converting scan-scoped table filters.
+    col_sub: Option<&'a Expression>,
+    /// The scan's fields, when known; geo lowering requires them to verify a geometry column.
```
myrrc: Nit: I think the user can infer why this field is needed, let's remove the "why" from the comment

myrrc (approved):
LGTM sans error handling

benchmarks/duckdb-bench/src/lib.rs
HarukiMoriarty: Let's make it internal for benchmark only for now, later I will systematically applied into vortex-duckdb.

vortex-duckdb/cpp/include/scalar_fn_pushdown.hpp
HarukiMoriarty: Let me add one example. The root cause here is that the DuckDB Spatial extension optimizer only recognizes the two-argument version of `ST_DWithin` as optimizable, allowing it to use a spatial index. If we keep the manually written three-argument version, DuckDB will fail to apply this optimization and will instead execute Query 8 using brute force.

robert3005 (approved):

PR #8627, opened 2026-06-30
Decode DateTimePart days with fixed-size chunks

myrrc:
- Decode DateTime parts to AnyColumnar instead of AnyCanonical to
  make use of constant seconds and subseconds.
- Use fixed-chunk iteration instead of from_iter which reallocates
  constantly.
- Unroll iteration in each chunk which allows LLVM to vectorize


encodings/datetime-parts/src/canonical.rs
```diff
@@ -97,6 +107,36 @@ pub fn decode_to_temporal(
...
+    /// widening to i64. If we process the code in fixed-size chunks and unroll
+    /// the chunks with seq_macro, vectorization happens.
+    const CHUNK: usize = 64;
+    let n = days.len();
+    let mut values = BufferMut::<i64>::with_capacity(n);
+    match_each_integer_ptype!(days.ptype(), |D| {
+        let src = days.as_slice::<D>();
+        let (src_chunks, src_rem) = src.as_chunks::<CHUNK>();
+        let (dst_chunks, _) = values.spare_capacity_mut()[..n].as_chunks_mut::<CHUNK>();
+        for (s_chunk, d_chunk) in src_chunks.iter().zip(dst_chunks.iter_mut()) {
```
0ax1, line 122: Did you see any diff in perf between using `seq` to manually unroll or letting the compiler do that based on the fixed range?

myrrc: Yea, without seq_macro with or without manual for i in 0..CHUNK the code is not vectorized.

encodings/datetime-parts/src/canonical.rs
```diff
@@ -45,45 +44,56 @@ pub fn decode_to_temporal(
...
-    let mut values: BufferMut<i64> = match_each_integer_ptype!(days.ptype(), |D| {
-        BufferMut::from_iter(days.as_slice::<D>().iter().map(|d| {
-            let d: i64 = d.as_();
-            d * 86_400 * divisor
-        }))
-    });
-
-    // Seconds/subseconds may be Constant — handle the fast path.
-    if let Some(seconds) = parts.seconds.as_constant() {
+    let seconds = parts.seconds.as_::<AnyColumnar>();
```
0ax1, line 47: Do we have a micro-benchmark exercising this change?

myrrc: clickbench q6 should improve

encodings/datetime-parts/src/canonical.rs
```diff
@@ -97,6 +107,36 @@ pub fn decode_to_temporal(
...
+    /// If "days" are u16 or u32, LLVM doesn't auto-vectorize the code due to
+    /// widening to i64. If we process the code in fixed-size chunks and unroll
+    /// the chunks with seq_macro, vectorization happens.
+    const CHUNK: usize = 64;
+    let n = days.len();
+    let mut values = BufferMut::<i64>::with_capacity(n);
+    match_each_integer_ptype!(days.ptype(), |D| {
+        let src = days.as_slice::<D>();
+        let (src_chunks, src_rem) = src.as_chunks::<CHUNK>();
+        let (dst_chunks, _) = values.spare_capacity_mut()[..n].as_chunks_mut::<CHUNK>();
```
robert3005, line 121: I don't think you can ignore the remainder here. We never check that n is a multiple of CHUNK

encodings/datetime-parts/src/canonical.rs
```diff
@@ -97,6 +107,36 @@ pub fn decode_to_temporal(
...
+        for (s_chunk, d_chunk) in src_chunks.iter().zip(dst_chunks.iter_mut()) {
+            seq_macro::seq!(I in 0..64 {
+                let day: i64 = s_chunk[I].as_();
+                d_chunk[I].write(day * day_to_unit + offset);
+            });
+        }
+        let tail_start = src_chunks.len() * CHUNK;
+        let dst_tail = &mut values.spare_capacity_mut()[tail_start..n];
```
robert3005, line 129: ah, you don't need this if you do my first comment

0ax1 (commented):
In theory makes sense, but I also think we should only land this if there's a measurable perf benefit.

encodings/datetime-parts/src/canonical.rs
myrrc: But I do process it later

robert3005:
I would like to have microbenchmarks so this doesn't randomly regress

myrrc:
I've rechecked and it doesn't improve on x86. Only rational improvement then is use from_trusted_len_iter and decode to AnyColumnar.

myrrc:
Closing due to no impact

PR #8631, opened 2026-06-30
Back VortexSession with ArcSwap for mutable but less errorprone session management

robert3005:
I think immutable session is nice but hard to integrate it with for a lot of
systems


gatesn:
Ok.... so instead of deadlocks we now have a footgun for getting stale data? Why have a mutable session if you make it easy to snapshot state out of the session that will never reflect mutability?

vortex-session/src/session.rs
```diff
@@ -0,0 +1,496 @@
...
+                .or_insert_with(|| Arc::clone(&default));
+            next
+        });
+    }
+
+    /// Register a session variable of type `V`, replacing any existing variable of that type.
+    ///
+    /// The mutation is applied in place to the shared backing store, so it is visible through every
+    /// clone of this session.
+    pub fn register<V: VortexSessionVar>(&self, var: V) {
```
robert3005: We should get rid of this method and require V to have Default

robert3005:
I don't understand what you're referring to? Can you maybe write a test to demonstrate?

AdamGS (approved):

PR #8643, opened 2026-07-02
Remove variants module

robert3005:
This module was never fully fleshed out and we have different ways to accomplish same things. This also migrates the PValue search sorted to be purely primitive based


vortex-array/src/search_sorted/primitive.rs
```diff
@@ -0,0 +1,92 @@
...
+use std::marker::PhantomData;
+
+use vortex_error::VortexResult;
+
+use crate::ArrayRef;
+use crate::ExecutionCtx;
+use crate::dtype::NativePType;
+use crate::search_sorted::IndexOrd;
+
+pub struct TypedPrimitiveArray<'a, T>(&'a ArrayRef, RefCell<&'a mut ExecutionCtx>, PhantomData<T>);
```
robert3005: this should be something less generic

lwwmanning:
@codex please review

encodings/runend/src/ops.rs
```diff
@@ -21,20 +22,24 @@ impl OperationsVTable<RunEnd> for RunEnd {
...
 
 /// Find the physical offset for and index that would be an end of the slice i.e., one past the last element.
 ///
 /// If the index exists in the array we want to take that position (as we are searching from the right)
 /// otherwise we want to take the next one
-pub(crate) fn find_slice_end_index(array: &ArrayRef, index: usize) -> VortexResult<usize> {
-    let result = array
-        .as_primitive_typed()
-        .search_sorted(&PValue::from(index), SearchSortedSide::Right)?;
+pub(crate) fn find_slice_end_index(
```
lwwmanning, line 34: this is always invoked with `array.ends()` as the `array` fn param, can we pass in typed ArrayView of RunEnd instead? and then internalize that complexity in the fn. current sig & invocation are a bit errorprone

encodings/runend/src/ops.rs
```diff
@@ -47,6 +52,18 @@ pub(crate) fn find_slice_end_index(array: &ArrayRef, index: usize) -> VortexResu
     })
 }
 
+pub(crate) fn find_physical_index(
+    array: &ArrayRef,
```
lwwmanning, line 56: same here

vortex-array/src/search_sorted/primitive.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::search_sorted::IndexOrd;
+
+/// A [`SearchSorted`](crate::search_sorted::SearchSorted) adapter over a sorted primitive-typed
+/// array, comparing elements as the native type `T`.
+///
+/// Values can be searched as `T`, `Option<T>`, or `usize`. Searching as `T` or `usize` treats
+/// null elements as `T::zero()`; use `Option<T>` when the array may contain nulls, in which case
+/// nulls sort before all non-null values.
+pub struct SearchSortedPrimitiveArray<'a, T>(
+    &'a ArrayRef,
```
lwwmanning, line 22: again, is there not a reason to use the existing typed ArrayView stuff instead?

lwwmanning (changes requested):

encodings/runend/src/ops.rs
robert3005: There's two cases where we want to call it on an array we are going to make an ends array, the array variant munges the parameters and delegates to this method

encodings/runend/src/ops.rs
robert3005: I have added a method that does this correctly on RunEndArray but we still need this for arrow import where we validate the array that arrow gives us

vortex-array/src/search_sorted/primitive.rs
robert3005: So this is a specialisation for SearchSorted over things that are known to be of primitive type. We have a general version of this for ArrayRef. In this pr I have been removed the silly variants adapters we had, and apart from this functionality, everything else had a better version. The view doesn't help beacuse I want "Any array that has primitive dtype". We should figure out what's a good way of implementing this functionality

lwwmanning (approved):

PR #8645, opened 2026-07-02
Ungrouped aggregate function pushdown for duckdb

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

myrrc:
This PR adds support for pushing down ungropped aggregates min,max,sum,avg,mean,first,any_value,count(col), and count(*).

It consists of roughly 3 parts:
1. Optimizer pass which works same as scalar function pushdown, but without visitors, because we can't express child extract with a visitor.
2. General accumulator handling in table_function.rs. This includes computing partials in every thread (so that we read data in parallel) and merging them, and also synchonization so that only one thread would write the global aggregated row count
3. Handling of count_star() (count(*)) a.k.a row count. Row count is a special accumulator because it doesn't need the array, and there's no column it aggregates.

I've tried the implementation with CountStart being a separate accumulator, and it is much more complex than current one, because you can't provide a real column to accumulate on in select(), and you need to create a dummy one. So, we pay the cost of a separate atomic and a count_star positions vector, but our expression projection handling stays simple.
We also reject count(*) if it's the only aggregate because duckdb calculating it natively (just using chunk lengths) is faster than our accumulator pipeline

On the API input note, nearly all aggregate functions except for count_star() in duckdb have one non-const argument so input is a pair of a column and an expression.

Depends on #8649

### DIFF
--- a/vortex-duckdb/cpp/aggregate_fn_pushdown.cpp
+++ b/vortex-duckdb/cpp/aggregate_fn_pushdown.cpp
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+#include "aggregate_fn_pushdown.hpp"
+#include "duckdb/planner/expression/bound_aggregate_expression.hpp"
+#include "duckdb/planner/expression/bound_columnref_expression.hpp"
+#include "duckdb/planner/operator/logical_aggregate.hpp"
+#include "optimizer.hpp"
+#include "table_function.hpp"
+
+using enum LogicalOperatorType;
+
+LogicalOperatorPtr TryPushdownAggregateFunctions(ClientContext &context, LogicalOperatorPtr plan) {
+    Analyses analyses;
+    Projections projections;
+    FindGetsAndProjections(*plan, analyses, projections);
+    if (analyses.empty()) {
+        return plan;
+    }
+    return RewriteAggregates(context, std::move(plan), analyses, projections);
+}
+
+LogicalOperatorPtr RewriteAggregates(ClientContext &context,
+                                     LogicalOperatorPtr op,
+                                     Analyses &analyses,
+                                     const Projections &projections) {
+    for (auto &child : op->children) {
+        child = RewriteAggregates(context, std::move(child), analyses, projections);
+    }
+    if (op->type == LOGICAL_AGGREGATE_AND_GROUP_BY) {
+        return TryReplaceAggregate(context, std::move(op), analyses, projections);
+    }
+    return op;
+}
+
+static bool IsUngrouped(const LogicalAggregate &agg) {
+    return agg.groups.empty() && agg.grouping_sets.empty() && agg.grouping_functions.empty() &&
+           !agg.expressions.empty();
+}
+
+constexpr inline idx_t COUNT_STAR_PROJ_IDX = std::numeric_limits<TableColumnStorageIndex>::max();
+
+LogicalOperatorPtr TryReplaceAggregate(ClientContext &context,
+                                       LogicalOperatorPtr op,
+                                       Analyses &analyses,
+                                       const Projections &projections) {
+    LogicalAggregate &agg = op->Cast<LogicalAggregate>();
+    if (!IsUngrouped(agg)) {
+        return op;
+    }
+
+    LogicalGet *const get = GetChildGet(agg);
+    if (get == nullptr) {
+        return op;
+    }
+
+    vector<std::pair<TableColumnScanIndex, const Expression &>> input;
+    const idx_t aggregates_len = agg.expressions.size();
+    input.reserve(aggregates_len);
+
+    for (const auto &expr : agg.expressions) {
+        if (expr->GetExpressionClass() != ExpressionClass::BOUND_AGGREGATE) {
+            return op;
+        }
+        const auto &bound_aggr = expr->Cast<BoundAggregateExpression>();
+        if (bound_aggr.IsDistinct() || bound_aggr.filter != nullptr || bound_aggr.order_bys != nullptr) {
+            return op;
+        }
+
+        if (bound_aggr.function.name == "count_star") {
+            input.emplace_back(COUNT_STAR_PROJ_IDX, *expr);
+            continue;
+        }
+
+        if (bound_aggr.children.size() != 1 ||
+            bound_aggr.children[0]->GetExpressionType() != ExpressionType::BOUND_COLUMN_REF) {
+            return op;
+        }
+        const auto &bound_col = bound_aggr.children[0]->Cast<BoundColumnRefExpression>();
+        const auto binding = Resolve(bound_col.binding, analyses, projections);
+        if (!binding || &binding->analysis.get != get) {
+            return op;
+        }
+        input.emplace_back(binding->column_index, *expr);
+    }
+
+    if (!aggregate_pushdown(context, {*get, input})) {
+        return op;
+    }
+
+    // GET now returns one column per aggregate. Expand existing columns
+    auto &column_ids = get->GetMutableColumnIds();
+    get->types.resize(aggregates_len);
+    get->returned_types.resize(aggregates_len);
+    column_ids.resize(aggregates_len);
+
+    vector<string> names(aggregates_len); // need a copy because we reference original names
+
+    for (idx_t i = 0; i < aggregates_len; i++) {
+        const auto &[column_index, expr] = input[i];
+        if (column_index == COUNT_STAR_PROJ_IDX) {
+            names[i] = "count_star()";
+        } else {
+            const TableColumnStorageIndex storage_index = get->GetColumnIds()[column_index].GetPrimaryIndex();
+            names[i] = get->names[storage_index];
+        }
+        get->types[i] = expr.return_type;
+        get->returned_types[i] = expr.return_type;
+        column_ids[i] = ColumnIndex {i};
+    }
+    get->names = std::move(names);
+    get->projection_ids.clear();
+    get->table_index = agg.aggregate_index;
+
+    unique_ptr<LogicalOperator> &child = agg.children[0];
+    if (child->type == LOGICAL_GET) {
+        return std::move(child);
+    }
+    D_ASSERT(child->type == LOGICAL_PROJECTION);
+    D_ASSERT(child->children.size() == 1);
+    D_ASSERT(child->children[0]->type == LOGICAL_GET);
+    return std::move(child->children[0]);
+}
+
+LogicalGet *GetChildGet(const LogicalAggregate &agg) {
+    if (agg.children.size() != 1) {
+        return nullptr;
+    }
+    LogicalOperator &child = *agg.children[0];
+    LogicalOperator *op;
+    if (child.type == LOGICAL_GET) {
+        op = &child;
+    } else if (child.type == LOGICAL_PROJECTION && child.children.size() == 1 &&
+               child.children[0]->type == LOGICAL_GET) {
+        op = child.children[0].get();
+    } else {
+        return nullptr;
+    }
+    LogicalGet &get = op->Cast<LogicalGet>();
+    return get.function.bind == duckdb_vx_table_function_bind ? &get : nullptr;
+}
--- a/vortex-duckdb/cpp/table_function.cpp
+++ b/vortex-duckdb/cpp/table_function.cpp
@@ -4,6 +4,7 @@
 #include "data.hpp"
 #include "error.hpp"
 #include "table_function.hpp"
+#include "expr.h"
 #include "vortex_duckdb.h"
 #include "table_function.h"
 #include "vortex.h"
@@ -171,12 +172,16 @@ struct CTableBindResult {
     vector<string> &names;
 };
 
+// This is a flaw of Duckdb API which doesn't allow passing non-const
+// expressions. We never modify the value on Rust side.
+static duckdb_vx_expr get_ffi_expr(const Expression &expr) {
+    return reinterpret_cast<duckdb_vx_expr>(const_cast<Expression *>(&expr));
+}
+
 bool projection_expression_pushdown(ClientContext &, const TableFunctionProjectionExpressionInput &input) {
     const auto &bind = input.get.bind_data->Cast<CTableBindData>();
 
-    // This is a flaw of Duckdb API which doesn't allow passing non-const
-    // expressions. We never modify the value on Rust side.
-    auto ffi_expr = reinterpret_cast<duckdb_vx_expr>(const_cast<Expression *>(&input.expression));
+    duckdb_vx_expr ffi_expr = get_ffi_expr(input.expression);
     void *const ffi_bind = bind.ffi_data->DataPtr();
     duckdb_vx_error error_out = nullptr;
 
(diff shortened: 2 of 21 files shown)


myrrc:
As we don't have statistics attached to arrays, and we don't use precomputed statistics as Datafusion does, this PR would likely sit idle for some time until these are done, because currently there's no performance improvement compared to the default versions.

There will be significant improvement once we add zone map statistics

vortex-sqllogictest/bin/sqllogictests-runner.rs
```diff
@@ -198,6 +198,11 @@ fn complete_files(
 }
 
 fn main() -> anyhow::Result<ExitCode> {
+    drop(
+        tracing_subscriber::fmt()
```
0ax1, line 202: Why the drop?

vortex-duckdb/cpp/aggregate_fn_pushdown.cpp
```diff
@@ -0,0 +1,136 @@
...
+        return op;
+    }
+
+    LogicalGet *const get = GetChildGet(agg);
+    if (get == nullptr) {
+        return op;
+    }
+
+    vector<std::pair<TableColumnStorageIndex, const Expression &>> input;
+    const idx_t N = agg.expressions.size();
```
0ax1: When I see `N`, I really expect a compile time constant. Can we give this a lower case more descriptive name?

vortex-duckdb/src/table_function.rs
```diff
@@ -299,11 +374,109 @@ pub fn init_local(global: &TableFunctionGlobal) -> TableFunctionLocal {
...
+                .compare_exchange(0, u64::MAX, Ordering::AcqRel, Ordering::Acquire)
+                .is_err()
+            {
+                return Ok(());
+            }
+
+            let mut accumulators = global_state.partials.lock();
+            let row_count = global_state.row_count.load(Ordering::Acquire) as i64;
+            let mut accum_iter = accumulators.iter_mut();
+            for (i, aggregate) in global_state.aggregates.iter().enumerate() {
```
0ax1: This loop looks a bit manual? How large does `accumulators` get, should we do sth along the lines of `zip` / `filter` here and iter twice?

vortex-duckdb/src/table_function.rs
```diff
@@ -299,11 +374,109 @@ pub fn init_local(global: &TableFunctionGlobal) -> TableFunctionLocal {
...
+            // is_err() means CAS didn't succeed
+            if global_state
+                .pending
+                .compare_exchange(0, u64::MAX, Ordering::AcqRel, Ordering::Acquire)
+                .is_err()
+            {
+                return Ok(());
+            }
+
+            let mut accumulators = global_state.partials.lock();
```
0ax1, line 482: Any way we can simplify the synchronization mechanism. Rather than using locks _and_ atomics.

0ax1 (commented):
Couple of nits and q.

vortex-sqllogictest/bin/sqllogictests-runner.rs
myrrc: Leftover from other pr, should be removed

vortex-duckdb/cpp/aggregate_fn_pushdown.cpp
myrrc: +1

vortex-duckdb/src/table_function.rs
myrrc: We support 6 distinct aggregations but there may be many more columns for them.

vortex-duckdb/src/table_function.rs
myrrc: We can't avoid mutex because DynAccumulator is not thread-safe, and can't use just mutex because every ArrayRef accumulation would trigger synchronization on partials.

Atomics are mostly for count_star(*) because it's a special accumulator

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -191,6 +196,33 @@ bool projection_expression_pushdown(ClientContext &, const TableFunctionProjecti
     return ret;
 }
 
+extern "C" {
+idx_t duckdb_vx_aggregate_len(duckdb_vx_agg_input ffi_input) {
+    return reinterpret_cast<const TableFunctionUngroupedAggregateInput *>(ffi_input)->projections.size();
+}
+
+duckdb_vx_expr duckdb_vx_aggregate_i(duckdb_vx_agg_input ffi_input, idx_t i, idx_t *proj_idx) {
```
0ax1: `duckdb_vx_aggregate_i`, let's make the suffix more descriptive.

vortex-duckdb/cpp/table_function.cpp
```diff
@@ -191,6 +196,33 @@ bool projection_expression_pushdown(ClientContext &, const TableFunctionProjecti
...
+    const auto &input = *reinterpret_cast<const TableFunctionUngroupedAggregateInput *>(ffi_input);
+    const auto &[scan_index, expr] = input.projections[i];
+    *proj_idx = scan_index == COUNT_STAR_PROJ_IDX ? scan_index
+                                                  : input.get.GetColumnIds()[scan_index].GetPrimaryIndex();
+    return get_ffi_expr(expr);
```
0ax1, line 209: Is this our convention, prefixing getters with `get_` consistently? In general, I think it's nice not doing that for getters.

vortex-duckdb/include/vortex.h
```diff
@@ -11,6 +11,8 @@
 
 #pragma once
 
+#define COUNT_STAR_PROJ_IDX UINT64_MAX
```
0ax1, line 14: Aren't all indices::max uint64. Can we get away with a single define? As in why specialize for `count *`? As discussed offline, maybe we can consider constexpr constants rather than pragma defines.

vortex-duckdb/src/convert/expr.rs
```diff
@@ -62,17 +76,18 @@ use crate::duckdb::ExpressionClass::BoundComparison;
 use crate::duckdb::ExpressionClass::BoundConjunction;
 use crate::duckdb::ExpressionClass::BoundConstant;
 use crate::duckdb::ExpressionClass::BoundRef;
+use crate::duckdb::ExpressionRef;
 use crate::projection::DuckdbField;
 
-fn from_bound_str(value: &duckdb::ExpressionRef) -> VortexResult<String> {
```
0ax1, line 67: tbh, I do kinda like knowing where the `duckdb::ExpressionRef` comes from in terms of namespace, given we have both vortex and duckdb expr in the mix.

vortex-duckdb/src/convert/expr.rs
```diff
@@ -319,7 +331,7 @@ pub fn try_from_bound_expression(
 }
 
 pub(super) fn try_from_bound_expression_with_col_sub(
-    value: &duckdb::ExpressionRef,
```
0ax1, line 322: -//-

vortex-duckdb/src/convert/expr.rs
```diff
@@ -463,10 +476,74 @@ pub fn try_from_projection_expression(
     })
 }
 
+// Aggregations we handle in Vortex and not in Duckdb
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PushedAggregate {
```
0ax1, line 483: Pushed, or Pushdown?

vortex-duckdb/src/convert/expr.rs
```diff
@@ -463,10 +476,74 @@ pub fn try_from_projection_expression(
...
+                PairOptions(opts, opts),
+                dtype,
+            )?),
+            Self::First => Box::new(Accumulator::try_new(First, AggregateEmptyOptions, dtype)?),
+            Self::Count => Box::new(Accumulator::try_new(Count, opts, dtype)?),
+        })
+    }
+}
+
+pub fn try_from_projection_aggregate(
```
0ax1, line 526: Should pub fns have docs in vortex-duckdb?

vortex-duckdb/src/duckdb/aggregate_pushdown.rs
```diff
@@ -0,0 +1,39 @@
...
+use num_traits::AsPrimitive;
+
+use crate::cpp;
+use crate::duckdb::Expression;
+use crate::duckdb::ExpressionRef;
+use crate::lifetime_wrapper;
+
+lifetime_wrapper!(AggregatePushdownInput, cpp::duckdb_vx_agg_input, |_| {});
+
+pub struct AggregateExpression<'a> {
```
0ax1, line 16: docs

vortex-duckdb/src/duckdb/aggregate_pushdown.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub expr: &'a ExpressionRef,
+    pub projection_id: u64,
+}
+
+impl AggregatePushdownInputRef {
+    pub fn len(&self) -> usize {
+        let len: usize = unsafe { cpp::duckdb_vx_aggregate_len(self.as_ptr()) }.as_();
```
0ax1: can't we return `unsafe { cpp::duckdb_vx_aggregate_len(self.as_ptr()) }.as_()` directly?

vortex-duckdb/src/projection.rs
```diff
@@ -187,6 +188,27 @@ impl Projection {
             file_row_number_column_pos,
         }
     }
+
+    pub fn new_aggregate(aggregates: &[ColumnAggregate], fields: &[DuckdbField]) -> Self {
```
0ax1, line 194: docs

vortex-duckdb/src/projection.rs
```diff
@@ -187,6 +188,27 @@ impl Projection {
...
+        let mut names = Vec::with_capacity(aggregates.len());
+        let mut seen: Vec<u64> = Vec::with_capacity(aggregates.len());
+        for aggregate in aggregates {
```
0ax1, line 197: This looks it can be a `filter` + `collect`.

vortex-duckdb/src/table_function.rs
```diff
@@ -124,6 +139,16 @@ pub struct TableFunctionGlobal {
     bytes_read: AtomicU64,
     file_index_column_pos: Option<usize>,
     file_row_number_column_pos: Option<usize>,
+
+    // Following fields are initialized and used only in aggregate scans
+    /// ArrayRef's scanned but not aggregated in "partials".
+    /// 0 means all arrays have been aggregated but output is not written.
+    /// u64::MAX means arrays have been aggregated and we've written output row
+    pending: Arc<AtomicU64>,
```
0ax1, line 147: `Following fields are initialized` Consider `Option` to represent this. Rust requires these to be initialized so the wording is slightly off.

vortex-duckdb/src/table_function.rs
```diff
@@ -239,6 +283,7 @@ pub fn init_global(init_input: &TableInitInput) -> VortexResult<TableFunctionGlo
         .partitions()
         .map(move |partition| {
             let tx = tx.clone();
+            let pending = Arc::clone(&pending_producer);
```
0ax1, line 286: @claude review, only this section, for threading / synchronization issues and whether the code can be simplified.

vortex-duckdb/src/table_function.rs
0ax1: Saying we could split it into 2 loops and make this simpler?

vortex-duckdb/src/table_function.rs
```diff
@@ -336,11 +411,109 @@ pub fn init_local(global: &TableFunctionGlobal) -> TableFunctionLocal {
...
+                    ColumnAggregate::CountStar => Value::from(row_count),
+                };
+                chunk.get_vector_mut(i).reference_value(&value);
+            }
+            chunk.set_len(1);
+            return Ok(());
+        };
+        let array = convert_result(result?.0, &mut ctx)?;
+
+        for (i, partial) in positions.iter().zip(local_state.partials.iter_mut()) {
```
0ax1, line 500: `i` => `idx`

vortex-duckdb/src/table_function.rs
```diff
@@ -336,11 +411,109 @@ pub fn init_local(global: &TableFunctionGlobal) -> TableFunctionLocal {
...
+                global.combine_partials(local.flush()?)?;
+            }
+        }
+
+        if has_count_star {
+            global_state
+                .row_count
+                .fetch_add(array.len() as u64, Ordering::Relaxed);
+        }
+        global_state.pending.fetch_sub(1, Ordering::Release);
```
0ax1, line 516: @claude also here check for threading / synchronization correctness

vortex-duckdb/src/table_function.rs
```diff
@@ -556,6 +769,31 @@ pub fn to_string(bind_data: &TableFunctionBind, map: &mut DuckdbStringMapRef) {
...
+    if !bind_data.aggregates.is_empty() {
+        let aggregations = bind_data
+            .aggregates
+            .iter()
+            .map(|agg| match agg {
+                ColumnAggregate::Real {
+                    projection_id,
+                    aggregate,
+                } => {
+                    let projection_id: usize = projection_id.as_();
```
0ax1, line 782: inline `let projection_id: usize = projection_id.as_();` ?

0ax1 (commented):
q and nits

vortex-duckdb/cpp/table_function.cpp
myrrc: +1, but let's leave it for further PRs

vortex-duckdb/include/vortex.h
myrrc: vortex.h is a C header and thus we can't use contexpr. In fact, this define is generated from a constexpr expression.

vortex-duckdb/src/convert/expr.rs
myrrc: Pushed, because when this enum is instantiated we have already pushed this aggregate to Vortex.

vortex-duckdb/src/table_function.rs
myrrc: Can't do that, as_() doesn't know it needs to cast to usize.

vortex-duckdb/src/table_function.rs
myrrc: +1

vortex-duckdb/src/table_function.rs
0ax1: return type?

0ax1: can prob also do turbofish notation otherwise

vortex-duckdb/src/duckdb/aggregate_pushdown.rs
myrrc: +1

vortex-duckdb/src/convert/expr.rs
myrrc: +1

vortex-duckdb/src/convert/expr.rs
myrrc: +1

vortex-duckdb/cpp/table_function.cpp
myrrc: +1

vortex-duckdb/src/projection.rs
myrrc: Update semantics in filter() are counter-intuitive so I'd leave this as is.

0ax1 (approved):

PR #8646, opened 2026-07-02
feat(vortex-geo): geometry `Bbox` zone-map statistic + distance-filter pruning

HarukiMoriarty:
## Summary

Adds spatial chunk-pruning to Vortex. A new `GeometryBounds` aggregate stores a per-chunk minimum bounding box (MBR) as a zone-map statistic, and a stats-rewrite rule uses it to skip chunks that cannot satisfy a `ST_Distance(geom, const) <= r` filter.

## Limitation
- Only the `<=` / `<` are pruned. `>` / `>=` are soundly prunable via the symmetric farthest-corner bound but are intentionally omitted (rarely?)
- Pruning is sound, but the performance is highly related with the geo column write order, selectivity depends on a spatially clustered layout (e.g. a Hilbert/Z-order sort) so chunk MBRs are tight and non-overlapping. 

## Testing

8 new vortex-geo tests. Point bbox across batches; Polygon bbox over all ring vertices, empty group → null, and registry self-declaration. only <=/< prune while >/>=/==/!= don't (parameterized), distance symmetry, non-distance comparisons ignored, and an end-to-end falsify.

## Performance

SF=1
```
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Query ┃ duckdb:parquet (base) ┃   duckdb:vortex ┃ duckdb:vortex-geo-native ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ 1     │                39.8ms │  15.5ms (0.39x) │            3.1ms (0.08x) │
│ 2     │               147.8ms │  39.9ms (0.27x) │           47.4ms (0.32x) │
│ 3     │                66.1ms │  24.0ms (0.36x) │            5.1ms (0.08x) │
│ 4     │               563.1ms │  60.2ms (0.11x) │          109.7ms (0.19x) │
│ 5     │               356.8ms │ 284.1ms (0.80x) │          287.9ms (0.81x) │
│ 6     │               677.3ms │  93.9ms (0.14x) │          149.5ms (0.22x) │
│ 7     │               174.0ms │  70.3ms (0.40x) │           96.7ms (0.56x) │
│ 8     │               142.4ms │  48.4ms (0.34x) │           66.1ms (0.46x) │
│ 9     │                18.7ms │  17.0ms (0.91x) │           19.3ms (1.03x) │
└───────┴───────────────────────┴─────────────────┴──────────────────────────┘
```

SF=3
```
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Query ┃ duckdb:parquet (base) ┃   duckdb:vortex ┃ duckdb:vortex-geo-native ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ 1     │                47.8ms │  36.4ms (0.76x) │            4.0ms (0.08x) │
│ 2     │               143.8ms │  61.0ms (0.42x) │          113.1ms (0.79x) │
│ 3     │                83.1ms │  68.2ms (0.82x) │            7.2ms (0.09x) │
│ 4     │               557.2ms │  67.4ms (0.12x) │          146.2ms (0.26x) │
│ 5     │               949.6ms │ 882.9ms (0.93x) │          897.8ms (0.95x) │
│ 6     │               674.5ms │ 124.5ms (0.18x) │          231.7ms (0.34x) │
│ 7     │               332.5ms │ 330.9ms (1.00x) │          277.2ms (0.83x) │
│ 8     │               183.7ms │ 162.0ms (0.88x) │          209.2ms (1.14x) │
│ 9     │                28.5ms │  25.9ms (0.91x) │           28.7ms (1.01x) │
└───────┴───────────────────────┴─────────────────┴──────────────────────────┘
```

SF=10
```
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Query ┃ duckdb:parquet (base) ┃   duckdb:vortex ┃ duckdb:vortex-geo-native ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ 1     │               161.6ms │ 109.2ms (0.68x) │            7.8ms (0.05x) │
│ 2     │               358.4ms │ 187.9ms (0.52x) │          329.1ms (0.92x) │
│ 3     │               270.7ms │ 237.7ms (0.88x) │           14.3ms (0.05x) │
│ 4     │               912.7ms │ 131.0ms (0.14x) │          157.8ms (0.17x) │
│ 5     │                 3.37s │   3.10s (0.92x) │            3.14s (0.93x) │
│ 6     │                 1.25s │ 301.7ms (0.24x) │          501.0ms (0.40x) │
│ 7     │                 1.27s │ 914.1ms (0.72x) │          969.6ms (0.76x) │
│ 8     │               693.5ms │ 918.7ms (1.32x) │          683.1ms (0.99x) │
│ 9     │                36.3ms │  35.7ms (0.98x) │           44.5ms (1.22x) │
└───────┴───────────────────────┴─────────────────┴──────────────────────────┘
```

Takeaway: when the data is pre-sorted, the bounding box pruning can significantly reduce the intermediate results DuckDB read in, make Q1 and Q3 significantly faster

### DIFF
--- a/vortex-array/src/aggregate_fn/plugin.rs
+++ b/vortex-array/src/aggregate_fn/plugin.rs
@@ -10,6 +10,7 @@ use crate::aggregate_fn::AggregateFn;
 use crate::aggregate_fn::AggregateFnId;
 use crate::aggregate_fn::AggregateFnRef;
 use crate::aggregate_fn::AggregateFnVTable;
+use crate::dtype::DType;
 
 /// Reference-counted pointer to an aggregate function plugin.
 pub type AggregateFnPluginRef = Arc<dyn AggregateFnPlugin>;
@@ -28,6 +29,12 @@ pub trait AggregateFnPlugin: 'static + Send + Sync {
     /// Deserialize an aggregate function from serialized metadata.
     fn deserialize(&self, metadata: &[u8], session: &VortexSession)
     -> VortexResult<AggregateFnRef>;
+
+    /// The default per-chunk zone statistic to store for a column of `input_dtype`, or `None` if
+    /// this aggregate isn't one.
+    fn zone_stat_default(&self, _input_dtype: &DType) -> Option<AggregateFnRef> {
+        None
+    }
 }
 
 impl std::fmt::Debug for dyn AggregateFnPlugin {
@@ -50,5 +57,9 @@ impl<V: AggregateFnVTable> AggregateFnPlugin for V {
     ) -> VortexResult<AggregateFnRef> {
         let options = AggregateFnVTable::deserialize(self, metadata, session)?;
         Ok(AggregateFn::new(self.clone(), options).erased())
+    }
+
+    fn zone_stat_default(&self, input_dtype: &DType) -> Option<AggregateFnRef> {
+        AggregateFnVTable::zone_stat_default(self, input_dtype)
     }
 }
--- a/vortex-array/src/aggregate_fn/plugin.rs
+++ b/vortex-array/src/aggregate_fn/plugin.rs
@@ -30,8 +30,15 @@ pub trait AggregateFnPlugin: 'static + Send + Sync {
     fn deserialize(&self, metadata: &[u8], session: &VortexSession)
     -> VortexResult<AggregateFnRef>;
 
-    /// The default per-chunk zone statistic to store for a column of `input_dtype`, or `None` if
-    /// this aggregate isn't one.
+    /// The default zone statistic (per-chunk) for a column of `input_dtype`, or `None` if the
+    /// dtype is not supported (or this aggregate is not a zone statistic at all).
+    ///
+    /// This is how a registered aggregate volunteers itself as a per-chunk statistic: when a
+    /// zoned layout writer opens a column, it collects every plugin's default via
+    /// [`AggregateFnSession::zone_stat_defaults`], so new statistics can be added without the
+    /// writer knowing about them.
+    ///
+    /// [`AggregateFnSession::zone_stat_defaults`]: crate::aggregate_fn::session::AggregateFnSession::zone_stat_defaults
     fn zone_stat_default(&self, _input_dtype: &DType) -> Option<AggregateFnRef> {
         None
     }
--- a/vortex-array/src/aggregate_fn/session.rs
+++ b/vortex-array/src/aggregate_fn/session.rs
@@ -10,6 +10,7 @@ use vortex_session::SessionVar;
 
 use crate::aggregate_fn::AggregateFnId;
 use crate::aggregate_fn::AggregateFnPluginRef;
+use crate::aggregate_fn::AggregateFnRef;
 use crate::aggregate_fn::AggregateFnVTable;
 use crate::aggregate_fn::fns::all_nan::AllNan;
 use crate::aggregate_fn::fns::all_non_distinct::AllNonDistinct;
@@ -44,6 +45,7 @@ use crate::arrays::chunked::compute::aggregate::ChunkedArrayAggregate;
 use crate::arrays::dict::compute::is_constant::DictIsConstantKernel;
 use crate::arrays::dict::compute::is_sorted::DictIsSortedKernel;
 use crate::arrays::dict::compute::min_max::DictMinMaxKernel;
+use crate::dtype::DType;
 
 /// Session state for aggregate functions and encoding-specific aggregate kernels.
 ///
@@ -132,6 +134,19 @@ impl AggregateFnSession {
         let id = vtable.id();
         let pluginref = Arc::new(vtable) as AggregateFnPluginRef;
         self.registry.insert(id, pluginref);
+    }
+
+    /// The default per-chunk zone statistics for a column of `input_dtype`, collected from every
+    /// registered aggregate's `zone_stat_default`.
+    pub fn zone_stat_defaults(&self, input_dtype: &DType) -> Vec<AggregateFnRef> {
+        self.registry.read(|registry| {
+            let mut fns: Vec<AggregateFnRef> = registry
+                .values()
+                .filter_map(|plugin| plugin.zone_stat_default(input_dtype))
+                .collect();
+            fns.sort_by_key(|f| f.id());
+            fns
+        })
     }
 
     /// Returns the aggregate kernel registered for `array_id` and `agg_fn_id`, if any.
--- a/vortex-array/src/aggregate_fn/session.rs
+++ b/vortex-array/src/aggregate_fn/session.rs
@@ -138,6 +138,9 @@ impl AggregateFnSession {
 
     /// The default per-chunk zone statistics for a column of `input_dtype`, collected from every
     /// registered aggregate's `zone_stat_default`.
+    ///
+    /// Each call scans the whole plugin registry, so this is intended to be called once per
+    /// column when a zoned writer is opened, not per chunk or per row.
     pub fn zone_stat_defaults(&self, input_dtype: &DType) -> Vec<AggregateFnRef> {
         self.registry.read(|registry| {
             let mut fns: Vec<AggregateFnRef> = registry
--- a/vortex-geo/src/aggregate_fn/aabb.rs
+++ b/vortex-geo/src/aggregate_fn/aabb.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-//! A zone-map statistic: the 2D bounding box of a native geometry column.
+//! The 2D axis-aligned bounding-box (AABB) aggregate for native geometry columns.
 
 use geo::Rect as GeoRect;
 use vortex_array::ArrayRef;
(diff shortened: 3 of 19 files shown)


vortex-array/src/aggregate_fn/plugin.rs
```diff
@@ -28,6 +29,12 @@ pub trait AggregateFnPlugin: 'static + Send + Sync {
     /// Deserialize an aggregate function from serialized metadata.
     fn deserialize(&self, metadata: &[u8], session: &VortexSession)
     -> VortexResult<AggregateFnRef>;
+
+    /// The default per-chunk zone statistic to store for a column of `input_dtype`, or `None` if
+    /// this aggregate isn't one.
```
connortsui20: Nit: this is kind of hard to understand, maybe say something like this or similar:

```rust
/// The default zone statistic (per-chunk) for a column of `input_dtype`,
/// or `None` if the dtype is not supported.
```rust

**Review by connortsui20 — COMMENTED** [review]
Seems fine to me, but someone who has worked on aggregates and stats more should probably review this

File: vortex-array/src/aggregate_fn/plugin.rs (hunk 1 above)
Comment (joseph-isaacs) [reply vortex-array/src/aggregate_fn/plugin.rs]: What is this for?

File: vortex-array/src/aggregate_fn/session.rs [hunk 2]
```diff
@@ -134,6 +136,19 @@ impl AggregateFnSession {
...
+    /// registered aggregate's `zone_stat_default`.
+    pub fn zone_stat_defaults(&self, input_dtype: &DType) -> Vec<AggregateFnRef> {
+        self.registry.read(|registry| {
+            let mut fns: Vec<AggregateFnRef> = registry
+                .values()
+                .filter_map(|plugin| plugin.zone_stat_default(input_dtype))
+                .collect();
+            fns.sort_by_key(|f| f.id());
+            fns
+        })
```
joseph-isaacs, line 152: worth nothing this is expensive if called a lot

vortex-geo/src/aggregate_fn/aabb.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+impl BoundsPartial {
+    /// Grow the accumulated box to also cover `other`.
+    fn merge(&mut self, other: GeoRect<f64>) {
+        self.rect = Some(match self.rect {
+            Some(cur) => GeoRect::new(
+                (
+                    cur.min().x.min(other.min().x),
+                    cur.min().y.min(other.min().y),
```
joseph-isaacs, line 54: can do and_then and unwrap?

vortex-geo/src/aggregate_fn/bounds.rs
```diff
@@ -0,0 +1,466 @@
...
+use crate::extension::GeoMetadata;
+use crate::extension::Rect;
+use crate::extension::box_storage_dtype;
+use crate::extension::coordinate::Dimension;
+use crate::extension::flatten_coordinates;
+use crate::extension::is_native_geometry;
+
+/// Aggregates a native geometry column's 2D minimum bounding box as the native `geoarrow.box` type.
+/// Stored as a zone statistic, it lets spatial filters skip chunks whose bounding box cannot hold a
+/// matching row.
```
joseph-isaacs: This is just the aggregate definition, its happens that the partial is used in a zone stat.

vortex-geo/src/aggregate_fn/bounds.rs
```diff
@@ -0,0 +1,466 @@
...
+    );
+    Scalar::extension::<Rect>(GeoMetadata::default(), storage)
+}
+
+impl AggregateFnVTable for GeometryBounds {
+    type Options = EmptyOptions;
+    type Partial = BoundsPartial;
+
+    fn id(&self) -> AggregateFnId {
+        static ID: CachedId = CachedId::new("vortex.geo.bounds");
```
joseph-isaacs: Why is this bounds and not box or rect?

vortex-geo/src/prune.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    /// Run the rule against `GeoDistance(root, origin) <operator> radius`, operands swapped when
+    /// `geom_first` is false.
+    fn falsify_distance(
+        operator: Operator,
+        geom_first: bool,
+        radius: f64,
+    ) -> VortexResult<Option<Expression>> {
+        let session = vortex_array::array_session();
+        crate::initialize(&session);
```
joseph-isaacs: this is odd here

vortex-geo/src/prune.rs
```diff
@@ -0,0 +1,523 @@
...
+        // The predicate root is the comparison, not `GeoDistance`, so key on `Binary`.
+        Binary.id()
+    }
+
+    fn falsify(
+        &self,
+        expr: &Expression,
+        ctx: &StatsRewriteCtx<'_>,
+    ) -> VortexResult<Option<Expression>> {
+        // Only the distance comparisons prune; `==` / `!=` and other operators fall through.
```
joseph-isaacs: can we prune other in the future or is it not possible, add comment here

vortex-geo/src/prune.rs
```diff
@@ -0,0 +1,523 @@
...
+        if distance.as_opt::<GeoDistance>().is_none() {
+            return Ok(None);
+        }
+        let Some((geom, constant)) = geometry_and_constant(distance, ctx)? else {
+            return Ok(None);
+        };
+        let Some(radius) = expr.child(1).as_opt::<Literal>() else {
+            return Ok(None);
+        };
+        let Ok(radius) = f64::try_from(radius) else {
```
joseph-isaacs, line 90: is this a bug? hitting this case

vortex-geo/src/prune.rs
```diff
@@ -0,0 +1,523 @@
...
+/// is needed.
+#[derive(Debug)]
+pub struct GeoDistanceBoundsPrune;
+
+impl StatsRewriteRule for GeoDistanceBoundsPrune {
+    fn scalar_fn_id(&self) -> ScalarFnId {
+        // The predicate root is the comparison, not `GeoDistance`, so key on `Binary`.
+        Binary.id()
+    }
+
```
joseph-isaacs: Its interesting this is really the `max` (a least upper bound) of the rect containing the geo like the max used for regular compare.

joseph-isaacs: I think this code with min/max over values may not exist any more

vortex-geo/src/aggregate_fn/bounds.rs
HarukiMoriarty: This name describes the operation, not the value. The aggregate computes the bounds; `box` and `rect` are what the result is. Similarly, `geo` crate calls it `BoundingRect`.

vortex-geo/src/prune.rs
HarukiMoriarty: - A null literal (distance <= NULL), no value to reason about. Scanning is sound: the filter evaluates to null at scan time and drops every row anyway; we just missed a chance to skip the chunk early.
- A non-primitive literal, mostly impossible in a typechecked plan (comparing f64 to a string fails planning), except extension-typed literals, which Binary's typecheck deliberately lets through.

joseph-isaacs: Did you test this I am not how `f64::try_from` and `Option::<f64>::try_from` differ?

joseph-isaacs: Also can you ever have a value that is not in fact f64 for a value typechecked pr

vortex-geo/src/prune.rs
```diff
@@ -0,0 +1,524 @@
...
+        Operator::Lte if radius < 0.0 => return Some(lit(true)),
+        Operator::Lt if radius <= 0.0 => return Some(lit(true)),
+        // `>= r` / `> r` with a negative radius (or zero, for `>=`) match all rows: never prune.
+        Operator::Gte if radius <= 0.0 => return None,
+        Operator::Gt if radius < 0.0 => return None,
+        _ => {}
+    }
+    // The stat is read through `ext_storage`/`get_item`, which propagate a missing stat (null) to
+    // "keep the chunk". Compared squared to avoid a `sqrt`; all operands are `>= 0`.
+    let aabb = ext_storage(stat(geom.clone(), GeometryAabb.bind(EmptyOptions)));
```
joseph-isaacs, line 162: use a custom ctor?

joseph-isaacs (approved):
LG, there are three or so optional comments

vortex-geo/src/prune.rs
HarukiMoriarty: > Did you test this I am not how `f64::try_from` and `Option::<f64>::try_from` differ?

They differ on both axes: `f64::try_from` casts any primitive ptype (integer radius → `10.0`), errors on null/non-primitive; `Option::<f64>::try_from` requires exactly F64 (an integer literal errors, no cast), null → `Ok(None)`. So the Option form would stop integer radii from pruning. Wasn't tested before, I added tests to pin what we tolerate.

> Also can you ever have a value that is not in fact f64 for a value typechecked pr

Through DuckDB, no. Its binder coerces to DOUBLE before pushdown. But Vortex itself never coerces filter expressions (`coerce_expression` exists but has no callers), and `Binary`'s typecheck exempts extension-typed operands, so the rule assumes uncoerced input from any engine: cast what's numeric, decline the rest (never error). The new tests pin exactly that.

PR #8649, opened 2026-07-03
mean() for decimals

myrrc:
Support mean() for decimals (keeps mean as Decimal). Support casting decimals to f64 (we need this for aggregate pushdown because avg(decimal) in duckdb is f64).

### DIFF
--- a/vortex-array/src/aggregate_fn/fns/mean/mod.rs
+++ b/vortex-array/src/aggregate_fn/fns/mean/mod.rs
@@ -18,10 +18,16 @@ use crate::aggregate_fn::combined::CombinedOptions;
 use crate::aggregate_fn::combined::PairOptions;
 use crate::aggregate_fn::fns::count::Count;
 use crate::aggregate_fn::fns::sum::Sum;
+use crate::aggregate_fn::fns::sum::sum_decimal_dtype;
 use crate::builtins::ArrayBuiltins;
 use crate::dtype::DType;
+use crate::dtype::DecimalDType;
+use crate::dtype::MAX_PRECISION;
+use crate::dtype::MAX_SCALE;
 use crate::dtype::Nullability;
 use crate::dtype::PType;
+use crate::dtype::i256;
+use crate::scalar::DecimalValue;
 use crate::scalar::Scalar;
 use crate::scalar_fn::fns::operators::Operator;
 
@@ -45,10 +51,7 @@ pub fn mean(array: &ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Scalar> {
 ///
 /// Implemented as `Sum / Count` via [`BinaryCombined`].
 ///
-/// Coercion / return type:
-/// - Booleans and primitive numeric types are coerced to `f64` and the result
-///   is a nullable `f64`.
-/// - Decimals are kept as decimals but not implemented currently
+/// Booleans and primitive numeric types are cast to f64. Decimals stay decimals.
 #[derive(Clone, Debug)]
 pub struct Mean;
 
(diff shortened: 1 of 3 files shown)


vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -88,20 +82,13 @@ impl BinaryCombined for Mean {
     }
 
     fn finalize(&self, sum: ArrayRef, count: ArrayRef) -> VortexResult<ArrayRef> {
-        let target = match sum.dtype() {
-            DType::Decimal(..) => sum.dtype().with_nullability(Nullability::Nullable),
-            _ => DType::Primitive(PType::F64, Nullability::Nullable),
-        };
+        let target = DType::Primitive(PType::F64, Nullability::Nullable);
```
robert3005, line 97: should decimal mean be f64 as well? The code previously kept them decimals, I think they should stay decimals

myrrc: Hm, okay, that makes sense, I'll rewrite to keep mean as decimal

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -156,13 +157,66 @@ fn mean_output_dtype(input_dtype: &DType) -> Option<DType> {
...
-        }
+        DType::Decimal(decimal_dtype, _) => Some(DType::Decimal(
+            mean_decimal_dtype(&sum_decimal_dtype(decimal_dtype)),
+            Nullability::Nullable,
+        )),
         _ => None,
     }
 }
 
+fn sum_decimal_dtype(input: &DecimalDType) -> DecimalDType {
```
robert3005: I think we want to keep this centralised

vortex-array/src/aggregate_fn/fns/mean/mod.rs
```diff
@@ -156,13 +157,66 @@ fn mean_output_dtype(input_dtype: &DType) -> Option<DType> {
...
+        sum_decimal.scale(),
+        target_decimal_dtype.scale(),
+    ) else {
+        return Ok(Scalar::null(target_dtype));
+    };
+    let mean = sum / i256::from_i128(i128::from(count));
+
+    let Ok(mean) = DecimalValue::try_from_i256(mean, target_decimal_dtype) else {
+        return Ok(Scalar::null(target_dtype));
+    };
```
robert3005, line 206: I just noticed that this is the cast logic + division in the middle. Maybe something to handle in the future

vortex-array/src/arrays/decimal/compute/cast.rs
```diff
@@ -143,6 +151,57 @@ impl CastKernel for Decimal {
...
+    let source_validity = array.validity()?;
+    let n = array.len();
+    let mask = source_validity.execute_mask(n, ctx)?;
+    let validity = source_validity.cast_nullability(nullability, n, ctx)?;
+
+    let scale: i32 = scale.as_();
+    let inv_factor: f64 = 10f64.powi(-scale);
+
+    let buffer = match &mask {
+        Mask::AllFalse(_) => BufferMut::<f64>::zeroed(n).freeze(),
```
robert3005: `Buffer::<f64>::zeroed(n)`

vortex-array/src/arrays/decimal/compute/cast.rs
```diff
@@ -143,6 +151,57 @@ impl CastKernel for Decimal {
     }
 }
 
+fn cast_to_f64(
```
robert3005, line 154: where is this logic used for mean? I think this could be useful but shouldn't we implement all of decimal -> primitive casting?

robert3005 (changes requested):
Mostly some clarifying questions

vortex-array/src/arrays/decimal/compute/cast.rs
myrrc: Not used in this PR, but in #8645 we need this, otherwise dtype for output won't match

myrrc: I have plans for general casting but I want to keep this PR small

robert3005 (approved):

PR #8651, opened 2026-07-03
[don't merge] New C++ api

myrrc:
- Replace C++ api with a new one which depends on vortex-ffi.
- Remove cxx crate from workspace. 
- Port FFI examples and tests to C++ examples as tests.
- Add a reader/writer guide for C++ api.
- Run C++ tests with ASAN and run examples in CI.

Depends: #8735


.github/workflows/ci.yml
```diff
@@ -475,8 +475,8 @@ jobs:
           command: check ${{ matrix.checks }}
 
   cxx-test:
-    name: "C++ build"
-    timeout-minutes: 30
+    name: "C++ API tests"
+    timeout-minutes: 5
```
0ax1, line 527: 5m is probably a  bit tight as an upper bound.

.github/workflows/ci.yml
```diff
@@ -490,20 +490,32 @@ jobs:
       - uses: ./.github/actions/setup-prebuild
         with:
           enable-sccache: "true"
```
0ax1, line 540: We didn't delete `vortex-cxx/src/read.rs`?

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+#pragma once
```
0ax1: Is this header checked by .clang-tidy? (and the src file)

vortex-cxx/.clang-format
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+NamespaceIndentation: None
```
0ax1: Why do we introduce a new formatting here? We have a ref formatting .clang-format at the repo top-level.

vortex-cxx/CMakeLists.txt
```diff
@@ -1,16 +1,16 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright the Vortex contributors
 
-cmake_minimum_required(VERSION 3.22)
+cmake_minimum_required(VERSION 3.10)
```
0ax1, line 4: How come we downgrade the version?

vortex-cxx/CMakeLists.txt
```diff
@@ -21,100 +21,122 @@ else ()
     message(STATUS "Sccache not found")
 endif ()
 
-if (NOT CMAKE_BUILD_TYPE)
-    set(CMAKE_BUILD_TYPE Debug)
-endif()
+# GCC's -Wmissing-field-initializers is a false positive on ScanOptions
```
0ax1: Should this be diagnostics push / pop rather than setting the flag globally?

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+ *
+ * Distinction between String/Binary and StringView/BytesView is that String
+ * needs to be UTF-8 but Binary doesn't need to. Thus a view over String is
+ * a std::string_view, and view over Binary is a std::span<const std::byte>.
+ */
+namespace vortex {
+
+using BinaryView = std::span<const std::byte>;
+
+class Array;
```
0ax1: Why forward decl?

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+class FileWriter;
+class Partition;
+class Scalar;
+class Scan;
+class Session;
+class StringView;
+class Validity;
+struct Field;
+
+namespace detail {
```
0ax1: docs docs docs

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+    AllInvalid = VX_VALIDITY_ALL_INVALID,
+    // Item validity is set in a boolean array: true = valid, false = invalid
+    Array = VX_VALIDITY_ARRAY,
+};
+
+inline constexpr ValidityType NonNullable = ValidityType::NonNullable;
+inline constexpr ValidityType AllValid = ValidityType::AllValid;
+inline constexpr ValidityType AllInvalid = ValidityType::AllInvalid;
+
+enum class DataTypeVariant {
```
0ax1: Shorten to DTypeVariant?

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+    I16 = PTYPE_I16,
+    I32 = PTYPE_I32,
+    I64 = PTYPE_I64,
+    F16 = PTYPE_F16,
+    F32 = PTYPE_F32,
+    F64 = PTYPE_F64,
+};
+
+// 16-bit float
+struct f16 {
```
0ax1: I guess we could detect at build time what C++ version the compiler supports and branch on using the C++23 std lib f16 or our custom one.

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+}
+}  // namespace detail
+
+// Types that can be stored in a Primitive array
+template <class T>
+concept primitive_type = detail::is_numeric_element<T>;
+
+// Types that a PrimitiveView can hold
+template <class T>
+concept view_type = primitive_type<T> || std::is_same_v<T, bool>;
```
0ax1: `primitive_view`? `view_type` in isolation reads vague.

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+    explicit Estimate(vx_estimate raw) : raw_(raw) {}
+
+    bool known() const noexcept { return raw_.type != VX_ESTIMATE_UNKNOWN; }
+    bool exact() const noexcept { return raw_.type == VX_ESTIMATE_EXACT; }
+
+    /**
+     * Estimated count. Throws if not known(). For inexact estimates this is
+     * is an upper bound.
+     */
+    uint64_t value() const;
```
0ax1: Not sure I understand this API. Why not just have value? What's the flow when the caller passes in the fallback. And why does that need to handled in a method of `Estimate`?

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+
+    /**
+     * For a Struct dtype, return its fields in order.
+     * Throws if DataType is not Struct.
+     */
+    std::vector<Field> fields() const;
+
+    // List accessors. Valid only on List and FixedSizeList dtypes
+
+    DataType list_element() const;
```
0ax1: Every public fn in this PR added needs docs.

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+
+// Field of a Struct dtype
+struct Field {
+    std::string name;
+    DataType dtype;
+};
+
+namespace dtype {
+
+inline constexpr bool Nullable = true;
```
0ax1: What's the naming convention we wanna go for? `Nullable` reads like a type. Google convention is `kNullable`, using the `k` prefix for constants.

0ax1 (commented):
first pass

vortex-cxx/CMakeLists.txt
myrrc: It works fine with cmake 3.10 so I don't see any point in setting up a higher version.

vortex-cxx/examples/.clang-tidy
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright the Vortex contributors
+InheritParentConfig: true
```
0ax1, line 3: Same here as for .clang-format. Whatever we do for vortex-cxx should be repo global I think.

vortex-cxx/examples/dtype.cpp
```diff
@@ -0,0 +1,112 @@
...
+                      << ", scale=" << static_cast<int>(dtype.decimal_scale())
+                      << ")";
+            break;
+    }
+    std::cout << (dtype.nullable() ? '?' : ' ') << '\n';
+}
+
+int main(int argc, char** argv) {
+    if (argc != 2) {
+        std::cerr << "Usage: dtype <file glob>\n";
```
0ax1, line 99: nit: Isn't it `Usage: <dtype> <file glob>\`

vortex-cxx/examples/hello-vortex.cpp
```diff
@@ -1,129 +1,63 @@
...
-        if (limit > 0) {
-            // prefer C++ way
-            builder.WithLimit(1);
-            // Rusty way is Ok, but you have to move the builder to an rvalue.
-            // builder = std::move(builder).WithLimit(3);
+    const fs::path out =
+        fs::temp_directory_path() / "hello_vortex_example.vortex";
+    std::cout << "Writing to: " << out.string() << '\n';
+
+    Session session;
```
0ax1: => rename to `VortexSession`,  given that session is a generic description?

vortex-cxx/tests/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+FetchContent_MakeAvailable(magic_enum)
+
+file(GLOB TEST_FILES CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
+
+add_executable(vortex_cxx_test ${TEST_FILES})
+target_include_directories(vortex_cxx_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+target_link_libraries(vortex_cxx_test PRIVATE
+    vortex_cxx_shared
+    Catch2::Catch2WithMain
+    magic_enum::magic_enum
```
0ax1, line 21: `magic_enum` why do we need this?

vortex-cxx/tests/common.hpp
```diff
@@ -0,0 +1,101 @@
...
+    static TempPath unique() {
+        auto dir = fs::temp_directory_path() / "vortex_cxx_test";
+        fs::create_directories(dir);
+        std::string name = std::to_string(std::random_device{}()) + ".vortex";
+        return TempPath{dir / name};
+    }
+
+private:
+    void reset() noexcept {
+        if (!path_.empty()) {
```
0ax1, line 59: Do we care about checking the error in tests?

vortex-cxx/tests/common.hpp
```diff
@@ -0,0 +1,101 @@
...
+            std::error_code ec;
+            fs::remove(path_, ec);
+        }
+    }
+
+    fs::path path_;
+};
+
+inline DataType sample_dtype() {
```
0ax1, line 68: Should we include sample fns headers more than once in a binary executable? ODR shouldn't matter if it's in diff binaries.

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+#include "vortex.hpp"
+
+#include <vortex.h>
```
0ax1: Why is vortex.h included as a system header? (comment on the header if we need to ignore diagnostics in that header)

.github/workflows/ci.yml
myrrc: Tests usually take 2.5 minutes so that's a 2x window.

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+
+#include <cstring>
+#include <mutex>
+#include <string>
+#include <vector>
+
+namespace vortex {
+
+namespace detail {
+void throw_on_error(vx_error* error) {
```
0ax1: We should doc the general error flow somewhere, maybe here.

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+    vx_error_free(error);
+    throw VortexException(message, code);
+}
+}  // namespace detail
+
+using detail::Access;
+using detail::throw_on_error;
+using detail::to_view;
+
+float f16::to_float() const {
```
0ax1: Is this thoroughly unit tested? With edge cases and all?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+
+int8_t DataType::decimal_scale() const {
+    return vx_dtype_decimal_scale(impl_.get());
+}
+
+namespace {
+const vx_struct_fields* struct_fields_or_throw(const vx_dtype* dtype) {
+    const vx_struct_fields* fields = vx_dtype_struct_dtype(dtype);
+    if (fields == nullptr) {
+        throw VortexException("dtype is not a struct",
```
0ax1: What is the general policy with exceptions vs. errors (like sum type Rust errors) etc?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+                              ErrorCode::MismatchedTypes);
+    }
+    return DataType(element);
+}
+
+uint32_t DataType::fixed_size_list_size() const {
+    return vx_dtype_fixed_size_list_size(impl_.get());
+}
+
+namespace dtype {
```
0ax1: Why don't we split logic across multiple files?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+        type_ = other.type_;
+        array_ = other.array_;
+        other.array_ = nullptr;
+        other.type_ = ValidityType::NonNullable;
+    }
+    return *this;
+}
+
+Validity::~Validity() {
+    if (array_ != nullptr) {
```
0ax1: Shouldn't free on a nullptr be a no-op within array, like it is otherwise for C++ in general when calling delete?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+                              ErrorCode::InvalidArgument);
+    }
+    return Access::adopt<Array>(vx_array_clone(array_));
+}
+
+Validity ValidityArray(const Array& bools) {
+    return Validity(ValidityType::Array, vx_array_clone(Access::c_ptr(bools)));
+}
+
+bool detail::ValidityBits::is_null(size_t i) const noexcept {
```
0ax1: Descriptive var names: `i` => `idx`

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+}
+
+Array Array::from_arrow(ArrowArray* array, ArrowSchema* schema, bool nullable) {
+    vx_error* error = nullptr;
+    const vx_array* out = vx_array_from_arrow(array, schema, nullable, &error);
+    throw_on_error(error);
+    return Access::adopt<Array>(out);
+}
+
+size_t Array::size() const { return vx_array_len(impl_.get()); }
```
0ax1: Is `impl_` a weird member var name? `inner_` or so? Maybe we can find sth more descriptive.

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+}
+
+Array Array::field(std::string_view name) const {
+    const DataType dt = dtype();
+    const std::unique_ptr<const vx_struct_fields,
+                          decltype(&vx_struct_fields_free)>
+        fields(struct_fields_or_throw(detail::Access::c_ptr(dt)),
+               &vx_struct_fields_free);
+    const uint64_t n = vx_struct_fields_nfields(fields.get());
+    for (uint64_t i = 0; i < n; ++i) {
```
0ax1: nit: `idx`

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+    return Access::adopt<Array>(out);
+}
+
+Array Array::field(std::string_view name) const {
+    const DataType dt = dtype();
+    const std::unique_ptr<const vx_struct_fields,
+                          decltype(&vx_struct_fields_free)>
+        fields(struct_fields_or_throw(detail::Access::c_ptr(dt)),
+               &vx_struct_fields_free);
+    const uint64_t n = vx_struct_fields_nfields(fields.get());
```
0ax1: `n` => something more descriptive

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+        const vx_view field = vx_struct_fields_field_name(fields.get(), i);
+        if (std::string_view{field.ptr, field.len} == name) {
+            return this->field(i);
+        }
+    }
+    throw VortexException("no field named \"" + std::string(name) + "\"",
+                          ErrorCode::InvalidArgument);
+}
+
+Array Array::slice(size_t start, size_t stop) const {
```
0ax1: begin end to use C++ terminology?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+    if (!vx_array_has_dtype(raw, DTYPE_BINARY)) {
+        throw VortexException("bytes(): array is not a Binary array",
+                              ErrorCode::MismatchedTypes);
+    }
+    detail::ValidityBits validity(session, raw);
+    const size_t n = vx_array_len(raw);
+    return BytesView(std::move(canonical), std::move(validity), n);
+}
+
+Array make_struct(
```
0ax1: Why actually not `&&` in the signature for the return type?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+    const Validity& validity) {
+    StructArrayBuilder builder(validity, fields.size());
+    for (const auto& [name, field] : fields) {
+        builder.add(name, field);
+    }
+    return std::move(builder).build();
+}
+
+void StructArrayBuilder::Deleter::operator()(
+    vx_struct_column_builder* ptr) const noexcept {
```
0ax1: What's our rule of thumb for sprinkling `noexcept`?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+ValidityBits& ValidityBits::operator=(ValidityBits&& other) noexcept {
+    if (this != &other) {
+        if (owner_ != nullptr) {
+            vx_array_free(owner_);
+        }
+        owner_ = other.owner_;
+        bits_ = other.bits_;
+        bit_offset_ = other.bit_offset_;
+        all_invalid_ = other.all_invalid_;
+        other.owner_ = nullptr;
```
0ax1: Why is `= nullptr` responsibility of the assignment operator?

vortex-cxx/vortex.cpp
```diff
@@ -0,0 +1,1065 @@
...
+}
+
+ArrowStream& ArrowStream::operator=(ArrowStream&& other) noexcept {
+    if (this != &other) {
+        if (stream_.release != nullptr) {
+            stream_.release(&stream_);
+        }
+        session_ = std::move(other.session_);
+        stream_ = other.stream_;
+        other.stream_ = ArrowArrayStream{};
```
0ax1: Also here we do why change `other`? `other.stream_ = ArrowArrayStream{};`

vortex-cxx/vortex.hpp
```diff
@@ -0,0 +1,1152 @@
...
+#include <optional>
+#include <span>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+/**
```
0ax1: Same here, let's think about whether all of this should go into a single header. We can still have a convenience vortex.hpp header with only include statements.

0ax1 (commented):
second pass

vortex-cxx/tests/CMakeLists.txt
myrrc: We need compile-time asserts that FFI bindings and C++-side enums have the same number of elements to avoid silent breaks when we change the FFI.
This is afaik the only way to do this before C++23

vortex-cxx/vortex.cpp
myrrc: handle_?

vortex-cxx/examples/hello-vortex.cpp
myrrc: We already have a namespace qualification, so it will be a vortex::Session in user code.

vortex-cxx/examples/dtype.cpp
myrrc: dtype is the name of the program, and <> usually denote parameters

vortex-cxx/vortex.hpp
myrrc: Yes

vortex-cxx/.clang-format
myrrc: +1

vortex-cxx/examples/.clang-tidy
myrrc: +1

vortex-cxx/vortex.cpp
myrrc: +1

vortex-cxx/vortex.cpp
myrrc: +1

vortex-cxx/vortex.hpp
myrrc: I'd leave this as is because we have DataType and not DType (DType comes from C api).

vortex-cxx/vortex.hpp
myrrc: I agree but using it as a type would produce a compilation error, and reading the code is also unambiguous because a type can't be a function argument.

vortex-cxx/vortex.cpp
myrrc: vortex-cxx uses vortex-ffi an an external library, so it is a system header indeed.

vortex-cxx/vortex.hpp
myrrc: +1

vortex-cxx/vortex.cpp
myrrc: Because we don't want a double-free? move constructor change of right side is pretty common in C++.

vortex-cxx/vortex.hpp
myrrc: I'll split it into multiple files

.github/workflows/ci.yml
```diff
@@ -538,20 +538,42 @@ jobs:
...
+      - name: Build and run examples
         run: |
-          cmake -S vortex-cxx/examples -B vortex-cxx/examples/build -DCMAKE_BUILD_TYPE=Release
+          cmake -S vortex-cxx -B vortex-cxx/examples/build -DSANITIZER=asan -DBUILD_EXAMPLES=1 -DTARGET_TRIPLE="x86_64-unknown-linux-gnu"
           cmake --build vortex-cxx/examples/build --parallel $(nproc)
-          vortex-cxx/examples/build/hello-vortex vortex-cxx/examples/goldenfiles/example.vortex
-      - uses: ./.github/actions/check-rebuild
-        with:
-          command: "cargo build --profile ci --locked -p vortex-cxx --lib"
+          cd vortex-cxx/examples/build/examples
```
0ax1, line 565: Why do we do this?
```
# filesystem loop 
rm -fr ../_deps/nanoarrow-src/python/subprojects/arrow-nanoarrow
```

.github/workflows/ci.yml
```diff
@@ -538,20 +538,42 @@ jobs:
...
+          # filesystem loop detected
+          rm -fr ../_deps/nanoarrow-src/python/subprojects/arrow-nanoarrow
+
+          ./writer people0.vortex
```
0ax1, line 569: Should this be invoked through a single entry point which knows that internals of how to exercise the example. Seems manual to invoke these 1 by 1 and brittle when changing the examples?

.github/workflows/ci.yml
```diff
@@ -538,20 +538,42 @@ jobs:
       - uses: ./.github/actions/setup-prebuild
         with:
           enable-sccache: "true"
-      - name: Build and run C++ unit tests
+      - name: Install Rust nightly toolchain
+        run: |
+          rustup toolchain install $NIGHTLY_TOOLCHAIN
+          rustup component add --toolchain $NIGHTLY_TOOLCHAIN rust-src rustfmt clippy llvm-tools-preview
+      - name: Build FFI library (asan)
```
0ax1, line 545: How do the build systems interplay again right now - we implicitly rely on the FFI Vortex lib being in the right spot?

.github/workflows/ci.yml
```diff
@@ -523,8 +523,8 @@ jobs:
           command: check ${{ matrix.checks }}
 
   cxx-test:
-    name: "C++ build"
-    timeout-minutes: 30
+    name: "C++ API tests"
+    timeout-minutes: 5
```
0ax1, line 527: `timeout-minutes: 5` This is our shortest timeout tier and in the repo and is usually only used for checking network connectivity.

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
 =======
 
-The Vortex C++ API provides an idiomatic C++ wrapper around the Vortex C FFI, built using
-`cxx <https://cxx.rs/>`_. It currently supports reading and writing Vortex files and integrates
-with the Arrow C Data Interface via `nanoarrow <https://arrow.apache.org/nanoarrow/>`_.
-
-In the future we will expand the C++ API to cover Vortex's plugin and extension points. Please
-reach out if you are interested in extending Vortex from C++ so we can prioritize these features.
+Vortex C++ API allows you to read and write `.vortex` files directly or via
+Arrow compatibility layer `nanoarrow <https://arrow.apache.org/nanoarrow/>`_.
```
0ax1: `like Nanoarrow` or only with Nanoarrow?

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+The only dependency apart from nanoarrow are Vortex C bindings.
 
 .. note::
-    Both the C++ API and this documentation are a work in progress. The API surface may change
-    significantly. Please reach out if you are interested in using Vortex from C++ so we can
-    prioritize stabilization.
-
+   C++ API is a work in progress. Please reach out to us if you are interested
+   in using Vortex from C++ or you want a feature not covered yet e.g.
```
0ax1, line 11: 👍

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
-* CMake 3.22 or higher
-* C++20 compatible compiler
-* Rust toolchain (for building the underlying Rust library)
+- C++20,
+- Rust toolchain,
+- and CMake 3.10.
 
 .. code-block:: bash
 
+    git clone --depth 1 https://github.com/vortex-data/vortex
```
0ax1, line 26: nit: Should we advocate for using the latest tagged release here. Is there a Git logic that allows for just fetching that maybe.

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+    cd vortex
+    cargo build --release -p vortex-ffi
     cd vortex-cxx
-    mkdir build && cd build
-    cmake ..
-    make -j$(nproc)
+    mkdir build
+    cmake -Bbuild -DCMAKE_BUILD_TYPE=Release
+    cmake --build build -j
```
0ax1, line 36: Is `-j` use parallelism of this machine?

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+For brevity we omit `main()`, system includes and some `using namespace`
+directives.
+
+Assuming you have Vortex files `people0`, `people1`, and `me` in a local folder,
+each containing U8 column "age" and U16 column "height", this is how you
+print all ages for specific heights:
+
+.. code-block:: cpp
+
+    using namespace vortex::expr::ops;
```
0ax1: Do we need the indent?

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+
+    const Session session;
```
0ax1, line 64: Should we const everything in the example. Even though we can, this makes the example more verbose and distracts from the more relevant parts.

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+                std::cout << int(value) << " ";
+            }
+        }
+    }
+    std::cout << "\n";
+
+DataSource and Scan
+^^^^^^^^^^^^^^^^^^^
+
+First, you need to create a Vortex session which holds stuff like object store
```
0ax1: `stuff` is too informal. Properly explain what the session is or drop that part.

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+credentials and does extension bookkeeping (we'll get back to it later). Then
+you need to create a DataSource, which is a view over multiple files which may
+also be remote. You can specify globs for every item as it's shown.
+
+Once you have a DataSource, you can create Scans out of it. A Scan is a single
```
0ax1, line 88: DataSource, symbols in ``?

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+is consumed by following calls so it needs to be non-`const`. ScanOptions which
+are passed to Scan are a simple C++ aggregate so you can initialize any fields
+you want or avoid them altogether (`auto scan = ds.scan()`). Options have
+reasonable defaults: empty projection returns all fields, empty filter doesn't
+filter rows, and so on.
+
+Expressions
+^^^^^^^^^^^
+
+We want to return both column so we omit `.projection`. If we'd want to return
```
0ax1: both column`s` ?

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
+only the "age" column we'd write `.projection = col("age")`. We pass an
+Expression to a filter which returns false for some "height" values, and the
+scan filters them out.
+
+Two additional things to look for in `.filter`: first, we allow overloading
+Expression operators which produce Expressions themselves a-la Eigen. This is
```
0ax1, line 105: Not sure we should assume that ref is known: `operators which produce Expressions themselves a-la Eigen`. We can just explain how it works.

docs/api/cpp/index.rst
```diff
@@ -1,82 +1,260 @@
...
-    #include "vortex/file.hpp"
-    #include "vortex/scan.hpp"
+    using dtype::Nullable;
 
-    // Open a Vortex file and scan with a row limit
-    auto stream = vortex::VortexFile::Open("data.vortex")
-        .CreateScanBuilder()
-        .WithLimit(1000)
-        .IntoStream();
+    const Session session;
```
0ax1, line 170: Same here is `const` really relevant for the example?

vortex-cuda/kernels/src/for.cu
```diff
@@ -9,9 +9,7 @@ template <typename T>
 struct ForOp {
     T reference;
 
-    __device__ inline T operator()(T value) const {
-        return value + reference;
-    }
+    __device__ inline T operator()(T value) const { return value + reference; }
```
0ax1, line 12: CUDA code caught in a stray bullet? 🙂

vortex-cxx/examples/scan_to_arrow.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+int main(int argc, char **argv) {
+    if (argc != 2) {
+        std::cerr << "Scan vortex files to Arrow\nUsage: scan_to_arrow <file glob>\n";
+        return 1;
+    }
+    const char *paths = argv[1];
+
+    Session session;
+    DataSource ds = DataSource::open(session, {paths});
```
0ax1, line 31: Can't i `pass` paths directly without braces? `{paths}` is a bit weird.

vortex-cxx/examples/scan_to_arrow.cpp
```diff
@@ -0,0 +1,65 @@
...
+    DataType out_dtype = ds.dtype();
+    ArrowSchema schema = out_dtype.to_arrow();
+    char schema_buf[10 * 1024];
+    const int64_t schema_len = ArrowSchemaToString(&schema, schema_buf, sizeof schema_buf, 1);
+    std::cout << "arrow schema: " << std::string_view {schema_buf, static_cast<size_t>(schema_len)} << '\n';
+    if (schema.release != nullptr) {
+        schema.release(&schema);
+    }
+
+    ArrowError arrow_error {};
```
0ax1: Does `ArrowError arrow_error {};` not default init without `{}`?

vortex-cxx/examples/scan_to_arrow.cpp
```diff
@@ -0,0 +1,65 @@
...
+    while (auto partition = scan.next_partition()) {
+        ArrowStream stream = std::move(*partition).into_arrow_stream();
+
+        size_t rows = 0;
+        size_t arrays = 0;
+        ArrowArray array = {};
+        while (ArrowArrayStreamGetNext(stream.raw(), &array, &arrow_error) == NANOARROW_OK &&
+               array.release != nullptr) {
+            rows += array.length;
+            ++arrays;
```
0ax1: `array_count` not `arrays`

vortex-cxx/include/vortex/array.hpp
```diff
@@ -0,0 +1,396 @@
...
+    NonNullable = VX_VALIDITY_NON_NULLABLE,
+    // All items are valid
+    AllValid = VX_VALIDITY_ALL_VALID,
+    // All items are invalid
+    AllInvalid = VX_VALIDITY_ALL_INVALID,
+    // Item validity is set in a boolean array: true = valid, false = invalid
+    Array = VX_VALIDITY_ARRAY,
+};
+
+inline constexpr ValidityType NonNullable = ValidityType::NonNullable;
```
0ax1: `constexpr` doesn't need `inline` for ODR right?

vortex-cxx/include/vortex/array.hpp
```diff
@@ -0,0 +1,396 @@
...
+ * function ValidityArray(validity_bools);
+ */
+class Validity {
+public:
+    // NOLINTNEXTLINE(google-explicit-constructor) for NonNullable constructor
+    Validity(ValidityType type) : type_(type), array_(nullptr) {}
+    Validity(ValidityType type, const Array &array);
+
+    Validity(const Validity &other);
+    Validity(Validity &&other) noexcept;
```
0ax1, line 64: Why `noexcept`?

vortex-cxx/include/vortex/array.hpp
```diff
@@ -0,0 +1,396 @@
...
+    Validity &operator=(Validity &&other) noexcept;
+    ~Validity();
+
+    ValidityType type() const noexcept { return type_; }
+
+    // Boolean validity array. Throws if type() != ValidityType::Array.
+    Array array() const;
+
+private:
+    friend struct detail::Access;
```
0ax1, line 77: @claude  Any better structure, such that we don't need `friend`. Think of diff solutions.

vortex-cxx/include/vortex/array.hpp
```diff
@@ -0,0 +1,396 @@
...
+
+    size_t size() const;
+    bool nullable() const;
+    bool has_dtype(DataTypeVariant variant) const;
+    bool is_primitive(PType ptype) const;
+    DataType dtype() const;
+    Validity validity() const;
+    size_t null_count() const;
+
+    // Struct field accessors. Throw if Array's DataType is not Struct
```
0ax1: Proper doxygen doc rather than the stray `//` with newline above?

vortex-cxx/include/vortex/array.hpp
```diff
@@ -0,0 +1,396 @@
...
+
+    Array field(size_t index) const;
+    Array field(std::string_view name) const;
+
+    /*
+     * Zero-copy view of rows [begin, end).
```
0ax1: `Zero-copy view of rows ` You mean non-owned?

vortex-cxx/include/vortex/array.hpp
```diff
@@ -0,0 +1,396 @@
...
+
+    // Bulk view over Utf8 values.
+    StringView strings(const Session &session) const;
+
+    // Bulk view over Binary values.
+    BytesView bytes(const Session &session) const;
+
+private:
+    friend struct detail::Access;
+    friend class StringView;
```
0ax1, line 219: @claude same here can we find a design where we don't need `friend`?

vortex-cxx/include/vortex/array.hpp
```diff
@@ -0,0 +1,396 @@
...
+ *
+ * StructArrayBuilder b(NonNullable, 2);
+ * Array result = builder
+ *     .add("age", ages)
+ *     .add("height", heights)
+ *     .build();
+ */
+class StructArrayBuilder {
+public:
+    explicit StructArrayBuilder(const Validity &validity, size_t capacity = 0);
```
0ax1, line 274: @claude for all constructors / destructuores, check their sanity and correctness.

vortex-cxx/vortex.cpp
myrrc: Array&&? But array clone is already cheap, and I assume && would produce issues with NRVO

vortex-cxx/include/vortex/common.hpp
```diff
@@ -0,0 +1,91 @@
...
+#pragma once
+
+#include <cstdint>
+#include <span>
+#include <string_view>
+#include <utility>
+#include <vortex.h>
+
+namespace vortex {
+#if __STDCPP_FLOAT64_T__ != 1
```
0ax1: Why not `ifdef` but check against `1`?

.github/workflows/ci.yml
myrrc: Build system issue. If you do glob expansion without it, the code errors with "filesystem loop detected"

vortex-cxx/include/vortex/common.hpp
myrrc: This is the example from https://en.cppreference.com/cpp/types/floating-point, but looks weird, I agree

.github/workflows/ci.yml
myrrc: Yes, the user must build the library by using cargo manually. This is better than implicit builds since we show the command.

docs/api/cpp/index.rst
myrrc: Yes. As alternative we can use -GNinja which uses -j by default.

docs/api/cpp/index.rst
myrrc: Yes, this is Doxygen/RST requirement

docs/api/cpp/index.rst
myrrc: I don't have a strong opinion here, but having a common sense for const-ness is quite useful for a C++ engineer

vortex-cxx/examples/scan_to_arrow.cpp
myrrc: This is the footgun we discussed before. If there's a constructor from string_view, `open({'file', 'file2'})` crashes because string_view can be constructed from a pointer (first argument) and a length (second on, const char* reinterpreted as size_t).
We can have a constructor from const std::string& but this would force a memory allocation even for string literals which IMO is a bit ugly. initializer_list is also ugly but at least it doesn't crash and doesn't allocate

vortex-cxx/include/vortex/array.hpp
myrrc: +1, since C++17 constexpr implies inline

vortex-cxx/include/vortex/array.hpp
myrrc: move constructors must be noexcept, otherwise they're not used. Imagine for T we have a vector<T> and T(T&&) is not noexcept. When a vector reallocates and moves elements, if T throws, vector can't roll back and can't continue https://stackoverflow.com/questions/9249781/are-move-constructors-required-to-be-noexcept

vortex-cxx/include/vortex/array.hpp
myrrc: It's owned in the sense we create a new Array. Zero-copy here means we don't copy the original range into the new one (which is the assumed behaviour), but I agree there should be some better wording.

vortex-cxx/examples/scan_to_arrow.cpp
myrrc: +1

vortex-cxx/include/vortex/array.hpp
myrrc: +1

vortex-cxx/examples/scan_to_arrow.cpp
myrrc: +1

docs/api/cpp/index.rst
myrrc: +1

docs/api/cpp/index.rst
myrrc: +1

.github/workflows/ci.yml
myrrc: It's a 2x window so I think it should be fine.

docs/api/cpp/index.rst
myrrc: We should, but git doesn't have latest tagged release handle, so we'd need to update the tagged version in the docs with every vortex release which is error-prone.

vortex-cxx/CMakeLists.txt
myrrc: +1

vortex-cxx/vortex.cpp
myrrc: I've added some tests and also the branch for std::float16_t.

vortex-cxx/vortex.hpp
myrrc: +1

vortex-cxx/tests/common.hpp
myrrc: Yes, I've added this.

vortex-cxx/vortex.cpp
myrrc: We always throw from C++ side (or panic if we can't throw reasonably).

vortex-cxx/vortex.cpp
myrrc: +1

.github/workflows/ci.yml
myrrc: Not sure. We can make something as run-examples.sh but it is just moving ugliness to other place.

docs/api/cpp/index.rst
myrrc: "like", +1

PR #8658, opened 2026-07-06
Remove borrowed types from FFI; fixed_size_list constructor accepts u32

myrrc:
FFI currently has two types of returned objects: ones are owned (inside of which
there's an Arc), others are borrowed (because there's no Arc inside).

One example of the latter is a struct field name which returns a borrowed
vx_string. We initially had a comment
caller shouldn't call the corresponding _free() function on these. This is not
enough as caller can call a _clone() function on a borrowed vx_string which is
UB because we try to increment a strong reference count on a pointer which isn't
an Arc. From a callee's perspective, we can't distinguish owned and borrowed
types, and this creates weird APIs like struct field dtype being owned but name
being borrowed.

This PR removes all borrowed/owned distinction from FFI layer. Each object now
returned from an FFI function is owned. It can be freely cloned, and each must
be freed explicitly with _free() functions.

On another change, fixed size list constructor previously accepted a size_t
(64-bit unsigned integer on most systems) but validated it against u32 because
that's our file format limitation for fixed size lists. This PR changes the
signature so that now it accepts uint32_t as length parameter.

As the changes are significant enough to trigger bugs in our own tests, this PR also re-enables FFI sanitizer tests in CI.

### DIFF
--- a/vortex-ffi/README.md
+++ b/vortex-ffi/README.md
@@ -1,4 +1,4 @@
-# Vortex C interface
+# Vortex C bindings
 
 ## Updating Headers
 
@@ -20,7 +20,7 @@ target_link_libraries(my_target, vortex_ffi_shared)
 # or target_link_libraries(my_target, vortex_ffi)
 ```
 
-## Running C examples:
+## Running C examples
 
 ```sh
 cmake -Bbuild -DBUILD_EXAMPLES=1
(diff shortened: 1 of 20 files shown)


0ax1 (commented):
Most importantly. Run under ASAN/leak sanitizer. And double check whether any benchmarks regress (bc we heap allocate now instead of borrowing in some cases).

myrrc:
I've reenabled ASAN, and it reports a leak as it should

vortex-ffi/README.md
```diff
@@ -99,7 +99,7 @@ cargo +nightly build -Zbuild-std --target=<target triple> \
 2. Build tests with target triple:
 
 ```sh
-cmake -Bbuild -DWITH_ASAN=1 -DTARGET_TRIPLE=<target triple>
+cmake -Bbuild -DSANITIZER=asan -DTARGET_TRIPLE=<target triple>
```
0ax1, line 102: Does `=asan` imply leak detection?

0ax1 (approved):

vortex-ffi/README.md
myrrc: Yes

PR #8661, opened 2026-07-07
Implement compare in vortex instead of falling back to arrow

robert3005:
Instead of going through arrow compare logic implement it in vortex

Lets us leverage optimised bitbuffer optimistaions and ultimately remove arrow
dependency from vortex-array


gatesn (approved):

ttenbrinkzenz:
Does this also mean these operations can in some cases skip decompression where in the past they couldn't?

robert3005:
They always could, this is reimplementing fallback mechanism where decompression wasn't skipped in vortex instead of using arrow

PR #8664, opened 2026-07-07
fix(array): don't abort on a NUL byte in a field name

### LINKED ISSUE #8652: NUL byte in a struct field name aborts the process (SIGABRT) via Arrow FFI schema export
## Summary

A struct field name containing a NUL byte (`U+0000`) causes a **non-unwinding panic → `SIGABRT`** that aborts the whole process when the file's schema is exported across the Arrow C Data FFI boundary. Because it aborts rather than returning an error, a single crafted or mislabeled file can take down any long-lived process (a query server, a JNI host) that opens it — so it reads as a robustness/DoS concern, not just a validation gap.

Observed via the JNI bindings (`vortex-jni` 0.75.0) on macOS aarch64, but the crash site is in `arrow-rs`'s FFI schema export, so it is not JNI-specific.

## Reproduction

1. Produce a Vortex file whose top-level struct has a field whose name contains an embedded NUL, e.g. the 10-char string `col` + `\0` + `hidden`. (The Vortex file *writer* rejects duplicate field names, but does not reject NUL/control characters in names, so such a file is producible — and any non-Vortex producer emitting the same `DType` would trip this too.)
2. Open it and scan through the Arrow export path (`DataSource::open` → `scan` → Arrow C schema export).

The process aborts during schema export.

## Backtrace (abridged)

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:225:5:
panic in a function that cannot unwind
...
  10: core::panicking::panic_nounwind_fmt
  11: core::panicking::panic_nounwind
  12: core::panicking::panic_cannot_unwind
  13: arrow_array::ffi_stream::get_schema
  14: Java_org_apache_arrow_c_jni_JniWrapper_getSchemaArrayStream
thread caused non-unwinding panic. aborting.
```

The proximate cause looks like a `CString::new` on the field name inside `get_schema` failing on the interior NUL, in a context declared not to unwind — so the failure escalates to `panic_cannot_unwind` → abort instead of surfacing as a `Result`.

## Expected

Opening or exporting the file should fail with a recoverable `VortexError` / `ArrowError`, not abort the process. Either reject NUL (and arguably other control characters) in field names at the boundary where a `DType`/schema is constructed or exported, or handle the `CString::new` error inside `get_schema` without crossing a non-unwinding boundary.

## Context

Found while hardening cross-compatibility for the pure-Java implementation ([vortex-java](https://github.com/dfa1/vortex-java), intro in #8250). vortex-java now rejects NUL/blank/control field names on both its writer and reader by policy, precisely so it never emits a file that aborts the reference toolchain — but the upstream abort seemed worth reporting on its own. Happy to provide the exact reproducer file or a standalone test if useful.

miniex:
## Summary

A struct field name with an interior NUL byte aborted the whole process (`SIGABRT`) when the schema was exported across the Arrow C Data FFI: arrow-rs copies the name into a `CString` in a non-unwinding context, so `CString::new` failing on the NUL escalated to an abort instead of an error (#8652). `to_arrow_schema` / `to_data_type_naive` and `ArrowSession::to_arrow_field` now reject a NUL byte in a field name up front, so a crafted or mislabeled file surfaces a recoverable `VortexError` instead of taking down the reader.

This rejects only NUL, the byte that actually aborts (other control characters are valid in a `CString`), and guards the Arrow export boundary rather than the writer, so it also covers a `DType` from a non-Vortex producer.

Closes: #8652

## Testing

Added tests that `to_arrow_schema` (top-level and nested struct) and `ArrowSession::to_arrow_field` reject a NUL field name. `cargo nextest run -p vortex-array` passes; `fmt --all` + `clippy --all-targets --all-features` clean.

---

### DIFF
--- a/vortex-array/src/dtype/arrow.rs
+++ b/vortex-array/src/dtype/arrow.rs
@@ -273,6 +273,7 @@ impl DType {
 
         let mut builder = SchemaBuilder::with_capacity(struct_dtype.names().len());
         for (field_name, field_dtype) in struct_dtype.names().iter().zip(struct_dtype.fields()) {
+            validate_arrow_field_name(field_name.as_ref())?;
             let field = if field_dtype.is_variant() {
                 let storage = DataType::Struct(variant_storage_fields_minimal());
                 Field::new(field_name.as_ref(), storage, field_dtype.is_nullable()).with_metadata(
@@ -306,6 +307,15 @@ impl DType {
     pub fn to_arrow_dtype(&self) -> VortexResult<DataType> {
         to_data_type_naive(self)
     }
+}
+
+/// Reject a field name Arrow cannot export: an interior NUL byte aborts the process during FFI
+/// schema export, where the name is copied into a NUL-terminated `CString`.
+pub(crate) fn validate_arrow_field_name(name: &str) -> VortexResult<()> {
+    if name.contains('\0') {
+        vortex_bail!("field name {name:?} cannot be exported to Arrow: contains a NUL byte");
+    }
+    Ok(())
 }
 
 /// Naive conversion from a Vortex `DType` to the nearest Arrow physical data type.
@@ -361,6 +371,7 @@ pub(crate) fn to_data_type_naive(dtype: &DType) -> VortexResult<DataType> {
         DType::Struct(struct_dtype, _) => {
             let mut fields = Vec::with_capacity(struct_dtype.names().len());
             for (field_name, field_dt) in struct_dtype.names().iter().zip(struct_dtype.fields()) {
+                validate_arrow_field_name(field_name.as_ref())?;
                 fields.push(FieldRef::from(Field::new(
                     field_name.as_ref(),
                     to_data_type_naive(&field_dt)?,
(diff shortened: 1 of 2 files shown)


vortex-array/src/dtype/arrow.rs
```diff
@@ -308,6 +309,15 @@ impl DType {
...
 }
 
+/// Reject a field name Arrow cannot export: an interior NUL byte aborts the process during FFI
+/// schema export, where the name is copied into a NUL-terminated `CString`.
+pub(crate) fn validate_arrow_field_name(name: &str) -> VortexResult<()> {
+    if name.contains('\0') {
+        vortex_bail!("field name {name:?} cannot be exported to Arrow: contains a NUL byte");
+    }
+    Ok(())
+}
```
joseph-isaacs, line 319: Does arrow use nul str or str + len

robert3005: I made this fix in arrow-rs https://github.com/apache/arrow-rs/pull/10328. Arrow supports names that are not valid c strings unfortunately

PR #8665, opened 2026-07-07
FFI: canonicalize array, data pointer for primitives

myrrc:
- Add canonicalize function for vx_array.
- Add data_ptr functions for canonicalized primitive and bool arrays.
- Add sink abort method which doesn't write file footer and leaves file invalid. Semantics here is we want to abort the sink in C++ destructor - if user hasn't called Finalize(), file will not be written correctly.
- Expose Rust's error code to FFI.

### DIFF
--- a/vortex-ffi/cinclude/vortex.h
+++ b/vortex-ffi/cinclude/vortex.h
@@ -202,6 +202,52 @@ typedef enum {
 } vx_estimate_type;
 
 /**
+ * Error category for vx_error.
+ */
+typedef enum {
+    /**
+     * All other errors
+     */
+    VX_ERROR_CODE_OTHER = 0,
+    /**
+     * Index out of bounds
+     */
+    VX_ERROR_CODE_OUT_OF_BOUNDS = 1,
+    /**
+     * Compute kernel execute error
+     */
+    VX_ERROR_CODE_COMPUTE = 2,
+    /**
+     * An invalid argument was provided.
+     */
+    VX_ERROR_CODE_INVALID_ARGUMENT = 3,
+    /**
+     * Serialization/deserialization error
+     */
+    VX_ERROR_CODE_SERIALIZATION = 4,
+    /**
+     * Unimplemented function
+     */
+    VX_ERROR_CODE_NOT_IMPLEMENTED = 5,
+    /**
+     * Type mismatch
+     */
+    VX_ERROR_CODE_MISMATCHED_TYPES = 6,
+    /**
+     * Assertion failed
+     */
+    VX_ERROR_CODE_ASSERTION_FAILED = 7,
+    /**
+     * IO error
+     */
+    VX_ERROR_CODE_IO = 8,
+    /**
+     * Panic inside FFI
+     */
+    VX_ERROR_CODE_PANIC = 9,
+} vx_error_code;
+
+/**
  * Equalities, inequalities, and boolean operations over possibly null values.
  * For most operations, if either side is null, the result is null.
  * VX_OPERATOR_KLEENE_AND, VX_OPERATOR_KLEENE_OR obey Kleene (three-valued)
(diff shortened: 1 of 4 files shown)


vortex-ffi/cinclude/vortex.h
```diff
@@ -201,6 +201,52 @@ typedef enum {
...
 } vx_estimate_type;
 
+/**
+ * Error category for vx_error.
+ */
+typedef enum {
+    /**
+     * All other errors
+     */
+    VX_ERROR_CODE_OTHER = 0,
```
0ax1, line 211: Should 0 be `other` or `no error`?

vortex-ffi/cinclude/vortex.h
```diff
@@ -201,6 +201,52 @@ typedef enum {
...
+     */
+    VX_ERROR_CODE_COMPUTE = 2,
+    /**
+     * An invalid argument was provided.
+     */
+    VX_ERROR_CODE_INVALID_ARGUMENT = 3,
+    /**
+     * Serialization/deserialization error
+     */
+    VX_ERROR_CODE_SERDE = 4,
```
0ax1: At the FFI layer we prob don't care about which serialization framework is used. Could consider suffixing this with SERIALIZATION or so rather than SERDE.

vortex-ffi/cinclude/vortex.h
```diff
@@ -201,6 +201,52 @@ typedef enum {
...
+     */
+    VX_ERROR_CODE_NOT_IMPLEMENTED = 5,
+    /**
+     * Type mismatch
+     */
+    VX_ERROR_CODE_MISMATCHED_TYPES = 6,
+    /**
+     * Assertion failed
+     */
+    VX_ERROR_CODE_ASSERTION_FAILED = 7,
```
0ax1, line 239: Should this be a separate case. Asserts could fire bc of reasons listed above.

vortex-ffi/cinclude/vortex.h
```diff
@@ -803,6 +849,43 @@ const vx_string *vx_array_get_utf8(const vx_array *array, uint32_t index);
  */
 const vx_binary *vx_array_get_binary(const vx_array *array, uint32_t index);
 
+/**
+ * For a canonical Bool array, return bit at "index".
```
0ax1: Empty line after first sentence?

vortex-ffi/cinclude/vortex.h
```diff
@@ -803,6 +849,43 @@ const vx_string *vx_array_get_utf8(const vx_array *array, uint32_t index);
...
+ * For invalid elements returned value is unspecified, check validity via
+ * vx_array_get_validity.
+ *
+ * Panics if "array" is not canonical - call vx_array_canonicalize first.
```
0ax1, line 857: Nit: We talk about a bool(not bit) array but then also about returning a bit at index.

vortex-ffi/cinclude/vortex.h
```diff
@@ -803,6 +849,43 @@ const vx_string *vx_array_get_utf8(const vx_array *array, uint32_t index);
...
+ * Panics if "array" is not a Bool array.
+ * Panics if "index" is out of bounds.
+ */
+bool vx_array_get_bool(const vx_array *array, size_t index);
+
+/**
+ * Decode array into its canonical (uncompressed) encoding.
```
0ax1: Why add `uncompressed` here in the doc?

vortex-ffi/src/array.rs
```diff
@@ -474,6 +482,100 @@ pub unsafe extern "C-unwind" fn vx_array_get_binary(
...
+/// For invalid elements returned value is unspecified, check validity via
+/// vx_array_get_validity.
+///
+/// Panics if "array" is not canonical - call vx_array_canonicalize first.
+/// Panics if "array" is not a Bool array.
+/// Panics if "index" is out of bounds.
+#[unsafe(no_mangle)]
+pub unsafe extern "C-unwind" fn vx_array_get_bool(array: *const vx_array, index: usize) -> bool {
+    let array = vx_array::as_ref(array);
+    let bool_array = array
```
0ax1, line 495: What is the policy rn in the ffi layer, panic on error explicitly? When do we return errors per output param, when do we crash in the general sense.

0ax1 (commented):
nits and questions

vortex-ffi/cinclude/vortex.h
myrrc: If there's no error, returned vx_error is NULL so we can't have a valid vx_error with "no error" code.

vortex-ffi/cinclude/vortex.h
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: We have this as a separate type in VortexError, I'm just exposing the contract. But I agree the general error type "assertion" shouldn't be present.

vortex-ffi/cinclude/vortex.h
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: +1

vortex-ffi/cinclude/vortex.h
myrrc: +1

vortex-ffi/src/array.rs
myrrc: Yet to be formalized, but generally accessors should panic on error and other functons should take an error_out.

0ax1 (approved):

PR #8668, opened 2026-07-07
FFI: replace vx_string/vx_binary with vx_view

myrrc:
vx_string/vx_binary is not a zero-copy abstraction. For every utf/binary view you allocate memory. This PR replaces both with a vx_view non-owning view akin to C++'s string_view.


vortex-ffi/cinclude/vortex.h
```diff
@@ -595,17 +585,34 @@ typedef struct {
...
+typedef struct {
+    /**
+     * NULL "ptr" requires len == 0
+     */
+    const char *ptr;
+    /**
+     * Length in bytes.
+     */
+    size_t len;
+} vx_str;
```
0ax1: `vx_bytes` then?

vortex-ffi/cinclude/vortex.h
```diff
@@ -595,17 +585,34 @@ typedef struct {
...
+     * Required: paths to files, tables, or layout trees. Each entry may be a
+     * glob pattern like "*.vortex". Must point to an array of size
+     * "paths_len". paths bytes are copied.
+     */
+    const vx_str *paths;
+    /**
+     * Number of entries in `paths`.
      */
-    const char *paths;
+    size_t paths_len;
```
0ax1, line 615: We can have multiple paths right, but only one length?

0ax1 (commented):
What is the value of splitting these types. It seems we now use str also for buffers that are just bytes.

vortex-ffi/cinclude/vortex.h
myrrc: Yes, this is the outer length. Inidividual lengths are tracked in each vx_str.

0ax1 (approved):
lgtm but lets rename to vx_view or span as discussed.

PR #8669, opened 2026-07-07
sccache cleanup

AdamGS:
## Rationale for this change

Make better use of sccache across our GH actions workflows.

1. Makes sccache opt in
2. When sccache is used, make sure its properly configured with runs-on.

None


0ax1:
Why should this not be turned on by default?

0ax1:
curious about a 2nd opinion here. @robert3005

0ax1 (approved):
in the name of consistency.

AdamGS:
c/cpp runs into some issues with our cmakelists, will fix in a followup.

PR #8674, opened 2026-07-07
Add TakeSlices for fixed-size-list take

danking:
The key issue here is that `fixed_size_list.take(indices)` explodes the indices into one index per _list element_. For a fixed-size list of length 4096, this means the number of take indices expands by three orders of magnitude. Moreover, the take kernel must either copy each index separately or reconstruct the fact that these are actually contiguous ranges so it can use range copies.

This PR now models that operation directly as a gather of contiguous runs:

```rust
fn take_slices(&self, starts: ArrayRef, lengths: ArrayRef) -> VortexResult<ArrayRef>;
```

`starts[i]` and `lengths[i]` describe each child run, and the output is the concatenation of `values[starts[i]..starts[i] + lengths[i]]`. The runs may overlap, repeat, and appear in any order; this is a gather of runs, not a normal slice.

For `FixedSizeList<100, u8>`, this:

```rust
fixed_size_list.take([10, 100, 0])
```

now becomes run selectors over the child elements:

```rust
starts  = [1000, 10000, 0]
lengths = [100, 100, 100]
```

instead of expanding to:

```rust
fixed_size_list.elements().take([1000, 1001, ..., 1099, 10000, 10001, ..., 10099, 0, 1, ..., 99])
```

This also gives `ListView -> List` packing a direct fast path: `ListView` rebuild now uses original offsets as starts and normalized sizes as lengths, so it can gather the referenced element runs without building a flat per-element take index buffer or falling back to per-list builders. `Struct` now pushes those run selectors into each field, and `VarBinView` now executes run gathers by copying contiguous view ranges while reusing the original data buffers.

This is a table of results for `FixedSizeList<N, f16>`:

list size | bytes/list | indices | per-index | take_slices | manual range copy | `per-index / take_slices`
-- | -- | -- | -- | -- | -- | --
1 | 2 B | 100 | 1.17 us | 1.04 us | 666 ns | 1.12x
1 | 2 B | 1000 | 3.08 us | 4.00 us | 2.42 us | 0.77x
16 | 32 B | 100 | 4.33 us | 1.21 us | 661 ns | 3.59x
16 | 32 B | 1000 | 35.3 us | 5.54 us | 2.71 us | 6.37x
256 | 512 B | 100 | 55.9 us | 1.83 us | 1.35 us | 30.5x
256 | 512 B | 1000 | 552 us | 12.7 us | 10.1 us | 43.4x
1024 | 2048 B | 100 | 220 us | 3.87 us | 3.42 us | 56.7x
1024 | 2048 B | 1000 | 2.40 ms | 105 us | 99.6 us | 22.8x
4096 | 8192 B | 100 | 927 us | 13.5 us | 12.9 us | 68.7x
4096 | 8192 B | 1000 | 9.18 ms | 391 us | 381 us | 23.5x

---

Codex:

## Summary

- Switch `TakeSlicesArray` from a `Vec<Range<usize>>` constructor to selector arrays: non-nullable unsigned `starts` and `lengths`.
- Route `FixedSizeList` take through contiguous child runs via `take_slices`, leaving child encodings responsible for optimization.
- Add selector validation/materialization helpers for equal-length, non-nullable unsigned selectors and in-bounds runs, including zero-length runs.
- Add a primitive `TakeSlicesExecute` kernel that copies each requested run directly from the primitive slice.
- Add `Struct` `TakeSlicesReduce` so run gathers push down into each field, matching the existing `take` shape.
- Add `VarBinView` `TakeSlicesExecute` so run gathers copy view ranges directly and reuse backing data buffers.
- Use `take_slices` to rebuild non-zero-copy `ListViewArray` layouts into packed, zero-copy-to-list layouts.
- Expand `take_fsl` benchmarks and `take_slices` coverage for selector arrays, overlapping/repeated runs, empty runs, nested `TakeSlices`, nullable validity, generic fallback, primitive execution, Struct pushdown, VarBinView view-buffer execution, and FSL take.

## Correctness and Coverage

- `TakeSlices` validates public construction selectors before creating the lazy array.
- Generic `TakeSlices` execution repeatedly slices the child and appends each run; specialized children can bypass that path.
- Nullable FSL take uses zero-length runs for null output rows and keeps nullness in outer FSL validity.
- `ListView` rebuild zeroes null-list lengths before gathering elements, preserving packed offsets without copying invalid-row payloads.
- `Struct` preserves struct-level validity while applying the same run selectors to every unmasked field.
- `VarBinView` preserves validity and shares the original data-buffer `Arc` while copying only selected views.

## Validation

- `cargo +nightly fmt --all`
- `cargo test -p vortex-array take_slices`
- `cargo test -p vortex-array fixed_size_list::tests::take`
- `cargo test -p vortex-array listview`
- `cargo check -p vortex-array --all-targets`
- `git diff --check`


danking:
@claude @codex please review.

vortex-array/src/arrays/take_slices/array.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Constructs a new validated `TakeSlicesArray` using the caller's execution context to
+    /// validate selector arrays and compute the output length.
+    pub fn try_new_with_ctx(
+        child: ArrayRef,
+        starts: ArrayRef,
+        lengths: ArrayRef,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Self> {
+        let len = selector_output_len_with_ctx(child.len(), &starts, &lengths, ctx)?;
+        Ok(Self::new_validated(child, starts, lengths, len))
```
robert3005: I don't think we should do this validation on construction. As long as starts, lengths are correct size then starts being out of bounds should be an execution error

vortex-array/src/arrays/take_slices/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+impl RunSelectors {
+    pub(super) fn new(
+        child_len: usize,
+        starts: &ArrayRef,
+        lengths: &ArrayRef,
+        ctx: &mut ExecutionCtx,
+    ) -> VortexResult<Self> {
+        check_selector_arrays(starts, lengths)?;
+        let starts = starts.clone().execute::<PrimitiveArray>(ctx)?;
+        let lengths = lengths.clone().execute::<PrimitiveArray>(ctx)?;
```
robert3005: this really should account for constant arrays, that's what you end up with for fixedsizelist

danking:
@robert3005 I believe I've addressed your comments. The benchmarks still aren't as good as I had hoped though.

robert3005:
I have some ideas how to make it faster on top of it. Especially for chunked array if you know you have FSL(Chunked(Primitive)) you can optimise the Chunked(Primitive) flow quite a lot. We can land that separately. I think the only remaining part is to reduce the size of codspeed benchmarks. We aim for less than 1ms of runtime

danking:
I get different results locally for the supposedly slower benchmarks:

Benchmark | develop local | HEAD local | Local result
-- | -- | -- | --
copy_nullable[65536] | 60.58 µs | 49.79 µs | HEAD ~18% faster
copy_non_nullable[65536] | 62.93 µs | 50.99 µs | HEAD ~19% faster
chunked_varbinview_canonical_into[(1000, 10)] | 6.208 µs | 5.166 µs | HEAD ~17% faster
rebuild_naive | 7.406 µs | 7.657 µs | HEAD ~3% slower
slice_empty_vortex | 2.903 ns | 2.410 ns | HEAD ~17% faster
varbinview_large | 5.937 µs | 3.708 µs | HEAD ~38% faster

</div></div>

robert3005:
yeah, codspeed is very sensitive to memory allocations

vortex-array/src/arrays/take_slices/vtable.rs
```diff
@@ -0,0 +1,308 @@
...
+        _buffers: &[BufferHandle],
+        _children: &dyn ArrayChildren,
+        _session: &VortexSession,
+    ) -> VortexResult<ArrayParts<Self>> {
+        vortex_bail!("TakeSlices array is not serializable")
+    }
+
+    fn execute(array: Array<Self>, ctx: &mut ExecutionCtx) -> VortexResult<ExecutionResult> {
+        let parent = array.clone().into_array();
+        if let Some(reduced) = array.child().reduce_parent(&parent, CHILD_SLOT)? {
```
robert3005: You don't need this, the outer loop will only invoke you if there's no reduce_parent/execute_parent rule, this is the final catch all

vortex-array/src/arrays/take_slices/vtable.rs
```diff
@@ -0,0 +1,308 @@
...
+    lengths: &[L],
+    output_len: usize,
+    builder: &mut dyn ArrayBuilder,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<()>
+where
+    S: IntegerPType,
+    L: IntegerPType,
+{
+    validate_index_ranges(child.len(), starts, lengths, output_len)?;
```
robert3005: slice will validate this, we don't have unchecked slice so this is double work

robert3005:
Fwiw I pushed a commit that reduced size of codspeed tests

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -145,77 +183,168 @@ fn take_non_nullable_fsl<I: IntegerPType, E: IntegerPType>(
...
+    let starts = PrimitiveArray::from_iter(starts).into_array();
+    let lengths = ConstantArray::new(length as u64, run_count).into_array();
+
+    // SAFETY: callers produced one start per output row after validating list indices against the
+    // source FSL length. `length` is the fixed list size, represented as a non-nullable unsigned
+    // constant array, and `output_len` was computed as `run_count * length`.
+    Ok(
+        unsafe { TakeSlicesArray::new_unchecked(elements.clone(), starts, lengths, output_len) }
+            .into_array()
+            .execute::<RecursiveCanonical>(ctx)?
```
robert3005: we shouldn't do this

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -145,77 +183,168 @@ fn take_non_nullable_fsl<I: IntegerPType, E: IntegerPType>(
...
+    length: usize,
+    output_len: usize,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<ArrayRef> {
+    let run_count = starts.len();
+    let starts = starts
+        .into_iter()
+        .map(|start| start as u64)
+        .collect::<Vec<_>>();
+    let starts = PrimitiveArray::from_iter(starts).into_array();
```
robert3005: hmmm?

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -145,77 +183,168 @@ fn take_non_nullable_fsl<I: IntegerPType, E: IntegerPType>(
...
 
+    // SAFETY: `new_elements` has `new_len * list_size` elements. `new_validity_builder` appends
+    // exactly one bit per output row, and `Validity::from` preserves that length when needed.
     Ok(unsafe {
         FixedSizeListArray::new_unchecked(new_elements, array.list_size(), new_validity, new_len)
     }
     .into_array())
 }
+
+fn indices_validity_mask(
```
robert3005: there's validity().execute_mask

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -2,138 +2,176 @@
...
-                    new_validity,
-                    new_len,
-                )
-            }
-            .into_array(),
+        return take_degenerate_fsl::<I>(array, indices, indices_array.as_view(), ctx);
+    }
+
+    if array.is_empty() {
+        return take_empty_non_degenerate_fsl::<I>(array, indices, indices_array.as_view(), ctx);
```
robert3005: I think you can handle this generically, all you need to do is to check that indices are also empty. I think this could already be happening

vortex-array/src/arrays/fixed_size_list/compute/take.rs
```diff
@@ -2,138 +2,176 @@
...
+    array: ArrayView<'_, FixedSizeList>,
+    indices_array: ArrayView<'_, Primitive>,
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<ArrayRef> {
+    let list_size = array.list_size() as usize;
+    let array_len = array.as_ref().len();
+    let indices: &[I] = indices_array.as_slice::<I>();
+    let new_len = indices.len();
+    let expected_elements_len = take_elements_len(new_len, list_size)?;
+    let mut starts = Vec::with_capacity(new_len);
```
robert3005: you should make it just a Buffer

vortex-array/src/arrays/take_slices/vtable.rs
```diff
@@ -0,0 +1,308 @@
...
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<()> {
+    let starts = array.starts();
+    let lengths = array.lengths();
+    check_index_arrays(starts, lengths)?;
+
+    match_each_unsigned_integer_ptype!(starts.dtype().as_ptype(), |S| {
+        match_each_unsigned_integer_ptype!(lengths.dtype().as_ptype(), |L| {
+            let starts = starts.clone().execute::<PrimitiveArray>(ctx)?;
+            let lengths = lengths.clone().execute::<PrimitiveArray>(ctx)?;
```
robert3005, line 240: I think this should still differentiate constant arrays

vortex-array/src/arrays/varbinview/compute/take_slices.rs
```diff
@@ -0,0 +1,106 @@
...
+    L: IntegerPType,
+{
+    validate_index_ranges(source.len(), starts, lengths, output_len)?;
+
+    let mut views = BufferMut::<BinaryView>::with_capacity(output_len);
+    for (&start, &length) in starts.iter().zip_eq(lengths) {
+        let start = index_value_to_usize("start", start)?;
+        let length = index_value_to_usize("length", length)?;
+        let end = start + length;
+        views.extend_from_slice(&source[start..end]);
```
robert3005: you can change this pattern to &source[start..][..length] but I don't feel too strongly about it

vortex-array/src/arrays/decimal/compute/take_slices.rs
```diff
@@ -0,0 +1,126 @@
...
+    L: IntegerPType,
+{
+    validate_index_ranges(source.len(), starts, lengths, output_len)?;
+
+    let mut values = BufferMut::<T>::with_capacity(output_len);
+    for (&start, &length) in starts.iter().zip_eq(lengths) {
+        let start = index_value_to_usize("start", start)?;
+        let length = index_value_to_usize("length", length)?;
+        let end = start + length;
+        values.extend_from_slice(&source[start..end]);
```
robert3005: since you validated them above this could all use unchecked

vortex-array/src/arrays/take_slices/array.rs
```diff
@@ -0,0 +1,93 @@
...
+pub(super) const CHILD_SLOT: usize = 0;
+/// The start index for each child run.
+pub(super) const STARTS_SLOT: usize = 1;
+/// The length for each child run.
+pub(super) const LENGTHS_SLOT: usize = 2;
+pub(super) const NUM_SLOTS: usize = 3;
+pub(super) const SLOT_NAMES: [&str; NUM_SLOTS] = ["child", "starts", "lengths"];
+
+/// Extension methods for [`TakeSlices`] arrays.
+pub trait TakeSlicesArrayExt: TypedArrayRef<TakeSlices> {
```
joseph-isaacs: this is old, please use the proc macro

joseph-isaacs (changes requested):
This could be implemented using `Take(values, codes=runend_array)` I was to be sure before we merge this that using a new Array and kernel is preferred

robert3005:
@joseph-isaacs this cannot be implemented using your suggestion. There’s a long conversation @danking and I had on this topic. This is more like codes=Chunked(Sequence)

vortex-array/src/arrays/fixed_size_list/compute/take.rs
danking: OK: https://github.com/vortex-data/vortex/pull/8674/changes/33a4fc3fba8b2b743bb7fa068a1b84c28437d6e9

vortex-array/src/arrays/fixed_size_list/compute/take.rs
danking: OK, this all changed. We handle nulls, bounds check, and convert from a list index to an element start index in one loop. Easiest thing is to look at latest commit: https://github.com/vortex-data/vortex/pull/8674/changes#diff-f0d52d538d5c7d147ab1d1fb20188562eadf08f5c7457b6c06175aad190c4284R195

vortex-array/src/arrays/fixed_size_list/compute/take.rs
danking: Done.

vortex-array/src/arrays/fixed_size_list/compute/take.rs
danking: Great point. `take_empty_fsl` now doesn't do any type dispatch.

vortex-array/src/arrays/fixed_size_list/compute/take.rs
danking: Great point, we now collect::<BufferMut<u64>>()

vortex-array/src/arrays/take_slices/vtable.rs
danking: OK, this is added but I'm skeptical all this extra code is worth it.

vortex-array/src/arrays/varbinview/compute/take_slices.rs
danking: done https://github.com/vortex-data/vortex/pull/8674/changes#diff-15c5d6c5ee3ded1d863ce5642dbc9b6f298663bd2c1d6523e370bf236e79d221R101

vortex-array/src/arrays/decimal/compute/take_slices.rs
danking: Done.

vortex-array/src/arrays/take_slices/array.rs
danking: Done.

danking:
@robert3005 @joseph-isaacs All comments addressed. I'm looking at the benchmarks results now.

danking:
Codex addressed the benchmarks in ways that don't seem copacetic. I need to carefully review before we merge.

vortex-array/src/arrays/take_slices/vtable.rs
```diff
@@ -149,6 +149,13 @@ impl VTable for TakeSlices {
     }
 
     fn execute(array: Array<Self>, ctx: &mut ExecutionCtx) -> VortexResult<ExecutionResult> {
+        if let Some(reduced) = array
+            .child()
+            .reduce_parent(array.as_ref(), TakeSlicesSlots::CHILD)?
```
robert3005, line 154: why wouldn't reduce rule help us here?

robert3005:
I think what Joe suggested for this pr is to remove TakeSlices array, have a SlicesArray and add generic rule dispatch (like TakeExecute/TakeReduce) that unpacks the take argument to SlicesArray if possible and lets you do the TakeSlices logic as it is no

vortex-array/src/arrays/take_slices/vtable.rs
joseph-isaacs, line 1: I thought we decided against this array

danking: cc: @robert3005 I'm not sure I follow you two. Is the objection to the name "TakeSlices"?

This array solves a very specific problem which is: efficiently implement `FixedSizeListArray.take([x0, ..., xn])`.

Currently, this produces `N_indices * LIST_LENGTH` individual index lookups. This is slow.

My first solution was to modify `FSL.take` to (1) canonicalize its elements and indices, and (2) loop over indices copying slices of primitive elements into a builder. This solution works but feels like the Wrong Way to do this?

I cannot reduce `Take(FSL, indices)` into `Take(elements, codes=runend_array)` because the element indices are not runs. For example, consider an FSL of length 3 and the take indices `[1, 5, 7]`. The element take indices are: `3, 4, 5,    15, 16, 17,   21, 22, 23`. This is like, PiecewiseSequential or something. Are you arguing in favor of creating a `PiecewiseSequential` array and modifying all the canonical array's `TakeExecute` implementation to check for PiecewiseSequential and use a slice-based algorithm?

danking: Codex is working on rebuilding this PR as two PRs: (1) PiecewiseSequentialArray and (2) TakeExecute specialization on PSA.

danking:
Replaced by:

1. https://github.com/vortex-data/vortex/pull/8799
2. https://github.com/vortex-data/vortex/pull/8800

PR #8676, opened 2026-07-07
Add `list_sum` expression

mhk197:
Adds a `vortex.list.sum` scalar function: the sum of each row's list elements, one output row per input row — the equivalent of DuckDB's `list_sum()` and DataFusion's `array_sum()`.

Note that list entries that are empty or all null need to be nullified. However, `GroupedAccumulator` returns 0 in these cases, so we need to either:
1. Change the behavior of `GroupedAccumulator` and grouped sum (or even sum) kernels or 
2. Do a second pass of list element validity and sizes, construct a mask for the sums, and apply it.

The first option is probably best but can be applied in a followup PR.


vortex-array/src/scalar_fn/fns/list_sum/mod.rs
```diff
@@ -0,0 +1,574 @@
...
+    ctx: &mut ExecutionCtx,
+) -> VortexResult<ArrayRef> {
+    let elem_dtype = match array.dtype() {
+        DType::List(elem, _) | DType::FixedSizeList(elem, ..) => elem.as_ref().clone(),
+        other => vortex_bail!("list_sum() requires List or FixedSizeList, got {other}"),
+    };
+
+    // Canonicalize upfront because `mask_empty_lists` needs access to list elements validity
+    // and sizes.
+    let canonical = array.clone().execute_until::<AnyCanonical>(ctx)?;
```
myrrc: I think we want AnyColumnar here and handle constant/canonical version at once

myrrc (commented):
Let's push list_sum down to Duckdb and write a sqllogictest for this (can be in a followup PR). I'd also like to see a performance benchmark query, I'll prepare the machinery for that.

vortex-array/src/scalar_fn/fns/list_sum/mod.rs
robert3005: Ah, this is handled above in execute

robert3005: ok no, the code above will be forced to canonicalise

mhk197: ah yeah we want to canonicalize a 1 row constant array, do sum, and then broadcast. will fix

robert3005 (approved):

PR #8677, opened 2026-07-07
Audit panic handling in detached tasks

AdamGS:
## Rationale for this change

Following #8672, I've audited a few other suspicous code paths that had similar shape, mostly on the read side.

1. Propagate panics and shutdown correctly for `FileSegmentSource`.
2. Handle panics correctly in the shared iterator used by DuckDB
3. Misc panic handling around our runtime abstraction. 

None

### DIFF
--- a/vortex-duckdb/src/table_function.rs
+++ b/vortex-duckdb/src/table_function.rs
@@ -14,10 +14,8 @@ use std::task::Poll;
 
 use custom_labels::CURRENT_LABELSET;
 use futures::FutureExt;
-use futures::SinkExt;
 use futures::Stream;
 use futures::StreamExt;
-use futures::channel::mpsc;
 use futures::future::BoxFuture;
 use itertools::Itertools;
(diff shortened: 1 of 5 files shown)


vortex-duckdb/src/table_function.rs
```diff
@@ -222,7 +230,7 @@ pub fn init_global(init_input: &TableInitInput) -> VortexResult<TableFunctionGlo
 
     // We create an async bounded channel so that all thread-local workers can pull the next
     // available array chunk regardless of which partition it came from.
-    let (tx, rx) = kanal::bounded_async(num_workers * 2);
+    let (tx, rx) = mpsc::channel(num_workers * 2);
```
robert3005: why the change?

AdamGS: codex had some stupid reason, reverted

robert3005 (approved):

PR #8679, opened 2026-07-07
feat(geo): add native `LineString`, `MultiPoint`, `MultiLineString` types

HarukiMoriarty:
Complete the set of OGC simple-feature geometries alongside the existing Point/Polygon/MultiPolygon:

- vortex.geo.linestring       - List<Struct<x, y[, z][, m]>>
- vortex.geo.multipoint       - List<Struct<x, y[, z][, m]>>
- vortex.geo.multilinestring  - List<List<Struct<x, y[, z][, m]>>>

Each mirrors the existing Polygon/MultiPolygon extension types.

Wire them through the DuckDB integration so they surface as GEOMETRY like the existing native types: the dtype -> GEOMETRY mapping, per-type WKB exporters, and native-geometry-column recognition for predicate pushdown.

Tests cover dtype validation across all four dimensions, Arrow field import/export (including that the shape-identical types resolve by name), and WKB-literal decode round-trips.


myrrc (approved):
LGTM for duckdb part

danking (approved):

PR #8683, opened 2026-07-08
ci: remove v3 benchmark ingest, make v4 Postgres ingest required

connortsui20:
## What

Removes the v3 (EC2/DuckDB) leg of the benchmark-results dual-write and promotes the v4 (RDS Postgres) ingest from best-effort to required.

- Deletes the three "Ingest ... to v3 server" steps (`bench.yml`, `sql-benchmarks.yml`, `commit-metadata.yml`), each gated on `vars.V3_INGEST_URL`.
- Removes `continue-on-error: true` from the v4 ingest step groups so a v4 ingest failure now fails the job (and alerts incident.io on develop).
- Rewrites the stale soak-era comments, which claimed v3 was hard-required and v4 best-effort — reality is now inverted.

## Why

The benchmarks website fully cut over to v4 (Vercel + RDS Postgres) on 2026-06-19 and has been stable since; the pre-cutover history was backfilled into v4 on 2026-07-07, so v3 holds nothing v4 lacks. This unblocks terminating the v3 EC2 host.

## Deliberately untouched

- The `--gh-json-v3 results.v3.jsonl` emitters, `scripts/post-ingest.py` (including its now-unused `--server` mode), and `scripts/_measurement_id.py` — the v3 JSONL format **is** v4's wire format; only the HTTP-to-EC2 leg is removed.
- The v2 legacy S3 uploads (`cat-s3.sh`, commit-metadata job in `bench.yml`) — still consumed by PR-benchmark comparison; separate teardown.

After merge, the repo-level `V3_INGEST_URL` var and `INGEST_BEARER_TOKEN`/`ADMIN_BEARER_TOKEN` secrets can be deleted (every removed step was gated on the var, so ordering is safe).

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -117,26 +117,11 @@ jobs:
         run: |
           bash scripts/cat-s3.sh vortex-ci-benchmark-results data.json.gz results.json
 
-      - name: Ingest results to v3 server
-        if: vars.V3_INGEST_URL != ''
-        shell: bash
-        env:
-          INGEST_BEARER_TOKEN: ${{ secrets.INGEST_BEARER_TOKEN }}
-        run: |
-          python3 scripts/post-ingest.py results.v3.jsonl \
-            --server "${{ vars.V3_INGEST_URL }}" \
-            --commit-sha "${{ github.sha }}" \
-            --benchmark-id "${{ matrix.benchmark.id }}" \
-            --repo-url "${{ github.server_url }}/${{ github.repository }}"
-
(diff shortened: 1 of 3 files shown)


AdamGS:
I have questions but I need to read it more closely first

connortsui20:
note that this is just deleting my experimental v3 stuff so it has nothing to do with the actual running server

AdamGS:
change seems reasonable, just running benchmarks as a sanity check

AdamGS (approved):

PR #8687, opened 2026-07-08
ci: pin the `measurement_id` hash with golden vectors in-repo

connortsui20:
## Summary

Right now, there is no way for us to ensure that our `measurement_id` key that we generate in CI matches the ones that we have on the [`benchmarks-website`](https://github.com/vortex-data/benchmarks-website) code.

This adds a bunch of "golden" hashes so that we can immediately see if something changes here that would break the contract with the benchmarks website server.

Merge this before vortex-data/benchmarks-website#7 so the hash is never unpinned.


0ax1:
@claude review

0ax1:
Can you make the PR desc a bit more concise (and maybe high level)?

0ax1 (approved):

Issue #7697, opened 2026-04-28
Epic: Compressor Improvements


connortsui20:
This epic is for making improvements to our compressor (including our default compressor, which is just all of our encodings/schemes + our sampling compressor).

## Status

**Proposed.**

We know that we want to improve this but it is not super clear what are the most meaningful changes we can make right now.

## Motivation

We made several large structural changes to the compressor (tracked in https://github.com/vortex-data/vortex/issues/7216), which both made the sampling compressor pluggable and also made observability into what the compressor is doing easier (perfetto tracing and a clearer sampling algorithm).

Those improvements gave a few insights in different ways we _could_ optimize the compressor, but it is unclear the degree to which those improvements would help.

Below are a list of ideas that I have that I think could be improvements to the compressor, in no particular order.

## TODO make subissues for these things

- [ ] Scheme priorities (i.e., trying one scheme before another)
- [ ] More exclusion rules (with more formal proofs on why they are correct)
- [ ] Hardcode `DictScheme` logic into the compressor? (likely more controversial...)
- [ ] Fine-grained stats collection. Not all schemes need all stats, we can be even more lazy with stats generation.
- [ ] Some scheme estimation has to compress the entire array to see if it makes sense (like `SequenceScheme`, but then we throw away the entire array and recompress later. That can definitely be fixed.
- [ ] Replace hardcoded `MAX_CASCADE = 3` with adaptive cascading that leverages the exclusion-bounded search tree (this is a stretch goal as the search space is potentially exponential)
- [ ] Unify stats (this is likely linked to a much larger change)

## Unresolved questions

- [ ] Should `MAX_CASCADE` be replaced with an adaptive search strategy now that the exclusion system bounds the search space?
- [ ] Is there a way we can improve FSST scheme estimation and compression? We currently sample for FSST compression every time, which is super expensive. But I tried being smarter with FSST estimation (try to not sample every single time and predict when FSST is good) but that basically only caused regressions (see https://github.com/vortex-data/vortex/pull/7611). It is unclear to me if there is a better way to quickly check if FSST is good for the whole array.
- [ ] Because we sample to 1024 values, it is possible for null-dominated sparse arrays to cause sampling to return 0 non-null values, which means the sampled compression ratio is technically infinite. This is obviously wrong, and it is extra wrong because the sampler will assume the entire array is all-null, which we check far in advance. It is unclear what the best solution to this is, see https://github.com/vortex-data/vortex/issues/7268.

## Implementation history

[...]

PR #8697, opened 2026-07-09
Add golden-corpus determinism tests for the default compressor

### LINKED ISSUE #6171: Better compressor tests
We should have a way to test that compression works as expected. Right now, all we can really do is assert that the display metadata string is as expected, which is not ideal.

It would be nice to assert for example that there is a specfic array node in the array tree, or even that the array tree does not contain something.

cc @joseph-isaacs

connortsui20:
Tracking Issue: https://github.com/vortex-data/vortex/issues/8701
Closes: https://github.com/vortex-data/vortex/issues/6171

## Rationale for this change

Groundwork for making the compressor's selection policy pluggable (#7697): before any refactoring of scheme selection, pin the compressor's current decisions with reviewable golden snapshots, so each subsequent PR can demonstrate unchanged selection behavior by showing zero snapshot churn. Test-only, and lands first so later PRs diff against a committed baseline.

This also closes #6171 by adding dedicated regression coverage for whether the compressor produces the expected array tree, across a representative corpus rather than one-off encoding assertions.

A golden test suite in `vortex-btrblocks`: 20 seed-generated corpus entries spanning the registered schemes' habitats, each compressed twice per run (direct determinism check) and snapshotted as its full encoding tree with exact array-level byte counts. Three feature variants cover the default scheme pool, `unstable_encodings`, and `with_compact` (+`zstd`+`pco`).

A dedicated `btrblocks-golden` CI job runs all three feature combinations explicitly. The variants are mutually exclusive at compile time (`unstable_encodings` changes `ALL_SCHEMES`), so the existing `--all-features` workspace test jobs compile the `default` variant out, and it would otherwise only run incidentally in the default-features musl job.

OnPair is excluded from the golden compressors: its dictionary training is nondeterministic run-to-run (randomly-seeded hash maps in the upstream crate), so it cannot serve as a baseline. That's a pre-existing bug worth its own issue.

None (tests and CI only).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -454,7 +454,7 @@ jobs:
 
   btrblocks-golden:
     name: "Rust (btrblocks golden corpus)"
-    timeout-minutes: 30
+    timeout-minutes: 10
     runs-on: >-
       ${{ github.repository == 'vortex-data/vortex'
           && format('runs-on={0}/runner=amd64-medium/image=ubuntu24-full-x64-pre-v2/extras=s3-cache/tag=btrblocks-golden', github.run_id)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,6 +452,38 @@ jobs:
           alert-title: "Rust tests (linux-arm64) failed on develop"
           deduplication-key: ci-rust-test-linux-arm64-failure
 
+  btrblocks-golden:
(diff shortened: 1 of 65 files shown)


robert3005 (approved):
Not sure if insta is the right way of asserting it but don't have a better idea without inventing whole new format

joseph-isaacs:
I would like us to be less strict about the golden files. 

Ideally we would support wild cards or something

PR #8698, opened 2026-07-09
Extract scheme-selection policy into a pluggable CostModel

connortsui20:
Tracking Issue: https://github.com/vortex-data/vortex/issues/8701

## Rationale for this change

The compressor's selection objective — maximize estimated compression ratio — is hardcoded into the selection loop, so it cannot be varied per workload (#7697). This PR separates candidate evaluation from selection policy: schemes produce model-independent candidate evidence, the configured `CostModel` converts each complete `Candidate` into `Cost`, and the compressor selects the lowest-cost candidate.

The default `SizeCost` preserves today's ratio ordering and canonical-acceptance threshold. It computes cost directly from the exact ratio evidence (`Cost(-ratio)`, canonical baseline `Cost(-1.0)`) rather than recomputing bytes, so IEEE rounding cannot turn a strict ratio ordering into a cost tie.

<details>

- `Scheme::evaluate` replaces the ratio-named `expected_compression_ratio` contract. It returns `SchemeEvaluation`: skip, force `AlwaysUse`, produce an opaque `CandidateEstimate`, or request a `DeferredEvaluation`.
- The compressor resolves deferred sampling/callback work, combines the resulting `CandidateEstimate` with compressor-owned context (scheme, input, optional compressed sample, and cascade ancestry), and constructs a borrowed `Candidate` for `CostModel::cost`.
- `vortex_compressor::cost` contains a finite, totally ordered `Cost`, the `CostModel` trait, the default `SizeCost`, and the model-facing `Candidate`. `Cost` is the only candidate-ranking value retained or compared by the selector; compression ratio is transient candidate evidence consumed by `SizeCost`, not selector state.
- Deferred callbacks no longer receive the current winner or a ratio threshold. Delta and Sequence therefore finish their evaluation before the active cost model decides whether their candidates are competitive. Model-aware pruning is deliberately deferred until a real non-size model establishes the lower-bound contract it needs.
- The initial model boundary and the decisions that remain outside it — constant handling, `AlwaysUse`, the final byte-acceptance gate, and extension-versus-storage selection — are documented explicitly.
- The `winner_compress` trace span gains model-defined `estimated_cost`; the old estimated-ratio winner field is removed. The achieved ratio remains output observability after compression, not selection policy.
- Selector-level tests prove that a custom model can change the winner, a deferred lower-ratio candidate reaches and can win under a custom model, candidates must beat `canonical_cost`, and sampled arrays/evidence are exposed correctly. The golden encoding-tree and size snapshots remain unchanged.

Removing the old ratio-specific deferred early exits may increase compression work for Delta and Sequence, but it does not change default winners or serialized output.

This PR does not add model-driven pruning, scheme priors, Delta-specific policy relocation, builder/session plumbing, or a production time-based model.

</details>

This adds `cost::{Cost, CostModel, SizeCost, Candidate}`, `evaluation::CandidateEstimate`, and `CascadingCompressor::with_cost_model`.

This deliberately changes the scheme-facing API: `Scheme::expected_compression_ratio` becomes `Scheme::evaluate`; `CompressionEstimate`, `EstimateVerdict`, `DeferredEstimate`, and `EstimateFn` are replaced by `SchemeEvaluation`, `ResolvedEvaluation`, `DeferredEvaluation`, and `DeferredEvaluationFn`. Deferred callbacks no longer receive a best-so-far argument. Downstream scheme implementations must migrate to the candidate-evaluation types.

The default `SizeCost` preserves existing selection ordering, canonical acceptance, encoding trees, and array sizes.


vortex-compressor/src/cost/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+//! The initial cost-model boundary covers scored candidates inside
+//! `CascadingCompressor::choose_best_scheme`. The following pre-existing decisions remain
+//! outside it:
+//!
+//! - Constant-array handling occurs before scheme selection.
+//! - [`EstimateVerdict::AlwaysUse`] is a forced-selection path that short-circuits priced
+//!   candidates.
+//! - The byte-acceptance gate: after the winning scheme compresses the full array, the result
+//!   is kept only if it is byte-wise smaller than its input. This is an axiom for **all**
+//!   models — compression never grows bytes. A model that prefers canonical (e.g. for speed)
```
connortsui20, line 26: I can't think of a case where we would have an encoding that is larger than the canonical encoding (even to speed up computation). The closest I can think of is byte bool as an encoding, but really that is like a new type.

PR #8703, opened 2026-07-09
Add `UnionVariants` type and embed it in `DType::Union`

connortsui20:
## Rationale for this change

Tracking Issue: https://github.com/vortex-data/vortex/issues/7882

We want to make `DType::Union` usable. This PR has been resurrected from https://github.com/vortex-data/vortex/pull/7902.

Replaces `DType::Union(Nullability)` with `DType::Union(UnionVariants, Nullability)`. `UnionVariants` carries a `FieldNames` list, per-variant `FieldDType`s, and an `i8` type-id vector (supporting non-consecutive tags like `[0, 5, 7]` so children can be removed without renumbering).

The flatbuffer and protobuf `Union` schemas are extended with the `names`/`dtypes`/`type_ids` fields, and serde, flatbuffers, and protobuf paths are filled in with round-trip tests and nullability-constraint checks. The flatbuffer `Union` table uses explicit field ids so that `nullable` keeps the id it had before this change, keeping the schema conformant with develop.

Existing `DType::Union(..)` match arms that reject or skip unions are left in place, and concrete implementations will come later.

`DType::Union` will now have a type parameter `UnionVariants`.

### DIFF
--- a/vortex-array/src/dtype/serde/serde.rs
+++ b/vortex-array/src/dtype/serde/serde.rs
@@ -27,6 +27,7 @@ use crate::dtype::FieldNames;
 use crate::dtype::Nullability;
 use crate::dtype::PType;
 use crate::dtype::StructFields;
+use crate::dtype::UnionVariants;
 use crate::dtype::decimal::DecimalDType;
 use crate::dtype::extension::ExtDTypeRef;
 use crate::dtype::extension::ExtId;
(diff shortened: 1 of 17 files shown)


vortex-array/src/dtype/union.rs
```diff
@@ -0,0 +1,612 @@
...
+///         DType::Primitive(PType::I32, Nullability::NonNullable),
+///         DType::Utf8(Nullability::NonNullable),
+///     ],
+/// )
+/// .unwrap();
+///
+/// assert_eq!(variants.len(), 2);
+/// assert_eq!(variants.type_ids(), &[0, 1]);
+/// ```
+#[allow(
```
connortsui20, line 56: for some reason expect doesn't work here

vortex-array/src/dtype/serde/serde.rs
```diff
@@ -391,6 +410,132 @@ impl<'de> DeserializeSeed<'de> for StructFieldsSeed<'_> {
...
+            2,
+            UnionVisitor {
+                session: self.session,
+            },
+        )
+    }
+}
+
+impl<'de> DeserializeSeed<'de> for DTypeSerde<'_, UnionVariants> {
+    type Value = UnionVariants;
```
connortsui20, line 465: this is all llm generated, I have no idea if it is correct

robert3005 (approved):

PR #8704, opened 2026-07-09
ST_Intersects pushdown: vortex.geo.intersects kernel, DuckDB lowering, spatial function overrides

HarukiMoriarty:
## Rationale for this change

`ST_Intersects` over native geometry columns currently evaluates inside DuckDB, exporting every row as `GEOMETRY` just to be tested. This adds a native `vortex.geo.intersects` kernel and pushes the filter into the scan: on SpatialBench SF1 (6M points vs a literal polygon), 406ms (DuckDB filter) / 36.6ms (SPATIAL_JOIN) -> 12.4ms pushed. Part of the native geospatial lane, following the `ST_Distance`/`ST_DWithin` pushdown.

- `vortex-geo`: `GeoIntersects` kernel.
- `vortex-duckdb`: lower `st_intersects(native column, GEOMETRY literal)` like `st_distance`.
- `vortex-duckdb`: shadow spatial's `ST_Intersects` with a non-fallible copy — DuckDB won't push can-throw filters through the projection every view-registered table has. The `ST_DWithin` override generalizes into a table-driven registry (`spatial_overrides.hpp`) shared by registration and the join-condition restore pass.

Note: drop `GeoDistance`'s columnar point fast paths. `geo` is row-oriented, so kernels materialize rows regardless; hand-rolled columnar paths belong in a future columnar geometry compute library (pushed Q1: 5.4ms -> 12.6ms, still far ahead of unpushed).


vortex-duckdb/cpp/scalar_fn_pushdown.cpp
```diff
@@ -125,15 +127,24 @@ ScalarFnReplace::ScalarFnReplace(Analyses &analyses, const Projections &projecti
 
 namespace {
 
-// See RestoreStDWithin: rebinding through spatial's own entry lets spatial build its own bind
-// data, so nothing here depends on its internals.
-class StDWithinRestore final : public LogicalOperatorVisitor {
+// Whether `expr` is a call bound through one of the Vortex user-catalog overrides (see
+// `SpatialOverrides` in spatial_overrides.hpp).
+bool IsSpatialOverrideCall(const BoundFunctionExpression &expr) {
```
myrrc: Let's move this code to a separate translation unit spatial_overrides.hpp/cpp as it's not really related to a scalar function pushdown, but overriding some of functions.
Moreover, when 2.0 comes it would be harder for me to remove scalar functions pushdown (upstreamed).

vortex-duckdb/cpp/expr.cpp
```diff
@@ -37,7 +38,59 @@ extern "C" const char *duckdb_vx_sfunc_name(duckdb_vx_sfunc ffi_func) {
...
+    {"st_intersects", 2, ClearErrorMode},
+};
+
+std::span<const SpatialOverride> SpatialOverrides() {
+    return SPATIAL_OVERRIDES;
+}
+
+/// Shadow spatial's `fn_override.name` in the user catalog with tweaked copies of its overloads;
+/// a no-op when the function is not registered (`spatial` not loaded).
+static void RegisterSpatialOverride(ClientContext &context, const SpatialOverride &fn_override) {
```
myrrc: Let's also move this code to a spatial translation unit

myrrc (changes requested):

myrrc:
Let's add a sqllogictest which checks our queries don't break if we override the functions

vortex-duckdb/cpp/include/expr.h
```diff
@@ -14,10 +14,10 @@ typedef struct duckdb_vx_agg_func_ *duckdb_vx_agg_func;
...
 const char *duckdb_vx_sfunc_name(duckdb_vx_sfunc ffi_func);
 
-/// Shadow `ST_DWithin` with a copy that keeps the radius as the third argument, so
-/// radius filters can push into Vortex scans. See `RestoreStDWithin` in
-/// scalar_fn_pushdown.hpp for the override/restore example.
-duckdb_state duckdb_vx_register_st_dwithin_override(duckdb_database ffi_db);
+/// Shadow the spatial functions that block filter pushdown with pushable copies; see
+/// `SPATIAL_OVERRIDES` in spatial_overrides.cpp for the list. Call after `LOAD spatial`;
+/// does nothing when spatial is not loaded.
+duckdb_state duckdb_vx_register_spatial_overrides(duckdb_database ffi_db);
```
myrrc: Let's move this into spatial_overrides.h

vortex-duckdb/cpp/spatial_overrides.cpp
```diff
@@ -0,0 +1,167 @@
...
+    }
+    return DuckDBSuccess;
+}
+
+namespace {
+
+// Rebinds overridden spatial calls in join conditions back to spatial's original, so spatial's
+// own machinery handles joins. Filters are left untouched: they keep the override and push to
+// Vortex scans.
+class SpatialOverrideRestore final : public LogicalOperatorVisitor {
```
myrrc: Let's move the class to .hpp file and implement its methods here for nicer indentation.

myrrc (approved):
LGTM!

vortex-geo/src/scalar_fn/intersects.rs
```diff
@@ -0,0 +1,307 @@
...
+        match child_idx {
+            0 => ChildName::from("a"),
+            1 => ChildName::from("b"),
+            _ => unreachable!("intersects has exactly two children"),
+        }
+    }
+
+    fn return_dtype(&self, _: &Self::Options, _: &[DType]) -> VortexResult<DType> {
+        Ok(DType::Bool(Nullability::NonNullable))
+    }
```
connortsui20: From what I saw in the other geo code, it seems like we disallow null geo arrays completely, so this method needs to validate that. This needs to be updated in `GeoDistance` to, I believe.

vortex-geo/src/scalar_fn/intersects.rs
```diff
@@ -0,0 +1,307 @@
...
+                    Scalar::bool(ga.intersects(&gb), Nullability::NonNullable),
+                    a.len(),
+                )
+                .into_array())
+            }
+            (Some(query), None) => intersects_constant(&b, query.scalar(), ctx),
+            (None, Some(query)) => intersects_constant(&a, query.scalar(), ctx),
+            (None, None) => {
+                vortex_ensure!(
+                    a.len() == b.len(),
```
connortsui20: Nit: `vortex_ensure_eq!`

vortex-geo/src/scalar_fn/distance.rs
```diff
@@ -81,7 +76,14 @@ impl ScalarFnVTable for GeoDistance {
...
 
-    fn return_dtype(&self, _: &Self::Options, _: &[DType]) -> VortexResult<DType> {
+    fn return_dtype(&self, _: &Self::Options, dtypes: &[DType]) -> VortexResult<DType> {
+        // Geometry arrays are never nullable; reject nullable operands up front.
+        for dtype in dtypes {
+            vortex_ensure!(
+                !dtype.is_nullable(),
+                "geo distance: nullable operand {dtype} is unsupported"
+            );
+        }
```
connortsui20: I think this validation needs to be more robust. So it should check that the input dtypes are actually things we can intersect (geo types) and that they are non-nullable. You can look at the scalar fns in `vortex-tensor` to see how we do it there

vortex-geo/src/scalar_fn/intersects.rs
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    fn return_dtype(&self, _: &Self::Options, dtypes: &[DType]) -> VortexResult<DType> {
+        // Geometry arrays are never nullable; reject nullable operands up front.
+        for dtype in dtypes {
+            vortex_ensure!(
+                !dtype.is_nullable(),
+                "geo intersects: nullable operand {dtype} is unsupported"
+            );
+        }
```
connortsui20: same thing here. And if the logic is similar, then you can extract that into a shared helper

connortsui20 (changes requested):

connortsui20 (approved):

PR #8705, opened 2026-07-09
Add `StandardSum`: a sum aggregate where zero valid values yields null

### LINKED ISSUE #9084: sum(all null) is 0 in vortex-duckdb
vortex-duckdb uses own accumulator for sum since https://github.com/vortex-data/vortex/pull/8645.
sum(all null) is 0 in vortex, and that's intended behaviour.

However, there's #8705 which adds the aggregate with behavior we want

mhk197:
Adds a new aggregate function `StandardSum` whose sum over zero valid values (empty or all-null input) is null rather than zero, matching `DuckDB`, `arrow-rs`, and `DataFusion`. Overflow still yields null. `list_sum` now uses it, which makes its empty/all-null-list second pass unnecessary.

The existing `Sum` is untouched. Since the zoned layout persists `AggregateSpecProto` { id: "vortex.sum", .. } and `Sum`'s partial is wire format, it makes more sense to keep its name for now. `Sum` keeps being what statistics store and merge (identity-on-empty is what stat merging requires now). 

`StandardSum`'s partial aggregate is`{sum, overflowed}` rather than just `sum`. A null `sum` means empty list OR overflow , and `overflow` poisons merges. This allows `merge(null, sum) = sum`.

Implementation notes:
- The accumulation kernels are shared with Sum (re-exported, unchanged)
- finalize produces a lazy array rather than materializing eagerly
- The grouped kernel emits `{sum, overflow}` rows directly

This also fixes `StructGetItemRule`. `get_item` on a nullable struct has a nullable result dtype even when every row is valid, but the rule returned the raw non-nullable field, causing a reduced-dtype mismatch (surfaced by the lazy finalize over null groups; it's a standalone bug).


gatesn (commented):
I think... with the new Zoned layout, the aggregate partial has become a wire API.

That said, we haven't updated FileStats to use aggregate partials. And I don't believe zone maps use Sum?

A safer option would be to add a separate Sum function with this new behavior and different partial. @robert3005 any thoughts on a preferred direction?

Either way, we should separately make sure we have forward/backward compatibility tests for zoned statistics!

robert3005:
we have SUM in zonemaps

robert3005:
@mhk197 I think simpler would be to add another function. Maybe even rename them such that the new sql semantic sum is called `sum` and the other is called `sum_<I have no idea what a good name is>`

joseph-isaacs:
Why can we not use `Sum` options?

robert3005:
We can but in order to support it you have to change the partial and the partial is being written to the files therefore you have to handle partial being in the old form even if you want to compute the new sum, maybe the semantics are clear but I thought it might be easier this way.

vortex-array/src/arrays/struct_/compute/rules.rs
```diff
@@ -103,8 +103,14 @@ impl ArrayParentReduceRule<Struct> for StructGetItemRule {
 
         match child.validity()? {
             Validity::NonNullable | Validity::AllValid => {
-                // If the struct is non-nullable or all valid, the field's validity is unchanged
-                Ok(Some(field.clone()))
+                // The field's values are unchanged, but `get_item` on a *nullable* struct has a
+                // nullable result dtype even when every struct row is valid, so a non-nullable
+                // field needs a nullability cast to match.
+                if child.as_ref().dtype().is_nullable() && !field.dtype().is_nullable() {
```
robert3005, line 109: I feel like this was a long standing bug

vortex-array/src/aggregate_fn/fns/standard_sum/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+            seen: false,
+            skip_nans: options.skip_nans,
+        })
+    }
+
+    fn combine_partials(&self, partial: &mut Self::Partial, other: Scalar) -> VortexResult<()> {
+        // Partials are `{sum, seen}` structs. A plain scalar is a cached `Stat::Sum` value,
+        // which cannot distinguish an empty sum from a zero sum, so it is treated as seen.
+        let (sum_value, other_seen) = if matches!(other.dtype(), DType::Struct(..)) {
+            if other.is_null() {
```
joseph-isaacs, line 166: How can this happen?

vortex-array/src/aggregate_fn/fns/standard_sum/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+use crate::scalar_fn::EmptyOptions;
+use crate::scalar_fn::fns::fill_null::FillNull;
+use crate::scalar_fn::fns::get_item::GetItem;
+use crate::scalar_fn::fns::mask::Mask;
+use crate::validity::Validity;
+
+/// Return the sum of an array. The result is null when the array has no valid values or the sum
+/// overflows.
+///
+/// See [`StandardSum`] for details.
```
joseph-isaacs, line 49: Can we instead store a overflowed and a nullable partial sum

`{partial: T, overflowed: bool}` that way you can first check overflow of both and return, then use `NULL op x = x` semantics. This seems much simpler.

I guess overflowed == seen /\ is_null(value). But this seems more complicated to me

joseph-isaacs (changes requested):
Lets make sure we get this right

joseph-isaacs:
I really want to avoid 100s of very similar partial. I am not sure we want to create a new one here vs using a good default for the other.

vortex-array/src/aggregate_fn/fns/standard_sum/mod.rs
mhk197: e.g. partial sum calculated over [-1,1], [0, 0, 0, 0] vs [null, null], []

vortex-array/src/aggregate_fn/fns/standard_sum/mod.rs
mhk197: Yep makes sense, this has been done.

mhk197: Wrt comment [below](https://github.com/vortex-data/vortex/pull/8705#issuecomment-4964811922), we're changing the semantics of partial `sum` field (NULL op x = x instead of poisoning merges) and partials are already persisted on wire. This change would result in old stored `Sum` stats/zone maps creating a lot of work, since would have to fetch and canonicalize underlying arrays to fetch validity + count non-null bits

gatesn:
The other one is part of the serialized ZonedLayout. You cannot change an aggregate partial

joseph-isaacs:
You can change it a forwards compat way.

vortex-array/src/aggregate_fn/fns/standard_sum/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+pub fn standard_sum(array: &ArrayRef, ctx: &mut ExecutionCtx) -> VortexResult<Scalar> {
+    // Short-circuit using cached array statistics. `Stat::Sum` is zero for an
+    // array with no valid values instead of null, so we need the null count for a nullable array.
```
robert3005, line 57: ah, this was how all of this used to behave before stats aggregate rewrite. Imho we should make this the SUM and the thing that's currently the sum be like SumZero

vortex-array/src/aggregate_fn/fns/standard_sum/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+/// - [DataFusion](https://github.com/apache/datafusion/blob/4153adf2c0f6e317ef476febfdc834208bd46622/datafusion/functions-aggregate/src/sum.rs#L370)
+#[derive(Clone, Debug)]
+pub struct StandardSum;
+
+impl AggregateFnVTable for StandardSum {
+    type Options = NumericalAggregateOpts;
+    type Partial = StandardSumPartial;
+
+    fn id(&self) -> AggregateFnId {
+        static ID: CachedId = CachedId::new("vortex.standard_sum");
```
robert3005, line 104: though the ids have to stay, aaaah annoying

vortex-array/src/aggregate_fn/fns/standard_sum/mod.rs
mhk197: That would be ideal but then we'd have to accept id mismatch (`SumZero` would have to keep `vortex.sum`) for backcompat :(. Which would be preferable?

robert3005: I think Joe is right and we need to somehow get the right behaviour forward. I missed this reviewing Nick's pr originally and now we have to pay the price

myrrc:
@mhk When you get back to this, can you also change it here https://github.com/vortex-data/vortex/blob/0462621dbd5b664cb1b9efd689136728cc8d4362/vortex-duckdb/src/convert/expr.rs#L531?

PR #8706, opened 2026-07-10
`ListLayout` baseline

mhk197:
# List Layout

Decomposes a `list<T>` column into three independently-written sub-columns — `elements`,
`offsets`, and (when nullable) `validity` — each backed by its own layout, under a single
`vortex.list` node.

```
vortex.list                       (ListLayout)
├── elements : <T's layout>       element space   — one entry per list *element*
├── offsets  : primitive u64      row space + 1   — n+1 entries for n lists, monotonic
└── validity : bool               row space       — present iff the list is nullable
```

There are three pieces: the **dispatcher** enables using `ListLayout` for arrays with `DType::List`, the **writer** shreds a list stream into the three child streams, and the **reader** reassembles only the children an expression actually needs.

Note that this a prototype and is not a performance improvement over flattening lists... yet. As such, this PR does not yet enable writing `ListLayout` for benchmarks.

---

## 1. `Recursive Writer Dispatch`

`TableStrategy` inspects the dtype of the stream it's handed and routes:

| dtype    | written by                                             |
| -------- | ------------------------------------------------------ |
| struct   | `StructStrategy`  |
| list     | `ListLayoutStrategy` (when list decomposition is on)   |
| anything | the configured leaf strategy                           |

To enable recursive handling of nested lists (elements are lists) or structs  it hands **a descended copy of itself** as
the child strategy. So a `list<struct<{ a, list<i32> }>>` recurses with no manual wiring — each
level dispatches on its own dtype. 

List decomposition is gated (`use_experimental_list_layout()` / `VORTEX_EXPERIMENTAL_LIST_LAYOUT`,
or an explicit builder call). When off, lists fall through to the leaf strategy unchanged.

---

## 2. `ListLayoutStrategy` 

A *structural* writer: it does not compress or inspect element dtypes; it shreds a list stream into
its three sub-streams and hands each to its own downstream strategy, producing one `ListLayout`.

To do so it must:

1. **Canonicalize** each incoming chunk to `List` parts. A gapped/reordered `ListView` is rebuilt
   into a gapless `List` first (`list_from_list_view`).
2. **Rebase offsets to global `u64`.** Each chunk's local offsets are shifted by `element_base`
   (the running count of elements emitted so far) and widened to `u64`, and the duplicated boundary
   offset is dropped on every chunk after the first — so the concatenation of all chunks is one
   monotonic `[0 … total_elements]` array of length `row_count + 1` that indexes into the single
   concatenated `elements` child.
3. **Fan out** `elements`, `offsets`, and `validity` onto three bounded (capacity-1) channels; each
   child is written concurrently by its own strategy via `spawn_nested`. The producer future is
   joined with the child writers (`try_join`) so a producer error surfaces instead of being hidden
   as an early channel close.

Non-list input is forwarded to a `fallback` strategy unchanged.

Because each child is written through an *independent* strategy, the children's physical shape is
decoupled: routing `elements` through the recursive dispatcher (→ repartition → chunked) makes it a
`ChunkedLayout` chunked in element space, while `offsets`/`validity` can be written with a different
(e.g. row-space) strategy. 

---

## 3. `ListReader` 

To read as little as possible, `projection_evaluation` first **classifies the
expression** (`get_necessary_list_children`) into the minimal set of children it needs, then routes
to a matching read path:

| class (`ListChildrenNeeded`) | example expr                    | reads                 | path                      |
| ---------------------------- | ------------------------------- | --------------------- | ------------------------- |
| `Validity`                   | `is_null(x)` / `is_not_null(x)` | validity only         | `project_validity`        |
| `OffsetsAndValidity`         | `list_length(x)`                | offsets + validity    | `project_offsets_validity`|
| `All`                        | `x`, or anything over elements  | elements + the rest   | `project_all`             |

`project_all` materializes the list, and itself picks between two sub-paths:

  - **Whole-chunk / concurrent** (`project_all_concurrent`) — used for a full-column range, or when
    `elements` is a single flat segment (nothing to skip). Fires the `elements`, `offsets`, and
    `validity` reads concurrently, assembles the list, slices to the requested range, and filters.
    No offsets→elements ordering dependency.
  - **Bounded** (`project_all_elements_bounded`) — used for a strict sub-range over **chunked**
    elements. Reads `offsets[range]`, decodes the first/last offset to bound the elements read to
    `[first … last)`, fetches only the element chunks that range overlaps, and rebases the offsets
    to index into the sliced buffer. Costs one offsets→elements round-trip but avoids reading the
    whole elements column for a partial scan split.

## TODOs:
* Add list-related stats for better pruning.
* Pruning for element zones.
* Correctly handle validity. Right now, validity is written as a flat layout. We also probably do not need a reader.
* Better handling of arbitrarily nested lists.
* Perf improvement.

### DIFF
--- a/vortex-file/src/strategy.rs
+++ b/vortex-file/src/strategy.rs
@@ -158,9 +158,10 @@ pub struct WriteStrategyBuilder {
     allow_encodings: Option<HashSet<ArrayId>>,
     flat_strategy: Option<Arc<dyn LayoutStrategy>>,
     probe_compressor: Option<Arc<dyn CompressorPlugin>>,
-    /// Force list-column decomposition on, overriding the
-    /// [`use_experimental_list_layout`] env-var default of off.
-    list_layout: bool,
+    /// Whether to write list fields using [`ListLayoutStrategy`].
+    ///
+    /// [`ListLayoutStrategy`]: vortex_layout::layouts::list::writer::ListLayoutStrategy
+    use_list_layout: bool,
 }
 
 impl Default for WriteStrategyBuilder {
@@ -174,7 +175,7 @@ impl Default for WriteStrategyBuilder {
             allow_encodings: Some(ALLOWED_ENCODINGS.clone()),
             flat_strategy: None,
             probe_compressor: None,
-            list_layout: false,
+            use_list_layout: use_experimental_list_layout(),
         }
     }
 }
@@ -189,11 +190,11 @@ impl WriteStrategyBuilder {
         self
     }
 
-    /// Enable list-column decomposition, overriding the env-var default of off. When enabled, list
-    /// columns are written as a `ListLayout` (independently compressed/chunked
-    /// elements/offsets/validity) rather than as flat arrays.
+    /// Enable writing list fields with [`ListLayoutStrategy`].
+    ///
+    /// [`ListLayoutStrategy`]: vortex_layout::layouts::list::writer::ListLayoutStrategy
     pub fn with_list_layout(mut self) -> Self {
-        self.list_layout = true;
+        self.use_list_layout = true;
         self
     }
 
@@ -263,9 +264,7 @@ impl WriteStrategyBuilder {
             Arc::new(FlatLayoutStrategy::default())
         };
 
-        // 7. for each chunk create a flat layout. List columns are decomposed above this point by
-        // the `TableStrategy` dispatcher (into independently-compressed elements/offsets/validity
-        // sub-columns), so the per-chunk leaf only ever sees flat, non-list chunks.
+        // 7. for each chunk create a flat layout
         let chunked = ChunkedLayoutStrategy::new(Arc::clone(&flat));
         // 6. buffer chunks so they end up with closer segment ids physically
         let buffered = BufferedStrategy::new(chunked, 2 * ONE_MEG); // 2MB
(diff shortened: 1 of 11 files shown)


vortex-file/src/strategy.rs
```diff
@@ -157,6 +158,10 @@ pub struct WriteStrategyBuilder {
     allow_encodings: Option<HashSet<ArrayId>>,
     flat_strategy: Option<Arc<dyn LayoutStrategy>>,
     probe_compressor: Option<Arc<dyn CompressorPlugin>>,
+    /// Whether to write list fields using [`ListLayoutStrategy`].
+    ///
+    /// [`ListLayoutStrategy`]: vortex_layout::layouts::list::writer::ListLayoutStrategy
+    use_list_layout: bool,
```
robert3005, line 165: Can you add a comment that this should only be enabled for experiments right now, will change and will lead to broken files in the future

a10y:
It is a bit unfortunate that the expressions all need to be in core to participate in push down. I guess that's something we can follow up on later?

vortex-file/src/strategy.rs
```diff
@@ -184,6 +190,14 @@ impl WriteStrategyBuilder {
         self
     }
 
+    /// Enable writing list fields with [`ListLayoutStrategy`].
+    ///
+    /// [`ListLayoutStrategy`]: vortex_layout::layouts::list::writer::ListLayoutStrategy
+    pub fn with_list_layout(mut self) -> Self {
```
robert3005, line 200: Can you add a comment to this method?

robert3005 (commented):
I think we can merge this as long as we add all the warnings that this is experimental

robert3005 (approved):

PR #8709, opened 2026-07-10
Implement LIKE in vortex instead of falling back to arrow

robert3005:
One more operation where we fallback to arrow instead of using our own
implementation. We can leverage view array layouts to speed up certain
comparisons


vortex-array/src/scalar_fn/fns/like/pattern.rs
```diff
@@ -0,0 +1,279 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Compiled SQL LIKE patterns evaluated over UTF-8 byte slices.
+//!
+//! A [`LikePattern`] is compiled once per pattern and then evaluated per value. Patterns that
+//! reduce to plain equality, prefix, suffix, or substring searches use direct byte comparisons
+//! (with `memchr`'s SIMD substring search for the contains case); everything else falls back to
```
AdamGS, line 8: ❤️

AdamGS (approved):
code is a bit dense but I think it all makes sense?

joseph-isaacs:
We should likely have the pattern logic/handling shared with fsst and onpair pushdown

PR #8711, opened 2026-07-10
vortex-file: Spawn tasks via runtime instead of directly via tokio

brancz:
## Rationale for this change

I upgraded to 0.78 and some automation we have that detects when dependencies use raw `tokio::spawn` got triggered. We need to control all spawn points, both for correctness testing and for instrumentation purposes. I realize that these spawns are in test code, but I thought since vortex already has the runtime abstractions, it wouldn't hurt to use them in the tests as well (admittedly our tooling is not crazily sophisticated so it can't distinguish whether it's testing code or not).

Use `TokioRuntime` instead of raw `tokio::spawn` in tests.

n/a


robert3005 (approved):

gatesn (approved):
This is great, would love to know what checks you're using so we can run them ourselves!

PR #8714, opened 2026-07-10
Remove stored `Union` `Nullability`

connortsui20:
## Summary

Tracking issue: #7882

A union has no independent parent validity. A row is null when its selected child is null, so the union's logical nullability is determined by its variants.

Storing the same information on `DType::Union` created two sources of truth and allowed the outer value to disagree with the variants. This change makes inconsistent union DTypes impossible.

The remaining question is how an operation such as `mask` should change a union's variants when it introduces nulls. That remains deferred and does not block the canonical sparse `UnionArray`.

The Arrow implementation notes and deferred operation scope are in [this comment on #7882](https://github.com/vortex-data/vortex/issues/7882#issuecomment-4935971614).

## Changes

- change `DType::Union` to store only `UnionVariants`
- derive union nullability at runtime from the variant DTypes
- name the non-zero-cost scan `derived_nullability`
- leave `with_nullability` unchanged for unions instead of rewriting variant schemas
- restrict union least-supertype coercion to identical variants for now
- remove Union nullability from Serde, protobuf, and FlatBuffers

Supersedes #8713.

### DIFF
--- a/vortex-array/src/dtype/dtype_impl.rs
+++ b/vortex-array/src/dtype/dtype_impl.rs
@@ -65,7 +65,7 @@ impl DType {
             | FixedSizeList(_, _, null)
             | Struct(_, null)
             | Variant(null) => matches!(null, Nullability::Nullable),
-            Union(variants) => variants.nullability().is_nullable(),
+            Union(variants) => variants.derived_nullability().is_nullable(),
             Extension(ext_dtype) => ext_dtype.storage_dtype().is_nullable(),
         }
     }
@@ -498,7 +498,7 @@ impl Display for DType {
                     .map(|(field_null, dt)| format!("{field_null}={dt}"))
                     .join(", "),
             ),
-            Union(uv) => write!(f, "union({uv}){}", uv.nullability()),
+            Union(uv) => write!(f, "union({uv}){}", uv.derived_nullability()),
             Variant(null) => write!(f, "variant{null}"),
             Extension(ext) => write!(f, "{}", ext),
         }
--- a/vortex-array/src/dtype/dtype_impl.rs
+++ b/vortex-array/src/dtype/dtype_impl.rs
@@ -64,8 +64,8 @@ impl DType {
             | List(_, null)
             | FixedSizeList(_, _, null)
             | Struct(_, null)
-            | Union(_, null)
             | Variant(null) => matches!(null, Nullability::Nullable),
+            Union(variants) => variants.nullability().is_nullable(),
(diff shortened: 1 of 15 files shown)


vortex-flatbuffers/flatbuffers/vortex-dtype/dtype.fbs
```diff
@@ -67,13 +67,10 @@ table Variant {
...
-// Explicit ids keep `nullable` at id 0, the id it had before the variant metadata fields were
-// added, so schemas written in between remain readable.
 table Union {
-    names: [string] (id: 1);
-    dtypes: [DType] (id: 2);
-    type_ids: [byte] (id: 3); // length must equal dtypes.len()
-    nullable: bool (id: 0);
+    names: [string];
+    dtypes: [DType];
+    type_ids: [byte]; // length must equal dtypes.len()
```
connortsui20, line 73: should be fine to break this since nobody is using this yet

vortex-array/src/dtype/dtype_impl.rs
```diff
@@ -92,7 +95,7 @@ impl DType {
             List(edt, _) => List(Arc::clone(edt), nullability),
             FixedSizeList(edt, size, _) => FixedSizeList(Arc::clone(edt), *size, nullability),
             Struct(sf, _) => Struct(sf.clone(), nullability),
-            Union(vs, _) => Union(vs.clone(), nullability),
+            Union(vs) => Union(vs.clone()),
```
connortsui20, line 98: So `with_nullability` now does nothing for `Union`. I think this is the correct semantics since it is similar for struct where it will not change child nullabilities?

connortsui20:
Is there a way to override the flatc check here?

Edit: making changes

robert3005 (approved):

.github/workflows/ci.yml
```diff
@@ -636,6 +636,9 @@ jobs:
           git fetch origin develop --depth 1
           git checkout origin/develop -- flatbuffers
       - name: "Verify flatbuffer back-compat"
+        if: >-
+          github.event_name != 'pull_request'
+          || !contains(github.event.pull_request.labels.*.name, 'changelog/break')
```
connortsui20: @robert3005 are you ok with this?

robert3005: we cannot do this

connortsui20: 😢

robert3005:
poke

PR #8715, opened 2026-07-10
Optimize dependencies using cargo-shear

AdamGS:
## Rationale for this change

Inspired by @0ax1, trying to see how many dependencies we can remove/simplify

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,6 @@ rust-version = "1.91.0"
 version = "0.1.0"
 
 [workspace.dependencies]
-aho-corasick = "1.1.3"
 anyhow = "1.0.97"
 arbitrary = "1.3.2"
 arc-swap = "1.9"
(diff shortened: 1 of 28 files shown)


Cargo.toml
```diff
@@ -316,7 +307,6 @@ vortex-onpair = { version = "0.1.0", path = "./encodings/experimental/onpair", d
 vortex-parquet-variant = { version = "0.1.0", path = "./encodings/parquet-variant" }
 vortex-pco = { version = "0.1.0", path = "./encodings/pco", default-features = false }
 vortex-proto = { version = "0.1.0", path = "./vortex-proto", default-features = false }
-vortex-row = { version = "0.1.0", path = "./vortex-row", default-features = false }
```
robert3005, line 319: this should stay here?

vortex-array/src/arrays/bool/vtable/canonical.rs
connortsui20, line 1: 🤣

Cargo.toml
AdamGS: sure

connortsui20 (approved):
seems fine to me, triggered benchmarks just in case features are weird

PR #8717, opened 2026-07-10
Split arrow dependency from vortex-array

robert3005:
Vortex-array doesn't need arrow for any particular functionality. We want to be
arrow compatible and support conversion from one to another but that logic
doesn't need to live with array implementations.

Instead we move all arrow related logic to vortex-arrow crate that users can
depend on which can depend on other crates not just vortex-array

N.B. we still have arrow-buffer dependency in vortex-array since we use their
i256. Having that dependency doesn't seem particularly problematic


robert3005:
~this pr includes #8709 and should only merge after that one~ This is now safe to merge

vortex-arrow/src/session.rs
```diff
@@ -537,22 +538,28 @@ impl ArrowSession {
...
             }
             DataType::LargeList(elem_field) => {
                 let list = array.as_list::<i64>();
                 let elements = self
                     .from_arrow_array(ArrowArrayRef::clone(list.values()), elem_field.as_ref())?;
                 let offsets = list.offsets().clone().into_array();
                 let validity = nulls(list.nulls(), field.is_nullable())?;
-                Ok(crate::arrays::ListArray::try_new(elements, offsets, validity)?.into_array())
+                Ok(
+                    vortex_array::arrays::ListArray::try_new(elements, offsets, validity)?
```
AdamGS: note for @robert3005

AdamGS (approved):

PR #8718, opened 2026-07-10
Assortment of `VortexSource` improvements and fixes

AdamGS:
## Rationale for this change

General improvements of the `FileSource`-based Apache DataFusion integration.

1. DF metrics are in use, so make better use of them both on reads and writes, which is a step towards having the same default observability that `ParquetSource` provides.
2. Fix a subtle issue where a dynamic expression only references partition literals, which we currently don't handle correctly.
3. Drop `batch_size` from the source, by default DataFusion splits the data coming out of `DataScanExec` nodes.
4. Make sure we keep schema and field metadata through all the projections paths.

More metrics will be reported in DF, and batch size needs to be done through the `DataSourceExec` and `FileScanConfig` level.

### DIFF
--- a/vortex-datafusion/src/persistent/opener.rs
+++ b/vortex-datafusion/src/persistent/opener.rs
@@ -111,6 +111,12 @@ pub(crate) struct VortexOpener {
 
 impl FileOpener for VortexOpener {
     fn open(&self, file: PartitionedFile) -> DFResult<FileOpenFuture> {
+        // Calculate the output schema before replacing partition columns with literals so it
+        // retains the table and partition-field metadata declared by the plan.
+        let output_schema = Arc::new(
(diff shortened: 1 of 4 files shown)


vortex-datafusion/src/persistent/opener.rs
```diff
@@ -612,6 +627,53 @@ mod tests {
 
     static SESSION: LazyLock<VortexSession> = LazyLock::new(VortexSession::default);
 
+    #[derive(Debug, Eq, Hash, PartialEq)]
+    struct SnapshotErrorExpr;
```
joseph-isaacs, line 632: add a doc str explaining why you added this

joseph-isaacs (approved):

PR #8719, opened 2026-07-10
Own SQL benchmark suite

### LINKED ISSUE #8329: Benchmarking individual queries performance for query engine integrations
We have two ways of verifying performance and correctness for our query engine integrations (namely duckdb and datafusion):

- SQL logic tests which are fast but verify correctness only, and
- Benchmarks, which are run on-demand in PRs, verify correctness and performance, and wihch are slow.

I propose adding a third verification:
- Only targeting vortex:duckdb and vortex:datafusion, since we don't care about baseline comparison.
- Run for every commit in PR without a separate tag,
- Checking performance of queries like `SELECT len(str) from '1.vortex'` on small-scale data to keep runtime fast.

myrrc:
Testing SQL queries' correctness is done in vortex-sqllogictest, but we also
want to test relative performance of queries not covered in existing benchmarks.
One example is aggregation with filter which should be much more performant
once we use zone maps statistics, but this type of query isn't present is
existing benchmarks.

This PR introduces a new query-per-file dataset and support for running it in CI

### DIFF
--- a/vortex-bench/src/vortex_queries.rs
+++ b/vortex-bench/src/vortex_queries.rs
@@ -1,0 +1,179 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright the Vortex contributors
+
+//! Benchmark that runs queries which aren't part of any existing benchmark
+//! suite but which performance we want to track.
+
+use std::fs;
+use std::path::PathBuf;
+use std::process::Command;
+
+use anyhow::Context;
+use anyhow::Result;
+use anyhow::anyhow;
+use anyhow::bail;
+use glob::Pattern;
+use tracing::debug;
+use url::Url;
+use vortex::error::VortexExpect;
+
+use crate::Benchmark;
+use crate::BenchmarkDataset;
+use crate::Format;
+use crate::TableSpec;
+use crate::bench_dir;
+use crate::resolve_data_url;
+
+// Path to script that creates Parquet data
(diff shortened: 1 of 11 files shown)


vortex-bench/.gitignore
```diff
@@ -1,2 +1,6 @@
-data/
+data/*
```
AdamGS: @connortsui20 had a recent PR that moved the scripts into some other directory, I think ideally we would keep the existing gitignore rule here and just place the new files/scripts/docs in that new `sql` directory

vortex-bench/src/vortex_queries.rs
```diff
@@ -0,0 +1,170 @@
...
+// Path to script that creates Parquet data
+const INIT_SQL: &str = "init.sql";
+
+pub struct VortexBenchmark {
+    data_url: Url,
+    queries_dir: PathBuf,
+    query: Option<PathBuf>,
+}
+
+impl VortexBenchmark {
```
AdamGS, line 36: add some doc either here or at the module level (`//!`) explaining what this is

vortex-bench/.gitignore
connortsui20: fyi https://github.com/vortex-data/vortex/pull/8684

myrrc:
As this is fast we should probably run this on every commit @AdamGS WDYT?

AdamGS (approved):

PR #8720, opened 2026-07-10
ST_Contains/ST_Within pushdown: vortex.geo.contains kernel, DuckDB lowering

HarukiMoriarty:
## Rationale for this change

Adds the third geo predicate kernel: OGC `ST_Contains`, following the `GeoDistance`/`GeoIntersects` pattern from #8704. One kernel serves both SQL spellings, since `ST_Within(a, b)` is `ST_Contains(b, a)`.

- `vortex-geo`: `GeoContains` kernel (`vortex.geo.contains`): OGC containment, boundary excluded, delegating to `geo::Contains`. Containment is asymmetric, so the operand roles are kept in separate helpers and pinned by a test. Nullable operands are rejected like the other kernels.
- `vortex-duckdb`: `st_contains` and `st_within` both lower to the kernel — within is contains with the operands swapped. Both get spatial override rows so their filters push through views.


myrrc (approved):
LGTM

connortsui20 (approved):

PR #8721, opened 2026-07-10
Fix decimal scalar Mul/Div to rescale at the shared scale

mhk197:
## What

`DecimalScalar::checked_binary_numeric` applied the checked operator to the raw stored integers with no scale handling, so Mul results were off by `10^scale` (its own test asserted `0.50 × 0.10 = "5.00"`; mathematically `0.05`) and Div dropped the scale entirely (`10.00 / 0.10 = "1.00"` instead of `100.00`).

This fixes Mul and Div to fixed-point semantics at the shared scale, truncating toward zero:

- **Mul**: the raw product (scale `2s`) is rescaled back to `s` via a new truncating rescale helper. The product is computed in i256, so it can exceed the operand storage width as long as the rescaled result fits the precision.
- **Div**: `(a · 10^s) / b` for `s ≥ 0`, `a / (b · 10^{-s})` for negative scales; division by zero returns `None`.
- **Add/Sub** are unchanged (raw ops are exact at a shared scale).

The raw `DecimalValue::checked_mul/checked_div` remain scale-oblivious and are now documented as such, pointing at the new `checked_mul_fixed_point`/`checked_div_fixed_point`.

The only production callers of the scalar op use Add (sum aggregate, stats), so this is a safe bug fix. It is also a prerequisite for native decimal array arithmetic (follow-up PR), whose conformance tests compare element-wise against this scalar implementation.

Note: the truncating rescale helper takes `i16` scales because intermediate scales (e.g. `2s` for the raw product) can exceed the `i8` range.

## Testing

- Fixed the Mul/Div test expectations and added truncation (`0.15 × 0.15 → 0.02`, toward-zero for negatives), negative-scale, wider-than-storage, and precision-overflow cases.
- `cargo nextest run -p vortex-array`, `cargo clippy --all-targets --all-features`, `cargo +nightly fmt --all`.

### DIFF
--- a/vortex-array/src/scalar/typed_view/decimal/dvalue.rs
+++ b/vortex-array/src/scalar/typed_view/decimal/dvalue.rs
@@ -154,6 +154,27 @@ impl DecimalValue {
         }
     }
 
+    /// Rescales a stored decimal value from one scale to another, truncating toward zero when
+    /// reducing the scale discards fractional digits.
+    ///
+    /// Scales are `i16` because intermediate scales (e.g. the raw product of a fixed-point
+    /// multiplication, whose scale is twice the operand scale) can exceed the `i8` range.
+    /// Returns `None` if increasing the scale overflows [`i256`].
+    pub(crate) fn rescale_i256_trunc(value: i256, from_scale: i16, to_scale: i16) -> Option<i256> {
+        if from_scale == to_scale || value == i256::ZERO {
+            return Some(value);
+        }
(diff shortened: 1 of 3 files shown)


mhk197:
Superseded by the per-operator restructure starting at #8724. The scalar Mul and Div fixes from this PR will land with the decimal Mul and Div kernel PRs respectively, since each fix is only observable through the op it corrects.

